Issue #1087, opened 2023-09-01
Investigate migrating custom codec from BinaryDocValuesFormat to KnnVectorsFormat


jmazanec15:
Currently, we integrate our native libraries with OpenSearch through Lucene's DocValuesFormat. At the time, Lucene did not have the [KnnVectorsFormat](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/codecs/KnnVectorsFormat.java) format (which was released in 9.0).

Now that it exists, I am wondering if we should move to use KnnVectorsFormat. KnnVectorsFormat has a KnnVectorsWriter and KnnVectorsReader. Migrating to KnnVectorsFormat would allow us to:
1. reduce branching logic between the different engines we support
2. avoid duplicate effort. For instance, we might be able to more easily support features such as incremental index building: https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/codecs/KnnFieldVectorsWriter.java#L37. 
3. support future features easier

In general, it would make the native library integrations more inline with the Lucene architecture which would have long-term benefits around maintainability and extendability.

All that being said, we need to do a deep dive into what switching means in terms of backwards compatibility and also scope out how much work needs to be done.

### Tracking list of benefits of moving to KnnVectorsFormat
1. #1467 
2. #1506 
3. #1050 
4. #675
5. Incremental index building for vector fields
6. Less blast radius for codec - right now, for any k-NN index, we will handle all binary doc values in our consumer. If there is an issue, potentially impacting non-knn fields. For this, we could just implement KnnVectorsFormat and only be responsible for the field's knn vectors
7. Open up possibility to use non FSDirectory backed indices (i.e. remote storage)
8. Use byte values with native libraries without reimplementation of int8 docvalues

heemin32:
## Overview

When we add KNN support using native engine library(Faiss, NMSLib), there was no KnnVectorsFormat available in the Lucene library. So we used BinaryDocValuesFormat to store the vector data and added knn index using the native engine library while indexing doc value. For searching, we implemented our own query and utilized native engine index for ANN and doc value data for exact search. 

Now as KnnVectorsFormat is available in the Lucene library, we are exploring an opportunity to move from DocValuesFormat to KnnVectorsFormat to see if there are any benefit of doing it.

[...]

PR #1573, opened 2024-03-21
Support script score when doc value is disabled

bugmakerrrrrr:
### Description
Today, script score is not supported if using lucene engine and doc value is disabled, because k-NN plugin uses doc value to execute exact query with script score. But actually, we can use `ByteVectorValues/FloatVectorValues` to execute this type of query.

In addition to , I think that we can totally disable doc value when using lucene engine, vector values can replace the function of doc value. Further, we can use `FlatVectorsFormat` to store vector values and implement our own native index functionality, so that we can avoid extra de/serialization and unify the script doc value interface. Any thoughts?

Btw, This PR has not added relevant tests yet. 
 
### Check List
  - [x] All tests pass

### DIFF
--- a/src/main/java/org/opensearch/knn/index/KNNVectorDVLeafFieldData.java
+++ b/src/main/java/org/opensearch/knn/index/KNNVectorDVLeafFieldData.java
@@ -5,9 +5,11 @@
 
 package org.opensearch.knn.index;
 
-import org.apache.lucene.index.BinaryDocValues;
 import org.apache.lucene.index.DocValues;
+import org.apache.lucene.index.FieldInfo;
 import org.apache.lucene.index.LeafReader;
+import org.apache.lucene.index.VectorEncoding;
+import org.apache.lucene.search.DocIdSetIterator;
 import org.opensearch.index.fielddata.LeafFieldData;
 import org.opensearch.index.fielddata.ScriptDocValues;
 import org.opensearch.index.fielddata.SortedBinaryDocValues;
@@ -39,10 +41,21 @@ public class KNNVectorDVLeafFieldData implements LeafFieldData {
     @Override
     public ScriptDocValues<float[]> getScriptValues() {
         try {
-            BinaryDocValues values = DocValues.getBinary(reader, fieldName);
-            return new KNNVectorScriptDocValues(values, fieldName, vectorDataType);
+            DocIdSetIterator values = null;
+            FieldInfo fieldInfo = reader.getFieldInfos().fieldInfo(fieldName);
+            System.out.println(fieldInfo);
+            if (fieldInfo.hasVectorValues()) {
+                values = fieldInfo.getVectorEncoding() == VectorEncoding.FLOAT32
+                    ? reader.getFloatVectorValues(fieldName)
+                    : reader.getByteVectorValues(fieldName);
+                System.out.println("use vector values");
+            } else {
+                values = DocValues.getBinary(reader, fieldName);
+                System.out.println("use binary values");
+            }
+            return KNNVectorScriptDocValues.create(values, fieldName, vectorDataType);
         } catch (IOException e) {
-            throw new IllegalStateException("Cannot load doc values for knn vector field: " + fieldName, e);
+            throw new IllegalStateException("Cannot load values for knn vector field: " + fieldName, e);
         }
     }
 
--- a/src/main/java/org/opensearch/knn/index/KNNVectorDVLeafFieldData.java
+++ b/src/main/java/org/opensearch/knn/index/KNNVectorDVLeafFieldData.java
@@ -41,17 +41,18 @@ public class KNNVectorDVLeafFieldData implements LeafFieldData {
     @Override
     public ScriptDocValues<float[]> getScriptValues() {
         try {
-            DocIdSetIterator values = null;
             FieldInfo fieldInfo = reader.getFieldInfos().fieldInfo(fieldName);
-            System.out.println(fieldInfo);
+            if (fieldInfo == null) {
+                return KNNVectorScriptDocValues.emptyValues(fieldName, vectorDataType);
+            }
+
+            DocIdSetIterator values = null;
             if (fieldInfo.hasVectorValues()) {
                 values = fieldInfo.getVectorEncoding() == VectorEncoding.FLOAT32
                     ? reader.getFloatVectorValues(fieldName)
                     : reader.getByteVectorValues(fieldName);
-                System.out.println("use vector values");
             } else {
                 values = DocValues.getBinary(reader, fieldName);
-                System.out.println("use binary values");
             }
             return KNNVectorScriptDocValues.create(values, fieldName, vectorDataType);
         } catch (IOException e) {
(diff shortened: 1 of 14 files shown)


bugmakerrrrrr:
Currently, if we disable the doc value and use script score, the following exception will be thrown:
```
{
    "error": {
        "root_cause": [
            {
                "type": "illegal_state_exception",
                "reason": "unexpected docvalues type NONE for field 'item_vector' (expected=BINARY). Re-index with correct docvalues type."
            }
        ],
        "type": "search_phase_execution_exception",
        "reason": "all shards failed",
        "phase": "query",
        "grouped": true,
        "failed_shards": [
            {
                "shard": 0,
                "index": "products-shirts",
                "node": "M-rVnb_0SiqtyJw2-04XNw",
                "reason": {
                    "type": "illegal_state_exception",
                    "reason": "unexpected docvalues type NONE for field 'item_vector' (expected=BINARY). Re-index with correct docvalues type."
                }
            }
        ],
        "caused_by": {
            "type": "illegal_state_exception",
            "reason": "unexpected docvalues type NONE for field 'item_vector' (expected=BINARY). Re-index with correct docvalues type.",
            "caused_by": {
                "type": "illegal_state_exception",
                "reason": "unexpected docvalues type NONE for field 'item_vector' (expected=BINARY). Re-index with correct docvalues type."
            }
        }
    },
    "status": 500
}
```

navneet1v:
> Further, we can use FlatVectorsFormat to store vector values and implement our own native index functionality, so that we can avoid extra de/serialization and unify the script doc value interface.

This is a great idea. There is a github issue created for the same. There some challenges around the same. Ref: https://github.com/opensearch-project/k-NN/issues/1087#issuecomment-1734568700

navneet1v:
@bugmakerrrrrr can you fix the gh actions.

navneet1v:
> I think that we can totally disable doc value when using lucene engine

Yes we should do this if possible. @jmazanec15 what your thought?

src/main/java/org/opensearch/knn/index/KNNVectorDVLeafFieldData.java
```diff
@@ -39,10 +41,21 @@ public long ramBytesUsed() {
     @Override
     public ScriptDocValues<float[]> getScriptValues() {
         try {
-            BinaryDocValues values = DocValues.getBinary(reader, fieldName);
-            return new KNNVectorScriptDocValues(values, fieldName, vectorDataType);
+            DocIdSetIterator values = null;
+            FieldInfo fieldInfo = reader.getFieldInfos().fieldInfo(fieldName);
+            System.out.println(fieldInfo);
```
navneet1v, line 46: remove this.

navneet1v:
I did a very high level look on the PR, the logic seems to be pretty neat. But please remove all sysout statements and add docs on public functions and new interfaces which are created. 

Along with fix the gh actions.

jmazanec15:
> Further, we can use FlatVectorsFormat to store vector values and implement our own native index functionality, so that we can avoid extra de/serialization and unify the script doc value interface.

Right, #1087 calls this out. I think it makes sense to implement our own KnnVectorsFormat that uses the correct FlatVectorsFormat.

jmazanec15:
> Yes we should do this if possible. @jmazanec15 what your thought?

Yes, I agree this can be disabled - no sense in storing both.

jmazanec15:
This looks pretty good to me. I think we will need to add a few test cases.

bugmakerrrrrr:
> > I think that we can totally disable doc value when using lucene engine
> 
> On this I would think we should do some BWC to see if for an index we suddenly stop sending doc values will it cause any issue.
> 
> @bugmakerrrrrr can we add a BWC tests for the same.

@navneet1v Of course. But in this PR, I will focus on supporting script score, and open another PR to disable doc value.

src/test/java/org/opensearch/knn/plugin/script/KNNScriptScoringIT.java
```diff
@@ -35,12 +38,29 @@
 import static org.hamcrest.Matchers.containsString;
 
 public class KNNScriptScoringIT extends KNNRestTestCase {
+    private void randomCreateKNNIndex() throws IOException {
```
navneet1v, line 41: move this function after all the public functions.

Also, why we are creating KNNIndex randomly?

navneet1v (commented):
Overall code looks good to me.

src/main/java/org/opensearch/knn/index/KNNVectorDVLeafFieldData.java
```diff
@@ -39,10 +41,22 @@ public long ramBytesUsed() {
...
+            FieldInfo fieldInfo = reader.getFieldInfos().fieldInfo(fieldName);
+            if (fieldInfo == null) {
+                return KNNVectorScriptDocValues.emptyValues(fieldName, vectorDataType);
+            }
+
+            DocIdSetIterator values = null;
+            if (fieldInfo.hasVectorValues()) {
+                values = fieldInfo.getVectorEncoding() == VectorEncoding.FLOAT32
+                    ? reader.getFloatVectorValues(fieldName)
+                    : reader.getByteVectorValues(fieldName);
```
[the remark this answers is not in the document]
bugmakerrrrrr: make sense

src/test/java/org/opensearch/knn/plugin/script/KNNScriptScoringIT.java
bugmakerrrrrr: I want to test different vector values by creating native knn index or Lucene knn index with/without doc values randomly

jmazanec15 (approved):
LGTM

src/test/java/org/opensearch/knn/plugin/script/KNNScriptScoringIT.java
navneet1v: this makes the test little unpredictable. Can we have separate tests for native engine and Lucene. In that sense we will know if test fails for which engine it is failing.

bugmakerrrrrr: I tend to keep this for two reasons: 1. the output log can help us reproduce the error with a specific random seed if the test fails; 2. using randomization to test different code paths is a common approach in OpenSearch, which can make test code shorter and clearer.

navneet1v: Randomization is good in data generation, but as part of UTs and ITs its better to test the concrete engine here. Because when index creation config is random we can never say that the script score is tested for nmslib, lucene and faiss engine. What if in all the CIs and build process nmslib engine never got hit and there is bug that get out in the release. Hence I wouldn't recommend creating index mappings at random here. 


Also, in this case I would not even generate random data for script score for ITs because we should validate the top results coming out of script score query to ensure that new DocValues are working perfectly.

navneet1v (approved):
Looks good to me. Thanks for fixing the tests.

navneet1v:
@jmazanec15 need approval from your side too on the PR.

jmazanec15 (approved):

navneet1v:
@bugmakerrrrrr can you manually raise the backport PR for this change.

bugmakerrrrrr:
> @bugmakerrrrrr can you manually raise the backport PR for this change.

@navneet1v Sure

PR #1575, opened 2024-03-22
Add Benchmarking Config for Faiss SQfp16

naveentatikonda:
### Description
Add configuration to run benchmarking test for Faiss SQfp16.

 
### Check List

### DIFF
--- a/benchmarks/perf-tool/release-configs/faiss-hnsw-sqfp16/index.json
+++ b/benchmarks/perf-tool/release-configs/faiss-hnsw-sqfp16/index.json
@@ -1,0 +1,33 @@
+{
+  "settings": {
+    "index": {
+      "knn": true,
+      "number_of_shards": 24,
+      "number_of_replicas": 1,
+      "knn.algo_param.ef_search": 100
+    }
+  },
+  "mappings": {
+    "properties": {
+      "target_field": {
+        "type": "knn_vector",
+        "dimension": 128,
+        "method": {
+          "name": "hnsw",
+          "space_type": "l2",
+          "engine": "faiss",
+          "parameters": {
+            "encoder": {
+              "name": "sq",
+              "parameters": {
+                "type": "fp16"
+              }
+            },
+            "ef_construction": 256,
+            "m": 16
+          }
+        }
+      }
+    }
+  }
+}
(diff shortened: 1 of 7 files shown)


ryanbogan (approved):

navneet1v:
@naveentatikonda Lets not add the SQ benchmarks in perf tool and instead put this in OSB. We have migrated our benchmarks to OSB repo. Lets use that repo to add new benchmarks. 

cc: @vamshin

naveentatikonda:
> @naveentatikonda Lets not add the SQ benchmarks in perf tool and instead put this in OSB. We have migrated our benchmarks to OSB repo. Lets use that repo to add new benchmarks.
> 
> cc: @vamshin

Sure I will add it in [OSB-workloads](https://github.com/opensearch-project/opensearch-benchmark-workloads/tree/main/vectorsearch). But, shall we also add it here just in case if we want to run tests using perf-tool. Also, if I'm not wrong as of now the OSB supports vector search with no train procedure only.

navneet1v:
> > @naveentatikonda Lets not add the SQ benchmarks in perf tool and instead put this in OSB. We have migrated our benchmarks to OSB repo. Lets use that repo to add new benchmarks.
> > cc: @vamshin
> 
> Sure I will add it in [OSB-workloads](https://github.com/opensearch-project/opensearch-benchmark-workloads/tree/main/vectorsearch). But, shall we also add it here just in case if we want to run tests using perf-tool. Also, if I'm not wrong as of now the OSB supports vector search with no train procedure only.

Yes it supports only no-train procedures. But SQFp-16 is no train workload right. So we should be able to support it with OSB.

navneet1v:
>  But, shall we also add it here just in case if we want to run tests using perf-tool. 

I would not recommend.

naveentatikonda:
> > > @naveentatikonda Lets not add the SQ benchmarks in perf tool and instead put this in OSB. We have migrated our benchmarks to OSB repo. Lets use that repo to add new benchmarks.
> > > cc: @vamshin
> > 
> > 
> > Sure I will add it in [OSB-workloads](https://github.com/opensearch-project/opensearch-benchmark-workloads/tree/main/vectorsearch). But, shall we also add it here just in case if we want to run tests using perf-tool. Also, if I'm not wrong as of now the OSB supports vector search with no train procedure only.
> 
> Yes it supports only no-train procedures. But SQFp-16 is no train workload right. So we should be able to support it with OSB.

We don't need training for hnsw. But, for ivf we need training. But, that makes sense that we want to migrate to OSB-workloads. Will add it there. Thanks!

Issue #1506, opened 2024-03-04
[FEATURE] Merge Optimization: Streaming vectors from Java Layer to JNI layer to avoid OOM/Circuit Breaker in native engines


navneet1v:
## Problem Statement

While testing on with below details, I was seeing CB exceptions.
1. dataset with 1536D and 5M records 
2. 1 data node with 128GB ram and 16vCPUs
3. OpenSearch heap size: 32gb, version 2.11.0
4. Shards: 1
5. Force merge segments to 1

### Logs:
If CB is enabled

```
024-02-24 03:36:09,149 | INFO: Start optimize (opensearch.py:169) (372349)                                                                                              
2024-02-24 03:42:42,914 | WARNING: VectorDB optimize error: TransportError(429, 'circuit_breaking_exception', '[parent] Data too large, data for [<http_request>] would b
e [32224869544/30gb], which is larger than the limit of [31111669350/28.9gb], real usage: [32224869544/30gb], new bytes reserved: [0/0b], usages [request=0/0b, fielddata
=0/0b, in_flight_requests=378/378b]') (task_runner.py:258) (371437)                                                                                                      
2024-02-24 03:42:42,992 | WARNING: Failed to run performance case, reason = TransportError(429, 'circuit_breaking_exception', '[parent] Data too large, data for [<http_r
equest>] would be [32224869544/30gb], which is larger than the limit of [31111669350/28.9gb], real usage: [32224869544/30gb], new bytes reserved: [0/0b], usages [request
=0/0b, fielddata=0/0b, in_flight_requests=378/378b]') (task_runner.py:186) (371437)  
```

If CB is not enabled
```
 WARN ][o.o.m.j.JvmGcMonitorService] [integTest-0] [gc][1433] overhead, spent [1.5s] collecting in the last [1.5s]                                                                                        [49/1803]
» WARN ][o.o.m.j.JvmGcMonitorService] [integTest-0] [gc][1434] overhead, spent [1s] collecting in the last [1s]
» ERROR][o.o.b.OpenSearchUncaughtExceptionHandler] [integTest-0] fatal error in thread [opensearch[integTest-0][refresh][T#1]], exiting
»  java.lang.OutOfMemoryError: Java heap space
»       at org.opensearch.common.util.concurrent.ThreadContext.propagateTransients(ThreadContext.java:574) ~[opensearch-2.11.1-SNAPSHOT.jar:2.11.1-SNAPSHOT]
[...]

PR #1583, opened 2024-03-28
Add micro-benchmark module in k-NN plugin for benchmark streaming vectors to JNI layer functionality.

navneet1v:
### Description
Add micro-benchmark module in k-NN plugin for benchmark streaming vectors to JNI layer functionality.

This is first of many PR that will be raised for Issue: https://github.com/opensearch-project/k-NN/issues/1506.

The benchmark results that will be used in #1506 will come from this change. Hence adding it upfront.

Along with this change I am adding a new interface in JNILayer: transferVectorsV2, which is actually getting benchmarked. This interface is experimental and will changed once I start the feature development. Right now I have marked the interface as experimental. 

#### What is micro-benchmarks? and how it is useful?
Micro-benchmark is a new module which is getting added in k-NN plugin that will host some client code to run the performance benchmarks. This module will help us do the benchmarking of a specific internal piece of code/functions/feature of k-NN plugin without setting up the whole cluster. Internally the module uses [JMH](http://openjdk.java.net/projects/code-tools/jmh/) as the benchmarking platform. 

### Results
`./gradlew -p micro-benchmarks run --args ' -gc true ' -Djvm.heap.size=16g`

| Benchmark                                 | 	(dimension) | 	(vectorsPerTransfer) | 	Mode | 	Cnt | 	Score | 	Units |
|-------------------------------------------|--------------|----------------------|-------|------|--------|------|
| TransferVectorsBenchmarks.transferVectors | 	128         | 	100000              | 	ss   | 	3   | 	1.904 | s/op |
| TransferVectorsBenchmarks.transferVectors | 128          | 500000               | ss    | 3    | 1.589  | s/op |
| TransferVectorsBenchmarks.transferVectors | 128          | 1000000              | ss    | 3    | 1.443  | s/op |
| TransferVectorsBenchmarks.transferVectors | 256          | 100000               | ss    | 3    | 3.986  | s/op |
| TransferVectorsBenchmarks.transferVectors | 256          | 500000               | ss    | 3    | 3.234  | s/op |
| TransferVectorsBenchmarks.transferVectors | 256          | 1000000              | ss    | 3    | 2.863  | s/op |
| TransferVectorsBenchmarks.transferVectors | 384          | 100000               | ss    | 3    | 6.306  | s/op |
| TransferVectorsBenchmarks.transferVectors | 384          | 500000               | ss    | 3    | 4.953  | s/op |
| TransferVectorsBenchmarks.transferVectors | 384          | 1000000              | ss    | 3    | 4.478  | s/op |
| TransferVectorsBenchmarks.transferVectors | 512          | 100000               | ss    | 3    | 7.74   | s/op |
| TransferVectorsBenchmarks.transferVectors | 512          | 500000               | ss    | 3    | 6.063  | s/op |
| TransferVectorsBenchmarks.transferVectors | 512          | 1000000              | ss    | 3    | 5.338  | s/op |

### Issues Resolved
NA
 
### Check List
  - [X] All tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 * Make the HitQueue size more appropriate for exact search [#1549](https://github.com/opensearch-project/k-NN/pull/1549)
 ### Bug Fixes
 ### Infrastructure
+* Add micro-benchmark module in k-NN plugin for benchmark streaming vectors to JNI layer functionality. [#1583](https://github.com/opensearch-project/k-NN/pull/1583)
 ### Documentation
 ### Maintenance
 ### Refactoring
--- a/micro-benchmarks/src/main/java/org/opensearch/knn/TransferVectorsBenchmarks.java
+++ b/micro-benchmarks/src/main/java/org/opensearch/knn/TransferVectorsBenchmarks.java
@@ -1,0 +1,87 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ * Modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+package org.opensearch.knn;
+
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.BenchmarkMode;
+import org.openjdk.jmh.annotations.Fork;
+import org.openjdk.jmh.annotations.Level;
+import org.openjdk.jmh.annotations.Measurement;
+import org.openjdk.jmh.annotations.Mode;
+import org.openjdk.jmh.annotations.OutputTimeUnit;
+import org.openjdk.jmh.annotations.Param;
+import org.openjdk.jmh.annotations.Scope;
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.annotations.State;
+import org.openjdk.jmh.annotations.Warmup;
+import org.opensearch.knn.jni.JNIService;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Random;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * The class provides runs some benchmarks and provide the performance data around how much time it will take to
+ * transfer vectors from java to jni layer for different configuration.
+ */
+@Warmup(iterations = 1, timeUnit = TimeUnit.SECONDS, time = 300)
+@Measurement(iterations = 1, timeUnit = TimeUnit.SECONDS, time = 300)
+@Fork(3)
+@BenchmarkMode(Mode.SingleShotTime)
+@OutputTimeUnit(TimeUnit.SECONDS)
+@State(Scope.Benchmark)
+public class TransferVectorsBenchmarks {
+    private static final Random random = new Random(1212121212);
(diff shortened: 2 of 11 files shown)


src/main/java/org/opensearch/knn/jni/FaissService.java
```diff
@@ -181,6 +181,21 @@ public static native KNNQueryResult[] queryIndexWithFilter(
...
     public static native long transferVectors(long vectorsPointer, float[][] trainingData);
 
+    /**
+     * Transfer vectors from Java to native layer. This is the version 2 of transfer vector functionality. The
+     * difference between this and the version 1 is, this version puts vectors at the end rather than in front.
+     * Keeping this name as V2 for now, will come up with better name going forward.
+     * <p>
+     * TODO: Rename the function
+     * <br>
+     * TODO: Make this function native function and use a common cpp file to host these functions.
```
jmazanec15, line 191: What does this mean?

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Features
 ### Enhancements
 * Make the HitQueue size more appropriate for exact search [#1549](https://github.com/opensearch-project/k-NN/pull/1549)
+* Add micro-benchmark module in k-NN plugin for benchmark streaming vectors to JNI layer functionality. [#1583](https://github.com/opensearch-project/k-NN/pull/1583)
```
jmazanec15, line 19: Is this infrastructure since it doesnt change prod code?

micro-benchmarks/src/main/java/org/opensearch/knn/TransferVectorsBenchmarks.java
```diff
@@ -0,0 +1,87 @@
...
+
+    @Param({ "128", "256", "384", "512" })
+    private int dimension;
+
+    @Param({ "100000", "500000", "1000000" })
+    private int vectorsPerTransfer;
+
+    private List<float[]> vectorList;
+
+    @Setup(Level.Invocation)
```
jmazanec15, line 55: Do we want this to be at the invocation level or the Trial level?  (ref: https://github.com/openjdk/jmh/blob/master/jmh-core/src/main/java/org/openjdk/jmh/annotations/Level.java)

Only thing with invocation level was I ran this yesterday and saw most of the cycles are consumed setting up the dataset (-prof stack option)

src/main/java/org/opensearch/knn/jni/FaissService.java
navneet1v: I added some todos on this function as this is not ready for production use. 

One todo is to rename and another is, as this function will be used for both nmslib and faiss. Having a function transfervectorv2 in faiss service doesn’t make sense. I want to create a common cpp class and common java class to access that function. Hence added this todo. Will resolve this in upcoming prs

micro-benchmarks/src/main/java/org/opensearch/knn/TransferVectorsBenchmarks.java
navneet1v: For me as I was using SingleShot benchmarks then there will always be 1 invocation of the benchmarks. But making it at iteration level will still be better. I don't want to put it at Trail level because each benchmark is running with different dimension and number of vectors to be transferred.

CHANGELOG.md
navneet1v: ahh. thanks will change this.

jmazanec15 (approved):
LGTM thanks!

ryanbogan (approved):

PR #1588, opened 2024-04-03
Integrating storeVectors interfaces with createIndex and createIndexTemplate functions.

### LINKED ISSUE #1506: [FEATURE] Merge Optimization: Streaming vectors from Java Layer to JNI layer to avoid OOM/Circuit Breaker in native engines

navneet1v:
### Description
Integrating storeVectors interfaces with createIndex and createIndexTemplate functions.

**This PR includes:**
1. I have changed the interface of createIndex and createIndexTemplate api to start using vectorAddress and dim, instead of vectors array.
2. As of now all the vectors are still getting passed to native memory at once. In the upcoming PR I will make the change to send the vectors in chunks, after updating the RFC.
3. I have remove the freeVector function from the code and started using JNICommons function. 
4. The transfer vector function is still there and will be removed in upcoming PRs after doing more testing.

**Critical decision**
As we have started sending vectorsAddress to Native layer as compared to vectordata, one of the critical code part here is who should free up the memory allocated to the vectors. To make sure that vector memory is freed once the index is created even in case there is exception to avoid possible memory leaks I have added try and finally block in addKNNBinaryField function. I don't want native code to free up this memory as 
1. The vector address generated by Java code and it becomes the responsibility of Java code to free up that memory.
2. Going forward if this createIndex interface starts to get exposed to external libraries that lib may or may not free up the memory. Hence it makes logical sense for Java aka addKNNBinaryField function to free up that memory.

### JNI tests
```
[==========] Running 22 tests from 20 test suites.
[----------] Global test environment set-up.
[----------] 1 test from FaissCreateIndexTest
[ RUN      ] FaissCreateIndexTest.BasicAssertions
[       OK ] FaissCreateIndexTest.BasicAssertions (5 ms)
[----------] 1 test from FaissCreateIndexTest (5 ms total)

[----------] 1 test from FaissCreateIndexFromTemplateTest
[ RUN      ] FaissCreateIndexFromTemplateTest.BasicAssertions
[       OK ] FaissCreateIndexFromTemplateTest.BasicAssertions (2 ms)
[----------] 1 test from FaissCreateIndexFromTemplateTest (2 ms total)

[----------] 3 tests from FaissLoadIndexTest
[ RUN      ] FaissLoadIndexTest.BasicAssertions
[       OK ] FaissLoadIndexTest.BasicAssertions (2 ms)
[ RUN      ] FaissLoadIndexTest.HNSWPQDisableSdcTable
WARNING clustering 256 points to 16 centroids: please provide at least 624 training points
[       OK ] FaissLoadIndexTest.HNSWPQDisableSdcTable (357 ms)
[ RUN      ] FaissLoadIndexTest.IVFPQDisablePrecomputeTable
WARNING clustering 256 points to 16 centroids: please provide at least 624 training points
[       OK ] FaissLoadIndexTest.IVFPQDisablePrecomputeTable (360 ms)
[----------] 3 tests from FaissLoadIndexTest (720 ms total)

[----------] 1 test from FaissQueryIndexTest
[ RUN      ] FaissQueryIndexTest.BasicAssertions
[       OK ] FaissQueryIndexTest.BasicAssertions (4 ms)
[----------] 1 test from FaissQueryIndexTest (4 ms total)

[----------] 1 test from FaissQueryIndexWithFilterTest1435
[ RUN      ] FaissQueryIndexWithFilterTest1435.BasicAssertions
[       OK ] FaissQueryIndexWithFilterTest1435.BasicAssertions (9 ms)
[----------] 1 test from FaissQueryIndexWithFilterTest1435 (9 ms total)

[----------] 1 test from FaissQueryIndexWithParentFilterTest
[ RUN      ] FaissQueryIndexWithParentFilterTest.BasicAssertions
[       OK ] FaissQueryIndexWithParentFilterTest.BasicAssertions (5 ms)
[----------] 1 test from FaissQueryIndexWithParentFilterTest (5 ms total)

[----------] 1 test from FaissFreeTest
[ RUN      ] FaissFreeTest.BasicAssertions
[       OK ] FaissFreeTest.BasicAssertions (0 ms)
[----------] 1 test from FaissFreeTest (0 ms total)

[----------] 1 test from FaissInitLibraryTest
[ RUN      ] FaissInitLibraryTest.BasicAssertions
[       OK ] FaissInitLibraryTest.BasicAssertions (0 ms)
[----------] 1 test from FaissInitLibraryTest (0 ms total)

[----------] 1 test from FaissTrainIndexTest
[ RUN      ] FaissTrainIndexTest.BasicAssertions
[       OK ] FaissTrainIndexTest.BasicAssertions (0 ms)
[----------] 1 test from FaissTrainIndexTest (0 ms total)

[----------] 1 test from FaissCreateHnswSQfp16IndexTest
[ RUN      ] FaissCreateHnswSQfp16IndexTest.BasicAssertions
[       OK ] FaissCreateHnswSQfp16IndexTest.BasicAssertions (5 ms)
[----------] 1 test from FaissCreateHnswSQfp16IndexTest (5 ms total)

[----------] 1 test from FaissIsSharedIndexStateRequired
[ RUN      ] FaissIsSharedIndexStateRequired.BasicAssertions
[       OK ] FaissIsSharedIndexStateRequired.BasicAssertions (0 ms)
[----------] 1 test from FaissIsSharedIndexStateRequired (0 ms total)

[----------] 1 test from FaissInitAndSetSharedIndexState
[ RUN      ] FaissInitAndSetSharedIndexState.BasicAssertions
WARNING clustering 256 points to 16 centroids: please provide at least 624 training points
[       OK ] FaissInitAndSetSharedIndexState.BasicAssertions (339 ms)
[----------] 1 test from FaissInitAndSetSharedIndexState (339 ms total)

[----------] 1 test from IDGrouperBitMapTest
[ RUN      ] IDGrouperBitMapTest.BasicAssertions
[       OK ] IDGrouperBitMapTest.BasicAssertions (0 ms)
[----------] 1 test from IDGrouperBitMapTest (0 ms total)

[----------] 1 test from NmslibIndexWrapperSearchTest
[ RUN      ] NmslibIndexWrapperSearchTest.BasicAssertions
[       OK ] NmslibIndexWrapperSearchTest.BasicAssertions (0 ms)
[----------] 1 test from NmslibIndexWrapperSearchTest (0 ms total)

[----------] 1 test from NmslibCreateIndexTest
[ RUN      ] NmslibCreateIndexTest.BasicAssertions
[       OK ] NmslibCreateIndexTest.BasicAssertions (2 ms)
[----------] 1 test from NmslibCreateIndexTest (2 ms total)

[----------] 1 test from NmslibLoadIndexTest
[ RUN      ] NmslibLoadIndexTest.BasicAssertions
[       OK ] NmslibLoadIndexTest.BasicAssertions (1 ms)
[----------] 1 test from NmslibLoadIndexTest (1 ms total)

[----------] 1 test from NmslibQueryIndexTest
[ RUN      ] NmslibQueryIndexTest.BasicAssertions
[       OK ] NmslibQueryIndexTest.BasicAssertions (2 ms)
[----------] 1 test from NmslibQueryIndexTest (2 ms total)

[----------] 1 test from NmslibFreeTest
[ RUN      ] NmslibFreeTest.BasicAssertions
[       OK ] NmslibFreeTest.BasicAssertions (0 ms)
[----------] 1 test from NmslibFreeTest (0 ms total)

[----------] 1 test from NmslibInitLibraryTest
[ RUN      ] NmslibInitLibraryTest.BasicAssertions
[       OK ] NmslibInitLibraryTest.BasicAssertions (0 ms)
[----------] 1 test from NmslibInitLibraryTest (0 ms total)

[----------] 1 test from CommonsTests
[ RUN      ] CommonsTests.BasicAssertions
[       OK ] CommonsTests.BasicAssertions (0 ms)
[----------] 1 test from CommonsTests (0 ms total)

[----------] Global test environment tear-down
[==========] 22 tests from 20 test suites ran. (1101 ms total)
[  PASSED  ] 22 tests.

```

 
### Issues Resolved
#1506 
 
### Check List
  - [X] All tests pass

### DIFF
--- a/jni/src/faiss_wrapper.cpp
+++ b/jni/src/faiss_wrapper.cpp
@@ -81,15 +81,19 @@ bool isIndexIVFPQL2(faiss::Index * index);
 // IndexIDMap which has member that will point to underlying index that stores the data
 faiss::IndexIVFPQ * extractIVFPQIndex(faiss::Index * index);
 
-void knn_jni::faiss_wrapper::CreateIndex(knn_jni::JNIUtilInterface * jniUtil, JNIEnv * env, jintArray idsJ,
-                                         jobjectArray vectorsJ, jstring indexPathJ, jobject parametersJ) {
+void knn_jni::faiss_wrapper::CreateIndex(knn_jni::JNIUtilInterface * jniUtil, JNIEnv * env, jintArray idsJ, jlong vectorsAddressJ, jint dimJ,
+                                         jstring indexPathJ, jobject parametersJ) {
 
     if (idsJ == nullptr) {
         throw std::runtime_error("IDs cannot be null");
     }
 
-    if (vectorsJ == nullptr) {
-        throw std::runtime_error("Vectors cannot be null");
+    if (vectorsAddressJ <= 0) {
+        throw std::runtime_error("VectorsAddress cannot be less than 0");
+    }
+
+    if(dimJ <= 0) {
+        throw std::runtime_error("Vectors dimensions cannot be less than or equal to 0");
     }
 
     if (indexPathJ == nullptr) {
@@ -109,15 +113,19 @@ void knn_jni::faiss_wrapper::CreateIndex(knn_jni::JNIUtilInterface * jniUtil, JN
     std::string spaceTypeCpp(jniUtil->ConvertJavaObjectToCppString(env, spaceTypeJ));
     faiss::MetricType metric = TranslateSpaceToMetric(spaceTypeCpp);
 
-    // Read data set
-    int numVectors = jniUtil->GetJavaObjectArrayLength(env, vectorsJ);
+    // Read vectors from memory address
+    auto *inputVectors = reinterpret_cast<std::vector<float>*>(vectorsAddressJ);
+    int dim = (int)dimJ;
+    // The number of vectors can be int here because a lucene segment number of total docs never crosses INT_MAX value
+    int numVectors = (int) (inputVectors->size() / (uint64_t) dim);
+    if(numVectors == 0) {
+        throw std::runtime_error("Number of vectors cannot be 0");
+    }
+
     int numIds = jniUtil->GetJavaIntArrayLength(env, idsJ);
     if (numIds != numVectors) {
         throw std::runtime_error("Number of IDs does not match number of vectors");
     }
-
-    int dim = jniUtil->GetInnerDimensionOf2dJavaFloatArray(env, vectorsJ);
-    auto dataset = jniUtil->Convert2dJavaObjectArrayToCppFloatVector(env, vectorsJ, dim);
 
     // Create faiss index
     jobject indexDescriptionJ = knn_jni::GetJObjectFromMapOrThrow(parametersCpp, knn_jni::INDEX_DESCRIPTION);
@@ -148,7 +156,7 @@ void knn_jni::faiss_wrapper::CreateIndex(knn_jni::JNIUtilInterface * jniUtil, JN
 
     auto idVector = jniUtil->ConvertJavaIntArrayToCppIntVector(env, idsJ);
     faiss::IndexIDMap idMap = faiss::IndexIDMap(indexWriter.get());
-    idMap.add_with_ids(numVectors, dataset.data(), idVector.data());
+    idMap.add_with_ids(numVectors, inputVectors->data(), idVector.data());
 
     // Write the index to disk
     std::string indexPathCpp(jniUtil->ConvertJavaStringToCppString(env, indexPathJ));
@@ -156,14 +164,18 @@ void knn_jni::faiss_wrapper::CreateIndex(knn_jni::JNIUtilInterface * jniUtil, JN
 }
 
 void knn_jni::faiss_wrapper::CreateIndexFromTemplate(knn_jni::JNIUtilInterface * jniUtil, JNIEnv * env, jintArray idsJ,
-                                                     jobjectArray vectorsJ, jstring indexPathJ,
+                                                     jlong vectorsAddressJ, jint dimJ, jstring indexPathJ,
                                                      jbyteArray templateIndexJ, jobject parametersJ) {
     if (idsJ == nullptr) {
         throw std::runtime_error("IDs cannot be null");
     }
 
-    if (vectorsJ == nullptr) {
-        throw std::runtime_error("Vectors cannot be null");
+    if (vectorsAddressJ <= 0) {
+        throw std::runtime_error("VectorsAddress cannot be less than 0");
+    }
+
+    if(dimJ <= 0) {
+        throw std::runtime_error("Vectors dimensions cannot be less than or equal to 0");
     }
 
     if (indexPathJ == nullptr) {
@@ -183,14 +195,14 @@ void knn_jni::faiss_wrapper::CreateIndexFromTemplate(knn_jni::JNIUtilInterface *
     jniUtil->DeleteLocalRef(env, parametersJ);
 
     // Read data set
-    int numVectors = jniUtil->GetJavaObjectArrayLength(env, vectorsJ);
+    // Read vectors from memory address
+    auto *inputVectors = reinterpret_cast<std::vector<float>*>(vectorsAddressJ);
+    int dim = (int)dimJ;
+    int numVectors = (int) (inputVectors->size() / (uint64_t) dim);
     int numIds = jniUtil->GetJavaIntArrayLength(env, idsJ);
     if (numIds != numVectors) {
         throw std::runtime_error("Number of IDs does not match number of vectors");
     }
-
-    int dim = jniUtil->GetInnerDimensionOf2dJavaFloatArray(env, vectorsJ);
-    auto dataset = jniUtil->Convert2dJavaObjectArrayToCppFloatVector(env, vectorsJ, dim);
 
     // Get vector of bytes from jbytearray
     int indexBytesCount = jniUtil->GetJavaBytesArrayLength(env, templateIndexJ);
@@ -208,7 +220,7 @@ void knn_jni::faiss_wrapper::CreateIndexFromTemplate(knn_jni::JNIUtilInterface *
 
     auto idVector = jniUtil->ConvertJavaIntArrayToCppIntVector(env, idsJ);
     faiss::IndexIDMap idMap =  faiss::IndexIDMap(indexWriter.get());
-    idMap.add_with_ids(numVectors, dataset.data(), idVector.data());
+    idMap.add_with_ids(numVectors, inputVectors->data(), idVector.data());
 
     // Write the index to disk
     std::string indexPathCpp(jniUtil->ConvertJavaStringToCppString(env, indexPathJ));
--- a/src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80DocValuesConsumer.java
+++ b/src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80DocValuesConsumer.java
@@ -16,6 +16,7 @@ import org.opensearch.core.xcontent.MediaTypeRegistry;
 import org.opensearch.core.xcontent.DeprecationHandler;
 import org.opensearch.core.xcontent.NamedXContentRegistry;
 import org.opensearch.knn.index.KNNSettings;
+import org.opensearch.knn.jni.JNICommons;
 import org.opensearch.knn.jni.JNIService;
 import org.opensearch.knn.index.SpaceType;
 import org.opensearch.knn.index.codec.util.KNNCodecUtil;
@@ -110,61 +111,67 @@ class KNN80DocValuesConsumer extends DocValuesConsumer implements Closeable {
         throws IOException {
         // Get values to be indexed
         BinaryDocValues values = valuesProducer.getBinary(field);
-        KNNCodecUtil.Pair pair = KNNCodecUtil.getFloats(values);
-        if (pair.vectors.length == 0 || pair.docs.length == 0) {
-            logger.info("Skipping engine index creation as there are no vectors or docs in the documents");
-            return;
-        }
-        long arraySize = calculateArraySize(pair.vectors, pair.serializationMode);
-        if (isMerge) {
-            KNNGraphValue.MERGE_CURRENT_OPERATIONS.increment();
-            KNNGraphValue.MERGE_CURRENT_DOCS.incrementBy(pair.docs.length);
-            KNNGraphValue.MERGE_CURRENT_SIZE_IN_BYTES.incrementBy(arraySize);
-        }
-        // Increment counter for number of graph index requests
-        KNNCounter.GRAPH_INDEX_REQUESTS.increment();
-        // Create library index either from model or from scratch
-        String engineFileName;
-        String indexPath;
-        NativeIndexCreator indexCreator;
-        final KNNEngine knnEngine = getKNNEngine(field);
-        if (field.attributes().containsKey(MODEL_ID)) {
-
-            String modelId = field.attributes().get(MODEL_ID);
-            Model model = ModelCache.getInstance().get(modelId);
-
-            engineFileName = buildEngineFileName(state.segmentInfo.name, knnEngine.getVersion(), field.name, knnEngine.getExtension());
-            indexPath = Paths.get(((FSDirectory) (FilterDirectory.unwrap(state.directory))).getDirectory().toString(), engineFileName)
-                .toString();
-
-            if (model.getModelBlob() == null) {
-                throw new RuntimeException("There is no trained model with id \"" + modelId + "\"");
+        KNNCodecUtil.Pair pair = null;
+        try {
+            pair = KNNCodecUtil.getFloats(values);
+            if (pair.getVectorAddress() == 0 || pair.docs.length == 0) {
+                logger.info("Skipping engine index creation as there are no vectors or docs in the segment");
+                return;
+            }
+            long arraySize = calculateArraySize(pair.docs.length, pair.getDimension(), pair.serializationMode);
+            if (isMerge) {
+                KNNGraphValue.MERGE_CURRENT_OPERATIONS.increment();
+                KNNGraphValue.MERGE_CURRENT_DOCS.incrementBy(pair.docs.length);
+                KNNGraphValue.MERGE_CURRENT_SIZE_IN_BYTES.incrementBy(arraySize);
+            }
+            // Increment counter for number of graph index requests
+            KNNCounter.GRAPH_INDEX_REQUESTS.increment();
+            final KNNEngine knnEngine = getKNNEngine(field);
+            final String engineFileName = buildEngineFileName(
+                state.segmentInfo.name,
+                knnEngine.getVersion(),
+                field.name,
+                knnEngine.getExtension()
+            );
+            final String indexPath = Paths.get(
+                ((FSDirectory) (FilterDirectory.unwrap(state.directory))).getDirectory().toString(),
+                engineFileName
+            ).toString();
+            KNNCodecUtil.Pair finalPair = pair;
+            NativeIndexCreator indexCreator;
+            // Create library index either from model or from scratch
+            if (field.attributes().containsKey(MODEL_ID)) {
+                String modelId = field.attributes().get(MODEL_ID);
+                Model model = ModelCache.getInstance().get(modelId);
+                if (model.getModelBlob() == null) {
+                    throw new RuntimeException(String.format("There is no trained model with id \"%s\"", modelId));
+                }
+                indexCreator = () -> createKNNIndexFromTemplate(model.getModelBlob(), finalPair, knnEngine, indexPath);
+            } else {
+                indexCreator = () -> createKNNIndexFromScratch(field, finalPair, knnEngine, indexPath);
             }
 
-            indexCreator = () -> createKNNIndexFromTemplate(model.getModelBlob(), pair, knnEngine, indexPath);
-        } else {
+            if (isMerge) {
+                recordMergeStats(pair.docs.length, arraySize);
+            }
 
-            engineFileName = buildEngineFileName(state.segmentInfo.name, knnEngine.getVersion(), field.name, knnEngine.getExtension());
-            indexPath = Paths.get(((FSDirectory) (FilterDirectory.unwrap(state.directory))).getDirectory().toString(), engineFileName)
-                .toString();
+            if (isRefresh) {
+                recordRefreshStats();
+            }
 
-            indexCreator = () -> createKNNIndexFromScratch(field, pair, knnEngine, indexPath);
+            // This is a bit of a hack. We have to create an output here and then immediately close it to ensure that
+            // engineFileName is added to the tracked files by Lucene's TrackingDirectoryWrapper. Otherwise, the file will
+            // not be marked as added to the directory.
+            state.directory.createOutput(engineFileName, state.context).close();
+            indexCreator.createIndex();
+            writeFooter(indexPath, engineFileName);
+        } finally {
+            // Freeing up the Native memory where vectors was stored. We added a try block here to ensure that even
+            // in case of exceptions we are freeing up the space to avoid memory leaks.
+            if (pair != null) {
+                JNICommons.freeVectorData(pair.getVectorAddress());
+            }
         }
-
-        if (isMerge) {
-            recordMergeStats(pair.docs.length, arraySize);
-        }
-
-        if (isRefresh) {
-            recordRefreshStats();
-        }
-
-        // This is a bit of a hack. We have to create an output here and then immediately close it to ensure that
-        // engineFileName is added to the tracked files by Lucene's TrackingDirectoryWrapper. Otherwise, the file will
-        // not be marked as added to the directory.
-        state.directory.createOutput(engineFileName, state.context).close();
-        indexCreator.createIndex();
-        writeFooter(indexPath, engineFileName);
     }
 
     private void recordMergeStats(int length, long arraySize) {
@@ -186,7 +193,15 @@ class KNN80DocValuesConsumer extends DocValuesConsumer implements Closeable {
             KNNSettings.state().getSettingValue(KNNSettings.KNN_ALGO_PARAM_INDEX_THREAD_QTY)
         );
         AccessController.doPrivileged((PrivilegedAction<Void>) () -> {
-            JNIService.createIndexFromTemplate(pair.docs, pair.vectors, indexPath, model, parameters, knnEngine);
+            JNIService.createIndexFromTemplate(
+                pair.docs,
+                pair.getVectorAddress(),
+                pair.getDimension(),
+                indexPath,
+                model,
+                parameters,
+                knnEngine
+            );
             return null;
         });
     }
@@ -228,7 +243,7 @@ class KNN80DocValuesConsumer extends DocValuesConsumer implements Closeable {
 
         // Pass the path for the nms library to save the file
         AccessController.doPrivileged((PrivilegedAction<Void>) () -> {
-            JNIService.createIndex(pair.docs, pair.vectors, indexPath, parameters, knnEngine);
+            JNIService.createIndex(pair.docs, pair.getVectorAddress(), pair.getDimension(), indexPath, parameters, knnEngine);
             return null;
         });
     }
--- a/src/main/java/org/opensearch/knn/index/codec/util/KNNCodecUtil.java
+++ b/src/main/java/org/opensearch/knn/index/codec/util/KNNCodecUtil.java
@@ -5,9 +5,13 @@
 
 package org.opensearch.knn.index.codec.util;
 
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+import lombok.Setter;
 import org.apache.lucene.index.BinaryDocValues;
 import org.apache.lucene.search.DocIdSetIterator;
 import org.apache.lucene.util.BytesRef;
+import org.opensearch.knn.jni.JNICommons;
 
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
@@ -26,21 +30,24 @@ public class KNNCodecUtil {
     // Java rounds each array size up to multiples of 8 bytes
     public static final int JAVA_ROUNDING_NUMBER = 8;
 
+    @AllArgsConstructor
     public static final class Pair {
-        public Pair(int[] docs, float[][] vectors, SerializationMode serializationMode) {
-            this.docs = docs;
-            this.vectors = vectors;
-            this.serializationMode = serializationMode;
-        }
-
         public int[] docs;
-        public float[][] vectors;
+        @Getter
+        @Setter
+        private long vectorAddress;
+        @Getter
+        @Setter
+        private int dimension;
         public SerializationMode serializationMode;
+
     }
 
     public static KNNCodecUtil.Pair getFloats(BinaryDocValues values) throws IOException {
         ArrayList<float[]> vectorList = new ArrayList<>();
         ArrayList<Integer> docIdList = new ArrayList<>();
+        long vectorAddress = 0;
+        int dimension = 0;
         SerializationMode serializationMode = SerializationMode.COLLECTION_OF_FLOATS;
         for (int doc = values.nextDoc(); doc != DocIdSetIterator.NO_MORE_DOCS; doc = values.nextDoc()) {
             BytesRef bytesref = values.binaryValue();
@@ -48,20 +55,22 @@ public class KNNCodecUtil {
                 serializationMode = KNNVectorSerializerFactory.serializerModeFromStream(byteStream);
                 final KNNVectorSerializer vectorSerializer = KNNVectorSerializerFactory.getSerializerByStreamContent(byteStream);
                 final float[] vector = vectorSerializer.byteToFloatArray(byteStream);
+                dimension = vector.length;
                 vectorList.add(vector);
             }
             docIdList.add(doc);
         }
-        return new KNNCodecUtil.Pair(
-            docIdList.stream().mapToInt(Integer::intValue).toArray(),
-            vectorList.toArray(new float[][] {}),
-            serializationMode
-        );
+        if (vectorList.isEmpty() == false) {
+            vectorAddress = JNICommons.storeVectorData(
+                vectorAddress,
+                vectorList.toArray(new float[][] {}),
+                (long) vectorList.size() * dimension
+            );
+        }
+        return new KNNCodecUtil.Pair(docIdList.stream().mapToInt(Integer::intValue).toArray(), vectorAddress, dimension, serializationMode);
     }
 
-    public static long calculateArraySize(float[][] vectors, SerializationMode serializationMode) {
-        int vectorLength = vectors[0].length;
-        int numVectors = vectors.length;
+    public static long calculateArraySize(int numVectors, int vectorLength, SerializationMode serializationMode) {
         if (serializationMode == SerializationMode.ARRAY) {
             int vectorSize = vectorLength * FLOAT_BYTE_SIZE + JAVA_ARRAY_HEADER_SIZE;
             if (vectorSize % JAVA_ROUNDING_NUMBER != 0) {
--- a/src/main/java/org/opensearch/knn/jni/NmslibService.java
+++ b/src/main/java/org/opensearch/knn/jni/NmslibService.java
@@ -42,11 +42,12 @@ class NmslibService {
      * Create an index for the native library
      *
      * @param ids array of ids mapping to the data passed in
-     * @param data array of float arrays to be indexed
+     * @param vectorsAddress address of native memory where vectors are stored
+     * @param dim dimension of the vector to be indexed
      * @param indexPath path to save index file to
      * @param parameters parameters to build index
      */
-    public static native void createIndex(int[] ids, float[][] data, String indexPath, Map<String, Object> parameters);
+    public static native void createIndex(int[] ids, long vectorsAddress, int dim, String indexPath, Map<String, Object> parameters);
 
     /**
      * Load an index into memory
--- a/jni/include/faiss_wrapper.h
+++ b/jni/include/faiss_wrapper.h
@@ -19,13 +19,13 @@ namespace knn_jni {
     namespace faiss_wrapper {
         // Create an index with ids and vectors. The configuration is defined by values in the Java map, parametersJ.
         // The index is serialized to indexPathJ.
-        void CreateIndex(knn_jni::JNIUtilInterface * jniUtil, JNIEnv * env, jintArray idsJ, jobjectArray vectorsJ,
+        void CreateIndex(knn_jni::JNIUtilInterface * jniUtil, JNIEnv * env, jintArray idsJ, jlong vectorsAddressJ, jint dimJ,
                          jstring indexPathJ, jobject parametersJ);
 
         // Create an index with ids and vectors. Instead of creating a new index, this function creates the index
         // based off of the template index passed in. The index is serialized to indexPathJ.
         void CreateIndexFromTemplate(knn_jni::JNIUtilInterface * jniUtil, JNIEnv * env, jintArray idsJ,
-                                     jobjectArray vectorsJ, jstring indexPathJ, jbyteArray templateIndexJ,
+                                     jlong vectorsAddressJ, jint dimJ, jstring indexPathJ, jbyteArray templateIndexJ,
                                      jobject parametersJ);
 
         // Load an index from indexPathJ into memory.
--- a/jni/include/nmslib_wrapper.h
+++ b/jni/include/nmslib_wrapper.h
@@ -25,7 +25,7 @@ namespace knn_jni {
     namespace nmslib_wrapper {
         // Create an index with ids and vectors. The configuration is defined by values in the Java map, parametersJ.
         // The index is serialized to indexPathJ.
-        void CreateIndex(knn_jni::JNIUtilInterface * jniUtil, JNIEnv * env, jintArray idsJ, jobjectArray vectorsJ,
+        void CreateIndex(knn_jni::JNIUtilInterface * jniUtil, JNIEnv * env, jintArray idsJ, jlong vectorsAddress, jint dim,
                          jstring indexPathJ, jobject parametersJ);
 
         // Load an index from indexPathJ into memory. Use parametersJ to set any query time parameters
--- a/jni/include/org_opensearch_knn_jni_FaissService.h
+++ b/jni/include/org_opensearch_knn_jni_FaissService.h
@@ -21,18 +21,18 @@ extern "C" {
 /*
  * Class:     org_opensearch_knn_jni_FaissService
  * Method:    createIndex
- * Signature: ([I[[FLjava/lang/String;Ljava/util/Map;)V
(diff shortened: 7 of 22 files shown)


jni/src/faiss_wrapper.cpp
```diff
@@ -109,16 +113,20 @@ void knn_jni::faiss_wrapper::CreateIndex(knn_jni::JNIUtilInterface * jniUtil, JN
     std::string spaceTypeCpp(jniUtil->ConvertJavaObjectToCppString(env, spaceTypeJ));
     faiss::MetricType metric = TranslateSpaceToMetric(spaceTypeCpp);
 
-    // Read data set
-    int numVectors = jniUtil->GetJavaObjectArrayLength(env, vectorsJ);
+    // Read vectors from memory address
+    auto *inputVectors = reinterpret_cast<std::vector<float>*>(vectorsAddressJ);
+    int dim = (int)dimJ;
+    // The number of vectors can be int here because a lucene segment number of total docs never crosses INT_MAX value
+    int numVectors = (int) (inputVectors->size() / (long long) dim);
```
jmazanec15, line 120: Instead of long long, should we just use uint64_t?

jni/src/faiss_wrapper.cpp
```diff
@@ -109,16 +113,20 @@ void knn_jni::faiss_wrapper::CreateIndex(knn_jni::JNIUtilInterface * jniUtil, JN
...
+    if(numVectors == 0) {
+        throw std::runtime_error("Number of vectors cannot be 0");
```
jmazanec15, line 122: Should we add this here? I remember discussing awhile ago that there was a possibility this could be called with 0 vectors.

jmazanec15 (commented):
Took a first pass on cpp side. Overall looks good - a few minor comments. Will take another pass on java side next

src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80DocValuesConsumer.java
```diff
@@ -110,61 +111,72 @@ public void addKNNBinaryField(FieldInfo field, DocValuesProducer valuesProducer,
...
-        String engineFileName;
-        String indexPath;
-        NativeIndexCreator indexCreator;
-        final KNNEngine knnEngine = getKNNEngine(field);
-        if (field.attributes().containsKey(MODEL_ID)) {
+        KNNCodecUtil.Pair pair = null;
+        try {
+            pair = KNNCodecUtil.getFloats(values);
+            if (pair.getVectorAddress() == 0 || pair.docs.length == 0) {
+                logger.info("Skipping engine index creation as there are no vectors or docs in the segment");
```
martin-gaievski, line 118: do we need to differentiate between those two conditions for if and log them individually with corresponding message?

jni/src/faiss_wrapper.cpp
navneet1v: Sure.. I will do that.

src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80DocValuesConsumer.java
```diff
@@ -110,61 +111,72 @@ public void addKNNBinaryField(FieldInfo field, DocValuesProducer valuesProducer,
...
+                return;
+            }
+            long arraySize = calculateArraySize(pair.docs.length, pair.getDimension(), pair.serializationMode);
+            if (isMerge) {
+                KNNGraphValue.MERGE_CURRENT_OPERATIONS.increment();
+                KNNGraphValue.MERGE_CURRENT_DOCS.incrementBy(pair.docs.length);
+                KNNGraphValue.MERGE_CURRENT_SIZE_IN_BYTES.incrementBy(arraySize);
```
martin-gaievski, line 125: for my understanding - is it ok if we mutate those three values individually? Seems change in a single value is thread safe but I wonder do we need to join all three changes and make it isolated?

src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80DocValuesConsumer.java
```diff
@@ -110,61 +111,72 @@ public void addKNNBinaryField(FieldInfo field, DocValuesProducer valuesProducer,
...
+
+                String modelId = field.attributes().get(MODEL_ID);
+                Model model = ModelCache.getInstance().get(modelId);
+
+                engineFileName = buildEngineFileName(state.segmentInfo.name, knnEngine.getVersion(), field.name, knnEngine.getExtension());
+                indexPath = Paths.get(((FSDirectory) (FilterDirectory.unwrap(state.directory))).getDirectory().toString(), engineFileName)
+                    .toString();
+
+                if (model.getModelBlob() == null) {
+                    throw new RuntimeException("There is no trained model with id \"" + modelId + "\"");
```
martin-gaievski, line 144: nit: you can use String.format with param

src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80DocValuesConsumer.java
```diff
@@ -110,61 +111,72 @@ public void addKNNBinaryField(FieldInfo field, DocValuesProducer valuesProducer,
...
+                KNNCodecUtil.Pair finalPair = pair;
+                indexCreator = () -> createKNNIndexFromTemplate(model.getModelBlob(), finalPair, knnEngine, indexPath);
+            } else {
 
-            engineFileName = buildEngineFileName(state.segmentInfo.name, knnEngine.getVersion(), field.name, knnEngine.getExtension());
-            indexPath = Paths.get(((FSDirectory) (FilterDirectory.unwrap(state.directory))).getDirectory().toString(), engineFileName)
-                .toString();
+                engineFileName = buildEngineFileName(state.segmentInfo.name, knnEngine.getVersion(), field.name, knnEngine.getExtension());
+                indexPath = Paths.get(((FSDirectory) (FilterDirectory.unwrap(state.directory))).getDirectory().toString(), engineFileName)
+                    .toString();
```
martin-gaievski, line 153: can these two lines be extracted in a method? I think it's a duplication of the same code you've added to the same class on lines 139-140

src/main/java/org/opensearch/knn/index/codec/util/KNNCodecUtil.java
```diff
@@ -26,42 +30,47 @@ public class KNNCodecUtil {
...
+        long vectorAddress = 0;
+        int dimension = 0;
         SerializationMode serializationMode = SerializationMode.COLLECTION_OF_FLOATS;
         for (int doc = values.nextDoc(); doc != DocIdSetIterator.NO_MORE_DOCS; doc = values.nextDoc()) {
             BytesRef bytesref = values.binaryValue();
             try (ByteArrayInputStream byteStream = new ByteArrayInputStream(bytesref.bytes, bytesref.offset, bytesref.length)) {
                 serializationMode = KNNVectorSerializerFactory.serializerModeFromStream(byteStream);
                 final KNNVectorSerializer vectorSerializer = KNNVectorSerializerFactory.getSerializerByStreamContent(byteStream);
                 final float[] vector = vectorSerializer.byteToFloatArray(byteStream);
+                dimension = vector.length;
```
martin-gaievski, line 58: in this case `dimension` will be taken from the last value from the binary doc values. Just making sure it's intentional and all those are always identical for all elements

src/main/java/org/opensearch/knn/jni/NmslibService.java
```diff
@@ -46,7 +46,7 @@ class NmslibService {
      * @param indexPath path to save index file to
```
martin-gaievski, line 46: nit: new method params are not reflected in the java doc

martin-gaievski (commented):
Overall looks good, left minor comments

jni/src/faiss_wrapper.cpp
navneet1v: I think this condition should never happen. We already have checks in the Java Layer, what would be the case when this will happen?

src/main/java/org/opensearch/knn/index/codec/util/KNNCodecUtil.java
navneet1v: Yes that is the understanding. We don't get vectors of different dimensions.

src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80DocValuesConsumer.java
navneet1v: This code was already there. Because I added try and finally block it is coming as a change.

src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80DocValuesConsumer.java
navneet1v: I think we can keep it like this only. not having docIds and still having vectors is not a use case for us.

jni/src/faiss_wrapper.cpp
jmazanec15: Im not sure what case is - I remember talking about it but cant find related discussion

jmazanec15 (approved):
This LGTM. I agree that Java code should free up the vector data.

navneet1v:
@martin-gaievski , @jmazanec15 added a new rev where the JNI tests are fixed and the comments are resolved.

martin-gaievski (approved):

Issue #814, opened 2023-03-21
[FEATURE] Support Radius Search in k-NN


Thijsvandepoll:
**Is your feature request related to a problem?**
I would be interested in the ability to provide a maximum radius filter to find all (approximate) nearest neighbors that lie within a certain distance from the query vector. For our use case we need to aggregate over this result. If I am not mistaken, this is currently not possible.

**What solution would you like?**
Something like:
```
GET my-knn-index-1/_search
{
  "query": {
    "knn": {
      "my_vector2": {
        "vector": [2, 3, 5, 6],
        "radius": 2.0
      }
    }
  }
}
```
This would allow me to aggregate over all results for the given radius.

**What alternatives have you considered?**
We are considering Elasticsearch as an alternative. They are working on such a feature: https://github.com/elastic/elasticsearch/issues/84929

navneet1v:
Assigning it to @vamshin

vamshin:
Please +1 if you are looking for this feature to help prioritize

junqiu-lei:
With leveraging both Lucene and FAISS libraries, we can find out a way to unify the radius search API for both enginess in OpenSearch k-NN.

* FAISS: It supports in radius searches by `range_search` API with distance parameter of radius for both IVF and HNSW methods. 
  * Library availability in k-NN:
    * IVF: Supported
    * HNSW: It will be supported after k-NN Faiss [bump PR](https://github.com/opensearch-project/k-NN/pull/1443) merged

* Lucene: Lucene recently added the feature of supporting similarity-based vector searches. It can find all vectors scoring above a resultSimilarity while traversing the HNSW graph till better-scoring nodes are available, or the best candidate is below a score of traversalSimilarity in the lowest level.
  * Library availability in k-NN:
    * HNSW: This feature will be released at Lucene 9.10([Feature release information](https://github.com/apache/lucene/pull/12679#issuecomment-1907958716), [Lucene release history](https://lucene.apache.org/core/corenews.html)), then OpenSearch will also be required to bump to Lucene 9.10.

Our initial release goal is to support both Lucene and Faiss engine for this feature.

cc: @vamshin @Thijsvandepoll

nknize:
Lucene [9.10 branch](https://github.com/apache/lucene/tree/branch_9_10) was cut yesterday and the release process will begin tomorrow. Given that OpenSearch release is starting today, it's unlikely this would make 2.12 w/o a delay. 

[...]

PR #1589, opened 2024-04-03
Support score type threshold in radial search

junqiu-lei:
### Description
This PR allows user to use `score` as threshold to search vectors, the search result contains all docs which score higher than the `score` for Lucene and Faiss engines ANN search.

### Usage
During index mapping and indexing stages no behavior change. Some query examples:

* Faiss engine with L2 space type
```
curl --location --request GET 'http://localhost:9200/target-index-faiss/_search' \
--header 'Content-Type: application/json' \
--data '{
  "size":12,  
  "query": {
    "knn": {
      "my_vector1": {
        "vector": [7.5, 8.5],
        "score": 0.5
      }
    }
  }
}'

// Query results
{
    "took": 4,
    "timed_out": false,
    "_shards": {
        "total": 1,
        "successful": 1,
        "skipped": 0,
        "failed": 0
    },
    "hits": {
        "total": {
            "value": 1,
            "relation": "eq"
        },
        "max_score": 0.7462686,
        "hits": [
            {
                "_index": "target-index-faiss",
                "_id": "1",
                "_score": 0.7462686,
                "_source": {
                    "my_vector1": [
                        7.0,
                        8.2
                    ],
                    "price": 4.4
                }
            }
        ]
    }
}
```

* Lucene engine with Inner product space type
```
curl --location --request GET 'http://localhost:9200/target-index-lucene/_search' \
--header 'Content-Type: application/json' \
--data '{
  "size":12,  
  "query": {
    "knn": {
      "my_vector1": {
        "vector": [7.0, 8.5],
        "score": 25
      }
    }
  }
}'
 
// Query results
{
    "took": 303,
    "timed_out": false,
    "_shards": {
        "total": 3,
        "successful": 3,
        "skipped": 0,
        "failed": 0
    },
    "hits": {
        "total": {
            "value": 4,
            "relation": "eq"
        },
        "max_score": 69.45,
        "hits": [
            {
                "_index": "target-index-lucene",
                "_id": "13",
                "_score": 69.45,
                "_source": {
                    "my_vector1": [
                        -2.0,
                        9.7
                    ],
                    "price": 18.0
                }
            },
            {
                "_index": "target-index-lucene",
                "_id": "17",
                "_score": 39.3,
                "_source": {
                    "my_vector1": [
                        -5.7,
                        9.2
                    ],
                    "price": 3.1
                }
            },
            {
                "_index": "target-index-lucene",
                "_id": "7",
                "_score": 34.45,
                "_source": {
                    "my_vector1": [
                        -1.9,
                        5.5
                    ],
                    "price": 11.0
                }
            },
            {
                "_index": "target-index-lucene",
                "_id": "3",
                "_score": 27.5,
                "_source": {
                    "my_vector1": [
                        -3.5,
                        6.0
                    ],
                    "price": 19.1
                }
            }
        ]
    }
}
```

### Issues Resolved
Part of https://github.com/opensearch-project/k-NN/issues/814
 
### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/SpaceType.java
+++ b/src/main/java/org/opensearch/knn/index/SpaceType.java
@@ -39,6 +39,9 @@ public enum SpaceType {
 
         @Override
         public float scoreToDistanceTranslation(float score) {
+            if (score == 0) {
+                throw new IllegalArgumentException(String.format(Locale.ROOT, "score cannot be 0 when space type is [%s]", getValue()));
+            }
             return 1 / score - 1;
         }
     },
--- a/src/main/java/org/opensearch/knn/index/SpaceType.java
+++ b/src/main/java/org/opensearch/knn/index/SpaceType.java
@@ -36,6 +36,11 @@ public enum SpaceType {
         public VectorSimilarityFunction getVectorSimilarityFunction() {
             return VectorSimilarityFunction.EUCLIDEAN;
         }
+
+        @Override
+        public float scoreToDistanceTranslation(float score) {
+            return 1 / score - 1;
+        }
     },
     COSINESIMIL("cosinesimil") {
         @Override
@@ -169,5 +174,15 @@ public enum SpaceType {
             }
         }
         throw new IllegalArgumentException("Unable to find space: " + spaceTypeName);
+    }
+
+    /**
+     * Translate a score to a distance for this space type
+     *
+     * @param score score to translate
+     * @return translated distance
+     */
+    public float scoreToDistanceTranslation(float score) {
+        throw new UnsupportedOperationException(String.format("Space [%s] does not have a score to distance translation", getValue()));
     }
 }
--- a/src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
+++ b/src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
@@ -52,6 +52,7 @@ public class KNNQueryBuilder extends AbstractQueryBuilder<KNNQueryBuilder> {
     public static final ParseField FILTER_FIELD = new ParseField("filter");
     public static final ParseField IGNORE_UNMAPPED_FIELD = new ParseField("ignore_unmapped");
     public static final ParseField DISTANCE_FIELD = new ParseField("distance");
+    public static final ParseField SCORE_FIELD = new ParseField("score");
     public static final int K_MAX = 10000;
     /**
      * The name for the knn query
@@ -64,6 +65,7 @@ public class KNNQueryBuilder extends AbstractQueryBuilder<KNNQueryBuilder> {
     private final float[] vector;
     private int k = 0;
     private Float distance = null;
+    private Float score = null;
     private QueryBuilder filter;
     private boolean ignoreUnmapped = false;
 
@@ -92,12 +94,13 @@ public class KNNQueryBuilder extends AbstractQueryBuilder<KNNQueryBuilder> {
      *
      * @param k K nearest neighbours for the given vector
      */
-    public KNNQueryBuilder k(int k) {
+    public KNNQueryBuilder k(Integer k) {
+        if (k == null) {
+            throw new IllegalArgumentException("[" + NAME + "] requires k to be set");
+        }
+        validSingleQueryType(k, distance, score);
         if (k <= 0 || k > K_MAX) {
             throw new IllegalArgumentException("[" + NAME + "] requires 0 < k <= " + K_MAX);
-        }
-        if (distance != null) {
-            throw new IllegalArgumentException("[" + NAME + "] requires either k or distance must be set");
         }
         this.k = k;
         return this;
@@ -112,10 +115,25 @@ public class KNNQueryBuilder extends AbstractQueryBuilder<KNNQueryBuilder> {
         if (distance == null) {
             throw new IllegalArgumentException("[" + NAME + "] requires distance to be set");
         }
-        if (k != 0) {
-            throw new IllegalArgumentException("[" + NAME + "] requires either k or distance must be set");
-        }
+        validSingleQueryType(k, distance, score);
         this.distance = distance;
+        return this;
+    }
+
+    /**
+     * Builder method for score
+     *
+     * @param score the score threshold for the nearest neighbours
+     */
+    public KNNQueryBuilder score(Float score) {
+        if (score == null) {
+            throw new IllegalArgumentException("[" + NAME + "] requires score to be set");
+        }
+        validSingleQueryType(k, distance, score);
+        if (score <= 0) {
+            throw new IllegalArgumentException("[" + NAME + "] requires score greater than 0");
+        }
+        this.score = score;
         return this;
     }
 
@@ -163,6 +181,7 @@ public class KNNQueryBuilder extends AbstractQueryBuilder<KNNQueryBuilder> {
         this.filter = filter;
         this.ignoreUnmapped = false;
         this.distance = null;
+        this.score = null;
     }
 
     public static void initialize(ModelDao modelDao) {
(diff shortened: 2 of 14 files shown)


src/main/java/org/opensearch/knn/index/util/Faiss.java
```diff
@@ -68,6 +69,14 @@ class Faiss extends NativeLibrary {
         rawScore -> SpaceType.INNER_PRODUCT.scoreTranslation(-1 * rawScore)
     );
 
+    // Map that transforms radial search score threshold to faiss required distance
+    private final static Map<SpaceType, Function<Float, Float>> SCORE_TO_DISTANCE_TRANSFORMATIONS = ImmutableMap.<
+        SpaceType,
+        Function<Float, Float>>builder()
+        .put(SpaceType.INNER_PRODUCT, score -> score > 1 ? 1 - score : 1 / score - 1)
+        .put(SpaceType.L2, score -> 1 / score - 1)
```
navneet1v, line 77: for L2 spacetype can we put this translation in SpaceType enum class?

src/main/java/org/opensearch/knn/index/util/Faiss.java
```diff
@@ -68,6 +69,14 @@ class Faiss extends NativeLibrary {
         rawScore -> SpaceType.INNER_PRODUCT.scoreTranslation(-1 * rawScore)
     );
 
+    // Map that transforms radial search score threshold to faiss required distance
+    private final static Map<SpaceType, Function<Float, Float>> SCORE_TO_DISTANCE_TRANSFORMATIONS = ImmutableMap.<
+        SpaceType,
+        Function<Float, Float>>builder()
+        .put(SpaceType.INNER_PRODUCT, score -> score > 1 ? 1 - score : 1 / score - 1)
```
navneet1v, line 76: Lets add some java doc here why this conversion make sense.

navneet1v:
Minor comments overall code looks good to me.

src/main/java/org/opensearch/knn/index/util/Faiss.java
junqiu-lei: Yes updated.

src/main/java/org/opensearch/knn/index/util/Faiss.java
junqiu-lei: Added knn space doc here for more details: https://opensearch.org/docs/latest/search-plugins/knn/approximate-knn/#spaces

jmazanec15:
In general, Im not in favor of "score" naming. Users are going to expect it to match the final score produced by the search results, regardless of the query type. So, they will be confused and cut issues when a result is returned from a more complex query with a hit score that is below the score they passed in the filter. I prefer the name "similarity" - this is what lucene refers to this as - https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/FloatVectorSimilarityQuery.java - it will disambiguate this case from the user experience.

junqiu-lei:
> In general, Im not in favor of "score" naming. Users are going to expect it to match the final score produced by the search results, regardless of the query type. So, they will be confused and cut issues when a result is returned from a more complex query with a hit score that is below the score they passed in the filter. I prefer the name "similarity" - this is what lucene refers to this as - https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/FloatVectorSimilarityQuery.java - it will disambiguate this case from the user experience.

I also thought about `similarity`. But I think user can easily understand the value mean of naming `score` in k-NN. The name `similarity` might be confused especially for feature new users for what value need provide. When say similarity, it could be related to score similarity or distance similarity. In Lucene, it only have score threshold, so I think it's fine to use similarity naming, in OpenSearch k-KNN we support both score threshold and distance threshold, so I think `score` as lower level words might be more suitable for us? 

cc: @vamshin @navneet1v @dylan-tong-aws

src/main/java/org/opensearch/knn/index/SpaceType.java
```diff
@@ -36,6 +36,11 @@ public float scoreTranslation(float rawScore) {
         public VectorSimilarityFunction getVectorSimilarityFunction() {
             return VectorSimilarityFunction.EUCLIDEAN;
         }
+
+        @Override
+        public float scoreToDistanceTranslation(float score) {
+            return 1 / score - 1;
```
jmazanec15, line 42: Can we have a check to confirm score != 0?

src/main/java/org/opensearch/knn/index/SpaceType.java
```diff
@@ -170,4 +175,14 @@ public static SpaceType getSpace(String spaceTypeName) {
...
         throw new IllegalArgumentException("Unable to find space: " + spaceTypeName);
     }
+
+    /**
+     * Translate a score to a distance for this space type
+     *
+     * @param score score to translate
+     * @return translated distance
+     */
+    public float scoreToDistanceTranslation(float score) {
```
jmazanec15, line 185: Why not make abstract and implement for all of the space types?

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
```diff
@@ -499,4 +541,24 @@ protected int doHashCode() {
     public String getWriteableName() {
         return NAME;
     }
+
+    private static void validSingleQueryType(Integer k, Float distance, Float score) {
```
jmazanec15, line 545: nit: valid -> validate

jmazanec15 (commented):
Per sync offline, I think "score" param is okay if we detail explanation in the documentation. Key is that score in this context is scoped to this query context.

src/main/java/org/opensearch/knn/index/SpaceType.java
junqiu-lei: ack

src/main/java/org/opensearch/knn/index/SpaceType.java
junqiu-lei: Because for other space types, currently I don't see the use cases for engines as common, we can change to abstract and add implement whenever needed.

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
junqiu-lei: ack

junqiu-lei: updated

src/main/java/org/opensearch/knn/index/SpaceType.java
junqiu-lei: Added

jmazanec15 (commented):
LGTM - thanks @junqiu-lei

src/main/java/org/opensearch/knn/index/SpaceType.java
```diff
@@ -36,6 +36,14 @@ public float scoreTranslation(float rawScore) {
         public VectorSimilarityFunction getVectorSimilarityFunction() {
             return VectorSimilarityFunction.EUCLIDEAN;
         }
+
+        @Override
+        public float scoreToDistanceTranslation(float score) {
```
navneet1v, line 41: are we adding the score to distance translation for L2 only? not for cosine and dot product?

junqiu-lei: @navneet1v IMO no need for cosine and dot product here. The SpaceType Enum class is used for common translation for different engines:

* Cosine's score to distance function doesn't needed because Lucene's radial search api just need score parameter, in k-NN faiss, it doesn't support Cosine's type.
* Dot product's score to distance function is only need in Faiss engine and in Faiss it has different translation from Lucene, so I just put the function inside Lucene.java where it needed.

navneet1v (approved):

PR #1603, opened 2024-04-09
Check if patch has already been applied

### LINKED ISSUE #1552: Patch files for nmslib and faiss creating enforcing compilation to happen during cmake
### Problem
To enable features in Faiss and Nmslib there are 2 patch files which have been added in k-NN repo. The patches gets applied while running `cmake` or gradle task `cmakeJniLib` ([nmslib](https://github.com/opensearch-project/k-NN/blob/main/jni/CMakeLists.txt#L91-L101), [faiss](https://github.com/opensearch-project/k-NN/blob/main/jni/CMakeLists.txt#L171-L181)). 

Due to this the cmake command assumes that files have changed and it leads recompile the files and dependent files leading to increase in running of all the gradle tasks like integTest, test, run etc.

This really adds significant time during development process as you might have to run the tests/integTest task multiple times. This is not a problem for CIs and actual opensearch build process because it builds the k-NN plugin only once.

### Solution
1. We should come up with a way to identify if patch is already applied on files and then skip adding the patches.
2. Another solution can be we should contribute back these features to respective libraries.

jmazanec15:
### Description
Checks if patch has already been applied before rebuilding
 
### Issues Resolved
#1552 
 
### Check List

### DIFF
--- a/jni/CMakeLists.txt
+++ b/jni/CMakeLists.txt
@@ -93,7 +93,11 @@ if (${CONFIG_NMSLIB} STREQUAL ON OR ${CONFIG_ALL} STREQUAL ON OR ${CONFIG_TEST} 
     # If it exists, apply patches
     if (EXISTS ${PATCH_FILE})
         message(STATUS "Applying custom patches.")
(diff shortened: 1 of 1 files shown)


jmazanec15:
Im going to close this and re-work

PR #1613, opened 2024-04-12
Remove unnecessary toString conversion of vector field and added some minor optimization in KNNCodec

### LINKED ISSUE #1598: [Enhancement][Build-Time] Remove conversion of KNN Field to string if stored Fields not required
### Description
While doing some benchmarks with Opensearch k-NN. I see that there is time going in to convert the kNN field to String even when storedFields is marked as false. On doing deep-dive I found out that, we convert the field to the string and then check if stored field is set as true or false.

Code ref:
https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java#L575

Here `point.toString` will convert the object to string which is expensive operation in high dimensionality use cases. 

Refer the Flame graph for the same for Vector dimension: 1536 ingestion

Screenshot 2024-04-08 at 12 58 58 PM

navneet1v:
### Description
Remove unnecessary toString conversion of vector field and added some minor optimization in KNNCodec
 
### Issues Resolved
#1598
 
### Check List
  - [X] All tests pass

### DIFF
--- a/src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapperUtil.java
+++ b/src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapperUtil.java
@@ -13,6 +13,7 @@ package org.opensearch.knn.index.mapper;
 
 import lombok.AccessLevel;
 import lombok.NoArgsConstructor;
+import org.apache.lucene.document.Field;
 import org.apache.lucene.document.FieldType;
 import org.apache.lucene.document.StoredField;
 import org.apache.lucene.index.DocValuesType;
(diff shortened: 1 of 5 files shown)


ryanbogan (approved):
LGTM!

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapperUtil.java
```diff
@@ -135,14 +136,9 @@ public static FieldType buildDocValuesFieldType(KNNEngine knnEngine) {
...
-    public static void addStoredFieldForVectorField(
-        ParseContext context,
-        FieldType fieldType,
-        String mapperName,
-        String vectorFieldAsString
-    ) {
+    public static void addStoredFieldForVectorField(ParseContext context, FieldType fieldType, String mapperName, Field vectorField) {
         if (fieldType.stored()) {
-            context.doc().add(new StoredField(mapperName, vectorFieldAsString));
+            context.doc().add(new StoredField(mapperName, vectorField.toString()));
```
martin-gaievski, line 141: so the idea is - we only covert to String if field is stored, only always. That's good optimization!

martin-gaievski (approved):
looks good, thank you

junqiu-lei (approved):

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapperUtil.java
navneet1v: Yes that is correct and if you look into the issue you can see the time it takes to convert a field to string.

PR #1630, opened 2024-04-18
Add stored fields for knn_vector type

### LINKED ISSUE #1606: [BUG] cannot retrieve stored kNN vector field
**What is the bug?**
I have an index where I set `store: true` on a field of type `knn_vector`. When I try to retrieve the field, the query succeeds but does not return the vector. 

1. Go to opensearch-dashboards
2. Create this test index
```
PUT knn_stored_field
{
    "aliases": {},
    "mappings": {
      "dynamic": "strict",
      "properties": {
        "knn_field": {
          "type": "knn_vector",
          "store": true,
          "dimension": 3,
          "method": {
            "engine": "faiss",
            "space_type": "innerproduct",
            "name": "hnsw",
            "parameters": {}
          }
        },
        "text_field": {
          "type": "text",
          "store": true
        }
      }
    },
    "settings": {
      "index": {
        "replication": {
          "type": "SEGMENT"
        },
        "refresh_interval": "30s",
        "number_of_shards": "5",
        "knn": "true",
        "number_of_replicas": "1"
      }
    }
  }
```
3. Index a document
```
PUT knn_stored_field/_doc/1
{
  "knn_field": [
    0.0,
    1.1,
    3.2
  ],
  "text_field": "abc123"
}
```
4. Query for all documents, attempting to retrieve the stored fields
```
GET knn_stored_field/_search
{
  "stored_fields": ["knn_field", "text_field"],
  "query": {
    "match_all": {}
  }
}
```
The result of this query is
```
{
  ...
  "hits": [
    {
      "_index": "knn_stored_field",
      "_id": "1",
      "_score": 1,
      "fields": {
        "text_field": [
          "abc123"
        ]
    }
  ]
}
```

I would expect each hit to include the `knn_field` vector as well as the `text_field` data.

 - Opensearch version 2.11, hosted on AWS

jmazanec15:
### Description
Fixes bug where we were not creating stored field type for knn_vector even when the mapping parameter is passed. To do this, we create a stored field from the bytes of the vectors. Then, implemented `valueForDisplay` in order to ensure that the vectors are displayed properly.

Example:
```
Request:
{
  "stored_fields": ["test_vector"],
  "query": {
    "match_all": {}
  }
}

Response:
{
  "took":248,
  "timed_out":false,
  "_shards":{
    "total":1,
    "successful":1,
    "skipped":0,
    "failed":0
  },
  "hits":{
    "total":{
      "value":1,
      "relation":"eq"
    },
    "max_score":1.0,
    "hits":[
      {
        "_index":"test_index",
        "_id":"1",
        "_score":1.0,
        "fields":{"test_vector":[[-128,0,1,127]]}
      }
    ]
  }
}
```

Along with this, clean up the field mapper implementations.

Add relevant uTs and iTs to ensure functionality is working as expected.
 
### Issues Resolved
#1606 
 
### Check List
  - [X] All tests pass

### DIFF
--- a/src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java
+++ b/src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java
@@ -11,6 +11,7 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.function.Supplier;
 import lombok.Getter;
@@ -20,6 +21,7 @@ import org.apache.lucene.index.DocValuesType;
 import org.apache.lucene.index.IndexOptions;
 import org.apache.lucene.search.DocValuesFieldExistsQuery;
 import org.apache.lucene.search.Query;
+import org.apache.lucene.util.BytesRef;
 import org.opensearch.Version;
 import org.opensearch.common.Explicit;
 import org.opensearch.common.Nullable;
@@ -52,16 +54,6 @@ import org.opensearch.knn.indices.ModelDao;
 import org.opensearch.search.aggregations.support.CoreValuesSourceType;
 import org.opensearch.search.lookup.SearchLookup;
 
-import java.io.IOException;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.List;
-import java.util.Locale;
-import java.util.Map;
-import java.util.Objects;
-import java.util.Optional;
-import java.util.function.Supplier;
-
 import static org.opensearch.knn.common.KNNConstants.DEFAULT_VECTOR_DATA_TYPE_FIELD;
 import static org.opensearch.knn.common.KNNConstants.ENCODER_SQ;
 import static org.opensearch.knn.common.KNNConstants.FAISS_SQ_CLIP;
@@ -74,19 +66,17 @@ import static org.opensearch.knn.common.KNNValidationUtil.validateByteVectorValu
 import static org.opensearch.knn.common.KNNValidationUtil.validateFloatVectorValue;
 import static org.opensearch.knn.common.KNNValidationUtil.validateVectorDimension;
 import static org.opensearch.knn.index.KNNSettings.KNN_INDEX;
-import static org.opensearch.knn.index.mapper.KNNVectorFieldMapperUtil.addStoredFieldForVectorField;
+import static org.opensearch.knn.index.mapper.KNNVectorFieldMapperUtil.createStoredFieldForByteVector;
+import static org.opensearch.knn.index.mapper.KNNVectorFieldMapperUtil.createStoredFieldForFloatVector;
 import static org.opensearch.knn.index.mapper.KNNVectorFieldMapperUtil.clipVectorValueToFP16Range;
+import static org.opensearch.knn.index.mapper.KNNVectorFieldMapperUtil.deserializeStoredVector;
 import static org.opensearch.knn.index.mapper.KNNVectorFieldMapperUtil.validateFP16VectorValue;
 import static org.opensearch.knn.index.mapper.KNNVectorFieldMapperUtil.validateVectorDataTypeWithEngine;
 import static org.opensearch.knn.index.mapper.KNNVectorFieldMapperUtil.validateVectorDataTypeWithKnnIndexSetting;
 
 /**
- * Field Mapper for KNN vector type.
- *
- * Extends ParametrizedFieldMapper in order to easily configure mapping parameters.
- *
- * Implementations of this class define what needs to be stored in Lucene's fieldType. This allows us to have
- * alternative mappings for the same field type.
+ * Field Mapper for KNN vector type. Implementations of this class define what needs to be stored in Lucene's fieldType.
+ * This allows us to have alternative mappings for the same field type.
  */
 @Log4j2
 public abstract class KNNVectorFieldMapper extends ParametrizedFieldMapper {
@@ -109,8 +99,8 @@ public abstract class KNNVectorFieldMapper extends ParametrizedFieldMapper {
     public static class Builder extends ParametrizedFieldMapper.Builder {
         protected Boolean ignoreMalformed;
 
-        protected final Parameter<Boolean> stored = Parameter.boolParam("store", false, m -> toType(m).stored, false);
-        protected final Parameter<Boolean> hasDocValues = Parameter.boolParam("doc_values", false, m -> toType(m).hasDocValues, true);
+        protected final Parameter<Boolean> stored = Parameter.storeParam(m -> toType(m).stored, false);
+        protected final Parameter<Boolean> hasDocValues = Parameter.docValuesParam(m -> toType(m).hasDocValues, true);
         protected final Parameter<Integer> dimension = new Parameter<>(KNNConstants.DIMENSION, false, () -> -1, (n, c, o) -> {
             if (o == null) {
                 throw new IllegalArgumentException("Dimension cannot be null");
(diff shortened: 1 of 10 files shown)


src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java
```diff
@@ -109,8 +98,8 @@ private static KNNVectorFieldMapper toType(FieldMapper in) {
     public static class Builder extends ParametrizedFieldMapper.Builder {
         protected Boolean ignoreMalformed;
 
-        protected final Parameter<Boolean> stored = Parameter.boolParam("store", false, m -> toType(m).stored, false);
-        protected final Parameter<Boolean> hasDocValues = Parameter.boolParam("doc_values", false, m -> toType(m).hasDocValues, true);
+        protected final Parameter<Boolean> stored = Parameter.storeParam(m -> toType(m).stored, false);
```
jmazanec15, line 101: changed these to use built ins to clean up the code.

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java
```diff
@@ -735,11 +733,6 @@ Optional<float[]> getFloatsFromContext(ParseContext context, int dimension, Meth
         return Optional.of(array);
     }
 
-    @Override
```
jmazanec15, line 738: Removed as it is default for ParametrizedFieldMapper

jmazanec15:
@luyuncheng could you take a look at this PR?

src/main/java/org/opensearch/knn/index/mapper/LuceneFieldMapper.java
```diff
@@ -92,7 +92,9 @@ protected void parseCreateField(ParseContext context, int dimension, SpaceType s
             KnnByteVectorField point = new KnnByteVectorField(name(), array, fieldType);
 
             context.doc().add(point);
-            addStoredFieldForVectorField(context, fieldType, name(), point);
+            if (this.stored) {
+                context.doc().add(createStoredFieldForVectorField(name(), array));
+            }
```
navneet1v, line 97: can we use the older abstraction where we were validating if we need to store the vector field or not. Rather than adding a if condition here?

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapperUtil.java
```diff
@@ -136,9 +136,39 @@ public static FieldType buildDocValuesFieldType(KNNEngine knnEngine) {
...
-    public static void addStoredFieldForVectorField(ParseContext context, FieldType fieldType, String mapperName, Field vectorField) {
-        if (fieldType.stored()) {
-            context.doc().add(new StoredField(mapperName, vectorField.toString()));
+    /**
+     * Creates a stored field for a byte vector
+     *
+     * @param name field name
+     * @param vector vector to be added to stored field
+     */
+    public static StoredField createStoredFieldForVectorField(String name, byte[] vector) {
```
VijayanB, line 145: nit:createStoredFieldForByteVectors

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapperUtil.java
```diff
@@ -136,9 +136,39 @@ public static FieldType buildDocValuesFieldType(KNNEngine knnEngine) {
...
+        return new StoredField(name, vector);
+    }
+
+    /**
+     * Creates a stored field for a float vector
+     *
+     * @param name field name
+     * @param vector vector to be added to stored field
+     */
+    public static StoredField createStoredFieldForVectorField(String name, float[] vector) {
```
VijayanB, line 155: nit:createStoredFieldForFloatVectors

Method overloading is not readable. It is better to create different name

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapperUtil.java
```diff
@@ -136,9 +136,39 @@ public static FieldType buildDocValuesFieldType(KNNEngine knnEngine) {
...
+     * @param vectorDataType type of vector
+     * @return either int[] or float[] of corresponding vector
+     */
+    public static Object deserializeStoredVector(BytesRef storedVector, VectorDataType vectorDataType) {
+        if (VectorDataType.BYTE == vectorDataType) {
+            byte[] bytes = storedVector.bytes;
+            int[] byteAsIntArray = new int[bytes.length];
+            Arrays.setAll(byteAsIntArray, i -> bytes[i]);
+            return byteAsIntArray;
+        } else {
```
VijayanB, line 170: nit: you don't need else

src/main/java/org/opensearch/knn/index/mapper/LuceneFieldMapper.java
jmazanec15: The old abstraction ended up breaking because in our parameter, in our field mapper build store parameter, we set the stored value for the [KNNFieldMapper parameter](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java#L112) instead of the lucene FieldType parameter:
```
protected final Parameter<Boolean> stored = Parameter.boolParam("store", false, m -> toType(m).stored, false);
```

So, with old abstraction, we would either need to modify it to
1. pass in the KNNVectorFieldMapper object
2. pass in the `bool stored` parameter
3. modify the Lucene FieldType to set stored there.

In terms of setting at [KNNVectorFieldMapper](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java#L489) or [FieldType](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/document/FieldType.java#L114-L117) level, Im not sure what is best practice. Ive seen it done both ways:
1. [TextFieldMapper](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/mapper/TextFieldMapper.java#L284)
2. [NumberFieldMapper](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/mapper/NumberFieldMapper.java#L1510)

@navneet1v do you have any opinions on this? Im going to dig a little bit more in Lucene to see if there are any pitfalls of not setting FieldType.stored but I dont see any obvious ones.

I can switch back to the old abstraction, but then Id probably rename from `addStoredFieldForVectorField` to `maybeAddStoredFieldForVectorField` so that it isnt misleading.

jmazanec15: > Im going to dig a little bit more in Lucene to see if there are any pitfalls of not setting FieldType.stored but I dont see any obvious ones.

https://github.com/apache/lucene/blob/3024e66e4aba942b039fcad7daf958aa4c90b8bf/lucene/core/src/java/org/apache/lucene/document/Field.java#L606-L625. I looked into this, I dont think its applicable. I think storing in the KNNFieldMapper parameter is fine. @navneet1v that being said, I can change `addStoredFieldForVectorField` -> `maybeAddStoredFieldForVectorField` and pass in the bool stored parameter to remove some redundant "if statements" in the code. Im hesistant to pass in the KNNVectorFieldMapper because I think it has a lot more functionality than is required for the helper method.

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapperUtil.java
```diff
@@ -136,9 +136,39 @@ public static FieldType buildDocValuesFieldType(KNNEngine knnEngine) {
...
+            return vectorDataType.getVectorFromDocValues(storedVector);
```
navneet1v, line 171: why we are getting this from docValues?

Wait this is just the conversion and we are reusing the same function. We should rename this function. This is misleading. May be it can be vectorDataType.getVectorFromByteRef something like that.

src/main/java/org/opensearch/knn/index/mapper/LuceneFieldMapper.java
navneet1v: >  Im hesistant to pass in the KNNVectorFieldMapper because I think it has a lot more functionality than is required for the helper method.


I agree on this.

I think now we have overloaded the functions the code bit more clean. Having if statements is fine here.

navneet1v (approved):

VijayanB (approved):

PR #1620, opened 2024-04-17
Separate common lib jni and non-jni into 2 libs

jmazanec15:
### Description
Creates a new shared non-jni library that the jni libraries can depend on. This is to resolve a build issue on windows that is coming from nmslib not being able to find symbols in the jni common library. The net result will be that we add one more shared library.

In addition, I removed all the CPack code from CMakeLists because we dont use it.
 
### Issues Resolved
#1610 
#1591 
 
### Check List
- [X] New functionality includes testing.
  - [X] All tests pass
- [X] New functionality has been documented.
  - [X] New functionality has javadoc added
- [X] Commits are signed as per the DCO using --signoff 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/k-NN/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #1636, opened 2024-04-19
Skip rebuild from scratch after cmake is ran

### LINKED ISSUE #1552: Patch files for nmslib and faiss creating enforcing compilation to happen during cmake

jmazanec15:
### Description
By commiting the patch changes (via `git am` instead of `git apply`), this commit prevents the libraries to be built from scratch every time cmake is ran. This will save additional build time.

from cmake, this looks like:
```
-- Applying custom patches.
Applying: Add IDGrouper for HNSW
Using index info to reconstruct a base tree...
M	faiss/CMakeLists.txt
M	faiss/Index.h
M	faiss/IndexHNSW.cpp
M	faiss/IndexIDMap.cpp
M	faiss/IndexIDMap.h
M	faiss/impl/HNSW.cpp
M	faiss/impl/ResultHandler.h
M	tests/CMakeLists.txt
Falling back to patching base and 3-way merge...
Auto-merging tests/CMakeLists.txt
Auto-merging faiss/impl/HNSW.cpp
Auto-merging faiss/IndexIDMap.cpp
Auto-merging faiss/Index.h
No changes -- Patch already applied.
Applying: Enable precomp table to be shared ivfpq
Using index info to reconstruct a base tree...
M	faiss/IndexIVFPQ.cpp
M	faiss/IndexIVFPQ.h
M	faiss/IndexIVFPQFastScan.cpp
M	faiss/IndexIVFPQFastScan.h
M	tests/CMakeLists.txt
M	tests/test_disable_pq_sdc_tables.cpp
Falling back to patching base and 3-way merge...
Auto-merging tests/CMakeLists.txt
No changes -- Patch already applied.
```

In addition to this, this commit separates out the different functionality to make the build system more readable. We can continue to iterate on this in the future.
 
### Issues Resolved
#1552
 
### Check List
  - [X] All tests pass

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -38,19 +38,11 @@ jobs:
         with:
           submodules: true
 
-      # Git functionality in CMAKE file does not work with given ubuntu image. Therefore, handling it here.
-      - name: Apply Git Patch
-        # Deleting file at the end to skip `git apply` inside CMAKE file
+      # Setup git user so that patches for native libraries can be applied and committed
+      - name: Setup git user
         run: |
-          cd jni/external/faiss
-          git apply --ignore-space-change --ignore-whitespace --3way ../../patches/faiss/0001-Custom-patch-to-support-multi-vector.patch
-          rm ../../patches/faiss/0001-Custom-patch-to-support-multi-vector.patch
-          git apply --ignore-space-change --ignore-whitespace --3way ../../patches/faiss/0002-Enable-precomp-table-to-be-shared-ivfpq.patch
-          rm ../../patches/faiss/0002-Enable-precomp-table-to-be-shared-ivfpq.patch
-          cd ../nmslib
-          git apply --ignore-space-change --ignore-whitespace --3way ../../patches/nmslib/0001-Initialize-maxlevel-during-add-from-enterpoint-level.patch
-          rm ../../patches/nmslib/0001-Initialize-maxlevel-during-add-from-enterpoint-level.patch
-        working-directory: ${{ github.workspace }}
+          su `id -un 1000` -c 'git config --global user.name "github-actions[bot]"'
+          su `id -un 1000` -c 'git config --global user.email "github-actions[bot]@users.noreply.github.com"'
 
       - name: Setup Java ${{ matrix.java }}
         uses: actions/setup-java@v1
@@ -89,6 +81,12 @@ jobs:
       - name: Checkout k-NN
         uses: actions/checkout@v1
 
+      # Setup git user so that patches for native libraries can be applied and committed
+      - name: Setup git user
+        run: |
+          git config --global user.name "github-actions[bot]"
+          git config --global user.email "github-actions[bot]@users.noreply.github.com"
+
       - name: Setup Java ${{ matrix.java }}
         uses: actions/setup-java@v1
         with:
@@ -122,6 +120,12 @@ jobs:
     steps:
       - name: Checkout k-NN
         uses: actions/checkout@v1
+
+      # Setup git user so that patches for native libraries can be applied and committed
+      - name: Setup git user
+        run: |
+          git config --global user.name "github-actions[bot]"
+          git config --global user.email "github-actions[bot]@users.noreply.github.com"
 
       - name: Setup Java ${{ matrix.java }}
         uses: actions/setup-java@v1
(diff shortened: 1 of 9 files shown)


navneet1v:
@jmazanec15 something seems to go wrong with BWC.. Not sure if it is because of this change

jmazanec15:
@navneet1v Windows was a GHA failure to setup Java. The BWC seem more related to this change. Taking a look

.gitignore
```diff
@@ -17,9 +17,7 @@ oss/*
 jni/CMakeCache.txt
 jni/CMakeFiles
 jni/Makefile
-jni/cmake*
-jni/CPack*
```
jmazanec15, line 21: We already removed CPack in #1620

.github/workflows/CI.yml
```diff
@@ -38,19 +38,10 @@ jobs:
         with:
           submodules: true
 
-      # Git functionality in CMAKE file does not work with given ubuntu image. Therefore, handling it here.
```
jmazanec15, line 41: This is no longer issue so I removed. All patches should be specified directly in build system.

navneet1v (dismissed):

jni/cmake/init-faiss.cmake
```diff
@@ -0,0 +1,66 @@
...
+find_package(LAPACK REQUIRED)
+
+# Set relevant properties
+set(BUILD_TESTING OFF)          # Avoid building faiss tests
+set(BLA_STATIC ON)              # Statically link BLAS
+set(FAISS_ENABLE_GPU OFF)
+set(FAISS_ENABLE_PYTHON OFF)
+
+if(NOT DEFINED SIMD_ENABLED)
+    set(SIMD_ENABLED true)   # set default value as true if the argument is not set
```
naveentatikonda, line 54: Have you verified if it is passing the argument properly for `SIMD_ENABLED` after these refactoring changes?

jni/CMakeLists.txt
```diff
@@ -65,144 +67,34 @@ endif()
...
-
-        if(CMAKE_CXX_COMPILER_ID MATCHES "Clang\$")
-            set(OpenMP_CXX_FLAGS "-Xpreprocessor -fopenmp -I/usr/local/opt/libomp/include")
-            set(OpenMP_CXX_LIB_NAMES "omp")
-            set(OpenMP_omp_LIBRARY /usr/local/opt/libomp/lib/libomp.dylib)
-        endif()
-    endif()
-
+    include(${CMAKE_CURRENT_SOURCE_DIR}/cmake/init-faiss.cmake)
     find_package(OpenMP REQUIRED)
```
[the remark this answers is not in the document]
jmazanec15: Good question - I held off here because we directly link the shared library (opensearch_faiss) against openmp here - so I figured it was good to keep in main cmake because its related to opensearch_faiss

.github/workflows/CI.yml
```diff
@@ -38,19 +38,10 @@ jobs:
...
-      - name: Apply Git Patch
-        # Deleting file at the end to skip `git apply` inside CMAKE file
+      - name: Setup git user
```
naveentatikonda, line 41: Can you add one comment for setting up a git user just for reference?

jni/cmake/init-faiss.cmake
jmazanec15: yes, include directive will just basically copy and paste the code into the section its included. See https://cmake.org/cmake/help/latest/command/include.html.

jni/CMakeLists.txt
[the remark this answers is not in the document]
naveentatikonda: makes sense. Thanks!

naveentatikonda (approved):

naveentatikonda (approved):

navneet1v (approved):

Issue #1554, opened 2024-03-18
Script score does not work with model based indices


jmazanec15:
## Description

If the field uses the modelfieldmapper, the field will now be able to work with script scoring:
```
opensearchpy.exceptions.TransportError: TransportError(500, 'search_phase_execution_exception', "Object's dimension=128 does not match the expected dimension=-1.")
```

This is because in script scoring, we validate the dimension and by default it is set to -1 for model-based indices: https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/plugin/script/KNNScoringSpaceUtil.java#L92-L96.

We should retrieve the dimension from model-metadata if this is the case, like in KNNQueryBuilder: https://github.com/opensearch-project/k-NN/blob/4734d88b8b12ff3edd676e3a1f337fc814c8932f/src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java#L289-L297. Or, figure out a more elegant way to handle model-based cases.

PR #1649, opened 2024-04-23
Enable script score to work with model based indices

ryanbogan:
### Description
Previously, an exception would be thrown when attempting to use script score with a model-based index.  Since the expected dimensions for a model-based index field are set to -1, there would be a mismatch between the actual dimension of the float array and the expected dimension.  This PR retrieves the expected dimensions from the model metadata for model-based indices, thus enabling script score to work as intended.
 
### Issues Resolved
https://github.com/opensearch-project/k-NN/issues/1554
 
### Check List
  - [x] All tests pass

### DIFF
--- a/src/main/java/org/opensearch/knn/plugin/script/KNNScoringSpaceUtil.java
+++ b/src/main/java/org/opensearch/knn/plugin/script/KNNScoringSpaceUtil.java
@@ -8,6 +8,9 @@ package org.opensearch.knn.plugin.script;
 import java.util.List;
 import org.opensearch.knn.index.VectorDataType;
 import org.opensearch.knn.index.mapper.KNNVectorFieldMapper;
+import org.opensearch.knn.indices.ModelDao;
+import org.opensearch.knn.indices.ModelMetadata;
+import org.opensearch.knn.indices.ModelUtil;
 import org.opensearch.knn.plugin.stats.KNNCounter;
 import org.opensearch.index.mapper.BinaryFieldMapper;
 import org.opensearch.index.mapper.MappedFieldType;
@@ -20,6 +23,12 @@ import static org.opensearch.index.mapper.NumberFieldMapper.NumberType.LONG;
 import static org.opensearch.knn.common.KNNValidationUtil.validateByteVectorValue;
 
 public class KNNScoringSpaceUtil {
+
+    private static ModelDao modelDao;
+
+    public static void initialize(ModelDao modelDao) {
+        KNNScoringSpaceUtil.modelDao = modelDao;
+    }
 
     /**
      * Check if the passed in fieldType is of type NumberFieldType with numericType being Long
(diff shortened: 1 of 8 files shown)


src/main/java/org/opensearch/knn/plugin/script/KNNScoringSpaceUtil.java
```diff
@@ -137,4 +146,43 @@ public static float getVectorMagnitudeSquared(float[] inputVector) {
...
     }
+
+    /**
+     * Get the expected dimensions from a specified knn vector field type.
+     *
+     * If the field is model-based, get dimensions from model metadata.
+     * @param knnVectorFieldType knn vector field type
+     * @return expected dimensions
+     */
+    public static int getExpectedDimensions(KNNVectorFieldMapper.KNNVectorFieldType knnVectorFieldType) {
```
jmazanec15, line 157: Can we move this to KNNVectorFieldMapperUtil?

src/main/java/org/opensearch/knn/plugin/script/KNNScoringSpaceUtil.java
```diff
@@ -21,6 +24,12 @@
 
 public class KNNScoringSpaceUtil {
 
+    private static ModelDao modelDao;
+
+    public static void initialize(ModelDao modelDao) {
```
martin-gaievski, line 29: method needs javadoc as it's public

src/main/java/org/opensearch/knn/plugin/script/KNNScoringSpaceUtil.java
```diff
@@ -137,4 +146,43 @@ public static float getVectorMagnitudeSquared(float[] inputVector) {
...
+
+    /**
+     * Get the expected dimensions from a specified knn vector field type.
+     *
+     * If the field is model-based, get dimensions from model metadata.
+     * @param knnVectorFieldType knn vector field type
+     * @return expected dimensions
+     */
+    public static int getExpectedDimensions(KNNVectorFieldMapper.KNNVectorFieldType knnVectorFieldType) {
+        int expectedDimensions = knnVectorFieldType.getDimension();
```
martin-gaievski, line 158: nit: you can make if check as a method and name it something like `isModelBasedIndex()`, this way the flow of getExpectedDimensions is easier to read

src/main/java/org/opensearch/knn/plugin/script/KNNScoringSpaceUtil.java
ryanbogan: Sure

src/main/java/org/opensearch/knn/plugin/script/KNNScoringSpaceUtil.java
martin-gaievski: all methods that you're adding, at least public ones: getExpectedDimensions, getModelMetadataForField

martin-gaievski (dismissed):

jmazanec15 (dismissed):

navneet1v:
@ryanbogan I don't see the IT for this change. Can you please add an IT for this change

ryanbogan:
BWC failures are not related to this PR, and were passing earlier

navneet1v (commented):
Overall code looks good to me. Let ensure all gh actions are successful.

jmazanec15 (approved):

martin-gaievski (approved):
Looks good to me, thx. Link checker is failing, but that's not related to changes in this PR

PR #1684, opened 2024-05-02
Add stats for radial search

junqiu-lei:
### Description
Add radial search stats into k-NN plugin stats api, the new stats have:

* max_distance_query_requests
* min_score_query_requests
* max_distance_query_with_filter_requests
* min_score_query_with_filter_requests

### Stats API example
```
GET /_plugins/_knn/stats HTTP/1.1
Host: localhost:9200
```
<details>
<summary>Response details</summary>

```
{
    "_nodes": {
        "total": 1,
        "successful": 1,
        "failed": 0
    },
    "cluster_name": "integTest",
    "circuit_breaker_triggered": false,
    "model_index_status": null,
    "nodes": {
        "nwOR9EKUQEGWmVhT56gDlg": {
            "max_distance_query_with_filter_requests": 3,
            "graph_memory_usage_percentage": 1.211015E-5,
            "graph_query_requests": 12,
            "graph_memory_usage": 2,
            "cache_capacity_reached": false,
            "load_success_count": 1,
            "training_memory_usage": 0,
            "indices_in_cache": {
                "knn-index-test": {
                    "graph_memory_usage_percentage": 1.211015E-5,
                    "graph_memory_usage": 2,
                    "graph_count": 1
                }
            },
            "script_query_errors": 0,
            "hit_count": 11,
            "knn_query_requests": 0,
            "total_load_time": 4045208,
            "miss_count": 1,
            "min_score_query_requests": 6,
            "knn_query_with_filter_requests": 0,
            "training_memory_usage_percentage": 0.0,
            "lucene_initialized": false,
            "max_distance_query_requests": 7,
            "graph_index_requests": 1,
            "faiss_initialized": true,
            "load_exception_count": 0,
            "training_errors": 0,
            "min_score_query_with_filter_requests": 1,
            "eviction_count": 0,
            "nmslib_initialized": false,
            "script_compilations": 0,
            "script_query_requests": 0,
            "graph_stats": {
                "merge": {
                    "current": 0,
                    "total": 0,
                    "total_time_in_millis": 0,
                    "current_docs": 0,
                    "total_docs": 0,
                    "total_size_in_bytes": 0,
                    "current_size_in_bytes": 0
                },
                "refresh": {
                    "total": 1,
                    "total_time_in_millis": 31
                }
            },
            "graph_query_errors": 0,
            "indexing_from_model_degraded": false,
            "graph_index_errors": 0,
            "training_requests": 0,
            "script_compilation_errors": 0
        }
    }
}
```

</details>

### Issues Resolved
[List any issues this PR will resolve]
 
### Check List
  - [ ] All tests pass


src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
```diff
@@ -566,4 +566,23 @@ private static void validateSingleQueryType(Integer k, Float distance, Float sco
             throw new IllegalArgumentException(String.format("[%s] requires exactly one of k, distance or score to be set", NAME));
         }
     }
+
+    private static void updateQueryStats(Integer k, Float minScore, Float maxDistance, QueryBuilder filter) {
```
shatejas, line 570: nit: you can clean this up a bit more to reduce if else. every parameter similar to k or maxDistance we add, will add to the if else statements.

```
validateSingleQuery()
updateQueryStats(k, ...)
updateQueryStats(minScore, ...)
updateQueryStats(maxDistance, ...)
```
```
static <T> void updateQueryStats(T queyParam, QueryBuilder filter, queryCounter, filtercounter) {
if (queyParam != null) {
            queryCounter.increment();
            if (filter != null) {
                filtercounter.increment();
            }
        }
}
```

junqiu-lei: Thanks @shatejas, this suggestion could reduce the function code itself, but looks like we still need extra multiple time calls at same place. Beside the `k` is Integer type, `maxDistance` and `minScore` are Float type, we couldn't directly use parameter `T queyParam` for all of them.

VijayanB: @junqiu-lei Why not use Object?

junqiu-lei: @VijayanB I don't see obviously benefit to introduce Object for this internal used function. Besides if we introduce the Object, we anyway need the condition check for which query type and query counter to use somewhere else.

shatejas: @junqiu-lei yeah extra calls is sometimes a good trade off for maintainability. the cost here shouldn't be high but correct me if I am missing something

@VijayanB On the object suggestion, a few thoughts - The method itself is doing quite a few things - parsing, building query builder, validating and adding stats. Its generally an indication to break it down. What we can actually do is start building query builder while parsing and once we build the KNNQueryBuilder we can validate the entire object as a whole and then update the stats for it as a whole, that way you don't need an extra object.

psuedo code to better understand 
```java
KnnQueryBuilder querybuilder = parseAndBuild(XContentparser) //Note there are no validations here
validate(querybuilder); //This throws if any of the validations fail
updateStats(querybuilder); //update stats
```

Having said this, its too much of a refactor so I wouldn't want to club it with this PR and can be punted till absolutely needed

junqiu-lei: Just had offline with @shatejas, I've updated this method to use generic parameter for queryParam

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
```diff
@@ -566,4 +568,13 @@ private static void validateSingleQueryType(Integer k, Float distance, Float sco
             throw new IllegalArgumentException(String.format("[%s] requires exactly one of k, distance or score to be set", NAME));
         }
     }
+
+    static <T> void updateQueryStats(T queryParam, QueryBuilder filter, KNNCounter queryCounter, KNNCounter queryWithFilterCounter) {
```
navneet1v, line 572: why this function is package private?

junqiu-lei: @navneet1v Updated to class scope private

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
```diff
@@ -566,4 +568,13 @@ private static void validateSingleQueryType(Integer k, Float distance, Float sco
...
     }
+
+    private static <T> void updateQueryStats(T queryParam, QueryBuilder filter, KNNCounter queryCounter, KNNCounter queryWithFilterCounter) {
+        if (queryParam != null) {
+            queryCounter.increment();
+            if (filter != null) {
+                queryWithFilterCounter.increment();
+            }
+        }
+    }
```
navneet1v, line 579: This is an interesting choice of abstraction. Would it be better if we have just created a list of KNNCounters that needs to be incremented, and then just pass on that list to a function may be incrementCounters(List<KNNCounters> countersToBeIncremented) of KNNCounter. I think that would be better simplified the logic for below reasons:
1. You will avoid extra null checks for every queryParam.
2. You will avoid null checks on the filter param. 
Because next time if a new query type will come we will need to repeat below thing for it too.

```
 updateQueryStats(maxDistance, filter, KNNCounter.MAX_DISTANCE_QUERY_REQUESTS, KNNCounter.MAX_DISTANCE_QUERY_WITH_FILTER_REQUESTS);
```

which not required.

src/main/java/org/opensearch/knn/plugin/stats/KNNStats.java
```diff
@@ -98,6 +100,28 @@ private void addQueryStats(ImmutableMap.Builder<String, KNNStat<?>> builder) {
...
+    private void addMaxDistanceQueryStats(ImmutableMap.Builder<String, KNNStat<?>> builder) {
+        builder.put(
+            StatNames.MAX_DISTANCE_QUERY_REQUESTS.getName(),
+            new KNNStat<>(false, new KNNCounterSupplier(KNNCounter.MAX_DISTANCE_QUERY_REQUESTS))
+        )
+            .put(
+                StatNames.MAX_DISTANCE_QUERY_WITH_FILTER_REQUESTS.getName(),
+                new KNNStat<>(false, new KNNCounterSupplier(KNNCounter.MAX_DISTANCE_QUERY_WITH_FILTER_REQUESTS))
+            );
+    }
```
navneet1v, line 123: can we merge these 2 functions with `addKNNQueryStats` and keep the same name which was `addQueryStats`

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
navneet1v: Please check my comment: https://github.com/opensearch-project/k-NN/pull/1684#discussion_r1595893366

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
junqiu-lei: @navneet1v Do you mean use it like below? it could simplify the logic, but I have light concern about the call parameter of countersToBeIncremented list order, we must be careful to prepare this parameter order to use.

```
    private static <T> void updateQueryStats(T queryParam, QueryBuilder filter, List<KNNCounter> countersToBeIncremented) {
        if (queryParam != null) {
            countersToBeIncremented.get(0).increment();
            if (filter != null) {
                countersToBeIncremented.get(1).increment();
            }
        }
    }

updateQueryStats(k, filter, List.of(KNNCounter.KNN_QUERY_REQUESTS, KNNCounter.KNN_QUERY_WITH_FILTER_REQUESTS));
updateQueryStats(minScore, filter, List.of(KNNCounter.MIN_SCORE_QUERY_REQUESTS, KNNCounter.MIN_SCORE_QUERY_WITH_FILTER_REQUESTS));
updateQueryStats(maxDistance, filter, List.of(KNNCounter.MAX_DISTANCE_QUERY_REQUESTS, KNNCounter.MAX_DISTANCE_QUERY_WITH_FILTER_REQUESTS));
```

src/main/java/org/opensearch/knn/plugin/stats/KNNStats.java
junqiu-lei: ack

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
junqiu-lei: Updated PR on your comment @navneet1v

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
```diff
@@ -300,6 +298,14 @@ public static KNNQueryBuilder fromXContent(XContentParser parser) throws IOExcep
 
         validateSingleQueryType(k, maxDistance, minScore);
 
+        updateQueryStats(k, filter, List.of(KNNCounter.KNN_QUERY_REQUESTS, KNNCounter.KNN_QUERY_WITH_FILTER_REQUESTS));
+        updateQueryStats(minScore, filter, List.of(KNNCounter.MIN_SCORE_QUERY_REQUESTS, KNNCounter.MIN_SCORE_QUERY_WITH_FILTER_REQUESTS));
+        updateQueryStats(
+            maxDistance,
+            filter,
+            List.of(KNNCounter.MAX_DISTANCE_QUERY_REQUESTS, KNNCounter.MAX_DISTANCE_QUERY_WITH_FILTER_REQUESTS)
+        );
```
navneet1v, line 307: rather than these 3 functions you can create a list of all the counters needs to be updated during the parsing of the Query. This is what I was saying and then in 1 single shot you can basically set all the counters.

junqiu-lei: "a list of all the counters" might not applicable.. because the counter usage logic of QUERY_REQUESTS and QUERY_WITH_FILTER_REQUESTS are different.

junqiu-lei: Had offline talk with @navneet1v, introduced VectorQueryType enum to handle which counter to use.

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
```diff
@@ -298,7 +296,11 @@ public static KNNQueryBuilder fromXContent(XContentParser parser) throws IOExcep
             }
         }
 
-        validateSingleQueryType(k, maxDistance, minScore);
+        VectorQueryType vectorQueryType = validateSingleQueryType(k, maxDistance, minScore);
```
navneet1v, line 299: rename the function of getQueryType.

navneet1v (commented):
One minor comment but overall looks good to me. Great job on abstracting the QueryType.

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
junqiu-lei: @navneet1v This function meanwhile is used in some other places for validation, so the name of `getQueryType` might isn't suitable I think. How about `handleQueryType`?

https://github.com/opensearch-project/k-NN/blob/9cdbcb2fd393ae77ec30603c929a772c10f3ea3a/src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java#L104

https://github.com/opensearch-project/k-NN/blob/9cdbcb2fd393ae77ec30603c929a772c10f3ea3a/src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java#L121

https://github.com/opensearch-project/k-NN/blob/9cdbcb2fd393ae77ec30603c929a772c10f3ea3a/src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java#L135

navneet1v (approved):

VijayanB (approved):

Issue #1424, opened 2024-01-26
[BUG]The exception is unfriendly(Mapper parsing exception) when k-NN circuit breaker is triggered


heemin32:
**What is the bug?**
During bulk indexing, when k-NN circuit breaker is triggered, client receive mapper_parsing_exception. This makes user hard to understand the actual root cause of the exception until they look in to a log file.

**How can one reproduce the bug?**
```
"type": "knn_vector",
"dimension": 256,
"method": {
    "name": "hnsw",
    "engine": "faiss",
    "space_type": "innerproduct",
    "parameters": {
        "ef_construction": 100,
        "m": 16
    }
}
```

Do bulk indexing of documents

```
Error message from client side
failed to parse field [passageVector] of type [knn_vector] in document with id 'asdfasdf'. mapper_parsing_exception'
Preview of fields's value: '[1.23232, 2.123232. ...]', statusCode: 400, type: mapper_parsing_exception
```

```
Log
MapperParsingException[failed to parse field [passageVector] of type [knn_vector] in document with id 'asdfasdf'. Preview of field's value: '[1.23232, 2.123232, ...]']; nested: IllegalStateException[Indexing knn vector fields is rejected as circuit breaker triggered. Check _opendistro/_knn/stats for detailed state];
```
**What is the expected behavior?**
It should provide the root cause of the exception, k-NN circuit breaker in client side error message

navneet1v:
@heemin32 how this is bug? I would consider this as an enhancement where messaging is not proper.

heemin32:
It is a bug because the error message is not correct. If the exception is InternalException or ThrottlingException, it is not a bug. However, mapper parsing exception does not encompass circuit breaker exception but those two are entirely different exceptions.

ryanbogan:
[...]

PR #1688, opened 2024-05-03
Add KnnCircuitBreakerException and modify exception message

ryanbogan:
### Description
Currently, when the circuit breaker is triggered and the user attempts to bulk index documents, an illegal state exception is thrown with a message that is not entirely accurate.  This PR introduces a new type of exception, `KnnCircuitBreakerException`, to be thrown for any action that fails because the circuit breaker is triggered.  In addition, it changes the exception message described above to clarify where exactly the exception is being thrown in the indexing process and provide more information about the circuit breaker.
 
### Issues Resolved
https://github.com/opensearch-project/k-NN/issues/1424
 
### Check List
  - [x] All tests pass


src/main/java/org/opensearch/knn/index/KnnCircuitBreakerException.java
```diff
@@ -0,0 +1,65 @@
...
+ * compatible open source license.
+ *
+ * Modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+package org.opensearch.knn.index;
+
+public class KnnCircuitBreakerException extends RuntimeException {
+    public KnnCircuitBreakerException() {
```
shatejas, line 15: Might be worth removing the default constructor so the client of this class is forced to add a message

ryanbogan: @shatejas What do you think about adding a default message in the constructor instead of removing it entirely?

ryanbogan:
Rolling upgrade failures are happening on other PR's too and are not related to this change.  Created an issue here: https://github.com/opensearch-project/k-NN/issues/1691

src/main/java/org/opensearch/knn/index/KnnCircuitBreakerException.java
shatejas: As long as clients have enough information to understand what went wrong the default message works. 

The intent is not to throw without a proper message considering its a custom exception and might be hard for someone to figure out what went wrong

ryanbogan: After trying to implement the default, it doesn't work because it's extending RuntimeException which needs the super to be called first.  I'll just remove the default constructor per your original recommendation.

shatejas (approved):

src/main/java/org/opensearch/knn/index/KnnCircuitBreakerException.java
```diff
@@ -0,0 +1,62 @@
...
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ * Modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+package org.opensearch.knn.index;
+
+public class KnnCircuitBreakerException extends RuntimeException {
```
martin-gaievski, line 14: Please add description for the class

src/main/java/org/opensearch/knn/index/KnnCircuitBreakerException.java
```diff
@@ -0,0 +1,62 @@
...
+
+    /**
+     * Constructs a KnnCircuitBreakerException with the specified detail
+     * message.  A detail message is a String that describes this particular
+     * exception.
+     *
+     * @param s the String that contains a detailed message
+     */
+    public KnnCircuitBreakerException(String s) {
```
martin-gaievski, line 23:
```suggestion
    public KnnCircuitBreakerException(final String message) {
```

martin-gaievski (approved):

jmazanec15 (approved):

Issue #1337, opened 2023-12-07
[BUG]Commas in model description cause ingestion to fail


heemin32:
**What is the bug?**
When model description contains a comma, knn plugin fails to parse model metadata properly and throw an exception in https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/indices/ModelMetadata.java#L240

```
"Illegal format for model metadata. Must be of the form \"<KNNEngine>,<SpaceType>,<Dimension>,<ModelState>,<Timestamp>,<Description>,<Error>\"."
```

**How can one reproduce the bug?**
Training a knn model with a description containing comma cause ingestion to fail using the model
```
POST /_plugins/_knn/models/test_model/_train
{
	"training_index": "train_index",
	"training_field": "train_field",
	"dimension": 768,
	"max_training_vector_count": 10000,
	"description": "teat A, B",
	"method": {
		"name": "hnsw",
		"engine": "faiss",
		"space_type": "l2",
		"parameters": {
			"encoder": {
				"name": "pq",
				"parameters": {
					"m": 64,
					"code_size": 8
				}
			}
		}
	}
}
```

**What is the expected behavior?**
The ingestion should not fail

vamshin:
Good catch!! Thanks @heemin32

simonhessner:
Thanks for looking into this so quickly. It's probably sufficient to escape the delimiter character in all fields or to prevent the user from using the delimiter in the description entirely

heemin32:
Changing delimiter might break backward compatibility. Blocking comma in description would be simpler to prevent the issue.

Issue #1479, opened 2024-02-10
Changing delimiter might break backward compatibility. Blocking comma in description would be simpler to prevent the issue.


michel344:
Changing delimiter might break backward compatibility. Blocking comma in description would be simpler to prevent the issue.

_Originally posted by @heemin32 in https://github.com/opensearch-project/k-NN/issues/1337#issuecomment-1847347488_

Issue #1691, opened 2024-05-06
BWC Rolling Upgrade tests failing


ryanbogan:
Currently, BWC rolling upgrade tests are failing with the following error:

```
↓ errors and warnings from /home/runner/work/k-NN/k-NN/qa/rolling-upgrade/build/testclusters/knnBwcCluster-rolling-2/logs/opensearch.stdout.log ↓
» WARN ][o.o.g.DanglingIndicesState] [knnBwcCluster-rolling-2] gateway.auto_import_dangling_indices is disabled, dangling indices will not be automatically detected or imported and must be managed manually
» WARN ][o.o.d.FileBasedSeedHostsProvider] [knnBwcCluster-rolling-2] expected, but did not find, a dynamic hosts list at [/home/runner/work/k-NN/k-NN/qa/rolling-upgrade/build/testclusters/knnBwcCluster-rolling-2/config/unicast_hosts.txt]
» WARN ][o.o.c.s.MasterService    ] [knnBwcCluster-rolling-2] failing [elected-as-cluster-manager ([3] nodes joined)[{knnBwcCluster-rolling-2}{o4vnPogpQqO6HsmBUtfNGw}{V_42OJ-oT8WWfb2Np4vUGA}{127.0.0.1}{127.0.0.1:45009}{dimr}{testattr=test, shard_indexing_pressure_enabled=true} elect leader, {knnBwcCluster-rolling-0}{Md9MTY4rTZiuFe6_CfeW1w}{9L3fYVNKQvaFHBtpd9LZpQ}{127.0.0.1}{127.0.0.1:42881}{dimr}{testattr=test, shard_indexing_pressure_enabled=true} elect leader, {knnBwcCluster-rolling-1}{fWwXwx_dSdqpKeu1eiCkZg}{cBQqyuvkQTyKg6ugNlabuw}{127.0.0.1}{127.0.0.1:32997}{dimr}{testattr=test, shard_indexing_pressure_enabled=true} elect leader, _BECOME_CLUSTER_MANAGER_TASK_, _FINISH_ELECTION_]]: failed to commit cluster state version [1]
»  org.opensearch.cluster.coordination.FailedToCommitClusterStateException: node is no longer cluster-manager for term 1 while handling publication
»  	at org.opensearch.cluster.coordination.Coordinator.publish(Coordinator.java:1294) ~[opensearch-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
[...]

PR #1692, opened 2024-05-06
Block commas in model description

ryanbogan:
### Description
When commas are present in the model description, the model metadata is not parsed correctly, resulting in an exception being thrown when ingesting data to or searching on a model-based index.  This PR throws an IllegalArgumentException if the user attempts to create a model with a comma in the model description.
 
### Issues Resolved
https://github.com/opensearch-project/k-NN/issues/1337
https://github.com/opensearch-project/k-NN/issues/1479
 
### Check List
  - [x] All tests pass

### DIFF
--- a/src/main/java/org/opensearch/knn/indices/ModelMetadata.java
+++ b/src/main/java/org/opensearch/knn/indices/ModelMetadata.java
@@ -67,9 +67,7 @@ public class ModelMetadata implements Writeable, ToXContentObject {
         // Description and error may be empty. However, reading the string will work as long as they are not null
         // which is checked in constructor and setters
         this.description = in.readString();
-        if (this.description.contains(",")) {
-            throw new IllegalArgumentException("Model description cannot contain any commas: ','.");
-        }
+        ModelUtil.blockCommasInModelDescription(this.description);
         this.error = in.readString();
 
         if (IndexUtil.isVersionOnOrAfterMinRequiredVersion(in.getVersion(), IndexUtil.MODEL_NODE_ASSIGNMENT_KEY)) {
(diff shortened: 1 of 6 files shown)


ryanbogan:
Rolling upgrade failures unrelated: https://github.com/opensearch-project/k-NN/issues/1691

src/main/java/org/opensearch/knn/indices/ModelMetadata.java
```diff
@@ -123,6 +126,9 @@ public ModelMetadata(
         this.state = new AtomicReference<>(Objects.requireNonNull(modelState, "modelState must not be null"));
         this.timestamp = Objects.requireNonNull(timestamp, "timestamp must not be null");
         this.description = Objects.requireNonNull(description, "description must not be null");
+        if (this.description.contains(",")) {
+            throw new IllegalArgumentException("Model description cannot contain any commas: ','");
```
VijayanB, line 130: nit: may be "model" instead of "Model"? i see every other exception message doesn't start with caps.

ryanbogan: All of the null check exceptions are lower case, but the other IllegalArgumentException earlier in the method that validates the max dimensions starts with a cap.

jmazanec15:
Can we add the check in the rest handler as well as ModelMetadata? Im just thinking we should fail fast on this when possible (i.e. before any transport requests are sent or state is serialized)

jmazanec15 (approved):

src/test/java/org/opensearch/knn/plugin/action/RestTrainModelHandlerIT.java
```diff
@@ -192,6 +193,68 @@ public void testTrainModel_fail_tooMuchData() throws Exception {
...
+
+        String modelId = "test-model-id";
+        String trainingIndexName = "train-index";
+        String trainingFieldName = "train-field";
+        int dimension = 8;
+
+        // Create a training index and randomly ingest data into it
+        createBasicKnnIndex(trainingIndexName, trainingFieldName, dimension);
+        int trainingDataCount = 200;
+        bulkIngestRandomVectors(trainingIndexName, trainingFieldName, trainingDataCount, dimension);
```
martin-gaievski, line 207: do we need to ingest vectors for this test? Training should fail because of the comma in model description.

ryanbogan: Good catch

martin-gaievski (approved):
looks good to me, thank you

jmazanec15 (approved):

PR #1701, opened 2024-05-14
[backport 2.x] Add stats for radial search

junqiu-lei:
### Description

Backport #1684 into 2.x branch

 
### Check List
  - [x] All tests pass


ryanbogan (approved):

src/test/java/org/opensearch/knn/plugin/action/RestKNNStatsHandlerIT.java
```diff
@@ -16,41 +17,24 @@
...
-import static org.opensearch.knn.common.KNNConstants.METHOD_HNSW;
-import static org.opensearch.knn.common.KNNConstants.METHOD_IVF;
-import static org.opensearch.knn.common.KNNConstants.METHOD_PARAMETER_NLIST;
-import static org.opensearch.knn.common.KNNConstants.METHOD_PARAMETER_SPACE_TYPE;
-import static org.opensearch.knn.common.KNNConstants.MODEL_ID;
-import static org.opensearch.knn.common.KNNConstants.MODEL_INDEX_NAME;
-import static org.opensearch.knn.common.KNNConstants.NAME;
-import static org.opensearch.knn.common.KNNConstants.NMSLIB_NAME;
-import static org.opensearch.knn.common.KNNConstants.PARAMETERS;
+import java.util.*;
```
Vikasht34, line 34: Can we avoid * and use whichever is required.

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
```diff
@@ -580,21 +583,27 @@ public String getWriteableName() {
         return NAME;
     }
 
-    private static void validateSingleQueryType(Integer k, Float distance, Float score) {
+    private static VectorQueryType validateSingleQueryType(Integer k, Float distance, Float score) {
         int countSetFields = 0;
```
Vikasht34, line 587: Can we avoid all together customeSetFields and fall early based on if all of the parameters are null or zero. Which is going to simply all 3 ifs we are putting together

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
```diff
@@ -322,7 +321,11 @@ public static KNNQueryBuilder fromXContent(XContentParser parser) throws IOExcep
             }
         }
 
-        validateSingleQueryType(k, maxDistance, minScore);
+        VectorQueryType vectorQueryType = validateSingleQueryType(k, maxDistance, minScore);
```
Vikasht34, line 324: Probably we can rename the method and move increment operation local to that function validated of any of the parameters are not null and set the enum as well the increment.

Vikasht34: It helps to localize the logic.

junqiu-lei:
@Vikasht34 This is a backport PR only, the original PR got approved from https://github.com/opensearch-project/k-NN/pull/1684

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
junqiu-lei: We have to verify these three parameters together because we need check only exactly one of them can be pass in.

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
junqiu-lei: @Vikasht34 Thanks for the comment, I had talked with @navneet1v on this part on original PR, if we separate out to validateSingleQuery and getQueryType, there will have duplicate code between the two methods. So I keep it as single method here for now.

src/test/java/org/opensearch/knn/plugin/action/RestKNNStatsHandlerIT.java
junqiu-lei: We probably can update in a future PR as it's a backport PR.

navneet1v (approved):

Issue #1537, opened 2024-03-14
[FEATURE] Enable setting ef_search parameter as part of k-NN Query


navneet1v:
**Is your feature request related to a problem?**
While using HNSW algorithm currently there is no support for sending the `ef_search` parameter as part of query. The ask here is to support functionality of setting the ef_search in the k-NN query.
Even though the downstream libraries like Faiss and Nmslib has support for this.

Faiss: https://github.com/facebookresearch/faiss/blob/dafdff110489db7587b169a0afee8470f220d295/faiss/impl/HNSW.h#L46

**What solution would you like?**
While doing a k-NN query as a user I should be able to provide the `ef_search` value.

**What alternatives have you considered?**
Currently in Nmslib I can update this value per index by updating the index settings. But this functionality is not there for faiss.

**Do you have any additional context?**
NA

jmazanec15:
@navneet1v does nmslib support passing this at search time? I dont think it does.

jmazanec15:
Also, IVF would benefit from this as well for the nprobe parameter. The nprobe should be easier to do once we have https://github.com/opensearch-project/k-NN/pull/1527 checked in as well. We can then just say for the nprobe parameter passed in at training is just the default.

navneet1v:
> @navneet1v does nmslib support passing this at search time? I dont think it does.

The way it works is you need to unload the graphs and then load again with new values. I do understand this has potential issues. But we should move towards making the change in nmslib to avoid making this as graph parameter.

shatejas:
Picking this up

shatejas:
# [RFC] Adding efSearch as query parameter for kNN

## Overview 

The efSearch parameter is utilized in the HNSW (Hierarchical Navigable Small World) algorithm to facilitate the exploration of more neighboring candidates, which in turn enhances the potential recall. A higher value for efSearch can often result in improved recall.

Opensearch k-NN plugin offers three engines options for HNSW-based kNN: NMSLIB, FAISS and Lucene. Currently customers can specify the efSearch option for NMSLIB and FAISS implementations as an index setting. This setting is established during index creation process .

FAISS ef_search value cannot be changed after the index is created. NMSLIB value can be changed. In the case of the Lucene implementation, the efSearch parameter is currently tied to the k parameter.

The objective of the feature is to enable efSearch as a query parameter across all three engines. This documents outlines the challenges associated with the goal, along with potential solutions and recommendations.

## API Contract changes

[...]

PR #1707, opened 2024-05-16
Adds support for ef_search query parameter in FAISS engine

shatejas:
### Description

This introduces `ef_search` as a query parameter for search based on hnsw algorithm. This PR has changes for FAISS engine. Support for other engines will be added in consecutive PRs. Radial search support for FAISS will also be added in consecutive PRs. These changes are focused on k based queries.
 
### Issues Resolved
https://github.com/opensearch-project/k-NN/issues/1537

### Local testing
```
curl -XGET "http://localhost:9200/faiss-hnsw-index/_search" -H 'Content-Type: application/json' -d'
{
  "query": {
    "knn": {
      "faiss_vector": {
        "vector": [2, 3, 5],
        "k": 2,
        "method_parameters": {
           "ef_search" : 5

          }
        }
      }
    }
  }'
```
```
{
  "took": 69,
  "timed_out": false,
  "_shards": {
    "total": 1,
    "successful": 1,
    "skipped": 0,
    "failed": 0
  },
  "hits": {
    "total": {
      "value": 2,
      "relation": "eq"
    },
    "max_score": 0.11918951,
    "hits": [
      {
        "_index": "faiss-hnsw-index",
        "_id": "3",
        "_score": 0.11918951,
        "_source": {
          "faiss_vector": [
            3.5,
            4.5,
            3.3
          ],
          "price": 12.9
        }
      },
      {
        "_index": "faiss-hnsw-index",
        "_id": "2",
        "_score": 0.10706638,
        "_source": {
          "faiss_vector": [
            2.5,
            3.5,
            2.2
          ],
          "price": 7.1
        }
      }
    ]
  }
}
```
 
### Check List
  - [x] All tests pass
- [ ] New functionality has been documented. - pending documentation
  - [] New functionality has javadoc added - added wherever I felt required. open to feedback and I can add it


jni/src/faiss_wrapper.cpp
```diff
@@ -371,12 +370,13 @@ jobjectArray knn_jni::faiss_wrapper::QueryIndex_WithFilter(knn_jni::JNIUtilInter
         std::unique_ptr<faiss::IDGrouperBitmap> idGrouper;
         std::vector<uint64_t> idGrouperBitmap;
         auto hnswReader = dynamic_cast<const faiss::IndexHNSW*>(indexReader->index);
-        if(hnswReader!= nullptr && parentIdsJ != nullptr) {
-            // Setting the ef_search value equal to what was provided during index creation. SearchParametersHNSW has a default
-            // value of ef_search = 16 which will then be used.
-            hnswParams.efSearch = hnswReader->hnsw.efSearch;
-            idGrouper = buildIDGrouperBitmap(jniUtil, env, parentIdsJ, &idGrouperBitmap);
-            hnswParams.grp = idGrouper.get();
+        if(hnswReader!= nullptr) {
```
shatejas, line 373: - Currently working on unit test for this class
- There is a minor behavior change here: Up until now if there are not filters or parent filters the efsearch parameter that was passed had a default value of 16. This makes sure it is picked from index or query. the default 16 will ever be used.

jni/src/faiss_wrapper.cpp
junqiu-lei, line 1: I think we also need support the queryEfSearch for Faiss radial search(range search in FAISS) 

https://github.com/opensearch-project/k-NN/blob/main/jni/src/faiss_wrapper.cpp#L596

shatejas: I am planning to create a a separate PR for radial search. The current changes are only for k based KNN

shatejas: Updated the description

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
```diff
@@ -216,6 +300,7 @@ public KNNQueryBuilder(StreamInput in) throws IOException {
             vector = in.readFloatArray();
             k = in.readInt();
             filter = in.readOptionalNamedWriteable(QueryBuilder.class);
+            efSearch = in.readOptionalInt();
```
junqiu-lei, line 303: For version compatibility for multiple nodes in difference versions OpenSearch cluster, we need version check to determine if it need read and write efSearch value. Example:
```
if (isClusterOnOrAfterMinRequiredVersion(KNNConstants.EF_SEARCH_KEY)) {
                efSearch = in.readOptionalFloat();
            }
```

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
```diff
@@ -324,6 +404,7 @@ protected void doWriteTo(StreamOutput out) throws IOException {
         out.writeFloatArray(vector);
         out.writeInt(k);
         out.writeOptionalNamedWriteable(filter);
+        out.writeOptionalInt(efSearch);
```
junqiu-lei, line 407: Save above, need version check.

jni/src/faiss_wrapper.cpp
junqiu-lei: Got it, thanks explanation.

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
shatejas: Got it. I saw that part and wasn't sure why it was required. Thanks for explanation, I will change it

jni/include/faiss_wrapper.h
```diff
@@ -45,17 +45,27 @@ namespace knn_jni {
...
-        // Return an array of KNNQueryResults
+         /**
+         *  Execute a query against the index located in memory at indexPointerJ
+         *  
+         * Parameters:
+         * queryEfSearch: -1 indicates to use efsearch value used during index setting 
+         * 
+         * Return an array of KNNQueryResults
+        */
         jobjectArray QueryIndex(knn_jni::JNIUtilInterface * jniUtil, JNIEnv * env, jlong indexPointerJ,
```
Vikasht34, line 56: Can we create Overloaded method for this with new parameter , Just to Make Sure ...It works with and without EfSearch passed with Query during whole deployment Process and till it goes out.

jni/include/faiss_wrapper.h
```diff
@@ -45,17 +45,27 @@ namespace knn_jni {
...
+
+        /**
+         *  Execute a query against the index located in memory at indexPointerJ along with Filters
+         *  
+         * Parameters:
+         * queryEfSearch: -1 indicates to use efsearch value used during index setting 
+         * 
+         * Return an array of KNNQueryResults
+        */
         jobjectArray QueryIndex_WithFilter(knn_jni::JNIUtilInterface * jniUtil, JNIEnv * env, jlong indexPointerJ,
```
Vikasht34, line 67: Can we create Overloaded method for this with new parameter , Just to Make Sure ...It works with and without EfSearch passed with Query during whole deployment Process and till it goes out.

jni/src/faiss_wrapper.cpp
Vikasht34, line 373: Did we test with cases what happnes when hnswReader is not null and parentIdsJ is Null and vice versa  ,  I feel earlier condition was fine to check both not null.  And Can we add test case for all Corner cased like when one is null and other is not , just to make sure it works , Since we are removing one condition from Top If Statment.

src/main/java/org/opensearch/knn/index/query/KNNQuery.java
```diff
@@ -23,26 +21,28 @@
...
  */
+@Getter
+@Builder
+@AllArgsConstructor
 public class KNNQuery extends Query {
 
     private final String field;
     private final float[] queryVector;
     private int k;
+    private Integer efSearch;
```
Vikasht34, line 39: It's good that we made Data Type Integer rather int , but are we checking null in efSearch?

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
```diff
@@ -93,61 +103,133 @@ public KNNQueryBuilder(String fieldName, float[] vector) {
...
-     * @param k K nearest neighbours for the given vector
+     * lombok SuperBuilder annotation requires a builder annotation on parent class to work well
+     * {@link AbstractQueryBuilder#boost()} and {@link AbstractQueryBuilder#queryName()} both need to be called
+     * A custom builder helps with the calls to the parent class, simultaneously addressing the problem of telescoping
+     * constructors in this class.
      */
-    public KNNQueryBuilder k(Integer k) {
-        if (k == null) {
-            throw new IllegalArgumentException(String.format("[%s] requires k to be set", NAME));
+    public static class Builder {
```
Vikasht34, line 111: Good Improvement to use Builder ..but we can avoid writing whole stuff by our self with little improvement 


``
```
@AllArgsConstructor
@Getter
@Setter
@Builder
public class KNNQueryBuilder extends AbstractQueryBuilder<KNNQueryBuilder> {
    // Existing fields remain unchanged
    @NonNull
    private final String fieldName;
    @NonNull
    private final float[] vector;
    @Builder.Default
    private int k = 0;
    private Integer efSearch;
    private Float maxDistance;
    private Float minScore;
    private QueryBuilder filter;
    @Builder.Default
    private boolean ignoreUnmapped = false;
    private String queryName;
    private float boost = DEFAULT_BOOST;

    // Custom validation method
    private void validate() {
        if (k < 0 || k > K_MAX) {
            throw new IllegalArgumentException(String.format("[%s] requires k to be in the range [0, %d]", NAME, K_MAX));
        }

        // Add more validation rules as needed
    }

    // Builder method for k with custom validation
    public KNNQueryBuilder k(int k) {
        this.k = k;
        validate();
        return this;
    }

    // Other builder methods remain unchanged
}
```

src/main/java/org/opensearch/knn/jni/FaissService.java
```diff
@@ -145,6 +145,7 @@ public static native KNNQueryResult[] queryIndexWithFilter(
         long indexPointer,
         float[] queryVector,
         int k,
+        int queryEfSearch,
```
Vikasht34, line 148: nit :- Can we have all Places either Integer or int

src/main/java/org/opensearch/knn/jni/JNIService.java
```diff
@@ -173,11 +175,13 @@ public static KNNQueryResult[] queryIndex(
         long indexPointer,
         float[] queryVector,
         int k,
+        @Nullable Integer efSearch,
```
Vikasht34, line 178: nit : Can we have nullable either at end of parameter or let's remove nullable , we can null any how With Integer.

src/main/java/org/opensearch/knn/index/query/KNNQueryFactory.java
```diff
@@ -79,12 +80,24 @@ public static Query create(CreateQueryRequest createQueryRequest) {
         }
 
         if (KNNEngine.getEnginesThatCreateCustomSegmentFiles().contains(createQueryRequest.getKnnEngine())) {
-            if (filterQuery != null && KNNEngine.getEnginesThatSupportsFilters().contains(createQueryRequest.getKnnEngine())) {
-                log.debug("Creating custom k-NN query with filters for index: {}, field: {} , k: {}", indexName, fieldName, k);
-                return new KNNQuery(fieldName, vector, k, indexName, filterQuery, parentFilter);
-            }
-            log.debug(String.format("Creating custom k-NN query for index: %s \"\", field: %s \"\", k: %d", indexName, fieldName, k));
-            return new KNNQuery(fieldName, vector, k, indexName, parentFilter);
+            final Query validatedFilterQuery = validateFilterQuerySupport(filterQuery, createQueryRequest.getKnnEngine());
```
Vikasht34, line 83: nit :- Can we just rename to filteredQuery.

jni/src/faiss_wrapper.cpp
shatejas: With the earlier condition efSearch was not getting updated and it was using the default value with the combination of filter and parent filter being null so I don't think its fine.

As mentioned, still working on the unit test

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
shatejas: 1. This is not going to work as mentioned in the comment above the `Builder` class definition. There is a parent class involved there which holds certain setter methods which won't seamlessly work with lombok. I wouldn't want to make changes in opensearch repo as there might be many queries using it.

Considering telescoping constructors, its worth maintaining a builder than multiple constructors which is more error prone and unreadable.

2.
```
public KNNQueryBuilder k(int k) { this.k = k; validate(); return this; }
```
This opens up a way to modify k (or any other request parameter), which I don't think is a good idea considering this object goes to parts of code that are not owned by us. This is an object which is formed after parsing json request. With buillder class and no setters I am making it immutable in a way.

3. Individual validations are avoided as suggested because there are validations which involve a combination of variables.

src/main/java/org/opensearch/knn/index/query/KNNQueryFactory.java
shatejas: `validated` suffix seems more appropriate

src/main/java/org/opensearch/knn/jni/FaissService.java
shatejas: I will prefer being consistent here, This is a JNI layer and we seem to have been using primitive data types here, there is no real advantage of using `Integer` as we will later have to add lines to convert it to cpp int. We won't be able to avoid a dummy value in either case

I am not using `int` elsewhere because there is a possibility of value being absent which I feel is better than a dummy value.

src/main/java/org/opensearch/knn/jni/JNIService.java
shatejas: - Why at the end of the parameter?
- Its a clear indication that this value can be null and modern editors like intellij helps out if you forget the check in code. I would keep it  as it is unless there is a strong reason to remove it.

src/main/java/org/opensearch/knn/index/query/KNNQuery.java
shatejas: I believe so wherever required

src/main/java/org/opensearch/knn/common/KNNConstants.java
```diff
@@ -24,6 +24,7 @@ public class KNNConstants {
     public static final String KNN = "knn";
     public static final String VECTOR = "vector";
     public static final String K = "k";
+    public static final String EF_SEARCH = "ef_Search";
```
VijayanB, line 27:
```suggestion
    public static final String EF_SEARCH = "ef_search";
```

VijayanB: We also have another constance METHOD_PARAMETER_EF_SEARCH , can we not re-factor and use it in both places?

qa/rolling-upgrade/src/test/java/org/opensearch/knn/bwc/QueryANNIT.java
```diff
@@ -0,0 +1,46 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ * Modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
```
VijayanB, line 10: nit: Did you generate this license or was it added by IDE? we moved to shorter version, if IDE added longer version, then we might need to fix this for another commit.

src/main/java/org/opensearch/knn/index/IndexUtil.java
```diff
@@ -51,6 +51,8 @@ public class IndexUtil {
             put(MODEL_NODE_ASSIGNMENT_KEY, MINIMAL_SUPPORTED_VERSION_FOR_MODEL_NODE_ASSIGNMENT);
             put(MODEL_METHOD_COMPONENT_CONTEXT_KEY, MINIMAL_SUPPORTED_VERSION_FOR_MODEL_METHOD_COMPONENT_CONTEXT);
             put(KNNConstants.RADIAL_SEARCH_KEY, MINIMAL_SUPPORTED_VERSION_FOR_RADIAL_SEARCH);
+            // TODO: Make 2.15 when we merge into 2.x. Making it 3.0.0 to make sure BWC passes
+            put(KNNConstants.EF_SEARCH, Version.V_3_0_0);
```
VijayanB, line 55: nit: MINIMAL_SUPPORTED_VERSION_FOR_EF_SEARCH_AS_QUERY_PARAM

jmazanec15:
```
{
  "query": {
    "knn": {
      "faiss_vector": {
        "vector": [2, 3, 5],
        "k": 2,
        "ef_search":5
      }
    }
  }
}
```

For interface, Im wondering if we should separate algo specific parameters into a nested `parameters` field. For instance, for faiss, we also will easily want to add parameters for `nprobes` to this. Additionally, as more algorithms are added, this main body will become cluttered.

src/main/java/org/opensearch/knn/common/KNNConstants.java
shatejas: I missed the existing constant, will refactor it for consistency

qa/rolling-upgrade/src/test/java/org/opensearch/knn/bwc/QueryANNIT.java
shatejas: I believe I just copy pasted it from somewhere, is there a preference I don't mind changing it

src/main/java/org/opensearch/knn/index/IndexUtil.java
shatejas: I saw that and it felt redundant considering we already have one constant and there is no reuse anticipated. I can change if you feel strongly about it

shatejas:
> For interface, Im wondering if we should separate algo specific parameters into a nested `parameters` field. For instance, for faiss, we also will easily want to add parameters for `nprobes` to this. Additionally, as more algorithms are added, this main body will become cluttered.

I am open to it. A few thoughts here
1. Client will need to have the knowledge of algorithm during query time. I don't have insights right now to say whether its good or bad.
2. Is there any impact on existing use cases, Will we have to deprecate any parameter?

One downside is it creates nesting of parameters depending on the approach, for instance
```
{
  "query": {
    "knn": {
      "faiss_vector": {
        "vector": [2, 3, 5],
        "hnsw": {
               k: 2,
               efSearch: 5
        }
      } 
    }
  }
}
```

Another one is that we might not get away with cluttering completely for instance lets say we introduce algo name in the query, something like. This moves the cluttering to a different block altogether
```
{
  "query": {
    "knn": {
      "faiss_vector": {
        "vector": [2, 3, 5],
      }
     "hnsw": {
           k: 2 // Not that this can be maxDist or minScore  for radial search
          efSearch: 5
      }
    }
  }
}
```
Also we will have to add additional validations. The `faiss_vector` might not be indexed with hnsw algorithm and might be `ivf`
Let me know your thoughts on it and I am open to suggestions here.

jmazanec15:
I think we can always validate against the field mapping that is being searched. That being said, I like the following:
```
{
  "query": {
    "knn": {
      "faiss_vector": {
        "vector": [2, 3, 5],
        "k": 2, // Note that this can be maxDist or minScore  for radial search
        "method_parameters": {
// method specific parameters we can pack in here
               "ef_search": 5,
               // "nprobes": 72
        }
      } 
    }
  }
}
```

With this, we can extend validate with the mapping being searched if the method parameter is valid. I really want to dynamically support nprobes because this is a value that needs to be tweaked during experimentation and to tweak it requires training a new model and rebuilding an index.

qa/rolling-upgrade/src/test/java/org/opensearch/knn/bwc/QueryANNIT.java
VijayanB:
```
/*
 */
 ```

File: src/main/java/org/opensearch/knn/index/IndexUtil.java (hunk 13 above)
Comment (VijayanB) [reply src/main/java/org/opensearch/knn/index/IndexUtil.java:55]: In this way you can only flip that value in 2.x and main at constant value instead of touching this method. what do you think?

Comment (shatejas) [reply src/main/java/org/opensearch/knn/index/IndexUtil.java:55]: The map itself is the constant, doesn't seem to be a big deal here. either way works for me

**Comment by jmazanec15** [comment]
@shatejas Can we also add support for dynamic nprobes for IVF? This will make iterating/tuning with IVF a lot easier. We will need to read model definition from cluster state, but @ryanbogan should be able to help provide some insights on that.

**Comment by shatejas** [comment]
> @shatejas Can we also add support for dynamic nprobes for IVF? This will make iterating/tuning with IVF a lot easier. We will need to read model definition from cluster state, but @ryanbogan should be able to help provide some insights on that.

@jmazanec15 I would want to keep it separate - have a separate PR and preferably a separate issue so we do our due diligence. Lets keep this PR and feature branch concentrated on efSearch

**Comment by navneet1v** [comment]
> > @shatejas Can we also add support for dynamic nprobes for IVF? This will make iterating/tuning with IVF a lot easier. We will need to read model definition from cluster state, but @ryanbogan should be able to help provide some insights on that.
> 
> @jmazanec15 I would want to keep it separate - have a separate PR and preferably a separate issue so we do our due diligence. Lets keep this PR and feature branch concentrated on efSearch

We can have a separate PR but I would like to ensure that we release these dynamic parameters supports other algos too and not only just HNSW. From this PR stand point can we ensure that the interface is open enough that it can take other algorithm related parameters.

**Comment by shatejas** [comment]
> > > @shatejas Can we also add support for dynamic nprobes for IVF? This will make iterating/tuning with IVF a lot easier. We will need to read model definition from cluster state, but @ryanbogan should be able to help provide some insights on that.
> > 
> > 
> > @jmazanec15 I would want to keep it separate - have a separate PR and preferably a separate issue so we do our due diligence. Lets keep this PR and feature branch concentrated on efSearch
> 
> We can have a separate PR but I would like to ensure that we release these dynamic parameters supports other algos too and not only just HNSW. From this PR stand point can we ensure that the interface is open enough that it can take other algorithm related parameters.

> I am currently trying to see how the structure of the code can be modified to fit this in a way that is open to extension. If I find a way that is simple enough you will see it in subsequent commits, if its a major refactor I will have a discussion

File: jni/src/faiss_wrapper.cpp [hunk 14]
```diff
@@ -409,6 +412,22 @@ jobjectArray knn_jni::faiss_wrapper::QueryIndex_WithFilter(knn_jni::JNIUtilInter
     return results;
 }
 
+int getQueryEfSearch(JNIEnv * env, knn_jni::JNIUtilInterface * jniUtil, jobject algoParams, int defaultEfSearch) {
+    if (algoParams == nullptr) {
+        return defaultEfSearch;
+    }
+    std::string hnswParameterJavaFilePath = "org/opensearch/knn/index/query/model/HNSWAlgoQueryParameters";
+    if (jniUtil->IsInstanceOf(env, algoParams, hnswParameterJavaFilePath)) {
```
jmazanec15, line 420: Is there any way to avoid multiple java calls?

jni/tests/faiss_wrapper_unit_test.cpp
```diff
@@ -0,0 +1,275 @@
...
+    };
+
+protected:
+    MockIndex index_;
+    MockIdMap id_map_;
+};
+
+namespace query_index_test {
+
+    TEST_P(FaissWrappeterParametrizedTestFixture, QueryIndexHNSWTests) {
```
jmazanec15, line 95: What is TEST_P?

jni/src/faiss_wrapper.cpp
shatejas: Not that I know of. We will have to go back to using primitive types on JNI layer by unpacking it in the java layer. How much is the overhead comparatively considering we are caching classes on initialization?

jni/tests/faiss_wrapper_unit_test.cpp
shatejas: [Macro for parametrized tests](https://google.github.io/googletest/reference/testing.html#TEST_P)

They are much more maintainable since it helps cover extensive use cases without copy pasta code. Especially when there is heavy branching

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
```diff
@@ -225,11 +323,21 @@ public KNNQueryBuilder(StreamInput in) throws IOException {
             if (isClusterOnOrAfterMinRequiredVersion(KNNConstants.RADIAL_SEARCH_KEY)) {
                 minScore = in.readOptionalFloat();
             }
+            streamInHnswParameters(in);
```
jmazanec15, line 326: How will this work for ivf/nprobes?

jni/src/faiss_wrapper.cpp
jmazanec15: > How much is the overhead comparatively considering we are caching classes on initialization?

From performance, Im not sure. But we are
1. Checking if instance type
2. Getting Optional

Why not let java pass a map to the JNI with the parameters?

jmazanec15 (commented):
Cant we do something similar to https://github.com/opensearch-project/k-NN/pull/1713 and delegate the parsing of search parameters to the engine/method combo?

jni/tests/faiss_wrapper_unit_test.cpp
jmazanec15: ah nice

jni/src/faiss_wrapper.cpp
shatejas: If the concern is not performance, whats the benefit of the map when we can have the entire object? Whats the concern with using an object

The map becomes a string to string property bag which will force us to validate again on the JNI layer and there is always a possibility of unknown parameters being added on accident. A typed value is seems  a better approach than passing around a generic map

src/main/java/org/opensearch/knn/index/query/parser/MethodParametersParser.java
```diff
@@ -0,0 +1,100 @@
...
+import static org.opensearch.knn.index.query.KNNQueryBuilder.EF_SEARCH_FIELD;
+import static org.opensearch.knn.index.query.KNNQueryBuilder.NAME;
+import static org.opensearch.knn.index.query.parser.KNNXParserUtil.parseJsonObject;
+
+@Getter
+public class MethodParametersParser {
+
+    private static final Set<String> VALID_METHOD_PARAMETERS = ImmutableSet.of(METHOD_PARAMETER_EF_SEARCH);
+
+    @SneakyThrows(IOException.class)
```
navneet1v, line 41: We don't add SneakyThrows in src code. Better to remove these annotations. You can use it in UTs and ITs.

src/main/java/org/opensearch/knn/index/query/parser/MethodParametersParser.java
```diff
@@ -0,0 +1,100 @@
...
+    public static AlgoQueryParameters streamInMethodParameters(StreamInput in) {
```
navneet1v, line 42: Here is an interesting choice of naming, we are calling function as streamInMethodParameters and rather then returning method parameters it is returning a class named as AlgoQueryParameters. We should be consistent in terms of what parameters these are. Are they algo related parameters or they are method related parameters. If you think algo and method is same thing, so use consistent naming.

src/main/java/org/opensearch/knn/index/query/parser/MethodParametersParser.java
```diff
@@ -0,0 +1,100 @@
...
+            throw new IllegalArgumentException("[" + NAME + "] method_parameter cannot be empty");
+        }
+
+        for (String jsonkey : parameters.keySet()) {
+            if (!VALID_METHOD_PARAMETERS.contains(jsonkey)) {
+                throw new IllegalArgumentException("[" + NAME + "] unknown parameter " + jsonkey + " found.");
+            }
+        }
+
+        return Optional.ofNullable((Integer) parameters.get(METHOD_PARAMETER_EF_SEARCH))
```
navneet1v, line 79: This good for 1 single attribute. But we might have to add more validations once we have more attributes that are getting parsed. I would recommend adding a todo here to handle that condition.

src/main/java/org/opensearch/knn/index/query/parser/MethodParametersParser.java
```diff
@@ -0,0 +1,100 @@
...
+            .filter(ef -> EF_SEARCH_FIELD.match(METHOD_PARAMETER_EF_SEARCH, parser.getDeprecationHandler()))
```
navneet1v, line 80: should we have some checks to ensure that ef_Search is a +ve value. my worry is if some once gives a value which is out of int range we typecasting it to int will lead to negative value. or someone my pass -ve values. So lets ensure we are throwing right exceptions in those cases.

src/main/java/org/opensearch/knn/index/query/model/AlgoQueryParameters.java
```diff
@@ -0,0 +1,18 @@
...
+ * GitHub history for details.
+ */
+
+package org.opensearch.knn.index.query.model;
+
+/**
+ * Functional Interface for algorithm specific parameters.
+ * These parameters are passed to the algorithm implementation while processing a query
+ */
+public interface AlgoQueryParameters {}
```
navneet1v, line 18: Should this be an interface or abstract class? I see this as an abstract class rather than interface.

jni/include/org_opensearch_knn_jni_FaissService.h
```diff
@@ -72,15 +72,15 @@ JNIEXPORT void JNICALL Java_org_opensearch_knn_jni_FaissService_setSharedIndexSt
  * Signature: (J[FI[I)[Lorg/opensearch/knn/index/query/KNNQueryResult;
  */
 JNIEXPORT jobjectArray JNICALL Java_org_opensearch_knn_jni_FaissService_queryIndex
-  (JNIEnv *, jclass, jlong, jfloatArray, jint, jintArray);
+  (JNIEnv *, jclass, jlong, jfloatArray, jint, jobject, jintArray);
```
navneet1v, line 75: as you are changing the signature of the function change this too.

```
Signature: (J[FI[I)[Lorg/opensearch/knn/index/query/KNNQueryResult;
```

.github/workflows/backwards_compatibility_tests_workflow.yml
```diff
@@ -15,7 +15,7 @@ jobs:
       matrix:
         java: [ 11, 17 ]
         os: [ubuntu-latest]
-        bwc_version : [ "2.0.1", "2.1.0", "2.2.1", "2.3.0", "2.4.1", "2.5.0", "2.6.0", "2.7.0", "2.8.0", "2.9.0", "2.10.0", "2.11.0", "2.12.0", "2.13.0", "2.14.0-SNAPSHOT"]
+        bwc_version : [ "2.0.1", "2.1.0", "2.2.1", "2.3.0", "2.4.1", "2.5.0", "2.6.0", "2.7.0", "2.8.0", "2.9.0", "2.10.0", "2.11.0", "2.12.0", "2.13.0", "2.14.0", "2.15.0-SNAPSHOT"]
```
navneet1v, line 18: this could have been added in a different PR. I have raised a PR for this: https://github.com/opensearch-project/k-NN/pull/1724

navneet1v (changes requested):
Completed my first pass on the src files. There are some good abstractions that have been added. I am seeing intensive use of streams and optionals. I personally feel we can avoid those things as search path of k-NN is very sensitive and these smaller abstractions of optional and stream adds latencies when called enough number of times doesn't matter from where they get called from.

navneet1v:
> For separation of responsibilites, I meant delegate handling of parsing to the method in the engine. For instance, do something like `Faiss.parseSearchParameters(KNNMethodContext, ThingToParse) -> return SearchParameters` and pass it down the chain of abstractions.

I haven't read the whole conversation. But I kind of agree on this part only, its good that we are building an algorithm specific abstraction. But once we have that abstraction we can pass it through 1 more level of check from Engines to see if they don't support a specific parameters they can fail the search call.

Now, should we allow Engines to parse the parameters I would say no. We should give them a parsed object and let them validate it. This will ensure that k-NN plugin controls the interface and engines just do the validation, rather than just giving engines control to define the interface.

src/test/java/org/opensearch/knn/index/FaissHNSWFlatE2EIT.java
```diff
@@ -0,0 +1,188 @@
...
+import java.util.Map;
+import java.util.Set;
+import java.util.TreeMap;
+
+import static com.carrotsearch.randomizedtesting.RandomizedTest.$;
+import static com.carrotsearch.randomizedtesting.RandomizedTest.$$;
+import static org.opensearch.knn.common.KNNConstants.*;
+
+@AllArgsConstructor
+public class FaissHNSWFlatE2EIT extends KNNRestTestCase {
```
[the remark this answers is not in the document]
shatejas: To make it parametrized. I really hate copy pasting tests - one change can potentially fail many tests and as side effect we need to change all the tests.

Since the plugin is not on junit 5 combined with the use of carrot search we can't use the same class to have parametrized tests. we need a separate class per test but we can add multiple test cases in there

src/main/java/org/opensearch/knn/index/query/parser/MethodParametersParser.java
shatejas: So the thought process was to make sure nothing other than algoquery parameters are there in this. From method parameters discussion things like overscorer can potential fit there but they shouldn't be clubbed and be added under algo params. The naming was chosen keeping in mind that. If something else get added in method params we should think about where to put it rather than using algo params as default. I can take a look at the structure again to see if I can make it more obvious

src/main/java/org/opensearch/knn/index/query/parser/MethodParametersParser.java
shatejas: Yeah the intention is to build on this when nProbs is added

src/main/java/org/opensearch/knn/index/query/parser/MethodParametersParser.java
shatejas: > should we have some checks to ensure that ef_Search is a +ve value

+ve check is present [here](https://github.com/opensearch-project/k-NN/pull/1707/files#diff-6d0df17543181185170f65c116cdb3514d910e1557f012b7959611ae92678080R222). Overall I would want to decouple parsing and validations moving forward and I have tried to reflect that with ef_search

jni/include/org_opensearch_knn_jni_FaissService.h
shatejas: Missed it. will do

src/main/java/org/opensearch/knn/index/query/model/AlgoQueryParameters.java
shatejas: Abstract classes open up possibility of abusing it for code reuse. Inheritance should only be used to enforce behavior, Composition should be used for code reuse. So default is interface. Abstract classes have limited use cases and they should be used sparingly.

Functional interfaces like these are generally used to have bounds on the type of objects passed and that helps not have type `Object` in the function signature which can be easily abused.

src/main/java/org/opensearch/knn/index/query/parser/MethodParametersParser.java
shatejas: Even with specific exceptions?....will remove

shatejas:
> Completed my first pass on the src files. There are some good abstractions that have been added. I am seeing intensive use of streams and optionals. I personally feel we can avoid those things as search path of k-NN is very sensitive and these smaller abstractions of optional and stream adds latencies when called enough number of times doesn't matter from where they get called from.

streams I have tried to avoid might have missed it, will take another pass. Optionals are a surprise honestly as they are pretty light weight. I like them because they help avoid branching. Will remove them

.github/workflows/backwards_compatibility_tests_workflow.yml
shatejas: This will be removed when I sync my branch

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
```diff
@@ -431,10 +498,34 @@ protected Query doToQuery(QueryShardContext context) {
...
+                    String.format(
+                        "Parameters not valid for [%s]:[%s] combination: [%s]",
+                        knnEngine,
+                        method,
+                        validationException.getMessage()
+                    )
+                );
+            }
+        } else {
+            throw new IllegalStateException("method not found");
```
shatejas, line 528: Need confirmation that this is the right behavior and there will never be a scenario which is valid when there is no method found

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
```diff
@@ -431,10 +498,34 @@ protected Query doToQuery(QueryShardContext context) {
             fieldDimension = modelMetadata.getDimension();
             knnEngine = modelMetadata.getKnnEngine();
             spaceType = modelMetadata.getSpaceType();
+            if (modelMetadata.getMethodComponentContext() != null) {
```
shatejas, line 501: Need inputs on whether this check is needed

ryanbogan: We added the method component context to `ModelMetadata` in 2.13.  Therefore, there could be a situation where a model is created before 2.13, the cluster is updated to 2.13 or later, and then that method will return null.

src/test/java/org/opensearch/knn/index/FaissHNSWFlatE2EIT.java
```diff
@@ -0,0 +1,189 @@
...
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.TreeMap;
+
+import static com.carrotsearch.randomizedtesting.RandomizedTest.$;
+import static com.carrotsearch.randomizedtesting.RandomizedTest.$$;
+import static org.opensearch.knn.common.KNNConstants.*;
```
ryanbogan, line 44: Can we list out the imports instead of using *?  There are a few other cases of this later in the PR as well

src/main/java/org/opensearch/knn/engine/method/EngineSpecificMethodContext.java
```diff
@@ -0,0 +1,31 @@
...
+
+package org.opensearch.knn.engine.method;
+
+import org.opensearch.knn.index.Parameter;
+
+import java.util.Collections;
+import java.util.Map;
+
+/**
+ * Holds context related to a method for a particular engine
```
jmazanec15, line 20: Currently, this information is stored in a [map](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/util/Faiss.java#L213) in the engines. Is the idea to migrate those maps to this abstraction?

src/main/java/org/opensearch/knn/engine/method/EngineSpecificMethodContext.java
```diff
@@ -0,0 +1,31 @@
...
+ * Each engine can have a specific set of parameters that it supports during index and build time. This context holds
+ * the information for each engine method combination.
+ *
+ * TODO: Move KnnMethod in here
+ */
+public interface EngineSpecificMethodContext {
```
jmazanec15, line 26: What do you think about name `MethodImplementationContext`?

src/main/java/org/opensearch/knn/engine/method/HnswContext.java
```diff
@@ -0,0 +1,30 @@
...
+
+package org.opensearch.knn.engine.method;
+
+import com.google.common.collect.ImmutableMap;
+import org.opensearch.knn.index.Parameter;
+import org.opensearch.knn.index.query.request.MethodParameter;
+
+import java.util.Map;
+
+public class HnswContext implements EngineSpecificMethodContext {
```
jmazanec15, line 20: So is idea for engine to implement HnswCOntext? Im not sure I understand how HnswContext is EngineSpecificMethodContext.

src/main/java/org/opensearch/knn/index/query/BaseQueryFactory.java
```diff
@@ -42,6 +43,7 @@ public static class CreateQueryRequest {
         private float[] vector;
         private byte[] byteVector;
         private VectorDataType vectorDataType;
+        private Map<String, ?> methodParameters;
```
jmazanec15, line 46: Why use wild card instead of Object?

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
shatejas: nevermind will have to change it because of an IT

src/main/java/org/opensearch/knn/engine/method/HnswContext.java
shatejas: > So is idea for engine to implement HnswCOntext

Not really. See [Faiss](https://github.com/opensearch-project/k-NN/pull/1707/files#diff-019b2da767e6ac2fa27b70e74d14a35936c7546562aeecd3fac7703142bd4405R338) and [doToQuery](https://github.com/opensearch-project/k-NN/pull/1707/files#diff-6d0df17543181185170f65c116cdb3514d910e1557f012b7959611ae92678080R512) validation for usage.

> Im not sure I understand how HnswContext is EngineSpecificMethodContext.

I understand the confusion. I had initially kept this FaissHnswContext. But then thinking about nmslib/lucene, considering there is no difference currently I wanted a way to reuse and let it have engine specific stuff when required. I have a hunch we will have to split it once we try and add KNNMethod in here. I will add a comment and change it to DefaultHnswContext. Thoughts?

src/main/java/org/opensearch/knn/engine/method/EngineSpecificMethodContext.java
shatejas: I really like having "engine/nativelibrary" and "method" both in there to clearly indicate that it shouldn't be used stand alone. Any other suggestions?

src/main/java/org/opensearch/knn/index/query/BaseQueryFactory.java
shatejas: The intent is to indicate that it contains multiple types and that it is already parsed. It's subtle and wouldn't have been my first choice but I still think its better than object. Since this will be object on compile time I don't see any overhead here unless I am missing something.

My initial though process was to store the type in `MethodParameter` enum  so we can cast it whenever we want but as of now there was no need since the responsibility for Faiss falls onto jni layer.

src/test/java/org/opensearch/knn/index/FaissHNSWFlatE2EIT.java
shatejas: For some reason intellij keeps doing it, I will try and find a way to have intellij not do it

src/main/java/org/opensearch/knn/engine/method/EngineSpecificMethodContext.java
jmazanec15: > I really like having "engine/nativelibrary" and "method" both in there to clearly indicate that it shouldn't be used stand alone ideally

What do you mean by stand alone here?

shatejas: For instance, I can see someone using it to use this to get HNSW method parameters in universal fashion and plug it somewhere other than engine.

src/main/java/org/opensearch/knn/index/query/parser/KNNXParserUtil.java
```diff
@@ -0,0 +1,47 @@
...
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.opensearch.core.xcontent.XContentParserUtils.parseFieldsValue;
+import static org.opensearch.knn.index.query.KNNQueryBuilder.NAME;
+
+public final class KNNXParserUtil {
+
+    public static Map<String, Object> parseJsonObject(XContentParser parser) throws IOException {
```
jmazanec15, line 26: Do we need this if XContentParser has map() method? https://github.com/opensearch-project/OpenSearch/blob/main/libs/core/src/main/java/org/opensearch/core/xcontent/XContentParser.java#L166

jmazanec15 (commented):
From a functionality perspective, I think it looks good.  I want to understand a little bit more on where method abstraction refactor is going - it seems partially complete in this PR (which is okay for feature branch). I just want to understand what it will look like in the end. A couple questions:
1. How will we eventually define supported methods for each engine? Is the plan to move away from current mechanism to new paths.
2. Will we be moving everything in index.util package to the new engine package?

src/main/java/org/opensearch/knn/index/query/parser/KNNXParserUtil.java
shatejas: Wasn't aware of the map, it pretty much does the same thing except the initial start object check. I will switch to that and delete this file

shatejas:
> 1. How will we eventually define supported methods for each engine? Is the plan to move away from current mechanism to new paths.

The KNNengine enum which defines the instance is fine, The native library has deep hierarchies which can turn problematic but for now I wouldn't want to remove it. What I would like to see us move away from is giant static maps that are defined in each engine class. Those aren't maintainable and hard to follow. I am trying to structure it in a way that these are more manageable at the same time easy to use. I will give psuedocode for FAISS maybe on how the final structure would look like
```java
class FAISS extends NativeLibrary {

     // singleton
     Map<String, EngineSpecificeMethodContext> methodContext = Map.of("hnsw", new FaissHnswContext(), "ivf", new FaissIVFContext());

//other engine specific stuff
}
```
FaissHnswContext.java
```java

class FaissHnswContext implements EngineSpecificMethodContext {

      //singleton
      KNNMethod supportedMethodComponents();
      Map<String, Parameter> supportedMethodParameters();
      //anything that is unique to engine and method specific that comes around maybe supportedQuantization() ? I know its encapsulated in knnmethod but just to give you an idea

}

```

ofcourse for reusability we can have something like DefaultHNSWContext. Again this is not set in stone and have tried to structure with the limited context I have. I am open to suggestions here.


> 2. Will we be moving everything in index.util package to the new engine package?

The package structure isn't intuitive right now and sure if it makes more sense we can move it

jmazanec15:
> What I would like to see us move away from is giant static maps that are defined in each engine class.

+1 want to remove this. I would like to move these to classes eventually.

> The package structure isn't intuitive right now and sure if it makes more sense we can move it

Right, this is overdue for refactor.

I think the long term potential goal is to encapsulate this information in a strong interface so that backend method/implementation can be easily extended by a third-party.

src/test/java/org/opensearch/knn/index/FaissHNSWFlatE2EIT.java
shatejas: made sure there are no * imports

jmazanec15 (approved):

src/main/java/org/opensearch/knn/index/query/parser/KNNXParserUtil.java
shatejas: Making this a part of another PR

src/test/java/org/opensearch/knn/index/FaissHNSWFlatE2EIT.java
ryanbogan: The same thing happens to me, it automatically changes anything with I believe 4 or more imports of the same class to a *

Issue #1075, opened 2023-08-30
Add validation for PQ based methods based on dimension


jmazanec15:
Related to https://github.com/opensearch-project/k-NN/pull/1074#discussion_r1309687747, right now, when validating a method to be used for training for PQ, we do not validate that the field's dimension is in fact divisble by the `M` parameter.

PQ encodes vectors by breaking them into `M` sub-vectors and then encoding each sub-vector independently. The behavior for the user will be that they get a vague excpetion. Something like: https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/training/TrainingJob.java#L196.

Instead, we should give a more descriptive error message that says exactly this. It will require some refactoring to https://github.com/opensearch-project/k-NN/blob/2.9.0.0/src/main/java/org/opensearch/knn/index/util/Faiss.java#L95

PR #1713, opened 2024-05-21
Add validation for pq m parameter before training starts

ryanbogan:
### Description
Currently, if the dimension of a model is not divisible by the pq m parameter (code count) provided in the training request, training fails with a detailed message in the logs.  In addition, the model state is update to failed and the model error is set to `"Failed to execute training. May be caused by an invalid method definition or not enough memory to perform training."`.

This PR adds a validation check for the above case before training starts or any memory is allocated.  The model state is still set to failed, but the error message is more specific.
 
### Issues Resolved
https://github.com/opensearch-project/k-NN/issues/1075
 
### Check List
  - [x] All tests pass


src/main/java/org/opensearch/knn/training/TrainingJob.java
```diff
@@ -115,6 +116,18 @@ public void run() {
         NativeMemoryAllocation modelAnonymousAllocation = null;
         ModelMetadata modelMetadata = model.getModelMetadata();
 
+        Map<String, Object> parameters = modelMetadata.getMethodComponentContext().getParameters();
```
jmazanec15, line 119: Lets do validation earlier - i.e. when building the KnnMethodContext? https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/plugin/rest/RestTrainModelHandler.java#L99

jmazanec15: I think we are going to want to change this structure here: https://github.com/opensearch-project/k-NN/blob/2.9.0.0/src/main/java/org/opensearch/knn/index/util/Faiss.java#L95

ryanbogan: I could do it a few lines lower in that file around https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/plugin/rest/RestTrainModelHandler.java#L119 , since we need the dimension and the method context to be set.  I think in that scenario, the model would never enter train state, so would we just throw an exception at that point?

ryanbogan: > I think we are going to want to change this structure here: https://github.com/opensearch-project/k-NN/blob/2.9.0.0/src/main/java/org/opensearch/knn/index/util/Faiss.java#L95

That's more of a limit check that occurs while building the KNNMethodContext right?  Why do we want to change that logic?

jmazanec15: Because that is where we already do parameter validation. Keeping it in one place would be ideal.

ryanbogan: I looked into that first, but we would have to do something like https://github.com/opensearch-project/k-NN/blob/2.9.0.0/src/main/java/org/opensearch/knn/index/util/Faiss.java#L106 where we create a series of methods and pass the dimension all the way down.  That's why I thought it would be better to validate closer to the top for this specifically since we have easy access to the dimension there.

src/main/java/org/opensearch/knn/common/KNNConstants.java
```diff
@@ -68,6 +68,8 @@ public class KNNConstants {
     public static final VectorDataType DEFAULT_VECTOR_DATA_TYPE_FIELD = VectorDataType.FLOAT;
 
     public static final String RADIAL_SEARCH_KEY = "radial_search";
+    public static final String INVALID_CODE_COUNT_ERROR_MESSAGE =
```
junqiu-lei, line 71: We can add PQ word in the constant to be clear, like `INVALID_PQ_CODE_COUNT_ERROR_MESSAGE`.

src/main/java/org/opensearch/knn/training/TrainingJob.java
jmazanec15: Right, but this checking and type casting is not extendable. We need to come up with a more maintainable solution. 

I think one possible solution could be to add a new chain of validation that validates a given KnnMethodContext against a given dimension: i.e. KnnMethodContext.validateForDimension(int dimension). In addition, we may also want to check if a given KnnMethodContext works for a particular data_type (https://opensearch.org/docs/latest/field-types/supported-field-types/knn-vector/#lucene-byte-vector). 

Maybe we could generalize this into KnnMethodContext.validateForDataSpec(DataSpec dataSpec) and then DataSpec is an object that contains information about the data that isnt included in the method definition. Then, we can call this from the field mapper or from the rest training api.

ryanbogan: Ah that's true.  Wouldn't validating a KnnMethodContext still require casting to check the parameters since the object would already be built?  I think the only way to avoid casting would be to extend the validation in the Faiss class that exists as the object is built, but I haven't been able to find a way to get the dimension there without modifying the addParameter method.

jmazanec15: I think we need to modify addParameter to contain one more validation that takes a DataSpec in. Along with this, we could easily provide validation on the amount of training data configured as well (Im thinking for future)

src/main/java/org/opensearch/knn/index/KNNMethod.java
```diff
@@ -77,6 +78,35 @@ public ValidationException validate(KNNMethodContext knnMethodContext) {
         return validationException;
     }
 
+    public ValidationException validateWithData(KNNMethodContext knnMethodContext, TrainingDataSpec trainingDataSpec) {
+        List<String> errorMessages = new ArrayList<>();
+        if (!containsSpace(knnMethodContext.getSpaceType())) {
```
VijayanB, line 83: nit: i though you are actually checking for space in string, isSpaceTypeSupported() might be good for readability ( FAR)

src/main/java/org/opensearch/knn/index/KNNMethod.java
```diff
@@ -77,6 +78,35 @@ public ValidationException validate(KNNMethodContext knnMethodContext) {
         return validationException;
     }
 
+    public ValidationException validateWithData(KNNMethodContext knnMethodContext, TrainingDataSpec trainingDataSpec) {
```
VijayanB, line 81: can we add documentation for public methods?

src/main/java/org/opensearch/knn/index/KNNMethod.java
```diff
@@ -77,6 +78,35 @@ public ValidationException validate(KNNMethodContext knnMethodContext) {
...
+            errorMessages.add(
+                String.format(
+                    "\"%s\" configuration does not support space type: " + "\"%s\".",
+                    this.methodComponent.getName(),
+                    knnMethodContext.getSpaceType().getValue()
+                )
+            );
+        }
+
+        ValidationException methodValidation = methodComponent.validateWithData(
```
VijayanB, line 93: do we have to validate further even if we don't support space type?

src/main/java/org/opensearch/knn/index/KNNMethod.java
ryanbogan: Good call, I meant to do that at the end and forgot about it

src/main/java/org/opensearch/knn/index/KNNMethod.java
ryanbogan: Even though validation will ultimately fail if space type is not supported, more error messages can be added to the errors based on any potential problems with the method component context

src/main/java/org/opensearch/knn/index/Parameter.java
```diff
@@ -89,6 +111,27 @@ public ValidationException validate(Object value) {
             }
             return validationException;
         }
+
+        @Override
+        public ValidationException validateWithData(Object value, TrainingDataSpec trainingDataSpec) {
+            ValidationException validationException = null;
+            if (!(value instanceof Boolean)) {
+                validationException = new ValidationException();
+                validationException.addValidationError(String.format("value not of type Boolean for Boolean parameter [%s].", getName()));
```
VijayanB, line 120:
```suggestion
                validationException.addValidationError(String.format("value is not an instance of Boolean for Boolean parameter [%s].", getName()));
```

src/main/java/org/opensearch/knn/index/Parameter.java
```diff
@@ -89,6 +111,27 @@ public ValidationException validate(Object value) {
...
+                return validationException;
+            }
+
+            if (validatorWithData == null) {
+                return validationException;
```
VijayanB, line 125:
```suggestion
                return null;
```

src/main/java/org/opensearch/knn/index/Parameter.java
```diff
@@ -118,6 +170,27 @@ public ValidationException validate(Object value) {
             }
             return validationException;
         }
+
+        @Override
+        public ValidationException validateWithData(Object value, TrainingDataSpec trainingDataSpec) {
+            ValidationException validationException = null;
+            if (!(value instanceof Integer)) {
+                validationException = new ValidationException();
+                validationException.addValidationError(String.format("value not of type Integer for Integer parameter [%s].", getName()));
```
VijayanB, line 179: same as above

src/test/java/org/opensearch/knn/index/FaissIT.java
```diff
@@ -1686,17 +1684,63 @@ public void testInvalidPQM_thenFail() {
...
+        // which is 2^8 = 256.
+        int trainingDataCount = 256;
+
+        int dimension = testData.indexData.vectors[0].length;
+        SpaceType spaceType = SpaceType.L2;
+        int ivfNlist = 4;
+        int ivfNprobes = 4;
+        int pqCodeSize = 8;
+
+        XContentBuilder xContentBuilder = XContentFactory.jsonBuilder()
```
VijayanB, line 1717: can you add json structure for below input? This will help to understand the structure. Thanks

src/main/java/org/opensearch/knn/training/TrainingDataSpec.java
```diff
@@ -0,0 +1,25 @@
...
+
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+
+@Getter
+@AllArgsConstructor
+public class TrainingDataSpec {
+    private int dimension;
+
+    public void setDimension(int dimension) {
```
VijayanB, line 22: nit: You can add @Setter instead of adding your own setter

src/main/java/org/opensearch/knn/index/util/Faiss.java
```diff
@@ -109,9 +109,6 @@ class Faiss extends NativeLibrary {
             .build()
     );
 
-    // TODO: To think about in future: for PQ, if dimension is not divisible by code count, PQ will fail. Right now,
```
jmazanec15, line 112: nice - finally getting rid of it

src/main/java/org/opensearch/knn/training/TrainingDataSpec.java
```diff
@@ -0,0 +1,26 @@
...
+import lombok.Getter;
+import lombok.Setter;
+
+/**
+ * A data spec containing relevant training data for validation.
+ */
+@Getter
+@Setter
+@AllArgsConstructor
+public class TrainingDataSpec {
```
jmazanec15, line 24: Can we change this to DataSetInfo or VectorSpaceInfo? I think we might want to use this for non-training data as well. In the future, I could see it having the following properties:
1. dimension
2. vector count (either training or index)
3. space type

ryanbogan: Sure, I think I like the second one better

jmazanec15 (approved):

VijayanB (approved):

Issue #1343, opened 2023-12-11
Block delete model requests if an index uses the model


jmazanec15:
## Description

Right now, our delete model functionality does not check if an index is currently using the model. This can lead users to get themselves into trouble if they delete a model that is in use by an index in the cluster. Instead, we should block model deletion and tell the user they first need to delete the backing indices.

PR #1722, opened 2024-05-30
Block delete model requests if an index uses the model

ryanbogan:
### Description
Adds a step in the deleteModel method within ModelDao to check if a model is in use by any indices.  If so, the delete request will fail with a new DeleteModelWhenInUseException containing the model name and the index using it.
 
### Issues Resolved
https://github.com/opensearch-project/k-NN/issues/1343
 
### Check List
  - [x] All tests pass


src/main/java/org/opensearch/knn/indices/ModelDao.java
```diff
@@ -531,6 +537,37 @@ public void delete(String modelId, ActionListener<DeleteModelResponse> listener)
                     return;
                 }
 
+                // Get index mappings to check if model is being used by any indices
+                getMappings(modelId, getMappingsStep);
+            }, listener::onFailure);
+
+            getMappingsStep.whenComplete(getMappingsResponse -> {
+                // The mapping metadata string will always have the model id followed by `,` and another parameter, or the end of the object
```
ryanbogan, line 545: An alternative solution would be to parse the map through casting and null checks, then check if the model_id field exists and then if it's equal to the modelId being deleted.  I chose to go with contains, since there are only two possible characters that can follow the id since the mapping string is in json format.  The other reason is that casting could get complicated with multiple fields beneath properties and would end up with a lot of nested loops.

src/main/java/org/opensearch/knn/indices/ModelDao.java
```diff
@@ -61,6 +65,7 @@
 import org.opensearch.knn.plugin.transport.UpdateModelMetadataAction;
 import org.opensearch.knn.plugin.transport.UpdateModelMetadataRequest;
 
+import javax.swing.*;
```
naveentatikonda, line 68: Pls avoid wildcard characters

src/main/java/org/opensearch/knn/indices/ModelDao.java
shatejas: can you try to use regex with a capturing group here something like `.*modelId=(.*)(?=\s|\t|,|})` and then reference the capturing group to get the id itself?  That would be less code than tokenizing everything

Regex will be safer here if there is a space or tab in between on accident considering its a json

shatejas (commented):
There doesn't seem to be any unit test updates. Does it need one? or are ModelTests unit tests?

ryanbogan:
> There doesn't seem to be any unit test updates. Does it need one? or are ModelTests unit tests?

It's a singleton node test, which is kind of halfway between a unit test and an integration test.  It's the only file we have other than bwc at the moment for testing models.

src/main/java/org/opensearch/knn/indices/ModelDao.java
```diff
@@ -531,6 +537,37 @@ public void delete(String modelId, ActionListener<DeleteModelResponse> listener)
...
+                getMappings(modelId, getMappingsStep);
+            }, listener::onFailure);
+
+            getMappingsStep.whenComplete(getMappingsResponse -> {
+                // The mapping metadata string will always have the model id followed by `,` and another parameter, or the end of the object
+                // `}`
+                // This check is to prevent an incorrect thrown exception when one model starts with the another model's id
+                // Ex: test-model and test-model1
+                String modelIdStringInMapping1 = String.format("model_id=%s,", modelId);
+                String modelIdStringInMapping2 = String.format("model_id=%s}", modelId);
```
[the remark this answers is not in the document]
ryanbogan: @naveentatikonda the difference is that it would be searching if the string contains `test-model,` or `test-model}` instead of just `test-model`.  I'll attempt the casting solution today.

src/main/java/org/opensearch/knn/indices/ModelDao.java
ryanbogan: That's true, I'm going to attempt the casting solution first, but will use this if that is not feasible.

src/main/java/org/opensearch/knn/common/exception/DeleteModelWhenInUseException.java
```diff
@@ -0,0 +1,30 @@
...
+
+package org.opensearch.knn.common.exception;
+
+import org.opensearch.OpenSearchException;
+import org.opensearch.core.common.logging.LoggerMessageFormat;
+
+/**
+ * Exception thrown when a model is deleted while it is in use by an index.
+ */
+public class DeleteModelWhenInUseException extends OpenSearchException {
```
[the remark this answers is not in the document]
ryanbogan: I made it specific to differentiate this exception from the delete model in train state exception here: https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/common/exception/DeleteModelWhenInTrainStateException.java

We could change that exception to be called `DeleteModelException` and then use that exception for both cases, just with different messages.  Thoughts?

src/main/java/org/opensearch/knn/indices/ModelDao.java
```diff
@@ -531,6 +537,37 @@ public void delete(String modelId, ActionListener<DeleteModelResponse> listener)
...
+                Map<String, MappingMetadata> mappings = getMappingsResponse.getMappings();
+                for (Map.Entry<String, MappingMetadata> entry : mappings.entrySet()) {
+                    MappingMetadata mappingMetadata = entry.getValue();
+                    String mappingMetadataString = mappingMetadata.getSourceAsMap().toString();
```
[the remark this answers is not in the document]
ryanbogan: Good catch.  `getMappings()` and `.getSourceAsMap()` use methods that require NonNull parameters, so it looks like the only part that could be null is `entry.getValue()`.  I'll add a check for that variable.

src/main/java/org/opensearch/knn/common/exception/DeleteModelWhenInUseException.java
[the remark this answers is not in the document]
shatejas: > Thoughts?

I would prefer that, the intent here is to strike the right balance. Its good that we are creating custom exception, at the same time we don't want separate exception for each use case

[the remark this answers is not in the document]
navneet1v: Suggestion:
if there is already an exception called as IllegalOperationException we can use that exception.

src/main/java/org/opensearch/knn/indices/ModelDao.java
```diff
@@ -531,6 +538,52 @@ public void delete(String modelId, ActionListener<DeleteModelResponse> listener)
...
                 }
 
+                // Get index mappings to check if model is being used by any indices
+                getMappings(modelId, getMappingsStep);
+            }, listener::onFailure);
+
+            getMappingsStep.whenComplete(getMappingsResponse -> {
+                String modelIdStringInMapping = String.join("model_id=", modelId);
+                Map<String, MappingMetadata> mappings = getMappingsResponse.getMappings();
+                for (Map.Entry<String, MappingMetadata> entry : mappings.entrySet()) {
```
shatejas, line 548: Looking at this code it desperately needs something like `Jackson-databind` as a dependency. Its not worth maintaining and parsing Json like this when there are opensource libraries written for this specific purpose. can you please look for something similar in Opensearch?

shatejas:
> > There doesn't seem to be any unit test updates. Does it need one? or are ModelTests unit tests?
> 
> It's a singleton node test, which is kind of halfway between a unit test and an integration test. It's the only file we have other than bwc at the moment for testing model logic.

I do feel unit and integration test should be there and tests like modelTests which are in a way component tests should be written whenever required. I can't see the coverage right now but please check the coverage considering the parsing of that json is getting complex

src/main/java/org/opensearch/knn/indices/ModelDao.java
ryanbogan: I think guava is better than jackson in terms of vulnerabilities, I can look into using that

ryanbogan:
> > > There doesn't seem to be any unit test updates. Does it need one? or are ModelTests unit tests?
> > 
> > 
> > It's a singleton node test, which is kind of halfway between a unit test and an integration test. It's the only file we have other than bwc at the moment for testing model logic.
> 
> I do feel unit and integration test should be there and tests like modelTests which are in a way component tests should be written whenever required. I can't see the coverage right now but please check the coverage considering the parsing of that json is getting complex

@shatejas I forgot to mention the `ModelDaoTests` class does have unit tests for all the methods, and I did add a unit test there.  We do also have https://github.com/opensearch-project/k-NN/blob/main/src/test/java/org/opensearch/knn/index/FaissIT.java that inherently ensures training works, although it is not specifically focused on `ModelDao`

jmazanec15:
I think this approach will be really expensive for clusters with a lot of indices with large mappings. 

Instead, we have a couple options:
1. We create a separate transport action, similar to [GetMappingTransportAction](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/action/admin/indices/mapping/get/TransportGetMappingsAction.java), that will perform the check and return true or false.
2. We use some kind tracking on index creation of the models in use. Did you consider this?

@heemin32 do you have any ideas?

ryanbogan:
> 2\. We use some kind tracking on index creation of the models in use. Did you consider this?

@jmazanec15 I thought about this initially, but I was worried about potential desynchronization of the indices in use.  To accomplish this solution, we would have to add a variable (either list or set) to the ModelMetadata.  Then, the create index and delete index actions would have to be modified to make an update call to update the model in ModelDao.  I was concerned about multiple indices being created or deleted at the same time, updating the same model, and potentially overwriting an index that was added or removed in that list.  That could theoretically result in a situation where the user has deleted the index, but would be unable to delete the model.

jmazanec15:
I see, then I think a custom transport action like GetMappingsTransportAction makes sense.

heemin32:
> I think this approach will be really expensive for clusters with a lot of indices with large mappings.
> 
> Instead, we have a couple options:
> 
> 1. We create a separate transport action, similar to [GetMappingTransportAction](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/action/admin/indices/mapping/get/TransportGetMappingsAction.java), that will perform the check and return true or false.
> 2. We use some kind tracking on index creation of the models in use. Did you consider this?
> 
> @heemin32 do you have any ideas?

Current implementation might have two concerns.
1. Scalability with large number of indices and big mappings.
2. Race condition. Creating a new index using the model right before the deletion of the model.

The first option won't provide any benefit. The index mapping data is maintained in cluster state metadata and it might already use GetMappingTransportAction behind the seen. For the second option, it could be more problematic when there is a inconsistency between two places where we store index-model mapping: one in new place and another in index mapping.

As 1. delete model won't be called with high frequency and 2. user can delete the index when race condition happen, we can accept those two concerns imo.

src/main/java/org/opensearch/knn/indices/ModelDao.java
```diff
@@ -527,10 +533,56 @@ public void delete(String modelId, ActionListener<DeleteModelResponse> listener)
...
```
heemin32, line 547: Will there be a way to fail fast? like if it is not knn index, skip and move on to the next index?

ryanbogan: Even on a k-NN index, there isn't necessarily a mapping that indicates if it's a k-NN index.
i.e.
```
{
  "mappings": {
    "properties": {
      "target-field": {
        "type": "knn_vector",
        "model_id": "my-model"
      }
    }
  }
}
```

The fastest fail I could think of to add is a contains() check to see if the model id is present in the string version of the mapping metadata before parsing through the entire map

jmazanec15:
> The first option won't provide any benefit. The index mapping data is maintained in cluster state metadata and it might already use GetMappingTransportAction behind the seen.

I dont think this is true. Creating our own transport action would allow us to check if the model is in use without requiring the mappings to be sent over the network. So, I think itd be much more efficient.

src/main/java/org/opensearch/knn/indices/ModelDao.java
jmazanec15: I think we can skip index if index setting "index.knn" is true

ryanbogan: That would require another transport call to get each index and check the settings.  Is that faster than checking the metadata string?

heemin32:
> > The first option won't provide any benefit. The index mapping data is maintained in cluster state metadata and it might already use GetMappingTransportAction behind the seen.
> 
> I dont think this is true. Creating our own transport action would allow us to check if the model is in use without requiring the mappings to be sent over the network. So, I think itd be much more efficient.

Yes. I will save some network bandwidth. To achieve that, instead of creating transport action, we can make the delete API to be handled only by coordinator node. Then, all will be handled locally.

src/main/java/org/opensearch/knn/indices/ModelDao.java
ryanbogan: Also, here is an example settings for a target index I created.  I don't see the index.knn setting:
```
{"index.creation_date":"1717708887058","index.number_of_replicas":"1","index.number_of_shards":"1","index.provided_name":"target-index","index.replication.type":"DOCUMENT","index.uuid":"cXM_TDJnSjO62HeJfGWDew","index.version.created":"137217827"}

```

jmazanec15:
@heemin32  Im not sure I think its worth moving delete api to cluster manager, but I do think we could just add to https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/plugin/transport/UpdateModelGraveyardTransportAction.java instead of creating a new transport action. This action already has access to the cluster metadata. It would also help provide some safety against race conditions. 

@ryanbogan what do you think?

ryanbogan:
I would lean towards creating a new transport action instead of modifying the existing one, since even though they both access the cluster metadata, the actions have very different goals.  I was able to get it working locally with a new transport action, so my vote is to go with that.

jmazanec15:
@ryanbogan Right now, the purpose of model graveyard is to basically prevent a model that is in the process of being deleted from being used in the interim. This might not be the exact usage right now, but I think it is the overall purpose (@naveentatikonda correct me if Im wrong). The update to model graveyard happens first: https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/indices/ModelDao.java#L534-L535.

In other words, as long as a model is in the model graveyard, it cannot be used (if it can, it is a bug). So, if we are to add the check if it is actively being used in the graveyard metadata update, we can provide better consistency. Separating into 2 separate actions can lead to inconsistency.

ryanbogan:
@jmazanec15 so in this case, we would still want the check to happen before it's added to the graveyard so I could add it within the clusterManagerOperation.  I also need to add the indices service to that class and then change the response type to a custom one with two booleans in order to throw the proper exception on failure.

jmazanec15:
Sounds good. Play around with it a little bit, but checking before adding to model graveyard is right high-level approach from consistency perspective. I think we should do it. 

> then change the response type to a custom one with two booleans in order to throw the proper exception on failure.

If the index is in use, it should throw an exception. Why do we need to change the response type?

ryanbogan:
We need to tell the user which index is using the model so they know which one to delete.  The custom response I created for a new transport action had two parameters in the response: a boolean representing if the model is in use, and a string representing the index name using the model.

jmazanec15:
Why cant the indices be in the exception message?

ryanbogan:
Ah right I could just use onFailure with the exception message instead.  I was throwing the exception in ModelDao once the response was returned.

heemin32:
> @heemin32 Im not sure I think its worth moving delete api to cluster manager, but I do think we could just add to https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/plugin/transport/UpdateModelGraveyardTransportAction.java instead of creating a new transport action. This action already has access to the cluster metadata. It would also help provide some safety against race conditions.
> 
> @ryanbogan what do you think?

What I meant by that is, we can just make `DeleteModelTransportAction` to extend `TransportClusterManagerNodeAction` and implement the logic inside `clusterManagerOperation`. I guess we might be talking about same thing though.
In my understanding there is 1:1 mapping between rest api and transport action in general. I cannot quite understand how delete model api can utilize `UpdateModelGraveyardTransportAction` unless we are going to map delete model and update modelgraveyard api to single `UpdateModelGraveyardTransportAction` which I don't think what you meant?

jmazanec15:
> In my understanding there is 1:1 mapping between rest api and transport action in general

End-to-end, a rest API may make multiple transport calls and require multiple transport actions. Think about search for instance.

heemin32:
> > In my understanding there is 1:1 mapping between rest api and transport action in general
> 
> End-to-end, a rest API may make multiple transport calls and require multiple transport actions. Think about search for instance.

Yeah. You are right. Was confused with ActionType to Action mapping. Thanks for correction.
Agree that we can check the usage of model before putting the model into graveyard.

src/main/java/org/opensearch/knn/plugin/transport/UpdateModelGraveyardRequest.java
```diff
@@ -18,7 +18,7 @@
 /**
  * Request for updating model graveyard while processing delete model request
  */
-public class UpdateModelGraveyardRequest extends AcknowledgedRequest<UpdateModelGraveyardRequest> {
+public class UpdateModelGraveyardRequest extends ClusterInfoRequest<UpdateModelGraveyardRequest> {
```
jmazanec15, line 21: why do we need to switch here? Cant we just keep the acked request and throw exception?

src/main/java/org/opensearch/knn/plugin/transport/UpdateModelGraveyardTransportAction.java
```diff
@@ -22,17 +25,16 @@
...
 
 import java.io.IOException;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Objects;
-import java.util.Set;
+import java.util.*;
 
-import static org.opensearch.knn.common.KNNConstants.PLUGIN_NAME;
+import static org.opensearch.knn.common.KNNConstants.*;
```
jmazanec15, line 37: remove wildcard

src/main/java/org/opensearch/knn/plugin/transport/UpdateModelGraveyardTransportAction.java
```diff
@@ -156,4 +168,72 @@ public ClusterTasksResult<UpdateModelGraveyardTask> execute(ClusterState cluster
             return new ClusterTasksResult.Builder<UpdateModelGraveyardTask>().successes(taskList).build(updatedClusterState);
         }
     }
+
+    private static void checkIfIndicesAreUsingModel(ClusterState clusterState, UpdateModelGraveyardTask task, String[] indicesArray)
```
jmazanec15, line 172: can we just make private method? Also, why use an array instead of list for indices?

src/main/java/org/opensearch/knn/plugin/transport/UpdateModelGraveyardTransportAction.java
```diff
@@ -156,4 +168,72 @@ public ClusterTasksResult<UpdateModelGraveyardTask> execute(ClusterState cluster
...
+        throws IOException {
+        Map<String, MappingMetadata> mappings = clusterState.metadata()
```
jmazanec15, line 174: Do we need to do this before we get list of indices which are using knn?

src/main/java/org/opensearch/knn/plugin/transport/UpdateModelGraveyardTransportAction.java
```diff
@@ -156,4 +168,72 @@ public ClusterTasksResult<UpdateModelGraveyardTask> execute(ClusterState cluster
...
+            .findMappings(indicesArray, task.getIndicesService().getFieldFilter());
+        Map<String, IndexMetadata> indices = clusterState.metadata().indices();
+        List<String> indicesUsingModel = new ArrayList<>();
+        // Parse indices and add to list if using the model
+        for (Map.Entry<String, MappingMetadata> entry : mappings.entrySet()) {
+            MappingMetadata mappingMetadata = entry.getValue();
+            Settings indexSettings = indices.get(entry.getKey()).getSettings();
+            if (mappingMetadata != null && Boolean.parseBoolean(indexSettings.get("index.knn", "false"))) {
```
jmazanec15, line 182: Would suggest first getting all of the indices with "index.knn" enabled in one loop and then checking the mapping in another.

src/main/java/org/opensearch/knn/plugin/transport/UpdateModelGraveyardTransportAction.java
```diff
@@ -156,4 +168,72 @@ public ClusterTasksResult<UpdateModelGraveyardTask> execute(ClusterState cluster
...
+                    indicesUsingModel = parseFieldsMap(fieldsMap, task, indicesUsingModel, index);
+                }
+            }
+        }
+        return indicesUsingModel;
+    }
+
+    private static List<String> parseFieldsMap(
+        Map<String, Object> fieldsMap,
+        UpdateModelGraveyardTask task,
```
jmazanec15, line 220: Pass modelId instead of task.

src/main/java/org/opensearch/knn/plugin/transport/UpdateModelGraveyardTransportAction.java
```diff
@@ -156,4 +168,72 @@ public ClusterTasksResult<UpdateModelGraveyardTask> execute(ClusterState cluster
...
+        List<String> indicesUsingModel,
+        String index
```
jmazanec15, line 222: Is this necessary?

src/main/java/org/opensearch/knn/plugin/transport/UpdateModelGraveyardTransportAction.java
ryanbogan: findMappings() later in the class takes a string array as a parameter.

src/main/java/org/opensearch/knn/plugin/transport/UpdateModelGraveyardRequest.java
ryanbogan: I think that was one of the things I was testing to try to get everything to work.  It's not necessary, I'll remove it

src/main/java/org/opensearch/knn/plugin/transport/UpdateModelGraveyardTransportAction.java
ryanbogan: I guess I could return true and add the index higher up, but I use it on line 232

naveentatikonda:
> @ryanbogan Right now, the purpose of model graveyard is to basically prevent a model that is in the process of being deleted from being used in the interim. This might not be the exact usage right now, but I think it is the overall purpose (@naveentatikonda correct me if Im wrong). The update to model graveyard happens first: https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/indices/ModelDao.java#L534-L535.
> 
> In other words, as long as a model is in the model graveyard, it cannot be used (if it can, it is a bug). So, if we are to add the check if it is actively being used in the graveyard metadata update, we can provide better consistency. Separating into 2 separate actions can lead to inconsistency.
> 
> So, I actually think that the best place for the check would be in the model graveyard update location.

Sorry for the late response. @jmazanec15 your explanation is correct

src/main/java/org/opensearch/knn/plugin/transport/UpdateModelGraveyardTransportAction.java
```diff
@@ -146,6 +161,7 @@ public ClusterTasksResult<UpdateModelGraveyardTask> execute(ClusterState cluster
                     modelGraveyard.remove(task.getModelId());
                     continue;
                 }
+                checkIfIndicesAreUsingModel(clusterState, task);
```
jmazanec15, line 164: Can we make this method return true and then we throw the exception in this method?

src/main/java/org/opensearch/knn/plugin/transport/UpdateModelGraveyardTransportAction.java
```diff
@@ -156,4 +172,71 @@ public ClusterTasksResult<UpdateModelGraveyardTask> execute(ClusterState cluster
...
+    private static void checkIfIndicesAreUsingModel(ClusterState clusterState, UpdateModelGraveyardTask task) throws IOException {
+        Map<String, IndexMetadata> indices = clusterState.metadata().indices();
+        List<String> knnIndicesList = new ArrayList<String>();
+        // Get list of knn indices
+        for (Map.Entry<String, IndexMetadata> entry : indices.entrySet()) {
+            Settings indexSettings = indices.get(entry.getKey()).getSettings();
+            if (Boolean.parseBoolean(indexSettings.get("index.knn", "false"))) {
+                knnIndicesList.add(entry.getKey());
+            }
+        }
```
jmazanec15, line 185: Return here if knnIndicesList is empty?

src/main/java/org/opensearch/knn/plugin/transport/UpdateModelGraveyardTransportAction.java
```diff
@@ -156,4 +172,71 @@ public ClusterTasksResult<UpdateModelGraveyardTask> execute(ClusterState cluster
...
+                String.format(
+                    "Cannot delete model [%s].  Model is in use by the following indices %s, which must be deleted first.",
+                    task.getModelId(),
+                    indicesUsingModel
+                )
+            );
+        }
+    }
+
+    private static boolean parseMappingMetadata(MappingMetadata mappingMetadata, UpdateModelGraveyardTask task) {
```
jmazanec15, line 211: rename to signal what the method does?

src/main/java/org/opensearch/knn/plugin/transport/UpdateModelGraveyardTransportAction.java
```diff
@@ -156,4 +172,71 @@ public ClusterTasksResult<UpdateModelGraveyardTask> execute(ClusterState cluster
...
+                if (sourceEntry.getKey() != null && sourceEntry.getKey().equals(PROPERTIES) && sourceEntry.getValue() instanceof Map) {
+                    Map<String, Object> fieldsMap = (Map<String, Object>) sourceEntry.getValue();
+                    return parseFieldsMap(fieldsMap, task.getModelId());
+                }
+            }
+        }
+        return false;
+    }
+
+    private static boolean parseFieldsMap(Map<String, Object> fieldsMap, String modelId) {
```
jmazanec15, line 226: same as above?

src/main/java/org/opensearch/knn/plugin/transport/UpdateModelGraveyardTransportAction.java
jmazanec15: that sounds good

ryanbogan:
Linux tests all passed on my dev desktop

jmazanec15 (dismissed):
lgtm

src/main/java/org/opensearch/knn/plugin/transport/UpdateModelGraveyardTransportAction.java
```diff
@@ -156,4 +182,65 @@ public ClusterTasksResult<UpdateModelGraveyardTask> execute(ClusterState cluster
             return new ClusterTasksResult.Builder<UpdateModelGraveyardTask>().successes(taskList).build(updatedClusterState);
         }
     }
+
+    private static List<String> getIndicesUsingModel(ClusterState clusterState, UpdateModelGraveyardTask task) throws IOException {
+        Map<String, IndexMetadata> indices = clusterState.metadata().indices();
+        List<String> knnIndicesList = new ArrayList<String>();
+        // Get list of knn indices
+        for (Map.Entry<String, IndexMetadata> entry : indices.entrySet()) {
+            Settings indexSettings = indices.get(entry.getKey()).getSettings();
```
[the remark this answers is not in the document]
ryanbogan: Changed the majority of the implementation to follow this style.  Will follow up with the rest in another PR

ryanbogan:
Skipping changelog because release notes for 2.15 are already out.  I update the release notes accordingly

jmazanec15 (approved):

src/main/java/org/opensearch/knn/plugin/transport/UpdateModelGraveyardTransportAction.java
```diff
@@ -155,5 +178,50 @@ public ClusterTasksResult<UpdateModelGraveyardTask> execute(ClusterState cluster
...
+                        return false;
+                    }
+                    Map propertiesMap = (Map<String, Object>) properties;
+                    return propertiesMapContainsModel(propertiesMap, task.getModelId());
+                })
+                .map(Map.Entry::getKey)
+                .collect(toList());
+        }
+
+        private boolean propertiesMapContainsModel(Map<String, Object> propertiesMap, String modelId) {
```
heemin32, line 210: Could you remove this method as well?

heemin32 (approved):
Will add more comments but functionally it is okay as it is.

src/main/java/org/opensearch/knn/plugin/transport/UpdateModelGraveyardTransportAction.java
heemin32:
```
    private static List<String> getIndicesUsingModel(ClusterState clusterState, UpdateModelGraveyardTask task) throws IOException {
        Map<String, IndexMetadata> indices = clusterState.metadata().indices();
        String[] knnIndicesList = indices.values().stream()
            .filter(metadata -> "true".equals(metadata.getSettings().get("index.knn", "false")))
            .map(metadata -> metadata.getIndex().getName())
            .toArray(String[]::new);

        if (knnIndicesList.length == 0) {
            return Collections.emptyList();
        }

        return clusterState.metadata()
            .findMappings(knnIndicesList, task.getIndicesService().getFieldFilter())
            .entrySet()
            .stream()
            .filter(entry -> entry.getValue() != null)
            .filter(entry -> {
                Object properties = entry.getValue().getSourceAsMap().get("properties");
                if (properties == null || properties instanceof Map == false) {
                    return false;
                }
                Map propertiesMap = (Map<String, Object>) properties;
                return propertiesMap.values().stream()
                    .filter(obj -> obj instanceof Map)
                    .filter(obj -> task.modelId.equals(((Map<String, Object>) obj).get(MODEL_ID)))
                    .findAny()
                    .isPresent();
            })
            .map(Map.Entry::getKey)
            .collect(Collectors.toList());
    }
```

src/test/java/org/opensearch/knn/KNNSingleNodeTestCase.java
```diff
@@ -36,13 +37,12 @@
 import org.opensearch.test.hamcrest.OpenSearchAssertions;
 
 import java.io.IOException;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.EnumSet;
-import java.util.Map;
+import java.util.*;
```
heemin32, line 40: Please change the IDE setting so that you don't use `*` in imports.

src/test/java/org/opensearch/knn/KNNSingleNodeTestCase.java
```diff
@@ -181,6 +181,38 @@ protected void addDoc(String index, String docId, String fieldName, String dummy
         assertEquals(response.status(), RestStatus.CREATED);
     }
 
+    /**
+     * Index a new model
+     */
+    protected void addDoc(Model model) throws IOException, ExecutionException, InterruptedException {
```
heemin32, line 187: Can we actually create a model instead of mimicking it by adding model metadata in system index?
Also, this method should not be named as `addDoc`. Should it be like `writeModelToModelSystemIndex`?

src/test/java/org/opensearch/knn/indices/ModelDaoTests.java
```diff
@@ -636,6 +629,91 @@ public void testDelete() throws IOException, InterruptedException {
         assertTrue(inProgressLatch3.await(100, TimeUnit.SECONDS));
     }
 
+    // Test Delete Model when the model is in use by an index
+    public void testDeleteModelInUse() throws IOException, ExecutionException, InterruptedException {
```
heemin32, line 633: You have no logic added to ModelDao. I think you just need to add your test in UpdateModelGraveyardTransactionActionTest.java

src/test/java/org/opensearch/knn/plugin/transport/UpdateModelGraveyardTransportActionTests.java
```diff
@@ -165,4 +188,201 @@ public void testCheckBlock() {
...
+
+        // Attempt to add model id to graveyard when one index is using model, should fail
+        updateModelGraveyardAndAssertDeleteModelException(
+            updateModelGraveyardTransportAction,
+            addModelGraveyardRequest,
+            indicesUsingModel.toString()
+        );
+    }
+
+    public void updateModelGraveyardAndAssertNoError(
```
heemin32, line 339: Let's make public only for the test entry methods.

src/test/java/org/opensearch/knn/plugin/transport/UpdateModelGraveyardTransportActionTests.java
```diff
@@ -165,4 +188,201 @@ public void testCheckBlock() {
...
+                "",
+                "",
+                MethodComponentContext.EMPTY
+            ),
+            modelBlob,
+            modelId
+        );
+
+        // created model and added it to index
+        addDoc(model);
```
heemin32, line 220: Is this how we add model in other tests inside this class? I don't think this is a right way to do it.

src/test/java/org/opensearch/knn/plugin/transport/UpdateModelGraveyardTransportActionTests.java
```diff
@@ -165,4 +188,201 @@ public void testCheckBlock() {
...
+
+        // Create basic index (not using k-NN)
+        String testIndex1 = "test-index1";
+        createIndex(testIndex1);
+
+        // Attempt to add model id to graveyard with one non-knn index present, should succeed
```
heemin32, line 226: Not that you need to make a change but this single test is too big imo.
I would rather write a test of
1. create model
2. create 1 knn index not using the model, create 2 knn index using the model
3. verify adding model to graveyard which will fail with correct list of 2 knn index using the model
4. delete the 2 knn index using the model
5. verify adding model to graveyard will succeed.

src/test/java/org/opensearch/knn/plugin/transport/UpdateModelGraveyardTransportActionTests.java
```diff
@@ -165,4 +188,201 @@ public void testCheckBlock() {
             .getInstance(UpdateModelGraveyardTransportAction.class);
         assertNull(updateModelGraveyardTransportAction.checkBlock(null, null));
     }
+
+    public void testGetIndicesUsingModel() throws IOException, ExecutionException, InterruptedException {
```
heemin32, line 192: The method name does not match.

src/test/java/org/opensearch/knn/plugin/transport/UpdateModelGraveyardTransportActionTests.java
ryanbogan: This is how we add models in ModelDaoTests.  I just moved the method from there to the superclass so I could use it in UpdateModelGraveyardTransportActionTests

ryanbogan: The model isn't created at all in the other tests within this class, they just use a model-id to simulate one existing.

PR #1731, opened 2024-06-07
Removes KNNXparser and uses map instead

shatejas:
Moves files from engine.method to index.util package

### Description
Follow up on https://github.com/opensearch-project/k-NN/pull/1707
 
### Check List
  - [ ] All tests pass
  - [ ] New functionality has javadoc added
- [ ] Commits are signed as per the DCO using --signoff 


src/main/java/org/opensearch/knn/index/MethodComponent.java
```diff
@@ -22,6 +22,8 @@
 import java.util.HashMap;
 import java.util.Map;
 import java.util.function.BiFunction;
+import java.util.List;
+import java.util.ArrayList;
```
shatejas, line 26: These two were somehow removed during the merge

jmazanec15 (approved):

PR #1742, opened 2024-06-11
Changes to be able to reuse MethodParameterParser class for neural search

shatejas:
### Description
MethodParameterParser needs to be used in neural search. Making it reusable
 
### Check List
  - [x] All tests pass

### DIFF
--- a/src/main/java/org/opensearch/knn/index/IndexUtil.java
+++ b/src/main/java/org/opensearch/knn/index/IndexUtil.java
@@ -47,7 +47,8 @@ public class IndexUtil {
     private static final Version MINIMAL_SUPPORTED_VERSION_FOR_MODEL_METHOD_COMPONENT_CONTEXT = Version.V_2_13_0;
     private static final Version MINIMAL_SUPPORTED_VERSION_FOR_RADIAL_SEARCH = Version.V_2_14_0;
     private static final Version MINIMAL_SUPPORTED_VERSION_FOR_METHOD_PARAMETERS = Version.V_3_0_0;
-    private static final Map<String, Version> minimalRequiredVersionMap = initializeMinimalRequiredVersionMap();
+    // public so neural search can access it
+    public static final Map<String, Version> minimalRequiredVersionMap = initializeMinimalRequiredVersionMap();
 
     private static Map<String, Version> initializeMinimalRequiredVersionMap() {
         final Map<String, Version> versionMap = new HashMap<>() {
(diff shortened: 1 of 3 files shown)


src/main/java/org/opensearch/knn/index/query/parser/MethodParametersParser.java
```diff
@@ -61,7 +62,7 @@ public static ValidationException validateMethodParameters(final Map<String, ?>
     }
 
     // deserialize for node to node communication
-    public static Map<String, ?> streamInput(StreamInput in) throws IOException {
+    public static Map<String, ?> streamInput(StreamInput in, Function<String, Boolean> minClusterVersionCheck) throws IOException {
```
ryanbogan, line 65: Is this parameter used anywhere?  I think it might be supposed to replace the check on line 72

src/main/java/org/opensearch/knn/index/IndexUtil.java
```diff
@@ -47,7 +47,8 @@ public class IndexUtil {
     private static final Version MINIMAL_SUPPORTED_VERSION_FOR_MODEL_METHOD_COMPONENT_CONTEXT = Version.V_2_13_0;
     private static final Version MINIMAL_SUPPORTED_VERSION_FOR_RADIAL_SEARCH = Version.V_2_14_0;
     private static final Version MINIMAL_SUPPORTED_VERSION_FOR_METHOD_PARAMETERS = Version.V_3_0_0;
-    private static final Map<String, Version> minimalRequiredVersionMap = initializeMinimalRequiredVersionMap();
+    // public so neural search can access it
+    public static final Map<String, Version> minimalRequiredVersionMap = initializeMinimalRequiredVersionMap();
```
ryanbogan, line 51: Out of curiosity, could neural search not access the map through the public methods below?

src/main/java/org/opensearch/knn/index/query/parser/MethodParametersParser.java
shatejas: Err, yes I missed replacing that. Nice catch!

src/main/java/org/opensearch/knn/index/IndexUtil.java
shatejas: So the methods below use KNNClusterState. And for neural search it uses NeuralClusterState. so can't reuse those methods. Since it was static I went with making it public instead of a getter

ryanbogan: Makes sense, thanks for the insight

ryanbogan (approved):
LGTM!

junqiu-lei (approved):

PR #1747, opened 2024-06-11
Add jni interface to use a binary hnsw index with faiss

heemin32:
### Description
Add jni interface to use a binary hnsw index with faiss

SearchParameter is not supported with binary hnsw index in faiss library yet. Therefore, multi vector, pre filtering, and ef-search parameter passing won't work with binary index. Support of SearchParameter in binary index with faiss will be handled either by a PR in faiss repo or faiss custom patch in a separate PR.
 
### Issues Resolved
N/A
 
### Check List
  - [X] All tests pass

### DIFF
--- a/jni/include/commons.h
+++ b/jni/include/commons.h
@@ -22,9 +22,23 @@ namespace knn_jni {
          * @param memoryAddress The address of the memory location where data will be stored.
          * @param data 2D float array containing data to be stored in native memory.
          * @param initialCapacity The initial capacity of the memory location.
-         * @return memory address where the data is stored.
+         * @return memory address of std::vector<float> where the data is stored.
          */
         jlong storeVectorData(knn_jni::JNIUtilInterface *, JNIEnv *, jlong , jobjectArray, jlong);
+
+        /**
+         * This is utility function that can be used to store data in native memory. This function will allocate memory for
+         * the data(rows*columns) with initialCapacity and return the memory address where the data is stored.
+         * If you are using this function for first time use memoryAddress = 0 to ensure that a new memory location is created.
+         * For subsequent calls you can pass the same memoryAddress. If the data cannot be stored in the memory location
+         * will throw Exception.
+         *
+         * @param memoryAddress The address of the memory location where data will be stored.
+         * @param data 2D byte array containing data to be stored in native memory.
+         * @param initialCapacity The initial capacity of the memory location.
+         * @return memory address of std::vector<uint8_t> where the data is stored.
+         */
+        jlong storeByteVectorData(knn_jni::JNIUtilInterface *, JNIEnv *, jlong , jobjectArray, jlong);
 
         /**
          * Free up the memory allocated for the data stored in memory address. This function should be used with the memory
@@ -33,5 +47,13 @@ namespace knn_jni {
          * @param memoryAddress address to be freed.
          */
         void freeVectorData(jlong);
+
+        /**
+         * Free up the memory allocated for the data stored in memory address. This function should be used with the memory
+         * address returned by {@link JNICommons#storeByteVectorData(long, byte[][], long, long)}
+         *
+         * @param memoryAddress address to be freed.
+         */
+        void freeByteVectorData(jlong);
     }
 }
--- a/jni/include/faiss_index_service.h
+++ b/jni/include/faiss_index_service.h
@@ -1,0 +1,113 @@
+// SPDX-License-Identifier: Apache-2.0
+//
+// The OpenSearch Contributors require contributions made to
+// this file be licensed under the Apache-2.0 license or a
+// compatible open source license.
+//
+// Modifications Copyright OpenSearch Contributors. See
+// GitHub history for details.
+
+/**
+ * This file contains classes for index operations which are free of JNI
+ */
+
+#ifndef OPENSEARCH_KNN_FAISS_INDEX_SERVICE_H
+#define OPENSEARCH_KNN_FAISS_INDEX_SERVICE_H
+
+#include <jni.h>
+#include "faiss/MetricType.h"
+#include "jni_util.h"
+#include "faiss_methods.h"
+#include <memory>
+
+namespace knn_jni {
+namespace faiss_wrapper {
+
+
+/**
+ * A class to provide operations on index
+ * This class should evolve to have only cpp object but not jni object
+ */
+class IndexService {
+public:
+    IndexService(std::unique_ptr<FaissMethods> faissMethods);
+    //TODO Remove dependency on JNIUtilInterface and JNIEnv
+    //TODO Reduce the number of parameters
+
+    /**
+     * Create index
+     *
+     * @param jniUtil jni util
+     * @param env jni environment
+     * @param metric space type for distance calculation
+     * @param indexDescription index description to be used by faiss index factory
+     * @param dim dimension of vectors
+     * @param numIds number of vectors
+     * @param threadCount number of thread count to be used while adding data
+     * @param vectorsAddress memory address which is holding vector data
+     * @param ids a list of document ids for corresponding vectors
+     * @param indexPath path to write index
+     * @param parameters parameters to be applied to faiss index
+     */
+    virtual void createIndex(
+        knn_jni::JNIUtilInterface * jniUtil,
+        JNIEnv * env,
+        faiss::MetricType metric,
+        std::string indexDescription,
+        int dim,
+        int numIds,
+        int threadCount,
+        int64_t vectorsAddress,
+        std::vector<int64_t> ids,
+        std::string indexPath,
+        std::unordered_map<std::string, jobject> parameters);
+    virtual ~IndexService() = default;
+protected:
+    std::unique_ptr<FaissMethods> faissMethods;
+};
+
+/**
+ * A class to provide operations on index
+ * This class should evolve to have only cpp object but not jni object
+ */
+class BinaryIndexService : public IndexService {
+public:
+    //TODO Remove dependency on JNIUtilInterface and JNIEnv
+    //TODO Reduce the number of parameters
+    BinaryIndexService(std::unique_ptr<FaissMethods> faissMethods);
+    /**
+     * Create binary index
+     *
+     * @param jniUtil jni util
+     * @param env jni environment
+     * @param metric space type for distance calculation
+     * @param indexDescription index description to be used by faiss index factory
+     * @param dim dimension of vectors
+     * @param numIds number of vectors
+     * @param threadCount number of thread count to be used while adding data
+     * @param vectorsAddress memory address which is holding vector data
+     * @param ids a list of document ids for corresponding vectors
+     * @param indexPath path to write index
+     * @param parameters parameters to be applied to faiss index
+     */
+    virtual void createIndex(
+        knn_jni::JNIUtilInterface * jniUtil,
+        JNIEnv * env,
+        faiss::MetricType metric,
+        std::string indexDescription,
+        int dim,
+        int numIds,
+        int threadCount,
+        int64_t vectorsAddress,
+        std::vector<int64_t> ids,
+        std::string indexPath,
+        std::unordered_map<std::string, jobject> parameters
+    ) override;
+    virtual ~BinaryIndexService() = default;
+};
+
+}
+}
+
+
+#endif //OPENSEARCH_KNN_FAISS_INDEX_SERVICE_H
--- a/jni/include/faiss_methods.h
+++ b/jni/include/faiss_methods.h
@@ -1,0 +1,42 @@
+// SPDX-License-Identifier: Apache-2.0
+//
+// The OpenSearch Contributors require contributions made to
+// this file be licensed under the Apache-2.0 license or a
+// compatible open source license.
+//
+// Modifications Copyright OpenSearch Contributors. See
+// GitHub history for details.
+
+#ifndef OPENSEARCH_KNN_FAISS_METHODS_H
+#define OPENSEARCH_KNN_FAISS_METHODS_H
+
+#include "faiss/Index.h"
+#include "faiss/IndexBinary.h"
+#include "faiss/IndexIDMap.h"
+#include "faiss/index_io.h"
+
+namespace knn_jni {
+namespace faiss_wrapper {
+
+/**
+ * A class having wrapped faiss methods
+ *
+ * This class helps to mock faiss methods during unit test
+ */
+class FaissMethods {
+public:
+    FaissMethods() = default;
+    virtual faiss::Index* indexFactory(int d, const char* description, faiss::MetricType metric);
+    virtual faiss::IndexBinary* indexBinaryFactory(int d, const char* description);
+    virtual faiss::IndexIDMapTemplate<faiss::Index>* indexIdMap(faiss::Index* index);
+    virtual faiss::IndexIDMapTemplate<faiss::IndexBinary>* indexBinaryIdMap(faiss::IndexBinary* index);
+    virtual void writeIndex(const faiss::Index* idx, const char* fname);
+    virtual void writeIndexBinary(const faiss::IndexBinary* idx, const char* fname);
+    virtual ~FaissMethods() = default;
+};
+
+} //namespace faiss_wrapper
+} //namespace knn_jni
+
+
+#endif //OPENSEARCH_KNN_FAISS_METHODS_H
--- a/jni/src/faiss_index_service.cpp
+++ b/jni/src/faiss_index_service.cpp
@@ -1,0 +1,164 @@
+// SPDX-License-Identifier: Apache-2.0
+//
+// The OpenSearch Contributors require contributions made to
+// this file be licensed under the Apache-2.0 license or a
+// compatible open source license.
+//
+// Modifications Copyright OpenSearch Contributors. See
+// GitHub history for details.
+
+#include "faiss_index_service.h"
+#include "faiss_methods.h"
+#include "faiss/index_factory.h"
+#include "faiss/Index.h"
+#include "faiss/IndexBinary.h"
+#include "faiss/IndexHNSW.h"
+#include "faiss/IndexBinaryHNSW.h"
+#include "faiss/IndexIVFFlat.h"
+#include "faiss/IndexBinaryIVF.h"
+#include "faiss/IndexIDMap.h"
+#include "faiss/index_io.h"
+#include <algorithm>
+#include <string>
+#include <vector>
+#include <memory>
+#include <type_traits>
+
+namespace knn_jni {
+namespace faiss_wrapper {
+
+template<typename INDEX, typename IVF, typename HNSW>
+void SetExtraParameters(knn_jni::JNIUtilInterface * jniUtil, JNIEnv *env,
+                        const std::unordered_map<std::string, jobject>& parametersCpp, INDEX * index) {
+    std::unordered_map<std::string,jobject>::const_iterator value;
+    if (auto * indexIvf = dynamic_cast<IVF*>(index)) {
+        if ((value = parametersCpp.find(knn_jni::NPROBES)) != parametersCpp.end()) {
+            indexIvf->nprobe = jniUtil->ConvertJavaObjectToCppInteger(env, value->second);
+        }
+
+        if ((value = parametersCpp.find(knn_jni::COARSE_QUANTIZER)) != parametersCpp.end()
+                && indexIvf->quantizer != nullptr) {
+            auto subParametersCpp = jniUtil->ConvertJavaMapToCppMap(env, value->second);
+            SetExtraParameters<INDEX, IVF, HNSW>(jniUtil, env, subParametersCpp, indexIvf->quantizer);
+        }
+    }
+
+    if (auto * indexHnsw = dynamic_cast<HNSW*>(index)) {
+
+        if ((value = parametersCpp.find(knn_jni::EF_CONSTRUCTION)) != parametersCpp.end()) {
+            indexHnsw->hnsw.efConstruction = jniUtil->ConvertJavaObjectToCppInteger(env, value->second);
+        }
+
+        if ((value = parametersCpp.find(knn_jni::EF_SEARCH)) != parametersCpp.end()) {
+            indexHnsw->hnsw.efSearch = jniUtil->ConvertJavaObjectToCppInteger(env, value->second);
+        }
+    }
+}
+
+IndexService::IndexService(std::unique_ptr<FaissMethods> faissMethods) : faissMethods(std::move(faissMethods)) {}
+
+void IndexService::createIndex(
+        knn_jni::JNIUtilInterface * jniUtil,
+        JNIEnv * env,
+        faiss::MetricType metric,
+        std::string indexDescription,
+        int dim,
+        int numIds,
+        int threadCount,
+        int64_t vectorsAddress,
+        std::vector<int64_t> ids,
+        std::string indexPath,
+        std::unordered_map<std::string, jobject> parameters
+    ) {
+    // Read vectors from memory address
+    auto *inputVectors = reinterpret_cast<std::vector<float>*>(vectorsAddress);
+
+    // The number of vectors can be int here because a lucene segment number of total docs never crosses INT_MAX value
+    int numVectors = (int) (inputVectors->size() / (uint64_t) dim);
+    if(numVectors == 0) {
+        throw std::runtime_error("Number of vectors cannot be 0");
+    }
+
+    if (numIds != numVectors) {
+        throw std::runtime_error("Number of IDs does not match number of vectors");
+    }
+
+    std::unique_ptr<faiss::Index> indexWriter(faissMethods->indexFactory(dim, indexDescription.c_str(), metric));
+
+    // Set thread count if it is passed in as a parameter. Setting this variable will only impact the current thread
+    if(threadCount != 0) {
+        omp_set_num_threads(threadCount);
+    }
+
+    // Add extra parameters that cant be configured with the index factory
+    SetExtraParameters<faiss::Index, faiss::IndexIVF, faiss::IndexHNSW>(jniUtil, env, parameters, indexWriter.get());
+
+    // Check that the index does not need to be trained
+    if(!indexWriter->is_trained) {
+        throw std::runtime_error("Index is not trained");
+    }
+
+    // Add vectors
+    std::unique_ptr<faiss::IndexIDMap> idMap(faissMethods->indexIdMap(indexWriter.get()));
+    idMap->add_with_ids(numVectors, inputVectors->data(), ids.data());
+
+    // Write the index to disk
+    faissMethods->writeIndex(idMap.get(), indexPath.c_str());
+}
+
+BinaryIndexService::BinaryIndexService(std::unique_ptr<FaissMethods> faissMethods) : IndexService(std::move(faissMethods)) {}
+
+void BinaryIndexService::createIndex(
+        knn_jni::JNIUtilInterface * jniUtil,
+        JNIEnv * env,
+        faiss::MetricType metric,
+        std::string indexDescription,
+        int dim,
+        int numIds,
+        int threadCount,
+        int64_t vectorsAddress,
+        std::vector<int64_t> ids,
+        std::string indexPath,
+        std::unordered_map<std::string, jobject> parameters
+    ) {
+    // Read vectors from memory address
+    auto *inputVectors = reinterpret_cast<std::vector<uint8_t>*>(vectorsAddress);
+
+    if (dim % 8 != 0) {
+        throw std::runtime_error("Dimensions should be multiply of 8");
+    }
+    // The number of vectors can be int here because a lucene segment number of total docs never crosses INT_MAX value
+    int numVectors = (int) (inputVectors->size() / (uint64_t) (dim / 8));
+    if(numVectors == 0) {
+        throw std::runtime_error("Number of vectors cannot be 0");
+    }
+
+    if (numIds != numVectors) {
+        throw std::runtime_error("Number of IDs does not match number of vectors");
+    }
+
+    std::unique_ptr<faiss::IndexBinary> indexWriter(faissMethods->indexBinaryFactory(dim, indexDescription.c_str()));
+
+    // Set thread count if it is passed in as a parameter. Setting this variable will only impact the current thread
+    if(threadCount != 0) {
+        omp_set_num_threads(threadCount);
+    }
+
+    // Add extra parameters that cant be configured with the index factory
+    SetExtraParameters<faiss::IndexBinary, faiss::IndexBinaryIVF, faiss::IndexBinaryHNSW>(jniUtil, env, parameters, indexWriter.get());
+
+    // Check that the index does not need to be trained
+    if(!indexWriter->is_trained) {
+        throw std::runtime_error("Index is not trained");
+    }
+
+    // Add vectors
+    std::unique_ptr<faiss::IndexBinaryIDMap> idMap(faissMethods->indexBinaryIdMap(indexWriter.get()));
+    idMap->add_with_ids(numVectors, inputVectors->data(), ids.data());
+
+    // Write the index to disk
+    faissMethods->writeIndexBinary(idMap.get(), indexPath.c_str());
+}
+
+} // namespace faiss_wrapper
+} // namesapce knn_jni
--- a/jni/src/faiss_wrapper.cpp
+++ b/jni/src/faiss_wrapper.cpp
@@ -12,6 +12,7 @@
 #include "jni_util.h"
 #include "faiss_wrapper.h"
 #include "faiss_util.h"
+#include "faiss_index_service.h"
 
 #include "faiss/impl/io.h"
 #include "faiss/index_factory.h"
@@ -22,6 +23,8 @@
 #include "faiss/Index.h"
 #include "faiss/impl/IDSelector.h"
 #include "faiss/IndexIVFPQ.h"
+#include "faiss/IndexBinaryIVF.h"
+#include "faiss/IndexBinaryHNSW.h"
 
 #include <algorithm>
 #include <jni.h>
@@ -82,8 +85,7 @@ bool isIndexIVFPQL2(faiss::Index * index);
 faiss::IndexIVFPQ * extractIVFPQIndex(faiss::Index * index);
 
 void knn_jni::faiss_wrapper::CreateIndex(knn_jni::JNIUtilInterface * jniUtil, JNIEnv * env, jintArray idsJ, jlong vectorsAddressJ, jint dimJ,
-                                         jstring indexPathJ, jobject parametersJ) {
-
+                                         jstring indexPathJ, jobject parametersJ, IndexService* indexService) {
     if (idsJ == nullptr) {
         throw std::runtime_error("IDs cannot be null");
     }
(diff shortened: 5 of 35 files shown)


CHANGELOG.md
```diff
@@ -27,3 +27,5 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Documentation
 ### Maintenance
 ### Refactoring
+* Add jni interface to use a binary hnsw index with faiss [#1747](https://github.com/opensearch-project/k-NN/pull/1747)
```
navneet1v, line 30: Should this be added in Enhancements or Features?

jni/src/faiss_wrapper.cpp
```diff
@@ -81,6 +85,85 @@ bool isIndexIVFPQL2(faiss::Index * index);
...
+        throw std::runtime_error("Vectors dimensions cannot be less than or equal to 0");
+    }
+
+    if (indexPathJ == nullptr) {
+        throw std::runtime_error("Index path cannot be null");
+    }
+
+    if (parametersJ == nullptr) {
+        throw std::runtime_error("Parameters cannot be null");
+    }
```
navneet1v, line 108: We should abstract the common code in separate functions. There is a lot of copy paste. Lets abstract things out.

src/main/java/org/opensearch/knn/jni/FaissService.java
```diff
@@ -49,6 +49,20 @@ class FaissService {
...
+     * created the memory address and that should only free up the memory. We are tracking the proper fix for this on this
+     * <a href="https://github.com/opensearch-project/k-NN/issues/1600">issue</a>
+     *
+     * @param ids array of ids mapping to the data passed in
+     * @param vectorsAddress address of native memory where vectors are stored
+     * @param dim dimension of the vector to be indexed
+     * @param indexPath path to save index file to
+     * @param parameters parameters to build index
+     */
+    public static native void createBinaryIndex(int[] ids, long vectorsAddress, int dim, String indexPath, Map<String, Object> parameters);
```
navneet1v, line 64: I don't think so we need this. Can we not use the parameters object to know if its a binary index or normal index.

src/main/java/org/opensearch/knn/jni/FaissService.java
```diff
@@ -90,6 +104,14 @@ public static native void createIndexFromTemplate(
      */
     public static native long loadIndex(String indexPath);
 
+    /**
+     * Load a binary index into memory
+     *
+     * @param indexPath path to index file
+     * @return pointer to location in memory the index resides in
+     */
+    public static native long loadBinaryIndex(String indexPath);
```
navneet1v, line 113: same as above.

navneet1v:
@heemin32 lets fix the GH actions too.

CHANGELOG.md
heemin32: Hmm. There is no enhancement or feature where end user can benefit from this PR as of now. Therefore I put it under refactoring.

src/main/java/org/opensearch/knn/jni/FaissService.java
heemin32: I believe such logic should be in Java side to make life easier. Also, the difference between binary index and non binary index is more than what it looks like at first glance. If we merge the two method, there will be lots of if/else and lots of different type of variables which will be used either by binary index or non binary index only.

src/main/java/org/opensearch/knn/jni/FaissService.java
heemin32: same as above as well :)

src/testFixtures/java/org/opensearch/knn/TestUtils.java
```diff
@@ -327,10 +332,46 @@ private float[][] readQueries(String path) throws IOException {
...
+            for (int i = 0; i < indexData.vectors.length; i++) {
+                for (int j = 0; j < indexData.vectors[i].length; j++) {
+                    flattenedVectors.add(indexData.vectors[i][j]);
+                }
+            }
+            Collections.sort(flattenedVectors);
+            Float median = flattenedVectors.get(flattenedVectors.size() / 2);
+
+            // Quantize + Packing for index data
+            indexBinaryData = new byte[indexData.vectors.length][(indexData.vectors[0].length + 7) / 8];
```
naveentatikonda, line 347: Can you add a comment explaining how you are quantizing and generating the data

heemin32: Quantization is obvious from the code I guess? `indexData.vectors[i][j] >= median ? 1 : 0`
Packing is also nothing special. Just packing 8 binary values into a single byte value.

heemin32: Let me add a comment.

shatejas (commented):
Reviewed this partly

src/main/java/org/opensearch/knn/jni/FaissService.java
shatejas: > If we merge the two method, there will be lots of if/else and lots of different type of variables

I would want to understand this further. Shouldn't we be able to avoid/minimize if-else by introducing an interface? 

```
- Convert everything to cpp type upfront
- Call the factory which can decide on the type of index and execute the operations required
- delete references and return results
```

heemin32:
> @heemin32 lets fix the GH actions too.

Will do. So far, the error is coming from https://github.com/opensearch-project/k-NN/pull/1747/files#diff-892c5ae9158d35b9f5fa9b67d53d7373ca6d2bf04d1052add99c329d03645409R110 as index description is not passed to the method.

jni/src/faiss_wrapper.cpp
```diff
@@ -81,6 +85,85 @@ bool isIndexIVFPQL2(faiss::Index * index);
 // IndexIDMap which has member that will point to underlying index that stores the data
 faiss::IndexIVFPQ * extractIVFPQIndex(faiss::Index * index);
 
+void knn_jni::faiss_wrapper::CreateBinaryIndex(knn_jni::JNIUtilInterface * jniUtil, JNIEnv * env, jintArray idsJ, jlong vectorsAddressJ, jint dimJ,
```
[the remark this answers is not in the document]
heemin32: There is a corresponding method in Java side which is FaissService -> JNIService. We are adding unit test there.

jni/src/faiss_wrapper.cpp
heemin32: @navneet1v Will reduce duplicated logic.
@shatejas As I mentioned in other comment, If there should be a class for index creation related logic, it should be in a java side. Adding more logic in JNI side will make it harder to maintain.

jni/src/faiss_wrapper.cpp
```diff
@@ -409,6 +505,114 @@ jobjectArray knn_jni::faiss_wrapper::QueryIndex_WithFilter(knn_jni::JNIUtilInter
     return results;
 }
 
+jobjectArray knn_jni::faiss_wrapper::QueryBinaryIndex_WithFilter(knn_jni::JNIUtilInterface * jniUtil, JNIEnv * env, jlong indexPointerJ,
```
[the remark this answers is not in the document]
heemin32: There is a java method which linked to this method and we can write a unit test in java side. Is there any reason why you want to write unit test in cpp side?

jni/src/jni_util.cpp
```diff
@@ -302,6 +335,23 @@ int knn_jni::JNIUtil::GetInnerDimensionOf2dJavaFloatArray(JNIEnv *env, jobjectAr
     return dim;
 }
 
+int knn_jni::JNIUtil::GetInnerDimensionOf2dJavaByteArray(JNIEnv *env, jobjectArray array2dJ) {
```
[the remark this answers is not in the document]
heemin32: Where do you see `index` in the signature?

src/main/java/org/opensearch/knn/jni/FaissService.java
heemin32: The parameter seems same but actually it is not because what we have in vectorsAddress is different between binary index and non binary index.

Also, the factory might end up having two different method(like what we have now) to create index at the end anyway.

src/main/java/org/opensearch/knn/jni/JNIService.java
```diff
@@ -50,7 +52,11 @@ public static void createIndex(
         }
 
         if (KNNEngine.FAISS == knnEngine) {
-            FaissService.createIndex(ids, vectorsAddress, dim, indexPath, parameters);
+            if (parameters.get(KNNConstants.INDEX_DESCRIPTION_PARAMETER).toString().startsWith(FAISS_BINARY_INDEX_PREFIX)) {
```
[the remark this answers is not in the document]
heemin32: index description is required in index factory. See
https://github.com/facebookresearch/faiss/wiki/The-index-factory
https://github.com/facebookresearch/faiss/wiki/Binary-indexes

Exact match does not work here.

src/main/java/org/opensearch/knn/jni/JNIService.java
```diff
@@ -50,7 +52,11 @@ public static void createIndex(
...
+                FaissService.createBinaryIndex(ids, vectorsAddress, dim, indexPath, parameters);
+            } else {
```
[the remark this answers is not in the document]
heemin32: I don't think having faiss specific logic in java side is a bad idea. I would rather make cpp side as simple as possible and add most of logic in java side. It is a pain to work on cpp side tbh.

src/main/java/org/opensearch/knn/jni/JNIService.java
```diff
@@ -195,6 +205,39 @@ public static KNNQueryResult[] queryIndex(
...
+     *
+     * @param indexPointer  pointer to index in memory
+     * @param queryVector   vector to be used for query
+     * @param k             neighbors to be returned
+     * @param knnEngine     engine to query index
+     * @param filteredIds   array of ints on which should be used for search.
+     * @param filterIdsType how to filter ids: Batch or BitMap
+     * @return KNNQueryResult array of k neighbors
+     */
+    public static KNNQueryResult[] queryBinaryIndex(
```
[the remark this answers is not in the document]
heemin32: There is no binary type. We pack the binary in byte type. Therefore the byte is just a place holder for 8 binary. It does not query byte.

src/test/java/org/opensearch/knn/jni/JNIServiceTests.java
```diff
@@ -647,6 +648,24 @@ public void testCreateIndex_faiss_valid() throws IOException {
         }
     }
 
+    @SneakyThrows
+    public void testCreateIndex_binary_faiss_valid() {
+        List<String> methods = ImmutableList.of(faissBinaryMethod);
```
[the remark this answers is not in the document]
heemin32: I can remove it. Just following the pattern of other tests.

src/test/java/org/opensearch/knn/jni/JNIServiceTests.java
```diff
@@ -647,6 +648,24 @@ public void testCreateIndex_faiss_valid() throws IOException {
...
+            long memoryAddr = testData.loadBinaryDataToMemoryAddress();
+            JNIService.createIndex(
+                testData.indexData.docs,
+                memoryAddr,
+                testData.indexData.getDimension(),
+                tmpFile1.toAbsolutePath().toString(),
+                ImmutableMap.of(INDEX_DESCRIPTION_PARAMETER, method),
+                KNNEngine.FAISS
+            );
+            assertTrue(tmpFile1.toFile().length() > 0);
```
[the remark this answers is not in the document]
heemin32: Agree that this is not a strong test case.
We should refactor the code to make those method as non static so the JNIService method is testable...

jni/src/faiss_wrapper.cpp
heemin32: @shatejas On a second thought, maybe having a factory class might be good ad you suggested...

src/main/java/org/opensearch/knn/jni/JNIService.java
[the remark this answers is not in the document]
shatejas: Can't we have the next best thing?
```
Static set of these strings
BFlat 
BIVF1024,
BHNSW16 
BIVF1024_HNSW16
```

```
if(Set.contains(indexDescription)) ....
```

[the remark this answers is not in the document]
heemin32: No. The number is not fixed. User can provide any.

jni/src/jni_util.cpp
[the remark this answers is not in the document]
heemin32: Yes. This method has an assumption that the length of element in the array is all same.
Still, opening up the index is unnecessary at this point imo.

jni/include/faiss_index_create.h
```diff
@@ -0,0 +1,100 @@
...
+            int threadCount,
+            int64_t vectorsAddress,
+            std::vector<int64_t> ids,
+            std::string indexPath,
+            std::unordered_map<std::string, jobject> parameters
+        );
+        void create() override;
+};
+
+class IndexCreateServiceFactory {
```
shatejas, line 90: nit: formatting

jni/include/faiss_util.h
```diff
@@ -14,11 +14,29 @@
...
+#include "faiss/impl/io.h"
+#include "faiss/index_factory.h"
+#include "faiss/index_io.h"
+#include "faiss/IndexHNSW.h"
+#include "faiss/IndexIVFFlat.h"
+#include "faiss/MetaIndexes.h"
+#include "faiss/Index.h"
+#include "faiss/impl/IDSelector.h"
+#include "faiss/IndexIVFPQ.h"
+#include "faiss/IndexBinaryIVF.h"
```
shatejas, line 27: nit: wondering if these are needed

jni/src/faiss_index_create.cpp
```diff
@@ -0,0 +1,265 @@
...
+    int dim,
+    int numIds,
+    int threadCount,
+    int64_t vectorsAddress,
+    std::vector<int64_t> ids,
+    std::string indexPath,
+    std::unordered_map<std::string, jobject> parameters
+) : jniUtil(jniUtil), env(env), metric(metric), indexDescription(indexDescription), dim(dim), numIds(numIds), threadCount(threadCount), vectorsAddress(vectorsAddress), ids(ids), indexPath(indexPath), parameters(parameters) {}
+
+void IndexCreateService::create() {
```
shatejas, line 78: Thanks for separating this out. The code structure looks much better

jni/include/faiss_index_create.h
heemin32: If you are referring the missing indentation with two namespaces, it is by intention to avoid having large space for each line.

jni/include/faiss_util.h
heemin32: Will check. Thanks.

jni/include/faiss_index_create.h
```diff
@@ -0,0 +1,100 @@
...
+#include <algorithm>
+#include <string>
+#include <vector>
+#include <memory>
+#include <jni.h>
+
+namespace knn_jni {
+namespace faiss_wrapper {
+
+class IndexCreateService {
```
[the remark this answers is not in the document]
navneet1v: My only question on this would be should we make this as IndexService rather than index create service so it can support both indexing and search. We can start with create functions for now. @heemin32 whats your thoughts?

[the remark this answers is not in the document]
heemin32: Yes. By looking at the code again, it is badly written. Instead of having all required values as class attributes, I can simply receive them. as create method parameter.

I think factory class is not needed and the code in factory class can be in JNI method. Let me refactor it.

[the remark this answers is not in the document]
navneet1v: Sure.. Lets give this a try. Thanks for taking up this effort. This is really tricky and was pending for a long time. Appreciate all the effort we can put to reduce the complexity

src/main/java/org/opensearch/knn/jni/JNIService.java
```diff
@@ -101,7 +103,12 @@ public static long loadIndex(String indexPath, Map<String, Object> parameters, K
         }
 
         if (KNNEngine.FAISS == knnEngine) {
-            return FaissService.loadIndex(indexPath);
+            if (parameters.get(KNNConstants.INDEX_DESCRIPTION_PARAMETER) != null
+                && parameters.get(KNNConstants.INDEX_DESCRIPTION_PARAMETER).toString().startsWith(FAISS_BINARY_INDEX_PREFIX)) {
+                return FaissService.loadBinaryIndex(indexPath);
+            } else {
+                return FaissService.loadIndex(indexPath);
+            }
```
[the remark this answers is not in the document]
heemin32: In my opinion, cpp code and JNI layer should be as thin as possible for better maintainability and development efficiency. 
Any code duplication in cpp code can be avoided if these two methods share single method internally.

heemin32:
Refactored and added unit tests for createIndex method. For queryIndex, it needs another big change as current code shape is not test friendly. Hope, we can improve it incrementally outside of this PR.

src/main/java/org/opensearch/knn/jni/JNIService.java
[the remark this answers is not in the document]
navneet1v: > Any code duplication in cpp code can be avoided if these two methods share single method internally.

Yes this is what my point was. So, main question is how difficult it is to say we create a functions called as


`FaissService.loadIndex(indexPath, indexType)`, here indexType can be a string or integer whatever suits us whose value can be binary or empty if not binary. Then in the cpp layer we check if the indexTypeString == 'binary' then we load the index as a binary index.

As you are making the change and more closer to the code then I am, would like to know if this is something we want to take right now if its a small change or it is big enough that we will need a separate PR.

jni/src/faiss_index_service.cpp
```diff
@@ -0,0 +1,174 @@
...
+#include <string>
+#include <vector>
+#include <memory>
+#include <type_traits>
+
+namespace knn_jni {
+namespace faiss_wrapper {
+
+template<typename INDEX, typename IVF, typename HNSW>
+void SetExtraParameters(knn_jni::JNIUtilInterface * jniUtil, JNIEnv *env,
```
navneet1v, line 31: really good abstraction.. Just by looking at it feels overkill, but when I looked at the reusability of the code looks absolutely awesome. But not sure if this is the best way to achieve this.

jni/src/faiss_wrapper.cpp
```diff
@@ -108,63 +110,61 @@ void knn_jni::faiss_wrapper::CreateIndex(knn_jni::JNIUtilInterface * jniUtil, JN
...
     // Releasing the vectorsAddressJ memory as that is not required once we have created the index.
     // This is not the ideal approach, please refer this gh issue for long term solution:
     // https://github.com/opensearch-project/k-NN/issues/1600
-    delete inputVectors;
+    if (isBinaryIndex) {
+        delete reinterpret_cast<std::vector<uint8_t>*>(vectorsAddress);
+    }
+    else {
+        delete reinterpret_cast<std::vector<float>*>(vectorsAddress);
+    }
```
navneet1v, line 167: why we need to `reinterpret_cast` the casting here? We can just call delete and then the address?

jni/include/faiss_index_service.h
```diff
@@ -0,0 +1,76 @@
...
+#include "jni_util.h"
+#include "faiss_methods.h"
+#include <memory>
+
+namespace knn_jni {
+namespace faiss_wrapper {
+
+class IndexService {
+public:
+    IndexService(std::unique_ptr<FaissMethods> faissMethods);
```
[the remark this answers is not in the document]
heemin32: For better testability. This help us to mock faiss method and write unit test easily.

jni/include/faiss_index_service.h
```diff
@@ -0,0 +1,76 @@
...
+#include <jni.h>
+#include "faiss/MetricType.h"
+#include "jni_util.h"
+#include "faiss_methods.h"
+#include <memory>
+
+namespace knn_jni {
+namespace faiss_wrapper {
+
+class IndexService {
```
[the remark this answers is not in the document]
heemin32: ack

src/main/java/org/opensearch/knn/jni/JNIService.java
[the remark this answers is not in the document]
heemin32: You are right. We make a decision in cpp for createIndex and make a decision in java for load/query index. Let me  move the decision role from cpp to java for create index method as well.

src/main/java/org/opensearch/knn/index/KNNIndexShard.java
```diff
@@ -171,6 +179,16 @@ List<EngineFileContext> getEngineFileContexts(IndexReader indexReader, KNNEngine
...
                     String modelId = fieldInfo.attributes().getOrDefault(MODEL_ID, null);
+                    String indexDescription = null;
+                    if (KNNEngine.FAISS.equals(knnEngine)
+                        && fieldInfo.attributes().get(KNNConstants.KNN_ENGINE).equals(KNNEngine.FAISS.getName())) {
+                        indexDescription = (String) XContentHelper.createParser(
+                            NamedXContentRegistry.EMPTY,
+                            DeprecationHandler.THROW_UNSUPPORTED_OPERATION,
+                            new BytesArray(fieldInfo.attributes().get(KNNConstants.PARAMETERS)),
+                            MediaTypeRegistry.getDefaultMediaType()
+                        ).map().getOrDefault(INDEX_DESCRIPTION_PARAMETER, null);
```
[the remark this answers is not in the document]
heemin32: ack

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -227,6 +233,16 @@ private Map<Integer, Float> doANNSearch(final LeafReaderContext context, final B
             spaceType = SpaceType.getSpace(spaceTypeName);
         }
 
+        String indexDescription = null;
+        if (KNNEngine.FAISS.equals(knnEngine) && modelId == null) {
```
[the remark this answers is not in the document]
heemin32: ack

jni/src/faiss_index_service.cpp
heemin32: Yeah. It was my best effort. There could be better a way though.

jni/include/faiss_index_service.h
```diff
@@ -0,0 +1,76 @@
...
+public:
+    IndexService(std::unique_ptr<FaissMethods> faissMethods);
+    virtual void createIndex(
```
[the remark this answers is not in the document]
heemin32: I didn't see it as an issue enough to be mentioned in TODO but I can add.

src/main/java/org/opensearch/knn/jni/JNIService.java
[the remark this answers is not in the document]
heemin32: > > Any code duplication in cpp code can be avoided if these two methods share single method internally.
> 
> Yes this is what my point was. So, main question is how difficult it is to say we create a functions called as
> 
> `FaissService.loadIndex(indexPath, indexType)`, here indexType can be a string or integer whatever suits us whose value can be binary or empty if not binary. Then in the cpp layer we check if the indexTypeString == 'binary' then we load the index as a binary index.
> 
> As you are making the change and more closer to the code then I am, would like to know if this is something we want to take right now if its a small change or it is big enough that we will need a separate PR.

Just passing indexType is easy work to do. It is more about our strategy on implementation; if we want add more logic to cpp or java. I think it is better to keep the logic as simple as possible in cpp side unless there are performance disadvantage.

src/main/java/org/opensearch/knn/index/IndexUtil.java
```diff
@@ -286,4 +303,13 @@ public static boolean isSharedIndexStateRequired(KNNEngine knnEngine, String mod
         return JNIService.isSharedIndexStateRequired(indexAddr, knnEngine);
     }
 
+    public static String getIndexDescription(FieldInfo fieldInfo) throws IOException {
```
shatejas, line 306: nit: You are extracting information from FieldInformation here. this in indexUtil seems out of place. Maybe consider creating a class `FieldInfoExtractor` and having a method there? That class will have a defined purpose. Having a generic Util class has overloaded responsibilities, we should start moving away from it

src/main/java/org/opensearch/knn/index/KNNIndexShard.java
```diff
@@ -171,6 +173,11 @@ List<EngineFileContext> getEngineFileContexts(IndexReader indexReader, KNNEngine
                     String spaceTypeName = fieldInfo.attributes().getOrDefault(SPACE_TYPE, SpaceType.L2.getValue());
                     SpaceType spaceType = SpaceType.getSpace(spaceTypeName);
                     String modelId = fieldInfo.attributes().getOrDefault(MODEL_ID, null);
+                    String indexDescription = null;
+                    if (KNNEngine.FAISS.equals(knnEngine)
```
shatejas, line 177: nit: Is this `if` branching required? let it return null if it doesn't find it in the fieldinfo. You have to handle null case irrespective of the branching

jni/include/faiss_methods.h
```diff
@@ -0,0 +1,40 @@
...
+#include "faiss/IndexIDMap.h"
+#include "faiss/index_io.h"
+
+namespace knn_jni {
+namespace faiss_wrapper {
+
+/**
+ * A class having wrapped faiss methods
+ */
+class FaissMethods {
```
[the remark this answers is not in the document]
heemin32: Might not possible. They are just simple method but not method under struct or class.

jni/include/commons.h
```diff
@@ -26,12 +26,34 @@ namespace knn_jni {
          */
         jlong storeVectorData(knn_jni::JNIUtilInterface *, JNIEnv *, jlong , jobjectArray, jlong);
 
+        /**
```
jmazanec15, line 29: Should we specify the data type the bytes will be stored in?

jni/include/faiss_index_service.h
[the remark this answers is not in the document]
jmazanec15: Should we put parameter definitions?

jni/include/faiss_index_service.h
```diff
@@ -0,0 +1,82 @@
+// SPDX-License-Identifier: Apache-2.0
```
jmazanec15, line 1: Like this class!

jni/include/faiss_methods.h
```diff
@@ -0,0 +1,42 @@
...
+
+namespace knn_jni {
+namespace faiss_wrapper {
+
+/**
+ * A class having wrapped faiss methods
+ *
+ * This class helps to mock faiss methods during unit test
+ */
+class FaissMethods {
```
jmazanec15, line 26: Are we missing some methods? i.e. load, free, search,...

jni/src/faiss_wrapper.cpp
```diff
@@ -409,6 +410,114 @@ jobjectArray knn_jni::faiss_wrapper::QueryIndex_WithFilter(knn_jni::JNIUtilInter
...
+        faiss::SearchParametersHNSW hnswParams;
+        faiss::SearchParametersIVF ivfParams;
+        std::unique_ptr<faiss::IDGrouperBitmap> idGrouper;
+        std::vector<uint64_t> idGrouperBitmap;
+        auto hnswReader = dynamic_cast<const faiss::IndexBinaryHNSW*>(indexReader->index);
+        if(hnswReader) {
+            // Setting the ef_search value equal to what was provided during index creation. SearchParametersHNSW has a default
+            // value of ef_search = 16 which will then be used.
+            hnswParams.efSearch = hnswReader->hnsw.efSearch;
+            hnswParams.sel = idSelector.get();
```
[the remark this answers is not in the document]
heemin32: TODO might not be enough to catch it. Let me create GH issue instead. https://github.com/opensearch-project/k-NN/issues/1769

jni/src/faiss_wrapper.cpp
```diff
@@ -247,6 +235,19 @@ jlong knn_jni::faiss_wrapper::LoadIndex(knn_jni::JNIUtilInterface * jniUtil, JNI
     return (jlong) indexReader;
 }
 
+jlong knn_jni::faiss_wrapper::LoadBinaryIndex(knn_jni::JNIUtilInterface * jniUtil, JNIEnv * env, jstring indexPathJ) {
```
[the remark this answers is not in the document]
heemin32: Yes. Didn't add it now because wasn't sure if current approach of using index service will be accepted.
Once this PR is merged, will move the load index method into index service in a separate PR.

jmazanec15 (commented):
I like the refactors and I think it looks good. One thing - can we run cpp unit tests with valgrind on linux to see if we are leaking any memory?

jni/include/faiss_methods.h
heemin32: I haven't moved all methods yet. We will add incrementally.

jni/include/commons.h
heemin32: It is in the parameter definition.
`@param data 2D byte array containing data to be stored in native memory.`

jni/include/faiss_methods.h
jmazanec15: Lets add todo

heemin32: I didn't want to go through and see what is needed to be in this class to add todo. We can add methods as needed base. For example, we don't need to add search method here because search is a method of Index struct which can be mocked without this class.

shatejas (approved):

heemin32:
Valgrind result
```
~/k-NN/jni$ valgrind --leak-check=yes ./bin/jni_test 
==54535== Memcheck, a memory error detector
==54535== Copyright (C) 2002-2022, and GNU GPL'd, by Julian Seward et al.
==54535== Using Valgrind-3.22.0 and LibVEX; rerun with -h for copyright info
==54535== Command: ./bin/jni_test
==54535== 
vex amd64->IR: unhandled instruction bytes: 0x62 0xF2 0xFD 0x8 0x7C 0xC0 0xC5 0xF9 0x7F 0x5
vex amd64->IR:   REX=0 REX.W=0 REX.R=0 REX.X=0 REX.B=0
vex amd64->IR:   VEX=0 VEX.L=0 VEX.nVVVV=0x0 ESC=NONE
vex amd64->IR:   PFX.66=0 PFX.F2=0 PFX.F3=0
==54535== valgrind: Unrecognised instruction at address 0x6a15c9e.
==54535==    at 0x6A15C9E: _GLOBAL__sub_I_distcomp_sparse_scalar_fast.cc (in /home/ubuntu/k-NN/jni/release/libopensearchknn_nmslib.so)
==54535==    by 0x400571E: call_init.part.0 (dl-init.c:74)
==54535==    by 0x4005823: call_init (dl-init.c:120)
==54535==    by 0x4005823: _dl_init (dl-init.c:121)
==54535==    by 0x401F59F: ??? (in /usr/lib/x86_64-linux-gnu/ld-linux-x86-64.so.2)
==54535== Your program just tried to execute an instruction that Valgrind
==54535== did not recognise.  There are two possible reasons for this.
==54535== 1. Your program has a bug and erroneously jumped to a non-code
==54535==    location.  If you are running Memcheck and you just saw a
==54535==    warning about a bad jump, it's probably your program's fault.
==54535== 2. The instruction is legitimate but Valgrind doesn't handle it,
==54535==    i.e. it's Valgrind's fault.  If you think this is the case or
==54535==    you are not sure, please let us know and we'll try to fix it.
==54535== Either way, Valgrind will now raise a SIGILL signal which will
==54535== probably kill your program.
==54535== 
==54535== Process terminating with default action of signal 4 (SIGILL)
==54535==  Illegal opcode at address 0x6A15C9E
==54535==    at 0x6A15C9E: _GLOBAL__sub_I_distcomp_sparse_scalar_fast.cc (in /home/ubuntu/k-NN/jni/release/libopensearchknn_nmslib.so)
==54535==    by 0x400571E: call_init.part.0 (dl-init.c:74)
==54535==    by 0x4005823: call_init (dl-init.c:120)
==54535==    by 0x4005823: _dl_init (dl-init.c:121)
==54535==    by 0x401F59F: ??? (in /usr/lib/x86_64-linux-gnu/ld-linux-x86-64.so.2)
==54535== 
==54535== HEAP SUMMARY:
==54535==     in use at exit: 83,256 bytes in 40 blocks
==54535==   total heap usage: 40 allocs, 0 frees, 83,256 bytes allocated
==54535== 
==54535== LEAK SUMMARY:
==54535==    definitely lost: 0 bytes in 0 blocks
==54535==    indirectly lost: 0 bytes in 0 blocks
==54535==      possibly lost: 0 bytes in 0 blocks
==54535==    still reachable: 83,256 bytes in 40 blocks
==54535==         suppressed: 0 bytes in 0 blocks
==54535== Reachable blocks (those to which a pointer was found) are not shown.
==54535== To see them, rerun with: --leak-check=full --show-leak-kinds=all
==54535== 
==54535== For lists of detected and suppressed errors, rerun with: -s
==54535== ERROR SUMMARY: 0 errors from 0 contexts (suppressed: 0 from 0)
Illegal instruction (core dumped)
```

jni/include/faiss_index_service.h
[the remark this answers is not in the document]
navneet1v: I would like to see if there is a better way to mock, not a blocker for now.

jni/src/faiss_index_service.cpp
navneet1v: I think I am good on this for now. Till someone contribute a better way that this.

Vikasht34 (approved):

jni/src/faiss_index_service.cpp
```diff
@@ -0,0 +1,164 @@
...
+            indexHnsw->hnsw.efConstruction = jniUtil->ConvertJavaObjectToCppInteger(env, value->second);
+        }
+
+        if ((value = parametersCpp.find(knn_jni::EF_SEARCH)) != parametersCpp.end()) {
+            indexHnsw->hnsw.efSearch = jniUtil->ConvertJavaObjectToCppInteger(env, value->second);
+        }
+    }
+}
+
+IndexService::IndexService(std::unique_ptr<FaissMethods> faissMethods) : faissMethods(std::move(faissMethods)) {}
```
navneet1v, line 58: The std::move is an interesting choice. But what my thinking is why we are using std::move here rather than just say a [unique pointer](https://en.cppreference.com/w/cpp/memory/unique_ptr) or [may do a call by reference](https://www.tutorialspoint.com/cplusplus/cpp_function_call_by_reference.htm) if we are concerned about copying of objects.

src/main/java/org/opensearch/knn/jni/JNIService.java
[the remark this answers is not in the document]
navneet1v: I agree in terms of keeping things as simple as possible. But I would like to have them at 1 place either in java layer or in cpp layer. In my true opinion I would like to have things in java layer as we get more control.

jni/src/faiss_wrapper.cpp
```diff
@@ -409,6 +410,114 @@ jobjectArray knn_jni::faiss_wrapper::QueryIndex_WithFilter(knn_jni::JNIUtilInter
     return results;
 }
 
+jobjectArray knn_jni::faiss_wrapper::QueryBinaryIndex_WithFilter(knn_jni::JNIUtilInterface * jniUtil, JNIEnv * env, jlong indexPointerJ,
```
navneet1v, line 413: will the abstraction on the query flow be added in the upcoming PRs?

src/main/java/org/opensearch/knn/jni/JNIService.java
```diff
@@ -50,7 +52,12 @@ public static void createIndex(
         }
 
         if (KNNEngine.FAISS == knnEngine) {
-            FaissService.createIndex(ids, vectorsAddress, dim, indexPath, parameters);
+            if (parameters.get(KNNConstants.INDEX_DESCRIPTION_PARAMETER) != null
```
navneet1v, line 55: This logic of identifying the binary index lets abstract this out in a utility. I think this is going to be used at other places too.

jmazanec15 (approved):

jni/src/faiss_index_service.cpp
heemin32: Call by reference increase a chance of memory leak by bug.

src/main/java/org/opensearch/knn/jni/JNIService.java
heemin32: As this is java side code, will make the change in the next PR.

src/main/java/org/opensearch/knn/jni/JNIService.java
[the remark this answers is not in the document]
heemin32: Yes. Moved them to java layer.

jni/src/faiss_wrapper.cpp
navneet1v: Lets do that.

jni/src/faiss_index_service.cpp
navneet1v: as this is an object, and you used the unique pointer, I doubt its the case. But if we 100% sure that this will not cause any performance issues and other memory leak I am okay to approve the code.

navneet1v (approved):

jni/src/faiss_index_service.cpp
heemin32: This std::move is passing its ownership. If you don't use it, the ownership won't be transferred to IndexService and the memory can be released while IndexService sill needs it.

jni/src/faiss_wrapper.cpp
heemin32: https://github.com/opensearch-project/k-NN/issues/1773

PR #1748, opened 2024-06-12
Adds ef_search support for Lucene kNN queries

shatejas:
### Description
Lucene queries use `k` as `efsearch`. So we take the max of two to pass is the right `k` value to Lucene. The `size` parameter will dictate the number of the hits

```
curl -XGET "http://localhost:9200/lucene-hnsw-index/_search" -H 'Content-Type: application/json' -d'
{
"size": 1,
  "query": {
    "knn": {
      "lucene_vector": {
        "vector": [2.5, 3.5, 2.2],
        "k": 3,
        "method_parameters": {
          "ef_search": 5
        }
        }
      }
    }
  }'
```
```
{
  "took": 4,
  "timed_out": false,
  "_shards": {
    "total": 1,
    "successful": 1,
    "skipped": 0,
    "failed": 0
  },
  "hits": {
    "total": {
      "value": 4,
      "relation": "eq"
    },
    "max_score": 1,
    "hits": [
      {
        "_index": "lucene-hnsw-index",
        "_id": "2",
        "_score": 1,
        "_source": {
          "lucene_vector": [
            2.5,
            3.5,
            2.2
          ],
          "price": 7.1
        }
      }
    ]
  }
}
```

 
### Issues Resolved
[#1537](https://github.com/opensearch-project/k-NN/issues/1537)
 
### Check List
  - [x] All tests pass

### DIFF
--- a/src/main/java/org/opensearch/knn/index/query/parser/MethodParametersParser.java
+++ b/src/main/java/org/opensearch/knn/index/query/parser/MethodParametersParser.java
@@ -29,7 +29,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.function.Function;
 
-import static org.opensearch.knn.index.IndexUtil.isClusterOnOrAfterMinRequiredVersion;
 import static org.opensearch.knn.index.query.KNNQueryBuilder.METHOD_PARAMS_FIELD;
 import static org.opensearch.knn.index.query.KNNQueryBuilder.NAME;
 
--- a/src/test/java/org/opensearch/knn/index/LuceneEngineIT.java
+++ b/src/test/java/org/opensearch/knn/index/LuceneEngineIT.java
@@ -499,13 +499,18 @@ public class LuceneEngineIT extends KNNRestTestCase {
         }
 
         validateQueries(spaceType, FIELD_NAME);
+        validateQueries(spaceType, FIELD_NAME, Map.of("ef_search", 100));
     }
 
     private void validateQueries(SpaceType spaceType, String fieldName) throws Exception {
+        validateQueries(spaceType, fieldName, null);
+    }
+
+    private void validateQueries(SpaceType spaceType, String fieldName, Map<String, ?> methodParameters) throws Exception {
 
         int k = LuceneEngineIT.TEST_INDEX_VECTORS.length;
         for (float[] queryVector : TEST_QUERY_VECTORS) {
-            Response response = searchKNNIndex(INDEX_NAME, new KNNQueryBuilder(fieldName, queryVector, k), k);
+            Response response = searchKNNIndex(INDEX_NAME, buildLuceneKSearchQuery(fieldName, k, queryVector, methodParameters), k);
             String responseBody = EntityUtils.toString(response.getEntity());
             List<KNNResult> knnResults = parseSearchResponse(responseBody, fieldName);
             assertEquals(k, knnResults.size());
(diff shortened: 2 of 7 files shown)


ryanbogan (approved):

src/main/java/org/opensearch/knn/index/query/parser/MethodParametersParser.java
```diff
@@ -29,7 +29,6 @@
 import java.util.Map;
 import java.util.function.Function;
 
-import static org.opensearch.knn.index.IndexUtil.isClusterOnOrAfterMinRequiredVersion;
```
naveentatikonda, line 32: Why are we removing this ?

src/test/java/org/opensearch/knn/index/query/KNNQueryFactoryTests.java
```diff
@@ -85,6 +86,98 @@ public void testCreateLuceneDefaultQuery() {
         }
     }
 
+    public void testLuceneFloatVectorQuery() {
+        Query actualQuery1 = KNNQueryFactory.create(
```
naveentatikonda, line 90: Can you add a function to create the query as it is duplicated multiple times

shatejas: It has a different request everytime. The create request itself is using a builder so wrapping it up with a function is moot

src/main/java/org/opensearch/knn/index/query/parser/MethodParametersParser.java
shatejas: It was a leftover from this PR https://github.com/opensearch-project/k-NN/pull/1742/files. The import is not needed. Sorry for the confusion

src/test/java/org/opensearch/knn/index/LuceneEngineIT.java
```diff
@@ -499,13 +499,22 @@ private void baseQueryTest(SpaceType spaceType) throws Exception {
...
+    }
+
+    private void validateQueries(SpaceType spaceType, String fieldName, Map<String, ?> methodParameters) throws Exception {
 
         int k = LuceneEngineIT.TEST_INDEX_VECTORS.length;
         for (float[] queryVector : TEST_QUERY_VECTORS) {
-            Response response = searchKNNIndex(INDEX_NAME, new KNNQueryBuilder(fieldName, queryVector, k), k);
+            Response response = searchKNNIndex(
+                INDEX_NAME,
+                KNNQueryBuilder.builder().fieldName(fieldName).k(k).vector(queryVector).methodParameters(methodParameters).build(),
```
junqiu-lei, line 515: For the end to end test, can we use XContentBuilder to build the query?

https://github.com/opensearch-project/k-NN/blob/main/src/testFixtures/java/org/opensearch/knn/KNNRestTestCase.java#L213

shatejas: Its much more convenient to use KNNBuilder. It converts to XContent [underneath](https://github.com/opensearch-project/k-NN/blob/main/src/testFixtures/java/org/opensearch/knn/KNNRestTestCase.java#L194). 

Not opposed to it but any specific reason we should move to xcontent for happy case tests?

junqiu-lei: Using XContentBuilder can better mock end-user API calls by simulating JSON content creation and ensuring end-to-end functionality.

KNNBuilder is an internal method, and relying on it in integration test has the possibility to lead to fragile tests that may break with internal changes, even if the public API remains stable.

shatejas: Fair, will change it

junqiu-lei (approved):

PR #1765, opened 2024-06-17
Add IllegalArgumentException for Lucene radial search with ef_search

junqiu-lei:
### Description
Add IllegalArgumentException for Lucene radial search with ef_search parameter. Context at https://github.com/opensearch-project/k-NN/issues/1537#issuecomment-2168724430
 
### Issues Resolved
Part of https://github.com/opensearch-project/k-NN/issues/1537
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
+++ b/src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
@@ -8,7 +8,6 @@ package org.opensearch.knn.index.query;
 import lombok.AccessLevel;
 import lombok.AllArgsConstructor;
 import lombok.Getter;
-import lombok.SneakyThrows;
 import lombok.extern.log4j.Log4j2;
 import org.apache.commons.lang.StringUtils;
 import org.apache.lucene.search.MatchNoDocsQuery;
@@ -97,8 +96,6 @@ public class KNNQueryBuilder extends AbstractQueryBuilder<KNNQueryBuilder> {
     private QueryBuilder filter;
     @Getter
     private boolean ignoreUnmapped;
-    @Getter
-    private VectorQueryType vectorQueryType;
 
     /**
      * Constructs a new query with the given field name and vector
@@ -193,23 +190,13 @@ public class KNNQueryBuilder extends AbstractQueryBuilder<KNNQueryBuilder> {
         }
 
         public KNNQueryBuilder build() {
-            VectorQueryType vectorQueryType = validate();
+            validate();
             int k = this.k == null ? 0 : this.k;
-            return new KNNQueryBuilder(
-                fieldName,
-                vector,
-                k,
-                maxDistance,
-                minScore,
-                methodParameters,
-                filter,
-                ignoreUnmapped,
-                vectorQueryType
-            ).boost(boost).queryName(queryName);
+            return new KNNQueryBuilder(fieldName, vector, k, maxDistance, minScore, methodParameters, filter, ignoreUnmapped).boost(boost)
+                .queryName(queryName);
         }
 
-        @SneakyThrows
-        private VectorQueryType validate() {
+        private void validate() {
             if (Strings.isNullOrEmpty(fieldName)) {
                 throw new IllegalArgumentException(String.format("[%s] requires fieldName", NAME));
             }
@@ -228,16 +215,13 @@ public class KNNQueryBuilder extends AbstractQueryBuilder<KNNQueryBuilder> {
                 throw new IllegalArgumentException(String.format("[%s] requires exactly one of k, distance or score to be set", NAME));
             }
 
-            VectorQueryType vectorQueryType = VectorQueryType.MAX_DISTANCE;
             if (k != null) {
-                vectorQueryType = VectorQueryType.K;
                 if (k <= 0 || k > K_MAX) {
                     throw new IllegalArgumentException(String.format("[%s] requires k to be in the range (0, %d]", NAME, K_MAX));
                 }
             }
 
             if (minScore != null) {
-                vectorQueryType = VectorQueryType.MIN_SCORE;
                 if (minScore <= 0) {
                     throw new IllegalArgumentException(String.format("[%s] requires minScore to be greater than 0", NAME));
                 }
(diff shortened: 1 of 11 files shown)


ryanbogan (approved):
LGTM!

junqiu-lei:
The CI failure in `Test k-NN on Secure Cluster` is related to https://github.com/opensearch-project/security-dashboards-plugin/issues/2009

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
```diff
@@ -558,6 +558,13 @@ protected Query doToQuery(QueryShardContext context) {
             radius = knnEngine.scoreToRadialThreshold(this.minScore, spaceType);
         }
 
+        if (radius != null
+            && methodParameters != null
+            && methodParameters.containsKey(METHOD_PARAMETER_EF_SEARCH)
+            && knnEngine.equals(KNNEngine.LUCENE)) {
+            throw new IllegalArgumentException(String.format("[" + NAME + "] Lucene engine does not support ef_search with radial search"));
+        }
```
shatejas, line 566: This is engine specific, we need to move it to EngineSpecificMethodContext. Currently the `supportedMethodParameters()` does not support this. We need to modify the interface to embed this logic in there. Can we change `supportedMethodParameters()` to add a context parameter where we can pass in the information and decide at run time which method parameters to return. Something like this

```
class LuceneHNSWContext implements EngineSpecificMethodContext {

Map<String, ?> supportedMethodParameters(Context ctx) {
HNSWContext ctx = (HNSWContext) ctx;
if(ctx.isRadial()) {
return emptyMap()
}

return supportedMethodParams
}

}
```

junqiu-lei: Good point, let me check.

junqiu-lei: Updated

src/main/java/org/opensearch/knn/validation/ParameterValidator.java
```diff
@@ -14,35 +14,89 @@
...
+
+    private static void validateNonNullParameters(Map<String, Parameter<?>> validParameters) {
         if (validParameters == null) {
             throw new IllegalArgumentException("validParameters cannot be null");
         }
+    }
 
-        if (requestParameters == null || requestParameters.isEmpty()) {
-            return null;
+    private static void checkEngineSpecificErrors(
```
shatejas, line 78: The method is kept generic enough to have validations of any kind lets keep it as it is. There is a generic message on the `doToQuery` layer, that should be enough for users to identify what went wrong, right? am I missing something here? if not lets revert this

src/main/java/org/opensearch/knn/index/util/EngineSpecificMethodContext.java
```diff
@@ -25,7 +25,11 @@
  */
 public interface EngineSpecificMethodContext {
 
-    Map<String, Parameter<?>> supportedMethodParameters();
+    Map<String, Parameter<?>> supportedMethodParameters(Context ctx);
 
-    EngineSpecificMethodContext EMPTY = Collections::emptyMap;
+    EngineSpecificMethodContext EMPTY = ctx -> Collections.emptyMap();
+
+    interface Context {
```
shatejas, line 32: Lets make this its own pojo and have `QueryContext` We can further reuse this context wherever required. We can use inheritance to further if required down the line, something like `RadialQueryContext` and `HNSWQueryContext`

src/main/java/org/opensearch/knn/index/util/EngineSpecificMethodContext.java
```diff
@@ -25,7 +25,11 @@
...
+        boolean isRadialSearch();
```
shatejas, line 33: is this valid for IVF?

junqiu-lei: radial search is supported in faiss IVF

src/main/java/org/opensearch/knn/validation/ParameterValidator.java
junqiu-lei: Updated to no change in this file. And added query type in the error message. https://github.com/opensearch-project/k-NN/pull/1765/files#diff-6d0df17543181185170f65c116cdb3514d910e1557f012b7959611ae92678080R549

src/main/java/org/opensearch/knn/index/util/EngineSpecificMethodContext.java
junqiu-lei: Separated out `QueryContext` from `EngineSpecificMethodContext`. Instead of using `isRadialSearch `, I updated to use `is_ef_search_parameter_supported` to have the check directly. I feel like currently we can use `QueryContext` without inheritance.

shatejas (changes requested):
Looks good overall. Some minor changes

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
```diff
@@ -521,16 +534,19 @@ protected Query doToQuery(QueryShardContext context) {
         final String method = methodComponentContext != null ? methodComponentContext.getName() : null;
         if (StringUtils.isNotBlank(method)) {
             final EngineSpecificMethodContext engineSpecificMethodContext = knnEngine.getMethodContext(method);
+            KNNEngine finalKnnEngine = knnEngine;
+            QueryContext methodContext = () -> (vectorQueryType.isRadialSearch() && KNNEngine.LUCENE.equals(finalKnnEngine)) == false;
```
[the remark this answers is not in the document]
junqiu-lei: Simplified the logic check in updated commit

ryanbogan (approved):

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
```diff
@@ -195,6 +197,7 @@ public KNNQueryBuilder build() {
                 .queryName(queryName);
         }
 
+        @SneakyThrows
```
shatejas, line 200: Lets remove sneaky throws from here

shatejas (approved):

PR #1770, opened 2024-06-25
Add rolling version upgrade support for ignore_unmapped

kotwanikunal:
### Description
- Adds in rolling version upgrade support for `ignore_unmapped`

### Issues Resolved
- N/A
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
+++ b/src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
@@ -331,7 +331,7 @@ public class KNNQueryBuilder extends AbstractQueryBuilder<KNNQueryBuilder> {
      */
     public KNNQueryBuilder(StreamInput in) throws IOException {
         super(in);
-        KNNQueryBuilder.Builder builder = KNNQueryBuilderParser.streamInput(in, IndexUtil::isClusterOnOrAfterMinRequiredVersion);
+        KNNQueryBuilder.Builder builder = KNNQueryBuilderParser.streamInput(in, IndexUtil::isStreamOnOrAfterMinRequiredVersion);
         fieldName = builder.fieldName;
         vector = builder.vector;
         k = builder.k;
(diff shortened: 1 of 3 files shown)


navneet1v:
@kotwanikunal thanks for raising the PR. Will wait for all BWC to gh actions to complete.

navneet1v:
@kotwanikunal can you fix the failed GH actions?

martin-gaievski:
Good finding, we should do same in other repos. Will approve once CI is fixed. Current error is from spotless, you can run `spotlessApply`

kotwanikunal:
Pushed linting fixes in.

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
```diff
@@ -324,7 +326,7 @@ protected void doWriteTo(StreamOutput out) throws IOException {
         out.writeFloatArray(vector);
         out.writeInt(k);
         out.writeOptionalNamedWriteable(filter);
-        if (isClusterOnOrAfterMinRequiredVersion("ignore_unmapped")) {
+        if (out.getVersion().onOrAfter(getMinVersionForFeature("ignore_unmapped"))) {
```
shatejas, line 329: Does this stream guarantee the min version of the cluster. Just verifying since the previous method has that check

kotwanikunal: Is there a historical reason that we look for the cluster version and not the stream version?
It should behave the same - a feature is on only when both the nodes are upgraded. 

@jmazanec15 / @navneet1v

heemin32: We haven't serialized the stream version when we write KNNQueryBuilder at first. Therefore, we were not able to use stream version. After that, we haven't tried to add stream version because cluster version just worked fine for us.

jmazanec15:
@kotwanikunal is there pending?

kotwanikunal:
> @kotwanikunal is there pending?

I had a rebase pending.
Updated the PR with latest references.

Issue #1767, opened 2024-06-19
[RFC] Binary vector support


heemin32:
## Overview

The increasing demand for binary format support from customers is becoming evident, with numerous instances demonstrating strong recall rates when using binary values generated from large language models (LLMs). For example, Cohere's introduction of the Cohere Embed embedding model, which inherently supports binary embeddings, has shown that binary vectors can retain 90-98% of the original search quality.

Given the impressive recall rates achieved with binary vectors, a growing number of users are seeking to leverage binary vectors in OpenSearch KNN indices to significantly reduce memory costs. By moving from float32 vectors to binary vectors, you can reduce the memory requirement by a factor of 32.

Implementing support for binary vectors in OpenSearch KNN indices is thus a highly beneficial feature, addressing customer demand and significantly lowering operational costs. This capability not only ensures high recall performance but also makes large-scale deployment more economically viable, facilitating greater adoption and efficiency.

## Scope

1. Support binary format with Faiss, HNSW
2. Script scoring on binary vector using hamming distance
3. Support binary format with Faiss, IVF
4. Binary format support with radial search on Faiss

## Out of scope

1. Support binary format with Nmslib
2. Binary quantization (Will be handled in a separate issue)
3. Rescoring (Will be handled in a separate issue)

## Future extension

1. Binary format support with Lucene
2. Support of other space type(ex. Jaccard)
3. Exploring of java long data type to store binary vector inside OpenSearch for any performance advantage

## Data flow diagram
![Screenshot 2024-06-18 at 4 01 40 PM](https://github.com/opensearch-project/k-NN/assets/1809492/bbe2cb01-8fc3-4c68-86e4-1b13622f0094)

### Input format
User should pack their binary into byte(int8). For example, for a binary value 0, 1, 1, 0, 0, 0, 1, 1, it will be 99. 

### Index setting
Because we are using int8 format as input, the dimension should be a multiple of 8. We are going to support new data_type, binary. With binary data type, the hammingdistance is the only space type that we are going to support as of now. If space type is not specified, the hammingdistance will be a default value for the binary data type.
```
PUT test-index
{
  "settings": {// no change
    "index": {
      "knn": true,
      "knn.algo_param.ef_search": 100
    }
  },
  "mappings": {
    "properties": {
      "my_vector1": {
        "type": "knn_vector", // no change
        "dimension": 24, // This should be multiple of 8
        "data_type": "binary",// new data type
        "method": {
          "name": "hnsw", // and also ivf
[...]

Issue #1779, opened 2024-06-28
[RFC] Optimized Disk-Based Vector Search


jmazanec15:
## Introduction 

This document outlines a high level proposal for providing efficient, yet easy to use k-NN in OpenSearch in low-memory environments. Many more details to come in individual component RFCs.

## Problem Statement

Vector search has become very popular in recent years due to recent advancements in LLMs. Typically, embedding models like Amazon Titan or Cohere or OpenAI will output fp32 embeddings in high dimensional space (>= 768). These embeddings are then ingested into vector search systems to provide semantic search. While semantic search can provide major benefits in search quality, it has historically come at a much higher cost than full text search. This is mainly due to the size of embeddings produced along with the high in-memory requirements of traditional Approximate Nearest Neighbor algorithms, such as HNSW.

With that, a major user ask is cost reduction. Many new solutions/algorithms are being released to reduce cost. General strategies include leveraging the disk to extend memory and/or using quantization techniques to reduce overall memory footprint per vector.  Currently, OpenSearch offers the following quantization techniques for the 2x, 4x, 8x+ compression levels (assuming input vectors are fp32 d-dimensional vectors, 2x would compress vectors into 16*d bits, etc):
| Compression Rate  | Options |
| ------------- | ------------- |
| 2x  | 1. [faiss] 16-bit Scalar Quantization  |
| 4x  | 1. [faiss] byte/8-bit Scalar Quantization (coming soon)<br> 2. [lucene] byte/8-bit Scalar Quantization (coming soon)  |
| 8x  | 1. [faiss] product quantization (m=dim/2)  |
| 16x  | 1. [faiss] product quantization (m=dim/4)  |
| 32x  | 1. [faiss] product quantization (m=dim/8)  |

Many users want 8x+ compression, meaning that the current scalar quantization techniques are not viable options for them. Product quantization can be employed to meet this requirement. However, the current product quantization implementation has a few drawbacks:

1. Product Quantization requires a separate training stage, which raises the bar for entry to get a working solution. Users need to maintain a separate workflow for creating the model in the cluster. Additionally, it is difficult to detect when the model no longer fits the distribution of the data and needs to be re-trained.
2. Product Quantization recall degrades as compression rate increases. It can be increased in OpenSearch with full-precision re-scoring via script queries, but this interface is very verbose

[...]

Issue #1600, opened 2024-04-09
[Enhancement] [Build-Time V1] Reduce Memory Footprint during Native Index Creation


navneet1v:
## Description
While running some experiments I can am able to see that we need more memory than final graph which gets created in a segment.

**Based on some research what I can see for Faiss**

1. We give faiss, all the vectors and ids for which graph needs to be created. here: https://github.com/navneet1v/k-NN/blob/badbb1d3438a37536ce6de29c4085cded0ef2723/jni/src/faiss_wrapper.cpp#L159 .

2. After this faiss, internally calls add on the index function. here: https://github.com/facebookresearch/faiss/blob/8898eabe9f16e7bbebdc19667c993f7dc55a6a0c/faiss/IndexIDMap.cpp#L80

3. This will call internally add function of IndexHNSW, which will [first store](https://github.com/facebookresearch/faiss/blob/f884ba240ea7ce948ce850c31f82bb2e9eac21b0/faiss/IndexHNSW.cpp#L395) the vector in a [flat index](https://github.com/facebookresearch/faiss/blob/2d8886cd4fb1bd0ebe5f8de7cae4c1ad91c8a40d/faiss/IndexFlatCodes.cpp#L23-L31)(which is nothing but a binary representation of vectors) and then [start creating](https://github.com/facebookresearch/faiss/blob/f884ba240ea7ce948ce850c31f82bb2e9eac21b0/faiss/IndexHNSW.cpp#L398) the HNSW graph.

So based on the above flow I can see that vectors which are stored in native memory in step 1 and passed to Faiss are stored first in FlatIndex of Faiss doubling the native memory.

**For Nmslib:**
1. We first take the vectors and copy vectors into [this new object](https://github.com/opensearch-project/k-NN/blob/feature/stream-vectors/jni/src/nmslib_wrapper.cpp#L140-L153) of nmslib.
2. After that, we [create the index for nmslib](https://github.com/opensearch-project/k-NN/blob/feature/stream-vectors/jni/src/nmslib_wrapper.cpp#L158). 
3. After the index is created we [save the index to a file](https://github.com/opensearch-project/k-NN/blob/feature/stream-vectors/jni/src/nmslib_wrapper.cpp#L159). This index writing makes a very large allocation before it writes the index on the file here: https://github.com/nmslib/nmslib/blob/a2d6624e1315402662025debfdd614b505d9c3ef/similarity_search/src/method/hnsw.cc#L417-L428

## Issue
[...]

PR #1781, opened 2024-07-01
 Add binary format support with HNSW method in Faiss Engine

heemin32:
#### What this PR has
Add binary format support with HNSW method in Faiss engine

#### What this PR does not have
1. Search parameter support which is required for dynamic ef_search value in query, efficient filtering, and multi vector support. This will be handled by a separate PR.
2. Script scoring

#### Changes
1. Introduction of `UNDEFINED` space type. This is needed because the space type is parsed before we know about data type and we are going to have a different default space type per `data_type`; `hamming` for binary and `l2` for all other. Therefore, I had to assign `UNDEFINED` space type when user does not provide any space type first. Then later inside mapper, I set proper default space type based on data type if the space type is `UNDEFINED`
2. Introduction of `isBinary` parameter in `JNIService.free` method. Because `faiss::Index` and `faiss::BinaryIndex` does not share common parent class, we cannot determine if it is `faiss::Index` or `faiss::BinaryIndex` using cast. Therefore, we need to know if it is binary index or not. The reason why we not having a separate `freeBinaryIndex` method is because it is single delete method in cpp side whereas, `createIndex` has separate method for binary index like `createBinaryIndex`.
3. Adding vector data type as parameter in `IndexEntryContext`. This data will be used when loading an index into memory to determine if it is binary index or not. For creation and query, we use vector data type value in `fieldInfo` to determine if it is binary index or not.
4. New `VectorTransfer`, `KNNIterator`, and `byteQueryVector` variable inside `Query` was needed as binary vector will have `byte[]` type when non binary vector will have `float[]` type. Refactoring will be needed to enhance maintainability which is captured in https://github.com/opensearch-project/k-NN/issues/1810

#### Testing
1. Create index
```
PUT /my-knn-index-1
{
  "settings": {
    "index": {
      "knn": true
    }
  },
  "mappings": {
    "properties": {
      "my_vector": {
        "type": "knn_vector",
        "dimension": 8,
        "data_type": "binary",
        "method": {
          "name": "hnsw",
          "engine": "faiss"
        }
      }
    }
  }
}
```

5. Ingest documents
```
POST /_bulk?refresh=true
{ "index": { "_index": "my-knn-index-1", "_id": "1" } }
{"my_vector" : [9]}
{ "index": { "_index": "my-knn-index-1", "_id": "2" } }
{"my_vector" : [14]}
{ "index": { "_index": "my-knn-index-1", "_id": "3" } }
{"my_vector" : [-100]}
```
6. Query
```
POST /my-knn-index-1/_search
{
  "query": {
    "knn": {
      "my_vector": {
        "vector": [13],
        "k": 2
      }
    }
  }
}
```
Response
```
{
  "took": 81,
  "timed_out": false,
  "_shards": {
    "total": 1,
    "successful": 1,
    "skipped": 0,
    "failed": 0
  },
  "hits": {
    "total": {
      "value": 2,
      "relation": "eq"
    },
    "max_score": 0.5,
    "hits": [
      {
        "_index": "my-knn-index-1",
        "_id": "1",
        "_score": 0.5,
        "_source": {
          "my_vector": [
            9
          ]
        }
      },
      {
        "_index": "my-knn-index-1",
        "_id": "2",
        "_score": 0.33333334,
        "_source": {
          "my_vector": [
            14
          ]
        }
      }
    ]
  }
}
```

 
### Issues Resolved
https://github.com/opensearch-project/k-NN/issues/1767
 
### Check List
  - [X] All tests pass


src/main/java/org/opensearch/knn/index/KNNVectorSimilarityFunction.java
```diff
@@ -0,0 +1,53 @@
...
+
+package org.opensearch.knn.index;
+
+import org.apache.lucene.index.VectorSimilarityFunction;
+import org.opensearch.knn.plugin.script.KNNScoringUtil;
+
+/**
+ * Wrapper class of VectorSimilarityFunction to support more function than what Lucene provides
+ */
+public enum KNNVectorSimilarityFunction {
```
navneet1v, line 14: We should think of a better way here rather than just having a thin wrapper. Should the space type or any other class present in KNN do the vector calculation rather than we telling the caller that they have to call `VectorSimilarityFunction`

src/main/java/org/opensearch/knn/index/SpaceType.java
```diff
@@ -104,18 +116,29 @@ public float scoreTranslation(float rawScore) {
...
     HAMMING_BIT("hammingbit") {
         @Override
         public float scoreTranslation(float rawScore) {
             return 1 / (1 + rawScore);
         }
+
+        @Override
+        public boolean isSupported(VectorDataType vectorDataType) {
+            return true;
+        }
```
navneet1v, line 132: why this is always returning true?

src/main/java/org/opensearch/knn/index/SpaceType.java
```diff
@@ -152,6 +175,10 @@ public void validateVector(float[] vector) {
         // do nothing
     }
 
+    public boolean isSupported(VectorDataType vectorDataType) {
+        return VectorDataType.FLOAT == vectorDataType || VectorDataType.BYTE == vectorDataType;
+    }
```
navneet1v, line 180: java docs? and what is the role of this function?

src/main/java/org/opensearch/knn/index/SpaceType.java
```diff
@@ -172,6 +199,12 @@ public static Set<String> getValues() {
 
     public static SpaceType getSpace(String spaceTypeName) {
         for (SpaceType currentSpaceType : SpaceType.values()) {
+            // UNDEFINED space type is a temporary value to be used only until we set proper default value when
+            // space type is not provided by user
+            // Therefore, we do not allow converting space type name to UNDEFINED space type
+            if (SpaceType.UNDEFINED == currentSpaceType) {
+                continue;
+            }
```
navneet1v, line 207: how will someone validate that space type is UNDEFINED and they need to set the space type?

src/main/java/org/opensearch/knn/index/VectorDataType.java
```diff
@@ -24,11 +24,25 @@
...
+
+        @Override
+        public FieldType createKnnVectorFieldType(int dimension, VectorSimilarityFunction vectorSimilarityFunction) {
+            throw new IllegalArgumentException("This is not supported");
+        }
+
+        @Override
+        public float[] getVectorFromBytesRef(BytesRef binaryValue) {
+            throw new IllegalArgumentException("This is not supported");
+        }
```
navneet1v, line 44: better error messages

src/main/java/org/opensearch/knn/index/codec/util/VectorTransfer.java
```diff
@@ -0,0 +1,32 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.knn.index.codec.util;
```
navneet1v, line 6: this class is not a utility.

src/main/java/org/opensearch/knn/index/codec/util/VectorTransfer.java
```diff
@@ -0,0 +1,32 @@
...
+
+import lombok.Data;
+
+import java.io.ByteArrayInputStream;
+
+@Data
+public abstract class VectorTransfer {
```
navneet1v, line 13: java docs needed.

src/main/java/org/opensearch/knn/index/codec/util/VectorTransfer.java
```diff
@@ -0,0 +1,32 @@
...
+    public VectorTransfer(final long vectorsStreamingMemoryLimit) {
+        this.vectorsStreamingMemoryLimit = vectorsStreamingMemoryLimit;
+        this.vectorsPerTransfer = Integer.MIN_VALUE;
+    }
+
+    abstract public void init(final long totalLiveDocs);
+
+    abstract public void addVector(final ByteArrayInputStream byteStream);
+
+    abstract public void flush();
```
navneet1v, line 29: rename this function to either transfer/stream. flush has different meaning in Opensearch world. So we should avoid it.

src/main/java/org/opensearch/knn/index/codec/util/VectorTransferByte.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    @Override
+    public void init(final long totalLiveDocs) {
+        this.totalLiveDocs = totalLiveDocs;
+        vectorList.clear();
+    }
+
+    @Override
+    public void addVector(final ByteArrayInputStream byteStream) {
```
navneet1v, line 29: the function is doing more than 1 thing, it is adding vectors and also streaming the vectors. We should break this either into 2 functions or rename the function in a fashion that it provides this information that we are doing 2 operations.

src/main/java/org/opensearch/knn/index/codec/util/VectorTransferByte.java
```diff
@@ -0,0 +1,57 @@
...
+
+        vectorList.add(vector);
+        if (vectorList.size() == vectorsPerTransfer) {
+            flush();
+        }
+    }
+
+    @Override
+    public void flush() {
+        vectorAddress = JNICommons.storeByteVectorData(vectorAddress, vectorList.toArray(new byte[][] {}), totalLiveDocs * dimension / 8);
```
navneet1v, line 49: Lets have proper brackets here to ensure that we know what dim needs to be divided first, this will improve the readability. `totalLiveDocs * dimension / 8`

src/main/java/org/opensearch/knn/common/KNNFaissUtil.java
```diff
@@ -0,0 +1,17 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.knn.common;
+
+import java.util.Map;
+
+import static org.opensearch.knn.index.util.Faiss.FAISS_BINARY_INDEX_DESCRIPTION_PREFIX;
+
+public class KNNFaissUtil {
```
navneet1v, line 12: this is not a faiss util. You can always pass the KNNEngine and the parameters map to know if the index is a binary index or not, this will avoid calling this utility as just faiss utility.

src/main/java/org/opensearch/knn/index/query/KNNQuery.java
```diff
@@ -37,9 +38,13 @@ public class KNNQuery extends Query {
 
     private final String field;
     private final float[] queryVector;
+    @Getter
+    private final byte[] byteQueryVector;
```
navneet1v, line 42: can we use <T> here rather than having 2 separate variables for query vectors and byteQueryVector.

src/main/java/org/opensearch/knn/index/query/filtered/FilteredIdsKNNByteIterator.java
```diff
@@ -0,0 +1,79 @@
...
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+
+/**
+ * Inspired by DiversifyingChildrenFloatKnnVectorQuery in lucene
+ * https://github.com/apache/lucene/blob/7b8aece125aabff2823626d5b939abf4747f63a7/lucene/join/src/java/org/apache/lucene/search/join/DiversifyingChildrenFloatKnnVectorQuery.java#L162
+ *
+ * The class is used in KNNWeight to score filtered KNN field by iterating filterIdsArray.
+ */
+public class FilteredIdsKNNByteIterator implements KNNIterator {
```
navneet1v, line 24: you don't need another class here. You can just pass the vector dataType to the iterator and then based on that you can easily find out the what the byte array should be converted to.

Also adding just a <T> in the FilterIds iterator will do the trick for computing the scores using space type.

navneet1v (commented):
Did first round of review for all the src files. Tests are still pending. 

On high level:
1. Add java docs on all the new classes and public functions.
2. Provide the I/O in PR description on how you tested this.
3. Add more details in PR description what this PR is for.

src/main/java/org/opensearch/knn/index/query/filtered/FilteredIdsKNNByteIterator.java
heemin32: Separate class is more cleaner. Otherwise you will need if/else statement here and there. Also, the query type is different between the two. `float[]` and `byte[]`.

src/main/java/org/opensearch/knn/index/query/KNNQuery.java
heemin32: Could you elaborate? Do you want it to be `KNNQuery<T>` or having another `KNNQueryByte` , or `private final Object queryVector`?

src/main/java/org/opensearch/knn/index/KNNMethodContext.java
```diff
@@ -50,19 +52,18 @@ public class KNNMethodContext implements ToXContentFragment, Writeable {
 
     public static synchronized KNNMethodContext getDefault() {
         if (defaultInstance == null) {
-            defaultInstance = new KNNMethodContext(
-                KNNEngine.DEFAULT,
-                SpaceType.DEFAULT,
-                new MethodComponentContext(METHOD_HNSW, Collections.emptyMap())
-            );
+            MethodComponentContext methodComponentContext = new MethodComponentContext(METHOD_HNSW, Collections.emptyMap());
+            methodComponentContext.setIndexVersion(Version.CURRENT);
```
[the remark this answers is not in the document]
navneet1v: can we add the same thing in the comment too. Otherwise it is really hard to understand why we did this. I would really like to think how we can avoid this.

src/main/java/org/opensearch/knn/index/SpaceType.java
```diff
@@ -26,15 +25,28 @@
  * nmslib calls the inner_product space "negdotprod". This translation should take place in the nmslib's jni layer.
  */
 public enum SpaceType {
+    // This undefined space type is used to indicate that space type is not provided by user
+    // Later, we need to assign a default value based on data type
+    UNDEFINED("undefined") {
+        @Override
+        public float scoreTranslation(final float rawScore) {
+            throw new IllegalStateException("This method should not be called with UNDEFINED space type");
```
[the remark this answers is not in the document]
navneet1v: The problem is this exception will be logged even when it is not bubbling up to the user. Given that case if the error message is not correct no one will be able to understand the log message. Lets just improve the wording here and be kind to future ourseleves

src/main/java/org/opensearch/knn/index/SpaceType.java
```diff
@@ -26,15 +25,28 @@
...
+        }
+
+        @Override
+        public boolean isSupported(VectorDataType vectorDataType) {
+            throw new IllegalStateException("This method should not be called with UNDEFINED space type");
```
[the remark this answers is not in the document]
navneet1v: same comment as above

src/main/java/org/opensearch/knn/index/KNNVectorSimilarityFunction.java
```diff
@@ -0,0 +1,53 @@
...
+    EUCLIDEAN(VectorSimilarityFunction.EUCLIDEAN),
+    DOT_PRODUCT(VectorSimilarityFunction.DOT_PRODUCT),
```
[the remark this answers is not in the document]
navneet1v: Caller is k-NN plugin only. So lets just ensure we provide right abstraction

src/main/java/org/opensearch/knn/index/KNNVectorSimilarityFunction.java
navneet1v: If we want to improve in a separate PR lets have a GH issue for this. This PR is already very big and brining a lot of new classes that needs to be thought through if we are just putting everything on future PRs then we might loose the context here.

src/main/java/org/opensearch/knn/index/SpaceType.java
navneet1v: Yes hamming bit was there before we added the support for fields. But Hamming bit is not supported with Floats. The field was not the knn field. Check this Link: https://opensearch.org/docs/latest/search-plugins/knn/knn-score-script/#getting-started-with-the-score-script-for-binary-data

src/main/java/org/opensearch/knn/index/VectorDataType.java
navneet1v: Doesn't matter is this message is user facing or not. Its a message which would be read by operators of Opensearch. So lets ensure the error messages are understandable.

src/main/java/org/opensearch/knn/index/query/KNNQuery.java
navneet1v: I am more of thinking this as `KNNQuery<T>` and not another `KNNQueryByte`

src/main/java/org/opensearch/knn/index/query/filtered/FilteredIdsKNNByteIterator.java
navneet1v: There is unnecessary interfaces which are added and just making this simple iteration logic duplicated across classes. Better idea might be to use <T> here.

src/main/java/org/opensearch/knn/plugin/script/KNNScoringUtil.java
```diff
@@ -195,6 +195,21 @@ public static float calculateHammingBit(Long queryLong, Long inputLong) {
         return Long.bitCount(queryLong ^ inputLong);
     }
 
+    /**
+     * This method calculates hamming distance between query vector
```
navneet1v, line 199: incomplete Java doc.

src/main/java/org/opensearch/knn/index/query/filtered/FilteredIdsKNNByteIterator.java
```diff
@@ -0,0 +1,79 @@
...
+
+    @Override
+    public float score() {
+        return currentScore;
+    }
+
+    protected float computeScore() throws IOException {
+        final BytesRef value = binaryDocValues.binaryValue();
+        final ByteArrayInputStream byteStream = new ByteArrayInputStream(value.bytes, value.offset, value.length);
+        final byte[] vector = byteStream.readAllBytes();
```
navneet1v, line 74: This should be behind the VectorSerializer interfaces otherwise there is no common way to read the binary vectors/float vectors.

src/main/java/org/opensearch/knn/index/KNNMethodContext.java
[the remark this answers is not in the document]
heemin32: What I mean is this `getDefault` method is used in unit test only. The reason that it worked without indexVersion before was there was no MethodFieldMapperTests. Let me just move this entire method inside test package....

src/main/java/org/opensearch/knn/index/codec/transfer/VectorTransferFloat.java
```diff
@@ -0,0 +1,69 @@
...
+import org.opensearch.knn.index.codec.util.KNNVectorSerializerFactory;
+import org.opensearch.knn.index.codec.util.SerializationMode;
+import org.opensearch.knn.jni.JNICommons;
+
+import java.io.ByteArrayInputStream;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Vector transfer for float
```
navneet1v, line 18: better java doc needed

src/main/java/org/opensearch/knn/index/codec/transfer/VectorTransfer.java
```diff
@@ -0,0 +1,55 @@
...
+    protected long vectorAddress;
+    protected int dimension;
+
+    public VectorTransfer(final long vectorsStreamingMemoryLimit) {
+        this.vectorsStreamingMemoryLimit = vectorsStreamingMemoryLimit;
+        this.vectorsPerTransfer = Integer.MIN_VALUE;
+    }
+
+    /**
+     * Initialize the transfer
```
navneet1v, line 30: `Initialize` the transfer of what? the doc is incomplete.

src/main/java/org/opensearch/knn/index/codec/transfer/VectorTransfer.java
```diff
@@ -0,0 +1,55 @@
...
+     *
+     * @param totalLiveDocs total number of vectors to be transferred
+     */
+    abstract public void init(final long totalLiveDocs);
+
+    /**
+     * Transfer a single vector
```
navneet1v, line 37: the documentation is incorrect, we are not transferring 1 vector here. We are transferring more than 1 vector.

src/main/java/org/opensearch/knn/index/codec/transfer/VectorTransfer.java
```diff
@@ -0,0 +1,55 @@
...
+     *
+     * @param byteStream a vector in byte stream format
+     */
+    abstract public void transfer(final ByteArrayInputStream byteStream);
+
+    /**
+     * Close the transfer
+     */
+    abstract public void close();
```
navneet1v, line 46: how we are ensuring that close method is always called?

src/main/java/org/opensearch/knn/index/SpaceType.java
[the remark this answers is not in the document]
heemin32: Updated

src/main/java/org/opensearch/knn/index/SpaceType.java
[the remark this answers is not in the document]
heemin32: Updated

src/main/java/org/opensearch/knn/index/KNNVectorSimilarityFunction.java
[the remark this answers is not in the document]
heemin32: DOT_PRODUCT is not being used. What would be wrong to have both of them? Or I can just remove DOT_PRODUCT entirely but no harm to have it here.

src/main/java/org/opensearch/knn/index/KNNVectorSimilarityFunction.java
heemin32: I am not making anything worse with this thin wrapper I believe. The caller calling `VectorSimilarityFunction` is what we have now.

src/main/java/org/opensearch/knn/index/SpaceType.java
heemin32: Updated the check

src/main/java/org/opensearch/knn/index/VectorDataType.java
heemin32: Updated. If you don't like it please suggest one.

src/main/java/org/opensearch/knn/plugin/script/KNNScoringUtil.java
heemin32: updated

src/main/java/org/opensearch/knn/index/codec/transfer/VectorTransfer.java
heemin32: I thought it is caller's responsibility to call the close. Any idea to enforce it?

src/main/java/org/opensearch/knn/index/codec/transfer/VectorTransfer.java
heemin32: I meant to say this method is called for a single vector.
Yes. We transfer more than 1 vector internally but it is hidden from caller.

src/main/java/org/opensearch/knn/index/codec/transfer/VectorTransfer.java
heemin32: ok.

src/main/java/org/opensearch/knn/index/codec/transfer/VectorTransferFloat.java
heemin32: This is child class of `VectorTransfer`. Wouldn't the java doc in `VectorTransfer` be enough?

src/main/java/org/opensearch/knn/index/codec/transfer/VectorTransfer.java
heemin32: Let me make it as closable class.

heemin32: I think I got your point. I think it should be refactored so that there is no dependency between methods.

heemin32:
@navneet1v I captured the all comments related with code refactoring in https://github.com/opensearch-project/k-NN/issues/1810. Please update in the issue if I missed any. I will follow up after completing this this feature implementation.
The reason that  I want to push this code as quickly as possible is to give enough time for reviewing IVF implementation which has developed on top of this PR so that the complete binary format support can be released in 2.16.

src/main/java/org/opensearch/knn/common/KNNValidationUtil.java
```diff
@@ -71,13 +71,32 @@ public static void validateByteVectorValue(float value) {
...
+     *
      * @param dimension dimension of vector
      * @param vectorSize size of the vector
+     * @param dataType vector data type
      */
-    public static void validateVectorDimension(int dimension, int vectorSize) {
-        if (dimension != vectorSize) {
-            String errorMessage = String.format(Locale.ROOT, "Vector dimension mismatch. Expected: %d, Given: %d", dimension, vectorSize);
-            throw new IllegalArgumentException(errorMessage);
+    public static void validateVectorDimension(final int dimension, final int vectorSize, final VectorDataType dataType) {
```
jmazanec15, line 80: do we need vector size and data type?

src/main/java/org/opensearch/knn/index/IndexUtil.java
```diff
@@ -310,4 +311,17 @@ public static boolean isSharedIndexStateRequired(KNNEngine knnEngine, String mod
...
     }
+
+    /**
+     * Tell if it is binary index or not
+     *
+     * @param knnEngine knn engine associated with an index
+     * @param parameters parameters associated with an index
+     * @return true if it is binary index
+     */
+    public static boolean isBinaryIndex(KNNEngine knnEngine, Map<String, Object> parameters) {
```
jmazanec15, line 322: Cant we check a field if its binary? Issue is that someday we may move away from the index description and manually build the index so I hate to add a dependence on it.

jmazanec15 (commented):
I think for a big PR like this, it would be helpful to give a high level summary of what you had to change in the implementation and why. For instance, explaining the different components changed, like why we are changing the free api to take a boolean. This will make review a lot easier and there will be less back and forth.

src/main/java/org/opensearch/knn/common/KNNValidationUtil.java
heemin32: Yes

src/main/java/org/opensearch/knn/index/SpaceType.java
```diff
@@ -104,18 +116,38 @@ public float scoreTranslation(float rawScore) {
         }
 
         @Override
-        public VectorSimilarityFunction getVectorSimilarityFunction() {
-            return VectorSimilarityFunction.MAXIMUM_INNER_PRODUCT;
+        public KNNVectorSimilarityFunction getKnnVectorSimilarityFunction() {
+            return KNNVectorSimilarityFunction.MAXIMUM_INNER_PRODUCT;
         }
     },
     HAMMING_BIT("hammingbit") {
```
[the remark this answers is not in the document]
heemin32: It would break backward compatibility with existing score script

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapperUtil.java
```diff
@@ -131,10 +170,10 @@ public static void validateVectorDataTypeWithKnnIndexSetting(
             throw new IllegalArgumentException(
                 String.format(
                     Locale.ROOT,
-                    "[%s] field with value [%s] is only supported for [%s] engine",
+                    "[%s] field with value [%s] is not supported for [%s] engine",
                     VECTOR_DATA_TYPE_FIELD,
                     vectorDataType.getValue().getValue(),
-                    LUCENE_NAME
+                    NMSLIB_NAME
```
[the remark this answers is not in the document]
heemin32: No. Exception message had to be updated because data_type field can be used for faiss engine now.

src/main/java/org/opensearch/knn/index/memory/NativeMemoryAllocation.java
```diff
@@ -171,7 +179,7 @@ private void cleanup() {
 
             // memoryAddress is sometimes initialized to 0. If this is ever the case, freeing will surely fail.
             if (memoryAddress != 0) {
-                JNIService.free(memoryAddress, knnEngine);
+                JNIService.free(memoryAddress, knnEngine, isBinaryIndex);
```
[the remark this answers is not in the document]
heemin32: Cast does not work. `reinterpret_cast` can be used only for child class to determine its type. `dynamic_cast` cannot do type check.
BinaryIndex and Index in faiss does not share common parent class.

src/main/java/org/opensearch/knn/index/util/Faiss.java
```diff
@@ -246,7 +247,7 @@ class Faiss extends NativeLibrary {
                     ).addParameter(METHOD_PARAMETER_M, "", "").addParameter(METHOD_ENCODER_PARAMETER, ",", "").build())
                 )
                 .build()
-        ).addSpaces(SpaceType.L2, SpaceType.INNER_PRODUCT).build(),
+        ).addSpaces(SpaceType.UNDEFINED, SpaceType.HAMMING_BIT, SpaceType.L2, SpaceType.INNER_PRODUCT).build(),
```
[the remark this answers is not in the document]
heemin32: When user does not define space type, it is set as UNDEFINED first. It is needed so that we can have different default space type based on data_type in the later phase.

src/main/java/org/opensearch/knn/index/IndexUtil.java
heemin32: Sure. Let me update the code.

heemin32: Updated.

navneet1v:
> @navneet1v I captured the all comments related with code refactoring in #1810. Please update in the issue if I missed any. I will follow up after completing this feature implementation. The reason that I want to push this code as quickly as possible without doing all the suggested refactoring is to give enough time for reviewing IVF implementation which has developed on top of this PR so that the complete binary format support can be released in 2.16.

I think there are multiple refactoring suggested in the code and my worry is there are new features that we are working on including [Disk based vector search](https://github.com/opensearch-project/k-NN/issues/1779) and [reducing memory foot print](https://github.com/opensearch-project/k-NN/issues/1600) during index creation they all touch the same code path. Doing refactoring later will impact those features.

src/main/java/org/opensearch/knn/common/KNNValidationUtil.java
jmazanec15: why?

heemin32:
> > @navneet1v I captured the all comments related with code refactoring in #1810. Please update in the issue if I missed any. I will follow up after completing this feature implementation. The reason that I want to push this code as quickly as possible without doing all the suggested refactoring is to give enough time for reviewing IVF implementation which has developed on top of this PR so that the complete binary format support can be released in 2.16.
> 
> I think there are multiple refactoring suggested in the code and my worry is there are new features that we are working on including [Disk based vector search](https://github.com/opensearch-project/k-NN/issues/1779) and [reducing memory foot print](https://github.com/opensearch-project/k-NN/issues/1600) during index creation they all touch the same code path. Doing refactoring later will impact those features.

We can discuss offline. My plan is working on refactoring after 2.16 release. Therefore, the delay won't be much between now and later.

src/main/java/org/opensearch/knn/common/KNNValidationUtil.java
heemin32: Talked in offline.

jmazanec15 (approved):

navneet1v (approved):

heemin32:
Ran `./gradlew :spotlessApply` to fix build failure.

PR #1783, opened 2024-07-02
Adds ef_search as a query parameter for Lucene, FAISS and NMSLIB

shatejas:
Currently ef_search is set at index level, this change gives the ability to have query time ef-search parameter without manipulating the index settings. query time value supersedes the value from index in FAISS and NMSLIB. For Lucene, max of k and ef_search is used as ef_search value

### Description
- Squashed commits from feature branch
     - Commits not included: Radial search and Nprobes related. Those will be squashed and put on top of these
- Changed version from 3 to 2.16 for ef_search and method_parameters
 
### Issues Resolved
[1537](https://github.com/opensearch-project/k-NN/issues/1537)
 
### Check List
  - [ ] All tests pass


jni/src/faiss_wrapper.cpp
```diff
@@ -24,6 +24,7 @@
 #include "faiss/IndexIVFPQ.h"
 
 #include <algorithm>
+#include <commons.h>
```
jmazanec15, line 27: "<>" usually indicates some kind of system header. Can we switch to "commons.h". See google style guide: https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes

jni/src/nmslib_wrapper.cpp
```diff
@@ -12,6 +12,8 @@
 #include "jni_util.h"
 #include "nmslib_wrapper.h"
 
+#include <commons.h>
```
jmazanec15, line 15: same as above

jni/src/nmslib_wrapper.cpp
```diff
@@ -243,14 +247,24 @@ jobjectArray knn_jni::nmslib_wrapper::QueryIndex(knn_jni::JNIUtilInterface * jni
...
+    }
 
-    similarity::KNNQuery<float> knnQuery(*(indexWrapper->space), queryObject.get(), kJ);
-    indexWrapper->index->Search(&knnQuery);
+    int queryEfSearch = knn_jni::commons::getIntegerMethodParameter(env, jniUtil, methodParams, EF_SEARCH, -1);
+    similarity::KNNQuery<float>* query;
+    if (queryEfSearch == -1) {
+        query = new similarity::KNNQuery<float>(*(indexWrapper->space), queryObject.get(), kJ);
+    } else {
+        query = new similarity::HNSWQuery<float>(*(indexWrapper->space), queryObject.get(), kJ, queryEfSearch);
```
jmazanec15, line 262: Dont we need eps? https://github.com/opensearch-project/k-NN/blob/9db146baba5804f3a49afdf3eb2c3de3ac9d1cb7/jni/patches/nmslib/0002-Adds-ability-to-pass-ef-parameter-in-the-query-for-h.patch#L121

jni/src/nmslib_wrapper.cpp
```diff
@@ -243,14 +247,24 @@ jobjectArray knn_jni::nmslib_wrapper::QueryIndex(knn_jni::JNIUtilInterface * jni
...
     jniUtil->ReleaseFloatArrayElements(env, queryVectorJ, rawQueryvector, JNI_ABORT);
+    std::unordered_map<std::string, jobject> methodParams;
+    if (methodParamsJ != nullptr) {
+        methodParams = jniUtil->ConvertJavaMapToCppMap(env, methodParamsJ);
+    }
 
-    similarity::KNNQuery<float> knnQuery(*(indexWrapper->space), queryObject.get(), kJ);
-    indexWrapper->index->Search(&knnQuery);
+    int queryEfSearch = knn_jni::commons::getIntegerMethodParameter(env, jniUtil, methodParams, EF_SEARCH, -1);
+    similarity::KNNQuery<float>* query;
```
jmazanec15, line 258: Can we use smart pointer to avoid having to delete at end?

jni/tests/faiss_wrapper_test.cpp
```diff
@@ -282,6 +287,7 @@ TEST(FaissQueryIndexTest, BasicAssertions) {
             delete it;
         }
     }
+    std::cout << "Test end";
```
jmazanec15, line 290: Do we need this?

jni/tests/faiss_wrapper_unit_test.cpp
```diff
@@ -0,0 +1,251 @@
...
+    FaissWrappeterParametrizedTestFixture() : index_(3), id_map_(&index_) {
+        index_.hnsw.efSearch = 100; // assigning 100 to make sure default of 16 is not used anywhere
+    };
+
+protected:
+    MockIndex index_;
+    MockIdMap id_map_;
+};
+
+namespace query_index_test {
```
jmazanec15, line 92: out of curiousity, what is reasoning for namespaces here?

jmazanec15 (commented):
Few comments on cpp side. But overall looks good. Going to look at Java side later.

jmazanec15 (commented):
Reviewed Java side. Looks good.

jni/src/nmslib_wrapper.cpp
heemin32: +1

src/test/java/org/opensearch/knn/index/query/KNNQueryBuilderTests.java
```diff
@@ -615,23 +792,61 @@ public void testDoToQuery_whenDoRadiusSearch_whenScoreThreshold_whenFilter_thenS
...
+
         Query query = knnQueryBuilder.doToQuery(mockQueryShardContext);
+
+        // Then
         assertNotNull(query);
         assertTrue(query.getClass().isAssignableFrom(KNNQuery.class));
+        assertEquals(HNSW_METHOD_PARAMS, ((KNNQuery) query).getMethodParameters());
     }
 
+    /** This test should be uncommented once we have nprobs. Considering engine instance is static its not possible to test this right now
```
heemin32, line 828: Should we just remove this test case? Do we have nprobs but cannot write a test because engine instance is static or we don't have nprobs yet?

jni/src/nmslib_wrapper.cpp
shatejas: There is a default present and we didn't seem to be using the defaults for KNNQuery, so kept the same behavior for hnsw. Can you elaborate why do you think we need it?

heemin32: I think @jmazanec15 missed this line where default value is defined. 
https://github.com/opensearch-project/k-NN/blob/9db146baba5804f3a49afdf3eb2c3de3ac9d1cb7/jni/patches/nmslib/0002-Adds-ability-to-pass-ef-parameter-in-the-query-for-h.patch#L50

jni/tests/faiss_wrapper_unit_test.cpp
shatejas: If I remember correctly, this allowed me to have multiple parametrized tests in the same file since I am using the same fixture for both. Could have gotten away with it if I was using different fixtures

jni/tests/faiss_wrapper_test.cpp
shatejas: not really. Seems to be a remnant of debugging

jni/src/nmslib_wrapper.cpp
heemin32: I think @jmazanec15 missed this line where default is defined?
https://github.com/opensearch-project/k-NN/blob/9db146baba5804f3a49afdf3eb2c3de3ac9d1cb7/jni/patches/nmslib/0002-Adds-ability-to-pass-ef-parameter-in-the-query-for-h.patch#L50

jmazanec15: Will this be using the HNSWQuery constructor? Or will it fall through to the parent class and somehow cast ef to eps? My concern here is that this signature doesnt look like it would support a default parameter for eps in c++:
```
HNSWQuery<dist_t>::HNSWQuery(const Space<dist_t> &space, const Object* query_object, const unsigned K, unsigned ef, float eps) 
```
I would expect defaults to be like `float eps = 0.0f`.

jni/src/nmslib_wrapper.cpp
shatejas: As far as I remember, I had a really hard time getting the unit test right with a smart pointer so had to switch to this. Suggestions are welcome here. I would like to keep it as it is otherwise

jni/src/nmslib_wrapper.cpp
jmazanec15: Im not sure how for HnswQuery(, the eps parameter can be inferred as optional based on superclass.

jni/src/nmslib_wrapper.cpp
shatejas: Will change

shatejas (commented):
This is simply merging the feature branch after squashing the commits. Just FYI if it wasn't obvious from the description

jni/src/nmslib_wrapper.cpp
jmazanec15: Oh I see you defaulted here: https://github.com/opensearch-project/k-NN/blob/9db146baba5804f3a49afdf3eb2c3de3ac9d1cb7/jni/patches/nmslib/0002-Adds-ability-to-pass-ef-parameter-in-the-query-for-h.patch#L50

jmazanec15: Thats good then.

Why did we set ef = 10 by default?

shatejas: I was attempting to integrate with the integration test in nmslib. I failed since its way too rigid to add additional parameters. As a part of it I had to give a default and gave it 10 and then kept it. I can change it if needed

jmazanec15: Will it ever fall through to default? If not, thats okay. If yes, then we should sync it with our current defaults (100)

jni/src/nmslib_wrapper.cpp
heemin32: Can we create a GH issue to refactor the code so that we can write unit test with smart pointer?

jni/src/nmslib_wrapper.cpp
shatejas: It won't as of now but I will change it 100 to be defensive wrt future code changes

jni/src/nmslib_wrapper.cpp
shatejas: https://github.com/opensearch-project/k-NN/issues/1785

jmazanec15 (approved):

heemin32 (approved):

Issue #1773, opened 2024-06-25
Refactoring of query method in JNI


heemin32:
There are two query method in jni layer. `knn_jni::faiss_wrapper::QueryBinaryIndex_WithFilter`, `knn_jni::faiss_wrapper::QueryIndex_WithFilter`. 

We want to refactor the code so that it can be easier to test and reduce code duplication.

PR #1784, opened 2024-07-02
Add binary format support with IVF method in Faiss Engine

junqiu-lei:
### Description
This PR will support using binary format with Faiss IVF method, it mainly have changes:
* introduce `data_type` field when train model
* support train model with binary format index
* create binary format IVF target index with with binary format model
* query IVF index with binary format

JNI layer related refactor works will be complete in another PR tracked by https://github.com/opensearch-project/k-NN/issues/1846

#### 1. Create binary format train index
```
PUT /train-index HTTP/1.1
Host: localhost:9200
Content-Type: application/json
Content-Length: 481

{
  "settings": {
    "number_of_shards": 3,
    "number_of_replicas": 0,
    "index.knn": true
  },
  "mappings": {
    "properties": {
      "train-field": {
        "type": "knn_vector",
        "dimension": 8,
        "data_type": "binary",
        "method": {
          "name": "hnsw",
          "space_type": "hammingbit",
          "engine": "faiss",
          "parameters": {
            "ef_construction": 128,
            "m": 24
          }
        }
      }
    }
  }
}
```

#### 2. Ingest tran index
```
POST /_bulk HTTP/1.1
Host: localhost:9200
Content-Type: application/json
Content-Length: 7067

{ "index": { "_index": "train-index", "_id": "1" } }
{ "train-field": [1]}
{ "index": { "_index": "train-index", "_id": "2" } }
{ "train-field": [2]}
{ "index": { "_index": "train-index", "_id": "3" } }
{ "train-field": [4]}
{ "index": { "_index": "train-index", "_id": "4" } }
........
```

#### 3. Create train model
```
POST /_plugins/_knn/models/my-model/_train HTTP/1.1
Host: localhost:9200
Content-Type: application/json
Content-Length: 313

{
  "training_index": "train-index",
  "training_field": "train-field",
  "dimension": 8,
  "description": "My model description",
  "data_type": "binary",
  "method": {
    "name": "ivf",
    "engine": "faiss",
    "space_type": "hammingbit",
    "parameters": {
      "nlist": 4,
      "nprobes":2 
    }
  }
}

```

#### 4. Create IVF binary format target index
```
PUT /target-index HTTP/1.1
Host: localhost:9200
Content-Type: application/json
Content-Length: 242

{
  "settings": {
    "number_of_shards": 1,
    "number_of_replicas": 1,
    "index.knn": true
  },
  "mappings": {
    "properties": {
      "target-field": {
        "type": "knn_vector",
        "model_id": "my-model"
      }
    }
  }
}
```

#### 5. Bulk target index
```
POST /_bulk HTTP/1.1
Host: localhost:9200
Content-Type: application/json
Content-Length: 931

{ "index": { "_index": "target-index", "_id": "1" } }
{ "target-field": [2]}
{ "index": { "_index": "target-index", "_id": "2" } }
{ "target-field": [3]}
.......
```

#### 6. Query target index
```
GET /target-index/_search HTTP/1.1
Host: localhost:9200
Content-Type: application/json
Content-Length: 110

{
  "query": {
    "knn": {
      "target-field": {
        "vector": [10],
        "k": 5
      }
    }
  }
}
```

#### 7. Query result
```
{
    "took": 2898,
    "timed_out": false,
    "_shards": {
        "total": 1,
        "successful": 1,
        "skipped": 0,
        "failed": 0
    },
    "hits": {
        "total": {
            "value": 5,
            "relation": "eq"
        },
        "max_score": 1.0,
        "hits": [
            {
                "_index": "target-index",
                "_id": "9",
                "_score": 1.0,
                "_source": {
                    "target-field": [
                        10
                    ]
                }
            },
            {
                "_index": "target-index",
                "_id": "1",
                "_score": 0.5,
                "_source": {
                    "target-field": [
                        2
                    ]
                }
            },
            {
                "_index": "target-index",
                "_id": "7",
                "_score": 0.5,
                "_source": {
                    "target-field": [
                        8
                    ]
                }
            },
            {
                "_index": "target-index",
                "_id": "10",
                "_score": 0.5,
                "_source": {
                    "target-field": [
                        11
                    ]
                }
            },
            {
                "_index": "target-index",
                "_id": "2",
                "_score": 0.33333334,
                "_source": {
                    "target-field": [
                        3
                    ]
                }
            }
        ]
    }
}
```

### Issues Resolved
part of https://github.com/opensearch-project/k-NN/issues/1767
 
### Check List
  - [x] All tests pass

### DIFF
--- a/jni/include/faiss_wrapper.h
+++ b/jni/include/faiss_wrapper.h
@@ -29,6 +29,12 @@ namespace knn_jni {
                                      jlong vectorsAddressJ, jint dimJ, jstring indexPathJ, jbyteArray templateIndexJ,
                                      jobject parametersJ);
 
+        // Create an index with ids and vectors. Instead of creating a new index, this function creates the index
+        // based off of the template index passed in. The index is serialized to indexPathJ.
+        void CreateBinaryIndexFromTemplate(knn_jni::JNIUtilInterface * jniUtil, JNIEnv * env, jintArray idsJ,
+                                     jlong vectorsAddressJ, jint dimJ, jstring indexPathJ, jbyteArray templateIndexJ,
+                                     jobject parametersJ);
+
         // Load an index from indexPathJ into memory.
         //
         // Return a pointer to the loaded index
@@ -95,6 +101,13 @@ namespace knn_jni {
         // Return the serialized representation
         jbyteArray TrainIndex(knn_jni::JNIUtilInterface * jniUtil, JNIEnv * env, jobject parametersJ, jint dimension,
                               jlong trainVectorsPointerJ);
+
+        // Create an empty binary index defined by the values in the Java map, parametersJ. Train the index with
+        // the vector of floats located at trainVectorsPointerJ.
+        //
+        // Return the serialized representation
+        jbyteArray TrainBinaryIndex(knn_jni::JNIUtilInterface * jniUtil, JNIEnv * env, jobject parametersJ, jint dimension,
+                         jlong trainVectorsPointerJ);
 
         /*
          * Perform a range search with filter against the index located in memory at indexPointerJ.
--- a/jni/src/faiss_wrapper.cpp
+++ b/jni/src/faiss_wrapper.cpp
@@ -70,6 +70,9 @@ void SetExtraParameters(knn_jni::JNIUtilInterface * jniUtil, JNIEnv *env,
 // Train an index with data provided
 void InternalTrainIndex(faiss::Index * index, faiss::idx_t n, const float* x);
 
+// Train a binary index with data provided
+void InternalTrainBinaryIndex(faiss::IndexBinary * index, faiss::idx_t n, const float* x);
+
 // Converts the int FilterIds to Faiss ids type array.
 void convertFilterIdsToFaissIdType(const int* filterIds, int filterIdsLength, faiss::idx_t* convertedFilterIds);
 
@@ -221,6 +224,76 @@ void knn_jni::faiss_wrapper::CreateIndexFromTemplate(knn_jni::JNIUtilInterface *
     // Write the index to disk
     std::string indexPathCpp(jniUtil->ConvertJavaStringToCppString(env, indexPathJ));
     faiss::write_index(&idMap, indexPathCpp.c_str());
+}
+
+void knn_jni::faiss_wrapper::CreateBinaryIndexFromTemplate(knn_jni::JNIUtilInterface * jniUtil, JNIEnv * env, jintArray idsJ,
+                                                     jlong vectorsAddressJ, jint dimJ, jstring indexPathJ,
+                                                     jbyteArray templateIndexJ, jobject parametersJ) {
+    if (idsJ == nullptr) {
+        throw std::runtime_error("IDs cannot be null");
+    }
+
+    if (vectorsAddressJ <= 0) {
+        throw std::runtime_error("VectorsAddress cannot be less than 0");
+    }
+
+    if(dimJ <= 0) {
+        throw std::runtime_error("Vectors dimensions cannot be less than or equal to 0");
+    }
+
+    if (indexPathJ == nullptr) {
+        throw std::runtime_error("Index path cannot be null");
+    }
+
+    if (templateIndexJ == nullptr) {
+        throw std::runtime_error("Template index cannot be null");
+    }
+
+    // Set thread count if it is passed in as a parameter. Setting this variable will only impact the current thread
+    auto parametersCpp = jniUtil->ConvertJavaMapToCppMap(env, parametersJ);
+    if(parametersCpp.find(knn_jni::INDEX_THREAD_QUANTITY) != parametersCpp.end()) {
+        auto threadCount = jniUtil->ConvertJavaObjectToCppInteger(env, parametersCpp[knn_jni::INDEX_THREAD_QUANTITY]);
+        omp_set_num_threads(threadCount);
+    }
+    jniUtil->DeleteLocalRef(env, parametersJ);
+
+    // Read data set
+    // Read vectors from memory address
+    auto *inputVectors = reinterpret_cast<std::vector<uint8_t>*>(vectorsAddressJ);
+    int dim = (int)dimJ;
+    if (dim % 8 != 0) {
+        throw std::runtime_error("Dimensions should be multiply of 8");
+    }
+    int numVectors = (int) (inputVectors->size() / (uint64_t) (dim / 8));
+    int numIds = jniUtil->GetJavaIntArrayLength(env, idsJ);
+    if (numIds != numVectors) {
+        throw std::runtime_error("Number of IDs does not match number of vectors");
+    }
+
+    // Get vector of bytes from jbytearray
+    int indexBytesCount = jniUtil->GetJavaBytesArrayLength(env, templateIndexJ);
+    jbyte * indexBytesJ = jniUtil->GetByteArrayElements(env, templateIndexJ, nullptr);
+
+    faiss::VectorIOReader vectorIoReader;
+    for (int i = 0; i < indexBytesCount; i++) {
+        vectorIoReader.data.push_back((uint8_t) indexBytesJ[i]);
+    }
+    jniUtil->ReleaseByteArrayElements(env, templateIndexJ, indexBytesJ, JNI_ABORT);
+
+    // Create faiss index
+    std::unique_ptr<faiss::IndexBinary> indexWriter;
+    indexWriter.reset(faiss::read_index_binary(&vectorIoReader, 0));
+
+    auto idVector = jniUtil->ConvertJavaIntArrayToCppIntVector(env, idsJ);
+    faiss::IndexBinaryIDMap idMap =  faiss::IndexBinaryIDMap(indexWriter.get());
+    idMap.add_with_ids(numVectors, reinterpret_cast<const uint8_t*>(inputVectors->data()), idVector.data());
+    // Releasing the vectorsAddressJ memory as that is not required once we have created the index.
+    // This is not the ideal approach, please refer this gh issue for long term solution:
+    // https://github.com/opensearch-project/k-NN/issues/1600
+    delete inputVectors;
+    // Write the index to disk
+    std::string indexPathCpp(jniUtil->ConvertJavaStringToCppString(env, indexPathJ));
+    faiss::write_index_binary(&idMap, indexPathCpp.c_str());
 }
 
 jlong knn_jni::faiss_wrapper::LoadIndex(knn_jni::JNIUtilInterface * jniUtil, JNIEnv * env, jstring indexPathJ) {
@@ -624,6 +697,57 @@ jbyteArray knn_jni::faiss_wrapper::TrainIndex(knn_jni::JNIUtilInterface * jniUti
     return ret;
 }
 
+jbyteArray knn_jni::faiss_wrapper::TrainBinaryIndex(knn_jni::JNIUtilInterface * jniUtil, JNIEnv * env, jobject parametersJ,
+                                              jint dimensionJ, jlong trainVectorsPointerJ) {
+    // First, we need to build the index
+    if (parametersJ == nullptr) {
+        throw std::runtime_error("Parameters cannot be null");
+    }
+
+    auto parametersCpp = jniUtil->ConvertJavaMapToCppMap(env, parametersJ);
+
+    jobject spaceTypeJ = knn_jni::GetJObjectFromMapOrThrow(parametersCpp, knn_jni::SPACE_TYPE);
+    std::string spaceTypeCpp(jniUtil->ConvertJavaObjectToCppString(env, spaceTypeJ));
+    faiss::MetricType metric = TranslateSpaceToMetric(spaceTypeCpp);
+
+    // Create faiss index
+    jobject indexDescriptionJ = knn_jni::GetJObjectFromMapOrThrow(parametersCpp, knn_jni::INDEX_DESCRIPTION);
+    std::string indexDescriptionCpp(jniUtil->ConvertJavaObjectToCppString(env, indexDescriptionJ));
+
+    std::unique_ptr<faiss::IndexBinary> indexWriter;
+    indexWriter.reset(faiss::index_binary_factory((int) dimensionJ, indexDescriptionCpp.c_str()));
+
+    // Set thread count if it is passed in as a parameter. Setting this variable will only impact the current thread
+    if(parametersCpp.find(knn_jni::INDEX_THREAD_QUANTITY) != parametersCpp.end()) {
+        auto threadCount = jniUtil->ConvertJavaObjectToCppInteger(env, parametersCpp[knn_jni::INDEX_THREAD_QUANTITY]);
+        omp_set_num_threads(threadCount);
+    }
+
+    // Train index if needed
+    auto *trainingVectorsPointerCpp = reinterpret_cast<std::vector<float>*>(trainVectorsPointerJ);
+    int numVectors = trainingVectorsPointerCpp->size()/(int) dimensionJ;
+    if(!indexWriter->is_trained) {
+        InternalTrainBinaryIndex(indexWriter.get(), numVectors, trainingVectorsPointerCpp->data());
+    }
+    jniUtil->DeleteLocalRef(env, parametersJ);
+
+    // Now that indexWriter is trained, we just load the bytes into an array and return
+    faiss::VectorIOWriter vectorIoWriter;
+    faiss::write_index_binary(indexWriter.get(), &vectorIoWriter);
+
+    // Wrap in smart pointer
+    std::unique_ptr<jbyte[]> jbytesBuffer;
+    jbytesBuffer.reset(new jbyte[vectorIoWriter.data.size()]);
+    int c = 0;
+    for (auto b : vectorIoWriter.data) {
+        jbytesBuffer[c++] = (jbyte) b;
+    }
+
+    jbyteArray ret = jniUtil->NewByteArray(env, vectorIoWriter.data.size());
+    jniUtil->SetByteArrayRegion(env, ret, 0, vectorIoWriter.data.size(), jbytesBuffer.get());
+    return ret;
+}
+
 faiss::MetricType TranslateSpaceToMetric(const std::string& spaceType) {
     if (spaceType == knn_jni::L2) {
         return faiss::METRIC_L2;
@@ -679,6 +803,15 @@ void InternalTrainIndex(faiss::Index * index, faiss::idx_t n, const float* x) {
 
     if (!index->is_trained) {
         index->train(n, x);
+    }
+}
+
+void InternalTrainBinaryIndex(faiss::IndexBinary * index, faiss::idx_t n, const float* x) {
+    if (auto * indexIvf = dynamic_cast<faiss::IndexBinaryIVF*>(index)) {
+        indexIvf->make_direct_map();
+    }
+    if (!index->is_trained) {
+        index->train(n, reinterpret_cast<const uint8_t*>(x));
     }
 }
 
--- a/src/main/java/org/opensearch/knn/index/IndexUtil.java
+++ b/src/main/java/org/opensearch/knn/index/IndexUtil.java
@@ -48,27 +48,9 @@ public class IndexUtil {
     private static final Version MINIMAL_SUPPORTED_VERSION_FOR_MODEL_METHOD_COMPONENT_CONTEXT = Version.V_2_13_0;
     private static final Version MINIMAL_SUPPORTED_VERSION_FOR_RADIAL_SEARCH = Version.V_2_14_0;
     private static final Version MINIMAL_SUPPORTED_VERSION_FOR_METHOD_PARAMETERS = Version.V_2_16_0;
+    private static final Version MINIMAL_SUPPORTED_VERSION_FOR_MODEL_VECTOR_DATA_TYPE = Version.V_2_16_0;
     // public so neural search can access it
     public static final Map<String, Version> minimalRequiredVersionMap = initializeMinimalRequiredVersionMap();
-
-    private static Map<String, Version> initializeMinimalRequiredVersionMap() {
-        final Map<String, Version> versionMap = new HashMap<>() {
-            {
-                put("ignore_unmapped", MINIMAL_SUPPORTED_VERSION_FOR_IGNORE_UNMAPPED);
-                put(MODEL_NODE_ASSIGNMENT_KEY, MINIMAL_SUPPORTED_VERSION_FOR_MODEL_NODE_ASSIGNMENT);
-                put(MODEL_METHOD_COMPONENT_CONTEXT_KEY, MINIMAL_SUPPORTED_VERSION_FOR_MODEL_METHOD_COMPONENT_CONTEXT);
-                put(KNNConstants.RADIAL_SEARCH_KEY, MINIMAL_SUPPORTED_VERSION_FOR_RADIAL_SEARCH);
-                put(KNNConstants.METHOD_PARAMETER, MINIMAL_SUPPORTED_VERSION_FOR_METHOD_PARAMETERS);
-            }
-        };
-
-        for (final MethodParameter methodParameter : MethodParameter.values()) {
-            if (methodParameter.getVersion() != null) {
-                versionMap.put(methodParameter.getName(), methodParameter.getVersion());
-            }
-        }
-        return Collections.unmodifiableMap(versionMap);
-    }
 
     /**
      * Determines the size of a file on disk in kilobytes
@@ -89,37 +71,6 @@ public class IndexUtil {
     }
 
     /**
-     * This method retrieves the field mapping by a given field path from the index metadata.
-     *
-     * @param properties Index metadata mapping properties.
-     * @param fieldPath The field path string that make up the path to the field mapping. e.g. "a.b.field" or "field".
-     *                  The field path is applied and checked in OpenSearch, so it is guaranteed to be valid.
-     *
-     * @return           The field mapping object if found, or null if the field is not found in the index metadata.
-     */
-    private static Object getFieldMapping(final Map<String, Object> properties, final String fieldPath) {
-        String[] fieldPaths = fieldPath.split("\\.");
-        Object currentFieldMapping = properties;
-
-        // Iterate through the field path list to retrieve the field mapping.
-        for (String path : fieldPaths) {
-            currentFieldMapping = ((Map<String, Object>) currentFieldMapping).get(path);
-            if (currentFieldMapping == null) {
-                return null;
-            }
-
-            if (currentFieldMapping instanceof Map<?, ?>) {
-                Object possibleProperties = ((Map<String, Object>) currentFieldMapping).get("properties");
-                if (possibleProperties instanceof Map<?, ?>) {
-                    currentFieldMapping = possibleProperties;
-                }
-            }
-        }
-
-        return currentFieldMapping;
-    }
-
-    /**
      * Validate that a field is a k-NN vector field and has the expected dimension
      *
      * @param indexMetadata metadata for index to validate
@@ -135,7 +86,8 @@ public class IndexUtil {
         IndexMetadata indexMetadata,
         String field,
         int expectedDimension,
-        ModelDao modelDao
+        ModelDao modelDao,
+        VectorDataType expectedVectorDataType
     ) {
         // Index metadata should not be null
         if (indexMetadata == null) {
@@ -188,6 +140,29 @@ public class IndexUtil {
         if (!(type instanceof String) || !KNNVectorFieldMapper.CONTENT_TYPE.equals(type)) {
             exception.addValidationError(String.format("Field \"%s\" is not of type %s.", field, KNNVectorFieldMapper.CONTENT_TYPE));
             return exception;
+        }
+
+        if (expectedVectorDataType != null) {
+            if (VectorDataType.BYTE == expectedVectorDataType) {
+                exception.addValidationError(
+                    String.format(Locale.ROOT, "vector data type \"%s\" is not supported for training.", expectedVectorDataType.getValue())
+                );
+                return exception;
+            }
+            VectorDataType trainIndexDataType = getVectorDataTypeFromFieldMapping(fieldMap);
+
+            if (trainIndexDataType != expectedVectorDataType) {
+                exception.addValidationError(
+                    String.format(
+                        Locale.ROOT,
+                        "Field \"%s\" has data type %s, which is different from data type used in the training request: %s",
+                        field,
+                        trainIndexDataType.getValue(),
+                        expectedVectorDataType.getValue()
+                    )
+                );
+                return exception;
+            }
         }
 
         // Return if dimension does not need to be checked
@@ -320,5 +295,95 @@ public class IndexUtil {
         return KNNEngine.FAISS == knnEngine
             && parameters.get(VECTOR_DATA_TYPE_FIELD) != null
             && parameters.get(VECTOR_DATA_TYPE_FIELD).toString().equals(VectorDataType.BINARY.getValue());
+    }
+
+    /**
+     * Tell if it is binary index or not
+     *
+     * @param vectorDataType vector data type
+     * @return true if it is binary index
+     */
+    public static boolean isBinaryIndex(VectorDataType vectorDataType) {
+        return VectorDataType.BINARY == vectorDataType;
+    }
+
+    /**
+     * Update vector data type into parameters
+     *
+     * @param parameters parameters associated with an index
+     * @param vectorDataType vector data type
+     */
+    public static void updateVectorDataTypeToParameters(Map<String, Object> parameters, VectorDataType vectorDataType) {
+        if (VectorDataType.BINARY == vectorDataType) {
+            parameters.put(VECTOR_DATA_TYPE_FIELD, VectorDataType.BINARY.getValue());
+        }
+    }
+
+    /**
+     * This method retrieves the field mapping by a given field path from the index metadata.
+     *
+     * @param properties Index metadata mapping properties.
+     * @param fieldPath The field path string that make up the path to the field mapping. e.g. "a.b.field" or "field".
+     *                  The field path is applied and checked in OpenSearch, so it is guaranteed to be valid.
+     *
+     * @return           The field mapping object if found, or null if the field is not found in the index metadata.
+     */
+    private static Object getFieldMapping(final Map<String, Object> properties, final String fieldPath) {
+        String[] fieldPaths = fieldPath.split("\\.");
+        Object currentFieldMapping = properties;
+
+        // Iterate through the field path list to retrieve the field mapping.
+        for (String path : fieldPaths) {
+            currentFieldMapping = ((Map<String, Object>) currentFieldMapping).get(path);
+            if (currentFieldMapping == null) {
+                return null;
+            }
+
+            if (currentFieldMapping instanceof Map<?, ?>) {
+                Object possibleProperties = ((Map<String, Object>) currentFieldMapping).get("properties");
+                if (possibleProperties instanceof Map<?, ?>) {
+                    currentFieldMapping = possibleProperties;
+                }
+            }
+        }
+
+        return currentFieldMapping;
+    }
+
+    /**
+     *  This method is used to get the vector data type from field mapping
+     * @param fieldMap field mapping
+     * @return vector data type
+     */
+    private static VectorDataType getVectorDataTypeFromFieldMapping(Map<String, Object> fieldMap) {
+        if (fieldMap.containsKey(VECTOR_DATA_TYPE_FIELD)) {
+            return VectorDataType.get((String) fieldMap.get(VECTOR_DATA_TYPE_FIELD));
+        }
+        return VectorDataType.DEFAULT;
+    }
+
+    /**
+     * Initialize the minimal required version map
+     *
+     * @return minimal required version map
+     */
+    private static Map<String, Version> initializeMinimalRequiredVersionMap() {
+        final Map<String, Version> versionMap = new HashMap<>() {
+            {
+                put("ignore_unmapped", MINIMAL_SUPPORTED_VERSION_FOR_IGNORE_UNMAPPED);
+                put(MODEL_NODE_ASSIGNMENT_KEY, MINIMAL_SUPPORTED_VERSION_FOR_MODEL_NODE_ASSIGNMENT);
+                put(MODEL_METHOD_COMPONENT_CONTEXT_KEY, MINIMAL_SUPPORTED_VERSION_FOR_MODEL_METHOD_COMPONENT_CONTEXT);
+                put(KNNConstants.RADIAL_SEARCH_KEY, MINIMAL_SUPPORTED_VERSION_FOR_RADIAL_SEARCH);
+                put(KNNConstants.METHOD_PARAMETER, MINIMAL_SUPPORTED_VERSION_FOR_METHOD_PARAMETERS);
+                put(KNNConstants.MODEL_VECTOR_DATA_TYPE_KEY, MINIMAL_SUPPORTED_VERSION_FOR_MODEL_VECTOR_DATA_TYPE);
+            }
+        };
+
+        for (final MethodParameter methodParameter : MethodParameter.values()) {
+            if (methodParameter.getVersion() != null) {
+                versionMap.put(methodParameter.getName(), methodParameter.getVersion());
+            }
+        }
+        return Collections.unmodifiableMap(versionMap);
     }
 }
--- a/src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80DocValuesConsumer.java
+++ b/src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80DocValuesConsumer.java
@@ -5,7 +5,6 @@
 
 package org.opensearch.knn.index.codec.KNN80Codec;
 
-import com.google.common.collect.ImmutableMap;
 import lombok.NonNull;
 import lombok.extern.log4j.Log4j2;
 import org.apache.lucene.store.ChecksumIndexInput;
@@ -15,6 +14,7 @@ import org.opensearch.core.common.bytes.BytesArray;
 import org.opensearch.core.xcontent.MediaTypeRegistry;
 import org.opensearch.core.xcontent.DeprecationHandler;
 import org.opensearch.core.xcontent.NamedXContentRegistry;
+import org.opensearch.knn.index.IndexUtil;
 import org.opensearch.knn.index.KNNSettings;
 import org.opensearch.knn.index.VectorDataType;
 import org.opensearch.knn.index.codec.transfer.VectorTransfer;
@@ -111,30 +111,10 @@ class KNN80DocValuesConsumer extends DocValuesConsumer implements Closeable {
         return KNNEngine.getEngine(engineName);
     }
 
-    private VectorTransfer getVectorTransfer(FieldInfo field) {
-        if (VectorDataType.BINARY.getValue().equalsIgnoreCase(field.attributes().get(KNNConstants.VECTOR_DATA_TYPE_FIELD))) {
-            return new VectorTransferByte(KNNSettings.getVectorStreamingMemoryLimit().getBytes());
-        }
-        return new VectorTransferFloat(KNNSettings.getVectorStreamingMemoryLimit().getBytes());
-    }
-
     public void addKNNBinaryField(FieldInfo field, DocValuesProducer valuesProducer, boolean isMerge, boolean isRefresh)
         throws IOException {
         // Get values to be indexed
         BinaryDocValues values = valuesProducer.getBinary(field);
-        KNNCodecUtil.Pair pair = KNNCodecUtil.getPair(values, getVectorTransfer(field));
-        if (pair.getVectorAddress() == 0 || pair.docs.length == 0) {
-            logger.info("Skipping engine index creation as there are no vectors or docs in the segment");
-            return;
-        }
-        long arraySize = calculateArraySize(pair.docs.length, pair.getDimension(), pair.serializationMode);
-        if (isMerge) {
-            KNNGraphValue.MERGE_CURRENT_OPERATIONS.increment();
-            KNNGraphValue.MERGE_CURRENT_DOCS.incrementBy(pair.docs.length);
-            KNNGraphValue.MERGE_CURRENT_SIZE_IN_BYTES.incrementBy(arraySize);
-        }
-        // Increment counter for number of graph index requests
-        KNNCounter.GRAPH_INDEX_REQUESTS.increment();
         final KNNEngine knnEngine = getKNNEngine(field);
         final String engineFileName = buildEngineFileName(
             state.segmentInfo.name,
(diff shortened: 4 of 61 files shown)


src/main/java/org/opensearch/knn/plugin/transport/TrainingModelRequest.java
```diff
@@ -103,6 +107,7 @@ public TrainingModelRequest(StreamInput in) throws IOException {
         this.maximumVectorCount = in.readInt();
         this.searchSize = in.readInt();
         this.trainingDataSizeInKB = in.readInt();
+        this.vectorDataType = VectorDataType.get(in.readOptionalString());
```
jmazanec15, line 110: This wont work in mixed cluster state - is that expected? If we want it to we cannot use readOptionalString we need to do version check.

src/main/java/org/opensearch/knn/training/TrainingJob.java
```diff
@@ -182,6 +187,14 @@ public void run() {
                 KNNSettings.state().getSettingValue(KNNSettings.KNN_ALGO_PARAM_INDEX_THREAD_QTY)
             );
 
+            if (VectorDataType.BINARY.equals(model.getModelMetadata().getVectorDataType())) {
+                trainParameters.put(
+                    KNNConstants.INDEX_DESCRIPTION_PARAMETER,
+                    FAISS_BINARY_INDEX_DESCRIPTION_PREFIX + trainParameters.get(KNNConstants.INDEX_DESCRIPTION_PARAMETER).toString()
```
jmazanec15, line 193: I see this happening in a couple places. We should abstract into a single method.

jmazanec15 (commented):
I think overall, it looks pretty good. A couple comments

jni/src/faiss_wrapper.cpp
```diff
@@ -223,6 +226,76 @@ void knn_jni::faiss_wrapper::CreateIndexFromTemplate(knn_jni::JNIUtilInterface *
     faiss::write_index(&idMap, indexPathCpp.c_str());
 }
 
+void knn_jni::faiss_wrapper::CreateBinaryIndexFromTemplate(knn_jni::JNIUtilInterface * jniUtil, JNIEnv * env, jintArray idsJ,
+                                                     jlong vectorsAddressJ, jint dimJ, jstring indexPathJ,
+                                                     jbyteArray templateIndexJ, jobject parametersJ) {
+    if (idsJ == nullptr) {
```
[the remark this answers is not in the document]
junqiu-lei: I will raise another PR to add it into faiss_index_service.cpp

src/main/java/org/opensearch/knn/plugin/transport/TrainingModelRequest.java
junqiu-lei: Can we use it like?
```
        if (IndexUtil.isVersionOnOrAfterMinRequiredVersion(in.getVersion(), KNNConstants.MODEL_VECTOR_DATA_TYPE_KEY)) {
            this.vectorDataType = VectorDataType.get(in.readOptionalString());
        } else {
            this.vectorDataType = VectorDataType.FLOAT;
        }
```

src/main/java/org/opensearch/knn/index/IndexUtil.java
```diff
@@ -248,6 +251,29 @@ public static ValidationException validateKnnField(
             return exception;
         }
 
+        // Return if vector data type does not need to be checked
+        if (expectedVectorDataType == null) {
+            return null;
+        }
+
+        // Determine the data type of the training index
+        VectorDataType trainIndexDataType = fieldMap.containsKey(VECTOR_DATA_TYPE_FIELD)
```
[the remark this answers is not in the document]
junqiu-lei: ack

src/main/java/org/opensearch/knn/training/TrainingJob.java
junqiu-lei: abstracted into a single method.

src/main/java/org/opensearch/knn/index/IndexUtil.java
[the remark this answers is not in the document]
junqiu-lei: updated

jni/src/faiss_wrapper.cpp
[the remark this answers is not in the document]
junqiu-lei: There is a related issue cut https://github.com/opensearch-project/k-NN/issues/1773

src/main/java/org/opensearch/knn/indices/ModelMetadata.java
```diff
@@ -351,12 +371,37 @@ public static ModelMetadata fromString(String modelMetadataString) {
...
+                error,
+                trainingNodeAssignment,
+                methodComponentContext,
+                vectorDataType
             );
         } else {
             throw new IllegalArgumentException(
                 "Illegal format for model metadata. Must be of the form "
-                    + "\"<KNNEngine>,<SpaceType>,<Dimension>,<ModelState>,<Timestamp>,<Description>,<Error>\" or \"<KNNEngine>,<SpaceType>,<Dimension>,<ModelState>,<Timestamp>,<Description>,<Error>,<NodeAssignment>\" or \"<KNNEngine>,<SpaceType>,<Dimension>,<ModelState>,<Timestamp>,<Description>,<Error>,<NodeAssignment>,<MethodContext>\"."
+                    + "\"<KNNEngine>,<SpaceType>,<Dimension>,<ModelState>,<Timestamp>,<Description>,<Error>,<VectorDataType>\" or \"<KNNEngine>,<SpaceType>,<Dimension>,<ModelState>,<Timestamp>,<Description>,<Error>,<NodeAssignment>,<VectorDataType>\" or \"<KNNEngine>,<SpaceType>,<Dimension>,<ModelState>,<Timestamp>,<Description>,<Error>,<NodeAssignment>,<MethodContext>,<VectorDataType>\"."
```
jmazanec15, line 404: Only the last one should have <VectorDataType> I believe

src/main/java/org/opensearch/knn/plugin/transport/TrainingModelRequest.java
```diff
@@ -103,6 +107,11 @@ public TrainingModelRequest(StreamInput in) throws IOException {
         this.maximumVectorCount = in.readInt();
         this.searchSize = in.readInt();
         this.trainingDataSizeInKB = in.readInt();
+        if (IndexUtil.isVersionOnOrAfterMinRequiredVersion(in.getVersion(), KNNConstants.MODEL_VECTOR_DATA_TYPE_KEY)) {
+            this.vectorDataType = VectorDataType.get(in.readOptionalString());
```
jmazanec15, line 111: Remove Optional here - for clusters above the min version, they should have it. Also, do same for writing.

src/main/java/org/opensearch/knn/training/VectorReader.java
```diff
@@ -89,7 +89,7 @@ public void read(
             throw validationException;
         }
 
-        ValidationException fieldValidationException = IndexUtil.validateKnnField(indexMetadata, fieldName, -1, null);
+        ValidationException fieldValidationException = IndexUtil.validateKnnField(indexMetadata, fieldName, -1, null, null);
```
jmazanec15, line 92: Why is this null at the end?

src/main/java/org/opensearch/knn/plugin/transport/TrainingModelRequest.java
junqiu-lei: Ack

src/main/java/org/opensearch/knn/training/VectorReader.java
junqiu-lei: I think the existing `IndexUtil.validateKnnField` function is shared for both index field and model field validation, in above case in the vectorReader, it's used in case of index field validation, similarly like the fourth parameter of `modalDao`, I pass expectedVectorDataType with null.

In ModelDaoRequest class, we use it as:
```
        ValidationException fieldValidation = IndexUtil.validateKnnField(
            indexMetadata,
            this.trainingField,
            this.dimension,
            modelDao,
            this.vectorDataType
        );
```

jni/src/faiss_wrapper.cpp
```diff
@@ -682,6 +806,15 @@ void InternalTrainIndex(faiss::Index * index, faiss::idx_t n, const float* x) {
     }
 }
 
+void InternalTrainBinaryIndex(faiss::IndexBinary * index, faiss::idx_t n, const float* x) {
+    if (auto * indexIvf = dynamic_cast<faiss::IndexBinaryIVF*>(index)) {
+        indexIvf->make_direct_map();
+    }
+    if (!index->is_trained) {
+        index->train(n, reinterpret_cast<const uint8_t*>(x));
```
heemin32, line 814: I don't think this reinterpret will work correctly. You cannot simply reinterpret `float*` to `uint8_t*`. The reinterpreted value in uint8_t will be something meaningless. Please check.

It need to be converted from float array to byte array through loop here. However this will use more memory for binary vector training compared to float vector training. Therefore, we need to store the training vector as byte from the start instead of float.

junqiu-lei: Thanks @heemin32. I updated PR to support passing byte array at start in [TrainingDataConsumer.java](https://github.com/opensearch-project/k-NN/compare/800c7c565a32419b4bcc539c83ca15e682fdc2f7..7294db4988d1d2e6111e4d4fbc401e77e57ce7b9#diff-6a607059c7d3f059ac5bf61ac1947bf96c4cdb7182b114a910cb9b4bb3f6a8dbR44-R68).

Also updated freeVectorData method to support binary index deletion.

jni/include/commons.h
```diff
@@ -45,8 +45,9 @@ namespace knn_jni {
          * address returned by {@link JNICommons#storeVectorData(long, float[][], long, long)}
          *
          * @param memoryAddress address to be freed.
+         * @param isBinaryIndex flag to indicate if the memory address is for binary index.
          */
-        void freeVectorData(jlong);
+        void freeVectorData(jlong, jboolean);
```
heemin32, line 50: It shouldn't have a concept of binary. It should be a type of underline memory. Like if it is float or uint8_t.

junqiu-lei: Updated to call the right jni method in java layer.

jni/src/faiss_wrapper.cpp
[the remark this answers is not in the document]
heemin32: @junqiu-lei You might want to cut a new issue and follow up after delivering this feature if you cannot make the change in this PR. The #1773 is only for searchWilthFilter method.

jni/src/faiss_wrapper.cpp
```diff
@@ -624,6 +697,57 @@ jbyteArray knn_jni::faiss_wrapper::TrainIndex(knn_jni::JNIUtilInterface * jniUti
     return ret;
 }
 
+jbyteArray knn_jni::faiss_wrapper::TrainBinaryIndex(knn_jni::JNIUtilInterface * jniUtil, JNIEnv * env, jobject parametersJ,
```
heemin32, line 700: We might need to refactor this to reduce code duplication

junqiu-lei: Will have another PR to have the refactor, we can follow on the issue https://github.com/opensearch-project/k-NN/issues/1773

src/main/java/org/opensearch/knn/index/IndexUtil.java
```diff
@@ -248,6 +263,28 @@ public static ValidationException validateKnnField(
             return exception;
         }
 
+        // Return if vector data type does not need to be checked
+        if (expectedVectorDataType == null) {
```
heemin32, line 267: I think expectedVectorDataType shouldn't be null?
Also, we should not just return here even if it is null so that the rest of the code can be executed to check other validation; I know there is no other validation now but still it is not future change friendly.

heemin32: When `expectedDimension` is <0 or null, your validation will never get hit.

src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80DocValuesConsumer.java
```diff
@@ -188,18 +199,19 @@ private void recordRefreshStats() {
...
-    private void createKNNIndexFromTemplate(byte[] model, KNNCodecUtil.Pair pair, KNNEngine knnEngine, String indexPath) {
-        Map<String, Object> parameters = ImmutableMap.of(
-            KNNConstants.INDEX_THREAD_QTY,
-            KNNSettings.state().getSettingValue(KNNSettings.KNN_ALGO_PARAM_INDEX_THREAD_QTY)
-        );
+    private void createKNNIndexFromTemplate(Model model, KNNCodecUtil.Pair pair, KNNEngine knnEngine, String indexPath) {
+        Map<String, Object> parameters = new HashMap<>();
+        parameters.put(KNNConstants.INDEX_THREAD_QTY, KNNSettings.state().getSettingValue(KNNSettings.KNN_ALGO_PARAM_INDEX_THREAD_QTY));
+
+        IndexUtil.updateVectorDataTypeToParameters(parameters, model.getModelMetadata().getVectorDataType());
```
heemin32, line 206: I think this method actually degrade the readability. Up to you though.

src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80DocValuesConsumer.java
```diff
@@ -248,7 +260,7 @@ private void createKNNIndexFromScratch(FieldInfo fieldInfo, KNNCodecUtil.Pair pa
                 KNNConstants.INDEX_DESCRIPTION_PARAMETER,
                 FAISS_BINARY_INDEX_DESCRIPTION_PREFIX + parameters.get(KNNConstants.INDEX_DESCRIPTION_PARAMETER).toString()
             );
-            parameters.put(KNNConstants.VECTOR_DATA_TYPE_FIELD, VectorDataType.BINARY.getValue());
+            IndexUtil.updateVectorDataTypeToParameters(parameters, VectorDataType.BINARY);
```
heemin32, line 263: Same here. By having this method, I need to go inside the method to see what it actually does which is just adding binary value in the parameters.

junqiu-lei: Answered above

src/main/java/org/opensearch/knn/index/IndexUtil.java
junqiu-lei: The existing validateKnnField method return one failed validation per time, so I followed same pattern here for vector type.

src/main/java/org/opensearch/knn/training/TrainingDataConsumer.java
```diff
@@ -24,23 +28,43 @@
...
 
     @Override
     public void accept(List<Float[]> floats) {
-        trainingDataAllocation.setMemoryAddress(
-            JNIService.transferVectors(
-                trainingDataAllocation.getMemoryAddress(),
-                floats.stream().map(ArrayUtils::toPrimitive).toArray(float[][]::new)
-            )
-        );
+        long memoryAddress = trainingDataAllocation.getMemoryAddress();
```
heemin32, line 45: Please add a unit test to validate the code.

src/main/java/org/opensearch/knn/plugin/transport/TrainingModelRequest.java
```diff
@@ -336,5 +355,10 @@ public void writeTo(StreamOutput out) throws IOException {
         out.writeInt(this.maximumVectorCount);
         out.writeInt(this.searchSize);
         out.writeInt(this.trainingDataSizeInKB);
+        if (IndexUtil.isVersionOnOrAfterMinRequiredVersion(out.getVersion(), KNNConstants.MODEL_VECTOR_DATA_TYPE_KEY)) {
+            out.writeString(this.vectorDataType.getValue());
+        } else {
+            out.writeString(VectorDataType.FLOAT.getValue());
```
heemin32, line 361: This is not going to work. The old node does not know how to parse this vector data type.
We should throw an error that binary data type is not supported when cluster has a node of old version.
Also, I think we also don't support Byte data type. Have we add a validation for that?

jni/include/faiss_wrapper.h
```diff
@@ -80,6 +86,12 @@ namespace knn_jni {
         jobjectArray QueryBinaryIndex_WithFilter(knn_jni::JNIUtilInterface * jniUtil, JNIEnv * env, jlong indexPointerJ,
                                                  jbyteArray queryVectorJ, jint kJ, jobject methodParamsJ, jlongArray filterIdsJ, jint filterIdsTypeJ, jintArray parentIdsJ);
 
+        // Execute a query against the binary index located in memory at indexPointerJ along with Filters
+        //
+        // Return an array of KNNQueryResults
+        jobjectArray QueryBinaryIndex_WithFilter(knn_jni::JNIUtilInterface * jniUtil, JNIEnv * env, jlong indexPointerJ,
```
jmazanec15, line 92: Why add this one as well? Do we need it?

src/main/java/org/opensearch/knn/indices/ModelMetadata.java
```diff
@@ -407,6 +453,10 @@ public static ModelMetadata getMetadataFromSourceMap(final Map<String, Object> m
             methodComponentContext = MethodComponentContext.EMPTY;
         }
 
+        if (vectorDataType == null) {
+            vectorDataType = VectorDataType.FLOAT.getValue();
```
heemin32, line 457: I see there are many place where we use float as default value. Can we define default value in VectorDataType and use it instead so that we know where we are using default value in the future?

src/main/java/org/opensearch/knn/index/memory/NativeMemoryLoadStrategy.java
```diff
@@ -167,11 +167,15 @@ public NativeMemoryAllocation.TrainingDataAllocation load(
             NativeMemoryAllocation.TrainingDataAllocation trainingDataAllocation = new NativeMemoryAllocation.TrainingDataAllocation(
                 executor,
                 0,
-                nativeMemoryEntryContext.calculateSizeInKB()
+                nativeMemoryEntryContext.calculateSizeInKB(),
```
heemin32, line 170: Shouldn't we modify this method? the size might be different for byte vector?

src/main/java/org/opensearch/knn/index/IndexUtil.java
heemin32: Try to write a unit test. You may find an issue with current implementation.
It is okay to return single exception but the issue here is you are returning when there is no exception. Then the rest of the validation code won't work.
Also, please check if your code works when expectedDimension is < 0 or null. In such case, your validation won't get executed.

src/main/java/org/opensearch/knn/training/TrainingDataConsumer.java
junqiu-lei: ack

src/main/java/org/opensearch/knn/indices/ModelMetadata.java
junqiu-lei: ack

src/main/java/org/opensearch/knn/index/memory/NativeMemoryLoadStrategy.java
junqiu-lei: ack

jni/src/faiss_wrapper.cpp
```diff
@@ -223,6 +226,76 @@ void knn_jni::faiss_wrapper::CreateIndexFromTemplate(knn_jni::JNIUtilInterface *
     faiss::write_index(&idMap, indexPathCpp.c_str());
 }
 
+void knn_jni::faiss_wrapper::CreateBinaryIndexFromTemplate(knn_jni::JNIUtilInterface * jniUtil, JNIEnv * env, jintArray idsJ,
```
Vikasht34, line 229: Can we refactor this method into 3 pieces to make more readable.

1. Validation 
2. AddPoints to Graph and WritertoDisk,

jni/src/faiss_wrapper.cpp
```diff
@@ -223,6 +226,76 @@ void knn_jni::faiss_wrapper::CreateIndexFromTemplate(knn_jni::JNIUtilInterface *
...
+    jniUtil->DeleteLocalRef(env, parametersJ);
+
+    // Read data set
+    // Read vectors from memory address
+    auto *inputVectors = reinterpret_cast<std::vector<uint8_t>*>(vectorsAddressJ);
+    int dim = (int)dimJ;
+    if (dim % 8 != 0) {
+        throw std::runtime_error("Dimensions should be multiply of 8");
+    }
+    int numVectors = (int) (inputVectors->size() / (uint64_t) (dim / 8));
```
Vikasht34, line 267: We can move these validations to common functions , these would be use in CreateBinaryFromScratch As well.

jni/src/faiss_wrapper.cpp
```diff
@@ -624,6 +697,57 @@ jbyteArray knn_jni::faiss_wrapper::TrainIndex(knn_jni::JNIUtilInterface * jniUti
     return ret;
 }
 
+jbyteArray knn_jni::faiss_wrapper::TrainBinaryIndex(knn_jni::JNIUtilInterface * jniUtil, JNIEnv * env, jobject parametersJ,
+                                              jint dimensionJ, jlong trainVectorsPointerJ) {
+    // First, we need to build the index
+    if (parametersJ == nullptr) {
```
Vikasht34, line 703: Can we simplify this whole function like this , It's more logically separated and reduced deuplications.

```
if (parametersJ == nullptr) {
        throw std::runtime_error("Parameters cannot be null");
    }

    // Convert Java map to C++ map
    auto parametersCpp = jniUtil->ConvertJavaMapToCppMap(env, parametersJ);

    // Get space type and metric
    faiss::MetricType metric = GetMetricFromParameters(env, parametersCpp, jniUtil);

    // Get index description
    std::string indexDescriptionCpp = GetIndexDescriptionFromParameters(env, parametersCpp, jniUtil);

    // Create faiss index
    auto indexWriter = CreateFaissIndexWriter(dimensionJ, indexDescriptionCpp);

    // Set thread count if provided
    SetThreadCountIfProvided(env, parametersCpp, jniUtil);

    // Train index if needed
    TrainIndexIfNeeded(env, trainVectorsPointerJ, dimensionJ, indexWriter.get());

    // Clean up local references
    jniUtil->DeleteLocalRef(env, parametersJ);

    // Serialize index and return byte array
    return SerializeIndex(env, indexWriter.get(), jniUtil);
```

jni/src/faiss_wrapper.cpp
Vikasht34, line 229: This class is not right place to have this , we already have BinaryIndexService can we have all Binary related functions to that BinaryIndexService.

src/main/java/org/opensearch/knn/index/IndexUtil.java
```diff
@@ -119,6 +121,18 @@ private static Object getFieldMapping(final Map<String, Object> properties, fina
         return currentFieldMapping;
     }
 
+    /**
+     *  This method is used to get the vector data type from field mapping
+     * @param fieldMap field mapping
+     * @return vector data type
+     */
+    private static VectorDataType getVectorDataTypeFromFieldMapping(Map<String, Object> fieldMap) {
```
Vikasht34, line 129: Move all Private Functions to end of the file .

src/main/java/org/opensearch/knn/index/IndexUtil.java
```diff
@@ -135,7 +149,8 @@ public static ValidationException validateKnnField(
         IndexMetadata indexMetadata,
         String field,
         int expectedDimension,
-        ModelDao modelDao
+        ModelDao modelDao,
+        VectorDataType expectedVectorDataType
```
Vikasht34, line 153: Since we are adding extra parameters to function , we will have make sure we are passing this extra param to all flow which is not required. Let's have overloaded function with extra params which makes sure that other flow is not affected and then refactor this method to avoid code duplicatation.

src/main/java/org/opensearch/knn/index/IndexUtil.java
Vikasht34, line 267: This is the reason why we would like to have overloaded function of validateKnnField , else we will have end up increasing Cyclomatic of this method.

src/main/java/org/opensearch/knn/index/IndexUtil.java
```diff
@@ -321,4 +358,26 @@ public static boolean isBinaryIndex(KNNEngine knnEngine, Map<String, Object> par
...
             && parameters.get(VECTOR_DATA_TYPE_FIELD).toString().equals(VectorDataType.BINARY.getValue());
     }
+
+    /**
+     * Tell if it is binary index or not
+     *
+     * @param vectorDataType vector data type
+     * @return true if it is binary index
+     */
+    public static boolean isBinaryIndex(VectorDataType vectorDataType) {
```
Vikasht34, line 368: Move all public functions above private functions in this class.

src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80DocValuesConsumer.java
```diff
@@ -112,7 +112,18 @@ private KNNEngine getKNNEngine(@NonNull FieldInfo field) {
     }
 
     private VectorTransfer getVectorTransfer(FieldInfo field) {
-        if (VectorDataType.BINARY.getValue().equalsIgnoreCase(field.attributes().get(KNNConstants.VECTOR_DATA_TYPE_FIELD))) {
```
Vikasht34, line 115: Let's move this check to Binary in differnt functions , It's make easy to understand the flow.

```
boolean isBinary = isFieldBinary(field);

private boolean isFieldBinary(Field field) {
    // Check if the field has a model ID and retrieve the model's vector data type
    if (field.attributes().containsKey(MODEL_ID)) {
        Model model = ModelCache.getInstance().get(field.attributes().get(MODEL_ID));
        return model.getModelMetadata().getVectorDataType() == VectorDataType.BINARY;
    } else if (VectorDataType.BINARY.getValue().equalsIgnoreCase(field.attributes().get(KNNConstants.VECTOR_DATA_TYPE_FIELD))) {
        return true;
    }
    return false;
}
```

Vikasht34 (commented):
Can we have Unit test for all c++ functions written.

src/main/java/org/opensearch/knn/indices/ModelMetadata.java
```diff
@@ -351,12 +371,37 @@ public static ModelMetadata fromString(String modelMetadataString) {
                 description,
                 error,
                 trainingNodeAssignment,
-                methodComponentContext
+                methodComponentContext,
+                VectorDataType.FLOAT
+            );
+        } else if (modelMetadataArray.length == 10) {
```
Vikasht34, line 377: This if -else is going to increase and it's not a good practice to use this , Can we use ReGistry Pattern to re-factor this , I will leave up to you to think how to avoid this if -else condition.

jni/src/faiss_wrapper.cpp
junqiu-lei: Refactor isn't in scope of this PR.

Vikasht34: I would prefer Rule of 30 , if functions length is going over 30 lines , we should sperate logical

jni/include/faiss_wrapper.h
jmazanec15: I dont understand - should it be included in this change?

src/main/java/org/opensearch/knn/index/IndexUtil.java
junqiu-lei: updated the validateKnnField function to have vector data type check when it's non null, the flows will be verified by adding tests

src/main/java/org/opensearch/knn/index/IndexUtil.java
junqiu-lei: done

jni/include/faiss_wrapper.h
junqiu-lei: It'll be removed, sorry for the confusion.

src/main/java/org/opensearch/knn/indices/ModelMetadata.java
Vikasht34: Simple Example of how can we avoid all duplications code ,,,

```
public static ModelMetadata fromString(String modelMetadataString) {
    String[] modelMetadataArray = modelMetadataString.split(DELIMITER, -1);
    int length = modelMetadataArray.length;

    if (length < 7 || length > 9) {
        // Throw Exception 
    }

    KNNEngine knnEngine = KNNEngine.getEngine(modelMetadataArray[0]);
    SpaceType spaceType = SpaceType.getSpace(modelMetadataArray[1]);
    int dimension = Integer.parseInt(modelMetadataArray[2]);
    ModelState modelState = ModelState.getModelState(modelMetadataArray[3]);
    String timestamp = modelMetadataArray[4];
    String description = modelMetadataArray[5];
    String error = modelMetadataArray[6];

    String trainingNodeAssignment = "";
    MethodComponentContext methodComponentContext = MethodComponentContext.EMPTY;

    if (length >= 8) {
        trainingNodeAssignment = modelMetadataArray[7];
    }
    if (length == 9) {
        methodComponentContext = MethodComponentContext.fromClusterStateString(modelMetadataArray[8]);
    }
    return new ModelMetadata(
        knnEngine,
        spaceType,
        dimension,
        modelState,
        timestamp,
        description,
        error,
        trainingNodeAssignment,
        methodComponentContext
    );
}
```

jni/src/faiss_wrapper.cpp
junqiu-lei: there will have a separate PR for the refactoring, created issue https://github.com/opensearch-project/k-NN/issues/1846

jni/src/faiss_wrapper.cpp
junqiu-lei: will be done in refactoring PR

jni/src/faiss_wrapper.cpp
junqiu-lei: will be done in refactoring PR

jni/src/faiss_wrapper.cpp
junqiu-lei: Yes, we will add it in BinaryIndexService by refactoring.

src/main/java/org/opensearch/knn/index/IndexUtil.java
junqiu-lei: done

src/main/java/org/opensearch/knn/index/IndexUtil.java
junqiu-lei: Updated

src/main/java/org/opensearch/knn/jni/JNIService.java
```diff
@@ -83,8 +83,13 @@ public static void createIndexFromTemplate(
         KNNEngine knnEngine
     ) {
         if (KNNEngine.FAISS == knnEngine) {
-            FaissService.createIndexFromTemplate(ids, vectorsAddress, dim, indexPath, templateIndex, parameters);
-            return;
+            if (IndexUtil.isBinaryIndex(knnEngine, parameters)) {
+                FaissService.createBinaryIndexFromTemplate(ids, vectorsAddress, dim, indexPath, templateIndex, parameters);
```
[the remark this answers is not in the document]
junqiu-lei: will be done in refactoring

src/main/java/org/opensearch/knn/jni/JNIService.java
```diff
@@ -308,6 +313,9 @@ public static void freeSharedIndexState(long shareIndexStateAddr, KNNEngine knnE
      */
     public static byte[] trainIndex(Map<String, Object> indexParameters, int dimension, long trainVectorsPointer, KNNEngine knnEngine) {
         if (KNNEngine.FAISS == knnEngine) {
+            if (IndexUtil.isBinaryIndex(knnEngine, indexParameters)) {
+                return FaissService.trainBinaryIndex(indexParameters, dimension, trainVectorsPointer);
```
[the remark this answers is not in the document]
junqiu-lei: I followed other functions check in this JNIService class

src/main/java/org/opensearch/knn/indices/ModelMetadata.java
junqiu-lei: refactored this method

src/main/java/org/opensearch/knn/training/TrainingDataConsumer.java
```diff
@@ -24,23 +28,43 @@
...
+
+        if (IndexUtil.isBinaryIndex(this.vectorDataType)) {
+            byte[][] byteArray = floats.stream()
```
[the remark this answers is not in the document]
junqiu-lei: Refactored to have FloatTrainingDataConsumer and ByteTrainingDataConsumer to handle train data consuming separately.

junqiu-lei:
> Can we have Unit test for all c++ functions written.

will do

src/main/java/org/opensearch/knn/plugin/transport/TrainingModelRequest.java
junqiu-lei: Added version validation at rest request step.

src/main/java/org/opensearch/knn/index/IndexUtil.java
junqiu-lei: Offline discussed with Heemin on this part, moved the vector data type check when it's non null before dimension check.

src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80DocValuesConsumer.java
```diff
@@ -354,4 +348,21 @@ private boolean isChecksumValid(long value) {
         // https://github.com/apache/lucene/blob/branch_9_0/lucene/core/src/java/org/apache/lucene/codecs/CodecUtil.java#L644-L647
         return (value & CRC32_CHECKSUM_SANITY) != 0;
     }
+
+    private boolean isBinaryField(FieldInfo field) {
+        if (field.attributes().containsKey(MODEL_ID)) {
+            Model model = ModelCache.getInstance().get(field.attributes().get(MODEL_ID));
```
jmazanec15, line 354: Model can be quite big because it contains the binary blob. Can we read from model metadata instead via ModelDao?

junqiu-lei: Updated code to use model only once in the addKNNBinaryField function

src/main/java/org/opensearch/knn/training/ByteTrainingDataConsumer.java
```diff
@@ -0,0 +1,90 @@
...
+    public void processTrainingVectors(SearchResponse searchResponse, int vectorsToAdd, String fieldName) {
+        List<byte[]> vectors = extractVectorsFromHits(searchResponse, vectorsToAdd, fieldName);
+
+        setTotalVectorsCountAdded(getTotalVectorsCountAdded() + vectors.size());
+
+        accept(vectors);
+    }
+
+    @Override
+    public List<byte[]> getTotalAddedVectors() {
```
heemin32, line 87: I see this method is called below. Where the implementation is?

src/main/java/org/opensearch/knn/training/ByteTrainingDataConsumer.java
```diff
@@ -0,0 +1,90 @@
...
+     * Constructor
+     *
+     * @param trainingDataAllocation NativeMemoryAllocation that contains information about native memory allocation.
+     */
+    public ByteTrainingDataConsumer(NativeMemoryAllocation.TrainingDataAllocation trainingDataAllocation) {
+        super(trainingDataAllocation);
+    }
+
+    @Override
+    public void accept(List<byte[]> byteVectors) {
```
heemin32, line 37: Why public?

src/main/java/org/opensearch/knn/training/ByteTrainingDataConsumer.java
```diff
@@ -0,0 +1,90 @@
...
+        long memoryAddress = trainingDataAllocation.getMemoryAddress();
+        memoryAddress = JNICommons.storeByteVectorData(memoryAddress, byteVectors.toArray(new byte[0][0]), byteVectors.size());
+        trainingDataAllocation.setMemoryAddress(memoryAddress);
+    }
+
+    @Override
+    public List<byte[]> extractVectorsFromHits(SearchResponse searchResponse, int vectorsToAdd, String fieldName) {
```
heemin32, line 44: Why public?

src/main/java/org/opensearch/knn/training/ByteTrainingDataConsumer.java
```diff
@@ -0,0 +1,90 @@
...
+
+        if (nullVectorCount > 0) {
+            logger.warn("Found {} documents with null byte vectors in field {}", nullVectorCount, fieldName);
+        }
+
+        return byteVectors;
+    }
+
+    @Override
+    public void processTrainingVectors(SearchResponse searchResponse, int vectorsToAdd, String fieldName) {
```
heemin32, line 78: I think only this method is needs to be public?

src/main/java/org/opensearch/knn/training/TrainingDataConsumer.java
```diff
@@ -11,19 +11,31 @@
...
 
 /**
- * Transfers vectors from JVM to native memory.
+ * TrainingDataConsumer is an abstract class that defines the interface for consuming training data.
+ * It is used to process training data and add it to the training data allocation.
+ *
+ * @param <T> The type of the training data
  */
-public class TrainingDataConsumer implements Consumer<List<Float[]>> {
+public abstract class TrainingDataConsumer<T> implements Consumer<List<T>> {
```
heemin32, line 32: Why does it have to implement Consumer?
I think `accept` method is called only inside `processTrainingVectors`

src/main/java/org/opensearch/knn/training/ByteTrainingDataConsumer.java
junqiu-lei: this method is used in TestFloatTrainingDataConsumer
```
    // create test float training data consumer class extending FloatTrainingDataConsumer
    private static class TestFloatTrainingDataConsumer extends FloatTrainingDataConsumer {
        @Getter
        private List<Float[]> totalAddedVectors = new ArrayList<>();

        public TestFloatTrainingDataConsumer(NativeMemoryAllocation.TrainingDataAllocation trainingDataAllocation) {
            super(trainingDataAllocation);
        }

        @Override
        public void processTrainingVectors(SearchResponse searchResponse, int vectorsToAdd, String fieldName) {
            List<Float[]> vectors = extractVectorsFromHits(searchResponse, vectorsToAdd, fieldName);
            totalAddedVectors.addAll(vectors);
            setTotalVectorsCountAdded(getTotalVectorsCountAdded() + vectors.size());
            accept(vectors);
        }
    }
```

src/main/java/org/opensearch/knn/training/ByteTrainingDataConsumer.java
junqiu-lei: previously implemented from Consumer, which are public, now removed parent implement from Consumer and changed it to protected

src/main/java/org/opensearch/knn/training/ByteTrainingDataConsumer.java
junqiu-lei: Updated

src/main/java/org/opensearch/knn/training/ByteTrainingDataConsumer.java
junqiu-lei: updated

src/main/java/org/opensearch/knn/training/TrainingDataConsumer.java
junqiu-lei: Updated without implement from Consumer

src/main/java/org/opensearch/knn/training/ByteTrainingDataConsumer.java
heemin32: It is wired to have a method only to be used in testing... Can we do differently?

src/main/java/org/opensearch/knn/training/ByteTrainingDataConsumer.java
heemin32: It can be even private. I don't see it being used outside of this class.

src/main/java/org/opensearch/knn/training/TrainingDataConsumer.java
```diff
@@ -11,19 +11,30 @@
...
 
 /**
- * Transfers vectors from JVM to native memory.
+ * TrainingDataConsumer is an abstract class that defines the interface for consuming training data.
+ * It is used to process training data and add it to the training data allocation.
+ *
+ * @param <T> The type of the training data
  */
-public class TrainingDataConsumer implements Consumer<List<Float[]>> {
+public abstract class TrainingDataConsumer<T> {
```
heemin32, line 31: I feel we don't need `T` here.
I see only public abstract void processTrainingVectors(SearchResponse searchResponse, int vectorsToAdd, String fieldName); is actually used.

src/main/java/org/opensearch/knn/training/ByteTrainingDataConsumer.java
junqiu-lei: I tried but couldn't find other way to read the whole vectors

src/main/java/org/opensearch/knn/training/TrainingDataConsumer.java
```diff
@@ -34,13 +45,29 @@ public TrainingDataConsumer(NativeMemoryAllocation.TrainingDataAllocation traini
...
 
-    @Override
-    public void accept(List<Float[]> floats) {
-        trainingDataAllocation.setMemoryAddress(
-            JNIService.transferVectors(
-                trainingDataAllocation.getMemoryAddress(),
-                floats.stream().map(ArrayUtils::toPrimitive).toArray(float[][]::new)
-            )
-        );
+    protected abstract void accept(List<T> vectors);
```
heemin32, line 48: We can remove accept, extractVectorsFromHits, getTotalAddedVectors which is not being called outside of this class.

src/main/java/org/opensearch/knn/training/ByteTrainingDataConsumer.java
heemin32: You don't need to. For unit test, all you need to verify is that JNILayer method is called properly with expected parameter.

src/main/java/org/opensearch/knn/training/TrainingDataConsumer.java
junqiu-lei: Removed extractVectorsFromHits, getTotalAddedVectors and kept accept to expose for some tests

src/main/java/org/opensearch/knn/training/TrainingDataConsumer.java
junqiu-lei: updated

src/main/java/org/opensearch/knn/training/ByteTrainingDataConsumer.java
junqiu-lei: accept method is used in some tests, but I removed other methods.

src/main/java/org/opensearch/knn/training/ByteTrainingDataConsumer.java
junqiu-lei: updated

src/main/java/org/opensearch/knn/training/ByteTrainingDataConsumer.java
heemin32: Thanks for the update but could you try to remove accept method as well? It is not great to have it just because it is used in the test.

src/test/java/org/opensearch/knn/training/VectorReaderTests.java
```diff
@@ -364,51 +363,69 @@ public void testRead_valid_NestedField() throws InterruptedException, ExecutionE
...
+        return ActionListener.wrap(response -> countDownLatch.countDown(), Throwable::printStackTrace);
+    }
+
+    private NativeMemoryAllocation.TrainingDataAllocation createMockTrainingDataAllocation() {
+        return new NativeMemoryAllocation.TrainingDataAllocation(null, 0, 0, VectorDataType.FLOAT);
+    }
 
-        List<Float[]> vectorsConsumed;
+    // create test float training data consumer class extending FloatTrainingDataConsumer
+    private static class TestFloatTrainingDataConsumer extends FloatTrainingDataConsumer {
```
heemin32, line 401: Why do we need this class? I think we can just use FloatTrainingDataConsumer?

src/main/java/org/opensearch/knn/training/FloatTrainingDataConsumer.java
```diff
@@ -0,0 +1,67 @@
...
+        trainingDataAllocation.setMemoryAddress(
+            JNIService.transferVectors(
+                trainingDataAllocation.getMemoryAddress(),
+                floats.stream().map(v -> ArrayUtils.toPrimitive((Float[]) v)).toArray(float[][]::new)
+            )
+        );
+    }
+
+    @Override
+    public void processTrainingVectors(SearchResponse searchResponse, int vectorsToAdd, String fieldName) {
```
heemin32, line 48: Is there a unit test for this method?

src/test/java/org/opensearch/knn/training/VectorReaderTests.java
junqiu-lei: In the TestFloatTrainingDataConsumer `processTrainingVectors` method, it additionally count the total vectors used for test only.

src/main/java/org/opensearch/knn/training/FloatTrainingDataConsumer.java
junqiu-lei: Yes, in [VectorReaderTests.java](https://github.com/opensearch-project/k-NN/blob/5685992f6b3266b34eac2c76576a59318461eea2/src/test/java/org/opensearch/knn/training/VectorReaderTests.java) by TestFloatTrainingDataConsumer.

heemin32: I think, we should name it as` FloatTrainingDataConsumerTests`

src/main/java/org/opensearch/knn/training/ByteTrainingDataConsumer.java
junqiu-lei: The `accept` is also used in the processTrainingVectors.

src/test/java/org/opensearch/knn/training/VectorReaderTests.java
heemin32: As I mentioned in another comment, I think we don't need to validate total vectors count. All we need is to verifying that JNIService is called with expected method and parameters.

src/main/java/org/opensearch/knn/training/FloatTrainingDataConsumer.java
junqiu-lei: There is a existing test class named `FloatTrainingDataConsumerTests`, but I think VectorReaderTests can cover the test of processTrainingVectors. Will have another PR for the refactor.

heemin32 (approved):
Approving. We need to refactor TrainingDataConsumer though.

jmazanec15 (approved):

PR #1802, opened 2024-07-09
[BUG] FIX Same Suffix Cause Recall Drop to zero

### LINKED ISSUE #1803: [BUG] FIX Same Suffix Cause Recall Drop to zero
**What is the bug?**
When we have two field in mapping like following shows:

```
        "MultipleMatrix": { 
          "type": "nested",
          "properties": { 
            "Vector": { 
              "type": "knn_vector",
              "dimension": 256
            }
          }
        },
        "PersonMultipleMatrix": { 
          "type": "nested",
          "properties": { 
            "Vector": { 
              "type": "knn_vector",
              "dimension": 128
            }
          }
        }
```
 
it would generate two field file in one segment like following shows:

```
_3_2011_MultipleMatrix.Vector.hnsw
_3_2011_PersonMultipleMatrix.Vector.hnsw
```

then we would do search in the KNNWeight with knnQuery at `MultipleMatrix.Vector` field

https://github.com/opensearch-project/k-NN/blob/5139b163bc4a36aef7c99ce53d5e8b14c43eb768/src/main/java/org/opensearch/knn/index/query/KNNWeight.java#L234-L251

but when filter engineSuffix, it would return two files which ends with `MultipleMatrix.Vector.hnsw` but `_3_2011_PersonMultipleMatrix.Vector.hnsw` is not expected.

```
.filter(fileName -> fileName.endsWith(engineSuffix)) 
```

so when `Path indexPath = PathUtils.get(directory, engineFiles.get(0));` get the error indexPath, it would return nothing results, and damage the recall.

### proposal
#1802

luyuncheng:
FIX Same Suffix BUG

### Description
When we have two field in mapping like following shows:

```
        "MultipleMatrix": { 
          "type": "nested",
          "properties": { 
            "Vector": { 
              "type": "knn_vector",
              "dimension": 256
            }
          }
        },
        "PersonMultipleMatrix": { 
          "type": "nested",
          "properties": { 
            "Vector": { 
              "type": "knn_vector",
              "dimension": 128
            }
          }
        }
```
 
it would generate two field file in one segment like following shows:

```
_3_2011_MultipleMatrix.Vector.hnsw
_3_2011_PersonMultipleMatrix.Vector.hnsw
```

then we would do search in the KNNWeight with knnQuery at `MultipleMatrix.Vector` field

https://github.com/opensearch-project/k-NN/blob/5139b163bc4a36aef7c99ce53d5e8b14c43eb768/src/main/java/org/opensearch/knn/index/query/KNNWeight.java#L234-L251

but when filter engineSuffix, it would return two files which ends with `MultipleMatrix.Vector.hnsw` but `_3_2011_PersonMultipleMatrix.Vector.hnsw` is not expected.

```
.filter(fileName -> fileName.endsWith(engineSuffix)) 
```

so when `Path indexPath = PathUtils.get(directory, engineFiles.get(0));` get the error indexPath, it would return nothing results, and damage the recall.

### Issues Resolved
#1803 
 
### Proposal

we can filter with `_MultipleMatrix.Vector.hnsw` which using under line to distinguish it.


navneet1v:
@luyuncheng this is a good catch. Let me review the code.

jmazanec15:
@luyuncheng Good find. Im not sure adding "_" check will fix for all though. For instance, if I have 2 fields "target_field" and "long_target_field", the segments could look like:
- _0_2011_target_field.hnsw
- _0_2011_long_target_field.hnsw
So, on search for target_field, I think it would also pickup long_target_field.

That being said, I am wondering if we can just match the file directly instead of trying to filter the files together.

luyuncheng:
> @luyuncheng Good find. Im not sure adding "_" check will fix for all though. For instance, if I have 2 fields "target_field" and "long_target_field", the segments could look like:
> 
> * _0_2011_target_field.hnsw
> * _0_2011_long_target_field.hnsw
>   So, on search for target_field, I think it would also pickup long_target_field.
> 
> That being said, I am wondering if we can just match the file directly instead of trying to filter the files together.

@jmazanec15  Correctly! 

Our file name format is `${Segment}_${KNNEngineVersion}_${FieldName}_${Extension}` , and KNNWeight using suffix `${FieldName}_${Extension}` to filter the path. if we want to our engine is forward compatible, we can not use `${KNNEngineVersion}` to filter file. that being said, we can not filter the file directly.

As i think, we can add a sort for the file length after filter suffix **`_${FieldName}_${Extension}`**

because, all suffix matched file in current segment, the shortest file must be the matched one. the prefix only the `SpecificSegment_${KNNEngineVersion}`. and `${KNNEngineVersion}` in one segment can not changed. 

 * _0_2011_target_field.hnsw
 * _0_2011_a_target_field.hnsw

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -235,10 +235,11 @@ private Map<Integer, Float> doANNSearch(final LeafReaderContext context, final B
             ? knnEngine.getExtension() + KNNConstants.COMPOUND_EXTENSION
             : knnEngine.getExtension();
         String engineSuffix = knnQuery.getField() + engineExtension;
+        String underLineEngineSuffix = "_" + engineSuffix;
         List<String> engineFiles = reader.getSegmentInfo()
             .files()
             .stream()
-            .filter(fileName -> fileName.endsWith(engineSuffix))
+            .filter(fileName -> fileName.endsWith(underLineEngineSuffix))
```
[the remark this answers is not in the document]
luyuncheng: Added at https://github.com/opensearch-project/k-NN/pull/1802/commits/28b1d359cf65bf4e139384ab7101b644655fe721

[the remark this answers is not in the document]
luyuncheng: Also i added a sort, the reason as comments shows: https://github.com/opensearch-project/k-NN/pull/1802#issuecomment-2219455156

jmazanec15:
@luyuncheng thats a really clever idea. Im wondering if there are any potential corner cases though - I cant really think of any. 

I have alternate idea that might be more direct. Assuming format is `${Segment}_${KNNEngineVersion}_${FieldName}${Extension}`, we should remove the segment name, KNNEngineVersion, and Extension so that just the fieldName is left and then directly match against the field name. 

Because we dont know the `KNNEngineVersion`, we need to remove it with more care. However, because this isnt user provided and we control it, I think we can impose the restriction that the KNNEngineVersion does not contain "_".

So, process would look like this:
```
## Step 1 - remove ${Segment}_ from ${Segment}_${KNNEngineVersion}_${FieldName}${Extension}
${Segment}_${KNNEngineVersion}_${FieldName}${Extension} -> ${KNNEngineVersion}_${FieldName}${Extension}
_0_2011_a_target_field.hnsw -> 2011_a_target_field.hnsw

## Step 2 - remove ${Extension} from ${KNNEngineVersion}_${FieldName}${Extension}
${KNNEngineVersion}_${FieldName}${Extension} -> ${KNNEngineVersion}_${FieldName}
2011_a_target_field.hnsw -> 2011_a_target_field

## Step 3 - remove ${KNNEngineVersion}_ from ${KNNEngineVersion}_${FieldName}
${KNNEngineVersion}_${FieldName} -> ${FieldName}
2011_a_target_field -> a_target_field

## Step 4 - match directly against field name
```

In step 3, we would need to write something that just basically removes everything up till and including the first "_".

@navneet1v what is your opinion on this?

luyuncheng:
> So, process would look like this:
> 4 steps...

@jmazanec15 
this implementation is a little bit complicate.

in `KNNWeight` we only read specific segment files with `reader.getSegmentInfo().files()` and it's EngineVersion must be same string. 
so we always have the same prefix `${Segment}_${KNNEngineVersion}` in any `reader.getSegmentInfo().files()` returns. 

that being said, we just need ensure that at `fileName.indexOf(${FieldName})  is the first matched index.

how about sort by this?

heemin32:
Option1. Sort is not needed. We can just pick the one with the shortest file name. This works only when `${Segment}`, `@{KNNEngineVersion}` are same for all engine files and I believe this will hold true in the future.

Option2. For striping the `${Segment}` and `${KNNEngineVersion}`, it works only when those two value does not have `_` and the format will be always `_${Segment}_${KNNEngineVersion}_${FieldName}.${Extension}` KNNEngineVersion is in our control is Segment value also under our control?

In performance wise, The Option2 could be slightly better because we don't need to iterate all files once we found matching one whereas the Option1 needs iteration of all files to find the shortest one.

jmazanec15:
> For striping the ${Segment} and ${KNNEngineVersion}, it works only when those two value does not have _

Actually, it does not matter for ${Segment} having _ or not because we have segment name. 

I think either approach is fine though. Will approve

jmazanec15 (approved):

heemin32:
With shortest file name approach, one extreme edge case I can think of is that if one of engine files is deleted it won't throw error but just use another engine file.

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -330,6 +320,25 @@ private Map<Integer, Float> doANNSearch(final LeafReaderContext context, final B
             .collect(Collectors.toMap(KNNQueryResult::getId, result -> knnEngine.score(result.getScore(), spaceType)));
     }
 
+    @VisibleForTesting
+    List<String> getEngineFiles(SegmentReader reader, String extension) throws IOException {
```
heemin32, line 324: 1. Should make this method to return `Path indexPath` of the engine file?
2. Is there a way to move this method outside of KNNWeight so that we can avoid `@VisibleForTesting`?

heemin32, line 324: 1. Shouldn't we make this method to return `Path indexPath` of the engine file?
2. Is there a way to move this method outside of KNNWeight so that we can avoid `@VisibleForTesting`?

luyuncheng:
> With shortest file name approach, one extreme edge case I can think of is that if one of engine files is deleted it won't throw error but just use another engine file.

@heemin32 this would throw error. because, we using `state.directory.createOutput(engineFileName, state.context).close();` method to ensure indexPath tracked by lucene. and it would commit into SegmentInfos. so the extreme edge you mentioned would throw exceptions when file deleted

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
luyuncheng: > Shouldn't we make this method to return Path indexPath of the engine file?

i prefer to return List<String>,  there is some other engines we may imported in the future, we can not assume that we only have one path.

luyuncheng: > Is there a way to move this method outside of KNNWeight so that we can avoid @VisibleForTesting?

how about move method into KNNCodecUtil as a static method?

heemin32: The main motive of moving the method out of this class is to make it easier to write a unit test of the caller of this method. In that sense, no much difference between current code and static method. I am fine with current code as it might need non trivial change to do that...

heemin32: > > Shouldn't we make this method to return Path indexPath of the engine file?
> 
> i prefer to return List, there is some other engines we may imported in the future, we can not assume that we only have one path.

We are already assuming that there is only one path if you the next code where use only use the first value from the list. 
```
Path indexPath = PathUtils.get(directory, engineFiles.get(0));
```

heemin32:
> > With shortest file name approach, one extreme edge case I can think of is that if one of engine files is deleted it won't throw error but just use another engine file.
> 
> @heemin32 this would throw error. because, we using `state.directory.createOutput(engineFileName, state.context).close();` method to ensure indexPath tracked by lucene. and it would commit into SegmentInfos. so the extreme edge you mentioned would throw exceptions when file deleted

It seems like the method has a precondition that there won't be such case where a file is missing. Please add that as a comment.

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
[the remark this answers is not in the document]
heemin32: I don't see integ test but only unit test.

navneet1v:
@heemin32 Lets try to complete the review on this PR. I think we won't be able to catch all the different combinations. Can you please add your final comments where you see the concerns?

heemin32:
> @heemin32 Lets try to complete the review on this PR. I think we won't be able to catch all the different combinations. Can you please add your final comments where you see the concerns?

1. Integ test is not there - From @navneet1v's comment
2. Adding precondition comment on the method; Example. "This method assume deleted file won't happen"
3. Modify the method to return `Path indexPath`. Okay with current approach though.

luyuncheng:
@heemin32 

> 2. Adding precondition comment on the method; Example. "This method assume deleted file won't happen"

it is not a precondition, it is an exception, when you delete a file, opensearch would throw exception with file deleted, and make shard unassigned. 

> 3. Modify the method to return `Path indexPath`. Okay with current approach though.

i prefer to using List<String> as return, when there is some other engines we may imported in the future, we can not assume that we only have one path. 
even 
> We are already assuming that there is only one path. If you see the following code, it uses only the first value from the list.
> ```Path indexPath = PathUtils.get(directory, engineFiles.get(0));```

it is because we using one file right now, but i say in the future we can have multi files, and we can only changed this, and without change `getEngineFiles`

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
luyuncheng: > We are already assuming that there is only one path. If you see the following code, it uses only the first value from the list.

it is because we using one file right now, but i say in the future we can have multi files, and we can only changed this, without change getEngineFiles

heemin32:
I might have used wrong term `precondition`. It is an assumption. The purpose of comment is to record that we have considered the edge case; it won't happen and handled by OS.

Anyway, @navneet1v if you are fine, you can go ahead and approve it. I don't have a hard objection if someone else are fine with current code.

navneet1v:
> I might have used wrong term `precondition`. It is an assumption. The purpose of comment is to record that we have considered the edge case; it won't happen and handled by OS.

The case where file is not present still happens today it is case when someone set index.knn to false and run a ANN search query. In that case we return an empty scorer telling 0 hits. Keeping that thing in mind, I don't think the exception will come. Now on the part where a file can be deleted, it means the index directory has been corrupted which will cause the shard init to fail. So, I wouldn't worry about that case for now.


> Anyway, @navneet1v if you are fine, you can go ahead and approve it. I don't have a hard objection if someone else are fine with current code.


I see that we have UTs, adding a new integ test won't be of much value. We can skip it.

navneet1v:
@luyuncheng I am good with the code please fix the conflicts so that we can move ahead for merging the code.

luyuncheng:
> @luyuncheng I am good with the code please fix the conflicts so that we can move ahead for merging the code.

@navneet1v @jmazanec15
FIXED at https://github.com/opensearch-project/k-NN/pull/1802/commits/06c1d6391cbe55e79a75682958e7a74c83044b2b

 pls re-approve it

jmazanec15 (approved):

jmazanec15:
thanks @luyuncheng!

PR #1811, opened 2024-07-10
Fixing BWC tests as 2.15.0 is already launched

navneet1v:
### Description
Fixing BWC tests as 2.15.0 is already launched
 
### Issues Resolved
NA
 
### Check List
- [x] New functionality includes testing.
  - [x] All tests pass
- [x] New functionality has been documented.
  - [x] New functionality has javadoc added
- [x] Commits are signed as per the DCO using --signoff 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/k-NN/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #1812, opened 2024-07-10
Setting locale to Locale.ROOT while using String.format to ensure that UTs doesn't fail when validating the exception messages

navneet1v:
### Description
Setting locale to Locale.ROOT while using String.format to ensure that UTs doesn't fail when validating the exception messages.

So the idea here is whenever we use String.format we should ensure that we are using the Locale.ROOT to set the locale of the string, otherwise if you are validating the exception messages then the string comparison can fail and lead to flaky tests.

To reproduce the issue run the below command on the main branch of K-NN plugin.
```
./gradlew ':test' --tests "org.opensearch.knn.index.query.KNNQueryBuilderInvalidParamsTests" -Dtests.seed=EF1B1919EE6D650B  -Dtests.locale=ar-IQ -Dtests.timezone=America/Managua
```
 
### Issues Resolved
NA
 
### Check List
  - [ ] All tests pass
  - [ ] New functionality has javadoc added
- [ ] Commits are signed as per the DCO using --signoff 

### DIFF
--- a/src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
+++ b/src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
@@ -44,6 +44,7 @@ import org.opensearch.knn.indices.ModelUtil;
 import java.io.IOException;
 import java.util.Arrays;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Objects;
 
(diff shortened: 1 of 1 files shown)


navneet1v:
The fix for BWC is present in this PR: https://github.com/opensearch-project/k-NN/pull/1811

martin-gaievski (approved):

junqiu-lei (approved):

ryanbogan (approved):

Issue #1789, opened 2024-07-03
Add a reference in the comment to confirm that the calculation logic is accurate


heemin32:
I think the logic here is not correct.
For example, if the java round up the memory usage by 8 bytes, at least the logic should be

```
int vectorSize = vectorLength * FLOAT_BYTE_SIZE;
            if (vectorSize % JAVA_ROUNDING_NUMBER != 0) {
                vectorSize += (JAVA_ROUNDING_NUMBER - vectorSize % JAVA_ROUNDING_NUMBER);
            }
            int vectorsSize = numVectors * (vectorSize + JAVA_REFERENCE_SIZE);
            if (vectorsSize % JAVA_ROUNDING_NUMBER != 0) {
                vectorsSize += (JAVA_ROUNDING_NUMBER - vectorsSize % JAVA_ROUNDING_NUMBER);
            }
            return vectorsSize;
```

https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/codec/util/KNNCodecUtil.java#L94

We need a comment and reference to document to show that the logic is accurate

jmazanec15:
@heemin32 what is potential impact of this bug? Does this need prioritization for 2.16?

heemin32:
The `current_size_in_bytes` and `total_size_in_bytes` in merge stats will be incorrect.

```
"graph_stats": {
    "merge": {
        "current": 0,
        "current_docs": 0,
        "current_size_in_bytes": 0,
        "total_size_in_bytes": 0,
        "total": 0,
        "total_time_in_millis": 0,
        "total_docs": 0
    },
    "refresh": {
        "total": 0,
        "total_time_in_millis": 0
    }
```

ryanbogan:
Forgot to comment on this last week.  I'll pick this up later this week, should be a minor change.

PR #1818, opened 2024-07-11
Fix graph merge stats size calculation

ryanbogan:
### Description
Fixes the calculations for size of merges in the graph stats section of KNNStats API.  This PR changes the logic to properly round values to the correct number of bytes.
 
### Issues Resolved
https://github.com/opensearch-project/k-NN/issues/1789
 
### Check List
  - [x] All tests pass


src/main/java/org/opensearch/knn/index/codec/util/KNNCodecUtil.java
```diff
@@ -94,24 +94,25 @@ public static KNNCodecUtil.Pair getFloats(BinaryDocValues values) throws IOExcep
     }
 
     public static long calculateArraySize(int numVectors, int vectorLength, SerializationMode serializationMode) {
+        // For more details on array memory usage in Java, visit https://www.javamex.com/tutorials/memory/array_memory_usage.shtml
         if (serializationMode == SerializationMode.ARRAY) {
             int vectorSize = vectorLength * FLOAT_BYTE_SIZE + JAVA_ARRAY_HEADER_SIZE;
             if (vectorSize % JAVA_ROUNDING_NUMBER != 0) {
```
heemin32, line 100: 1. Let's make is as a private method.
2. Please add a unit test.

src/main/java/org/opensearch/knn/index/codec/util/KNNCodecUtil.java
```diff
@@ -94,24 +94,25 @@ public static KNNCodecUtil.Pair getFloats(BinaryDocValues values) throws IOExcep
     }
 
     public static long calculateArraySize(int numVectors, int vectorLength, SerializationMode serializationMode) {
+        // For more details on array memory usage in Java, visit https://www.javamex.com/tutorials/memory/array_memory_usage.shtml
```
heemin32, line 97: Please mention that following is for 2-d array.
Also, it is true regardless 32bits and 64bits machine?

src/main/java/org/opensearch/knn/index/codec/util/KNNCodecUtil.java
```diff
@@ -94,24 +94,25 @@ public static KNNCodecUtil.Pair getFloats(BinaryDocValues values) throws IOExcep
...
-                vectorSize += vectorSize % JAVA_ROUNDING_NUMBER;
+                vectorSize += (JAVA_ROUNDING_NUMBER - vectorSize % JAVA_ROUNDING_NUMBER);
             }
             int vectorsSize = numVectors * (vectorSize + JAVA_REFERENCE_SIZE) + JAVA_ARRAY_HEADER_SIZE;
             if (vectorsSize % JAVA_ROUNDING_NUMBER != 0) {
-                vectorsSize += vectorsSize % JAVA_ROUNDING_NUMBER;
+                vectorsSize += (JAVA_ROUNDING_NUMBER - vectorsSize % JAVA_ROUNDING_NUMBER);
             }
             return vectorsSize;
         } else {
```
heemin32, line 108: Could you provide a reference for this case?

src/main/java/org/opensearch/knn/index/codec/util/KNNCodecUtil.java
ryanbogan: Why would 32 bits and 64 handle it differently?

ryanbogan: I didn't see any mention of a difference for array size in my research

src/main/java/org/opensearch/knn/index/codec/util/KNNCodecUtil.java
ryanbogan: It's the same link as above, under the 2-D array section

heemin32: 2-D array is for the first one. This one is not 2-D array I believe?

src/main/java/org/opensearch/knn/index/codec/util/KNNCodecUtil.java
heemin32: The reference size or rounding could be different between 32 and 64.

heemin32: https://blog.ycrash.io/how-much-memory-does-your-java-object-occupy/

src/main/java/org/opensearch/knn/index/codec/util/KNNCodecUtil.java
ryanbogan: The first one is a 1D array since it's an individual vector.  The 2D is an array of vectors

src/main/java/org/opensearch/knn/index/codec/util/KNNCodecUtil.java
ryanbogan: Reference size would only matter if we had 3-D arrays, and we only have 2-D since it's an array of vectors.
For the rounding number, I'm a little confused because I thought no matter what in Java it always rounds up to a multiple of 8

Issue #1659, opened 2024-04-26
[FEATURE] Support for Faiss byte vector


naveentatikonda:
**Is your feature request related to a problem?**
For lucene engine we have [Lucene byte vector feature](https://opensearch.org/docs/latest/field-types/supported-field-types/knn-vector#lucene-byte-vector), which accepts byte vectors in the range [-128 to 127] providing memory savings upto 75% when compared with fp32 vectors. But, for large scale workloads we usually prefer to use faiss engine and as of today Faiss only supports fp32 and fp16 vectors(using SQfp16). So, adding byte vector support to faiss engine helps to reduce memory requirements especially for those users who are using LLM like [Cohere Embed](https://cohere.com/blog/int8-binary-embeddings) that generates signed int8 embeddings ranging from [-128 to 127].

**What solution would you like?**
Add a new Faiss ScalarQuantizer like QT_8bit_direct which doesn't require training and quantizes fp32 vector values (within signed byte range and without any precision) into byte sized vectors reducing memory footprints by a factor of 4.
https://faiss.ai/cpp_api/struct/structfaiss_1_1ScalarQuantizer.html

https://github.com/facebookresearch/faiss/issues/3488

jmazanec15:
@naveentatikonda what quantization technique is used?

jmazanec15:
Right, but how do they implement to 8-bit. I dont think they can quantize into fp8 because too much precision would be lost

jmazanec15:
> they are serializing fp32 values into uint8(0 to 255)
But how are they doing this? Would they take 0.2212 -> 0?

naveentatikonda:
> But how are they doing this? Would they take 0.2212 -> 0?

Yes, you are right. For `QT_8bit_direct`(without training) they are casting float to uint8 to serialize it. That's the reason I mentioned, it leads to complete loss of precision.
https://github.com/facebookresearch/faiss/blob/main/faiss/impl/ScalarQuantizer.cpp#L521-L531

[...]

PR #1823, opened 2024-07-12
Add HNSW changes to support Faiss byte vector

naveentatikonda:
### Description
Add HNSW changes to support Faiss byte vector which behind the scenes uses Faiss `SQ8_direct_signed` scalar quantizer. 
 
### Issues Resolved
https://github.com/opensearch-project/k-NN/issues/1659
 
### Check List
  - [x] All tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Features
 * Integrate Lucene Vector field with native engines to use KNNVectorFormat during segment creation [#1945](https://github.com/opensearch-project/k-NN/pull/1945)
 * k-NN query rescore support for native engines [#1984](https://github.com/opensearch-project/k-NN/pull/1984)
+* Add support for byte vector with Faiss Engine HNSW algorithm [#1823](https://github.com/opensearch-project/k-NN/pull/1823)
 ### Enhancements
 * Adds iterative graph build capability into a faiss index to improve the memory footprint during indexing and Integrates KNNVectorsFormat for native engines[#1950](https://github.com/opensearch-project/k-NN/pull/1950)
 ### Bug Fixes
--- a/jni/include/commons.h
+++ b/jni/include/commons.h
@@ -49,6 +49,7 @@ namespace knn_jni {
          * @param memoryAddress The address of the memory location where data will be stored.
          * @param data 2D byte array containing binary data to be stored in native memory.
          * @param initialCapacity The initial capacity of the memory location.
+         * @param append whether to append or start from index 0 when called subsequently with the same address
          * @return memory address of std::vector<uint8_t> where the data is stored.
          */
         jlong storeBinaryVectorData(knn_jni::JNIUtilInterface *, JNIEnv *, jlong , jobjectArray, jlong, jboolean);
@@ -63,6 +64,7 @@ namespace knn_jni {
         * @param memoryAddress The address of the memory location where data will be stored.
         * @param data 2D byte array containing int8 data to be stored in native memory.
         * @param initialCapacity The initial capacity of the memory location.
+        * @param append whether to append or start from index 0 when called subsequently with the same address
         * @return memory address of std::vector<int8_t> where the data is stored.
         */
         jlong storeByteVectorData(knn_jni::JNIUtilInterface *, JNIEnv *, jlong , jobjectArray, jlong, jboolean);
--- a/jni/include/commons.h
+++ b/jni/include/commons.h
@@ -47,10 +47,24 @@ namespace knn_jni {
          * CAUTION: The behavior is undefined if the memory address is deallocated and the method is called
          *
          * @param memoryAddress The address of the memory location where data will be stored.
-         * @param data 2D byte array containing data to be stored in native memory.
+         * @param data 2D byte array containing binary data to be stored in native memory.
          * @param initialCapacity The initial capacity of the memory location.
          * @return memory address of std::vector<uint8_t> where the data is stored.
          */
+        jlong storeBinaryVectorData(knn_jni::JNIUtilInterface *, JNIEnv *, jlong , jobjectArray, jlong, jboolean);
+
+        /**
+        * This is utility function that can be used to store signed int8 data in native memory. This function will allocate memory for
+        * the data(rows*columns) with initialCapacity and return the memory address where the data is stored.
+        * If you are using this function for first time use memoryAddress = 0 to ensure that a new memory location is created.
+        * For subsequent calls you can pass the same memoryAddress. If the data cannot be stored in the memory location
+        * will throw Exception.
+        *
+        * @param memoryAddress The address of the memory location where data will be stored.
+        * @param data 2D byte array containing int8 data to be stored in native memory.
+        * @param initialCapacity The initial capacity of the memory location.
+        * @return memory address of std::vector<int8_t> where the data is stored.
+        */
         jlong storeByteVectorData(knn_jni::JNIUtilInterface *, JNIEnv *, jlong , jobjectArray, jlong, jboolean);
 
         /**
@@ -68,6 +82,14 @@ namespace knn_jni {
          * @param memoryAddress address to be freed.
          */
         void freeByteVectorData(jlong);
+
+        /**
+         * Free up the memory allocated for the data stored in memory address. This function should be used with the memory
+         * address returned by {@link JNICommons#storeBinaryVectorData(long, byte[][], long, long)}
+         *
+         * @param memoryAddress address to be freed.
+         */
+        void freeBinaryVectorData(jlong);
 
         /**
          * Extracts query time efSearch from method parameters
--- a/jni/include/jni_util.h
+++ b/jni/include/jni_util.h
@@ -71,8 +71,10 @@ namespace knn_jni {
 
         virtual void Convert2dJavaObjectArrayAndStoreToFloatVector(JNIEnv *env, jobjectArray array2dJ,
                                                                    int dim, std::vector<float> *vect ) = 0;
-        virtual void Convert2dJavaObjectArrayAndStoreToByteVector(JNIEnv *env, jobjectArray array2dJ,
+        virtual void Convert2dJavaObjectArrayAndStoreToBinaryVector(JNIEnv *env, jobjectArray array2dJ,
                                                                    int dim, std::vector<uint8_t> *vect ) = 0;
+        virtual void Convert2dJavaObjectArrayAndStoreToByteVector(JNIEnv *env, jobjectArray array2dJ,
+                                                                           int dim, std::vector<int8_t> *vect ) = 0;
 
         virtual std::vector<int64_t> ConvertJavaIntArrayToCppIntVector(JNIEnv *env, jintArray arrayJ) = 0;
 
@@ -173,7 +175,8 @@ namespace knn_jni {
         void SetObjectArrayElement(JNIEnv *env, jobjectArray array, jsize index, jobject val);
         void SetByteArrayRegion(JNIEnv *env, jbyteArray array, jsize start, jsize len, const jbyte * buf);
         void Convert2dJavaObjectArrayAndStoreToFloatVector(JNIEnv *env, jobjectArray array2dJ, int dim, std::vector<float> *vect);
-        void Convert2dJavaObjectArrayAndStoreToByteVector(JNIEnv *env, jobjectArray array2dJ, int dim, std::vector<uint8_t> *vect);
+        void Convert2dJavaObjectArrayAndStoreToBinaryVector(JNIEnv *env, jobjectArray array2dJ, int dim, std::vector<uint8_t> *vect);
+        void Convert2dJavaObjectArrayAndStoreToByteVector(JNIEnv *env, jobjectArray array2dJ, int dim, std::vector<int8_t> *vect);
 
     private:
         std::unordered_map<std::string, jclass> cachedClasses;
--- a/jni/src/commons.cpp
+++ b/jni/src/commons.cpp
@@ -37,7 +37,7 @@ jlong knn_jni::commons::storeVectorData(knn_jni::JNIUtilInterface *jniUtil, JNIE
     return (jlong) vect;
 }
 
-jlong knn_jni::commons::storeByteVectorData(knn_jni::JNIUtilInterface *jniUtil, JNIEnv *env, jlong memoryAddressJ,
+jlong knn_jni::commons::storeBinaryVectorData(knn_jni::JNIUtilInterface *jniUtil, JNIEnv *env, jlong memoryAddressJ,
                                         jobjectArray dataJ, jlong initialCapacityJ, jboolean appendJ) {
     std::vector<uint8_t> *vect;
     if ((long) memoryAddressJ == 0) {
@@ -49,6 +49,26 @@ jlong knn_jni::commons::storeByteVectorData(knn_jni::JNIUtilInterface *jniUtil, 
 
     if (appendJ == JNI_FALSE) {
         vect->clear();
+    }
+
+    int dim = jniUtil->GetInnerDimensionOf2dJavaByteArray(env, dataJ);
+    jniUtil->Convert2dJavaObjectArrayAndStoreToBinaryVector(env, dataJ, dim, vect);
+
+    return (jlong) vect;
+}
+
+jlong knn_jni::commons::storeByteVectorData(knn_jni::JNIUtilInterface *jniUtil, JNIEnv *env, jlong memoryAddressJ,
+                                        jobjectArray dataJ, jlong initialCapacityJ, jboolean appendJ) {
+    std::vector<int8_t> *vect;
+    if (memoryAddressJ == 0) {
+        vect = new std::vector<int8_t>();
+        vect->reserve(static_cast<long>(initialCapacityJ));
+    } else {
+        vect = reinterpret_cast<std::vector<int8_t>*>(memoryAddressJ);
+    }
+
+    if (appendJ == JNI_FALSE) {
+            vect->clear();
     }
 
     int dim = jniUtil->GetInnerDimensionOf2dJavaByteArray(env, dataJ);
@@ -64,9 +84,16 @@ void knn_jni::commons::freeVectorData(jlong memoryAddressJ) {
     }
 }
 
-void knn_jni::commons::freeByteVectorData(jlong memoryAddressJ) {
+void knn_jni::commons::freeBinaryVectorData(jlong memoryAddressJ) {
     if (memoryAddressJ != 0) {
         auto *vect = reinterpret_cast<std::vector<uint8_t>*>(memoryAddressJ);
+        delete vect;
+    }
+}
+
+void knn_jni::commons::freeByteVectorData(jlong memoryAddressJ) {
+    if (memoryAddressJ != 0) {
+        auto *vect = reinterpret_cast<std::vector<int8_t>*>(memoryAddressJ);
         delete vect;
     }
 }
--- a/jni/src/faiss_index_service.cpp
+++ b/jni/src/faiss_index_service.cpp
@@ -250,5 +250,120 @@ void BinaryIndexService::writeIndex(
     }
 }
 
+ByteIndexService::ByteIndexService(std::unique_ptr<FaissMethods> faissMethods) : IndexService(std::move(faissMethods)) {}
+
+void ByteIndexService::allocIndex(faiss::Index * index, size_t dim, size_t numVectors) {
+    if(auto * indexHNSWSQ = dynamic_cast<faiss::IndexHNSWSQ *>(index)) {
+        if(auto * indexScalarQuantizer = dynamic_cast<faiss::IndexScalarQuantizer *>(indexHNSWSQ->storage)) {
+            indexScalarQuantizer->codes.reserve(indexScalarQuantizer->code_size * numVectors);
+        }
+        return;
+    }
+}
+
+jlong ByteIndexService::initIndex(
+        knn_jni::JNIUtilInterface * jniUtil,
+        JNIEnv * env,
+        faiss::MetricType metric,
+        std::string indexDescription,
+        int dim,
+        int numVectors,
+        int threadCount,
+        std::unordered_map<std::string, jobject> parameters
+    ) {
+    // Create index using Faiss factory method
+    std::unique_ptr<faiss::Index> index(faissMethods->indexFactory(dim, indexDescription.c_str(), metric));
+
+    // Set thread count if it is passed in as a parameter. Setting this variable will only impact the current thread
+    if(threadCount != 0) {
+        omp_set_num_threads(threadCount);
+    }
+
+    // Add extra parameters that cant be configured with the index factory
+    SetExtraParameters<faiss::Index, faiss::IndexIVF, faiss::IndexHNSW>(jniUtil, env, parameters, index.get());
+
+    // Check that the index does not need to be trained
+    if(!index->is_trained) {
+        throw std::runtime_error("Index is not trained");
+    }
+
+    std::unique_ptr<faiss::IndexIDMap> idMap (faissMethods->indexIdMap(index.get()));
+    //Makes sure the index is deleted when the destructor is called, this cannot be passed in the constructor
+    idMap->own_fields = true;
+
+    allocIndex(dynamic_cast<faiss::Index *>(idMap->index), dim, numVectors);
+
+    //Release the ownership so as to make sure not delete the underlying index that is created. The index is needed later
+    //in insert and write operations
+    index.release();
+    return reinterpret_cast<jlong>(idMap.release());
+}
+
+void ByteIndexService::insertToIndex(
+        int dim,
+        int numIds,
+        int threadCount,
+        int64_t vectorsAddress,
+        std::vector<int64_t> & ids,
+        jlong idMapAddress
+    ) {
+    // Read vectors from memory address
+    auto *inputVectors = reinterpret_cast<std::vector<int8_t>*>(vectorsAddress);
+
+    // The number of vectors can be int here because a lucene segment number of total docs never crosses INT_MAX value
+    int numVectors = inputVectors->size() / dim;
+    if(numVectors == 0) {
+        throw std::runtime_error("Number of vectors cannot be 0");
+    }
+
+    if (numIds != numVectors) {
+        throw std::runtime_error("Number of IDs does not match number of vectors");
+    }
+
+    // Set thread count if it is passed in as a parameter. Setting this variable will only impact the current thread
+    if(threadCount != 0) {
+        omp_set_num_threads(threadCount);
+    }
+
+    faiss::IndexIDMap * idMap = reinterpret_cast<faiss::IndexIDMap *> (idMapAddress);
+
+    // Add vectors in batches by casting int8 vectors into float with a batch size of 1000
+    int batchSize = 1000;
+    std::vector <float> inputFloatVectors(batchSize * dim);
+    std::vector <int64_t> floatVectorsIds(batchSize);
+    int id = 0;
+    auto iter = inputVectors->begin();
+
+    for (int id = 0; id < numVectors; id += batchSize) {
+        if (numVectors - id < batchSize) {
+            batchSize = numVectors - id;
+            inputFloatVectors.resize(batchSize * dim);
+            floatVectorsIds.resize(batchSize);
+        }
+
+        for (int i = 0; i < batchSize; ++i) {
+            floatVectorsIds[i] = ids[id + i];
+            for (int j = 0; j < dim; ++j, ++iter) {
+                inputFloatVectors[i * dim + j] = static_cast<float>(*iter);
+            }
+        }
+        idMap->add_with_ids(batchSize, inputFloatVectors.data(), floatVectorsIds.data());
+    }
+}
+
+void ByteIndexService::writeIndex(
+        std::string indexPath,
+        jlong idMapAddress
+    ) {
+    std::unique_ptr<faiss::IndexIDMap> idMap (reinterpret_cast<faiss::IndexIDMap *> (idMapAddress));
+
+    try {
+        // Write the index to disk
+        faissMethods->writeIndex(idMap.get(), indexPath.c_str());
+    } catch(std::exception &e) {
+        throw std::runtime_error("Failed to write index to disk");
+    }
+}
+
 } // namespace faiss_wrapper
 } // namesapce knn_jni
--- a/jni/src/faiss_index_service.cpp
+++ b/jni/src/faiss_index_service.cpp
@@ -327,7 +327,8 @@ void ByteIndexService::insertToIndex(
 
     faiss::IndexIDMap * idMap = reinterpret_cast<faiss::IndexIDMap *> (idMapAddress);
 
-    // Add vectors in batches by casting int8 vectors into float with a batch size of 1000
+    // Add vectors in batches by casting int8 vectors into float with a batch size of 1000 to avoid additional memory spike.
+    // Refer to this github issue for more details https://github.com/opensearch-project/k-NN/issues/1659#issuecomment-2307390255
     int batchSize = 1000;
     std::vector <float> inputFloatVectors(batchSize * dim);
     std::vector <int64_t> floatVectorsIds(batchSize);
@@ -337,8 +338,6 @@ void ByteIndexService::insertToIndex(
     for (int id = 0; id < numVectors; id += batchSize) {
         if (numVectors - id < batchSize) {
             batchSize = numVectors - id;
-            inputFloatVectors.resize(batchSize * dim);
-            floatVectorsIds.resize(batchSize);
         }
 
         for (int i = 0; i < batchSize; ++i) {
@@ -364,6 +363,5 @@ void ByteIndexService::writeIndex(
         throw std::runtime_error("Failed to write index to disk");
     }
 }
-
 } // namespace faiss_wrapper
 } // namesapce knn_jni
--- a/jni/src/jni_util.cpp
+++ b/jni/src/jni_util.cpp
@@ -261,7 +261,7 @@ void knn_jni::JNIUtil::Convert2dJavaObjectArrayAndStoreToFloatVector(JNIEnv *env
     env->DeleteLocalRef(array2dJ);
 }
 
-void knn_jni::JNIUtil::Convert2dJavaObjectArrayAndStoreToByteVector(JNIEnv *env, jobjectArray array2dJ,
+void knn_jni::JNIUtil::Convert2dJavaObjectArrayAndStoreToBinaryVector(JNIEnv *env, jobjectArray array2dJ,
                                                                      int dim, std::vector<uint8_t> *vect) {
 
     if (array2dJ == nullptr) {
@@ -288,6 +288,38 @@ void knn_jni::JNIUtil::Convert2dJavaObjectArrayAndStoreToByteVector(JNIEnv *env,
         for(int j = 0; j < dim; ++j) {
             vect->push_back(vector[j]);
         }
+        env->ReleaseByteArrayElements(vectorArray, reinterpret_cast<int8_t*>(vector), JNI_ABORT);
+    }
+    this->HasExceptionInStack(env);
+    env->DeleteLocalRef(array2dJ);
+}
+
+void knn_jni::JNIUtil::Convert2dJavaObjectArrayAndStoreToByteVector(JNIEnv *env, jobjectArray array2dJ,
+                                                                     int dim, std::vector<int8_t> *vect) {
+
+    if (array2dJ == nullptr) {
+        throw std::runtime_error("Array cannot be null");
+    }
+
+    int numVectors = env->GetArrayLength(array2dJ);
+    this->HasExceptionInStack(env, "Unable to get array length");
+
+    for (int i = 0; i < numVectors; ++i) {
+        auto vectorArray = static_cast<jbyteArray>(env->GetObjectArrayElement(array2dJ, i));
+        this->HasExceptionInStack(env, "Unable to get object array element");
+
+        if (dim != env->GetArrayLength(vectorArray)) {
+            env->DeleteLocalRef(array2dJ);
+            throw std::runtime_error("Dimension of vectors is inconsistent");
+        }
+
+        int8_t* vector = reinterpret_cast<int8_t*>(env->GetByteArrayElements(vectorArray, nullptr));
+        if (vector == nullptr) {
+            this->HasExceptionInStack(env);
+            throw std::runtime_error("Unable to get byte array elements");
+        }
+
+        vect->insert(vect->end(), vector, vector + dim);
         env->ReleaseByteArrayElements(vectorArray, reinterpret_cast<int8_t*>(vector), JNI_ABORT);
     }
     this->HasExceptionInStack(env);
--- a/jni/src/org_opensearch_knn_jni_FaissService.cpp
+++ b/jni/src/org_opensearch_knn_jni_FaissService.cpp
@@ -67,6 +67,20 @@ JNIEXPORT jlong JNICALL Java_org_opensearch_knn_jni_FaissService_initBinaryIndex
     return (jlong)0;
 }
 
+JNIEXPORT jlong JNICALL Java_org_opensearch_knn_jni_FaissService_initByteIndex(JNIEnv * env, jclass cls,
+                                                                            jlong numDocs, jint dimJ,
+                                                                            jobject parametersJ)
+{
+    try {
+        std::unique_ptr<knn_jni::faiss_wrapper::FaissMethods> faissMethods(new knn_jni::faiss_wrapper::FaissMethods());
+        knn_jni::faiss_wrapper::ByteIndexService byteIndexService(std::move(faissMethods));
+        return knn_jni::faiss_wrapper::InitIndex(&jniUtil, env, numDocs, dimJ, parametersJ, &byteIndexService);
+    } catch (...) {
+        jniUtil.CatchCppExceptionAndThrowJava(env);
+    }
+    return (jlong)0;
+}
+
 JNIEXPORT void JNICALL Java_org_opensearch_knn_jni_FaissService_insertToIndex(JNIEnv * env, jclass cls, jintArray idsJ,
                                                                             jlong vectorsAddressJ, jint dimJ,
                                                                             jlong indexAddress, jint threadCount)
@@ -95,6 +109,20 @@ JNIEXPORT void JNICALL Java_org_opensearch_knn_jni_FaissService_insertToBinaryIn
     }
 }
 
+JNIEXPORT void JNICALL Java_org_opensearch_knn_jni_FaissService_insertToByteIndex(JNIEnv * env, jclass cls, jintArray idsJ,
+                                                                            jlong vectorsAddressJ, jint dimJ,
+                                                                            jlong indexAddress, jint threadCount)
+{
+    try {
+        std::unique_ptr<knn_jni::faiss_wrapper::FaissMethods> faissMethods(new knn_jni::faiss_wrapper::FaissMethods());
+        knn_jni::faiss_wrapper::ByteIndexService byteIndexService(std::move(faissMethods));
+        knn_jni::faiss_wrapper::InsertToIndex(&jniUtil, env, idsJ, vectorsAddressJ, dimJ, indexAddress, threadCount, &byteIndexService);
+    } catch (...) {
+        // NOTE: ADDING DELETE STATEMENT HERE CAUSES A CRASH!
+        jniUtil.CatchCppExceptionAndThrowJava(env);
+    }
+}
+
 JNIEXPORT void JNICALL Java_org_opensearch_knn_jni_FaissService_writeIndex(JNIEnv * env, jclass cls,
                                                                             jlong indexAddress,
                                                                             jstring indexPathJ)
@@ -116,6 +144,19 @@ JNIEXPORT void JNICALL Java_org_opensearch_knn_jni_FaissService_writeBinaryIndex
         std::unique_ptr<knn_jni::faiss_wrapper::FaissMethods> faissMethods(new knn_jni::faiss_wrapper::FaissMethods());
         knn_jni::faiss_wrapper::BinaryIndexService binaryIndexService(std::move(faissMethods));
         knn_jni::faiss_wrapper::WriteIndex(&jniUtil, env, indexPathJ, indexAddress, &binaryIndexService);
+    } catch (...) {
+        jniUtil.CatchCppExceptionAndThrowJava(env);
+    }
+}
+
+JNIEXPORT void JNICALL Java_org_opensearch_knn_jni_FaissService_writeByteIndex(JNIEnv * env, jclass cls,
+                                                                            jlong indexAddress,
+                                                                            jstring indexPathJ)
+{
+    try {
+        std::unique_ptr<knn_jni::faiss_wrapper::FaissMethods> faissMethods(new knn_jni::faiss_wrapper::FaissMethods());
+        knn_jni::faiss_wrapper::ByteIndexService byteIndexService(std::move(faissMethods));
+        knn_jni::faiss_wrapper::WriteIndex(&jniUtil, env, indexPathJ, indexAddress, &byteIndexService);
     } catch (...) {
         jniUtil.CatchCppExceptionAndThrowJava(env);
     }
--- a/src/main/java/org/opensearch/knn/index/codec/transfer/OffHeapBinaryVectorTransfer.java
+++ b/src/main/java/org/opensearch/knn/index/codec/transfer/OffHeapBinaryVectorTransfer.java
@@ -5,6 +5,8 @@
 
 package org.opensearch.knn.index.codec.transfer;
 
+import org.opensearch.knn.jni.JNICommons;
+
 import java.io.IOException;
 import java.util.List;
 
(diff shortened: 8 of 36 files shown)


src/main/java/org/opensearch/knn/index/codec/transfer/VectorTransferByteToFloat.java
```diff
@@ -0,0 +1,78 @@
...
+        return SerializationMode.COLLECTION_OF_FLOATS;
+    }
+
+    private void transfer() {
+        vectorAddress = JNICommons.storeVectorData(vectorAddress, vectorList.toArray(new float[][] {}), totalLiveDocs * dimension);
+        vectorList.clear();
+    }
+
+    // cast byte vector array into float vector array
+    private static float[] byteToFloatArray(ByteArrayInputStream byteStream) {
```
Vikasht34, line 69: ByeBuffer is special class given for Manipulating Bytes . ByteBuffer to read bytes as floats directly, leveraging native optimizations and reducing the number of operations required

```
private static float[] byteToFloatArray(ByteArrayInputStream byteStream) {
        byte[] vectorAsByteArray = byteStream.readAllBytes();
        int sizeOfFloatArray = vectorAsByteArray.length;
        float[] vector = new float[sizeOfFloatArray];

        ByteBuffer byteBuffer = ByteBuffer.wrap(vectorAsByteArray).order(ByteOrder.nativeOrder());
        for (int i = 0; i < sizeOfFloatArray; i++) {
            vector[i] = byteBuffer.get(i);
        }

        return vector;
    } 
```

src/main/java/org/opensearch/knn/index/codec/transfer/VectorTransferByteToFloat.java
```diff
@@ -0,0 +1,78 @@
...
+import org.opensearch.knn.jni.JNICommons;
+
+import java.io.ByteArrayInputStream;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Vector transfer for byte vectors by casting them as float vectors
+ */
+public class VectorTransferByteToFloat extends VectorTransfer {
```
Vikasht34, line 18: Why we need to cast then for vector transfer?

src/main/java/org/opensearch/knn/index/mapper/MethodFieldMapper.java
```diff
@@ -58,10 +62,16 @@ public class MethodFieldMapper extends KNNVectorFieldMapper {
...
-            this.fieldType.putAttribute(
-                PARAMETERS,
-                XContentFactory.jsonBuilder().map(knnEngine.getMethodAsMap(knnMethodContext)).toString()
-            );
+            Map<String, Object> methodParamsMap = knnEngine.getMethodAsMap(knnMethodContext);
+
+            // If VectorDataType is Byte using Faiss engine then manipulate Index Description to use "SQ8_direct_signed" scalar quantizer
+            // For example, Index Description "HNSW16,Flat" will be updated as "HNSW16,SQ8_direct_signed"
+            if (VectorDataType.BYTE.equals(vectorDataType) && methodParamsMap.containsKey(INDEX_DESCRIPTION_PARAMETER)) {
+                String indexDescriptionValue = (String) methodParamsMap.get(INDEX_DESCRIPTION_PARAMETER);
```
Vikasht34, line 70: There should be null check on indexDescriptionValue , as we are calling split on that , it can throw NPE

src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80DocValuesConsumer.java
```diff
@@ -115,6 +116,10 @@ private VectorTransfer getVectorTransfer(FieldInfo field) {
         if (VectorDataType.BINARY.getValue().equalsIgnoreCase(field.attributes().get(KNNConstants.VECTOR_DATA_TYPE_FIELD))) {
             return new VectorTransferByte(KNNSettings.getVectorStreamingMemoryLimit().getBytes());
         }
+        if (VectorDataType.BYTE.getValue().equalsIgnoreCase(field.attributes().get(KNNConstants.VECTOR_DATA_TYPE_FIELD))
```
Vikasht34, line 119: This function is getting overloaded with multiples ifs , Can we use Switch for more readability  and what if vectorDataType is null, Can we also check Null

```
private VectorTransfer getVectorTransfer(FieldInfo field) {
    String vectorDataType = field.attributes().get(KNNConstants.VECTOR_DATA_TYPE_FIELD);
    String knnEngine = field.attributes().get(KNNConstants.KNN_ENGINE);
    long memoryLimit = KNNSettings.getVectorStreamingMemoryLimit().getBytes();

    if (vectorDataType == null) {
        throw new IllegalArgumentException("VECTOR_DATA_TYPE_FIELD attribute is missing");
    }

    switch (vectorDataType.toLowerCase()) {
        case "binary":
            return new VectorTransferByte(memoryLimit);
        case "byte":
            if (KNNEngine.FAISS.getName().equalsIgnoreCase(knnEngine)) {
                return new VectorTransferByteToFloat(memoryLimit);
            }
        default:
            return new VectorTransferFloat(memoryLimit);
    }
}
```

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
```diff
@@ -605,8 +609,16 @@ protected Query doToQuery(QueryShardContext context) {
                 .knnEngine(knnEngine)
                 .indexName(indexName)
                 .fieldName(this.fieldName)
-                .vector(VectorDataType.FLOAT == vectorDataType ? this.vector : null)
-                .byteVector(VectorDataType.BYTE == vectorDataType || VectorDataType.BINARY == vectorDataType ? byteVector : null)
+                .vector(
```
Vikasht34, line 612: Can we simplify here , conditions are getting broader ? Probably we can get requestbuilder first with params which which is common and then setting specific one based on condition.

src/main/java/org/opensearch/knn/index/codec/transfer/VectorTransferByteToFloat.java
naveentatikonda: As we are using scalar quantizer behind the scenes which expects input as float vector. So, while deserializing it back to byte vector casting it as float vector and then transferring it

src/main/java/org/opensearch/knn/index/mapper/MethodFieldMapper.java
naveentatikonda: Ack

src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80DocValuesConsumer.java
naveentatikonda: Refactored it to use switch case. But, I don't think we need a null check here because vectorDataType is an optional field which is by default set to float.

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
naveentatikonda: Refactored it by extracting the conditions into 2 separate methods

src/main/java/org/opensearch/knn/index/codec/transfer/VectorTransferByteToFloat.java
```diff
@@ -0,0 +1,81 @@
...
+import java.io.ByteArrayInputStream;
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Vector transfer for byte vectors by casting them as float vectors
+ */
+public class VectorTransferByteToFloat extends VectorTransfer {
```
navneet1v, line 20: This class is unnecessary. Already have classes to transfer bytes and floats to JNI layer. Let reuse those classes to do so. 

If we have to do this, it means the interface of VectorTransfer is not solid enough and needs more thought process.

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java
```diff
@@ -349,7 +349,7 @@ private void validateEncoder(final KNNMethodContext knnMethodContext, final Vect
                 return;
             }
 
-            if (VectorDataType.BINARY != vectorDataType) {
+            if (VectorDataType.FLOAT == vectorDataType) {
```
navneet1v, line 352: why this change?

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapperUtil.java
```diff
@@ -105,16 +105,17 @@ public static void validateVectorDataType(KNNMethodContext methodContext, Vector
         }
 
         if (VectorDataType.BYTE == vectorDataType) {
-            if (KNNEngine.LUCENE == methodContext.getKnnEngine()) {
+            if (KNNEngine.LUCENE == methodContext.getKnnEngine() || KNNEngine.FAISS == methodContext.getKnnEngine()) {
```
navneet1v, line 108: better to put these engines in a list which says what engines support ByteVectors

src/main/java/org/opensearch/knn/index/mapper/MethodFieldMapper.java
```diff
@@ -58,10 +62,18 @@ public class MethodFieldMapper extends KNNVectorFieldMapper {
...
         try {
-            this.fieldType.putAttribute(
-                PARAMETERS,
-                XContentFactory.jsonBuilder().map(knnEngine.getMethodAsMap(knnMethodContext)).toString()
-            );
+            Map<String, Object> methodParamsMap = knnEngine.getMethodAsMap(knnMethodContext);
+
+            // If VectorDataType is Byte using Faiss engine then manipulate Index Description to use "SQ8_direct_signed" scalar quantizer
+            // For example, Index Description "HNSW16,Flat" will be updated as "HNSW16,SQ8_direct_signed"
+            if (VectorDataType.BYTE.equals(vectorDataType) && methodParamsMap.containsKey(INDEX_DESCRIPTION_PARAMETER)) {
```
navneet1v, line 69: anywhere you use enum use == rather than equals its faster and provides you complie time validations.

src/main/java/org/opensearch/knn/index/mapper/MethodFieldMapper.java
```diff
@@ -58,10 +62,18 @@ public class MethodFieldMapper extends KNNVectorFieldMapper {
...
+                String indexDescriptionValue = (String) methodParamsMap.get(INDEX_DESCRIPTION_PARAMETER);
+                if (indexDescriptionValue != null && !indexDescriptionValue.isEmpty()) {
```
navneet1v, line 71:
```suggestion
                if (indexDescriptionValue != null && indexDescriptionValue.isEmpty() == false) {
```

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
```diff
@@ -582,10 +582,14 @@ protected Query doToQuery(QueryShardContext context) {
             }
             spaceType.validateVector(byteVector);
         } else if (VectorDataType.BYTE == vectorDataType) {
-            byteVector = new byte[vector.length];
+            if (KNNEngine.LUCENE.equals(knnEngine)) {
```
navneet1v, line 585: use ==

src/main/java/org/opensearch/knn/index/codec/transfer/VectorTransferByteToFloat.java
```diff
@@ -0,0 +1,81 @@
...
+
+    @Override
+    public SerializationMode getSerializationMode(ByteArrayInputStream byteStream) {
+        return SerializationMode.COLLECTION_OF_FLOATS;
+    }
+
+    private void transfer() {
+        vectorAddress = JNICommons.storeVectorData(vectorAddress, vectorList.toArray(new float[][] {}), totalLiveDocs * dimension);
+        vectorList.clear();
+    }
```
navneet1v, line 68: If you are just going to transfer vectors are floats to JNI layer then the memory savings at the indexing layer will not be present because you are still transferring byte values as float point values. This is a big concern for me for enabling the byte vector support in faiss. and I can see these 2 paths here:

1. why we are not exposing a byte array based interfaces from faiss
2. Lets just transfer bytes to JNI layer and then on the JNI layer we convert byte vector to float vector one by one and send it faiss one by one for graph creation. This will ensure that memory is not spiking during indexing.

if 1 is not possible then we should definitely use 2 here otherwise benefits of byte vector for floats will not be same as that of Lucene.

navneet1v (changes requested):

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 * Adds dynamic query parameter ef_search [#1783](https://github.com/opensearch-project/k-NN/pull/1783)
 * Adds dynamic query parameter ef_search in radial search faiss engine [#1790](https://github.com/opensearch-project/k-NN/pull/1790)
 * Add binary format support with HNSW method in Faiss Engine [#1781](https://github.com/opensearch-project/k-NN/pull/1781)
+* Add HNSW changes to support Faiss byte vector [#1823](https://github.com/opensearch-project/k-NN/pull/1823)
```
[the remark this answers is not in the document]
naveentatikonda: For now we are supporting byte vector using HNSW and IVF as we are implementing it using SQ. This PR will only add support for HNSW changes. For IVF, we need to add `data_type` field in train model API which Junqiu is working on in this PR(#1784). Once it is merged, will raise a separate PR on top of that.

src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80DocValuesConsumer.java
```diff
@@ -112,10 +113,20 @@ private KNNEngine getKNNEngine(@NonNull FieldInfo field) {
...
-        if (VectorDataType.BINARY.getValue().equalsIgnoreCase(field.attributes().get(KNNConstants.VECTOR_DATA_TYPE_FIELD))) {
-            return new VectorTransferByte(KNNSettings.getVectorStreamingMemoryLimit().getBytes());
+        String vectorDataType = field.attributes().getOrDefault(KNNConstants.VECTOR_DATA_TYPE_FIELD, VectorDataType.FLOAT.getValue());
+        String knnEngine = field.attributes().get(KNNConstants.KNN_ENGINE);
+        long memoryLimit = KNNSettings.getVectorStreamingMemoryLimit().getBytes();
+
+        switch (vectorDataType.toLowerCase()) {
+            case "binary":
+                return new VectorTransferByte(memoryLimit);
+            case "byte":
```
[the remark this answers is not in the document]
naveentatikonda: I have tried but that gave me some syntax errors. Will check again

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java
naveentatikonda: Like Binary we are not supporting any encoder for byte vector as we are already using SQ. So, updated the conditions.

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapperUtil.java
naveentatikonda: Ack

src/main/java/org/opensearch/knn/index/codec/transfer/VectorTransferByteToFloat.java
naveentatikonda: Got it. The Scalar quantizer in faiss accepts input as fp32 datatype. So, it might need a lot of changes and first approach may not be feasible. Will implement the second approach. Thanks!

jni/src/faiss_index_service.cpp
```diff
@@ -160,5 +160,119 @@ void BinaryIndexService::createIndex(
...
+        // Check that the index does not need to be trained
+        if (!indexWriter->is_trained) {
+            throw std::runtime_error("Index is not trained");
+        }
+
+        return indexWriter;
+
+    }
+
+    void ByteIndexService::addVectorsToIndex(
```
heemin32, line 210: You don't need smart pointer here for indexWriter. The caller might want to keep the ownership and release it in their own scope.

```
    void ByteIndexService::addVectorsToIndex(
            faiss::Index* indexWriter,
            std::vector <int8_t> *inputVectors,
            int dim,
            std::vector <int64_t> ids,
            std::string indexPath
    ) {
        std::unique_ptr <faiss::IndexIDMap> idMap(faissMethods->indexIdMap(indexWriter.get()));
        int numVectors = (int) (inputVectors->size() / (uint64_t) dim);

        int batchSize = 1000;
        std::vector <float> inputFloatVectors;
        vector.reserve(batchSize * dim);
        std::vector <int64_t> floatVectorsIds;
        floatVectorsIds.reserve(batchSize);
        int pos = 0;
        std::vector<int8_t>::iterator iter = inputVectors->begin();

        while (pos < numVectors) {
            for (int i = 0;i < batchSize && pos < numVectors; i++, pos++) {
                floatVectorsIds.push_back(ids[pos]);
                for (int j = 0; j < dim; j++) {
                    inputFloatVectors->push_back((float) *iter);
                    iter++;
                }
            }
            idMap->add_with_ids(batchSize, inputFloatVectors.data(), floatVectorsIds.data());
            inputFloatVectors.clear();
            floatVectorsIds.clear();
        }

        // Write the index to disk
        faissMethods->writeIndex(idMap.get(), indexPath.c_str());
    }
```

src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80DocValuesConsumer.java
```diff
@@ -359,10 +360,18 @@ private boolean isChecksumValid(long value) {
...
-    private VectorTransfer getVectorTransfer(VectorDataType vectorDataType) {
-        if (VectorDataType.BINARY == vectorDataType) {
-            return new VectorTransferByte(KNNSettings.getVectorStreamingMemoryLimit().getBytes());
+    private VectorTransfer getVectorTransfer(VectorDataType vectorDataType, KNNEngine knnEngine) {
+        long memoryLimit = KNNSettings.getVectorStreamingMemoryLimit().getBytes();
+        switch (vectorDataType) {
+            case BINARY:
+                return new VectorTransferBinary(memoryLimit);
+            case BYTE:
+                if (KNNEngine.FAISS == knnEngine) {
```
heemin32, line 369: I think this is only either faiss or nmslib. nmslib does not support byte. Therefore, we don't need this condition check.

src/main/java/org/opensearch/knn/index/util/IndexUtil.java
```diff
@@ -412,4 +412,17 @@ private static Map<String, Version> initializeMinimalRequiredVersionMap() {
...
+
+    /**
+     * Tell if it is byte index or not
+     *
+     * @param knnEngine knn engine associated with an index
+     * @param parameters parameters associated with an index
+     * @return true if it is binary index
+     */
+    public static boolean isByteIndex(KNNEngine knnEngine, Map<String, Object> parameters) {
+        return KNNEngine.FAISS == knnEngine
```
heemin32, line 424: But we have byte index for Lucene right?
Shouldn't this method be `isFaissByteIndex`?

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
```diff
@@ -441,10 +441,14 @@ protected Query doToQuery(QueryShardContext context) {
             }
             spaceType.validateVector(byteVector);
         } else if (VectorDataType.BYTE == vectorDataType) {
-            byteVector = new byte[vector.length];
+            if (KNNEngine.LUCENE == knnEngine) {
```
heemin32, line 444: This is quite confusing. Can we just use byte here same as lucene and only convert it back to float right before calling JNI?

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased 2.x](https://github.com/opensearch-project/k-NN/compare/2.16...2.x)
 ### Features
+* Add HNSW changes to support Faiss byte vector [#1823](https://github.com/opensearch-project/k-NN/pull/1823)
```
navneet1v, line 18: will this be better as `Add support for byte vector with Faiss Engine HNSW algorithm`

jni/include/commons.h
```diff
@@ -48,6 +48,20 @@ namespace knn_jni {
...
+        * If you are using this function for first time use memoryAddress = 0 to ensure that a new memory location is created.
+        * For subsequent calls you can pass the same memoryAddress. If the data cannot be stored in the memory location
+        * will throw Exception.
+        *
+        * @param memoryAddress The address of the memory location where data will be stored.
+        * @param data 2D byte array containing int8 data to be stored in native memory.
+        * @param initialCapacity The initial capacity of the memory location.
+        * @return memory address of std::vector<int8_t> where the data is stored.
+        */
+        jlong storeByteVectorData(knn_jni::JNIUtilInterface *, JNIEnv *, jlong , jobjectArray, jlong);
```
navneet1v, line 63: should we just have storeByteVectorData rather than having 2. Because at the lowest level both are storing vectors as ByteArray.

navneet1v: NVM I saw that binary uses Signed int_8 and byte uses uint_8. I think we might need to move all these storing and freeing up of the memory in their respective IndexService. It can be huge refactor. @heemin32 , @jmazanec15 what is your thought. 

Reason of having all of them in common place is not because of code duplication, I am more worried about ensuring there are not leaks and edge cases are handled during this store and release of memory address.

jni/src/faiss_index_service.cpp
```diff
@@ -160,5 +160,119 @@ void BinaryIndexService::createIndex(
...
+            std::unique_ptr <faiss::Index> indexWriter,
+            std::vector <int8_t> *inputVectors,
+            int dim,
+            std::vector <int64_t> ids,
+            std::string indexPath
+    ) {
+        std::unique_ptr <faiss::IndexIDMap> idMap(faissMethods->indexIdMap(indexWriter.get()));
+        int numVectors = (int) (inputVectors->size() / (uint64_t) dim);
+
+        int batchSize = 1000;
```
navneet1v, line 220: why 1000?

jni/src/faiss_index_service.cpp
```diff
@@ -160,5 +160,119 @@ void BinaryIndexService::createIndex(
...
+                for (int j = 0; j < dim; j++) {
+                    inputFloatVectors->push_back((float) *iter);
+                    iter++;
+                }
+            }
+            idMap->add_with_ids(batchSize, inputFloatVectors->data(), floatVectorsIds.data());
+
+            totalNumVecs = totalNumVecs - batchSize;
+            inputFloatVectors->clear();
+            floatVectorsIds.clear();
```
navneet1v, line 245: clear doesn't free up the memory and also run with linear time complexity, as per the docs here:https://cplusplus.com/reference/vector/vector/clear/.

Instead I would recommend either overriding the values or using an array and override the values. This will ensure that same memory location is getting used and then we can destroy it later.

jni/src/faiss_index_service.cpp
```diff
@@ -160,5 +160,119 @@ void BinaryIndexService::createIndex(
...
+        std::vector <int64_t> floatVectorsIds;
+        int id = 0;
+        std::vector<int8_t>::iterator iter = inputVectors->begin();
+
+        while (id < numVectors) {
+            if (totalNumVecs < batchSize) {
+                batchSize = totalNumVecs;
+            }
+
+            inputFloatVectors->reserve(batchSize * dim);
```
navneet1v, line 232: not sure what will happen if we call multiple times reserve on an already created vector.

jni/src/faiss_index_service.cpp
```diff
@@ -160,5 +160,119 @@ void BinaryIndexService::createIndex(
...
+                floatVectorsIds.push_back(ids[id++]);
+                for (int j = 0; j < dim; j++) {
+                    inputFloatVectors->push_back((float) *iter);
+                    iter++;
+                }
+            }
+            idMap->add_with_ids(batchSize, inputFloatVectors->data(), floatVectorsIds.data());
+
+            totalNumVecs = totalNumVecs - batchSize;
+            inputFloatVectors->clear();
```
Vikasht34, line 244: with Resize approach , we don't have to worry about cleraring , it re-uses same memory and resize if needed.

```
for (int id = 0; id < numVectors; id += batchSize) {
    if (numVectors - id < batchSize) {
        batchSize = numVectors - id;
        inputFloatVectors.resize(batchSize * dim);
        floatVectorsIds.resize(batchSize);
    }

    for (int i = 0; i < batchSize; ++i) {
        floatVectorsIds[i] = ids[id + i];
        for (int j = 0; j < dim; ++j, ++iter) {
            inputFloatVectors[i * dim + j] = static_cast<float>(*iter);
        }
    }

    idMap->add_with_ids(batchSize, inputFloatVectors.data(), floatVectorsIds.data());
}
```

jni/src/jni_util.cpp
```diff
@@ -294,6 +294,40 @@ void knn_jni::JNIUtil::Convert2dJavaObjectArrayAndStoreToByteVector(JNIEnv *env,
...
 
+void knn_jni::JNIUtil::Convert2dJavaObjectArrayAndStoreToByteVector(JNIEnv *env, jobjectArray array2dJ,
+                                                                     int dim, std::vector<int8_t> *vect) {
+
+    if (array2dJ == nullptr) {
+        throw std::runtime_error("Array cannot be null");
+    }
+
+    int numVectors = env->GetArrayLength(array2dJ);
+    this->HasExceptionInStack(env);
```
Vikasht34, line 305: Should we add message like ""Failed to get array length" here

jni/src/jni_util.cpp
```diff
@@ -294,6 +294,40 @@ void knn_jni::JNIUtil::Convert2dJavaObjectArrayAndStoreToByteVector(JNIEnv *env,
...
+
+    for (int i = 0; i < numVectors; ++i) {
+        auto vectorArray = (jbyteArray)env->GetObjectArrayElement(array2dJ, i);
```
Vikasht34, line 308: Please avoid c_casts and start usinhg c++ casts like static_cast or another one.

jni/src/jni_util.cpp
```diff
@@ -294,6 +294,40 @@ void knn_jni::JNIUtil::Convert2dJavaObjectArrayAndStoreToByteVector(JNIEnv *env,
...
+            throw std::runtime_error("Dimension of vectors is inconsistent");
+        }
+
+        int8_t* vector = reinterpret_cast<int8_t*>(env->GetByteArrayElements(vectorArray, nullptr));
+        if (vector == nullptr) {
+            this->HasExceptionInStack(env);
+            throw std::runtime_error("Unable to get byte array elements");
+        }
+
+        for(int j = 0; j < dim; ++j) {
```
Vikasht34, line 322: We can use standard library given 

like 

vect->insert(vect->end(), vector, vector + dim);

jni/src/org_opensearch_knn_jni_FaissService.cpp
```diff
@@ -75,6 +75,24 @@ JNIEXPORT void JNICALL Java_org_opensearch_knn_jni_FaissService_createBinaryInde
     }
 }
 
+JNIEXPORT void JNICALL Java_org_opensearch_knn_jni_FaissService_createByteIndex(JNIEnv * env, jclass cls, jintArray idsJ,
+                                                                            jlong vectorsAddressJ, jint dimJ,
+                                                                            jstring indexPathJ, jobject parametersJ)
+{
+    try {
+        std::unique_ptr<knn_jni::faiss_wrapper::FaissMethods> faissMethods(new knn_jni::faiss_wrapper::FaissMethods());
```
Vikasht34, line 83: let's use std::make_unique to create objects which is already given for unique pointers

```
 auto faissMethods = std::make_unique<knn_jni::faiss_wrapper::FaissMethods>();
```

jni/src/org_opensearch_knn_jni_FaissService.cpp
```diff
@@ -75,6 +75,24 @@ JNIEXPORT void JNICALL Java_org_opensearch_knn_jni_FaissService_createBinaryInde
...
+        knn_jni::faiss_wrapper::ByteIndexService byteIndexService(std::move(faissMethods));
+        knn_jni::faiss_wrapper::CreateIndex(&jniUtil, env, idsJ, vectorsAddressJ, dimJ, indexPathJ, parametersJ, &byteIndexService);
+
+        // Releasing the vectorsAddressJ memory as that is not required once we have created the index.
+        // This is not the ideal approach, please refer this gh issue for long term solution:
+        // https://github.com/opensearch-project/k-NN/issues/1600
+        delete reinterpret_cast<std::vector<int8_t>*>(vectorsAddressJ);
```
Vikasht34, line 90: this automatically deleted when vectorPtr goes out of scope
```
std::unique_ptr<std::vector<int8_t>> vectorPtr(reinterpret_cast<std::vector<int8_t>*>(vectorsAddressJ));
```

Vikasht34 (commented):
Finished the First leg

jni/src/org_opensearch_knn_jni_FaissService.cpp
naveentatikonda: discussed offline... `std::make_unique` is supported only from C++14

jni/src/org_opensearch_knn_jni_FaissService.cpp
naveentatikonda: discussed offline and going with existing change as we are doing the same everywhere

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
```diff
@@ -441,10 +441,14 @@ protected Query doToQuery(QueryShardContext context) {
...
+                byteVector = new byte[vector.length];
+            }
             for (int i = 0; i < vector.length; i++) {
                 validateByteVectorValue(vector[i], knnVectorFieldType.getVectorDataType());
-                byteVector[i] = (byte) vector[i];
+                if (KNNEngine.LUCENE == knnEngine) {
+                    byteVector[i] = (byte) vector[i];
```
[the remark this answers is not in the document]
naveentatikonda: because vector is of type float, so we need to cast it into byte

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
```diff
@@ -441,10 +441,14 @@ protected Query doToQuery(QueryShardContext context) {
...
             spaceType.validateVector(byteVector);
         } else if (VectorDataType.BYTE == vectorDataType) {
-            byteVector = new byte[vector.length];
+            if (KNNEngine.LUCENE == knnEngine) {
+                byteVector = new byte[vector.length];
+            }
             for (int i = 0; i < vector.length; i++) {
                 validateByteVectorValue(vector[i], knnVectorFieldType.getVectorDataType());
-                byteVector[i] = (byte) vector[i];
+                if (KNNEngine.LUCENE == knnEngine) {
```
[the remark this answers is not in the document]
naveentatikonda: We can't use it here as we need to add separate logic for Lucene engine. Anyways refactored this into a switch case

src/main/java/org/opensearch/knn/index/mapper/MethodFieldMapper.java
```diff
@@ -60,6 +64,17 @@ public class MethodFieldMapper extends KNNVectorFieldMapper {
 
         try {
             Map<String, Object> libParams = knnEngine.getKNNLibraryIndexingContext(knnMethodContext).getLibraryParameters();
+
+            // If VectorDataType is Byte using Faiss engine then manipulate Index Description to use "SQ8_direct_signed" scalar quantizer
+            // For example, Index Description "HNSW16,Flat" will be updated as "HNSW16,SQ8_direct_signed"
+            if (VectorDataType.BYTE == vectorDataType && libParams.containsKey(INDEX_DESCRIPTION_PARAMETER)) {
```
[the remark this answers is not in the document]
naveentatikonda: yes, that's right

src/main/java/org/opensearch/knn/index/codec/transfer/VectorTransferBinary.java
```diff
@@ -0,0 +1,72 @@
...
+            vectorsPerTransfer = vectorsStreamingMemoryLimit / bytesRef.length;
+            // If vectorsPerTransfer comes out to be 0, then we set number of vectors per transfer to 1, to ensure that
+            // we are sending minimum number of vectors.
+            if (vectorsPerTransfer == 0) {
+                vectorsPerTransfer = 1;
+            }
+        }
+
+        vectorList.add(ArrayUtil.copyOfSubArray(bytesRef.bytes, bytesRef.offset, bytesRef.offset + bytesRef.length));
+        if (vectorList.size() == vectorsPerTransfer) {
```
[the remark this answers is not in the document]
naveentatikonda: As discussed offline all the vector transfer logic will be refactored in this PR - https://github.com/opensearch-project/k-NN/pull/1950

jni/include/jni_util.h
```diff
@@ -73,6 +73,8 @@ namespace knn_jni {
                                                                    int dim, std::vector<float> *vect ) = 0;
         virtual void Convert2dJavaObjectArrayAndStoreToByteVector(JNIEnv *env, jobjectArray array2dJ,
                                                                    int dim, std::vector<uint8_t> *vect ) = 0;
+        virtual void Convert2dJavaObjectArrayAndStoreToSignedByteVector(JNIEnv *env, jobjectArray array2dJ,
```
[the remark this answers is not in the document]
naveentatikonda: refactored everything into bytevector and binaryvector

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
naveentatikonda: refactored it with a switch case

jni/include/commons.h
jmazanec15: I think we need to have a clear ownership understanding for the data. Someone has to decide when this memory gets freed up. Id probably say its best to keep generic stores here, but have the engines handle cleanup.

jni/src/faiss_index_service.cpp
naveentatikonda: I ran some tests comparing a batch size of 1 vs 1000, I could see good difference in the forcemerge time where a batch size of 1000 performs faster by consuming little bit of extra memory

Vikasht34 (approved):
Looks Fine !!

jni/include/commons.h
navneet1v: @jmazanec15 want to understand your comment here. Where are we seeing that ownership of pointers are not correct?

jmazanec15: Im just saying that we need to ensure the interface makes it clear who has to free what. If we allocate it generically, then free it per engine, this could cause errors

shatejas (approved):
Looks good from the integration stand point

src/main/java/org/opensearch/knn/index/codec/transfer/OffHeapBinaryVectorTransfer.java
```diff
@@ -21,12 +23,16 @@ public OffHeapBinaryVectorTransfer(int transferLimit) {
...
 
     @Override
-    protected long transfer(List<byte[]> vectorsToTransfer, boolean append) throws IOException {
-        // TODO: call to JNIService to transfer vector
-        return 0L;
+    protected long transfer(List<byte[]> batch, boolean append) throws IOException {
+        return JNICommons.storeBinaryVectorData(
+            getVectorAddress(),
+            batch.toArray(new byte[][] {}),
+            (long) batch.get(0).length * transferLimit,
```
jmazanec15, line 34: Should we check if batch is empty?

src/main/java/org/opensearch/knn/index/engine/KNNEngine.java
```diff
@@ -34,6 +34,7 @@ public enum KNNEngine implements KNNLibrary {
     private static final Set<KNNEngine> CUSTOM_SEGMENT_FILE_ENGINES = ImmutableSet.of(KNNEngine.NMSLIB, KNNEngine.FAISS);
     private static final Set<KNNEngine> ENGINES_SUPPORTING_FILTERS = ImmutableSet.of(KNNEngine.LUCENE, KNNEngine.FAISS);
     public static final Set<KNNEngine> ENGINES_SUPPORTING_RADIAL_SEARCH = ImmutableSet.of(KNNEngine.LUCENE, KNNEngine.FAISS);
+    public static final Set<KNNEngine> ENGINES_SUPPORTING_BYTE_VECTOR = ImmutableSet.of(KNNEngine.FAISS, KNNEngine.LUCENE);
```
jmazanec15, line 37: Do we need this? I actually recently refactored to make it on the method level: https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/engine/faiss/FaissHNSWMethod.java#L38

src/main/java/org/opensearch/knn/index/mapper/MethodFieldMapper.java
```diff
@@ -119,10 +122,23 @@ private MethodFieldMapper(
             knnMethodConfigContext
         );
         try {
-            this.fieldType.putAttribute(
-                PARAMETERS,
-                XContentFactory.jsonBuilder().map(knnLibraryIndexingContext.getLibraryParameters()).toString()
-            );
+            Map<String, Object> libParams = knnLibraryIndexingContext.getLibraryParameters();
+
+            // If VectorDataType is Byte using Faiss engine then manipulate Index Description to use "SQ8_direct_signed" scalar quantizer
```
jmazanec15, line 127: Can we do this in the FaissHNSWMethod. See here: https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/engine/faiss/AbstractFaissMethod.java#L99-L103

src/main/java/org/opensearch/knn/index/engine/KNNEngine.java
naveentatikonda: removed

jni/include/commons.h
```diff
@@ -47,10 +47,24 @@ namespace knn_jni {
          * CAUTION: The behavior is undefined if the memory address is deallocated and the method is called
          *
          * @param memoryAddress The address of the memory location where data will be stored.
-         * @param data 2D byte array containing data to be stored in native memory.
+         * @param data 2D byte array containing binary data to be stored in native memory.
          * @param initialCapacity The initial capacity of the memory location.
          * @return memory address of std::vector<uint8_t> where the data is stored.
          */
+        jlong storeBinaryVectorData(knn_jni::JNIUtilInterface *, JNIEnv *, jlong , jobjectArray, jlong, jboolean);
```
navneet1v, line 54: what is the use use of jboolean parameter in the end? I am not seeing any java docs related to that.

jni/src/commons.cpp
```diff
@@ -51,6 +51,26 @@ jlong knn_jni::commons::storeByteVectorData(knn_jni::JNIUtilInterface *jniUtil,
...
+    std::vector<int8_t> *vect;
+    if (memoryAddressJ == 0) {
+        vect = new std::vector<int8_t>();
+        vect->reserve(static_cast<long>(initialCapacityJ));
+    } else {
+        vect = reinterpret_cast<std::vector<int8_t>*>(memoryAddressJ);
+    }
+
+    if (appendJ == JNI_FALSE) {
+            vect->clear();
```
navneet1v, line 71: The clear operation doesn't come for free and it runs in linear complexity.

Ref: https://cplusplus.com/reference/vector/vector/clear/ I think this append feature will add latency in the whole streaming operation.

cc: @shatejas

jni/src/faiss_index_service.cpp
```diff
@@ -250,5 +250,120 @@ void BinaryIndexService::writeIndex(
...
+
+    // Set thread count if it is passed in as a parameter. Setting this variable will only impact the current thread
+    if(threadCount != 0) {
+        omp_set_num_threads(threadCount);
+    }
+
+    faiss::IndexIDMap * idMap = reinterpret_cast<faiss::IndexIDMap *> (idMapAddress);
+
+    // Add vectors in batches by casting int8 vectors into float with a batch size of 1000
+    int batchSize = 1000;
```
navneet1v, line 331: since the vector are already coming in batches here why we do we need to batching again?

jni/src/faiss_index_service.cpp
```diff
@@ -250,5 +250,120 @@ void BinaryIndexService::writeIndex(
...
+            floatVectorsIds.resize(batchSize);
+        }
+
+        for (int i = 0; i < batchSize; ++i) {
+            floatVectorsIds[i] = ids[id + i];
+            for (int j = 0; j < dim; ++j, ++iter) {
+                inputFloatVectors[i * dim + j] = static_cast<float>(*iter);
+            }
+        }
+        idMap->add_with_ids(batchSize, inputFloatVectors.data(), floatVectorsIds.data());
```
navneet1v, line 350: You can use float[] and int64_t[] here. and then you would never need to do resize which you are doing in line number 340 and 341. 

As resize has Linear complexity and can trigger a reallocation of memory, which will further spike the memory consumption. Ref: https://cplusplus.com/reference/vector/vector/resize/

Also if you see you don't even need resize, as you updating the vectors at same reference and updating the batchSize attribute. 
In that case even when `inputFloatVectors` is lets say has size 100, but batchSize is 10, then only 10 elements will be picked from inputFloatVectors. You can check in Faiss. Last I checked I remember something like this.

```
idMap->add_with_ids(batchSize, inputFloatVectors.data(), floatVectorsIds.data());
```

src/main/java/org/opensearch/knn/index/engine/faiss/AbstractFaissMethod.java
```diff
@@ -105,6 +106,19 @@ static KNNLibraryIndexingContext adjustPrefix(
         if (knnMethodConfigContext.getVectorDataType() == VectorDataType.BINARY) {
             prefix = FAISS_BINARY_INDEX_DESCRIPTION_PREFIX;
         }
+        if (knnMethodConfigContext.getVectorDataType() == VectorDataType.BYTE) {
+
+            // If VectorDataType is Byte using Faiss engine then manipulate Index Description to use "SQ8_direct_signed" scalar quantizer
+            // For example, Index Description "HNSW16,Flat" will be updated as "HNSW16,SQ8_direct_signed"
+            String indexDescription = methodAsMapBuilder.indexDescription;
+            if (indexDescription != null && indexDescription.isEmpty() == false) {
```
navneet1v, line 114: why not just use StringUtils.isEmpty() from apache commons.

Also what will happen in the case indexDescription is null or empty. I think we should throw an exception rather than just silently not doing anything.

src/main/java/org/opensearch/knn/index/codec/transfer/OffHeapBinaryVectorTransfer.java
```diff
@@ -21,12 +23,19 @@ public OffHeapBinaryVectorTransfer(int transferLimit) {
...
+    protected long transfer(List<byte[]> batch, boolean append) throws IOException {
+        if (!batch.isEmpty()) {
+            return JNICommons.storeBinaryVectorData(
+                getVectorAddress(),
+                batch.toArray(new byte[][] {}),
+                (long) batch.get(0).length * transferLimit,
+                append
+            );
+        }
+        return 0;
```
jmazanec15, line 39: Should this return the vector address? Just thinking in append scenario

src/main/java/org/opensearch/knn/index/engine/faiss/AbstractFaissMethod.java
jmazanec15: Just a heads up - as part of the mode/compression support, Im going to adjust the default logic to provide kinda of a default provider as opposed to just a default. So we wont need this additional logic once thats changed, because we will be able to supply default quantizer based on data type.

jni/src/faiss_index_service.cpp
```diff
@@ -250,5 +250,120 @@ void BinaryIndexService::writeIndex(
...
+    int batchSize = 1000;
+    std::vector <float> inputFloatVectors(batchSize * dim);
+    std::vector <int64_t> floatVectorsIds(batchSize);
+    int id = 0;
+    auto iter = inputVectors->begin();
+
+    for (int id = 0; id < numVectors; id += batchSize) {
+        if (numVectors - id < batchSize) {
+            batchSize = numVectors - id;
+            inputFloatVectors.resize(batchSize * dim);
```
vibrantvarun, line 340: Just for understanding purpose, why this resize is needed?

jni/src/commons.cpp
shatejas: We are trading a memory optimization for latencies here. If the user doesn't want to rely on underlying implementation optimization and be absolutely sure that this doesn't incur latencies, the can always use append == true which is still the default in [JNICommons](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/jni/JNICommons.java#L48).

This was done to have no fragmentation for iterative builds. Alternative is to allow memory fragmentation, considering indexing is not latency sensitive and is memory sensitive, it makes sense to keep it. Thoughts?

I also dug into 
> This may be optimized to constant complexity for [trivially-destructible types](https://cplusplus.com/is_trivially_destructible) (such as scalar or PODs), where elements need not be destroyed.

Here is a [good explanation](https://stackoverflow.com/questions/11235975/what-is-the-complexity-of-stdvectortclear-when-t-is-a-primitive-type#:~:text=Therefore%2C%20setting%20vector%20's%20size,by%20employing%20partial%20template%20specialization) of why it might possible be O(1) for primitive types considering they are trivially destructible. So we might possibly get away with added latencies but cannot say for sure

jni/src/faiss_index_service.cpp
```diff
@@ -329,16 +329,14 @@ void ByteIndexService::insertToIndex(
 
     // Add vectors in batches by casting int8 vectors into float with a batch size of 1000
     int batchSize = 1000;
-    std::vector <float> inputFloatVectors(batchSize * dim);
-    std::vector <int64_t> floatVectorsIds(batchSize);
+    float inputFloatVectors[batchSize * dim];
```
navneet1v, line 332: Try using malloc here to create the array.

jni/include/commons.h
naveentatikonda: updated docs for append

jni/src/faiss_index_service.cpp
naveentatikonda: As discussed offline, if we don't use batching from JNI to faiss it leads to a 1.2GB memory spike in an avg production cluster(using 32gb of heap) when we cast these int8 vectors into float. To avoid that spike we need batching here. I will add these details on a github issue and will add that comment here.

jni/src/faiss_index_service.cpp
naveentatikonda: Tested with arrays, which was running into segmentation fault with benchmarking test. We can avoid resize with std::vector which works based on batchsize and I have tested it
```
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGSEGV (0xb) at pc=0x00007f20a93e33a0, pid=80, tid=295
#
# JRE version: OpenJDK Runtime Environment Temurin-21.0.1+12 (21.0.1+12) (build 21.0.1+12-LTS)
# Java VM: OpenJDK 64-Bit Server VM Temurin-21.0.1+12 (21.0.1+12-LTS, mixed mode, sharing, tiered, compressed oops, compressed class ptrs, g1 gc, linux-amd64)
# Problematic frame:
# C  [libopensearchknn_faiss_avx2.so+0x1de3a0]  knn_jni::faiss_wrapper::ByteIndexService::insertToIndex(int, int, int, long, std::vector<long, std::allocator<long> >&, long)+0x210
#
# Core dump will be written. Default location: Core dumps may be processed with "/usr/lib/systemd/systemd-coredump %P %u %g %s %t %c %h" (or dumping to /home/ci-runner/opensearch/core.80)
#
# An error report file with more information is saved as:
# logs/hs_err_pid80.log
[23.597s][warning][os] Loading hsdis library failed
#
# If you would like to submit a bug report, please visit:
#   https://github.com/adoptium/adoptium-support/issues
# The crash happened outside the Java Virtual Machine in native code.
# See problematic frame for where to report the bug.
#
```

src/main/java/org/opensearch/knn/index/engine/faiss/AbstractFaissMethod.java
naveentatikonda: Updated to use StringUtils for now. But, I'm not throwing an exception for the else case as Jack mentioned that we will be refactoring it.

src/main/java/org/opensearch/knn/index/codec/transfer/OffHeapBinaryVectorTransfer.java
naveentatikonda: Actually, we don't need to check if list is empty or not because we are calling this if list size is equal to transfer limit or during flush (where we are verifying if list is not empty)

jni/include/commons.h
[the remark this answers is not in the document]
naveentatikonda: The first 2 are not params

jni/src/faiss_index_service.cpp
naveentatikonda: resize is used to reallocate the memory based on the new batch size, just in case if the last batch is less than the usual size 1000

jni/src/faiss_index_service.cpp
naveentatikonda: malloc is also working and I have verified with a benchmarking test. But, going with std::vector changes as either of them works without resize. Thanks!

jni/src/faiss_index_service.cpp
navneet1v: Using malloc instead of `arr[batchSize*dim]` may cause issue as malloc will create array on heap but `arr[batchSize*dim]` create on stack. Please use that.

naveentatikonda: Update: Arrays also will work using malloc
https://github.com/opensearch-project/k-NN/pull/1823#discussion_r1728537543

vibrantvarun (approved):

jni/src/commons.cpp
jmazanec15: I think this operation is fine to have - int8_t should be trivially destructable. In future, we should move away from using vector so we can more explicitly control the allocation for the buffer.

jmazanec15 (approved):
LGTM

jni/src/commons.cpp
navneet1v: If we are confident on the destruction that I agree it should not be a big performance concern. Resolving the conversation.

navneet1v (approved):

PR #1825, opened 2024-07-15
Switch from byte stream to byte ref for serde

### LINKED ISSUE #1736: [k-NN] Avoid additional copy to stream during binary doc values deserialization
## Description
Related #1709 

During deserialization, we have an intermediate step where we copy a bytesref to a bytesstream. This can be removed as it  adds some overhead and does add any benefit. With #1087, this will be no longer necessary. But until then, it could add some benefit.

From the experiments in #1709 , the diff with and without the optimization is:

### Results -  w/o change
Run # | p50 latency (ms) | p90 latency (ms) | p99 latency (ms) | Recall
-- | -- | -- | -- | --
1 | 674 | 684 | 692 | 0.99998
2 | 674 | 684 | 692 | 0.99998

### Results -  w/ change
Run # | p50 latency (ms) | p90 latency (ms) | p99 latency (ms) | Recall
-- | -- | -- | -- | --
1 | 568 | 584 | 594 | 0.99998
2 | 568 | 584 | 596 | 0.99998

PoC commit: https://github.com/jmazanec15/k-NN-1/commit/7984618de9369228af388148d023475b0c58c061.

jmazanec15:
### Description
Avoids copy during serialization and deserialization by switching from requiring byte streams to only requiring byte refs. From [KNNCodecUtil](https://github.com/opensearch-project/k-NN/blob/2.15.0.0/src/main/java/org/opensearch/knn/index/codec/util/KNNCodecUtil.java#L57-L62), we can see that we convert byte ref to byte stream just to convert it back in the [KNNVectorSerializer](https://github.com/opensearch-project/k-NN/blob/2.15.0.0/src/main/java/org/opensearch/knn/index/codec/util/KNNVectorAsCollectionOfFloatsSerializer.java#L29-L39) logic - in other words, it is not necessary. This can speed up operations by 15-20% for exact search (see #1736).

This will not have impact on KnnVectorsFormat structures. Those will use lucenes vector serde which is already optimized. This change is meant to add a boost in performance until that is used completely.

New functionality is not added - existing tests provide coverage.

### Issues Resolved
#1736 
 
### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/codec/util/KNNVectorAsArraySerializer.java
+++ b/src/main/java/org/opensearch/knn/index/codec/util/KNNVectorAsArraySerializer.java
@@ -5,6 +5,8 @@
 
 package org.opensearch.knn.index.codec.util;
 
+import org.apache.lucene.util.BytesRef;
+
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
@@ -31,8 +33,8 @@ public class KNNVectorAsArraySerializer implements KNNVectorSerializer {
     }
 
     @Override
-    public float[] byteToFloatArray(ByteArrayInputStream byteStream) {
-        try {
+    public float[] byteToFloatArray(BytesRef bytesRef) {
+        try (ByteArrayInputStream byteStream = new ByteArrayInputStream(bytesRef.bytes, bytesRef.offset, bytesRef.length)) {
             final ObjectInputStream objectStream = new ObjectInputStream(byteStream);
             final float[] vector = (float[]) objectStream.readObject();
             return vector;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 * Adds dynamic query parameter ef_search in radial search faiss engine [#1790](https://github.com/opensearch-project/k-NN/pull/1790)
 * Add binary format support with HNSW method in Faiss Engine [#1781](https://github.com/opensearch-project/k-NN/pull/1781)
 ### Enhancements
+* Switch from byte stream to byte ref for serde [#1825](https://github.com/opensearch-project/k-NN/pull/1825)
 ### Bug Fixes
 * Fixing the arithmetic to find the number of vectors to stream from java to jni layer.[#1804](https://github.com/opensearch-project/k-NN/pull/1804)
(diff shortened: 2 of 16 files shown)


heemin32 (approved):

src/main/java/org/opensearch/knn/index/codec/util/KNNVectorAsArraySerializer.java
```diff
@@ -31,8 +33,8 @@ public byte[] floatToByteArray(float[] input) {
     }
 
     @Override
-    public float[] byteToFloatArray(ByteArrayInputStream byteStream) {
-        try {
+    public float[] byteToFloatArray(BytesRef bytesRef) {
+        try (ByteArrayInputStream byteStream = new ByteArrayInputStream(bytesRef.bytes, bytesRef.offset, bytesRef.length)) {
```
martin-gaievski, line 37: what's the meaning of `bytesRef.offset`, are there any edge case scenarios where behavior is different from ByteArrayInputStream?

martin-gaievski: adding to previous comment - should we add more checks if with offset we get out of the stream's boundaries? I mean in the code itself and also add a unit test that models such scenario.

jmazanec15: Class is defined here: https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/util/BytesRef.java. 

> are there any edge case scenarios where behavior is different from ByteArrayInputStream

I dont think so - ByteArrayInputStream is pretty lightweight.

The constructor does do a validity check on BytesRef, so Im not sure we need to also check here for validity of bytes ref: https://github.com/apache/lucene/blob/5e52b8094aca4dab2898c7dd17084fc7cf7fd016/lucene/core/src/java/org/apache/lucene/util/BytesRef.java#L58. 

That being said, I think there is a mistake for this here: https://github.com/opensearch-project/k-NN/blob/50539182580aff1b4a26b684da39841bb78cfd1b/src/main/java/org/opensearch/knn/index/codec/util/KNNVectorAsCollectionOfFloatsSerializer.java#L31-L34. length already factors in offset so I think this is wrong. It has not showed up yet because offset is always 0 and length=length (i believe). 

Anyway, I will fix and add a unit test.

heemin32 (approved):
LGTM. Thanks for rebasing the code which might have taken an extra effort.

navneet1v (approved):

PR #1840, opened 2024-07-16
Iterative Vector Insertion

MrFlap:
### Description
This change is an memory-based optimization of KNN index construction. Instead of loading all of the vectors into memory, then all into the index, we only load one (much smaller) batch at a time in order to avoid having copies of vector memory.

This also implements memory allocation on index initialization on HNSW to avoid vector resizing.

Support for creating index from template has not been added yet and will be added once this change is made to the main branch: https://github.com/opensearch-project/k-NN/pull/1784

Currently no planned support for nmslib.

### Issues Resolved
https://github.com/opensearch-project/k-NN/issues/1600

### Benchmarking
Here are some current benchmarks and metrics:
#### 1 MB Vector Streaming Limit on SIFT dataset:
- Memory usage: graph_sift_1mb
- Recall @k10: 0.983
- Ingestion time: 908s

#### 10 MB Vector Streaming Limit Memory Usage
- Memory usage: graph_sift_10mb
- Recall @k10: 0.983
- Ingestion time: 908s

#### 100 MB Vector Streaming Limit Memory Usage
- Memory usage: graph_sift_100mb
- Recall @k10: 0.983
- Ingestion time: 909s

 
### Check List
  - [x] All tests pass


navneet1v:
@MrFlap can you attach the benchmarking which you have done on the issue description

MrFlap:
> @MrFlap can you attach the benchmarking which you have done on the issue description

Will add more after OSB.

junqiu-lei:
> Anyone knows why tests haven't been executed for this PR?

Might need to resolve conflicts firstly.

navneet1v:
@heemin32 and @junqiu-lei the PR is more of an early look into the code. Since the tests can change based on the src files, the idea here is first we review the src and then same PR can be updated with tests. GH actions may keep on failing because there has been a lot of changes on going in the similar files.

src/main/java/org/opensearch/knn/jni/JNIService.java
```diff
@@ -23,6 +23,55 @@
...
+        if (KNNEngine.FAISS == knnEngine) {
+            if (IndexUtil.isBinaryIndex(knnEngine, parameters)) {
+                return FaissService.initBinaryIndex(size, dim, parameters);
+            } else {
+                return FaissService.initIndex(size, dim, parameters);
+            }
+        }
+
+        throw new IllegalArgumentException(
+            String.format("initIndexFromScratch not supported for provided engine : %s", knnEngine.getName())
```
navneet1v, line 38: Wherever you are using String.format pass Locale.ROOT as the first argument to ensure that correct locale is set for the exception

navneet1v (commented):
Reviewed the JNI layer and some java files. KNN80DocValuesConsumer class is bit hacky. Lets try to modularize the code bit more to ensure that it is maintainable.

jni/cmake/init-faiss.cmake-e
```diff
@@ -0,0 +1,69 @@
+#
+# Copyright OpenSearch Contributors
+# SPDX-License-Identifier: Apache-2.0
+#
+
+# Check if faiss exists
+find_path(FAISS_REPO_DIR NAMES faiss PATHS ${CMAKE_CURRENT_SOURCE_DIR}/external/faiss NO_DEFAULT_PATH)
```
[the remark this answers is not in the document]
MrFlap: Didn't mean to add. Don't know why it's here. Will be removed.

jni/src/.idea/modules.xml
MrFlap, line 1: Will remove xml files

jni/src/faiss_index_service.cpp
```diff
@@ -57,6 +57,109 @@ void SetExtraParameters(knn_jni::JNIUtilInterface * jniUtil, JNIEnv *env,
...
+    faiss::IndexIDMap * idMap(faissMethods->indexIdMap(indexWriter));
+
+    faiss::IndexHNSW * hnsw = dynamic_cast<faiss::IndexHNSW *>(idMap->index);
+
+    if(hnsw != NULL) {
+        faiss::IndexFlat * storage = dynamic_cast<faiss::IndexFlat *>(hnsw->storage);
+        if(storage != NULL) {
+            storage->codes.reserve(dim * numVectors);
+        }
+    }
```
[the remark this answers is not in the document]
MrFlap: Will add a comment, but this is not creating an index. This is checking if the underlying storage _is_ HNSWFlat, then allocating for it. This technique can be used for other index types.

jni/src/faiss_index_service.cpp
```diff
@@ -57,6 +57,109 @@ void SetExtraParameters(knn_jni::JNIUtilInterface * jniUtil, JNIEnv *env,
...
+        omp_set_num_threads(threadCount);
+    }
+
+    faiss::IndexIDMap * idMap = reinterpret_cast<faiss::IndexIDMap *> (idMapAddress);
+
+    // Write the index to disk
+    faissMethods->writeIndex(idMap, indexPath.c_str());
+
+    // Free the memory used by the index
+    delete idMap;
```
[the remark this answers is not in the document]
MrFlap: Yes. I need to add that back.

jni/src/faiss_wrapper.cpp
```diff
@@ -85,6 +85,220 @@ bool isIndexIVFPQL2(faiss::Index * index);
...
+    if(parametersCpp.find(knn_jni::PARAMETERS) != parametersCpp.end()) {
+        subParametersCpp = jniUtil->ConvertJavaMapToCppMap(env, parametersCpp[knn_jni::PARAMETERS]);
+    }
+    // end parameters to pass
+
+    // Create index
+    return indexService->initIndex(jniUtil, env, metric, indexDescriptionCpp, dim, numDocs, threadCount, subParametersCpp);
+}
+/*
+jlong knn_jni::faiss_wrapper::InitIndexFromTemplate(knn_jni::JNIUtilInterface * jniUtil, JNIEnv * env, jlong numDocs, jint dimJ,
```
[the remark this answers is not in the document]
MrFlap: Will do

jni/src/faiss_wrapper.cpp
```diff
@@ -85,6 +85,220 @@ bool isIndexIVFPQL2(faiss::Index * index);
...
+        subParametersCpp = jniUtil->ConvertJavaMapToCppMap(env, parametersCpp[knn_jni::PARAMETERS]);
+    }
+    // end parameters to pass
+
+    // Create index
+    return indexService->initIndex(jniUtil, env, metric, indexDescriptionCpp, dim, numDocs, threadCount, subParametersCpp);
+}
+*/
+void knn_jni::faiss_wrapper::InsertToIndex(knn_jni::JNIUtilInterface * jniUtil, JNIEnv * env, jintArray idsJ, jlong vectorsAddressJ, jint dimJ,
+                                         jlong index_ptr, jobject parametersJ, IndexService* indexService) {
```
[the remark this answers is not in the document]
MrFlap: Would it not be used at least for the thread count? I could just pass thread count though.

jni/src/faiss_wrapper.cpp
```diff
@@ -85,6 +85,220 @@ bool isIndexIVFPQL2(faiss::Index * index);
...
+    // so that it is easier to access.
+    auto parametersCpp = jniUtil->ConvertJavaMapToCppMap(env, parametersJ);
+
+    // Parameters to pass
+    // Metric type
+    jobject spaceTypeJ = knn_jni::GetJObjectFromMapOrThrow(parametersCpp, knn_jni::SPACE_TYPE);
+    std::string spaceTypeCpp(jniUtil->ConvertJavaObjectToCppString(env, spaceTypeJ));
+    faiss::MetricType metric = TranslateSpaceToMetric(spaceTypeCpp);
+    jniUtil->DeleteLocalRef(env, spaceTypeJ);
+
```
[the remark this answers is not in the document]
MrFlap: No. Thought I deleted that.

src/main/java/org/opensearch/knn/index/codec/transfer/VectorTransferByte.java
```diff
@@ -60,7 +60,11 @@ public SerializationMode getSerializationMode(final BytesRef bytesRef) {
 
     private void transfer() {
         int lengthOfVector = dimension / 8;
-        vectorAddress = JNICommons.storeByteVectorData(vectorAddress, vectorList.toArray(new byte[][] {}), totalLiveDocs * lengthOfVector);
+        if(totalLiveDocs > 0) {
```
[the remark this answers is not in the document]
MrFlap: This is so that we have an option to only allocate enough memory for one transfer. This is as opposed to allocating enough memory for all of the live docs. This seemed like the easiest way to do this.

src/main/java/org/opensearch/knn/index/codec/util/KNNCodecUtil.java
```diff
@@ -38,32 +38,47 @@ public static final class Pair {
...
         SerializationMode serializationMode = SerializationMode.COLLECTION_OF_FLOATS;
-        vectorTransfer.init(getTotalLiveDocsCount(values));
-        for (int doc = values.nextDoc(); doc != DocIdSetIterator.NO_MORE_DOCS; doc = values.nextDoc()) {
+        if(iterative) {
+            vectorTransfer.init(0);
+        } else {
+            vectorTransfer.init(getTotalLiveDocsCount(values));
+        }
+        int doc = values.nextDoc();
+        for (; doc != DocIdSetIterator.NO_MORE_DOCS; doc = values.nextDoc()) {
```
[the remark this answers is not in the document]
MrFlap: Currently I'm using the docID outside of the loop to see if we have finished iterating through the live docs. Could be a nicer way to do this.

src/main/java/org/opensearch/knn/jni/JNIService.java
MrFlap: I need to look into this. This would need to be changed in the already existing functions as this was copied from createIndex.

src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80DocValuesConsumer.java
```diff
@@ -120,19 +120,18 @@ private VectorTransfer getVectorTransfer(FieldInfo field) {
 
     public void addKNNBinaryField(FieldInfo field, DocValuesProducer valuesProducer, boolean isMerge, boolean isRefresh)
         throws IOException {
+        // Don't know why this would be null. Can't do any creation when it is, so best to just return.
+        if (field == null) {
+            logger.info("Field is null!\n");
+            return;
+        }
```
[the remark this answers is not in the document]
MrFlap: It used to be null for unit tests. I haven't seen it recently. Will remove.

jni/include/faiss_index_service.h
```diff
@@ -31,8 +31,50 @@ namespace faiss_wrapper {
...
+     * @param env jni environment
+     * @param metric space type for distance calculation
+     * @param indexDescription index description to be used by faiss index factory
+     * @param dim dimension of vectors
+     * @param numVectors number of vectors
+     * @param threadCount number of thread count to be used while adding data
+     * @param parameters parameters to be applied to faiss index
+     * @return pointer to the native index object
+     */
+    virtual jlong initIndex(knn_jni::JNIUtilInterface *jniUtil, JNIEnv *env, faiss::MetricType metric, std::string indexDescription, int dim, int numVectors, int threadCount, std::unordered_map<std::string, jobject> parameters);
```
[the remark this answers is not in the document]
MrFlap: Ideally yes. Should I remove it or deprecate it?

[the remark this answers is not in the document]
navneet1v: I think we should remove it, if it is not getting used. and if it is still getting used then marking it as deprecated will be right thing to do.

jni/src/faiss_index_service.cpp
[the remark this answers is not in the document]
navneet1v: > Will add a comment, but this is not creating an index. 

Oh yes it is not creating the FlatIndex. What I mean to say is lets add a comment which mention why we are getting the codes and then reserving the space.

jni/src/faiss_index_service.cpp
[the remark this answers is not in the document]
navneet1v: Please then make the change,

jni/src/faiss_wrapper.cpp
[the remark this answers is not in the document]
navneet1v: You have 2 options here:
1. Pass the whole parametersJ and then only parse the threadCount out of it.
2. Pass only threadCount. 

I think 1 is better here since we need threadCount.

jni/src/faiss_wrapper.cpp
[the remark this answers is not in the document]
navneet1v: Seems like its not. `InsertToIndex` should just insert the data into FaissIndex. Let simplify the code.

src/main/java/org/opensearch/knn/index/codec/util/KNNCodecUtil.java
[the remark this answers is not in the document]
navneet1v: Sure. Looking at the code I feel we can make the logic to stream vectors and build graphs iteratively more cleaner. Please look into how we can clean up the whole process of streaming the vectors and also create the graphs iteratively.

src/main/java/org/opensearch/knn/jni/JNIService.java
navneet1v: Sure.

jni/src/faiss_wrapper.cpp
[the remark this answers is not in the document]
MrFlap: Okay. I was thinking more of doing option 2 and parsing threadCount in the java part of the JNI layer, though. This was one of the optimizations I was planning on doing later since every time we insert to the index, we have to recreate the entire Java map in c++.

jni/include/faiss_index_service.h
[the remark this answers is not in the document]
MrFlap: I just tried this. There are a few integration tests that directly call this function. I don't think it would be too hard to change these to use the iterative method. Would this be best?

jni/include/faiss_index_service.h
```diff
@@ -31,8 +31,50 @@ namespace faiss_wrapper {
...
+     *
+     * @param jniUtil jni util
+     * @param env jni environment
+     * @param metric space type for distance calculation
+     * @param indexDescription index description to be used by faiss index factory
+     * @param dim dimension of vectors
+     * @param numVectors number of vectors
+     * @param threadCount number of thread count to be used while adding data
+     * @param parameters parameters to be applied to faiss index
+     * @return pointer to the native index object
```
Vikasht34, line 45: nit:- memory address to native Index.

jni/include/faiss_index_service.h
```diff
@@ -31,8 +31,50 @@ namespace faiss_wrapper {
 class IndexService {
 public:
     IndexService(std::unique_ptr<FaissMethods> faissMethods);
-    //TODO Remove dependency on JNIUtilInterface and JNIEnv
-    //TODO Reduce the number of parameters
+    /**
```
Vikasht34, line 34: Right now functions are taking 8 parameters , Ideally we should not increase more than 2. This is not scalable , 

Shall we wrap it like this , It makes api more cleaner.
```

struct IndexConfiguration {
    knn_jni::JNIUtilInterface* jniUtil;
    JNIEnv* env;
    faiss::MetricType metric;
    std::string indexDescription;
    int dim;
    int numVectors;
    int threadCount;
    std::unordered_map<std::string, std::string> parameters;
};

class IndexService {
public:
    virtual jlong initIndex(const IndexConfiguration& config);
};
```

src/main/java/org/opensearch/knn/index/codec/transfer/VectorTransferByte.java
```diff
@@ -34,11 +34,12 @@ public void init(final long totalLiveDocs) {
     public void transfer(final BytesRef bytesRef) {
         dimension = bytesRef.length * 8;
         if (vectorsPerTransfer == Integer.MIN_VALUE) {
-            // if vectorsStreamingMemoryLimit is 100 bytes and we have 50 vectors with length of 5, then per
-            // transfer we have to send 100/5 => 20 vectors.
             vectorsPerTransfer = vectorsStreamingMemoryLimit / bytesRef.length;
-            // If vectorsPerTransfer comes out to be 0, then we set number of vectors per transfer to 1, to ensure that
-            // we are sending minimum number of vectors.
+            if (totalLiveDocs > 0) {
```
jmazanec15, line 38: Why change this in this PR?

jni/include/faiss_wrapper.h
```diff
@@ -18,10 +18,11 @@
 
 namespace knn_jni {
     namespace faiss_wrapper {
-        // Create an index with ids and vectors. The configuration is defined by values in the Java map, parametersJ.
-        // The index is serialized to indexPathJ.
-        void CreateIndex(knn_jni::JNIUtilInterface * jniUtil, JNIEnv * env, jintArray idsJ, jlong vectorsAddressJ, jint dimJ,
-                         jstring indexPathJ, jobject parametersJ, IndexService* indexService);
+        jlong InitIndex(knn_jni::JNIUtilInterface *jniUtil, JNIEnv *env, jlong numDocs, jint dimJ, jobject parametersJ, IndexService *indexService);
```
Vikasht34, line 21: Let's wrap function parameter in struct , More than 2 parameters  for a function  will be making hard time to scale.

src/main/java/org/opensearch/knn/index/codec/util/KNNCodecUtil.java
```diff
@@ -38,32 +38,52 @@ public static final class Pair {
...
             BytesRef bytesref = values.binaryValue();
             serializationMode = vectorTransfer.getSerializationMode(bytesref);
             vectorTransfer.transfer(bytesref);
             docIdList.add(doc);
+            // Semi-hacky way to check if the streaming limit has been reached
+            if (iterative && vectorTransfer.getVectorAddress() != 0) {
```
jmazanec15, line 73: Are we transferring one vector at a time if its iterative? Can we do in batches? Also, instead of the iterative arg, can we just have how many vectors should be fetched? That way, this function does not need to be concerned with what the vectors are being used for.

src/main/java/org/opensearch/knn/jni/FaissService.java
```diff
@@ -50,32 +50,64 @@ class FaissService {
...
-     * Create an index for the native library The memory occupied by the vectorsAddress will be freed up during the
-     * function call. So Java layer doesn't need to free up the memory. This is not an ideal behavior because Java layer
-     * created the memory address and that should only free up the memory. We are tracking the proper fix for this on this
-     * <a href="https://github.com/opensearch-project/k-NN/issues/1600">issue</a>
+     * Initialize an index for the native library. Takes in numDocs to
+     * allocate the correct amount of memory.
      *
-     * @param ids array of ids mapping to the data passed in
-     * @param vectorsAddress address of native memory where vectors are stored
+     * @param numDocs number of documents to be added
```
jmazanec15, line 56: Does this need to be fixed? What if a user doesnt know up front?

src/main/java/org/opensearch/knn/jni/FaissService.java
```diff
@@ -50,32 +50,64 @@ class FaissService {
...
+     * @param dim dimension of the vector to be indexed
+     * @param parameters parameters to build index
+     */
+    public static native long initBinaryIndex(long numDocs, int dim, Map<String, Object> parameters);
+
+    /**
+     * Inserts to a faiss index.
+     *
+     * @param ids ids of documents
+     * @param vectorsAddress address of native memory where vectors are stored
```
jmazanec15, line 76: The number of ids will always match the number of vectors?

src/main/java/org/opensearch/knn/jni/FaissService.java
```diff
@@ -50,32 +50,64 @@ class FaissService {
...
+
+    /**
+     * Writes a faiss index.
+     *
+     * @param indexAddress address of native memory where index is stored
+     * @param indexPath path to save index file to
+     * @param threadCount number of threads to use for insertion
      */
-    public static native void createBinaryIndex(int[] ids, long vectorsAddress, int dim, String indexPath, Map<String, Object> parameters);
+    public static native void writeBinaryIndex(long indexAddress, String indexPath, int threadCount);
```
jmazanec15, line 110: Do we need thread count for writing?

src/main/java/org/opensearch/knn/jni/FaissService.java
```diff
@@ -50,32 +50,64 @@ class FaissService {
...
+    /**
+     * Inserts to a faiss index.
+     *
+     * @param ids ids of documents
      * @param vectorsAddress address of native memory where vectors are stored
      * @param dim dimension of the vector to be indexed
+     * @param indexAddress address of native memory where index is stored
+     * @param threadCount number of threads to use for insertion
+     */
+    public static native void insertToBinaryIndex(int[] ids, long vectorsAddress, int dim, long indexAddress, int threadCount);
```
jmazanec15, line 92: This is great that we are separating insert and write! We might be able to do more creative things in the future with this.

src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80DocValuesConsumer.java
```diff
@@ -127,142 +135,254 @@ public void addKNNBinaryField(FieldInfo field, DocValuesProducer valuesProducer,
...
     }
 
     private void recordRefreshStats() {
         KNNGraphValue.REFRESH_TOTAL_OPERATIONS.increment();
     }
 
-    private void createKNNIndexFromTemplate(Model model, KNNCodecUtil.Pair pair, KNNEngine knnEngine, String indexPath) {
+    private Map<String, Object> genParameters(boolean fromScratch, FieldInfo fieldInfo, KNNEngine knnEngine) throws IOException {
         Map<String, Object> parameters = new HashMap<>();
+        ;
```
jmazanec15, line 170: nit: remove?

src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80DocValuesConsumer.java
```diff
@@ -127,142 +135,254 @@ public void addKNNBinaryField(FieldInfo field, DocValuesProducer valuesProducer,
...
+        if (fromScratch) {
```
jmazanec15, line 171: Can we just have branch between fromScratch and not during index initialization? Then, everything else should proceed the same.

jni/src/faiss_index_service.cpp
```diff
@@ -57,6 +57,104 @@ void SetExtraParameters(knn_jni::JNIUtilInterface * jniUtil, JNIEnv *env,
...
+        JNIEnv * env,
+        faiss::MetricType metric,
+        std::string indexDescription,
+        int dim,
+        int numVectors,
+        int threadCount,
+        std::unordered_map<std::string, jobject> parameters
+    ) {
+    // Removed the check for number of vectors
+    // Don't use unique_ptr here since we want to access the index in the future.
```
Vikasht34, line 71: When using std::unique_ptr, the object it manages is automatically destroyed when the unique_ptr goes out of scope or is reset. This automatic management ensures that resources are freed, preventing memory leaks.

It has lots of other benefits over naked pointer.

However to address the concern of getting aceess later , we can transfer the ownership . Let me give you refactored code on how to use unique pointer to get automatic memory free and don't loose transfer.

jni/src/faiss_index_service.cpp
```diff
@@ -57,6 +57,104 @@ void SetExtraParameters(knn_jni::JNIUtilInterface * jniUtil, JNIEnv *env,
 
 IndexService::IndexService(std::unique_ptr<FaissMethods> faissMethods) : faissMethods(std::move(faissMethods)) {}
 
+jlong IndexService::initIndex(
```
Vikasht34, line 60: Below are the things we can improve in this function 

1. What happens when there is error from this line , we should gracefully handle the error and let caller know what error happened , this is just an example , we should wrap the complete function in try catch
```
faiss::Index * indexWriter(faissMethods->indexFactory(dim, indexDescription.c_str(), metric));
```
2. I see we are avoiding smart pointers and using naked pointers,  and the concern is about accesbilty , but we can achiever using smart pointer too , this is called transfering the owenership.  This is really good artricle to follow https://bartoszmilewski.com/2009/05/21/unique_ptr-how-unique-is-it/



I somehow refactored this code to make it more error handling with smart pointers , let me know if this makes sense.

```
jlong IndexService::initIndex(
    knn_jni::JNIUtilInterface *jniUtil,
    JNIEnv *env,
    faiss::MetricType metric,
    std::string indexDescription,
    int dim,
    int numVectors,
    int threadCount,
    std::unordered_map<std::string, jobject> parameters
) {
    try {
        // Create index using Faiss factory method
        std::unique_ptr<faiss::Index> index(faissMethods->indexFactory(dim, indexDescription.c_str(), metric));
        if (!index) {
            throw std::runtime_error("Failed to create index");
        }

        // Set thread count for OpenMP if specified
        if (threadCount > 0) {
            omp_set_num_threads(threadCount);
        }

        // Set extra parameters for specific index types
        SetExtraParameters<faiss::Index, faiss::IndexIVF, faiss::IndexHNSW>(jniUtil, env, parameters, index.get());

        // Check if the index needs training
        if (!index->is_trained) {
            throw std::runtime_error("Index is not trained");
        }

        std::unique_ptr<faiss::IndexIDMap> idMap(new faiss::IndexIDMap(index.release()));

        // Check if the index is of type IndexHNSW and set storage reservation if needed
        if (faiss::IndexHNSW *hnsw = dynamic_cast<faiss::IndexHNSW *>(idMap->index)) {
            if (faiss::IndexFlat *storage = dynamic_cast<faiss::IndexFlat *>(hnsw->storage)) {
                storage->codes.reserve(dim * numVectors * 4); // Reserving space to avoid reallocations
            }
        }

     
        return reinterpret_cast<jlong>(idMap.release());
    } catch (const std::exception &e) {
        // On exception , let's release all the rescources.
    }
}

```

jni/src/faiss_index_service.cpp
```diff
@@ -57,6 +57,104 @@ void SetExtraParameters(knn_jni::JNIUtilInterface * jniUtil, JNIEnv *env,
...
+            // Allocate enough memory for all of the vectors we plan on inserting
+            // We do this to avoid unnecessary memory allocations during insert
+            storage->codes.reserve(dim * numVectors * 4);
+        }
+    }
+
+    return (jlong)idMap;
+}
+
+void IndexService::insertToIndex(
```
Vikasht34, line 106: Same comment as above on making more error handling friendly , releasing memory when exception happned.

src/test/java/org/opensearch/knn/jni/JNIServiceTests.java
```diff
@@ -66,6 +66,17 @@ public class JNIServiceTests extends KNNTestCase {
     private String faissMethod = "HNSW32,Flat";
     private String faissBinaryMethod = "BHNSW32";
 
+    void createIndex(int[] ids, long address, int dimension, String name, Map<String, Object> parameters, KNNEngine engine) {
```
Vikasht34, line 69: This is good , but let's move this function to end of the all test classes.

src/test/java/org/opensearch/knn/index/memory/NativeMemoryLoadStrategyTests.java
```diff
@@ -40,6 +42,17 @@
 
 public class NativeMemoryLoadStrategyTests extends KNNTestCase {
 
+    void createIndex(int[] ids, long address, int dimension, String name, Map<String, Object> parameters, KNNEngine engine) {
```
Vikasht34, line 45: Let's have public functions up and move private functions down the file.

src/main/java/org/opensearch/knn/index/codec/transfer/VectorTransferByte.java
MrFlap: The logic for calculating vectorsPerTransfer was incorrect when I first made this change. One of the new test cases also was not passing unless I limited vectorsPerTransfer to the number of documents, but this might be due to something else. Currently looking into this.

src/main/java/org/opensearch/knn/jni/JNIService.java
```diff
@@ -13,16 +13,99 @@
...
+     * Initialize an index for the native library. Takes in numDocs to
+     * allocate the correct amount of memory.
+     *
+     * @param numDocs number of documents to be added
+     * @param dim dimension of the vector to be indexed
+     * @param parameters parameters to build index
+     * @param knnEngine knn engine
+     * @return address of the index in memory
+     */
+    public static long initIndexFromScratch(long numDocs, int dim, Map<String, Object> parameters, KNNEngine knnEngine) {
```
Vikasht34, line 38: Since we are creating this functions from Scratch , Let's wrap these parameters in class.

src/main/java/org/opensearch/knn/index/codec/util/KNNCodecUtil.java
MrFlap: If the iterative flag is set to true, this means to return after we have reached vectorStreamingLimit worth of vectors. Maybe it would be best to rename the flag to reduce confusion? We could also theoretically change this to how big the batch size is, but I'm not sure that would be better. This flag accounts for the two cases of:
1. Push all of the vectors into memory
2. Push one batch with specified size into memory.

src/main/java/org/opensearch/knn/jni/JNIService.java
```diff
@@ -13,16 +13,99 @@
...
+        if (KNNEngine.FAISS == knnEngine) {
```
Vikasht34, line 39: Jni Service is gateway for all native services like NMSLIB, Faiss , Let's give responsibility for deciding which index to intialize to respective service , in this case let's faiss service decide , I need to call binary vs normal.

I would even prefer , giving this responsibility to actually the ones who is taking action , Let's create faiss factory at cpp layer which decides what method to call. with this all the Api service layer is clean and there is no branching in service layer.

else for every function we keep checking if else and if someting else comes up , we will end up having one more else.and this changes will be at multiple places like we are having here and below functions as well.

src/main/java/org/opensearch/knn/jni/FaissService.java
MrFlap: If the user doesn't know how many documents will be in the index, they could set numDocs to zero. All numDocs is used for is to make sure there is enough capacity to hold all of the data we will insert. I will add a comment here to make this clear 👍

src/main/java/org/opensearch/knn/jni/FaissService.java
MrFlap: I don't think that read_index or write_index in faiss use OpenMP, but I could be wrong. If they don't, then threadCount isn't necessary.

src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80DocValuesConsumer.java
MrFlap: I don't know if I'm following. You mean have two different functions, one for fromScratch, one for fromTemplate?

jni/include/faiss_index_service.h
MrFlap: This is a nice way to do it. It would require some work, though, since we need a solution that doesn't require remaking `parameters` on each InsertToIndex. In order to do this, then, we would probably need InitIndex to return the address to the IndexConfiguration, then use the address as the input for subsequent functions. I will look into this.

Vikasht34: Sure . let me know if you need any help , we can write together

src/main/java/org/opensearch/knn/jni/FaissService.java
MrFlap: Ideally yes, but the function won't break if a different number for numDocs would be used.

jni/src/faiss_index_service.cpp
MrFlap: This is nice. I didn't know about unique_ptr::release. As far as resources go, I don't think there are any that need to be freed, but I could look into it.

Vikasht34: this what I meant index.release() , and here also idMap.release()

MrFlap: Currently exceptions are caught at the top of the JNI layer here: https://github.com/MrFlap/k-NN/blob/95aec46d702f842a36ed599adf7c37cae4ca00b7/jni/src/org_opensearch_knn_jni_FaissService.cpp#L50. Should I move this down?

Vikasht34: U can keep that there also , and from here u can wrap in try catch , throw the error and the benefit u get that in catch , U can release all the objects are you working on and that releases memory

Vikasht34: NVM u don't have to worry about releasing , that's the beauty of unique pointer , once it's out of scope , it's get freed.

heemin32:
Could you explain how to interpret the graph in benchmark? Like what does time mean and why there is a spike at the end? Also, it would be nice to have a memory usage graph of previous implementation(current state).

In addition to memory usage, I think indexing time comparison between new and old approach is also important. Can we have that data as well?

jni/src/faiss_index_service.cpp
```diff
@@ -108,6 +205,103 @@ void IndexService::createIndex(
...
+        throw std::runtime_error("Index is not trained");
+    }
+
+    // Add vectors
+    std::unique_ptr<faiss::IndexBinaryIDMap> idMap(faissMethods->indexBinaryIdMap(indexWriter.release()));
+
+    // Check to see if the current index is BinaryHNSW
+    faiss::IndexBinaryHNSW * hnsw = dynamic_cast<faiss::IndexBinaryHNSW *>(idMap->index);
+
+    if(hnsw != NULL) {
```
heemin32, line 240: Shouldn't we throw an exception if it is not binaryHNSW?

jni/src/faiss_index_service.cpp
```diff
@@ -108,6 +205,103 @@ void IndexService::createIndex(
...
+        // Check to see if the HNSW storage is IndexBinaryFlat
+        faiss::IndexBinaryFlat * storage = dynamic_cast<faiss::IndexBinaryFlat *>(hnsw->storage);
+        if(storage != NULL) {
```
heemin32, line 243: Same here. Do we expect the code to work fine even if storage is null? Otherwise we should throw an exception.

MrFlap:
> Could you explain how to interpret the graph in benchmark? Like what does time mean and why there is a spike at the end? Also, it would be nice to have a memory usage graph of previous implementation(current state).

Will add to the issue, I will also redo the benchmarks once the PR is completed

> In addition to memory usage, I think indexing time comparison between new and old approach is also important. Can we have that data as well?

Yes. Will add. I am almost certain that indexing time is unaffected even for small batch sizes.

jni/src/faiss_index_service.cpp
```diff
@@ -57,6 +57,103 @@ void SetExtraParameters(knn_jni::JNIUtilInterface * jniUtil, JNIEnv *env,
...
+    // Add extra parameters that cant be configured with the index factory
+    SetExtraParameters<faiss::Index, faiss::IndexIVF, faiss::IndexHNSW>(jniUtil, env, parameters, indexWriter.get());
+
+    // Check that the index does not need to be trained
+    if(!indexWriter->is_trained) {
+        throw std::runtime_error("Index is not trained");
+    }
+
+    // Add vectors
+    std::unique_ptr<faiss::IndexIDMap> idMap (faissMethods->indexIdMap(indexWriter.release()));
```
heemin32, line 87: Nit. If we are going to keep the memory, I think smart pointer only makes it complex. We can simply use plain pointer. 

Here, it would be better to set `idMap->own_fields = true;` ihere nstead of we call `delete idMap-index` later.

heemin32, line 87: Nit. If we are going to keep the memory, I think smart pointer only makes it complex. We can simply use plain pointer. 

Here, it would be better to set `idMap->own_fields = true;` here nstead of we call `delete idMap-index` later.

jni/src/faiss_index_service.cpp
```diff
@@ -57,6 +57,103 @@ void SetExtraParameters(knn_jni::JNIUtilInterface * jniUtil, JNIEnv *env,
...
+        omp_set_num_threads(threadCount);
+    }
+
+    faiss::IndexIDMap * idMap = reinterpret_cast<faiss::IndexIDMap *> (idMapAddress);
+
+    // Write the index to disk
+    faissMethods->writeIndex(idMap, indexPath.c_str());
+
+    // Free the memory used by the index
+    delete idMap;
```
heemin32, line 153: One concern here is the case where we fail to release this index memory due to some exception during indexing process. 

I think this delete part should have its own method and we need to put the entire indexing process under try block in java side and call this delete method in finally block.

jni/src/faiss_index_service.cpp
MrFlap: I talked with @navneet1v about this, but own_fields doesn't work as intended. Don't know if this is a bug in faiss or a bug in how we initialize an index.

MrFlap: The smart pointer is a good idea, though, since we don't have to worry about releasing memory for all possible exceptions.

jni/src/faiss_index_service.cpp
MrFlap: This is a good catch. I think that doing the same smart pointer trick should work here though.

jni/src/faiss_index_service.cpp
MrFlap: No. The underlying index could have many different types, for example IndexBinaryFlat. I think it will be out of scope for my internship, but all other index types need to be accounted for here.

jni/src/faiss_index_service.cpp
MrFlap: Yes. The underlying storage could be something other than IndexBinaryFlat, such as in the case for composite indices. Again, a deep dive would be required to properly allocate memory for all types of indices. This is just an implementation for one case.

jni/src/faiss_index_service.cpp
Vikasht34: Yes , Smart Pointer releases memory automatically  when it goes out of scope, we don't have to write delete idmap, it's automatically taken care.
https://fastercapital.com/content/Smart-Pointers-in-C----Managing-Memory-Safely.html#:~:text=Instead%20of%20returning%20a%20raw,overall%20safety%20of%20your%20code.

jni/src/faiss_index_service.cpp
Vikasht34: Yes , Smart Pointer releases memory automatically when it goes out of scope, we don't have to write delete idmap, it's automatically taken care.
https://fastercapital.com/content/Smart-Pointers-in-C----Managing-Memory-Safely.html#:~:text=Instead%20of%20returning%20a%20raw,overall%20safety%20of%20your%20code.

jni/src/faiss_index_service.cpp
MrFlap: There is an issue though with needing to release the memory of the underlying index as well in the case of an exception. Going to think about the best way to do that.

heemin32: > This is a good catch. I think that doing the same smart pointer trick should work here though.
Smart pointer is one thing to handle exception inside this method itself. How about the case where we are missing to call this method from java side? Are we making sure that this method is called once the init index is called?

jni/include/faiss_index_service.h
```diff
@@ -31,8 +31,50 @@ namespace faiss_wrapper {
...
+     * @param jniUtil jni util
+     * @param env jni environment
+     * @param metric space type for distance calculation
+     * @param indexDescription index description to be used by faiss index factory
+     * @param threadCount number of thread count to be used while adding data
+     * @param indexPath path to write index
+     * @param idMap a map of document id and vector id
+     * @param parameters parameters to be applied to faiss index
+     */
+    virtual void writeIndex(int threadCount, std::string indexPath, jlong idMapAddress);
```
heemin32, line 75: Could you add unit test for all these new cpp methods?

jni/include/org_opensearch_knn_jni_FaissService.h
```diff
@@ -18,23 +18,54 @@
...
  */
-JNIEXPORT void JNICALL Java_org_opensearch_knn_jni_FaissService_createIndex
-        (JNIEnv *, jclass, jintArray, jlong, jint, jstring, jobject);
-
+JNIEXPORT jlong JNICALL Java_org_opensearch_knn_jni_FaissService_initIndex(JNIEnv * env, jclass cls,
+                                                                            jlong numDocs, jint dimJ,
+                                                                            jobject parametersJ);
 /*
  * Class:     org_opensearch_knn_jni_FaissService
- * Method:    createBinaryIndex
```
heemin32, line 32: Did we remove this method? I think we are still using it? 
```
JNIService.createIndex(batch.docs, batch.getVectorAddress(), batch.getDimension(), indexPath, parameters, knnEngine);
```

jni/include/faiss_index_service.h
MrFlap: I was going to just take the previous unit tests that use createIndex and rewrite them to use the new functions. I don't know what I would test for if I called these functions individually, as they don't make sense without calling the others.

jni/include/org_opensearch_knn_jni_FaissService.h
MrFlap: To be removed.

heemin32: I think you already removed createIndex method but I think we are still using it??
```
    private void createKNNIndexFromScratch() throws IOException {
        KNNCodecUtil.VectorBatch batch = KNNCodecUtil.getVectorBatch(values, vectorTransfer, false);
        AccessController.doPrivileged((PrivilegedAction<Void>) () -> {
            JNIService.createIndex(batch.docs, batch.getVectorAddress(), batch.getDimension(), indexPath, parameters, knnEngine);
            return null;
        });
    }
```

jni/include/faiss_index_service.h
heemin32: The index service take faiss method as parameter during creation. You can mock faiss method and make sure that correct method is being called with expected parameter.

jni/include/faiss_index_service.h
MrFlap: This is to be resolved in a later PR when the JNI layer fixes are implemented 👍

jni/include/faiss_wrapper.h
MrFlap: This is to be resolved in a later PR when the JNI layer fixes are implemented 👍

jni/include/org_opensearch_knn_jni_FaissService.h
MrFlap: This is for nmslib.

MrFlap: I need to keep compatibility with both libraries 😅

jni/include/faiss_index_service.h
MrFlap: I'll look into how to do this.

jni/src/faiss_index_service.cpp
MrFlap: Ah, this could be an issue. I'll wrap the insertion method inside of a try catch block that will free the index.

MrFlap: Pushed

shatejas (commented):
Seems to be significant change, does this need to be behind a feature flag?

jni/src/faiss_index_service.cpp
```diff
@@ -104,10 +206,113 @@ void IndexService::createIndex(
...
+        int threadCount,
+        std::unordered_map<std::string, jobject> parameters
+    ) {
+    // Create index using Faiss factory method
+    std::unique_ptr<faiss::IndexBinary> indexWriter(faissMethods->indexBinaryFactory(dim, indexDescription.c_str()));
+
+    // Set thread count if it is passed in as a parameter. Setting this variable will only impact the current thread
+    if(threadCount != 0) {
+        omp_set_num_threads(threadCount);
+    }
```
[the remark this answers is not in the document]
MrFlap: This was originally a remnant of copy paste, and I am not sure if any faiss indices use OpenMP to initialize. I could be wrong, though, and have kept it until now in case there is an index that uses OpenMP to initialize.

.idea/copyright/SPDX_ALv2.xml
```diff
@@ -1,6 +0,0 @@
-<component name="CopyrightManager">
```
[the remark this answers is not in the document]
MrFlap: These needed to be added back. I thought that I accidentally committed these.

jni/src/faiss_index_service.cpp
MrFlap: I don't think there should be a warning since it would be perfectly valid to make an IndexBinaryFlat instead of an IndexHNSW.

jni/src/faiss_index_service.cpp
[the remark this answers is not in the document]
MrFlap: Resolving this for now since it doesn't affect program flow and could be useful

jni/include/faiss_index_service.h
MrFlap: I changed the old unit tests to mock these methods and make sure they were called correctly with the right amount of parameters. Could maybe use some work, though. Resolving for now.

jni/src/faiss_index_service.cpp
MrFlap: I am now handling exceptions and freeing memory in the java layer. Resolving for now.

jni/include/faiss_index_service.h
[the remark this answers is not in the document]
MrFlap: Removed createIndex. Resolving.

jni/src/faiss_index_service.cpp
MrFlap: Resolving for now

jni/src/faiss_index_service.cpp
MrFlap: Resolved since this is the best fix for freeing index storage

src/main/java/org/opensearch/knn/jni/JNIService.java
MrFlap: Resolved as this will be a change when JNI refactoring is done.

src/main/java/org/opensearch/knn/jni/JNIService.java
MrFlap: Resolved as this will be a change when JNI refactoring is done.

src/test/java/org/opensearch/knn/jni/JNIServiceTests.java
MrFlap: Done

src/main/java/org/opensearch/knn/index/codec/transfer/VectorTransferByte.java
MrFlap: Changed back

src/main/java/org/opensearch/knn/index/codec/util/KNNCodecUtil.java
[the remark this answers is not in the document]
MrFlap: Fixed the for loop.

src/main/java/org/opensearch/knn/index/codec/transfer/VectorTransferByte.java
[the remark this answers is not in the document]
MrFlap: Resolved as this is still necessary.

MrFlap:
> Seems to be significant change, does this need to be behind a feature flag?

I don't know. This change isn't really adding new functionality for the user, so I'd think it would be only an enhancement.

src/main/java/org/opensearch/knn/index/codec/native_index/NativeIndexBuilder.java
```diff
@@ -0,0 +1,177 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.knn.index.codec.native_index;
```
navneet1v, line 6:
```suggestion
package org.opensearch.knn.index.codec.nativeindex;
```

src/main/java/org/opensearch/knn/index/codec/native_index/NativeIndexBuilder.java
```diff
@@ -0,0 +1,177 @@
...
+import org.opensearch.knn.plugin.stats.KNNGraphValue;
+
+import lombok.NonNull;
+
+import static org.opensearch.knn.common.KNNConstants.MODEL_ID;
+
+/**
+ * Abstract class to build the KNN index
+ */
+public abstract class NativeIndexBuilder {
```
navneet1v, line 35: should we think to rename this class to NativeIndexWriter

src/main/java/org/opensearch/knn/index/codec/native_index/NativeIndexBuilder.java
```diff
@@ -0,0 +1,177 @@
...
+
+    /**
+     * Class that holds info about vectors
+     */
+    protected class NativeVectorInfo {
```
navneet1v, line 40:
```suggestion
    protected class VectorInfo {
```

src/main/java/org/opensearch/knn/index/codec/native_index/NativeIndexBuilder.java
```diff
@@ -0,0 +1,177 @@
...
+        protected VectorDataType vectorDataType;
+        protected int dimension;
+        protected SerializationMode serializationMode;
+    }
+
+    /**
+     * Class that holds info about the native index
+     */
+    protected class NativeIndexInfo {
```
navneet1v, line 49:
```suggestion
    protected class IndexInfo {
```

src/main/java/org/opensearch/knn/index/codec/native_index/NativeIndexBuilder.java
```diff
@@ -0,0 +1,177 @@
...
+        protected FieldInfo fieldInfo;
+        protected KNNEngine knnEngine;
+        protected int numDocs;
+        protected long arraySize;
+        protected Map<String, Object> parameters;
+        protected NativeVectorInfo vectorInfo;
+        protected String indexPath;
+    }
+
+    protected final Logger logger = LogManager.getLogger(NativeIndexBuilder.class);
```
navneet1v, line 59: use @Log4j2

src/main/java/org/opensearch/knn/index/codec/native_index/NativeIndexBuilder.java
```diff
@@ -0,0 +1,177 @@
...
+
+    private long getVectorSize(NativeVectorInfo vectorInfo) {
+        if (vectorInfo.vectorDataType == VectorDataType.BINARY) {
+            return vectorInfo.dimension / 8;
+        } else {
+            return vectorInfo.dimension * 4;
+        }
+    }
+
+    protected KNNEngine getKNNEngine(@NonNull FieldInfo field) {
```
navneet1v, line 149: remove @NonNull from private functions.

src/main/java/org/opensearch/knn/index/codec/native_index/NativeIndexBuilder.java
```diff
@@ -0,0 +1,177 @@
...
+        return indexInfo;
+    }
+
+    private long getVectorSize(NativeVectorInfo vectorInfo) {
+        if (vectorInfo.vectorDataType == VectorDataType.BINARY) {
+            return vectorInfo.dimension / 8;
+        } else {
+            return vectorInfo.dimension * 4;
+        }
+    }
```
navneet1v, line 147: interesting I think you want the bytes for a single vector here. I would rename this to getBytesPerVector

src/main/java/org/opensearch/knn/index/codec/native_index/NativeIndexBuilder.java
```diff
@@ -0,0 +1,177 @@
...
+     * Class that holds info about the native index
+     */
+    protected class NativeIndexInfo {
+        protected FieldInfo fieldInfo;
+        protected KNNEngine knnEngine;
+        protected int numDocs;
+        protected long arraySize;
+        protected Map<String, Object> parameters;
+        protected NativeVectorInfo vectorInfo;
+        protected String indexPath;
```
navneet1v, line 56: make all attributes private here

src/main/java/org/opensearch/knn/index/codec/native_index/NativeIndexBuilder.java
navneet1v, line 49: use @Builder and @Value annotation of lombok here.

src/main/java/org/opensearch/knn/index/codec/native_index/NativeIndexBuilderScratch.java
```diff
@@ -0,0 +1,117 @@
...
+import org.opensearch.knn.index.VectorDataType;
+import org.opensearch.knn.index.codec.transfer.VectorTransfer;
+import org.opensearch.knn.index.codec.util.KNNCodecUtil;
+import org.opensearch.knn.index.util.KNNEngine;
+import org.opensearch.knn.jni.JNIService;
+
+import static org.opensearch.knn.common.KNNConstants.PARAMETERS;
+import static org.opensearch.knn.index.util.Faiss.FAISS_BINARY_INDEX_DESCRIPTION_PREFIX;
+
+public class NativeIndexBuilderScratch extends NativeIndexBuilder {
```
navneet1v, line 35: lets add java doc on all public classes and public functions

src/main/java/org/opensearch/knn/index/codec/native_index/NativeIndexBuilderScratchIter.java
```diff
@@ -0,0 +1,74 @@
...
+    protected void createIndex(NativeIndexInfo indexInfo, BinaryDocValues values) throws IOException {
+        long indexAddress = initIndexFromScratch(
+            indexInfo.numDocs,
+            indexInfo.vectorInfo.dimension,
+            indexInfo.knnEngine,
+            indexInfo.parameters
+        );
+        KNNCodecUtil.VectorBatch batch = KNNCodecUtil.getVectorBatch(values, getVectorTransfer(indexInfo.vectorInfo.vectorDataType), true);
+        try {
+            for (; !batch.finished; batch = KNNCodecUtil.getVectorBatch(
```
navneet1v, line 32:
```suggestion
            for (; batch.finished == false; batch = KNNCodecUtil.getVectorBatch(
```

navneet1v (commented):
Good work on the NativeEngineBuilder abstraction. Few comment on that interface just to align with how things are named at codec level. Please fix them

navneet1v:
@MrFlap add an entry in the Changelog.md file so that Changelog verifier GH action is successful.

src/main/java/org/opensearch/knn/index/codec/native_index/NativeIndexBuilderScratchIter.java
MrFlap: I get this would be more readable, but isn't it convention to avoid == false?

src/main/java/org/opensearch/knn/jni/JNIService.java
```diff
@@ -17,12 +17,73 @@
...
+
+import java.util.Locale;
 import java.util.Map;
 
 /**
  * Service to distribute requests to the proper engine jni service
  */
 public class JNIService {
+
+    public static long initIndexFromScratch(long size, int dim, Map<String, Object> parameters, KNNEngine knnEngine) {
```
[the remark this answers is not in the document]
MrFlap: I think they are all added unless I am missing some.

src/main/java/org/opensearch/knn/jni/FaissService.java
```diff
@@ -49,33 +49,17 @@ class FaissService {
...
-     * function call. So Java layer doesn't need to free up the memory. This is not an ideal behavior because Java layer
-     * created the memory address and that should only free up the memory. We are tracking the proper fix for this on this
-     * <a href="https://github.com/opensearch-project/k-NN/issues/1600">issue</a>
-     *
-     * @param ids array of ids mapping to the data passed in
-     * @param vectorsAddress address of native memory where vectors are stored
-     * @param dim dimension of the vector to be indexed
-     * @param indexPath path to save index file to
-     * @param parameters parameters to build index
-     */
```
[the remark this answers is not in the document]
MrFlap: What should I do to make sure they aren't lost?

src/main/java/org/opensearch/knn/index/codec/native_index/NativeIndexBuilder.java
MrFlap: That makes more sense.

src/main/java/org/opensearch/knn/index/codec/native_index/NativeIndexBuilder.java
MrFlap: Ok 👍 was unaware of the convention.

MrFlap:
> @MrFlap add an entry in the Changelog.md file so that Changelog verifier GH action is successful.

Done. Did I put it in the right place?

src/main/java/org/opensearch/knn/index/codec/native_index/NativeIndexBuilderScratchIter.java
MrFlap: Resolved

navneet1v:
> > @MrFlap add an entry in the Changelog.md file so that Changelog verifier GH action is successful.
> 
> Done. Did I put it in the right place?

No it should be under 2.x changes

jni/src/faiss_index_service.cpp
```diff
@@ -57,76 +57,172 @@ void SetExtraParameters(knn_jni::JNIUtilInterface * jniUtil, JNIEnv *env,
...
+
+    std::unique_ptr<faiss::IndexIDMap> idMap (reinterpret_cast<faiss::IndexIDMap *> (idMapAddress));
 
     // Write the index to disk
     faissMethods->writeIndex(idMap.get(), indexPath.c_str());
+
+    // Free the memory used by the index
+    delete idMap->index;
```
shatejas, line 159: Does this need to be in a try catch?

jni/src/faiss_index_service.cpp
```diff
@@ -137,27 +233,34 @@ void BinaryIndexService::createIndex(
...
+        omp_set_num_threads(threadCount);
+    }
+
+    std::unique_ptr<faiss::IndexBinaryIDMap> idMap (reinterpret_cast<faiss::IndexBinaryIDMap *> (idMapAddress));
 
     // Write the index to disk
     faissMethods->writeIndexBinary(idMap.get(), indexPath.c_str());
+
+    // Free the memory used by the index
+    delete idMap->index;
```
shatejas, line 263: same here, we should make sure we free up the memory when there is an exception

src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexWriter.java
```diff
@@ -0,0 +1,186 @@
...
+     * Method for creating a KNN index in the specified native library
+     *
+     * @param fieldInfo
+     * @param valuesProducer
+     * @param indexPath
+     * @param isMerge
+     * @param isRefresh
+     * @throws IOException
+     */
+    public void createKNNIndex(FieldInfo fieldInfo, DocValuesProducer valuesProducer, String indexPath, boolean isMerge, boolean isRefresh)
```
shatejas, line 77: This is favoring polymorphism over composition. Not ideal since its used for code reuse

MrFlap: I can change the classes to follow a composition-based design, but I think this is the way that @navneet1v wants it.

MrFlap: Should we meet to discuss which design would be better?

jni/src/faiss_index_service.cpp
MrFlap: How should I free the memory if delete is not working?

shatejas: wrap the call with try catch and delete it in catch

https://github.com/opensearch-project/k-NN/blob/main/jni/src/nmslib_wrapper.cpp#L269

src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexWriter.java
shatejas: Its a basic principle which is often overlooked. Its a no brainer that composition should take precedence unless you are enforcing behavior. I don't think there is a big discussion required here. I will leave it up to @navneet1v and you to decide. 

As a rule I only use polymorphism if it ties to [Liskov substitution](https://en.wikipedia.org/wiki/Liskov_substitution_principle). Exceptions will be only when there is some sort of a constraint

https://en.wikipedia.org/wiki/Composition_over_inheritance#:~:text=Composition%20over%20inheritance%20(or%20composite,from%20a%20base%20or%20parent

MrFlap: Ok. I think polymorphism would be better here since we can hide the internal type behind the abstract class so the reader doesn't have to worry about which NativeIndexWriter they are using.

It seems like this is a case Liskov subsititution. If we state the property is having the function createKNNIndex that makes an index and writes it to disk, then all subclasses share this property. That should be the only property of the class, since it holds no fields and only has one public function.

jni/src/faiss_index_service.cpp
MrFlap: Ah. Are you saying to wrap the whole function in try catch?

MrFlap: Added. Let me know if I did it right. Resolving for now.

src/main/java/org/opensearch/knn/jni/FaissService.java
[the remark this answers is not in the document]
navneet1v: What I meant was move them to new functions which you have created/right place where the functionality is provided now.

src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexWriter.java
navneet1v: On taking a deeper look I can see there are both pros and cons (one of the cons of composition here is the createIndex and write Index functions will not be used outside the codec, hence putting them in place makes sense), which will require further discussion on LLD of this class. Hence I propose we should merge this change in feature branch and then work on improving the LLD on the java side. Reason I want this change to be merged in feature branch is because this change is already became quite big. So lets just iterate over it in a feature branch. @shatejas please let me know your thoughts.

navneet1v: @shatejas I have changed the branch to https://github.com/opensearch-project/k-NN/tree/feature/iterative-index-build for merging this feature.

src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80DocValuesConsumer.java
```diff
@@ -127,142 +113,24 @@ public void addKNNBinaryField(FieldInfo field, DocValuesProducer valuesProducer,
...
-            if (efConstruction != null) {
-                algoParams.put(KNNConstants.METHOD_PARAMETER_EF_CONSTRUCTION, Integer.parseInt(efConstruction));
-            }
-
-            String m = fieldAttributes.get(KNNConstants.HNSW_ALGO_M);
-            if (m != null) {
-                algoParams.put(KNNConstants.METHOD_PARAMETER_M, Integer.parseInt(m));
-            }
-            parameters.put(PARAMETERS, algoParams);
+    public static NativeIndexWriter getWriter(FieldInfo fieldInfo, KNNEngine knnEngine) {
```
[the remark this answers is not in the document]
MrFlap: No. Will fix.

src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80DocValuesConsumer.java
```diff
@@ -127,142 +113,24 @@ public void addKNNBinaryField(FieldInfo field, DocValuesProducer valuesProducer,
...
-                pair.getVectorAddress(),
-                pair.getDimension(),
-                indexPath,
-                model.getModelBlob(),
-                parameters,
-                knnEngine
-            );
-            return null;
-        });
+        writeFooter(indexPath, engineFileName);
```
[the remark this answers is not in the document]
MrFlap: The thought was that this functionality doesn't interact with Faiss or nmslib, it just adds opensearch-specific info to the file. The functionality for NativeIndexWriter should be to get parameters, create the native index, then write the native index to disk. Whatever opensearch needs to do to these files should be outside of this class.

src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80DocValuesConsumer.java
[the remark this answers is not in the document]
MrFlap: Will do. Just didn't know standard practice.

src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexWriter.java
```diff
@@ -0,0 +1,186 @@
...
+    /**
+     * Method that gets the native index info from a given field
+     * @param fieldInfo
+     * @param valuesProducer
+     * @param indexPath
+     * @return native index info
+     * @throws IOException
+     */
+    private NativeIndexInfo getIndexInfo(FieldInfo fieldInfo, DocValuesProducer valuesProducer, String indexPath) throws IOException {
+        int numDocs = (int) KNNCodecUtil.getTotalLiveDocsCount(valuesProducer.getBinary(fieldInfo));
```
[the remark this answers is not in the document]
MrFlap: numDocs should never exceed INT_MAX, and this better demonstrates that fact than a long. We already represented numDocs as an int [here](https://github.com/opensearch-project/k-NN/blob/245995a6abe64b93bc44fef61e9d13df53696bcd/src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80DocValuesConsumer.java#L181) and [here](https://github.com/opensearch-project/k-NN/blob/245995a6abe64b93bc44fef61e9d13df53696bcd/jni/src/faiss_index_service.cpp#L76) for that reason.

src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexWriter.java
```diff
@@ -0,0 +1,186 @@
...
+        private FieldInfo fieldInfo;
+        private KNNEngine knnEngine;
+        private int numDocs;
+        private long arraySize;
+        private Map<String, Object> parameters;
+        private NativeVectorInfo vectorInfo;
+        private String indexPath;
+    }
+
+    protected final Logger logger = LogManager.getLogger(NativeIndexWriter.class);
```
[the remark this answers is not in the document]
MrFlap: Should I use it for all classes? This annotation makes a private logger.

src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexWriter.java
shatejas: >  So lets just iterate over it in a feature branch

I don't mind

>  benefit in terms of performance

Best perfomance can be achieved by the most unmaintainable code. Its not about performance. We should sync up offline as this PR is not the right place

src/main/java/org/opensearch/knn/jni/FaissService.java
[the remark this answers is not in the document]
MrFlap: Moved the docs here to insertToIndex.

src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80DocValuesConsumer.java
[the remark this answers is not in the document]
MrFlap: Resolving for now as I think this is the way to go.

src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexWriter.java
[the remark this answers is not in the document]
MrFlap: Resolving for now as I think this is the way to go.

src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexWriter.java
MrFlap: Going to resolve this for now.

shatejas (commented):
Looks good overall. Waiting for CI to pass before approval

Meanwhile there are 90 commits, can you squash them please into 1 commit before we merge it into a branch

src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexWriter.java
navneet1v: > Best perfomance can be achieved by the most unmaintainable code. Its not about performance. We should sync up offline as this PR is not the right place

My main motive to ask about performance was if there is perf benefits then its no brainer that we should go with composition for sure. But if there is no, then its open for discussion. 

See one of the reason why I was on fence for composition here is the write and create index functionality is not needed outside of this writer as of now. Hence, putting them in 1 place provide a tighter control. But if there are use-cases of those composition else where then I agree we should go with composition. Hence lets have a gh issue for this discussion.

src/main/java/org/opensearch/knn/index/codec/util/KNNCodecUtil.java
```diff
@@ -38,32 +38,51 @@ public static final class Pair {
...
-        return new KNNCodecUtil.Pair(
+
+        boolean finished = values.docID() == DocIdSetIterator.NO_MORE_DOCS;
+
+        return new KNNCodecUtil.VectorBatch(
             docIdList.stream().mapToInt(Integer::intValue).toArray(),
             vectorTransfer.getVectorAddress(),
             vectorTransfer.getDimension(),
-            serializationMode
+            serializationMode,
```
navneet1v, line 84: serializationMode is no longer required I think. Lets remove this from the code.

src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexWriterScratchIter.java
```diff
@@ -0,0 +1,72 @@
...
+            KNNCodecUtil.VectorBatch batch = KNNCodecUtil.getVectorBatch(
+                values,
+                getVectorTransfer(indexInfo.getVectorInfo().getVectorDataType()),
+                true
+            );
+            insertToIndex(batch, indexInfo.getKnnEngine(), indexAddress, indexInfo.getParameters());
+            if (batch.finished) {
+                break;
+            }
+        }
```
navneet1v, line 44: instead of having a while loop @MrFlap do you think if we can find out how trips to JNI will be required and then we use that as counter/terminating condition for breaking out of the loop? while(true) is giving me scares of infinite loop.

src/main/java/org/opensearch/knn/jni/FaissService.java
navneet1v: @MrFlap so did we check if we need thread count for writing the index?

src/main/java/org/opensearch/knn/index/codec/util/KNNCodecUtil.java
MrFlap: Ok. Will remove.

src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexWriterScratchIter.java
MrFlap: Right. I don't think there is a great way to do this since different batches could have different numbers of documents. The most maintainable way may be to accumulate the number of docs until we've reached totalLiveDocsCount?

MrFlap:
> @MrFlap so did we check if we need thread count for writing the index?

I just checked, and we currently don't. I can remove it.

src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexWriter.java
shatejas: > if there is perf benefits then its no brainer

Maintainability of the code is also a no brainer, especially when its not affecting performance

> I was on fence for composition here is the write and create index functionality is not needed outside of this writer as of now.

There are ways to do it. private-package with strategy pattern is one. The problem here is has a potential to violate [OCP](https://en.wikipedia.org/wiki/Open%E2%80%93closed_principle). Its open to modification, an implementation can simply override the public createIndex since the class cannot be made final. Composition generally helps to stick to [SRP](https://en.wikipedia.org/wiki/Single-responsibility_principle) and makes testing easier. For instance, it forces to test public createIndex in unit tests for every implementation, composition simplifies that.

navneet1v:
> > @MrFlap so did we check if we need thread count for writing the index?
> 
> I just checked, and we currently don't. I can remove it.

remove it then, if its needed.

src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexWriterScratchIter.java
navneet1v: Make sense. Not blocking this PR because of this change,

shatejas (approved):

navneet1v (approved):

PR #1141, opened 2023-09-18
Add graph creation stats to the KNNStats API

ryanbogan:
### Description
Adds graph creation stats to the KNNStats API

Old:
```
{
  "_nodes": {
    "total": 1,
    "successful": 1,
    "failed": 0
  },
  "cluster_name": "903360760764:navneet-test-domain",
  "circuit_breaker_triggered": false,
  "model_index_status": null,
  "nodes": {
    "H4Bf02YSRr2UBc_kjuEduw": {
      "graph_memory_usage_percentage": 0.00015215651,
      "graph_query_requests": 0,
      "graph_memory_usage": 3,
      "cache_capacity_reached": false,
      "load_success_count": 3,
      "training_memory_usage": 0,
      "indices_in_cache": {
        "products-shirts": {
          "graph_memory_usage_percentage": 0.00015215651,
          "graph_memory_usage": 3,
          "graph_count": 3
        }
      },
      "script_query_errors": 0,
      "hit_count": 0,
      "knn_query_requests": 0,
      "total_load_time": 9907321,
      "miss_count": 3,
      "knn_query_with_filter_requests": 0,
      "training_memory_usage_percentage": 0,
      "lucene_initialized": false,
      "graph_index_requests": 3,
      "faiss_initialized": true,
      "load_exception_count": 0,
      "training_errors": 0,
      "eviction_count": 0,
      "nmslib_initialized": false,
      "script_compilations": 0,
      "script_query_requests": 0,
      "graph_query_errors": 0,
      "indexing_from_model_degraded": false,
      "graph_index_errors": 0,
      "training_requests": 0,
      "script_compilation_errors": 0
    }
  }
}
```

New stats being added: 
```
"graph_stats": {
    "merge": {
        "current": 0,
        "current_docs": 0,
        "current_size_in_bytes": 0,
        "total": 0,
        "total_time_in_millis": 0,
        "total_docs": 0
    },
    "refresh": {
        "total": 0,
        "total_time_in_millis": 0
    }
}
```
### Issues Resolved
https://github.com/opensearch-project/k-NN/issues/999

 
### Check List
- [X] New functionality includes testing.
  - [X] All tests pass
- [X] New functionality has been documented.
  - [X] New functionality has javadoc added
- [X] Commits are signed as per the DCO using --signoff 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
[...]

PR #1844, opened 2024-07-17
Fix graph merge stats size calculation

ryanbogan:
### Description
Fixes the calculations for size of merges in the graph stats section of KNNStats API. This PR changes the logic to properly round values to the correct number of bytes.

Continuation of https://github.com/opensearch-project/k-NN/pull/1818, which had too many merge conflicts to fix cleanly.
 
### Issues Resolved
https://github.com/opensearch-project/k-NN/issues/1789
 
### Check List
  - [x] All tests pass


ryanbogan:
BWC failures are unrelated to this PR

luyuncheng (dismissed):
LGTM

src/main/java/org/opensearch/knn/index/codec/util/KNNCodecUtil.java
```diff
@@ -67,42 +67,44 @@ public static KNNCodecUtil.Pair getPair(final BinaryDocValues values, final Vect
...
 
+    /**
+     * This method provides a rough estimate of the number of bytes used for storing an array with the given parameters.
+     * @param numVectors number of vectors in the array
+     * @param vectorLength the length of each vector
+     * @param serializationMode serialization mode
+     * @return rough estimate of number of bytes used to store an array with the given parameters
+     */
     public static long calculateArraySize(int numVectors, int vectorLength, SerializationMode serializationMode) {
         if (serializationMode == SerializationMode.ARRAY) {
```
heemin32, line 78: It would be nice to add a comment on how this serialization mode store the vector in memory. Also please add a link to a reference where this formula came from.

If it is 2d array, float[][], the calculation here is not correct.
it should be something like(REFERENCE_SIZE or JAVA_ARRAY_HEADER_SIZE could be different between 32bit and 64bit)
```
int sizeOfVectorArray = vectorLength * FLOAT_BYTE_SIZE + JAVA_ARRAY_HEADER_SIZE;
sizeOfVectorArray = roundVectorSize(sizeOfVectorArray) * numVectors;

int sizeOfReferenceArray = numVectors * JAVA_REFERENCE_SIZE + JAVA_ARRAY_HEADER_SIZE;
sizeOfReferenceArray = roundVectorSize(sizeOfReferenceArray);

return sizeOfVectorArray + sizeOfReferenceArray;
```

ryanbogan: Yeah you're right, will change.  Forgot to add the reference on this recreated PR

heemin32: Hmm. You are right.. It was same :)

heemin32: Sorry. Let me look at this further. I still feel there is something incorrectness here..

heemin32: I looked at the code and this method mainly used to calculate how many off heap memory is being used during merge because we are allocating off heap memory to transfer vector data to JNI layer to construct a new graph. Then, I wonder why we need to distinguish between serialization mode. It is not about java memory or disk usage in serialized form but memory usage in pure float/byte array. Therefore, the memory size calculation should be simply `numVector * vectorLength * FLOAT_BYTE_SIZE` for float and `numVector * vectorLength` for byte.

Please let me know if I am understanding it incorrectly. @navneet1v, @martin-gaievski?
Original PR where this feature in introduced. https://github.com/opensearch-project/k-NN/pull/1141

ryanbogan: @navneet1v @martin-gaievski Specifically, I believe this is the conversation thread @heemin32 is referring to on that PR: https://github.com/opensearch-project/k-NN/pull/1141#discussion_r1336155264

navneet1v: this is an interesting question. Now I am trying to think for what exact purpose we want these metrics. I have to think over it. Earlier when it was added because everything was in JVM so having JVM stats was good. Now with streaming change and iterative graph build what these metrics should represent. I don't have an exact ans here I have to think more on this.

navneet1v: Okay now I am remember what it was for. I think we can remove the serialization mode and java things from the size calculation. Lets remove the java overhead. I agree with @heemin32  here.

heemin32 (approved):

src/main/java/org/opensearch/knn/index/codec/util/KNNCodecUtil.java
```diff
@@ -67,39 +61,18 @@ public static KNNCodecUtil.Pair getPair(final BinaryDocValues values, final Vect
...
-            int vectorSize = vectorLength;
-            if (vectorSize % JAVA_ROUNDING_NUMBER != 0) {
-                vectorSize += vectorSize % JAVA_ROUNDING_NUMBER;
-            }
-            int vectorsSize = numVectors * (vectorSize + JAVA_REFERENCE_SIZE);
-            if (vectorsSize % JAVA_ROUNDING_NUMBER != 0) {
-                vectorsSize += vectorsSize % JAVA_ROUNDING_NUMBER;
-            }
-            return vectorsSize;
+        if (serializationMode == SerializationMode.COLLECTIONS_OF_BYTES) {
```
navneet1v, line 72: can we get rid of this `serializationMode` attribute completely?

ryanbogan: This method is used to calculate array size from a Pair class typically.  The issue is that the KNNCodecUtil.Pair class only has doc id's, a vector address, dimension, and serialization mode as instance variables.  Therefore, without reading from the vector address I don't think it's possible to differentiate whether the data is floats or bytes without the serialization mode.

navneet1v: Can we use vector datatype to know if the vector is byte or float?

ryanbogan: Yeah I think that would work, binary type would be the same calculation as byte right?

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Bug Fixes
 * Corrected search logic for scenario with non-existent fields in filter [#1874](https://github.com/opensearch-project/k-NN/pull/1874)
 * Add script_fields context to KNNAllowlist [#1917] (https://github.com/opensearch-project/k-NN/pull/1917)
+* * Fix graph merge stats size calculation [#1844](https://github.com/opensearch-project/k-NN/pull/1844)
```
navneet1v, line 22: nit remove 2 `*`

src/main/java/org/opensearch/knn/index/codec/util/KNNCodecUtil.java
```diff
@@ -67,39 +62,18 @@ public static KNNCodecUtil.Pair getPair(final BinaryDocValues values, final Vect
...
+     * @param vectorLength the length of each vector
+     * @param vectorDataType type of data stored in each vector
+     * @return rough estimate of number of bytes used to store an array with the given parameters
+     */
+    public static long calculateArraySize(int numVectors, int vectorLength, VectorDataType vectorDataType) {
+        if (vectorDataType == VectorDataType.FLOAT) {
+            return numVectors * vectorLength * FLOAT_BYTE_SIZE;
         } else {
-            throw new IllegalStateException("Unreachable code");
+            return numVectors * vectorLength;
```
navneet1v, line 76: lets just say `VectorDataType.BYTE` or `VectorDataType.BINARY` then do `return numVectors * vectorLength;` and else throw exception. this will ensure that if any other  vector data type comes in and case is not handled then we throw exception to ensure that author handles it.

heemin32 (approved):

navneet1v (approved):

Issue #1277, opened 2023-10-25
[FEATURE] Lucene Inbuilt Scalar quantizer to convert  float 32 bits to 7 bits


vamshin:
**Is your feature request related to a problem?**
Inbuilt scalar quantizer to convert  float 32 bits to 7 bit using lucene engine.
Related to https://forum.opensearch.org/t/byte-size-vector-with-neural-search-on-the-fly/16416/2 

**What solution would you like?**
Inbuilt Byte quantizer to convert  float 32 bits to 7 bit.

**What alternatives have you considered?**
Another approach could be a ingestion processor "Byte quantizer" that takes 32bit float vectors and scalar quantize to Byte vectors

jmazanec15:
Lucene recently added scalar quantization support inside codec: https://github.com/apache/lucene/pull/12582. Would this solve the use case?

vamshin:
@jmazanec15 thanks for pointing. We should have this exposed from k-NN mappings/index settings or some other way. 
Given Lucene already has support, we could prioritize for 2.11 launch

jmazanec15:
@vamshin its a newer feature. Would require some testing on our end. Pretty interesting though - they found that they could avoid recomputing scalar quantization parameters for every segment while preserving recall.

Galilyou:
looking forward for this one guys; hopefully it makes it to v2.14.0 without having to be pushed to further releases

asfoorial:
Is this feature work with neural-search plugin out-of-box and hassle free? If I understood correctly, this feature will enable the _predict api to return quantized embeddings in, say int8, such that neural-search will automatically understand it and no need for manual quantizations

naveentatikonda:
> Does this feature work with neural-search plugin out-of-box and hassle free? 

Yes this feature will work out of box with Neural search plugin where you are using an ingest processor to convert the text to embeddings during ingestion. Will the same be true if you are using a neural query I will check and get back to you on that.

> If I understood correctly, this feature will enable the _predict api to return quantized embeddings in, say int8, such that neural-search will automatically understand it and no need for manual quantizations

no this is not how this feature will work. So the way it will be working is, at the lowest level (aka segments), we will quantize the float 32 bit floats to int8. All you would be needing to is create the kNN vector field with right quantizer during index creation. Rest all will be take care. No changes in predict api is required.

asfoorial:
> > Does this feature work with neural-search plugin out-of-box and hassle free?
> 
[...]

PR #1848, opened 2024-07-17
Add support for Lucene inbuilt Scalar Quantizer

naveentatikonda:
### Description
This PR adds support for inbuilt Scalar Quantizer with Lucene engine which takes fp32 vectors as input and dynamically quantizes them based on the number of bits(default `7`) and other parameters like confidence_interval provided in the index mapping. For now, we are not supporting 8 bits due to a recall issue and only supporting 7 bits. 

#### UX changes
To keep it consistent with Faiss, we are adding `sq` encoder under `parameters` field. The `sq` encoder in Lucene supports 3 optional parameters:

- `bits` - Which decides the number of buckets for the input vector data to be quantized into; Defaults to `7`.
- `confidence_interval` - Which is used to determine the `minQuantile` and `maxQuantile` params used in the Quantization process. Acceptable values are 0, >=0.9 && <=1.0. If the confidence_interval is not provided we will set default value as `null` which is computed later in Lucene from the dimension of the vector as `1 - 1/(1 + dimension)`.

Sample Index mapping using `sq` encoder is shown below:

```
    "mappings": {
        "properties": {
            "my_vector1": {
                "type": "knn_vector",
                "dimension": 2,
                "method": {
                    "name": "hnsw",
                    "space_type": "l2",
                    "engine": "lucene",
                    "parameters": {
                        "encoder": {
                            "name": "sq",
                            "parameters": {
                                "bits": 7,
                                "confidence_interval": 1.0
                            }
                        }
                        "ef_construction": 128,
                        "m": 24,
                    }
                }
            }
        }
    }
```

### Related Issues
https://github.com/opensearch-project/k-NN/issues/1277

### Check List


naveentatikonda:
Note - I'm still running some benchmarking tests for 4 bits, will complete by weekend and include 4 bits if there are no performance issues. That will be a very small change.

src/main/java/org/opensearch/knn/index/Parameter.java
```diff
@@ -204,6 +205,71 @@ public ValidationException validateWithData(Object value, VectorSpaceInfo vector
...
+            Predicate<Double> validator,
+            BiFunction<Double, VectorSpaceInfo, Boolean> validatorWithData
+        ) {
+            super(name, defaultValue, validator, validatorWithData);
+        }
+
+        @Override
+        public ValidationException validate(Object value) {
+            ValidationException validationException = null;
+            if (value.equals(0)) value = 0.0;
```
Vikasht34, line 228: It will throw NPE when Object is Null.

src/main/java/org/opensearch/knn/index/Parameter.java
```diff
@@ -204,6 +205,71 @@ public ValidationException validateWithData(Object value, VectorSpaceInfo vector
...
+
+            if (!(value instanceof Double)) {
+                validationException = new ValidationException();
```
Vikasht34, line 231: I see this code is getting duplicated 

 ```
validationException = new ValidationException();
                validationException.addValidationError(
                    String.format(Locale.ROOT, "Value not of type Double for Double " + "parameter \"%s\".", getName())
                );
                return validationException;
```

Let's create a private function createValidationException which takes string and return exception.

File: src/main/java/org/opensearch/knn/index/Parameter.java [hunk 3]
```diff
@@ -204,6 +205,71 @@ public ValidationException validateWithData(Object value, VectorSpaceInfo vector
...
+                    String.format(Locale.ROOT, "Parameter validation failed for Double " + "parameter \"%s\".", getName())
+                );
+            }
+            return validationException;
+        }
+
+        @Override
+        public ValidationException validateWithData(Object value, VectorSpaceInfo vectorSpaceInfo) {
+            ValidationException validationException = null;
+            if (!(value instanceof Double)) {
```
Vikasht34, line 250: Let's have  Null check on Value to differentiate errors

src/main/java/org/opensearch/knn/index/codec/Function5Arity.java
```diff
@@ -0,0 +1,11 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.knn.index.codec;
+
+@FunctionalInterface
+public interface Function5Arity<S, T, U, V, X, R> {
```
Vikasht34, line 9: This is going to solve the immediate problem where we need five params and return VectorFormat Class . These are two draw back on this interface.

1. If tomorrow I need sixth Param , I need to create separate interface with six param .
2. On every Field Vector format class , we are going to write like the way it is getting written now 

```
(maxConnm, beamWidth) -> new Lucene99HnswVectorsFormat(maxConnm, beamWidth),
            (maxConnm, beamWidth, confidenceInterval, bits, compress) -> new Lucene99HnswScalarQuantizedVectorsFormat(
                maxConnm,
                beamWidth,
                1,
                bits,
                compress,
                confidenceInterval,
                null
            `)`
```

We can make this extensible like this 

Create a class like this 

```
public class KnnVectorsFormatParams {
    private final int maxConnections;
    private final int beamWidth;
    private final float confidenceInterval;
    private final int bits;
    private final boolean compressFlag;

    public KnnVectorsFormatParams(int maxConnections, int beamWidth, float confidenceInterval, int bits, boolean compressFlag) {
        this.maxConnections = maxConnections;
        this.beamWidth = beamWidth;
        this.confidenceInterval = confidenceInterval;
        this.bits = bits;
        this.compressFlag = compressFlag;
    }

    public int getMaxConnections() {
        return maxConnections;
    }

    public int getBeamWidth() {
        return beamWidth;
    }

    public float getConfidenceInterval() {
        return confidenceInterval;
    }

    public int getBits() {
        return bits;
    }

    public boolean isCompressFlag() {
        return compressFlag;
    }
}

```

Modify Validate Method 

```
private final Function<KnnVectorsFormatParams, KnnVectorsFormat> quantizedVectorsFormatSupplier;

private KnnVectorsFormat validateAndApplyQuantizedVectorsFormatForLuceneEngine(
        final Map<String, Object> params,
        final String field,
        final int maxConnections,
        final int beamWidth
    ) {

        return null;
    }

        return null;
    }

    MethodComponentContext encoderMethodComponentContext = (MethodComponentContext) params.get(METHOD_ENCODER_PARAMETER);
    if (!ENCODER_SQ.equals(encoderMethodComponentContext.getName())) {
        return null;
    }

    Map<String, Object> sqEncoderParams = encoderMethodComponentContext.getParameters();
    Float confidenceInterval = getConfidenceInterval(sqEncoderParams);
    int bits = getBits(sqEncoderParams);
    boolean compressFlag = getCompressFlag(sqEncoderParams);

    log.debug(
        "Initialize KNN vector format for field [{}] with params [{}] = \"{}\", [{}] = \"{}\", [{}] = \"{}\", [{}] = \"{}\", [{}] = \"{}\"",
        field, MAX_CONNECTIONS, maxConnections, BEAM_WIDTH, beamWidth, LUCENE_SQ_CONFIDENCE_INTERVAL, confidenceInterval, LUCENE_SQ_BITS, bits, LUCENE_SQ_COMPRESS, compressFlag
    );

    KnnVectorsFormatParams formatParams = new KnnVectorsFormatParams(maxConnections, beamWidth, confidenceInterval, bits, compressFlag);
    return quantizedVectorsFormatSupplier.apply(formatParams);
}








```


and Now 

This is simplified like this 

```
public class KNN990PerFieldKnnVectorsFormat extends BasePerFieldKnnVectorsFormat {
    public KNN990PerFieldKnnVectorsFormat(final Optional<MapperService> mapperService) {
        super(
            mapperService,
            Lucene99HnswVectorsFormat.DEFAULT_MAX_CONN,
            Lucene99HnswVectorsFormat.DEFAULT_BEAM_WIDTH,
            () -> new Lucene99HnswVectorsFormat(),
            params -> new Lucene99HnswVectorsFormat(
                params.getMaxConnections(),
                params.getBeamWidth()
            ),
            params -> new Lucene99HnswScalarQuantizedVectorsFormat(
                params.getMaxConnections(),
                params.getBeamWidth(),
                1,
                params.getBits(),
                params.isCompressFlag(),
                params.getConfidenceInterval(),
                null
            )
        );
    }
```

With this , 
1. we won't end up creating multiple function inteface , we can just modify param class to include new param.
2. It's cleaner and more readble way.

src/main/java/org/opensearch/knn/index/codec/BasePerFieldKnnVectorsFormat.java
```diff
@@ -30,6 +41,7 @@ public abstract class BasePerFieldKnnVectorsFormat extends PerFieldKnnVectorsFor
     private final int defaultBeamWidth;
     private final Supplier<KnnVectorsFormat> defaultFormatSupplier;
     private final BiFunction<Integer, Integer, KnnVectorsFormat> formatSupplier;
+    private final Function5Arity<Integer, Integer, Float, Integer, Boolean, KnnVectorsFormat> quantizedVectorsFormatSupplier;
```
navneet1v, line 44: this is complete abuse of Functional interface, it provides no understanding what is the meaning of `Integer, Integer, Float, Integer, Boolean`. Put it under a class and use that class.

src/main/java/org/opensearch/knn/index/codec/BasePerFieldKnnVectorsFormat.java
```diff
@@ -59,6 +86,46 @@ public KnnVectorsFormat getKnnVectorsFormatForField(final String field) {
         return formatSupplier.apply(maxConnections, beamWidth);
     }
 
+    private KnnVectorsFormat validateAndApplyQuantizedVectorsFormatForLuceneEngine(
```
navneet1v, line 89: function with multiple responsibilities. Please ensure that functions have a single responsibility here otherwise the scope of function will just increase and it becomes unmanagable.

src/main/java/org/opensearch/knn/index/codec/KNN920Codec/KNN920PerFieldKnnVectorsFormat.java
```diff
@@ -22,7 +22,9 @@ public KNN920PerFieldKnnVectorsFormat(final Optional<MapperService> mapperServic
             Lucene92HnswVectorsFormat.DEFAULT_MAX_CONN,
             Lucene92HnswVectorsFormat.DEFAULT_BEAM_WIDTH,
             () -> new Lucene92HnswVectorsFormat(),
-            (maxConnm, beamWidth) -> new Lucene92HnswVectorsFormat(maxConnm, beamWidth)
+            (maxConnm, beamWidth) -> new Lucene92HnswVectorsFormat(maxConnm, beamWidth),
+            (maxConnm, beamWidth, confidenceInterval, bits, compress) -> new Lucene92HnswVectorsFormat(maxConnm, beamWidth)
```
navneet1v, line 26: why we are updating the older codec? is it needed?

navneet1v:
> * `bits` - Which decides the number of buckets for the input vector data to be quantized into; Defaults to `7`.
> * `confidence_interval` - Which is used to determine the `minQuantile` and `maxQuantile` params used in the Quantization process. Acceptable values are 0, >=0.9 && <=1.0. If the confidence_interval is not provided we will set default value as `null` which is computed later in Lucene from the dimension of the vector as `1 - 1/(1 + dimension)`.
> * `compress` - Default value is `false`, which is useful only with 4 bits to compress and provide more memory reduction.

Seeing these different parameters and how they work. Lets ensure documentation is very clear otherwise there will be a lot of confusions.

src/main/java/org/opensearch/knn/index/codec/Function5Arity.java
navneet1v: The idea of having a params class is really good. One thing I would like to add here is we should have a base class of parms of HNSW and then for different format we should have child classes.

I would also like to see how we can abstract the validation logic too in those classes to ensure maximum code reusability.

src/main/java/org/opensearch/knn/common/KNNConstants.java
```diff
@@ -16,6 +16,8 @@ public class KNNConstants {
     public static final String KNN_METHOD = "method";
     public static final String NAME = "name";
     public static final String PARAMETERS = "parameters";
+    public static final String MAX_CONNECTIONS = "max_connections";
```
jmazanec15, line 19: These arent going in the API right?

src/main/java/org/opensearch/knn/index/codec/BasePerFieldKnnVectorsFormat.java
```diff
@@ -59,6 +86,46 @@ public KnnVectorsFormat getKnnVectorsFormatForField(final String field) {
...
+        if (!(params.get(METHOD_ENCODER_PARAMETER) instanceof MethodComponentContext)) {
+            return null;
+        }
+        MethodComponentContext encoderMethodComponentContext = (MethodComponentContext) params.get(METHOD_ENCODER_PARAMETER);
+        if (!ENCODER_SQ.equals(encoderMethodComponentContext.getName())) {
+            return null;
+        }
+        Map<String, Object> sqEncoderParams = encoderMethodComponentContext.getParameters();
+        Float confidenceInterval = getConfidenceInterval(sqEncoderParams);
+        int bits = getBits(sqEncoderParams);
```
jmazanec15, line 110: Is there validation on this?

jmazanec15 (commented):
Whats the validation on the bits parameter?

jmazanec15:
Is 7-bit quantization implemented significantly different than 8-bit?

src/main/java/org/opensearch/knn/index/Parameter.java
naveentatikonda: Ack

src/main/java/org/opensearch/knn/index/codec/Function5Arity.java
naveentatikonda: Thanks for calling it out. Based on the suggestions provided, I created a base class for Params and a separate child class for quantizedformat params. Please take a look

src/main/java/org/opensearch/knn/index/codec/BasePerFieldKnnVectorsFormat.java
naveentatikonda: Ack

src/main/java/org/opensearch/knn/index/codec/KNN920Codec/KNN920PerFieldKnnVectorsFormat.java
naveentatikonda: It is forcing me to do that as part of initializing the constructor. Still working on that to see if we can skip it somehow. Vikash is also helping me here looking for alternate options.

src/main/java/org/opensearch/knn/index/codec/KNNVectorsFormatParams.java
```diff
@@ -0,0 +1,45 @@
...
+@Getter
+public class KNNVectorsFormatParams {
+    private int maxConnections;
+    private int beamWidth;
+
+    boolean validate(final Map<String, Object> params) {
+        return false;
+    }
+
+    void initialize(final Map<String, Object> params, int defaultMaxConnections, int defaultBeamWidth) {
```
navneet1v, line 27: why package private?

naveentatikonda:
> Whats the validation on the bits parameter?

In Lucene.java as part of parameter validation, created a list of integers which has the bits that we are supporting and validating against it.

src/main/java/org/opensearch/knn/index/codec/Function5Arity.java
navneet1v: since you have separated the params out why do we still need this interface?

src/main/java/org/opensearch/knn/common/KNNConstants.java
naveentatikonda: They are part of API, I didn't make any change here. This is the `m` value right ?

src/main/java/org/opensearch/knn/index/codec/BasePerFieldKnnVectorsFormat.java
naveentatikonda: The parameter validations are in Lucene.java class

naveentatikonda:
> Is 7-bit quantization implemented significantly different than 8-bit?

no, the difference is the number of buckets we are quantizing the given vectors into, which is `2^(bits) - 1` buckets

src/main/java/org/opensearch/knn/index/codec/Function5Arity.java
naveentatikonda: not required. I removed it now

src/main/java/org/opensearch/knn/index/codec/KNNVectorsFormatParams.java
naveentatikonda: changed to protected

src/main/java/org/opensearch/knn/index/codec/KNNVectorsFormatParams.java
```diff
@@ -0,0 +1,45 @@
...
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+import org.opensearch.knn.common.KNNConstants;
+
+import java.util.Map;
+
+/**
+ * Class provides params for LuceneHNSWVectorsFormat
+ */
+@NoArgsConstructor
```
navneet1v, line 17: why no args constructor?

src/main/java/org/opensearch/knn/index/codec/KNNVectorsFormatParams.java
```diff
@@ -0,0 +1,45 @@
...
+    private int beamWidth;
+
+    protected boolean validate(final Map<String, Object> params) {
+        return false;
+    }
+
+    protected void initialize(final Map<String, Object> params, int defaultMaxConnections, int defaultBeamWidth) {
+        this.maxConnections = getMaxConnections(params, defaultMaxConnections);
+        this.beamWidth = getBeamWidth(params, defaultBeamWidth);
+    }
```
navneet1v, line 30: why this cannot be done in constructor?

src/main/java/org/opensearch/knn/index/codec/KNNVectorsFormatParams.java
```diff
@@ -0,0 +1,45 @@
...
+        }
+        return defaultMaxConnections;
+    }
+
+    private int getBeamWidth(final Map<String, Object> params, int defaultBeamWidth) {
+        if (params != null && params.containsKey(KNNConstants.METHOD_PARAMETER_EF_CONSTRUCTION)) {
+            return (int) params.get(KNNConstants.METHOD_PARAMETER_EF_CONSTRUCTION);
+        }
+        return defaultBeamWidth;
+    }
```
navneet1v, line 44: shouldn't the default parameter be the part of the public getter rather than the private getter?

src/main/java/org/opensearch/knn/index/codec/KNNScalarQuantizedVectorsFormatParams.java
```diff
@@ -0,0 +1,89 @@
...
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ * Modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+package org.opensearch.knn.index.codec;
```
navneet1v, line 12: move the params to org.opensearch.knn.index.codec.model or something similar

src/main/java/org/opensearch/knn/index/codec/KNNVectorsFormatParams.java
```diff
@@ -0,0 +1,45 @@
...
+/**
+ * Class provides params for LuceneHNSWVectorsFormat
+ */
+@NoArgsConstructor
+@Getter
+public class KNNVectorsFormatParams {
+    private int maxConnections;
+    private int beamWidth;
+
+    protected boolean validate(final Map<String, Object> params) {
```
navneet1v, line 23: why this is protected? Shouldn't this be a public function?

src/main/java/org/opensearch/knn/index/codec/KNNScalarQuantizedVectorsFormatParams.java
```diff
@@ -0,0 +1,89 @@
...
+        this.bits = getBits(sqEncoderParams);
+        this.compressFlag = getCompressFlag(sqEncoderParams);
+    }
+
+    private Float getConfidenceInterval(final Map<String, Object> params) {
+
+        if (params != null && params.containsKey(LUCENE_SQ_CONFIDENCE_INTERVAL)) {
+            if (params.get(LUCENE_SQ_CONFIDENCE_INTERVAL).equals(0)) return Float.valueOf(0);
+
+            return ((Double) params.get(LUCENE_SQ_CONFIDENCE_INTERVAL)).floatValue();
```
navneet1v, line 70:
```suggestion
            if (params.get(LUCENE_SQ_CONFIDENCE_INTERVAL).equals(0)) {
              return Float.valueOf(0);
            }
            return ((Double) params.get(LUCENE_SQ_CONFIDENCE_INTERVAL)).floatValue();
```


nit: better readability

src/main/java/org/opensearch/knn/index/codec/KNNScalarQuantizedVectorsFormatParams.java
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    private int getBits(final Map<String, Object> params) {
+        if (params != null && params.containsKey(LUCENE_SQ_BITS)) {
+            return (int) params.get(LUCENE_SQ_BITS);
+        }
+        return LUCENE_SQ_DEFAULT_BITS;
+    }
+
+    private boolean getCompressFlag(final Map<String, Object> params) {
```
navneet1v, line 83: instead of get, you could have named these functions as initBits initCompressFlag and then set the required class variables in these functions.

src/main/java/org/opensearch/knn/index/codec/KNNVectorsFormatParams.java
naveentatikonda: removed it

src/main/java/org/opensearch/knn/index/codec/KNNVectorsFormatParams.java
naveentatikonda: Now, after creating a separate package we need to make it public. Earlier protected is good enough

src/main/java/org/opensearch/knn/index/codec/KNN920Codec/KNN920PerFieldKnnVectorsFormat.java
Vikasht34: There are two way to avoid changing in base classes , 
1. Overload constructor in baseFormat Class and this is the other way ...

Here is how base class will looks like 

``
```
@AllArgsConstructor
@Log4j2
public abstract class BasePerFieldKnnVectorsFormat extends PerFieldKnnVectorsFormat {
    private final Optional<MapperService> mapperService;
    private final int defaultMaxConnections;
    private final BiFunction<Integer, Integer, KnnVectorsFormat> formatSupplier;

    @Override
    public KnnVectorsFormat getKnnVectorsFormatForField(final String field) {
        var type = mapperService.map(service -> service.documentMapper().mappers().getMapper(field)).orElseThrow(
            () -> new IllegalStateException("No mapper for field [" + field + "]")
        ).fieldType(field);
        var params = type.getKnnMethodContext().getMethodComponentContext().getParameters();
        int maxConnections = getMaxConnections(params);
        int beamWidth = getBeamWidth(params);

        if (type.getKnnMethodContext().getKnnEngine() == KNNEngine.LUCENE
            && params != null
            && params.containsKey(METHOD_ENCODER_PARAMETER)) {
            KNNScalarQuantizedVectorsFormatParams knnScalarQuantizedVectorsFormatParams = new KNNScalarQuantizedVectorsFormatParams(
                params,
                defaultMaxConnections,
                defaultBeamWidth
            );
            if (knnScalarQuantizedVectorsFormatParams.validate(params)) {
                log.debug(
                    "Initialize KNN vector format for field [{}] with params [{}] = \"{}\", [{}] = \"{}\", [{}] = \"{}\", [{}] = \"{}\"",
                    field,
                    MAX_CONNECTIONS,
                    knnScalarQuantizedVectorsFormatParams.getMaxConnections(),
                    BEAM_WIDTH,
                    knnScalarQuantizedVectorsFormatParams.getBeamWidth(),
                    LUCENE_SQ_CONFIDENCE_INTERVAL,
                    knnScalarQuantizedVectorsFormatParams.getConfidenceInterval(),
                    LUCENE_SQ_BITS,
                    knnScalarQuantizedVectorsFormatParams.getBits()
                );
                return getScalarQuantizedVectorsFormatSupplier().apply(knnScalarQuantizedVectorsFormatParams);
            }
        }

        KNNVectorsFormatParams knnVectorsFormatParams = new KNNVectorsFormatParams(params, defaultMaxConnections, defaultBeamWidth);
        log.debug(
            "Initialize KNN vector format for field [{}] with params [max_connections] = \"{}\" and [beam_width] = \"{}\"",
            field,
            maxConnections,
            beamWidth,
            knnVectorsFormatParams.getMaxConnections(),
            knnVectorsFormatParams.getBeamWidth()
        );
        return vectorsFormatSupplier.apply(knnVectorsFormatParams);
    }

    protected Function<KNNScalarQuantizedVectorsFormatParams, KnnVectorsFormat> getScalarQuantizedVectorsFormatSupplier() {
        throw new UnsupportedOperationException("Scalar Quantized Vectors Format Supplier not supported");
    }

    private int getMaxConnections(Map<String, String> params) {
        // Implementation to get max connections from params or return defaultMaxConnections
        return defaultMaxConnections;
    }

    private int getBeamWidth(Map<String, String> params) {
        // Implementation to get beam width from params or return defaultBeamWidth
        return defaultBeamWidth;
    }
}
```
```
public class KNN990PerFieldKnnVectorsFormat extends BasePerFieldKnnVectorsFormat {
    private final Function<KNNScalarQuantizedVectorsFormatParams, KnnVectorsFormat> scalarQuantizedVectorsFormatSupplier;

    public KNN990PerFieldKnnVectorsFormat(final Optional<MapperService> mapperService) {
        super(
            mapperService,
            Lucene99HnswVectorsFormat.DEFAULT_MAX_CONN,
            Lucene99HnswVectorsFormat.DEFAULT_BEAM_WIDTH,
            () -> new Lucene99HnswVectorsFormat(),
            (maxConn, beamWidth) -> new Lucene99HnswVectorsFormat(maxConn, beamWidth),
            knnVectorsFormatParams -> new Lucene99HnswVectorsFormat(
                knnVectorsFormatParams.getMaxConnections(),
                knnVectorsFormatParams.getBeamWidth()
            )
        );
        this.scalarQuantizedVectorsFormatSupplier = knnScalarQuantizedVectorsFormatParams -> new Lucene99HnswScalarQuantizedVectorsFormat(
            knnScalarQuantizedVectorsFormatParams.getMaxConnections(),
            knnScalarQuantizedVectorsFormatParams.getBeamWidth(),
            1,
            knnScalarQuantizedVectorsFormatParams.getBits(),
            knnScalarQuantizedVectorsFormatParams.isCompressFlag(),
            knnScalarQuantizedVectorsFormatParams.getConfidenceInterval(),
            null
        );
    }

    @Override
    protected Function<KNNScalarQuantizedVectorsFormatParams, KnnVectorsFormat> getScalarQuantizedVectorsFormatSupplier() {
        return scalarQuantizedVectorsFormatSupplier;
    }
}
```

src/main/java/org/opensearch/knn/index/codec/KNN920Codec/KNN920PerFieldKnnVectorsFormat.java
```diff
@@ -22,7 +22,15 @@ public KNN920PerFieldKnnVectorsFormat(final Optional<MapperService> mapperServic
             Lucene92HnswVectorsFormat.DEFAULT_MAX_CONN,
             Lucene92HnswVectorsFormat.DEFAULT_BEAM_WIDTH,
             () -> new Lucene92HnswVectorsFormat(),
-            (maxConnm, beamWidth) -> new Lucene92HnswVectorsFormat(maxConnm, beamWidth)
+            knnVectorsFormatParams -> new Lucene92HnswVectorsFormat(
+                knnVectorsFormatParams.getMaxConnections(),
+                knnVectorsFormatParams.getBeamWidth()
+            ),
+            knnScalarQuantizedVectorsFormatParams -> new Lucene92HnswVectorsFormat(
```
Vikasht34, line 29: We can avoid changing in other child classes in two way , Create Overload constructor in base class or something like below

```
@AllArgsConstructor
@Log4j2
public abstract class BasePerFieldKnnVectorsFormat extends PerFieldKnnVectorsFormat {
    private final Optional<MapperService> mapperService;
    private final int defaultMaxConnections;
    private final BiFunction<Integer, Integer, KnnVectorsFormat> formatSupplier;

    @Override
    public KnnVectorsFormat getKnnVectorsFormatForField(final String field) {
        var type = mapperService.map(service -> service.documentMapper().mappers().getMapper(field)).orElseThrow(
            () -> new IllegalStateException("No mapper for field [" + field + "]")
        ).fieldType(field);
        var params = type.getKnnMethodContext().getMethodComponentContext().getParameters();
        int maxConnections = getMaxConnections(params);
        int beamWidth = getBeamWidth(params);

        if (type.getKnnMethodContext().getKnnEngine() == KNNEngine.LUCENE
            && params != null
            && params.containsKey(METHOD_ENCODER_PARAMETER)) {
            KNNScalarQuantizedVectorsFormatParams knnScalarQuantizedVectorsFormatParams = new KNNScalarQuantizedVectorsFormatParams(
                params,
                defaultMaxConnections,
                defaultBeamWidth
            );
            if (knnScalarQuantizedVectorsFormatParams.validate(params)) {
                log.debug(
                    "Initialize KNN vector format for field [{}] with params [{}] = \"{}\", [{}] = \"{}\", [{}] = \"{}\", [{}] = \"{}\"",
                    field,
                    MAX_CONNECTIONS,
                    knnScalarQuantizedVectorsFormatParams.getMaxConnections(),
                    BEAM_WIDTH,
                    knnScalarQuantizedVectorsFormatParams.getBeamWidth(),
                    LUCENE_SQ_CONFIDENCE_INTERVAL,
                    knnScalarQuantizedVectorsFormatParams.getConfidenceInterval(),
                    LUCENE_SQ_BITS,
                    knnScalarQuantizedVectorsFormatParams.getBits()
                );
                return getScalarQuantizedVectorsFormatSupplier().apply(knnScalarQuantizedVectorsFormatParams);
            }
        }

        KNNVectorsFormatParams knnVectorsFormatParams = new KNNVectorsFormatParams(params, defaultMaxConnections, defaultBeamWidth);
        log.debug(
            "Initialize KNN vector format for field [{}] with params [max_connections] = \"{}\" and [beam_width] = \"{}\"",
            field,
            maxConnections,
            beamWidth,
            knnVectorsFormatParams.getMaxConnections(),
            knnVectorsFormatParams.getBeamWidth()
        );
        return vectorsFormatSupplier.apply(knnVectorsFormatParams);
    }

    protected Function<KNNScalarQuantizedVectorsFormatParams, KnnVectorsFormat> getScalarQuantizedVectorsFormatSupplier() {
        throw new UnsupportedOperationException("Scalar Quantized Vectors Format Supplier not supported");
    }

    private int getMaxConnections(Map<String, String> params) {
        // Implementation to get max connections from params or return defaultMaxConnections
        return defaultMaxConnections;
    }

    private int getBeamWidth(Map<String, String> params) {
        // Implementation to get beam width from params or return defaultBeamWidth
        return defaultBeamWidth;
    }
}
```
```
public class KNN990PerFieldKnnVectorsFormat extends BasePerFieldKnnVectorsFormat {
    private final Function<KNNScalarQuantizedVectorsFormatParams, KnnVectorsFormat> scalarQuantizedVectorsFormatSupplier;

    public KNN990PerFieldKnnVectorsFormat(final Optional<MapperService> mapperService) {
        super(
            mapperService,
            Lucene99HnswVectorsFormat.DEFAULT_MAX_CONN,
            Lucene99HnswVectorsFormat.DEFAULT_BEAM_WIDTH,
            () -> new Lucene99HnswVectorsFormat(),
            (maxConn, beamWidth) -> new Lucene99HnswVectorsFormat(maxConn, beamWidth),
            knnVectorsFormatParams -> new Lucene99HnswVectorsFormat(
                knnVectorsFormatParams.getMaxConnections(),
                knnVectorsFormatParams.getBeamWidth()
            )
        );
        this.scalarQuantizedVectorsFormatSupplier = knnScalarQuantizedVectorsFormatParams -> new Lucene99HnswScalarQuantizedVectorsFormat(
            knnScalarQuantizedVectorsFormatParams.getMaxConnections(),
            knnScalarQuantizedVectorsFormatParams.getBeamWidth(),
            1,
            knnScalarQuantizedVectorsFormatParams.getBits(),
            knnScalarQuantizedVectorsFormatParams.isCompressFlag(),
            knnScalarQuantizedVectorsFormatParams.getConfidenceInterval(),
            null
        );
    }

    @Override
    protected Function<KNNScalarQuantizedVectorsFormatParams, KnnVectorsFormat> getScalarQuantizedVectorsFormatSupplier() {
        return scalarQuantizedVectorsFormatSupplier;
    }
}
```

navneet1v: +1, just creating 2 constructors should solve the problem of not touching the older codecs.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNN990PerFieldKnnVectorsFormat.java
```diff
@@ -23,7 +24,19 @@ public KNN990PerFieldKnnVectorsFormat(final Optional<MapperService> mapperServic
...
             () -> new Lucene99HnswVectorsFormat(),
-            (maxConnm, beamWidth) -> new Lucene99HnswVectorsFormat(maxConnm, beamWidth)
+            knnVectorsFormatParams -> new Lucene99HnswVectorsFormat(
+                knnVectorsFormatParams.getMaxConnections(),
+                knnVectorsFormatParams.getBeamWidth()
+            ),
+            knnScalarQuantizedVectorsFormatParams -> new Lucene99HnswScalarQuantizedVectorsFormat(
+                knnScalarQuantizedVectorsFormatParams.getMaxConnections(),
+                knnScalarQuantizedVectorsFormatParams.getBeamWidth(),
+                1,
```
navneet1v, line 34: move this to a constant. better to avoid the magic numbers.

src/main/java/org/opensearch/knn/common/KNNConstants.java
```diff
@@ -73,6 +75,14 @@ public class KNNConstants {
...
     // Lucene specific constants
     public static final String LUCENE_NAME = "lucene";
+    public static final String LUCENE_SQ_CONFIDENCE_INTERVAL = "confidence_interval";
+    public static final int DYNAMIC_CONFIDENCE_INTERVAL = 0;
+    public static final double MINIMUM_CONFIDENCE_INTERVAL = 0.9;
+    public static final double MAXIMUM_CONFIDENCE_INTERVAL = 1.0;
+    public static final String LUCENE_SQ_BITS = "bits";
+    public static final Integer LUCENE_SQ_DEFAULT_BITS = 7;
+    public static final List<Integer> LUCENE_SQ_BITS_SUPPORTED = List.of(7);
+    public static final String LUCENE_SQ_COMPRESS = "compress";
```
navneet1v, line 85: we are not using this constant. lets remove it.

src/main/java/org/opensearch/knn/common/KNNConstants.java
```diff
@@ -73,6 +75,14 @@ public class KNNConstants {
 
     // Lucene specific constants
     public static final String LUCENE_NAME = "lucene";
+    public static final String LUCENE_SQ_CONFIDENCE_INTERVAL = "confidence_interval";
+    public static final int DYNAMIC_CONFIDENCE_INTERVAL = 0;
+    public static final double MINIMUM_CONFIDENCE_INTERVAL = 0.9;
+    public static final double MAXIMUM_CONFIDENCE_INTERVAL = 1.0;
+    public static final String LUCENE_SQ_BITS = "bits";
+    public static final Integer LUCENE_SQ_DEFAULT_BITS = 7;
+    public static final List<Integer> LUCENE_SQ_BITS_SUPPORTED = List.of(7);
```
navneet1v, line 84: if the constant is getting used once, its better to move the constant in that class only rather than putting it in a generic constants class.

src/main/java/org/opensearch/knn/index/codec/KNN920Codec/KNN920PerFieldKnnVectorsFormat.java
naveentatikonda: Ack

src/test/java/org/opensearch/knn/index/LuceneEngineIT.java
```diff
@@ -466,6 +475,200 @@ public void testRadiusSearch_usingScoreThreshold_withFilter_usingCosineMetrics_u
...
+                SpaceType.L2,
+                VectorDataType.FLOAT,
+                LUCENE_SQ_DEFAULT_BITS,
+                confidenceInterval
+            )
+        );
+    }
+
+    @SneakyThrows
+    public void testSQ_withByteVectorDataType_thenThrowException() {
```
navneet1v, line 509: combine all 3 of them in 1 test with name testSQ_withInvalidParam_thenThrowException

navneet1v (commented):
Not seeing any Unit test, please add the unit test for codec classes.

navneet1v (dismissed):

src/main/java/org/opensearch/knn/index/util/Lucene.java
```diff
@@ -17,16 +17,43 @@
...
+    private static final List<Integer> LUCENE_SQ_BITS_SUPPORTED = List.of(7);
+
+    private final static Map<String, MethodComponent> HNSW_ENCODERS = ImmutableMap.of(
+        ENCODER_SQ,
+        MethodComponent.Builder.builder(ENCODER_SQ)
+            .addParameter(
+                LUCENE_SQ_CONFIDENCE_INTERVAL,
+                new Parameter.DoubleParameter(
+                    LUCENE_SQ_CONFIDENCE_INTERVAL,
+                    null,
```
jmazanec15, line 47: Why null default?

src/main/java/org/opensearch/knn/index/util/Lucene.java
```diff
@@ -44,6 +71,10 @@ public class Lucene extends JVMLibrary {
                         v -> v > 0
                     )
                 )
+                .addParameter(
+                    METHOD_ENCODER_PARAMETER,
+                    new Parameter.MethodComponentContextParameter(METHOD_ENCODER_PARAMETER, null, HNSW_ENCODERS)
```
jmazanec15, line 76: Why null here as well?

src/main/java/org/opensearch/knn/index/util/Lucene.java
naveentatikonda: For Lucene, the default value for confidence_interval is null which will later compute the confidence_interval based on dimension of vector (1-1/(1+d))

src/main/java/org/opensearch/knn/index/util/Lucene.java
naveentatikonda: Lucene doesn't have any default encoder like faiss. Should I set it to HNSW_ENCODERS ?

jmazanec15: For faiss, I set it to the flat encoding awhile ago. I just worry a little bit around NPE. Can we have a flat encoder be a noop?

naveentatikonda: Sure

jmazanec15 (approved):

navneet1v (approved):

src/main/java/org/opensearch/knn/index/Parameter.java
```diff
@@ -204,6 +206,85 @@ public ValidationException validateWithData(Object value, VectorSpaceInfo vector
...
+            return null;
+        }
+
+        @Override
+        public ValidationException validateWithData(Object value, VectorSpaceInfo vectorSpaceInfo) {
+            if (Objects.isNull(value)) {
+                String validationErrorMsg = String.format(Locale.ROOT, "Null value provided for Double " + "parameter \"%s\".", getName());
+                return getValidationException(validationErrorMsg);
+            }
+
```
heemin32, line 260: No conversion here?
```
if (value.equals(0)) {
  value = 0.0;
  }
```

heemin32 (approved):

heemin32:
> For now, we are not supporting 8 bits due to a recall issue and only supporting 7 bits.
Does that mean, user can only set 7 bits but nothing else?

naveentatikonda:
> > For now, we are not supporting 8 bits due to a recall issue and only supporting 7 bits.
> 
> Does that mean, user can only set 7 bits but nothing else?

yes, for now only 7 bits. We will support 8 bits and 4 bits after the underlying quantization issues are resolved in lucene.

Issue #1810, opened 2024-07-10
[FEATURE]Refactoring of code for better maintainability


heemin32:
After introducing binary vector format of which has its own code path in across multiple files, the needs of refactoring is required to enhance code maintainability. 

Related PR. https://github.com/opensearch-project/k-NN/pull/1781

Few action items are as follow
1. We have two query vector in `KNNQuery` class. One for float/byte vectors and another for binary vector. We want to refactor it so that it won't have two vector variables inside single KNNQuery. 
2. We have `VectorTransfer` class. The issue with this class is that the caller should call `close()` method at the end to make sure the transfer complete properly. We want to avoid such requirement.
3. We have `NestedFilteredIdsKNNByteIterator`, `FilteredIdsKNNByteIterator`, `NestedFilteredIdsKNNIterator`, and `FilteredIdsKNNIterator`. We want to remove the `Byte` iterator and reduce code duplication.
4. We introduced thin wrapper class `KNNVectorSimilarityFunction` to wrap lucene's `VectorSimilarityFunction`. We might want to see if there is other ways to avoid this thin wrapper.

PR #1850, opened 2024-07-18
Fix few bugs on binary index with Faiss HNSW

heemin32:
### Description
Did a manual test on following items. Will raise a new PR implementing this as integ test.

#### Test case
1. Can create binary index using Faiss HNSW
2. Can create binary index with knn index as false
3. Cannot create binary index using Lucene
4. Cannot create binary index using Nmslib
5. Cannot run training API on binary index
6. Cannot create binary index using encoding(sq)
7. Cannot run range query on binary index
8. Can run knn query on binary index
9. Can run knn query with nested field on binary index
10. Can run knn query with filtering on binary index
11. Can run knn query with filtering on nested field with binary index
12. Can run script scoring with hamming on binary index
13. Cannot run script scoring with hamming on non binary index
14. Cannot run script scoring with non hamming on binary index
15. Can run painless script with hamming on binary index
16. Cannot run painless script with hamming on non binary index
17. Cannot run painless script with non hamming on binary index
18. Can run script scoring with hamming on binary index with knn set as false
19. Can run painless script with hamming on binary index with knn set as false

#### Bug1 (2. Can create binary index with knn index as false)
* Actual: Following index creation succeeded but got an error while ingesting as default space was set as l2.
* Expected: It should set hamming as default space type and ingesting should succeed.
```
{
  "settings": {
    "index": {
      "knn": false
    }
  },
  "mappings": {
    "properties": {
      "my_vector": {
        "type": "knn_vector",
        "dimension": 8,
        "data_type": "binary"
      }
    }
  }
}
```

#### Bug2 (3. Cannot create binary index using Lucene)
* Actual: I got error  `\"hnsw\" configuration does not support space type: \"hammingbit\"` which is not appropriate without mentioning what engine it uses.
* Expected: The error message should have engine name as well because hnsw with faiss support hammingbit.
```
Create index with knn as false
{
  "settings": {
    "index": {
      "knn": false
    }
  },
  "mappings": {
    "properties": {
      "my_vector": {
        "type": "knn_vector",
        "dimension": 8,
        "data_type": "binary",
        "method": {
          "name": "hnsw",
          "engine": "lucene",
          "space_type": "hammingbit"
        }
      }
    }
  }
}
```

#### Bug 3. (7. Cannot run range query on binary index)
* Actual: Runtime error
* Expected: It should be Unsupported operation error with a message like "radial search is not supported with binary index"
```
"root_cause": [
    {
    "type": "runtime_exception",
    "reason": "java.lang.Exception: Query Vector cannot be null"
    }
],
```

#### Bug 4 (11. Can run knn query with filtering on nested field with binary index)
* Actual: Failed to run filtered query with binary Index. Error was happening while validating if exact search is required or not based on calculation cost
* Expected: Filtered query with binary index should work.
```
    "root_cause": [
      {
        "type": "null_pointer_exception",
        "reason": "Cannot read the array length because the return value of \"org.opensearch.knn.index.query.KNNQuery.getQueryVector()\" is null"
      }
    ],
```

### Related Issues
N/A

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/IndexUtil.java
+++ b/src/main/java/org/opensearch/knn/index/IndexUtil.java
@@ -190,6 +190,19 @@ public class IndexUtil {
             return exception;
         }
 
+        String vectorDataType = (String) fieldMap.get(VECTOR_DATA_TYPE_FIELD);
+        if (VectorDataType.BINARY.toString().equalsIgnoreCase(vectorDataType)) {
+            exception.addValidationError(
+                String.format(
+                    Locale.ROOT,
+                    "Field \"%s\" is of data type %s. Only FLOAT or BYTE is supported.",
+                    field,
+                    VectorDataType.BINARY
+                )
+            );
+            return exception;
+        }
+
         // Return if dimension does not need to be checked
         if (expectedDimension < 0) {
             return null;
--- a/src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
+++ b/src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
@@ -622,6 +622,9 @@ public class KNNQueryBuilder extends AbstractQueryBuilder<KNNQueryBuilder> {
                     String.format(Locale.ROOT, "Engine [%s] does not support radial search", knnEngine)
                 );
             }
+            if (vectorDataType == VectorDataType.BINARY) {
+                throw new UnsupportedOperationException(String.format(Locale.ROOT, "Binary data type does not support radial search"));
+            }
             RNNQueryFactory.CreateQueryRequest createQueryRequest = RNNQueryFactory.CreateQueryRequest.builder()
                 .knnEngine(knnEngine)
                 .indexName(indexName)
--- a/src/main/java/org/opensearch/knn/index/query/KNNWeight.java
+++ b/src/main/java/org/opensearch/knn/index/query/KNNWeight.java
@@ -476,8 +476,15 @@ public class KNNWeight extends Weight {
         if (isExactSearchThresholdSettingSet(filterThresholdValue)) {
             return filterThresholdValue >= filterIdsCount;
         }
+
         // if no setting is set, then use the default max distance computation value to see if we can do exact search.
-        return KNNConstants.MAX_DISTANCE_COMPUTATIONS >= filterIdsCount * knnQuery.getQueryVector().length;
+        /**
+         * TODO we can have a different MAX_DISTANCE_COMPUTATIONS for binary index as computation cost for binary index
+         * is cheaper than computation cost for non binary vector
+         */
+        return KNNConstants.MAX_DISTANCE_COMPUTATIONS >= filterIdsCount * (knnQuery.getVectorDataType() == VectorDataType.FLOAT
+            ? knnQuery.getQueryVector().length
+            : knnQuery.getByteQueryVector().length);
     }
 
     /**
--- a/src/main/java/org/opensearch/knn/index/KNNMethod.java
+++ b/src/main/java/org/opensearch/knn/index/KNNMethod.java
@@ -22,6 +22,7 @@ import java.util.Arrays;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
 
@@ -57,8 +58,10 @@ public class KNNMethod {
         if (!isSpaceTypeSupported(knnMethodContext.getSpaceType())) {
             errorMessages.add(
                 String.format(
-                    "\"%s\" configuration does not support space type: " + "\"%s\".",
+                    Locale.ROOT,
+                    "\"%s\" with \"%s\" configuration does not support space type: " + "\"%s\".",
                     this.methodComponent.getName(),
+                    knnMethodContext.getKnnEngine().getName().toLowerCase(Locale.ROOT),
                     knnMethodContext.getSpaceType().getValue()
                 )
             );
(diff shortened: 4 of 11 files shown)


junqiu-lei (dismissed):

src/main/java/org/opensearch/knn/index/IndexUtil.java
```diff
@@ -190,6 +190,14 @@ public static ValidationException validateKnnField(
             return exception;
         }
 
+        String vectorDataType = (String) fieldMap.get(VECTOR_DATA_TYPE_FIELD);
```
navneet1v, line 193: can `fieldMap.get(VECTOR_DATA_TYPE_FIELD)` be null?

src/main/java/org/opensearch/knn/index/IndexUtil.java
```diff
@@ -190,6 +190,14 @@ public static ValidationException validateKnnField(
...
+        if (VectorDataType.BINARY.toString().equalsIgnoreCase(vectorDataType)) {
```
navneet1v, line 194: we should convert the datatype to enum and then use `==` in `if` condition

src/main/java/org/opensearch/knn/index/IndexUtil.java
```diff
@@ -190,6 +190,14 @@ public static ValidationException validateKnnField(
...
+            exception.addValidationError(
+                String.format("Field \"%s\" is of data type %s. Only FLOAT or BYTE is supported.", field, VectorDataType.BINARY)
```
navneet1v, line 196: when using String.format set Locale.ROOT to ensure that UTs don't break if we validate the exception messages

ryanbogan (approved):
LGTM!  The integration tests for these cases will be in 2.16 right?

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
```diff
@@ -622,6 +622,11 @@ protected Query doToQuery(QueryShardContext context) {
                     String.format(Locale.ROOT, "Engine [%s] does not support radial search", knnEngine)
                 );
             }
+            if (vectorDataType == VectorDataType.BINARY) {
+                throw new UnsupportedOperationException(
+                    String.format(Locale.ROOT, "Binary data type does not support radial search", knnEngine)
```
navneet1v, line 627: knnEngine is not getting added in the exception message. You missed setting %s in the messsage,

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -477,7 +477,9 @@ private boolean canDoExactSearch(final int filterIdsCount) {
             return filterThresholdValue >= filterIdsCount;
         }
         // if no setting is set, then use the default max distance computation value to see if we can do exact search.
-        return KNNConstants.MAX_DISTANCE_COMPUTATIONS >= filterIdsCount * knnQuery.getQueryVector().length;
+        return KNNConstants.MAX_DISTANCE_COMPUTATIONS >= filterIdsCount * (knnQuery.getVectorDataType() == VectorDataType.FLOAT
+            ? knnQuery.getQueryVector().length
+            : knnQuery.getByteQueryVector().length);
```
navneet1v, line 482: with hamming distance we can do more computation for exact search. I would recommend adding a TODO here to ensure that we revisit the number of distance computation for different fields.

heemin32: ack

heemin32:
> LGTM! The integration tests for these cases will be in 2.16 right?
Will try. As the last resort, will do the manual test after code freeze once more.

heemin32:
> @heemin32 Can you provide intended behavior for the bugs as well?

Added.

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -476,8 +476,15 @@ private boolean canDoExactSearch(final int filterIdsCount) {
...
+
         // if no setting is set, then use the default max distance computation value to see if we can do exact search.
-        return KNNConstants.MAX_DISTANCE_COMPUTATIONS >= filterIdsCount * knnQuery.getQueryVector().length;
+        /**
+         * TODO we can have a different MAX_DISTANCE_COMPUTATIONS for binary index as computation cost for binary index
+         * is cheaper than computation cost for non binary vector
+         */
+        return KNNConstants.MAX_DISTANCE_COMPUTATIONS >= filterIdsCount * (knnQuery.getVectorDataType() == VectorDataType.FLOAT
+            ? knnQuery.getQueryVector().length
+            : knnQuery.getByteQueryVector().length);
```
shatejas, line 487: I don't like the fact that we have two vectors. This will create a lot of branching in the code. We should have gone with generics or something with a different type of query. Something similar to lucene.

shatejas (commented):
So I see three unit tests added. Does that cover all cases mentioned in the description?

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
heemin32: We are already aware of it and captured it in https://github.com/opensearch-project/k-NN/issues/1810

src/main/java/org/opensearch/knn/index/IndexUtil.java
navneet1v: if it can be null then the type casting to string will cause a NPE. Please handle this gracefully.

navneet1v:
@heemin32 lets ensure that GH action are successful before we merge this PR

src/main/java/org/opensearch/knn/index/IndexUtil.java
heemin32: It won't cause a NPE.

navneet1v: why?? casting a null to string cause NPE. I am missing something here.

heemin32: I tested it and it didn't cause NPE.

navneet1v (approved):

vamshin (approved):
LGTM! Thanks

Issue #1853, opened 2024-07-18
[RFC] Integrating KNNVectorsFormat in Native Vector Search Engine


navneet1v:
## Introduction

The issue focuses on providing detailed design for integrating KNNVectorsFormat in Native Vector Search Engines (like NMSLib and FAISSLib). Apart from just moving to the new vector format, the document takes another step forward to improve exact search user experience too. At the end issue touch bases on the plan of implementation to give an iterative way to implement things. 

## Background

KNNPlugin(aka Vector Engine) was added in OpenSearch back 2019, at that time Lucene didn’t support any native VectorFormat. To go around this, the decision was taken to represent vector as Binary DocValues and then override the 
BinaryDocValuesFormat to store vectors and build the vector data structures. With Lucene Version 9.0 Lucene added a new format optimized for Vectors. Since that the format has evolved and optimized with features like iterative graph builds, in built scaler quantization, optimized support for reading vectors from disk etc. 

### Earlier Investigation

In September 2023, we did [an investigation](https://github.com/opensearch-project/k-NN/issues/1087#issuecomment-1734568700)(thanks to @heemin32 who did the investigation) to get some details on what it takes to move from BinaryDocValuesFormat to KNNVectorsFormat. Below were the main concerns in a summarized fashion as per the older deep-dive.

#### Summarizing the Main Concerns from the earlier investigation(ref: Cons section of [this](https://github.com/opensearch-project/k-NN/issues/1087#issuecomment-1734568700))

1. Strong dependency on the Lucene with native engine implementation.
2. Increased disk space, as we will be storing vectors as doc values and also in KNNVectorsFormat.
3. Increased indexing latency because we will be adding vectors in both doc values and also in KNNVectorsFormat.
4. Increased code size: We used doc value codec to do all the works except indexing and searching. After the migration, we need our own implementation of codec which will increase the line of codes. However, as the code evolve and we keep feature parity between native engine and the Lucene engine, initial increase in code size will be paid off.
5. Effort to migration: The effort needs for this migration is not small as we need to write our own codec from scratch.
6. Not enough feature that we can get benefited out of the box: Graph build on the fly during indexing won’t happen for native engine even after the migration. Native engine should support adding node to an existing graph first and we need to implement to call the method in addField method.
7. With iterative graph builds in place we can also add more validation to ensure that we throttle request during indexing if memory is going above the limit of CB. Thanks to @jmazanec15 for suggesting this.

[...]

PR #1855, opened 2024-07-18
Introduce NativeEngineKNNVectorsFormat as a KNNVectorsFormat for Native engines

navneet1v:
### Description
Introduced NativeEngineKNNVectorsFormat as a KNNVectorsFormat for Native engines(faiss and nmslib). This PR doesn't change any code path or do the integration of the new format in the actual flow. This is more of a base work I am adding for next step of changes which are coming. 

**Changes include:**
1. Introduce the NativeEngineKNNVectorsFormat, readers and writers.
2. Added the unit test for this new format.
3. Currently the format is only writing the FlatVectors and not building the hnsw vector files. I will add this change in next few PRs.
4. Not adding the backport label on this PR. I will add the backport label once 2.16 is released, as this change is not intended for 2.16 version of Opensearch.

### Related Issues
https://github.com/opensearch-project/k-NN/issues/1853

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngineFieldVectorsWriter.java
+++ b/src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngineFieldVectorsWriter.java
@@ -1,0 +1,107 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ * Modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+package org.opensearch.knn.index.codec.KNN990Codec;
+
+import lombok.Getter;
+import org.apache.lucene.codecs.KnnFieldVectorsWriter;
+import org.apache.lucene.index.DocsWithFieldSet;
+import org.apache.lucene.index.FieldInfo;
+import org.apache.lucene.util.InfoStream;
+import org.apache.lucene.util.RamUsageEstimator;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * NativeEngineVectorFieldsWriter is a class that will be used to accumulate all the vectors during ingestion before
+ * lucene does a flush. This class ensures that KNNVectorWriter is free from generics and this class can encapsulate
+ * all the details related to vectors types and docIds.
+ *
+ * @param <T> float[] or byte[]
+ */
+class NativeEngineFieldVectorsWriter<T> extends KnnFieldVectorsWriter<T> {
+    private static final long SHALLOW_SIZE = RamUsageEstimator.shallowSizeOfInstance(NativeEngineFieldVectorsWriter.class);
+    private final FieldInfo fieldInfo;
+    /**
+     * We are using a map here instead of list, because for sampler interface for quantization we have to advance the iterator
+     * to a specific docId, there a list cannot be useful because a docId != index of the vector in the list. Similar
+     * thing is true when we have vector field in child document. There doc Ids will not be consistent. Hence, we need to
+     * use the map here.
+     */
+    @Getter
+    private final Map<Integer, T> vectors;
+    private int lastDocID = -1;
+    @Getter
+    private final DocsWithFieldSet docsWithField;
+    private final InfoStream infoStream;
+
+    static NativeEngineFieldVectorsWriter<?> create(final FieldInfo fieldInfo, final InfoStream infoStream) {
+        switch (fieldInfo.getVectorEncoding()) {
+            case FLOAT32:
+                return new NativeEngineFieldVectorsWriter<float[]>(fieldInfo, infoStream);
+            case BYTE:
+                return new NativeEngineFieldVectorsWriter<byte[]>(fieldInfo, infoStream);
+        }
+        throw new IllegalStateException("Unsupported Vector encoding : " + fieldInfo.getVectorEncoding());
+    }
+
+    NativeEngineFieldVectorsWriter(final FieldInfo fieldInfo, final InfoStream infoStream) {
+        this.fieldInfo = fieldInfo;
+        this.infoStream = infoStream;
+        vectors = new HashMap<>();
+        this.docsWithField = new DocsWithFieldSet();
+    }
+
+    /**
+     * Add new docID with its vector value to the given field for indexing. Doc IDs must be added in
+     * increasing order.
+     *
+     * @param docID int
+     * @param vectorValue T
+     */
+    @Override
+    public void addValue(int docID, T vectorValue) {
+        if (docID == lastDocID) {
+            throw new IllegalArgumentException(
+                "[NativeEngineKNNVectorWriter]VectorValuesField \""
+                    + fieldInfo.name
+                    + "\" appears more than once in this document (only one value is allowed per field)"
+            );
+        }
+        assert docID > lastDocID;
+        vectors.put(docID, vectorValue);
+        docsWithField.add(docID);
+        lastDocID = docID;
+    }
+
+    /**
+     * Used to copy values being indexed to internal storage.
+     *
+     * @param vectorValue an array containing the vector value to add
+     * @return a copy of the value; a new array
+     */
+    @Override
+    public T copyValue(T vectorValue) {
+        throw new UnsupportedOperationException("NativeEngineVectorFieldsWriter doesn't support copyValue operation");
+    }
+
+    /**
+     * Return the memory usage of this object in bytes. Negative values are illegal.
+     */
+    @Override
+    public long ramBytesUsed() {
+        return SHALLOW_SIZE + docsWithField.ramBytesUsed() + (long) this.vectors.size() * (long) (RamUsageEstimator.NUM_BYTES_OBJECT_REF
+            + RamUsageEstimator.NUM_BYTES_ARRAY_HEADER) + (long) this.vectors.size() * RamUsageEstimator.shallowSizeOfInstance(
+                Integer.class
+            ) + (long) vectors.size() * fieldInfo.getVectorDimension() * fieldInfo.getVectorEncoding().byteSize;
+    }
+}
--- a/src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsFormat.java
+++ b/src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsFormat.java
@@ -1,0 +1,68 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ * Modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+package org.opensearch.knn.index.codec.KNN990Codec;
+
+import org.apache.lucene.codecs.KnnVectorsFormat;
+import org.apache.lucene.codecs.KnnVectorsReader;
+import org.apache.lucene.codecs.KnnVectorsWriter;
+import org.apache.lucene.codecs.hnsw.DefaultFlatVectorScorer;
+import org.apache.lucene.codecs.hnsw.FlatVectorsFormat;
+import org.apache.lucene.codecs.lucene99.Lucene99FlatVectorsFormat;
+import org.apache.lucene.index.SegmentReadState;
+import org.apache.lucene.index.SegmentWriteState;
+
+import java.io.IOException;
+
+/**
+ * This is a Vector format that will be used for Native engines like Faiss and Nmslib for reading and writing vector
+ * related data structures.
+ */
+public class NativeEngines990KnnVectorsFormat extends KnnVectorsFormat {
+    /** The format for storing, reading, merging vectors on disk */
+    private static FlatVectorsFormat flatVectorsFormat;
+    private static final String FORMAT_NAME = "NativeEngines99KnnVectorsFormat";
+
+    public NativeEngines990KnnVectorsFormat() {
+        super(FORMAT_NAME);
+        flatVectorsFormat = new Lucene99FlatVectorsFormat(new DefaultFlatVectorScorer());
+    }
+
+    public NativeEngines990KnnVectorsFormat(final FlatVectorsFormat lucene99FlatVectorsFormat) {
+        super(FORMAT_NAME);
+        flatVectorsFormat = lucene99FlatVectorsFormat;
+    }
+
+    /**
+     * Returns a {@link KnnVectorsWriter} to write the vectors to the index.
+     *
+     * @param state {@link SegmentWriteState}
+     */
+    @Override
+    public KnnVectorsWriter fieldsWriter(final SegmentWriteState state) throws IOException {
+        return new NativeEngines990KnnVectorsWriter(state, flatVectorsFormat.fieldsWriter(state));
+    }
+
+    /**
+     * Returns a {@link KnnVectorsReader} to read the vectors from the index.
+     *
+     * @param state {@link SegmentReadState}
+     */
+    @Override
+    public KnnVectorsReader fieldsReader(final SegmentReadState state) throws IOException {
+        return new NativeEngines990KnnVectorsReader(state, flatVectorsFormat.fieldsReader(state));
+    }
+
+    @Override
+    public String toString() {
+        return "NativeEngines99KnnVectorsFormat(name=" + this.getClass().getSimpleName() + ", flatVectorsFormat=" + flatVectorsFormat + ")";
+    }
+}
--- a/src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
+++ b/src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
@@ -1,0 +1,162 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ * Modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+package org.opensearch.knn.index.codec.KNN990Codec;
+
+import org.apache.lucene.codecs.KnnVectorsReader;
+import org.apache.lucene.codecs.hnsw.FlatVectorsReader;
+import org.apache.lucene.index.ByteVectorValues;
+import org.apache.lucene.index.FieldInfo;
+import org.apache.lucene.index.FloatVectorValues;
+import org.apache.lucene.index.SegmentReadState;
+import org.apache.lucene.search.KnnCollector;
+import org.apache.lucene.search.ScoreDoc;
+import org.apache.lucene.search.TopDocs;
+import org.apache.lucene.search.TotalHits;
+import org.apache.lucene.util.Bits;
+import org.apache.lucene.util.IOUtils;
+
+import java.io.IOException;
+
+/**
+ * Vectors reader class for reading the flat vectors for native engines. The class provides methods for iterating
+ * over the vectors and retrieving their values.
+ */
+public class NativeEngines990KnnVectorsReader extends KnnVectorsReader {
+
+    private final FlatVectorsReader flatVectorsReader;
+
+    public NativeEngines990KnnVectorsReader(final SegmentReadState state, final FlatVectorsReader flatVectorsReader) {
+        this.flatVectorsReader = flatVectorsReader;
+    }
+
+    /**
+     * Checks consistency of this reader.
+     *
+     * <p>Note that this may be costly in terms of I/O, e.g. may involve computing a checksum value
+     * against large data files.
+     *
+     */
+    @Override
+    public void checkIntegrity() throws IOException {
+        flatVectorsReader.checkIntegrity();
+    }
+
+    /**
+     * Returns the {@link FloatVectorValues} for the given {@code field}. The behavior is undefined if
+     * the given field doesn't have KNN vectors enabled on its {@link FieldInfo}. The return value is
+     * never {@code null}.
+     *
+     * @param field {@link String}
+     */
+    @Override
+    public FloatVectorValues getFloatVectorValues(final String field) throws IOException {
+        return flatVectorsReader.getFloatVectorValues(field);
+    }
+
+    /**
+     * Returns the {@link ByteVectorValues} for the given {@code field}. The behavior is undefined if
+     * the given field doesn't have KNN vectors enabled on its {@link FieldInfo}. The return value is
+     * never {@code null}.
+     *
+     * @param field {@link String}
+     */
+    @Override
+    public ByteVectorValues getByteVectorValues(final String field) throws IOException {
+        return flatVectorsReader.getByteVectorValues(field);
+    }
+
+    /**
+     * Return the k nearest neighbor documents as determined by comparison of their vector values for
+     * this field, to the given vector, by the field's similarity function. The score of each document
+     * is derived from the vector similarity in a way that ensures scores are positive and that a
+     * larger score corresponds to a higher ranking.
+     *
+     * <p>The search is allowed to be approximate, meaning the results are not guaranteed to be the
+     * true k closest neighbors. For large values of k (for example when k is close to the total
+     * number of documents), the search may also retrieve fewer than k documents.
+     *
+     * <p>The returned {@link TopDocs} will contain a {@link ScoreDoc} for each nearest neighbor, in
+     * order of their similarity to the query vector (decreasing scores). The {@link TotalHits}
+     * contains the number of documents visited during the search. If the search stopped early because
+     * it hit {@code visitedLimit}, it is indicated through the relation {@code
+     * TotalHits.Relation.GREATER_THAN_OR_EQUAL_TO}.
+     *
+     * <p>The behavior is undefined if the given field doesn't have KNN vectors enabled on its {@link
+     * FieldInfo}. The return value is never {@code null}.
+     *
+     * @param field        the vector field to search
+     * @param target       the vector-valued query
+     * @param knnCollector a KnnResults collector and relevant settings for gathering vector results
+     * @param acceptDocs   {@link Bits} that represents the allowed documents to match, or {@code null}
+     *                     if they are all allowed to match.
+     */
+    @Override
+    public void search(String field, float[] target, KnnCollector knnCollector, Bits acceptDocs) throws IOException {
+        throw new UnsupportedOperationException("Search functionality using codec is not supported with Native Engine Reader");
+    }
+
+    /**
+     * Return the k nearest neighbor documents as determined by comparison of their vector values for
+     * this field, to the given vector, by the field's similarity function. The score of each document
+     * is derived from the vector similarity in a way that ensures scores are positive and that a
+     * larger score corresponds to a higher ranking.
+     *
+     * <p>The search is allowed to be approximate, meaning the results are not guaranteed to be the
+     * true k closest neighbors. For large values of k (for example when k is close to the total
+     * number of documents), the search may also retrieve fewer than k documents.
+     *
+     * <p>The returned {@link TopDocs} will contain a {@link ScoreDoc} for each nearest neighbor, in
+     * order of their similarity to the query vector (decreasing scores). The {@link TotalHits}
+     * contains the number of documents visited during the search. If the search stopped early because
+     * it hit {@code visitedLimit}, it is indicated through the relation {@code
+     * TotalHits.Relation.GREATER_THAN_OR_EQUAL_TO}.
+     *
+     * <p>The behavior is undefined if the given field doesn't have KNN vectors enabled on its {@link
+     * FieldInfo}. The return value is never {@code null}.
+     *
+     * @param field        the vector field to search
+     * @param target       the vector-valued query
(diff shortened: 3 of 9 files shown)


src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngineVectorFieldsWriter.java
```diff
@@ -0,0 +1,102 @@
...
+import java.util.Map;
+
+/**
+ * NativeEngineVectorFieldsWriter is a class that will be used to accumulate all the vectors during ingestion before
+ * lucene does a flush. This class ensures that KNNVectorWriter is free from generics and this class can encapsulate
+ * all the details related to vectors types and docIds.
+ *
+ * @param <T> float[] or byte[]
+ */
+class NativeEngineVectorFieldsWriter<T> extends KnnFieldVectorsWriter<T> {
```
jmazanec15, line 31: nit: NativeEngineVectorFieldsWriter -> NativeEngineFieldVectorsWriter (in line with KnnFieldVectorsWriter)

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsFormat.java
```diff
@@ -0,0 +1,68 @@
...
+    /** The format for storing, reading, merging vectors on disk */
+    private static FlatVectorsFormat flatVectorsFormat;
+    private static final String FORMAT_NAME = "NativeEngines99KnnVectorsFormat";
+
+    public NativeEngines990KnnVectorsFormat() {
+        super(FORMAT_NAME);
+        flatVectorsFormat = new Lucene99FlatVectorsFormat(new DefaultFlatVectorScorer());
+    }
+
+    public NativeEngines990KnnVectorsFormat(final Lucene99FlatVectorsFormat lucene99FlatVectorsFormat) {
```
jmazanec15, line 39: Does arg need to be Lucene99FlatVectorsFormat? Can it be FlatVectorsFormat?

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
```diff
@@ -0,0 +1,162 @@
...
+     * close may fail require careful attention. It is strongly advised
+     * to relinquish the underlying resources and to internally
+     * <em>mark</em> the {@code Closeable} as closed, prior to throwing
+     * the {@code IOException}.
+     *
+     * @throws IOException if an I/O error occurs
+     */
+    @Override
+    public void close() throws IOException {
+        IOUtils.close(flatVectorsReader);
```
jmazanec15, line 152: FAR: I wonder if we can remove the whole evict from cache on file deletion and call it here even though we do not use this Reader for search?

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
```diff
@@ -0,0 +1,162 @@
...
+     * @param knnCollector a KnnResults collector and relevant settings for gathering vector results
+     * @param acceptDocs   {@link Bits} that represents the allowed documents to match, or {@code null}
+     *                     if they are all allowed to match.
+     */
+    @Override
+    public void search(String field, float[] target, KnnCollector knnCollector, Bits acceptDocs) throws IOException {
+        throw new UnsupportedOperationException("Search functionality using codec is not supported with Native Engine Reader");
+    }
+
+    /**
```
jmazanec15, line 107: nit: Do we need copy and paste java docs for unsupported?

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriter.java
```diff
@@ -0,0 +1,115 @@
...
+     * Flush all buffered data on disk. This is not fsync. This is lucene flush.
+     *
+     * @param maxDoc int
+     * @param sortMap {@link Sorter.DocMap}
+     */
+    @Override
+    public void flush(int maxDoc, final Sorter.DocMap sortMap) throws IOException {
+        // simply write data in the flat file
+        flatVectorsWriter.flush(maxDoc, sortMap);
+        // TODO: add code for creating Vector datastructures during lucene flush operation
```
jmazanec15, line 65: Oh so this doesnt actually create the native indices, right?

navneet1v: No its not creating any native index. Right now this is just creating flat vectors file.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
navneet1v: its just my preference, I feel we will implement it later on hence keeping the doc.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
navneet1v: Yeah that can be done. But since I have not integrated the HNSW graphs creation and eviction here. Will do it once we have that integration

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsFormat.java
navneet1v: it can be changed, doing it.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriter.java
navneet1v: Once the 2.16 is released I will start working on those changes. Keeping the PRs small to ensure that we can review the code in detail.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsFormat.java
```diff
@@ -0,0 +1,68 @@
...
+     * @param state {@link SegmentReadState}
+     */
+    @Override
+    public KnnVectorsReader fieldsReader(final SegmentReadState state) throws IOException {
+        return new NativeEngines990KnnVectorsReader(state, flatVectorsFormat.fieldsReader(state));
+    }
+
+    @Override
+    public String toString() {
+        return "NativeEngines99KnnVectorsFormat(name=NativeEngines99KnnVectorsFormat, flatVectorsFormat=" + flatVectorsFormat + ")";
```
shatejas, line 66: nit: This might print the fully qualified name for flatVectorFormat. maybe change to `.getClass().getSimpleName`

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
```diff
@@ -0,0 +1,162 @@
...
+     *
+     * @param field        the vector field to search
+     * @param target       the vector-valued query
+     * @param knnCollector a KnnResults collector and relevant settings for gathering vector results
+     * @param acceptDocs   {@link Bits} that represents the allowed documents to match, or {@code null}
+     *                     if they are all allowed to match.
+     */
+    @Override
+    public void search(String field, float[] target, KnnCollector knnCollector, Bits acceptDocs) throws IOException {
+        throw new UnsupportedOperationException("Search functionality using codec is not supported with Native Engine Reader");
```
shatejas, line 104: Is this temporary or are we planning to implement it?

With plans to move calling aproximate search to `rewrite`, it might make sense to move the query calls to jni layer in this method. We might have to have more information like engine which we might be able to get from field. This seems to be the right place theoretically. Might be worth looking into

navneet1v: Atleast, my plans were not to move the query here. I mentioned the same in the [RFC](https://github.com/opensearch-project/k-NN/issues/1853) too. reason being older indices using binarydocvalues doesn't support this interface in codec and also in the LeafReader.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsFormat.java
navneet1v: done

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
navneet1v: > With plans to move calling aproximate search to rewrite, it might make sense to move the query calls to jni layer in this method. We might have to have more information like engine which we might be able to get from field. This seems to be the right place theoretically. Might be worth looking into

Once we have some proposal we can take a look into this. I have some other plans for `rewrite` once we have some proposal I can provide more details on what I am thinking.

navneet1v:
Not adding the backport label on this PR. I will add the backport label once 2.16 is released, as this change is not intended for 2.16 version of Opensearch.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngineFieldVectorsWriter.java
```diff
@@ -0,0 +1,107 @@
...
+ * lucene does a flush. This class ensures that KNNVectorWriter is free from generics and this class can encapsulate
+ * all the details related to vectors types and docIds.
+ *
+ * @param <T> float[] or byte[]
+ */
+class NativeEngineFieldVectorsWriter<T> extends KnnFieldVectorsWriter<T> {
+    private static final long SHALLOW_SIZE = RamUsageEstimator.shallowSizeOfInstance(NativeEngineFieldVectorsWriter.class);
+    private final FieldInfo fieldInfo;
+    /**
+     * We are using a map here instead of list, because for sampler interface for quantization we have to advance the iterator
```
jmazanec15, line 35: Im a little skeptical on this point. Are the docids going to be ordered as they are added? Does quantization require random access or can it use an iterator?

My main concern is storing the vectors in a hashmap may have suboptimal performance. We can come back to this later, but I think it worth thinking about. An alternative could be storing vectors in an array and then having their array index as value in the map.

navneet1v: > Are the docids going to be ordered as they are added? Does quantization require random access or can it use an iterator?

No they are no necessarily ordered. Consider the case of nested fields, where child docs has the vector field but not the parent doc. So we will have docids 1,2 but 3(which is parent docId) it will not be present.


> My main concern is storing the vectors in a hashmap may have suboptimal performance

We are not hashing the vectors, we are just hashing the docId which is int. The Hashmap will store the reference of the vector. So if we use List<T> or HashMap<Integer,T> the difference is we are hashing the integer, which is key of hashmap.

> An alternative could be storing vectors in an array and then having their array index as value in the map.

Yes this was done earlier, but as I mentioned earlier the docIds may not be in the order, as there can be some ids which are missing.


>  Does quantization require random access or can it use an iterator?

not random access per say but it is more like quantization sample out few docIds(sort them) and then ask vectors for those docIds. So it needs to access vectors with interface like `advance`.

jmazanec15: > No they are no necessarily ordered.

I do think that they will be ordered. See https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/codecs/KnnFieldVectorsWriter.java#L33-L37. 

> We are not hashing the vectors, we are just hashing the docId which is int. The Hashmap will store the reference of the vector. So if we use List or HashMap<Integer,T> the difference is we are hashing the integer, which is key of hashmap.

I think here Im more wondering about position in memory of vectors, not about hashing. If we are copying the vectors it would be good to copy in a way that positions the vectors next to each other. 

Overall, I feel that setting up in a list fashion might give a better access pattern for iterative use case. But I dont know for certain.  I think Lucene does this here: https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99FlatVectorsWriter.java#L432-L445. Also, I see lucene makes a copy. I wonder if this is something we should do as well.

navneet1v: > > No they are no necessarily ordered.
> 
> I do think that they will be ordered. See https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/codecs/KnnFieldVectorsWriter.java#L33-L37.

I added an example for the clarification. We will get the docIds ordered but they will not be contiguous, this is the case when we have nested fields. I added the example too. 

> If we are copying the vectors it would be good to copy in a way that positions the vectors next to each other.

Atleast in the NativeEngineFieldWriter we are not doing the copy and we are just storing the reference of the vector. 

> Overall, I feel that setting up in a list fashion might give a better access pattern for iterative use case

As the vectors are present in heap, I don't see how this will give better access pattern. If this would have been disk then yes may be. 


> Also, I see lucene makes a copy. I wonder if this is something we should do as well.

If you check here they are not making a copy. So this let me believe that copy is not required.
https://github.com/apache/lucene/blob/branch_9_11/lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsWriter.java#L605-L617

navneet1v: Overall what I can see 2 options here to resolve this:
1. I can go ahead move from Map<Intger,T> to List<T>. and then when I implement the VectorValues I can come back and show that why we need to make this as a map here. 
2. I can keep using Map<Intger,T> here.

I am okay with both, because if don't make this as a Map here then iteration and advance(docId) logic will be complex. Keeping this as a Map makes logic clean without any big downside.

[POC to show how map was getting used](https://github.com/opensearch-project/k-NN/blob/feature/disk-based-knn/src/main/java/org/opensearch/knn/index/vectorvalues/KNNVectorValuesIterator.java#L60-L86) and its ease of use.

jmazanec15: Sure, Im fine just going with the map initially. I think array list iteration is in general faster than map, but maybe negligibly so.

jmazanec15 (approved):

ryanbogan (approved):

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriter.java
```diff
@@ -0,0 +1,115 @@
...
+    private boolean finished;
+
+    /**
+     * Add new field for indexing.
+     * In Lucene, we use single file for all the vector fields so here we need to see how we are going to make things
+     * work.
+     * @param fieldInfo {@link FieldInfo}
+     */
+    @Override
+    public KnnFieldVectorsWriter<?> addField(final FieldInfo fieldInfo) throws IOException {
```
heemin32, line 47: Can't we avoid using `?`. It is hard to know what the `?` can be and what we can do with `?`.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/UnitTestCodec.java
```diff
@@ -0,0 +1,37 @@
...
+
+package org.opensearch.knn.index.codec.KNN990Codec;
+
+import org.apache.lucene.codecs.FilterCodec;
+import org.apache.lucene.codecs.KnnVectorsFormat;
+import org.apache.lucene.codecs.perfield.PerFieldKnnVectorsFormat;
+import org.opensearch.knn.index.codec.KNNCodecVersion;
+
+/**
+ * This codec is for testing. The reason for putting this codec here is SPI is only scanning the src folder and not
```
heemin32, line 20: What is SPI?

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/UnitTestCodec.java
```diff
@@ -0,0 +1,37 @@
+/*
```
heemin32, line 1: I think we are using this simpler form of license header.

```
/*
 */
```

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngineFieldVectorsWriter.java
```diff
@@ -0,0 +1,107 @@
...
+        switch (fieldInfo.getVectorEncoding()) {
+            case FLOAT32:
+                return new NativeEngineFieldVectorsWriter<float[]>(fieldInfo, infoStream);
+            case BYTE:
+                return new NativeEngineFieldVectorsWriter<byte[]>(fieldInfo, infoStream);
+        }
+        throw new IllegalStateException("Unsupported Vector encoding : " + fieldInfo.getVectorEncoding());
+    }
+
+    NativeEngineFieldVectorsWriter(final FieldInfo fieldInfo, final InfoStream infoStream) {
```
heemin32, line 57: I think this should be private so that we can enforce the use of `create` method which enforce the restriction on the class?

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngineFieldVectorsWriter.java
```diff
@@ -0,0 +1,107 @@
...
+import java.util.Map;
+
+/**
+ * NativeEngineVectorFieldsWriter is a class that will be used to accumulate all the vectors during ingestion before
+ * lucene does a flush. This class ensures that KNNVectorWriter is free from generics and this class can encapsulate
+ * all the details related to vectors types and docIds.
+ *
+ * @param <T> float[] or byte[]
+ */
+class NativeEngineFieldVectorsWriter<T> extends KnnFieldVectorsWriter<T> {
```
heemin32, line 31: I was thinking if it would be better to have KNNVector class and make class look like 
```
class NativeEngineFieldVectorsWriter<KNNVector> extends KnnFieldVectorsWriter<KNNVector> {
```

```
public interface KNNVector{
  // common methods
}
public class KNNFloatVector implement KNNVector{
  private float[];
}
public class KNNByteVector implement KNNVector{
  private byte[];
}
```

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/UnitTestCodec.java
```diff
@@ -0,0 +1,37 @@
...
+import org.apache.lucene.codecs.FilterCodec;
+import org.apache.lucene.codecs.KnnVectorsFormat;
+import org.apache.lucene.codecs.perfield.PerFieldKnnVectorsFormat;
+import org.opensearch.knn.index.codec.KNNCodecVersion;
+
+/**
+ * This codec is for testing. The reason for putting this codec here is SPI is only scanning the src folder and not
+ * able to pick this class if its in test folder. Don't use this codec outside of testing
+ */
+public class UnitTestCodec extends FilterCodec {
```
heemin32, line 23: Why not using mock instead of having this class in source folder?

navneet1v: We cannot use Mock, as we are using indexwriter.add docs functionality to test the codec. This has been a practice in Lucene where they build a light weight codec to test the code which is better than doing mock.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngineFieldVectorsWriter.java
navneet1v: not able to understand the benefit of this over float[]. Having another object on top of floats means more objects in heap when this class is not providing any other abstraction.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/UnitTestCodec.java
navneet1v: I don't think so the header is updated in the repo ref: https://github.com/opensearch-project/k-NN/blob/main/.idea/copyright/SPDX_ALv2.xml. otherwise it would have come automatically from IDE. Let me check my IDE again.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/UnitTestCodec.java
navneet1v: SPI is service provider interface. This is used to create the object of a class by using name of the classes. This is how Lucene creates the Codecs without actually calling new on those classes.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriter.java
navneet1v: No we cannot avoid. If I avoid it and lets say make it just KnnFieldVectorsWriter, then warning comes up saying Raw usage of Parametrized class KnnFieldVectorsWriter. Hence we have to use `?`

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngineFieldVectorsWriter.java
navneet1v: We can make it private but then in all UTs I have to use create method. But I think UTs can be changed. I will add in the followup PR.

src/test/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsFormatTests.java
```diff
@@ -0,0 +1,209 @@
...
+        // do it at the end so that all search is completed
+        indexReader.close();
+    }
+
+    private List<String> getFilesFromSegment(Directory dir, String fileFormat) throws IOException {
+        return Arrays.stream(dir.listAll()).filter(x -> x.contains(fileFormat)).collect(Collectors.toList());
+    }
+
+    /**
+     * This should have been annotated with @Before, but somehow when I annotate with @Before apart from running
```
heemin32, line 161: Isn't it because you have two tests in this class? Then, this will be called twice for each test.
If it runs more than that, it could be due to method name is being conflicted where it is being called by parent class?

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/UnitTestCodec.java
heemin32: @martin-gaievski ?

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriter.java
heemin32: That is one place where wrapping might be good over primitive array.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngineFieldVectorsWriter.java
heemin32: 1. It can enforce the interface instead of opening up for any object for `NativeEngineFieldVectorsWriter` Currently, we are enforcing it through constructor which might not intuitive.
2. More flexibility. For example, if we need some other method specific to float or byte, current architecture might not support it smoothly but we will have to add if/else inside `NativeEngineFieldVectorsWriter`.
3. For the return value, we don't know what the T will be so we used `?`. The caller have to validate its instance type before handling it.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/UnitTestCodec.java
navneet1v: Also if we use mock then when the segments are read lucene will create object of the codec then since spi is not present for codec the code will fail.

src/test/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsFormatTests.java
navneet1v: Could be. the problem was when I run it there is some test called as classMethod which was getting called and I was not able to figure that at the time of raising the PR. I will debug this more but as it is in test this was not a big issue. Hence I kept it. But I understand your concern let me try to give it 1 more shot to see if I can fix the problem.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriter.java
navneet1v: conversation continued here: https://github.com/opensearch-project/k-NN/pull/1855#discussion_r1692070831

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngineFieldVectorsWriter.java
navneet1v: Resolved in this PR: https://github.com/opensearch-project/k-NN/pull/1897/files

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngineFieldVectorsWriter.java
navneet1v: As discussed, having a base interface like KNNVector just over native types creates an overhead. I did the following change and found out that there is ~60byte overhead that we will get per vector if we add any extra abstraction. You can use the `ramBytesUsed` function to get these details.

```
/*
 *
 * The OpenSearch Contributors require contributions made to
 * this file be licensed under the Apache-2.0 license or a
 * compatible open source license.
 *
 * Modifications Copyright OpenSearch Contributors. See
 * GitHub history for details.
 */

package org.opensearch.knn.common;

import lombok.EqualsAndHashCode;
import lombok.Getter;
import lombok.ToString;
import org.apache.lucene.util.RamUsageEstimator;

@Getter
@EqualsAndHashCode
@ToString
abstract class KNNVector {
    protected int dimension;

    public long ramBytesUsed() {
        return RamUsageEstimator.sizeOf(dimension);
    }
}


/*
 *
 * The OpenSearch Contributors require contributions made to
 * this file be licensed under the Apache-2.0 license or a
 * compatible open source license.
 *
 * Modifications Copyright OpenSearch Contributors. See
 * GitHub history for details.
 */

package org.opensearch.knn.common;

import lombok.EqualsAndHashCode;
import lombok.Getter;
import lombok.ToString;

@Getter
@EqualsAndHashCode(callSuper = true)
@ToString(callSuper = true)
class KNNFloatVector extends KNNVector {
    private final float[] vector;

    public KNNFloatVector(float[] vector) {
        if (vector == null || vector.length == 0) {
            throw new IllegalStateException("The input vector for KNNFloatVector Object cannot be null");
        }
        this.dimension = vector.length;
        this.vector = vector;
    }
}


/*
 *
 * The OpenSearch Contributors require contributions made to
 * this file be licensed under the Apache-2.0 license or a
 * compatible open source license.
 *
 * Modifications Copyright OpenSearch Contributors. See
 * GitHub history for details.
 */

package org.opensearch.knn.common;

import lombok.EqualsAndHashCode;
import lombok.Getter;
import org.apache.lucene.util.RamUsageEstimator;

@Getter
@EqualsAndHashCode(callSuper = true)
class KNNByteVector extends KNNVector {
    private static final long SHALLOW_SIZE = RamUsageEstimator.shallowSizeOfInstance(KNNByteVector.class);

    private final byte[] vector;

    public KNNByteVector(byte[] vector) {
        if (vector == null || vector.length == 0) {
            throw new IllegalStateException("The input vector for KNNByteVector Object cannot be null");
        }
        this.dimension = vector.length;
        this.vector = vector;
    }

    @Override
    public long ramBytesUsed() {
        return SHALLOW_SIZE + ramBytesWithoutInstanceSize();
    }

    protected long ramBytesWithoutInstanceSize() {
        return super.ramBytesUsed() + RamUsageEstimator.shallowSizeOf(vector);
    }
}


/*
 *
 * The OpenSearch Contributors require contributions made to
 * this file be licensed under the Apache-2.0 license or a
 * compatible open source license.
 *
 * Modifications Copyright OpenSearch Contributors. See
 * GitHub history for details.
 */

package org.opensearch.knn.common;

import lombok.EqualsAndHashCode;
import lombok.ToString;
import org.apache.lucene.util.RamUsageEstimator;

@EqualsAndHashCode(callSuper = true)
@ToString(callSuper = true)
class KNNBinaryVector extends KNNByteVector {
    private static final long SHALLOW_SIZE = RamUsageEstimator.shallowSizeOfInstance(KNNBinaryVector.class);

    public KNNBinaryVector(final byte[] vector) {
        super(vector);
        this.dimension = getDimension() * Byte.SIZE;
    }

    @Override
    public long ramBytesUsed() {
        long s = SHALLOW_SIZE;
        return super.ramBytesWithoutInstanceSize() + SHALLOW_SIZE;
    }
}

```

Another thing is NativeEngineFieldWriter generic is limited to codec and not exposed outside. Plus for iterating over vectors I have added concrete classes in this PR: https://github.com/opensearch-project/k-NN/pull/1897/files. This should resolve all the issues.

Based on this I am closing the issue. If we deicide later that we need to add a low level interface will add it.

Issue #1286, opened 2023-10-25
[BUG] When submitting a KNN search with a filter that refers to a non existing field, a non-descriptive `failed to create query: Rewrite first` error is returned


de-code:
**Describe the bug**

When submitting a KNN search with a filter that refers to a non existing field, then I am getting a non-descriptive `failed to create query: Rewrite first`

**To Reproduce**

1. Create Index with KNN

```text
PUT /test-index-knn1
{
  "settings": {
    "index": {
      "knn": true
    }
  },
  "mappings": {
    "properties": {
      "vector": {
        "type": "knn_vector",
        "dimension": 3,
        "method": {
          "name": "hnsw",
          "space_type": "l2",
          "engine": "lucene"
        }
      },
      "field1": {
        "type": "integer"
      }
    }
  }
}
```

2. Query with non-existing field:

```text
POST test-index-knn1/_search
{
  "query": {
    "knn": {
      "vector": {
        "vector": [0, 0, 0],
        "k": 10,
        "filter": {
          "bool": {
            "must": [
              {"range": {"non-existing-field1": {"gte": 1}}}
            ]
          }
        }
      }
    }
  }
}
```

3. This results in `failed to create query: Rewrite first` error

```json
{
  "error": {
    "root_cause": [
      {
        "type": "query_shard_exception",
        "reason": "failed to create query: Rewrite first",
        "index": "test-index-knn1",
        "index_uuid": "E5XiPqWQTuiQBsKTz2bEqw"
      }
    ],
    "type": "search_phase_execution_exception",
    "reason": "all shards failed",
    "phase": "query",
    "grouped": true,
    "failed_shards": [
      {
        "shard": 0,
        "index": "test-index-knn1",
        "node": "UOMb6bKjR1qZRdNKuM2ueQ",
        "reason": {
          "type": "query_shard_exception",
          "reason": "failed to create query: Rewrite first",
          "index": "test-index-knn1",
          "index_uuid": "E5XiPqWQTuiQBsKTz2bEqw",
          "caused_by": {
            "type": "illegal_state_exception",
            "reason": "Rewrite first"
          }
        }
      }
    ]
  },
  "status": 400
}
```

**Expected behavior**

Correct behaviour could be one of:
* The error message should point out what field name doesn't exist . Similar to when specifying a non existing field as the vector.
* No error message, similar to a non-KNN query

e.g. the following query:

```text
POST test-index-knn1/_search
{
  "query": {
    "knn": {
[...]

PR #1874, opened 2024-07-24
Corrected search logic for scenario with non-existent fields in filter

martin-gaievski:
### Description
Changing behavior of the k-NN query with filters. Changes related to following areas:
- false negative scenario when query return an error instead of legit results
- more user friendly response in case filter has only one clause with non-existent field, return 0 total hits instead of 400 error

Today in scenario when filter refers to filed that does not exist query return 400 error and exception in the response body. 

```
{
  "error": {
    "root_cause": [
      {
        "type": "query_shard_exception",
        "reason": "failed to create query: Rewrite first",
        "index": "test-index-knn1",
        "index_uuid": "E5XiPqWQTuiQBsKTz2bEqw"
      }
    ],
    "type": "search_phase_execution_exception",
    "reason": "all shards failed",
    "phase": "query",
    "grouped": true,
    "failed_shards": [
      {
        "shard": 0,
        "index": "test-index-knn1",
        "node": "UOMb6bKjR1qZRdNKuM2ueQ",
        "reason": {
          "type": "query_shard_exception",
          "reason": "failed to create query: Rewrite first",
          "index": "test-index-knn1",
          "index_uuid": "E5XiPqWQTuiQBsKTz2bEqw",
          "caused_by": {
            "type": "illegal_state_exception",
            "reason": "Rewrite first"
          }
        }
      }
    ]
  },
  "status": 400
}
```

In the same scenario other queries have more complex logic. Filter is rewritten earlier in the execution and in it's simplest form  passed to the low level code. 

In following example the query will return empty response:

```
{
    "query": {
     ...
                "filter": {
                    "bool": {
                        "should": [
                            {
                                "range": {
                                    "non-existing-field1": {
                                        "gte": 1
```

one more example where query can return some results because non existent field is part of the optional clause:

```
{
    "query": {
     .....
                "filter": {
                    "bool": {
                        "must": [
                            {
                                "range": {
                                    "existing-field1": {
                                        "gte": 1
                                    }
                                }
                            }
                        ],
                        "should": [
                            {
                                "range": {
                                    "non-existing-field1": {
                                        "gte": 1
                                    }
``` 

The approach taken in this PR is based on usage on rewrite query functionality. This allows to avoid runtime exceptions as they are handled gracefully by rewrite logic.

### Related Issues
Resolves https://github.com/opensearch-project/k-NN/issues/1286

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
+++ b/src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
@@ -24,6 +24,7 @@ import org.opensearch.index.mapper.MappedFieldType;
 import org.opensearch.index.mapper.NumberFieldMapper;
 import org.opensearch.index.query.AbstractQueryBuilder;
 import org.opensearch.index.query.QueryBuilder;
+import org.opensearch.index.query.QueryRewriteContext;
 import org.opensearch.index.query.QueryShardContext;
 import org.opensearch.knn.common.KNNConstants;
 import org.opensearch.knn.index.IndexUtil;
@@ -485,7 +486,7 @@ public class KNNQueryBuilder extends AbstractQueryBuilder<KNNQueryBuilder> {
     }
 
     @Override
-    protected Query doToQuery(QueryShardContext context) throws IOException {
+    protected Query doToQuery(QueryShardContext context) {
         MappedFieldType mappedFieldType = context.fieldMapper(this.fieldName);
 
         if (mappedFieldType == null && ignoreUnmapped) {
@@ -600,11 +601,6 @@ public class KNNQueryBuilder extends AbstractQueryBuilder<KNNQueryBuilder> {
             throw new IllegalArgumentException(String.format(Locale.ROOT, "Engine [%s] does not support filters", knnEngine));
         }
 
-        // rewrite filter query if it exists to avoid runtime errors in next steps of query phase
-        if (Objects.nonNull(filter)) {
-            filter = filter.rewrite(context);
-        }
-
         String indexName = context.index().getName();
 
         if (k != 0) {
@@ -714,5 +710,14 @@ public class KNNQueryBuilder extends AbstractQueryBuilder<KNNQueryBuilder> {
     @Override
     public String getWriteableName() {
         return NAME;
+    }
+
+    @Override
+    protected QueryBuilder doRewrite(QueryRewriteContext queryShardContext) throws IOException {
+        // rewrite filter query if it exists to avoid runtime errors in next steps of query phase
+        if (Objects.nonNull(filter)) {
+            filter = filter.rewrite(queryShardContext);
+        }
+        return super.doRewrite(queryShardContext);
     }
 }
--- a/src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
+++ b/src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
@@ -485,7 +485,7 @@ public class KNNQueryBuilder extends AbstractQueryBuilder<KNNQueryBuilder> {
     }
 
     @Override
-    protected Query doToQuery(QueryShardContext context) {
+    protected Query doToQuery(QueryShardContext context) throws IOException {
         MappedFieldType mappedFieldType = context.fieldMapper(this.fieldName);
 
         if (mappedFieldType == null && ignoreUnmapped) {
@@ -598,6 +598,11 @@ public class KNNQueryBuilder extends AbstractQueryBuilder<KNNQueryBuilder> {
             && filter != null
             && !KNNEngine.getEnginesThatSupportsFilters().contains(knnEngine)) {
             throw new IllegalArgumentException(String.format(Locale.ROOT, "Engine [%s] does not support filters", knnEngine));
+        }
+
+        // rewrite filter query if it exists to avoid runtime errors in next steps of query phase
+        if (Objects.nonNull(filter)) {
+            filter = filter.rewrite(context);
         }
 
         String indexName = context.index().getName();
--- a/src/test/java/org/opensearch/knn/index/LuceneEngineIT.java
+++ b/src/test/java/org/opensearch/knn/index/LuceneEngineIT.java
@@ -17,6 +17,7 @@ import org.opensearch.client.ResponseException;
 import org.opensearch.common.Nullable;
 import org.opensearch.core.xcontent.XContentBuilder;
 import org.opensearch.common.xcontent.XContentFactory;
+import org.opensearch.index.query.QueryBuilder;
 import org.opensearch.index.query.QueryBuilders;
 import org.opensearch.knn.KNNRestTestCase;
 import org.opensearch.knn.KNNResult;
@@ -86,6 +87,9 @@ public class LuceneEngineIT extends KNNRestTestCase {
     private static final String KNN_VECTOR_TYPE = "knn_vector";
     private static final String PROPERTIES_FIELD_NAME = "properties";
     private static final String TYPE_FIELD_NAME = "type";
+    private static final String INTEGER_FIELD_NAME = "int_field";
+    private static final String FILED_TYPE_INTEGER = "integer";
+    private static final String NON_EXISTENT_INTEGER_FIELD_NAME = "nonexistent_int_field";
 
     @After
     public final void cleanUp() throws IOException {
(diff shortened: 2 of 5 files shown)


ryanbogan (dismissed):

heemin32 (dismissed):

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
```diff
@@ -600,6 +600,11 @@ protected Query doToQuery(QueryShardContext context) {
             throw new IllegalArgumentException(String.format(Locale.ROOT, "Engine [%s] does not support filters", knnEngine));
         }
 
+        // rewrite filter query if it exists to avoid runtime errors in next steps of query phase
+        if (Objects.nonNull(filter)) {
```
shatejas, line 604: Should we be rewriting the filter query in doRewrite method of querybuilder? A caveat of that is we have to replace the filter with rewritten one, which we are doing here as well. Just asking because there is dedicated method to rewrite querybuillders

src/test/java/org/opensearch/knn/index/query/KNNQueryBuilderTests.java
```diff
@@ -485,6 +486,7 @@ public void testDoToQuery_Normal() throws Exception {
         assertEquals(knnQueryBuilder.vector(), query.getQueryVector());
     }
 
+    @SneakyThrows
```
shatejas, line 489: Seems like we skipped adding unit test, any particular reason? if not lets add it

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
navneet1v: > Should we be rewriting the filter query in doRewrite method of querybuilder?

+1 on this. I am also little surprised why we are doing the rewrite in doToQuery function.

src/test/java/org/opensearch/knn/index/query/KNNQueryBuilderTests.java
navneet1v: +1

navneet1v (changes requested):
@martin-gaievski putting this PR as request changes, till we resolve comments from @shatejas .

src/test/java/org/opensearch/knn/index/query/KNNQueryBuilderTests.java
martin-gaievski: we cannot test real behavior of the rewrite for the query in the unit test, only in integ tests. We already have unit tests for general logic for cases when filter is null and not null.

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
martin-gaievski: other queries do rewrite as part of `doToQuery`, it's even part of the abstract [query builder class](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/query/AbstractQueryBuilder.java#L250). I don't have strong preference, just we already have override for `doToQuery`, but not for `doRewrite`. 
Let me check if there are any caveats, if not I can move the logic to `doRewrite`

navneet1v: great.

shatejas: > it's even part of the abstract [query builder class](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/query/AbstractQueryBuilder.java#L250).

This seems to be the actual execution of OS, where `.rewrite` calls doRewrite of underlying querybuilder and `toQuery` will call doToQuery. They are chained but called one after the other

martin-gaievski: yup, I checked, override in `rewrite` does the job, pushed a commit that implements the change

src/test/java/org/opensearch/knn/index/query/KNNQueryBuilderTests.java
martin-gaievski: added unit tests that follow this logic, just using mocked context objects

src/test/java/org/opensearch/knn/index/LuceneEngineIT.java
```diff
@@ -278,6 +282,77 @@ public void testQueryWithFilterUsingByteVectorDataType() {
...
+        List<KNNResult> resultsQuery1 = parseSearchResponse(
+            EntityUtils.toString(searchWithRequiredNonExistentFiledInFilterResponse.getEntity()),
+            FIELD_NAME
+        );
+        assertTrue(resultsQuery1.isEmpty());
+
+        // use filter with non existent field as optional, we should have some results
+        QueryBuilder filterWithOptionalNonExistentField = QueryBuilders.boolQuery()
+            .should(QueryBuilders.rangeQuery(NON_EXISTENT_INTEGER_FIELD_NAME).gte(1))
+            .must(QueryBuilders.rangeQuery(INTEGER_FIELD_NAME).gte(1));
```
navneet1v, line 343: why we are getting a successful result if the field doesn't exist? am I missing something here

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
```diff
@@ -710,4 +711,13 @@ protected int doHashCode() {
     public String getWriteableName() {
         return NAME;
     }
+
+    @Override
+    protected QueryBuilder doRewrite(QueryRewriteContext queryShardContext) throws IOException {
+        // rewrite filter query if it exists to avoid runtime errors in next steps of query phase
+        if (Objects.nonNull(filter)) {
+            filter = filter.rewrite(queryShardContext);
```
navneet1v, line 719: can it happen that after filter.rewrite is done the filter query is null?

navneet1v:
> This allows to avoid runtime exceptions as they are handled gracefully by rewrite logic.

@martin-gaievski can you elaborate more on this? When you say non existent field is handled gracefully you mean it will throw exception? or nothing will happen.

navneet1v:
> > This allows to avoid runtime exceptions as they are handled gracefully by rewrite logic.
> 
> @martin-gaievski can you elaborate more on this? When you say non existent field is handled gracefully you mean it will throw exception? or nothing will happen.

NVM I check the behavior, so atleast in my simple testing on an index if the field is not present on the index then it gives no results. So if a filter has non existent fields then filter will become null right?

@martin-gaievski

martin-gaievski:
> > > This allows to avoid runtime exceptions as they are handled gracefully by rewrite logic.
> > 
> > 
> > @martin-gaievski can you elaborate more on this? When you say non existent field is handled gracefully you mean it will throw exception? or nothing will happen.
> 
> NVM I check the behavior, so atleast in my simple testing if the field is not present on the index then it gives no results. So if a filter has non existent fields then filter will become null right?
> 
> @martin-gaievski

in a simple form when there is only one such field - yes, filter is null. If filter is more sophisticated like bool with two shoulds, where one has non existent field and other clause is legit then filter will not be null.

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
martin-gaievski: yes, it's possible for a simple filters with invalid clauses (like initial reported case with non existing field)

src/test/java/org/opensearch/knn/index/query/KNNQueryBuilderTests.java
```diff
@@ -1293,4 +1309,19 @@ public void testDoToQuery_whenBinaryWithInvalidDimension_thenException() throws
...
+        assertEquals(knnQueryBuilder, rewritten);
+    }
+
+    @SneakyThrows
+    public void testDoRewrite_whenFilterSet_thenSuccessful() {
+        QueryBuilder filter = QueryBuilders.termQuery(TEXT_FIELD_NAME, TEXT_VALUE);
+        KNNQueryBuilder knnQueryBuilder = new KNNQueryBuilder(FIELD_NAME, QUERY_VECTOR, K, filter);
+        QueryBuilder rewritten = knnQueryBuilder.rewrite(mock(QueryRewriteContext.class));
+        assertEquals(knnQueryBuilder, rewritten);
+    }
```
shatejas, line 1326:
```suggestion
    public void testDoRewrite_whenFilterSet_thenSuccessful() {
        //Given
        QueryBuilder filter = mock(QueryBuilder.class);
        QueryBuilder rewrittenFilter = mock(QueryBuilder.class);
        QueryRewriteContext context = mock(QueryRewriteContext.class);
        when(filter.rewrite(context)).thenReturn(rewrittenFilter);
        KNNQueryBuilder expected = KNNQueryBuilder.builder()
                .fieldName(FIELD_NAME)
                .vector(QUERY_VECTOR)
                .filter(rewrittenFilter)
                .k(K)
                .build();



        //When
        KNNQueryBuilder knnQueryBuilder = KNNQueryBuilder.builder()
                .fieldName(FIELD_NAME)
                .vector(QUERY_VECTOR)
                .filter(filter)
                .k(K)
                .build();
        
        QueryBuilder actual = knnQueryBuilder.rewrite(context);
        
        //Then
        assertEquals(expected, actual);
    }
```

shatejas: The previous still has room to make it tighter. 

Note: please make sure you add imports if you accept the suggestion

navneet1v:
> > > > This allows to avoid runtime exceptions as they are handled gracefully by rewrite logic.
> > > 
> > > 
> > > @martin-gaievski can you elaborate more on this? When you say non existent field is handled gracefully you mean it will throw exception? or nothing will happen.
> > 
> > 
> > NVM I check the behavior, so atleast in my simple testing if the field is not present on the index then it gives no results. So if a filter has non existent fields then filter will become null right?
> > @martin-gaievski
> 
> in a simple form when there is only one such field - yes, filter is null. If filter is more sophisticated like bool with two shoulds, where one has non existent field and other clause is legit then filter will not be null.

I have 1 more question on this behavior, lets say in the should clause if both the fields are non existent then what will happen to the bool query. Can we please check it? In my head I am not able to think what will happen in such a case.

src/test/java/org/opensearch/knn/index/LuceneEngineIT.java
navneet1v: This is answered in other comments. Resolving the comment.

src/test/java/org/opensearch/knn/index/query/KNNQueryBuilderTests.java
martin-gaievski: ack

martin-gaievski:
> > > > > This allows to avoid runtime exceptions as they are handled gracefully by rewrite logic.
> > > > 
> > > > 
> > > > @martin-gaievski can you elaborate more on this? When you say non existent field is handled gracefully you mean it will throw exception? or nothing will happen.
> > > 
> > > 
> > > NVM I check the behavior, so atleast in my simple testing if the field is not present on the index then it gives no results. So if a filter has non existent fields then filter will become null right?
> > > @martin-gaievski
> > 
> > 
> > in a simple form when there is only one such field - yes, filter is null. If filter is more sophisticated like bool with two shoulds, where one has non existent field and other clause is legit then filter will not be null.
> 
> I have 1 more question on this behavior, lets say in the should clause if both the fields are non existent then what will happen to the bool query. Can we please check it? In my head I am not able to think what will happen in such a case.

in this case same logic applies to all bool-should clauses. If we used  non existent fields for both clauses then for the filter overall there will be no docs that match such criteria and final the result will be empty. 

checked this on real data, below are results:

request: 

```
{
    "query": {
        "knn": {
            "vector": {
                "vector": [
                    0,
                    0,
                    0
                ],
                "k": 10,
                "filter": {
                    "bool": {
                        "should": [
                            {
                                "range": {
                                    "non-existing-field1": {
                                        "gte": 1
                                    }
                                }
                            },
                            {
                                "range": {
                                    "non-existing-field2": {
                                        "gte": 1
                                    }
                                }
                            }
                        ]
                    }
                }
            }
        }
    }
}
```

response:
```
{
    "took": 5,
    "timed_out": false,
    "_shards": {
        "total": 1,
        "successful": 1,
        "skipped": 0,
        "failed": 0
    },
    "hits": {
        "total": {
            "value": 0,
            "relation": "eq"
        },
        "max_score": null,
        "hits": []
    }
}
```

shatejas (approved):
Thanks for making the tests tighter!

navneet1v:
@martin-gaievski based on the real request can you just tell me 1 more thing: so for these 2 non-existent field filter.rewrite gave a null response or not. Its is more from my understanding standpoint. other than that code looks good to me.

navneet1v (approved):

martin-gaievski:
> @martin-gaievski based on the real request can you just tell me 1 more thing: so for these 2 non-existent field filter.rewrite gave a null response or not. Its is more from my understanding standpoint. other than that code looks good to me.

in this scenario filter object gonna be `match_none` query

navneet1v:
> in this scenario filter object gonna be `match_none` query

Thanks, this is what I was exactly looking for. Already approved the code.

heemin32 (approved):

PR #1876, opened 2024-07-24
Adding integration tests for binary index

heemin32:
### Description
Adding integration tests for binary index with following scenarios

* Can create binary index using Faiss HNSW
* Can create binary index with knn index as false
* Cannot create binary index using Lucene
* Cannot create binary index using Nmslib
* Cannot create binary index using encoding(sq)
* Cannot run range query on binary index
* Can run knn query on binary index
* Can run knn query with nested field on binary index
* Can run knn query with filtering on binary index
* Can run knn query with filtering on nested field with binary index
* Can run script scoring with hamming on binary index
* Cannot run script scoring with hamming on non binary index
* Cannot run script scoring with non hamming on binary index
* Can run painless script with hamming on binary index
* Cannot run painless script with hamming on non binary index
* Cannot run painless script with non hamming on binary index
* Can run script scoring with hamming on binary index with knn set as false
* Can run painless script with hamming on binary index with knn set as false

### Notes
* Also moved some of existing integ tests files under the integ package. I think we should move all integration test files under integ package. Currently, they are scattered across all packages which makes it hard to find correct class to add a new test.
* Moved radius search validation before translating max distance and min score value. Otherwise, it fails during score translation with hamming distance type and the error message is not user friendly. 
#### Before
```
Space [hamming] does not have a score to distance translation
```
#### After
```
Binary data type does not support radial search
```

### Related Issues
N/A

### Check List

For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/k-NN/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).* *


src/test/java/org/opensearch/knn/integ/AutoIndexCleanUpTestCase.java
```diff
@@ -0,0 +1,53 @@
...
+
+import java.io.IOException;
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * This class provides the convenience of automatically deleting indices created during tests once they are completed
+ */
+@Log4j2
+public class AutoIndexCleanUpTestCase extends KNNRestTestCase {
```
shatejas, line 21: Why make this a test case? Can you create a utility class and plug it in the required test. Deep hierarchies are not good for code reuse, this is adding on to already existing hierarchies. 

Cleanup can be different for each test so whatever tests require it can call the utility by passing in indices

src/test/java/org/opensearch/knn/integ/BinaryIndexIT.java
```diff
@@ -0,0 +1,242 @@
...
+        Byte[] vector4 = { 0b00001111, 0b00000001 };
+        addKnnDoc(indexName, "1", fieldName, vector1);
+        addKnnDoc(indexName, "2", fieldName, vector2);
+        addKnnDoc(indexName, "3", fieldName, vector3);
+        addKnnDoc(indexName, "4", fieldName, vector4);
+
+        // Query
+        float[] queryVector = { (byte) 0b10001111, (byte) 0b10000000 };
+        int k = 4;
+        Response searchResponse = searchKNNIndex(indexName, new KNNQueryBuilder(fieldName, queryVector, k), k);
```
shatejas, line 87: Use XcontentBuilder here as well?

src/test/java/org/opensearch/knn/integ/BinaryIndexIT.java
```diff
@@ -0,0 +1,242 @@
...
+
+    @SneakyThrows
+    public void testFaissHnswBinary_whenKnnIndexIsTrue_thenCreateIngestQueryWorks() {
+        // Create Index
+        String indexName = getCurrentMethodName().toLowerCase(Locale.ROOT);
+        String fieldName = "my-vector";
+        createKnnHnswBinaryIndex(true, KNNEngine.FAISS, indexName, fieldName, 16, null);
+
+        // Ingest
+        Byte[] vector1 = { 0b00000001, 0b00000001 };
```
shatejas, line 75: wondering why we did not use the resources that are in setup in setupClass() for this one

src/test/java/org/opensearch/knn/integ/BinaryIndexIT.java
```diff
@@ -0,0 +1,242 @@
...
+        String fieldName = "my-vector";
+        Exception e = expectThrows(
+            Exception.class,
+            () -> createKnnHnswBinaryIndex(true, KNNEngine.FAISS, indexName, fieldName, 16, ENCODER_SQ)
+        );
+        assertTrue(e.getMessage(), e.getMessage().contains("does not support sq encoder"));
+    }
+
+    @SneakyThrows
+    public void testFaissHnswBinary_whenRadialSearch_thenThrowException() {
```
shatejas, line 153: nit: For all throwException tests, a parametrized test is much easier to maintain 

https://github.com/opensearch-project/k-NN/blob/main/src/test/java/org/opensearch/knn/index/query/InvalidSearchQueryIT.java

src/test/java/org/opensearch/knn/integ/BinaryIndexIT.java
```diff
@@ -0,0 +1,242 @@
...
+        for (int i = 0; i < k; i++) {
+            assertEquals(k - i, Integer.parseInt(results.get(i).getDocId()));
+        }
+    }
+
+    @SneakyThrows
+    public void testFaissHnswBinary_when1000Data_thenRecallIsAboveNinePointZero() {
+        // Create Index
+        String indexName = getCurrentMethodName().toLowerCase(Locale.ROOT);
+        String fieldName = "my-vector";
```
junqiu-lei, line 101: Looks like indexName and fieldName can be shared use for all tests.

src/test/java/org/opensearch/knn/integ/BinaryIndexIT.java
```diff
@@ -0,0 +1,242 @@
...
+            .field(VECTOR, queryVector)
+            .field(MIN_SCORE, minScore)
+            .endObject().endObject().endObject().endObject();
+        Response response = searchKNNIndex(indexName, query, size);
+        return parseSearchResponse(EntityUtils.toString(response.getEntity()), fieldName);
+    }
+
+    private List<KNNResult> runKnnQuery(final String indexName, final String fieldName, final float[] queryVector, final int k)
+        throws Exception {
+        Response response = searchKNNIndex(indexName, KNNQueryBuilder.builder().fieldName(fieldName).vector(queryVector).k(k).build(), k);
```
junqiu-lei, line 190: Better to use `searchKNNIndex(String index, XContentBuilder xContentBuilder, int resultSize)` to mock the request from end

src/test/java/org/opensearch/knn/integ/FilteredSearchBinaryIT.java
```diff
@@ -0,0 +1,82 @@
...
+import org.opensearch.knn.index.VectorDataType;
+import org.opensearch.knn.index.query.KNNQueryBuilder;
+import org.opensearch.knn.index.util.KNNEngine;
+
+import java.util.List;
+import java.util.Locale;
+
+import static org.opensearch.knn.common.KNNConstants.METHOD_HNSW;
+
+public class FilteredSearchBinaryIT extends AutoIndexCleanUpTestCase {
```
junqiu-lei, line 25: I saw there also have a new test class `BinaryIndexIT`, can we merge binary related ITs into one test file?

src/test/java/org/opensearch/knn/integ/BinaryIndexIT.java
heemin32: Why not KNNQueryBuilder?

src/test/java/org/opensearch/knn/integ/FilteredSearchBinaryIT.java
heemin32: I think multiple smaller classes are better than single big class.

src/test/java/org/opensearch/knn/integ/BinaryIndexIT.java
martin-gaievski: seems indexName is specific to the test method, but fieldName can be common

src/test/java/org/opensearch/knn/integ/BinaryIndexIT.java
heemin32: The resource in setup is large(1000) compared to this one.
In this test, I wanted to validate the query result with specific order of doc ids which is not easy with 1000 data.

src/test/java/org/opensearch/knn/integ/BinaryIndexIT.java
heemin32: I didn't want to share the index name between methods but I guess, it will make less noise if I just follow what we have done before. Will update it.

src/test/java/org/opensearch/knn/integ/BinaryIndexIT.java
heemin32: What do you mean with `mock the request from end`?

src/test/java/org/opensearch/knn/integ/BinaryIndexIT.java
shatejas: KNNQueryBuilder is an internal implementation, is that to right to use in integ test? we are creating interdependencies with actual code. 

Ideally we should just be passing a json or use a java client to do so.

src/test/java/org/opensearch/knn/integ/BinaryIndexIT.java
junqiu-lei: I mean using XContentBuilder can better mock request from end user api call than use KNNQueryBuilder

src/test/java/org/opensearch/knn/integ/BinaryIndexIT.java
heemin32: That make sense.

src/test/java/org/opensearch/knn/integ/BinaryIndexIT.java
heemin32: Got it. Fait enough.

src/test/java/org/opensearch/knn/integ/BinaryIndexIT.java
heemin32: By the way, where is this being used? https://github.com/opensearch-project/k-NN/blob/main/src/test/java/org/opensearch/knn/index/query/InvalidSearchQueryIT.java#L37

heemin32: Got it. I missed the method as the method name does not match with what it actually does..

heemin32: I updatedUsed parameterized test. However, now I remember on reason that I don't like about the parameterized test. When one of test fail, there is no easy way to run that specific test to debug without code change.

src/testFixtures/java/org/opensearch/knn/KNNJsonQueryBuilder.java
```diff
@@ -0,0 +1,67 @@
...
+import org.opensearch.core.xcontent.XContentBuilder;
+
+import java.io.IOException;
+
+/**
+ * Helper method to create knn query in json string
+ * Here, we don't use any of the internal class so that it can mimic user request more closely.
+ */
+@Builder
+public class KNNJsonQueryBuilder {
```
shatejas, line 20: Nice!

shatejas (approved):

junqiu-lei (dismissed):

martin-gaievski (dismissed):

junqiu-lei (dismissed):

junqiu-lei (dismissed):

heemin32:
The parameter "explain" setting in search request have made the search request failed for nested field. Removed it which seems unnecessary for all existing test.

martin-gaievski (approved):

junqiu-lei (approved):

martin-gaievski (approved):

Issue #1845, opened 2024-07-17
Investigate migrating from native ANN search at weight.scorer create level to query.rewrite leve


jmazanec15:
## Description

Currently, the ANN search is executed at [KNNWeight.scorer()](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/query/KNNWeight.java#L130) stage. Scorers are created at the per segment level. The scorer returned more or less just iterates over the matches that are returned. 

In Lucene, they do the Ann search the [query rewrite](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java#L70) phase. This will build scorers for each segment, but limit the number of hits for the shard to k. 

I want to investigate feasibility of migrating to ANN search execution at rewrite state in the plugin. Off the top of my head, it has the following benefits:
1. Number of results returned per shard are always k. For the plugin, we will do num_segments*k results for matching. Just returning k results will therefore reduce the search behavior dependency on the number of segments. This would let us also limit the number of re-score operations that would take place as well as part of #1779 
2. Ability to share state between segments during search. Lucene introduced a really cool feature in https://github.com/apache/lucene/pull/12962/files that relies on collectors in order to share best hit information between segments. With this, segment count can be less of a factor for performance. The shared state across segments is implemented via the KnnCollector interface. This may be possible via creating something new in the KNNWeight class, but would require duplicate effort
3. On duplicate effort, moving to follow convention of Lucene to execute search at re-write would give us the ability to more easily migrate fully to the KnnVectorFormat provided by Lucene (see #1087).  

That being said, I still need to figure out what the downsides may be. At the moment, I cannot think of any, but will update here if I find any.

In order to do this migration, I think we would need to
1. Build PoC - see what tests break
2. A/B performance test the PoC
3. Feature compatibility deep dive (concurrent segment search, neural search, hybrid search, etc.)

PR #1877, opened 2024-07-25
Introduces NativeEngineKNNQuery which executes ANN on rewrite

shatejas:
### Description

All integration tests are passing. Cannot support radial currently since k is not significant, there are ways to do it but since its not planned for 2.17 this should work

### Related Issues
Resolves #[1845](https://github.com/opensearch-project/k-NN/issues/1845)

### Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,3 +30,4 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 * Refactor method structure and definitions [#1920](https://github.com/opensearch-project/k-NN/pull/1920)
 * Refactor KNNVectorFieldType from KNNVectorFieldMapper to a separate class for better readability. [#1931](https://github.com/opensearch-project/k-NN/pull/1931)
 * Generalize lib interface to return context objects [#1925](https://github.com/opensearch-project/k-NN/pull/1925)
+* Move k search k-NN query to re-write phase of vector search query for Native Engines [#1877](https://github.com/opensearch-project/k-NN/pull/1877)
--- a/src/main/java/org/opensearch/knn/common/featureflags/KNNFeatureFlags.java
+++ b/src/main/java/org/opensearch/knn/common/featureflags/KNNFeatureFlags.java
@@ -1,0 +1,46 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ */
+
+package org.opensearch.knn.common.featureflags;
+
+import com.google.common.annotations.VisibleForTesting;
+import lombok.experimental.UtilityClass;
+import org.opensearch.common.settings.Setting;
+import org.opensearch.knn.index.KNNSettings;
+
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import static org.opensearch.common.settings.Setting.Property.Dynamic;
+import static org.opensearch.common.settings.Setting.Property.NodeScope;
+
+/**
+ * Class to manage KNN feature flags
+ */
+@UtilityClass
+public class KNNFeatureFlags {
+
+    // Feature flags
+    private static final String KNN_LAUNCH_QUERY_REWRITE_ENABLED = "knn.feature.query.rewrite.enabled";
+    private static final boolean KNN_LAUNCH_QUERY_REWRITE_ENABLED_DEFAULT = true;
+
+    @VisibleForTesting
+    public static final Setting<Boolean> KNN_LAUNCH_QUERY_REWRITE_ENABLED_SETTING = Setting.boolSetting(
+        KNN_LAUNCH_QUERY_REWRITE_ENABLED,
+        KNN_LAUNCH_QUERY_REWRITE_ENABLED_DEFAULT,
+        NodeScope,
+        Dynamic
+    );
+
+    public static List<Setting<?>> getFeatureFlags() {
+        return Stream.of(KNN_LAUNCH_QUERY_REWRITE_ENABLED_SETTING).collect(Collectors.toUnmodifiableList());
+    }
+
+    public static boolean isKnnQueryRewriteEnabled() {
+        return Boolean.parseBoolean(KNNSettings.state().getSettingValue(KNN_LAUNCH_QUERY_REWRITE_ENABLED).toString());
+    }
+}
--- a/src/main/java/org/opensearch/knn/index/KNNSettings.java
+++ b/src/main/java/org/opensearch/knn/index/KNNSettings.java
@@ -9,17 +9,17 @@ import lombok.extern.log4j.Log4j2;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.opensearch.OpenSearchParseException;
-import org.opensearch.cluster.metadata.IndexMetadata;
-import org.opensearch.core.action.ActionListener;
 import org.opensearch.action.admin.cluster.settings.ClusterUpdateSettingsRequest;
 import org.opensearch.action.admin.cluster.settings.ClusterUpdateSettingsResponse;
 import org.opensearch.client.Client;
+import org.opensearch.cluster.metadata.IndexMetadata;
 import org.opensearch.cluster.service.ClusterService;
 import org.opensearch.common.settings.Setting;
 import org.opensearch.common.settings.Settings;
+import org.opensearch.common.unit.TimeValue;
+import org.opensearch.core.action.ActionListener;
 import org.opensearch.core.common.unit.ByteSizeUnit;
 import org.opensearch.core.common.unit.ByteSizeValue;
-import org.opensearch.common.unit.TimeValue;
 import org.opensearch.index.IndexModule;
 import org.opensearch.knn.index.memory.NativeMemoryCacheManager;
 import org.opensearch.knn.index.memory.NativeMemoryCacheManagerDto;
@@ -28,20 +28,22 @@ import org.opensearch.monitor.jvm.JvmInfo;
 import org.opensearch.monitor.os.OsProbe;
 
 import java.security.InvalidParameterException;
-import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
+import static java.util.stream.Collectors.toUnmodifiableMap;
 import static org.opensearch.common.settings.Setting.Property.Dynamic;
 import static org.opensearch.common.settings.Setting.Property.IndexScope;
 import static org.opensearch.common.settings.Setting.Property.NodeScope;
-import static org.opensearch.core.common.unit.ByteSizeValue.parseBytesSizeValue;
 import static org.opensearch.common.unit.MemorySizeValue.parseBytesSizeValueOrHeapRatio;
+import static org.opensearch.core.common.unit.ByteSizeValue.parseBytesSizeValue;
+import static org.opensearch.knn.common.featureflags.KNNFeatureFlags.getFeatureFlags;
 
 /**
  * This class defines
@@ -289,6 +291,9 @@ public class KNNSettings {
         }
     };
 
+    private final static Map<String, Setting<?>> FEATURE_FLAGS = getFeatureFlags().stream()
+        .collect(toUnmodifiableMap(Setting::getKey, Function.identity()));
+
     private ClusterService clusterService;
     private Client client;
 
@@ -326,7 +331,7 @@ public class KNNSettings {
             );
 
             NativeMemoryCacheManager.getInstance().rebuildCache(builder.build());
-        }, new ArrayList<>(dynamicCacheSettings.values()));
+        }, Stream.concat(dynamicCacheSettings.values().stream(), FEATURE_FLAGS.values().stream()).collect(Collectors.toUnmodifiableList()));
     }
 
     /**
@@ -344,6 +349,10 @@ public class KNNSettings {
     private Setting<?> getSetting(String key) {
         if (dynamicCacheSettings.containsKey(key)) {
             return dynamicCacheSettings.get(key);
+        }
+
+        if (FEATURE_FLAGS.containsKey(key)) {
+            return FEATURE_FLAGS.get(key);
         }
 
         if (KNN_CIRCUIT_BREAKER_TRIGGERED.equals(key)) {
@@ -390,7 +399,8 @@ public class KNNSettings {
             KNN_FAISS_AVX2_DISABLED_SETTING,
             KNN_VECTOR_STREAMING_MEMORY_LIMIT_PCT_SETTING
         );
-        return Stream.concat(settings.stream(), dynamicCacheSettings.values().stream()).collect(Collectors.toList());
+        return Stream.concat(settings.stream(), Stream.concat(getFeatureFlags().stream(), dynamicCacheSettings.values().stream()))
+            .collect(Collectors.toList());
     }
 
     public static boolean isKNNPluginEnabled() {
--- a/src/main/java/org/opensearch/knn/index/query/KNNScorer.java
+++ b/src/main/java/org/opensearch/knn/index/query/KNNScorer.java
@@ -87,6 +87,13 @@ public class KNNScorer extends Scorer {
             public int docID() {
                 return docIdsIter.docID();
             }
+
+            @Override
+            public boolean equals(Object obj) {
+                if (!(obj instanceof Scorer)) return false;
+                return getWeight().equals(((Scorer) obj).getWeight());
+            }
         };
+
     }
 }
--- a/src/main/java/org/opensearch/knn/index/query/KNNWeight.java
+++ b/src/main/java/org/opensearch/knn/index/query/KNNWeight.java
@@ -108,6 +108,22 @@ public class KNNWeight extends Weight {
 
     @Override
     public Scorer scorer(LeafReaderContext context) throws IOException {
+        final Map<Integer, Float> docIdToScoreMap = searchLeaf(context);
+        if (docIdToScoreMap.isEmpty()) {
+            return KNNScorer.emptyScorer(this);
+        }
+
+        return convertSearchResponseToScorer(docIdToScoreMap);
+    }
+
+    /**
+     * Executes k nearest neighbor search for a segment to get the top K results
+     * This is made public purely to be able to be reused in {@link org.opensearch.knn.index.query.nativelib.NativeEngineKnnVectorQuery}
+     *
+     * @param context LeafReaderContext
+     * @return A Map of docId to scores for top k results
+     */
+    public Map<Integer, Float> searchLeaf(LeafReaderContext context) throws IOException {
 
         final BitSet filterBitSet = getFilteredDocsBitSet(context);
         int cardinality = filterBitSet.cardinality();
@@ -115,7 +131,7 @@ public class KNNWeight extends Weight {
         // We should give this condition a deeper look that where it should be placed. For now I feel this is a good
         // place,
         if (filterWeight != null && cardinality == 0) {
-            return KNNScorer.emptyScorer(this);
+            return Collections.emptyMap();
         }
         final Map<Integer, Float> docIdsToScoreMap = new HashMap<>();
 
@@ -129,7 +145,7 @@ public class KNNWeight extends Weight {
         } else {
             Map<Integer, Float> annResults = doANNSearch(context, filterBitSet, cardinality);
             if (annResults == null) {
-                return null;
+                return Collections.emptyMap();
             }
             if (canDoExactSearchAfterANNSearch(cardinality, annResults.size())) {
                 log.debug(
@@ -144,9 +160,9 @@ public class KNNWeight extends Weight {
             docIdsToScoreMap.putAll(annResults);
         }
         if (docIdsToScoreMap.isEmpty()) {
-            return KNNScorer.emptyScorer(this);
+            return Collections.emptyMap();
         }
-        return convertSearchResponseToScorer(docIdsToScoreMap);
+        return docIdsToScoreMap;
     }
 
     private BitSet getFilteredDocsBitSet(final LeafReaderContext ctx) throws IOException {
--- a/src/main/java/org/opensearch/knn/index/query/nativelib/DocAndScoreQuery.java
+++ b/src/main/java/org/opensearch/knn/index/query/nativelib/DocAndScoreQuery.java
@@ -1,0 +1,185 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.knn.index.query.nativelib;
+
+import org.apache.lucene.index.LeafReaderContext;
+import org.apache.lucene.search.DocIdSetIterator;
+import org.apache.lucene.search.Explanation;
+import org.apache.lucene.search.IndexSearcher;
+import org.apache.lucene.search.Query;
+import org.apache.lucene.search.QueryVisitor;
+import org.apache.lucene.search.ScoreMode;
+import org.apache.lucene.search.Scorer;
+import org.apache.lucene.search.Weight;
+
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.Objects;
+
+import static org.apache.lucene.search.DocIdSetIterator.NO_MORE_DOCS;
+
+/**
+ * This is the same as {@link org.apache.lucene.search.AbstractKnnVectorQuery.DocAndScoreQuery}
+ */
+final class DocAndScoreQuery extends Query {
+
+    private final int k;
+    private final int[] docs;
+    private final float[] scores;
+    private final int[] segmentStarts;
+    private final Object contextIdentity;
+
+    DocAndScoreQuery(int k, int[] docs, float[] scores, int[] segmentStarts, Object contextIdentity) {
+        this.k = k;
+        this.docs = docs;
+        this.scores = scores;
+        this.segmentStarts = segmentStarts;
+        this.contextIdentity = contextIdentity;
+    }
+
+    @Override
+    public Weight createWeight(IndexSearcher searcher, ScoreMode scoreMode, float boost) {
+        if (searcher.getIndexReader().getContext().id() != contextIdentity) {
+            throw new IllegalStateException("This DocAndScore query was created by a different reader");
+        }
+        return new Weight(this) {
+            @Override
+            public Explanation explain(LeafReaderContext context, int doc) {
+                int found = Arrays.binarySearch(docs, doc + context.docBase);
+                if (found < 0) {
+                    return Explanation.noMatch("not in top " + k);
+                }
+                return Explanation.match(scores[found] * boost, "within top " + k);
+            }
+
+            @Override
+            public int count(LeafReaderContext context) {
+                return segmentStarts[context.ord + 1] - segmentStarts[context.ord];
+            }
+
+            @Override
+            public Scorer scorer(LeafReaderContext context) {
+                if (segmentStarts[context.ord] == segmentStarts[context.ord + 1]) {
+                    return null;
+                }
+                return new Scorer(this) {
+                    final int lower = segmentStarts[context.ord];
+                    final int upper = segmentStarts[context.ord + 1];
+                    int upTo = -1;
+
+                    @Override
+                    public DocIdSetIterator iterator() {
+                        return new DocIdSetIterator() {
+                            @Override
+                            public int docID() {
+                                return docIdNoShadow();
+                            }
+
+                            @Override
+                            public int nextDoc() {
+                                if (upTo == -1) {
+                                    upTo = lower;
+                                } else {
+                                    ++upTo;
+                                }
+                                return docIdNoShadow();
+                            }
+
+                            @Override
+                            public int advance(int target) throws IOException {
+                                return slowAdvance(target);
+                            }
+
+                            @Override
+                            public long cost() {
+                                return upper - lower;
+                            }
+                        };
+                    }
+
+                    @Override
+                    public float getMaxScore(int docId) {
+                        docId += context.docBase;
+                        float maxScore = 0;
+                        for (int idx = Math.max(0, upTo); idx < upper && docs[idx] <= docId; idx++) {
+                            maxScore = Math.max(maxScore, scores[idx]);
+                        }
+                        return maxScore * boost;
+                    }
+
+                    @Override
+                    public float score() {
+                        return scores[upTo] * boost;
+                    }
+
+                    @Override
+                    public int advanceShallow(int docid) {
+                        int start = Math.max(upTo, lower);
+                        int docidIndex = Arrays.binarySearch(docs, start, upper, docid + context.docBase);
+                        if (docidIndex < 0) {
+                            docidIndex = -1 - docidIndex;
+                        }
+                        if (docidIndex >= upper) {
+                            return NO_MORE_DOCS;
+                        }
+                        return docs[docidIndex];
+                    }
+
+                    /**
+                     * move the implementation of docID() into a differently-named method so we can call it
+                     * from DocIDSetIterator.docID() even though this class is anonymous
+                     *
+                     * @return the current docid
+                     */
+                    private int docIdNoShadow() {
+                        if (upTo == -1) {
+                            return -1;
+                        }
+                        if (upTo >= upper) {
+                            return NO_MORE_DOCS;
+                        }
+                        return docs[upTo] - context.docBase;
+                    }
+
+                    @Override
+                    public int docID() {
+                        return docIdNoShadow();
+                    }
+                };
+            }
+
+            @Override
+            public boolean isCacheable(LeafReaderContext ctx) {
+                return true;
+            }
+        };
+    }
+
+    @Override
+    public String toString(String field) {
+        return "DocAndScore[" + k + "][docs:" + Arrays.toString(docs) + ", scores:" + Arrays.toString(scores) + "]";
+    }
+
+    @Override
+    public void visit(QueryVisitor visitor) {
+        visitor.visitLeaf(this);
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (!sameClassAs(obj)) {
+            return false;
+        }
+        return contextIdentity == ((DocAndScoreQuery) obj).contextIdentity
+            && Arrays.equals(docs, ((DocAndScoreQuery) obj).docs)
+            && Arrays.equals(scores, ((DocAndScoreQuery) obj).scores);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(classHash(), contextIdentity, Arrays.hashCode(docs), Arrays.hashCode(scores));
+    }
+}
--- a/src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java
+++ b/src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java
@@ -1,0 +1,140 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.knn.index.query.nativelib;
+
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
+import org.apache.lucene.index.IndexReader;
+import org.apache.lucene.index.LeafReaderContext;
+import org.apache.lucene.search.IndexSearcher;
+import org.apache.lucene.search.MatchNoDocsQuery;
+import org.apache.lucene.search.Query;
+import org.apache.lucene.search.QueryVisitor;
+import org.apache.lucene.search.ScoreDoc;
+import org.apache.lucene.search.ScoreMode;
+import org.apache.lucene.search.TopDocs;
+import org.apache.lucene.search.TotalHits;
+import org.apache.lucene.util.Bits;
+import org.opensearch.knn.index.query.KNNQuery;
+import org.opensearch.knn.index.query.KNNWeight;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Comparator;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.concurrent.Callable;
+
+/**
+ * {@link KNNQuery} executes approximate nearest neighbor search (ANN) on a segment level.
+ * {@link NativeEngineKnnVectorQuery} executes approximate nearest neighbor search but gives
+ * us the control to combine the top k results in each leaf and post process the results just
+ * for k-NN query if required. This is done by overriding rewrite method to execute ANN on each leaf
+ * {@link KNNQuery} does not give the ability to post process segment results.
+ */
+@Getter
+@RequiredArgsConstructor
+public class NativeEngineKnnVectorQuery extends Query {
+
+    private final KNNQuery knnQuery;
+
+    @Override
+    public Query rewrite(final IndexSearcher indexSearcher) throws IOException {
+        final IndexReader reader = indexSearcher.getIndexReader();
+        final KNNWeight knnWeight = (KNNWeight) knnQuery.createWeight(indexSearcher, ScoreMode.COMPLETE, 1);
+        List<LeafReaderContext> leafReaderContexts = reader.leaves();
+
+        List<Callable<TopDocs>> tasks = new ArrayList<>(leafReaderContexts.size());
+        for (LeafReaderContext leafReaderContext : leafReaderContexts) {
+            tasks.add(() -> searchLeaf(leafReaderContext, knnWeight));
+        }
+        TopDocs[] perLeafResults = indexSearcher.getTaskExecutor().invokeAll(tasks).toArray(TopDocs[]::new);
+        // TopDocs.merge requires perLeafResults to be sorted in descending order.
+        TopDocs topK = TopDocs.merge(knnQuery.getK(), perLeafResults);
+        if (topK.scoreDocs.length == 0) {
+            return new MatchNoDocsQuery();
+        }
+        return createRewrittenQuery(reader, topK);
+    }
+
+    private Query createRewrittenQuery(IndexReader reader, TopDocs topK) {
+        int len = topK.scoreDocs.length;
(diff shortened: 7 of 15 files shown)


src/main/java/org/opensearch/knn/search/NativeEngineKnnVectorQuery.java
```diff
@@ -0,0 +1,282 @@
...
+        List<Callable<TopDocs>> tasks = new ArrayList<>(leafReaderContexts.size());
+        for (LeafReaderContext leafReaderContext : leafReaderContexts) {
+            tasks.add(() -> searchLeaf(leafReaderContext, knnWeight));
+        }
+        TopDocs[] perLeafResults = indexSearcher.getTaskExecutor().invokeAll(tasks).toArray(TopDocs[]::new);
+        TopDocs topK = TopDocs.merge(knnQuery.getK(), perLeafResults);
+        if (topK.scoreDocs.length == 0) {
+            return new MatchNoDocsQuery();
+        }
+        return createRewrittenQuery(reader, topK);
```
shatejas, line 51: nit: Rescoring should happen before this

src/main/java/org/opensearch/knn/search/NativeEngineKnnVectorQuery.java
```diff
@@ -0,0 +1,282 @@
...
+import static org.apache.lucene.search.DocIdSetIterator.NO_MORE_DOCS;
+
+@RequiredArgsConstructor
+public class NativeEngineKnnVectorQuery extends Query {
+
+    private final KNNQuery knnQuery;
+
+    public Query rewrite(final IndexSearcher indexSearcher) throws IOException {
+        final IndexReader reader = indexSearcher.getIndexReader();
+        final KNNWeight knnWeight = (KNNWeight) knnQuery.createWeight(indexSearcher, ScoreMode.COMPLETE, 1);
```
navneet1v, line 39: [Query]: what is the significance  `1` in function param?

src/main/java/org/opensearch/knn/search/NativeEngineKnnVectorQuery.java
```diff
@@ -0,0 +1,282 @@
...
+
+    @Override
+    public String toString(String field) {
+        return this.getClass().getSimpleName() + "[" + field + "]..." + KNNQuery.class.getSimpleName() + "[" + knnQuery.toString() + "]";
+    }
+
+    @Override
+    public void visit(QueryVisitor visitor) {
+        //No-op
+    }
```
navneet1v, line 105: do this: visitor.visitLeaf(this);

src/main/java/org/opensearch/knn/search/NativeEngineKnnVectorQuery.java
```diff
@@ -0,0 +1,282 @@
...
+        TopDocs topK = TopDocs.merge(knnQuery.getK(), perLeafResults);
+        if (topK.scoreDocs.length == 0) {
+            return new MatchNoDocsQuery();
+        }
+        return createRewrittenQuery(reader, topK);
+    }
+
+    private Query createRewrittenQuery(IndexReader reader, TopDocs topK) {
+        int len = topK.scoreDocs.length;
+        Arrays.sort(topK.scoreDocs, Comparator.comparingInt(a -> a.doc));
```
navneet1v, line 56: why we are comparing docId here? shouldn't this be score.

src/main/java/org/opensearch/knn/search/NativeEngineKnnVectorQuery.java
```diff
@@ -0,0 +1,282 @@
...
+
+    private TopDocs searchLeaf(LeafReaderContext ctx, KNNWeight queryWeight) throws IOException {
+        int totalHits = 0;
+        KNNScorer scorer = (KNNScorer) queryWeight.scorer(ctx);
+        DocIdSetIterator docIdSetIterator = scorer.iterator();
+        List<ScoreDoc> scoreDocs = new ArrayList<>();
+        while (docIdSetIterator.nextDoc() != DocIdSetIterator.NO_MORE_DOCS) {
+            scoreDocs.add(new ScoreDoc(docIdSetIterator.docID(), scorer.score()));
+            totalHits++;
+        }
```
navneet1v, line 93: I would really love to see how we can avoid this iteration. may be instead of calling `queryWeight.scorer` we can add another function in KNNWeightClass which can by `queryWeight.scorer` and searchLeaf function to get docs and Score in an easy format.

src/main/java/org/opensearch/knn/search/NativeEngineKnnVectorQuery.java
```diff
@@ -0,0 +1,282 @@
...
+        return new TopDocs(new TotalHits(totalHits, TotalHits.Relation.EQUAL_TO), scoreDocs.toArray(ScoreDoc[]::new));
```
navneet1v, line 94: this should be equal_to only in the case when totalHits size == k.

navneet1v (commented):
Overall code looks good to me. I would like to see other usecase of KNNQuery getting added here like currently filters are not present here. Similarly Radial search, other ITs and BWC tests.

src/main/java/org/opensearch/knn/search/NativeEngineKnnVectorQuery.java
shatejas: Its a boost, I still have to test if thats the right value there

src/main/java/org/opensearch/knn/search/NativeEngineKnnVectorQuery.java
shatejas: So the DocAndScoreQuery expects docs in sorted order. DocAndScoreQuery uses this sorted docIds to divide the results per leaf. This leverages concurrent execution to then put docs in TopDocsCollector faster. 

I did not change it because Its an optimization that will help control latencies when k is large and concurrent segment search is on. Alternative is to have a Query which has a scorer that returns all results in the first leaf and then returns empty scorer for subsequent leaves

src/main/java/org/opensearch/knn/search/NativeEngineKnnVectorQuery.java
shatejas: Took a deeper look into this, I initially thought that DISI is sorting scores with the BulkAdder. But its sorting Ids and not scores. I will have to revisit this since `TopDocs.merge` expects sorted TopDocs by score.

src/main/java/org/opensearch/knn/search/NativeEngineKnnVectorQuery.java
shatejas: > this should be equal_to only in the case when totalHits size == k.

These are leaf level topdoc values so EQUAL_TO is correct here

src/main/java/org/opensearch/knn/search/NativeEngineKnnVectorQuery.java
```diff
@@ -0,0 +1,282 @@
...
+    public Query rewrite(final IndexSearcher indexSearcher) throws IOException {
+        final IndexReader reader = indexSearcher.getIndexReader();
+        final KNNWeight knnWeight = (KNNWeight) knnQuery.createWeight(indexSearcher, ScoreMode.COMPLETE, 1);
+        List<LeafReaderContext> leafReaderContexts = reader.leaves();
+
+        List<Callable<TopDocs>> tasks = new ArrayList<>(leafReaderContexts.size());
+        for (LeafReaderContext leafReaderContext : leafReaderContexts) {
+            tasks.add(() -> searchLeaf(leafReaderContext, knnWeight));
+        }
+        TopDocs[] perLeafResults = indexSearcher.getTaskExecutor().invokeAll(tasks).toArray(TopDocs[]::new);
```
Vikasht34, line 46: Can we have this to handle cases where task is failing for better error handling

```
TopDocs[] perLeafResults = indexSearcher.getTaskExecutor().invokeAll(tasks).stream()
                .map(future -> {
                    try {
                        return future.get();
                    } catch (Exception e) {
                    }
                }).toArray(TopDocs[]::new);
```

Vikasht34: Nvm , I see InvokeAll is already throwing future error ...

src/main/java/org/opensearch/knn/search/NativeEngineKnnVectorQuery.java
```diff
@@ -0,0 +1,282 @@
...
+import java.util.concurrent.Callable;
+
+import static org.apache.lucene.search.DocIdSetIterator.NO_MORE_DOCS;
+
+@RequiredArgsConstructor
+public class NativeEngineKnnVectorQuery extends Query {
+
+    private final KNNQuery knnQuery;
+
+    public Query rewrite(final IndexSearcher indexSearcher) throws IOException {
```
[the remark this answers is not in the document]
shatejas: the delegate will never be executed by IndexSearcher so I don't think so. Am I missing anything in terms of rewrite

src/main/java/org/opensearch/knn/index/query/nativelib/DocAndScoreQuery.java
```diff
@@ -0,0 +1,182 @@
...
+import org.apache.lucene.search.Scorer;
+import org.apache.lucene.search.Weight;
+
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.Objects;
+
+import static org.apache.lucene.search.DocIdSetIterator.NO_MORE_DOCS;
+
+class DocAndScoreQuery extends Query {
```
jmazanec15, line 24: Add link to reference implementation for future readability.

src/main/java/org/opensearch/knn/index/query/KNNScorer.java
```diff
@@ -87,6 +87,13 @@ public float score() throws IOException {
             public int docID() {
                 return docIdsIter.docID();
             }
+
+            @Override
+            public boolean equals(Object obj) {
```
jmazanec15, line 92: qq: Why add?

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -108,14 +108,23 @@ public Explanation explain(LeafReaderContext context, int doc) {
...
     public Scorer scorer(LeafReaderContext context) throws IOException {
+        final Map<Integer, Float> docIdToScoreMap = searchLeaf(context);
+        if (docIdToScoreMap.isEmpty()) {
+            return KNNScorer.emptyScorer(this);
+        }
+
+        return convertSearchResponseToScorer(docIdToScoreMap);
+    }
+
+    public Map<Integer, Float> searchLeaf(LeafReaderContext context) throws IOException {
```
jmazanec15, line 119: javadoc

src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java
```diff
@@ -0,0 +1,130 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Comparator;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.concurrent.Callable;
+
+@RequiredArgsConstructor
+public class NativeEngineKnnVectorQuery extends Query {
```
jmazanec15, line 33: javadoc - add a lot of context about this class

src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java
```diff
@@ -0,0 +1,130 @@
...
+
+    private final KNNQuery knnQuery;
+
+    public Query rewrite(final IndexSearcher indexSearcher) throws IOException {
```
jmazanec15, line 37: override

jmazanec15 (commented):
Overall looks good. Few minor comments

src/main/java/org/opensearch/knn/search/NativeEngineKnnVectorQuery.java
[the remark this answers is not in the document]
jmazanec15: Youre right I think its good.

src/main/java/org/opensearch/knn/index/query/KNNScorer.java
shatejas: - Instead of returning null I am returning [empty map](https://github.com/opensearch-project/k-NN/pull/1877/files#diff-b0eb527e944c912b803db37ed9dd3dfa7689864b77d7146a0d45d7b73138d7b8R125)
- as a consequence I had to [change the unit test](https://github.com/opensearch-project/k-NN/pull/1877/files#diff-1cd3834579e8736ab1ebfec9ffeaef363d96f76443ba40d166ed368cf6c6cedaR353)

`assertEquals` uses equals method to match objects. Comparing objects directly makes for a tight test

jmazanec15 (approved):
LGTM

jmazanec15:
@shatejas do you think it doesnt need changelog?

src/main/java/org/opensearch/knn/common/launchcontrol/KNNLaunchControl.java
```diff
@@ -0,0 +1,39 @@
...
+        Setting.Property.NodeScope,
+        Setting.Property.Dynamic
+    );
+
+    public static Stream<Setting<?>> getFeatureFlags() {
+        return Stream.of(KNN_LAUNCH_QUERY_REWRITE_ENABLED_SETTING);
+    }
+
+    public static boolean isFeatureEnabled(final String featureName) {
+        return Boolean.parseBoolean(KNNSettings.state().getSettingValue(featureName).toString());
```
Vikasht34, line 37: Do we anticipate , parsing failure , can we wrap around try catch , in case failure , can we return false.

src/main/java/org/opensearch/knn/common/launchcontrol/KNNLaunchControl.java
```diff
@@ -0,0 +1,39 @@
...
+public final class KNNLaunchControl {
+
+    // Feature flags
+    public static final String KNN_LAUNCH_QUERY_REWRITE_ENABLED = "knn.launch.query.rewrite.enabled";
+    private static final boolean KNN_LAUNCH_QUERY_REWRITE_ENABLED_DEFAULT = true;
+
+    public static Setting<Boolean> KNN_LAUNCH_QUERY_REWRITE_ENABLED_SETTING = Setting.boolSetting(
+        KNN_LAUNCH_QUERY_REWRITE_ENABLED,
+        KNN_LAUNCH_QUERY_REWRITE_ENABLED_DEFAULT,
+        Setting.Property.NodeScope,
```
Vikasht34, line 28: can we use static imports and have leaf level passed here , just to make more concise.

src/main/java/org/opensearch/knn/common/launchcontrol/KNNLaunchControl.java
```diff
@@ -0,0 +1,39 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
```
Vikasht34, line 5: nit :- I guess signature has been changed 

```
/*
 *
 */
```

src/main/java/org/opensearch/knn/common/launchcontrol/KNNLaunchControl.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+package org.opensearch.knn.common.launchcontrol;
+
+import org.opensearch.common.settings.Setting;
+import org.opensearch.knn.index.KNNSettings;
+
+import java.util.stream.Stream;
+
+public final class KNNLaunchControl {
```
navneet1v, line 19: General terminology is feature flag, let use that in class name,

shatejas:
> @shatejas do you think it doesnt need changelog?

Wasn't planning on adding it. Doesn't seem to be significant enough for users to make it to change log

src/main/java/org/opensearch/knn/common/launchcontrol/KNNLaunchControl.java
Vikasht34, line 19: Do we anticipate object from this class else can have private constructor to block that?

src/main/java/org/opensearch/knn/common/launchcontrol/LaunchControlUtils.java
```diff
@@ -0,0 +1,26 @@
...
+ * GitHub history for details.
+ */
+
+package org.opensearch.knn.common.launchcontrol;
+
+import org.apache.lucene.search.Query;
+import org.opensearch.knn.index.query.KNNQuery;
+import org.opensearch.knn.index.query.nativelib.NativeEngineKnnVectorQuery;
+
+public final class LaunchControlUtils {
```
Vikasht34, line 18: Do we anticipate object from this class else can have private constructor to block that?

src/main/java/org/opensearch/knn/common/launchcontrol/KNNLaunchControl.java
```diff
@@ -0,0 +1,39 @@
...
+        Setting.Property.Dynamic
+    );
+
+    public static Stream<Setting<?>> getFeatureFlags() {
+        return Stream.of(KNN_LAUNCH_QUERY_REWRITE_ENABLED_SETTING);
+    }
+
+    public static boolean isFeatureEnabled(final String featureName) {
+        return Boolean.parseBoolean(KNNSettings.state().getSettingValue(featureName).toString());
+    }
```
navneet1v, line 38: rather than having such generic function, I would recommend moving to isFeatureABCEnabled(). 
1. This ensures client doesn't need to look for string is required to be passed for this feature flag.
2. If this logic of enabling/disabling a feature become complex we just have to make change at 1 place.

src/main/java/org/opensearch/knn/index/KNNSettings.java
```diff
@@ -289,6 +292,9 @@ public class KNNSettings {
         }
     };
 
+    private static Map<String, Setting<?>> FEATURE_FLAGS = KNNLaunchControl.getFeatureFlags()
```
Vikasht34, line 295: Can we use  Collections.unmodifiableMap to ensure that the map is immutable, which is a good practice for feature flags. Additionally can we mark this as final , just to prevent this getting re-assigned.

src/main/java/org/opensearch/knn/common/launchcontrol/KNNLaunchControl.java
```diff
@@ -0,0 +1,39 @@
...
+    // Feature flags
+    public static final String KNN_LAUNCH_QUERY_REWRITE_ENABLED = "knn.launch.query.rewrite.enabled";
+    private static final boolean KNN_LAUNCH_QUERY_REWRITE_ENABLED_DEFAULT = true;
+
+    public static Setting<Boolean> KNN_LAUNCH_QUERY_REWRITE_ENABLED_SETTING = Setting.boolSetting(
+        KNN_LAUNCH_QUERY_REWRITE_ENABLED,
+        KNN_LAUNCH_QUERY_REWRITE_ENABLED_DEFAULT,
+        Setting.Property.NodeScope,
+        Setting.Property.Dynamic
+    );
```
navneet1v, line 30: Looking at the usage of this, this is not a true feature flag. because if we want to make this as feature flag we would have register it with KNNplugin class. 

But if we want to treat this as a setting then lets move this in settings.

src/main/java/org/opensearch/knn/common/launchcontrol/KNNLaunchControl.java
```diff
@@ -0,0 +1,39 @@
...
+
+import java.util.stream.Stream;
+
+public final class KNNLaunchControl {
+
+    // Feature flags
+    public static final String KNN_LAUNCH_QUERY_REWRITE_ENABLED = "knn.launch.query.rewrite.enabled";
+    private static final boolean KNN_LAUNCH_QUERY_REWRITE_ENABLED_DEFAULT = true;
+
+    public static Setting<Boolean> KNN_LAUNCH_QUERY_REWRITE_ENABLED_SETTING = Setting.boolSetting(
```
navneet1v, line 25: why public?

src/main/java/org/opensearch/knn/common/launchcontrol/LaunchControlUtils.java
navneet1v, line 18: Please add Java docs on all public classes and functions.

src/main/java/org/opensearch/knn/index/KNNSettings.java
```diff
@@ -326,7 +332,7 @@ private void setSettingsUpdateConsumers() {
             );
 
             NativeMemoryCacheManager.getInstance().rebuildCache(builder.build());
-        }, new ArrayList<>(dynamicCacheSettings.values()));
+        }, Stream.concat(dynamicCacheSettings.values().stream(), FEATURE_FLAGS.values().stream()).collect(Collectors.toList()));
```
Vikasht34, line 335: can we make this list immutable , since we are not anticipating , any caller to mutate future flag and that ensures that this is also not getting mutated by mistake.

src/main/java/org/opensearch/knn/common/launchcontrol/LaunchControlUtils.java
```diff
@@ -0,0 +1,26 @@
...
+
+    public static Query getNativeKNNQuery(final KNNQuery knnQuery) {
+        if (KNNLaunchControl.isFeatureEnabled(KNNLaunchControl.KNN_LAUNCH_QUERY_REWRITE_ENABLED_SETTING.getKey())) {
+            return new NativeEngineKnnVectorQuery(knnQuery);
+        }
+        return knnQuery;
+    }
```
navneet1v, line 25: The responsibility of this class is not aligned with the function which it has. If some class says `LaunchControlUtils` it would be saying as per my opinion is a feature A/B is in launched mode or not. It should not be giving query in the return type.

src/main/java/org/opensearch/knn/common/launchcontrol/KNNLaunchControl.java
Vikasht34, line 25: QQ :- Will that be any use case , where mutiple threads trying to create setting ? and If yes then can we prevent that? or  do we allow to create setting for each thread?

src/main/java/org/opensearch/knn/common/launchcontrol/LaunchControlUtils.java
```diff
@@ -0,0 +1,26 @@
...
+
+package org.opensearch.knn.common.launchcontrol;
+
+import org.apache.lucene.search.Query;
+import org.opensearch.knn.index.query.KNNQuery;
+import org.opensearch.knn.index.query.nativelib.NativeEngineKnnVectorQuery;
+
+public final class LaunchControlUtils {
+
+    public static Query getNativeKNNQuery(final KNNQuery knnQuery) {
```
navneet1v, line 20: This function can easily be part of KNNQueryFactory.java and `KNNLaunchControl.isFeatureEnabled(KNNLaunchControl.KNN_LAUNCH_QUERY_REWRITE_ENABLED_SETTING.getKey())` can be added as isNativeQueryRewriteEnabled()

src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java
```diff
@@ -0,0 +1,138 @@
...
+        float[] scores = new float[len];
+        for (int i = 0; i < len; i++) {
+            docs[i] = topK.scoreDocs[i].doc;
+            scores[i] = topK.scoreDocs[i].score;
+        }
+        int[] segmentStarts = findSegmentStarts(reader, docs);
+        return new DocAndScoreQuery(knnQuery.getK(), docs, scores, segmentStarts, reader.getContext().id());
+    }
+
+    static int[] findSegmentStarts(IndexReader reader, int[] docs) {
```
navneet1v, line 76: why package private?

navneet1v:
Please make the change in changelog.md file

src/main/java/org/opensearch/knn/index/query/nativelib/DocAndScoreQuery.java
```diff
@@ -0,0 +1,185 @@
...
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.Objects;
+
+import static org.apache.lucene.search.DocIdSetIterator.NO_MORE_DOCS;
+
+/**
+ * This is the same as {@link org.apache.lucene.search.AbstractKnnVectorQuery.DocAndScoreQuery}
+ */
+class DocAndScoreQuery extends Query {
```
Vikasht34, line 27: nit:- can we have make this class final.

src/main/java/org/opensearch/knn/index/query/nativelib/DocAndScoreQuery.java
```diff
@@ -0,0 +1,185 @@
...
+    public void visit(QueryVisitor visitor) {
+        visitor.visitLeaf(this);
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (!sameClassAs(obj)) {
+            return false;
+        }
+        return contextIdentity == ((DocAndScoreQuery) obj).contextIdentity
```
Vikasht34, line 176: nit:- can we just use equals method from contextIndentity

src/main/java/org/opensearch/knn/common/launchcontrol/KNNLaunchControl.java
shatejas: Its declaring a setting object which is done once and defaults are written, multithreading concerns should be handled by settings API not here

src/main/java/org/opensearch/knn/common/launchcontrol/KNNLaunchControl.java
shatejas: Its a read, so we don't. boolean ensure type safety while its being written

src/main/java/org/opensearch/knn/common/launchcontrol/KNNLaunchControl.java
shatejas: I wanted to distinguish between a setting and launch control flag. I used FeatureFlag before but the process of flipping it required a new distribution build and adding a new flag as a setting required a change in core. Both aren't ideal so chose plugin related settings route. 

A separate class makes the intent clear.

src/main/java/org/opensearch/knn/common/launchcontrol/KNNLaunchControl.java
shatejas: It feels like a wrapper which we can get away with, I don't have a preference here

src/main/java/org/opensearch/knn/common/launchcontrol/KNNLaunchControl.java
shatejas: LaunchControl was to not confuse it with existing Feature flag. I can change it back

src/main/java/org/opensearch/knn/common/launchcontrol/KNNLaunchControl.java
shatejas: `LaunchControlUtils` was using it

src/main/java/org/opensearch/knn/common/launchcontrol/LaunchControlUtils.java
shatejas: I couldn't think of a better name. suggestions welcome

src/main/java/org/opensearch/knn/common/launchcontrol/LaunchControlUtils.java
shatejas: Separate should work here especially if we end up making changes for radial query

src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java
shatejas: will add private

shatejas:
> Please make the change in changelog.md file

https://github.com/opensearch-project/k-NN/pull/1877#issuecomment-2264201750

navneet1v:
> > @shatejas do you think it doesnt need changelog?
> 
> Wasn't planning on adding it. Doesn't seem to be significant enough for users to make it to change log

I think we should put a change log.

src/main/java/org/opensearch/knn/common/launchcontrol/LaunchControlUtils.java
navneet1v: Even in that case isn't KNNQueryBuilder going to use this flag only. This being a public interface and just being used by KNNQueryFactory doesn't seem correct. and lets say Radial search wants to use, then it should use isRewriteFeatureEnabled and doesn't need to know if query conversion is needed or not.

src/main/java/org/opensearch/knn/common/launchcontrol/LaunchControlUtils.java
navneet1v: KNNPluginFeatureFlags

src/main/java/org/opensearch/knn/common/launchcontrol/KNNLaunchControl.java
navneet1v: We can put them in 1 package and then lets make everything package private.

src/main/java/org/opensearch/knn/common/launchcontrol/KNNLaunchControl.java
navneet1v: When you say launch control? what launch it is controlling? 

> I used FeatureFlag before but the process of flipping it required a new distribution build 

Why? can you share some reference where you read it?

> adding a new flag as a setting required a change in core

Please share reference.

src/main/java/org/opensearch/knn/common/launchcontrol/KNNLaunchControl.java
navneet1v: Lets put a wrapper. Lets make it easy for the caller.

src/main/java/org/opensearch/knn/common/launchcontrol/KNNLaunchControl.java
navneet1v: Lets use the semantics of Opensearch here it helps in better communication during discussions on github/any other maintainer as everyone understand what a feature flag is. It ensures consistency.

Otherwise at every place we have to write:
Enable/disable this Launch control, which is nothing but similar to feature flag.

src/main/java/org/opensearch/knn/common/launchcontrol/KNNLaunchControl.java
shatejas: > When you say launch control? what launch it is controlling?

NativeEngineQuery. I am changing it back to feature flag if thats the preference

> Why? can you share some reference where you read it?

https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/common/util/FeatureFlags.java#L169

> Please share reference.

https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/common/util/FeatureFlags.java#L138

src/main/java/org/opensearch/knn/common/launchcontrol/LaunchControlUtils.java
shatejas: > Even in that case isn't KNNQueryBuilder going to use this flag only. This being a public interface and just being used by KNNQueryFactory doesn't seem correct.

Adding it in querybuilder will cause cyclic dependency. Querybuilder depends on queryfactory which will depend on a method in query builder

> This being a public interface and just being used by KNNQueryFactory doesn't seem correct
> then it should use isRewriteFeatureEnabled and doesn't need to know if query conversion is needed or not.

The idea is to use a common method to decide if native engine query or not irrespective what code base accesses it. the definition is tight enough to not accept any other query. There isn't a good place, isolating seems like a right approach. I will change the name of the class to narrow down its responsibility

src/main/java/org/opensearch/knn/index/query/nativelib/DocAndScoreQuery.java
shatejas: saves us from npe. will keep it as it is

src/main/java/org/opensearch/knn/common/featureflags/KNNNativeQueryLauncher.java
```diff
@@ -0,0 +1,28 @@
...
+import org.opensearch.knn.index.query.KNNQuery;
+import org.opensearch.knn.index.query.nativelib.NativeEngineKnnVectorQuery;
+
+import static org.opensearch.knn.common.featureflags.KNNFeatureFlags.isKnnLaunchQueryRewriteEnabled;
+
+/**
+ * Utility class to decide on whether to use NativeEngineKnnVectorQuery based on a feature flag
+ */
+@UtilityClass
+public class KNNNativeQueryLauncher {
```
Vikasht34, line 20: Do we really need this class , it feels like wrapping isKnnLaunchQueryRewriteEnabled() over another class , 
on the other hand at caller side , it could be easily 
`return isKnnLaunchQueryRewriteEnabled() ? NativeEngineKnnVectorQuery(knnQuery): knnQuery`

I feel this class can be avoided.  It takes burden to write unit test for same thing again over isKnnLaunchQueryRewriteEnabled which was already unit tested

src/main/java/org/opensearch/knn/index/KNNSettings.java
```diff
@@ -289,6 +292,9 @@ public class KNNSettings {
         }
     };
 
+    private final static Map<String, Setting<?>> FEATURE_FLAGS = KNNFeatureFlags.getFeatureFlags()
```
Vikasht34, line 295: I guess u forgot to remove KNNFeatureFlags here , it's just like getFeatureFlags(), since it's already added in import

src/main/java/org/opensearch/knn/index/KNNSettings.java
```diff
@@ -390,7 +400,8 @@ public List<Setting<?>> getSettings() {
             KNN_FAISS_AVX2_DISABLED_SETTING,
             KNN_VECTOR_STREAMING_MEMORY_LIMIT_PCT_SETTING
         );
-        return Stream.concat(settings.stream(), dynamicCacheSettings.values().stream()).collect(Collectors.toList());
+        return Stream.concat(settings.stream(), Stream.concat(getFeatureFlags(), dynamicCacheSettings.values().stream()))
```
Vikasht34, line 403: Why not FEATURE_FLAGS.values().stream() ? and I guess we missed to make imutable?

src/main/java/org/opensearch/knn/common/featureflags/KNNFeatureFlags.java
```diff
@@ -0,0 +1,48 @@
...
+        KNN_LAUNCH_QUERY_REWRITE_ENABLED_DEFAULT,
+        NodeScope,
+        Dynamic
+    );
+
+    public static Stream<Setting<?>> getFeatureFlags() {
+        return Stream.of(KNN_LAUNCH_QUERY_REWRITE_ENABLED_SETTING);
+    }
+
+    public static boolean isFeatureEnabled(final String featureName) {
```
Vikasht34, line 41: Are we planning to use this function in future for any other feature? Right now with flow it seems redundant , since another public function call another public? and in future if need this to avoid feature name to function ,it could be private.

src/main/java/org/opensearch/knn/index/KNNSettings.java
```diff
@@ -9,39 +9,42 @@
...
 import org.opensearch.cluster.service.ClusterService;
 import org.opensearch.common.settings.Setting;
 import org.opensearch.common.settings.Settings;
+import org.opensearch.common.unit.TimeValue;
+import org.opensearch.core.action.ActionListener;
 import org.opensearch.core.common.unit.ByteSizeUnit;
 import org.opensearch.core.common.unit.ByteSizeValue;
-import org.opensearch.common.unit.TimeValue;
 import org.opensearch.index.IndexModule;
+import org.opensearch.knn.common.featureflags.KNNFeatureFlags;
```
Vikasht34, line 24: so we don't need this import right?

src/main/java/org/opensearch/knn/common/featureflags/KNNNativeQueryLauncher.java
shatejas: > I feel this class can be avoided. It takes burden to write unit test for same thing again ove

The class is for reusability in case we end up adding radial with this. I don't think a small static class is a burden and as far as unit tests are concerned its already written. This class will be gone by 2.19, so I unless there is a valid reason to remove I will prefer to keep it this way

> it could be easily
return isKnnLaunchQueryRewriteEnabled() ? NativeEngineKnnVectorQuery(knnQuery): knnQuery

And we don't have to unit test this?

src/main/java/org/opensearch/knn/index/KNNSettings.java
shatejas: I think if we want imports a certain way we should try and encapsulate in checkstyle, its going to be hard to keep a track of it.

src/main/java/org/opensearch/knn/common/featureflags/KNNFeatureFlags.java
shatejas: > Are we planning to use this function in future for any other feature?

Its made for reuse so yes 

> Right now with flow it seems redundant , since another public function call another public?

if anything I think `isKnnLaunchQueryRewriteEnabled` is redundant IMO.  With each feature we add there will be one extra method and can turn into a long list of methods. There is nothing wrong in forcing the user to know what setting needs to be used and its also hardly reusable but it's not a hill to die on so I have added it for now. 

I would rather prefer reusability and maintainability over convenience and get rid of `isKnnLaunchQueryRewriteEnabled` method than making this private.

Vikasht34: yeah that was my concern , either of two has to be here , not both , if you are removing isKnnLaunchQueryRewriteEnabled , that's also fine by me

src/main/java/org/opensearch/knn/common/launchcontrol/KNNLaunchControl.java
shatejas: I double checked this and it doesn't work so have to go with the settings approach.

Vikasht34 (approved):
Looks Good to me .

CHANGELOG.md
```diff
@@ -27,4 +27,5 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 * Clean up parsing for query [#1824](https://github.com/opensearch-project/k-NN/pull/1824)
 * Refactor engine package structure [#1913](https://github.com/opensearch-project/k-NN/pull/1913)
 * Refactor method structure and definitions [#1920](https://github.com/opensearch-project/k-NN/pull/1920)
-* Generalize lib interface to return context objects [#1925](https://github.com/opensearch-project/k-NN/pull/1925)
\ No newline at end of file
+* Generalize lib interface to return context objects [#1925](https://github.com/opensearch-project/k-NN/pull/1925)
+* Moving query execution to shard level [#1877](https://github.com/opensearch-project/k-NN/pull/1877)
```
navneet1v, line 31: That is not what we are doing. I would recommend changing this to 

Move `k` search k-NN query to re-write phase of vector search query for Native Engines.

Reason:
1. we are not making change in the radial search.
2. We are doing this for native engines.

src/main/java/org/opensearch/knn/index/KNNSettings.java
```diff
@@ -289,6 +291,9 @@ public class KNNSettings {
         }
     };
 
+    private final static Map<String, Setting<?>> FEATURE_FLAGS = getFeatureFlags().stream()
+        .collect(toUnmodifiableMap(Setting::getKey, Function.identity()));
```
navneet1v, line 295: do we need this? can we just use the functions directly on line 402 of this file?

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -108,14 +108,30 @@ public Explanation explain(LeafReaderContext context, int doc) {
...
+     * @param context LeafReaderContext
+     * @return A Map of docId to scores for top k results
+     */
+    public Map<Integer, Float> searchLeaf(LeafReaderContext context) throws IOException {
 
         final BitSet filterBitSet = getFilteredDocsBitSet(context);
         int cardinality = filterBitSet.cardinality();
         // We don't need to go to JNI layer if no documents are found which satisfy the filters
         // We should give this condition a deeper look that where it should be placed. For now I feel this is a good
         // place,
```
vibrantvarun, line 132: I think this comment does not describe the new change added. Can you describe why are we returning an empty Map?

src/main/java/org/opensearch/knn/index/KNNSettings.java
shatejas: Yes we need this, this is used in line 354 when we are trying to get the settings

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
shatejas: It still applies - Its an optimization where we don't search (don't go to JNI layer). filters being empty indicates no results were found. Doing additional search will always results in empty response.

We are returning empty Map to indicate that there were no matches. This is for both NativeEngineKnnVectorQuery as well as `scorer ` in this class. The emtyScorer part is simply moved to line 113

navneet1v:
@shatejas please resolve conflicts.

Vikasht34 (approved):

CHANGELOG.md
```diff
@@ -28,5 +28,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 * Clean up parsing for query [#1824](https://github.com/opensearch-project/k-NN/pull/1824)
 * Refactor engine package structure [#1913](https://github.com/opensearch-project/k-NN/pull/1913)
 * Refactor method structure and definitions [#1920](https://github.com/opensearch-project/k-NN/pull/1920)
+* Generalize lib interface to return context objects [#1925](https://github.com/opensearch-project/k-NN/pull/1925)
```
jmazanec15, line 31: duplicate changelog entry

jmazanec15 (commented):
Can you fix changelog? Other than that looks good

vibrantvarun (approved):

navneet1v (approved):

PR #1888, opened 2024-07-26
Add triage.md based off of security

jmazanec15:
### Description
Adds a triaging markdown file for a new triage meeting we are proposing for k-NN and neural search bi-weekly (please help check if I missed any copy/paste errors).

See security for details: https://github.com/opensearch-project/security/blob/main/TRIAGING.md

Highlights:
- Hosted by Amazon Chime
- 5 pm PST bi-weekly on Tuesdays
- scope: neural-search and k-NN - reason for combining is because there is a lot of maintainer overlap

Feedback is welcome. Whatever process decided upon is not set in stone forever and can be adjusted as we go. 

Will open up a similar PR in neural-search

### Check List

### DIFF
--- a/TRIAGING.md
+++ b/TRIAGING.md
@@ -1,0 +1,89 @@
+<img src="https://opensearch.org/assets/img/opensearch-logo-themed.svg" height="64px">
+
+The maintainers of the k-NN/neural-search Repo's seek to promote an inclusive and engaged community of contributors. In 
+order to facilitate this, bi-weekly triage meetings are open-to-all and attendance is encouraged for anyone who hopes to 
+contribute, discuss an issue, or learn more about the project. To learn more about contributing to the 
+k-NN/neural-search Repo visit the [Contributing](./CONTRIBUTING.md) documentation.
+
+### Do I need to attend for my issue to be addressed/triaged?
+
+Attendance is not required for your issue to be triaged or addressed. All new issues are triaged bi-weekly.
+
+### What happens if my issue does not get covered this time?
+
+Each meeting we seek to  address all new issues. However, should we run out of time before your issue is discussed, you 
+are always welcome to attend the next meeting or to follow up on the issue post itself.
+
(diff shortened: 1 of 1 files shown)


TRIAGING.md
```diff
@@ -0,0 +1,89 @@
+<img src="https://opensearch.org/assets/img/opensearch-logo-themed.svg" height="64px">
+
+The maintainers of the k-NN/neural-search Repo's seek to promote an inclusive and engaged community of contributors. In
```
heemin32, line 3: Can we separate meetings between k-NN and neural-search? I think more focused meeting for each repo could be more efficient. Two groups has different maintainers and different interests I believe.

heemin32: Mentioning of triaging issue `neural-search` repo in k-NN repo sounds a bit weird. Not sure what others think. 
How can we handle if there are many issues in knn and we use most of the time on knn issue only? Should we divide it like 8:00AM to 8:30AM to knn and 8:30AM to 9:00AM to neural? Some people might only have interest in one of repo but not both and to save their time clear separation might be good.

vamshin: Makes sense. How about we focus every alternate week for each plugin? Not sure if 30 min is enough

heemin32: If we want to separate the meetings, we can decide the frequency and length of the meetings on a per-repo basis. I agree that 30 minutes might not be enough.

jmazanec15: We had a discussion with the team and everyone said to have 1 bi-weekly meeting for the 2 repos. I think we should start out with this plan and then adjust as we see fit in the future.

heemin32 (dismissed):

TRIAGING.md
heemin32: Fair enough.

vamshin (approved):

heemin32 (approved):

PR #1897, opened 2024-07-29
Introduce KNNVectorValues interface to iterate on different types of Vector values during indexing and search

navneet1v:
### Description
Added KNNVectorValues interface to iterate on different types of Vector values during indexing and search.

After adding this interface, the other parts of plugin code can reuse this new VectorValues to get and instance for iterating over the vector.

I also went ahead and put all code related to VectorValues generation for test in 1 class. This will help any UT code to get any type of Vector Values very easily.

### Things added in this PR:
1. Added the interface for iterating over the docIds for KNNVectorValues
2. Added the interface to retrieve the vectors and other information from KNNVectorValues.
3. Added the strategies to extract the vector from iterators.
4. Added relevant unit test for the code. For all these interfaces I tried to keep UTs less and use the VectorValues to do all the UTs. This will keep the UTs count in control and given that all lines are tested/covered we should be good here.

### Related Issues
Resolves https://github.com/opensearch-project/k-NN/issues/1853

### Earlier PR:
https://github.com/opensearch-project/k-NN/pull/1855

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngineFieldVectorsWriter.java
+++ b/src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngineFieldVectorsWriter.java
@@ -54,7 +54,7 @@ class NativeEngineFieldVectorsWriter<T> extends KnnFieldVectorsWriter<T> {
         throw new IllegalStateException("Unsupported Vector encoding : " + fieldInfo.getVectorEncoding());
     }
 
-    NativeEngineFieldVectorsWriter(final FieldInfo fieldInfo, final InfoStream infoStream) {
+    private NativeEngineFieldVectorsWriter(final FieldInfo fieldInfo, final InfoStream infoStream) {
         this.fieldInfo = fieldInfo;
         this.infoStream = infoStream;
         vectors = new HashMap<>();
--- a/src/main/java/org/opensearch/knn/index/codec/util/KNNCodecUtil.java
+++ b/src/main/java/org/opensearch/knn/index/codec/util/KNNCodecUtil.java
@@ -115,7 +115,7 @@ public class KNNCodecUtil {
         return String.format("_%s%s", fieldName, extension);
     }
 
-    private static long getTotalLiveDocsCount(final BinaryDocValues binaryDocValues) {
+    public static long getTotalLiveDocsCount(final BinaryDocValues binaryDocValues) {
         long totalLiveDocs;
         if (binaryDocValues instanceof KNN80BinaryDocValues) {
             totalLiveDocs = ((KNN80BinaryDocValues) binaryDocValues).getTotalLiveDocs();
--- a/src/main/java/org/opensearch/knn/index/vectorvalues/KNNBinaryVectorValues.java
+++ b/src/main/java/org/opensearch/knn/index/vectorvalues/KNNBinaryVectorValues.java
@@ -1,0 +1,41 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.knn.index.vectorvalues;
+
+import lombok.ToString;
+import org.apache.lucene.codecs.KnnFieldVectorsWriter;
+import org.apache.lucene.index.BinaryDocValues;
+import org.apache.lucene.index.ByteVectorValues;
+
+import java.io.IOException;
+
+/**
+ * Concrete implementation of {@link KNNVectorValues} that returns byte[] as vector where binary vector is stored and
+ * provides an abstraction over {@link BinaryDocValues}, {@link ByteVectorValues}, {@link KnnFieldVectorsWriter} etc.
+ */
+@ToString(callSuper = true)
+public class KNNBinaryVectorValues extends KNNVectorValues<byte[]> {
+    KNNBinaryVectorValues(KNNVectorValuesIterator vectorValuesIterator) {
+        super(vectorValuesIterator);
+    }
+
+    @Override
+    public byte[] getVector() throws IOException {
+        final byte[] vector = VectorValueExtractorStrategy.extractBinaryVector(vectorValuesIterator);
+        this.dimension = vector.length;
+        return vector;
+    }
+
+    /**
+     * Binary Vector values gets stored as byte[], hence for dimension of the binary vector we have to multiply the
+     * byte[] size with {@link Byte#SIZE}
+     * @return int
+     */
+    @Override
+    public int dimension() {
+        return super.dimension() * Byte.SIZE;
+    }
+}
--- a/src/main/java/org/opensearch/knn/index/vectorvalues/KNNByteVectorValues.java
+++ b/src/main/java/org/opensearch/knn/index/vectorvalues/KNNByteVectorValues.java
@@ -1,0 +1,31 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.knn.index.vectorvalues;
+
+import lombok.ToString;
+import org.apache.lucene.codecs.KnnFieldVectorsWriter;
+import org.apache.lucene.index.BinaryDocValues;
+import org.apache.lucene.index.ByteVectorValues;
+
+import java.io.IOException;
+
+/**
+ * Concrete implementation of {@link KNNVectorValues} that returns float[] as vector and provides an abstraction over
+ * {@link BinaryDocValues}, {@link ByteVectorValues}, {@link KnnFieldVectorsWriter} etc.
+ */
+@ToString(callSuper = true)
+public class KNNByteVectorValues extends KNNVectorValues<byte[]> {
+    KNNByteVectorValues(KNNVectorValuesIterator vectorValuesIterator) {
+        super(vectorValuesIterator);
+    }
+
+    @Override
+    public byte[] getVector() throws IOException {
+        final byte[] vector = VectorValueExtractorStrategy.extractByteVector(vectorValuesIterator);
+        this.dimension = vector.length;
+        return vector;
+    }
+}
--- a/src/main/java/org/opensearch/knn/index/vectorvalues/KNNFloatVectorValues.java
+++ b/src/main/java/org/opensearch/knn/index/vectorvalues/KNNFloatVectorValues.java
@@ -1,0 +1,29 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.knn.index.vectorvalues;
+
+import org.apache.lucene.codecs.KnnFieldVectorsWriter;
+import org.apache.lucene.index.BinaryDocValues;
+import org.apache.lucene.index.FloatVectorValues;
+
+import java.io.IOException;
+
+/**
+ * Concrete implementation of {@link KNNVectorValues} that returns float[] as vector and provides an abstraction over
+ * {@link BinaryDocValues}, {@link FloatVectorValues}, {@link KnnFieldVectorsWriter} etc.
+ */
+public class KNNFloatVectorValues extends KNNVectorValues<float[]> {
+    KNNFloatVectorValues(final KNNVectorValuesIterator vectorValuesIterator) {
+        super(vectorValuesIterator);
+    }
+
+    @Override
+    public float[] getVector() throws IOException {
+        final float[] vector = VectorValueExtractorStrategy.extractFloatVector(vectorValuesIterator);
+        this.dimension = vector.length;
+        return vector;
+    }
+}
--- a/src/main/java/org/opensearch/knn/index/vectorvalues/KNNVectorValues.java
+++ b/src/main/java/org/opensearch/knn/index/vectorvalues/KNNVectorValues.java
@@ -1,0 +1,85 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.knn.index.vectorvalues;
+
+import lombok.ToString;
+import org.apache.lucene.codecs.KnnFieldVectorsWriter;
+import org.apache.lucene.index.BinaryDocValues;
+import org.apache.lucene.index.ByteVectorValues;
+import org.apache.lucene.index.FloatVectorValues;
+
+import java.io.IOException;
+
+/**
+ * An abstract class to iterate over KNNVectors, as KNNVectors are stored as different representation like
+ * {@link BinaryDocValues}, {@link FloatVectorValues}, {@link ByteVectorValues}, {@link KnnFieldVectorsWriter} etc.
+ * @param <T>
+ */
+@ToString
+public abstract class KNNVectorValues<T> {
+
+    protected final KNNVectorValuesIterator vectorValuesIterator;
+    protected int dimension;
+
+    protected KNNVectorValues(final KNNVectorValuesIterator vectorValuesIterator) {
+        this.vectorValuesIterator = vectorValuesIterator;
+    }
+
+    /**
+     * Return a vector reference. If you are adding this address in a List/Map ensure that you are copying the vector first.
+     * This is to ensure that we keep the heap and latency in check by reducing the copies of vectors.
+     *
+     * @return T an array of byte[], float[]
+     * @throws IOException if we are not able to get the vector
+     */
+    public abstract T getVector() throws IOException;
+
+    /**
+     * Dimension of vector is returned. Do call getVector function first before calling this function otherwise you will get 0 value.
+     * @return int
+     */
+    public int dimension() {
+        assert docId() != -1 && dimension != 0 : "Cannot get dimension before we retrieve a vector from KNNVectorValues";
+        return dimension;
+    }
+
+    /**
+     * Returns the total live docs for KNNVectorValues.
+     * @return long
+     */
+    public long totalLiveDocs() {
+        return vectorValuesIterator.liveDocs();
+    }
+
+    /**
+     * Returns the current docId where the iterator is pointing to.
+     * @return int
+     */
+    public int docId() {
+        return vectorValuesIterator.docId();
+    }
+
+    /**
+     * Advances to a specific docId. Ensure that the passed docId is greater than current docId where Iterator is
+     * pointing to, otherwise
+     * {@link IOException} will be thrown
+     * @return int
+     * @throws IOException if we are not able to move to the passed docId.
+     */
+    public int advance(int docId) throws IOException {
+        return vectorValuesIterator.advance(docId);
+    }
+
+    /**
+     * Move to nextDocId.
+     * @return int
+     * @throws IOException if we cannot move to next docId
+     */
+    public int nextDoc() throws IOException {
+        return vectorValuesIterator.nextDoc();
+    }
+
+}
--- a/src/main/java/org/opensearch/knn/index/vectorvalues/KNNVectorValuesFactory.java
+++ b/src/main/java/org/opensearch/knn/index/vectorvalues/KNNVectorValuesFactory.java
@@ -1,0 +1,60 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.knn.index.vectorvalues;
+
+import org.apache.lucene.index.DocsWithFieldSet;
+import org.apache.lucene.search.DocIdSetIterator;
+import org.opensearch.knn.index.VectorDataType;
+
+import java.util.Map;
+
+/**
+ * A factory class that provides various methods to create the {@link KNNVectorValues}.
+ */
+public final class KNNVectorValuesFactory {
+
+    /**
+     * Returns a {@link KNNVectorValues} for the given {@link DocIdSetIterator} and {@link VectorDataType}
+     *
+     * @param vectorDataType {@link VectorDataType}
+     * @param docIdSetIterator {@link DocIdSetIterator}
+     * @return {@link KNNVectorValues} of type float[]
+     */
+    public static <T> KNNVectorValues<T> getVectorValues(final VectorDataType vectorDataType, final DocIdSetIterator docIdSetIterator) {
+        return getVectorValues(vectorDataType, new KNNVectorValuesIterator.DocIdsIteratorValues(docIdSetIterator));
+    }
+
+    /**
+     * Returns a {@link KNNVectorValues} for the given {@link DocIdSetIterator} and a Map of docId and vectors.
+     *
+     * @param vectorDataType {@link VectorDataType}
+     * @param docIdWithFieldSet {@link DocsWithFieldSet}
+     * @return {@link KNNVectorValues} of type float[]
+     */
+    public static <T> KNNVectorValues<T> getVectorValues(
+        final VectorDataType vectorDataType,
+        final DocsWithFieldSet docIdWithFieldSet,
+        final Map<Integer, T> vectors
+    ) {
+        return getVectorValues(vectorDataType, new KNNVectorValuesIterator.FieldWriterIteratorValues<T>(docIdWithFieldSet, vectors));
+    }
+
+    @SuppressWarnings("unchecked")
+    private static <T> KNNVectorValues<T> getVectorValues(
+        final VectorDataType vectorDataType,
+        final KNNVectorValuesIterator knnVectorValuesIterator
+    ) {
+        switch (vectorDataType) {
+            case FLOAT:
+                return (KNNVectorValues<T>) new KNNFloatVectorValues(knnVectorValuesIterator);
+            case BYTE:
+                return (KNNVectorValues<T>) new KNNByteVectorValues(knnVectorValuesIterator);
+            case BINARY:
+                return (KNNVectorValues<T>) new KNNBinaryVectorValues(knnVectorValuesIterator);
+        }
+        throw new IllegalArgumentException("Invalid Vector data type provided, hence cannot return VectorValues");
+    }
+}
--- a/src/main/java/org/opensearch/knn/index/vectorvalues/KNNVectorValuesIterator.java
+++ b/src/main/java/org/opensearch/knn/index/vectorvalues/KNNVectorValuesIterator.java
@@ -1,0 +1,188 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.knn.index.vectorvalues;
+
+import lombok.NonNull;
+import org.apache.lucene.codecs.KnnFieldVectorsWriter;
+import org.apache.lucene.index.BinaryDocValues;
+import org.apache.lucene.index.ByteVectorValues;
+import org.apache.lucene.index.DocsWithFieldSet;
+import org.apache.lucene.index.FloatVectorValues;
+import org.apache.lucene.search.DocIdSetIterator;
+import org.opensearch.knn.index.codec.util.KNNCodecUtil;
+
+import java.io.IOException;
+import java.util.List;
+import java.util.Map;
+import java.util.function.Function;
+
+/**
+ * An abstract class that provides an iterator to iterate over KNNVectors, as KNNVectors are stored as different
+ * representation like {@link BinaryDocValues}, {@link FloatVectorValues}, FieldWriter etc. How to iterate using this
+ * iterator please refer {@link DocIdsIteratorValues} java docs.
+ */
+public interface KNNVectorValuesIterator {
+
+    /**
+     * Returns the current docId where the iterator is pointing to.
+     * @return int
+     */
+    int docId();
+
+    /**
+     * Advances to a specific docId. Ensure that the passed docId is greater than current docId where Iterator is
+     * pointing to, otherwise
+     * {@link IOException} will be thrown
+     * @return int
+     * @throws IOException if we are not able to move to the passed docId.
+     */
+    int advance(int docId) throws IOException;
+
+    /**
+     * Move to nextDocId. If no more docs are present then {@link DocIdSetIterator#NO_MORE_DOCS} will be returned.
+     * @return int
+     * @throws IOException if we cannot move to next docId
+     */
+    int nextDoc() throws IOException;
+
+    /**
+     * Return a {@link DocIdSetIterator}
+     * @return {@link DocIdSetIterator}
+     */
+    DocIdSetIterator getDocIdSetIterator();
+
+    /**
+     * Total number of live doc which will the iterator will iterate upon.
+     * @return long: total number of live docs
+     */
+    long liveDocs();
+
+    /**
+     * Returns the {@link VectorValueExtractorStrategy} to extract the vector from the iterator.
+     * @return VectorValueExtractorStrategy
+     */
+    VectorValueExtractorStrategy getVectorExtractorStrategy();
+
+    /**
+     * A DocIdsIteratorValues provides a common iteration logic for all Values that implements
+     * {@link DocIdSetIterator} interface. Example: {@link BinaryDocValues}, {@link FloatVectorValues} etc.
+     */
+    class DocIdsIteratorValues implements KNNVectorValuesIterator {
+        protected DocIdSetIterator docIdSetIterator;
+        private static final List<Function<DocIdSetIterator, Boolean>> VALID_ITERATOR_INSTANCE = List.of(
+            (itr) -> itr instanceof BinaryDocValues,
+            (itr) -> itr instanceof FloatVectorValues,
+            (itr) -> itr instanceof ByteVectorValues
+        );
+
+        DocIdsIteratorValues(@NonNull final DocIdSetIterator docIdSetIterator) {
+            validateIteratorType(docIdSetIterator);
+            this.docIdSetIterator = docIdSetIterator;
+        }
+
+        @Override
+        public int docId() {
+            return docIdSetIterator.docID();
+        }
+
+        @Override
+        public int advance(int docId) throws IOException {
+            return docIdSetIterator.advance(docId);
+        }
+
+        @Override
+        public int nextDoc() throws IOException {
+            return docIdSetIterator.nextDoc();
+        }
+
+        @Override
+        public DocIdSetIterator getDocIdSetIterator() {
+            return docIdSetIterator;
+        }
+
+        @Override
+        public long liveDocs() {
+            if (docIdSetIterator instanceof BinaryDocValues) {
+                return KNNCodecUtil.getTotalLiveDocsCount((BinaryDocValues) docIdSetIterator);
+            } else if (docIdSetIterator instanceof FloatVectorValues || docIdSetIterator instanceof ByteVectorValues) {
+                return docIdSetIterator.cost();
+            }
+            throw new IllegalArgumentException(
+                "DocIdSetIterator present is not of valid type. Valid types are: BinaryDocValues, FloatVectorValues and ByteVectorValues"
+            );
+        }
+
+        @Override
+        public VectorValueExtractorStrategy getVectorExtractorStrategy() {
+            return new VectorValueExtractorStrategy.DISIVectorExtractor();
+        }
+
+        private void validateIteratorType(final DocIdSetIterator docIdSetIterator) {
+            VALID_ITERATOR_INSTANCE.stream()
+                .map(v -> v.apply(docIdSetIterator))
+                .filter(Boolean::booleanValue)
+                .findFirst()
+                .orElseThrow(
+                    () -> new IllegalArgumentException(
+                        "DocIdSetIterator present is not of valid type. Valid types are: BinaryDocValues, FloatVectorValues and ByteVectorValues"
+                    )
+                );
+        }
+    }
+
+    /**
+     * A FieldWriterIteratorValues is mainly used when Vectors are stored in {@link KnnFieldVectorsWriter} interface.
+     */
+    class FieldWriterIteratorValues<T> implements KNNVectorValuesIterator {
+        private final DocIdSetIterator docIdSetIterator;
+        private final Map<Integer, T> vectors;
+
+        FieldWriterIteratorValues(@NonNull final DocsWithFieldSet docsWithFieldSet, @NonNull final Map<Integer, T> vectors) {
+            assert docsWithFieldSet.iterator().cost() == vectors.size();
+            this.vectors = vectors;
+            this.docIdSetIterator = docsWithFieldSet.iterator();
+        }
+
+        @Override
+        public int docId() {
+            return docIdSetIterator.docID();
+        }
+
+        @Override
+        public int advance(int docId) throws IOException {
+            return docIdSetIterator.advance(docId);
+        }
+
+        @Override
+        public int nextDoc() throws IOException {
+            return docIdSetIterator.nextDoc();
+        }
+
+        /**
+         * Returns a Map of docId and vector.
+         * @return {@link Map}
+         */
+        public T vectorsValue() {
+            return vectors.get(docId());
+        }
+
+        @Override
+        public DocIdSetIterator getDocIdSetIterator() {
+            return docIdSetIterator;
+        }
+
+        @Override
+        public long liveDocs() {
+            return docIdSetIterator.cost();
+        }
+
+        @Override
+        public VectorValueExtractorStrategy getVectorExtractorStrategy() {
+            return new VectorValueExtractorStrategy.FieldWriterIteratorVectorExtractor();
+        }
+    }
+
+}
--- a/src/main/java/org/opensearch/knn/index/vectorvalues/VectorValueExtractorStrategy.java
+++ b/src/main/java/org/opensearch/knn/index/vectorvalues/VectorValueExtractorStrategy.java
@@ -1,0 +1,126 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.knn.index.vectorvalues;
+
+import org.apache.lucene.index.BinaryDocValues;
+import org.apache.lucene.index.ByteVectorValues;
+import org.apache.lucene.index.FloatVectorValues;
+import org.apache.lucene.search.DocIdSetIterator;
+import org.apache.lucene.util.ArrayUtil;
+import org.apache.lucene.util.BytesRef;
+import org.opensearch.knn.index.VectorDataType;
+import org.opensearch.knn.index.codec.util.KNNVectorSerializer;
+import org.opensearch.knn.index.codec.util.KNNVectorSerializerFactory;
+
+import java.io.IOException;
+
+/**
+ * Provides different strategies to extract the vectors from different {@link KNNVectorValuesIterator}
+ */
+interface VectorValueExtractorStrategy {
+
+    /**
+     * Extract a float vector from KNNVectorValuesIterator.
+     * @param iterator {@link KNNVectorValuesIterator}
+     * @return float[]
+     * @throws IOException exception while retrieving the vectors
+     */
+    static float[] extractFloatVector(final KNNVectorValuesIterator iterator) throws IOException {
+        return iterator.getVectorExtractorStrategy().extract(VectorDataType.FLOAT, iterator);
+    }
+
+    /**
+     * Extract a byte vector from KNNVectorValuesIterator.
+     * @param iterator {@link KNNVectorValuesIterator}
+     * @return byte[]
+     * @throws IOException exception while retrieving the vectors
+     */
+    static byte[] extractByteVector(final KNNVectorValuesIterator iterator) throws IOException {
+        return iterator.getVectorExtractorStrategy().extract(VectorDataType.BYTE, iterator);
+    }
+
+    /**
+     * Extract a binary vector which is represented as byte[] from KNNVectorValuesIterator.
+     * @param iterator {@link KNNVectorValuesIterator}
+     * @return byte[]
+     * @throws IOException exception while retrieving the vectors
+     */
+    static byte[] extractBinaryVector(final KNNVectorValuesIterator iterator) throws IOException {
+        return iterator.getVectorExtractorStrategy().extract(VectorDataType.BINARY, iterator);
+    }
+
+    /**
+     * Extract Vector based on the vector datatype and vector values iterator.
+     * @param vectorDataType {@link VectorDataType}
+     * @param vectorValuesIterator {@link KNNVectorValuesIterator}
+     * @return vector
+     * @param <T> could be of type float[], byte[]
+     * @throws IOException exception during extracting the vectors
+     */
+    <T> T extract(VectorDataType vectorDataType, KNNVectorValuesIterator vectorValuesIterator) throws IOException;
+
+    /**
+     * Strategy to extract the vector from {@link KNNVectorValuesIterator.DocIdsIteratorValues}
+     */
+    class DISIVectorExtractor implements VectorValueExtractorStrategy {
+        @Override
+        public <T> T extract(final VectorDataType vectorDataType, final KNNVectorValuesIterator vectorValuesIterator) throws IOException {
+            final DocIdSetIterator docIdSetIterator = vectorValuesIterator.getDocIdSetIterator();
+            switch (vectorDataType) {
+                case FLOAT:
+                    if (docIdSetIterator instanceof BinaryDocValues) {
+                        final BinaryDocValues values = (BinaryDocValues) docIdSetIterator;
+                        return (T) getFloatVectorFromByteRef(values.binaryValue());
+                    } else if (docIdSetIterator instanceof FloatVectorValues) {
+                        return (T) ((FloatVectorValues) docIdSetIterator).vectorValue();
+                    }
+                    throw new IllegalArgumentException(
+                        "VectorValuesIterator is not of a valid type. Valid Types are: BinaryDocValues and FloatVectorValues"
+                    );
+                case BYTE:
+                case BINARY:
+                    if (docIdSetIterator instanceof BinaryDocValues) {
+                        final BinaryDocValues values = (BinaryDocValues) docIdSetIterator;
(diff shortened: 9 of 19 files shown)


src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngineFieldVectorsWriter.java
```diff
@@ -54,7 +54,7 @@ static NativeEngineFieldVectorsWriter<?> create(final FieldInfo fieldInfo, final
         throw new IllegalStateException("Unsupported Vector encoding : " + fieldInfo.getVectorEncoding());
     }
 
-    NativeEngineFieldVectorsWriter(final FieldInfo fieldInfo, final InfoStream infoStream) {
+    private NativeEngineFieldVectorsWriter(final FieldInfo fieldInfo, final InfoStream infoStream) {
```
navneet1v, line 57: making this change resolves this comment: https://github.com/opensearch-project/k-NN/pull/1855#discussion_r1692067872

src/main/java/org/opensearch/knn/index/vectorvalues/KNNVectorValues.java
```diff
@@ -0,0 +1,85 @@
...
+    public int dimension() {
+        return dimension;
+    }
+
+    /**
+     * Returns the total live docs for KNNVectorValues.
+     * @return long
+     */
+    public long totalLiveDocs() {
+        if (vectorValuesIterator instanceof KNNVectorValuesIterator.DocIdsIteratorValues) {
```
shatejas, line 63: KNNVectorValuesIterator interface is defined in KNN, so instead of instance checks delegate to its implementation to reduce branching.

```
KNNVectorValuesIterator {
........//existing methods
      default long totalLiveDocs() {
             return getDocsIdSetITerator().cost(); // If you don't like this you can move it to FieldsValueIterator
      }
}
```
```
DocIdsIteratorValues {
    
long totalLiveDocs() {
      if (instance of KNN80BinaryDocValues) { // Yes, don't depend on knncodecutil unless there is a strong need, // this simplifies the logic for you
             return getDocsIdSetITerator().getTotalLiveDocs
      } 
     return getDocsIdSetITerator().cost()
}
}
```

and then this will become
```
public long totalLiveDocs() {
       return vectorValuesIterator.totalLiveDocs();
}
```
I understand that there are certain validations in place here but think through it, is it a dead validation or not. If there is really a need move it to the implementation of vectorvaluesiterator

navneet1v: after doing that, we will be just removing 1 level of if condition. But as it is 1 less nesting, I think its fine. and iterator giving live docs make sense as an interface. BTW lucene uses cost/cardinality for this function.

src/main/java/org/opensearch/knn/index/codec/util/KNNCodecUtil.java
```diff
@@ -115,7 +115,7 @@ public static String buildEngineFileSuffix(String fieldName, String extension) {
         return String.format("_%s%s", fieldName, extension);
     }
 
-    private static long getTotalLiveDocsCount(final BinaryDocValues binaryDocValues) {
+    public static long getTotalLiveDocsCount(final BinaryDocValues binaryDocValues) {
```
navneet1v, line 118: This KNNCodecUtil class will be refactored, once I integrate the NativeIndexWriter functionality with KNNVectorsFormatCodec.

Vikasht34 (commented):
Will Continue with remaining Part tomorrow

src/main/java/org/opensearch/knn/index/vectorvalues/KNNVectorValuesFactory.java
```diff
@@ -0,0 +1,111 @@
...
+import org.apache.lucene.index.DocsWithFieldSet;
+import org.apache.lucene.index.FloatVectorValues;
+import org.apache.lucene.search.DocIdSetIterator;
+
+import java.util.Map;
+
+/**
+ * A factory class that provides various methods to create the {@link KNNVectorValues}.
+ */
+public final class KNNVectorValuesFactory {
```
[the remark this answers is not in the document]
navneet1v: there is no use of Vector datatype parameter to be passed as the factory class is providing concrete functions. The problem with the suggested methods is introduction of `?` in the return type. Which make it very hard for caller to know what is the return type. Hence its not needed.

src/main/java/org/opensearch/knn/index/vectorvalues/KNNByteVectorValues.java
```diff
@@ -0,0 +1,73 @@
...
+ * Concrete implementation of {@link KNNVectorValues} that returns float[] as vector and provides an abstraction over
+ * {@link BinaryDocValues}, {@link ByteVectorValues}, {@link KnnFieldVectorsWriter} etc.
+ */
+@ToString(callSuper = true)
+public class KNNByteVectorValues extends KNNVectorValues<byte[]> {
+    KNNByteVectorValues(KNNVectorValuesIterator vectorValuesIterator) {
+        super(vectorValuesIterator);
+    }
+
+    public int docId() {
```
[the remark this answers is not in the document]
navneet1v: because these functions are coming from interface and not from abstract class.

src/main/java/org/opensearch/knn/index/vectorvalues/KNNFloatVectorValues.java
```diff
@@ -0,0 +1,78 @@
...
+import org.opensearch.knn.index.codec.util.KNNVectorSerializerFactory;
+
+import java.io.IOException;
+
+/**
+ * Concrete implementation of {@link KNNVectorValues} that returns float[] as vector and provides an abstraction over
+ * {@link BinaryDocValues}, {@link FloatVectorValues}, {@link KnnFieldVectorsWriter} etc.
+ */
+
+public class KNNFloatVectorValues extends KNNVectorValues<float[]> {
```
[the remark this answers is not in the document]
navneet1v: > Every Vector Value Class has 3 responsibilities

Let me clarify something here getVector idea is to get the vector from various iterators. I can move the functionality inside iterator classes, which may reduce top level if else but as VectorValues uses generic and DocIdsIteratorValues doesn't. Also in that case even if I introduce generic in DocIdsIteratorValues, I still have to do instanceof checks and then for BinaryDocValues there will be no way to know whether I have to return a float[] or byte array.


> Apart from this for every vector value class, this if else is common pattern except for data type and return type change 

No that is not the case. You just looked at the high level if else condition. Look closely, the way we convert byteref to vectors and what functions we call to get the vectors is different. Yes there is some common code for fieldsWriter which can be moved but since its was small I decided to keep it.

src/main/java/org/opensearch/knn/index/vectorvalues/KNNVectorValuesIterator.java
```diff
@@ -0,0 +1,157 @@
...
+import org.opensearch.knn.index.codec.util.KNNCodecUtil;
+
+import java.io.IOException;
+import java.util.Map;
+
+/**
+ * An abstract class that prvides an iterator to iterate over KNNVectors, as KNNVectors are stored as different
+ * representation like {@link BinaryDocValues}, {@link FloatVectorValues}, FieldWriter etc.
+ */
+public interface KNNVectorValuesIterator {
```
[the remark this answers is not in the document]
navneet1v: > it makes harder to read the file classes

That is a subjective thing. 
Plus having this type of personally, I see reduces the number of files to maintain. Also, if you see the code is pretty less, if required later we can move these classes in separate file. But ideally if interface implementation is small I like to keep them in single file. It helps in managing and reading code easy(again a subjective thing).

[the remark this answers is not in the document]
navneet1v: Also, I don't see these iterators getting exposed to any class outside, hence keeping them in a single file is a good idea.

src/main/java/org/opensearch/knn/index/vectorvalues/KNNVectorValuesFactory.java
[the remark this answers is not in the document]
Vikasht34: If I have 10 child classes , By using the implemented way for factory pattern , I can have 10 methods which return instance of 10 child objects individually  and that's factory pattern?

We have something called generics in Java , that helps creating generic methods , here is how we can do that 

```
package org.opensearch.knn.index.vectorvalues;

public enum VectorType {
    FLOAT,
    BYTE,
    BINARY
}
```
package org.opensearch.knn.index.vectorvalues;

import org.apache.lucene.search.DocIdSetIterator;

import java.util.Map;

public class KNNVectorValuesFactory {

    public static <T> KNNVectorValues<T> getVectorValues(VectorType vectorType, DocIdSetIterator docIdSetIterator) {
        KNNVectorValuesIterator vectorValuesIterator = new KNNVectorValuesIterator.DocIdsIteratorValues(docIdSetIterator);

        switch (vectorType) {
            case FLOAT:
                return (KNNVectorValues<T>) new KNNFloatVectorValues(vectorValuesIterator);
            case BYTE:
                return (KNNVectorValues<T>) new KNNByteVectorValues(vectorValuesIterator);
            case BINARY:
                return (KNNVectorValues<T>) new KNNBinaryVectorValues(vectorValuesIterator);
            default:
                throw new IllegalArgumentException("Unknown vector type: " + vectorType);
        }
    }

    public static <T> KNNVectorValues<T> getVectorValues(VectorType vectorType, DocsWithFieldSet docIdWithFieldSet, Map<Integer, T> vectors) {
        KNNVectorValuesIterator vectorValuesIterator = new KNNVectorValuesIterator.FieldWriterIteratorValues<>(docIdWithFieldSet, vectors);

        switch (vectorType) {
            case FLOAT:
                return (KNNVectorValues<T>) new KNNFloatVectorValues(vectorValuesIterator);
            case BYTE:
                return (KNNVectorValues<T>) new KNNByteVectorValues(vectorValuesIterator);
            case BINARY:
                return (KNNVectorValues<T>) new KNNBinaryVectorValues(vectorValuesIterator);
            default:
                throw new IllegalArgumentException("Unknown vector type: " + vectorType);
        }
    }
}


Here is how Client can call, 

```
DocIdSetIterator iterator = ...; // Your iterator instance
KNNVectorValues<float[]> floatVectors = KNNVectorValuesFactory.getVectorValues(VectorType.FLOAT, iterator);
KNNVectorValues<byte[]> byteVectors = KNNVectorValuesFactory.getVectorValues(VectorType.BYTE, iterator);

// For the case with DocsWithFieldSet and a map of vectors
DocsWithFieldSet docsWithFieldSet = ...;
Map<Integer, float[]> floatVectorMap = ...;
KNNVectorValues<float[]> floatVectorsWithMap = KNNVectorValuesFactory.getVectorValues(VectorType.FLOAT, docsWithFieldSet, floatVectorMap);

```

src/main/java/org/opensearch/knn/index/vectorvalues/KNNFloatVectorValues.java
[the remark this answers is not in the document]
Vikasht34: Sure , there is something called DRY ..this is quote what it says 

```
"The Pragmatic Programmer" discusses the importance of code organization and the principles of DRY (Don't Repeat Yourself) and modularity.

On managing complexity and organizing code:

    "If you find yourself thinking that you must repeat code to make it work in a new context, you are probably missing an abstraction or two." (Chapter 3, "The DRY Principle")
```

let me give you simple implementation of the provided code , 
Which addresses , 
1. Separation of Concern 
2. Single Responsibly pattern
3. And Easily Maintaible 
4. Shared Common logic of get vector 


My Vector Values classes is simple , you can compare to what you have right now , I used generics and composition pattern to address this issue in your code.

Note :- Below is simple implementaion , it can be further simplified with functional interface


```
public class KNNFloatVectorValues {
    private final KNNVectorValuesIterator vectorValuesIterator;
    private int dimension;

    public KNNFloatVectorValues(KNNVectorValuesIterator vectorValuesIterator) {
        this.vectorValuesIterator = vectorValuesIterator;
    }

    public float[] getVector() throws IOException {
        return VectorValuesExtractor.extractVector(vectorValuesIterator, VectorValuesExtractor::extractFloatVector);
    }
}

public class KNNByteVectorValues {
    private final KNNVectorValuesIterator vectorValuesIterator;
    private int dimension;

    public KNNByteVectorValues(KNNVectorValuesIterator vectorValuesIterator) {
        this.vectorValuesIterator = vectorValuesIterator;
    }

    public byte[] getVector() throws IOException {
        return VectorValuesExtractor.extractVector(vectorValuesIterator, VectorValuesExtractor::extractByteVector);
    }
}

```

I sperated out common extration logic from vector values classes to from here so that it can be re-used and if we can avoid duplication of if else in all vector values classes 

```
import org.apache.lucene.index.BinaryDocValues;
import org.apache.lucene.index.FloatVectorValues;
import org.apache.lucene.index.ByteVectorValues;
import org.apache.lucene.search.DocIdSetIterator;
import org.apache.lucene.util.ArrayUtil;
import org.apache.lucene.util.BytesRef;

import java.io.IOException;
import java.util.function.Function;

public class VectorValuesExtractor {

    public static <T> T extractVector(KNNVectorValuesIterator vectorValuesIterator, Function<DocIdSetIterator, T> extractor) throws IOException {
        T vector = null;

        if (vectorValuesIterator instanceof KNNVectorValuesIterator.DocIdsIteratorValues) {
            DocIdSetIterator docIdSetIterator = vectorValuesIterator.getDocIdSetIterator();
            vector = extractor.apply(docIdSetIterator);
        } else if (vectorValuesIterator instanceof KNNVectorValuesIterator.FieldWriterIteratorValues) {
            KNNVectorValuesIterator.FieldWriterIteratorValues<T> fieldWriterIteratorValues =
                (KNNVectorValuesIterator.FieldWriterIteratorValues<T>) vectorValuesIterator;
            vector = fieldWriterIteratorValues.vectorsValue();
        }

        if (vector == null) {
            throw new IllegalArgumentException("Unsupported DocIdSetIterator type");
        }

        return vector;
    }

    public static float[] extractFloatVector(DocIdSetIterator docIdSetIterator) {
        try {
            if (docIdSetIterator instanceof BinaryDocValues) {
                BinaryDocValues values = (BinaryDocValues) docIdSetIterator;
                BytesRef bytesRef = values.binaryValue();
                return getFloatVectorFromByteRef(bytesRef);
            } else if (docIdSetIterator instanceof FloatVectorValues) {
                return ((FloatVectorValues) docIdSetIterator).vectorValue();
            }
        } catch (IOException e) {
            throw new RuntimeException("Error extracting float vector", e);
        }
        throw new IllegalArgumentException("Unsupported DocIdSetIterator type for float vector extraction");
    }

    public static byte[] extractByteVector(DocIdSetIterator docIdSetIterator) {
        try {
            if (docIdSetIterator instanceof BinaryDocValues) {
                BinaryDocValues values = (BinaryDocValues) docIdSetIterator;
                BytesRef bytesRef = values.binaryValue();
                return ArrayUtil.copyOfSubArray(bytesRef.bytes, bytesRef.offset, bytesRef.offset + bytesRef.length);
            } else if (docIdSetIterator instanceof ByteVectorValues) {
                return ((ByteVectorValues) docIdSetIterator).vectorValue();
            }
        } catch (IOException e) {
            throw new RuntimeException("Error extracting byte vector", e);
        }
        throw new IllegalArgumentException("Unsupported DocIdSetIterator type for byte vector extraction");
    }

    private static float[] getFloatVectorFromByteRef(BytesRef bytesRef) {
        // Implement the conversion logic from BytesRef to float[]
        // This is a placeholder
        return new float[0];
    }
}

```

As you can see 
1. I have one classes soley resposible for getting vector out 
2. now Every vector value class does not have to write same if else condition.

src/main/java/org/opensearch/knn/index/vectorvalues/KNNVectorValuesIterator.java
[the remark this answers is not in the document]
Vikasht34: Sure , here is what some of the best pratices say about the code that you have Written

1. "Clean Code" by Robert C. Martin 

``` 
"The larger and more complex a class, the harder it is to understand. Code that is hard to understand is harder to maintain. It’s harder to ensure that it functions correctly. It’s harder to ensure that it doesn't have bugs." (Chapter 2, "Meaningful Names")
```
2.  "The Pragmatic Programmer" by Andrew Hunt and David Thomas
```
On breaking down functionality:

    "Divide your work into small chunks that you can accomplish, review, and test. This approach is more likely to lead to modular, maintainable code." (Chapter 5,)
```

This is not subjective thing , that's well re-commened by renouce authors on braking functions and having modular. 

I don't buy the idea of "iterators getting exposed to any class outside" , that's upto us how to use it , out side of package definitely restrict it , and inside the package depends on usibility

src/main/java/org/opensearch/knn/index/vectorvalues/KNNFloatVectorValues.java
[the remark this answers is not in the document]
heemin32: I think it is time to set some guideline on how to code should look like and make it official for knn repo so that we can save our energy for every PR.

For this case, something like
* Each class, interface, enum should be defined in its own file
vs
* Single file can have multiple class, interface, and enum. It can be determined solely by the author.

src/main/java/org/opensearch/knn/index/vectorvalues/KNNVectorValuesFactory.java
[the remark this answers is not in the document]
heemin32: +1 with current code. 

When you call `KNNVectorValuesFactory.getVectorValues(VectorType.BINARY, iterator);` it is hard to know if its return type is `KNNVectorValues<float[]>` or `KNNVectorValues<byte[]>`. You need to go inside the method and check what is its return value. 
By having `public static KNNVectorValues<byte[]> getBinaryVectorValues(final DocIdSetIterator docIdSetIterator)`, it is more explicit and you just know what it will return without going inside the method and check the actual return value.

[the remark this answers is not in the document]
heemin32: Returnning `?` type is not ideal in static type system where you cannot catch the issue during compile time but only runtime.

src/main/java/org/opensearch/knn/index/vectorvalues/KNNFloatVectorValues.java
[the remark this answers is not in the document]
Vikasht34: Make Sense

src/main/java/org/opensearch/knn/index/vectorvalues/KNNVectorValuesFactory.java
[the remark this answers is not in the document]
Vikasht34: well when u call KNNVectorValuesFactory.getVectorValues(VectorType.BINARY, iterator) , it says that vectorr value is type Binary as you hve already speicified. while calling , I don't see any reason on confusion 

So here was origianl thing 

KNNVectorValues<float[]> x = KNNVectorValuesFactory.getFloatVectorValue()
vs 
KNNVectorValues<float[]> floatVectorsWithMap = KNNVectorValuesFactory.getVectorValues(VectorType.FLOAT, docsWithFieldSet, floatVectorMap);

where is the confusion?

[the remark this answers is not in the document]
Vikasht34: > Returnning `?` type is not ideal in static type system where you cannot catch the issue during compile time but only runtime.

I slightly disagree on this , this is from Java Creator Effective Java (3rd Edition)  Joshua Bloch

Item number :- 29 ,  Consider typesafe heterogeneous containers and Favor generic methods, 
Bloch advocates for using generic methods to increase flexibility and reusability. He explains how generic methods can help avoid casting and improve type safety.

I don't find any documents which says generics is bad  and we should not use , rather I find useful to use that when we have lots of similar code to make it consistent.

src/main/java/org/opensearch/knn/index/vectorvalues/KNNVectorValuesIterator.java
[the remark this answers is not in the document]
navneet1v: This comment should resolve your concern: https://github.com/opensearch-project/k-NN/pull/1897/files#r1697233708

src/main/java/org/opensearch/knn/index/vectorvalues/KNNFloatVectorValues.java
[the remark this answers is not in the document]
Vikasht34: We should always emphasis on best code guildlines which is already define. Like 
1. Every Class should have one task to do.
2. Class should be open for extension 
3. We should not repeat our self , if we are then we need to find abstactions.

and if we are not following the same ,then we should justify why?

src/main/java/org/opensearch/knn/index/vectorvalues/KNNVectorValuesIterator.java
```diff
@@ -0,0 +1,157 @@
...
+
+        @Override
+        public long liveDocs() {
+            if (docIdSetIterator instanceof BinaryDocValues) {
+                return KNNCodecUtil.getTotalLiveDocsCount((BinaryDocValues) docIdSetIterator);
+            } else if (docIdSetIterator instanceof FloatVectorValues || docIdSetIterator instanceof ByteVectorValues) {
+                return docIdSetIterator.cost();
+            }
+            throw new IllegalArgumentException(
+                "DocIdSetIterator present is not of valid type. Valid types are: BinaryDocValues, FloatVectorValues and ByteVectorValues"
```
shatejas, line 105: I think the validation should be in the constructor. This might force us to validate in other methods or logic being added in this class. The earlier the validation the better. Let me know what you think

src/main/java/org/opensearch/knn/index/vectorvalues/KNNByteVectorValues.java
```diff
@@ -0,0 +1,63 @@
...
+    @SuppressWarnings("unchecked")
+    public byte[] getVector() throws IOException {
+        final byte[] vector;
+        if (vectorValuesIterator instanceof KNNVectorValuesIterator.DocIdsIteratorValues) {
+            final DocIdSetIterator docIdSetIterator = vectorValuesIterator.getDocIdSetIterator();
+            if (docIdSetIterator instanceof BinaryDocValues) {
+                final BinaryDocValues values = (BinaryDocValues) docIdSetIterator;
+                final BytesRef bytesRef = values.binaryValue();
+                vector = ArrayUtil.copyOfSubArray(bytesRef.bytes, bytesRef.offset, bytesRef.offset + bytesRef.length);
+                dimension = vector.length;
```
vibrantvarun, line 44: I see repetition of this part of code in below if and else. Can we create a method here to reduce repetition?

src/main/java/org/opensearch/knn/index/vectorvalues/KNNByteVectorValues.java
```diff
@@ -0,0 +1,63 @@
...
+ */
+@ToString(callSuper = true)
+public class KNNByteVectorValues extends KNNVectorValues<byte[]> {
+    KNNByteVectorValues(KNNVectorValuesIterator vectorValuesIterator) {
+        super(vectorValuesIterator);
+    }
+
+    @Override
+    @SuppressWarnings("unchecked")
+    public byte[] getVector() throws IOException {
```
vibrantvarun, line 36: As this is a public method, Can we add method definition above?

src/main/java/org/opensearch/knn/index/vectorvalues/KNNByteVectorValues.java
```diff
@@ -0,0 +1,63 @@
...
+        final byte[] vector;
+        if (vectorValuesIterator instanceof KNNVectorValuesIterator.DocIdsIteratorValues) {
```
vibrantvarun, line 38: Can we add a one liner comment on explaining how does functionality changes in each if and else case?

Like what happens if docIdSetIterator is an instance of BinaryDocValues.

src/main/java/org/opensearch/knn/index/vectorvalues/KNNFloatVectorValues.java
```diff
@@ -0,0 +1,68 @@
...
+    @Override
+    @SuppressWarnings("unchecked")
+    public float[] getVector() throws IOException {
+        final float[] vector;
+        if (vectorValuesIterator instanceof KNNVectorValuesIterator.DocIdsIteratorValues) {
+            final DocIdSetIterator docIdSetIterator = vectorValuesIterator.getDocIdSetIterator();
+            if (docIdSetIterator instanceof BinaryDocValues) {
+                final BinaryDocValues values = (BinaryDocValues) docIdSetIterator;
+                vector = getFloatVectorFromByteRef(values.binaryValue());
+                dimension = vector.length;
```
vibrantvarun, line 42: Please apply DRY principle for this part of code.

src/main/java/org/opensearch/knn/index/vectorvalues/KNNFloatVectorValues.java
```diff
@@ -0,0 +1,68 @@
...
+                "VectorValuesIterator is not of a valid type. Valid Types are: " + "BinaryDocValues and FloatVectorValues"
+            );
+        } else if (vectorValuesIterator instanceof KNNVectorValuesIterator.FieldWriterIteratorValues) {
+            final KNNVectorValuesIterator.FieldWriterIteratorValues<float[]> fieldWriterIteratorValues =
+                (KNNVectorValuesIterator.FieldWriterIteratorValues<float[]>) vectorValuesIterator;
+            vector = fieldWriterIteratorValues.vectorsValue();
+            dimension = vector.length;
+            return vector;
+        }
+        throw new IllegalArgumentException(
```
vibrantvarun, line 59: Can we early return when this condition is hit?

src/main/java/org/opensearch/knn/index/vectorvalues/KNNFloatVectorValues.java
vibrantvarun: Can we think of segregation these if elses to different methods which represents the functionality. This way we can reduce clumsiness in this method and achieve better readabitlity.

src/main/java/org/opensearch/knn/index/vectorvalues/KNNVectorValues.java
```diff
@@ -0,0 +1,90 @@
...
+
+import java.io.IOException;
+
+/**
+ * An abstract class to iterate over KNNVectors, as KNNVectors are stored as different representation like
+ * {@link BinaryDocValues}, {@link FloatVectorValues}, {@link ByteVectorValues}, {@link KnnFieldVectorsWriter} etc.
+ * @param <T>
+ */
+@ToString
+public abstract class KNNVectorValues<T> {
```
vibrantvarun, line 28: Can all methods defined in this class can be package-private?

src/main/java/org/opensearch/knn/index/vectorvalues/KNNVectorValuesFactory.java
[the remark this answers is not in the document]
heemin32: You can compile following code as well and you will find out that it break somewhere during run time.
```
KNNVectorValues<float[]> floatVectorsWithMap = KNNVectorValuesFactory.getVectorValues(VectorType.BINARY, docsWithFieldSet, floatVectorMap);
```

I am not saying generic is bad. However, you need to use is properly for it to be good. Returning ? does not help to avoid type casting and improve type safety.

src/main/java/org/opensearch/knn/index/vectorvalues/KNNVectorValuesIterator.java
```diff
@@ -0,0 +1,157 @@
...
+        @Override
+        public int nextDoc() throws IOException {
+            return docIdSetIterator.nextDoc();
+        }
+
+        /**
+         * Returns a Map of docId and vector.
+         * @return {@link Map}
+         */
+        public T vectorsValue() {
```
shatejas, line 142: I like this, was wondering if we can have something similar for `DocIdsIteratorValues`, we probably will need separate implementations for float and byte there but might simplify the logic in KNNVectorValues end

src/main/java/org/opensearch/knn/index/vectorvalues/KNNVectorValuesFactory.java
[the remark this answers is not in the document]
vibrantvarun: > ```
> public static <T> KNNVectorValues<T> getVectorValues(VectorType vectorType, DocIdSetIterator docIdSetIterator) {
>     KNNVectorValuesIterator vectorValuesIterator = new KNNVectorValuesIterator.DocIdsIteratorValues(docIdSetIterator);
> 
>     switch (vectorType) {
>         case FLOAT:
>             return (KNNVectorValues<T>) new KNNFloatVectorValues(vectorValuesIterator);
>         case BYTE:
>             return (KNNVectorValues<T>) new KNNByteVectorValues(vectorValuesIterator);
>         case BINARY:
>             return (KNNVectorValues<T>) new KNNBinaryVectorValues(vectorValuesIterator);
>         default:
>             throw new IllegalArgumentException("Unknown vector type: " + vectorType);
>     }
> }
> 
> public static <T> KNNVectorValues<T> getVectorValues(VectorType vectorType, DocsWithFieldSet docIdWithFieldSet, Map<Integer, T> vectors) {
>     KNNVectorValuesIterator vectorValuesIterator = new KNNVectorValuesIterator.FieldWriterIteratorValues<>(docIdWithFieldSet, vectors);
> 
>     switch (vectorType) {
>         case FLOAT:
>             return (KNNVectorValues<T>) new KNNFloatVectorValues(vectorValuesIterator);
>         case BYTE:
>             return (KNNVectorValues<T>) new KNNByteVectorValues(vectorValuesIterator);
>         case BINARY:
>             return (KNNVectorValues<T>) new KNNBinaryVectorValues(vectorValuesIterator);
>         default:
>             throw new IllegalArgumentException("Unknown vector type: " + vectorType);
>     }
> }
> ```

@Vikasht34 in your implementation why are we repeating the code mentioned below? A simple method getVectorType which has method definition mentioned below will solve the purpose.
> ```
>         case FLOAT:
>             return (KNNVectorValues<T>) new KNNFloatVectorValues(vectorValuesIterator);
>         case BYTE:
>             return (KNNVectorValues<T>) new KNNByteVectorValues(vectorValuesIterator);
>         case BINARY:
>             return (KNNVectorValues<T>) new KNNBinaryVectorValues(vectorValuesIterator);
>         default:
>             throw new IllegalArgumentException("Unknown vector type: " + vectorType);
> ```

vibrantvarun (commented):
Did the first round of review.

src/main/java/org/opensearch/knn/index/vectorvalues/KNNFloatVectorValues.java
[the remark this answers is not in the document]
navneet1v: In terms of extracting the vector extractor code out, I would go ahead and change it. I was trying to hold it off, thinking that we will not have more than 2 ways to extract the vector. But since there has been a lot of discussion already happened on this piece of code I think making the change is better to avoid long discussion.

src/main/java/org/opensearch/knn/index/vectorvalues/KNNVectorValuesFactory.java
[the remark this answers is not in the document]
navneet1v: by seeing all the comments I will go ahead and first try this:

```
KNNVectorValues<float[]> x = KNNVectorValuesFactory.getFloatVectorValue(docsWithFieldSet, floatVectorMap)
vs
KNNVectorValues<float[]> floatVectorsWithMap = KNNVectorValuesFactory.getVectorValues(VectorType.FLOAT, docsWithFieldSet, floatVectorMap);
```
I am not particularly happy with sending VectorDataType in params, but since its not a one way door, and very much internal to k-NN hence I will change it to use switch, only if we can avoid `?` in the method signature.

src/main/java/org/opensearch/knn/index/vectorvalues/KNNVectorValuesIterator.java
navneet1v: @shatejas in terms of logic simplification, there are already some good suggestion added, let me try that first and how code comes out. 
on your suggestion I was in a lot of dilemma whether to give Iterator the interface of getVector and I decided not to because I wanted iterator to just iterating over the docIds. But I can give another thought on this.

src/main/java/org/opensearch/knn/index/vectorvalues/KNNVectorValuesIterator.java
navneet1v: I agree that validation can be moved to the constructor, but even in that case this exception will be present that is curse of `if/else`. we need to have a else condition. Lets say even if I move this logic via Map<itr, functional interface> I would still need to throw the exception there as the output of map can be null for an iterator which is not present in map.

src/main/java/org/opensearch/knn/index/vectorvalues/KNNByteVectorValues.java
navneet1v: All the java docs are added on the base class. Hence not adding them on the child classes.

src/main/java/org/opensearch/knn/index/vectorvalues/KNNFloatVectorValues.java
navneet1v: actually if you look closely it looks similar but they are not. Nevertheless the code has been refactored now, which should solve your concern too.

src/main/java/org/opensearch/knn/index/vectorvalues/KNNVectorValues.java
navneet1v: No they cannot be package private because this KNNVectorValues will be used by other packages to iterate over the vectors.

src/main/java/org/opensearch/knn/index/vectorvalues/KNNByteVectorValues.java
navneet1v: I think the code is self explanatory there. Just try to take a look 1 more time and see if you still feel the same.

jmazanec15 (dismissed):
From functionality perspective, I think it looks good. One minor comment, but can probably be handled in another review.

src/main/java/org/opensearch/knn/index/vectorvalues/KNNBinaryVectorValues.java
```diff
@@ -0,0 +1,38 @@
+/*
```
heemin32, line 1: Could you update this licensing?
```
/*
 */
```

src/main/java/org/opensearch/knn/index/vectorvalues/KNNBinaryVectorValues.java
```diff
@@ -0,0 +1,38 @@
...
+import org.apache.lucene.codecs.KnnFieldVectorsWriter;
+import org.apache.lucene.index.BinaryDocValues;
+import org.apache.lucene.index.ByteVectorValues;
+
+/**
+ * Concrete implementation of {@link KNNVectorValues} that returns byte[] as vector where binary vector is stored and
+ * provides an abstraction over {@link BinaryDocValues}, {@link ByteVectorValues}, {@link KnnFieldVectorsWriter} etc.
+ */
+@ToString(callSuper = true)
+public class KNNBinaryVectorValues extends KNNByteVectorValues {
```
heemin32, line 24: Personally, it looks better to understand the code by making KNNBinaryVectorValues in the same hierarchical level with other Byte and Float even if we have to repeat the getVector() method. However, I will just leave it to you to make a final decision. Avoiding a repetition of code is not a thing we should pursue the most imo.

```suggestion
public class KNNBinaryVectorValues extends KNNVectorValues<byte[]> {
```

src/main/java/org/opensearch/knn/index/vectorvalues/KNNVectorValues.java
```diff
@@ -0,0 +1,90 @@
...
+     * Return a vector reference. If you are adding this address in a List/Map ensure that you are copying the vector first.
+     * This is to ensure that we keep the heap and latency in check by reducing the copies of vectors.
+     *
+     * @return T an array
+     * @throws IOException if we are not able to get the vector
+     */
+    public abstract T getVector() throws IOException;
+
+    /**
+     * Dimension of vector is returned. Do call getVector function first before calling this function otherwise you will get 0 value.
```
heemin32, line 47: Would be nice if we can avoid this restriction. If getVector function is not called, we should call it before returning dimension.

src/main/java/org/opensearch/knn/index/vectorvalues/KNNVectorValuesFactory.java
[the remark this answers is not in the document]
heemin32: I am not happy with this change becase there is no difference between `?` and `T` but won't push further if it is already decided.

src/main/java/org/opensearch/knn/index/vectorvalues/VectorValueExtractorStrategy.java
```diff
@@ -0,0 +1,96 @@
...
+interface VectorValueExtractorStrategy {
+    /**
+     * Extract Vector based on the vector datatype and vector values iterator.
+     * @param vectorDataType {@link VectorDataType}
+     * @param vectorValuesIterator {@link KNNVectorValuesIterator}
+     * @return vector
+     * @param <T> could be of type float[], byte[]
+     * @throws IOException exception during extracting the vectors
+     */
+    <T> T extract(VectorDataType vectorDataType, KNNVectorValuesIterator vectorValuesIterator) throws IOException;
```
heemin32, line 33: I still against this pattern here. 
It is like you have to go inside extract method and check what it returns based on your parameter. 
It would be much easier to read code by having
```
byte[] extractBinaryVector(KNNVectorValuesIterator vectorValuesIterator)
byte[] extractByteVector(KNNVectorValuesIterator vectorValuesIterator)
float[] extractFloatVector(KNNVectorValuesIterator vectorValuesIterator)
```

src/main/java/org/opensearch/knn/index/vectorvalues/KNNBinaryVectorValues.java
navneet1v: Seems like the change of license just went, after these files are created. Let me fix it now.

src/main/java/org/opensearch/knn/index/vectorvalues/KNNBinaryVectorValues.java
navneet1v: Let me try to see if I can remove code duplication by keeping them at same level. I can see BitVectorFormat at same level as the byteVectors format. I will change it.

https://github.com/apache/lucene/tree/main/lucene/codecs/src/java/org/apache/lucene/codecs/bitvectors

src/main/java/org/opensearch/knn/index/vectorvalues/KNNVectorValues.java
navneet1v: This restriction mainly comes from BinaryDocValues. As, before reading the BDV we don't know what the vector is and hence we don't know what is the dimension. For FVV, and ByteVV this is not a problem because the dimension gets serialized with VV.

If we want to fix this, we have to read a doc and put some complexity in the nextDoc functions. if we are okay with complexity on the nextDoc for BDV I can make the change. 

Another option is have a cloneable interface on BDV which we Lucene doesn't provide. So only option is have make the nextdoc function complex. What your thought on this?

@heemin32

src/main/java/org/opensearch/knn/index/vectorvalues/VectorValueExtractorStrategy.java
navneet1v: Lets keep it like this for now, but I think this way we avoid a lot of extraction methods/classes per IteratorType and then per vector datatype. As use of generics is quite new in k-NN plugin but a prominent thing in Java. Its more around getting used to in reading the code with generics

src/main/java/org/opensearch/knn/index/vectorvalues/KNNFloatVectorValues.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+public class KNNFloatVectorValues extends KNNVectorValues<float[]> {
+    KNNFloatVectorValues(final KNNVectorValuesIterator vectorValuesIterator) {
+        super(vectorValuesIterator);
+    }
+
+    @Override
+    public float[] getVector() throws IOException {
+        final float[] vector = vectorValuesIterator.getVectorExtractorStrategy().extract(VectorDataType.FLOAT, vectorValuesIterator);
+        this.dimension = vector.length;
```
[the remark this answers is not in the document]
navneet1v: since it was just a setting of int, I was not fond of making it lazily as it will add some branches. Hence I did it like this. Since for BDV, till you don't read the vector you don't know what is the dimension. It is coming from there

src/main/java/org/opensearch/knn/index/vectorvalues/VectorValueExtractorStrategy.java
heemin32: It is not about getting used to generic. Just this is not a best practice to use generic. It is better for caller to know about return type of method from its signature. Otherwise, we always have to look inside the implementation and see what will be the return type for each parameter combinations.

If we really want to use generic at least it would be better to look like
```
<T> T extract(T value, KNNVectorValuesIterator vectorValuesIterator)
```
But again, passing T value is strange because we only use it to determine return type...

src/main/java/org/opensearch/knn/index/vectorvalues/KNNVectorValues.java
heemin32: Would it be okay for caller to determine dimension using the returned vector? Does dimension have to be inside this class?

src/main/java/org/opensearch/knn/index/vectorvalues/VectorValueExtractorStrategy.java
Vikasht34: @heemin32 , VectorDataType is already getting passed in this function ...so at caller already knows what data types got passed, I don't foresee any confusion around implementation and generics use cases. 

I slightly disagree on this 

```
<T> T extract(T value, KNNVectorValuesIterator vectorValuesIterator)
```

Here are my reasons 

1. The use of <T> T in this method signature means that the type T is determined by the caller. However, without a clear relationship between T and the KNNVectorValuesIterator or the data being extracted, the method becomes ambiguous. There's no guarantee that the type of T matches the type of data the iterator can provide.
2. The parameter T value doesn't seem to serve a purpose in the context of an extraction operation. Typically, an extractor method would only require the source of data (in this case, KNNVectorValuesIterator) and any necessary metadata (like data type). The value parameter might confuse users, as it's unclear why they would provide an instance of T when they expect to receive one , on the other hand it is getting up fronted provided vector data type.

heemin32: `VectorDataType is already getting passed in this function` but who enforce it and which type does each VectorDataType map to? You know that because you are familiar with the inside code. 

Let's say I have a method `<T> T getSomthing(VectorDataType type, KNNVectorValuesIterator iterator, SomeotherClass class);` Can you guess what the T will be for each parameter?

Vikasht34: @heemin32  Yes this is how :- 

```
KNNVectorValues<float[]> floatVectorsWithMap = KNNVectorValuesFactory.getVectorValues(VectorType.FLOAT, docsWithFieldSet, floatVectorMap);
```

src/main/java/org/opensearch/knn/index/vectorvalues/VectorValueExtractorStrategy.java
```diff
@@ -0,0 +1,96 @@
...
+     */
+    class DISIVectorExtractor implements VectorValueExtractorStrategy {
+        @Override
+        public <T> T extract(final VectorDataType vectorDataType, final KNNVectorValuesIterator vectorValuesIterator) throws IOException {
+            final DocIdSetIterator docIdSetIterator = vectorValuesIterator.getDocIdSetIterator();
+            switch (vectorDataType) {
+                case FLOAT:
+                    if (docIdSetIterator instanceof BinaryDocValues) {
+                        final BinaryDocValues values = (BinaryDocValues) docIdSetIterator;
+                        return (T) getFloatVectorFromByteRef(values.binaryValue());
```
heemin32, line 46: Can we write a unit test calling this code part? This code part might throw `ClassCastException`

src/main/java/org/opensearch/knn/index/vectorvalues/KNNVectorValues.java
navneet1v: > Would it be okay for caller to determine dimension using the returned vector? Does dimension have to be inside this class?

So for binary vector they need to identify the type of vector and then use the dimension. Having it like this make it easy to use.

heemin32: I think we can enforce it.
Have method 
```
public int getDimension(T vector) {} 
```
Then, the caller should provide vector to know about dimension.

navneet1v: This is an interesting idea. Let me do this.

src/main/java/org/opensearch/knn/index/vectorvalues/VectorValueExtractorStrategy.java
navneet1v: this is already covered by the UTs present here: KNNVectorValuesTests. If you see I have added tests for types of vectors, byte[], binary and float[]. so this casting case is covered.

Vikasht34 (approved):
LGTM

src/main/java/org/opensearch/knn/index/vectorvalues/KNNVectorValues.java
navneet1v: @heemin32 the problem with this approach is I have to write the implementation of get dimension in each that extends KNNVectorValues. Because T.length is not supported. Do you know if there a way to find the length of T

src/main/java/org/opensearch/knn/index/vectorvalues/VectorValueExtractorStrategy.java
navneet1v: I think I got it what you mean.  I will add a test for exception.

src/main/java/org/opensearch/knn/index/vectorvalues/KNNVectorValues.java
heemin32: I don't think there is a way to do that. I would like to say this is why generic is not used properly here though not sure if you agree on it :)

src/main/java/org/opensearch/knn/index/vectorvalues/VectorValueExtractorStrategy.java
heemin32: My bad. Never mind. I thought (T) casting is not allowed but it was not..

navneet1v: I thought you were asking for the last error condition. So I added tests for that.

src/main/java/org/opensearch/knn/index/vectorvalues/KNNVectorValues.java
navneet1v: > I don't think there is a way to do that. I would like to say this is why generic is not used properly here

See this cannot be the reason, because if that is the case then T as generic will never be used then. See this problem of dimension doesn't come if we are using FloatVector Values or ByteVectorValues. So the problem is not with generics, its just we are expecting the dimension before even reading a vector. Its like calling a `docId()` on the Iterator without calling `nextDoc()`. I just added the comment in java doc to ensure that caution is taken.

heemin32: If we use FloatVectorValue and ByteVectorValues, the method will be look like `public int getDimension(float[] vector) {}` `public int getDimension(byte[] vector) {}` each.

heemin32: Or maybe, it is happening because float[] and byte[] is primitive array, and not List<Float> and List<Byte>.

heemin32: IMO, it is fine to have 
```
public abstract int getDimensions(T vector) throws IOException;
```
and let child class to implement it.

navneet1v: @heemin32 on thinking more on this, its better to align the interface with `values` interface of Lucene. Hence I am keeping the dimension interface free from `vector` param, but I am not shutting door on the idea of sending vector to get the dimension. 

So for this PR I am adding some validation upfront, which will ensure improper use of interface will throw errors. In the next PR, let me take a stab at completely getting rid of these validations by implementing the dimension logic for BDV. 


Resolving this comment. feel free to un-resolve the comment if you think otherwise. :)

src/main/java/org/opensearch/knn/index/vectorvalues/KNNFloatVectorValues.java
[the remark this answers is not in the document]
navneet1v: @jmazanec15 I added some details on how to resolve this concern here: https://github.com/opensearch-project/k-NN/pull/1897#discussion_r1699570249 please let me know if this an acceptable approach here.

src/main/java/org/opensearch/knn/index/vectorvalues/KNNVectorValuesIterator.java
navneet1v: Added the validations in constructor too.

jmazanec15 (approved):

PR #1900, opened 2024-07-30
[DO NOT MERGE][NEED INITIAL FEEDBACK] Adds NativeEngineKnnFloatVectorQuery

shatejas:
### Description
This approach extends Lucene's KNNFloatVectorQuery. Testing in progress

There are some differences and its hard to achieve status quo with current implementation without overriding rewrite or making changes in lucene
Differences
- We loose the [current exact search ](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/query/KNNWeight.java#L497)setting we provide to our customers. This is a regression

### Related Issues

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/query/KNNQueryFactory.java
+++ b/src/main/java/org/opensearch/knn/index/query/KNNQueryFactory.java
@@ -16,6 +16,7 @@ import org.apache.lucene.search.join.DiversifyingChildrenFloatKnnVectorQuery;
 import org.opensearch.index.query.QueryShardContext;
 import org.opensearch.knn.index.VectorDataType;
 import org.opensearch.knn.index.util.KNNEngine;
+import org.opensearch.knn.search.NativeEngineKnnFloatVectorQuery;
 
 import java.util.Locale;
 import java.util.Map;
@@ -100,27 +101,31 @@ public class KNNQueryFactory extends BaseQueryFactory {
 
             switch (vectorDataType) {
                 case BINARY:
-                    return KNNQuery.builder()
-                        .field(fieldName)
-                        .byteQueryVector(byteVector)
-                        .indexName(indexName)
-                        .parentsFilter(parentFilter)
-                        .k(k)
-                        .methodParameters(methodParameters)
-                        .filterQuery(validatedFilterQuery)
-                        .vectorDataType(vectorDataType)
-                        .build();
+                    return new NativeEngineKnnFloatVectorQuery(
+                        KNNQuery.builder()
+                            .field(fieldName)
+                            .byteQueryVector(byteVector)
+                            .indexName(indexName)
+                            .parentsFilter(parentFilter)
+                            .k(k)
+                            .methodParameters(methodParameters)
+                            .filterQuery(validatedFilterQuery)
+                            .vectorDataType(vectorDataType)
+                            .build()
+                    );
                 default:
-                    return KNNQuery.builder()
-                        .field(fieldName)
-                        .queryVector(vector)
-                        .indexName(indexName)
-                        .parentsFilter(parentFilter)
-                        .k(k)
-                        .methodParameters(methodParameters)
-                        .filterQuery(validatedFilterQuery)
-                        .vectorDataType(vectorDataType)
-                        .build();
+                    return new NativeEngineKnnFloatVectorQuery(
+                        KNNQuery.builder()
+                            .field(fieldName)
+                            .queryVector(vector)
+                            .indexName(indexName)
+                            .parentsFilter(parentFilter)
+                            .k(k)
+                            .methodParameters(methodParameters)
+                            .filterQuery(validatedFilterQuery)
+                            .vectorDataType(vectorDataType)
+                            .build()
+                    );
             }
         }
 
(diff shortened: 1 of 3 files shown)


src/main/java/org/opensearch/knn/search/NativeEngineKnnFloatVectorQuery.java
```diff
@@ -0,0 +1,112 @@
...
+        if (leafDocScores == null) {
+            leafDocScores = Collections.emptyMap();
+        }
+        final Bits liveDocs = context.reader().getLiveDocs();
+
+        final List<Map.Entry<Integer, Float>> topScores = new ArrayList<>(leafDocScores.entrySet());
+        topScores.sort(Map.Entry.<Integer, Float>comparingByValue().reversed());
+
+        //This is to trick the implementation to force exactsearch after approxsearch
+        TotalHits.Relation relation = query.getFilterQuery() != null
```
shatejas, line 73: I will have to push this later as this does not consider deleted docs. will do it in next iteration

src/main/java/org/opensearch/knn/index/query/KNNQueryFactory.java
```diff
@@ -100,27 +101,31 @@ public static Query create(CreateQueryRequest createQueryRequest) {
...
-                        .field(fieldName)
-                        .byteQueryVector(byteVector)
-                        .indexName(indexName)
-                        .parentsFilter(parentFilter)
-                        .k(k)
-                        .methodParameters(methodParameters)
-                        .filterQuery(validatedFilterQuery)
-                        .vectorDataType(vectorDataType)
-                        .build();
+                    return new NativeEngineKnnFloatVectorQuery(
```
shatejas, line 104: There will be a feature flag here

src/main/java/org/opensearch/knn/search/NativeEngineKnnFloatVectorQuery.java
```diff
@@ -0,0 +1,112 @@
...
+import org.opensearch.knn.index.query.KNNQuery;
+import org.opensearch.knn.index.query.KNNWeight;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+
+public class NativeEngineKnnFloatVectorQuery extends KnnFloatVectorQuery {
```
navneet1v, line 34: if I am not wrong, KnnFloatVectorQuery just checks the FiltersIds<=K to do the exact search but not the other condition like AdvancedFilterThreshold. Thats a bumer.

Also, currently the way ParentIdsFilters and radial search work for Lucene VectorQuery and that of NativeEngine query is different.


Another thing, by extending KnnFloatVectorQuery, we have just reused the DocAndScore query if I am not wrong. Anything else that we reused here which I am missing?

src/main/java/org/opensearch/knn/search/NativeEngineKnnFloatVectorQuery.java
```diff
@@ -0,0 +1,112 @@
...
+                totalHits++;
+            }
+        }
+
+        return new TopDocs(new TotalHits(totalHits, relation), scoreDocs.toArray(ScoreDoc[]::new));
+    }
+
+    @Override
+    protected TopDocs mergeLeafResults(TopDocs[] perLeafResults) {
+        // Rescore here
```
jmazanec15, line 109: Issue with re-scoring here is we need access to the LeafReaderContext for each set of results to lookup the vectors. So, not sure we can re-score here. 

Instead, I think we will need to re-score from rewrite. If we want to share functionality, I think we could call super.rewrite(), get the resulting Query, and then rescore. Otherwise, we should just override the rewrite method.

jmazanec15:
I dont think we can get rid of that setting without following deprecation path. Why is it blocked? If we cant get around it, I think this is a non-starter then.

src/main/java/org/opensearch/knn/search/NativeEngineKnnFloatVectorQuery.java
shatejas: > if I am not wrong, KnnFloatVectorQuery just checks the FiltersIds<=K to do the exact search but not the other condition like AdvancedFilterThreshold. Thats a bumer.

Correct, losing support for a setting. so now backward compatible

> Also, currently the way ParentIdsFilters and radial search work for Lucene VectorQuery and that of NativeEngine query is different.

- parentIds: I missed this one, nice catch
- radial: this needs some thought so didn't plug it in here

> Anything else that we reused here which I am missing?

Yeah DocAndScore and some boiler plate filtering

src/main/java/org/opensearch/knn/search/NativeEngineKnnFloatVectorQuery.java
shatejas: Ah I see there is not way to cleanly rescore in this approach then

shatejas:
> I dont think we can get rid of that setting without following deprecation path. Why is it blocked? If we cant get around it, I think this is a non-starter then.

Its blocked we are trying to use lucenes rewrite method which has a simpler condition and no way to extend

jmazanec15:
Is it worth extending AbstractKnnFloatQuery and overwriting everything just for sake of not having to copy DocAndScoreQuery code? 

One comment around setting: in future, if we move towards implementing codec reader search methods, we could pass the setting to the codec if its an index setting. But, thats out of scope for this.

PR #1908, opened 2024-07-30
Refactor package containing engine

jmazanec15:
### Description
Moves engine components from util to engine and makes subpackages for individual engines. Collect utils into util package. Move QueryContext into query package.

Overall a lot of changes, but no real functional changes. Benefit of this is (1) more descript naming of packages (2) better ability to separate between engines. Hopefully, at some point, we will be able to move all JNI components to individual engines.

One downside of this is that if another project is using any of these interfaces, we will have to change. Another downside will be that it will require everyone to rebase.

I wanted feedback from team on whether this change is worth it. I think its an incremental step in organizing the code better and separating the engine specific functionality into its own module. Its always annoyed me that the util package contains all of the engine functionality, but it hasnt been changed because itll touch so many files. 

### Check List


jmazanec15:
Im going to wait until after some more changes are in to refactor package like this.

PR #1913, opened 2024-07-31
Refactor engine package structure

jmazanec15:
### Description
Refactors package structure for engines:
1. First, it renames the package from util to engine
2. Next, it moves actual utils to the util package
3. Lastly, it puts the engines (faiss, lucene, nmslib) in separate sub-packages. Purpose for this is to separate between engines.

In order to support quantization framework, some refactoring needs to be done. This is the first step. After this, I am planning on changing up the KNNMethod interface and how Methods are defined for engines. 

Keeping scope of this PR limited to change of package only, just because it touches so many files. For comments that are not related to package change, we can address in upcoming PRs.

Ill make sure neural-search is updated as well.

### Related Issues
Previous PR I opened then closed yesterday: #1908. I changed my mind and think it would be good to do this refactor upfront. 

### Check List


heemin32 (dismissed):

qa/restart-upgrade/src/test/java/org/opensearch/knn/bwc/FaissSQIT.java
```diff
@@ -20,10 +20,10 @@
 import org.opensearch.core.xcontent.XContentBuilder;
 import org.opensearch.knn.KNNResult;
 import org.opensearch.knn.common.KNNConstants;
-import org.opensearch.knn.index.KNNMethod;
+import org.opensearch.knn.index.engine.KNNMethod;
 import org.opensearch.knn.index.SpaceType;
 import org.opensearch.knn.index.query.KNNQueryBuilder;
-import org.opensearch.knn.index.util.KNNEngine;
+import org.opensearch.knn.index.engine.KNNEngine;
```
shatejas, line 26: Why does it still have to be under index package?

src/main/java/org/opensearch/knn/index/engine/QueryContext.java
```diff
@@ -0,0 +1,17 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.knn.index.engine;
```
shatejas, line 6: create another package `engine.model` and have all the request pojos in there

src/main/java/org/opensearch/knn/index/engine/LuceneHNSWContext.java
```diff
@@ -1,18 +1,11 @@
...
- * Modifications Copyright OpenSearch Contributors. See
- * GitHub history for details.
  */
 
-package org.opensearch.knn.index.util;
+package org.opensearch.knn.index.engine;
 
 import com.google.common.collect.ImmutableMap;
-import org.opensearch.knn.index.Parameter;
 import org.opensearch.knn.index.query.request.MethodParameter;
```
shatejas, line 10: This file needs to be under lucene if we are going with one package per engine

src/main/java/org/opensearch/knn/index/engine/MethodComponent.java
```diff
@@ -1,15 +1,9 @@
...
- * The OpenSearch Contributors require contributions made to
- * this file be licensed under the Apache-2.0 license or a
- * compatible open source license.
- *
- * Modifications Copyright OpenSearch Contributors. See
- * GitHub history for details.
  */
 
-package org.opensearch.knn.index;
+package org.opensearch.knn.index.engine;
```
shatejas, line 6: Not engine specific, need a better place for this, this holds request parameters for indexing

src/main/java/org/opensearch/knn/index/engine/MethodComponentContext.java
```diff
@@ -1,15 +1,9 @@
...
- * The OpenSearch Contributors require contributions made to
- * this file be licensed under the Apache-2.0 license or a
- * compatible open source license.
- *
- * Modifications Copyright OpenSearch Contributors. See
- * GitHub history for details.
  */
 
-package org.opensearch.knn.index;
+package org.opensearch.knn.index.engine;
```
shatejas, line 6: Same with this, not engine specific

src/main/java/org/opensearch/knn/index/engine/KNNMethodContext.java
```diff
@@ -20,7 +14,7 @@
 import org.opensearch.core.common.io.stream.StreamInput;
 import org.opensearch.core.common.io.stream.StreamOutput;
 import org.opensearch.core.common.io.stream.Writeable;
-import org.opensearch.knn.index.util.KNNEngine;
+import org.opensearch.knn.index.SpaceType;
```
shatejas, line 17: Again, not engine specific

shatejas (commented):
> Keeping scope of this PR limited to change of package only, just because it touches so many files.

The scope of the refactor doesn't seem to be clear. this PR is package related but to what extent are things being moved around, are they tied to specific release?

I wouldn't want this in main until its complete. Does it make more sense to push this in quantization framework branch since they seem to be related from description?

heemin32:
> The scope of the refactor doesn't seem to be clear. this PR is package related but to what extent are things being moved around, are they tied to specific release?

> I wouldn't want this in main until its complete. Does it make more sense to push this in quantization framework branch since they seem to be related from description?

As long as we are sure that there won't be a complete new package layout change again during quantization implementation, I think this change is harmless. Doing this early in main will reduce merging effort compared to do it later.

qa/restart-upgrade/src/test/java/org/opensearch/knn/bwc/FaissSQIT.java
heemin32: I think `org.opensearch.knn.index` is base package for knn index in general. I guess we can move it one more up `org.opensearch.knn.engne` as @shatejas mentioned. What do you think @jmazanec15 ?

jmazanec15: Thats a good question. Its been mostly convention to have index apply to Lucene index/shard related operations in OpenSearch (see https://github.com/opensearch-project/OpenSearch/tree/main/server/src/main/java/org/opensearch/index). I could move it out, but might be better to consider later as engine is pretty shard specific.

src/main/java/org/opensearch/knn/index/engine/QueryContext.java
jmazanec15: sure

src/main/java/org/opensearch/knn/index/engine/LuceneHNSWContext.java
jmazanec15: Makes sense. Will move

src/main/java/org/opensearch/knn/index/engine/MethodComponent.java
jmazanec15: I think with the current hierarchy it is specific to Engine. Engines are composed of KNNMethods wqhich are composed of MethodComponents which are composed of Parameters. I want to move towards encapsulating details around methods behind engines. Hence, I think it makes sense for them to be in the engine package.

src/main/java/org/opensearch/knn/index/engine/MethodComponentContext.java
jmazanec15: Lets discuss on https://github.com/opensearch-project/k-NN/pull/1913#discussion_r1698810135

src/main/java/org/opensearch/knn/index/engine/KNNMethodContext.java
jmazanec15: Lets discuss on https://github.com/opensearch-project/k-NN/pull/1913#discussion_r1698810135

jmazanec15:
> The scope of the refactor doesn't seem to be clear. this PR is package related but to what extent are things being moved around, are they tied to specific release?

I wanted to address the following in this PR:
1. util is a bad package for all of the engine components to be in. I wanted to move the engine components to their own package. Then, keep the utils package for actual utilities. Better defining these constructs will make it easier to develop on top of
2. Putting engines in separate packages. This is an incremental step in separating the engines, which I think from a long term maintainability perspective we are going to want to do more and more. 
3. Move KNNMethod and MethodComponents into engine. I forgot to mention this in description. These are the components that define the methods the engines support. So, I think it makes sense to define them in the engine. Yes, they may be used elsewhere, but I think they are a central part of the engine construct.

I didnt include any other code in this PR just to keep focus narrow - functionality should not change.

In general, my goal is to make incremental changes to improve structure of code.

src/main/java/org/opensearch/knn/index/engine/MethodComponent.java
shatejas: > I want to move towards encapsulating details around methods behind engines.

I took a deeper look. MethodComponent encapsulates properties of engine in a way and the context is the one which is a request object. So MethodComponent makes sense to be in the engine but MethodComponentContext doesn't because it holds the request given by the user, that isn't related to engine. 

So I would say anything suffixed `Context` should not be in engine as it has parsing logic of user request and then holds the information to create index 

P.S we need naming which doesn't cause confusion

shatejas:
> In general, my goal is to make incremental changes to improve structure of code.

+1 on this, the repo needs better package structure. My concern is more over reverts if we have to do it. 1 commit to revert is better than multiple

qa/restart-upgrade/src/test/java/org/opensearch/knn/bwc/FaissSQIT.java
shatejas: I will leave this up to you. 

The concern is majority of code inside index package. For instance, KNNQueryBuilder which isn't shard specific goes under index

jmazanec15:
> +1 on this, the repo needs better package structure. My concern is more over reverts if we have to do it. 1 commit to revert is better than multiple

Right, if this change had to be reverted it would be a pain for everyone after. I think this change is pretty low-risk because functionality is not changing just package structure. Issues will be caught at compile time. So I think that there is a very small chance we have to revert as opposed to making a minor fix.

src/main/java/org/opensearch/knn/index/engine/MethodComponent.java
jmazanec15: I initially wrote the KNNMethodContext to be user configuration of KNNMethod and similarly for MethodComponentContext and MethodComponent. So KNNMethod and MethodComponent define the structure and KNNMethodContext and MethodComponentContext provide the configuration. They have roughly followed this, but not completely. That being said, do you have recommendation for which package they should be in? It seemed to me that for now they could be in the same package because of this parallel.

qa/restart-upgrade/src/test/java/org/opensearch/knn/bwc/FaissSQIT.java
jmazanec15: Yes, that isnt great. I think this is probably something that can be addressed later in separate PR refactoring index package.

heemin32 (approved):

src/main/java/org/opensearch/knn/index/engine/MethodComponent.java
shatejas: My preference would keep it outside `engine` to start with. `index.request` might be a good place

shatejas (approved):

Issue #1889, opened 2024-07-29
[RFC] Quantization Framework For Disk Optimized Vector Search


Vikasht34:
# Introduction
This document proposes a general-purpose framework for vector quantization. This framework aims to utilize various quantization techniques, including product quantization, binary quantization, and scalar quantization. The goal is to compress floating-point vectors into bytes or bits, significantly reducing memory usage for storing and processing vector data in semantic search applications.

---

## Requirements

| Requirement ID | Description
| -------------- | ----------- |
| R-1            | Effective Binary Quantizer for both float and byte vectors | 
| R-2            | Extensible framework for various quantization techniques | 
| R-3            | Memory-efficient handling of large datasets |
| R-4            | Extensible for future quantization techniques and integrations |
| R-5            | Client-selectable quantization techniques based on needs |
| R-6            | Implement sampler for training data |

---

## Success Metrics

| Metric ID | Description |
| --------- | ----------- |
| M-1       | Quantization latency should be less than 20ms |
| M-2       | Memory usage of the framework (dependent on vector dimension and quantization type) |

---

## State of the System

Currently, we have the following quantization techniques:
1. **Product Quantization**: Requires an additional training step.
2. **Byte Quantization**: Implemented through Lucene.
3. **Scalar Quantization**: Implemented through FAISS (FP16).

---

### Responsibilities of Quantizer Framework

1. **Support Multiple Quantization Types**:
   - Supports space-compression (e.g., PQ) and value-compression (e.g., Scalar Quantization) techniques.
   - Extensible for future methods like PCA or 2-bit quantization.
   - New methods can be added with minimal changes to existing code.

2. **Clear Interface for Quantization and Training**:
   - Clients can specify and configure parameters for different techniques.
   - Data sources for quantization can be specified.

3. **Efficiency**:
   - Optimize memory usage during the quantization process.

4. **Quantization Framework will sample records from training.**

5. **Stateless**:
   - Does not store quantization state or thresholds.

6. **Metrics Tracking**:
   - Time taken during training and quantization.
   - Memory consumption.

7. **Decision Making**:
   - The framework does not decide when to re-quantize; it accepts quantization states.

### Components of Quantization Framework

1. **Quantization Orchestrator**
2. **Sampler**
3. **Quantizers**
![Untitled Diagram-Page-7 (7) (1)](https://github.com/user-attachments/assets/4a71004d-558d-40e6-9ffb-8f140f32842b)

[...]

PR #1920, opened 2024-08-01
Refactor method structure and definitions

jmazanec15:
### Description
This is a refactor around KNNMethod Definitions. It pre-work for interface for #1889. Functionally, nothing is changing. The overall scope/goal is to make it easier to integrate the quantization framework functionality and also make the plugin easier to maintain in the future. 

This refactor has 4 components (I did one commit per)
1. Remove getMethod from KNNLibrary interface. The motivation here is that the KNNMethod object is an implementation detail for an engine, so it does not make sense to expose it. The functionality exposed should be around validating configurations/params, and/or providing information relevant to building and/or searching the index produced by the KNNLibrary
2. Move EngineSpecificMethodContext into KNNMethod. Motivation is that the KNNMethod is meant to encapsulate a libraries ANN configuration. This includes the parameters necessary for search.
3. Change KNNMethod to interface and implement per Method we support. This resolves the dreaded configuration complexity we have in the engines by pulling the method definitions out into their own classes (see https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/engine/faiss/Faiss.java#L217-L311). 
4. Creating basic interface for Encoders. Similar to making KNNMethod an interface and implementing per method we support, I did the same with the Encoders. In the future, this will make it a lot easy to add Quantizers from the quantization framework. 

The next steps after this is to change the KNNLibrary.getMethodAsMap to return a new Object that will both contain information to be passed to the library as well as the map needed to build the index - specifically, it will return the information needed by the quantization framework to configure the quantization. This will simplify the experience for building more complex objects. 

### Related Issues
#1889 

### Check List


navneet1v (dismissed):
@jmazanec15 code looks good to me. At some point of time I really think we need a good class diagram on Method and Library class hierarchy for easy understanding. May be you can use IDE tool to build this and paste it with the PR. See if that is possible

jmazanec15:
Sure Ill look into it - class hierarchy will help.

src/main/java/org/opensearch/knn/index/engine/faiss/FaissHNSWMethod.java
```diff
@@ -0,0 +1,93 @@
...
+    /**
+     * Constructor for FaissHNSWMethod
+     *
+     * @see AbstractKNNMethod
+     */
+    public FaissHNSWMethod() {
+        super(initMethodComponent(), Set.copyOf(SUPPORTED_SPACES), new DefaultHnswContext());
+    }
+
+    private static MethodComponent initMethodComponent() {
```
heemin32, line 57: nit. It doesn't have to be static.

jmazanec15: its called for super constructor so it ends up actually having to

src/main/java/org/opensearch/knn/index/engine/Encoder.java
```diff
@@ -0,0 +1,25 @@
...
+ * Interface representing an encoder. An encoder generally refers to a vector quantizer.
+ */
+public interface Encoder {
+    /**
+     * The name of the encoder does not have to be unique. Howevwer, when using within a method, there cannot be
+     * 2 encoders with the same name.
+     *
+     * @return Name of the encoder
+     */
+    String getName();
```
heemin32, line 18:
```suggestion
    String getName() {
      return getMEthodComponent().getName();
    }
```
and you can remove override function in all implementation class.

jmazanec15: Makes sense. Added default

heemin32 (approved):
LGTM. Thanks!

navneet1v (approved):

naveentatikonda (approved):

PR #1925, opened 2024-08-02
Generalize lib interface to return context objects

jmazanec15:
### Description
Generalizes the KNNLibrary interface to return an object for both search and indexing so that the plugin can search/index against them. The names of the methods are "getKNNLibraryIndexBuildContext" and "getKNNLibrarySearchContext". In the future, I could see these methods changing into something like getKNNLibraryIndexBuilder and getKNNLibrarySearcher so that the engine actually encapsulates the interactions with the libraries, but that needs some more thought. For now, this change makes it more clear what one of the central roles of the KNNLibrary interface is - telling the plugin how to index/search. 

For the quantization framework, making the interface return objects will allow us to properly configure both quantization as well as whatever index we want to use with it. See the following locations for more details where we would do this:
1.  https://github.com/jmazanec15/k-NN-1/blob/method-stage/src/main/java/org/opensearch/knn/index/mapper/MethodFieldMapper.java#L63
2. https://github.com/jmazanec15/k-NN-1/blob/method-stage/src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java#L376

Again, no functionality change - just more refactoring to better support the quantization framework. 

One thing I have realized from these changes is that he KNNMethodContext class needs to be sured-up a bit to be more robust. I will create a separate issue for this.

### Check List


naveentatikonda (dismissed):

heemin32:
How about `getKNNLibraryIndexingContext`?

heemin32 (approved):

src/main/java/org/opensearch/knn/index/engine/DefaultHnswContext.java
```diff
@@ -14,7 +14,7 @@
 /**
  * Default HNSW context for all engines. Have a different implementation if engine context differs.
  */
-public final class DefaultHnswContext implements EngineSpecificMethodContext {
+public final class DefaultHnswContext implements KNNLibrarySearchContext {
```
heemin32, line 17: Maybe DefaultHnswSearchContext?

src/main/java/org/opensearch/knn/index/engine/DefaultIVFContext.java
```diff
@@ -11,7 +11,7 @@
 
 import java.util.Map;
 
-public final class DefaultIVFContext implements EngineSpecificMethodContext {
+public final class DefaultIVFContext implements KNNLibrarySearchContext {
```
heemin32, line 14: Maybe DefaultIVFSearchContext?

src/main/java/org/opensearch/knn/index/engine/lucene/LuceneHNSWContext.java
```diff
@@ -6,15 +6,15 @@
...
+import org.opensearch.knn.index.engine.KNNLibrarySearchContext;
 import org.opensearch.knn.index.engine.Parameter;
 import org.opensearch.knn.index.engine.model.QueryContext;
 import org.opensearch.knn.index.query.request.MethodParameter;
 
 import java.util.Collections;
 import java.util.Map;
 
-public class LuceneHNSWContext implements EngineSpecificMethodContext {
+public class LuceneHNSWContext implements KNNLibrarySearchContext {
```
heemin32, line 17: LuceneHNSWSearchContext

jmazanec15:
Addressed @heemin32 !

src/main/java/org/opensearch/knn/index/engine/KNNLibraryIndexingContextImpl.java
```diff
@@ -0,0 +1,24 @@
...
+
+import lombok.Builder;
+
+import java.util.Map;
+
+/**
+ * Simple implementation of {@link KNNLibraryIndexBuildContext}
+ */
+@Builder
+public class KNNLibraryIndexBuildContextImpl implements KNNLibraryIndexBuildContext {
```
heemin32, line 16:
```suggestion
public class KNNLibraryIndexingContextImpl implements KNNLibraryIndexBuildContext {
```

heemin32 (approved):
LGTM.

naveentatikonda (approved):

junqiu-lei (approved):

PR #1929, opened 2024-08-03
Quantization Framework Implementation with 1bit and MultiBit Binary Quantizer

Vikasht34:
### Description
This pull request introduces a new quantization framework, including the MultiBitScalarQuantizer and OneBitScalarQuantizer, to provide efficient quantization methods for vectors in OpenSearch.

#### Key features:
* Memory Efficiency: Designed the train method in MultiBitScalarQuantizer to compute mean and standard deviation in a single pass, reducing redundant memory allocations and improving efficiency.
* Code Structure: Improved code readability and maintainability by reusing arrays for multiple calculations and employing a clear, modular approach.

#### OneBitScalarQuantizer
*  Training: Computes the mean of each dimension from sampled vectors, using these means as thresholds for quantization.
* Quantization: Compares each dimension of the vector against the corresponding mean (threshold) to determine the quantized value, resulting in a binary representation (1 bit per dimension).

#### MultiBitScalarQuantizer

* Training: Calculates mean, and standard deviation for each dimension from sampled vectors in a single pass. Determines multiple thresholds based on these bits per encoded in quantization.
* Quantization: Uses the calculated thresholds to quantize the vector into multi-bit representations per dimension, allowing for more precise quantization compared to the one-bit approach.

### Related Issues
Resolves [#1889](https://github.com/opensearch-project/k-NN/issues/1889)

### Check List


src/main/java/org/opensearch/knn/quantization/util/QuantizationStateSerializer.java
```diff
@@ -0,0 +1,109 @@
...
+     * @param specificDataDeserializer The deserializer for the specific data associated with the state.
+     * @return The deserialized QuantizationState including its specific data.
+     * @throws IOException            If an I/O error occurs during deserialization.
+     * @throws ClassNotFoundException If the class of the serialized object cannot be found.
+     */
+    public static QuantizationState deserialize(byte[] bytes, SerializableDeserializer specificDataDeserializer)
+            throws IOException, ClassNotFoundException {
+        try (ByteArrayInputStream bis = new ByteArrayInputStream(bytes);
+             ObjectInputStream in = new ObjectInputStream(bis)) {
+            int parentLength = in.readInt(); // Read the length of the
```
ryanbogan, line 68: Nit: duplicate comment fragment

ryanbogan:
Looks like DCO and spotless are failing.

Running `./gradlew spotlessApply` and pushing the commit will fix the spotless part.

For DCO, every commit needs to be signed off with your name and email, which can be done using -s when committing if you have the git configure set up.  This specific commit can be fixed by following the instructions found here: https://github.com/opensearch-project/k-NN/pull/1929/checks?check_run_id=28295803029

src/main/java/org/opensearch/knn/quantization/models/quantizationState/OneBitScalarQuantizationState.java
```diff
@@ -0,0 +1,60 @@
...
+
+import java.io.IOException;
+
+/**
+ * OneBitScalarQuantizationState represents the state of one-bit scalar quantization,
+ * including the mean values used for quantization.
+ */
+public final class OneBitScalarQuantizationState implements QuantizationState {
+    private final SQParams quantizationParams;
+    private final float[] mean;
```
ryanbogan, line 19: Thoughts on renaming this variable to `meanArray`? When I see a variable named mean I would assume it's a float not a float array inherently

src/main/java/org/opensearch/knn/quantization/quantizer/MultiBitScalarQuantizer.java
```diff
@@ -0,0 +1,175 @@
...
+    @Override
+    public QuantizationState train(final TrainingRequest<float[]> trainingRequest) {
+        if (!IS_TRAINING_REQUIRED) {
+            return new DefaultQuantizationState(trainingRequest.getParams());
+        }
+        SQParams params = QuantizerHelper.validateAndExtractParams(trainingRequest);
+        int[] sampledIndices = sampler.sample(trainingRequest.getTotalNumberOfVectors(), samplingSize);
+
+        int dimension = trainingRequest.getVectorByDocId(sampledIndices[0]).length;
+        float[] sumAndMean = new float[dimension];
```
ryanbogan, line 84: Suggest the same renaming as my question above

src/main/java/org/opensearch/knn/quantization/models/requests/TrainingRequest.java
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    /**
+     * Sets the sampled indices for this training request.
+     *
+     * @param sampledIndices the sampled indices.
+     */
+    public void setSampledIndices(int[] sampledIndices) {
+        this.sampledIndices = sampledIndices;
+    }
```
shatejas, line 59: This setter makes the request mutable, any specific reason for this? 

Is the caller supposed to sample before calling the train of the `Quantizer`?

src/main/java/org/opensearch/knn/quantization/quantizer/Quantizer.java
```diff
@@ -0,0 +1,40 @@
...
+/**
+ * The Quantizer interface defines the methods required for training and quantizing vectors
+ * in the context of K-Nearest Neighbors (KNN) and similar machine learning tasks.
+ * It supports training to determine quantization parameters and quantizing data vectors
+ * based on these parameters.
+ *
+ * @param <T> The type of the vector or data to be quantized.
+ * @param <R> The type of the quantized output, typically a compressed or encoded representation.
+ */
+public interface Quantizer<T, R> {
```
shatejas, line 21: Based on method signature, this seems to a bifunction, is it worth re-considering?

```
public interface Quantizer<T, U extends QuantizationState, R> 
```

src/main/java/org/opensearch/knn/quantization/models/quantizationOutput/BinaryQuantizationOutput.java
```diff
@@ -0,0 +1,31 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.knn.quantization.models.quantizationOutput;
+
+/**
+ * The BinaryQuantizationOutput class represents the output of a quantization process in binary format.
+ * It implements the QuantizationOutput interface to handle byte arrays specifically.
+ */
+public class BinaryQuantizationOutput implements QuantizationOutput<byte[]> {
```
shatejas, line 12: I want to see if we can avoid wrapping the array with and object. @navneet1v had tried something similar with KNNVector and an increase in memory usage was seen with the approach. Maybe he can share some insights.

Can you throw some light on how this is intended to be used when integrated with say indexing. Will we be able to get away with simply returning byte array without this wrapper in quantizer?

src/main/java/org/opensearch/knn/quantization/util/BitPackingUtils.java
```diff
@@ -0,0 +1,60 @@
...
+import lombok.experimental.UtilityClass;
+
+import java.util.List;
+
+/**
+ * Utility class for bit packing operations.
+ * Provides methods for packing arrays of bits into byte arrays for efficient storage or transmission.
+ */
+@UtilityClass
+public class BitPackingUtils {
```
shatejas, line 18: nit: A suffix `Utils` generally opens up a possibility of misuse. Lets try to define the name in a way that it constraints the responsibility. Consider `BitPacker` whole responsibility can be pack and unpack as its an inverse function

src/main/java/org/opensearch/knn/quantization/sampler/ReservoirSampler.java
```diff
@@ -0,0 +1,84 @@
...
+     * of being included in the sample.
+     *
+     * @param numVectors the total number of vectors.
+     * @param sampleSize the number of indices to sample.
+     * @return an array of sampled indices.
+     */
+    private int[] reservoirSampleIndices(final int numVectors, final int sampleSize) {
+        int[] indices = IntStream.range(0, sampleSize).toArray();
+        for (int i = sampleSize; i < numVectors; i++) {
+            int j = random.nextInt(i + 1);
```
shatejas, line 76: Lets consider using `ThreadLocalRandom`?

https://docs.oracle.com/javase/8/docs/api/java/util/Random.html

> Instances of java.util.Random are threadsafe. However, the concurrent use of the same java.util.Random instance across threads may encounter contention and consequent poor performance. Consider instead using [ThreadLocalRandom](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ThreadLocalRandom.html) in multithreaded designs.

shatejas (commented):
Took the first pass. Need to still look at a few more details and unit testing

A few overall things
- **Statics**: QuantizationFactory is static. unit testing with static gets much harder as it affects multiple components and gets harder to mock. Is there a way we can avoid it and initialize things like registrar in KNNPlugin.createComponents
- **Enums**: I see values which are not yet supported by framework. Enums are a good way to validate whats supported. Consider only having the values which are currently supported by framework and add it as and when the support is added

Vikasht34:
> Took the first pass. Need to still look at a few more details and unit testing
> 
> A few overall things
> 
> * **Statics**: QuantizationFactory is static. unit testing with static gets much harder as it affects multiple components and gets harder to mock. Is there a way we can avoid it and initialize things like registrar in KNNPlugin.createComponents
> * **Enums**: I see values which are not yet supported by framework. Enums are a good way to validate whats supported. Consider only having the values which are currently supported by framework and add it as and when the support is added

For Point 2 , My vision was to give overall picture , but I don't mind removing it , the ones which is nit getting used. 
For Point 1 :- I would be prefer Keeping registration close to Framework , Let Client not worry on registry.  I would not prefer registration via Create Components as I wanted to have lazy and and not all the clients would be Quantization.

src/main/java/org/opensearch/knn/quantization/sampler/ReservoirSampler.java
Vikasht34: I had this in mind , Missed at last moment. Thanks for adding this. Will Change this.

src/main/java/org/opensearch/knn/quantization/enums/SQTypes.java
```diff
@@ -0,0 +1,60 @@
...
+    /**
+     * FOUR_BIT quantization uses four bits per coordinate.
+     */
+    FOUR_BIT,
+
+    /**
+     * UNSUPPORTED_TYPE is used to denote quantization types that are not supported.
+     * This can be used as a placeholder or default value.
+     */
+    UNSUPPORTED_TYPE
```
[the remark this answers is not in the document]
Vikasht34: I had this for Unit tests , let me see how can I remove this.

src/main/java/org/opensearch/knn/quantization/models/requests/TrainingRequest.java
Vikasht34: Good Catch , It was just  for some testing and forgot to remove this .

src/main/java/org/opensearch/knn/quantization/enums/QuantizationType.java
```diff
@@ -0,0 +1,34 @@
...
+ *       values, which can be useful for compressing data or reducing the precision
+ *       of the representation.</li>
+ * </ul>
+ */
+public enum QuantizationType {
+    /**
+     * Represents space quantization, typically involving dimensionality reduction
+     * or space partitioning techniques.
+     */
+    SPACE_QUANTIZATION,
```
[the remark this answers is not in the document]
jmazanec15: +1

src/main/java/org/opensearch/knn/quantization/models/requests/TrainingRequest.java
```diff
@@ -0,0 +1,55 @@
...
+
+import org.opensearch.knn.quantization.models.quantizationParams.QuantizationParams;
+
+/**
+ * TrainingRequest represents a request for training a quantizer.
+ *
+ * @param <T> the type of vectors to be trained.
+ */
+public abstract class TrainingRequest<T> {
+    private final QuantizationParams params;
```
jmazanec15, line 16: Can we use lombok on these?

src/main/java/org/opensearch/knn/quantization/models/requests/TrainingRequest.java
```diff
@@ -0,0 +1,55 @@
...
+    private final int totalNumberOfVectors;
+
+    /**
+     * Constructs a TrainingRequest with the given parameters and total number of vectors.
+     *
+     * @param params              the quantization parameters.
+     * @param totalNumberOfVectors the total number of vectors.
```
jmazanec15, line 23: What does total number of vectors mean?

src/main/java/org/opensearch/knn/quantization/quantizer/MultiBitScalarQuantizer.java
```diff
@@ -0,0 +1,148 @@
...
+    // to having consistent, Will revisit
+    // if this requires change
+    private static final int DEFAULT_SAMPLE_SIZE = 25000;
+
+    /**
+     * Constructs a MultiBitScalarQuantizer with a specified number of bits per coordinate.
+     *
+     * @param bitsPerCoordinate the number of bits used per coordinate for quantization.
+     */
+    public MultiBitScalarQuantizer(final int bitsPerCoordinate) {
```
jmazanec15, line 43: Do we need multiple constructors or can we just have one?

src/main/java/org/opensearch/knn/quantization/quantizer/OneBitScalarQuantizer.java
```diff
@@ -0,0 +1,107 @@
...
+    /**
+     * Constructs a OneBitScalarQuantizer with a specified sampling size.
+     *
+     * @param samplingSize the number of samples to use for training.
+     */
+    public OneBitScalarQuantizer(final int samplingSize, final Sampler sampler) {
+
+        this.samplingSize = samplingSize;
+        this.sampler = sampler;
+        ;
```
jmazanec15, line 52: nit: remove

src/main/java/org/opensearch/knn/quantization/quantizer/OneBitScalarQuantizer.java
```diff
@@ -0,0 +1,107 @@
...
+import org.opensearch.knn.quantization.models.requests.TrainingRequest;
+import org.opensearch.knn.quantization.sampler.Sampler;
+import org.opensearch.knn.quantization.sampler.SamplingFactory;
+import org.opensearch.knn.quantization.util.BitPacker;
+import org.opensearch.knn.quantization.util.QuantizerHelper;
+
+import java.util.Collections;
+
+/**
+ * OneBitScalarQuantizer is responsible for quantizing vectors using a single bit per dimension.
```
jmazanec15, line 22: Why not just configure multi-bit number of bits to 1 and get rid of this class?

src/main/java/org/opensearch/knn/quantization/sampler/ReservoirSampler.java
```diff
@@ -0,0 +1,85 @@
...
+ * ReservoirSampler implements the Sampler interface and provides a method for sampling
+ * a specified number of indices from a total number of vectors using the reservoir sampling algorithm.
+ * This algorithm is particularly useful for randomly sampling a subset of data from a larger set
+ * when the total size of the dataset is unknown or very large.
+ */
+final class ReservoirSampler implements Sampler {
+
+    private final Random random;
+
+    /**
```
jmazanec15, line 23: Do we need 3 constructors? Can we reduce?

src/main/java/org/opensearch/knn/quantization/sampler/ReservoirSampler.java
```diff
@@ -0,0 +1,85 @@
...
+     * Samples indices from the range [0, totalNumberOfVectors).
+     * If the total number of vectors is less than or equal to the sample size, it returns all indices.
+     * Otherwise, it uses the reservoir sampling algorithm to select a random subset.
+     *
+     * @param totalNumberOfVectors the total number of vectors to sample from.
+     * @param sampleSize           the number of indices to sample.
+     * @return an array of sampled indices.
+     */
+    @Override
+    public int[] sample(final int totalNumberOfVectors, final int sampleSize) {
```
jmazanec15, line 58: This will be about 100 KB per array right? Can we do something to either share or cut down on this?

src/main/java/org/opensearch/knn/quantization/sampler/ReservoirSampler.java
```diff
@@ -0,0 +1,85 @@
...
+        if (totalNumberOfVectors <= sampleSize) {
+            return IntStream.range(0, totalNumberOfVectors).toArray();
+        }
+        return reservoirSampleIndices(totalNumberOfVectors, sampleSize);
+    }
+
+    /**
+     * Applies the reservoir sampling algorithm to select a random sample of indices.
```
jmazanec15, line 66: maybe link to algo for reservoir sampling

src/main/java/org/opensearch/knn/quantization/sampler/ReservoirSampler.java
```diff
@@ -0,0 +1,85 @@
...
+     * This method ensures that each index in the range [0, numVectors) has an equal probability
+     * of being included in the sample.
+     *
+     * @param numVectors the total number of vectors.
+     * @param sampleSize the number of indices to sample.
+     * @return an array of sampled indices.
+     */
+    private int[] reservoirSampleIndices(final int numVectors, final int sampleSize) {
+        int[] indices = IntStream.range(0, sampleSize).toArray();
```
jmazanec15, line 75: This might be a big array as well. Can we think of making more efficient?

src/main/java/org/opensearch/knn/quantization/sampler/ReservoirSampler.java
```diff
@@ -0,0 +1,85 @@
...
+        for (int i = sampleSize; i < numVectors; i++) {
+            int j = random.nextInt(i + 1);
+            if (j < sampleSize) {
+                indices[j] = i;
+            }
+        }
+        Arrays.sort(indices);
```
jmazanec15, line 82: Can we make this more efficient?

src/main/java/org/opensearch/knn/quantization/sampler/Sampler.java
```diff
@@ -0,0 +1,10 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.knn.quantization.sampler;
+
+public interface Sampler {
+    int[] sample(int totalNumberOfVectors, int sampleSize);
```
jmazanec15, line 9: javadoc

src/main/java/org/opensearch/knn/quantization/sampler/SamplingFactory.java
```diff
@@ -0,0 +1,46 @@
...
+     * The class is not meant to be instantiated, as it provides static methods only.
+     */
+    private SamplingFactory() {
+
+    }
+
+    /**
+     * SamplerType is an enumeration of the different types of samplers that can be created by the factory.
+     */
+    public enum SamplerType {
```
jmazanec15, line 25: Separate class?

jmazanec15 (commented):
Took an initial pass. Will need a couple more.

src/main/java/org/opensearch/knn/quantization/models/quantizationParams/SQParams.java
```diff
@@ -0,0 +1,83 @@
...
+ */
+public class SQParams implements QuantizationParams {
+    private final ScalarQuantizationType sqType;
+
+    /**
+     * Constructs an SQParams instance with the specified scalar quantization type.
+     *
+     * @param sqType The specific type of scalar quantization (e.g., ONE_BIT, TWO_BIT, FOUR_BIT).
+     */
+    public SQParams(final ScalarQuantizationType sqType) {
```
navneet1v, line 25: why not use @RequiredArgsConstructor here.

src/main/java/org/opensearch/knn/quantization/factory/QuantizerRegistry.java
```diff
@@ -0,0 +1,82 @@
...
+     */
+    public static <P extends QuantizationParams> void register(
+        final Class<P> paramClass,
+        final QuantizationType quantizationType,
+        final ScalarQuantizationType sqType,
+        final Supplier<? extends Quantizer<?, ?>> quantizerSupplier
+    ) {
+        String identifier = createIdentifier(quantizationType, sqType);
+        // Ensure that the quantizer for this identifier is registered only once
+        registry.computeIfAbsent(identifier, key -> quantizerSupplier);
```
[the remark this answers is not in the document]
navneet1v: What will happen in this case:
```
key1 -> QT1
key1 -> QT2
```
?

src/main/java/org/opensearch/knn/quantization/factory/QuantizerRegistry.java
```diff
@@ -0,0 +1,82 @@
...
+
+    /**
+     * Creates a unique identifier for the quantizer based on the quantization type and specific quantization subtype.
+     *
+     * @param quantizationType the quantization type
+     * @param sqType           the specific quantization subtype
+     * @return a string identifier
+     */
+    private static String createIdentifier(final QuantizationType quantizationType, final ScalarQuantizationType sqType) {
+        return quantizationType.name() + "_" + sqType.name();
```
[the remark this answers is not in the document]
navneet1v: Example:
ONE_BIT -> VALUE_ONE_BIT -> old segment QS

ONEBIT -> VALUE_ONEBIT -> new segment QS

registry -> VALUE_ONE_BIT (old code)

registry -> VALUE_ONEBIT (new code)

src/main/java/org/opensearch/knn/quantization/models/quantizationParams/QuantizationParams.java
```diff
@@ -0,0 +1,39 @@
...
+import java.io.Serializable;
+
+/**
+ * Interface for quantization parameters.
+ * This interface defines the basic contract for all quantization parameter types.
+ * It provides methods to retrieve the quantization type and a unique type identifier.
+ * Implementations of this interface are expected to provide specific configurations
+ * for various quantization strategies.
+ */
+public interface QuantizationParams extends Serializable {
```
navneet1v, line 19: Question: Just extending the Serializable will that be sufficient? I think we should add the functions writeObject/readObject to ensure that we are controlling how the serializing is happening for the param class.

Example:
```
// old
Param{
   var1 => abc
   var2 => def
}

// new code
Param{
   var2 => bbb
   var1 => aaa
   var3 => asdfasdf
}
```

navneet1v (commented):
Clean code. Please take a look at the comments provided.

src/main/java/org/opensearch/knn/quantization/quantizer/OneBitScalarQuantizer.java
Vikasht34: It Does not work with OneBit ..if Bit is 1 , it gives two thresholds and we don't need to calculate stdDev , so we can not use for oneBit

src/main/java/org/opensearch/knn/quantization/sampler/ReservoirSampler.java
navneet1v: can we use bitset here?

src/main/java/org/opensearch/knn/quantization/models/quantizationParams/QuantizationParams.java
shatejas: +1 this should help us manage backward incompatible changes

src/main/java/org/opensearch/knn/quantization/factory/QuantizerRegistry.java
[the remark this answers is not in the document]
Vikasht34: Behavior of computeIfAbsent
* First Call with key1 -> QT1:
        If paramIdentifier (key1) is not present in the registry, the lambda function (key -> quantizerSupplier) will be called, and the result (QT1) will be inserted into the registry with the key (key1).
* Subsequent Call with key1 -> QT2:
        If paramIdentifier (key1) is already present in the registry, computeIfAbsent will not call the lambda function. Instead, it will return the existing value associated with the key (QT1), and QT2 will be ignored.

src/main/java/org/opensearch/knn/quantization/enums/ScalarQuantizationType.java
```diff
@@ -0,0 +1,34 @@
...
+
+    /**
+     * TWO_BIT quantization uses two bits per coordinate.
+     */
+    TWO_BIT,
+
+    /**
+     * FOUR_BIT quantization uses four bits per coordinate.
+     */
+    FOUR_BIT,
```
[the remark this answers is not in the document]
Vikasht34: This is the class where all SqTypes will be , I had it before , I deleted because it was not getting used.

src/main/java/org/opensearch/knn/quantization/quantizer/MultiBitScalarQuantizer.java
Vikasht34: I had this for easy configuration till we have multiple version for Sampler , and I wanted to avoid passing sampler from Registry , It does not harm with Overloaded Constructor.

src/main/java/org/opensearch/knn/quantization/quantizer/Quantizer.java
Vikasht34: Quantization State is something Type Independent , It wil be overKill if I define that as U , right now there are no strong use cases requiring different types of QuantizationState or QuantizationOutput, the added complexity might not be necessary. Can revisit in future if needed.

src/main/java/org/opensearch/knn/quantization/quantizer/OneBitScalarQuantizer.java
```diff
@@ -0,0 +1,114 @@
...
+     * Trains the quantizer by calculating the mean of each dimension from the sampled vectors.
+     * These means are used as thresholds in the quantization process.
+     *
+     * @param trainingRequest the request containing the data and parameters for training.
+     * @return a OneBitScalarQuantizationState containing the calculated means.
+     */
+    @Override
+    public QuantizationState train(final TrainingRequest<float[]> trainingRequest) {
+        BitSet sampledDocIds = sampler.sample(trainingRequest.getTotalNumberOfVectors(), samplingSize);
+        float[] mean = QuantizerHelper.calculateMean(trainingRequest, sampledDocIds);
```
ryanbogan, line 64: Could we rename this to meanArray as well?

src/main/java/org/opensearch/knn/quantization/quantizer/QuantizerHelper.java
```diff
@@ -0,0 +1,93 @@
...
+     * Each element of the mean vector is computed as the average of the corresponding elements
+     * of the sampled vectors.</p>
+     *
+     * @param samplingRequest The {@link TrainingRequest} containing the dataset and methods to access vectors by their indices.
+     * @param sampledIndices An array of indices representing the sampled vectors to be used for mean calculation.
+     * @return A float array representing the mean vector of the sampled vectors.
+     * @throws IllegalArgumentException If any of the vectors at the sampled indices are null.
+     * @throws IllegalStateException If the mean array is unexpectedly null after processing the vectors.
+     */
+    static float[] calculateMean(TrainingRequest<float[]> samplingRequest, BitSet sampledIndices) {
```
ryanbogan, line 34: And all the ones here too for meanArray

ryanbogan: There are also some in the tests but that's not as important

src/main/java/org/opensearch/knn/quantization/sampler/ReservoirSampler.java
```diff
@@ -0,0 +1,79 @@
...
+ */
+final class ReservoirSampler implements Sampler {
+
+    private final Random random;
+
+    /**
+     * Constructs a ReservoirSampler with a new Random instance.
+     */
+    public ReservoirSampler() {
+        this.random = ThreadLocalRandom.current();
```
shatejas, line 26: Won't caching this as a member variable defeat the purpose of isolating random to a single thread? 

The sampler is a member variable for quantizer. so there is one instance of this created. If train is called in parallel, will there be a risk of contention here?

Typical use of this will be wherever a new random number needs to be generated in this case on line 67

```
int j = ThreadLocalRandom.current().nextInt(i + 1);
```

src/test/java/org/opensearch/knn/quantization/enums/SQTypesTests.java
```diff
@@ -0,0 +1,24 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.knn.quantization.enums;
+
+import org.opensearch.knn.KNNTestCase;
+
+public class SQTypesTests extends KNNTestCase {
```
shatejas, line 10: nit: will be a good idea to change the name here

src/main/java/org/opensearch/knn/quantization/models/quantizationState/MultiBitScalarQuantizationState.java
```diff
@@ -0,0 +1,97 @@
...
+    private static final int CURRENT_VERSION = 1; // Current version of SQParams
+
+    @Override
+    public ScalarQuantizationParams getQuantizationParams() {
+        return quantizationParams;
+    }
+
+    @Override
+    public void writeExternal(ObjectOutput out) throws IOException {
+        int combinedVersion = (CURRENT_VERSION << 16) | quantizationParams.getVersion();
```
shatejas, line 53: Is`|` done to save memory. I am wondering why is this coupled and if there are any unforeseen consequences . Should  `quantizationParams.writeExternal(out)` be taking care of quantizationparams version?

src/main/java/org/opensearch/knn/quantization/models/quantizationState/OneBitScalarQuantizationState.java
```diff
@@ -0,0 +1,75 @@
...
+import java.io.ObjectInput;
+import java.io.ObjectOutput;
+
+/**
+ * OneBitScalarQuantizationState represents the state of one-bit scalar quantization,
+ * including the mean values used for quantization.
+ */
+@Getter
+@NoArgsConstructor // No-argument constructor for deserialization
+@AllArgsConstructor // Constructor with all arguments
```
vibrantvarun, line 23: Can we have this extra comments of 
```
// No-argument constructor for deserialization
// Constructor with all arguments
```

as they are self explanatory.

src/main/java/org/opensearch/knn/quantization/models/quantizationState/OneBitScalarQuantizationState.java
```diff
@@ -0,0 +1,75 @@
...
+public final class OneBitScalarQuantizationState implements QuantizationState {
+    private ScalarQuantizationParams quantizationParams;
+    private float[] meanThresholds;
+    private static final long serialVersionUID = 1L; // Version ID for serialization
+    private static final int CURRENT_VERSION = 1; // Current version of SQParams
+
```
vibrantvarun, line 29: Can we add a similar type of description from MultiBitScalarQuantization here as it helped me a lot to understand about the class?

shatejas:
> Point 1 :- I would be prefer Keeping registration close to Framework , Let Client not worry on registry. I would not prefer registration via Create Components as I wanted to have lazy and and not all the clients would be Quantization.

Fair, my main concern is with static factory is the unit testing, I would still want to see how this is plugged in. If this is not plugged in as a member variable I see unit tests turning more into a component tests since this piece will be used in the active path of both indexing and search.

src/main/java/org/opensearch/knn/quantization/quantizer/Quantizer.java
```diff
@@ -0,0 +1,37 @@
...
+    QuantizationState train(TrainingRequest<T> trainingRequest);
+
+    /**
+     * Quantizes the provided vector using the specified quantization state.
+     *
+     * @param vector the vector to quantize.
+     * @param state  the quantization state containing parameters for quantization.
+     * @param output the QuantizationOutput object to store the quantized representation of the vector.
+     */
+    void quantize(T vector, QuantizationState state, QuantizationOutput<R> output) throws IOException;
```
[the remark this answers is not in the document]
Vikasht34: Yes , it will be different per vector.

src/main/java/org/opensearch/knn/quantization/models/quantizationParams/QuantizationParams.java
```diff
@@ -0,0 +1,55 @@
...
+
+    /**
+     * Serializes the QuantizationParams object to an external output.
+     * Default implementation is no-op.
+     *
+     * @param out the ObjectOutput to write the object to.
+     * @throws IOException if an I/O error occurs during serialization.
+     */
+    @Override
+    default void writeExternal(ObjectOutput out) throws IOException {
```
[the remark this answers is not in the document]
Vikasht34: We are not , This is the interface we are implementating 

```
public interface Externalizable extends java.io.Serializable {
    /**
     * The object implements the writeExternal method to save its contents
     * by calling the methods of DataOutput for its primitive values or
     * calling the writeObject method of ObjectOutput for objects, strings,
     * and arrays.
     *
     * @serialData Overriding methods should use this tag to describe
     *             the data layout of this Externalizable object.
     *             List the sequence of element types and, if possible,
     *             relate the element to a public/protected field and/or
     *             method of this Externalizable class.
     *
     * @param     out the stream to write the object to
     * @throws    IOException Includes any I/O exceptions that may occur
     */
    void writeExternal(ObjectOutput out) throws IOException;

    /**
     * The object implements the readExternal method to restore its
     * contents by calling the methods of DataInput for primitive
     * types and readObject for objects, strings and arrays.  The
     * readExternal method must read the values in the same sequence
     * and with the same types as were written by writeExternal.
     *
     * @param     in the stream to read data from in order to restore the object
     * @throws    IOException if I/O errors occur
     * @throws    ClassNotFoundException If the class for an object being
     *            restored cannot be found.
     */
    void readExternal(ObjectInput in) throws IOException, ClassNotFoundException;
}
```

src/main/java/org/opensearch/knn/quantization/models/quantizationState/MultiBitScalarQuantizationState.java
Vikasht34: We are writing version once , and we don't want write version for every child classes and yes it's for saving memory

src/main/java/org/opensearch/knn/quantization/quantizer/Quantizer.java
Vikasht34:
```public interface QuantizationState extends Externalizable {
    /**
     * Returns the quantization parameters associated with this state.
     *
     * @return the quantization parameters.
     */
    QuantizationParams getQuantizationParams();

    /**
     * Serializes the quantization state to a byte array.
     *
     * @return a byte array representing the serialized state.
     * @throws IOException if an I/O error occurs during serialization.
     */
    byte[] toByteArray() throws IOException;
}
```

This is quantization State , It is not any Type specific, I really don't want to abuse generics and follow KISS.

src/main/java/org/opensearch/knn/quantization/models/quantizationState/MultiBitScalarQuantizationState.java
shatejas: How will the version be handled in case quantizationParams is written independently? Being a public isn't that a possibility?

src/main/java/org/opensearch/knn/quantization/models/quantizationParams/QuantizationParams.java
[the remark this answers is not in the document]
shatejas: My question is more on why is there a default implementation for QuantizationParams, Shouldn't we be forcing the implementation, to have its own implementation to manage proper read and write?

src/main/java/org/opensearch/knn/quantization/models/quantizationParams/ScalarQuantizationParams.java
```diff
@@ -0,0 +1,96 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.knn.quantization.models.quantizationParams;
+
+import lombok.*;
```
shatejas, line 8: Remove * imports?

src/main/java/org/opensearch/knn/quantization/models/quantizationParams/QuantizationParams.java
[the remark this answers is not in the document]
Vikasht34: Oh this , Yeah I can remove that ...It's just came out of practice ..

src/main/java/org/opensearch/knn/quantization/enums/ScalarQuantizationType.java
```diff
@@ -0,0 +1,44 @@
...
+     */
+    FOUR_BIT(4);
+
+    private final int id;
+
+    ScalarQuantizationType(int id) {
+        this.id = id;
+    }
+
+    public int getId() {
```
jmazanec15, line 41: Javadoc or can we use lombok getter anotation?

src/main/java/org/opensearch/knn/quantization/factory/QuantizerRegistry.java
```diff
@@ -0,0 +1,57 @@
...
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+final class QuantizerRegistry {
+    // ConcurrentHashMap for thread-safe access
+    private static final Map<String, Supplier<? extends Quantizer<?, ?>>> registry = new ConcurrentHashMap<>();
+
+    /**
+     * Registers a quantizer with the registry.
+     *
+     * @param paramIdentifier   the unique identifier for the quantization parameters
+     * @param quantizerSupplier a supplier that provides instances of the quantizer
```
jmazanec15, line 31: Do we need to create a new quantizer for each request or can we reuse?

src/main/java/org/opensearch/knn/quantization/models/quantizationOutput/QuantizationOutput.java
```diff
@@ -0,0 +1,30 @@
...
+     */
+    T getQuantizedVector();
+
+    /**
+     * Updates the quantized vector with new data.
+     *
+     * @param newQuantizedVector the new quantized vector data.
+     * @throws IOException if an I/O error occurs during the update.
+     */
+    void updateQuantizedVector(T newQuantizedVector) throws IOException;
```
jmazanec15, line 29: qq: Why is this needed?

src/main/java/org/opensearch/knn/quantization/models/quantizationOutput/BinaryQuantizationOutput.java
```diff
@@ -0,0 +1,40 @@
...
+import java.io.IOException;
+
+/**
+ * The BinaryQuantizationOutput class represents the output of a quantization process in binary format.
+ * It implements the QuantizationOutput interface to handle byte arrays specifically.
+ */
+@NoArgsConstructor
+public class BinaryQuantizationOutput implements QuantizationOutput<byte[]> {
+    @Getter
+    private final ByteArrayOutputStream byteArrayOutputStream = new ByteArrayOutputStream();
```
jmazanec15, line 21: Is a stream like this going to add overhead? Can we just use a fix-sized byte array upfront?

src/main/java/org/opensearch/knn/quantization/models/quantizationParams/QuantizationParams.java
```diff
@@ -0,0 +1,27 @@
...
+public interface QuantizationParams extends Externalizable {
+    /**
+     * Provides a unique identifier for the quantization parameters.
+     * This identifier is typically a combination of the quantization type
+     * and additional specifics, and it serves to distinguish between different
+     * configurations or modes of quantization.
+     *
+     * @return a string representing the unique type identifier.
+     */
+    String getTypeIdentifier();
```
jmazanec15, line 26: qq: Why not return the quantization type directly?

src/main/java/org/opensearch/knn/quantization/models/quantizationParams/QuantizationParams.java
```diff
@@ -0,0 +1,27 @@
...
+import java.io.Externalizable;
+
+/**
+ * Interface for quantization parameters.
+ * This interface defines the basic contract for all quantization parameter types.
+ * It provides methods to retrieve the quantization type and a unique type identifier.
+ * Implementations of this interface are expected to provide specific configurations
+ * for various quantization strategies.
+ */
+public interface QuantizationParams extends Externalizable {
```
jmazanec15, line 17: Should we implement OpenSearch Writeable interface for this instead of Externalizable? https://github.com/opensearch-project/OpenSearch/blob/main/libs/core/src/main/java/org/opensearch/core/common/io/stream/Writeable.java

src/main/java/org/opensearch/knn/quantization/models/quantizationParams/ScalarQuantizationParams.java
```diff
@@ -0,0 +1,114 @@
...
+     *
+     * @param in the ObjectInput to read the object from.
+     * @throws IOException if an I/O error occurs during deserialization.
+     * @throws ClassNotFoundException if the class of the serialized object cannot be found.
+     */
+    @Override
+    public void readExternal(ObjectInput in) throws IOException, ClassNotFoundException {
+        // The version is already read by the parent state class and set in VersionContext
+        // Retrieve the current version from VersionContext to handle version-specific deserialization logic
+        // int versionId = VersionContext.getVersion();
```
jmazanec15, line 80: Why commented out? Same with other things in the class?

src/main/java/org/opensearch/knn/quantization/models/quantizationState/OneBitScalarQuantizationState.java
```diff
@@ -0,0 +1,181 @@
...
+    @Override
+    public ScalarQuantizationParams getQuantizationParams() {
+        return quantizationParams;
+    }
+
+    /**
+     * This method is responsible for writing the state of the OneBitScalarQuantizationState object to an external output.
+     * It includes versioning information to ensure compatibility between different versions of the serialized object.
+     *
+     * <p>Versioning is managed using the {@link VersionContext} class. This allows other classes that are serialized
```
jmazanec15, line 52: Are you sure we need this VersionContext? Seems like it could have some issues and rather not need to deal with it.

src/main/java/org/opensearch/knn/quantization/models/quantizationState/OneBitScalarQuantizationState.java
```diff
@@ -0,0 +1,181 @@
...
+     * OneBitScalarQuantizationState state = new OneBitScalarQuantizationState(params, meanThresholds);
+     * byte[] serializedState = state.toByteArray();
+     * }
+     * </pre>
+     *
+     * @return a byte array representing the serialized state of this object.
+     * @throws IOException if an I/O error occurs during serialization.
+     */
+    @Override
+    public byte[] toByteArray() throws IOException {
```
jmazanec15, line 148: Do we need this if we implement Writeable interface?

src/main/java/org/opensearch/knn/quantization/quantizer/Quantizer.java
```diff
@@ -0,0 +1,37 @@
+package org.opensearch.knn.quantization.quantizer;
+
+import org.opensearch.knn.quantization.models.quantizationOutput.QuantizationOutput;
+import org.opensearch.knn.quantization.models.quantizationState.QuantizationState;
+import org.opensearch.knn.quantization.models.requests.TrainingRequest;
+
+import java.io.IOException;
+
+/**
+ * The Quantizer interface defines the methods required for training and quantizing vectors
```
jmazanec15, line 10: If we wanted to do some kind of asymetric distance computation in the future, could we change this interface easily to do that?

src/main/java/org/opensearch/knn/quantization/sampler/SamplingFactory.java
```diff
@@ -0,0 +1,34 @@
...
+     * Creates and returns a Sampler instance based on the specified SamplerType.
+     *
+     * @param samplerType the type of sampler to create.
+     * @return a Sampler instance.
+     * @throws IllegalArgumentException if the sampler type is not supported.
+     */
+    public static Sampler getSampler(final SamplerType samplerType) {
+        switch (samplerType) {
+            case RESERVOIR:
+                return new ReservoirSampler();
```
jmazanec15, line 28: Can we reuse/singleton?

jmazanec15 (commented):
Overall I think it looks really good. A few more comments. One thing in particular though, do we really need VersionContext or can we avoid this. Im a little bit wary of the static setting of variable even though its thread local.

Im going to take a closer look at the quantizer implementations next, but didnt have time right now and wanted to provide more incremental feedback.

src/main/java/org/opensearch/knn/quantization/models/quantizationOutput/QuantizationOutput.java
Vikasht34: We did not want to create new object of Quantized output for every request during quantize , from Client we will just pass refrence of quantized output and keep refreshing quantized vector.

src/main/java/org/opensearch/knn/quantization/models/quantizationOutput/BinaryQuantizationOutput.java
Vikasht34: The problem with fixed size is we don't know the buffer size , but stream gives us flexibility.

src/main/java/org/opensearch/knn/quantization/models/quantizationParams/QuantizationParams.java
Vikasht34: Idea is isloate framewoek as much as isloated with intefaces , and keep interfces clean to remove tight coupling ,  Externalizable is also same as writable which inturns uses objectstream.

src/main/java/org/opensearch/knn/quantization/models/quantizationParams/ScalarQuantizationParams.java
Vikasht34: This comment is for any future backward comapatibility , so In functure if we introduce , this will help to how to get version and comapre.

src/main/java/org/opensearch/knn/quantization/models/quantizationState/OneBitScalarQuantizationState.java
Vikasht34: There are two options . Each class writes version, which takes more memory , that's why we are writing version once , now when we read the version , we need to pass to child class for desealize according to version. either I pass that version in method to accomplish that and to do that I will write something like 

ReadExternal( In, version) which is not given by Externaible interface so it will overide default implementation , which creates much confusion. ThreadLocal is local to caller thread , I am not forcesing any potential issue. I can revisit in future if we need to.

src/main/java/org/opensearch/knn/quantization/models/quantizationState/OneBitScalarQuantizationState.java
Vikasht34: I commented on top for this.

src/main/java/org/opensearch/knn/quantization/quantizer/Quantizer.java
Vikasht34: yes ..we cab have default here so that otheres are not affected.

src/main/java/org/opensearch/knn/quantization/models/quantizationOutput/BinaryQuantizationOutput.java
jmazanec15: Why dont we know the size? Wont QuantizationOutput always output an array of fixed size? Would it ever operate on different dimensional data?

src/main/java/org/opensearch/knn/quantization/factory/QuantizerRegistry.java
Vikasht34: Good Catch ....We can re-use, since Quantizer is stateless we can definitely do.

src/main/java/org/opensearch/knn/quantization/models/quantizationParams/QuantizationParams.java
Vikasht34: I took depper look in Writable , seems promising though we have coupled with OpenSearch Core , Implemented using Writable.

src/main/java/org/opensearch/knn/quantization/models/quantizationState/OneBitScalarQuantizationState.java
Vikasht34: Removed it .

src/main/java/org/opensearch/knn/quantization/models/quantizationOutput/BinaryQuantizationOutput.java
Vikasht34: Done

src/main/java/org/opensearch/knn/quantization/enums/ScalarQuantizationType.java
```diff
@@ -0,0 +1,46 @@
...
+import lombok.Getter;
+
+/**
+ * The ScalarQuantizationType enum defines the various scalar quantization types that can be used
+ * for vector quantization. Each type corresponds to a different bit-width representation of the quantized values.
+ *
+ * <p>
+ * Future Developers: If you change the name of any enum constant, do not change its associated value.
+ * Serialization and deserialization depend on these values to maintain compatibility.
+ * </p>
```
navneet1v, line 17: can we also add a test case which will break if any of the Quantization type has same Id?

src/main/java/org/opensearch/knn/quantization/factory/QuantizerRegistrar.java
```diff
@@ -0,0 +1,46 @@
...
+final class QuantizerRegistrar {
+
+    /**
+     * Registers default quantizers if not already registered.
+     * <p>
+     * This method is synchronized to ensure that registration occurs only once,
+     * even in a multi-threaded environment.
+     * </p>
+     */
+    static synchronized void registerDefaultQuantizers() {
```
navneet1v, line 29: why do we need to synchronized this code block? when the caller of this block is already synchronized? any special reason here?

src/main/java/org/opensearch/knn/quantization/factory/QuantizerRegistrar.java
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * The QuantizerRegistrar class is responsible for registering default quantizers.
+ * This class ensures that the registration happens only once in a thread-safe manner.
+ */
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+final class QuantizerRegistrar {
+
+    /**
+     * Registers default quantizers if not already registered.
```
navneet1v, line 23: this function is not checking if quantizers are registered or not. Lets just fix this doc to say register default quantizers

src/main/java/org/opensearch/knn/quantization/factory/QuantizerRegistry.java
```diff
@@ -0,0 +1,56 @@
...
+    /**
+     * Registers a quantizer with the registry.
+     *
+     * @param paramIdentifier the unique identifier for the quantization parameters
+     * @param quantizer       an instance of the quantizer
+     */
+    static void register(final String paramIdentifier, final Quantizer<?, ?> quantizer) {
+        // Ensure that the quantizer for this identifier is registered only once
+        registry.putIfAbsent(paramIdentifier, quantizer);
+    }
```
navneet1v, line 35: should we throw an exception if using a same paramIdentifier a quantizer is already registered?

src/main/java/org/opensearch/knn/quantization/models/quantizationOutput/BinaryQuantizationOutput.java
navneet1v: Un-resolving the comment as the comment is not resolved to say what approach we are going to follow here.

src/main/java/org/opensearch/knn/quantization/models/quantizationOutput/BinaryQuantizationOutput.java
```diff
@@ -0,0 +1,41 @@
...
+public class BinaryQuantizationOutput implements QuantizationOutput<byte[]> {
+    @Getter
+    private byte[] quantizedVector;
+
+    /**
+     * Updates the quantized vector with a new byte array.
+     *
+     * @param newQuantizedVector the new quantized vector represented as a byte array.
+     */
+    public void updateQuantizedVector(final byte[] newQuantizedVector) {
```
navneet1v, line 25: wondering what could be a better way to doing this and more importantly how clients are supposed to use this, how clients will know what length of byte vector in this case needs to provided

src/main/java/org/opensearch/knn/quantization/models/quantizationOutput/QuantizationOutput.java
navneet1v, line 29: why this function is throwing an exception? when the only implementation throws an `IllegalArgumentException` and no IOException. What kind of IO we are excepting in this function?

src/main/java/org/opensearch/knn/quantization/models/quantizationParams/ScalarQuantizationParams.java
```diff
@@ -0,0 +1,78 @@
...
+ * The ScalarQuantizationParams class represents the parameters specific to scalar quantization (SQ).
+ * This class implements the QuantizationParams interface and includes the type of scalar quantization.
+ */
+@Getter
+@AllArgsConstructor
+@NoArgsConstructor // No-argument constructor for deserialization
+@EqualsAndHashCode
+public class ScalarQuantizationParams implements QuantizationParams {
+    private ScalarQuantizationType sqType;
+    private static final long serialVersionUID = 1L; // Version ID for serialization
```
navneet1v, line 29: do we need this now?

src/main/java/org/opensearch/knn/quantization/models/quantizationParams/ScalarQuantizationParams.java
```diff
@@ -0,0 +1,78 @@
...
+
+    /**
+     * Static method to generate type identifier based on ScalarQuantizationType.
+     *
+     * @param sqType the scalar quantization type.
+     * @return A string representing the unique type identifier.
+     */
+    public static String generateTypeIdentifier(ScalarQuantizationType sqType) {
```
navneet1v, line 37: why this is a public function of `ScalarQuantizationParams`? who will use this outside of QF? is this function better be present in an utility/helper?

src/main/java/org/opensearch/knn/quantization/models/quantizationParams/ScalarQuantizationParams.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    /**
+     * Provides a unique type identifier for the ScalarQuantizationParams, combining the SQ type.
+     * This identifier is useful for distinguishing between different configurations of scalar quantization parameters.
+     *
+     * @return A string representing the unique type identifier.
+     */
+    @Override
+    public String getTypeIdentifier() {
```
navneet1v, line 48: will client ever use a public function like this?

src/main/java/org/opensearch/knn/quantization/models/quantizationParams/ScalarQuantizationParams.java
```diff
@@ -0,0 +1,78 @@
...
+        return generateIdentifier(sqType.getId());
+    }
+
+    private static String generateIdentifier(int id) {
+        return String.format(Locale.ROOT, "ScalarQuantizationParams_%d", id);
```
navneet1v, line 53: we use Locale.ROOT generally for exception messages as all users might not use English language. Given this identifier will be used as key, I think its better not to use locale and also not use String.format to keep identifiers consistent.

just do

```
return "ScalarQuantizationParams_" + id
```

src/main/java/org/opensearch/knn/quantization/models/quantizationParams/ScalarQuantizationParams.java
```diff
@@ -0,0 +1,78 @@
...
+    /**
+     * Writes the object to the output stream.
+     * This method is part of the Writeable interface and is used to serialize the object.
+     *
+     * @param out the output stream to write the object to.
+     * @throws IOException if an I/O error occurs.
+     */
+    @Override
+    public void writeTo(StreamOutput out) throws IOException {
+        out.writeEnum(sqType);
```
navneet1v, line 65: writeEnum doesn't write the value in this case `id` of the enum. It is writing the `ordinal` of the enum. Ref: https://github.com/opensearch-project/OpenSearch/blob/main/libs/core/src/main/java/org/opensearch/core/common/io/stream/StreamOutput.java#L1240-L1242

Reason why this is working for you is because currently the ordinal and the ids are same. If you reverse the ids in the enum your tests should fail.

src/main/java/org/opensearch/knn/quantization/models/quantizationParams/ScalarQuantizationParams.java
```diff
@@ -0,0 +1,78 @@
...
+    /**
+     * Reads the object from the input stream.
+     * This method is part of the Writeable interface and is used to deserialize the object.
+     *
+     * @param in the input stream to read the object from.
+     * @throws IOException if an I/O error occurs.
+     */
+    public ScalarQuantizationParams(StreamInput in, int version) throws IOException {
+        this.sqType = in.readEnum(ScalarQuantizationType.class);
+    }
```
navneet1v, line 77: same as above. Ref: https://github.com/opensearch-project/OpenSearch/blob/main/libs/core/src/main/java/org/opensearch/core/common/io/stream/StreamInput.java#L1280-L1289

What you should be doing here is first reading the id as vInt and then using that Id to map to right `ScalarQuantizationType`.

src/main/java/org/opensearch/knn/quantization/models/quantizationState/MultiBitScalarQuantizationState.java
```diff
@@ -0,0 +1,166 @@
...
+     *
+     * @param out the StreamOutput to write the object to.
+     * @throws IOException if an I/O error occurs during serialization.
+     */
+    @Override
+    public void writeTo(StreamOutput out) throws IOException {
+        out.writeInt(Version.CURRENT.id); // Write the version
+        quantizationParams.writeTo(out);
+        out.writeInt(thresholds.length);
+        out.writeInt(thresholds[0].length);
```
navneet1v, line 78: why not VInt?

src/main/java/org/opensearch/knn/quantization/models/quantizationState/MultiBitScalarQuantizationState.java
```diff
@@ -0,0 +1,166 @@
...
+        for (float[] row : thresholds) {
+            for (float value : row) {
+                out.writeFloat(value);
+            }
+        }
```
navneet1v, line 83: why we are not using `writeArray` function of StreamOutput here?

src/main/java/org/opensearch/knn/quantization/models/quantizationState/MultiBitScalarQuantizationState.java
```diff
@@ -0,0 +1,166 @@
...
+    public MultiBitScalarQuantizationState(StreamInput in) throws IOException {
+        int version = in.readInt(); // Read the version
+        this.quantizationParams = new ScalarQuantizationParams(in, version);
+        int rows = in.readInt();
+        int cols = in.readInt();
+        thresholds = new float[rows][cols];
+        for (int i = 0; i < rows; i++) {
+            for (int j = 0; j < cols; j++) {
+                thresholds[i][j] = in.readFloat();
+            }
```
navneet1v, line 119: same as above.

navneet1v (changes requested):
Given the size of the PR and amount of changes was able to complete only half. Will check other half of the PR next day.

My main concern currently in the PR is use of Ordinal in the serialization of Enums(this is done by SteamInput/StreamOutput class) rather than id which was added just for serialization.

src/main/java/org/opensearch/knn/quantization/quantizer/QuantizerHelper.java
```diff
@@ -0,0 +1,93 @@
...
+            throw new IllegalStateException("Mean array should not be null after processing vectors.");
+        }
+        for (int j = 0; j < mean.length; j++) {
+            mean[j] /= totalSamples;
+        }
+        return mean;
+    }
+
+    /**
+     * Calculates the sum, sum of squares, mean, and standard deviation for each dimension in a single pass.
```
jmazanec15, line 59: This is just mean and std dev now right?

src/main/java/org/opensearch/knn/quantization/quantizer/OneBitScalarQuantizer.java
```diff
@@ -0,0 +1,104 @@
...
+     * Quantizes the provided vector using the given quantization state.
+     * It compares each dimension of the vector against the corresponding mean (threshold) to determine the quantized value.
+     *
+     * @param vector the vector to quantize.
+     * @param state  the quantization state containing the means for each dimension.
+     * @param output the QuantizationOutput object to store the quantized representation of the vector.
+     * @return a BinaryQuantizationOutput containing the quantized data.
+     */
+    @Override
+    public void quantize(final float[] vector, final QuantizationState state, final QuantizationOutput<byte[]> output) throws IOException {
```
jmazanec15, line 78: Why not make state a constructor parameter instead of passing in here?

src/main/java/org/opensearch/knn/quantization/quantizer/BitPacker.java
```diff
@@ -0,0 +1,65 @@
...
+        }
+
+        return packedBits;
+    }
+
+    /**
+     * Overloaded method to pack bits for one-bit quantization.
+     *
+     * @param vector             the vector to quantize.
+     * @param thresholds         the thresholds for quantization.
```
jmazanec15, line 59: nit: add a little bit more on what the thresholds are

src/main/java/org/opensearch/knn/quantization/quantizer/QuantizerHelper.java
```diff
@@ -0,0 +1,93 @@
...
+
+import java.util.BitSet;
+
+/**
+ * Utility class providing common methods for quantizer operations, such as parameter validation and
+ * extraction. This class is designed to be used with various quantizer implementations that require
+ * consistent handling of training requests and sampled indices.
+ */
+@UtilityClass
+class QuantizerHelper {
```
jmazanec15, line 19: Are we able to leverage lucene's VectorUtil class at all (https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/util/VectorUtil.java) so we can get SIMD performance benefits by default?

src/main/java/org/opensearch/knn/quantization/quantizer/BitPacker.java
```diff
@@ -0,0 +1,65 @@
...
+     * Packs bits into a byte array based on the provided thresholds and vector.
+     *
+     * @param vector             the vector to quantize.
+     * @param thresholds         the thresholds for quantization.
+     * @param bitsPerCoordinate  the number of bits used per coordinate.
+     * @return the packed bits as a byte array.
+     */
+    byte[] packBits(final float[] vector, final float[][] thresholds, final int bitsPerCoordinate) {
+        int totalBits = bitsPerCoordinate * vector.length;
+        int byteLength = (totalBits + 7) >> 3; // Calculate byte length needed
```
jmazanec15, line 38: Might be worth just adding pseudo code formula in comment. Took me a second to get it.

src/main/java/org/opensearch/knn/quantization/quantizer/BitPacker.java
```diff
@@ -0,0 +1,65 @@
...
+        byte[] packedBits = new byte[byteLength];
+
+        for (int i = 0; i < bitsPerCoordinate; i++) {
+            for (int j = 0; j < vector.length; j++) {
+                if (vector[j] > thresholds[i][j]) {
```
jmazanec15, line 43: This might be good to just have a big comment above just because it is really the substance of the class.

src/main/java/org/opensearch/knn/quantization/quantizer/OneBitScalarQuantizer.java
Vikasht34: The reason why I am making like this is make Quantizer Stateless , so that we don't have to get quantizer per quantizer state.

src/main/java/org/opensearch/knn/quantization/sampler/ReservoirSampler.java
```diff
@@ -0,0 +1,88 @@
...
+            indices[i] = i;
+        }
+        for (int i = sampleSize; i < numVectors; i++) {
+            int j = ThreadLocalRandom.current().nextInt(i + 1);
+            if (j < sampleSize) {
+                indices[j] = i;
+            }
+        }
+        // Using BitSet to track the presence of indices
+        BitSet bitSet = new BitSet(numVectors);
```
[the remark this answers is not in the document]
Vikasht34: Yeah ..you are right , For now I will have simple array , becuase it's more memory beneficial, and can re-visit in future for possible optimizations.

```
    int[] indices Array:
        The indices array holds sampleSize elements, so in your example with a sample size of 25,000, the array will contain 25,000 integers.
        int  4 bytes, so the memory usage for the indices array would be 25,000 * 4 = 100,000 bytes (approximately 100 KB).

    BitSet bitSet:
        The BitSet object is used to track the presence of indices and has a size proportional to numVectors, even if not all bits are set.
        For 10 million vectors, the BitSet would require 10 million bits, which is 10,000,000 / 8 = 1,250,000 bytes (approximately 1.25 MB).

Total Memory Usage:

    For 10 million vectors and a sample size of 25,000, the total memory usage would be approximately 1.35 MB (1.25 MB for the BitSet + 100 KB for the indices array)
```

src/main/java/org/opensearch/knn/quantization/quantizer/QuantizerHelper.java
Vikasht34: I wanted to decopule framework as much as from Lucene , will take a second look on this  and if needed I can have this in future pr

src/main/java/org/opensearch/knn/quantization/factory/QuantizerRegistrar.java
Vikasht34: The synchronized  the registerDefaultQuantizers() method adds another layer of thread-safety by ensuring that the registration process itself is atomic. I prefer Redundancy for Safety.

src/main/java/org/opensearch/knn/quantization/models/quantizationOutput/BinaryQuantizationOutput.java
Vikasht34: Sure , we are following the approach where we will pass QuantizationOutput in quantize Method , which avoids creating new object for everryQuantization and I had implemented quantized output in such as way that even the underlying quantized vector input reference will also be same .

src/main/java/org/opensearch/knn/quantization/models/quantizationOutput/BinaryQuantizationOutput.java
Vikasht34: Yeah I made update to this , with the new one , we are even keeping the same refrence for quantized vector.

src/main/java/org/opensearch/knn/quantization/models/quantizationParams/ScalarQuantizationParams.java
Vikasht34: As Discussed , we are keeping this now , will Revisit this later , if we need to make some more refactor.

src/main/java/org/opensearch/knn/quantization/models/quantizationParams/ScalarQuantizationParams.java
Vikasht34: As Discussed , we are keeping this now , will Revisit this later , if we need to make some more refactor.

jmazanec15 (approved):
LGTM - thanks

ryanbogan (approved):
LGTM

navneet1v (approved):

PR #1934, opened 2024-08-05
Fix KNN module to avoid invalid characters to be included as a part of file name.

0ctopus13prime:
### Description
Issue : https://github.com/opensearch-project/k-NN/issues/1859.

#### Issue
While OpenSearch does allow for a field name to have an empty space within it and it disallows an empty space to be contained in a physical file name, [KNNCodecUtil::buildEngineFileName](https://github.com/opensearch-project/k-NN/blob/a16b8aa965f66268799c001756c78174c7baba19/src/main/java/org/opensearch/knn/index/codec/util/KNNCodecUtil.java#L106C26-L106C45) uses the field name as a part of a vector file name directly without any proper adjustment to be made. As a result, if the field name does not meet the valid criteria, it can fail due to failing the validation in [BlobStoreIndexShardSnapshot](https://github.com/opensearch-project/OpenSearch/blob/7cbff4f9fd8ae53b1672aa8e2582e23bb2c16def/server/src/main/java/org/opensearch/index/snapshots/blobstore/BlobStoreIndexShardSnapshot.java). For example, _0_2011_my vector.hnswc ('my vector' is the field name). As a result, [BlobStoreIndexShardSnapshot](https://github.com/opensearch-project/OpenSearch/blob/7cbff4f9fd8ae53b1672aa8e2582e23bb2c16def/server/src/main/java/org/opensearch/index/snapshots/blobstore/BlobStoreIndexShardSnapshot.java#L343) throws an exception complaining file name is not valid.

#### Solution
In KNNCodecUtil, we need to add a method to escape invalid characters in the field name when creating file name.
In this solution proposal, I'm suggesting to escape it to '.'. For example, `_0_2011_my vector.hnswc` → `_0_2011_my.vector.hnswc`.

Field name does not have "." inside, as it indicates more than one field name if the dot presents. 
For example, `a.b.c` indicates three fields with `a`, `b` and `c`.

AS-IS:
```
public static String buildEngineFileSuffix(String fieldName, String extension) {
    return String.format("_%s%s", fieldName, extension);
}
```

TO-BE:

```
public static String buildEngineFileSuffix(String fieldName, String extension) {
    return String.format("_%s%s", escapeSpaceInFieldName(fieldName), extension);
}

private static String escapeSpaceInFieldName(String fieldName) {
    char[] characters = fieldName.toCharArray();
    for (int i = 0 ; i < characters.length ; ++i) {
        if (Strings.INVALID_FILENAME_CHARS.contains(characters[i])) {
            characters[i] = '.';  // Overwrite with an escaping character.
        }
    }
    return new String(characters);
}
```

### Related Issues
Issue : https://github.com/opensearch-project/k-NN/issues/1859.

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/codec/util/KNNCodecUtil.java
+++ b/src/main/java/org/opensearch/knn/index/codec/util/KNNCodecUtil.java
@@ -5,12 +5,15 @@
 
 package org.opensearch.knn.index.codec.util;
 
+import com.google.common.annotations.VisibleForTesting;
 import lombok.AllArgsConstructor;
 import lombok.Getter;
 import lombok.Setter;
 import org.apache.lucene.index.BinaryDocValues;
 import org.apache.lucene.search.DocIdSetIterator;
 import org.apache.lucene.util.BytesRef;
+import org.opensearch.core.common.Strings;
+import org.opensearch.knn.common.KNNConstants;
 import org.opensearch.knn.index.codec.KNN80Codec.KNN80BinaryDocValues;
 import org.opensearch.knn.index.codec.transfer.VectorTransfer;
 
@@ -111,8 +114,45 @@ public class KNNCodecUtil {
         return String.format("%s_", segmentName);
     }
 
+    /**
+     * Build a suffix of file name with provided field name and custom extension.
+     * In case where the given field name contains invalid characters for the file name, it will escape them to '.'.
+     * Note that, field name MUST NOT contain '.' in it.
+     * <p>
+     * Ex:
+     * With 'my_vector' as field name and '.hnswc' as an extension -> _0_2011_my_vector.hnswc
+     * With 'my vector' (blank) as field name and '.hnswc' as an extension -> _0_2011_my.vector.hnswc
+     * With 'my/vector' (it has '/') as field name and '.hnswc' as an extension -> _0_2011_my.vector.hnswc
+     *
+     * @param fieldName : Vector field name
+     * @param extension : File extension.
+     * @return File suffix containing a field name that escaped with invalid characters.
+     */
     public static String buildEngineFileSuffix(String fieldName, String extension) {
-        return String.format("_%s%s", fieldName, extension);
+        return String.format("_%s%s", escapeInvalidFileNameCharsInFieldName(fieldName), extension);
+    }
+
+    /**
+     * In case of compound file, extension would be {engine-extension} + 'c' otherwise just be {engine-extension}
+     * Ex: _0_2011_my_vector.hnswc, where engine-extension is 'hnsw'.
+     */
+    public static String buildCompoundFile(String extension, boolean isCompoundFile) {
+        if (isCompoundFile) {
+            return extension + KNNConstants.COMPOUND_EXTENSION;
+        } else {
+            return extension;
+        }
+    }
+
+    @VisibleForTesting
+    static String escapeInvalidFileNameCharsInFieldName(final String fieldName) {
+        char[] characters = fieldName.toCharArray();
+        for (int i = 0; i < characters.length; ++i) {
+            if (Strings.INVALID_FILENAME_CHARS.contains(characters[i])) {
+                characters[i] = '.';
+            }
+        }
+        return new String(characters);
     }
 
     public static long getTotalLiveDocsCount(final BinaryDocValues binaryDocValues) {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Bug Fixes
 * Corrected search logic for scenario with non-existent fields in filter [#1874](https://github.com/opensearch-project/k-NN/pull/1874)
 * Add script_fields context to KNNAllowlist [#1917] (https://github.com/opensearch-project/k-NN/pull/1917)
+* Fix KNN module to avoid invalid characters to be included as a part of file name. [#1934] (https://github.com/opensearch-project/k-NN/pull/1934)
 ### Infrastructure
 ### Documentation
 ### Maintenance
--- a/src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80CompoundFormat.java
+++ b/src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80CompoundFormat.java
@@ -6,12 +6,12 @@
 package org.opensearch.knn.index.codec.KNN80Codec;
 
 import org.apache.lucene.backward_codecs.lucene50.Lucene50CompoundFormat;
-import org.opensearch.knn.common.KNNConstants;
 import org.apache.lucene.codecs.CompoundDirectory;
 import org.apache.lucene.codecs.CompoundFormat;
 import org.apache.lucene.index.SegmentInfo;
 import org.apache.lucene.store.Directory;
 import org.apache.lucene.store.IOContext;
+import org.opensearch.knn.index.codec.util.KNNCodecUtil;
 import org.opensearch.knn.index.engine.KNNEngine;
 
 import java.io.IOException;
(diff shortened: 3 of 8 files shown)


navneet1v:
@0ctopus13prime please fix the conflicts. and DCO check.

src/main/java/org/opensearch/knn/index/codec/util/KNNCodecUtil.java
```diff
@@ -112,7 +113,17 @@ public static String buildEngineFilePrefix(String segmentName) {
     }
 
     public static String buildEngineFileSuffix(String fieldName, String extension) {
-        return String.format("_%s%s", fieldName, extension);
+        return String.format("_%s%s", escapeSpaceInFieldName(fieldName), extension);
+    }
+
+    private static String escapeSpaceInFieldName(String fieldName) {
```
heemin32, line 119:
```suggestion
    private static String escapeSpaceInFieldName(final String fieldName) {
```

0ctopus13prime: Sounds good! Will update in the next revision.

heemin32 (approved):

navneet1v (approved):

src/main/java/org/opensearch/knn/index/codec/util/KNNCodecUtil.java
```diff
@@ -112,7 +113,17 @@ public static String buildEngineFilePrefix(String segmentName) {
     }
 
     public static String buildEngineFileSuffix(String fieldName, String extension) {
-        return String.format("_%s%s", fieldName, extension);
+        return String.format("_%s%s", escapeInvalidFileNameCharsInFieldName(fieldName), extension);
+    }
+
+    private static String escapeInvalidFileNameCharsInFieldName(final String fieldName) {
```
navneet1v, line 119: if we are escaping the characters here, do we need to make changes here too: https://github.com/opensearch-project/k-NN/blob/27f31684e4672f5bdb7026ac06601c4a7a123743/src/main/java/org/opensearch/knn/index/query/KNNWeight.java#L348

when we are creating the path for this file for query?

Also, what will be the exact file name on the disk will it has the escape characters?

0ctopus13prime: Q1. `if we are escaping the characters here, do we need to make changes here too?`
Ans : Yes, I think so. Will include the change in weight class in this PR as well.


Q2. `what will be the exact file name on the disk will it has the escape characters?`
The file format would be `<Version>_<Escaped Field Name><Ext>`.

Ex:
Field name=`my_vector`, then file name `_0_2011_my_vector.hnswc`
Field name=`my vector` (it includes invalid char for filename), then the resulting file name would be `_0_2011_my.vector.hnswc`

heemin32: > Q1. if we are escaping the characters here, do we need to make changes here too?
Ans : Yes, I think so. Will include the change in weight class in this PR as well.

I wonder why this is not caught through our existing test? Could you add test if the case is missing?

navneet1v: reason for that it, what being added is unit test and not IT. Hence this was missed.

navneet1v: @0ctopus13prime lets also add the unit test for KNNWeight apart from IT where we are doing unit test for getEngineFiles(). This will ensure that nothing is missed

0ctopus13prime: Coming right up

src/main/java/org/opensearch/knn/index/codec/util/KNNCodecUtil.java
```diff
@@ -111,8 +114,45 @@ public static String buildEngineFilePrefix(String segmentName) {
...
 
+    /**
+     * Build a suffix of file name with provided field name and custom extension.
+     * In case where the given field name contains invalid characters for the file name, it will escape them to '.'.
+     * Note that, field name MUST NOT contain '.' in it.
+     * <p>
+     * Ex:
+     * With 'my_vector' as field name and '.hnswc' as an extension -> _0_2011_my_vector.hnswc
+     * With 'my vector' (blank) as field name and '.hnswc' as an extension -> _0_2011_my.vector.hnswc
+     * With 'my/vector' (it has '/') as field name and '.hnswc' as an extension -> _0_2011_my.vector.hnswc
```
navneet1v, line 125: what is the behavior of Opensearch with such fields? does Opensearch adds `.` or they do something else?

heemin32: OpenSearch does not create its own file except lucene's segment file and lucene does not use field name in its file name but use random value.

src/main/java/org/opensearch/knn/index/codec/util/KNNCodecUtil.java
```diff
@@ -111,8 +114,45 @@ public static String buildEngineFilePrefix(String segmentName) {
...
+     *
+     * @param fieldName : Vector field name
+     * @param extension : File extension.
+     * @return File suffix containing a field name that escaped with invalid characters.
+     */
     public static String buildEngineFileSuffix(String fieldName, String extension) {
```
heemin32, line 131: Should we keep backward compatibility? For example, with field name `test field`, we first look for `test.field` and if there is not such file we should look for `test field`.

src/main/java/org/opensearch/knn/index/codec/util/KNNCodecUtil.java
heemin32: I think we just need to block the space in the field name. `.` is used for nested field.
For example
```
"mappings": {
    "properties": {
      "my_field": {
        "type": "nested",
        "properties": {
          "my_vector": {
            "type": "knn_vector",
            "dimension": 3
          }
      }
    }
  }
}
```
Then, file name will be `	_0_165_ my_field.my_vector.faissc`

Also, I was able to create field name using `.`. 
```
"mappings": {
    "properties": {
      "my_field.my_vector": {
        "type": "knn_vector",
        "dimension": 2,
        "method": {
          "name": "hnsw",
          "space_type": "innerproduct",
          "engine": "faiss"
        }
      }
    }
  }
```

0ctopus13prime: `a.b.c` will be splitted into three "actual" fields within Lucene.
So it will be translated to three fields where `a`, `b` and `c`.
Eventually, the field name to be passed down to `escapeSpaceInFieldName` does not contain dot ('.') character.

0ctopus13prime: For example, `my_field.my_vector` will call `escapeSpaceInFieldName` twice with each `my_field` and `my_vector`.

heemin32: KNN field might behave differently with lucene. Try to create knn index and test it locally. 
`escapeSpaceInFieldName` will be called only one time with field name `my_field.my_vector`.

0ctopus13prime: Can you share the query input you've tested?
I've tested it, and I got two fields to be passed into it.

heemin32:
```
PUT /test_index
{
  "settings": {
    "index": {
      "knn": true,
      "number_of_shards": 1
      
    }
  },
  "mappings": {
    "properties": {
      "my_field.my_vector": {
        "type": "knn_vector",
        "dimension": 2,
        "method": {
          "name": "hnsw",
          "space_type": "innerproduct",
          "engine": "lucene"
        }
      }
    }
  }
}
```

```
PUT /_bulk?refresh=true
{ "index": { "_index": "test_index", "_id": "1" } }
{"my_field.my_vector":[-2.0, 2.0], "text": "a"}
{ "index": { "_index": "test_index", "_id": "2" } }
{"my_field.my_vector":[2.0, -2.0], "text": "d"}
```

```
POST /test_index/_search
{
  "query": {
    "knn": {
      "my_field.my_vector": {
        "vector": [
          -2.0,
          2.0
        ],
        "k": 2
      }
    }
  }
}
```

heemin32: I mean in `KNNWeight` not in `KNN80CompountFormat`

0ctopus13prime:
Will discard this PR and let me re-raise a new PR that prohibits all disallowed characters for any vector fields.

0ctopus13prime:
Raised a new PR to block invalid characters for a physical file name entirely

https://github.com/opensearch-project/k-NN/pull/1936

PR #1936, opened 2024-08-07
Disallow invalid characters for physical file name to be included within vector field name.

0ctopus13prime:
### Description
Issue : https://github.com/opensearch-project/k-NN/issues/1859.

#### Issue
While OpenSearch does allow for a field name to have an empty space within it and it disallows an empty space to be contained in a physical file name, [KNNCodecUtil::buildEngineFileName](https://github.com/opensearch-project/k-NN/blob/a16b8aa965f66268799c001756c78174c7baba19/src/main/java/org/opensearch/knn/index/codec/util/KNNCodecUtil.java#L106C26-L106C45) uses the field name directly as a part of a vector file name. As a result, in case where the field name had one of disallowed character for a physical file name, it fails in validation of [BlobStoreIndexShardSnapshot](https://github.com/opensearch-project/OpenSearch/blob/7cbff4f9fd8ae53b1672aa8e2582e23bb2c16def/server/src/main/java/org/opensearch/index/snapshots/blobstore/BlobStoreIndexShardSnapshot.java). For example, `_0_2011_my vector.hnswc` (where 'my vector' is the field name). As a result, [BlobStoreIndexShardSnapshot](https://github.com/opensearch-project/OpenSearch/blob/7cbff4f9fd8ae53b1672aa8e2582e23bb2c16def/server/src/main/java/org/opensearch/index/snapshots/blobstore/BlobStoreIndexShardSnapshot.java#L343) throws an exception complaining file name is not valid.

#### Solution
Add a validation logic to throw an exception in case provided vector field name has any invalid characters.

```
private void validateFullFieldName(BuilderContext context) {
    final String fullFieldName = buildFullName(context);
    for (char ch : fullFieldName.toCharArray()) {
        if (Strings.INVALID_FILENAME_CHARS.contains(ch)) {
            throw new IllegalArgumentException(...);
        }
    }
}


public abstract class KNNVectorFieldMapper extends ParametrizedFieldMapper {
    ...
    public static class Builder extends ParametrizedFieldMapper.Builder {
        @Override
        public KNNVectorFieldMapper build(BuilderContext context) {
            validateFullFieldName(context);
            ...
```

### Related Issues
Issue : https://github.com/opensearch-project/k-NN/issues/1859.

### Check List


src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java
```diff
@@ -416,6 +421,32 @@ private KNNEngine validateDimensions(final KNNMethodContext knnMethodContext, fi
...
+
+        /**
+         * Validate whether provided full field name contain any invalid characters for physical file name.
+         * At the moment, we use a field name as a part of file name while we throw an exception
+         * if a physical file name contains any invalid characters when creating snapshot.
+         * To prevent from this happening, we restrict vector field name and make sure generated file to have a valid name.
+         *
+         * @param context : Builder context to have field name info.
+         */
+        private void validateFullFieldName(BuilderContext context) {
```
heemin32, line 433:
```suggestion
        private void validateFullFieldName(final BuilderContext context) {
```

src/test/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapperTests.java
```diff
@@ -1049,6 +1050,22 @@ public void testBuilder_whenBinaryWithLegacyKNNEnabled_thenException() {
         assertTrue(ex.getMessage(), ex.getMessage().contains("is not supported for"));
     }
 
+    public void testWhenInvalidCharsInFieldName() {
```
heemin32, line 1053:
```suggestion
    public void testBuild_whenInvalidCharsInFieldName_thenThrowException() {
```

src/test/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapperTests.java
```diff
@@ -1049,6 +1050,22 @@ public void testBuilder_whenBinaryWithLegacyKNNEnabled_thenException() {
...
+            final String invalidVectorFieldName = "fieldname" + disallowChar;
+
+            // Prepare context.
+            Settings settings = Settings.builder().put(settings(CURRENT).build()).put(KNN_INDEX, true).build();
+            Mapper.BuilderContext builderContext = new Mapper.BuilderContext(settings, new ContentPath());
+
+            // IllegalArgumentException should be thrown.
+            assertThrows(IllegalArgumentException.class, () -> {
+                new KNNVectorFieldMapper.Builder(invalidVectorFieldName, null, CURRENT).build(builderContext);
+            });
```
heemin32, line 1065: Let's verify error message as well.

```
Exception e = assertThrows(...
assertTrue(e.getMessage, e.getMessage.contains("Vector field name must not include"));
```

0ctopus13prime: Sounds good!

src/test/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapperTests.java
0ctopus13prime: Sounds good!

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java
0ctopus13prime: Thank you for the suggestion, but I don't think I would add `final` keyword unless you have strong opinion on this.
Unlike `const` keyword in C++, I hardly think introducing `final` adds any values other than prohibiting reassignment.
As you can see in the method, it just merely bypasses the reference, and it would not make any differences whether if we added `final`.

Of course, please feel free drop any thoughts on this! :)

heemin32: It is just coding convention we are following. KNN does not have published document about it though. We might need one..

0ctopus13prime: Oh.. if that is the case then I will update accordingly!

0ctopus13prime: Updated!

src/test/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapperTests.java
```diff
@@ -1049,6 +1050,23 @@ public void testBuilder_whenBinaryWithLegacyKNNEnabled_thenException() {
...
+
+            // Prepare context.
+            Settings settings = Settings.builder().put(settings(CURRENT).build()).put(KNN_INDEX, true).build();
+            Mapper.BuilderContext builderContext = new Mapper.BuilderContext(settings, new ContentPath());
+
+            // IllegalArgumentException should be thrown.
+            Exception e = assertThrows(IllegalArgumentException.class, () -> {
+                new KNNVectorFieldMapper.Builder(invalidVectorFieldName, null, CURRENT).build(builderContext);
+            });
+            assertTrue(e.getMessage().contains("Vector field name must not include"));
```
heemin32, line 1066: nit. print `e.getMessage()` so that we know what is the original message.
```suggestion
            assertTrue(e.getMessage(), e.getMessage().contains("Vector field name must not include"));
```

heemin32 (dismissed):

src/test/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapperTests.java
0ctopus13prime: Sounds good! Will update it shortly

heemin32 (dismissed):

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java
```diff
@@ -416,6 +421,32 @@ private KNNEngine validateDimensions(final KNNMethodContext knnMethodContext, fi
...
+
+        /**
+         * Validate whether provided full field name contain any invalid characters for physical file name.
+         * At the moment, we use a field name as a part of file name while we throw an exception
+         * if a physical file name contains any invalid characters when creating snapshot.
+         * To prevent from this happening, we restrict vector field name and make sure generated file to have a valid name.
+         *
+         * @param context : Builder context to have field name info.
+         */
+        private void validateFullFieldName(final BuilderContext context) {
```
navneet1v, line 433: is this check happening during every document ingestion or happening once during index creation?

0ctopus13prime: That depends on the whether dynamic mapping is enabled.
If it was enabled, then `MapperService` will walk through the documents to identify whether there are any new fields added, and during which, the check will be done along with other validation checks.
If it was not enabled, it won't be called.

navneet1v: @0ctopus13prime make sense. What tests we have performed to validate this behavior? Reason I am asking this is if this check is performed per document then this check will incur latency in the overall indexing time. Hence I want to make sure that this check is done once.

0ctopus13prime: I don't think it will add up trivial latency because of the changes, as the validation check that I added is merely walk through field char[] to look up hash set. Overall time complexity is bounded by `O(N)` where N = the number of chars in vector field. Presumably this will only add up nano scale latency in the end, but only affecting vector fields.

But if we're being conservative, then I can run TPS benchmark and compare the metrics between baseline.
What's your thought?

navneet1v: Yeah I understand the latency is in nanos. But would be interesting is if you can validate/provide details how you tested to know that this function runs once. We don't need benchmarks here.

0ctopus13prime: Mostly done in my local dev environment with debugger attached in IDE.

Launch a server in IDE (I got rid of all segments and repos every time I launch a new server)
1. Tried to create a vector index with invalid vector field name. Ex: `my vector`. And validated the request is failing.
```
curl -X PUT 'http://localhost:9200/knn-index/' -H 'Content-Type: application/json' -d '
{
  "settings": {
    "index": {
      "knn": true,
      "knn.algo_param.ef_search": 100
    }
  },
  "mappings": {
    "properties": {
      "my vector": { <------------ This must fail
        "type": "knn_vector",
        "dimension": 2
      }
    },
    "dynamic": false | true <-------- Enable or disable dynamic mapping
  }
}
'
```

Response:
```
{
  "error": {
    "root_cause": [
      {
        "type": "resource_already_exists_exception",
        "reason": "index [knn-index/kOKM8elaSeacwnNGF11ZMg] already exists",
        "index": "knn-index",
        "index_uuid": "kOKM8elaSeacwnNGF11ZMg"
      }
    ],
    "type": "resource_already_exists_exception",
    "reason": "index [knn-index/kOKM8elaSeacwnNGF11ZMg] already exists",
    "index": "knn-index",
    "index_uuid": "kOKM8elaSeacwnNGF11ZMg"
  },
  "status": 400
}
```

2. Tried to ingest with both cases where dynamic mapping is enabled. In this case, I created a vector index with field name `my_vector` (valid one)
```
curl -X POST 'http://localhost:9200/_bulk' -H 'Content-Type: application/json' -d '{ "index": { "_index": "knn-index", "_id": "1" } }
   { "my_vector": [1.5, 2.5], "price": 12.2 }
   { "index": { "_index": "knn-index", "_id": "2" } }
   { "my_vector": [2.5, 3.5], "price": 7.1 }
   { "index": { "_index": "knn-index", "_id": "3" } }
   { "my_vector": [3.5, 4.5], "price": 12.9 }
   { "index": { "_index": "knn-index", "_id": "4" } }
   { "my_vector": [5.5, 6.5], "price": 1.2 }
   { "index": { "_index": "knn-index", "_id": "5" } }
   { "my_vector": [4.5, 5.5], "price": 3.7 }
'
```

Validated : 
- `validateFullFieldName` was called for each `my_vector`.


I tried to put everything I have tested so far, but if you need additional info, please feel free to drop a comment!

navneet1v: @0ctopus13prime this tells me that for every document indexing this function will be called. What would be the case in which this function won't be called.

0ctopus13prime: @navneet1v 
As I said, if `dynamic` is disabled, the validation check that I add will not be called.
But if `dynamic` is enabled, during the document parsing, it will invoke the function along with others.

```
        public KNNVectorFieldMapper build(Mapper.BuilderContext context) {
            this.validateFullFieldName(context); <------- This
            KNNMethodContext knnMethodContext = (KNNMethodContext)this.knnMethodContext.getValue();
            this.setDefaultSpaceType(knnMethodContext, (VectorDataType)this.vectorDataType.getValue());
            this.validateSpaceType(knnMethodContext, (VectorDataType)this.vectorDataType.getValue());
            this.validateDimensions(knnMethodContext, (VectorDataType)this.vectorDataType.getValue());
            this.validateEncoder(knnMethodContext, (VectorDataType)this.vectorDataType.getValue());
```

navneet1v (approved):
Code looks good to me. Please fix the conflicts.

ryanbogan (approved):
LGTM!

heemin32 (approved):

navneet1v (approved):

PR #1945, opened 2024-08-09
Integrate Lucene Vector field with native engines to use KNNVectorFormat during segment creation

navneet1v:
### Description
Integrate Lucene Vector field with native engines, to use KNNVectorFormat during segment creation

### What has changed and why:
1. With this change I added the capability to use Lucene based vector field to for native engines. This feature is currently added behind a cluster setting to ensure that 2.x and main branch builds don't break. Once the code for adding the vector data structures is added for new KNNVectorsFormat this setting will be removed and new ITs will also be added.
2. There is a loose attribute with name isIndexKNN added which will be refactored once this PR https://github.com/opensearch-project/k-NN/pull/1939 is merged with FlatVectorsMapper class.
3. If `index.knn` is false then we will use DocValuesBased Vector Field. Because if we use the Lucene based vector field then KNNCodec will not be triggered and default KNNFormat will be used which will create the HNSW graph as that is what default behavior of Lucene library. This is the reason why indexKNN check is added while deciding which VectorField to use.

### Related Issues
https://github.com/opensearch-project/k-NN/issues/1853

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java
+++ b/src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java
@@ -16,7 +16,8 @@ import java.util.function.Supplier;
 import lombok.extern.log4j.Log4j2;
 import org.apache.lucene.document.Field;
 import org.apache.lucene.document.FieldType;
-import org.apache.lucene.index.DocValuesType;
+import org.apache.lucene.document.KnnByteVectorField;
+import org.apache.lucene.document.KnnFloatVectorField;
 import org.apache.lucene.index.IndexOptions;
 import org.opensearch.Version;
 import org.opensearch.common.Explicit;
@@ -456,6 +457,7 @@ public abstract class KNNVectorFieldMapper extends ParametrizedFieldMapper {
     protected boolean hasDocValues;
     protected VectorDataType vectorDataType;
     protected ModelDao modelDao;
+    protected boolean useLuceneBasedVectorField;
 
     // We need to ensure that the original KNNMethodContext as parsed is stored to initialize the
     // Builder for serialization. So, we need to store it here. This is mainly to ensure that the legacy field mapper
@@ -497,16 +499,29 @@ public abstract class KNNVectorFieldMapper extends ParametrizedFieldMapper {
         parseCreateField(context, fieldType().getKnnMappingConfig().getDimension(), fieldType().getVectorDataType());
     }
 
+    private Field createVectorField(float[] vectorValue) {
+        if (useLuceneBasedVectorField) {
+            return new KnnFloatVectorField(name(), vectorValue, fieldType);
+        }
+        return new VectorField(name(), vectorValue, fieldType);
+    }
+
+    private Field createVectorField(byte[] vectorValue) {
+        if (useLuceneBasedVectorField) {
+            return new KnnByteVectorField(name(), vectorValue, fieldType);
+        }
+        return new VectorField(name(), vectorValue, fieldType);
+    }
+
     /**
      * Function returns a list of fields to be indexed when the vector is float type.
      *
      * @param array array of floats
-     * @param fieldType {@link FieldType}
      * @return {@link List} of {@link Field}
      */
-    protected List<Field> getFieldsForFloatVector(final float[] array, final FieldType fieldType) {
+    protected List<Field> getFieldsForFloatVector(final float[] array) {
         final List<Field> fields = new ArrayList<>();
-        fields.add(new VectorField(name(), array, fieldType));
+        fields.add(createVectorField(array));
         if (this.stored) {
             fields.add(createStoredFieldForFloatVector(name(), array));
         }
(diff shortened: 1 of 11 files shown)


navneet1v:
Checking why CIs are failing. it doesn't seem the issue with the code.

jmazanec15:
> 3. If index.knn is false and we use the Lucene based vector field then KNNCodec will not be triggered and default KNNFormat will be used which will create the HNSW graph as that is what default behavior of Lucene library.

Does this mean if someone upgrades their index with knn=false, it is going to switch from binary doc values to lucene?

navneet1v:
> > 3. If index.knn is false and we use the Lucene based vector field then KNNCodec will not be triggered and default KNNFormat will be used which will create the HNSW graph as that is what default behavior of Lucene library.
> 
> Does this mean if someone upgrades their index with knn=false, it is going to switch from binary doc values to lucene?

To limit the scope of this pr and refactoring you are doing if its not a knn index we will still BDV

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java
```diff
@@ -532,18 +550,55 @@ protected List<Field> getFieldsForFloatVector(final float[] array, final FieldTy
...
-        fields.add(new VectorField(name(), array, fieldType));
+        fields.add(createVectorField(array, dimension));
         if (this.stored) {
             fields.add(createStoredFieldForByteVector(name(), array));
         }
         return fields;
     }
 
+    private Field createVectorField(float[] vectorValue, int dimension) {
+        final FieldType tempFieldType = new FieldType(fieldType);
```
jmazanec15, line 565: Why do we need to create this here as opposed to KNNFieldMapper Construction? This will create temp field per doc parsed right?

navneet1v: This is a valid question. So, I have to do it because I have to add the vector in the fieldType and once the constructor a mapper is done it freeze the field. Here: https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/mapper/MethodFieldMapper.java#L68, same is there for other mappers.

So, if we want to remove this temp field I have to remove the freeze part from all the mappers and then do it here. I think that is a better option and I can avoid the tempFieldType creation. 

Does that sound good?

jmazanec15: Im not sure Im following. We need to add the vector in the field not the field type. We can reuse the fieldType instead of recreating.

navneet1v: @jmazanec15 yes we can reuse the fieldType for all the mappers except ModelFieldMapper. Because when the constructor of the ModelFieldMapper is called we just have modelId and as per this comment cluster state is not initialized at that time [ref](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/mapper/ModelFieldMapper.java#L49-L52) so we don't have access to information like VectorDataType, dimensions etc, which is required for Lucene Vector Field.


Now I see couple of ways here:
1. We can keep this code of creating the fieldType using a temp for every document ingestion for all the mappers.
2. We can skip doing the `fieldType.freeze()` for ModelFieldMapper and set attributes during [here](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/mapper/ModelFieldMapper.java#L64).
3. We use this tempFieldType behavior for only ModelFieldMapper.

jmazanec15: > We can skip doing the fieldType.freeze() for ModelFieldMapper and set attributes during [here](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/mapper/ModelFieldMapper.java#L64).

I would be in favior of this approach similar to how were handling the init's (i.e. initPerDimensionValidator)

navneet1v: Let me take a look. Since the code has been updated now.

navneet1v: Fixed.

navneet1v:
> > 3. If index.knn is false and we use the Lucene based vector field then KNNCodec will not be triggered and default KNNFormat will be used which will create the HNSW graph as that is what default behavior of Lucene library.
> 
> Does this mean if someone upgrades their index with knn=false, it is going to switch from binary doc values to lucene?

No.

jmazanec15 (dismissed):
Looks good. Needs a rebase but approving

navneet1v:
> Looks good. Needs a rebase but approving

Thanks. I am fixing the conflicts. Will raise the PR in few hours.

jmazanec15 (approved):

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapperUtil.java
```diff
@@ -238,6 +238,22 @@ static void validateIfKNNPluginEnabled() {
...
+     * We will use LuceneKNNVectorsFormat when these below condition satisfy:
+     * <ol>
+     *  <li>Index is created with Version of opensearch >= 2.17</li>
+     *  <li>Cluster setting is enabled to use Lucene KNNVectors format. This condition is temporary condition and will be
+     * removed before release.</li>
+     * </ol>
+     * @param indexCreatedVersion {@link Version}
+     * @return true if vector field should use KNNVectorsFormat
+     */
+    static boolean useLuceneKNNVectorsFormat(final Version indexCreatedVersion) {
```
ryanbogan, line 253: Once the second check is removed before release, should we move this check to `IndexUtil` with the rest of our version/feature maps?

ryanbogan (approved):
One small comment for the future, but LGTM!

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapperUtil.java
navneet1v: Sure.

Issue #1885, opened 2024-07-25
Move away from file watcher for releasing memory


jmazanec15:
## Description

Right now, we free memory for segments whose files get deleted by having a resource watcher watch for those files:
1. Init of resource watcher - https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/plugin/KNNPlugin.java#L196
2. File change watcher initialization - https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/memory/NativeMemoryLoadStrategy.java#L84
3. File watcher - https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/memory/NativeMemoryLoadStrategy.java#L97

This file watching can actually be completely removed and it would do some good in simplifying the code (less static initialization, decoupling between cache manager and native memory load strategy, moving away from tight coupling of FSDirectory, etc). And additionally, how we do it now is a bug - see #1012.

To fix this, we just need to ensure that for either our DocValuesProducer and our KnnVectorsReader, when close is called, we evict any indices from memory (similar to how we do it now with the filewatcher). We would actually need to implement a really light weight DocValuesProducer that delegates for everything.

## Related issues
#1012

luyuncheng:
hi @jmazanec15 i am wring a DocValuesProducer for something optimize. i like this very much. we can release memory efficiently. i can help contribute my DocValuesProducer

heemin32:
Is the lifecycle of native memory(native knn index) same as the lifecycle of DocValuesProducer or KnnVectorsReader? 
I thought the native memory stays in memory until either we clear cache or the index is unusable(delete or close). 

Does `close` is called for either DocValuesProducer or KnnVectorsReader when the index is closed or deleted?

jmazanec15:
> hi @jmazanec15 i am writing a DocValuesProducer for KNN80Codec for save store usage. i like this issues very much. it can release memory efficiently. i can help contribute my DocValuesProducer if you like

Awesome! What custom functionality did you add  in DVProducer to reduce storage? I think the change for this issue would be quite small so I do not see it conflicting.

> Is the lifecycle of native memory(native knn index) same as the lifecycle of DocValuesProducer or KnnVectorsReader?

[...]

PR #1946, opened 2024-08-09
Add DocValuesProducers for releasing memory when close index

luyuncheng:
### Description
in #1885 we talk about a method that we need release memory when a producer closed. so i open this PR and added `KNN80DocValuesProducer`. this producer can release memory when reader closed a segment. 

also i added `refCount`  as the [comments](https://github.com/opensearch-project/k-NN/issues/1885#issuecomment-2254576901) cares about.

i think this pr is the 1st step, we only added a producers, because we talked in #1885, that we need get `binaryDocValues` in `DocValuesProducers` from native engine. 

i see there is an #1853 on going, so i prefer to in next step to read binaryDocValues from native engines. 

### Related Issues
https://github.com/opensearch-project/k-NN/issues/1885

### Check List


navneet1v:
@luyuncheng can we fix the build CIs. BWC CIs is failing across other PRs will check whats happening there.

src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80CompoundDirectory.java
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    @Override
+    public long fileLength(String name) throws IOException {
+        return delegate.fileLength(name);
+    }
+
+    @Override
+    public IndexInput openInput(String name, IOContext context) throws IOException {
+        return delegate.openInput(name, context);
```
jmazanec15, line 46: For future extendability, could we check if this is a k-NN file and open it from the directory? I know we typically dont open k-NN files via openInput, but if we do, this might fail:
```
        if (KNNEngine.getEnginesThatCreateCustomSegmentFiles().stream().anyMatch(engine -> name.endsWith(engine.getCompoundExtension()))) {
            return directory.openInput(name, context);
        }

        return delegate.openInput(name, context);
```

src/main/java/org/opensearch/knn/index/memory/NativeMemoryAllocation.java
```diff
@@ -159,17 +181,24 @@ class IndexAllocation implements NativeMemoryAllocation {
...
             executor.execute(() -> {
                 writeLock();
                 cleanup();
                 writeUnlock();
             });
         }
 
+        @Override
+        public void close() {
+            if (!closed && refCounted.refCount() > 0) {
```
jmazanec15, line 197: Wait if refCount is > 0, we dont want to close right?

src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80DocValuesProducer.java
```diff
@@ -0,0 +1,155 @@
...
+     * Get KNNEngine From FieldInfo
+     * @param field which field we need produce from engine
+     * @return if and only if Native Engine we return specific engine, else return Lucene Engine
+     */
+    private KNNEngine getKNNEngine(@NonNull FieldInfo field) {
+
+        final String modelId = field.attributes().get(MODEL_ID);
+        if (modelId != null) {
+            var model = ModelCache.getInstance().get(modelId);
+            return KNNEngine.LUCENE;
```
jmazanec15, line 131: Why Lucene here?

src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80DocValuesProducer.java
```diff
@@ -0,0 +1,155 @@
...
+        }
+        final String engineName = field.attributes().get(KNNConstants.KNN_ENGINE);
+        if (engineName == null) {
+            return KNNEngine.LUCENE;
```
jmazanec15, line 135: Also why would this be Lucene?

jmazanec15 (commented):
Thanks for making this @luyuncheng. Overall looks pretty good. A few comments. I think it is okay to initially focus on DocValues and then add functionality for KnnVectorsFormat in another PR (and get rid of file watcher in this PR).

src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80DocValuesProducer.java
```diff
@@ -0,0 +1,155 @@
...
+        this.delegate = delegate;
+        this.state = state;
+        this.nativeMemoryCacheManager = NativeMemoryCacheManager.getInstance();
+
+        Directory directory = state.directory;
+        // directory would be CompoundDirectory, we need get directory firstly and then unwrap
+        if (state.directory instanceof KNN80CompoundDirectory) {
+            directory = ((KNN80CompoundDirectory) state.directory).getDir();
+        }
+        String directoryPath = ((FSDirectory) FilterDirectory.unwrap(directory)).getDirectory().toString();
```
navneet1v, line 63: will this be backward compaitble?

src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80DocValuesProducer.java
```diff
@@ -0,0 +1,155 @@
...
+            return KNNEngine.LUCENE;
+        }
+        final String engineName = field.attributes().get(KNNConstants.KNN_ENGINE);
+        if (engineName == null) {
+            return KNNEngine.LUCENE;
+        }
+        return KNNEngine.getEngine(engineName);
+    }
+
+    List<String> getEngineFiles(String extension, String fieldName) {
```
navneet1v, line 140: I see similar logic in KNNWeight class too, can we put all the common logic in helpers/utils.

kotwanikunal:
Added some thoughts related to the cache free up: https://github.com/opensearch-project/k-NN/issues/1885#issuecomment-2284961955

src/main/java/org/opensearch/knn/index/memory/NativeMemoryAllocation.java
luyuncheng: in concurrent scenarios, there is an search happen before the close. and we do can not assume `KNNWeight#doANNSearch` or `IndexShard#Close` or `ClearCache` has already returned like async close shard/segment scenarios. so i prefer to keep the memory incase `SegmentFault`.

src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80DocValuesProducer.java
luyuncheng: LGTM, at https://github.com/opensearch-project/k-NN/blob/ee0b7d4217b0db76626973fe4209cc429a6147c3/src/main/java/org/opensearch/knn/index/codec/util/KNNCodecUtil.java#L132-L146 i moved into `KNNCodecUtil.java`

src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80DocValuesProducer.java
luyuncheng: nice catch, i just want to get nativeEngine file path. it is truly confused, so at https://github.com/opensearch-project/k-NN/blob/ee0b7d4217b0db76626973fe4209cc429a6147c3/src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80DocValuesProducer.java#L123-L144 i using specific KNNEngine enum or null

src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80DocValuesProducer.java
luyuncheng: nice catch, i just want to get nativeEngine file path. it is truly confused, so at https://github.com/opensearch-project/k-NN/blob/ee0b7d4217b0db76626973fe4209cc429a6147c3/src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80DocValuesProducer.java#L123-L144 i using specific KNNEngine enum or null

src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80CompoundDirectory.java
```diff
@@ -0,0 +1,59 @@
...
+import lombok.Getter;
+import org.apache.lucene.codecs.CompoundDirectory;
+import org.apache.lucene.store.Directory;
+import org.apache.lucene.store.IOContext;
+import org.apache.lucene.store.IndexInput;
+
+import java.io.IOException;
+import java.util.Set;
+
+public class KNN80CompoundDirectory extends CompoundDirectory {
```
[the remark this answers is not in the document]
luyuncheng: in the refresh stack
`IndexShard.openEngineAndRecoverFromTranslog` -> `InternalEngine.refresh`  -> `StandardDirectoryReader.open` -> `SegmentDocValues.newDocValuesProducer` -> `KNN80DocValuesFormat.fieldsProducer` it would throw Exception

> java.lang.ClassCastException: class org.apache.lucene.codecs.lucene90.Lucene90CompoundReader cannot be cast to class org.apache.lucene.store.FSDirectory (org.apache.lucene.codecs.lucene90.Lucene90CompoundReader and org.apache.lucene.store.FSDirectory are in unnamed module of loader 'app')

i think it is because `Lucene90CompoundReader` is an final class, we can not cast it directly across module. we need an wrap for `Lucene#Codec#CompoundDirectory` which reader is already be an wrap directory

src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80DocValuesProducer.java
luyuncheng: i think it has been backward compatible as CI tests passed

jmazanec15 (commented):
One minor comment - other than that looks good!

luyuncheng:
> Can we check if FilterDirectory unwraps to FSDirectory before casting? If it does not, can we warn log and NO-OP? My main concern here is around remove directory implementations.

@jmazanec15  at https://github.com/opensearch-project/k-NN/pull/1946/commits/2bcc13946a5c26a25f0183321ecdba876dc1d40a how about added try catch `ClassCastException`

src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80DocValuesProducer.java
luyuncheng: @jmazanec15 at https://github.com/opensearch-project/k-NN/pull/1946/commits/2bcc13946a5c26a25f0183321ecdba876dc1d40a how about added try catch `ClassCastException`

jmazanec15 (approved):
That looks good. Thanks approving!

src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80CompoundDirectory.java
navneet1v: @jmazanec15 why we want to do this? this can lead to unnecessary mapping of the file in memory. My thoughts will be not to do this.

src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80DocValuesProducer.java
```diff
@@ -0,0 +1,154 @@
...
+            directory = ((KNN80CompoundDirectory) state.directory).getDir();
+        }
+
+        FSDirectory fsDirectory = null;
+        try {
+            fsDirectory = ((FSDirectory) FilterDirectory.unwrap(directory));
+        } catch (ClassCastException ex) {
+            log.warn("{} can not casting to FSDirectory", directory.toString());
+            return;
+        }
```
navneet1v, line 68: can we change this behavior of class casting to say instance of check.
Directory dir = FilterDirectory.unwrap(directory))
if ((dir instanceof FSDirectory) == false) {
   log.warn("{} can not casting to FSDirectory", directory);
}

src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80DocValuesProducer.java
```diff
@@ -0,0 +1,154 @@
...
+    @Override
+    public void close() {
+        try {
+            delegate.close();
+            for (String path : indexPathMap.values()) {
+                nativeMemoryCacheManager.invalidate(path);
+            }
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
```
navneet1v, line 125: if during the invalidation we get exception from 1 file then it will lead to throwing of exceptions and other graph files will still remain in cache. How we are handling that case?

src/main/java/org/opensearch/knn/index/codec/util/KNNCodecUtil.java
```diff
@@ -102,4 +106,20 @@ public static long getTotalLiveDocsCount(final BinaryDocValues binaryDocValues)
         }
         return totalLiveDocs;
     }
+
+    public static List<String> getEngineFiles(String extension, String fieldName, SegmentInfo segmentInfo) {
```
navneet1v, line 110: please add java docs on the public function.

and also unit test for this function.

src/main/java/org/opensearch/knn/index/memory/NativeMemoryAllocation.java
```diff
@@ -80,6 +81,26 @@ public interface NativeMemoryAllocation {
...
+     * Decreases the refCount of this  instance. If the refCount drops to 0, then this
+     * instance is considered as closed and should not be used anymore.
+     *
+     * @see #incRef
+     *
+     * @return returns {@code true} if the ref count dropped to 0 as a result of calling this method
+     */
+    default boolean decRef() {
+        return true;
+    }
```
navneet1v, line 102: > and should not be used anymore.

what is the meaning of this? Are you suggesting that is ref is not available for eviction?

luyuncheng: this comes from `org.opensearch.common.util.concurrent.RefCounted#decRef` and in `NativeMemoryAllocation.IndexAllocation.RefCountedReleasable#decRef` it would calling `closeInternal` when `count == 0`.  so this means: when count drops to 0, this allocation can not use anymore.

src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80DocValuesProducer.java
luyuncheng: LGTM

src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80DocValuesProducer.java
luyuncheng: nice catch, let's move invalidate cache ahead

src/main/java/org/opensearch/knn/index/codec/util/KNNCodecUtil.java
luyuncheng: > please add java docs on the public function.

LGTM

src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80DocValuesProducer.java
```diff
@@ -0,0 +1,154 @@
...
+            return;
+        }
+        assert fsDirectory != null;
+        String directoryPath = fsDirectory.getDirectory().toString();
+        for (FieldInfo field : state.fieldInfos) {
+            if (!field.attributes().containsKey(KNN_FIELD)) {
+                continue;
+            }
+            // Only Native Engine put into indexPathMap
+            KNNEngine knnEngine = getNativeKNNEngine(field);
```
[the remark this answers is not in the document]
luyuncheng: getNativeKNNEngine  just want to verify the native Engine type, and get engine extension like `KNN80DocValuesConsumer` do, but we need ignore model's engine. so i think `FieldInfoExtractor` would not be hear

src/main/java/org/opensearch/knn/index/codec/util/KNNCodecUtil.java
luyuncheng: FIXED at https://github.com/opensearch-project/k-NN/pull/1946/commits/3b5751c08732d0a147880627fd44ae8b3196fe98

src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80DocValuesProducer.java
luyuncheng: FIXED at https://github.com/opensearch-project/k-NN/pull/1946/commits/3b5751c08732d0a147880627fd44ae8b3196fe98

src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80DocValuesProducer.java
luyuncheng: FIXED at https://github.com/opensearch-project/k-NN/pull/1946/commits/3b5751c08732d0a147880627fd44ae8b3196fe98

src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80CompoundDirectory.java
jmazanec15: I dont think anyone should call openInput on k-NN files. But if they do, `return delegate.openInput(name, context);
` will throw exception because file isnt actually in compound file

navneet1v: I am not sure if exception will come. @luyuncheng have you seen an exception here? I think it will open the input. 

But even if there is no exception we should not open the input, it is just unnecessary mapping of file. One we start to use IndexInput to read the graph file(ref: https://github.com/opensearch-project/k-NN/issues/1951) we can start opening the files here.

src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80DocValuesProducer.java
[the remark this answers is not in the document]
navneet1v: > getNativeKNNEngine just want to verify the native Engine type

Then in that case the name of the function is misleading. More so you are not just validating the Native engines you are validating if they are faiss/Nmslib. Atleast the way I would recommend here to do is, use the FieldInfoExtractor to find the engine and then compare it againsts the list of engines you want this logic to run.

This will ensure that code is clearly telling what it is doing, also the logic of extracting KNNEngine is in 1 place. :)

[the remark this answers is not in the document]
luyuncheng: > use the FieldInfoExtractor to find the engine

i am a little bit confused, `FieldInfoExtractor` only enum the field type, when there is Lucene engine which is not native engine, that there is no need to added into `indexPathMap` which in `KNN80DocValuesProducer`, but `FieldInfoExtractor` always return `FLOAT` as the fact we need nativeEngine type.

and `getNativeKNNEngine` function java docs says `if and only if Native Engine we return specific engine, else return null` so we can get native engine type and use common logic `KNNCodecUtil.getEngineFiles`

src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80CompoundDirectory.java
luyuncheng: i did not find any exception. but i do want to implement an nativeEngine indexInput like #1951 says. 
how about:
1. we return `return delegate.openInput(name, context);` 
2. and added a assert when name is native engines. 
3. also added an TODO comments here: TODO: using the native engine IndexInput.

@navneet1v @jmazanec15

jmazanec15: @navneet1v See how we write: https://github.com/luyuncheng/k-NN-1/blob/DocValuesProducers/src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80CompoundFormat.java#L48-L53. We dont use the delegate to write the file, we use directory directly.

src/main/java/org/opensearch/knn/index/codec/util/KNNCodecUtil.java
navneet1v: Having a separate UT will be better KNNWeight now will not be the only consumer of this function

src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80CompoundDirectory.java
navneet1v: @jmazanec15 I understand we would be directly writing the file. I have tested a small code change where in KNNWeight class I was opening .hnsw file using IndexInput and it worked. I didn't get the exception. So want to know when you say there will be exception what will be that exception.

Neverthless I believe we should not even open the nativeIndex files in the producer as it will unnecessary map the file on RAM.

jmazanec15: In your test was it an .hnsw file or .hnswc? I think .hnswc should fail with this line

src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80DocValuesProducer.java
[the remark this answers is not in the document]
navneet1v: @luyuncheng I think you need to pull in new changes from main. There is a new function added in FieldInfoExtractor. Ref: https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/common/FieldInfoExtractor.java#L30-L36

> also getNativeKNNEngine function java docs says if and only if Native Engine we return specific engine, else return null so we can get native engine type and use common logic KNNCodecUtil.getEngineFiles

I understand your concern and yes `FieldInfoExtractor.extractKNNEngine` will give the engine if present or give the default engine. But in this case we want to do 2 things:
1. Get the engine if the engine is set.
2. If the engine is set then it should not be Lucene Engine.


[Suggestion] Here is something I think we can do:
1. `field.attributes().containsKey(KNNVectorFieldMapper.KNN_FIELD)` use this to know if the field is KNNType or not.
2. If the field is KNNType then get the engine out of FieldInfo.
3. Validate if engine == FAISS or NMSLIB.

I see 2 is already taken care by `FieldInfoExtractor.extractKNNEngine` so if we enhance `FieldInfoExtractor.extractKNNEngine` to add 1 we can reuse the code, and ensure we have a very tight implementation of extracting engine from FieldInfo. Rather than having logic to extract kNN engine at multiple places.

The 3 you can do in a private function in this class.

Let me know if this sounds good. My intension here is to ensure that this logic of getting KNNEngine is at one place and not duplicated.

navneet1v:
@luyuncheng can you please look at the comment added reply on them.

luyuncheng:
> @luyuncheng can you please look at the comment added reply on them.

@navneet1v Sorry for the late reply, i just get back from vacation. i'll do it ASAP.

src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80DocValuesProducer.java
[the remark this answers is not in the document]
luyuncheng: i like the idea! so in https://github.com/opensearch-project/k-NN/pull/1946/commits/07806261e34cd4d22fa98897fad4b60a65181d57 i try to use `FieldInfoExtractor.extractKNNEngine`

src/main/java/org/opensearch/knn/index/codec/util/KNNCodecUtil.java
luyuncheng: Added at https://github.com/opensearch-project/k-NN/pull/1946/commits/439c5b7e01f4bf1d75c8f9e369c1674be21f2d50#

jmazanec15 (approved):
LGTM

navneet1v (approved):

luyuncheng:
@jmazanec15 @navneet1v Thanks guys for reviewing

0ctopus13prime:
Hi @luyuncheng 
Could you share why we don't include a field having 'model_id' in its field attribute into `indexPathMap`??
Wondering having 'model_id' in its field attribute mean something else??

Link : https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80DocValuesProducer.java#L140

```
// Only Native Engine put into indexPathMap
KNNEngine knnEngine = getNativeKNNEngine(field);
if (knnEngine == null) {
    continue;
}
```

```
private KNNEngine getNativeKNNEngine(@NonNull FieldInfo field) {
    final String modelId = field.attributes().get(MODEL_ID);
    if (modelId != null) {
        return null; <------------- This
    }
    ...
```

luyuncheng:
> Could you share why we don't include a field having 'model_id' in its field attribute into `indexPathMap`??
> Wondering having 'model_id' in its field attribute mean something else??

because in model engine, do not need docValuesProducer for merge. so i filter it out

luyuncheng:
@0ctopus13prime 

> What is model engine??

i think only faiss can have a  model engine?

> What do we do with a model engine??
> How we differentiate model engine from the plain vector engine (ex: Hnsw)??

sorry i do not understand this question.

> Why a model engine does not need a resource clean up in cache manager when DVReader is being closed??

i see there are 3 memory allocation `IndexAllocation`, `TrainingDataAllocation`, `AnonymousAllocation`. cache manager do not mange all of these?

0ctopus13prime:
@luyuncheng 
Thank you, but it still does not answer the questions. :) 
Overall, I'm trying to understand what 'model engine' is.  And how it's different from a plain graph based vector engine like HNSW.

**Could you give more contexts why did you add the logic so that those 'model engines' are excluded from invalidating the cache??**

```
    private KNNEngine getNativeKNNEngine(@NonNull FieldInfo field) {
        final String modelId = field.attributes().get(MODEL_ID);
        if (modelId != null) { <------------ This one. Why did you add it?
            return null;
        }
        KNNEngine engine = FieldInfoExtractor.extractKNNEngine(field);
        if (KNNEngine.getEnginesThatCreateCustomSegmentFiles().contains(engine)) {
            return engine;
        }
        return null;
    }
```

0ctopus13prime:
Hi @luyuncheng,
could you answer ^^ questions when you have time?

What I understand is modelId coming from cluster state also I confirmed OpneSearch build its index internally.
I recently removed the code so that allocated memory can be released after we deprecated FileWatcher.
I think this is safe to do that, but hoping to get your motivation for the if-branch just in case.

Thank you

PR #1950, opened 2024-08-12
Integrates FAISS iterative builds with NativeEngines990KnnVectorsFormat

shatejas:
### Description
- The commit merges [feature/iterative-index-build](https://github.com/opensearch-project/k-NN/tree/feature/iterative-index-build) branch
- It integrates with KNNVectorValues and NativeEngines990KnnVectorsFormat
- There are changes to reuse offheap vector buffer during iterative vector transfer

### Related Issues
Resolves #[1853](https://github.com/opensearch-project/k-NN/issues/1853)

### Check List


navneet1v:
@shatejas lets fix the CIs first

jni/src/faiss_index_service.cpp
```diff
@@ -57,76 +57,179 @@ void SetExtraParameters(knn_jni::JNIUtilInterface * jniUtil, JNIEnv *env,
...
+
+    /*
+     * NOTE: The process of memory allocation is currently only implemented for HNSW.
+     * This technique of checking the types of the index and subindices should be generalized into
+     * another function.
+     */
+
+    // Check to see if the current index is HNSW
+    faiss::IndexHNSWFlat * hnsw = dynamic_cast<faiss::IndexHNSWFlat *>(idMap->index);
+    if(hnsw != NULL) {
```
navneet1v, line 97: I think we started to use nullptr instead of NULL.

jni/src/faiss_index_service.cpp
```diff
@@ -57,76 +57,179 @@ void SetExtraParameters(knn_jni::JNIUtilInterface * jniUtil, JNIEnv *env,
...
+    faiss::IndexHNSWSQ * hnswSq = dynamic_cast<faiss::IndexHNSWSQ *>(idMap->index);
+    if(hnswSq != NULL) {
+        // Check to see if the HNSW storage is IndexFlat
+        faiss::IndexFlat * storage = dynamic_cast<faiss::IndexFlat *>(hnswSq->storage);
+        if(storage != NULL) {
+            // Allocate enough memory for all of the vectors we plan on inserting
+            // We do this to avoid unnecessary memory allocations during insert
+            storage->codes.reserve(dim * numVectors * 2);
+        }
+    }
```
navneet1v, line 115: Andrew updated the code to fix the SQ. Can you refer this commit: https://github.com/opensearch-project/k-NN/commit/e2204c29248685b1e053efed50e00208a9afed28 I think we have not included this change in this PR.

navneet1v (changes requested):
Added initial comments, will be reviewing the code further in next few hours.

src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexWriter.java
```diff
@@ -0,0 +1,287 @@
...
+        return indexBuilt;
+    }
+
+    /**
+     * flushes the index
+     * @param knnVectorValues
+     * @return boolean to indicate if the index was written
+     * @throws IOException
+     */
+    public boolean flushIndex(final KNNVectorValues<?> knnVectorValues) throws IOException {
```
Vikasht34, line 112: FlushIndex and MergeIndex is doing most of the things same except the metrics emission , Can we Create common function like ProcessIndex and call this ProcessIndex from both function with extra parameter with functional Interface that takes which graph emission method to run

src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexWriter.java
```diff
@@ -0,0 +1,287 @@
...
+            .build();
+    }
+
+    private Map<String, Object> getParameters(FieldInfo fieldInfo, VectorDataType vectorDataType, KNNEngine knnEngine) throws IOException {
+        Map<String, Object> parameters = new HashMap<>();
+        Map<String, String> fieldAttributes = fieldInfo.attributes();
+        String parametersString = fieldAttributes.get(KNNConstants.PARAMETERS);
+
+        // parametersString will be null when legacy mapper is used
+        if (parametersString == null) {
```
Vikasht34, line 173: Can we simply this to make more understandable 

if (Parameter == null) then buildLegacyParms() else put that para string in paramter 
next  updateFaissBinaryParameters
next update indexing thread

Vikasht34 (commented):
I have covered 4 files , will continue with remaining file tomoroww.

jni/src/org_opensearch_knn_jni_FaissService.cpp
```diff
@@ -39,37 +39,84 @@ void JNI_OnUnload(JavaVM *vm, void *reserved) {
...
-        // Releasing the vectorsAddressJ memory as that is not required once we have created the index.
-        // This is not the ideal approach, please refer this gh issue for long term solution:
-        // https://github.com/opensearch-project/k-NN/issues/1600
-        delete reinterpret_cast<std::vector<float>*>(vectorsAddressJ);
+JNIEXPORT jlong JNICALL Java_org_opensearch_knn_jni_FaissService_initBinaryIndex(JNIEnv * env, jclass cls,
+                                                                            jlong numDocs, jint dimJ,
+                                                                            jobject parametersJ)
+{
+    try {
+        std::unique_ptr<knn_jni::faiss_wrapper::FaissMethods> faissMethods(new knn_jni::faiss_wrapper::FaissMethods());
```
jmazanec15, line 61: Why are we creating a new one of these per index? Couldnt we reuse?

jni/src/faiss_index_service.cpp
```diff
@@ -137,28 +218,35 @@ void BinaryIndexService::createIndex(
...
-    std::unique_ptr<faiss::IndexBinaryIDMap> idMap(faissMethods->indexBinaryIdMap(indexWriter.get()));
     idMap->add_with_ids(numVectors, inputVectors->data(), ids.data());
+}
+
+void BinaryIndexService::writeIndex(
+        std::string indexPath,
+        jlong idMapAddress
+    ) {
+
+    std::unique_ptr<faiss::IndexBinaryIDMap> idMap (reinterpret_cast<faiss::IndexBinaryIDMap *> (idMapAddress));
```
jmazanec15, line 237: Are we sure we want to transfer ownership of the index here? Seems this could be misleading

jni/src/faiss_index_service.cpp
```diff
@@ -137,28 +218,35 @@ void BinaryIndexService::createIndex(
...
+        faissMethods->writeIndexBinary(idMap.get(), indexPath.c_str());
+    } catch(std::exception &e) {
+        delete idMap->index;
+        throw std::runtime_error("Failed to write index to disk");
+    }
 
-    // Write the index to disk
-    faissMethods->writeIndexBinary(idMap.get(), indexPath.c_str());
+    // Free the memory used by the index
+    delete idMap->index;
```
jmazanec15, line 248: Can we set owns field to true in https://github.com/facebookresearch/faiss/blob/main/faiss/IndexIDMap.h so we dont have to do this?

jni/src/faiss_index_service.cpp
```diff
@@ -57,76 +57,157 @@ void SetExtraParameters(knn_jni::JNIUtilInterface * jniUtil, JNIEnv *env,
...
+
+    // Check that the index does not need to be trained
+    if(!indexWriter->is_trained) {
+        throw std::runtime_error("Index is not trained");
+    }
+
+    std::unique_ptr<faiss::IndexBinaryIDMap> idMap(faissMethods->indexBinaryIdMap(indexWriter.get()));
+
+    allocIndex(dynamic_cast<faiss::Index *>(idMap->index), dim, numVectors);
+    indexWriter.release();
```
jmazanec15, line 196: do we need this?

jni/src/faiss_index_service.cpp
```diff
@@ -57,76 +57,157 @@ void SetExtraParameters(knn_jni::JNIUtilInterface * jniUtil, JNIEnv *env,
...
+
+    try {
+        // Write the index to disk
+        faissMethods->writeIndex(idMap.get(), indexPath.c_str());
+    } catch(std::exception &e) {
+        delete idMap->index;
+        throw std::runtime_error("Failed to write index to disk");
+    }
+    // Free the memory used by the index
+    delete idMap->index;
```
jmazanec15, line 154: Same as below

jni/src/faiss_index_service.cpp
```diff
@@ -57,76 +57,157 @@ void SetExtraParameters(knn_jni::JNIUtilInterface * jniUtil, JNIEnv *env,
...
 
-void IndexService::createIndex(
+void IndexService::allocIndex(faiss::Index * index, size_t dim, size_t numVectors) {
+    if(auto * indexHNSWSQ = dynamic_cast<faiss::IndexHNSWSQ *>(index)) {
+        if(auto * indexScalarQuantizer = dynamic_cast<faiss::IndexScalarQuantizer *>(indexHNSWSQ->storage)) {
+            indexScalarQuantizer->codes.reserve(indexScalarQuantizer->code_size * numVectors);
+        }
+        return;
+    }
+    if(auto * indexHNSW = dynamic_cast<faiss::IndexHNSW *>(index)) {
```
jmazanec15, line 67: Are we not supporting IVF?

src/main/java/org/opensearch/knn/common/KNNVectorUtil.java
```diff
@@ -42,4 +44,18 @@ public static boolean isZeroVector(float[] vector) {
...
+    /**
+     * Creates an int overflow safe arraylist. If there is an overflow it will create a list with default initial size
+     * @param batchSize size to allocate
+     * @return an arrayList
+     */
+    public static <T> ArrayList<T> createArrayList(long batchSize) {
+        try {
+            return new ArrayList<>(Math.toIntExact(batchSize));
+        } catch (Exception exception) {
+            // No-op
```
jmazanec15, line 57: This seems risky. Shouldnt we just throw?

src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80DocValuesConsumer.java
```diff
@@ -95,174 +59,26 @@ public void addBinaryField(FieldInfo field, DocValuesProducer valuesProducer) th
...
-        // Skip index creation if no vectors or docs in segment
-        if (pair.getVectorAddress() == 0 || pair.docs.length == 0) {
-            logger.info("Skipping engine index creation as there are no vectors or docs in the segment");
-            return;
-        }
-
-        long arraySize = calculateArraySize(pair.docs.length, pair.getDimension(), vectorDataType);
+    public void addKNNBinaryField(FieldInfo field, DocValuesProducer valuesProducer, boolean isMerge) throws IOException {
+        final VectorDataType vectorDataType = extractVectorDataType(field);
+        final KNNVectorValues<?> knnVectorValues = KNNVectorValuesFactory.getVectorValues(vectorDataType, valuesProducer.getBinary(field));
```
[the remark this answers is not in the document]
shatejas: How is this supposed to know the type without the wildcard use? What am I missing?

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriter.java
```diff
@@ -62,14 +70,48 @@ public KnnFieldVectorsWriter<?> addField(final FieldInfo fieldInfo) throws IOExc
...
+        for (final NativeEngineFieldVectorsWriter<?> field : fields) {
+            final VectorDataType vectorDataType = extractVectorDataType(field.getFieldInfo());
+            final KNNVectorValues<?> knnVectorValues = KNNVectorValuesFactory.getVectorValues(
+                vectorDataType,
+                field.getDocsWithField(),
+                field.getVectors()
+            );
+
+            // TODO: Extract quantization state here
+            boolean indexWritten = NativeIndexWriter.getWriter(field.getFieldInfo(), segmentWriteState).refreshIndex(knnVectorValues);
```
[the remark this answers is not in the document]
shatejas: Its changed to flushIndex and mergeIndex now. Its still writing the index and the name is mapped according to where it should be called. Suggestions on the names are welcome and I can reconsider

src/main/java/org/opensearch/knn/index/codec/nativeindex/BulkVectorTransferIndexBuildStrategy.java
```diff
@@ -0,0 +1,91 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.knn.index.codec.nativeindex;
```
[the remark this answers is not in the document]
shatejas: > but it would be completed if we get this Strategies via Factory rather than new BulkVectorTransferIndexBuildStrategy()

The intention is to keep it singleton. A factory isn't required here since there are two cases and a simpler if else is managing it. If more cases are ever added, sure it can manage it better. For now I would prefer keeping it simple

src/main/java/org/opensearch/knn/index/codec/nativeindex/BulkVectorTransferIndexBuildStrategy.java
```diff
@@ -0,0 +1,91 @@
...
+
+    public static BulkVectorTransferIndexBuildStrategy getInstance() {
+        return INSTANCE;
+    }
+
+    private BulkVectorTransferIndexBuildStrategy() {}
+
+    public boolean buildIndex(final BuildIndexParams indexInfo, final KNNVectorValues<?> knnVectorValues) {
+
+        try (final VectorTransfer vectorTransfer = getVectorTransfer(indexInfo.getVectorDataType(), knnVectorValues)) {
```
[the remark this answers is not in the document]
shatejas: I anticipate logic to getVectorTransfer a bit more involved when quantization is introduced, I chose to punt it for the integration so we have a better idea on managing it

src/main/java/org/opensearch/knn/index/codec/nativeindex/BulkVectorTransferIndexBuildStrategy.java
```diff
@@ -0,0 +1,91 @@
...
+import java.security.AccessController;
+import java.security.PrivilegedAction;
+import java.util.Map;
+
+import static org.opensearch.knn.common.KNNConstants.MODEL_ID;
+
+/**
+ * Transfers all vectors to offheap and then builds an index
+ */
+final class BulkVectorTransferIndexBuildStrategy implements NativeIndexBuildStrategy {
```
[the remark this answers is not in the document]
shatejas: > Functionality for both strategy is not same, Bulk has two things , tranfer and create , while other one has 3 , trasfer create and write

The responsibility is to build and write index (graphs) for the segment. How many parts the strategy breaks it into is its own responsibility. Both are building and writing the index irrespective of the names and the number of steps

> Either we break into transfer , create and write in separate functions at Inteface level , or we should think something differnt.

There is little room for separation of responsibilities here considering its all intertwined. Each of the responsibilities has been delegated, this is aggregating things together. The transfer if delegated to vector transfer, its only managing the lifecycle of vector transfer object. The index creation and writing is delegated to JNI layer, this is only managing the sequence

src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexWriter.java
shatejas: The logic for this is kept as it is for now. This surely needs a refactor but considering the amount of changes going in this PR I just added a comment above  `indexParams` I can follow up after the PR in terms of cleaning this part up

jni/src/faiss_index_service.cpp
shatejas: Not as of now, it will still use bulk vector transfer followed by createIndexFromTemplate

src/main/java/org/opensearch/knn/common/KNNVectorUtil.java
shatejas: The [intent](https://github.com/opensearch-project/k-NN/pull/1950#discussion_r1715603078). This is being defensive while being consistent with current approach of unchecked arraylist expansion. The risk is unchanged

src/main/java/org/opensearch/knn/index/codec/transfer/VectorTransfer.java
```diff
@@ -5,50 +5,43 @@
...
 
     /**
-     * Initialize the transfer
-     *
-     * @param totalLiveDocs total number of vectors to be transferred
+     * Transfer a single vector from one location to another
+     * @throws IOException
      */
-    abstract public void init(final long totalLiveDocs);
+    void transfer() throws IOException;
```
navneet1v, line 21: Not seeing where this function is getting used. I am missing something here.

src/main/java/org/opensearch/knn/index/codec/transfer/OffHeapFloatVectorTransfer.java
```diff
@@ -0,0 +1,50 @@
...
+import org.opensearch.knn.jni.JNICommons;
+
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.List;
+
+/**
+ * Transfer float vectors to off heap memory.
+ */
+public final class OffHeapFloatVectorTransfer extends OffHeapQuantizedVectorTransfer<float[], float[]> {
```
navneet1v, line 19: If we have to convert and transfer a float[] to byte[] do we create a new class for that kind of transfer? This is the reason I think 1 generic is good, but other generic I am not sure how useful it is.

src/main/java/org/opensearch/knn/index/codec/transfer/OffHeapByteVectorTransfer.java
```diff
@@ -0,0 +1,46 @@
...
+public final class OffHeapByteVectorTransfer extends OffHeapQuantizedVectorTransfer<byte[], byte[]> {
+
+    // TODO: Replace with KNNByteVectorValues
+    public OffHeapByteVectorTransfer(KNNVectorValues<byte[]> vectorValues, final Long batchSize) throws IOException {
+        super(vectorValues, batchSize, (vector, state) -> vector, StringUtils.EMPTY, DEFAULT_COMPRESSION_FACTOR);
+    }
+
+    // TODO: Replace with KNNByteVectorValues
+    public OffHeapByteVectorTransfer(KNNVectorValues<byte[]> vectorValues) throws IOException {
+        this(vectorValues, null);
```
navneet1v, line 28: what will be the case when we need to pass the batchSize, just wondering?

navneet1v:
Some of the comments can be old as there were updates in the PR. Please ignore comments if they are already answered.

src/main/java/org/opensearch/knn/index/codec/transfer/OffHeapQuantizedVectorTransfer.java
```diff
@@ -0,0 +1,149 @@
...
+        this.batchSize = batchSize == null ? transferLimit : batchSize;
+        this.vectorsToTransfer = createArrayList(this.batchSize);
+        this.transferredDocIdsList = createArrayList(this.batchSize);
+        this.vectorValues = vectorValues;
+        this.vectorAddress = 0; // we can allocate initial memory here, currently storeVectorData takes care of it
+    }
+
+    @Override
+    public void transfer() throws IOException {
+        if (vectorValues.docId() == NO_MORE_DOCS) {
```
navneet1v, line 74: why we are doing this check? Plus this transfer function is only called from Transferbatch do we need to make this as public? Seeing public I am trying to see who is calling this function

src/main/java/org/opensearch/knn/index/codec/transfer/OffHeapFloatVectorTransfer.java
```diff
@@ -0,0 +1,50 @@
...
+    protected float[] copyVector(float[] vector) {
+        return Arrays.copyOf(vector, vector.length);
+    }
+
+    @Override
+    protected long transfer(List<float[]> vectorsToTransfer, long bytesPerVector, boolean append) throws IOException {
+        return JNICommons.storeVectorData(
+            getVectorAddress(),
+            vectorsToTransfer.toArray(new float[][] {}),
+            this.batchSize * bytesPerVector,
```
navneet1v, line 39: I want to know how bytesPerVector * batchSize gives the initialCapacity for the memory location created in C++ layer? can you clarify few things here.

src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80DocValuesConsumer.java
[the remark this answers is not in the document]
navneet1v: I also want to know how we can avoid this. because the underline VectorValues class uses generic/ or we write a wrapper over the primitive arrays. But then in that case VectorTransfer classes has to figure out what is that generics.

src/main/java/org/opensearch/knn/index/codec/transfer/OffHeapQuantizedVectorTransfer.java
```diff
@@ -0,0 +1,149 @@
...
+            // Throwing instead of returning so there is no way client can go into an infinite loop
+            throw new IllegalStateException("No more vectors available to transfer");
+        }
+
+        transferredDocsIds = new int[1];
+        V vector = quantizer.apply(vectorValues.getVector(), quantizationState);
+        vectorsToTransfer.add(vector);
```
navneet1v, line 81: we need copy of vectors here too.

src/main/java/org/opensearch/knn/index/codec/nativeindex/MemOptimizedNativeIndexBuildStrategy.java
```diff
@@ -0,0 +1,94 @@
...
+    private static MemOptimizedNativeIndexBuildStrategy INSTANCE = new MemOptimizedNativeIndexBuildStrategy();
+
+    public static MemOptimizedNativeIndexBuildStrategy getInstance() {
+        return INSTANCE;
+    }
+
+    private MemOptimizedNativeIndexBuildStrategy() {}
+
+    public boolean buildIndex(BuildIndexParams indexInfo, final KNNVectorValues<?> knnVectorValues) throws IOException {
+        if (knnVectorValues.docId() == -1) {
```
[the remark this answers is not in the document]
shatejas: Its being defensive if this class is used standalone

src/main/java/org/opensearch/knn/index/codec/nativeindex/MemOptimizedNativeIndexBuildStrategy.java
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    // TODO: Will probably need a factory once quantization is added
+    private VectorTransfer getVectorTransfer(VectorDataType vectorDataType, KNNVectorValues<?> knnVectorValues) throws IOException {
+        switch (vectorDataType) {
+            case FLOAT:
+                return new OffHeapFloatVectorTransfer((KNNFloatVectorValues) knnVectorValues);
+            case BINARY:
+            case BYTE:
+                return new OffHeapByteVectorTransfer((KNNVectorValues<byte[]>) knnVectorValues);
```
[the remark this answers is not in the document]
shatejas: Currently Binary and Byte have different classes but same implementation. Which will change with https://github.com/opensearch-project/k-NN/pull/1823

src/main/java/org/opensearch/knn/index/codec/transfer/VectorTransfer.java
```diff
@@ -5,50 +5,43 @@
...
-     * Transfer a single vector
-     *
-     * @param bytesRef a vector in bytes format
+     * Transfer a batch of vectors from one location to another
+     * The batch size here is intended to be constant for multiple transfers so should be encapsulated in the
+     * implementation. A new batch size should require another instance
+     * @throws IOException
      */
-    abstract public void transfer(final BytesRef bytesRef);
+    void transferBatch() throws IOException;
```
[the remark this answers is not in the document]
shatejas: transfer will transfer one vector at a time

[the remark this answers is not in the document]
navneet1v: why we need to transfer 1 vector at a time?

src/main/java/org/opensearch/knn/index/codec/nativeindex/MemOptimizedNativeIndexBuildStrategy.java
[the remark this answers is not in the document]
navneet1v: its okay to be defensive, but we will loose this context in sometime. So anytime we are adding a defensive check its better to add the java docs to tell that this is defensive check and where it is already handled. Otherwise what will happen is we will just forget why we did this.

Another thing is if we can make this function only accessible internally we can avoid these checks.

src/main/java/org/opensearch/knn/index/codec/transfer/OffHeapQuantizedVectorTransfer.java
```diff
@@ -0,0 +1,149 @@
...
+import static org.opensearch.knn.common.KNNVectorUtil.createArrayList;
+
+/**
+ * The class is intended to transfer {@link KNNVectorValues} to off heap memory. If also provides and ability to quantize the vector
+ * before it is transferred to offHeap memory. The class is not thread safe
+ *
+ * @param <T> an array of primitive type
+ * @param <V> an array of primitive type after being quantized
+ */
+abstract class OffHeapQuantizedVectorTransfer<T, V> implements VectorTransfer {
```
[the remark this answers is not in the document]
shatejas: > and then expose a default implementation of retireveVector/getVector in this class which can be default to just getting vector from VectorValues and if quantization Object is present in any implementation of VectorTransfer then they can override this getVector function to apply quantization.

I think this should be possible,  I can have an interface preprocessTransfer which should be able to have quantization implementation in there

src/main/java/org/opensearch/knn/index/codec/transfer/OffHeapFloatVectorTransfer.java
shatejas: For now float[] to float[] does not have a use case so kept it as it is

src/main/java/org/opensearch/knn/index/codec/transfer/OffHeapByteVectorTransfer.java
shatejas: Yes. [VectorTransferIndexBuildStrategy](https://github.com/opensearch-project/k-NN/pull/1950/files#diff-6047a6c31b628a0edc4bd7eba164bdfc35fcc53de703a25a295e91aa04b94d20R85)

src/main/java/org/opensearch/knn/index/codec/transfer/OffHeapQuantizedVectorTransfer.java
shatejas: > why we are doing this check?

Its being defensive so client does not go into infinite loop

>  Plus this transfer function is only called from Transferbatch do we need to make this as public? Seeing public I am trying to see who is calling this function

https://github.com/opensearch-project/k-NN/pull/1950#discussion_r1715970582

src/main/java/org/opensearch/knn/index/codec/transfer/OffHeapQuantizedVectorTransfer.java
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    @Override
+    public void transfer() throws IOException {
+        if (vectorValues.docId() == NO_MORE_DOCS) {
+            // Throwing instead of returning so there is no way client can go into an infinite loop
+            throw new IllegalStateException("No more vectors available to transfer");
+        }
+
+        transferredDocsIds = new int[1];
```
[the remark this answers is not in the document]
shatejas: Its intended to only transfer one vector at a time

src/main/java/org/opensearch/knn/index/codec/transfer/VectorTransfer.java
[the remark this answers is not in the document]
navneet1v: > Its intended to transfer one vector at a time, this will be useful when we are doing iterative build on addValue

It should be handled when we are building that functionality. Because at that time we can focus on that functionality in more details around what is needed. Like what if we want to reuse batch=1 and transferLimit=1 to send 1 vector at time, rather than this. I may not be thinking optimal here, but I do think inserting 1 vector at a time will have more things to look at hence its better to not add this functionality. 

Since this is my opinion. I am happy to know more on this.

src/main/java/org/opensearch/knn/index/codec/transfer/OffHeapFloatVectorTransfer.java
navneet1v: Sure will check.

But in my thought process on this `OffHeapFloatVectorTransfer` should just say I am going to transfer a batch of float vectors to a memory address. How those floats are getting created it should not know about it.

Please let me know what you think here. Since you are more close to the code than I am.

src/main/java/org/opensearch/knn/index/codec/transfer/OffHeapByteVectorTransfer.java
navneet1v: thanks found it later. Its hard to grasp from code.

src/main/java/org/opensearch/knn/index/codec/transfer/OffHeapFloatVectorTransfer.java
navneet1v: may be the variable name is not correct. In c++, it is representing what is the length of the 1D array we should have which is calculated as `dimensions per vector * batchSize` `bytesPerVector` tells me that if I have a float[] of 5 dimension, then this value should be 5 * 4 = 20.

Please check what it is. I think variable name is not correct.

src/main/java/org/opensearch/knn/index/codec/transfer/OffHeapQuantizedVectorTransfer.java
navneet1v: Lets mention this in the java doc. Just to know that why these conditions are there.

jni/src/faiss_index_service.cpp
navneet1v: +1, if it works then yes we should do this. But lets ensure all the internal indices are getting freed, otherwise it will be a memory leak.

jni/src/faiss_index_service.cpp
navneet1v: are you asking for why we need `allocIndex(dynamic_cast<faiss::Index *>(idMap->index), dim, numVectors);`?

src/main/java/org/opensearch/knn/index/codec/nativeindex/MemOptimizedNativeIndexBuildStrategy.java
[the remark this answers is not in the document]
shatejas: will add javadocs, I encapsulated this in KNNVectorValues. let me know what you think about it

src/main/java/org/opensearch/knn/index/codec/transfer/VectorTransfer.java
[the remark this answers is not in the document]
shatejas: I can remove it for now

src/main/java/org/opensearch/knn/index/codec/transfer/OffHeapQuantizedVectorTransfer.java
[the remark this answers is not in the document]
shatejas: I tried this and the issue comes in during transferLimit calculations where it requires bytesPerVector. If we can get around it somehow thins can work

src/main/java/org/opensearch/knn/index/codec/transfer/OffHeapQuantizedVectorTransfer.java
shatejas: There is a comment there, anywhere else you want me to put it?

src/main/java/org/opensearch/knn/index/codec/transfer/OffHeapFloatVectorTransfer.java
shatejas: > In c++, it is representing what is the length of the 1D array we should have which is calculated as dimensions per vector * batchSize

Good catch, This is reserving way more capacity than required

jni/src/org_opensearch_knn_jni_FaissService.cpp
shatejas: To reuse we will somehow need to use dependency injection, which doesn't seem to be possible as this is not inside a class. This was created to enable unit testing and making it static won't make sense. Any other suggestions?

jni/src/faiss_index_service.cpp
shatejas: @jmazanec15 My understanding is `indexWriter.release();` is specifically done to not delete the the index created. The unique ptr will go out of scope once this method finishes, this is initializing the index, we need the index in future to make sure we can insert and write

shatejas: Added a comment to make it more evident

jni/src/faiss_index_service.cpp
shatejas: @navneet1v Seems like the [destructor deletes it](https://github.com/facebookresearch/faiss/blob/33c0ba5d002a7cd9761513f06ecc9822079d4a2f/faiss/IndexIDMap.cpp#L211)

navneet1v: @shatejas so what you are saying is we are able to set the own_fields to true? and if it is set to true I can see the index is getting deleted.

jni/src/faiss_index_service.cpp
shatejas: The concern is more about if a code path is accessing it after write is complete. There is no mechanism currently to enforce it and this makes memory management simpler so keeping it as it is

jni/src/faiss_index_service.cpp
```diff
@@ -57,76 +57,163 @@ void SetExtraParameters(knn_jni::JNIUtilInterface * jniUtil, JNIEnv *env,
...
+
+    std::unique_ptr<faiss::IndexIDMap> idMap (faissMethods->indexIdMap(index.get()));
+    //Makes sure the index is deleted when the destructor is called
+    idMap->own_fields = true;
+
+    allocIndex(dynamic_cast<faiss::Index *>(idMap->index), dim, numVectors);
+
+    //Release the ownership so as to make sure not delete the underlying index that is created. The index is needed later
+    //in insert and write operations
+    index.release();
```
navneet1v, line 109: if we have not created the index via unique pointer then it won't be required. See if we can just avoid using Unique pointer so we don't have to do this release here

jni/src/faiss_index_service.cpp
shatejas: We are able to set it to true. The link is the traced code path and seems like its getting deleted. 

Additional verifications will need an component test which is not currently present, current unit test will not be able to test the delete as its mocked currently. I can revert back to a manual delete if component test is a blocker

navneet1v: But if that is the case, if I see this code which builds the index it is already setting the own_fields to true. ref: https://github.com/facebookresearch/faiss/blob/33c0ba5d002a7cd9761513f06ecc9822079d4a2f/faiss/index_factory.cpp#L746-L760 , so we setting that to true should not impact.

and in that case the internal index should have been deleted. But this check was explicitly added because the memory was not getting free when tested in memory constraint env. I would recommend keeping this check as it ensures memory is getting free, and do a separate benchmarks to provide setting own fields is working or not.


If there is any code piece that is reverting this value to value please share. I didn't find any thing. Because I do remember me not freeing the memory caused the leak in my tests and same happened with Andrew too. 


Please let me know I am missing anything. This is what based on my experience working on this code I have seen

navneet1v: I think we can keep it. So what we are doing if I see the code correctly, we are using a factory to build the HNSW index, which sets own_field=true. But after that we are putting that HNSW index in the IDMap index there we are not setting own_fields=true. With your change I think it will work.

src/main/java/org/opensearch/knn/index/codec/transfer/OffHeapQuantizedVectorTransfer.java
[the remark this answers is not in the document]
navneet1v: @shatejas can we calculate the bytesPerVector on the quantized vector using a function that can take a look at the output type of quantization and give us that value? 

will that solve it assuming we have a way to find the bytes per vector of the quantized vector? thoughts?

jni/src/faiss_index_service.cpp
shatejas: Considering the push towards using smart pointers, I think this should work

src/main/java/org/opensearch/knn/index/codec/transfer/OffHeapFloatVectorTransfer.java
```diff
@@ -0,0 +1,37 @@
...
+
+    public OffHeapFloatVectorTransfer(int transferLimit) {
+        super(transferLimit);
+    }
+
+    @Override
+    protected long transfer(final List<float[]> vectorsToTransfer, boolean append) throws IOException {
+        return JNICommons.storeVectorData(
+            getVectorAddress(),
+            vectorsToTransfer.toArray(new float[][] {}),
```
jmazanec15, line 26: [Not blocking - can take in another review] Instead of this (which I think will copy), should we transfer one vector at a time?

src/main/java/org/opensearch/knn/index/codec/transfer/OffHeapVectorTransfer.java
```diff
@@ -0,0 +1,63 @@
...
+            return true;
+        }
+        return false;
+    }
+
+    public boolean flush(boolean append) throws IOException {
+        // flush before closing
+        if (!vectorsToTransfer.isEmpty()) {
+            vectorAddress = transfer(vectorsToTransfer, append);
+            return true;
```
jmazanec15, line 57: We should call `vectorsToTransfer.clear();` to just in case there are 100s of MBs of vectors queued up to be flushed

src/main/java/org/opensearch/knn/index/codec/transfer/OffHeapFloatVectorTransfer.java
shatejas: This is a shallow copy, so actual vectors won't be copied. The memory gains might not be much

Vikasht34 (approved):
LGTM. , if there is Major Improvement needed , let's have in Follow up PR.

jmazanec15 (dismissed):
lgtm!

src/main/java/org/opensearch/knn/index/codec/nativeindex/MemOptimizedNativeIndexBuildStrategy.java
```diff
@@ -0,0 +1,110 @@
...
+                knnVectorValues.dimension(),
+                indexParameters,
+                engine
+            )
+        );
+
+        int transferLimit = (int) Math.max(1, KNNSettings.getVectorStreamingMemoryLimit().getBytes() / knnVectorValues.bytesPerVector());
+        try (final OffHeapVectorTransfer vectorTransfer = getVectorTransfer(indexInfo.getVectorDataType(), transferLimit)) {
+
+            final List<Integer> tranferredDocIds = new ArrayList<>();
```
navneet1v, line 58: when you know the transfer limit then why cannot we directly make int[] rather than a List<Integer>

src/main/java/org/opensearch/knn/index/codec/transfer/OffHeapVectorTransfer.java
```diff
@@ -0,0 +1,73 @@
...
+
+    private final List<T> vectorsToTransfer;
+
+    public OffHeapVectorTransfer(final int transferLimit) {
+        this.transferLimit = transferLimit;
+        this.vectorsToTransfer = new ArrayList<>(transferLimit);
+        this.vectorAddress = 0;
+    }
+
+    public boolean transfer(T vector, boolean append) throws IOException {
```
navneet1v, line 43: Please add java docs on all the public functions atleast on the base class functions so that it is easy to know what these functions are supposed to do.

src/main/java/org/opensearch/knn/index/codec/transfer/OffHeapFloatVectorTransfer.java
```diff
@@ -0,0 +1,36 @@
...
+            (long) vectorsToTransfer.get(0).length * this.transferLimit,
+            append
+        );
+    }
+
+    @Override
+    public void deallocate() {
```
navneet1v, line 33: why public?

src/main/java/org/opensearch/knn/jni/JNIService.java
```diff
@@ -13,28 +13,110 @@
...
+     */
+    public static void insertToIndex(
+        int[] docs,
+        long vectorsAddress,
+        int dimension,
+        Map<String, Object> parameters,
+        long indexAddress,
+        KNNEngine knnEngine
+    ) {
+        int threadCount = (int) parameters.getOrDefault(KNNConstants.INDEX_THREAD_QTY, 0);
```
navneet1v, line 70: why default is 0 here?

jni/src/nmslib_wrapper.cpp
```diff
@@ -157,12 +157,6 @@ void knn_jni::nmslib_wrapper::CreateIndex(knn_jni::JNIUtilInterface * jniUtil, J
         }
         jniUtil->ReleaseIntArrayElements(env, idsJ, idsCpp, JNI_ABORT);
 
-        // Releasing the vectorsAddressJ memory as that is not required once we have created the index.
-        // This is not the ideal approach, please refer this gh issue for long term solution:
-        // https://github.com/opensearch-project/k-NN/issues/1600
-        //commons::freeVectorData(vectorsAddressJ);
-        delete inputVectors;
```
navneet1v, line 164: if you don't release the memory here then during the createIndex and save index the memory will shoot up. Hence this was added specially here to avoid the memory shoots ups. Lets not remove this.

jni/src/faiss_index_service.cpp
```diff
@@ -57,76 +57,163 @@ void SetExtraParameters(knn_jni::JNIUtilInterface * jniUtil, JNIEnv *env,
...
+    SetExtraParameters<faiss::Index, faiss::IndexIVF, faiss::IndexHNSW>(jniUtil, env, parameters, index.get());
+
+    // Check that the index does not need to be trained
+    if(!index->is_trained) {
+        throw std::runtime_error("Index is not trained");
+    }
+
+    std::unique_ptr<faiss::IndexIDMap> idMap (faissMethods->indexIdMap(index.get()));
+    //Makes sure the index is deleted when the destructor is called
+    idMap->own_fields = true;
```
navneet1v, line 103: can we pass value of own_fields in the constructor?

jni/src/faiss_index_service.cpp
navneet1v: Smart pointers are used to ensure memory gets freed up. Here the use case is not here. So I don't agree we should use smart pointers here doesn't matter whatever the push is because its not the right usage of smart pointers.

jni/src/faiss_index_service.cpp
```diff
@@ -57,76 +57,163 @@ void SetExtraParameters(knn_jni::JNIUtilInterface * jniUtil, JNIEnv *env,
...
-        std::vector<int64_t> ids,
-        std::string indexPath,
-        std::unordered_map<std::string, jobject> parameters
+        std::vector<int64_t> & ids,
+        jlong idMapAddress
     ) {
-    // Read vectors from memory address
-    auto *inputVectors = reinterpret_cast<std::vector<float>*>(vectorsAddress);
+    // Read vectors from memory address (unique ptr since we want to remove from memory after use)
+    std::vector<float> * inputVectors = reinterpret_cast<std::vector<float>*>(vectorsAddress);
```
navneet1v, line 122: this input vectors memory gets freed by the Java classes as per this code. So, what the use of smart pointers here?

jni/src/faiss_index_service.cpp
```diff
@@ -57,76 +57,163 @@ void SetExtraParameters(knn_jni::JNIUtilInterface * jniUtil, JNIEnv *env,
...
 
     // The number of vectors can be int here because a lucene segment number of total docs never crosses INT_MAX value
     int numVectors = (int) (inputVectors->size() / (uint64_t) dim);
     if(numVectors == 0) {
-        throw std::runtime_error("Number of vectors cannot be 0");
+        return;
```
navneet1v, line 127: why this change?

jni/src/faiss_wrapper.cpp
```diff
@@ -217,10 +241,6 @@ void knn_jni::faiss_wrapper::CreateIndexFromTemplate(knn_jni::JNIUtilInterface *
     auto idVector = jniUtil->ConvertJavaIntArrayToCppIntVector(env, idsJ);
     faiss::IndexIDMap idMap =  faiss::IndexIDMap(indexWriter.get());
     idMap.add_with_ids(numVectors, inputVectors->data(), idVector.data());
-    // Releasing the vectorsAddressJ memory as that is not required once we have created the index.
-    // This is not the ideal approach, please refer this gh issue for long term solution:
-    // https://github.com/opensearch-project/k-NN/issues/1600
-    delete inputVectors;
```
navneet1v, line 223: there was a reason for deallocating the memory here. Please read the gh issue: https://github.com/opensearch-project/k-NN/issues/1600 . The iterative graph build solution is only for HNSW index and not for training based indices. So we should keep this de-allocation here.

navneet1v (changes requested):

src/main/java/org/opensearch/knn/indices/ModelUtil.java
```diff
@@ -48,7 +50,7 @@ public static ModelMetadata getModelMetadata(final String modelId) {
         }
         final Model model = ModelCache.getInstance().get(modelId);
         final ModelMetadata modelMetadata = model.getModelMetadata();
-        if (ModelUtil.isModelCreated(modelMetadata) == false) {
+        if (!ModelUtil.isModelCreated(modelMetadata)) {
```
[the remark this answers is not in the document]
shatejas: Seems like opensearch is deviating from the most used `!` which is considered more concise and readable, is there a coding conventions specific to opensearch that are documented?

jni/src/faiss_index_service.cpp
shatejas: unfortunately no

jni/src/faiss_index_service.cpp
shatejas: This was a part of this change https://github.com/opensearch-project/k-NN/pull/1840/files#diff-f3c66e67165ff091239e0c736a78da59871ad6aa4565bce541911b2da5807bcfR134?

Why do we want to throw if the length is 0?

jni/src/faiss_wrapper.cpp
shatejas: > The iterative graph build solution is only for HNSW index and not for training based indices. So we should keep this de-allocation here.

The code path of vector transfer doesn't defer, the same applies for allocating and deallocating of memory for these input vectors.

I strongly believe that memory management shouldn't be at abrupt locations in the code base. We should be trying to keep the control of the delete with the location where it was allocated to keep it centralized and make the code readable, moreover it removes the responsibility of deallocating the memory from various places in the code

As far as the comment goes, the github issue will track the iterative approach for nmslib and other cases as well

src/main/java/org/opensearch/knn/index/vectorvalues/KNNVectorValues.java
```diff
@@ -23,11 +23,24 @@ public abstract class KNNVectorValues<T> {
...
     protected final KNNVectorValuesIterator vectorValuesIterator;
     protected int dimension;
+    protected long bytesPerVector;
 
     protected KNNVectorValues(final KNNVectorValuesIterator vectorValuesIterator) {
         this.vectorValuesIterator = vectorValuesIterator;
     }
 
+    /**
+     * Iterates the values once only if docIds is not at start position
```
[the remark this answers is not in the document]
shatejas: It iterates

src/main/java/org/opensearch/knn/index/vectorvalues/KNNVectorValues.java
```diff
@@ -23,11 +23,24 @@ public abstract class KNNVectorValues<T> {
...
+     * Also populates dimension and bytesPerVector in the process
+     * @throws IOException
+     */
+    public void init() throws IOException {
+        if (docId() == -1) {
+            nextDoc();
```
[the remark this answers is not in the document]
shatejas: This is a side effect of the decision that we can only get dimensions once we get vectors. We cannot get vectors without calling `nextDoc`. Similar to the fact that the user needs to be aware of the fact that you need to iterate once to get dimensions, the user needs to be aware when to iterate once. Moving this to client leads to repeated code which adds to maintenance

We can rethink the name here but this seems to be the right place rather than a random utility

src/main/java/org/opensearch/knn/index/codec/nativeindex/model/BuildIndexParams.java
```diff
@@ -0,0 +1,23 @@
...
+import lombok.Builder;
+import lombok.Value;
+import org.opensearch.knn.index.VectorDataType;
+import org.opensearch.knn.index.engine.KNNEngine;
+
+import java.util.Map;
+
+@Value
+@Builder
+public class BuildIndexParams {
```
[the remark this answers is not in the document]
shatejas: BuildIndexParams seems more appropriate

src/main/java/org/opensearch/knn/index/codec/nativeindex/MemOptimizedNativeIndexBuildStrategy.java
shatejas: No we can't, the transferDocIds is not always `transferLimit`. We also clear this list once inserted

src/main/java/org/opensearch/knn/jni/JNIService.java
shatejas: https://github.com/opensearch-project/k-NN/pull/1840/files#diff-892c5ae9158d35b9f5fa9b67d53d7373ca6d2bf04d1052add99c329d03645409R70

jni/src/faiss_index_service.cpp
shatejas: I don't see the use of smart pointer, what am I missing?

src/main/java/org/opensearch/knn/index/codec/transfer/OffHeapVectorTransferFactory.java
```diff
@@ -0,0 +1,26 @@
...
+    public static <T> OffHeapVectorTransfer<T> getVectorTransfer(final VectorDataType vectorDataType, final int transferLimit) {
+        switch (vectorDataType) {
+            case FLOAT:
+                return (OffHeapVectorTransfer<T>) new OffHeapFloatVectorTransfer(transferLimit);
+            case BINARY:
+                // TODO: Add binary here
+            case BYTE:
+                return (OffHeapVectorTransfer<T>) new OffHeapByteVectorTransfer(transferLimit);
+            default:
+                throw new IllegalArgumentException("Unsupported vector data type: " + vectorDataType);
```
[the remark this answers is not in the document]
shatejas: "No valid vector transfer implementation" is an implementation detail and doesn't make sense for the user.

src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexWriter.java
```diff
@@ -0,0 +1,295 @@
...
+                    MediaTypeRegistry.getDefaultMediaType()
+                ).map()
+            );
+        }
+
+        parameters.put(KNNConstants.VECTOR_DATA_TYPE_FIELD, vectorDataType.getValue());
+        // In OpenSearch 2.16, we added the prefix for binary indices in the index description in the codec logic.
+        // After 2.16, we added the binary prefix in the faiss library code. However, to ensure backwards compatibility,
+        // we need to ensure that if the description does not contain the prefix but the type is binary, we add the
+        // description.
```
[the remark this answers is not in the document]
shatejas: This is just a part of rebasing from main

jni/src/nmslib_wrapper.cpp
shatejas: https://github.com/opensearch-project/k-NN/pull/1950#discussion_r1721170735

jni/src/faiss_index_service.cpp
shatejas: We need some guidelines on where to use smart pointers in that case so there is not confusion. This seems to be done [here](https://github.com/opensearch-project/k-NN/pull/1840/files#diff-f3c66e67165ff091239e0c736a78da59871ad6aa4565bce541911b2da5807bcfR116) which seemed to be have been approved before. 

I think as long as the memory is managed correctly and is not unambiguous, smart pointers should work.

src/main/java/org/opensearch/knn/indices/ModelUtil.java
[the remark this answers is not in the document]
navneet1v: > is there a coding conventions specific to opensearch that are documented?

I haven't seen something like that. But in general, `!` is not used in Opensearch project. Just a convention, but I have seen benefits of this in long term. So upto you. Just want to point out as you changed the code too.

src/main/java/org/opensearch/knn/index/vectorvalues/KNNVectorValues.java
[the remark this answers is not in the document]
navneet1v: Better you say it moves the docId to first docId location, iterate means something repeat.

src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexWriter.java
```diff
@@ -0,0 +1,295 @@
...
+     */
+    public void mergeIndex(final KNNVectorValues<?> knnVectorValues) throws IOException {
+        knnVectorValues.init();
+        if (knnVectorValues.docId() == NO_MORE_DOCS) {
+            // This is in place so we do not add metrics
+            return;
+        }
+
+        long arraySize = knnVectorValues.bytesPerVector();
+        startMergeStats(knnVectorValues.dimension(), arraySize);
```
navneet1v, line 104: Here you need knnVectorValues.totalLiveDocs and not dimension. because method signature is:
```
 private void startMergeStats(int numDocs, long arraySize) {}
```

src/main/java/org/opensearch/knn/index/vectorvalues/KNNVectorValues.java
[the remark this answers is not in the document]
navneet1v: > We can rethink the name here but this seems to be the right place rather than a random utility

I am not saying to put it in the utility, put it in the NativeIndexWriter. Apart from indexing code, none needs to know what is the dimension. BTW, even on indexing you can use FieldInfo to get the dimension of the vectors.

On checking further on the code this line: https://github.com/opensearch-project/k-NN/pull/1950/files#diff-75a39a358693703da64ecd93181897b1e20ad25f951f0f3813148c21257a999cR104

for binary vector datatype will give size of array * 8 dimension value. See my comment on the same line there you don't need dimension. Hence just putting this on the fact that we cannot get dimension else we read the vector doesn't seem a correct reason for having this function.

src/main/java/org/opensearch/knn/index/codec/transfer/OffHeapVectorTransferFactory.java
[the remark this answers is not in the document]
navneet1v: > "No valid vector transfer implementation" is an implementation detail and doesn't make sense for the user.

The exceptions from these layer doesn't propagate to users. This will be added in logs which will be used by admin and operators. Just saying `Unsupported vector data type` gives no clue what the exception and what are the valid vector datatypes here.

src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexWriter.java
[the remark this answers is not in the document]
navneet1v: @jmazanec15 and @heemin32 why this change is coming in 2.16?

src/main/java/org/opensearch/knn/index/codec/nativeindex/MemOptimizedNativeIndexBuildStrategy.java
navneet1v: > No we can't, the transferDocIds is not always the size of transferLimit. We also clear this list once inserted

actually it is either == transfer limit or whatever is left at the end of the transfer aka last batch. I feel it can optimized. I will still leave it upto you to change it and when you want to change it.

jni/src/faiss_wrapper.cpp
navneet1v: > I strongly believe that memory management shouldn't be at abrupt locations in the code base. 

I also agree on this. But in this case you have to release the memory here, otherwise you will go to the next steps in index writing and still holding the memory which is not required. Hence the comment was added clearly stating why we should do this.

Also this code path is for **template index** and not for iterative builds we should not change the memory release pattern otherwise you will hold more memory even when you start writing the index.

> As far as the comment goes, the github issue will track the iterative approach for nmslib and other cases as well

Yes it should but since we are not making the change for nmslib and other cases we should free up the memory in java layer. Specially in case of nmslib which increases the memory footprint when it write the index on the file.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriter.java
```diff
@@ -62,14 +70,41 @@ public KnnFieldVectorsWriter<?> addField(final FieldInfo fieldInfo) throws IOExc
...
         // This will ensure that we are merging the FlatIndex during force merge.
         flatVectorsWriter.mergeOneField(fieldInfo, mergeState);
-        // TODO: add code for creating Vector datastructures during merge operation
+
+        // For merge, pick values from flat vector and reindex again. This will use the flush operation to create graphs
+        final VectorDataType vectorDataType = extractVectorDataType(fieldInfo);
+        final KNNVectorValues<?> knnVectorValues;
+        switch (fieldInfo.getVectorEncoding()) {
+            case FLOAT32:
+                final FloatVectorValues mergedFloats = MergedVectorValues.mergeFloatVectorValues(fieldInfo, mergeState);
```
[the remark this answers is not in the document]
shatejas: So I looked deeper into this
- line 88: Creates a DocIdSetIterator (FloatVectorValues/ByteVectorValues) from readers and writes in into a merged flat vectors file.
- line 107: Builds the graph. We need a fresh iterator for this hence the code which is merging, its a lazy implementation so its not merging right away.

In this entire process, we are reading from the segment file twice, I wasn't able to trace the code path but my understanding is until we use the same reader, lucene will not delete the files. line 88 merge will create a merged flat vectors file and we can keep using the older files for merging graphs. Please correct me if I have misunderstood something here

The side-effect with this impl is that we are reading twice from segment files - once for merging flat vectors and second for merging the graphs. Alternative would be to move writing of merged flat vectors code here and iteratively build the graph here simultaneously. The caveat here is we won't be able to delegate to flat vectors implementation. 

Thoughts?

src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexWriter.java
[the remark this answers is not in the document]
jmazanec15: This has already been merged. In 2.16, we modified the description in KNN80DocValuesConsumer to have the binary prefix. However, in 2.17, we are building the index description behind the faiss KNNLibrary abstraction, where the rest of the index description is generated.

src/main/java/org/opensearch/knn/index/codec/nativeindex/MemOptimizedNativeIndexBuildStrategy.java
shatejas: Sure it comes with additional array management as a caveat. This was kept the way it was based on previous implementation of Vector transfer

src/main/java/org/opensearch/knn/index/codec/transfer/OffHeapFloatVectorTransfer.java
shatejas: Will make it protected

src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexWriter.java
shatejas: Its confusing because of [this](https://github.com/opensearch-project/k-NN/pull/1950/files#diff-75a39a358693703da64ecd93181897b1e20ad25f951f0f3813148c21257a999cR244) where it says bytes

```
private void startMergeStats(int numDocs, long arraySize) {
        KNNGraphValue.MERGE_CURRENT_OPERATIONS.increment();
        KNNGraphValue.MERGE_CURRENT_DOCS.incrementBy(numDocs);
        KNNGraphValue.MERGE_CURRENT_SIZE_IN_BYTES.incrementBy(arraySize);
        KNNGraphValue.MERGE_TOTAL_OPERATIONS.increment();
        KNNGraphValue.MERGE_TOTAL_DOCS.incrementBy(numDocs);
        KNNGraphValue.MERGE_TOTAL_SIZE_IN_BYTES.incrementBy(arraySize);
    }
```
need clarification here

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriter.java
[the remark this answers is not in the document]
shatejas: Reading twice is okay, Lucene has the same implementation where it will read twice.

Merge is only happening once

CHANGELOG.md
```diff
@@ -17,11 +17,16 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
...
 ### Enhancements
+* Add functionality to iteratively insert vectors into a faiss index to improve the memory footprint during indexing. [#1840](https://github.com/opensearch-project/k-NN/pull/1840)
 ### Bug Fixes
 * Corrected search logic for scenario with non-existent fields in filter [#1874](https://github.com/opensearch-project/k-NN/pull/1874)
 * Add script_fields context to KNNAllowlist [#1917] (https://github.com/opensearch-project/k-NN/pull/1917)
 * Fix graph merge stats size calculation [#1844](https://github.com/opensearch-project/k-NN/pull/1844) 
 * Disallow a vector field to have an invalid character for a physical file name. [#1936](https://github.com/opensearch-project/k-NN/pull/1936)
+* Fix graph merge stats size calculation [#1844](https://github.com/opensearch-project/k-NN/pull/1844)
+* Integrate Lucene Vector field with native engines to use KNNVectorFormat during segment creation [#1945](https://github.com/opensearch-project/k-NN/pull/1945)
+* Disallow a vector field to have an invalid character for a physical file name. [#1936] (https://github.com/opensearch-project/k-NN/pull/1936)
```
navneet1v, line 28: Few of these changelogs were moved to separate sections. Please check the latest changelog. Like `Integrate Lucene Vector field with native engines to use KNNVectorFormat during segment creation` is not a bug fix.

navneet1v:
Overall code looks good to me. CIs are failing due to some backward incompatible changes in main branch of Opensearch. That needs to be fixed first and then CIs needs to be run again. Approving this PR assuming CIs will pass. But lets not merge the PR till CIs are passing.

navneet1v (approved):

jmazanec15 (dismissed):

navneet1v (approved):

ryanbogan (approved):
LGTM! CIs are passing now

PR #1957, opened 2024-08-14
Encapsulate dimension, vector data type validation/processing inside Library

jmazanec15:
### Description
As part of #1779 , we need to have the ability to take a user config for a float based space type (i.e. l2) and configure the quantization framework to go from float to bit and build a faiss binary index. In the current implementation, the data type validation is outside of the KNNLibrary/engine abstractions. This will make it difficult do the complex configuration around the quantization framework index building and search. The main theme of this PR is to move towards handling Library specific configuration to the library.

This PR does the following:
1. Adds `KNNMethodConfigContext` that is passed to `KNNLibrary` for validation as well as processing. The `KNNMethodConfigContext` object has information outside of what is passed by the user for configuring the method. This will allow the `KNNLibrary` specific components to do more complex validation/processing logic such as configuring a binary index, or figuring out how to validate/process incoming vectors. This wasnt added as a part of `KNNMethodContext` because it would bloat the class further. This is somewhat of an extension to the `VectorSpaceInfo` abstraction
2. For per-dimension processing/validation, faiss fp16, we moved the different validators/processors to the Library so we dont have to extract encoder information from the mapping. 
3. Dimension check's and version created were added to `KNNMethodConfigContext`

### Check List


src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80DocValuesConsumer.java
```diff
@@ -243,18 +242,6 @@ private void createKNNIndexFromScratch(FieldInfo fieldInfo, KNNCodecUtil.Pair pa
             );
         }
 
-        // Update index description of Faiss for binary data type
-        if (KNNEngine.FAISS == knnEngine
```
navneet1v, line 247: what will happen if segment is an old segment. I think we will still need this update for old segments. We might need to find a way to check if the index description is correct for binary index and if it not correct then update it otherwise don't update the index description.

navneet1v (commented):
Overall looks good to me. Just 1 comment on the BWC at the codec level

jmazanec15:
@navneet1v ah I forgot about that - good catch. Let me fix

qa/restart-upgrade/src/test/java/org/opensearch/knn/bwc/AbstractRestartUpgradeTestCase.java
```diff
@@ -59,4 +59,10 @@ protected final Optional<String> getBWCVersion() {
         return Optional.ofNullable(System.getProperty(BWC_VERSION, null));
     }
 
+    protected String maybeRemoveSnapshotSuffix(String version) {
```
navneet1v, line 62: why we are doing this?

qa/restart-upgrade/src/test/java/org/opensearch/knn/bwc/IndexingIT.java
```diff
@@ -68,6 +71,37 @@ public void testKNNIndexDefaultLegacyFieldMappingForceMerge() throws Exception {
         }
     }
 
+    // Ensure bwc works for binary force merge
+    public void testKNNIndexBinaryForceMerge() throws Exception {
+        // Binary is only supported on or after 2.16. Need to skip the others
+        if (getBWCVersion().isPresent() && Version.fromString(maybeRemoveSnapshotSuffix(getBWCVersion().get())).before(Version.V_2_16_0)) {
+            return;
+        }
```
navneet1v, line 79: There is another way to do this skipping through brazil.gradle. wondering why we choose this way?

https://github.com/opensearch-project/neural-search/blob/main/qa/restart-upgrade/build.gradle#L134-L142

jmazanec15: fixed it

navneet1v (approved):

navneet1v:
@jmazanec15 seems like a build failure.. can you check once whats happening

jmazanec15:
@navneet1v I dont see anything interesting in the logs and am not able to reproduce it locally. Im guessing its unrelated/flaky. Captured the logs: https://gist.github.com/jmazanec15/ec985b4aeeae3325d5159974cfb375ab. Ill retry

navneet1v (approved):

junqiu-lei (approved):

PR #1960, opened 2024-08-14
Add quantization state cache

ryanbogan:
### Description
Adds a cache for storing quantization states.  This PR contains the cache and tests without integration.  In a later PR, this cache will be integrated with a quantization state reader and the query flow.

### Check List


Vikasht34:
Quantization Cache Class Looks good , but we need to improve on unit tests for following test cases
1. testSingleThreadedAddAndRetrieve
2.testMultiThreadedAddAndRetrieve
3.testMultiThreadedEvict
4.testConcurrentAddAndEvict
5.testCacheClearFromMultipleThreads

Let's have all these tests added.

Vikasht34 (approved):
Looks Good to me .

src/main/java/org/opensearch/knn/quantization/models/quantizationState/QuantizationStateCache.java
```diff
@@ -0,0 +1,70 @@
...
+
+package org.opensearch.knn.quantization.models.quantizationState;
+
+import com.google.common.cache.Cache;
+import com.google.common.cache.CacheBuilder;
+
+/**
+ * A thread-safe singleton cache that contains quantization states.
+ */
+public class QuantizationStateCache {
```
navneet1v, line 14: Questions as this cache will be storing information which is at a segment level:
1. How we are controlling the max size of the cache? and given that this cache will be present in Heap, will the size of the cache be static? How we are controlling the Cache thrashing use case and how as a Opensearch user I will know there is cache thrashing?
2. During eviction, what is the eviction policy getting used for this cache?
3. What controls we have to increase and decrease the size of cache?
4. What controls we are providing to clear the cache do we need it?
5. Should we also have a TTL based eviction?
6. Should we integrate stats of this cache in knn stats api?

ryanbogan:
The most recent commits should address @navneet1v 's comments 1-5.  Will add to stats in follow up PR.  I also still need to add more unit testing for the new functionality.

src/main/java/org/opensearch/knn/index/KNNSettings.java
```diff
@@ -106,6 +108,11 @@ public class KNNSettings {
     public static final String KNN_DEFAULT_VECTOR_STREAMING_MEMORY_LIMIT_PCT = "1%";
 
     public static final Integer ADVANCED_FILTERED_EXACT_SEARCH_THRESHOLD_DEFAULT_VALUE = -1;
+    public static final Integer KNN_DEFAULT_QUANTIZATION_STATE_CACHE_SIZE_LIMIT_PERCENTAGE = 5; // By default, set aside 5% of the JVM for
+                                                                                                // the limit
+    public static final Integer KNN_MAX_QUANTIZATION_STATE_CACHE_SIZE_LIMIT_PERCENTAGE = 10; // Quantization state cache limit cannot exceed
+                                                                                             // 10% of the JVM heap
+    public static final Integer KNN_DEFAULT_QUANTIZATION_STATE_CACHE_EXPIRY_TIME_MINUTES = 60;
```
navneet1v, line 115: what is reasoning behind these limits?

src/main/java/org/opensearch/knn/quantization/models/quantizationState/QuantizationStateCache.java
```diff
@@ -0,0 +1,132 @@
...
+    private QuantizationStateCache() {
+        maxCacheSizeInKB = QUANTIZATION_STATE_CACHE_SIZE_LIMIT_SETTING.get(clusterService.getSettings()).getKb();
+        clusterService.getClusterSettings().addSettingsUpdateConsumer(QUANTIZATION_STATE_CACHE_SIZE_LIMIT_SETTING, it -> {
+            maxCacheSizeInKB = it.getKb();
+            rebuildCache();
+        });
+        clusterService.getClusterSettings().addSettingsUpdateConsumer(QUANTIZATION_STATE_CACHE_EXPIRY_TIME_MINUTES_SETTING, it -> {
+            rebuildCache();
+        });
+        buildCache();
```
navneet1v, line 43: Question: should this be logic present in the KNNSettings class?

@jmazanec15 thoughts?

src/main/java/org/opensearch/knn/quantization/models/quantizationState/QuantizationStateCache.java
```diff
@@ -0,0 +1,132 @@
...
+ * A thread-safe singleton cache that contains quantization states.
+ */
+@Log4j2
+public class QuantizationStateCache {
+
+    private static volatile QuantizationStateCache instance;
+    private static ClusterService clusterService;
+    private Cache<String, QuantizationState> cache;
+    private long maxCacheSizeInKB;
+    private Instant evictedDueToSizeAt;
```
navneet1v, line 32: shouldn't this just be a counter of type long?

src/main/java/org/opensearch/knn/quantization/models/quantizationState/QuantizationStateCache.java
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    /**
+     * Retrieves the quantization state associated with a given field name.
+     * @param fieldName The name of the field.
+     * @return The associated QuantizationState, or null if not present.
+     */
+    public QuantizationState getQuantizationState(String fieldName) {
+        return cache.getIfPresent(fieldName);
+    }
```
navneet1v, line 97: Shouldn't this be more of a logic where if entry is not present in the cache then we load the value in the cache and then return that value?

Any reason of not making this as Read-Through cache?

@Vikasht34 thoghts?

src/main/java/org/opensearch/knn/quantization/models/quantizationState/QuantizationStateCache.java
```diff
@@ -0,0 +1,132 @@
...
+
+    private void updateEvictedDueToSizeAt() {
+        evictedDueToSizeAt = Instant.now();
+    }
+
+    /**
+     * Clears all entries from the cache.
+     */
+    public void clear() {
```
navneet1v, line 129: any reason of making this function public?

src/main/java/org/opensearch/knn/index/KNNSettings.java
ryanbogan: I added initial ones based on conversation with @Vikasht34 .  We'll probably need to think more on this with benchmarking but for now I needed a value to put

src/main/java/org/opensearch/knn/quantization/models/quantizationState/QuantizationStateCache.java
ryanbogan: I could put it there.  `ModelCache` has it in the constructor while `NativeMemoryCacheManager` puts it in `KNNSettings`.

src/main/java/org/opensearch/knn/quantization/models/quantizationState/QuantizationStateCache.java
ryanbogan: Not sure I understand, why would it be a counter instead of just a long holding the max size (which is calculated from the percent setting)

src/main/java/org/opensearch/knn/quantization/models/quantizationState/QuantizationStateCache.java
ryanbogan: Eventually, it will be used when indices are closed, which separates it from the rebuild() method.

src/main/java/org/opensearch/knn/index/KNNSettings.java
Vikasht34: Here is the reasoning behind this 

Number of segments :- 2500 ( this must be exterme) 
Dimesion 16000 
Binary Quantization 4 bit 
Threshould will looks like this 
```
thresholds[0] {0.1f, 0.2f, 0.3f  upto 16k }
     *   thresholds[1] {0.4f, 0.5f, 0.6f upto 16k}
     *   thresholds[2] {0.7f, 0.8f, 0.9f upto 16k}
     *   thresholds[3] {1.0f, 1.1f, 1.2f upto 16k}
``` 
For 2500 segments, the total size of the thresholds array would be:
2500 segments x 16,000 columns x 4 bytes/element = 1,600,000,000 bytes
Converting this to more familiar units:

    1,600,000,000 bytes = 1.6 GB (gigabytes)

So, the total size of your thresholds array for 2500 segments would be approximately 1.6 GB.
If you have a 32 GB heap, this array would occupy:
(1.6 GB / 32 GB) x 100% ≈ 5%

this is extreme case assuming all vectors are 16000 size and 2500 segments 

Assuming P90 case where avregae Dimesion is 5000 with 4bit :- 0.5 GB which 1.5625% of 32gb heap , so we choose 5% as safe limit which can handle exterme case as well.

src/main/java/org/opensearch/knn/quantization/models/quantizationState/QuantizationStateCache.java
Vikasht34: Well here is my thoughts on this , I am really not in favour of clubbing cache with Quantization Reader, since we don't know what would be future for Cache interms of unifying intefaces across multiple caches and implementation , I would keep this totally light weight, 

Neither I want my NativeQuantizationState Reader to know about cache .

QuantizationStateCache: Manages in-memory caching of quantization states.
QuantizationStateReader: Handles reading quantization states from persistent storage.
QuantizationStateWriter: Manages writing quantization states to persistent storage.

From Query flow 
I will have my Interface 

```
public interface QuantizationStateLoader {
    QuantizationState getQuantizationState(String fieldName, String typeofQuantization);
}
```

and this inteface will handle both cache and NativeQuantizationStateReader , to read it from cache , if not read from file , deseralize based on type of Quantization and save it Cache. 


So my Cache is clean , neither it has to know about type of Quantization State to deseralize and neither about how it was originally stored.

src/main/java/org/opensearch/knn/quantization/models/quantizationState/QuantizationStateCache.java
Vikasht34: My thought on this , If we could decouple ClusterService from here would be great? I am not sure if that's possible.

ryanbogan: Done

src/main/java/org/opensearch/knn/quantization/models/quantizationState/QuantizationStateCache.java
```diff
@@ -0,0 +1,123 @@
...
+    private QuantizationStateCache() {
+        maxCacheSizeInKB = ((ByteSizeValue) KNNSettings.state().getSettingValue(QUANTIZATION_STATE_CACHE_SIZE_LIMIT)).getKb();
+        buildCache();
+    }
+
+    /**
+     * Gets the singleton instance of the cache.
+     * @return QuantizationStateCache
+     */
+    public static QuantizationStateCache getInstance() {
```
shatejas, line 49: `QuantizationStateCache ` creates another cache layer in the code base for a specific purpose

This can be simplified to `Cache<String, QuantizationState> getInstance()`. methods like `getQuatizationState` are just wrappers over the exising Cache methods, with each functionality we add, we need another wrapper function. Would encourage you to rethink this so we don't have wrapper methods and this can be simply be used as initialize and tear down of the cache

src/main/java/org/opensearch/knn/index/KNNSettings.java
```diff
@@ -272,6 +280,44 @@ public class KNNSettings {
         NodeScope
     );
 
+    /*
+     * Quantization state cache settings
+     */
+    public static final Setting<ByteSizeValue> QUANTIZATION_STATE_CACHE_SIZE_LIMIT_SETTING = new Setting<ByteSizeValue>(
```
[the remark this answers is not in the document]
ryanbogan: Giving the user the option to change these settings can help with debugging clusters and limiting the situation where clusters being under-provisioned can hurt the cache.  It also allows the users to choose how much latency they want to achieve versus the extra heap space they'll have to allocate.

src/main/java/org/opensearch/knn/quantization/models/quantizationState/QuantizationStateCache.java
ryanbogan: This is modeled after other similar caches in the plugin such as ModelCache and NativeMemoryCacheManager.  There is more functionality here, such as the expired time and max size in kb that will eventually be present in the KNNStats API, which will help users debug and have more insight into potential issues with their clusters.

shatejas (approved):

src/main/java/org/opensearch/knn/quantization/models/quantizationState/QuantizationStateCache.java
ryanbogan: We'll take this as a follow-up PR after thinking about it more

ryanbogan:
There's one flaky test I'm still trying to fix

src/main/java/org/opensearch/knn/index/KNNSettings.java
jmazanec15: >  2500 segments x 16,000 columns x 4 bytes/element = 1,600,000,000 bytes

This is 160,000,000 bytes - 150 MB. 5k is 50 MB. That being said, I dont think we need these limits.

jmazanec15: Also, if we do need the limits, can we not make them settings? Settings are hard to remove. It seems in the future, we may want to rely on IndexInput and mmap to handle caching. Id prefer not to expose these as settings

src/main/java/org/opensearch/knn/quantization/models/quantizationState/QuantizationStateCache.java
```diff
@@ -0,0 +1,123 @@
...
+                }
+            }
+        }
+        return instance;
+    }
+
+    private void buildCache() {
+        this.cache = CacheBuilder.newBuilder().concurrencyLevel(1).maximumWeight(maxCacheSizeInKB).weigher((k, v) -> {
+            try {
+                return ((QuantizationState) v).toByteArray().length;
```
jmazanec15, line 65: We need to serialize to get the weight? That seems too expensive

ryanbogan: As of right now, the only guaranteed methods to be in Quantization State are getting the params and toByteArray.  This was the only way I could find to estimate the size with those methods

jmazanec15: We can take in another review on this - but need to track.

Vikasht34: Let me take a look on this. Will address this in upcoming pr.

jmazanec15 (approved):

src/main/java/org/opensearch/knn/quantization/models/quantizationState/QuantizationStateCache.java
navneet1v: > So Cache is clean , neither it has to know about type of Quantization State to deseralize and neither about how it was originally stored.

if are looking for making cache clean then the current implementation is not clean. It already has knowledge of `QuantizationState` object and other QS details. 

But whether we want to make cache clean or not that is separate question. But it doesn't ans the question why this cache is not a Read-Through cache? You can just provide a loader functional interface here which is just loads the entry in cache if its not present. Then no-one needs to know about QS.

navneet1v: The problem with this kind of design is, we will have to write a wrapper on top of it which will first check if the entry in cache is present or not, and then load data in the cache. Which can be just avoided here.

navneet1v (approved):

Issue #1771, opened 2024-06-25
[FEATURE] Support rolling upgrades for version based changes


kotwanikunal:
**Is your feature request related to a problem?**
- Currently, version specific features are enabled by fetching the minimum node version from the cluster: [here](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/IndexUtil.java#L258-L263)
- The current mechanism of checking for versions does not support rolling upgrades and can be enhanced to be consistent with the @opensearch-project/OpenSearch core logic

**What solution would you like?**
- Utilize the transport stream version for enabling version support rather than using the cluster state
- Sample PR: https://github.com/opensearch-project/k-NN/pull/1770

**What alternatives have you considered?**
None

**Do you have any additional context?**
- N/A

PR #1969, opened 2024-08-15
Adds rescore parameter to KNNQuery

jmazanec15:
### Description
Adds rescore parameter to knn query. With this commit, the rescore is a no-op. The functionality and validation will be added in a later commit.

The valid options look like:
```
"query": {
  "knn": {
    "target_field": {
      "vector": [0, 0],
      "k": 100,
      "rescore": true
    }
  }
}

"query": {
  "knn": {
    "target_field": {
      "vector": [0, 0],
      "k": 100,
      "rescore": {}
    }
  }
}

  "query": {
    "knn": {
      "target_field": {
        "vector": [0, 0],
        "k": 100,
        "rescore": {
            "oversample_factor": 1.2
        }
      }
    }
  }
```

Will update changelog once feature is completed

### Related Issues
Related to #1779 

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/query/parser/KNNQueryBuilderParser.java
+++ b/src/main/java/org/opensearch/knn/index/query/parser/KNNQueryBuilderParser.java
@@ -30,8 +30,8 @@ import static org.opensearch.index.query.AbstractQueryBuilder.BOOST_FIELD;
 import static org.opensearch.index.query.AbstractQueryBuilder.NAME_FIELD;
 import static org.opensearch.index.query.AbstractQueryBuilder.parseInnerQueryBuilder;
 import static org.opensearch.knn.common.KNNConstants.METHOD_PARAMETER;
-import static org.opensearch.knn.common.KNNConstants.RESCORE_PARAMETER;
 import static org.opensearch.knn.index.query.KNNQueryBuilder.RESCORE_FIELD;
+import static org.opensearch.knn.index.query.parser.RescoreParser.RESCORE_PARAMETER;
 import static org.opensearch.knn.index.util.IndexUtil.isClusterOnOrAfterMinRequiredVersion;
 import static org.opensearch.knn.index.query.KNNQueryBuilder.FILTER_FIELD;
 import static org.opensearch.knn.index.query.KNNQueryBuilder.IGNORE_UNMAPPED_FIELD;
@@ -125,7 +125,7 @@ public final class KNNQueryBuilderParser {
         }
 
         if (minClusterVersionCheck.apply(RESCORE_PARAMETER)) {
-            builder.rescoreContext(RescoreParser.streamInput(in, IndexUtil::isClusterOnOrAfterMinRequiredVersion));
+            builder.rescoreContext(RescoreParser.streamInput(in));
         }
 
         return builder;
@@ -158,7 +158,7 @@ public final class KNNQueryBuilderParser {
             MethodParametersParser.streamOutput(out, builder.getMethodParameters(), IndexUtil::isClusterOnOrAfterMinRequiredVersion);
         }
         if (minClusterVersionCheck.apply(RESCORE_PARAMETER)) {
-            RescoreParser.streamOutput(out, builder.getRescoreContext(), IndexUtil::isClusterOnOrAfterMinRequiredVersion);
+            RescoreParser.streamOutput(out, builder.getRescoreContext());
         }
     }
 
--- a/src/main/java/org/opensearch/knn/index/query/parser/KNNQueryBuilderParser.java
+++ b/src/main/java/org/opensearch/knn/index/query/parser/KNNQueryBuilderParser.java
@@ -16,6 +16,7 @@ import org.opensearch.core.xcontent.XContentLocation;
 import org.opensearch.core.xcontent.XContentParser;
 import org.opensearch.index.query.QueryBuilder;
 import org.opensearch.knn.common.KNNConstants;
+import org.opensearch.knn.index.query.rescore.RescoreContext;
 import org.opensearch.knn.index.util.IndexUtil;
 import org.opensearch.knn.index.query.KNNQueryBuilder;
 
@@ -29,6 +30,8 @@ import static org.opensearch.index.query.AbstractQueryBuilder.BOOST_FIELD;
 import static org.opensearch.index.query.AbstractQueryBuilder.NAME_FIELD;
 import static org.opensearch.index.query.AbstractQueryBuilder.parseInnerQueryBuilder;
 import static org.opensearch.knn.common.KNNConstants.METHOD_PARAMETER;
+import static org.opensearch.knn.common.KNNConstants.RESCORE_PARAMETER;
+import static org.opensearch.knn.index.query.KNNQueryBuilder.RESCORE_FIELD;
 import static org.opensearch.knn.index.util.IndexUtil.isClusterOnOrAfterMinRequiredVersion;
 import static org.opensearch.knn.index.query.KNNQueryBuilder.FILTER_FIELD;
 import static org.opensearch.knn.index.query.KNNQueryBuilder.IGNORE_UNMAPPED_FIELD;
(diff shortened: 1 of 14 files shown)


src/main/java/org/opensearch/knn/index/query/rescore/RescoreContext.java
```diff
@@ -0,0 +1,68 @@
...
+
+    @Builder.Default
+    private float oversampleFactor = DEFAULT_OVERSAMPLE_FACTOR;
+
+    /**
+     * Validate the rescore context
+     *
+     * @return ValidationException if validation fails, null otherwise
+     */
+    public ValidationException validate() {
```
shatejas, line 34: Out of curiosity, there are places in the code base where there is a similar pattern, The POJO is coupled with validation. Ideally these should just have getters and setters, validations should be a separate concern, we ideally wouldn't want to have access to any logic apart from getters and setters to wherever this object is being passed around

Thoughts?

src/main/java/org/opensearch/knn/index/query/parser/KNNQueryBuilderParser.java
```diff
@@ -110,6 +124,10 @@ public static KNNQueryBuilder.Builder streamInput(StreamInput in, Function<Strin
             builder.methodParameters(MethodParametersParser.streamInput(in, IndexUtil::isClusterOnOrAfterMinRequiredVersion));
         }
 
+        if (minClusterVersionCheck.apply(RESCORE_PARAMETER)) {
```
shatejas, line 127: I think this will affect neural, just a heads up

src/main/java/org/opensearch/knn/index/query/parser/RescoreParser.java
```diff
@@ -0,0 +1,103 @@
...
+        if (rescoreContext == null) {
+            out.writeBoolean(false);
+            return;
+        }
+
+        out.writeBoolean(true);
+        if (minClusterVersionCheck.apply(RESCORE_PARAMETER)) {
+            out.writeFloat(rescoreContext.getOversampleFactor());
+        }
+    }
```
navneet1v, line 81: I trying to see how this code is backward compatible. We are adding a new boolean on the stream before checking the min cluster version. Would you please provide some details how this will be BWC? Shouldn't you check first the min cluster version and then see you want to read or write the boolean.

Another point:
I know we use the min cluster version everywhere but that is not the best way, we have a min stream version also available. I would recommend using it. Ref: https://github.com/opensearch-project/k-NN/issues/1771 it has a sample PR also on how to do it. Please see if we can do that.

src/main/java/org/opensearch/knn/index/query/rescore/RescoreContext.java
```diff
@@ -0,0 +1,68 @@
...
+
+@Slf4j
+@Getter
+@AllArgsConstructor
+@Builder
+@EqualsAndHashCode
+public final class RescoreContext {
+
+    public static final float DEFAULT_OVERSAMPLE_FACTOR = 1.0f;
+    public static final float MAX_OVERSAMPLE_FACTOR = 5.0f;
```
navneet1v, line 23: @jmazanec15 will 5 be good enough may be 100? I know nothing is good enough. but I was thinking with 100 as max and a cap of 10K on the k*oversampling factor should cover most of the cases.

Thoughts?

src/main/java/org/opensearch/knn/index/util/IndexUtil.java
```diff
@@ -48,6 +48,8 @@ public class IndexUtil {
     private static final Version MINIMAL_SUPPORTED_VERSION_FOR_RADIAL_SEARCH = Version.V_2_14_0;
     private static final Version MINIMAL_SUPPORTED_VERSION_FOR_METHOD_PARAMETERS = Version.V_2_16_0;
     private static final Version MINIMAL_SUPPORTED_VERSION_FOR_MODEL_VECTOR_DATA_TYPE = Version.V_2_16_0;
+    // TODO: Will update once 2.17 backport change is merged
+    private static final Version MINIMAL_RESCORE_FEATURE = Version.V_3_0_0;
```
shatejas, line 52: nit: Look around if there is a need to have this in neural plugin

src/main/java/org/opensearch/knn/index/query/parser/RescoreParser.java
jmazanec15: Let me take a look

src/test/java/org/opensearch/knn/index/query/KNNQueryBuilderTests.java
```diff
@@ -776,19 +777,23 @@ public void testDoToQuery_InvalidZeroByteVector() {
 
     public void testSerialization() throws Exception {
```
shatejas, line 778: nit: Does KNNQueryBuilderValidParamsTests and KNNQueryBuilderInvalidParamsTests need an update?

src/main/java/org/opensearch/knn/index/query/rescore/RescoreContext.java
jmazanec15: Sure I can increase. Let me update

src/main/java/org/opensearch/knn/index/query/rescore/RescoreContext.java
```diff
@@ -0,0 +1,68 @@
...
+                    RescoreContext.MIN_OVERSAMPLE_FACTOR
+                )
+            );
+            return validationException;
+        }
+
+        if (oversampleFactor > RescoreContext.MAX_OVERSAMPLE_FACTOR) {
+            ValidationException validationException = new ValidationException();
+            validationException.addValidationError(
+                String.format(
```
shatejas, line 50: Local.ROOT please or convert it to string

src/main/java/org/opensearch/knn/index/query/rescore/RescoreContext.java
```diff
@@ -0,0 +1,68 @@
...
+    /**
+     * Validate the rescore context
+     *
+     * @return ValidationException if validation fails, null otherwise
+     */
+    public ValidationException validate() {
+        if (oversampleFactor < RescoreContext.MIN_OVERSAMPLE_FACTOR) {
+            ValidationException validationException = new ValidationException();
+            validationException.addValidationError(
+                String.format(
```
shatejas, line 38: same

src/main/java/org/opensearch/knn/index/query/rescore/RescoreContext.java
```diff
@@ -0,0 +1,68 @@
...
+package org.opensearch.knn.index.query.rescore;
+
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.extern.slf4j.Slf4j;
+import org.opensearch.common.ValidationException;
+
+@Slf4j
```
shatejas, line 15: Is this needed?

naveentatikonda (approved):

src/main/java/org/opensearch/knn/index/query/rescore/RescoreContext.java
jmazanec15: I think that makes sense. Let me update to move it to parser similar to method params

src/main/java/org/opensearch/knn/index/query/rescore/RescoreContext.java
jmazanec15: no let me fix

src/test/java/org/opensearch/knn/index/query/KNNQueryBuilderTests.java
jmazanec15: Ill add some basic ones, but do most of testing in the Rescore validation tests

src/main/java/org/opensearch/knn/index/query/parser/RescoreParser.java
jmazanec15: updated both

src/main/java/org/opensearch/knn/index/query/rescore/RescoreContext.java
jmazanec15: done

ryanbogan (approved):
LGTM

src/test/java/org/opensearch/knn/integ/QueryParseIT.java
```diff
@@ -0,0 +1,148 @@
...
+import org.opensearch.core.xcontent.XContentBuilder;
+import org.opensearch.knn.KNNRestTestCase;
+
+import java.io.IOException;
+import java.util.Locale;
+
+import static org.opensearch.knn.index.query.parser.RescoreParser.RESCORE_OVERSAMPLE_PARAMETER;
+import static org.opensearch.knn.index.query.parser.RescoreParser.RESCORE_PARAMETER;
+
+public class QueryParseIT extends KNNRestTestCase {
```
shatejas, line 21: nit: https://github.com/opensearch-project/k-NN/blob/main/src/test/java/org/opensearch/knn/index/query/InvalidSearchQueryIT.java 

Seems like a duplicate in a way, maybe added a todo on one of them to merge for later

shatejas (approved):

Vikasht34 (approved):
LGTM !!

src/test/java/org/opensearch/knn/integ/QueryParseIT.java
jmazanec15: Yes, that probably makes sense. Let me make a note on this and come back with it in another PR, with some other test cleanup.

naveentatikonda (approved):

Issue #1594, opened 2024-04-05
[Enhancement] Create annotation in K-NN plugin to measure the execution time and logging k-nn Stats


navneet1v:
### Description
Currently in k-NN code base we time different functions like graph creation etc, and also we put some stats in k-NN stats api related to different invocations of different code path. To achieve this we have duplicated code all over the place.

To make it easier we can build some custom annotation in java that can do this work for us and which can be reused at different place.

Some annotation that I am thinking:
1. @LogExecutionTime
2. @EmitStats(name="FaissInit")

There can be many more.

Issue #1491, opened 2024-02-21
Auto-filter deleted documents on search for native engines


jmazanec15:
# Description

Recently, we added a lot of filtering capabilities for faiss so that users can pass filters to hnsw, ivf, ivfpq, etc. When searching a segment, we should automatically add deleted documents to this filter, so that faiss can skip these as well instead of returning results and getting filtered out after. Currently, lucene implements this behavior for HNSW: https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsReader.java#L236-L238. There was some discussion around this in #1003

navneet1v:
Thanks @jmazanec15 for opening up this issue. I was going to open this issue. I think we should definitely implement a deleted documents filtering capability for Faiss.

But before we implement we should optimize the filtering in general for k-NN. I am working on detailing all the different optimization that is required to be done in filtering. Once we have those optimizations in place we should implement this feature.

jmazanec15:
Moving to 2.15. I dont think we are going to get in 2.14

vamshin:
Moving to 2.16.

kr4sv:
Hey,
with faiss becoming the default knn engine, is there a target release version for this enhancement?
thx

PR #1984, opened 2024-08-19
k-NN query rescore support for native engines

jmazanec15:
### Description
Implements re-scoring. It uses the rescoring context in the query builder to execute a two-phased search. First, it oversamples the ANN index and then reduces the results down to the oversample factor, and then rescores and returns.

Shared code for exact search was pulled out into a class called exact searcher. Rescore search functionality is included in a class called RescoreSearcher. A few sanity integ tests have been added along with existing ones, but intend to add more once quantization framework is integrated. 

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/common/FieldInfoExtractor.java
+++ b/src/main/java/org/opensearch/knn/common/FieldInfoExtractor.java
@@ -8,6 +8,7 @@ package org.opensearch.knn.common;
 import lombok.experimental.UtilityClass;
 import org.apache.commons.lang.StringUtils;
 import org.apache.lucene.index.FieldInfo;
+import org.opensearch.knn.index.SpaceType;
 import org.opensearch.knn.index.VectorDataType;
 import org.opensearch.knn.index.engine.KNNEngine;
 import org.opensearch.knn.indices.ModelMetadata;
@@ -19,6 +20,11 @@ import org.opensearch.knn.index.engine.qframe.QuantizationConfig;
 import org.opensearch.knn.index.engine.qframe.QuantizationConfigParser;
 
 import static org.opensearch.knn.common.KNNConstants.QFRAMEWORK_CONFIG;
+import org.opensearch.knn.indices.ModelDao;
+
+import java.util.Locale;
+
+import static org.opensearch.knn.common.KNNConstants.SPACE_TYPE;
 
 /**
  * A utility class to extract information from FieldInfo.
@@ -69,5 +75,32 @@ public class FieldInfoExtractor {
             return QuantizationConfig.EMPTY;
         }
         return QuantizationConfigParser.fromCsv(quantizationConfigString);
+    }
+
+    /**
+     * Get the space type for the given field info.
+     *
+     * @param modelDao ModelDao instance to retrieve model metadata
+     * @param fieldInfo FieldInfo instance to extract space type from
+     * @return SpaceType for the given field info
+     */
+    public static SpaceType getSpaceType(final ModelDao modelDao, final FieldInfo fieldInfo) {
+        final String spaceTypeString = fieldInfo.getAttribute(SPACE_TYPE);
+        if (StringUtils.isNotEmpty(spaceTypeString)) {
+            return SpaceType.getSpace(spaceTypeString);
+        }
+
+        final String modelId = fieldInfo.getAttribute(MODEL_ID);
+        if (StringUtils.isEmpty(modelId)) {
+            throw new IllegalArgumentException(
+                String.format(Locale.ROOT, "Unable to find the Space Type from Field Info attribute for field %s", fieldInfo.getName())
+            );
+        }
+
+        ModelMetadata modelMetadata = modelDao.getMetadata(modelId);
+        if (modelMetadata == null) {
+            throw new IllegalArgumentException(String.format(Locale.ROOT, "Unable to find the model metadata for model id %s", modelId));
+        }
+        return modelMetadata.getSpaceType();
     }
 }
--- a/src/main/java/org/opensearch/knn/index/query/KNNQuery.java
+++ b/src/main/java/org/opensearch/knn/index/query/KNNQuery.java
@@ -21,6 +21,7 @@ import org.apache.lucene.search.Weight;
 import org.apache.lucene.search.join.BitSetProducer;
 import org.opensearch.knn.index.KNNSettings;
 import org.opensearch.knn.index.VectorDataType;
+import org.opensearch.knn.index.query.rescore.RescoreContext;
 
 import java.io.IOException;
 import java.util.Arrays;
@@ -38,13 +39,12 @@ public class KNNQuery extends Query {
 
     private final String field;
     private final float[] queryVector;
-    @Getter
     private final byte[] byteQueryVector;
     private int k;
     private Map<String, ?> methodParameters;
     private final String indexName;
-    @Getter
     private final VectorDataType vectorDataType;
+    private final RescoreContext rescoreContext;
 
     @Setter
     private Query filterQuery;
@@ -59,7 +59,7 @@ public class KNNQuery extends Query {
         final String indexName,
         final BitSetProducer parentsFilter
     ) {
-        this(field, queryVector, null, k, indexName, null, parentsFilter, VectorDataType.FLOAT);
+        this(field, queryVector, null, k, indexName, null, parentsFilter, VectorDataType.FLOAT, null);
     }
 
     public KNNQuery(
@@ -68,9 +68,10 @@ public class KNNQuery extends Query {
         final int k,
         final String indexName,
         final Query filterQuery,
-        final BitSetProducer parentsFilter
+        final BitSetProducer parentsFilter,
+        final RescoreContext rescoreContext
     ) {
-        this(field, queryVector, null, k, indexName, filterQuery, parentsFilter, VectorDataType.FLOAT);
+        this(field, queryVector, null, k, indexName, filterQuery, parentsFilter, VectorDataType.FLOAT, rescoreContext);
     }
 
     public KNNQuery(
@@ -80,9 +81,10 @@ public class KNNQuery extends Query {
         final String indexName,
         final Query filterQuery,
         final BitSetProducer parentsFilter,
-        final VectorDataType vectorDataType
+        final VectorDataType vectorDataType,
+        final RescoreContext rescoreContext
     ) {
-        this(field, null, byteQueryVector, k, indexName, filterQuery, parentsFilter, vectorDataType);
+        this(field, null, byteQueryVector, k, indexName, filterQuery, parentsFilter, vectorDataType, rescoreContext);
     }
 
     private KNNQuery(
@@ -93,7 +95,8 @@ public class KNNQuery extends Query {
         final String indexName,
         final Query filterQuery,
         final BitSetProducer parentsFilter,
-        final VectorDataType vectorDataType
+        final VectorDataType vectorDataType,
+        final RescoreContext rescoreContext
     ) {
         this.field = field;
         this.queryVector = queryVector;
@@ -103,6 +106,7 @@ public class KNNQuery extends Query {
         this.filterQuery = filterQuery;
         this.parentsFilter = parentsFilter;
         this.vectorDataType = vectorDataType;
+        this.rescoreContext = rescoreContext;
     }
 
     /**
@@ -114,7 +118,7 @@ public class KNNQuery extends Query {
      * @param parentsFilter parent filter
      */
     public KNNQuery(String field, float[] queryVector, String indexName, BitSetProducer parentsFilter) {
-        this(field, queryVector, null, 0, indexName, null, parentsFilter, VectorDataType.FLOAT);
+        this(field, queryVector, null, 0, indexName, null, parentsFilter, VectorDataType.FLOAT, null);
     }
 
     /**
--- a/src/main/java/org/opensearch/knn/index/query/KNNQueryFactory.java
+++ b/src/main/java/org/opensearch/knn/index/query/KNNQueryFactory.java
@@ -5,7 +5,6 @@
 
 package org.opensearch.knn.index.query;
 
-import com.google.common.annotations.VisibleForTesting;
 import lombok.extern.log4j.Log4j2;
 import org.apache.lucene.search.KnnByteVectorQuery;
 import org.apache.lucene.search.KnnFloatVectorQuery;
@@ -17,6 +16,7 @@ import org.opensearch.index.query.QueryShardContext;
 import org.opensearch.knn.index.VectorDataType;
 import org.opensearch.knn.index.engine.KNNEngine;
 import org.opensearch.knn.index.query.nativelib.NativeEngineKnnVectorQuery;
+import org.opensearch.knn.index.query.rescore.RescoreContext;
 
 import java.util.Locale;
 import java.util.Map;
@@ -31,39 +31,6 @@ import static org.opensearch.knn.index.VectorDataType.SUPPORTED_VECTOR_DATA_TYPE
  */
 @Log4j2
 public class KNNQueryFactory extends BaseQueryFactory {
-
-    /**
-     * Note. This method should be used only for test.
-     * Should use {@link #create(CreateQueryRequest)} instead.
-     *
-     * Creates a Lucene query for a particular engine.
-     *
-     * @param knnEngine Engine to create the query for
-     * @param indexName Name of the OpenSearch index that is being queried
-     * @param fieldName Name of the field in the OpenSearch index that will be queried
-     * @param vector The query vector to get the nearest neighbors for
-     * @param k the number of nearest neighbors to return
-     * @return Lucene Query
-     */
-    @VisibleForTesting
-    public static Query create(
-        KNNEngine knnEngine,
-        String indexName,
-        String fieldName,
-        float[] vector,
-        int k,
-        VectorDataType vectorDataType
-    ) {
-        final CreateQueryRequest createQueryRequest = CreateQueryRequest.builder()
-            .knnEngine(knnEngine)
-            .indexName(indexName)
-            .fieldName(fieldName)
-            .vector(vector)
-            .vectorDataType(vectorDataType)
-            .k(k)
-            .build();
-        return create(createQueryRequest);
-    }
 
     /**
      * Creates a Lucene query for a particular engine.
@@ -81,6 +48,7 @@ public class KNNQueryFactory extends BaseQueryFactory {
         final VectorDataType vectorDataType = createQueryRequest.getVectorDataType();
         final Query filterQuery = getFilterQuery(createQueryRequest);
         final Map<String, ?> methodParameters = createQueryRequest.getMethodParameters();
+        final RescoreContext rescoreContext = createQueryRequest.getRescoreContext().orElse(null);
 
         BitSetProducer parentFilter = null;
         if (createQueryRequest.getContext().isPresent()) {
@@ -112,6 +80,7 @@ public class KNNQueryFactory extends BaseQueryFactory {
                         .methodParameters(methodParameters)
                         .filterQuery(validatedFilterQuery)
                         .vectorDataType(vectorDataType)
+                        .rescoreContext(rescoreContext)
                         .build();
                     break;
                 default:
@@ -124,6 +93,7 @@ public class KNNQueryFactory extends BaseQueryFactory {
                         .methodParameters(methodParameters)
                         .filterQuery(validatedFilterQuery)
                         .vectorDataType(vectorDataType)
+                        .rescoreContext(rescoreContext)
                         .build();
             }
             return isKnnQueryRewriteEnabled() ? new NativeEngineKnnVectorQuery(knnQuery) : knnQuery;
--- a/src/main/java/org/opensearch/knn/index/query/KNNWeight.java
+++ b/src/main/java/org/opensearch/knn/index/query/KNNWeight.java
@@ -7,15 +7,12 @@ package org.opensearch.knn.index.query;
 
 import com.google.common.annotations.VisibleForTesting;
 import lombok.extern.log4j.Log4j2;
-import org.apache.commons.lang.StringUtils;
 import org.apache.lucene.index.FieldInfo;
 import org.apache.lucene.index.LeafReaderContext;
 import org.apache.lucene.index.SegmentReader;
 import org.apache.lucene.search.DocIdSetIterator;
 import org.apache.lucene.search.Explanation;
 import org.apache.lucene.search.FilteredDocIdSetIterator;
-import org.apache.lucene.search.HitQueue;
-import org.apache.lucene.search.ScoreDoc;
 import org.apache.lucene.search.Scorer;
 import org.apache.lucene.search.Weight;
 import org.apache.lucene.store.FSDirectory;
@@ -23,7 +20,6 @@ import org.apache.lucene.store.FilterDirectory;
 import org.apache.lucene.util.BitSet;
 import org.apache.lucene.util.BitSetIterator;
 import org.apache.lucene.util.Bits;
-import org.apache.lucene.util.DocIdSetBuilder;
 import org.apache.lucene.util.FixedBitSet;
 import org.opensearch.common.io.PathUtils;
 import org.opensearch.common.lucene.Lucene;
@@ -37,16 +33,7 @@ import org.opensearch.knn.index.memory.NativeMemoryAllocation;
 import org.opensearch.knn.index.memory.NativeMemoryCacheManager;
 import org.opensearch.knn.index.memory.NativeMemoryEntryContext;
 import org.opensearch.knn.index.memory.NativeMemoryLoadStrategy;
-import org.opensearch.knn.index.query.filtered.FilteredIdsKNNByteIterator;
-import org.opensearch.knn.index.query.filtered.FilteredIdsKNNIterator;
-import org.opensearch.knn.index.query.filtered.KNNIterator;
-import org.opensearch.knn.index.query.filtered.NestedFilteredIdsKNNByteIterator;
-import org.opensearch.knn.index.query.filtered.NestedFilteredIdsKNNIterator;
 import org.opensearch.knn.index.engine.KNNEngine;
-import org.opensearch.knn.index.vectorvalues.KNNBinaryVectorValues;
-import org.opensearch.knn.index.vectorvalues.KNNFloatVectorValues;
-import org.opensearch.knn.index.vectorvalues.KNNVectorValues;
-import org.opensearch.knn.index.vectorvalues.KNNVectorValuesFactory;
 import org.opensearch.knn.indices.ModelDao;
 import org.opensearch.knn.indices.ModelMetadata;
 import org.opensearch.knn.indices.ModelUtil;
@@ -58,9 +45,7 @@ import java.nio.file.Path;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.Comparator;
-import java.util.HashMap;
 import java.util.List;
-import java.util.Locale;
 import java.util.Map;
 import java.util.concurrent.ExecutionException;
 import java.util.stream.Collectors;
@@ -84,6 +69,9 @@ public class KNNWeight extends Weight {
 
     private final NativeMemoryCacheManager nativeMemoryCacheManager;
     private final Weight filterWeight;
+    private final ExactSearcher exactSearcher;
+
+    private static ExactSearcher DEFAULT_EXACT_SEARCHER;
 
     public KNNWeight(KNNQuery query, float boost) {
         super(query);
@@ -91,6 +79,7 @@ public class KNNWeight extends Weight {
         this.boost = boost;
         this.nativeMemoryCacheManager = NativeMemoryCacheManager.getInstance();
         this.filterWeight = null;
+        this.exactSearcher = DEFAULT_EXACT_SEARCHER;
     }
 
     public KNNWeight(KNNQuery query, float boost, Weight filterWeight) {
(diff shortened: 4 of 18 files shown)


src/main/java/org/opensearch/knn/index/query/rescore/RescoreSearcher.java
```diff
@@ -0,0 +1,96 @@
...
+        KNNWeight knnWeight,
+        List<Map<Integer, Float>> perLeafResults,
+        int k
+    ) throws IOException {
+        List<Callable<TopDocs>> topDocTasks = new ArrayList<>(leafReaderContexts.size());
+        for (int i = 0; i < perLeafResults.size(); i++) {
+            BitSet convertedBitSet = resultUtil.resultMapToMatchBitSet(perLeafResults.get(i));
+            LeafReaderContext leafReaderContext = leafReaderContexts.get(i);
+            topDocTasks.add(
+                () -> resultUtil.resultMapToTopDocs(
```
navneet1v, line 88: this conversion of mapping docs to bitset and then bitset to iterator has potential to incur latency in case of large number segments and high value of first pass of K. 

If you see these bitsets are getting converted to iterators only at the end here: https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/query/filtered/FilteredIdsKNNIterator.java#L39

So, I see there is a potential to reduce all this extra conversion.

Another idea could be to write our own implementation of DISI which underline uses a Map/List<Integer> for iteration.

src/main/java/org/opensearch/knn/index/query/rescore/RescoreSearcher.java
```diff
@@ -0,0 +1,96 @@
...
+     */
+    public TopDocs[] search(
+        final IndexSearcher indexSearcher,
+        List<LeafReaderContext> leafReaderContexts,
+        KNNWeight knnWeight,
+        int firstPassK,
+        int finalK
+    ) throws IOException {
+        List<Map<Integer, Float>> perLeafResults = firstPassSearch(indexSearcher, leafReaderContexts, knnWeight, firstPassK);
+        return rescoreResults(indexSearcher, leafReaderContexts, knnWeight, perLeafResults, finalK);
```
navneet1v, line 47: should we add some debug logs which logs the time taken to run these 2 functions. I know there is no good way to record time for these queries. But if we can add some debug logs they might come handy later on.

src/main/java/org/opensearch/knn/index/query/ResultUtil.java
```diff
@@ -0,0 +1,118 @@
...
+    public static final ResultUtil DEFAULT = new ResultUtil();
+
+    /**
+     * Reduce the results to only include the top k results across all leaf results
+     *
+     * @param perLeafResults Results from the list
+     * @param k the number of results across all leaf results to return
+     * @return Reduced results across all leaf results
+     */
+    public List<Map<Integer, Float>> reduceToTopK(List<Map<Integer, Float>> perLeafResults, int k) {
```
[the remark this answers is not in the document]
jmazanec15: Will discuss in https://github.com/opensearch-project/k-NN/pull/1984#discussion_r1721264017

src/main/java/org/opensearch/knn/index/query/ResultUtil.java
```diff
@@ -0,0 +1,118 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.List;
+import java.util.Map;
+
+public final class ResultUtil {
+
+    public static final ResultUtil DEFAULT = new ResultUtil();
```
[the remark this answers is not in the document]
jmazanec15: main reasoning was to make testing easier via mocking.

src/main/java/org/opensearch/knn/index/codec/util/KNNCodecUtil.java
```diff
@@ -102,4 +111,20 @@ public static long getTotalLiveDocsCount(final BinaryDocValues binaryDocValues)
         }
         return totalLiveDocs;
     }
+
+    public static SpaceType getSpaceType(final ModelDao modelDao, final FieldInfo fieldInfo) {
```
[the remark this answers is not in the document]
jmazanec15: yeah thatd make more sense.

src/main/java/org/opensearch/knn/index/query/KNNQueryFactory.java
```diff
@@ -112,6 +80,7 @@ public static Query create(CreateQueryRequest createQueryRequest) {
                         .methodParameters(methodParameters)
                         .filterQuery(validatedFilterQuery)
                         .vectorDataType(vectorDataType)
+                        .rescoreContext(rescoreContext)
```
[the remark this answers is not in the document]
jmazanec15: Its optionally null. Not sure I understand question

src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java
```diff
@@ -49,11 +50,16 @@ public Query rewrite(final IndexSearcher indexSearcher) throws IOException {
...
-        for (LeafReaderContext leafReaderContext : leafReaderContexts) {
-            tasks.add(() -> searchLeaf(leafReaderContext, knnWeight));
+        TopDocs[] perLeafResults;
+        RescoreContext rescoreContext = knnQuery.getRescoreContext();
+        if (rescoreContext == null) {
+            perLeafResults = doSearch(indexSearcher, leafReaderContexts, knnWeight);
+        } else {
+            int firstPassK = rescoreContext.getFirstPassK(knnQuery.getK());
+            perLeafResults = rescoreContext.getRescoreSearcher()
+                .search(indexSearcher, leafReaderContexts, knnWeight, firstPassK, knnQuery.getK());
```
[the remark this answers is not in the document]
jmazanec15: sure, I can change it up

src/main/java/org/opensearch/knn/index/query/rescore/RescoreSearcher.java
```diff
@@ -0,0 +1,96 @@
...
+        Map<Integer, Float> leafDocScores = queryWeight.searchLeaf(ctx);
+        if (leafDocScores.isEmpty()) {
+            return leafDocScores;
+        }
+
+        Bits liveDocs = ctx.reader().getLiveDocs();
+        return leafDocScores.entrySet()
+            .stream()
+            .filter(entry -> liveDocs == null || liveDocs.get(entry.getKey()))
+            .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
```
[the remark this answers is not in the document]
jmazanec15: > Also, why we are not adding this in the function: queryWeight.searchLeaf(ctx) at the end ?

I didnt make the change in k-NN Weight, because it would change the behavior of how we currently support non-rescored search. I figured itd be best in this change to avoid chaning behavior of that path just to reduce blast radius.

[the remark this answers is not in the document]
jmazanec15: I can add comment

src/main/java/org/opensearch/knn/index/query/ResultUtil.java
```diff
@@ -0,0 +1,118 @@
...
+                    continue;
+                }
+
+                if (score > topKScores[k - 1]) {
+                    topKScores[k - 1] = score;
+                    Arrays.sort(topKScores, Comparator.comparingDouble(Float::doubleValue).reversed());
+                }
+                count++;
+            }
+        }
```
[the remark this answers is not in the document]
jmazanec15: Ah that simplifies a lot. Will update.

src/main/java/org/opensearch/knn/index/query/ResultUtil.java
```diff
@@ -0,0 +1,118 @@
...
+
+        // Reduce the results based on min competitive score
+        // List<Map<Integer, Float>> perLeafResultsCopy = perLeafResults.stream().map(HashMap::new).collect(Collectors.toList());
+        perLeafResults.forEach(results -> results.entrySet().removeIf(entry -> entry.getValue() < topKScores[k - 1]));
```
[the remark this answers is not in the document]
jmazanec15: > I see that you are removing the docs with score < minCompetitive score. But what will happen if more than 1 doc has == min competitive score, will it create any problem?

In that case, it will return > k results and everything above the min competitive score. This wont be an issue for later functionality. 

> Also, I see that we are iterating over the perLeafResults twice, see if we can reduce it to 1 to save latency.

Problem is, we need first iteration to find the min score and then the next iteration to filter based on min score. There's probably a more complex way to do this that would give better overall time, but I dont see it being very expensive.

src/main/java/org/opensearch/knn/index/query/rescore/RescoreSearcher.java
jmazanec15: It seems it would be easiest to switch BitSet to DocIdSetIterator in the FilteredIdsKNNIterator (https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/query/filtered/FilteredIdsKNNIterator.java#L33) and then create a wrapper of the DocIDSetIterator around the Map of results to avoid the conversion from map to bitset. I can make this change - but before I do, want to get your thoughts?

src/main/java/org/opensearch/knn/index/query/rescore/RescoreSearcher.java
jmazanec15: Debug logs are good. Which method did you wrap with a timer before? I can model it based off of that

src/main/java/org/opensearch/knn/index/query/rescore/RescoreSearcher.java
jmazanec15: or, I can take for another review - what do you think?

navneet1v: Lets do it in separate review. But it should be done before the release of disk based vector search feature.

src/main/java/org/opensearch/knn/index/query/KNNQueryFactory.java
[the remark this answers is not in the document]
navneet1v: what I mean to say here is: CreateQueryRequest has the rescoreContext as Optional<RescoreContext>, but in KNNQuery object it is not Optional<RescoreContext>. So was thinking why we are doing this? We should keep the RescoreContext wrapped in optional for both the POJO.

src/main/java/org/opensearch/knn/index/query/ResultUtil.java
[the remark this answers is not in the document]
navneet1v: I see, I thought we might have another implementation of ResultUtil in future thats why it is called as default. But what we are trying to do here follow a singleton pattern and ease our testing. If that is the case, just change variable name from DEFAULT to INSTANCE.

I am on split opinion here on using such thing. The way I see this is a util class and should not be mocked. If we are mocking it then there are things in this utility which are doing much more than some transformations. 

Also, if you really want to mock it you can use MockStatic functionality. I see that is also pretty handy.

src/main/java/org/opensearch/knn/index/query/rescore/RescoreSearcher.java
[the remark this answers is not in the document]
navneet1v: Please add a comment. I think after adding deletedFilters logic this code should go away. Ref: https://github.com/opensearch-project/k-NN/issues/1491

src/main/java/org/opensearch/knn/index/query/ResultUtil.java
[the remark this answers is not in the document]
navneet1v: > In that case, it will return > k results and everything above the min competitive score. This wont be an issue for later functionality.

Good to know. :) 


> > Also, I see that we are iterating over the perLeafResults twice, see if we can reduce it to 1 to save latency.

> Problem is, we need first iteration to find the min score and then the next iteration to filter based on min score. There's probably a more complex way to do this that would give better overall time, but I don't see it being very expensive.

I think if you keep more information related to segment, doc and score in the Heap You can recreate the perLeafResults easily. But its still an optimization that can be taken as the code evolves. I would recommend adding a comment here saying it can be optimized and we will do it.

These can become expensive once we have more segments and large vales of K.

src/main/java/org/opensearch/knn/index/query/rescore/RescoreSearcher.java
navneet1v: I think I wrap functions in a feature branch and not in the main branch. So the code might not be a good. :P 

I had this code: https://github.com/navneet1v/k-NN/blob/perf-deep-dive/src/main/java/org/opensearch/knn/jni/JNIService.java#L110-L117 long time back. But not a good way I think.

src/main/java/org/opensearch/knn/index/query/rescore/RescoreSearcher.java
```diff
@@ -0,0 +1,98 @@
...
+    private List<Map<Integer, Float>> firstPassSearch(
+        final IndexSearcher indexSearcher,
+        List<LeafReaderContext> leafReaderContexts,
+        KNNWeight knnWeight,
+        int k
+    ) throws IOException {
+        List<Callable<Map<Integer, Float>>> tasks = new ArrayList<>(leafReaderContexts.size());
+        for (LeafReaderContext leafReaderContext : leafReaderContexts) {
+            tasks.add(() -> firstPassSearchLeaf(leafReaderContext, knnWeight));
+        }
```
shatejas, line 62: For first pass, why can't we reuse the existing logic in [NativeEngineKnnVectorQuery.java](https://github.com/opensearch-project/k-NN/pull/1984/files#diff-9cfe412357ba56b3ef216427d491fc653535686a760e8ba19ea1aa00fc0e0338). we just have to manipulate the k or is there some performance benefits which I missed?

One thing you will have to take care is you need to subtract docBase per leaf in rescoreResults 
This causes duplication adding on to the maintainance

```
// current knnweight will manipulate k if rescorercontext is in query
List<Callable<TopDocs>> tasks = new ArrayList<>(leafReaderContexts.size());
            tasks.add(() -> searchLeaf(leafReaderContext, knnWeight));
        }
        TopDocs[] perLeafResults = indexSearcher.getTaskExecutor().invokeAll(tasks).toArray(TopDocs[]::new);
        // TopDocs.merge requires perLeafResults to be sorted in descending order.
        TopDocs topK = TopDocs.merge(knnQuery.getK(), perLeafResults);
       TopDocs rescoredResults = rescore context == null? tokK : rescoresearcher.rescoreResults(...)?

  
```

src/main/java/org/opensearch/knn/index/query/rescore/RescoreSearcher.java
jmazanec15: I thought you added an annotation somewhere. Might be mistaken.

navneet1v: I created the GH issue for the work. But I didn't pick that up. 

Ref: https://github.com/opensearch-project/k-NN/issues/1594

src/main/java/org/opensearch/knn/index/query/rescore/RescoreSearcher.java
shatejas: I think I figured it, we need access to doc Ids of per leaf results so we get the flat values

src/main/java/org/opensearch/knn/index/query/rescore/RescoreSearcher.java
```diff
@@ -0,0 +1,101 @@
...
+    private TopDocs[] rescoreResults(
+        final IndexSearcher indexSearcher,
+        List<LeafReaderContext> leafReaderContexts,
+        KNNWeight knnWeight,
+        List<Map<Integer, Float>> perLeafResults,
+        int k
+    ) throws IOException {
+        List<Callable<TopDocs>> topDocTasks = new ArrayList<>(leafReaderContexts.size());
+        for (int i = 0; i < perLeafResults.size(); i++) {
+            BitSet convertedBitSet = resultUtil.resultMapToMatchBitSet(perLeafResults.get(i));
```
shatejas, line 90: Any reason this is not inside the callable? Verify once but I think there is a loop inside this when creating bitset, throwing inside the callable will help with latencies

src/main/java/org/opensearch/knn/index/query/rescore/RescoreSearcher.java
```diff
@@ -0,0 +1,101 @@
...
+            LeafReaderContext leafReaderContext = leafReaderContexts.get(i);
+            topDocTasks.add(
+                () -> resultUtil.resultMapToTopDocs(
+                    knnWeight.exactSearch(leafReaderContext, convertedBitSet, false, k),
```
shatejas, line 94: ExactSearch doesn't seem to be happening in parallel, intended?

src/main/java/org/opensearch/knn/index/query/ResultUtil.java
```diff
@@ -0,0 +1,116 @@
...
+     */
+    public List<Map<Integer, Float>> reduceToTopK(List<Map<Integer, Float>> perLeafResults, int k) {
+        // Iterate over all scores to get min competitive score
+        PriorityQueue<Float> topKScoresQueue = new PriorityQueue<>(k);
+        for (int i = 0; i < k; i++) {
+            topKScoresQueue.add(-Float.MAX_VALUE);
+        }
+
+        for (Map<Integer, Float> perLeafResult : perLeafResults) {
+            for (Float score : perLeafResult.values()) {
```
shatejas, line 44: This seems expensive, is there a possibility that we can maintain a global priority queue of with a pair of id in a  and score and keep adding the results in `firstPassSearchLeaf` when we are looping through for filtering deleted docs? At the end of parallel execution we should be able to just poll the priority queue and rescore. 

Its a high level thought and in a way inspired from Lucene's `TopKnnCollectorManager`. There is a slight modification though where we need doc ids (with leaf index I suppose). Worth time boxing it as we can reduce latencies significantly when concurrent search is on

shatejas (commented):
Overall I think there is a potential to merge firstPass to reduce repeated search logic
Here is the psuedo code, see if it works out

```
public Query rewrite(final IndexSearcher indexSearcher) throws IOException {
        final IndexReader reader = indexSearcher.getIndexReader();
        final KNNWeight knnWeight = (KNNWeight) knnQuery.createWeight(indexSearcher, ScoreMode.COMPLETE, 1);

        List<Callable<Map<Integer, Float>>> tasks = new ArrayList<>(leafReaderContexts.size());
            tasks.add(() -> searchLeaf(leafReaderContext, knnWeight));
        }
        List<Map<Integer, Float>> perLeafResults = indexSearcher.getTaskExecutor().invokeAll(tasks);
        Map<Integer, Float> topk = rescoreContext ? reduceToTopK : { reduceTooversampledK(), rescore(), topk  }
                
        if (topK.scoreDocs.length == 0) {
            return new MatchNoDocsQuery();
        }
        return createRewrittenQuery(reader, topK);
    }
```

src/main/java/org/opensearch/knn/index/query/KNNQueryFactory.java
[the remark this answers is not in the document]
jmazanec15: Ah the getter is Optional but not the member. I prefer to use Optionals to wrap potentially null member variables as opposed to making them member variables. Intellij gives the following suggestion:
```
'Optional<QuantizationConfig>' used as type for field 'quantizationConfig' 
 Inspection info: Reports any cases in which java. util. Optional<T>, java. util. OptionalDouble, java. util. OptionalInt, java. util. OptionalLong, or com. google. common. base. Optional are used as types for fields or parameters.
Optional was designed to provide a limited mechanism for library method return types in which a clear way to represent "no result" was needed.
```

src/main/java/org/opensearch/knn/index/query/rescore/RescoreContext.java
```diff
@@ -20,8 +21,14 @@ public final class RescoreContext {
...
     public static final float MIN_OVERSAMPLE_FACTOR = 0.0f;
 
+    public static final int MAX_FIRST_PASS_RESULTS = 10000;
+
+    private static final RescoreSearcher DEFAULT_RESCORE_SEARCHER = new RescoreSearcher(ResultUtil.DEFAULT);
+
     @Builder.Default
     private float oversampleFactor = DEFAULT_OVERSAMPLE_FACTOR;
+    @Builder.Default
+    private RescoreSearcher rescoreSearcher = DEFAULT_RESCORE_SEARCHER;
```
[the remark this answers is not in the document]
jmazanec15: Sure, I can remove and put in the NativeQuery where it needs to be called. That makes sense.

src/main/java/org/opensearch/knn/index/query/rescore/RescoreSearcher.java
```diff
@@ -0,0 +1,98 @@
...
+    private Map<Integer, Float> firstPassSearchLeaf(LeafReaderContext ctx, KNNWeight queryWeight) throws IOException {
+        Map<Integer, Float> leafDocScores = queryWeight.searchLeaf(ctx);
+        if (leafDocScores.isEmpty()) {
+            return leafDocScores;
+        }
+
+        Bits liveDocs = ctx.reader().getLiveDocs();
+        // TODO: This filters out the live docs from the knnWeight. Once knnWeight.searchLeaf returns results with
+        // deleted docs filtered out, this can be removed.
+        return leafDocScores.entrySet()
```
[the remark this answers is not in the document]
jmazanec15: Is alternative, just a for loop over entries and removing?

src/main/java/org/opensearch/knn/index/query/rescore/RescoreSearcher.java
jmazanec15: I see that makes sense to have inside.

src/main/java/org/opensearch/knn/index/query/ResultUtil.java
jmazanec15: @navneet1v had similar idea. I think this is probably a good optimization to make, but will take in a future PR.

src/main/java/org/opensearch/knn/index/query/ResultUtil.java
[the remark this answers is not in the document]
jmazanec15: Sure, I will switch to static utility class

src/main/java/org/opensearch/knn/index/codec/util/KNNCodecUtil.java
```diff
@@ -102,4 +111,20 @@ public static long getTotalLiveDocsCount(final BinaryDocValues binaryDocValues)
...
+        final String spaceTypeString = fieldInfo.getAttribute(SPACE_TYPE);
+        if (StringUtils.isNotEmpty(spaceTypeString)) {
+            return SpaceType.getSpace(spaceTypeString);
+        }
+
+        final String modelId = fieldInfo.getAttribute(MODEL_ID);
+        if (StringUtils.isNotEmpty(modelId)) {
```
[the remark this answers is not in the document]
jmazanec15: I held off on adding this here for now just because ModelUtil gets the model metadata from the system index instead of the cluster state. I need to fix this, but will leave it for later.

jmazanec15:
> Overall I think there is a potential to merge firstPass to reduce repeated search logic
Here is the psuedo code, see if it works out

@shatejas I didnt want to have to modify the k parameter in the k-NN query, so i kept them as separate.

shatejas:
> > Overall I think there is a potential to merge firstPass to reduce repeated search logic
> > Here is the psuedo code, see if it works out
> 
> @shatejas I didnt want to have to modify the k parameter in the k-NN query, so i kept them as separate.

Can't you create a local k parameter either inside knnweight or nativequery?

src/main/java/org/opensearch/knn/index/query/KNNQuery.java
```diff
@@ -38,13 +39,12 @@ public class KNNQuery extends Query {
...
     private final String field;
     private final float[] queryVector;
-    @Getter
     private final byte[] byteQueryVector;
     private int k;
     private Map<String, ?> methodParameters;
     private final String indexName;
-    @Getter
     private final VectorDataType vectorDataType;
+    private final RescoreContext rescoreContext;
```
Vikasht34, line 47: In our code, the RescoreContext is wrapped in an Optional in some places but not others. It's generally a good idea to be consistent in how you use Optional. If you're using it to represent a potentially absent value

public KNNQuery(...., Optional<RescoreContext> rescoreContext) {
    this.rescoreContext = rescoreContext;
}

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -82,13 +68,17 @@ public class KNNWeight extends Weight {
 
     private final NativeMemoryCacheManager nativeMemoryCacheManager;
     private final Weight filterWeight;
+    private final ExactSearcher exactSearcher;
+
+    private static ExactSearcher DEFAULT_EXACT_SEARCHER;
```
Vikasht34, line 73: This is redundant , we are having two variable for ExactSearcher just to save DI , why should we not considering lazy intialization of ExactSearcher  by Singleton and get Instance here ..

Vikasht34 (commented):
The searchLeaf method seems to have different implementations for different scenarios (like with or without rescoring) and it's all over places like one in KNN Weight and other on Exacat Search , which internally calls exact search class search leaf only , we should consider unifiying this expreince , either by using startegy or use of  function interface and have one class which has searchleaf method where we can look into.

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
jmazanec15: I can change to that. I guess testing and the fact that ExactSearcher has dependency on ModelDao

jmazanec15: I might defer changing this to later after I make a change to the ModelUtil to get modelMetadata properly

jmazanec15:
> The searchLeaf method seems to have different implementations for different scenarios (like with or without rescoring) and it's all over places like one in KNN Weight and other on Exacat Search , which internally calls exact search class search leaf only , we should consider unifiying this expreince , either by using startegy or use of function interface and have one class which has searchleaf method where we can look into.

@Vikasht34 I think the main goal for using KNNWeight calls in NativeKNNVectorQuery was to avoid too much change around that. I think we can change this up as well, but will do in future review, just so that I can move on to working on mode and compression. Does that sound good?

src/main/java/org/opensearch/knn/common/FieldInfoExtractor.java
```diff
@@ -34,4 +41,27 @@ public static VectorDataType extractVectorDataType(final FieldInfo fieldInfo) {
...
+     */
+    public static SpaceType getSpaceType(final ModelDao modelDao, final FieldInfo fieldInfo) {
+        final String spaceTypeString = fieldInfo.getAttribute(SPACE_TYPE);
+        if (StringUtils.isNotEmpty(spaceTypeString)) {
+            return SpaceType.getSpace(spaceTypeString);
+        }
+
+        final String modelId = fieldInfo.getAttribute(MODEL_ID);
+        if (StringUtils.isNotEmpty(modelId)) {
+            ModelMetadata modelMetadata = modelDao.getMetadata(modelId);
```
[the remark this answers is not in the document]
jmazanec15: Updating modelMetadata - space type wont be null if model metadata isnt so it should be good

src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java
```diff
@@ -49,19 +50,63 @@ public Query rewrite(final IndexSearcher indexSearcher) throws IOException {
...
+            perLeafResults = doRescore(indexSearcher, leafReaderContexts, knnWeight, perLeafResults, finalK);
+        }
+        ResultUtil.reduceToTopK(perLeafResults, finalK);
+        TopDocs[] topDocs = new TopDocs[perLeafResults.size()];
+        for (int i = 0; i < perLeafResults.size(); i++) {
+            topDocs[i] = ResultUtil.resultMapToTopDocs(perLeafResults.get(i), leafReaderContexts.get(i).docBase);
         }
-        TopDocs[] perLeafResults = indexSearcher.getTaskExecutor().invokeAll(tasks).toArray(TopDocs[]::new);
+
         // TopDocs.merge requires perLeafResults to be sorted in descending order.
```
ryanbogan, line 70: Nit: perLeafResults variable was removed

ryanbogan (approved):
One minor comment, but otherwise LGTM

ryanbogan (dismissed):

ryanbogan (approved):

shatejas (approved):
Lgtm

src/main/java/org/opensearch/knn/index/query/ResultUtil.java
```diff
@@ -0,0 +1,112 @@
...
+                if (topKMinQueue.peek() != null && score > topKMinQueue.peek()) {
+                    topKMinQueue.poll();
+                    topKMinQueue.add(score);
+                }
+            }
+        }
+
+        // Reduce the results based on min competitive score
+        float minScore = topKMinQueue.peek() == null ? -Float.MAX_VALUE : topKMinQueue.peek();
```
navneet1v, line 50: you might have to put a loop here to keep on getting elements from priority queue till the score is -FLOAT.MAX_VALUE

navneet1v: I think this looks good. We don't need a loop here. I was thinking of some other case.

But we can do this, if minScore == Float.MAX we can return directly perLeafResults and we don't need to iterate over perLeafResults.

navneet1v (dismissed):

src/main/java/org/opensearch/knn/index/query/ResultUtil.java
jmazanec15: sure, I added a count just because if minScore == -Float.MAX check wont be true if there are exactly k results

navneet1v (approved):

ryanbogan (approved):

PR #1994, opened 2024-08-21
Add support for qframework from interface

jmazanec15:
### Description
Adds support for parsing qframework for faiss in the interface for IVF and HNSW. Does not yet integrate them with the framework.

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/engine/qframe/QuantizationConfigParser.java
+++ b/src/main/java/org/opensearch/knn/index/engine/qframe/QuantizationConfigParser.java
@@ -1,0 +1,85 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.knn.index.engine.qframe;
+
+import org.apache.lucene.analysis.util.CSVUtil;
+import org.opensearch.knn.index.engine.faiss.QFrameBitEncoder;
+import org.opensearch.knn.quantization.enums.ScalarQuantizationType;
+
+import java.util.Locale;
+
+/**
+ * Parse util for quantization config
+ */
+public class QuantizationConfigParser {
+
+    public static final String SEPARATOR = "=";
+    public static final String TYPE_NAME = "type";
+    public static final String BINARY_TYPE = QFrameBitEncoder.NAME;
+    public static final String BIT_COUNT_NAME = QFrameBitEncoder.BITCOUNT_PARAM;
+
+    /**
+     * Parse quantization config to csv format
+     * Example: type=binary,bits=2
+     * @param quantizationConfig Quantization config
+     * @return Csv format of quantization config
+     */
+    public static String toCsv(QuantizationConfig quantizationConfig) {
+        if (quantizationConfig == null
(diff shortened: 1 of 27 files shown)


src/main/java/org/opensearch/knn/index/engine/qframe/QuantizationConfigParser.java
```diff
@@ -0,0 +1,85 @@
...
+import org.apache.lucene.analysis.util.CSVUtil;
+import org.opensearch.knn.index.engine.faiss.QFrameBitEncoder;
+import org.opensearch.knn.quantization.enums.ScalarQuantizationType;
+
+import java.util.Locale;
+
+/**
+ * Parse util for quantization config
+ */
+public class QuantizationConfigParser {
```
heemin32, line 17: I think it would be better to have all these methods inside `QuantizationConfig` class itself?
```
// Creation from csv
QuantizationConfig config = QuantizationConfig.Builder.newInstance().addCsv(csv).build();
// or
QuantizationConfig config = new QuantizationConfig(csv);

// Back to csv
String configInString = config.toCsv();
```

jmazanec15: I decided to keep QuantizationConfig simple and have other class, as @shatejas had suggested in https://github.com/opensearch-project/k-NN/pull/1969#discussion_r1718900321 and in the field mapping parser changes. Im good with either, but dont see a major benefit for one vs other.

heemin32: This is more like we are having serialization and deserialization method inside KNNBuilderQuery and few other requests classes rather than having a separate static methods.

Those methods are tightly coupled with what field we have in the class. Whenever you make change in those field, you have to modify these methods. Having them in multiple classes is like broadening the scope of changes from one class to multiple classes whenever there is an update.

shatejas: In general serialization and deserialization can be expensive, this code doesn't seem to super expensive. Having said that, the question would be, what part of code base are we okay having access to this especially if we are not using java serialization or something with requires this to be in the POJO. Adding it in POJO itself, gives an easy access. on contrary client has to be mindful to use this parser and is not an easy access also keeping the responsibility different. 

These are my two cents, its trivial enough and I am okay if maintainers feel okay about adding it in pojo

heemin32: Fair enough. Will just leave it to the author of the code to decide.
However, I still believe it would be natural to have them inside Config class. Not this case right now but suppose if QuantizationConfig has child classes for example, then this approach might get complicated.

heemin32 (approved):

junqiu-lei (approved):

src/main/java/org/opensearch/knn/index/engine/qframe/QuantizationConfigParser.java
jmazanec15: Ill keep as is for now.

PR #1996, opened 2024-08-21
Integration of Quantization Framework for Binary Quantization with Indexing Flow

Vikasht34:
### Description
This PR introduces changes to the indexing flow by integrating it with the Quantization Framework, specifically adding support for Binary Quantization. Additionally, the PR includes updates to the Quantization State to provide accurate RAM usage estimates for Quantization State instances. 

### Related Issues

### Check List


Vikasht34:
This failure in Integ tests for Window is un related to this PR 

```
REPRODUCE WITH: gradlew ':test' --tests "org.opensearch.knn.indices.ModelDaoTests.testCreate" -Dtests.seed=4AD3C3C3F1DEFD7A -Dtests.security.manager=false -Dtests.locale=teo-UG -Dtests.timezone=America/Eirunepe -Druntime.java=21
org.opensearch.knn.indices.ModelDaoTests > testCreate FAILED
    java.lang.AssertionError
        at __randomizedtesting.SeedInfo.seed([4AD3C3C3F1DEFD7A:759903A4454CD0FB]:0)
        at org.junit.Assert.fail(Assert.java:87)
        at org.junit.Assert.assertTrue(Assert.java:42)
        at org.junit.Assert.assertTrue(Assert.java:53)
        at org.opensearch.knn.indices.ModelDaoTests.testCreate(ModelDaoTests.java:110)
```

src/main/java/org/opensearch/knn/index/quantizationIntegration/QuantizationIntegration.java
```diff
@@ -0,0 +1,122 @@
...
+
+/**
+ * A singleton class responsible for handling the quantization process, including training a quantizer
+ * and applying quantization to vectors. This class is designed to be thread-safe.
+ *
+ * @param <T> The type of the input vectors to be quantized.
+ * @param <R> The type of the quantized output vectors.
+ */
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+public class QuantizationIntegration<T, R> {
```
jmazanec15, line 31: Can we call it something other than integration? Doesnt seem quite right - for instance, there are QuantizationIntegrationTests that are unit tests. A couple ideas:
1. QuantizationService
2. QuantizationHandler
3. QuantizationManager

src/main/java/org/opensearch/knn/quantization/models/quantizationState/MultiBitScalarQuantizationState.java
```diff
@@ -124,4 +125,40 @@ public byte[] toByteArray() throws IOException {
...
+    public int getBytesStoredPerVector() {
+        // Calculate the number of bytes required for multi-bit quantization
+        return thresholds.length * thresholds[0].length;
+    }
+
+    @Override
+    public int getDimensionForIndexing() {
+        // For multi-bit quantization, the dimension for indexing is the number of rows * columns in the thresholds array.
+        // Where number of column reprensents Dimesion of Original vector and number of rows equals to number of bits
+        return thresholds.length * thresholds[0].length;
```
jmazanec15, line 144: Do we need to ensure threshold length > 0?

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriter.java
```diff
@@ -146,4 +132,103 @@ public long ramBytesUsed() {
...
+     * @param <T>            The type of vectors being processed.
+     * @return The {@link KNNVectorValues} associated with the field during the merge.
+     * @throws IOException If an I/O error occurs during the retrieval.
+     */
+    private <T> KNNVectorValues<T> getKNNVectorValuesForMerge(
+        final VectorDataType vectorDataType,
+        final FieldInfo fieldInfo,
+        final MergeState mergeState
+    ) throws IOException {
+        switch (vectorDataType) {
```
jmazanec15, line 162: I think this needs to be vector encoding not vectorDataType. Otherwise, I dont think this will work for Binary

src/test/java/org/opensearch/knn/quantization/quantizationState/QuantizationStateTests.java
```diff
@@ -65,4 +66,30 @@ public void testSerializationWithDifferentVersions() throws IOException {
...
+
+        assertEquals(expectedRamBytesUsed, state.ramBytesUsed());
+    }
+
+    public void testMultiBitScalarQuantizationStateRamBytesUsed() {
+        ScalarQuantizationParams params = new ScalarQuantizationParams(ScalarQuantizationType.TWO_BIT);
+        float[][] thresholds = { { 0.5f, 1.5f, 2.5f }, { 1.0f, 2.0f, 3.0f } };
+
+        MultiBitScalarQuantizationState state = new MultiBitScalarQuantizationState(params, thresholds);
+        long expectedRamBytesUsed = RamUsageEstimator.shallowSizeOfInstance(MultiBitScalarQuantizationState.class) + RamUsageEstimator
```
jmazanec15, line 86: Could we test based on a formula as opposed to RamUsageEstimator? I just want to ensure that these are giving us what we are expecting. If not, could you paste on this comment via log what the estimate is for the tests and see if we can ensure theyre about what we expect?

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriter.java
```diff
@@ -146,4 +132,103 @@ public long ramBytesUsed() {
...
+     */
+    private <T, C> void processField(
+        final FieldInfo fieldInfo,
+        final VectorValuesRetriever<VectorDataType, FieldInfo, C, KNNVectorValues<T>> vectorValuesRetriever,
+        final IndexOperation<T> indexOperation,
+        final C context
+    ) throws IOException {
+        final VectorDataType vectorDataType = extractVectorDataType(fieldInfo);
+        KNNVectorValues<T> knnVectorValuesForTraining = vectorValuesRetriever.apply(vectorDataType, fieldInfo, context);
+        KNNVectorValues<T> knnVectorValuesForIndexing = vectorValuesRetriever.apply(vectorDataType, fieldInfo, context);
```
shatejas, line 220: Will this be holding twice the memory for same set of vectors? or underneath is it the same reference on the vectors

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriter.java
```diff
@@ -146,4 +132,103 @@ public long ramBytesUsed() {
...
+     * @param vectorValuesRetriever  A functional interface that retrieves {@link KNNVectorValues} based on the vector data type,
+     *                                field information, and additional context (e.g., merge state or field writer).
+     * @param indexOperation         A functional interface that performs the indexing operation using the retrieved
+     *                                {@link KNNVectorValues}.
+     * @param context                The additional context required for retrieving the vector values (e.g., {@link MergeState} or {@link NativeEngineFieldVectorsWriter}).
+     * @param <T>                    The type of vectors being processed.
+     * @param <C>                    The type of the context needed for retrieving the vector values.
+     * @throws IOException If an I/O error occurs during the processing.
+     */
+    private <T, C> void processField(
```
shatejas, line 212: nit: seems too generic, trainAndIndex

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriter.java
```diff
@@ -146,4 +132,103 @@ public long ramBytesUsed() {
...
+    }
+
+    /**
+     * Functional interface representing an operation that indexes the provided {@link KNNVectorValues}.
+     *
+     * @param <T> The type of vectors being processed.
+     */
+    @FunctionalInterface
+    private interface IndexOperation<T> {
+        void execute(NativeIndexWriter writer, KNNVectorValues<T> knnVectorValues) throws IOException;
```
shatejas, line 181: nit: buildAndWrite

shatejas (changes requested):

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriter.java
Vikasht34: Underneath Same refrence is getting used , Discussed with Navneet on this

src/main/java/org/opensearch/knn/index/codec/nativeindex/DefaultIndexBuildStrategy.java
```diff
@@ -60,9 +82,9 @@ public void buildAndWriteIndex(final BuildIndexParams indexInfo, final KNNVector
             if (params.containsKey(MODEL_ID)) {
                 AccessController.doPrivileged((PrivilegedAction<Void>) () -> {
                     JNIService.createIndexFromTemplate(
-                        intListToArray(tranferredDocIds),
+                        intListToArray(transferredDocIds),
                         vectorAddress,
-                        knnVectorValues.dimension(),
+                        quantizationState.isPresent() ? quantizationState.get().getDimensionForIndexing() : knnVectorValues.dimension(),
```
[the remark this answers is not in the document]
Vikasht34: Model , it's self is Quantization like it's PQ or IVF PQ ,  right now so this quantisation is not related to Model , In future we may remove Model PQ and ON-BOARD with this.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriter.java
shatejas: okay, my concern is the use of interface, we have made sure in current implementation that it does not clone. with changing code this will be easy to miss for a different implementation. There are few ways to go about this to be defensive
1. Make sure there is unit or integ test which compares the references underneath somehow with the use of interface.
2. Be defensive and reuse the same reference so we can leverage garbage collector even the references are not same underneath (this is still tricky)
```
vectorValues = retriever.apply
train()
vectorValues = retriever.apply
index()
```
3. Another option would be to have actual implementation without interface with the caveat of duplication

src/main/java/org/opensearch/knn/index/codec/nativeindex/DefaultIndexBuildStrategy.java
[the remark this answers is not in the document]
shatejas: In that case can you double check once if this line is needed right away. I think createIndexFromTemplate is only used for PQ

maybe add a validation at the start 
```
assert model != null && quantizationstate == null
```

src/main/java/org/opensearch/knn/index/quantizationIntegration/QuantizationIntegration.java
```diff
@@ -0,0 +1,122 @@
...
+
+    /**
+     * Applies quantization to the given vector using the specified {@link QuantizationState} and
+     * {@link QuantizationOutput}.
+     *
+     * @param quantizationState The {@link QuantizationState} containing the state of the trained quantizer.
+     * @param vector The vector to be quantized.
+     */
+    public R quantize(final QuantizationState quantizationState, final T vector) {
+        QuantizationOutput<R> quantizationOutput = createQuantizationOutput(quantizationState.getQuantizationParams());
```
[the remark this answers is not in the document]
Vikasht34: Good Catch ...I completely missed it .

src/main/java/org/opensearch/knn/index/quantizationIntegration/QuantizationIntegration.java
```diff
@@ -0,0 +1,122 @@
...
+    /**
+     * Retrieves the appropriate {@link VectorDataType} to be used during the transfer of vectors for indexing or merging.
+     * This method is intended to determine the correct vector data type based on the provided {@link FieldInfo}.
+     *
+     * @param fieldInfo The {@link FieldInfo} object containing metadata about the field for which the vector data type
+     *                  is being determined.
+     * @return The {@link VectorDataType} to be used during the vector transfer process
+     */
+    public VectorDataType getVectorDataTypeForTransfer(final FieldInfo fieldInfo) {
+        // TODO: Replace this with actual logic to extract quantization parameters from FieldInfo
```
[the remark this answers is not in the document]
Vikasht34: Will add in quip today.

ryanbogan (dismissed):
LGTM!

src/main/java/org/opensearch/knn/index/quantizationService/KNNVectorQuantizationTrainingRequest.java
```diff
@@ -0,0 +1,55 @@
...
+                knnVectorValues.nextDoc();
+                lastIndex++;
+            }
+            if (knnVectorValues.docId() == NO_MORE_DOCS) {
+                return null;
+            }
+            // Return the vector and the updated index
+            return knnVectorValues.getVector();
+        } catch (Exception e) {
+            throw new RuntimeException("Failed to retrieve vector for docId: " + docId, e);
```
jmazanec15, line 52: Why do we catch all exceptions?

src/main/java/org/opensearch/knn/index/codec/nativeindex/MemOptimizedNativeIndexBuildStrategy.java
```diff
@@ -56,29 +87,37 @@ public void buildAndWriteIndex(final BuildIndexParams indexInfo, final KNNVector
...
-            final List<Integer> tranferredDocIds = new ArrayList<>(transferLimit);
+            final List<Integer> transferredDocIds = new ArrayList<>(transferLimit);
+
             while (knnVectorValues.docId() != NO_MORE_DOCS) {
                 // append is false to be able to reuse the memory location
-                boolean transferred = vectorTransfer.transfer(knnVectorValues.conditionalCloneVector(), false);
-                tranferredDocIds.add(knnVectorValues.docId());
+                boolean transferred;
+                if (quantizationState != null && quantizationOutput != null) {
+                    quantizationHandler.quantize(quantizationState, knnVectorValues.getVector(), quantizationOutput);
```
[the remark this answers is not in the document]
Vikasht34: if by any reason Quantizer is not able to quantize will throw error from there , I don't think we need any check here..

src/main/java/org/opensearch/knn/index/codec/nativeindex/IndexBuildHelper.java
```diff
@@ -0,0 +1,76 @@
...
+        if (indexBuildSetup.getQuantizationState() != null && indexBuildSetup.getQuantizationOutput() != null) {
+            quantizationService.quantize(
+                indexBuildSetup.getQuantizationState(),
+                knnVectorValues.getVector(),
+                indexBuildSetup.getQuantizationOutput()
+            );
+            return vectorTransfer.transfer(indexBuildSetup.getQuantizationOutput().getQuantizedVector(), append);
+        } else {
+            return vectorTransfer.transfer(knnVectorValues.conditionalCloneVector(), append);
+        }
```
shatejas, line 47: IMO we should not be separating out the transfer, flush and close of OffHeapVectorTransfer. The intent of the previous implementation was to allocate and deallocate memory as well as manage transfers in a centralized location. Breaking it this way increases the risk of not managing the lifecycle of transfer properly irrespective of the class being package-private. 

I understand this was done to reduce duplication, but I think the risk of mismanaged lifecycle of the transfer here should be reduced by having transfer, flush, reset, close clearly being managed in one class. 

Let me know your thoughts on this

src/main/java/org/opensearch/knn/index/quantizationService/KNNVectorQuantizationTrainingRequest.java
Vikasht34: Just to give meaningful Information about the exception.

src/main/java/org/opensearch/knn/index/codec/nativeindex/IndexBuildHelper.java
Vikasht34: I don't mind isolating it ....Since you have most context on transfer .

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriter.java
Vikasht34: One way to use reflection to test private ones , It's tough to have tests around that ..I would keep it as of now . Will take an AI if we need further refactor.

Vikasht34: I used the second suggestion to isolate it

src/main/java/org/opensearch/knn/index/quantizationService/QuantizationService.java
```diff
@@ -0,0 +1,128 @@
...
+
+/**
+ * A singleton class responsible for handling the quantization process, including training a quantizer
+ * and applying quantization to vectors. This class is designed to be thread-safe.
+ *
+ * @param <T> The type of the input vectors to be quantized.
+ * @param <R> The type of the quantized output vectors.
+ */
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+public final class QuantizationService<T, R> {
```
navneet1v, line 33: The meaning of Service is pretty different in Opensearch. This is more of a manager and not service. If it is service then why it is has generics in it.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriter.java
```diff
@@ -146,4 +132,101 @@ public long ramBytesUsed() {
...
+     * @param indexOperation         A functional interface that performs the indexing operation using the retrieved
+     *                                {@link KNNVectorValues}.
+     * @param context                The additional context required for retrieving the vector values (e.g., {@link MergeState} or {@link NativeEngineFieldVectorsWriter}).
+     * @param <T>                    The type of vectors being processed.
+     * @param <C>                    The type of the context needed for retrieving the vector values.
+     * @throws IOException If an I/O error occurs during the processing.
+     */
+    private <T, C> void trainAndIndex(
+        final FieldInfo fieldInfo,
+        final VectorValuesRetriever<VectorDataType, FieldInfo, C, KNNVectorValues<T>> vectorValuesRetriever,
```
navneet1v, line 214: This is really abuse of generics. Just because 2 functions are returning same thing doesn't mean that we should use generic to make them same. If you use some overloaded functions it could have been easy to read, understand and extend.

src/main/java/org/opensearch/knn/index/quantizationService/KNNVectorQuantizationTrainingRequest.java
```diff
@@ -0,0 +1,55 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.knn.index.quantizationService;
```
navneet1v, line 6: this should be `org.opensearch.knn.index.quantization;`

src/main/java/org/opensearch/knn/index/quantizationService/KNNVectorQuantizationTrainingRequest.java
navneet1v: why it is a runtime exception but not something like `ABCException extends RuntimeException`

src/main/java/org/opensearch/knn/index/codec/nativeindex/IndexBuildSetup.java
```diff
@@ -0,0 +1,40 @@
...
+final class IndexBuildSetup {
+    /**
+     * The number of bytes per vector.
+     */
+    private final int bytesPerVector;
+
+    /**
+     * The number of dimensions in the vector.
+     */
+    private final int dimensions;
```
navneet1v, line 29: is this number of dimension in input vector or the quantized vector?

src/main/java/org/opensearch/knn/index/quantizationService/KNNVectorQuantizationTrainingRequest.java
```diff
@@ -0,0 +1,55 @@
...
+    private final KNNVectorValues<T> knnVectorValues;
+    private int lastIndex;
+
+    /**
+     * Constructs a new QuantizationFloatVectorTrainingRequest.
+     *
+     * @param knnVectorValues the KNNVectorValues instance containing the vectors.
+     */
+    KNNVectorQuantizationTrainingRequest(KNNVectorValues<T> knnVectorValues) {
+        super((int) knnVectorValues.totalLiveDocs());
```
navneet1v, line 28: why you are casting this to int?

src/main/java/org/opensearch/knn/index/quantizationService/KNNVectorQuantizationTrainingRequest.java
```diff
@@ -0,0 +1,55 @@
...
+        this.knnVectorValues = knnVectorValues;
+        this.lastIndex = 0;
+    }
+
+    /**
+     * Retrieves the float vector associated with the specified document ID.
```
navneet1v, line 34:
```suggestion
     * Retrieves the vector associated with the specified document ID.
```

src/main/java/org/opensearch/knn/index/quantizationService/KNNVectorQuantizationTrainingRequest.java
```diff
@@ -0,0 +1,55 @@
...
+     *
+     * @param docId the document ID.
+     * @return the float vector corresponding to the specified document ID, or null if the docId is invalid.
+     */
+    @Override
+    public T getVectorByDocId(int docId) {
+        try {
+            while (lastIndex <= docId) {
```
navneet1v, line 42: is there a possibility that sampler returns the docId which is `0`.

src/main/java/org/opensearch/knn/index/quantizationService/KNNVectorQuantizationTrainingRequest.java
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    /**
+     * Retrieves the float vector associated with the specified document ID.
+     *
+     * @param docId the document ID.
+     * @return the float vector corresponding to the specified document ID, or null if the docId is invalid.
+     */
+    @Override
+    public T getVectorByDocId(int docId) {
```
navneet1v, line 40: actually this is not getVectorByDocId. This is `getVectorAtThePosition(int position)`. Your sampler is not giving the docId, it is actually giving the position the position of the docs that needs to be picked for sampling.

src/main/java/org/opensearch/knn/index/quantizationService/KNNVectorQuantizationTrainingRequest.java
```diff
@@ -0,0 +1,55 @@
...
+        try {
+            while (lastIndex <= docId) {
+                knnVectorValues.nextDoc();
```
navneet1v, line 43: while doing nextDoc you can get `NO_MORE_DOCS` but this while loop will not break and will keep on going forward. I think we should break on that abrupt condition.

src/main/java/org/opensearch/knn/index/quantizationService/KNNVectorQuantizationTrainingRequest.java
```diff
@@ -0,0 +1,55 @@
...
+                lastIndex++;
+            }
+            if (knnVectorValues.docId() == NO_MORE_DOCS) {
+                return null;
+            }
```
navneet1v, line 48: isn't this an exception scenario, if current docId is NO_MORE_DOCS it means the number of vectors provided to sampler was wrong and the whole quantization should fail rather than returning null.

src/main/java/org/opensearch/knn/index/quantizationService/KNNVectorQuantizationTrainingRequest.java
```diff
@@ -0,0 +1,55 @@
...
+            // Return the vector and the updated index
```
navneet1v, line 49: you are just returning the vector here. where we are returning the updated index?

src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexWriter.java
```diff
@@ -68,13 +73,29 @@ public class NativeIndexWriter {
...
+     * @param state              The SegmentWriteState representing the current segment's writing context.
+     * @param quantizationState  The QuantizationState that contains  quantization state required for quantization
+     * @return                   A NativeIndexWriter instance appropriate for the specified field, configured with or without quantization.
+     */
+    public static NativeIndexWriter getWriter(
+        final FieldInfo fieldInfo,
+        final SegmentWriteState state,
+        final QuantizationState quantizationState
+    ) {
+        return createWriter(fieldInfo, state, quantizationState);
```
navneet1v, line 98: why do we need a private function when all you are doing is calling that function from here.

src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexWriter.java
```diff
@@ -68,13 +73,29 @@ public class NativeIndexWriter {
...
     }
```
navneet1v, line 99: I don't think so you need createWriter function. because all you are doing is calling the that function. actually you can do something like this.

```
public static NativeIndexWriter getWriter(final FieldInfo fieldInfo, SegmentWriteState state) {
    getWriter(fieldInfo, state, null)
}

public static NativeIndexWriter getWriter(
        final FieldInfo fieldInfo,
        final SegmentWriteState state,
        final QuantizationState quantizationState
    ) {
... code from create Writer here.
} 

```

src/main/java/org/opensearch/knn/index/codec/nativeindex/IndexBuildHelper.java
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    /**
+     * Prepares the quantization setup including bytes per vector and dimensions.
+     *
+     * @param knnVectorValues the KNN vector values.
+     * @param indexInfo the index build parameters.
+     * @return an instance of QuantizationSetup containing relevant information.
+     */
+    static IndexBuildSetup prepareIndexBuild(KNNVectorValues<?> knnVectorValues, BuildIndexParams indexInfo) {
```
navneet1v, line 49: This function can be easily moved to IndexBuildSetup class, and renamed as `static IndexBuildSetup build(KNNVectorValues<?> knnVectorValues, BuildIndexParams indexInfo)`.

and can be used as way to build the IndexBuildSetup object.

src/main/java/org/opensearch/knn/index/codec/nativeindex/IndexBuildHelper.java
```diff
@@ -0,0 +1,68 @@
...
+
+    /**
+     * Processes and returns the vector based on whether quantization is applied or not.
+     *
+     * @param knnVectorValues the KNN vector values to be processed.
+     * @param indexBuildSetup the setup containing quantization state and output, along with other parameters.
+     * @return the processed vector, either quantized or original.
+     * @throws IOException if an I/O error occurs during processing.
+     */
+    static Object processAndReturnVector(KNNVectorValues<?> knnVectorValues, IndexBuildSetup indexBuildSetup) throws IOException {
```
navneet1v, line 28: why not do this?
```suggestion
    static <T> T processAndReturnVector(KNNVectorValues<T> knnVectorValues, IndexBuildSetup indexBuildSetup) throws IOException {
```

src/main/java/org/opensearch/knn/index/codec/nativeindex/IndexBuildHelper.java
```diff
@@ -0,0 +1,68 @@
...
+        QuantizationService quantizationService = QuantizationService.getInstance();
```
navneet1v, line 29: why QuantizationService is without its generics? shouldn't this be:

```suggestion
        QuantizationService<?,?> quantizationService = QuantizationService.getInstance();
```

src/main/java/org/opensearch/knn/quantization/models/quantizationState/MultiBitScalarQuantizationState.java
```diff
@@ -124,4 +125,45 @@ public byte[] toByteArray() throws IOException {
...
+
+        // Calculate the number of bytes required for multi-bit quantization
+        return thresholds.length * thresholds[0].length;
+    }
+
+    @Override
+    public int getDimensions() {
+        // For multi-bit quantization, the dimension for indexing is the number of rows * columns in the thresholds array.
+        // Where number of column reprensents Dimesion of Original vector and number of rows equals to number of bits
+        return thresholds.length * thresholds[0].length;
```
navneet1v, line 149: why getDimension is not checking `thresholds are null or have invalid structure`

src/main/java/org/opensearch/knn/index/codec/nativeindex/IndexBuildHelper.java
Vikasht34: I can't make it T , If I will make this T and the method also has to be T and Class TOO , not only here , Native Index Build strategy and Default also has to be Type erasure and then it will go to NativeindexWrite ...and I did not want to make ? so I would keep as it is.

src/main/java/org/opensearch/knn/index/codec/nativeindex/IndexBuildHelper.java
Vikasht34: I can't make it T , If I will make this T and the method also has to be T and Class TOO , not only here , Native Index Build strategy and Default also has to be Type erasure and then it will go to NativeindexWrite ...and I did not want to make ? so I would keep as it is.

src/main/java/org/opensearch/knn/index/codec/nativeindex/IndexBuildHelper.java
Vikasht34: I would my POJO Simple.  and act as Simple Model Class. So I would using KISS here.

src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexWriter.java
Vikasht34: It is required so if tomorrow if I need to create another overload , I can easily extend private method , rather changing all public , so it's clean.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriter.java
Vikasht34: IMO , it's looks fine.

src/main/java/org/opensearch/knn/index/codec/nativeindex/IndexBuildSetup.java
Vikasht34: No , It could be normal flow as well as quantized.

src/main/java/org/opensearch/knn/index/quantizationService/KNNVectorQuantizationTrainingRequest.java
Vikasht34: This is getting handled while Calculating thresholds, if vector is null , it will throw exception.

src/main/java/org/opensearch/knn/index/quantizationService/KNNVectorQuantizationTrainingRequest.java
Vikasht34: no need , for null it's gona throw exception

src/main/java/org/opensearch/knn/index/quantizationService/KNNVectorQuantizationTrainingRequest.java
Vikasht34: It's exception training , we are handling that there .

src/main/java/org/opensearch/knn/index/quantizationService/KNNVectorQuantizationTrainingRequest.java
navneet1v: then the java doc is incorrect here.

src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexWriter.java
navneet1v: cannot say that. Its one and the same thing. Leaving it upto your choice

src/main/java/org/opensearch/knn/index/codec/nativeindex/IndexBuildHelper.java
navneet1v: if that is the case then Builder classes in the Builder patter will be separate from POJO. Moving this function in POJO class has it advantages in terms of restricting the usage of the class like making constructor private, any user of this class knowing how to create the `IndexBuildSetup` POJO.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriter.java
navneet1v: Would like to know the thoughts of other maintainers here and sorry it doesn't follow the KISS which you are trying to follow in other parts of the code.

cc: @jmazanec15 , @heemin32

src/main/java/org/opensearch/knn/index/codec/nativeindex/IndexBuildSetup.java
navneet1v: In that case Java doc needs to be updated to reflect this.

src/main/java/org/opensearch/knn/index/quantizationService/KNNVectorQuantizationTrainingRequest.java
jmazanec15: Why cant we just log the message in 52 and then re-throw e?

src/main/java/org/opensearch/knn/index/quantizationService/KNNVectorQuantizationTrainingRequest.java
navneet1v: Please don't resolve the comment without adding any response whether you are going to make the change or not.

src/main/java/org/opensearch/knn/index/quantizationService/KNNVectorQuantizationTrainingRequest.java
navneet1v: Here is the thing:
You function has 3 behaviors:
1. Return a valid vector.
2. Return a null vector
3. Throw exception.

You can make it just 2 and free the caller from knowing that vector could be null. This will reduce the branches on caller and on this function and make the code simpler. 

Also, this logic then will not need java docs on why we are adding it and any other conditions
```
if (knnVectorValues.docId() == NO_MORE_DOCS) {
                return null;
            }
```

src/main/java/org/opensearch/knn/index/quantizationService/KNNVectorQuantizationTrainingRequest.java
navneet1v: > It's exception training , we are handling that there .

Where?

src/main/java/org/opensearch/knn/index/codec/nativeindex/IndexBuildSetup.java
Vikasht34: It's already says The number of dimensions in the vector. ....which is pretty generic here.....

src/main/java/org/opensearch/knn/index/quantizationService/KNNVectorQuantizationTrainingRequest.java
Vikasht34: Sampler needs total number of Numbers which is int .. @

src/main/java/org/opensearch/knn/index/quantizationService/KNNVectorQuantizationTrainingRequest.java
Vikasht34:
```
static float[] calculateMeanThresholds(TrainingRequest<float[]> samplingRequest, int[] sampledIndices) {
        int totalSamples = sampledIndices.length;
        float[] mean = null;
        int lastIndex = 0;
        for (int docId : sampledIndices) {
            float[] vector = samplingRequest.getVectorAtThePosition(docId);
            if (vector == null) {
                throw new IllegalArgumentException("Vector at sampled index " + docId + " is null.");
            }
            if (mean == null) {
                mean = new float[vector.length];
            }
            for (int j = 0; j < vector.length; j++) {
                mean[j] += vector[j];
            }
        }
        if (mean == null) {
            throw new IllegalStateException("Mean array should not be null after processing vectors.");
        }
        for (int j = 0; j < mean.length; j++) {
            mean[j] /= totalSamples;
        }
        return mean;
    }
```

src/main/java/org/opensearch/knn/index/quantizationService/KNNVectorQuantizationTrainingRequest.java
Vikasht34: first and two is same ..which is return vector ..it could be null or anything else...

Vikasht34: this is what any function does ...throwing exception from function is pretty normal thing.

src/main/java/org/opensearch/knn/index/codec/nativeindex/IndexBuildHelper.java
Vikasht34: This is called Builder design pattern which tell if you have any complex Pojo , then how to build it , or for any normal pojo also ..It would be ...I would prefer keeping Model class Simple.

src/main/java/org/opensearch/knn/index/codec/nativeindex/IndexBuildHelper.java
Vikasht34: Because all the downstream is well defined generics but in this case ..,NativeBuildSratgey , Default Strategy and Nativewriter is not.. ..

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriter.java
Vikasht34: Well using generics and Functional Inteeface is also part of KISS. I would like to read any literature  where it says functional inteface is not following KISS.

src/main/java/org/opensearch/knn/index/quantizationService/KNNVectorQuantizationTrainingRequest.java
navneet1v: Let me simplify the question here:
What will happen if we keep on doing nextDoc() when doc is already at NO_MORE_DOCS? Are we expecting the iterator to throw the exception here? Have we validated this scenario?

src/main/java/org/opensearch/knn/index/quantizationService/KNNVectorQuantizationTrainingRequest.java
navneet1v: > first and two is same ..which is return vector ..it could be null or anything else...

This is the reason I said returns a valid Vector. Null is not a valid vector. Returning null when we already know that it is not the right behavior for caller why don't just throw the exception. I am not able to understand the downside here.

src/main/java/org/opensearch/knn/index/codec/nativeindex/IndexBuildHelper.java
navneet1v: The doesn't it make sense to have well defined named functions that can bridge this gap rather than avoiding generics in the object creation.

src/main/java/org/opensearch/knn/index/quantizationService/KNNVectorQuantizationTrainingRequest.java
navneet1v: All the iterators(BDV, FVV, BVV) gives long as the totalLiveDocs. Don't you think we should use long here?

src/main/java/org/opensearch/knn/index/codec/nativeindex/IndexBuildSetup.java
navneet1v: I think you miss the point, the question here is not about whether the java doc is generic or not. Its about will which dimension this variable represent. 
When you say this:
```
The number of dimensions in the vector.
```
it is unclear which vector you are taking about. Quantized vector/non quantized/ if its a binary vector type so is it representing dimension for vector which Faiss will take or length of byte array.

src/main/java/org/opensearch/knn/index/quantizationService/KNNVectorQuantizationTrainingRequest.java
Vikasht34: Sure , I can rely on VectorValue exception.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriter.java
navneet1v: What I mean to say is functional Interface will be following KISS but this function doesn't. I don't know from where you get this idea I am saying Functional interface are not part of KISS.

Vikasht34: As you suggested Writing total 6 functions for flush and merge ...that's where we have generics and functional interface to solve the problem of writing 6 functions vs 1.

src/main/java/org/opensearch/knn/index/codec/nativeindex/IndexBuildHelper.java
navneet1v: I don't agree here. But neither I am blocking the PR because of this. So, leaving it upto you.

src/main/java/org/opensearch/knn/index/codec/nativeindex/IndexBuildSetup.java
Vikasht34: I changed it to Dimension of Vector for Indexing

src/main/java/org/opensearch/knn/index/quantizationService/KNNVectorQuantizationTrainingRequest.java
Vikasht34: This casting has to be done somewhere , for Sampler it's just an total number of items which needs to be sampled , I think int should be fine.

src/main/java/org/opensearch/knn/index/quantizationService/KNNVectorQuantizationTrainingRequest.java
Vikasht34: sure , I can break the loop.

src/main/java/org/opensearch/knn/index/codec/nativeindex/IndexBuildHelper.java
Vikasht34: Thanks

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriter.java
shatejas: This felt like a bit of an over-engineering IMO. Its creating 2 functional interfaces with little advantage and little possibility of reuse. Having said that there isn't a performance disadvantage here but I would still encourage you to simplify this in terms of readability considering the code base is not functional

src/main/java/org/opensearch/knn/index/codec/nativeindex/IndexBuildHelper.java
Vikasht34: I would have to make changes to all three classes to make that ..I am avoiding that change , because that will change all of tests where it's getting used and that's not part of this PR..We can come back later to make it , if we have to.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriter.java
jmazanec15: Is there any way we can avoid the Context arguement? I think everything else looks good, but the context is a bit hard to follow. Can VectorValuesRetriever just be configured from calling code and avoid context?

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriter.java
```diff
@@ -146,4 +132,101 @@ public long ramBytesUsed() {
...
+        QuantizationParams quantizationParams = quantizationService.getQuantizationParams(fieldInfo);
+        QuantizationState quantizationState = null;
+        if (quantizationParams != null) {
+            quantizationState = quantizationService.train(quantizationParams, knnVectorValues);
+        }
+        NativeIndexWriter writer = (quantizationParams != null)
+            ? NativeIndexWriter.getWriter(fieldInfo, segmentWriteState, quantizationState)
+            : NativeIndexWriter.getWriter(fieldInfo, segmentWriteState);
+
+        knnVectorValues = vectorValuesRetriever.apply(vectorDataType, fieldInfo, context);
```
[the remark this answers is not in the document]
Vikasht34: No we need to get different vectorvalues for Training and Indexing because ,we would have already iterated.

src/main/java/org/opensearch/knn/index/codec/nativeindex/IndexBuildHelper.java
```diff
@@ -0,0 +1,68 @@
...
+import org.opensearch.knn.index.codec.nativeindex.model.BuildIndexParams;
+import org.opensearch.knn.index.quantizationService.QuantizationService;
+import org.opensearch.knn.index.vectorvalues.KNNVectorValues;
+import org.opensearch.knn.quantization.models.quantizationOutput.QuantizationOutput;
+import org.opensearch.knn.quantization.models.quantizationState.QuantizationState;
+
+import java.io.IOException;
+
+@UtilityClass
+class IndexBuildHelper {
```
shatejas, line 18: - This is more of a quantization related utility. Also please rethink the name, A `helper` suffix opens the possibility of misuse
- Also I did not see a unit test around this, seems like we are relying on indexbuild strategy test. Might be a good idea to have an independent unit test considering there is branching in the code

shatejas (commented):
Functionally it looks good. A few minor things

src/main/java/org/opensearch/knn/index/codec/nativeindex/MemOptimizedNativeIndexBuildStrategy.java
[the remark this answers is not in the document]
shatejas: Got it, The suggestion is to be more defensive. I am fine with this approach as its not too alarming right now considering quantization is heavily coupled. Would encourage you to take another opinion

src/main/java/org/opensearch/knn/index/codec/nativeindex/IndexBuildHelper.java
Vikasht34: Added Unit tests.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriter.java
Vikasht34: I changed it to have more meaningful name and added more Java Docs around that to have better understanding , if In case we need there is more understanding issue , I can come back and change in future.

jmazanec15 (approved):

ryanbogan (dismissed):

shatejas (approved):
lgtm

src/main/java/org/opensearch/knn/index/quantizationService/KNNVectorQuantizationTrainingRequest.java
jmazanec15: I think we do need to change this - if num vecs > 2B, I think we can run into issues. Im fine adding as a follow up in another review, but we should represent total number of vectors in an index as long because 2B is somewhat reasonable.

ryanbogan (approved):

src/main/java/org/opensearch/knn/index/quantizationService/KNNVectorQuantizationTrainingRequest.java
Vikasht34: This is from Lucene 

[https://github.com/apache/lucene/blob/91f45b53e6b4a9b68d110ee128797e74a6578379/luc[…]c/java/org/apache/lucene/util/quantization/ScalarQuantizer.java](https://github.com/apache/lucene/blob/91f45b53e6b4a9b68d110ee128797e74a6578379/lucene/core/src/java/org/apache/lucene/util/quantization/ScalarQuantizer.java#L306-L325)

and this is https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/DocIdSetIterator.java#L43
[DocIdSetIterator.java](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/DocIdSetIterator.java)



if you see this link https://issues.apache.org/jira/browse/LUCENE-4607 
because the current scorer/disi interface can never return anything > Integer.MAX_VALUE, so the estimatedDocCount can never be 64 bits.

I think we should change estimation to int rather than long.

jmazanec15 (approved):

src/main/java/org/opensearch/knn/index/quantizationService/KNNVectorQuantizationTrainingRequest.java
Vikasht34: I have removed any catching the exception and throwing and relying on upstream, right now this function simple enough to return vector or null , any how when I throw any expception from here ..Client has to know about that ..I will let Clinet handle Null and take call appropriately . I am closing this , If you have any further improvement I can take in  follow up PR.

src/main/java/org/opensearch/knn/index/quantizationService/KNNVectorQuantizationTrainingRequest.java
Vikasht34: It's part of Client not Quantization Framework , I would keep it on Quantization service.

src/main/java/org/opensearch/knn/index/quantizationService/KNNVectorQuantizationTrainingRequest.java
Vikasht34: Sorry I missed , Will remove this in follow up PR.

src/main/java/org/opensearch/knn/index/quantizationService/KNNVectorQuantizationTrainingRequest.java
navneet1v: If you want to keep service then atleast make it all small case or quantization.service.

Camel case in package name I have not seen anywhere that was the whole point

Vikasht34: Sure , I will fix this in followup PR.

PR #1997, opened 2024-08-21
Add quantization state reader and writer

ryanbogan:
### Description
Adds a quantization state writer and reader for native engines

### Check List


ryanbogan:
I still need to write unit tests, but raising now for initial feedback

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNNQuantizationStateReader.java
```diff
@@ -0,0 +1,122 @@
...
+
+    /**
+     * Read quantization states and return list of fieldNames and bytes
+     * @param state the read state to read from
+     */
+    public Map<String, byte[]> read(SegmentReadState state) {
+        String quantizationStateFileName = IndexFileNames.segmentFileName(state.segmentInfo.name, state.segmentSuffix, "qs");
+        Map<String, byte[]> readQuantizationStateInfos = new HashMap<>();
+
+        try (IndexInput input = state.directory.openInput(quantizationStateFileName, IOContext.READ)) {
```
jmazanec15, line 36: Should we open with checksum? https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/store/Directory.java#L147-L156

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNNQuantizationStateReader.java
```diff
@@ -0,0 +1,122 @@
...
+
+            long footerStart = input.length() - CodecUtil.footerLength();
```
jmazanec15, line 38: Should we put this block in a method?

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNNQuantizationStateReader.java
```diff
@@ -0,0 +1,122 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Reads quantization states
+ */
+public class KNNQuantizationStateReader {
+
+    /**
```
jmazanec15, line 28: Can you add a layout of the format of the file in the comment?

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNNQuantizationStateWriter.java
```diff
@@ -0,0 +1,89 @@
...
+    /**
+     * Constructor
+     * @param segmentWriteState segment write state containing segment information
+     * @throws IOException exception could be thrown while creating the output
+     */
+    public KNNQuantizationStateWriter(SegmentWriteState segmentWriteState) throws IOException {
+        String quantizationStateFileName = IndexFileNames.segmentFileName(
+            segmentWriteState.segmentInfo.name,
+            segmentWriteState.segmentSuffix,
+            "qs"
```
jmazanec15, line 36: Make this a constnt

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNNQuantizationStateReader.java
ryanbogan: I don't think that will work because I'm seeking backwards at times during the file read

ryanbogan: https://github.com/opensearch-project/k-NN/pull/1997

jmazanec15: Oh good to know then thats good

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNNQuantizationStateReader.java
```diff
@@ -0,0 +1,145 @@
...
+                long position = input.readVLong();
+                positions.add(position);
+            }
+            // Read each field's bytes
+            for (int i = 0; i < numFields; i++) {
+                input.seek(positions.get(i));
+                byte[] stateBytes = new byte[lengths.get(i)];
+                input.readBytes(stateBytes, 0, lengths.get(i));
+                readQuantizationStateInfos.put(fieldNames.get(i), stateBytes);
+            }
```
jmazanec15, line 79: Can we validate footer check sum? i.e. - https://github.com/apache/lucene/blob/b4a06770c996c2f5263d6006bc6d62a1c9f45b19/lucene/core/src/java/org/apache/lucene/codecs/lucene90/Lucene90DocValuesProducer.java#L139

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNNQuantizationStateReader.java
```diff
@@ -0,0 +1,145 @@
...
+        } catch (IOException e) {
+            throw new RuntimeException(e);
```
jmazanec15, line 81: We can just throw IOException

jmazanec15:
```
public NativeEngines990KnnVectorsWriter(SegmentWriteState segmentWriteState, FlatVectorsWriter flatVectorsWriter) throws IOException {
```

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNNQuantizationStateReader.java
```diff
@@ -0,0 +1,145 @@
...
+    public Map<String, byte[]> read(SegmentReadState state) {
+        String quantizationStateFileName = IndexFileNames.segmentFileName(
+            state.segmentInfo.name,
+            state.segmentSuffix,
+            KNNConstants.QUANTIZATION_STATE_FILE_SUFFIX
+        );
+        Map<String, byte[]> readQuantizationStateInfos = new HashMap<>();
+
+        try (IndexInput input = state.directory.openInput(quantizationStateFileName, IOContext.READ)) {
+
```
jmazanec15, line 58: Do we need to read the header?

ryanbogan: No, we don't actually need the header since we're working backwards from the footer.

ryanbogan: But it has to be written for the input to work

ryanbogan:
I'll add end to end integration tests to fully test the quantization reader and writer once the integration is complete with indexing and query flow.  Additionally, I'll add another unit test for the second read method with the PR for integration to the query flow.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNNQuantizationStateReader.java
```diff
@@ -0,0 +1,143 @@
...
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Reads quantization states
+ */
+public class KNNQuantizationStateReader {
```
Vikasht34, line 28: Let's make this class Final , Restrict Construstor.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNNQuantizationStateWriter.java
```diff
@@ -0,0 +1,90 @@
...
+import org.opensearch.knn.quantization.models.quantizationState.QuantizationState;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Writes quantization states to off heap memory
+ */
+public class KNNQuantizationStateWriter {
```
Vikasht34, line 23: Final Class with Lombok ?

src/main/java/org/opensearch/knn/common/KNNConstants.java
```diff
@@ -71,6 +71,7 @@ public class KNNConstants {
     public static final VectorDataType DEFAULT_VECTOR_DATA_TYPE_FIELD = VectorDataType.FLOAT;
 
     public static final String RADIAL_SEARCH_KEY = "radial_search";
+    public static final String QUANTIZATION_STATE_FILE_SUFFIX = "qs";
```
navneet1v, line 74: should we add some plugin extension too here. I feel qs being so generic what will happen if Lucene has this file extension at some point of time. 

@jmazanec15 , @Vikasht34 thoughts?

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriter.java
```diff
@@ -78,8 +86,12 @@ public void flush(int maxDoc, final Sorter.DocMap sortMap) throws IOException {
                 field.getVectors()
             );
 
+            // TODO: Extract quantization state here, uncomment below line once implemented
+            // quantizationStateWriter.writeState(field.getFieldInfo().getName(), quantizationState);
```
navneet1v, line 90: Should remove the comments now?

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNNQuantizationStateReader.java
```diff
@@ -0,0 +1,143 @@
...
+
+    /**
+     * Reads an individual quantization state for a given field
+     * @param directory directory to open input
+     * @param segmentName segment name
+     * @param segmentSuffix segment suffix
+     * @param fieldInfo field information
+     * @return quantization state
+     */
+    public QuantizationState read(Directory directory, String segmentName, String segmentSuffix, FieldInfo fieldInfo) throws IOException {
```
Vikasht34, line 93: Let's wrap all four arguments in Model , and Opening of Index Input Let's have in Constructor  for both reads.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNNQuantizationStateWriter.java
```diff
@@ -0,0 +1,90 @@
...
+        for (FieldQuantizationState fieldQuantizationState : fieldQuantizationStates) {
+            output.writeString(fieldQuantizationState.fieldName);
+            output.writeInt(fieldQuantizationState.stateBytes.length);
+            output.writeVLong(fieldQuantizationState.position);
+        }
+        output.writeLong(indexStartPosition);
+        output.writeInt(-1);
+        CodecUtil.writeFooter(output);
+        output.close();
+        fieldQuantizationStates = new ArrayList<>();
```
navneet1v, line 81: once we have written the footer why we need to make `fieldQuantizationStates = new Arraylist?`

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNNQuantizationStateWriter.java
```diff
@@ -0,0 +1,90 @@
...
+        output.writeInt(fieldQuantizationStates.size());
+        for (FieldQuantizationState fieldQuantizationState : fieldQuantizationStates) {
+            output.writeString(fieldQuantizationState.fieldName);
+            output.writeInt(fieldQuantizationState.stateBytes.length);
+            output.writeVLong(fieldQuantizationState.position);
+        }
+        output.writeLong(indexStartPosition);
+        output.writeInt(-1);
+        CodecUtil.writeFooter(output);
+        output.close();
```
navneet1v, line 80: shouldn't this be in a separate close function, which should be triggered by the Close of NativeIndexWriter.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNNQuantizationStateReader.java
navneet1v: Also can you please add same in the Writer Class too.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNNQuantizationStateReader.java
navneet1v: @ryanbogan in most of the remote store implementation we download the headers of the file which provide all the details on where the data is present for a particular filed so that right blocks can be downloaded. Are we storing this information in header too for QS?

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNNQuantizationStateReader.java
```diff
@@ -0,0 +1,143 @@
...
+            positions.add(position);
+        }
+        // Read each field's bytes
+        for (int i = 0; i < numFields; i++) {
+            input.seek(positions.get(i));
+            byte[] stateBytes = new byte[lengths.get(i)];
+            input.readBytes(stateBytes, 0, lengths.get(i));
+            readQuantizationStateInfos.put(fieldNames.get(i), stateBytes);
+        }
+        input.close();
```
navneet1v, line 81: why we are closing this input?

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNNQuantizationStateReader.java
```diff
@@ -0,0 +1,143 @@
...
+     * QS2 field name
+     * QS2 state bytes length
+     * QS2 position of state bytes
+     * Position of index section (where QS1 field name is located)
+     * -1 (marker)
+     * Footer
+     *
+     * @param state the read state to read from
+     */
+    public Map<String, byte[]> read(SegmentReadState state) throws IOException {
```
navneet1v, line 49: This is interesting function. We are always reading all the QS at once for all the fields. Shouldn't this be more like read QS for a field name?

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNNQuantizationStateWriter.java
```diff
@@ -0,0 +1,90 @@
...
+
+    /**
+     * Writes index footer and other index information for parsing later
+     * @throws IOException could be thrown while writing
+     */
+    public void writeFooter() throws IOException {
+        long indexStartPosition = output.getFilePointer();
+        output.writeInt(fieldQuantizationStates.size());
+        for (FieldQuantizationState fieldQuantizationState : fieldQuantizationStates) {
+            output.writeString(fieldQuantizationState.fieldName);
```
navneet1v, line 73: why we are writing the field name? we can write the ordinal of the field which will be just int and will save space for us.

navneet1v (commented):
I see that writer is only added for flush and not for merge. Will there be separate PR for that?

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNNQuantizationStateWriter.java
```diff
@@ -0,0 +1,90 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Writes quantization states to off heap memory
+ */
+public class KNNQuantizationStateWriter {
+
+    private final IndexOutput output;
+    private List<FieldQuantizationState> fieldQuantizationStates = new ArrayList<>();
```
navneet1v, line 26: why we need this array list?

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriter.java
ryanbogan: I need to wait until after Vikash's next PR is merged to uncomment.  The quantizationState variable doesn't exist yet

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNNQuantizationStateWriter.java
ryanbogan: It clears the variable if writeState() is ever called again

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNNQuantizationStateWriter.java
ryanbogan: This array list stores information that is written as part of the footer for navigating the file during read

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNNQuantizationStateWriter.java
ryanbogan: Does that still work for the use case where we want to read a specific field from a given segment?

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNNQuantizationStateReader.java
ryanbogan: I don't think so.  We write the field names, length of bytes, and position of the data in an index section right before the footer.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNNQuantizationStateReader.java
ryanbogan: The input is specific to the method, so it needs to be closed once the read is complete.

ryanbogan:
> I see that writer is only added for flush and not for merge. Will there be separate PR for that?

I thought that the merge method eventually uses flush based on https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriter.java#L90

If that is not the case, I can add in separate PR

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNNQuantizationStateReader.java
ryanbogan: The other method only reads until it finds a specific field.

navneet1v:
> > I see that writer is only added for flush and not for merge. Will there be separate PR for that?
> 
> I thought that the merge method eventually uses flush based on https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriter.java#L90
> 
> If that is not the case, I can add in separate PR

No it doesn't use same method. I would not recommend creating another PR. Please add it in 1 PR as this PR is very small and I would like to see both flush and merge for QS writer in 1 shot.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNNQuantizationStateWriter.java
navneet1v: Got it. But why we are using footer for navigation? Is this the practice in lucene?

ryanbogan: It's not the actual footer, but an indexing section that we created right before it

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNNQuantizationStateWriter.java
navneet1v: Yes it works. What you need to do is find the ordinal of the field and then use that ordinal to know where the field is present. I would recommend looking at some of the implementations of Lucene for flat vector writers. If we are not able to figure that out in that case using field_name is still fine.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNNQuantizationStateReader.java
navneet1v: I think in that case this function should be a static function. Plus, if at any place exception occurs then this input.close will never be called. It would be better if we use try with resource here.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNNQuantizationStateWriter.java
navneet1v: but the StateWriter lifecycle is with the NativeIndexWriter, so why to do this? if this would have been a c++ code I can understand we want to clear the memory but in this case is it really needed?

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriter.java
navneet1v: make sense.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNNQuantizationStateWriter.java
ryanbogan: I guess not, I can get rid of it

src/main/java/org/opensearch/knn/common/KNNConstants.java
jmazanec15: agree - maybe qframe or pluginqs

ryanbogan: qstate?

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNNQuantizationStateWriter.java
ryanbogan: Looks like FieldInfo contains an attribute `number` that is the internal field number.  Would using that be better?

Vikasht34:
> > > I see that writer is only added for flush and not for merge. Will there be separate PR for that?
> > 
> > 
> > I thought that the merge method eventually uses flush based on https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriter.java#L90
> > If that is not the case, I can add in separate PR
> 
> No it doesn't use same method. I would not recommend creating another PR. Please add it in 1 PR as this PR is very small and I would like to see both flush and merge for QS writer in 1 shot.


With My CR lots of things has been changed with Flush and Merge , We don't have to save Quantization State at two places , it will be in one Function , We should even hold on Flush also .

ryanbogan:
Still need a unit test for the second read method now that it's fully implemented.

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -256,6 +274,20 @@ private Map<Integer, Float> doANNSearch(
             );
         }
 
+        if (quantizationState != null) {
```
Vikasht34, line 277: We don't need all this ..let me sync with you.

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -256,6 +258,25 @@ private Map<Integer, Float> doANNSearch(
             );
         }
 
+        QuantizationParams quantizationParams = quantizationService.getQuantizationParams(fieldInfo);
+
+        byte[] quantizedVector = null;
```
Vikasht34, line 263: This will not be the case always , It depends on type of Quantizer. It might be ok for BinaryQuantization but it wil change of others.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNNQuantizationStateReader.java
```diff
@@ -0,0 +1,152 @@
...
+
+            if (position == -1 || length == 0) {
+                throw new IllegalArgumentException(String.format("Field %s not found", readConfig.getFieldInfo().getName()));
+            }
+
+            input.seek(position);
+            byte[] stateBytes = new byte[length];
+            input.readBytes(stateBytes, 0, length);
+            // Deserialize the byte array to a quantization state object
+            ScalarQuantizationType scalarQuantizationType = ScalarQuantizationType.fromId(
```
Vikasht34, line 128: This not gona work , and we don't need this plus here we need to make work of other types of Quantization Params.

Vikasht34 (commented):
I think we missed the code where we are loading the cache from NativeEngine990knnVectorReader and evicting the cache from close function

src/main/java/org/opensearch/knn/quantization/models/quantizationState/QuantizationStateCacheManager.java
```diff
@@ -0,0 +1,81 @@
...
+
+        return quantizationState;
+    }
+
+    /**
+     * Adds or updates a quantization state in the cache.
+     * @param fieldName The name of the field.
+     * @param quantizationState The quantization state to store.
+     */
+    private void addQuantizationState(String fieldName, QuantizationState quantizationState) {
```
Vikasht34, line 59: I think this function has to be public so that , from NativeReader we can load all cache.

src/main/java/org/opensearch/knn/quantization/models/quantizationState/QuantizationStateCacheManager.java
```diff
@@ -0,0 +1,81 @@
...
+package org.opensearch.knn.quantization.models.quantizationState;
+
+import lombok.AccessLevel;
+import lombok.NoArgsConstructor;
+import org.opensearch.knn.index.codec.KNN990Codec.KNNQuantizationStateReader;
+
+import java.io.IOException;
+
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+public final class QuantizationStateCacheManager {
```
Vikasht34, line 15: Please add unit test for this class .

src/main/java/org/opensearch/knn/index/KNNSettings.java
```diff
@@ -396,11 +396,11 @@ private void setSettingsUpdateConsumers() {
             NativeMemoryCacheManager.getInstance().rebuildCache(builder.build());
         }, Stream.concat(dynamicCacheSettings.values().stream(), FEATURE_FLAGS.values().stream()).collect(Collectors.toUnmodifiableList()));
         clusterService.getClusterSettings().addSettingsUpdateConsumer(QUANTIZATION_STATE_CACHE_SIZE_LIMIT_SETTING, it -> {
-            QuantizationStateCache.getInstance().setMaxCacheSizeInKB(it.getKb());
-            QuantizationStateCache.getInstance().rebuildCache();
+            QuantizationStateCacheManager.getInstance().setMaxCacheSizeInKB(it.getKb());
```
Vikasht34, line 399: let's not duplicate the getInstance call multiple times?

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNNQuantizationStateReader.java
```diff
@@ -0,0 +1,153 @@
...
+     * QS2 field number
+     * QS2 state bytes length
+     * QS2 position of state bytes
+     * Position of index section (where QS1 field name is located)
+     * -1 (marker)
+     * Footer
+     *
+     * @param state the read state to read from
+     */
+    public static Map<String, byte[]> read(SegmentReadState state) throws IOException {
```
Vikasht34, line 52: between Read and other read methods lots of things are common , let abstract out common functionality? and after that please ensure to test end to end.

ryanbogan: Added abstractions and manually tested end to end.

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
ryanbogan: We can change this as we add more types of quantization.  Eventually, we might have to run checks on the params and cast further down in the class.

navneet1v: Lets add a TODO on top of this that this needs to be changed.

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -256,6 +258,24 @@ private Map<Integer, Float> doANNSearch(
...
+
+        if (quantizationParams != null) {
+            QuantizationConfigKNNCollector tempCollector = new QuantizationConfigKNNCollector();
+            reader.searchNearestVectors(knnQuery.getField(), new float[0], tempCollector, null);
+            if (tempCollector.getQuantizationState() == null) {
+                throw new IllegalStateException("No quantization state for file");
+            }
+            QuantizationOutput quantizationOutput = quantizationService.createQuantizationOutput(quantizationParams);
+            quantizedVector = (byte[]) quantizationService.quantize(
```
navneet1v, line 272: Please add a TODO on top of this line which suggest that in future it is not necessary that we will have byte[] as an output from quantize function.

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -273,7 +293,12 @@ private Map<Integer, Float> doANNSearch(
                 new NativeMemoryEntryContext.IndexEntryContext(
                     indexPath.toString(),
                     NativeMemoryLoadStrategy.IndexLoadStrategy.getInstance(),
-                    getParametersAtLoading(spaceType, knnEngine, knnQuery.getIndexName(), vectorDataType),
+                    getParametersAtLoading(
+                        spaceType,
+                        knnEngine,
+                        knnQuery.getIndexName(),
+                        quantizationParams == null ? vectorDataType : VectorDataType.BINARY
```
navneet1v, line 300: same as above.

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -296,10 +321,11 @@ private Map<Integer, Float> doANNSearch(
             }
             int[] parentIds = getParentIdsArray(context);
             if (k > 0) {
-                if (knnQuery.getVectorDataType() == VectorDataType.BINARY) {
+                if (knnQuery.getVectorDataType() == VectorDataType.BINARY
+                    || quantizationParams != null && quantizationService.getVectorDataTypeForTransfer(fieldInfo) == VectorDataType.BINARY) {
                     results = JNIService.queryBinaryIndex(
                         indexAllocation.getMemoryAddress(),
-                        knnQuery.getByteQueryVector(),
+                        quantizationParams == null ? knnQuery.getByteQueryVector() : quantizedVector,
```
navneet1v, line 328: same as above

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/QuantizationConfigKNNCollector.java
```diff
@@ -0,0 +1,59 @@
...
+
+import lombok.Getter;
+import lombok.Setter;
+import org.apache.lucene.search.KnnCollector;
+import org.apache.lucene.search.TopDocs;
+import org.opensearch.knn.quantization.models.quantizationState.QuantizationState;
+
+@Setter
+@Getter
+public class QuantizationConfigKNNCollector implements KnnCollector {
```
navneet1v, line 16: add Java docs on top of this.

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -217,7 +221,8 @@ private Map<Integer, Float> doANNSearch(
         final int cardinality,
         final int k
     ) throws IOException {
-        final SegmentReader reader = Lucene.segmentReader(context.reader());
+        LeafReader reader2 = context.reader();
+        final SegmentReader reader = Lucene.segmentReader(reader2);
```
navneet1v, line 225:
```suggestion
        final SegmentReader reader = Lucene.segmentReader(context.reader());
```

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -217,7 +221,8 @@ private Map<Integer, Float> doANNSearch(
         final int cardinality,
         final int k
     ) throws IOException {
-        final SegmentReader reader = Lucene.segmentReader(context.reader());
+        LeafReader reader2 = context.reader();
```
navneet1v, line 224: Please add UTs for KNNWeightClass for the new flow.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriter.java
```diff
@@ -42,15 +41,22 @@
...
+    private final KNNQuantizationStateWriter quantizationStateWriter;
     private final List<NativeEngineFieldVectorsWriter<?>> fields = new ArrayList<>();
     private boolean finished;
     private final QuantizationService quantizationService = QuantizationService.getInstance();
 
+    public NativeEngines990KnnVectorsWriter(SegmentWriteState segmentWriteState, FlatVectorsWriter flatVectorsWriter) throws IOException {
+        this.segmentWriteState = segmentWriteState;
+        this.flatVectorsWriter = flatVectorsWriter;
+        this.quantizationStateWriter = new KNNQuantizationStateWriter(segmentWriteState);
+        quantizationStateWriter.writeHeader(segmentWriteState);
```
navneet1v, line 57: shouldn't we do this in the merge and refresh function, rather than doing this in the constructor?

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
```diff
@@ -101,6 +117,21 @@ public ByteVectorValues getByteVectorValues(final String field) throws IOExcepti
      */
     @Override
     public void search(String field, float[] target, KnnCollector knnCollector, Bits acceptDocs) throws IOException {
+        if (knnCollector instanceof QuantizationConfigKNNCollector) {
```
navneet1v, line 120: lets put a TODO here suggesting that this a hack, where we are using KNNCollector to init the QS.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNNQuantizationStateWriter.java
```diff
@@ -0,0 +1,109 @@
...
+        output = segmentWriteState.directory.createOutput(quantizationStateFileName, segmentWriteState.context);
+    }
+
+    /**
+     * Writes an index header
+     * @param segmentWriteState state containing segment information
+     * @throws IOException exception could be thrown while writing header
+     */
+    public void writeHeader(SegmentWriteState segmentWriteState) throws IOException {
+        CodecUtil.writeIndexHeader(output, "QuantizationCodec", 0, segmentWriteState.segmentInfo.getId(), segmentWriteState.segmentSuffix);
```
navneet1v, line 64: `QuantizationCodec` not the correct value here. We need to add the perfieldFormat or the  Codec name which we are using to create this file.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNNQuantizationStateReader.java
```diff
@@ -0,0 +1,155 @@
...
+
+    @VisibleForTesting
+    static byte[] readStateBytes(IndexInput input, long position, int length) throws IOException {
+        input.seek(position);
+        byte[] stateBytes = new byte[length];
+        input.readBytes(stateBytes, 0, length);
+        return stateBytes;
+    }
+
+    @VisibleForTesting
```
navneet1v, line 151: why we are enabling @VisibleForTesting on these functions? These should be tested along with read functions

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNNQuantizationStateReader.java
```diff
@@ -0,0 +1,155 @@
...
+        }
+        return readQuantizationStateInfos;
+    }
+
+    /**
+     * Reads an individual quantization state for a given field
+     * @param readConfig a config class that contains necessary information for reading the state
+     * @return quantization state
+     */
+    public static QuantizationState read(QuantizationStateReadConfig readConfig) throws IOException {
```
navneet1v, line 88: who is the caller of this function?

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNNQuantizationStateReader.java
```diff
@@ -0,0 +1,155 @@
...
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Reads quantization states
+ */
+public final class KNNQuantizationStateReader {
```
navneet1v, line 31: I still think we should configure this class a proper reader rather than having a bunch static functions in it and then giving all responsibility of populating cache to `NativeEngines990KnnVectorsReader` as NativeEngines990KnnVectorsReader has more responsibilities. I feel `NativeEngines990KnnVectorsReader` is more like an uber Reader that just provide ways to read values using specialized readers like FlatReaders or KNNQSReaders.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriter.java
ryanbogan: Is there a downside to initializing in the constructor?  There is only one writer per segment so there will only need to be one file for each of the writers.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNNQuantizationStateWriter.java
ryanbogan: So KNN990Codec should be here instead?

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNNQuantizationStateReader.java
ryanbogan: Visible for testing doesn't actually do anything as it is just an annotation.  These methods would otherwise be private, but since they need to be tested I made them package level access.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNNQuantizationStateReader.java
ryanbogan: It's called by the cache manager when attempting to get a quantization state from the cache.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNNQuantizationStateReader.java
```diff
@@ -0,0 +1,155 @@
...
+                throw new IllegalArgumentException(String.format("Field %s not found", field));
+            }
+
+            byte[] stateBytes = readStateBytes(input, position, length);
+
+            // Deserialize the byte array to a quantization state object
+            ScalarQuantizationType scalarQuantizationType = ((ScalarQuantizationParams) readConfig.getQuantizationParams()).getSqType();
+            if (scalarQuantizationType == ScalarQuantizationType.ONE_BIT) {
+                return OneBitScalarQuantizationState.fromByteArray(stateBytes);
+            } else if (scalarQuantizationType == ScalarQuantizationType.TWO_BIT
```
shatejas, line 123: nit: switch case would have been cleaner here

src/test/java/org/opensearch/knn/index/codec/KNN990Codec/KNNQuantizationStateReaderTests.java
```diff
@@ -0,0 +1,236 @@
...
+            fieldInfos,
+            Mockito.mock(IOContext.class),
+            segmentSuffix
+        );
+
+        try (MockedStatic<KNNQuantizationStateReader> mockedStaticReader = Mockito.mockStatic(KNNQuantizationStateReader.class)) {
+            mockedStaticReader.when(() -> KNNQuantizationStateReader.getNumFields(input)).thenReturn(2);
+            mockedStaticReader.when(() -> KNNQuantizationStateReader.read(segmentReadState)).thenCallRealMethod();
+            try (MockedStatic<CodecUtil> mockedStaticCodecUtil = mockStatic(CodecUtil.class)) {
+                KNNQuantizationStateReader.read(segmentReadState);
```
shatejas, line 84: is it possible to create an expected map and do assertEquals(expectedMap, KNNQuantizationStateReader.read(segmentReadState)) for a stronger check?

src/test/java/org/opensearch/knn/index/codec/KNN990Codec/KNNQuantizationStateReaderTests.java
```diff
@@ -0,0 +1,236 @@
...
+
+                    Mockito.when(quantizationStateReadConfig.getQuantizationParams()).thenReturn(scalarQuantizationParams2);
+                    Mockito.when(quantizationStateReadConfig.getQuantizationParams()).thenReturn(scalarQuantizationParams2);
+                    QuantizationState quantizationState = KNNQuantizationStateReader.read(quantizationStateReadConfig);
+                    assertTrue(quantizationState instanceof MultiBitScalarQuantizationState);
+
+                    Mockito.when(quantizationStateReadConfig.getQuantizationParams()).thenReturn(scalarQuantizationParams4);
+                    Mockito.when(quantizationStateReadConfig.getQuantizationParams()).thenReturn(scalarQuantizationParams4);
+                    quantizationState = KNNQuantizationStateReader.read(quantizationStateReadConfig);
+                    assertTrue(quantizationState instanceof MultiBitScalarQuantizationState);
```
shatejas, line 174: instead of instanceof, can we do `assertEquals(expectedState, quantizationState)`?

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNNQuantizationStateWriter.java
```diff
@@ -0,0 +1,109 @@
...
+        output = segmentWriteState.directory.createOutput(quantizationStateFileName, segmentWriteState.context);
+    }
+
+    /**
+     * Writes an index header
+     * @param segmentWriteState state containing segment information
+     * @throws IOException exception could be thrown while writing header
+     */
+    public void writeHeader(SegmentWriteState segmentWriteState) throws IOException {
+        CodecUtil.writeIndexHeader(output, "KNN990Codec", 0, segmentWriteState.segmentInfo.getId(), segmentWriteState.segmentSuffix);
```
shatejas, line 64: can we use `KNNCodecVersion` to get the codecName instead of a string here, 

Also if this writer is tied to the codec, 
- should we be encapsulating that in the enum like the vector formats?
- Should this be renamed to `KNN990QuantizationStateWriter`?

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNNQuantizationStateReader.java
navneet1v: So basically this code part will hit after eviction of the key. Right?

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
```diff
@@ -33,8 +44,13 @@
 public class NativeEngines990KnnVectorsReader extends KnnVectorsReader {
 
     private final FlatVectorsReader flatVectorsReader;
+    private final SegmentReadState segmentReadState;
+    private final QuantizationStateCacheManager quantizationStateCacheManager = QuantizationStateCacheManager.getInstance();
+    private Map<String, String> fieldToUniqueCacheId;
```
shatejas, line 49: consider `quatizationStateCacheKeyPerField`

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
```diff
@@ -159,4 +194,31 @@ public void close() throws IOException {
     public long ramBytesUsed() {
         return flatVectorsReader.ramBytesUsed();
     }
+
+    private void populateFieldMapAndQuantizationStateCache() throws IOException {
```
shatejas, line 198: consider `primeQuantizationStateCache`

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNNQuantizationStateWriter.java
navneet1v: Looking at some references in Lucene I can see this is what being used:
https://github.com/apache/lucene/blob/branch_9x/lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99ScalarQuantizedVectorsFormat.java#L47

So we should have something like `NativeEngines99KnnVectorsFormatQSData`

navneet1v: > Should this be renamed to KNN990QuantizationStateWriter

+1 on this. This should happen for both readers and writers

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -296,10 +324,12 @@ private Map<Integer, Float> doANNSearch(
             }
             int[] parentIds = getParentIdsArray(context);
             if (k > 0) {
-                if (knnQuery.getVectorDataType() == VectorDataType.BINARY) {
+                if (knnQuery.getVectorDataType() == VectorDataType.BINARY
+                    || quantizationParams != null && quantizationService.getVectorDataTypeForTransfer(fieldInfo) == VectorDataType.BINARY) {
```
shatejas, line 328: Instead of this create a variable `finalVectorDataType`/ `processedVectorDataType` and compute the vector data and then use it everywhere

```
finalVectorDataType = knnQuery or quantizedService as a source

```

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -256,6 +258,26 @@ private Map<Integer, Float> doANNSearch(
             );
         }
 
+        QuantizationParams quantizationParams = quantizationService.getQuantizationParams(fieldInfo);
+
+        // TODO: Change type of vector once more quantization methods are supported
+        byte[] quantizedVector = null;
+
+        if (quantizationParams != null) {
```
shatejas, line 266: nit: abstract this out into a method

```
byte[] quantizedVector = getQuantizedVector(quantParams)
```

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -256,6 +258,26 @@ private Map<Integer, Float> doANNSearch(
...
+            QuantizationConfigKNNCollector tempCollector = new QuantizationConfigKNNCollector();
+            reader.searchNearestVectors(knnQuery.getField(), new float[0], tempCollector, null);
```
shatejas, line 268: Should we be adding an instance of check for the reader to be safe?

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
shatejas: nit: Would have preferred type casting the reader instead of this. not super urgent

shatejas (commented):
Most things aren't a blocker

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNNQuantizationStateReader.java
ryanbogan: Yeah, if the key is evicted already, then this is the method that is called

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNNQuantizationStateReader.java
```diff
@@ -0,0 +1,155 @@
...
+                lengths.add(length);
+                long position = input.readVLong();
+                positions.add(position);
+            }
+            // Read each field's bytes
+            for (int i = 0; i < numFields; i++) {
+                byte[] stateBytes = readStateBytes(input, positions.get(i), lengths.get(i));
+                String fieldName = state.fieldInfos.fieldInfo(fieldNumbers.get(i)).getName();
+                readQuantizationStateInfos.put(fieldName, stateBytes);
+            }
```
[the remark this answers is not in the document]
ryanbogan: I don't think this works because we are reading all of the positions and lists back to back in an index section.  All of the bytes are in their own section.

src/test/java/org/opensearch/knn/index/codec/KNN990Codec/KNNQuantizationStateReaderTests.java
ryanbogan: I tried to get it to work but wasn't able to.  The issue is I have to simulate the file to write in order to get the proper response out of the read function since I can't globally return a response of bytes for input.readInt() or input.readLong().

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNNQuantizationStateReader.java
```diff
@@ -0,0 +1,155 @@
...
+        }
+    }
+
+    @VisibleForTesting
+    static int getNumFields(IndexInput input) throws IOException {
+        long footerStart = input.length() - CodecUtil.footerLength();
+        long markerAndIndexPosition = footerStart - Integer.BYTES - Long.BYTES;
+        input.seek(markerAndIndexPosition);
+        long indexStartPosition = input.readLong();
+        input.readInt();
```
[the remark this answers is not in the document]
ryanbogan: I did some testing, and I still need to write the marker, but I don't have to read it. Will delete this line

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
ryanbogan: The reader is always a SegmentReader

ryanbogan:
BWC failures are not due to this PR

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
ryanbogan: As of right now, we only use this check here so it would only be used once.  Everywhere else it just reads from the query vector type and does not need the influence of the quantization service.

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
navneet1v: The reader is converted above to segment reader..

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNNQuantizationStateReader.java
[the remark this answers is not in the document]
ryanbogan: I misunderstood, will push change shortly to be more efficient.

Vikasht34 (approved):
Functionally looks good to me.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNN990QuantizationStateWriter.java
```diff
@@ -0,0 +1,115 @@
...
+
+    /**
+     * Writes an index header
+     * @param segmentWriteState state containing segment information
+     * @throws IOException exception could be thrown while writing header
+     */
+    public void writeHeader(SegmentWriteState segmentWriteState) throws IOException {
+        CodecUtil.writeIndexHeader(
+            output,
+            "NativeEngines99KnnVectorsFormatQSData",
```
shatejas, line 66: `NativeEngines990KnnVectorsFormatQSData` to be consistent

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriter.java
shatejas: The problem here is you are initializing the quantization file and writing the header even when quantization is not required. Can you push it to flush and merge so we don't end up writing an empty file with header and footer?

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNN990QuantizationStateWriter.java
naveentatikonda, line 66: Should this be named as `99` or `990` like `NativeEngines990KnnVectorsFormatQSData` to keep it consistent with `NativeEngines990KnnVectorsFormat` ?

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNN990QuantizationStateWriter.java
```diff
@@ -0,0 +1,115 @@
...
+
+    @AllArgsConstructor
+    private static class FieldQuantizationState {
+        final int fieldNumber;
+        final byte[] stateBytes;
+        @Setter
+        Long position;
+    }
+
+    public void closeOutput() throws IOException {
```
shatejas, line 112: Maybe implement Closeable so you can leverage try with resources wherever needed

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNN990QuantizationStateWriter.java
naveentatikonda: Also, create a constant for that

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNN990QuantizationStateWriter.java
```diff
@@ -0,0 +1,115 @@
...
+        long position = output.getFilePointer();
+        output.writeBytes(stateBytes, stateBytes.length);
+        fieldQuantizationStates.add(new FieldQuantizationState(fieldNumber, stateBytes, position));
+    }
+
+    /**
+     * Writes index footer and other index information for parsing later
+     * @throws IOException could be thrown while writing
+     */
+    public void writeFooter() throws IOException {
```
shatejas, line 91: thinking out loud here,

There is a small risk here that writing of footer is dependent on client. To safe guard that the footer is written you can check if its written in close (if this class implement closeable). One of the downsides I see there is that if its used with try with resources the footer will be written even in exception cases and I am not sure if thats considered as a corrupt file

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
```diff
@@ -159,4 +194,31 @@ public void close() throws IOException {
     public long ramBytesUsed() {
         return flatVectorsReader.ramBytesUsed();
     }
+
+    private void primeQuantizationStateCache() throws IOException {
```
naveentatikonda, line 198: Can you add a comment explaining what we are doing in this method

shatejas: So the method is priming the quantization state cache and that was the intent of the name change

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/QuantizationConfigKNNCollector.java
```diff
@@ -0,0 +1,62 @@
...
+ */
+@Setter
+@Getter
+public class QuantizationConfigKNNCollector implements KnnCollector {
+
+    private QuantizationState quantizationState;
+
+    @Override
+    public boolean earlyTerminated() {
+        throw new UnsupportedOperationException("Search functionality using codec is not supported with Native Engine Reader");
```
naveentatikonda, line 25: Can we create a constant for this error message in this class and use it in all the methods ?

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNN990QuantizationStateWriter.java
navneet1v: +1

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriter.java
```diff
@@ -53,10 +51,17 @@ public class NativeEngines990KnnVectorsWriter extends KnnVectorsWriter {
...
     private final FlatVectorsWriter flatVectorsWriter;
+    private KNN990QuantizationStateWriter quantizationStateWriter;
     private final List<NativeEngineFieldVectorsWriter<?>> fields = new ArrayList<>();
     private boolean finished;
     private final QuantizationService quantizationService = QuantizationService.getInstance();
 
+    public NativeEngines990KnnVectorsWriter(SegmentWriteState segmentWriteState, FlatVectorsWriter flatVectorsWriter) {
+        this.segmentWriteState = segmentWriteState;
+        this.flatVectorsWriter = flatVectorsWriter;
+        this.quantizationStateWriter = null;
```
shatejas, line 62: nit: default is null so not needed

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriter.java
```diff
@@ -116,6 +126,7 @@ public void finish() throws IOException {
             throw new IllegalStateException("NativeEnginesKNNVectorsWriter is already finished");
         }
         finished = true;
+        quantizationStateWriter.writeFooter();
```
shatejas, line 129: Needs a null check

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriter.java
```diff
@@ -134,6 +145,7 @@ public void finish() throws IOException {
      */
     @Override
     public void close() throws IOException {
+        quantizationStateWriter.closeOutput();
```
shatejas, line 148: needs a null check

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriter.java
```diff
@@ -79,6 +84,9 @@ public KnnFieldVectorsWriter<?> addField(final FieldInfo fieldInfo) throws IOExc
     @Override
     public void flush(int maxDoc, final Sorter.DocMap sortMap) throws IOException {
         flatVectorsWriter.flush(maxDoc, sortMap);
+
+        initQuantizationStateWriterIfNecessary();
```
shatejas, line 88: So this will still open and write the header without checking if quantization is necessary
- you can check it inside the method `or`
- you do it after line 252 as the first step

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNN990QuantizationStateReader.java
```diff
@@ -0,0 +1,144 @@
...
+
+            int numFields = getNumFields(input);
+
+            // Read each field's metadata from the index section and then read bytes
+            for (int i = 0; i < numFields; i++) {
+                int fieldNumber = input.readInt();
+                int length = input.readInt();
+                long position = input.readVLong();
+                byte[] stateBytes = readStateBytes(input, position, length);
+                String fieldName = state.fieldInfos.fieldInfo(fieldNumber).getName();
```
navneet1v, line 65: what will happen in the nest field case?

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriter.java
navneet1v: This is making me think why ITs are not failing, as QSW will be null for nmslib and faiss engine HNSW

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
```diff
@@ -159,4 +194,31 @@ public void close() throws IOException {
...
+        quantizationStateCacheKeyPerField = new HashMap<>();
+        Map<String, byte[]> stateMap = KNN990QuantizationStateReader.read(segmentReadState);
+        for (Map.Entry<String, byte[]> entry : stateMap.entrySet()) {
```
navneet1v, line 201: this is causing the BWC to fail. So right now without checking the file is present or not you are reading the QS. In the read function we should first check if file is present or not. Then only we should open the file.

navneet1v:
```
Caused by: java.nio.file.NoSuchFileException: /home/runner/work/k-NN/k-NN/qa/restart-upgrade/build/testclusters/knnBwcCluster-restart-0/data/nodes/0/indices/7GO1YWCXQt62LsLnXY0JoQ/0/index/_2j_NativeEngines990KnnVectorsFormat_0.osknnqstate
»  	at java.base/sun.nio.fs.UnixException.translateToIOException(UnixException.java:92) ~[?:?]
»  	at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:106) ~[?:?]
»  	at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:111) ~[?:?]
»  	at java.base/sun.nio.fs.UnixFileSystemProvider.newFileChannel(UnixFileSystemProvider.java:224) ~[?:?]
»  	at java.base/java.nio.channels.FileChannel.open(FileChannel.java:309) ~[?:?]
»  	at java.base/java.nio.channels.FileChannel.open(FileChannel.java:369) ~[?:?]
»  	at org.apache.lucene.store.MemorySegmentIndexInputProvider.openInput(MemorySegmentIndexInputProvider.java:58) ~[lucene-core-9.12.0-snapshot-847316d.jar:9.12.0-snapshot-847316d 847316dd1eec2258ccf5da52dca51c01b9df1c4c - 2024-06-27 17:07:28]
»  	at org.apache.lucene.store.MMapDirectory.openInput(MMapDirectory.java:330) ~[lucene-core-9.12.0-snapshot-847316d.jar:9.12.0-snapshot-847316d 847316dd1eec2258ccf5da52dca51c01b9df1c4c - 2024-06-27 17:07:28]
»  	at org.opensearch.index.store.FsDirectoryFactory$HybridDirectory.openInput(FsDirectoryFactory.java:164) ~[opensearch-3.0.0-SNAPSHOT.jar:3.0.0-SNAPSHOT]
»  	at org.apache.lucene.store.FilterDirectory.openInput(FilterDirectory.java:101) ~[lucene-core-9.12.0-snapshot-847316d.jar:9.12.0-snapshot-847316d 847316dd1eec2258ccf5da52dca51c01b9df1c4c - 2024-06-27 17:07:28]
»  	at org.apache.lucene.store.FilterDirectory.openInput(FilterDirectory.java:101) ~[lucene-core-9.12.0-snapshot-847316d.jar:9.12.0-snapshot-847316d 847316dd1eec2258ccf5da52dca51c01b9df1c4c - 2024-06-27 17:07:28]
»  	at org.opensearch.knn.index.codec.KNN990Codec.KNN990QuantizationStateReader.read(KNN990QuantizationStateReader.java:54) ~[?:?]
»  	at org.opensearch.knn.index.codec.KNN990Codec.NativeEngines990KnnVectorsReader.primeQuantizationStateCache(NativeEngines990KnnVectorsReader.java:200) ~[?:?]
»  	at org.opensearch.knn.index.codec.KNN990Codec.NativeEngines990KnnVectorsReader.<init>(NativeEngines990KnnVectorsReader.java:53) ~[?:?]
»  	at org.opensearch.knn.index.codec.KNN990Codec.NativeEngines990KnnVectorsFormat.fieldsReader(NativeEngines990KnnVectorsFormat.java:61) ~[?:?]
»  	at org.apache.lucene.codecs.perfield.PerFieldKnnVectorsFormat$FieldsReader.<init>(PerFieldKnnVectorsFormat.java:222) ~[lucene-core-9.12.0-snapshot-847316d.jar:9.12.0-
```

ryanbogan: Added

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriter.java
ryanbogan: Added

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
Vikasht34: Got caught in Unit test only ...It's fixed now.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNN990QuantizationStateReader.java
ryanbogan: Checking

ryanbogan: It works normally

ryanbogan: The quantization state was written and read correctly

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNN990QuantizationStateReader.java
```diff
@@ -0,0 +1,146 @@
...
+                readQuantizationStateInfos.put(fieldName, stateBytes);
+            }
+        } catch (Exception e) {
+            return readQuantizationStateInfos;
```
shatejas, line 69: I am assuming an exception is expected if the file is not found. In that case the catch all is risky here, change to catch a specific exception and then return `Collections.emptyMap()`. Creating a map on line 52 is not ideal as it allocates memory, `Collections.emptyMap` will not. Probably move the `new map()` to line 55. 

For all other exceptions log an error and please throw

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNN990QuantizationStateReader.java
```diff
@@ -0,0 +1,146 @@
...
+        input.seek(position);
+        byte[] stateBytes = new byte[length];
+        input.readBytes(stateBytes, 0, length);
+        return stateBytes;
+    }
+
+    @VisibleForTesting
+    static String getQuantizationStateFileName(SegmentReadState state) {
+        return IndexFileNames.segmentFileName(state.segmentInfo.name, state.segmentSuffix, KNNConstants.QUANTIZATION_STATE_FILE_SUFFIX);
+    }
```
shatejas, line 145: nit: See if you can have a common place for reader and writer. Idea is to have a common place for both to avoid errors

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNN990QuantizationStateReader.java
```diff
@@ -0,0 +1,146 @@
...
+     * @param readConfig a config class that contains necessary information for reading the state
+     * @return quantization state
+     */
+    public static QuantizationState read(QuantizationStateReadConfig readConfig) throws IOException {
+        SegmentReadState segmentReadState = readConfig.getSegmentReadState();
+        String field = readConfig.getField();
+        String quantizationStateFileName = getQuantizationStateFileName(segmentReadState);
+        int fieldNumber = segmentReadState.fieldInfos.fieldInfo(field).getFieldNumber();
+
+        try (IndexInput input = segmentReadState.directory.openInput(quantizationStateFileName, IOContext.READ)) {
```
shatejas, line 85: This is a try without a catch, The same applies here, return null for a specific exception and throw for others

shatejas (commented):
Looks good overall, some minor changes

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNN990QuantizationStateReader.java
Vikasht34: I am thinking to return empty map for all type of errors , and let's not load the cache , we really don't want to throw the excpetion and break and query flow. let it run usual.

Collections.emptyMap() also allocates memory in the same way new does , the only differnce is it's singleton and does not allocate every time , but here it does not matter.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNN990QuantizationStateReader.java
Vikasht34: Ideally we can have it , but this piece of code is getting called from KnnWeight only when QuantizationParm is not null.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNN990QuantizationStateReader.java
ryanbogan: I thought about this, but one uses SegmentWriteState and the other uses SegmentReadState

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNN990QuantizationStateReader.java
shatejas: > I am thinking to return empty map for all type of errors , and let's not load the cache , we really don't want to throw the excpetion and break and query flow. let it run usual.

Why? Silent failures are not good

> Collections.emptyMap() also allocates memory in the same way new does , the only differnce is it's singleton and does not allocate every time , but here it does not matter.

For every request its creating new map which allocates memory, so we need that singleton

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNN990QuantizationStateReader.java
shatejas: Same for this https://github.com/opensearch-project/k-NN/pull/1997#discussion_r1744563505

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNN990QuantizationStateReader.java
Vikasht34: Read functionality is getting called from NativeReader Constructor for both cases , like with quantization and without , without quantization we are returing empty map and with also in case of any error empty hash map basically don't load the hash map to load cache. Which in my case it's fine ..becuase flow is same. 

Segment reader is opened only once , and it's not going to open for each request , so it does not matter.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNN990QuantizationStateReader.java
Vikasht34: File not found excpetion will never be for this case , as in KNN wight we already have check before calling this function

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNN990QuantizationStateReader.java
shatejas: 1. This is a public function, so the code needs to be defensive
2. The segment reader is opened once per request. If there is enough requests we are unnecessarily allocating memory when not needed. So it might add up.

Lets make sure we aren't assigning memory where its not needed

navneet1v: @Vikasht34 and @ryanbogan if we are catching the exception we should have logs too on what was the exception. Lets say if there is genuinely some error, then just eating up the exception is not correct.

More over, we are using IndexInput's exception to say if the file is present or not. Why not just get all the segmentfiles from the segment info and check if the desired file is present or not.

navneet1v: I would suggest to have another check in both the read functions which check if QS file is present or not. and then safely return an empty map. and then we don't need to catch all the exceptions.

Vikasht34: Make Sense ...

ryanbogan: Will address in follow-up

ryanbogan:
Adding skip-changelog because entry is in the release notes instead

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/QuantizationConfigKNNCollector.java
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * Collector used for passing the quantization state during query flow.
+ */
+@Setter
+@Getter
+public class QuantizationConfigKNNCollector implements KnnCollector {
+
+    private QuantizationState quantizationState;
+
+    private final String NATIVE_ENGINE_SEARCH_ERROR_MESSAGE = "Search functionality using codec is not supported with Native Engine Reader";
```
naveentatikonda, line 23: use `static`

shatejas (approved):

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/QuantizationConfigKNNCollector.java
ryanbogan: Will add to follow-up

naveentatikonda (approved):

jmazanec15 (approved):

navneet1v:
Merging the PR to unblock others. The test which is failing will be fixed in next iteration

PR #1999, opened 2024-08-21
Fix string for qframe parser key to not use String.format

jmazanec15:
### Description
Fixes String for to/from csv parsing to not use String.format. This will ensure we do not run into Locale issues. 

### DIFF
--- a/src/main/java/org/opensearch/knn/index/engine/qframe/QuantizationConfigParser.java
+++ b/src/main/java/org/opensearch/knn/index/engine/qframe/QuantizationConfigParser.java
@@ -33,14 +33,8 @@ public class QuantizationConfigParser {
             || quantizationConfig.getQuantizationType() == null) {
             return "";
         }
-        return String.format(
-            Locale.ROOT,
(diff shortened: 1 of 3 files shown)


src/test/java/org/opensearch/knn/jni/JNIServiceTests.java
```diff
@@ -60,6 +61,7 @@
 import static org.opensearch.knn.common.KNNConstants.NAME;
 import static org.opensearch.knn.common.KNNConstants.PARAMETERS;
 
+@Ignore
```
navneet1v, line 64: why this Ignore?

jmazanec15: oops skipping for tests. Let me fix

jmazanec15: fixed

ryanbogan (approved):

navneet1v (approved):

PR #2005, opened 2024-08-27
Add model version to model metadata and change model metadata reads to be from cluster metadata

ryanbogan:
### Description
Adds the OpenSearch version with which a model is created to its metadata.  Additionally, changes modelMetadata reads to read from cluster metadata in order to eliminate an unnecessary get model call.

This change helps prevent null pointer exceptions, since we can tell which variables are present in the metadata based on the version. Additionally, it is required for re-scoring features that are being added as part of disk-based search.

### Check List


src/main/java/org/opensearch/knn/common/KNNConstants.java
```diff
@@ -73,6 +73,7 @@ public class KNNConstants {
     public static final VectorDataType DEFAULT_VECTOR_DATA_TYPE_FIELD = VectorDataType.FLOAT;
 
     public static final String RADIAL_SEARCH_KEY = "radial_search";
+    public static final String MODEL_VERSION = "model_version";
```
naveentatikonda, line 76: What is the purpose of this model version ?

ryanbogan: It helps prevent null pointer exceptions, since we can tell which variables are present in the metadata based on the version.  Additionally, it is required for re-scoring features that are being added as part of disk-based search.

naveentatikonda (dismissed):

src/main/java/org/opensearch/knn/indices/ModelMetadata.java
```diff
@@ -304,13 +323,14 @@ public static ModelMetadata fromString(String modelMetadataString) {
         String[] modelMetadataArray = modelMetadataString.split(DELIMITER, -1);
         int length = modelMetadataArray.length;
 
-        if (length < 7 || length > 10) {
+        if (length < 7 || length > 11) {
```
jmazanec15, line 326: Are we able to get any more information around the model version based on what is stored in model metadata? Itd be useful to know what models are created after 2.14

src/main/java/org/opensearch/knn/training/TrainingJob.java
```diff
@@ -90,7 +91,8 @@ public TrainingJob(
                 "",
                 nodeAssignment,
                 knnMethodContext.getMethodComponentContext(),
-                knnMethodConfigContext.getVectorDataType()
+                knnMethodConfigContext.getVectorDataType(),
+                Version.CURRENT
```
jmazanec15, line 95: Can we set version during knnMethodConfigContext configuration?

src/main/java/org/opensearch/knn/indices/ModelMetadata.java
ryanbogan: For anything going forward, we should be able to, but anything before 2.17 is going to have an empty version

src/main/java/org/opensearch/knn/training/TrainingJob.java
ryanbogan: Oh I didn't even see that, we should just be able to call getVersionCreated()

jmazanec15:
@ryanbogan I think we might need to update https://github.com/opensearch-project/k-NN/blob/main/src/main/resources/mappings/model-index.json

ryanbogan:
> @ryanbogan I think we might need to update https://github.com/opensearch-project/k-NN/blob/main/src/main/resources/mappings/model-index.json

Ah good catch, will push fix

navneet1v (approved):

ryanbogan:
Skipping changelog because change is in release notes

src/main/java/org/opensearch/knn/indices/ModelUtil.java
```diff
@@ -48,8 +48,9 @@ public static ModelMetadata getModelMetadata(final String modelId) {
         if (StringUtils.isEmpty(modelId)) {
             return null;
         }
-        final Model model = ModelCache.getInstance().get(modelId);
-        final ModelMetadata modelMetadata = model.getModelMetadata();
+        ModelDao modelDao = ModelDao.OpenSearchKNNModelDao.getInstance();
+        final ModelMetadata modelMetadata = modelDao.getMetadata(modelId);
+        System.out.println(modelMetadata.toString());
```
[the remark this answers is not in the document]
jmazanec15: (I removed it)

jmazanec15 (dismissed):
LGTM

Vikasht34 (approved):
LGTM

jmazanec15 (approved):

PR #2015, opened 2024-08-30
Fix memory overflow caused by cache behavior

### LINKED ISSUE #1582: Investigate rearchitecture of the native memory circuit breaker
## Description

Wanted to put out an issue to track potentially rearch our current circuit breaker/native memory management system. Currently, our circuit breaking logic has a couple flaws and can be confusing to users. I think that we should consider re-architecting it to be more in line with the JVM heap based circuit breaker that OpenSearch uses (https://github.com/opensearch-project/OpenSearch/blob/main/libs/core/src/main/java/org/opensearch/core/common/breaker/CircuitBreaker.java)

A couple issues I have with our circuit breaker:
1. Caching allocations and limiting memory is tied together in a guava cache. This makes it difficult to impose custom limits per-allocation type
2. All allocation types sit in a single allocation manager. This can make it tricky for complex management like we had in #1507. Also, it makes it very difficult if another plugin perhaps wanted to introduce their own native memory allocation type
3. Guava cache does not free memory allocations immediately when they are evicted. This can lead to going over circuit breaker limit in native memory consumption
4. We do not track large allocations during indexing time
5. We require consumers of native memory allocations to acquire locks. This can make the code very hard to manage and requires delicate handling of exceptions etc.

## Proposed Solution
I think we should investigate migrating towards the approach OpenSearch has taken for the jvm circuit breaker limits. This would allow:
1. Through the a hierarchical mem management system, we could separate management of diff memory allocations while still imposing a global limit. This would make the system more extendable
2. Per allocation type memory management would allow us to take more control over memory free life cycle. This would allow us to migrate away from problems associated with guava cache freeing.
3. Provide a more consistent OpenSearch user experience by following conventions set for existing circuit breaker

kotwanikunal:
### Description
- Adds in a force evict mechanism before entry to prevent memory overflow for native memory usage within the cache
- This solution works in tandem with `jemalloc` enabled. In case of default `malloc`, there are potential delays with memory free up operations leading to similar scenarios
- The solution has been tested and benchmarked - described in the Testing section below

### Problem
- The problem stems from the behavior of the cache where in entries of the cache might not be cleaned up as soon as they are evictable
- This delay between new entry and cleanup can, in our case, cause the memory to overflow - leading to a node failure

### Solution
- This solution aims at maintaining an additional recency list within the cache manager
- On every `get` call - 
  - in case of a cache hit - it just marks the key as recent, returns the value
  - in case of a cache miss - it adds the key to the recency list, loads the item into the cache and returns
  - in case of a cache miss & a potential overflow - eviction process kicks off before the item is loaded into the cache 

### Testing

1. **Case: Enough memory available**
  - Scenario description: Indexed 1M 768d vectors on a 3 node cluster with 16GB memory - 8GB heap, 8GB for graph files, ran an OSB query load with and without the fix
  - TLDR; Not a ton of deviation from current code in terms of raw numbers. Results below

| Index Type    | Dimension | Number Of Vectors | Min Throughput | Mean Throughput | Median Throughput | Max Throughput | 50th percentile latency | 90th percentile latency | 99th percentile latency | 100th percentile latency | 50th percentile service time | 90th percentile service time | 99th percentile service time | 100th percentile service time | Mean recall@k | Mean recall@1 |
| ------------- | --------- | ----------------- | -------------- | --------------- | ----------------- | -------------- | ----------------------- | ----------------------- | ----------------------- | ------------------------ | ---------------------------- | ---------------------------- | ---------------------------- | ----------------------------- | ------------- | ------------- |
| Current Code  | 768       | 1,000,000         | 2.46           | 2.68            | 2.69              | 2.74           | 358.268                 | 382.036                 | 414.636                 | 482.332                  | 358.268                      | 382.036                      | 414.636                      | 482.332                       | 0.98          | 1             |
| Code with Fix | 768       | 1,000,000         | 2.55           | 2.71            | 2.7               | 2.75           | 363.62                  | 398.338                 | 432.811                 | 447.278                  | 363.62                       | 398.338                      | 432.811                      | 447.278                       | 0.98          | 1             |

2. **Case: Heavy query load scenario**
  - Scenario: Created 3 indexes (2 FAISS, 1 NMSLIB), Indexed 1M 768d vectors on a 3 node cluster with 16GB memory - 8GB heap, 8GB for graph files, ran 3 parallel query workloads for all the indexes at once. 
  - TLDR; **Without the fix** - noticed node crashes. **With the fix** - there was an uptick in error due to timeouts, latency increased but no node crash was observed

| Index Type      | Dimension | Number Of Vectors | Min Throughput | Mean Throughput | Median Throughput | Max Throughput | 50th percentile latency | 90th percentile latency | 99th percentile latency | 100th percentile latency | 50th percentile service time | 90th percentile service time | 99th percentile service time | 100th percentile service time | Error rate % | Mean recall@k | Mean recall@1 |
| --------------- | --------- | ----------------- | -------------- | --------------- | ----------------- | -------------- | ----------------------- | ----------------------- | ----------------------- | ------------------------ | ---------------------------- | ---------------------------- | ---------------------------- | ----------------------------- | ------------ | ------------- | ------------- |
| Index1 with fix | 768       | 1,000,000         | 0.05           | 0.06            | 0.06              | 0.09           | 13228.9                 | 19363.3                 | 24602.8                 | 35710.1                  | 13228.9                      | 19363.3                      | 24602.8                      | 35710.1                       | 11           | 0.99          | 1             |
| Index2 with fix | 768       | 1,000,000         | 0.1            | 0.25            | 0.14              | 1.02           | 11728.5                 | 15634.2                 | 23228.2                 | 34283.1                  | 11728.5                      | 15634.2                      | 23228.2                      | 34283.1                       | 0            | 0.98          | 1             |
| Index3 with fix | 768       | 1,000,000         | 0.06           | 0.07            | 0.07              | 0.09           | 12575                   | 17641.8                 | 25928.4                 | 27722.8                  | 12575                        | 17641.8                      | 25928.4                      | 27722.8                       | 3            | 0.98          | 1             |

### Related Issues
Partially resolves #1582 (Case 3)

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/common/featureflags/KNNFeatureFlags.java
+++ b/src/main/java/org/opensearch/knn/common/featureflags/KNNFeatureFlags.java
@@ -1,0 +1,49 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ */
+
+package org.opensearch.knn.common.featureflags;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.ImmutableList;
+import lombok.experimental.UtilityClass;
+import org.opensearch.common.Booleans;
+import org.opensearch.common.settings.Setting;
+import org.opensearch.knn.index.KNNSettings;
+
+import java.util.List;
+
+import static org.opensearch.common.settings.Setting.Property.Dynamic;
+import static org.opensearch.common.settings.Setting.Property.NodeScope;
+
+/**
+ * Class to manage KNN feature flags
+ */
+@UtilityClass
+public class KNNFeatureFlags {
+
+    // Feature flags
+    private static final String KNN_FORCE_EVICT_CACHE_ENABLED = "knn.feature.cache.force_evict.enabled";
+
+    @VisibleForTesting
+    public static final Setting<Boolean> KNN_FORCE_EVICT_CACHE_ENABLED_SETTING = Setting.boolSetting(
+        KNN_FORCE_EVICT_CACHE_ENABLED,
+        false,
+        NodeScope,
+        Dynamic
+    );
+
+    public static List<Setting<?>> getFeatureFlags() {
+        return ImmutableList.of(KNN_FORCE_EVICT_CACHE_ENABLED_SETTING);
+    }
+
+    /**
+     * Checks if force evict for cache is enabled by executing a check against cluster settings
+     * @return true if force evict setting is set to true
+     */
+    public static boolean isForceEvictCacheEnabled() {
+        return Booleans.parseBoolean(KNNSettings.state().getSettingValue(KNN_FORCE_EVICT_CACHE_ENABLED).toString(), false);
+    }
+}
--- a/src/main/java/org/opensearch/knn/index/memory/NativeMemoryAllocation.java
+++ b/src/main/java/org/opensearch/knn/index/memory/NativeMemoryAllocation.java
@@ -13,6 +13,7 @@ package org.opensearch.knn.index.memory;
 
 import lombok.Getter;
 import org.apache.lucene.index.LeafReaderContext;
+import org.opensearch.knn.common.featureflags.KNNFeatureFlags;
 import org.opensearch.knn.index.VectorDataType;
 import org.opensearch.knn.index.query.KNNWeight;
 import org.opensearch.knn.jni.JNIService;
@@ -161,11 +162,19 @@ public interface NativeMemoryAllocation {
 
         @Override
         public void close() {
-            executor.execute(() -> {
+            Runnable onClose = () -> {
                 writeLock();
                 cleanup();
                 writeUnlock();
-            });
+            };
+
+            // The close operation needs to be blocking to prevent overflow
+            // This blocks any entry until the close has completed, preventing creation before close scenarios
+            if (KNNFeatureFlags.isForceEvictCacheEnabled()) {
+                onClose.run();
+            } else {
+                executor.execute(onClose);
+            }
         }
 
         private void cleanup() {
--- a/src/main/java/org/opensearch/knn/index/memory/NativeMemoryCacheManager.java
+++ b/src/main/java/org/opensearch/knn/index/memory/NativeMemoryCacheManager.java
@@ -21,13 +21,17 @@ import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.opensearch.common.unit.TimeValue;
 import org.opensearch.knn.common.exception.OutOfNativeMemoryException;
+import org.opensearch.knn.common.featureflags.KNNFeatureFlags;
 import org.opensearch.knn.index.KNNSettings;
 import org.opensearch.knn.plugin.stats.StatNames;
 
 import java.io.Closeable;
+import java.util.Deque;
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.Map;
 import java.util.Optional;
+import java.util.concurrent.ConcurrentLinkedDeque;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
@@ -45,6 +49,7 @@ public class NativeMemoryCacheManager implements Closeable {
     private static NativeMemoryCacheManager INSTANCE;
 
     private Cache<String, NativeMemoryAllocation> cache;
+    private Deque<String> accessRecencyQueue;
     private final ExecutorService executor;
     private AtomicBoolean cacheCapacityReached;
     private long maxWeight;
@@ -97,7 +102,7 @@ public class NativeMemoryCacheManager implements Closeable {
         }
 
         cacheCapacityReached = new AtomicBoolean(false);
-
+        accessRecencyQueue = new ConcurrentLinkedDeque<>();
         cache = cacheBuilder.build();
     }
 
@@ -301,7 +306,52 @@ public class NativeMemoryCacheManager implements Closeable {
             );
         }
 
-        return cache.get(nativeMemoryEntryContext.getKey(), nativeMemoryEntryContext::load);
+        if (KNNFeatureFlags.isForceEvictCacheEnabled()) {
+            // Utilizes a force eviction mechanism to free up memory before the entry can be added to the cache
+            // In case of a cache hit, the operation just updates the locally maintained recency list
+            // In case of a cache miss, least recently accessed entries are evicted in a blocking manner
+            // before the new entry can be added to the cache.
+            String key = nativeMemoryEntryContext.getKey();
+            NativeMemoryAllocation result = cache.getIfPresent(key);
+
+            // Cache Hit
+            // In case of a cache hit, moving the item to the end of the recency queue adds
+            // some overhead to the get operation. This can be optimized further to make this operation
+            // as lightweight as possible. Multiple approaches and their outcomes were documented
+            // before moving forward with the current solution.
+            // The details are outlined here: https://github.com/opensearch-project/k-NN/pull/2015#issuecomment-2327064680
+            if (result != null) {
+                accessRecencyQueue.remove(key);
+                accessRecencyQueue.addLast(key);
+                return result;
+            }
+
+            // Cache Miss
+            // Evict before put
+            synchronized (this) {
+                if (getCacheSizeInKilobytes() + nativeMemoryEntryContext.calculateSizeInKB() >= maxWeight) {
+                    Iterator<String> lruIterator = accessRecencyQueue.iterator();
+                    while (lruIterator.hasNext()
+                        && (getCacheSizeInKilobytes() + nativeMemoryEntryContext.calculateSizeInKB() >= maxWeight)) {
+
+                        String keyToRemove = lruIterator.next();
+                        NativeMemoryAllocation allocationToRemove = cache.getIfPresent(keyToRemove);
+                        if (allocationToRemove != null) {
+                            allocationToRemove.close();
+                            cache.invalidate(keyToRemove);
+                        }
+                        lruIterator.remove();
+                    }
+                }
+
+                result = cache.get(key, nativeMemoryEntryContext::load);
+                accessRecencyQueue.addLast(key);
+
+                return result;
+            }
+        } else {
+            return cache.get(nativeMemoryEntryContext.getKey(), nativeMemoryEntryContext::load);
+        }
     }
 
     /**
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 * Add script_fields context to KNNAllowlist [#1917] (https://github.com/opensearch-project/k-NN/pull/1917)
 * Fix graph merge stats size calculation [#1844](https://github.com/opensearch-project/k-NN/pull/1844)
 * Disallow a vector field to have an invalid character for a physical file name. [#1936](https://github.com/opensearch-project/k-NN/pull/1936)
+* Fix memory overflow caused by cache behavior [#2015](https://github.com/opensearch-project/k-NN/pull/2015)
 ### Infrastructure
 * Parallelize make to reduce build time [#2006] (https://github.com/opensearch-project/k-NN/pull/2006)
 ### Documentation
--- a/src/main/java/org/opensearch/knn/index/KNNSettings.java
+++ b/src/main/java/org/opensearch/knn/index/KNNSettings.java
@@ -34,14 +34,17 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
+import static java.util.stream.Collectors.toUnmodifiableMap;
 import static org.opensearch.common.settings.Setting.Property.Dynamic;
 import static org.opensearch.common.settings.Setting.Property.IndexScope;
 import static org.opensearch.common.settings.Setting.Property.NodeScope;
 import static org.opensearch.common.unit.MemorySizeValue.parseBytesSizeValueOrHeapRatio;
 import static org.opensearch.core.common.unit.ByteSizeValue.parseBytesSizeValue;
+import static org.opensearch.knn.common.featureflags.KNNFeatureFlags.getFeatureFlags;
 
 /**
  * This class defines
@@ -351,6 +354,9 @@ public class KNNSettings {
         }
     };
 
+    private final static Map<String, Setting<?>> FEATURE_FLAGS = getFeatureFlags().stream()
+        .collect(toUnmodifiableMap(Setting::getKey, Function.identity()));
+
     private ClusterService clusterService;
     private Client client;
 
@@ -388,7 +394,7 @@ public class KNNSettings {
             );
 
             NativeMemoryCacheManager.getInstance().rebuildCache(builder.build());
-        }, dynamicCacheSettings.values().stream().collect(Collectors.toUnmodifiableList()));
+        }, Stream.concat(dynamicCacheSettings.values().stream(), FEATURE_FLAGS.values().stream()).collect(Collectors.toUnmodifiableList()));
         clusterService.getClusterSettings().addSettingsUpdateConsumer(QUANTIZATION_STATE_CACHE_SIZE_LIMIT_SETTING, it -> {
             QuantizationStateCache.getInstance().setMaxCacheSizeInKB(it.getKb());
             QuantizationStateCache.getInstance().rebuildCache();
@@ -413,6 +419,10 @@ public class KNNSettings {
     private Setting<?> getSetting(String key) {
         if (dynamicCacheSettings.containsKey(key)) {
             return dynamicCacheSettings.get(key);
+        }
+
+        if (FEATURE_FLAGS.containsKey(key)) {
+            return FEATURE_FLAGS.get(key);
         }
 
         if (KNN_CIRCUIT_BREAKER_TRIGGERED.equals(key)) {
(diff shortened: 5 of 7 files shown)


src/main/java/org/opensearch/knn/index/memory/NativeMemoryAllocation.java
```diff
@@ -163,11 +164,19 @@ class IndexAllocation implements NativeMemoryAllocation {
...
+            Runnable onClose = () -> {
                 writeLock();
                 cleanup();
                 writeUnlock();
-            });
+            };
+
+            // The close operation needs to be blocking to prevent overflow
+            // This blocks any entry until the close has completed, preventing creation before close scenarios
+            if (KNNFeatureFlags.isForceEvictCacheEnabled()) {
```
heemin32, line 175: Have you tested with only this change without a code which try to evict before adding value to cache to see if node crash or not?

src/main/java/org/opensearch/knn/index/memory/NativeMemoryCacheManager.java
```diff
@@ -301,7 +306,47 @@ public NativeMemoryAllocation get(NativeMemoryEntryContext<?> nativeMemoryEntryC
...
+            // Utilizes a force eviction mechanism to free up memory before the entry can be added to the cache
+            // In case of a cache hit, the operation just updates the locally maintained recency list
+            // In case of a cache miss, least recently accessed entries are evicted in a blocking manner
+            // before the new entry can be added to the cache.
+            String key = nativeMemoryEntryContext.getKey();
+            NativeMemoryAllocation result = cache.getIfPresent(key);
+
+            // Cache Hit
+            if (result != null) {
+                accessRecencyQueue.remove(key);
```
heemin32, line 319: This is concerning part when there are many graphs in a single node. This remove is linear complexity and this code path will be hit for every query. I think the benchmark case might not have enough segment files to see the degradation.

src/main/java/org/opensearch/knn/index/memory/NativeMemoryCacheManager.java
```diff
@@ -301,7 +306,47 @@ public NativeMemoryAllocation get(NativeMemoryEntryContext<?> nativeMemoryEntryC
...
+                accessRecencyQueue.addLast(key);
```
heemin32, line 320: There is a chance that same key can be added more than once in the queue.

src/main/java/org/opensearch/knn/index/memory/NativeMemoryCacheManager.java
```diff
@@ -301,7 +306,47 @@ public NativeMemoryAllocation get(NativeMemoryEntryContext<?> nativeMemoryEntryC
...
+                    Iterator<String> lruIterator = accessRecencyQueue.iterator();
+                    while (lruIterator.hasNext()
+                        && (getCacheSizeInKilobytes() + nativeMemoryEntryContext.calculateSizeInKB() >= maxWeight)) {
+
+                        String keyToRemove = lruIterator.next();
+                        NativeMemoryAllocation allocationToRemove = cache.getIfPresent(keyToRemove);
+                        if (allocationToRemove != null) {
+                            allocationToRemove.close();
+                            cache.invalidate(keyToRemove);
+                            lruIterator.remove();
```
heemin32, line 337: If we remove key from dequeue only when it is available in cache, some key will stay in the dequeue forever.

src/main/java/org/opensearch/knn/index/memory/NativeMemoryAllocation.java
```diff
@@ -163,11 +164,19 @@ class IndexAllocation implements NativeMemoryAllocation {
 
         @Override
         public void close() {
-            executor.execute(() -> {
+            Runnable onClose = () -> {
```
jmazanec15, line 167: Oh I like this - I was always wondering if we actually need to run this in an executor context. Anecdotally from you testing, how long would this method typically run?

src/main/java/org/opensearch/knn/index/memory/NativeMemoryCacheManager.java
```diff
@@ -97,7 +102,7 @@ private void initialize(NativeMemoryCacheManagerDto nativeMemoryCacheDTO) {
         }
 
         cacheCapacityReached = new AtomicBoolean(false);
-
+        accessRecencyQueue = new ConcurrentLinkedDeque<>();
```
jmazanec15, line 105: So - why cant we rely on guava cache to do this? I forget some of the details around how size based evictions take place.

src/main/java/org/opensearch/knn/index/memory/NativeMemoryAllocation.java
kotwanikunal: Yes. As I pointed out in the testing section - the nodes did crash.

src/main/java/org/opensearch/knn/index/memory/NativeMemoryCacheManager.java
kotwanikunal: I think most of your concerns are around queue efficiency/operations. 

Let me see if I can replace it to a ordered set and make it work good enough. We do not have an out of the box concurrent ordered set solution.
We can always use `Collections.synchronizedSet(new LinkedHashSet<>())` - but the efficiency of wrapped concurrent collections is not as efficient.

src/main/java/org/opensearch/knn/index/memory/NativeMemoryAllocation.java
kotwanikunal: I did not run numbers specifically for this block. Do you want me to get those numbers?

kotwanikunal: The other general problem with this today is that when the cache calls the `onRemoval` method for an entry - based on the Guava's documentation - it is supposed to be blocking. 
So when the cache is trying to clean up - forking the operation on this thread enables the cache to have a different usage metric compared to what is held by native memory indexes, further exacerbating the overflow problem.

src/main/java/org/opensearch/knn/index/memory/NativeMemoryCacheManager.java
```diff
@@ -301,7 +306,47 @@ public NativeMemoryAllocation get(NativeMemoryEntryContext<?> nativeMemoryEntryC
...
+                if (getCacheSizeInKilobytes() + nativeMemoryEntryContext.calculateSizeInKB() >= maxWeight) {
+                    Iterator<String> lruIterator = accessRecencyQueue.iterator();
+                    while (lruIterator.hasNext()
+                        && (getCacheSizeInKilobytes() + nativeMemoryEntryContext.calculateSizeInKB() >= maxWeight)) {
+
+                        String keyToRemove = lruIterator.next();
+                        NativeMemoryAllocation allocationToRemove = cache.getIfPresent(keyToRemove);
+                        if (allocationToRemove != null) {
+                            allocationToRemove.close();
+                            cache.invalidate(keyToRemove);
```
heemin32, line 336: Will this call `close()` one more time by Guava?

kotwanikunal: Yes. Which should be fine since `close()` is idempotent. 

https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/memory/NativeMemoryAllocation.java#L173-L177

src/main/java/org/opensearch/knn/index/memory/NativeMemoryCacheManager.java
heemin32: Can we consider utilizing removal listener?

navneet1v: +1 on @heemin32 comment. Is there a way we can avoid having our own queue/set and use removal listener from Guava cache? Does Guava cache provides a sync way to remove the items from cache?

src/main/java/org/opensearch/knn/index/memory/NativeMemoryCacheManager.java
navneet1v: There should not be the case where 1 key is added more than once in the queue.

But looking at the current code I think 319 and 320 should be in a sync block so that only 1 thread is updating the key Even when the queue is `ConcurrentLinkedDeque`, as this is a transaction on the key

src/main/java/org/opensearch/knn/index/memory/NativeMemoryCacheManager.java
navneet1v: +1. Even if there is not allocation present on the key the key should be removed from the queue.

src/main/java/org/opensearch/knn/index/memory/NativeMemoryCacheManager.java
kotwanikunal: That is the root of our problems - there are currently two async operations that happen for cache entry removal - 

1. The notification from Guava is not guaranteed to be before entry i.e the object cleanup can happen after an entry is added
2. The `close` on `NativeMemoryAllocation` objects is forked off into a separate thread (`onRemoval` from the cache is supposed to be sync with the removal flow)

This PR solves both of these problems. If we could utilize the listener directly - we would not be in this problem in the first place :) 

Details: https://github.com/google/guava/wiki/CachesExplained#when-does-cleanup-happen)

src/main/java/org/opensearch/knn/index/memory/NativeMemoryCacheManager.java
kotwanikunal: Ack. This will be updated with the next rev.

src/main/java/org/opensearch/knn/index/memory/NativeMemoryCacheManager.java
heemin32: I think removal listener should be called earlier enough unlike cleanup? Can we test it?
```
If your cache should not grow beyond a certain size, just use [CacheBuilder.maximumSize(long)](https://guava.dev/releases/snapshot/api/docs/com/google/common/cache/CacheBuilder.html#maximumSize-long-). The cache will try to evict entries that haven't been used recently or very often. Warning: the cache may evict entries before this limit is exceeded -- typically when the cache size is approaching the limit.
```

```
You may specify a removal listener for your cache to perform some operation when an entry is removed, via [CacheBuilder.removalListener(RemovalListener)](https://guava.dev/releases/snapshot/api/docs/com/google/common/cache/CacheBuilder.html#removalListener-com.google.common.cache.RemovalListener-). The [RemovalListener](https://guava.dev/releases/snapshot/api/docs/com/google/common/cache/RemovalListener.html) gets passed a [RemovalNotification](https://guava.dev/releases/snapshot/api/docs/com/google/common/cache/RemovalNotification.html), which specifies the [RemovalCause](https://guava.dev/releases/snapshot/api/docs/com/google/common/cache/RemovalCause.html), key, and value.
```

src/main/java/org/opensearch/knn/index/memory/NativeMemoryCacheManager.java
kotwanikunal: https://github.com/opensearch-project/k-NN/pull/2015#discussion_r1742382795

src/main/java/org/opensearch/knn/index/memory/NativeMemoryAllocation.java
kotwanikunal: Ran some numbers - I could only get averages.

With default code (current code) - 4.26076555024 ms averaged across nodes 
With modified code - 1.63179895861 ms averaged across nodes  (Possibly because the block is sync already)

kotwanikunal:
Ran some tests for LinkedHashSet vs the current queue implementation. Results:

| Index Type              | Dimension | Number Of Vectors | Min Throughput | Mean Throughput | Median Throughput | Max Throughput | 50th percentile latency | 90th percentile latency | 99th percentile latency | 100th percentile latency | Error rate % | Mean recall@k | Mean recall@1 |
| ----------------------- | --------- | ----------------- | -------------- | --------------- | ----------------- | -------------- | ----------------------- | ----------------------- | ----------------------- | ------------------------ | ------------ | ------------- | ------------- |
| Single Query Queue      | 768       | 1,000,000         | 1.28           | 2.37            | 2.46              | 2.58           | 368.768                 | 417.358                 | 483.694                 | 761.651                  | 0            | 0.98          | 0.97          |
| Single Query Sync Set   | 768       | 1,000,000         | 1.1            | 1.97            | 2.07              | 2.3            | 407.133                 | 516.032                 | 658.99                  | 889.347                  | 0            | 0.98          | 0.97          |
|                         |           |                   |                |                 |                   |                |                         |                         |                         |                          |              |               |               |
| Parallel Query Queue    | 768       | 1,000,000         | 0.07           | 0.07            | 0.07              | 0.09           | 13465.6                 | 18455.7                 | 31858                   | 33136.6                  | 0            | 0.99          | 1             |
| Parallel Query Sync Set | 768       | 1,000,000         | 0.06           | 0.07            | 0.06              | 0.08           | 13937                   | 18936.5                 | 28190.5                 | 28804.4                  | 8            | 0.99          | 1             |

src/main/java/org/opensearch/knn/index/memory/NativeMemoryCacheManager.java
kotwanikunal: Ran some more tests for the comparison between the two approaches (set vs optimized concurrent queue) - https://github.com/opensearch-project/k-NN/pull/2015#issuecomment-2327064680

It does look like the concurrent optimized version of the queue helps with perf.

kotwanikunal: @heemin32 We do have an `onRemoval` listener attached: [Ref code](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/memory/NativeMemoryCacheManager.java#L383-L394)

The problem is that this is called after an entry is made into the cache - so there is still a potential for memory overflow - leading to node drops.

To validate this - I tested a theory using minimal changes from this solution
- Made no changes to the `get` method from `NativeMemoryCacheManager` 
- Made the `close()` method from `NativeMemoryAllocation` to be sync with the `onRemoval` method

This still led to memory overflows and node drop scenarios.
We need both the things to work in tandem.
The solutions are using an additional DS (like this solution) or design a cache that has strict enforcement.

navneet1v (approved):
Approving the PR. I do understand that by adding updating element in the queue is not fully optimized but as the solution provides more stability(in memory constraint situation)I am good with the approach.

I would suggest turning on the feature flag so that we can see the impact of this in the nightlies. But I will leave this upto the author.

src/main/java/org/opensearch/knn/index/memory/NativeMemoryCacheManager.java
heemin32: I would like to explore other cache option like caffeine or ehcache but I guess it can be discussed in a separate thread.

The main concern here is removing and adding entry in dequeue. 
Because this code path is called for every query, it could impact to query latency with high number of segments compared to current implementation.
One other approach would be having last access time in the cache item itself and updating it for every get operation. When eviction is needed, we scan the entire entry and evict least accessed item.

kotwanikunal: The general idea here was to add in correctness.
A minor increase in latency is a lot better than node drops in my opinion. And this can be looked at as an incremental solution. 

With the feature flag turned off - you get the same performance as you do today. But for use cases where node drops are a bigger problem, we can turn on force eviction.

And your concerns are valid - this is not the most optimized solution currently. But it is a step in that direction.
We can work on improving it before enabling the feature by default.

src/main/java/org/opensearch/knn/common/featureflags/KNNFeatureFlags.java
```diff
@@ -0,0 +1,49 @@
...
+@UtilityClass
+public class KNNFeatureFlags {
+
+    // Feature flags
+    private static final String KNN_FORCE_EVICT_CACHE_ENABLED = "knn.feature.cache.force_evict.enabled";
+
+    @VisibleForTesting
+    public static final Setting<Boolean> KNN_FORCE_EVICT_CACHE_ENABLED_SETTING = Setting.boolSetting(
+        KNN_FORCE_EVICT_CACHE_ENABLED,
+        false,
```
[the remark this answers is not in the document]
kotwanikunal: Given @heemin32's callouts and concerns, I think we can let it default to `false` for now. I can turn it on by default once I have some more optimizations in.

src/main/java/org/opensearch/knn/index/memory/NativeMemoryCacheManager.java
heemin32: - Is there any urgency on this feature? Otherwise, implementing it without feature flag is the best.

navneet1v: > Is there any urgency on this feature? Otherwise, implementing it without feature flag is the best.

Feature flag is there to be more of a defensive mechanism, since all tests cannot be covered and this code touches the hot code path. What I am not able to understand is why this question is coming up? Because this is a good feature that adds stability to cluster.

navneet1v: > I would like to explore other cache option like caffeine or ehcache but I guess it can be discussed in a separate thread.

@heemin32 
Do you know that the limitations of Guava cache is not present in these caches?

heemin32: >Feature flag is there to be more of a defensive mechanism, since all tests cannot be covered and this code touches the hot code path. What I am not able to understand is why this question is coming up? Because this is a good feature that adds stability to cluster.

The question is more about the urgency not the need of this feature. I feel feature flag is mostly for user to configure the behavior. As a method of defensive mechanism, we should invest more on testing.
I think we can avoid using dequeue which puts latency for every query. Instead we can have last accessed time en cache entry and evict them when the cache is full where some latency won't have a big impact.

I haven't checked if they don't have the limitation of Guava cache.

navneet1v: I see this as a bug in k-NN plugin, and with more. 


> As a method of defensive mechanism, we should invest more on testing.

There has been testing around it but not all cases can be covered. Hence having this helps.

> I think we can avoid using dequeue which puts latency for every query. Instead we can have last accessed time en cache entry and evict them when the cache is full where some latency won't have a big impact.

What you mean by this?


> I haven't checked if they don't have the limitation of Guava cache.

Then this is the best solution we have for now.

heemin32 (approved):
Approving it as this is behind feature flag and it is false by default. We might need a test with higher number of segments.

kotwanikunal:
Backport: https://github.com/opensearch-project/k-NN/pull/2032

PR #2018, opened 2024-08-31
Adds graph build time metrics in NativeEngines990KnnVectorsWriter

shatejas:
### Description
Adds graph build time. This was missed in the initial PR

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriter.java
+++ b/src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriter.java
@@ -24,11 +24,13 @@ import org.apache.lucene.index.SegmentWriteState;
 import org.apache.lucene.index.Sorter;
 import org.apache.lucene.util.IOUtils;
 import org.apache.lucene.util.RamUsageEstimator;
+import org.opensearch.common.StopWatch;
 import org.opensearch.knn.index.quantizationservice.QuantizationService;
 import org.opensearch.knn.index.VectorDataType;
 import org.opensearch.knn.index.codec.nativeindex.NativeIndexWriter;
 import org.opensearch.knn.index.vectorvalues.KNNVectorValues;
 import org.opensearch.knn.index.vectorvalues.KNNVectorValuesFactory;
+import org.opensearch.knn.plugin.stats.KNNGraphValue;
 import org.opensearch.knn.quantization.models.quantizationParams.QuantizationParams;
 import org.opensearch.knn.quantization.models.quantizationState.QuantizationState;
 
@@ -45,6 +47,10 @@ import static org.opensearch.knn.common.FieldInfoExtractor.extractVectorDataType
 @RequiredArgsConstructor
 public class NativeEngines990KnnVectorsWriter extends KnnVectorsWriter {
     private static final long SHALLOW_SIZE = RamUsageEstimator.shallowSizeOfInstance(NativeEngines990KnnVectorsWriter.class);
+
+    private static final String FLUSH_OPERATION = "flush";
+    private static final String MERGE_OPERATION = "merge";
+
     private final SegmentWriteState segmentWriteState;
     private final FlatVectorsWriter flatVectorsWriter;
     private final List<NativeEngineFieldVectorsWriter<?>> fields = new ArrayList<>();
@@ -78,7 +84,9 @@ public class NativeEngines990KnnVectorsWriter extends KnnVectorsWriter {
                 field.getFieldInfo(),
                 (vectorDataType, fieldInfo, fieldVectorsWriter) -> getKNNVectorValues(vectorDataType, fieldVectorsWriter),
                 NativeIndexWriter::flushIndex,
-                field
+                field,
+                KNNGraphValue.REFRESH_TOTAL_TIME_IN_MILLIS,
+                FLUSH_OPERATION
             );
         }
     }
@@ -88,7 +96,14 @@ public class NativeEngines990KnnVectorsWriter extends KnnVectorsWriter {
         // This will ensure that we are merging the FlatIndex during force merge.
         flatVectorsWriter.mergeOneField(fieldInfo, mergeState);
         // For merge, pick values from flat vector and reindex again. This will use the flush operation to create graphs
-        trainAndIndex(fieldInfo, this::getKNNVectorValuesForMerge, NativeIndexWriter::mergeIndex, mergeState);
+        trainAndIndex(
+            fieldInfo,
+            this::getKNNVectorValuesForMerge,
+            NativeIndexWriter::mergeIndex,
+            mergeState,
+            KNNGraphValue.MERGE_TOTAL_TIME_IN_MILLIS,
+            MERGE_OPERATION
+        );
 
     }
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriter.java
```diff
@@ -228,6 +238,12 @@ private <T, C> void trainAndIndex(
             : NativeIndexWriter.getWriter(fieldInfo, segmentWriteState);
 
         knnVectorValues = vectorValuesRetriever.apply(vectorDataType, fieldInfo, VectorProcessingContext);
+
+        StopWatch stopWatch = new StopWatch();
+        stopWatch.start();
         indexOperation.buildAndWrite(writer, knnVectorValues);
+        long time_in_millis = stopWatch.totalTime().millis();
+        graphBuildTime.set(graphBuildTime.getValue() + time_in_millis);
+        log.warn("Graph build complete in " + time_in_millis + " ms for " + graphBuildTime.getName());
```
navneet1v, line 247: this code will not log what was the time taken for refresh and what time was taken during merge as `graphBuildTime.getName()` will always give `total_time_in_millis` for both of them.

I tested the code via this:

```
public void testABC() {
        System.out.println(KNNGraphValue.MERGE_TOTAL_TIME_IN_MILLIS.getName());
        Assert.fail();
    }
```

Output:
```
[2024-09-03T02:51:53,983][INFO ][o.o.k.i.c.KNNCodecFactoryTests] [testABC] before test
total_time_in_millis
[2024-09-03T02:51:54,037][INFO ][o.o.k.i.c.KNNCodecFactoryTests] [testABC] after test
```

navneet1v (dismissed):
Approving the change. But I do believe this function: trainAndIndex is becoming hard to extend. @Vikasht34 lets do some refactoring on this after the freeze.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriter.java
navneet1v: Two separate enums are used to track times for refresh and merge. This could be the resason.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriter.java
```diff
@@ -228,6 +245,12 @@ private <T, C> void trainAndIndex(
             : NativeIndexWriter.getWriter(fieldInfo, segmentWriteState);
 
         knnVectorValues = vectorValuesRetriever.apply(vectorDataType, fieldInfo, VectorProcessingContext);
+
+        StopWatch stopWatch = new StopWatch();
+        stopWatch.start();
         indexOperation.buildAndWrite(writer, knnVectorValues);
+        long time_in_millis = stopWatch.totalTime().millis();
+        graphBuildTime.set(graphBuildTime.getValue() + time_in_millis);
+        log.warn("Graph build took " + time_in_millis + " ms for " + operationName);
```
heemin32, line 254: Why are we publishing this as warn level but not info level?

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriter.java
```diff
@@ -228,6 +245,12 @@ private <T, C> void trainAndIndex(
             : NativeIndexWriter.getWriter(fieldInfo, segmentWriteState);
 
         knnVectorValues = vectorValuesRetriever.apply(vectorDataType, fieldInfo, VectorProcessingContext);
+
+        StopWatch stopWatch = new StopWatch();
+        stopWatch.start();
         indexOperation.buildAndWrite(writer, knnVectorValues);
+        long time_in_millis = stopWatch.totalTime().millis();
+        graphBuildTime.set(graphBuildTime.getValue() + time_in_millis);
```
heemin32, line 253:
```suggestion 
        graphBuildTime.incrementBy(time_in_millis);
```

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriter.java
```diff
@@ -88,7 +96,14 @@ public void mergeOneField(final FieldInfo fieldInfo, final MergeState mergeState
         // This will ensure that we are merging the FlatIndex during force merge.
         flatVectorsWriter.mergeOneField(fieldInfo, mergeState);
         // For merge, pick values from flat vector and reindex again. This will use the flush operation to create graphs
-        trainAndIndex(fieldInfo, this::getKNNVectorValuesForMerge, NativeIndexWriter::mergeIndex, mergeState);
+        trainAndIndex(
```
heemin32, line 99: Can we add a unit test?

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriter.java
shatejas: Its maintaining the [status quo](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80DocValuesConsumer.java#L57)

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriter.java
shatejas: The class seems to only have a component test. Will try to cover it in the existing component test and create an issue to add unit test around this so it can be taken as a follow up

navneet1v: @heemin32 do we need UTs to check if the log is coming?

heemin32: Not about logging but about adding build time in metrics.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriter.java
heemin32: I think that status quo is not correct.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriter.java
heemin32: > The class seems to only have a component test. Will try to cover it in the existing component test and create an issue to add unit test around this so it can be taken as a follow up

I believe, all these static method is making it harder to write a unit test. I wish we can rely more on dependency injection using Guice.

navneet1v: > I believe, all these static method is making it harder to write a unit test. I wish we can rely more on dependency injection using Guice.

The stats needs to be static as of now as the are gathered at node level. and if we want to make these things work via DI then its a change which is out of scope of this PR.

navneet1v (approved):

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriter.java
heemin32: I know. Just sharing my long term desire and testing the water :)

heemin32 (approved):

Issue #1949, opened 2024-08-12
[RFC] Disk-based Mode Design


jmazanec15:
## Introduction
Document proposes design for mode parameter for disk-based features. For engineers reviewing, please read [Appendix A: Existing Configuration Concepts in k-NN Plugin](#appendix-a-existing-configuration-concepts-in-knn-plugin) for background on important concepts for supporting multiple methods in the k-NN plugin. Refer to [Appendix B: Terminology](#appendix-b-terminology) for common terminology.

## Problem Statement

k-NN search is a vast problem space and users have a diverse set of requirements. For instance, one user may really only care about cost and accuracy, while another has strict search latency requirements. In the current version of the k-NN plugin, users need to tune low-level algorithmic parameters in order to achieve the tradeoffs they want. For example, to create an index now, the mapping may look like:

```
PUT my-vector-index
{
  "mappings": {
    "properties": {
      "my_vector_field": {
        "type": "knn_vector",
        "dimension": 8,
        "method": {
            "name": "hnsw",
            "engine": "faiss",
            "space_type": "l2",
            "params": {
              "m": 16,
              "ef_construction": 512,
              "encoder": {
                "name": "sq",
                "params": {
                  "bits": 8
                }
              }
            }
        }
      }
    }
  }
}
```

This method of fine-tuning creates a complex user experience. In other words, there is a significant amount of effort required to achieve desired performance profile.

As we introduce new disk-based features, this experience is only going to get more complex. So, as a part of this project, we want to ensure that we provide an experience that makes it very easy to onboard, while still giving users flexibility to tune as far as they want to. 

### Functional

1. User should be able to provide a minimal indication that they want to use disk-optimized features for their k-NN field at index creation, and the features get enabled with sensible defaults
2. User should be able to make adjustments to their index configuration to tune for certain performance characteristics. For instance, if a user wants to increase the recall, they should be able to adjust parameters to do so
3. On upgrade, a user’s index configuration should stay constant (i.e. when upgrading, we should not switch underlying parameters).

### Non-functional

1. Documentation should clearly call out what the default values are for certain workload configurations
2. Plugin developers should be able to easily and incrementally improve upon default configurations

## Out of scope

1. Compression parameter design
2. Space type location in mapping. Will assume it is outside of the method, without loss of generality
3. Experimentation for final parameter configuration for disk-based mode.

## Proposed Solution

[...]

PR #2019, opened 2024-09-03
Introduce compression and mode mapping parms

jmazanec15:
### Description
Introduces new params for mapping and training, called compression_level and mode. These parameters are high level parameters that give the plugin a hint as to what the user wants to configure their system like without exposing algorithmic details. This change just adds these parameters to the plugin as noops. In future change, we will add the functionality for parameter resolution.

Along with this, I added a class to more easily manage the original parameters that a user passes. This will help ensure our mapper maintains good compatibility.

Adding tests now, but wanted to raise PR to get early feedback.

### Related Issues
#1949 

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/mapper/CompressionLevel.java
+++ b/src/main/java/org/opensearch/knn/index/mapper/CompressionLevel.java
@@ -20,7 +20,7 @@ import java.util.stream.Collectors;
  */
 @AllArgsConstructor
 public enum CompressionLevel {
-    NOT_CONFIGURED(-1, null),
+    NOT_CONFIGURED(-1, ""),
     x1(1, "1x"),
     x2(2, "2x"),
     x4(4, "4x"),
--- a/src/main/java/org/opensearch/knn/index/mapper/CompressionLevel.java
+++ b/src/main/java/org/opensearch/knn/index/mapper/CompressionLevel.java
@@ -1,0 +1,84 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.knn.index.mapper;
+
+import lombok.AllArgsConstructor;
+
+import java.util.Locale;
+
+/**
+ * Enum representing the compression level for float vectors. Compression in this sense refers to compressing a
+ * full precision value into a smaller number of bits. For instance. "16x" compression would mean that 2 bits would
+ * need to be used to represent a 32-bit floating point number.
+ */
+@AllArgsConstructor
+public enum CompressionLevel {
+    NOT_CONFIGURED(-1),
+    x1(1),
+    x2(2),
+    x4(4),
+    x8(8),
+    x16(16),
+    x32(32);
+
+    /**
+     * Default is set to 1x and is a noop
+     */
+    private static final CompressionLevel DEFAULT = x1;
+
+    /**
+     * Get the compression level from a string representation. The format for the string should be "Nx", where N is
+     * the factor by which compression should take place
+     *
+     * @param name String representation of the compression level
+     * @return CompressionLevel enum value
+     */
+    public static CompressionLevel fromString(String name) {
+        if (name == null) {
+            return NOT_CONFIGURED;
+        }
+        for (CompressionLevel config : CompressionLevel.values()) {
+            if (config.toString() != null && config.toString().equals(name)) {
+                return config;
+            }
+        }
+        throw new IllegalArgumentException(String.format(Locale.ROOT, "Invalid compression level: \"[%s]\"", name));
+    }
+
+    private final int compressionLevel;
+
+    @Override
+    public String toString() {
+        if (this == NOT_CONFIGURED) {
+            return null;
+        }
+        return compressionLevel + "x";
+    }
+
+    /**
+     * Gets the number of bits used to represent a float in order to achieve this compression. For instance, for
+     * 32x compression, each float would need to be encoded in a single bit.
+     *
+     * @return number of bits to represent a float at this compression level
+     */
+    public int numBitsForFloat() {
+        if (this == NOT_CONFIGURED) {
+            return DEFAULT.numBitsForFloat();
+        }
+
+        return (Float.BYTES * Byte.SIZE) * compressionLevel;
+    }
+
+    /**
+     * Utility method that checks if compression is configured.
+     *
+     * @param compressionLevel Compression to check
+     * @return true if compression is configured, false otherwise
+     */
+    public static boolean isConfigured(CompressionLevel compressionLevel) {
+        return compressionLevel != null && compressionLevel != NOT_CONFIGURED;
+    }
+}
--- a/src/main/java/org/opensearch/knn/index/mapper/CompressionLevel.java
+++ b/src/main/java/org/opensearch/knn/index/mapper/CompressionLevel.java
@@ -28,8 +28,10 @@ public enum CompressionLevel {
     x16(16, "16x"),
     x32(32, "32x");
 
+    // Internally, an empty string is easier to deal with them null. However, from the mapping,
+    // we do not want users to pass in the empty string and instead want null. So we make the conversion herex
     static final String[] NAMES_ARRAY = Arrays.stream(CompressionLevel.values())
-        .map(CompressionLevel::getName)
+        .map(compressionLevel -> compressionLevel == NOT_CONFIGURED ? null : compressionLevel.getName())
         .collect(Collectors.toList())
         .toArray(new String[0]);
 
--- a/src/main/java/org/opensearch/knn/index/mapper/CompressionLevel.java
+++ b/src/main/java/org/opensearch/knn/index/mapper/CompressionLevel.java
@@ -6,8 +6,12 @@
 package org.opensearch.knn.index.mapper;
 
 import lombok.AllArgsConstructor;
+import lombok.Getter;
+import org.opensearch.core.common.Strings;
 
+import java.util.Arrays;
 import java.util.Locale;
+import java.util.stream.Collectors;
 
 /**
  * Enum representing the compression level for float vectors. Compression in this sense refers to compressing a
@@ -16,13 +20,18 @@ import java.util.Locale;
  */
 @AllArgsConstructor
 public enum CompressionLevel {
-    NOT_CONFIGURED(-1),
-    x1(1),
-    x2(2),
-    x4(4),
-    x8(8),
-    x16(16),
-    x32(32);
+    NOT_CONFIGURED(-1, null),
+    x1(1, "1x"),
+    x2(2, "2x"),
+    x4(4, "4x"),
+    x8(8, "8x"),
+    x16(16, "16x"),
+    x32(32, "32x");
+
+    static final String[] NAMES_ARRAY = Arrays.stream(CompressionLevel.values())
+        .map(CompressionLevel::getName)
+        .collect(Collectors.toList())
+        .toArray(new String[0]);
 
     /**
      * Default is set to 1x and is a noop
@@ -36,12 +45,12 @@ public enum CompressionLevel {
      * @param name String representation of the compression level
      * @return CompressionLevel enum value
      */
-    public static CompressionLevel fromString(String name) {
-        if (name == null) {
+    public static CompressionLevel fromName(String name) {
+        if (Strings.isEmpty(name)) {
             return NOT_CONFIGURED;
         }
         for (CompressionLevel config : CompressionLevel.values()) {
-            if (config.toString() != null && config.toString().equals(name)) {
+            if (config.getName() != null && config.getName().equals(name)) {
                 return config;
             }
         }
@@ -49,14 +58,8 @@ public enum CompressionLevel {
     }
 
     private final int compressionLevel;
-
-    @Override
-    public String toString() {
-        if (this == NOT_CONFIGURED) {
-            return null;
-        }
-        return compressionLevel + "x";
-    }
+    @Getter
+    private final String name;
 
     /**
      * Gets the number of bits used to represent a float in order to achieve this compression. For instance, for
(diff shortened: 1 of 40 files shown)


src/main/java/org/opensearch/knn/index/mapper/CompressionLevel.java
```diff
@@ -0,0 +1,84 @@
...
+
+    private final int compressionLevel;
+
+    @Override
+    public String toString() {
+        if (this == NOT_CONFIGURED) {
+            return null;
+        }
+        return compressionLevel + "x";
+    }
```
navneet1v, line 59: we are returning 32x here when enum is x32. Can we have like 1 notation. My recommendation will be to use 32x rather than x32.

jmazanec15: Enum needs to start with alpha character unfortunately.

jmazanec15: So I think 32x as string is fine, then x32 is good for name

src/main/java/org/opensearch/knn/indices/ModelDao.java
```diff
@@ -293,7 +295,12 @@ private void putInternal(Model model, ActionListener<IndexResponse> listener, Do
                     put(KNNConstants.MODEL_ERROR, modelMetadata.getError());
                     put(KNNConstants.MODEL_NODE_ASSIGNMENT, modelMetadata.getNodeAssignment());
                     put(KNNConstants.VECTOR_DATA_TYPE_FIELD, modelMetadata.getVectorDataType());
-
+                    if (Mode.isConfigured(modelMetadata.getMode())) {
```
heemin32, line 298: Wonder if `isConfigured` method is really required.
Can we just simply call `put(KNNConstants.MODE_PARAMETER, modelMetadata.getMode())`?

src/main/java/org/opensearch/knn/index/mapper/CompressionLevel.java
```diff
@@ -0,0 +1,84 @@
...
+                return config;
+            }
+        }
+        throw new IllegalArgumentException(String.format(Locale.ROOT, "Invalid compression level: \"[%s]\"", name));
+    }
+
+    private final int compressionLevel;
+
+    @Override
+    public String toString() {
```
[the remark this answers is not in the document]
jmazanec15: I beleive that they are compatible with each other. What do you mean that they arent?

src/main/java/org/opensearch/knn/index/mapper/CompressionLevel.java
```diff
@@ -0,0 +1,84 @@
...
+
+import java.util.Locale;
+
+/**
+ * Enum representing the compression level for float vectors. Compression in this sense refers to compressing a
+ * full precision value into a smaller number of bits. For instance. "16x" compression would mean that 2 bits would
+ * need to be used to represent a 32-bit floating point number.
+ */
+@AllArgsConstructor
+public enum CompressionLevel {
```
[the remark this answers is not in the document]
jmazanec15: I think given that we would use the same parameter for future, we should leave as is

src/main/java/org/opensearch/knn/indices/ModelDao.java
jmazanec15: Do you think its okay to write null here? I was thinking we could avoid doing it.

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java
```diff
@@ -231,18 +239,18 @@ public KNNVectorFieldMapper build(BuilderContext context) {
...
                     hasDocValues.get(),
                     modelDao,
-                    indexCreatedVersion
+                    indexCreatedVersion,
+                    originalParameters
                 );
             }
 
-            if (resolvedKNNMethodContext == null) {
+            if (originalParameters.getResolvedKnnMethodContext() == null) {
```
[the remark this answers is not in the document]
jmazanec15: I like this - but, might take in another review

src/main/java/org/opensearch/knn/index/mapper/CompressionLevel.java
navneet1v: > Enum needs to start with alpha character unfortunately.

Didn't know this. Feels pretty weird.

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java
```diff
@@ -162,6 +156,20 @@ public static class Builder extends ParametrizedFieldMapper.Builder {
             }
         });
 
+        protected final Parameter<String> mode = Parameter.restrictedStringParam(
```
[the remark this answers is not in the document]
jmazanec15: yes

navneet1v (commented):
Overall looks good to me. I am little skeptical on the part of having NotConfigured along with null values for new Enums. Would like to know what other think about this.

cc: @heemin32 , @shatejas

src/main/java/org/opensearch/knn/indices/ModelDao.java
heemin32: I think it should be okay. Anyway, we are not checking if the key exist while reading it but checking the value is null or not.

src/main/java/org/opensearch/knn/index/mapper/CompressionLevel.java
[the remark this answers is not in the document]
heemin32: You cannot call like
```
CompressionLevel.fromString(x86.toString());
```

src/main/java/org/opensearch/knn/indices/ModelMetadata.java
```diff
@@ -89,6 +96,15 @@ public ModelMetadata(StreamInput in) throws IOException {
         } else {
             this.vectorDataType = VectorDataType.DEFAULT;
         }
+
+        if (IndexUtil.isVersionOnOrAfterMinRequiredVersion(in.getVersion(), KNNConstants.MINIMAL_MODE_AND_COMPRESSION_FEATURE)) {
+            this.mode = Mode.fromString(in.readOptionalString());
+            this.compressionLevel = CompressionLevel.fromString(in.readOptionalString());
+        } else {
+            this.mode = Mode.NOT_CONFIGURED;
+            this.compressionLevel = CompressionLevel.NOT_CONFIGURED;
```
shatejas, line 105: nit: could have defaulted to `in_memory` and `1x` here. They are still valid values for before 2.17

src/main/java/org/opensearch/knn/index/mapper/CompressionLevel.java
```diff
@@ -0,0 +1,84 @@
...
+        return compressionLevel + "x";
+    }
+
+    /**
+     * Gets the number of bits used to represent a float in order to achieve this compression. For instance, for
+     * 32x compression, each float would need to be encoded in a single bit.
+     *
+     * @return number of bits to represent a float at this compression level
+     */
+    public int numBitsForFloat() {
```
heemin32, line 67: ?
```suggestion
    public int numBitsForFloat32() {
```

src/main/java/org/opensearch/knn/indices/ModelMetadata.java
```diff
@@ -304,13 +324,14 @@ public static ModelMetadata fromString(String modelMetadataString) {
...
-        if (length < 7 || length > 10) {
+        if (length < 7 || length > 12) {
             throw new IllegalArgumentException(
                 "Illegal format for model metadata. Must be of the form "
                     + "\"<KNNEngine>,<SpaceType>,<Dimension>,<ModelState>,<Timestamp>,<Description>,<Error>\" or "
                     + "\"<KNNEngine>,<SpaceType>,<Dimension>,<ModelState>,<Timestamp>,<Description>,<Error>,<NodeAssignment>\" or "
                     + "\"<KNNEngine>,<SpaceType>,<Dimension>,<ModelState>,<Timestamp>,<Description>,<Error>,<NodeAssignment>,<MethodContext>\" or "
-                    + "\"<KNNEngine>,<SpaceType>,<Dimension>,<ModelState>,<Timestamp>,<Description>,<Error>,<NodeAssignment>,<MethodContext>,<VectorDataType>\"."
+                    + "\"<KNNEngine>,<SpaceType>,<Dimension>,<ModelState>,<Timestamp>,<Description>,<Error>,<NodeAssignment>,<MethodContext>,<VectorDataType>\". or "
+                    + "\"<KNNEngine>,<SpaceType>,<Dimension>,<ModelState>,<Timestamp>,<Description>,<Error>,<NodeAssignment>,<MethodContext>,<VectorDataType>,<WorkloadConfig>,<CompressionConfig>\"."
```
shatejas, line 334: nit: `<Mode>,<CompressionLevel>`

src/main/java/org/opensearch/knn/index/mapper/CompressionLevel.java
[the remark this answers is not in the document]
heemin32: Fair enough. The javadoc says this is only for fp32 so..

heemin32:
> Overall looks good to me. I am little skeptical on the part of having NotConfigured along with null values for new Enums. Would like to know what other think about this.
> 
> cc: @heemin32 , @shatejas

Yes. By having NotConfigured, we should be able to assume that the enum won't be a null for it to be useful.

shatejas:
> Overall looks good to me. I am little skeptical on the part of having NotConfigured along with null values for new Enums. Would like to know what other think about this.
> 
> cc: @heemin32 , @shatejas

@navneet1v https://github.com/jmazanec15/k-NN-1/pull/1#discussion_r1739289392

Ideally we shouldn't have placeholders, I think we can get away with it unless I am missing something. We can revisit removing as long as its not a one way door (make sure we are not passing `NOT_CONFIGURED` in streams used for bwc)

src/main/java/org/opensearch/knn/index/mapper/CompressionLevel.java
```diff
@@ -0,0 +1,84 @@
...
+import java.util.Locale;
+
+/**
+ * Enum representing the compression level for float vectors. Compression in this sense refers to compressing a
+ * full precision value into a smaller number of bits. For instance. "16x" compression would mean that 2 bits would
+ * need to be used to represent a 32-bit floating point number.
+ */
+@AllArgsConstructor
+public enum CompressionLevel {
+    NOT_CONFIGURED(-1),
```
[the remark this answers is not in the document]
jmazanec15: Right. The value isnt exposed outside of this class and is handled below

src/main/java/org/opensearch/knn/indices/ModelMetadata.java
jmazanec15: I prefer NOT_CONFIGURED just because I want to differentiate when someone actually supplies as config "1x" or "in_memory" and handle that case a bit differently.

shatejas: > handle that case a bit differently

Would appreciate more context on this part. I was of the opinion it will execute the same code path as the current one

src/main/java/org/opensearch/knn/indices/ModelDao.java
```diff
@@ -293,7 +295,12 @@ private void putInternal(Model model, ActionListener<IndexResponse> listener, Do
...
+                        put(KNNConstants.MODE_PARAMETER, modelMetadata.getMode().toString());
+                    }
+                    if (CompressionLevel.isConfigured(modelMetadata.getCompressionLevel())) {
+                        put(KNNConstants.COMPRESSION_LEVEL_PARAMETER, modelMetadata.getCompressionLevel().toString());
```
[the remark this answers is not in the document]
jmazanec15: I was a little bit concerned here that we are ingesting the value into the system index and was concerned on having an object indexed. Do you think thats fine though?

shatejas (approved):
Overall no concerns with functionality 

- Please make sure NOT_CONFIGURED is not passed in streams for bwc in any case

src/main/java/org/opensearch/knn/indices/ModelMetadata.java
```diff
@@ -465,6 +503,14 @@ public XContentBuilder toXContent(XContentBuilder builder, Params params) throws
         if (IndexUtil.isClusterOnOrAfterMinRequiredVersion(KNNConstants.MODEL_VECTOR_DATA_TYPE_KEY)) {
             builder.field(KNNConstants.VECTOR_DATA_TYPE_FIELD, vectorDataType.getValue());
         }
+        if (IndexUtil.isClusterOnOrAfterMinRequiredVersion(KNNConstants.MINIMAL_MODE_AND_COMPRESSION_FEATURE)) {
+            if (Mode.isConfigured(mode)) {
+                builder.field(KNNConstants.MODE_PARAMETER, mode.toString());
+            }
+            if (CompressionLevel.isConfigured(compressionLevel)) {
+                builder.field(KNNConstants.COMPRESSION_LEVEL_PARAMETER, compressionLevel.toString());
```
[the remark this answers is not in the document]
jmazanec15: For toXContent method, this will be potentially exposed to user. So, I was thinking it would be better to return what they would put in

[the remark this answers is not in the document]
navneet1v: In that case, we should have another parameter in enum which tells what is user provided value. having toString here is not what I would recommend

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java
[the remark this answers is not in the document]
jmazanec15: I thought @heemin32  meant just the last stream

src/main/java/org/opensearch/knn/index/mapper/CompressionLevel.java
```diff
@@ -0,0 +1,87 @@
...
+import java.util.stream.Collectors;
+
+/**
+ * Enum representing the compression level for float vectors. Compression in this sense refers to compressing a
+ * full precision value into a smaller number of bits. For instance. "16x" compression would mean that 2 bits would
+ * need to be used to represent a 32-bit floating point number.
+ */
+@AllArgsConstructor
+public enum CompressionLevel {
+    NOT_CONFIGURED(-1, null),
```
navneet1v, line 23: interesting that we using null here as name. In that case we can never get NOT_CONFIGURED from `fromName` function is that correct?

heemin32 (dismissed):

navneet1v (approved):

naveentatikonda (approved):

PR #2034, opened 2024-09-04
Introduce mode and compression param resolution

jmazanec15:
### Description
Adds mode and compression based parameter resolution. With this, if a user specifies the mode and/or compression params, we will create a default configuration with the aim of meeting those hints.

Currently, it does not contain support for overriding any of the parameters. This will be taken in a future commit.

Still working on tests, but figured it would be good to get the PR out. Heres what I still need to do:
- [ ] Resolve test failures
- [x] Block mode/compression on non-float data types
- [ ] Add test coverage

### Related Issues
#1949 

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/mapper/ModeBasedResolver.java
+++ b/src/main/java/org/opensearch/knn/index/mapper/ModeBasedResolver.java
@@ -1,0 +1,209 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.knn.index.mapper;
+
+import org.opensearch.knn.index.KNNSettings;
+import org.opensearch.knn.index.SpaceType;
+import org.opensearch.knn.index.engine.KNNEngine;
+import org.opensearch.knn.index.engine.KNNMethodContext;
+import org.opensearch.knn.index.engine.MethodComponentContext;
+import org.opensearch.knn.index.engine.faiss.QFrameBitEncoder;
+import org.opensearch.knn.index.query.rescore.RescoreContext;
+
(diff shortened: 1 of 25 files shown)


src/main/java/org/opensearch/knn/index/mapper/ModeBasedResolver.java
```diff
@@ -0,0 +1,158 @@
...
+                        KNNSettings.INDEX_KNN_DEFAULT_ALGO_PARAM_EF_SEARCH,
+                        METHOD_ENCODER_PARAMETER,
+                        encoderContext
+                    )
+                )
+            );
+        }
+
+        if (knnEngine == KNNEngine.FAISS) {
+            return new KNNMethodContext(
```
naveentatikonda, line 79: The last two return statements look alike, then why do we need this condition ?

jmazanec15: Good catch - I updated. Difference is that nmslib does not take ef_search as index param while faiss does

navneet1v:
Code looks good to me. Waiting for CI to be successful

src/test/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsFormatTests.java
```diff
@@ -95,6 +96,7 @@ public void tearDown() throws Exception {
         super.tearDown();
     }
 
+    @Ignore
```
jmazanec15, line 99: Ignoring - there is some broken functionality around the quantization state that is being fixed.

ryanbogan (approved):

PR #2035, opened 2024-09-04
Adds index and query BWC tests for missing engines

shatejas:
### Description

This is done to make sure that KNN80DocsValueConsumer code path is hit

### Check List

### DIFF
--- a/qa/restart-upgrade/build.gradle
+++ b/qa/restart-upgrade/build.gradle
@@ -35,6 +35,7 @@ testClusters {
     }
 }
 
+   def versionsBelow2_3 = ["1.", "2.0.", "2.1.", "2.2."]
 // Task to run BWC tests against the old cluster
     task testAgainstOldCluster(type: StandaloneRestIntegTestTask) {
         dependsOn "zipBwcPlugin"
@@ -86,6 +87,13 @@ testClusters {
                 knn_bwc_version.startsWith("2.15.")) {
             filter {
                 excludeTestsMatching "org.opensearch.knn.bwc.IndexingIT.testKNNIndexBinaryForceMerge"
+            }
+        }
+
+        if (versionsBelow2_3.any {knn_bwc_version.startsWith(it) }) {
+            filter {
+                excludeTestsMatching "org.opensearch.knn.bwc.QueryANNIT.testQueryOnLuceneIndex"
+                excludeTestsMatching "org.opensearch.knn.bwc.IndexingIT.testKNNIndexLuceneForceMerge"
             }
         }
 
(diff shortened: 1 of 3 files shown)


qa/restart-upgrade/build.gradle
```diff
@@ -35,6 +35,7 @@ testClusters {
     }
 }
 
+   def versionsBelow2_3 = ["1.", "2.0.", "2.1.", "2.2."]
```
navneet1v, line 38: I see that we have added test for restart upgrades here. Do we have plans to add test for rolling upgrades too in follow up PR?

qa/restart-upgrade/src/test/java/org/opensearch/knn/bwc/QueryANNIT.java
```diff
@@ -36,4 +35,28 @@ public void testQueryANN() throws Exception {
...
+    }
+
+    public void testQueryOnLuceneIndex() throws Exception {
+        if (isRunningAgainstOldCluster()) {
+            createKnnIndex(testIndex, getKNNDefaultIndexSettings(), createKnnIndexMapping(TEST_FIELD, DIMENSIONS, ALGORITHM, LUCENE_NAME));
+            addKNNDocs(testIndex, TEST_FIELD, DIMENSIONS, 0, NUM_DOCS);
+            validateKNNSearch(testIndex, TEST_FIELD, DIMENSIONS, NUM_DOCS, K);
+        } else {
+            validateKNNSearch(testIndex, TEST_FIELD, DIMENSIONS, NUM_DOCS, K);
+            validateKNNSearch(testIndex, TEST_FIELD, DIMENSIONS, NUM_DOCS, K, Map.of(METHOD_PARAMETER_EF_SEARCH, EF_SEARCH));
```
navneet1v, line 58: I thought we didn't support MethodParameters for Lucene in the query. But I can see a test here. am I missing something here?

shatejas: We do, it overrides k based on max value of ef and k

qa/restart-upgrade/build.gradle
shatejas: rolling upgrade will only hit native code path after 2.17 since we just allow 1 version upgrade so its moot unless I am missing something

navneet1v: Make sense.. i forgot that

qa/restart-upgrade/src/test/java/org/opensearch/knn/bwc/QueryANNIT.java
navneet1v: Make sense

navneet1v (approved):

heemin32 (approved):

PR #2055, opened 2024-09-05
Throws and exception for radial search when mapping is for on-disk mode

shatejas:
### Description
- Adds a debug log for time spent rescoring

### Related Issues

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
+++ b/src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
@@ -25,6 +25,7 @@ import org.opensearch.index.query.QueryRewriteContext;
 import org.opensearch.index.query.QueryShardContext;
 import org.opensearch.knn.index.engine.KNNMethodConfigContext;
 import org.opensearch.knn.index.engine.model.QueryContext;
+import org.opensearch.knn.index.engine.qframe.QuantizationConfig;
 import org.opensearch.knn.index.mapper.KNNMappingConfig;
 import org.opensearch.knn.index.mapper.KNNVectorFieldType;
 import org.opensearch.knn.index.query.parser.RescoreParser;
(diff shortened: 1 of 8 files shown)


ryanbogan (dismissed):

heemin32 (approved):

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
```diff
@@ -451,6 +452,10 @@ protected Query doToQuery(QueryShardContext context) {
             if (vectorDataType == VectorDataType.BINARY) {
                 throw new UnsupportedOperationException(String.format(Locale.ROOT, "Binary data type does not support radial search"));
             }
+
```
jmazanec15, line 455: Im thinking that we add in the KNNMappingConfig a value that says whether quantization framework is being used and then expose it via the the KNNVectorFieldType.


To get if its being used, in the model and method field mapper, itll look something like:
```
        KNNLibraryIndexingContext knnLibraryIndexingContext = knnMethodContext.getKnnEngine()
            .getKNNLibraryIndexingContext(knnMethodContext, knnMethodConfigContext);
        if (knnLibraryIndexingContext.getQuantizationConfig() != QuantizationConfig.EMPTY)
```

shatejas: I couldn't grok why quantization is involved here can you elaborate?

Is the idea that if they have compression 1x we allow radial search, at that point is mode value relevant?

jmazanec15: For radial search, user passes in scoring threshold for which results should or shouldnt be returned. The problem is, for binary quantization, the configured distance is "l2" and then we quantize vecs and use the "hamming" distance. The scale of the "l2" distance and "hamming" will be completely different - i.e. hamming doesnt approximate the actual l2 distance. So, radial search doesnt make sense if we are doing this binary quantization

jmazanec15 (approved):

ryanbogan (approved):

PR #2071, opened 2024-09-08
Re-Call Issue Fix with Binary Quantized Vectors

Vikasht34:
### Description
This PR addresses a critical issue that was identified during benchmarking, where the recall performance unexpectedly dropped below 1. The root cause of the issue was traced to two main problems in the quantization and vector handling process:

1. Bit Packing of Quantized Vector:  The quantized vector was not correctly updated during the bit-packing process. Instead, the same vector values were being reused, resulting in incorrect quantization of subsequent vectors.
2. Vector Transfer to JNI: When transferring vectors to the JNI layer, vectors were passed by reference, causing all the vectors in the VectorTransfer list to reference the same object. This led to unintended behavior where all vectors became identical, severely affecting the recall accuracy.
### Testing
Performed Benchmarking with NQ Data Set Results are here s3://disk-based-ann-bq/NQ-1M-768/
Re-Call for One Bit Quantization with 2x Oversampling us 0.94

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/codec/nativeindex/QuantizationIndexUtils.java
+++ b/src/main/java/org/opensearch/knn/index/codec/nativeindex/QuantizationIndexUtils.java
@@ -33,7 +33,7 @@ class QuantizationIndexUtils {
                 knnVectorValues.getVector(),
                 indexBuildSetup.getQuantizationOutput()
             );
-            return indexBuildSetup.getQuantizationOutput().getQuantizedVector();
+            return indexBuildSetup.getQuantizationOutput().getQuantizedVectorCopy();
         } else {
             return knnVectorValues.conditionalCloneVector();
         }
--- a/src/main/java/org/opensearch/knn/index/codec/nativeindex/QuantizationIndexUtils.java
+++ b/src/main/java/org/opensearch/knn/index/codec/nativeindex/QuantizationIndexUtils.java
@@ -18,12 +18,12 @@ import java.io.IOException;
 class QuantizationIndexUtils {
 
     /**
-     * Processes and returns the vector based on whether quantization is applied or not.
+     * Processes the vector from {@link KNNVectorValues} and returns either a cloned quantized vector or a cloned original vector.
      *
-     * @param knnVectorValues the KNN vector values to be processed.
-     * @param indexBuildSetup the setup containing quantization state and output, along with other parameters.
-     * @return the processed vector, either quantized or original.
-     * @throws IOException if an I/O error occurs during processing.
+     * @param knnVectorValues The KNN vector values containing the original vector.
+     * @param indexBuildSetup The setup containing the quantization state and output details.
+     * @return The quantized vector (as a byte array) or the original/cloned vector.
+     * @throws IOException If an I/O error occurs while processing the vector.
      */
     static Object processAndReturnVector(KNNVectorValues<?> knnVectorValues, IndexBuildSetup indexBuildSetup) throws IOException {
         QuantizationService quantizationService = QuantizationService.getInstance();
@@ -33,6 +33,10 @@ class QuantizationIndexUtils {
                 knnVectorValues.getVector(),
                 indexBuildSetup.getQuantizationOutput()
             );
+            /**
+             * Returns a copy of the quantized vector. This is because of during transfer same vectors was getting
+             * added due to reference.
+             */
             return indexBuildSetup.getQuantizationOutput().getQuantizedVectorCopy();
         } else {
             return knnVectorValues.conditionalCloneVector();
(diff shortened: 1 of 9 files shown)


src/main/java/org/opensearch/knn/quantization/models/quantizationOutput/BinaryQuantizationOutput.java
```diff
@@ -63,4 +63,16 @@ public boolean isPrepared(int vectorLength) {
...
         return quantizedVector;
     }
+
+    /**
+     * Returns a copy of the quantized vector.
+     *
+     * @return a copy of the quantized vector byte array.
+     */
+    @Override
+    public byte[] getQuantizedVectorCopy() {
```
navneet1v, line 73: [Not a blocker]
Lets add the documentation of the `getQuantizedVector` that it gives the same reference object in return. This will ensure that person using the function knows how to use these functions.

navneet1v, line 73: [Not a blocker]
Also lets have some unit test for `getQuantizedVector` and `getQuantizedVectorCopy` functions to ensure that same reference is returned when `getQuantizedVector` is called and different is returned when `getQuantizedVectorCopy`.

src/main/java/org/opensearch/knn/index/codec/nativeindex/QuantizationIndexUtils.java
```diff
@@ -33,7 +33,7 @@ static Object processAndReturnVector(KNNVectorValues<?> knnVectorValues, IndexBu
                 knnVectorValues.getVector(),
                 indexBuildSetup.getQuantizationOutput()
             );
-            return indexBuildSetup.getQuantizationOutput().getQuantizedVector();
+            return indexBuildSetup.getQuantizationOutput().getQuantizedVectorCopy();
```
navneet1v, line 36: [Not a blocker]
Lets add a java doc on top of this line to tell why we are calling VectorCopy and not `Vector()` function

navneet1v (commented):
Overall code looks good to me. There are some minor comments. See how you want to resolve it.

Apart from this I would recommend doing 2 things:
1. Update the description of the PR with details on what are the bugs which was found.
2. Also add the details on how you tested these, like if you have ran a benchmarks to see the recall add that too.

src/main/java/org/opensearch/knn/index/codec/nativeindex/QuantizationIndexUtils.java
```diff
@@ -33,7 +33,11 @@ static Object processAndReturnVector(KNNVectorValues<?> knnVectorValues, IndexBu
                 knnVectorValues.getVector(),
                 indexBuildSetup.getQuantizationOutput()
             );
-            return indexBuildSetup.getQuantizationOutput().getQuantizedVector();
+            /**
+             * Returns a copy of the quantized vector. This is because of during transfer same vectors was getting a
+             * dded due to reference.
```
ryanbogan, line 38: Nit: added on the same line

ryanbogan (dismissed):
One minor nit but not blocking.  Otherwise, LGTM!

ryanbogan (commented):
Looks like there are some merge conflicts that need to be fixed @Vikasht34

ryanbogan (dismissed):

ryanbogan (dismissed):

ryanbogan (approved):

navneet1v (approved):

src/test/java/org/opensearch/knn/quantization/output/BinaryQuantizationOutputTests.java
```diff
@@ -0,0 +1,121 @@
...
+
+public class BinaryQuantizationOutputTests extends KNNTestCase {
+
+    private static final int BITS_PER_COORDINATE = 1;
+    private BinaryQuantizationOutput quantizationOutput;
+
+    @Before
+    public void setUp() throws Exception {
+        super.setUp();
+        quantizationOutput = new BinaryQuantizationOutput(BITS_PER_COORDINATE);
```
shatejas, line 20: This member variable seems to be creating interdependency between tests, can we create this for every test please

src/main/java/org/opensearch/knn/index/codec/nativeindex/QuantizationIndexUtils.java
```diff
@@ -33,7 +33,11 @@ static Object processAndReturnVector(KNNVectorValues<?> knnVectorValues, IndexBu
                 knnVectorValues.getVector(),
                 indexBuildSetup.getQuantizationOutput()
             );
-            return indexBuildSetup.getQuantizationOutput().getQuantizedVector();
+            /**
+             * Returns a copy of the quantized vector. This is because of during transfer same vectors was getting
+             * added due to reference.
+             */
+            return indexBuildSetup.getQuantizationOutput().getQuantizedVectorCopy();
```
shatejas, line 40: Can we update the unit test for this class to make sure we always return a copy? we need a `assertNotSame` check in the unit test of this class

PR #2046, opened 2024-09-05
Fix the force merge with Quantization failures when a segment has deleted docs in it

navneet1v:
### Description
Fix the force merge with Quantization failures when a segment has deleted docs in it.

#### Issue:
When documents are deleted from the segment and segments are force merged, the FlatVectorValues are getting merged to 1 vectorValues but that `vectorvalue.getLiveDocs()` doesn't provide correct live docs but it provides the `cost` which includes the deleted docs too. 

Due to this when we are doing quantization, we are hitting the vectors which are not present and leading to NPE. 

```
org.apache.lucene.index.MergePolicy$MergeException: java.lang.IllegalStateException: this writer hit an unrecoverable error; cannot merge
        at org.opensearch.index.engine.InternalEngine$EngineMergeScheduler$2.doRun(InternalEngine.java:2494) [opensearch-3.0.0-SNAPSHOT.jar:3.0.0-SNAPSHOT]
        at org.opensearch.common.util.concurrent.ThreadContext$ContextPreservingAbstractRunnable.doRun(ThreadContext.java:982) [opensearch-3.0.0-SNAPSHOT.jar:3.0.0-SNAPSHOT]
        at org.opensearch.common.util.concurrent.AbstractRunnable.run(AbstractRunnable.java:52) [opensearch-3.0.0-SNAPSHOT.jar:3.0.0-SNAPSHOT]
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144) [?:?]
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642) [?:?]
        at java.base/java.lang.Thread.run(Thread.java:1583) [?:?]
Caused by: java.lang.IllegalStateException: this writer hit an unrecoverable error; cannot merge
        at org.apache.lucene.index.IndexWriter.hasPendingMerges(IndexWriter.java:2436) ~[lucene-core-9.12.0-snapshot-847316d.jar:9.12.0-snapshot-847316d 847316dd1eec2258ccf5da52dca51c01b9df1c4c - 2024-06-27 17:07:28]
        at org.opensearch.index.engine.InternalEngine$EngineMergeScheduler.afterMerge(InternalEngine.java:2452) ~[opensearch-3.0.0-SNAPSHOT.jar:3.0.0-SNAPSHOT]
[...]

PR #2072, opened 2024-09-09
[BugFix] Fixing KNNMethodContext resolution when documents are ingested and a flaky test

navneet1v:
### Description
Fixing KNNMethodContext resolution when documents are ingested and a flaky test

### Bug Fixes:
1. When mode and compression is provided and documents are ingested, the rescore parameter was not getting passed in during search, as in the FieldMapper mergeBuilder Mode and compression attributes were not getting set.
2. There was a flaky test `testDeletedDocsWithSegmentMerge_whenValid_ThenSucceed`. It was added in the PR: https://github.com/opensearch-project/k-NN/pull/2046 to test that merges are happening without any failures. But as this test was doing search at the end it is not necessary that search will always give consistent results as the documents are deleted. Hence the correct test for this is ensuring the index is green after the merge. The flakiness was observed here: https://github.com/opensearch-project/k-NN/pull/2046

### Check List

### DIFF
--- a/src/test/java/org/opensearch/knn/integ/ModeAndCompressionIT.java
+++ b/src/test/java/org/opensearch/knn/integ/ModeAndCompressionIT.java
@@ -7,16 +7,18 @@ package org.opensearch.knn.integ;
 
 import lombok.SneakyThrows;
 import org.apache.hc.core5.http.io.entity.EntityUtils;
+import org.junit.Assert;
 import org.junit.Ignore;
+import org.opensearch.client.Request;
 import org.opensearch.client.Response;
 import org.opensearch.client.ResponseException;
 import org.opensearch.common.xcontent.XContentFactory;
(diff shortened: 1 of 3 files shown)


src/test/java/org/opensearch/knn/integ/ModeAndCompressionIT.java
```diff
@@ -50,7 +52,7 @@ public class ModeAndCompressionIT extends KNNRestTestCase {
 
     private static final int DIMENSION = 16;
     private static final int NUM_DOCS = 20;
-    private static final int K = 2;
+    private static final int K = NUM_DOCS;
```
navneet1v, line 55: we need to use K=NUM_DOCS here because there can be equal scores in the firstPass of search hence the results are not consistent in runs. With this atleast we ensure that all docs are searched and their scores are matched.

jmazanec15 (approved):

PR #2076, opened 2024-09-09
[Bug-Fix] Fix efficient filtering of vector search when quantization is used

navneet1v:
### Description
[Bug-Fix] Fix efficient filtering of vector search when quantization is used

### Issue
When efficient filtering happens then at per segment level k-NN plugin takes a decision to either do filtered ANN search or Exact search. But when quantization is present then the first pass search happens on the quantized vectors. Currently only ANN search happens with quantized vectors and not exact search. This can lead of different scores for a shard for first pass search if its segments takes different route to do the filtered search.

### What this PR do?
This PR adds the capability to do quantized exact search when exact search happens during filtered vector search. The way it is achieved is, by passing the segment level quantization information to KNNIterators which iterates over the vectors. It also add a capability now where we can specify if we want to search on quantized vectors or non quantized vectors. 

### Testing
Performed a Manual testing for filtered vector search. Will be adding more ITs and UTs for the change. Raising the PR for doing first pass on the code.

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/query/ExactSearcher.java
+++ b/src/main/java/org/opensearch/knn/index/query/ExactSearcher.java
@@ -6,6 +6,8 @@
 package org.opensearch.knn.index.query;
 
 import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.Value;
 import lombok.extern.log4j.Log4j2;
 import org.apache.lucene.index.FieldInfo;
 import org.apache.lucene.index.LeafReaderContext;
@@ -42,27 +44,18 @@ public class ExactSearcher {
     /**
      * Execute an exact search on a subset of documents of a leaf
      *
-     * @param leafReaderContext LeafReaderContext to be searched over
-     * @param matchedDocs matched documents
-     * @param knnQuery KNN Query
-     * @param k number of results to return
-     * @param isParentHits whether the matchedDocs contains parent ids or child ids. This is relevant in the case of
-     *                     filtered nested search where the matchedDocs contain the parent ids and {@link NestedFilteredIdsKNNIterator}
-     *                     needs to be used.
+     * @param leafReaderContext {@link LeafReaderContext}
+     * @param exactSearcherContext {@link ExactSearcherContext}
      * @return Map of re-scored results
+     * @throws IOException exception during execution of exact search
      */
-    public Map<Integer, Float> searchLeaf(
-        final LeafReaderContext leafReaderContext,
-        final BitSet matchedDocs,
-        final KNNQuery knnQuery,
-        int k,
-        boolean isParentHits
-    ) throws IOException {
-        KNNIterator iterator = getMatchedKNNIterator(leafReaderContext, matchedDocs, knnQuery, isParentHits);
-        if (matchedDocs.cardinality() <= k) {
+    public Map<Integer, Float> searchLeaf(final LeafReaderContext leafReaderContext, final ExactSearcherContext exactSearcherContext)
+        throws IOException {
+        KNNIterator iterator = getMatchedKNNIterator(leafReaderContext, exactSearcherContext);
+        if (exactSearcherContext.getMatchedDocs().cardinality() <= exactSearcherContext.getK()) {
             return scoreAllDocs(iterator);
         }
-        return searchTopK(iterator, k);
+        return searchTopK(iterator, exactSearcherContext.getK());
     }
 
     private Map<Integer, Float> scoreAllDocs(KNNIterator iterator) throws IOException {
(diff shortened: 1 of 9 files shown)


src/main/java/org/opensearch/knn/index/query/ExactSearcher.java
```diff
@@ -145,10 +142,43 @@ private KNNIterator getMatchedKNNIterator(
...
+     */
+    @Value
+    @Builder
+    public static class ExactSearcherContext {
+        /**
+         * controls whether we should use Quantized vectors during exact search or not. This is useful because when we do
+         * re-scoring we need to re-score using full precision vectors and not quantized vectors.
+         */
+        boolean useQuantizedVectorsForSearch;
+        LeafReaderContext leafReaderContext;
```
shatejas, line 173: [Not a blocker for release] nit: searchLeaf should be a function of LeafReaderContext and exactSearchParams, would have preferred it separate

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -140,8 +137,16 @@ public Map<Integer, Float> searchLeaf(LeafReaderContext context, int k) throws I
          * This improves the recall.
          */
         Map<Integer, Float> docIdsToScoreMap;
+        final ExactSearcher.ExactSearcherContext exactSearcherContext = ExactSearcher.ExactSearcherContext.builder()
```
shatejas, line 140: nit: Seems like we are building it early

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -140,8 +137,16 @@ public Map<Integer, Float> searchLeaf(LeafReaderContext context, int k) throws I
...
+            .k(k)
+            .isParentHits(true)
```
shatejas, line 142: nit: Is this correct? if yes I think it needs a comment why is it kept true

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -140,8 +137,16 @@ public Map<Integer, Float> searchLeaf(LeafReaderContext context, int k) throws I
...
+            .matchedDocs(filterBitSet)
+            .leafReaderContext(context)
+            .useQuantizedVectorsForSearch(true)
```
shatejas, line 145: Same for this, add a comment of why its kept true. Is there a way we can pass in QuantizationConfig here? its available in mappedFieldType

shatejas (commented):
No major blockers as such, just need clarification of the hard coded  true part

Any latency considerations considering each vector from flat values will be quantized for efficient filters. It might be worth calling out if some impact is expected and can be considered for optimizations of efficient filters

src/main/java/org/opensearch/knn/index/query/ExactSearcher.java
navneet1v: Sure. I don't see why we want to do it like that? Its just better to have 1 single class encapsulating the whole context of the search

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
navneet1v: As this is used in the both if and else condition, hence I am building it outside of if.

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
navneet1v: These were the defaults which was present. I can add a comment here.

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
navneet1v: Its not about passing the Quantized config. Even when Quantized Config is present, we might not want to do a quantization based exact search. So every caller knows on what type of vectors it wants to do exact search.

navneet1v:
> Any latency considerations considering each vector from flat values will be quantized for efficient filters. It might be worth calling out if some impact is expected and can be considered for optimizations of efficient filters

As the quantized vectors are not stored in the segment, and we have to do on the fly quantization there will be latency impact. But thing is we don't have a baseline for this. Since QF is added with on-disk support so this is the first time we will be having this flow. For all the older cases there will no latency impact ideally at-least from this change.

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
navneet1v: NVM. the comment is here: https://github.com/opensearch-project/k-NN/pull/2076/files#diff-68bee8536a6d51809bc0408f057e47c6f2c1921a0eef56488f2b54cc1986a85bR177-R182

src/main/java/org/opensearch/knn/index/query/ExactSearcher.java
shatejas: function signature can often be representations of what it does, in this case the function represents "search on a leaf(segment) given these params or a query".  It doesn't seem logical to club both data and params to search on that data especially when not needed.

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
navneet1v: will add some comment here too.

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
shatejas: So the idea is to make the decision early with quantization config (if its possible). In this case, we are adding true as default and then falling back on quantization params as null. Checking quantization params checks for quantization config. anyways not a blocker as its not expensive if its not present

shatejas (approved):

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
navneet1v: > So the idea is to make the decision early with quantization config (if its possible)

Here I took the other approach. The flow tells exact search to do search on Quantized vectors or not. This has some benefits: 
1. The logic of what vectors to use for search is defined by the flow and not dependent on Quantization param. Biggest reason for that is : for rescoring if you look at we have the Quantization params but we want to search via full precision vectors. 

If we move towards the use of QuantizationParams being null or not in fieldInfo to say what vectors to use for quantization then it will not work for rescoring.

naveentatikonda (approved):

jmazanec15 (approved):

PR #2087, opened 2024-09-11
Improving Defaults Hyper Parameter for Binary Quantization Indexes

Vikasht34:
### Description
This PR contains changes to change default hyper parameter in case of Binary Quantization from 100 to 256 for ef Search and ef Construction and also changes the default oversampling factor from 2.x to 3.x.

### Related Issues

https://github.com/opensearch-project/k-NN/issues/1949

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/engine/MethodComponent.java
+++ b/src/main/java/org/opensearch/knn/index/engine/MethodComponent.java
@@ -11,6 +11,8 @@ import org.opensearch.common.TriFunction;
 import org.opensearch.common.ValidationException;
 import org.opensearch.knn.common.KNNConstants;
 import org.opensearch.knn.index.VectorDataType;
+import org.opensearch.knn.index.mapper.CompressionLevel;
+import org.opensearch.knn.index.mapper.Mode;
 import org.opensearch.knn.index.util.IndexHyperParametersUtil;
 
 import java.util.HashMap;
@@ -328,6 +330,15 @@ public class MethodComponent {
         Map<String, Object> parametersWithDefaultsMap = new HashMap<>();
         Map<String, Object> userProvidedParametersMap = methodComponentContext.getParameters();
         Version indexCreationVersion = knnMethodConfigContext.getVersionCreated();
+        Mode mode = knnMethodConfigContext.getMode();
+        CompressionLevel compressionLevel = knnMethodConfigContext.getCompressionLevel();
+
+        // Check if the mode is ON_DISK and the compression level is one of the binary quantization levels (x32, x16, or x8).
+        // This determines whether to use binary quantization-specific values for parameters like ef_search and ef_construction.
+        boolean isOnDiskWithBinaryQuantization = (compressionLevel == CompressionLevel.x32
+            || compressionLevel == CompressionLevel.x16
+            || compressionLevel == CompressionLevel.x8);
+
         for (Parameter<?> parameter : methodComponent.getParameters().values()) {
             if (methodComponentContext.getParameters().containsKey(parameter.getName())) {
                 parametersWithDefaultsMap.put(parameter.getName(), userProvidedParametersMap.get(parameter.getName()));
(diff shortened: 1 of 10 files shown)


src/main/java/org/opensearch/knn/index/engine/MethodComponent.java
```diff
@@ -328,19 +330,42 @@ public static Map<String, Object> getParameterMapWithDefaultsAdded(
         Map<String, Object> parametersWithDefaultsMap = new HashMap<>();
         Map<String, Object> userProvidedParametersMap = methodComponentContext.getParameters();
         Version indexCreationVersion = knnMethodConfigContext.getVersionCreated();
+        Mode mode = knnMethodConfigContext.getMode();
+        CompressionLevel compressionLevel = knnMethodConfigContext.getCompressionLevel();
+
+        boolean isOnDiskWithBinaryQuantization = mode == Mode.ON_DISK
```
navneet1v, line 336: why we are checking for on_disk? we should just look for compression parameter to set these parameters.

navneet1v, line 336: also lets add some comment on top of it around what it means.

src/main/java/org/opensearch/knn/index/engine/MethodComponent.java
```diff
@@ -328,19 +330,42 @@ public static Map<String, Object> getParameterMapWithDefaultsAdded(
...
+            && (compressionLevel.equals(CompressionLevel.x32)
+                || compressionLevel.equals(CompressionLevel.x16)
+                || compressionLevel.equals(CompressionLevel.x8));
```
navneet1v, line 339: better to use == rather than equals.

src/main/java/org/opensearch/knn/index/engine/MethodComponent.java
Vikasht34: I was trying to safe ...and going was definition...It does not harm ..

src/main/java/org/opensearch/knn/index/engine/MethodComponent.java
```diff
@@ -328,19 +330,42 @@ public static Map<String, Object> getParameterMapWithDefaultsAdded(
         Map<String, Object> parametersWithDefaultsMap = new HashMap<>();
         Map<String, Object> userProvidedParametersMap = methodComponentContext.getParameters();
         Version indexCreationVersion = knnMethodConfigContext.getVersionCreated();
+        Mode mode = knnMethodConfigContext.getMode();
+        CompressionLevel compressionLevel = knnMethodConfigContext.getCompressionLevel();
```
navneet1v, line 334: if mode and compression is not used and just encoder is provided then how we are changing the hyper parameters?

src/main/java/org/opensearch/knn/index/engine/MethodComponent.java
navneet1v: but then it is not consistent. The idea is BQ is lossy so we want to increase the hyper parameter values.

src/main/java/org/opensearch/knn/index/engine/MethodComponent.java
Vikasht34: this isOnDiskWithBinaryQuantization will be false ..and flow as it is like now

Vikasht34: that's why I was very specific to BinaryQuantization

navneet1v: Ref: https://github.com/opensearch-project/k-NN/pull/2087/files#r1755277111

src/main/java/org/opensearch/knn/index/engine/MethodComponent.java
```diff
@@ -328,19 +330,44 @@ public static Map<String, Object> getParameterMapWithDefaultsAdded(
...
                 if (parameter.getName().equals(KNNConstants.METHOD_PARAMETER_EF_SEARCH)) {
-                    parametersWithDefaultsMap.put(parameter.getName(), IndexHyperParametersUtil.getHNSWEFSearchValue(indexCreationVersion));
+                    if (isOnDiskWithBinaryQuantization) {
+                        parametersWithDefaultsMap.put(parameter.getName(), IndexHyperParametersUtil.getBinaryQuantizationEFSearchValue());
+                    } else {
+                        parametersWithDefaultsMap.put(
+                            parameter.getName(),
+                            IndexHyperParametersUtil.getHNSWEFSearchValue(indexCreationVersion)
+                        );
+                    }
```
shatejas, line 357: Would prefer this logic inside `IndexHyperParametersUtils` for better readability and testability. It can take compression and version (maybe mode if required) and give the right values

src/main/java/org/opensearch/knn/index/engine/MethodComponent.java
Vikasht34: Addressed!!

src/main/java/org/opensearch/knn/index/engine/MethodComponent.java
Vikasht34: Addressed

src/main/java/org/opensearch/knn/index/engine/MethodComponent.java
Vikasht34: Make Sense , but I would keep this PR same as current flow , Would take it after this release.

navneet1v (approved):

naveentatikonda (approved):

PR #2110, opened 2024-09-16
Add changes for AVX-512 support in k-NN. 

### LINKED ISSUE #2056: [FEATURE] Add support for FAISS AVX512
Improve performance of vector search and indexing when using the FAISS library, by using AVX512 accelerators which are supported in faiss. 

Add support to use AVX512, initially as an optional feature and eventually enable by default on machines where AVX512 is supported on subsequent OpenSearch releases.
This will be a 2 part change: 
1) Adds necessary code in k-NN to check for AVX512 support on the machine.
2) A separate change made in FAISS to add missing functions support for AVX512.

akashsha1:
### Description
This change adds support to speed up vector search and indexing in faiss using AVX512 hardware accelerator.

### Related Issues
Resolves #2056 

### Check List
- [x] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose). https://github.com/opensearch-project/documentation-website/pull/8307

### DIFF
--- a/DEVELOPER_GUIDE.md
+++ b/DEVELOPER_GUIDE.md
@@ -283,19 +283,22 @@ make -j 4
 
 ### Enable SIMD Optimization
 SIMD(Single Instruction/Multiple Data) Optimization is enabled by default on Linux and Mac which boosts the performance
-by enabling `AVX2` on `x86 architecture` and `NEON` on `ARM64 architecture` while building the Faiss library. But to enable SIMD, the underlying processor
-should support this (AVX2 or NEON). It can be disabled by setting the parameter `simd.enabled` to `false`. As of now, it is not supported on Windows OS.
+by enabling `AVX2` and `AVX512` on `x86 architecture` and `NEON` on `ARM64 architecture` where applicable while building the Faiss library. But to enable SIMD,
+the underlying processor should support these capabilities (AVX512, AVX2 or NEON). It can be disabled by setting the parameter `avx2.enabled` to `false` and
+`avx512.enabled` to `false`. If your processor supports `AVX512` or `AVX2`, they can be set by enabling the setting . By default, these values are enabled on 
+OpenSearch. Some exceptions: As of now, SIMD support is not supported on Windows OS, and AVX512 is not present on MAC systems due to hardware not supporting the
+feature.
 
 ```
 # While building OpenSearch k-NN
-./gradlew build -Dsimd.enabled=true
+./gradlew build -Davx2.enabled=true -Davx512.enabled=true
 
 # While running OpenSearch k-NN
-./gradlew run -Dsimd.enabled=true
+./gradlew run -Davx2.enabled=true -Davx512.enabled=true
 
 # While building the JNI libraries
 cd jni
-cmake . -DSIMD_ENABLED=true
+cmake . -DAVX2_ENABLED=true -DAVX512_ENABLED=true
 ```
 
 ## Run OpenSearch k-NN
--- a/src/main/java/org/opensearch/knn/index/KNNSettings.java
+++ b/src/main/java/org/opensearch/knn/index/KNNSettings.java
@@ -512,20 +512,12 @@ public class KNNSettings {
         }
     }
 
-    public static boolean isFaissAVX512Disabled() {
-        try {
-            return KNNSettings.state().getSettingValue(KNNSettings.KNN_FAISS_AVX512_DISABLED);
-        } catch (Exception e) {
-            // In some UTs we identified that cluster setting is not set properly an leads to NPE. This check will avoid
-            // those cases and will still return the default value.
-            log.warn(
-                "Unable to get setting value {} from cluster settings. Using default value as {}",
-                KNN_FAISS_AVX512_DISABLED,
-                KNN_DEFAULT_FAISS_AVX512_DISABLED_VALUE,
-                e
-            );
-            return KNN_DEFAULT_FAISS_AVX512_DISABLED_VALUE;
-        }
+    public static boolean isFaissAVX512Disabled() {        
+            String sValue = KNNSettings.state().getSettingValue(KNNSettings.KNN_FAISS_AVX512_DISABLED);
+            if (sValue != null) {
+                return KNNSettings.state().getSettingValue(KNNSettings.KNN_FAISS_AVX512_DISABLED);
+            }
+            return KNN_DEFAULT_FAISS_AVX512_DISABLED_VALUE;        
     }
 
     public static Integer getFilteredExactSearchThreshold(final String indexName) {
--- a/src/main/java/org/opensearch/knn/index/KNNSettings.java
+++ b/src/main/java/org/opensearch/knn/index/KNNSettings.java
@@ -14,6 +14,7 @@ import org.opensearch.action.admin.cluster.settings.ClusterUpdateSettingsRespons
 import org.opensearch.client.Client;
 import org.opensearch.cluster.metadata.IndexMetadata;
 import org.opensearch.cluster.service.ClusterService;
+import org.opensearch.common.Booleans;
 import org.opensearch.common.settings.Setting;
 import org.opensearch.common.settings.Settings;
 import org.opensearch.common.unit.TimeValue;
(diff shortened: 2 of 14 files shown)


assanedi:
Benchmark was run using opensearch-benchmark with cohere dataset(768 dimensions). 
The opensearch cluster was deployed with 2 data nodes (r7i.2xlarges),  1 replica and 4 shards. 
Using this setup AVX512 shows 15% improvement over AVX2 as shown below: 

image

naveentatikonda:
> Benchmark was run using opensearch-benchmark with cohere dataset(768 dimensions). The opensearch cluster was deployed with 2 data nodes (r7i.2xlarges), 1 replica and 4 shards. Using this setup AVX512 shows 15% improvement over AVX2 on indexing and 7 % on search as shown below:
> 
> image

@assanedi Can you also pls add other configuration details like the indexing clients, query clients, ef_construction, ef_search, etc

naveentatikonda (dismissed):
LGTM! Thanks @akashsha1 @assanedi

naveentatikonda:
> "target_index_bulk_indexing_clients": 20,
"target_index_max_num_segments": 10,
"hnsw_ef_search": 256,
"hnsw_ef_construction": 256

@assanedi Isn't the max_num_segments was 1 during forcemerge ?

assanedi:
> > "target_index_bulk_indexing_clients": 20,
> > "target_index_max_num_segments": 10,
> > "hnsw_ef_search": 256,
> > "hnsw_ef_construction": 256
> 
> @assanedi Isn't the max_num_segments was 1 during forcemerge ?

Yes I run the forcemerge API, here is the results of it:
curl -X POST -k --user admin:admin http://10.0.0.80:9200/_forcemerge?max_num_segments=1
{"_shards":{"total":8,"successful":8,"failed":0}}

naveentatikonda:
> 
> Yes I run the forcemerge API, here is the results of it: curl -X POST -k --user admin:admin http://10.0.0.80:9200/_forcemerge?max_num_segments=1 {"_shards":{"total":8,"successful":8,"failed":0}}

Yes, but in the configuration you mentioned it as 10 instead of 1 for `target_index_max_num_segments `

DEVELOPER_GUIDE.md
```diff
@@ -283,19 +283,22 @@ make -j 4
...
+by enabling `AVX2` and `AVX512` on `x86 architecture` and `NEON` on `ARM64 architecture` where applicable while building the Faiss library. But to enable SIMD,
+the underlying processor should support these capabilities (AVX512, AVX2 or NEON). It can be disabled by setting the parameter `avx2.enabled` to `false` and
+`avx512.enabled` to `false`. If your processor supports `AVX512` or `AVX2`, they can be set by enabling the setting . By default, these values are enabled on 
+OpenSearch. Some exceptions: As of now, SIMD support is not supported on Windows OS, and AVX512 is not present on MAC systems due to hardware not supporting the
+feature.
 
 ```
 # While building OpenSearch k-NN
-./gradlew build -Dsimd.enabled=true
+./gradlew build -Davx2.enabled=true -Davx512.enabled=true
```
jmazanec15, line 294: Shouldnt these be mutually exclusive?

naveentatikonda:
For FP32 we don’t need to make any changes in Faiss as they are using auto-vectorization to achieve the optimization with AVX512. But, for Scalar Quantization Intel have raised a PR to Faiss which is under review
https://github.com/facebookresearch/faiss/pull/3853

src/main/java/org/opensearch/knn/index/KNNSettings.java
```diff
@@ -499,6 +512,22 @@ public static boolean isFaissAVX2Disabled() {
         }
     }
 
+    public static boolean isFaissAVX512Disabled() {
+        try {
+            return KNNSettings.state().getSettingValue(KNNSettings.KNN_FAISS_AVX512_DISABLED);
```
jmazanec15, line 517: Can we do proper null checks here? In general, I think its best to avoid catching all exceptions.

src/main/java/org/opensearch/knn/jni/FaissService.java
```diff
@@ -30,14 +32,17 @@
  *      src/main/java/org/opensearch/knn/index/query/KNNQueryResult.java
  *      src/main/java/org/opensearch/knn/common/KNNConstants.java
  */
+@SuppressWarnings("deprecation")
```
jmazanec15, line 35: Why was this needed?

src/main/java/org/opensearch/knn/jni/PlatformUtils.java
```diff
@@ -80,4 +81,38 @@ public static boolean isAVX2SupportedBySystem() {
         }
         return false;
     }
+
+    public static boolean isAVX512SupportedBySystem() {
+
+        if (!Platform.isIntel() || Platform.isMac() || Platform.isWindows()) {
```
jmazanec15, line 87: does mac support avx2 but not avx512?

src/main/java/org/opensearch/knn/jni/PlatformUtils.java
```diff
@@ -80,4 +81,38 @@ public static boolean isAVX2SupportedBySystem() {
...
+
+            try {
+                return AccessController.doPrivileged((PrivilegedExceptionAction<Boolean>) () -> {
+                    Stream<String> linestream = Files.lines(Paths.get(fileName));
+                    String flags = linestream.filter(line -> line.startsWith("flags")).findFirst().orElse("");
+                    return Arrays.stream(avx512).allMatch(flags::contains);
+                });
+
+            } catch (Exception e) {
+                logger.error("[KNN] Error reading file [{}]. [{}]", fileName, e.getMessage(), e);
```
jmazanec15, line 113: Same as above

src/main/java/org/opensearch/knn/jni/PlatformUtils.java
akashsha1: that's right. mac moved away from x86 right around the time avx512 was introduced. 
I don't have mac systems to verify this, and @naveentatikonda confirmed his mac system doesn't have avx512. 
if you know of any mac systems supporting avx512, I can update the check here.

src/main/java/org/opensearch/knn/jni/FaissService.java
akashsha1: added by VSCode accidentally. Fixed with the latest commit.

src/main/java/org/opensearch/knn/jni/PlatformUtils.java
akashsha1: updated to PrivilegedActionException. 
However, do note that AccessController API itself is deprecated by Java, and hence the IDE prompts the warning. This'll require a broader change across the codebase.

DEVELOPER_GUIDE.md
akashsha1: Ideally we want simd (avx2, avx512 - highest of whichever is present) to be enabled by default. The order of checks would be: 
if (AVX512 enabled and present) { use avx512 } 
else if (AVX2 enabled and present) { use avx2 }
else { use generic version }

by making it mutually exclusive, the intermediate step cannot be achieved.

naveentatikonda: I think Jack's point is to update it as something like this `./gradlew build -Davx2.enabled=false -Davx512.enabled=true` as we can't use both of them at the same time

src/main/java/org/opensearch/knn/index/KNNSettings.java
akashsha1: Looks like java boolean cannot be null. So null check won't be possible. 
Your second point on exceptions is valid, and this code shouldn't throw exceptions as a default value is set. I've removed the try/catch block.

naveentatikonda: @akashsha1 as users will manually set this setting in opensearch.yml, there is a chance of getting null. To avoid it shall we change it to 
`return Booleans.parseBoolean(KNNSettings.state().getSettingValue(KNNSettings.KNN_FAISS_AVX512_DISABLED).toString(), false);`

naveentatikonda: `Booleans.parseBoolean` will do the null validation and if it is null, it will return the default value

akashsha1: spoke with Naveen on slack, and updated to 
return Booleans.parseBoolean(KNNSettings.state().getSettingValue(KNNSettings.KNN_FAISS_AVX512_DISABLED).toString(), KNN_DEFAULT_FAISS_AVX512_DISABLED_VALUE);

jmazanec15 (approved):
LGTM thanks

naveentatikonda (approved):

PR #2111, opened 2024-09-17
Integrates KNN plugin with ConcurrentSearchRequestDecider interface

shatejas:
This allows knn queries to enable concurrency when `index.search.concurrent_segment_search.mode` or `search.concurrent_segment_search.mode` in `auto` mode. Without this the default behavior of auto mode is non-concurrent search

### Description
More details in https://github.com/opensearch-project/OpenSearch/issues/15259

#### Functional
- Manually tested with debugger to make sure `ConcurrentQueryPhaseSearcher` is used
- Unit test and sanity IntegTest added

#### Performance
**Baseline (no settings update)**

50th percentile latency,prod-queries,58.86101468404134,ms
90th percentile latency,prod-queries,85.32014465332031,ms
99th percentile latency,prod-queries,95.10733413696289,ms

**Concurrent_segment_search.mode: auto**

50th percentile latency,prod-queries,45.04893729613377,ms
90th percentile latency,prod-queries,48.14952836545217,ms
99th percentile latency,prod-queries,50.41835594177246,ms
   
### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/plugin/search/KNNConcurrentSearchRequestDecider.java
+++ b/src/main/java/org/opensearch/knn/plugin/search/KNNConcurrentSearchRequestDecider.java
@@ -1,0 +1,65 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.knn.plugin.search;
+
+import lombok.EqualsAndHashCode;
+import org.opensearch.index.IndexSettings;
+import org.opensearch.index.query.QueryBuilder;
+import org.opensearch.knn.index.KNNSettings;
+import org.opensearch.knn.index.query.KNNQueryBuilder;
+import org.opensearch.search.deciders.ConcurrentSearchDecision;
+import org.opensearch.search.deciders.ConcurrentSearchRequestDecider;
+
+import java.util.Optional;
+
+/**
+ * Decides if the knn query uses concurrent segment search
+ * As of 2.17, this is only used when
+ *  - "index.search.concurrent_segment_search.mode": "auto" or
+ *  - "search.concurrent_segment_search.mode": "auto"
+ *
+ *  Note: the class is not thread-safe and a new instance needs to be created for each request
+ */
+@EqualsAndHashCode(callSuper = true)
+public class KNNConcurrentSearchRequestDecider extends ConcurrentSearchRequestDecider {
+
+    private static final ConcurrentSearchDecision DEFAULT_KNN_DECISION = new ConcurrentSearchDecision(
+        ConcurrentSearchDecision.DecisionStatus.NO_OP,
+        "Default decision"
+    );
+    private static final ConcurrentSearchDecision YES = new ConcurrentSearchDecision(
+        ConcurrentSearchDecision.DecisionStatus.YES,
+        "Enable concurrent search for knn as Query has k-NN query in it and index is k-nn index"
+    );
+
+    private ConcurrentSearchDecision knnDecision = DEFAULT_KNN_DECISION;
+
+    @Override
+    public void evaluateForQuery(final QueryBuilder queryBuilder, final IndexSettings indexSettings) {
+        if (queryBuilder instanceof KNNQueryBuilder && indexSettings.getValue(KNNSettings.IS_KNN_INDEX_SETTING)) {
+            knnDecision = YES;
+        } else {
+            knnDecision = DEFAULT_KNN_DECISION;
+        }
+    }
+
+    @Override
+    public ConcurrentSearchDecision getConcurrentSearchDecision() {
+        return knnDecision;
+    }
+
+    /**
+     * Returns {@link KNNConcurrentSearchRequestDecider} when index.knn is true
+     */
+    public static class Factory implements ConcurrentSearchRequestDecider.Factory {
+        public Optional<ConcurrentSearchRequestDecider> create(final IndexSettings indexSettings) {
+            if (indexSettings.getValue(KNNSettings.IS_KNN_INDEX_SETTING)) {
+                return Optional.of(new KNNConcurrentSearchRequestDecider());
+            }
+            return Optional.empty();
+        }
+    }
+}
--- a/src/test/java/org/opensearch/knn/integ/search/ConcurrentSegmentSearchIT.java
+++ b/src/test/java/org/opensearch/knn/integ/search/ConcurrentSegmentSearchIT.java
@@ -1,0 +1,141 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.knn.integ.search;
+
+import com.google.common.primitives.Floats;
+import lombok.SneakyThrows;
+import org.apache.hc.core5.http.io.entity.EntityUtils;
+import org.junit.BeforeClass;
+import org.opensearch.client.Response;
+import org.opensearch.common.settings.Settings;
+import org.opensearch.core.xcontent.XContentBuilder;
+import org.opensearch.knn.KNNJsonIndexMappingsBuilder;
+import org.opensearch.knn.KNNRestTestCase;
+import org.opensearch.knn.KNNResult;
+import org.opensearch.knn.TestUtils;
+import org.opensearch.knn.index.SpaceType;
+import org.opensearch.knn.index.engine.KNNEngine;
+import org.opensearch.knn.index.query.KNNQueryBuilder;
+import org.opensearch.knn.plugin.script.KNNScoringUtil;
+
+import java.io.IOException;
+import java.net.URL;
+import java.util.List;
+import java.util.Map;
+import java.util.TreeMap;
+
+import static org.opensearch.knn.common.KNNConstants.METHOD_HNSW;
+
+/**
+ * Note that this is simply a sanity test to make sure that concurrent search code path is hit E2E and scores are intact
+ * There is no latency verification as it can be better encapsulated in nightly runs.
+ */
+public class ConcurrentSegmentSearchIT extends KNNRestTestCase {
+
+    static TestUtils.TestData testData;
+
+    @BeforeClass
+    public static void setUpClass() throws IOException {
+        if (ConcurrentSegmentSearchIT.class.getClassLoader() == null) {
+            throw new IllegalStateException("ClassLoader of ConcurrentSegmentSearchIT Class is null");
+        }
+        URL testIndexVectors = ConcurrentSegmentSearchIT.class.getClassLoader().getResource("data/test_vectors_1000x128.json");
+        URL testQueries = ConcurrentSegmentSearchIT.class.getClassLoader().getResource("data/test_queries_100x128.csv");
+        assert testIndexVectors != null;
+        assert testQueries != null;
+        testData = new TestUtils.TestData(testIndexVectors.getPath(), testQueries.getPath());
+    }
+
+    @SneakyThrows
+    public void testConcurrentSegmentSearch_thenSucceed() {
+        String indexName = "test-concurrent-segment";
+        String fieldName = "test-field-1";
+        int dimension = testData.indexData.vectors[0].length;
+        final XContentBuilder indexBuilder = createFaissHnswIndexMapping(fieldName, dimension);
+        Map<String, Object> mappingMap = xContentBuilderToMap(indexBuilder);
+        String mapping = indexBuilder.toString();
+        createKnnIndex(indexName, mapping);
+        assertEquals(new TreeMap<>(mappingMap), new TreeMap<>(getIndexMappingAsMap(indexName)));
+
+        // Index the test data
+        for (int i = 0; i < testData.indexData.docs.length; i++) {
+            addKnnDoc(
+                indexName,
+                Integer.toString(testData.indexData.docs[i]),
+                fieldName,
+                Floats.asList(testData.indexData.vectors[i]).toArray()
+            );
+        }
+        refreshAllNonSystemIndices();
+        updateIndexSettings(indexName, Settings.builder().put("index.search.concurrent_segment_search.mode", "auto"));
+
+        // Test search queries
+        int k = 10;
+        verifySearch(indexName, fieldName, k);
+
+        updateIndexSettings(indexName, Settings.builder().put("index.search.concurrent_segment_search.mode", "all"));
+        verifySearch(indexName, fieldName, k);
+
+        deleteKNNIndex(indexName);
+    }
+
+    /*
+    {
+      "properties": {
+        "<fieldName>": {
+            "type": "knn_vector",
+            "dimension": <dimension>,
+            "method": {
+            "name": "hnsw",
+            "space_type": "l2",
+            "engine": "faiss",
+            "parameters": {
+                "m": 16,
+                "ef_construction": 128,
+                "ef_search": 128
+            }
+          }
+        }
+      }
+     */
+    @SneakyThrows
+    private XContentBuilder createFaissHnswIndexMapping(String fieldName, int dimension) {
+        return KNNJsonIndexMappingsBuilder.builder()
+            .fieldName(fieldName)
+            .dimension(dimension)
+            .method(
+                KNNJsonIndexMappingsBuilder.Method.builder()
+                    .engine(KNNEngine.FAISS.getName())
+                    .methodName(METHOD_HNSW)
+                    .spaceType(SpaceType.L2.getValue())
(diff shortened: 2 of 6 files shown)


navneet1v:
@shatejas for perf runs will need more details:
1. How many shards were used?
2. What was the dataset used?
3. What was the machine used?

src/main/java/org/opensearch/knn/plugin/search/KNNConcurrentSearchRequestDecider.java
```diff
@@ -0,0 +1,58 @@
...
+import org.opensearch.index.IndexSettings;
+import org.opensearch.index.query.QueryBuilder;
+import org.opensearch.knn.index.query.KNNQueryBuilder;
+import org.opensearch.search.deciders.ConcurrentSearchDecision;
+import org.opensearch.search.deciders.ConcurrentSearchRequestDecider;
+
+import java.util.Optional;
+
+/**
+ * Decides if the knn query uses concurrent search
```
navneet1v, line 18:
```suggestion
 * Decides if the knn query uses concurrent segment search
```

src/main/java/org/opensearch/knn/plugin/search/KNNConcurrentSearchRequestDecider.java
```diff
@@ -0,0 +1,58 @@
...
+        ConcurrentSearchDecision.DecisionStatus.YES,
+        "Enable concurrent search for knn"
+    );
+
+    private ConcurrentSearchDecision knnDecision = DEFAULT_KNN_DECISION;
+
+    @Override
+    public void evaluateForQuery(final QueryBuilder queryBuilder, final IndexSettings indexSettings) {
+        if (queryBuilder instanceof KNNQueryBuilder) {
+            knnDecision = YES;
```
navneet1v, line 42: we should also check index.knn = true here because a user can send a k-NN query even when its index.knn is false.

src/main/java/org/opensearch/knn/plugin/search/KNNConcurrentSearchRequestDecider.java
```diff
@@ -0,0 +1,58 @@
...
+            knnDecision = DEFAULT_KNN_DECISION;
+        }
+    }
+
+    @Override
+    public ConcurrentSearchDecision getConcurrentSearchDecision() {
+        return knnDecision;
+    }
+
+    public static class Factory implements ConcurrentSearchRequestDecider.Factory {
```
navneet1v, line 53: add java docs,

shatejas:
> * How many shards were used?

1 shard

> * What was the dataset used?

cohere-1m

> * What was the machine used?

This was run on docker with this configuration
JVM=36g
CPU_COUNT=8
MEM_SIZE=48g

src/main/java/org/opensearch/knn/plugin/search/KNNConcurrentSearchRequestDecider.java
```diff
@@ -0,0 +1,58 @@
...
+        public Optional<ConcurrentSearchRequestDecider> create(IndexSettings indexSettings) {
+            return Optional.of(new KNNConcurrentSearchRequestDecider());
+        }
```
navneet1v, line 56: I think we should not always send the new instance of `KNNConcurrentSearchRequestDecider`. Only if indexSettings has index.knn = true we should send the new instance. Otherwise even for non k-NN indices this create method will be called, which means that for every querybuilder `evaluateForQuery` function of KNNConcurrentSearchRequestDecider class.

navneet1v (changes requested):

navneet1v:
> JVM=36g

Curious on why JVM was 36g?

shatejas:
> > JVM=36g
> 
> Curious on why JVM was 36g?

Didn't think memory made a difference as there is no change in amount of threads (as of 2.17) allocated by OS when concurrent segment search code path is used, so gave sufficient memory

src/main/java/org/opensearch/knn/plugin/search/KNNConcurrentSearchRequestDecider.java
shatejas: Was of the assumption that index.knn is not set when IVF is used, turns out its not true

src/main/java/org/opensearch/knn/plugin/search/KNNConcurrentSearchRequestDecider.java
navneet1v: @shatejas before resolving the comment can you add a comment if you are going to make the change or not?

src/main/java/org/opensearch/knn/plugin/search/KNNConcurrentSearchRequestDecider.java
navneet1v: all indices of knn has index.knn as true. Only training index can have index.knn as false.

src/main/java/org/opensearch/knn/plugin/search/KNNConcurrentSearchRequestDecider.java
shatejas: @navneet1v Resolved after I made the change :) 

https://github.com/opensearch-project/k-NN/pull/2111/files#diff-3ecc756cf2b8824250072443c0a8d1953227eccefcfce7216848290d03de81bbR59

src/test/java/org/opensearch/knn/integ/search/ConcurrentSegmentSearchIT.java
```diff
@@ -0,0 +1,151 @@
...
+ */
+public class ConcurrentSegmentSearchIT extends KNNRestTestCase {
+
+    static TestUtils.TestData testData;
+
+    @BeforeClass
+    public static void setUpClass() throws IOException {
+        if (ConcurrentSegmentSearchIT.class.getClassLoader() == null) {
+            throw new IllegalStateException("ClassLoader of FaissIT Class is null");
+        }
```
navneet1v, line 49: didn't understand what we have this and exception says ClassLoader of FaissIT class is null

src/test/java/org/opensearch/knn/integ/search/ConcurrentSegmentSearchIT.java
```diff
@@ -0,0 +1,151 @@
...
+        URL testIndexVectors = ConcurrentSegmentSearchIT.class.getClassLoader().getResource("data/test_vectors_1000x128.json");
+        URL testQueries = ConcurrentSegmentSearchIT.class.getClassLoader().getResource("data/test_queries_100x128.csv");
+        assert testIndexVectors != null;
+        assert testQueries != null;
+        testData = new TestUtils.TestData(testIndexVectors.getPath(), testQueries.getPath());
+    }
+
+    @SneakyThrows
+    public void testConcurrentSegmentSearch() {
```
navneet1v, line 58: better to use the convention testABC_whenLMN_thenXYZ()

src/test/java/org/opensearch/knn/integ/search/ConcurrentSegmentSearchIT.java
```diff
@@ -0,0 +1,151 @@
...
+                "m": 16,
+                "ef_construction": 128,
+                "ef_search": 128
+            }
+          }
+        }
+      }
+     */
+    @SneakyThrows
+    private XContentBuilder createFaissHnswIndexMapping(String fieldName, int dimension) {
```
navneet1v, line 108: You can do something like this:
```
createKnnIndex(testIndex, getKNNDefaultIndexSettings(), createKnnIndexMapping(TEST_FIELD, DIMENSIONS));
```

I would suggest checking KNNRestTestCase.java class as it has a lot of helper functions to create the index and not create another function in ITs for creating the mappings.

src/test/java/org/opensearch/knn/integ/search/ConcurrentSegmentSearchIT.java
```diff
@@ -0,0 +1,151 @@
...
+            );
+        }
+        refreshAllNonSystemIndices();
+        updateIndexSettings(indexName, Settings.builder().put("index.search.concurrent_segment_search.mode", "auto"));
+
+        // Test search queries
+        int k = 10;
+        verifySearch(indexName, fieldName, k);
+
+        updateIndexSettings(indexName, Settings.builder().put("index.search.concurrent_segment_search.mode", "all"));
```
navneet1v, line 84: the indexsetting should be reverted back to default otherwise for all other tests CSS will start to work.

src/test/java/org/opensearch/knn/plugin/search/KNNConcurrentSearchRequestDeciderTests.java
```diff
@@ -0,0 +1,65 @@
...
+import org.opensearch.knn.index.KNNSettings;
+import org.opensearch.knn.index.query.KNNQueryBuilder;
+import org.opensearch.search.deciders.ConcurrentSearchDecision;
+
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+public class KNNConcurrentSearchRequestDeciderTests extends KNNTestCase {
+
+    public void testDecider() {
```
navneet1v, line 20: same as above use testABC_whenLMN_thenXYZ() for all the test functions

src/main/java/org/opensearch/knn/plugin/search/KNNConcurrentSearchRequestDecider.java
```diff
@@ -0,0 +1,65 @@
...
+@EqualsAndHashCode(callSuper = true)
+public class KNNConcurrentSearchRequestDecider extends ConcurrentSearchRequestDecider {
+
+    private static final ConcurrentSearchDecision DEFAULT_KNN_DECISION = new ConcurrentSearchDecision(
+        ConcurrentSearchDecision.DecisionStatus.NO_OP,
+        "Default decision"
+    );
+    private static final ConcurrentSearchDecision YES = new ConcurrentSearchDecision(
+        ConcurrentSearchDecision.DecisionStatus.YES,
+        "Enable concurrent search for knn"
```
navneet1v, line 35:
```suggestion
        "Enable concurrent search for knn as Query has k-NN query in it and index is k-nn index"
```

src/main/java/org/opensearch/knn/plugin/search/KNNConcurrentSearchRequestDecider.java
```diff
@@ -0,0 +1,65 @@
...
+
+import java.util.Optional;
+
+/**
+ * Decides if the knn query uses concurrent segment search
+ * As of 2.17, this is only used when
+ *  - "index.search.concurrent_segment_search.mode": "auto" or
+ *  - "search.concurrent_segment_search.mode": "auto"
+ *
+ *  Note: the class is not thread-safe and a new instance needs to be created for each request
```
jmazanec15, line 24: Why not just making this one singleton? Is any state being modified?

src/main/java/org/opensearch/knn/plugin/search/KNNConcurrentSearchRequestDecider.java
```diff
@@ -0,0 +1,65 @@
...
+    private static final ConcurrentSearchDecision YES = new ConcurrentSearchDecision(
+        ConcurrentSearchDecision.DecisionStatus.YES,
+        "Enable concurrent search for knn"
+    );
+
+    private ConcurrentSearchDecision knnDecision = DEFAULT_KNN_DECISION;
+
+    @Override
+    public void evaluateForQuery(final QueryBuilder queryBuilder, final IndexSettings indexSettings) {
+        if (queryBuilder instanceof KNNQueryBuilder && indexSettings.getValue(KNNSettings.IS_KNN_INDEX_SETTING)) {
```
jmazanec15, line 42: Will we be able to look into cpu and memory? Is this done in a chain like fashion?

shatejas: > Will we be able to look into cpu and memory

The control is with core and core decides based on CPU and couple other conditions. I don't think they have any immediate plans to give that control to plugin as per the [RFC](https://github.com/opensearch-project/OpenSearch/issues/14781) and plugin [RFC](https://github.com/opensearch-project/OpenSearch/issues/15259)

> Is this done in a chain like fashion?

I don't completely understand the question. There will be multiple deciders based on number of plugins, it will loop through each of them and evaluate the query builder with visitor pattern using `QueryBuilder.visit` which evaluates and sets a decision. Even if one decider decides `NO` it **will not**  execute concurrent search. Let me know if that answers your question

src/main/java/org/opensearch/knn/plugin/search/KNNConcurrentSearchRequestDecider.java
shatejas: Unfortunately we can't, The plugin gives a factory extension and creates a new object which maintains state for a request, So returning same object for all requests can cause race conditions with unpredictable results.

src/main/java/org/opensearch/knn/plugin/search/KNNConcurrentSearchRequestDecider.java
shatejas: This can keep changing if the logic changes, so I chose to be concise

src/test/java/org/opensearch/knn/integ/search/ConcurrentSegmentSearchIT.java
shatejas: Its using a specific index name for the IT to make sure no other tests are affected. 

Will delete the index to be safe

src/main/java/org/opensearch/knn/plugin/search/KNNConcurrentSearchRequestDecider.java
navneet1v: In that case we should update this message. because right now if we keep it like this, we don't know on what bases k-NN said CSS should be true.

src/main/java/org/opensearch/knn/plugin/search/KNNConcurrentSearchRequestDecider.java
navneet1v: @jmazanec15 the decider Object is at per request level. So, if it is singleton then how will be the decision made?

src/test/java/org/opensearch/knn/integ/search/ConcurrentSegmentSearchIT.java
navneet1v: sure.

src/test/java/org/opensearch/knn/integ/search/ConcurrentSegmentSearchIT.java
shatejas: > You can do something like this

The mapping is being asserted so I need to separate it out

> I would suggest checking KNNRestTestCase.java class as it has a lot of helper functions to create the index and not create another function in ITs for creating the mappings.

The requests for tests should be ideally localized to tests so we unknowingly don't affect multiple different test scenarios. Switching to `KNNJsonIndexMappingsBuilder` but I would prefer this mapping to be localized to this test

src/test/java/org/opensearch/knn/integ/search/ConcurrentSegmentSearchIT.java
shatejas: This is a similar approach as `FaissIT.java` so keeping the status quo will change the error message

src/main/java/org/opensearch/knn/plugin/search/KNNConcurrentSearchRequestDecider.java
jmazanec15: This decider object does not have any state. So the factory could return the same object every time.

jmazanec15: but Im fine creating a new object - seems like it could be risky to share if state is added in future.

navneet1v: No the object has the state. If you see getDeciderObject is returning the state which was decided on the evaluateQuery function.

shatejas: Resolved with @jmazanec15 offline 

Here is the link which helps understand why its per request. https://github.com/opensearch-project/OpenSearch/blob/4802d0d744cfd19827ab61c0807f557b768bd1aa/server/src/main/java/org/opensearch/search/DefaultSearchContext.java#L948

jmazanec15 (approved):
This looks good to me! Thanks!

navneet1v (approved):

jmazanec15 (approved):

Issue #2134, opened 2024-09-20
[BUG] Regression in cohere-10m force merge latency after switching to NativeEngines990KnnVectorsWriter


shatejas:
**What is the bug?**

After the switch to `NativeEngines990KnnVectorsWriter` we saw force merge latencies increased approximately by [20%](https://benchmarks.opensearch.org/app/dashboards#/view/f9ef2c40-ee11-11ee-a589-033f2c21734f?_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:now-15m,to:now))&_a=(description:'',filters:!(('$state':(store:appState),meta:(alias:!n,disabled:!f,index:'333e2950-edec-11ed-a453-51f8c622cf9a',key:query,negate:!f,type:custom,value:'%7B%22bool%22:%7B%22minimum_should_match%22:1,%22should%22:%5B%7B%22match_phrase%22:%7B%22user-tags.cluster-config%22:%22arm64-r6g.4xlarge-3-data-6-shards-1-replica-nmslib-cohere-10m%22%7D%7D,%7B%22match_phrase%22:%7B%22meta.tag_cluster-config%22:%22arm64-r6g.4xlarge-3-data-6-shards-1-replica-nmslib-cohere-10m%22%7D%7D%5D%7D%7D'),query:(bool:(minimum_should_match:1,should:!((match_phrase:(user-tags.cluster-config:arm64-r6g.4xlarge-3-data-6-shards-1-replica-nmslib-cohere-10m)),(match_phrase:(meta.tag_cluster-config:arm64-r6g.4xlarge-3-data-6-shards-1-replica-nmslib-cohere-10m))))))),fullScreenMode:!f,options:(hidePanelTitles:!f,useMargins:!t),query:(language:kuery,query:''),timeRestore:!f,title:'%5BVector%20Search%5D%20%5BNmslib%5D%201OM%20Cohere%20768D',viewMode:view)) in nightly runs

The increase has been consistent

**How can one reproduce the bug?**
Running a benchmark for force-merge against 2.17 vs 2.16 with cohere 10m dataset takes 2000seconds (~30 mins) more

**What is the expected behavior?**
It should take the same time or less

**What is your host/environment?**
[...]

PR #2135, opened 2024-09-23
Optimizes live docs computes for force merge in NativeEngines990KNNVectorWriter

shatejas:
### Description
Currently live docs are computed with linear complexity even when it does not have deleted docs. This PR skips computing it

#### Testing 
- UTs and javadocs in progression. Need initial feed back

#### Related Issues
https://github.com/opensearch-project/k-NN/issues/2134

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/vectorvalues/KNNMergeVectorValues.java
+++ b/src/main/java/org/opensearch/knn/index/vectorvalues/KNNMergeVectorValues.java
@@ -1,0 +1,120 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.knn.index.vectorvalues;
+
+import org.apache.lucene.codecs.KnnVectorsReader;
+import org.apache.lucene.index.ByteVectorValues;
+import org.apache.lucene.index.DocIDMerger;
+import org.apache.lucene.index.FieldInfo;
+import org.apache.lucene.index.FloatVectorValues;
+import org.apache.lucene.index.MergeState;
+import org.apache.lucene.index.VectorEncoding;
+import org.apache.lucene.search.DocIdSetIterator;
+import org.apache.lucene.util.Bits;
+import org.apache.lucene.util.FixedBitSet;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Utility class to get merged VectorValues from MergeState
+ */
+public final class KNNMergeVectorValues {
+
+    /**
+     * Gets list of {@link KNNVectorValuesSub} for {@link FloatVectorValues} from a merge state and returns the iterator which
+     * iterates over live docs from all segments while mapping docIds.
+     *
+     * @param fieldInfo
+     * @param mergeState
+     * @return List of KNNVectorSub
+     * @throws IOException
+     */
+    public static KNNVectorValuesIterator.MergeFloat32VectorValuesIterator mergeFloatVectorValues(
+        FieldInfo fieldInfo,
+        MergeState mergeState
+    ) throws IOException {
+        assert fieldInfo != null && fieldInfo.hasVectorValues();
+        if (fieldInfo.getVectorEncoding() != VectorEncoding.FLOAT32) {
+            throw new UnsupportedOperationException("Cannot merge vectors encoded as [" + fieldInfo.getVectorEncoding() + "] as FLOAT32");
+        }
+        final List<KNNVectorValuesSub<FloatVectorValues>> subs = new ArrayList<>();
+        for (int i = 0; i < mergeState.knnVectorsReaders.length; i++) {
+            KnnVectorsReader knnVectorsReader = mergeState.knnVectorsReaders[i];
+            if (knnVectorsReader != null) {
+                FloatVectorValues values = knnVectorsReader.getFloatVectorValues(fieldInfo.getName());
+                if (values != null) {
+                    final Bits liveDocs = mergeState.liveDocs[i];
+                    final int liveDocsCardinality = cardinality(liveDocs, Math.toIntExact(values.cost()));
+                    subs.add(new KNNVectorValuesSub<>(mergeState.docMaps[i], values, liveDocsCardinality));
+                }
+            }
+        }
+        return new KNNVectorValuesIterator.MergeFloat32VectorValuesIterator(subs, mergeState);
+    }
+
+    /**
+     * Gets list of {@link KNNVectorValuesSub} for {@link ByteVectorValues} from a merge state. This can be further
+     * used to create an iterator for getting the docs and its vector values
+     * @param fieldInfo
+     * @param mergeState
+     * @return List of KNNVectorSub
+     * @throws IOException
+     */
+    public static KNNVectorValuesIterator.MergeByteVectorValuesIterator mergeByteVectorValues(FieldInfo fieldInfo, MergeState mergeState)
+        throws IOException {
+        assert fieldInfo != null && fieldInfo.hasVectorValues();
+        if (fieldInfo.getVectorEncoding() != VectorEncoding.BYTE) {
+            throw new UnsupportedOperationException("Cannot merge vectors encoded as [" + fieldInfo.getVectorEncoding() + "] as BYTE");
+        }
+        final List<KNNVectorValuesSub<ByteVectorValues>> subs = new ArrayList<>();
+        for (int i = 0; i < mergeState.knnVectorsReaders.length; i++) {
+            KnnVectorsReader knnVectorsReader = mergeState.knnVectorsReaders[i];
+            if (knnVectorsReader != null) {
+                ByteVectorValues values = knnVectorsReader.getByteVectorValues(fieldInfo.getName());
+                if (values != null) {
+                    final Bits liveDocs = mergeState.liveDocs[i];
+                    final int liveDocsCardinality = cardinality(liveDocs, Math.toIntExact(values.cost()));
+                    subs.add(new KNNVectorValuesSub<>(mergeState.docMaps[i], values, liveDocsCardinality));
+                }
+            }
+        }
+        return new KNNVectorValuesIterator.MergeByteVectorValuesIterator(subs, mergeState);
+    }
+
+    private static int cardinality(final Bits liveDocs, final int defaultCount) {
+        if (liveDocs == null) {
+            return defaultCount;
+        }
+
+        if (liveDocs instanceof FixedBitSet) {
+            return ((FixedBitSet) liveDocs).cardinality();
+        }
+
+        int count = 0;
+        for (int index = 0; index < liveDocs.length(); index++) {
+            count += liveDocs.get(index) ? 1 : 0;
+        }
+        return count;
+    }
+
+    static class KNNVectorValuesSub<T extends DocIdSetIterator> extends DocIDMerger.Sub {
+        final T values;
+        final int liveDocs;
+
+        KNNVectorValuesSub(MergeState.DocMap docMap, T values, int liveDocs) {
+            super(docMap);
+            this.values = values;
+            this.liveDocs = liveDocs;
+        }
+
+        @Override
+        public int nextDoc() throws IOException {
+            return values.nextDoc();
+        }
+    }
+}
--- a/src/main/java/org/opensearch/knn/index/vectorvalues/KNNVectorValues.java
+++ b/src/main/java/org/opensearch/knn/index/vectorvalues/KNNVectorValues.java
@@ -82,7 +82,6 @@ public abstract class KNNVectorValues<T> {
      * </pre>
      * @return long
      */
-    @Deprecated
     public long totalLiveDocs() {
         return vectorValuesIterator.liveDocs();
     }
--- a/src/main/java/org/opensearch/knn/index/vectorvalues/KNNVectorValuesFactory.java
+++ b/src/main/java/org/opensearch/knn/index/vectorvalues/KNNVectorValuesFactory.java
@@ -5,10 +5,13 @@
 
 package org.opensearch.knn.index.vectorvalues;
 
+import com.google.common.annotations.VisibleForTesting;
+import lombok.extern.log4j.Log4j2;
 import org.apache.lucene.index.DocValues;
 import org.apache.lucene.index.DocsWithFieldSet;
 import org.apache.lucene.index.FieldInfo;
 import org.apache.lucene.index.LeafReader;
+import org.apache.lucene.index.MergeState;
 import org.apache.lucene.index.VectorEncoding;
 import org.apache.lucene.search.DocIdSetIterator;
 import org.opensearch.knn.common.FieldInfoExtractor;
@@ -17,9 +20,13 @@ import org.opensearch.knn.index.VectorDataType;
 import java.io.IOException;
 import java.util.Map;
 
+import static org.opensearch.knn.index.vectorvalues.KNNMergeVectorValues.mergeByteVectorValues;
+import static org.opensearch.knn.index.vectorvalues.KNNMergeVectorValues.mergeFloatVectorValues;
+
 /**
  * A factory class that provides various methods to create the {@link KNNVectorValues}.
  */
+@Log4j2
 public final class KNNVectorValuesFactory {
 
     /**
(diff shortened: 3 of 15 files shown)


src/main/java/org/opensearch/knn/index/vectorvalues/KNNMergeVectorValues.java
```diff
@@ -0,0 +1,113 @@
...
+import org.apache.lucene.index.VectorEncoding;
+import org.apache.lucene.search.DocIdSetIterator;
+import org.apache.lucene.util.Bits;
+import org.apache.lucene.util.FixedBitSet;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+
+public final class KNNMergeVectorValues {
```
navneet1v, line 23: please add java docs on top of this class.

src/main/java/org/opensearch/knn/index/vectorvalues/KNNMergeVectorValues.java
```diff
@@ -0,0 +1,113 @@
...
+                FloatVectorValues values = knnVectorsReader.getFloatVectorValues(fieldInfo.name);
+                if (values != null) {
+                    final Bits liveDocs = mergeState.liveDocs[i];
+                    final int liveDocsInt;
+                    if (liveDocs != null) {
+                        if (liveDocs instanceof FixedBitSet) {
+                            liveDocsInt = ((FixedBitSet) liveDocs).cardinality();
+                        } else {
+                            liveDocsInt = computeLiveDocs(values, liveDocs);
+                            values = knnVectorsReader.getFloatVectorValues(fieldInfo.name);
```
navneet1v, line 44: let add comment around this code on why we are doing it.

src/main/java/org/opensearch/knn/index/vectorvalues/KNNMergeVectorValues.java
```diff
@@ -0,0 +1,113 @@
...
+        if (liveDocs != null) {
+            while (values.docID() != DocIdSetIterator.NO_MORE_DOCS) {
+                count++;
+                values.nextDoc();
+            }
+        }
+        return count;
+    }
+
+    static class KNNVectorValuesSub<T extends DocIdSetIterator> extends DocIDMerger.Sub {
```
navneet1v, line 98: why this is package private?

src/main/java/org/opensearch/knn/index/vectorvalues/KNNMergeVectorValues.java
```diff
@@ -0,0 +1,113 @@
...
+                        liveDocsInt = Math.toIntExact(values.cost());
+                    }
+                    subs.add(new KNNVectorValuesSub<>(mergeState.docMaps[i], values, liveDocsInt));
+                }
+            }
+        }
+        return subs;
+    }
+
+    public static List<KNNVectorValuesSub<ByteVectorValues>> mergeByteVectorValues(FieldInfo fieldInfo, MergeState mergeState)
```
navneet1v, line 56: add java doc.

src/main/java/org/opensearch/knn/index/vectorvalues/KNNVectorValuesFactory.java
```diff
@@ -45,7 +52,36 @@ public static <T> KNNVectorValues<T> getVectorValues(
         final DocsWithFieldSet docIdWithFieldSet,
         final Map<Integer, T> vectors
     ) {
-        return getVectorValues(vectorDataType, new KNNVectorValuesIterator.FieldWriterIteratorValues<T>(docIdWithFieldSet, vectors));
+        return getVectorValues(vectorDataType, new KNNVectorValuesIterator.FieldWriterIteratorValues<>(docIdWithFieldSet, vectors));
+    }
+
+    public static <T> KNNVectorValues<T> getVectorValues(
```
navneet1v, line 58: java doc.

src/main/java/org/opensearch/knn/index/vectorvalues/KNNVectorValuesFactory.java
```diff
@@ -72,8 +108,9 @@ public static <T> KNNVectorValues<T> getVectorValues(final FieldInfo fieldInfo,
         return getVectorValues(FieldInfoExtractor.extractVectorDataType(fieldInfo), vectorValuesIterator);
     }
 
+    @VisibleForTesting
     @SuppressWarnings("unchecked")
-    private static <T> KNNVectorValues<T> getVectorValues(
+    static <T> KNNVectorValues<T> getVectorValues(
```
navneet1v, line 113: why to make it package private?

navneet1v (commented):
Please add java doc on all the new added classes.

navneet1v:
Please also update the changelog.

src/main/java/org/opensearch/knn/index/vectorvalues/KNNMergeVectorValues.java
shatejas: Needed in [KNNVectorValuesIterator](https://github.com/opensearch-project/k-NN/pull/2135/files#diff-5b15a95bce01061c109452cff568a77fbef8d9a79f5efe7c3161b5d745c567d0R194)

src/main/java/org/opensearch/knn/index/vectorvalues/KNNVectorValuesFactory.java
shatejas: Needed for [unit test](https://github.com/opensearch-project/k-NN/pull/2135/files#diff-d70259d4ce2116e0b254c69590c9fd2bec228e43e4bd0eec20d7d38693250b0fR128)

navneet1v: lets try to avoid this and use one of the public function to test this function.

src/main/java/org/opensearch/knn/index/vectorvalues/KNNMergeVectorValues.java
```diff
@@ -0,0 +1,120 @@
...
+        }
+
+        if (liveDocs instanceof FixedBitSet) {
+            return ((FixedBitSet) liveDocs).cardinality();
+        }
+
+        int count = 0;
+        for (int index = 0; index < liveDocs.length(); index++) {
+            count += liveDocs.get(index) ? 1 : 0;
+        }
```
navneet1v, line 101: This will iterate over all the bits in the bitset, rather I would suggest VectorValues to iterate over the docIds and then see if the docIds is present in bitset or not. Because it can happen for a field only few docIds are present in liveDocs.

src/main/java/org/opensearch/knn/index/vectorvalues/KNNMergeVectorValues.java
```diff
@@ -0,0 +1,120 @@
...
+            throw new UnsupportedOperationException("Cannot merge vectors encoded as [" + fieldInfo.getVectorEncoding() + "] as BYTE");
+        }
+        final List<KNNVectorValuesSub<ByteVectorValues>> subs = new ArrayList<>();
+        for (int i = 0; i < mergeState.knnVectorsReaders.length; i++) {
+            KnnVectorsReader knnVectorsReader = mergeState.knnVectorsReaders[i];
+            if (knnVectorsReader != null) {
+                ByteVectorValues values = knnVectorsReader.getByteVectorValues(fieldInfo.getName());
+                if (values != null) {
+                    final Bits liveDocs = mergeState.liveDocs[i];
+                    final int liveDocsCardinality = cardinality(liveDocs, Math.toIntExact(values.cost()));
```
navneet1v, line 81: for understanding on how much time we are spending on this operation its better to add some debug logs.

src/main/java/org/opensearch/knn/index/vectorvalues/KNNMergeVectorValues.java
navneet1v: If this is a utility, lets rename the class to reflect that. Before you added the java doc I was under impression that its some kind of vectorValues.

src/main/java/org/opensearch/knn/index/vectorvalues/KNNMergeVectorValues.java
```diff
@@ -0,0 +1,120 @@
...
+    /**
+     * Gets list of {@link KNNVectorValuesSub} for {@link FloatVectorValues} from a merge state and returns the iterator which
+     * iterates over live docs from all segments while mapping docIds.
+     *
+     * @param fieldInfo
+     * @param mergeState
+     * @return List of KNNVectorSub
+     * @throws IOException
+     */
+    public static KNNVectorValuesIterator.MergeFloat32VectorValuesIterator mergeFloatVectorValues(
```
navneet1v, line 37:
```suggestion
    public static KNNVectorValuesIterator.MergeFloat32VectorValuesIterator getMergedFloatVectorValuesIterator(
```

src/main/java/org/opensearch/knn/index/vectorvalues/KNNMergeVectorValues.java
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    private static int cardinality(final Bits liveDocs, final int defaultCount) {
+        if (liveDocs == null) {
+            return defaultCount;
+        }
+
+        if (liveDocs instanceof FixedBitSet) {
+            return ((FixedBitSet) liveDocs).cardinality();
+        }
```
navneet1v, line 96: why this special case for fixedbitset and not for `SparseFixedBitSet` ?

src/main/java/org/opensearch/knn/index/vectorvalues/KNNVectorValues.java
```diff
@@ -82,7 +82,6 @@ public int bytesPerVector() {
      * </pre>
      * @return long
      */
-    @Deprecated
     public long totalLiveDocs() {
```
navneet1v, line 85: if we removing deprecated from here, we should also update the java docs.

src/main/java/org/opensearch/knn/index/vectorvalues/KNNVectorValuesIterator.java
```diff
@@ -185,4 +189,96 @@ public VectorValueExtractorStrategy getVectorExtractorStrategy() {
...
+        }
+
+        @Override
+        public int docId() {
+            return docId;
+        }
+
+        @Override
+        public int advance(int docId) throws IOException {
+            throw new UnsupportedOperationException();
```
navneet1v, line 221: add the exception message.

src/main/java/org/opensearch/knn/index/vectorvalues/KNNVectorValuesIterator.java
```diff
@@ -185,4 +189,96 @@ public VectorValueExtractorStrategy getVectorExtractorStrategy() {
...
+
+        @Override
+        public VectorValueExtractorStrategy getVectorExtractorStrategy() {
+            return VECTOR_VALUES_STRATEGY;
+        }
+
+        public abstract U vectorValue() throws IOException;
+    }
+
+    class MergeFloat32VectorValuesIterator extends MergeSegmentVectorValuesIterator<FloatVectorValues, float[]> {
```
navneet1v, line 255: Why we went with creating the iterator route here? rather we could have just extended the FloatVectorValues and ByteVectorValues. This would have avoided all these iterators and new strategy classes. 

Ref: https://github.com/apache/lucene/blob/branch_9x/lucene/core/src/java/org/apache/lucene/codecs/KnnVectorsWriter.java#L286

src/main/java/org/opensearch/knn/index/vectorvalues/KNNMergeVectorValues.java
shatejas: I don't think its being used. I was looking at the usages of [LiveDocsFormats write](https://github.com/apache/lucene/blob/53d1c2bd2fb3e6b9da590bee360996dbbdc8ea34/lucene/core/src/java/org/apache/lucene/codecs/LiveDocsFormat.java#L44) 
- [DocumentWritersPerThread](https://github.com/apache/lucene/blob/53d1c2bd2fb3e6b9da590bee360996dbbdc8ea34/lucene/core/src/java/org/apache/lucene/index/DocumentsWriterPerThread.java#L661)
- [Pending Deletes](https://github.com/apache/lucene/blob/53d1c2bd2fb3e6b9da590bee360996dbbdc8ea34/lucene/core/src/java/org/apache/lucene/index/PendingDeletes.java#L194)

read only returns FixedBits. Moreover while debugging I did not see SparseFixedBitSet usage. I can change it to `instanceof BitSet` to be safe

src/main/java/org/opensearch/knn/index/vectorvalues/KNNMergeVectorValues.java
shatejas: That requires another call to `knnVectorsReader.get[Float/Byte]VectorValues` for each segment, which is an anticipated latency driver. IO operation might be more expensive than looping through considering merges

src/main/java/org/opensearch/knn/index/vectorvalues/KNNVectorValues.java
shatejas: Missed as there was no @deprecated javadoc tag. The convention is to add @deprecated in javadocs tag  and give a reason next to it instead of combining it with existing

src/main/java/org/opensearch/knn/index/vectorvalues/KNNVectorValuesIterator.java
shatejas: Didn't realize its a DISI, changing it

We can take up the lucene contribution as a follow up

navneet1v: > We can take up the lucene contribution as a follow up

Please create a gh issue and ref it here.

src/main/java/org/opensearch/knn/index/vectorvalues/KNNMergeVectorValues.java
navneet1v: I am not super confident with this and also the next loop to find the live docs. I would check with lucene community first before making any assumptions here to avoid any possible regression, since this is on the write path.

shatejas: > community first before making any assumptions

Can you elaborate about the concerns with BitSet usage, as in performance or functionality? cardinality is supposed to return set bits and lucene seems to have optimized it for Fixed and Sparse with bit shifting

> the next loop to find the live docs

For this is the concern performance or functionality?

Asking because we should do our due diligence before reaching out to the lucene community

src/main/java/org/opensearch/knn/index/vectorvalues/KNNMergeVectorValues.java
shatejas: Changing this to return MergedFloatVectorValues which extends FloatVectorValues so iterator will not make sense

src/main/java/org/opensearch/knn/index/vectorvalues/KNNMergeVectorValues.java
navneet1v: @shatejas I think we don't need to reach out to Lucene, I thought of the case where this bitset count will fail and you can try this out with your cluster.

Live docs tells the total number of docs which are not deleted on the segment. Now a doc may or may not contain the vector field. So consider a case where there are 3 docs in the segment.
doc1 : has vector field : doc is deleted
doc2: has no vector field : doc is not deleted
doc3: has vector field : doc is not deleted.

With the above case the bitset will give live docs as 2 but for the vector field the live docs is only 1. 

Hence, using the bitset.cardinality() will never give you right number of live docs.

and this is the reason why lucene doesn't use this during segment merges during the training.

cc: @jmazanec15 and @heemin32

src/main/java/org/opensearch/knn/index/vectorvalues/KNNMergeVectorValues.java
navneet1v: Check my comment here: https://github.com/opensearch-project/k-NN/pull/2135#discussion_r1775784103 using live docs will not give you correct doc count.

Issue #1942, opened 2024-08-08
[FEATURE] Creating Vector data structures Greedily


navneet1v:
## Description
As of version 2.13 of Opensearch, whenever a segment is created we create the data structures which are required to do vector search(aka graphs for HNSW algorithm, buckets for IVF algorithm etc.). When the segments gets merged unlink inverted file index, BKDs these data structures are not merged, rather we create them from scratch(true for native engines, and Lucene(if deletes are there)). Example: if we are merging 2 segments with 1k documents each, the graphs which are created in both the segments are ignored and a new graph with 2K documents will newly be created. This leads to waste of compute(as build vector search data structures is very expensive) and slows down the build time for Vector indices.

Hence the idea is we should build these data structures greedily.

1. **Delaying Vector Data structures Creation:** 
    1. For bulk ingestion the proposal is we should completely stop creating vector data structures during segment creation ~and merges of segments~. We should only create vector data structures once the whole indexing ~and merges(including force merge if possible)~ is completed. This will ensure that we are creating vector data structures once. Refer Appendix A on how we can do this of GH issue: https://github.com/opensearch-project/k-NN/issues/1599.
    2. If a user is constantly indexing and searching on data then not building graphs will lead to high latency + user might not have time to do force merge to trigger the graph creation, hence, we should build the capability where based on a certain threshold we can Opensearch can decide for which segment graph needs to be created. This will improve the ingestion speed. Ref this community request: https://github.com/opensearch-project/k-NN/issues/1577

Having the capability to disable graph creation is extreme and will be used for cases where we need high speed indexing, index re-builds etc. On top of this feature next capability will be added is threshold based graph builds. This will ensure that this greed graph build based capability is used for more general use-cases with search also possible if graph not present.

2. **Incremental graph creation:** For native engines(Nmslib and Faiss) the vector data structures are build during segment creation(aka OS refresh), which leads to spikes in CPU and sometimes throttling due to high CPU(we saw this in various benchmarks). On the other time when segments are not getting created the CPU util stays very low. This leads uneven usage of CPU(basically a SAW tooth curve of CPU utilization is created) and users are not able to push more documents at Opensearch at steady state. Specifically for Streaming Ingestion use cases we can create graphs incrementally during ingestion(Lucene engine already implements this) this will ensure that we spread out the CPU utilization peak over a period of time and when OS refresh happens we already have graph created. Which will ensure that it is available for search for users hence reduce the Search after Indexing time.

References:
[...]

PR #2136, opened 2024-09-23
Add exact search if no native engine files are available

VijayanB:
### Description
Add exact search if no native engine files are available . This PR is part of issue where users will have option to decide when to build graph, which helps them to reduce build time. When graph is not available, plugin will return empty results. With this PR, exact search will be performed when only no engine file is available in segment.

We also don't need version check or feature flag because, option to not build vector data structure will only be available post 2.17. If an index is created using pre 2.17 version, segment will always have engine files and this feature will never be called during search.

### Related Issues
Part of #1942 

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/query/KNNWeight.java
+++ b/src/main/java/org/opensearch/knn/index/query/KNNWeight.java
@@ -22,6 +22,7 @@ import org.apache.lucene.util.Bits;
 import org.apache.lucene.util.FixedBitSet;
 import org.opensearch.common.io.PathUtils;
 import org.opensearch.common.lucene.Lucene;
+import org.opensearch.knn.common.FieldInfoExtractor;
 import org.opensearch.knn.common.KNNConstants;
 import org.opensearch.knn.index.KNNSettings;
 import org.opensearch.knn.index.SpaceType;
@@ -33,6 +34,7 @@ import org.opensearch.knn.index.memory.NativeMemoryEntryContext;
 import org.opensearch.knn.index.memory.NativeMemoryLoadStrategy;
 import org.opensearch.knn.index.engine.KNNEngine;
 import org.opensearch.knn.index.quantizationservice.QuantizationService;
+import org.opensearch.knn.index.query.ExactSearcher.ExactSearcherContext.ExactSearcherContextBuilder;
 import org.opensearch.knn.indices.ModelDao;
 import org.opensearch.knn.indices.ModelMetadata;
 import org.opensearch.knn.indices.ModelUtil;
@@ -129,42 +131,18 @@ public class KNNWeight extends Weight {
         if (filterWeight != null && cardinality == 0) {
             return Collections.emptyMap();
         }
-
         /*
-         * The idea for this optimization is to get K results, we need to atleast look at K vectors in the HNSW graph
+         * The idea for this optimization is to get K results, we need to at least look at K vectors in the HNSW graph
          * . Hence, if filtered results are less than K and filter query is present we should shift to exact search.
          * This improves the recall.
          */
-        Map<Integer, Float> docIdsToScoreMap;
-        final ExactSearcher.ExactSearcherContext exactSearcherContext = ExactSearcher.ExactSearcherContext.builder()
-            .k(k)
-            .isParentHits(true)
-            .matchedDocs(filterBitSet)
-            // setting to true, so that if quantization details are present we want to do search on the quantized
-            // vectors as this flow is used in first pass of search.
-            .useQuantizedVectorsForSearch(true)
-            .knnQuery(knnQuery)
-            .build();
-        if (filterWeight != null && canDoExactSearch(cardinality)) {
-            docIdsToScoreMap = exactSearch(context, exactSearcherContext);
-        } else {
-            docIdsToScoreMap = doANNSearch(context, filterBitSet, cardinality, k);
-            if (docIdsToScoreMap == null) {
-                return Collections.emptyMap();
-            }
-            if (canDoExactSearchAfterANNSearch(cardinality, docIdsToScoreMap.size())) {
-                log.debug(
-                    "Doing ExactSearch after doing ANNSearch as the number of documents returned are less than "
-                        + "K, even when we have more than K filtered Ids. K: {}, ANNResults: {}, filteredIdCount: {}",
-                    k,
-                    docIdsToScoreMap.size(),
-                    cardinality
-                );
-                docIdsToScoreMap = exactSearch(context, exactSearcherContext);
-            }
+        if (isFilteredExactSearchPreferred(cardinality)) {
+            return doExactSearch(context, filterBitSet, k);
         }
-        if (docIdsToScoreMap.isEmpty()) {
-            return Collections.emptyMap();
+        Map<Integer, Float> docIdsToScoreMap = doANNSearch(context, filterBitSet, cardinality, k);
+        if (isExactSearchRequire(context, cardinality, docIdsToScoreMap.size())) {
+            final BitSet docs = filterWeight != null ? filterBitSet : null;
+            return doExactSearch(context, docs, k);
         }
         return docIdsToScoreMap;
     }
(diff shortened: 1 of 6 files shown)


src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80DocValuesProducer.java
```diff
@@ -75,6 +75,9 @@ public KNN80DocValuesProducer(DocValuesProducer delegate, SegmentReadState state
                 continue;
             }
             List<String> engineFiles = KNNCodecUtil.getEngineFiles(knnEngine.getExtension(), field.name, state.segmentInfo);
+            if (engineFiles.isEmpty()) {
+                continue;
+            }
```
navneet1v, line 80: why we need this?

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -441,14 +445,72 @@ private boolean isExactSearchThresholdSettingSet(int filterThresholdValue) {
...
     }
 
+    /**
+     * This condition mainly checks whether exact search should be performed or not
+     * @param context LeafReaderContext
+     * @param filterIdsCount count of filtered Doc ids
+     * @param annResultCount Count of Nearest Neighbours we got after doing filtered ANN Search.
+     * @return boolean - true if exactSearch needs to be done after ANNSearch.
+     */
+    private boolean canDoExactSearch(final LeafReaderContext context, final int filterIdsCount, final int annResultCount) {
```
navneet1v, line 455: better to name the function isExactSearchRequired

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -441,14 +445,72 @@ private boolean isExactSearchThresholdSettingSet(int filterThresholdValue) {
...
+        if (annResultCount == 0 && isMissingNativeEngineFiles(context)) {
```
navneet1v, line 456: why we are looking for `annResultCount == 0` ?

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -441,14 +445,72 @@ private boolean isExactSearchThresholdSettingSet(int filterThresholdValue) {
...
+        final KNNEngine knnEngine = extractKNNEngine(fieldInfo);
+        final List<String> engineFiles = KNNCodecUtil.getEngineFiles(
+            knnEngine.getExtension(),
+            knnQuery.getField(),
+            reader.getSegmentInfo().info
+        );
+        return engineFiles.isEmpty();
+    }
+
+    private KNNEngine extractKNNEngine(FieldInfo fieldInfo) {
```
navneet1v, line 504: there is already a class called as FieldInfoExtracter and this extractKNNEngine function is present there, use that funciton.

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -121,54 +124,35 @@ public Scorer scorer(LeafReaderContext context) throws IOException {
...
+            // if filter is provided uses filtered docs else use get all docs
+            final BitSet docs = filterWeight != null ? acceptedDocs : getAllDocsBitSet(context);
+            return doExactSearch(context, docs, k);
         }
         return docIdsToScoreMap;
     }
 
+    static {
+        EMPTY_DOC_IDS_TO_SCORE_MAP = Collections.emptyMap();
+    }
```
navneet1v, line 154: may be move this on top of the class. and why do we even need this? what was the problem with Collections.emptyMap()?

src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80DocValuesProducer.java
VijayanB: Line 81 assumes that engine files will always be available, that is not the case anymore.

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -185,13 +169,21 @@ private BitSet getFilteredDocsBitSet(final LeafReaderContext ctx) throws IOExcep
         return createBitSet(scorer.iterator(), liveDocs, maxDoc);
     }
 
-    private BitSet createBitSet(final DocIdSetIterator filteredDocIdsIterator, final Bits liveDocs, int maxDoc) throws IOException {
-        if (liveDocs == null && filteredDocIdsIterator instanceof BitSetIterator) {
+    private BitSet getAllDocsBitSet(final LeafReaderContext ctx) throws IOException {
+        final FloatVectorValues floatVectorValues = ctx.reader().getFloatVectorValues(this.knnQuery.getField());
```
navneet1v, line 173: There can be ByteVectorValues too. You should actually create KNNVectorValues here using KNNVectorValuesFactory class.

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -185,13 +169,21 @@ private BitSet getFilteredDocsBitSet(final LeafReaderContext ctx) throws IOExcep
...
+        final Bits liveDocs = ctx.reader().getLiveDocs();
+        final int maxDoc = ctx.reader().maxDoc();
+        final VectorScorer scorer = floatVectorValues.scorer(this.knnQuery.getQueryVector());
```
navneet1v, line 176: this is pretty weird, since all we need is the iterator, you can get the iterator from KNNVectorValues too.

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
VijayanB: Moved to the top. To avoid repeating Collections.emptyMap() in every return statement

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
VijayanB: To narrow the calls to checkNativeEngine files are missing or not? For ex: if result != 0, we know that it is not missing.

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
VijayanB: Ack

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
VijayanB: Ack

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
jmazanec15: why have the block and not initialize in declaration?

jmazanec15: also, are we using this at all?

src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80DocValuesProducer.java
VijayanB: Created new PR #2147 to address this change

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
VijayanB: Removed

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
VijayanB: Updated iterator to accept non filter use cases. We don't need this anymore

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -129,42 +131,18 @@ public Map<Integer, Float> searchLeaf(LeafReaderContext context, int k) throws I
...
-                );
-                docIdsToScoreMap = exactSearch(context, exactSearcherContext);
-            }
+        if (isFilteredExactSearchPreferred(cardinality)) {
+            return doExactSearch(context, filterBitSet, k);
         }
-        if (docIdsToScoreMap.isEmpty()) {
-            return Collections.emptyMap();
+        Map<Integer, Float> docIdsToScoreMap = doANNSearch(context, filterBitSet, cardinality, k);
+        if (isExactSearchRequire(context, cardinality, docIdsToScoreMap.size())) {
```
jmazanec15, line 143: add comment on this?

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -441,14 +428,59 @@ private boolean isExactSearchThresholdSettingSet(int filterThresholdValue) {
...
+    /**
+     * This condition mainly checks whether exact search should be performed or not
+     * @param context LeafReaderContext
+     * @param filterIdsCount count of filtered Doc ids
+     * @param annResultCount Count of Nearest Neighbours we got after doing filtered ANN Search.
+     * @return boolean - true if exactSearch needs to be done after ANNSearch.
+     */
+    private boolean isExactSearchRequire(final LeafReaderContext context, final int filterIdsCount, final int annResultCount) {
+        if (annResultCount == 0 && isMissingNativeEngineFiles(context)) {
+            log.info("Perform exact search after approximate search since no native engine files are available");
```
jmazanec15, line 440: I would say dont info log per query

VijayanB: Ack

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
VijayanB: Ack

navneet1v (approved):

Issue #2145, opened 2024-09-24
[Enhancement] Optimize searchTopK Method in ExactSearcher for Improved Performance


junqiu-lei:
**What is the enhancement?**
We can optimize the reduceToTopK method by removing unnecessary pre-filling, reducing redundant peek() calls for better performance.

Code ref: https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/query/ResultUtil.java#L33-L59

PR #2146, opened 2024-09-24
Optimize reduceToTopK in ResultUtil by removing pre-filling and reducing peek calls

junqiu-lei:
### Description

This PR optimizes the reduceToTopK method by eliminating unnecessary pre-filling of the priority queue, reducing redundant peek() calls, and adding null safety checks for better performance.

### Related Issues
Closes https://github.com/opensearch-project/k-NN/issues/2145

### Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 * Add AVX512 support to k-NN for FAISS library [#2069](https://github.com/opensearch-project/k-NN/pull/2069)
 ### Enhancements
 * Add short circuit if no live docs are in segments [#2059](https://github.com/opensearch-project/k-NN/pull/2059)
+* Optimize reduceToTopK in ResultUtil by removing pre-filling and reducing peek calls [#2146](https://github.com/opensearch-project/k-NN/pull/2146)
 ### Bug Fixes
 ### Infrastructure
 ### Documentation
--- a/src/main/java/org/opensearch/knn/index/query/ResultUtil.java
+++ b/src/main/java/org/opensearch/knn/index/query/ResultUtil.java
@@ -33,15 +33,14 @@ public final class ResultUtil {
     public static void reduceToTopK(List<Map<Integer, Float>> perLeafResults, int k) {
         // Iterate over all scores to get min competitive score
         PriorityQueue<Float> topKMinQueue = new PriorityQueue<>(k);
-        for (int i = 0; i < k; i++) {
-            topKMinQueue.add(-Float.MAX_VALUE);
-        }
 
         int count = 0;
         for (Map<Integer, Float> perLeafResult : perLeafResults) {
             count += perLeafResult.size();
             for (Float score : perLeafResult.values()) {
-                if (topKMinQueue.peek() != null && score > topKMinQueue.peek()) {
+                if (topKMinQueue.size() < k) {
+                    topKMinQueue.add(score);
+                } else if (topKMinQueue.peek() != null && score > topKMinQueue.peek()) {
                     topKMinQueue.poll();
                     topKMinQueue.add(score);
                 }


src/main/java/org/opensearch/knn/index/query/ExactSearcher.java
```diff
@@ -68,26 +68,25 @@ private Map<Integer, Float> scoreAllDocs(KNNIterator iterator) throws IOExceptio
...
         while ((docId = iterator.nextDoc()) != DocIdSetIterator.NO_MORE_DOCS) {
-            if (iterator.score() > topDoc.score) {
-                topDoc.score = iterator.score();
-                topDoc.doc = docId;
-                // As the HitQueue is min heap, updating top will bring the doc with -INF score or worst score we
-                // have seen till now on top.
-                topDoc = queue.updateTop();
+            float score = iterator.score();
+
+            if (score < 0) {
```
navneet1v, line 78: is the score conversion happening? because inner product can give negative scores

navneet1v: Also, I didn't understand the logic on why score < 0, we should continue.

src/main/java/org/opensearch/knn/index/query/ExactSearcher.java
```diff
@@ -68,26 +68,25 @@ private Map<Integer, Float> scoreAllDocs(KNNIterator iterator) throws IOExceptio
...
             }
-        }
 
-        // If scores are negative we will remove them.
-        // This is done, because there can be negative values in the Heap as we init the heap with Score as -INF.
-        // If filterIds < k, the some values in heap can have a negative score.
-        while (queue.size() > 0 && queue.top().score < 0) {
-            queue.pop();
+            if (queue.size() < k) {
+                queue.add(new ScoreDoc(docId, score));
```
navneet1v, line 83: this condition will never be true because of this.
```
final HitQueue queue = new HitQueue(k, true);
```

src/main/java/org/opensearch/knn/index/query/ExactSearcher.java
```diff
@@ -68,26 +68,25 @@ private Map<Integer, Float> scoreAllDocs(KNNIterator iterator) throws IOExceptio
...
+            } else if (score > queue.top().score) {
```
navneet1v, line 84: what will happen in the case when score == queue.top().score ?

src/main/java/org/opensearch/knn/index/query/ExactSearcher.java
```diff
@@ -68,26 +68,25 @@ private Map<Integer, Float> scoreAllDocs(KNNIterator iterator) throws IOExceptio
     }
 
     private Map<Integer, Float> searchTopK(KNNIterator iterator, int k) throws IOException {
-        // Creating min heap and init with MAX DocID and Score as -INF.
```
navneet1v, line 71: why we are removing this comment? this comment is still true.

junqiu-lei:
Offline synced with @navneet1v and @jmazanec15, we now focused on optimizing reduceToTopK function in ResultUtil, updated the PR.

shatejas (commented):
@junqiu-lei Did we compare removing the map with this approach to see which is faster?  or is it not possible? I feel like its a simpler code that way if its yielding better results

- removing the map approach: See if we can preserve the order of the results for each leaf [instead of converting it to map](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/query/KNNWeight.java#L378). 

We might have to first check if KNNResults array is in order. but that way we can do something similar to what lucene does and just pick the top elements and stop at k in reduceToTopK

junqiu-lei:
> @junqiu-lei Did we compare removing the map with this approach to see which is faster? or is it not possible? I feel like its a simpler code that way if its yielding better results
> 
> * removing the map approach: See if we can preserve the order of the results for each leaf [instead of converting it to map](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/query/KNNWeight.java#L378).
> 
> We might have to first check if KNNResults array is in order. but that way we can do something similar to what lucene does and just pick the top elements and stop at k in reduceToTopK

I think this is another place we can optimize, no we don't have comparing the difference so far. For this PR, we might can use micro benchmark for testing the improvement if possible.

jmazanec15 (commented):
So, I think the most optimal way to do this would be to require that each perLeafResults comes in order from top scores to worst scores. Then, a simple algorithm that basically just scans the top of each set of results and then only takes the top k. That being, that would require how results are returned in the different methods. Do you think we should do this @navneet1v ?

jmazanec15 (dismissed):

jmazanec15 (approved):

junqiu-lei:
> So, I think the most optimal way to do this would be to require that each perLeafResults comes in order from top scores to worst scores. Then, a simple algorithm that basically just scans the top of each set of results and then only takes the top k. That being, that would require how results are returned in the different methods. Do you think we should do this @navneet1v ?

I can raise other PR for this part optimization.

navneet1v:
> So, I think the most optimal way to do this would be to require that each perLeafResults comes in order from top scores to worst scores. Then, a simple algorithm that basically just scans the top of each set of results and then only takes the top k. That being, that would require how results are returned in the different methods. Do you think we should do this @navneet1v ?

Yes I think we should change the return types

navneet1v (approved):

PR #2147, opened 2024-09-25
KNN80DocValues should only be considered for BinaryDocValues fields

VijayanB:
### Description
KNN80DocValuesProducer should consider adding files from fields that has BinaryDocValues and doesn't have vector values in producer. By default, k-NN field types post 2.17 don't create BinaryDocValues for knn field anymore. However, users can set doc_values = true, to create BinaryDocValues explicitly like any other field, though it is not used for approximate search. Hence, we want to avoid considering fields that is created using NativeEngines990KnnVectorsFormat. Checking doc Type not equal to BinayDocValues and field doesn't have vectors will make sure we don't accidentally add native engine files that are created from NativeEngines990KnnVectorsFormat.

### Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Enhancements
 * Add short circuit if no live docs are in segments [#2059](https://github.com/opensearch-project/k-NN/pull/2059)
 ### Bug Fixes
+* KNN80DocValues should only be considered for BinaryDocValues fields [#2147](https://github.com/opensearch-project/k-NN/pull/2147)
 ### Infrastructure
 ### Documentation
 ### Maintenance
--- a/src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80DocValuesProducer.java
+++ b/src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80DocValuesProducer.java
@@ -15,6 +15,7 @@ import lombok.NonNull;
 import lombok.extern.log4j.Log4j2;
 import org.apache.lucene.codecs.DocValuesProducer;
 import org.apache.lucene.index.BinaryDocValues;
+import org.apache.lucene.index.DocValuesType;
 import org.apache.lucene.index.FieldInfo;
 import org.apache.lucene.index.NumericDocValues;
 import org.apache.lucene.index.SegmentReadState;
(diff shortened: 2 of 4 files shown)


navneet1v:
@VijayanB thanks for catching and fixing this bug.

navneet1v:
@VijayanB I won't recommend doing skip-change log for this and add the entry in the change log.

src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80DocValuesProducer.java
```diff
@@ -69,6 +70,10 @@ public KNN80DocValuesProducer(DocValuesProducer delegate, SegmentReadState state
             if (!field.attributes().containsKey(KNN_FIELD)) {
                 continue;
             }
+            // Only segments that contains BinaryDocValues and doesn't have vector values
```
navneet1v, line 73: elaborate the comment little bit more for our futurseleves.

navneet1v: also lets add a UT for this

src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80DocValuesProducer.java
```diff
@@ -69,6 +70,10 @@ public KNN80DocValuesProducer(DocValuesProducer delegate, SegmentReadState state
...
+            if( field.getDocValuesType() != DocValuesType.BINARY || field.hasVectorValues()){
```
jmazanec15, line 74: why would this happen both times?

VijayanB:
> @VijayanB I won't recommend doing skip-change log for this and add the entry in the change log.

Ack

src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80DocValuesProducer.java
VijayanB: @jmazanec15 I updated comment. Let me know if that is not clear or missing some more context

src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80DocValuesProducer.java
VijayanB: Ack

src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80DocValuesProducer.java
jmazanec15: I guess I was thinking its redundant if we already check the KNN_FIELD attribute

jmazanec15 (approved):

CHANGELOG.md
```diff
@@ -23,6 +23,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Enhancements
 * Add short circuit if no live docs are in segments [#2059](https://github.com/opensearch-project/k-NN/pull/2059)
 ### Bug Fixes
+* KNN80DocValues should only be considered for BinaryDocValues fields [#2147](https://github.com/opensearch-project/k-NN/pull/2147)
```
navneet1v, line 26: this is making no sense.

VijayanB: i updated PR description.

navneet1v: PR description is fine. But change log should also reflect what you are doing.

navneet1v (approved):

PR #2155, opened 2024-09-27
KNNIterators should support with and without filters

VijayanB:
### Description
Update VectorIterator, ByteIterator, NestedVectorIterator, NestedByteIterator to iterate even if there are no filters provided.
Currently this is used by exact search to score either topK docs or all docs when filter is provided by users.
However, in future we will be allowing exact search even if there are no filters. Hence, decouple filter
and make it option to support both cases.

### Related Issues
Pre-requisite for  #1942 

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/query/iterators/ByteVectorIdsKNNIterator.java
+++ b/src/main/java/org/opensearch/knn/index/query/iterators/ByteVectorIdsKNNIterator.java
@@ -38,6 +38,8 @@ public class ByteVectorIdsKNNIterator implements KNNIterator {
         this.queryVector = queryVector;
         this.binaryVectorValues = binaryVectorValues;
         this.spaceType = spaceType;
+        // This cannot be moved inside nextDoc() method since it will break when we have nested field, where
+        // nextDoc should already be referring to next knnVectorValues
         this.docId = getNextDocId();
     }
 
--- a/src/main/java/org/opensearch/knn/index/query/iterators/ByteVectorIdsKNNIterator.java
+++ b/src/main/java/org/opensearch/knn/index/query/iterators/ByteVectorIdsKNNIterator.java
@@ -20,8 +20,6 @@ import java.io.IOException;
  * The class is used in KNNWeight to score all docs, but, it iterates over filterIdsArray if filter is provided
  */
 public class ByteVectorIdsKNNIterator implements KNNIterator {
-    // Array of doc ids to iterate
-    protected final BitSet filterIdsBitSet;
     protected final BitSetIterator bitSetIterator;
     protected final byte[] queryVector;
     protected final KNNBinaryVectorValues binaryVectorValues;
@@ -34,13 +32,17 @@ public class ByteVectorIdsKNNIterator implements KNNIterator {
         final byte[] queryVector,
         final KNNBinaryVectorValues binaryVectorValues,
         final SpaceType spaceType
-    ) {
-        this.filterIdsBitSet = filterIdsBitSet;
-        this.bitSetIterator = new BitSetIterator(filterIdsBitSet, filterIdsBitSet.length());
+    ) throws IOException {
+        this.bitSetIterator = filterIdsBitSet == null ? null : new BitSetIterator(filterIdsBitSet, filterIdsBitSet.length());
         this.queryVector = queryVector;
         this.binaryVectorValues = binaryVectorValues;
         this.spaceType = spaceType;
-        this.docId = bitSetIterator.nextDoc();
+        this.docId = getNextDocId();
+    }
+
+    ByteVectorIdsKNNIterator(final byte[] queryVector, final KNNBinaryVectorValues binaryVectorValues, final SpaceType spaceType)
+        throws IOException {
+        this(null, queryVector, binaryVectorValues, spaceType);
     }
 
     /**
@@ -55,10 +57,13 @@ public class ByteVectorIdsKNNIterator implements KNNIterator {
         if (docId == DocIdSetIterator.NO_MORE_DOCS) {
             return DocIdSetIterator.NO_MORE_DOCS;
         }
-        int doc = binaryVectorValues.advance(docId);
+        if (bitSetIterator != null) {
+            binaryVectorValues.advance(docId);
+        }
         currentScore = computeScore();
-        docId = bitSetIterator.nextDoc();
-        return doc;
+        int currentDocId = docId;
+        docId = getNextDocId();
+        return currentDocId;
     }
 
     @Override
@@ -71,5 +76,12 @@ public class ByteVectorIdsKNNIterator implements KNNIterator {
         // Calculates a similarity score between the two vectors with a specified function. Higher similarity
         // scores correspond to closer vectors.
         return spaceType.getKnnVectorSimilarityFunction().compare(queryVector, vector);
+    }
+
+    protected int getNextDocId() throws IOException {
+        if (bitSetIterator != null) {
+            return this.bitSetIterator.nextDoc();
+        }
+        return this.binaryVectorValues.nextDoc();
     }
 }
--- a/src/main/java/org/opensearch/knn/index/query/iterators/ByteVectorIdsKNNIterator.java
+++ b/src/main/java/org/opensearch/knn/index/query/iterators/ByteVectorIdsKNNIterator.java
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-package org.opensearch.knn.index.query.filtered;
+package org.opensearch.knn.index.query.iterators;
 
 import org.apache.lucene.search.DocIdSetIterator;
 import org.apache.lucene.util.BitSet;
@@ -17,9 +17,9 @@ import java.io.IOException;
  * Inspired by DiversifyingChildrenFloatKnnVectorQuery in lucene
  * https://github.com/apache/lucene/blob/7b8aece125aabff2823626d5b939abf4747f63a7/lucene/join/src/java/org/apache/lucene/search/join/DiversifyingChildrenFloatKnnVectorQuery.java#L162
  *
- * The class is used in KNNWeight to score filtered KNN field by iterating filterIdsArray.
+ * The class is used in KNNWeight to score all docs, but, it iterates over filterIdsArray if filter is provided
  */
-public class FilteredIdsKNNByteIterator implements KNNIterator {
+public class ByteVectorIdsKNNIterator implements KNNIterator {
     // Array of doc ids to iterate
     protected final BitSet filterIdsBitSet;
     protected final BitSetIterator bitSetIterator;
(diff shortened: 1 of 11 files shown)


src/main/java/org/opensearch/knn/index/query/iterators/ByteVectorIdsKNNIterator.java
```diff
@@ -17,30 +17,32 @@
...
     protected final BitSetIterator bitSetIterator;
     protected final byte[] queryVector;
     protected final KNNBinaryVectorValues binaryVectorValues;
     protected final SpaceType spaceType;
     protected float currentScore = Float.NEGATIVE_INFINITY;
     protected int docId;
 
-    public FilteredIdsKNNByteIterator(
+    public ByteVectorIdsKNNIterator(
         final BitSet filterIdsBitSet,
```
navneet1v, line 31: should we make filterIdsBitSet as Optional? or atleast annotate as @Nullable to provide a signal that this parameter can be null.

VijayanB: Ack

src/main/java/org/opensearch/knn/index/query/iterators/VectorIdsKNNIterator.java
```diff
@@ -33,33 +31,44 @@ public class FilteredIdsKNNIterator implements KNNIterator {
...
+        this.docId = getNextDocId();
         this.quantizedQueryVector = quantizedQueryVector;
         this.segmentLevelQuantizationInfo = segmentLevelQuantizationInfo;
     }
 
+    protected int getNextDocId() throws IOException {
+        if (bitSetIterator != null) {
+            return this.bitSetIterator.nextDoc();
+        }
+        return this.knnFloatVectorValues.nextDoc();
```
navneet1v, line 70: should we advance the vector values also in this function which will ensure that we don't need this code in nextDoc() function.

```
if (bitSetIterator != null) {
    knnFloatVectorValues.advance(docId);
}
```

navneet1v (commented):
Thanks for the refactoring.

src/main/java/org/opensearch/knn/index/query/iterators/VectorIdsKNNIterator.java
VijayanB: i thought about it, but I decided against that since advance will do nothing but pointing to itself in non filter cases. With this way i feel it is clear and conveys exactly why we are advancing during filter. However, do you see potential issue with impact in performance ?

navneet1v: its not about impact in performance. if we want t advance to advance to next doc then we should actually do that for both bitsetIterator and also knnFloatVectorValues. Hence I believe either we should remove the call of nextDoc from constructors and make things clean or move this logic to getNextDocId(). I would prefer doing first one where we remove the call of nextDoc from constructor.

VijayanB: I was planning towards remove that call from constructor but NestedIterator will break  if i move docId inside nextDoc. With regards to your suggestion, i believe it will look like below
```
    if (bitSetIterator == null) {
          return this.knnFloatVectorValues.nextDoc();
    }
    int docId = this.bitSetIterator.nextDoc();
   if(docId != DocIdSetIterator.NO_MORE_DOCS) {
          knnFloatVectorValues.advance(docId);
   }
   return docId
```
i was hesitant since i had to to add NO_MORE_DOCS check again.

src/main/java/org/opensearch/knn/index/query/iterators/ByteVectorIdsKNNIterator.java
```diff
@@ -17,30 +18,32 @@
...
+    ) throws IOException {
+        this.bitSetIterator = filterIdsBitSet == null ? null : new BitSetIterator(filterIdsBitSet, filterIdsBitSet.length());
         this.queryVector = queryVector;
         this.binaryVectorValues = binaryVectorValues;
         this.spaceType = spaceType;
-        this.docId = bitSetIterator.nextDoc();
+        this.docId = getNextDocId();
+    }
+
+    ByteVectorIdsKNNIterator(final byte[] queryVector, final KNNBinaryVectorValues binaryVectorValues, final SpaceType spaceType)
```
heemin32, line 44:
```suggestion
    public ByteVectorIdsKNNIterator(final byte[] queryVector, final KNNBinaryVectorValues binaryVectorValues, final SpaceType spaceType)
```

src/main/java/org/opensearch/knn/index/query/iterators/NestedByteVectorIdsKNNIterator.java
```diff
@@ -3,33 +3,45 @@
...
         final KNNBinaryVectorValues binaryVectorValues,
         final SpaceType spaceType,
         final BitSet parentBitSet
-    ) {
+    ) throws IOException {
         super(filterIdsArray, queryVector, binaryVectorValues, spaceType);
         this.parentBitSet = parentBitSet;
     }
 
+    NestedByteVectorIdsKNNIterator(
```
heemin32, line 35:
```suggestion
    public NestedByteVectorIdsKNNIterator(
```

src/main/java/org/opensearch/knn/index/query/iterators/NestedVectorIdsKNNIterator.java
```diff
@@ -3,42 +3,53 @@
...
+ * However, it dedupe docs per each parent doc
  * of which ID is set in parentBitSet and only return best child doc with the highest score.
  */
-public class NestedFilteredIdsKNNIterator extends FilteredIdsKNNIterator {
+public class NestedVectorIdsKNNIterator extends VectorIdsKNNIterator {
     private final BitSet parentBitSet;
 
-    NestedFilteredIdsKNNIterator(
-        final BitSet filterIdsArray,
+    NestedVectorIdsKNNIterator(
```
heemin32, line 25:
```suggestion
    public NestedVectorIdsKNNIterator(
```

src/main/java/org/opensearch/knn/index/query/iterators/NestedVectorIdsKNNIterator.java
```diff
@@ -3,42 +3,53 @@
...
         final KNNFloatVectorValues knnFloatVectorValues,
         final SpaceType spaceType,
         final BitSet parentBitSet
-    ) {
+    ) throws IOException {
         this(filterIdsArray, queryVector, knnFloatVectorValues, spaceType, parentBitSet, null, null);
     }
 
-    public NestedFilteredIdsKNNIterator(
+    NestedVectorIdsKNNIterator(
```
heemin32, line 35:
```suggestion
    public NestedVectorIdsKNNIterator(
```

src/main/java/org/opensearch/knn/index/query/iterators/VectorIdsKNNIterator.java
```diff
@@ -33,33 +32,44 @@ public class FilteredIdsKNNIterator implements KNNIterator {
     protected int docId;
     private final SegmentLevelQuantizationInfo segmentLevelQuantizationInfo;
 
-    FilteredIdsKNNIterator(
-        final BitSet filterIdsBitSet,
+    VectorIdsKNNIterator(
```
heemin32, line 35:
```suggestion
    public VectorIdsKNNIterator(
```

src/main/java/org/opensearch/knn/index/query/iterators/VectorIdsKNNIterator.java
```diff
@@ -33,33 +32,44 @@ public class FilteredIdsKNNIterator implements KNNIterator {
...
         final float[] queryVector,
         final KNNFloatVectorValues knnFloatVectorValues,
         final SpaceType spaceType
-    ) {
+    ) throws IOException {
         this(filterIdsBitSet, queryVector, knnFloatVectorValues, spaceType, null, null);
     }
 
-    public FilteredIdsKNNIterator(
+    VectorIdsKNNIterator(final float[] queryVector, final KNNFloatVectorValues knnFloatVectorValues, final SpaceType spaceType)
```
heemin32, line 44:
```suggestion
    public VectorIdsKNNIterator(final float[] queryVector, final KNNFloatVectorValues knnFloatVectorValues, final SpaceType spaceType)
```

src/main/java/org/opensearch/knn/index/query/iterators/ByteVectorIdsKNNIterator.java
```diff
@@ -55,10 +58,13 @@ public int nextDoc() throws IOException {
         if (docId == DocIdSetIterator.NO_MORE_DOCS) {
             return DocIdSetIterator.NO_MORE_DOCS;
         }
-        int doc = binaryVectorValues.advance(docId);
+        if (bitSetIterator != null) {
+            binaryVectorValues.advance(docId);
```
heemin32, line 62: Can we move this line inside `getNextDocId()`? See line#84 for example.

src/main/java/org/opensearch/knn/index/query/iterators/ByteVectorIdsKNNIterator.java
```diff
@@ -72,4 +78,11 @@ protected float computeScore() throws IOException {
         // scores correspond to closer vectors.
         return spaceType.getKnnVectorSimilarityFunction().compare(queryVector, vector);
     }
+
+    protected int getNextDocId() throws IOException {
+        if (bitSetIterator != null) {
+            return this.bitSetIterator.nextDoc();
```
heemin32, line 84: Let's remove `this` for consistency. `this` is mostly needed only inside constructor to distinguish the variable from parameter.

```suggestion
            int docId = bitSetIterator.nextDoc();
            binaryVectorValues.advance(docId);
            return docId;
```

VijayanB: @heemin32 It is missing NO_DOCUMENTS check. Please check my response here https://github.com/opensearch-project/k-NN/pull/2155#discussion_r1778178001

heemin32: I am good with having NO_DOCUMENTS check here if needed.

src/main/java/org/opensearch/knn/index/query/iterators/VectorIdsKNNIterator.java
VijayanB: As discussed offline, adding comment to nextDoc method

src/main/java/org/opensearch/knn/index/query/iterators/VectorIdsKNNIterator.java
VijayanB: Ack

src/main/java/org/opensearch/knn/index/query/iterators/VectorIdsKNNIterator.java
VijayanB: Ack

heemin32 (approved):

navneet1v (approved):

PR #2172, opened 2024-09-30
 Adding Support to Enable/Disble Share level Rescoring and Update Oversampling Factor

Vikasht34:
### Description
Issue:- During rescoring at the shard level in a multi-segment case, the recall of the search results drops to zero.
Fix :- To address the issue of reduced recall during shard-level rescoring in multi-segment cases, we are introducing an index-level setting that allows users to enable or disable shard-level rescoring. This setting gives finer control over whether rescoring should be performed at the shard level or if the system should rely on a more global rescoring process across all segments. By disabling shard-level rescoring, the recall issue is mitigated, as the rescoring process will take into account the complete set of results from all segments, rather than focusing on top k results. 

Oversampling Update :- With this change we are updating default oversampling factor as well. For dimensions less than 768, an oversample factor of 3.0x is applied to ensure higher recall for smaller vector spaces. For dimensions between 768 and 1000, the oversampling factor is set to 2.0x to balance recall and performance. For dimensions above 1000, no oversampling (1.0x) is applied.

### Related Issues

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/KNNSettings.java
+++ b/src/main/java/org/opensearch/knn/index/KNNSettings.java
@@ -88,6 +88,7 @@ public class KNNSettings {
     public static final String QUANTIZATION_STATE_CACHE_SIZE_LIMIT = "knn.quantization.cache.size.limit";
     public static final String QUANTIZATION_STATE_CACHE_EXPIRY_TIME_MINUTES = "knn.quantization.cache.expiry.minutes";
     public static final String KNN_FAISS_AVX512_DISABLED = "knn.faiss.avx512.disabled";
+    public static final String KNN_DISK_VECTOR_SHARD_LEVEL_RESCORING_DISABLED = "index.knn.disk.vector.shard_level_rescoring_disabled";
 
     /**
      * Default setting values
@@ -112,10 +113,30 @@ public class KNNSettings {
     public static final Integer KNN_MAX_QUANTIZATION_STATE_CACHE_SIZE_LIMIT_PERCENTAGE = 10; // Quantization state cache limit cannot exceed
                                                                                              // 10% of the JVM heap
     public static final Integer KNN_DEFAULT_QUANTIZATION_STATE_CACHE_EXPIRY_TIME_MINUTES = 60;
+    public static final boolean KNN_DISK_VECTOR_SHARD_LEVEL_RESCORING_DISABLED_VALUE = true;
 
     /**
      * Settings Definition
      */
+
+    /**
+     * This setting controls whether shard-level re-scoring for KNN disk-based vectors is turned off.
+     * The setting uses:
+     * <ul>
+     *     <li><b>KNN_DISK_VECTOR_SHARD_LEVEL_RESCORING_DISABLED:</b> The name of the setting.</li>
+     *     <li><b>KNN_DISK_VECTOR_SHARD_LEVEL_RESCORING_DISABLED_VALUE:</b> The default value (true or false).</li>
+     *     <li><b>IndexScope:</b> The setting works at the index level.</li>
+     *     <li><b>Dynamic:</b> This setting can be changed without restarting the cluster.</li>
+     * </ul>
+     *
+     * @see Setting#boolSetting(String, boolean, Setting.Property...)
+     */
+    public static final Setting<Boolean> KNN_DISK_VECTOR_SHARD_LEVEL_RESCORING_DISABLED_SETTING = Setting.boolSetting(
+        KNN_DISK_VECTOR_SHARD_LEVEL_RESCORING_DISABLED,
+        KNN_DISK_VECTOR_SHARD_LEVEL_RESCORING_DISABLED_VALUE,
+        IndexScope,
+        Dynamic
+    );
 
     // This setting controls how much memory should be used to transfer vectors from Java to JNI Layer. The default
     // 1% of the JVM heap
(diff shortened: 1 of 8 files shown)


src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java
```diff
@@ -63,7 +63,12 @@ public Weight createWeight(IndexSearcher indexSearcher, ScoreMode scoreMode, flo
         } else {
             int firstPassK = rescoreContext.getFirstPassK(finalK);
             perLeafResults = doSearch(indexSearcher, leafReaderContexts, knnWeight, firstPassK);
-            ResultUtil.reduceToTopK(perLeafResults, firstPassK);
+            if (KNNSettings.isShardLevelRescoringDisabledForDiskBasedVector(knnQuery.getIndexName()) == false) {
+                log.info("Shard Level Rescoring is enabled. Reducing docs to {}", firstPassK);
```
jmazanec15, line 67: Id say no info log here

src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java
```diff
@@ -63,7 +63,12 @@ public Weight createWeight(IndexSearcher indexSearcher, ScoreMode scoreMode, flo
...
+                ResultUtil.reduceToTopK(perLeafResults, firstPassK);
+            } else {
+                log.info("Shard Level Rescoring is not enabled.");
```
jmazanec15, line 70: or here

src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java
```diff
@@ -63,7 +63,9 @@ public Weight createWeight(IndexSearcher indexSearcher, ScoreMode scoreMode, flo
         } else {
             int firstPassK = rescoreContext.getFirstPassK(finalK);
             perLeafResults = doSearch(indexSearcher, leafReaderContexts, knnWeight, firstPassK);
-            ResultUtil.reduceToTopK(perLeafResults, firstPassK);
+            if (KNNSettings.isShardLevelRescoringDisabledForDiskBasedVector(knnQuery.getIndexName()) == false) {
+                ResultUtil.reduceToTopK(perLeafResults, firstPassK);
+            }
```
navneet1v, line 68: can we have some logs may be deubg logs/warm logs to say that shard level rescoring is disabled.

jmazanec15 (approved):

src/main/java/org/opensearch/knn/index/mapper/CompressionLevel.java
```diff
@@ -97,32 +97,35 @@ public static boolean isConfigured(CompressionLevel compressionLevel) {
...
+            if (dimension >= RescoreContext.DIMENSION_THRESHOLD_1000) {
+                // No oversampling for dimensions >= 1000
+                return RescoreContext.builder().oversampleFactor(RescoreContext.OVERSAMPLE_FACTOR_1000).build();
+            } else if (dimension >= RescoreContext.DIMENSION_THRESHOLD_768) {
+                // 2x oversampling for dimensions >= 768 but < 1000
+                return RescoreContext.builder().oversampleFactor(RescoreContext.OVERSAMPLE_FACTOR_768).build();
             } else {
-                return defaultRescoreContext;
+                // 3x oversampling for dimensions < 768
+                return RescoreContext.builder().oversampleFactor(RescoreContext.OVERSAMPLE_FACTOR_BELOW_768).build();
```
navneet1v, line 125: should we go with these thresholds only when shard level rescoring is disabled?

Vikasht34: I thought about it and disccused with Jack ..., we don't have way to know Index Name ..to have this ...

src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java
Vikasht34: Hahaha , I just deleted from Jack comment .....

src/main/java/org/opensearch/knn/index/mapper/CompressionLevel.java
Vikasht34: As Discussed right now we don't have way to get Index Name so this can not be done.

src/main/java/org/opensearch/knn/index/KNNSettings.java
```diff
@@ -88,6 +88,7 @@ public class KNNSettings {
     public static final String QUANTIZATION_STATE_CACHE_SIZE_LIMIT = "knn.quantization.cache.size.limit";
     public static final String QUANTIZATION_STATE_CACHE_EXPIRY_TIME_MINUTES = "knn.quantization.cache.expiry.minutes";
     public static final String KNN_FAISS_AVX512_DISABLED = "knn.faiss.avx512.disabled";
+    public static final String KNN_DISK_VECTOR_SHARD_LEVEL_RESCORING_DISABLED = "index.knn.disk.vector.shard_level_rescoring_disabled";
```
shatejas, line 91: nit: A disabled suffix might cause confusion here if we ever expose this to customers. consider `index.knn.shard_level_rescoring` and the default can be false

jmazanec15 (approved):

naveentatikonda (approved):

PR #2174, opened 2024-10-01
Add support for radial search in exact search

VijayanB:
### Description
When threshold value is set, knn plugin will not be creating graph. Hence, when search request is trigged during that time, exact search will return valid results. However, radial search was never included as part of exact search. This will break radial search when threshold is added and radial search is requested. In this commit, new method
is introduced to accept min score and return documents that are greater than min score, similar to how radial search is performed by native engines. This search is independent of engine, but, radial search is supported only for FAISS engine out of all native engines.

### Related Issues
Part of #1942 

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/query/ExactSearcher.java
+++ b/src/main/java/org/opensearch/knn/index/query/ExactSearcher.java
@@ -21,6 +21,7 @@ import org.opensearch.knn.common.FieldInfoExtractor;
 import org.opensearch.knn.index.SpaceType;
 import org.opensearch.knn.index.VectorDataType;
 import org.opensearch.knn.index.query.iterators.BinaryVectorIdsKNNIterator;
+import org.opensearch.knn.index.engine.KNNEngine;
 import org.opensearch.knn.index.query.iterators.ByteVectorIdsKNNIterator;
 import org.opensearch.knn.index.query.iterators.NestedBinaryVectorIdsKNNIterator;
 import org.opensearch.knn.index.query.iterators.VectorIdsKNNIterator;
@@ -36,7 +37,9 @@ import org.opensearch.knn.indices.ModelDao;
 
 import java.io.IOException;
 import java.util.HashMap;
+import java.util.Locale;
 import java.util.Map;
+import java.util.function.Predicate;
 
 @Log4j2
 @AllArgsConstructor
@@ -55,11 +58,41 @@ public class ExactSearcher {
     public Map<Integer, Float> searchLeaf(final LeafReaderContext leafReaderContext, final ExactSearcherContext exactSearcherContext)
         throws IOException {
         KNNIterator iterator = getKNNIterator(leafReaderContext, exactSearcherContext);
+        if (exactSearcherContext.getKnnQuery().getRadius() != null) {
+            return doRadialSearch(leafReaderContext, exactSearcherContext, iterator);
+        }
         if (exactSearcherContext.getMatchedDocs() != null
             && exactSearcherContext.getMatchedDocs().cardinality() <= exactSearcherContext.getK()) {
             return scoreAllDocs(iterator);
         }
         return searchTopK(iterator, exactSearcherContext.getK());
+    }
+
+    /**
+     * Perform radial search by comparing scores with min score. Currently, FAISS from native engine supports radial search.
+     * Hence, we assume that Radius from knnQuery is always distance, and we convert it to score since we do exact search uses scores
+     * to filter out the documents that does not have given min score.
+     * @param leafReaderContext
+     * @param exactSearcherContext
+     * @param iterator {@link KNNIterator}
+     * @return Map of docId and score
+     * @throws IOException exception raised by iterator during traversal
+     */
+    private Map<Integer, Float> doRadialSearch(
+        LeafReaderContext leafReaderContext,
+        ExactSearcherContext exactSearcherContext,
+        KNNIterator iterator
+    ) throws IOException {
+        final SegmentReader reader = Lucene.segmentReader(leafReaderContext.reader());
+        final KNNQuery knnQuery = exactSearcherContext.getKnnQuery();
+        final FieldInfo fieldInfo = reader.getFieldInfos().fieldInfo(knnQuery.getField());
+        final KNNEngine engine = FieldInfoExtractor.extractKNNEngine(fieldInfo);
+        if (KNNEngine.FAISS != engine) {
+            throw new IllegalArgumentException(String.format(Locale.ROOT, "Engine [%s] does not support radial search", engine));
+        }
+        final SpaceType spaceType = FieldInfoExtractor.getSpaceType(modelDao, fieldInfo);
+        final float minScore = spaceType.scoreTranslation(knnQuery.getRadius());
+        return filterDocsByMinScore(exactSearcherContext, iterator, minScore);
     }
 
     private Map<Integer, Float> scoreAllDocs(KNNIterator iterator) throws IOException {
@@ -71,15 +104,19 @@ public class ExactSearcher {
         return docToScore;
     }
 
-    private Map<Integer, Float> searchTopK(KNNIterator iterator, int k) throws IOException {
+    private Map<Integer, Float> searchTopCandidates(KNNIterator iterator, int limit, Predicate<Float> filterScore) throws IOException {
         // Creating min heap and init with MAX DocID and Score as -INF.
-        final HitQueue queue = new HitQueue(k, true);
+        final HitQueue queue = new HitQueue(limit, true);
         ScoreDoc topDoc = queue.top();
         final Map<Integer, Float> docToScore = new HashMap<>();
         int docId;
         while ((docId = iterator.nextDoc()) != DocIdSetIterator.NO_MORE_DOCS) {
-            if (iterator.score() > topDoc.score) {
-                topDoc.score = iterator.score();
+            final float currentScore = iterator.score();
+            if (filterScore != null && Predicate.not(filterScore).test(currentScore)) {
+                continue;
+            }
+            if (currentScore > topDoc.score) {
+                topDoc.score = currentScore;
                 topDoc.doc = docId;
                 // As the HitQueue is min heap, updating top will bring the doc with -INF score or worst score we
                 // have seen till now on top.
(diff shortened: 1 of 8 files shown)


VijayanB:
This depends on https://github.com/opensearch-project/k-NN/pull/2136

src/main/java/org/opensearch/knn/index/query/ExactSearcher.java
```diff
@@ -55,29 +58,67 @@ public class ExactSearcher {
...
+        final SegmentReader reader = Lucene.segmentReader(leafReaderContext.reader());
+        final KNNQuery knnQuery = exactSearcherContext.getKnnQuery();
+        final FieldInfo fieldInfo = reader.getFieldInfos().fieldInfo(knnQuery.getField());
+        final KNNEngine engine = FieldInfoExtractor.extractKNNEngine(fieldInfo);
+        if (KNNEngine.FAISS != engine) {
+            throw new IllegalArgumentException(String.format(Locale.ROOT, "Engine [%s] does not support radial search", engine));
+        }
+        final SpaceType spaceType = FieldInfoExtractor.getSpaceType(modelDao, fieldInfo);
+        final float minScore = spaceType.scoreTranslation(knnQuery.getRadius());
+        return filterDocsByMinScore(exactSearcherContext, iterator, minScore);
```
junqiu-lei, line 95: Is it big effort to directly use distance to calculate in the exact search?

VijayanB: Yes. We use similarity function from lucene. If we want to use distance, it is not straight forward,

src/main/java/org/opensearch/knn/index/query/ExactSearcher.java
```diff
@@ -55,29 +58,67 @@ public class ExactSearcher {
...
+        LeafReaderContext leafReaderContext,
+        ExactSearcherContext exactSearcherContext,
+        KNNIterator iterator
+    ) throws IOException {
+        final SegmentReader reader = Lucene.segmentReader(leafReaderContext.reader());
+        final KNNQuery knnQuery = exactSearcherContext.getKnnQuery();
+        final FieldInfo fieldInfo = reader.getFieldInfos().fieldInfo(knnQuery.getField());
+        final KNNEngine engine = FieldInfoExtractor.extractKNNEngine(fieldInfo);
+        if (KNNEngine.FAISS != engine) {
+            throw new IllegalArgumentException(String.format(Locale.ROOT, "Engine [%s] does not support radial search", engine));
```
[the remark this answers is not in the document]
VijayanB: Correct. This is only called by Native Engines. Hence we can assume that only faiss is supported.

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -441,14 +431,59 @@ private boolean isExactSearchThresholdSettingSet(int filterThresholdValue) {
...
+    /**
+     * This condition mainly checks whether exact search should be performed or not
+     * @param context LeafReaderContext
+     * @param filterIdsCount count of filtered Doc ids
+     * @param annResultCount Count of Nearest Neighbours we got after doing filtered ANN Search.
+     * @return boolean - true if exactSearch needs to be done after ANNSearch.
+     */
+    private boolean isExactSearchRequire(final LeafReaderContext context, final int filterIdsCount, final int annResultCount) {
+        if (annResultCount == 0 && isMissingNativeEngineFiles(context)) {
+            log.debug("Perform exact search after approximate search since no native engine files are available");
```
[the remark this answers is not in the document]
VijayanB: Since this is in path of query, recommendation is to avoid info.

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -441,14 +431,59 @@ private boolean isExactSearchThresholdSettingSet(int filterThresholdValue) {
...
+            return true;
+        }
+        if (isFilteredExactSearchRequireAfterANNSearch(filterIdsCount, annResultCount)) {
+            log.debug(
```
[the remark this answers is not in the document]
VijayanB: Since this is in path of query, recommendation is to avoid info.

src/main/java/org/opensearch/knn/index/query/ExactSearcher.java
```diff
@@ -98,10 +135,20 @@ private Map<Integer, Float> searchTopK(KNNIterator iterator, int k) throws IOExc
             final ScoreDoc doc = queue.pop();
             docToScore.put(doc.doc, doc.score);
         }
-
         return docToScore;
     }
 
+    private Map<Integer, Float> searchTopK(KNNIterator iterator, int k) throws IOException {
+        return searchTopCandidates(iterator, k, null);
```
shatejas, line 142: consider passing in `score -> true` to avoid nulls

nit: probably a wrapper method is not needed here since its 1 line of code

src/main/java/org/opensearch/knn/index/query/ExactSearcher.java
```diff
@@ -71,15 +104,19 @@ private Map<Integer, Float> scoreAllDocs(KNNIterator iterator) throws IOExceptio
...
         final Map<Integer, Float> docToScore = new HashMap<>();
         int docId;
         while ((docId = iterator.nextDoc()) != DocIdSetIterator.NO_MORE_DOCS) {
-            if (iterator.score() > topDoc.score) {
-                topDoc.score = iterator.score();
+            final float currentScore = iterator.score();
+            if (filterScore != null && Predicate.not(filterScore).test(currentScore)) {
+                continue;
+            }
+            if (currentScore > topDoc.score) {
```
shatejas, line 118: compiler will make this `if () else { if() }` so this is not reducing nesting. would recommend making filter score @Nonnull and do `if (predicate.test && currentScore > topDoc.score)` if the intent is to reduce nesting

src/main/java/org/opensearch/knn/index/query/ExactSearcher.java
```diff
@@ -55,13 +58,43 @@ public class ExactSearcher {
...
+    private Map<Integer, Float> doRadialSearch(
+        LeafReaderContext leafReaderContext,
+        ExactSearcherContext exactSearcherContext,
+        KNNIterator iterator
+    ) throws IOException {
+        final SegmentReader reader = Lucene.segmentReader(leafReaderContext.reader());
+        final KNNQuery knnQuery = exactSearcherContext.getKnnQuery();
+        final FieldInfo fieldInfo = reader.getFieldInfos().fieldInfo(knnQuery.getField());
+        final KNNEngine engine = FieldInfoExtractor.extractKNNEngine(fieldInfo);
+        if (KNNEngine.FAISS != engine) {
```
shatejas, line 90: I would try to avoid this and rely on RNNQueryFactory to throw upfront if engine is not correct. And to cover the case with a negative IT where it throws. Just feel this is way too late to throw a validation error and in a way redundant since its covered [here](https://github.com/opensearch-project/k-NN/blob/07f4df2015a0af5158859c42c98511b865005eac/src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java#L447)

shatejas (commented):
There doesn't seem to be updates in unit tests, will it be present in another iteration?

src/main/java/org/opensearch/knn/index/query/ExactSearcher.java
VijayanB: It is mainly added because of assumption that radius always returns distance instead of score, hence, in future if our current assumption became invalid, this check will make sure we don't overlook. Like you mentioned, we will throw exception in Query builder for index with nmslib as engine

src/main/java/org/opensearch/knn/index/query/ExactSearcher.java
VijayanB: Ack

src/main/java/org/opensearch/knn/index/query/ExactSearcher.java
VijayanB: updated

VijayanB:
> There doesn't seem to be updates in unit tests, will it be present in another iteration?

Added unit test

src/test/java/org/opensearch/knn/index/query/KNNWeightTests.java
```diff
@@ -763,6 +763,95 @@ public void validateANNWithFilterQuery_whenExactSearch_thenSuccess(final boolean
         }
     }
 
+    @SneakyThrows
+    public void testRadialSearch_whenNoEngineFiles_thenPerformExactSearch() {
```
navneet1v, line 767: Thought I see we are using this test to check ExactSearch for radial search and avoided the Unit test for ExactSearch with radial search threshold. See if you want to mock ExactSearcher search class and all its invocation to make it proper Unit test. Since exact search is not getting added more and more extra logic do you see we should have separate unit test class for ExactSearcher class.

VijayanB: Good point. I can create Github issue for refactor. I see two option, 1) Keep this PR as it is and update as part of GH since there are more test cases that uses exact search, or 2) I can only do it for radial search, as part of GH others can be moved.

navneet1v: I like the option 2. Lets do atleast for this test case and create a GH issue too for the test refactor.

VijayanB: updated this PR. Will create GH issue

src/test/java/org/opensearch/knn/index/query/ExactSearcherTests.java
```diff
@@ -0,0 +1,139 @@
...
+import static org.opensearch.knn.common.KNNConstants.KNN_ENGINE;
+import static org.opensearch.knn.common.KNNConstants.PARAMETERS;
+import static org.opensearch.knn.common.KNNConstants.SPACE_TYPE;
+
+public class ExactSearcherTests extends KNNTestCase {
+
+    private static final String SEGMENT_NAME = "0";
+
+    @SneakyThrows
+    public void testRadialSearch_whenNoEngineFiles_thenSuccess() {
```
shatejas, line 54: Thanks for adding these tests

shatejas (approved):

src/test/java/org/opensearch/knn/index/query/KNNWeightTests.java
VijayanB: https://github.com/opensearch-project/k-NN/issues/2180

src/main/java/org/opensearch/knn/index/query/ExactSearcher.java
```diff
@@ -71,15 +106,17 @@ private Map<Integer, Float> scoreAllDocs(KNNIterator iterator) throws IOExceptio
         return docToScore;
     }
 
-    private Map<Integer, Float> searchTopK(KNNIterator iterator, int k) throws IOException {
+    private Map<Integer, Float> searchTopCandidates(KNNIterator iterator, int limit, @NonNull Predicate<Float> filterScore)
```
vibrantvarun, line 109: Curious why k variable is renamed to limit?

VijayanB: This method will be called by radialSearch as well. Hence, renamed it to K to be more generic.

vibrantvarun (approved):

jmazanec15 (approved):

PR #2183, opened 2024-10-03
Score Fix for Binary Quantized Vector and Setting Default value in case of shard level rescoring is disabled for oversampling factor

Vikasht34:
### Description
**Change1**

Problem
In Disk Vector Mode, while calculating the score for a binary quantized vector, the system was incorrectly using Inner Product as the similarity function. However, for binary quantized vectors, the correct distance metric should be Hamming Distance.

Impact:
This incorrect scoring mechanism led to poor results, particularly in cases with multiple segments. During the process of reducing to the top k candidates, the best candidates (those with the most relevant results) were being filtered out. As a result, the recall was zero, meaning no relevant results were being returned, which critically affected the accuracy of search results.

Expected Behavior:
The system should use Hamming Distance for binary quantized vectors, ensuring that the correct similarity metric is applied. This would allow the best candidates to be retained during the top-k reduction process, thereby maintaining high recall and improving overall search performance in multi-segment scenarios.

**Change 2:**

Problem:
The system already had a setting to disable shard-level rescoring, allowing for rescoring at the segment level. However, with a recent PR, a default oversampling factor was reintroduced to be applied when the user did not explicitly provide one. Unfortunately, this default oversampling factor was being applied in both cases — regardless of whether shard-level rescoring was enabled or disabled. With Segment level , we don't need high oversampling factor. 

Solution:
This PR introduces changes to ensure that the default oversampling factor is only applied when shard-level rescoring is disabled. When segment-level rescoring is enabled, the system overrides the default oversampling factor, ensuring the correct sampling factor is applied based on dimesion.

### Related Issues

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/query/KNNWeight.java
+++ b/src/main/java/org/opensearch/knn/index/query/KNNWeight.java
@@ -375,6 +375,10 @@ public class KNNWeight extends Weight {
             return null;
         }
 
+        if (quantizedVector != null) {
+            return Arrays.stream(results)
+                .collect(Collectors.toMap(KNNQueryResult::getId, result -> knnEngine.score(result.getScore(), SpaceType.HAMMING)));
+        }
         return Arrays.stream(results)
             .collect(Collectors.toMap(KNNQueryResult::getId, result -> knnEngine.score(result.getScore(), spaceType)));
     }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 * Adding Support to Enable/Disble Share level Rescoring and Update Oversampling Factor[#2172](https://github.com/opensearch-project/k-NN/pull/2172)
 ### Bug Fixes
 * KNN80DocValues should only be considered for BinaryDocValues fields [#2147](https://github.com/opensearch-project/k-NN/pull/2147)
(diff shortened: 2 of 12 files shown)


src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -375,6 +375,10 @@ private Map<Integer, Float> doANNSearch(
             return null;
         }
 
+        if (quantizedVector != null) {
+            return Arrays.stream(results)
+                .collect(Collectors.toMap(KNNQueryResult::getId, result -> knnEngine.score(result.getScore(), SpaceType.HAMMING)));
```
jmazanec15, line 380: Should we just override spaceType?

src/main/java/org/opensearch/knn/index/query/parser/RescoreParser.java
```diff
@@ -106,6 +107,7 @@ public static void streamOutput(StreamOutput out, RescoreContext rescoreContext)
             return;
         }
         out.writeOptionalFloat(rescoreContext == null ? null : rescoreContext.getOversampleFactor());
+        out.writeOptionalBoolean(rescoreContext == null ? null : rescoreContext.isUserProvided());
```
jmazanec15, line 110: Why do we need this? This could lead to a break in BWC. Could we just ensure the optional float encapsulates this info?

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
Vikasht34: It takes final , we can not override.

jmazanec15 (approved):

naveentatikonda (approved):

PR #2200, opened 2024-10-09
Add CompressionLevel Calculation for PQ

jmazanec15:
### Description
Currently, for product quantization, we set the calculated compression level to NOT_CONFIGURED. The main issue with this is that if a user sets up a disk-based index with PQ, no re-scoring will happen by default.

This change adds the calculation so that the proper re-scoring will happen. The formula is fairly straightforward =>
actual compression = (d * 32) / (m * code_size). Then, we round to the neareste compression level (because we only support discrete compression levels).

One small issue with this is that if PQ is configured to have compression > 32x, the value will be 32x. Functionally, the only issue will be that we may not be as aggressive on oversampling for on disk mode.

### Check List


navneet1v:
> One small issue with this is that if PQ is configured to have compression > 32x, the value will be 32x. Functionally, the only issue will be that we may not be as aggressive on oversampling for on disk mode.

should we allow more compression level?

src/main/java/org/opensearch/knn/index/mapper/CompressionLevel.java
```diff
@@ -29,6 +29,8 @@ public enum CompressionLevel {
     x16(16, "16x", new RescoreContext(3.0f, false), Set.of(Mode.ON_DISK)),
     x32(32, "32x", new RescoreContext(3.0f, false), Set.of(Mode.ON_DISK));
 
+    public static final CompressionLevel MAX_COMPRESSION_LEVEL = CompressionLevel.x32;
```
navneet1v, line 32: should we have 64 as a max compression level? I don't have a solid point to have 64 but I think have 1 more extra compression is always good.

navneet1v (commented):
overall looks good to me. Just 1 minor comment.

heemin32 (commented):
LGTM. Do we want to add integ test for it? Recall should be higher with rescoring than baseline.

heemin32 (approved):

src/main/java/org/opensearch/knn/index/mapper/CompressionLevel.java
jmazanec15: I think that makes sense. I guess for 64x, default for all dimensions should probably be 5x.

navneet1v (approved):

heemin32 (approved):

jmazanec15:
@navneet1v we did here: https://github.com/opensearch-project/k-NN/issues/1779#user-content-appendix-b-baseline-rescore-experiments

navneet1v:
> @navneet1v we did here: [#1779 (comment)](https://github.com/opensearch-project/k-NN/issues/1779#user-content-appendix-b-baseline-rescore-experiments)

Thanks. I completely forgot about that.

PR #2212, opened 2024-10-16
Remove explicit K from exact search context

VijayanB:
### Description
K can be extracted from knn query instead of explicitly passing as input to ExactSearch Context. Previously this was
added mainly with assumption that 'k' can be different from Query if rescore context is not null. However, the main reason we have different K from Query is due to quantization. Having larger K is apt for Approx search, for exact search we don't need to collect more than expected top k from query.

### Check List


jmazanec15:
> Having larger K is apt for Approx search, for exact search we don't need to collect more than expected top k from query.

If we are approximating distance with exact search, why not maintain oversample?

VijayanB:
Closing this PR since this is required for oversampling.

Issue #2163, opened 2024-09-27
Switch default vector engine from nmslib to faiss


vamshin:
**Is your feature request related to a problem?**
The nmslib vector engine is currently in maintenance mode as of 10/27, and we plan to deprecate it in the 3.x version. All the feature development/optimizations are focused in faiss engine since 2.7 version. To align the new vector indexes with these improvements, it would be beneficial to switch the default vector engine from nmslib to faiss.

**What solution would you like?**
Switch the default vector engine from nmslib to faiss.

navneet1v:
@vamshin I see this issue is put in 2.18 version? is the feature planned for 2.18?

vamshin:
@navneet1v yes, idea is to do for 2.18

vamshin:
As part of this default switch to faiss, I feel we should also switch to fp_16 with in faiss.

jmazanec15:
I think that until we can automatically handle cases where the data may not fit inside fp_16, we should not do this yet. Instead, for OOTB experience, users can specify compression and/or mode. If needed, we can introduce a new mode or similar meta parameter.

vamshin:
For 2.18, we can just make faiss default and later figure out plan for making fp_16 as default.

navneet1v:
closing this issue as 2.18 is released.

dblock:
[[Catch All Triage](https://github.com/opensearch-project/.github/blob/main/TRIAGING.md) - [1](https://github.com/dblock), [2](https://github.com/reta), [3](https://github.com/andrross/), [4](https://github.com/dolla24)]

PR #2221, opened 2024-10-20
Update default engine to FAISS

VijayanB:
### Description
- Update default engine to FAISS.
- Fixed tests that assumed default engine to be nmslib to either use nmslib explicitly or update test assertion to look for faiss specific artifacts like file extension instead of nmslib artifacts
- removed tests that are just validating default (provided it is nmslib) doesn't support certain features
### Related Issues
#2163

### Check List


VijayanB:
Doc issue: https://github.com/opensearch-project/documentation-website/issues/8589

vibrantvarun (commented):
Looks good! Will wait for other engineers review

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapperUtil.java
```diff
@@ -203,7 +203,7 @@ static KNNMethodContext createKNNMethodContextFromLegacy(
             ? topLevelSpaceType
             : KNNVectorFieldMapperUtil.getSpaceType(indexSettings);
         return new KNNMethodContext(
-            KNNEngine.NMSLIB,
```
jmazanec15, line 206: Keep as NMSLIB - nmslib is only legacy support

jmazanec15:
Can you update this too: https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/engine/EngineResolver.java#L51

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapperUtil.java
jmazanec15: maybe add comment to like I should have done haha

src/main/java/org/opensearch/knn/index/engine/KNNEngine.java
```diff
@@ -29,7 +29,7 @@ public enum KNNEngine implements KNNLibrary {
     FAISS(FAISS_NAME, Faiss.INSTANCE),
     LUCENE(LUCENE_NAME, Lucene.INSTANCE);
 
-    public static final KNNEngine DEFAULT = NMSLIB;
+    public static final KNNEngine DEFAULT = FAISS;
```
navneet1v, line 32: How we are handling the backward compatibility here? Also can we validate if we have proper BWC tests to test this default engine change.

src/test/java/org/opensearch/knn/index/KNNIndexShardTests.java
```diff
@@ -114,7 +115,7 @@ public void testGetAllEngineFileContexts() throws IOException, ExecutionExceptio
         engineFileContexts = knnIndexShard.getAllEngineFileContexts(searcher.getIndexReader());
         assertEquals(1, engineFileContexts.size());
         List<String> paths = engineFileContexts.stream().map(KNNIndexShard.EngineFileContext::getIndexPath).collect(Collectors.toList());
-        assertTrue(paths.get(0).contains("hnsw") || paths.get(0).contains("hnswc"));
+        assertTrue(paths.get(0).contains(FAISS_HNSW_EXTENSION));
```
shatejas, line 118: is the compound extension covered?

src/test/java/org/opensearch/knn/index/VectorDataTypeIT.java
```diff
@@ -263,29 +263,6 @@ public void testByteVectorDataTypeWithNmslibEngine() {
         assertTrue(ex.getMessage().contains("is not supported for vector data type"));
     }
 
-    @SneakyThrows
-    public void testByteVectorDataTypeWithLegacyFieldMapperKnnIndexSetting() {
```
shatejas, line 267: Just checking if this case is covered when engine is explicitly NMSLIB

src/test/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapperTests.java
```diff
@@ -146,7 +146,7 @@ public void testTypeParser_build_fromKnnMethodContext() throws IOException {
         // Check that knnMethodContext takes precedent over both model and legacy
         ModelDao modelDao = mock(ModelDao.class);
 
-        SpaceType spaceType = SpaceType.COSINESIMIL;
+        SpaceType spaceType = SpaceType.DEFAULT;
```
shatejas, line 149: Out of curiosity, why this change?

src/main/java/org/opensearch/knn/index/engine/KNNEngine.java
Vikasht34: I was just curious to know , why do we need to handle BWC in this case , So i created a Index , without specifying any engine so it picked NMSLIB , so now this information is already there in Cluster State and Serialised and passed to all Nodes, Now During query time , we do get it information from there and it does not depend on the new Default.

src/test/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapperTests.java
jmazanec15: @vamshin is it okay default engine doesnt support cosine?

jmazanec15: We could add this by taking a normalization and then setting to innerproduct. But this would take some effort.

vamshin: ah good catch! I think still we need to make faiss as default based on all the advanced capabilities like nested fields, efficient filters, disk optimized search, in future GPUs. In long term makes sense to build the normalization.

src/main/java/org/opensearch/knn/index/engine/KNNEngine.java
navneet1v: >  So i created a Index , without specifying any engine so it picked NMSLIB , so now this information is already there in Cluster State and Serialised and passed to all Nodes,

the engine gets persisted with indexMapping only when you create the index and also provide the key `method`. But if you don't specify the key `method` while creating the index and lets say create an index like this:
```
PUT my-knn-index-1
{
  "settings": {
    "index": {
      "knn": true
    }
  },
  "mappings": {
    "properties": {
        "my_vector1": {
          "type": "knn_vector",
          "dimension": 2
        }
    }
  }
}
```

then we infer the engine. Ref: https://github.com/opensearch-project/k-NN/blob/HEAD/src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapperUtil.java#L196-L219 for more details.

src/test/java/org/opensearch/knn/index/VectorDataTypeIT.java
VijayanB: Yes, it is covered explicitly for nmslib engine

src/test/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapperTests.java
VijayanB: @shatejas FAISS doesn't support cosine yet.

src/test/java/org/opensearch/knn/index/KNNIndexShardTests.java
VijayanB: Updated test to use nmslib

jmazanec15 (approved):

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java
```diff
@@ -550,6 +549,12 @@ private void setEngine(final KNNMethodContext knnMethodContext, KNNEngine knnEng
         }
     }
 
+    static boolean useKNNMethodContextFromLegacy(Builder builder, Mapper.TypeParser.ParserContext parserContext) {
```
navneet1v, line 552: why this is static? and could we move this function to : KNNVectorFieldMapperUtil class?

navneet1v (approved):
I just have a minor comment. I see

vibrantvarun (approved):

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java
VijayanB: Ack. Moved to Util class

jmazanec15 (dismissed):

navneet1v (approved):

VijayanB:
Tests are failing not related to this PR. This PR https://github.com/opensearch-project/k-NN/pull/2236 will fix those issue in main. Will rebase once the other PR is merged.

VijayanB:
Test failure is not due to this change.

jmazanec15 (approved):

naveentatikonda (approved):

VijayanB:
Failure is due to waiting for this PR to be backported https://github.com/opensearch-project/k-NN/pull/2248

PR #2241, opened 2024-10-29
Introduced writing layer, getting rid of writing logic that uses an absolute path in the filesystem.

0ctopus13prime:
### Description
This PR introduces an abstract writing layer into native engines (NMSLIB, Faiss). 
In which, made native engines rely on an write interface to do the IO instead of directly depending on File API.
Due to this layer, we can use a different kind of underlying storage to save vector index. The default one that is currently being used is FSDirectory saving the index in the local file system.

### Related Issues
RFC : https://github.com/opensearch-project/k-NN/issues/2033

### Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ## [Unreleased 2.x](https://github.com/opensearch-project/k-NN/compare/2.18...2.x)
 ### Features
 ### Enhancements
+- Introduced a writing layer in native engines where relies on the writing interface to process IO. (#2241)[https://github.com/opensearch-project/k-NN/pull/2241]
 ### Bug Fixes
 ### Infrastructure
 ### Documentation
--- a/jni/include/faiss_stream_support.h
+++ b/jni/include/faiss_stream_support.h
@@ -15,6 +15,7 @@
 #include "faiss/impl/io.h"
 #include "jni_util.h"
 #include "native_engines_stream_support.h"
+#include "parameter_utils.h"
 
 #include <jni.h>
 #include <stdexcept>
@@ -34,7 +35,7 @@ class FaissOpenSearchIOReader final : public faiss::IOReader {
  public:
   explicit FaissOpenSearchIOReader(NativeEngineIndexInputMediator *_mediator)
       : faiss::IOReader(),
-        mediator(_mediator) {
+        mediator(knn_jni::util::ParameterCheck::require_non_null(_mediator, "mediator")) {
     name = "FaissOpenSearchIOReader";
   }
 
@@ -54,6 +55,40 @@ class FaissOpenSearchIOReader final : public faiss::IOReader {
  private:
   NativeEngineIndexInputMediator *mediator;
 };  // class FaissOpenSearchIOReader
+
+
+/**
+ * A glue component inheriting IOWriter to delegate IO processing down to the given
+ * mediator. The mediator is expected to do write bytes via the provided Lucene's IndexOutput.
+ */
+class FaissOpenSearchIOWriter final : public faiss::IOWriter {
+ public:
+  explicit FaissOpenSearchIOWriter(NativeEngineIndexOutputMediator *_mediator)
+      : faiss::IOWriter(),
+        mediator(knn_jni::util::ParameterCheck::require_non_null(_mediator, "mediator")) {
+    name = "FaissOpenSearchIOWriter";
+  }
+
+  size_t operator()(const void *ptr, size_t size, size_t nitems) final {
+    const auto writeBytes = size * nitems;
+    if (writeBytes > 0) {
+      mediator->writeBytes(reinterpret_cast<const uint8_t *>(ptr), writeBytes);
+    }
+    return nitems;
+  }
+
+  // return a file number that can be memory-mapped
+  int filedescriptor() final {
+    throw std::runtime_error("filedescriptor() is not supported in FaissOpenSearchIOWriter.");
+  }
+
+  void flush() {
+    mediator->flush();
+  }
+
+ private:
+  NativeEngineIndexOutputMediator *mediator;
+};  // class FaissOpenSearchIOWriter
 
 
 
--- a/jni/include/jni_util.h
+++ b/jni/include/jni_util.h
@@ -18,6 +18,7 @@
 #include <vector>
 #include <unordered_map>
 #include <cstdint>
+#include <functional>
 
 namespace knn_jni {
 
@@ -33,7 +34,7 @@ namespace knn_jni {
         virtual void HasExceptionInStack(JNIEnv* env) = 0;
 
         // HasExceptionInStack with ability to specify message
-        virtual void HasExceptionInStack(JNIEnv* env, const std::string& message) = 0;
+        virtual void HasExceptionInStack(JNIEnv* env, const char *message) = 0;
 
         // Catches a C++ exception and throws the corresponding exception to the JVM
         virtual void CatchCppExceptionAndThrowJava(JNIEnv* env) = 0;
@@ -144,6 +145,9 @@ namespace knn_jni {
         virtual jlong CallNonvirtualLongMethodA(JNIEnv * env, jobject obj, jclass clazz,
                                                 jmethodID methodID, jvalue* args) = 0;
 
+        virtual void CallNonvirtualVoidMethodA(JNIEnv * env, jobject obj, jclass clazz,
+                                               jmethodID methodID, jvalue* args) = 0;
+
         // --------------------------------------------------------------------------
     };
 
@@ -158,7 +162,7 @@ namespace knn_jni {
 
         void ThrowJavaException(JNIEnv* env, const char* type = "", const char* message = "") final;
         void HasExceptionInStack(JNIEnv* env) final;
-        void HasExceptionInStack(JNIEnv* env, const std::string& message) final;
+        void HasExceptionInStack(JNIEnv* env, const char* message) final;
         void CatchCppExceptionAndThrowJava(JNIEnv* env) final;
         jclass FindClass(JNIEnv * env, const std::string& className) final;
         jmethodID FindMethod(JNIEnv * env, const std::string& className, const std::string& methodName) final;
@@ -200,13 +204,38 @@ namespace knn_jni {
         jfieldID GetFieldID(JNIEnv * env, jclass clazz, const char *name, const char *sig) final;
         jint CallNonvirtualIntMethodA(JNIEnv *env, jobject obj, jclass clazz, jmethodID methodID, jvalue *args) final;
         jlong CallNonvirtualLongMethodA(JNIEnv * env, jobject obj, jclass clazz, jmethodID methodID, jvalue* args) final;
+        void CallNonvirtualVoidMethodA(JNIEnv * env, jobject obj, jclass clazz, jmethodID methodID, jvalue* args) final;
         void * GetPrimitiveArrayCritical(JNIEnv * env, jarray array, jboolean *isCopy) final;
         void ReleasePrimitiveArrayCritical(JNIEnv * env, jarray array, void *carray, jint mode) final;
 
     private:
         std::unordered_map<std::string, jclass> cachedClasses;
         std::unordered_map<std::string, jmethodID> cachedMethods;
-    };
+    };  // class JNIUtil
+
+    /**
+     * It's common cleaner to release a primitive array within its destructor.
+     * Ex: JNIReleaseElements release_int_array_elements {[=](){
+     *   jniUtil->ReleaseIntArrayElements(env, idsJ, idsCpp, JNI_ABORT);
+     * }};
+     */
+    struct JNIReleaseElements {
+      explicit JNIReleaseElements(std::function<void()> _release_func)
+          : release_func(std::move(_release_func)) {
+      }
+
+      ~JNIReleaseElements() {
+        try {
+          if (release_func) {
+            release_func();
+          }
+        } catch (...) {
+          // Ignore
+        }
+      }
+
+      std::function<void()> release_func;
+    };  // struct ReleaseIntArrayElements
 
     // ------------------------------- CONSTANTS --------------------------------
     extern const std::string FAISS_NAME;
--- a/jni/include/native_engines_stream_support.h
+++ b/jni/include/native_engines_stream_support.h
@@ -13,6 +13,8 @@
 #define OPENSEARCH_KNN_JNI_STREAM_SUPPORT_H
 
 #include "jni_util.h"
+#include "parameter_utils.h"
+#include "memory_util.h"
 
 #include <jni.h>
 #include <stdexcept>
@@ -21,8 +23,6 @@
 
 namespace knn_jni {
 namespace stream {
-
-
 
 /**
  * This class contains Java IndexInputWithBuffer reference and calls its API to copy required bytes into a read buffer.
@@ -33,9 +33,10 @@ class NativeEngineIndexInputMediator {
   NativeEngineIndexInputMediator(JNIUtilInterface *_jni_interface,
                                  JNIEnv *_env,
                                  jobject _indexInput)
-      : jni_interface(_jni_interface),
-        env(_env),
-        indexInput(_indexInput),
+      : jni_interface(knn_jni::util::ParameterCheck::require_non_null(
+          _jni_interface, "jni_interface")),
+        env(knn_jni::util::ParameterCheck::require_non_null(_env, "env")),
+        indexInput(knn_jni::util::ParameterCheck::require_non_null(_indexInput, "indexInput")),
         bufferArray((jbyteArray) (_jni_interface->GetObjectField(_env,
                                                                  _indexInput,
                                                                  getBufferFieldId(_jni_interface, _env)))),
@@ -43,7 +44,7 @@ class NativeEngineIndexInputMediator {
         remainingBytesMethod(getRemainingBytesMethod(_jni_interface, _env)) {
   }
 
-  void copyBytes(int64_t nbytes, uint8_t *destination) {
+  void copyBytes(int64_t nbytes, uint8_t * RESTRICT destination) {
     auto jclazz = getIndexInputWithBufferClass(jni_interface, env);
 
     while (nbytes > 0) {
@@ -52,11 +53,12 @@ class NativeEngineIndexInputMediator {
       args.j = nbytes;
       const auto readBytes =
           jni_interface->CallNonvirtualIntMethodA(env, indexInput, jclazz, copyBytesMethod, &args);
+      jni_interface->HasExceptionInStack(env, "Reading bytes via IndexInput has failed.");
 
       // === Critical Section Start ===
 
       // Get primitive array pointer, no copy is happening in OpenJDK.
-      auto primitiveArray =
+      jbyte * RESTRICT primitiveArray =
           (jbyte *) jni_interface->GetPrimitiveArrayCritical(env, bufferArray, nullptr);
 
       // Copy Java bytes to C++ destination address.
@@ -65,6 +67,7 @@ class NativeEngineIndexInputMediator {
       // Release the acquired primitive array pointer.
       // JNI_ABORT tells JVM to directly free memory without copying back to Java byte[].
       // Since we're merely copying data, we don't need to copying back.
+      // Note than when we received an internal primitive array pointer, then the mode will be ignored.
       jni_interface->ReleasePrimitiveArrayCritical(env, bufferArray, primitiveArray, JNI_ABORT);
 
       // === Critical Section End ===
@@ -75,11 +78,13 @@ class NativeEngineIndexInputMediator {
   }
 
   int64_t remainingBytes() {
-      return jni_interface->CallNonvirtualLongMethodA(env,
-                                                      indexInput,
-                                                      getIndexInputWithBufferClass(jni_interface, env),
-                                                      remainingBytesMethod,
-                                                      nullptr);
+    auto bytes = jni_interface->CallNonvirtualLongMethodA(env,
+                                                          indexInput,
+                                                          getIndexInputWithBufferClass(jni_interface, env),
+                                                          remainingBytesMethod,
+                                                          nullptr);
+    jni_interface->HasExceptionInStack(env, "Checking remaining bytes has failed.");
+    return bytes;
   }
 
  private:
@@ -116,6 +121,110 @@ class NativeEngineIndexInputMediator {
   jmethodID copyBytesMethod;
   jmethodID remainingBytesMethod;
 }; // class NativeEngineIndexInputMediator
+
+
+
+/**
+ * This class delegates the provided index output to do IO processing.
+ * In most cases, it is expected that IndexOutputWithBuffer was passed down to this,
+ * which eventually have Lucene's IndexOutput to write bytes.
+ */
+class NativeEngineIndexOutputMediator {
+ public:
+  NativeEngineIndexOutputMediator(JNIUtilInterface *_jni_interface,
+                                  JNIEnv *_env,
+                                  jobject _indexOutput)
+      : jni_interface(knn_jni::util::ParameterCheck::require_non_null(_jni_interface, "jni_interface")),
+        env(knn_jni::util::ParameterCheck::require_non_null(_env, "env")),
+        indexOutput(knn_jni::util::ParameterCheck::require_non_null(_indexOutput, "indexOutput")),
+        bufferArray((jbyteArray) (_jni_interface->GetObjectField(_env,
+                                                                 _indexOutput,
+                                                                 getBufferFieldId(_jni_interface, _env)))),
+        writeBytesMethod(getWriteBytesMethod(_jni_interface, _env)),
+        bufferLength(jni_interface->GetJavaBytesArrayLength(env, bufferArray)),
+        nextWriteIndex() {
+  }
+
+  void writeBytes(const uint8_t * RESTRICT source, size_t nbytes) {
+    auto left = nbytes;
+    while (left > 0) {
+      const auto writeBytes = std::min(bufferLength - nextWriteIndex, left);
+
+      // === Critical Section Start ===
+
+      // Get primitive array pointer, no copy is happening in OpenJDK.
+      jbyte * RESTRICT primitiveArray =
+          (jbyte *) jni_interface->GetPrimitiveArrayCritical(env, bufferArray, nullptr);
+
+      // Copy the given bytes to Java byte[] address.
+      std::memcpy(primitiveArray + nextWriteIndex, source, writeBytes);
+
+      // Release the acquired primitive array pointer.
+      // 0 tells JVM to copy back the content, and to free the pointer. It will be ignored if we acquired an internal
+      // primitive array pointer instead of a copied version.
+      // From JNI docs:
+      // Mode 0 : copy back the content and free the elems buffer
+      // The mode argument provides information on how the array buffer should be released. mode has no effect if elems
+      // is not a copy of the elements in array.
+      jni_interface->ReleasePrimitiveArrayCritical(env, bufferArray, primitiveArray, 0);
+
+      // === Critical Section End ===
+
+      nextWriteIndex += writeBytes;
+      if (nextWriteIndex >= bufferLength) {
+        callWriteBytesInIndexOutput();
+      }
+
+      source += writeBytes;
+      left -= writeBytes;
+    }  // End while
+  }
+
+  void flush() {
+    if (nextWriteIndex > 0) {
+      callWriteBytesInIndexOutput();
+    }
+  }
+
+ private:
+  static jclass getIndexOutputWithBufferClass(JNIUtilInterface *jni_interface, JNIEnv *env) {
+    static jclass INDEX_OUTPUT_WITH_BUFFER_CLASS =
+        jni_interface->FindClassFromJNIEnv(env, "org/opensearch/knn/index/store/IndexOutputWithBuffer");
+    return INDEX_OUTPUT_WITH_BUFFER_CLASS;
+  }
+
+  static jmethodID getWriteBytesMethod(JNIUtilInterface *jni_interface, JNIEnv *env) {
+    static jmethodID WRITE_METHOD_ID =
+        jni_interface->GetMethodID(env, getIndexOutputWithBufferClass(jni_interface, env), "writeBytes", "(I)V");
+    return WRITE_METHOD_ID;
+  }
+
+  static jfieldID getBufferFieldId(JNIUtilInterface *jni_interface, JNIEnv *env) {
+    static jfieldID BUFFER_FIELD_ID =
+        jni_interface->GetFieldID(env, getIndexOutputWithBufferClass(jni_interface, env), "buffer", "[B");
+    return BUFFER_FIELD_ID;
+  }
+
+  void callWriteBytesInIndexOutput() {
+    auto jclazz = getIndexOutputWithBufferClass(jni_interface, env);
+    // Initializing the first integer parameter of `writeBytes`.
+    // `i` represents an integer parameter.
+    jvalue args {.i = nextWriteIndex};
+    jni_interface->CallNonvirtualVoidMethodA(env, indexOutput, jclazz, writeBytesMethod, &args);
+    jni_interface->HasExceptionInStack(env, "Writing bytes via IndexOutput has failed.");
+    nextWriteIndex = 0;
+  }
+
+  JNIUtilInterface *jni_interface;
+  JNIEnv *env;
+
+  // `IndexOutputWithBuffer` instance having `IndexOutput` instance obtained from `Directory` for reading.
+  jobject indexOutput;
+  jbyteArray bufferArray;
+  jmethodID writeBytesMethod;
+  size_t bufferLength;
+  int32_t nextWriteIndex;
+};  // NativeEngineIndexOutputMediator
 
 
 
--- a/jni/include/nmslib_stream_support.h
+++ b/jni/include/nmslib_stream_support.h
@@ -13,11 +13,11 @@
 #define OPENSEARCH_KNN_JNI_NMSLIB_STREAM_SUPPORT_H
 
 #include "native_engines_stream_support.h"
+#include "utils.h"  // This is from NMSLIB
+#include "parameter_utils.h"
 
 namespace knn_jni {
 namespace stream {
-
-
 
 /**
  * NmslibIOReader implementation delegating NativeEngineIndexInputMediator to read bytes.
@@ -25,7 +25,8 @@ namespace stream {
 class NmslibOpenSearchIOReader final : public similarity::NmslibIOReader {
  public:
   explicit NmslibOpenSearchIOReader(NativeEngineIndexInputMediator *_mediator)
-      : mediator(_mediator) {
+      : similarity::NmslibIOReader(),
+        mediator(knn_jni::util::ParameterCheck::require_non_null(_mediator, "mediator")) {
   }
 
   void read(char *bytes, size_t len) final {
@@ -43,6 +44,27 @@ class NmslibOpenSearchIOReader final : public similarity::NmslibIOReader {
   NativeEngineIndexInputMediator *mediator;
 };  // class NmslibOpenSearchIOReader
 
+
+class NmslibOpenSearchIOWriter final : public similarity::NmslibIOWriter {
+ public:
+  explicit NmslibOpenSearchIOWriter(NativeEngineIndexOutputMediator *_mediator)
+      : similarity::NmslibIOWriter(),
+        mediator(knn_jni::util::ParameterCheck::require_non_null(_mediator, "mediator")) {
+  }
+
+  void write(char *bytes, size_t len) final {
+    if (len > 0) {
+      mediator->writeBytes((uint8_t *) bytes, len);
+    }
+  }
+
+  void flush() final {
+    mediator->flush();
+  }
+
+ private:
+  NativeEngineIndexOutputMediator *mediator;
+};  // class NmslibOpenSearchIOWriter
 
 
 }
--- a/jni/src/faiss_index_service.cpp
+++ b/jni/src/faiss_index_service.cpp
@@ -9,7 +9,6 @@
 
 #include "faiss_index_service.h"
 #include "faiss_methods.h"
-#include "faiss/index_factory.h"
 #include "faiss/Index.h"
 #include "faiss/IndexBinary.h"
 #include "faiss/IndexHNSW.h"
@@ -17,8 +16,7 @@
 #include "faiss/IndexIVFFlat.h"
 #include "faiss/IndexBinaryIVF.h"
 #include "faiss/IndexIDMap.h"
-#include "faiss/index_io.h"
-#include <algorithm>
+
 #include <string>
 #include <vector>
 #include <memory>
@@ -55,20 +53,18 @@ void SetExtraParameters(knn_jni::JNIUtilInterface * jniUtil, JNIEnv *env,
     }
 }
 
-IndexService::IndexService(std::unique_ptr<FaissMethods> faissMethods) : faissMethods(std::move(faissMethods)) {}
+IndexService::IndexService(std::unique_ptr<FaissMethods> _faissMethods) : faissMethods(std::move(_faissMethods)) {}
 
 void IndexService::allocIndex(faiss::Index * index, size_t dim, size_t numVectors) {
-    if(auto * indexHNSWSQ = dynamic_cast<faiss::IndexHNSWSQ *>(index)) {
-        if(auto * indexScalarQuantizer = dynamic_cast<faiss::IndexScalarQuantizer *>(indexHNSWSQ->storage)) {
+    if (auto * indexHNSWSQ = dynamic_cast<faiss::IndexHNSWSQ *>(index)) {
+        if (auto * indexScalarQuantizer = dynamic_cast<faiss::IndexScalarQuantizer *>(indexHNSWSQ->storage)) {
             indexScalarQuantizer->codes.reserve(indexScalarQuantizer->code_size * numVectors);
         }
-        return;
     }
-    if(auto * indexHNSW = dynamic_cast<faiss::IndexHNSW *>(index)) {
+    if (auto * indexHNSW = dynamic_cast<faiss::IndexHNSW *>(index)) {
         if(auto * indexFlat = dynamic_cast<faiss::IndexFlat *>(indexHNSW->storage)) {
             indexFlat->codes.reserve(indexFlat->code_size * numVectors);
         }
-        return;
     }
 }
 
@@ -86,7 +82,7 @@ jlong IndexService::initIndex(
     std::unique_ptr<faiss::Index> index(faissMethods->indexFactory(dim, indexDescription.c_str(), metric));
 
     // Set thread count if it is passed in as a parameter. Setting this variable will only impact the current thread
-    if(threadCount != 0) {
+    if (threadCount != 0) {
         omp_set_num_threads(threadCount);
     }
 
@@ -94,7 +90,7 @@ jlong IndexService::initIndex(
     SetExtraParameters<faiss::Index, faiss::IndexIVF, faiss::IndexHNSW>(jniUtil, env, parameters, index.get());
 
     // Check that the index does not need to be trained
-    if(!index->is_trained) {
+    if (!index->is_trained) {
         throw std::runtime_error("Index is not trained");
     }
 
@@ -123,7 +119,7 @@ void IndexService::insertToIndex(
 
     // The number of vectors can be int here because a lucene segment number of total docs never crosses INT_MAX value
     int numVectors = (int) (inputVectors->size() / (uint64_t) dim);
-    if(numVectors == 0) {
+    if (numVectors == 0) {
         throw std::runtime_error("Number of vectors cannot be 0");
     }
 
@@ -132,7 +128,7 @@ void IndexService::insertToIndex(
     }
 
     // Set thread count if it is passed in as a parameter. Setting this variable will only impact the current thread
-    if(threadCount != 0) {
+    if (threadCount != 0) {
         omp_set_num_threads(threadCount);
     }
 
@@ -143,26 +139,30 @@ void IndexService::insertToIndex(
 }
 
 void IndexService::writeIndex(
-        std::string indexPath,
-        jlong idMapAddress
-    ) {
+    faiss::IOWriter* writer,
+    jlong idMapAddress
+) {
     std::unique_ptr<faiss::IndexIDMap> idMap (reinterpret_cast<faiss::IndexIDMap *> (idMapAddress));
 
     try {
         // Write the index to disk
-        faissMethods->writeIndex(idMap.get(), indexPath.c_str());
+        faissMethods->writeIndex(idMap.get(), writer);
+        if (auto openSearchIOWriter = dynamic_cast<knn_jni::stream::FaissOpenSearchIOWriter*>(writer)) {
+            openSearchIOWriter->flush();
+        }
     } catch(std::exception &e) {
         throw std::runtime_error("Failed to write index to disk");
     }
 }
 
-BinaryIndexService::BinaryIndexService(std::unique_ptr<FaissMethods> faissMethods) : IndexService(std::move(faissMethods)) {}
+BinaryIndexService::BinaryIndexService(std::unique_ptr<FaissMethods> _faissMethods)
+  : IndexService(std::move(_faissMethods)) {
+}
 
 void BinaryIndexService::allocIndex(faiss::Index * index, size_t dim, size_t numVectors) {
-    if(auto * indexBinaryHNSW = dynamic_cast<faiss::IndexBinaryHNSW *>(index)) {
+    if (auto * indexBinaryHNSW = dynamic_cast<faiss::IndexBinaryHNSW *>(index)) {
         auto * indexBinaryFlat = dynamic_cast<faiss::IndexBinaryFlat *>(indexBinaryHNSW->storage);
         indexBinaryFlat->xb.reserve(dim * numVectors / 8);
-        return;
     }
 }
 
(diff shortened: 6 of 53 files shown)


0ctopus13prime:
768 dimension 10m data testing result is coming shortly.

jni/include/native_engines_stream_support.h
```diff
@@ -119,6 +120,105 @@ class NativeEngineIndexInputMediator {
...
+                                  jobject _indexOutput)
+      : jni_interface(_jni_interface),
+        env(_env),
+        indexOutput(_indexOutput),
+        bufferArray((jbyteArray) (_jni_interface->GetObjectField(_env,
+                                                                 _indexOutput,
+                                                                 getBufferFieldId(_jni_interface, _env)))),
+        writeBytesMethod(getWriteBytesMethod(_jni_interface, _env)),
+        bufferLength(jni_interface->GetJavaBytesArrayLength(env, bufferArray)),
+        nextWriteIndex() {
```
jmazanec15, line 141: Should we explicitly initialize this?

jni/include/native_engines_stream_support.h
```diff
@@ -119,6 +120,105 @@ class NativeEngineIndexInputMediator {
...
+
+  void writeBytes(uint8_t *source, size_t nbytes) {
+    auto left = nbytes;
+    while (left > 0) {
+      const auto writeBytes = std::min(bufferLength - nextWriteIndex, left);
+
+      // === Critical Section Start ===
+
+      // Get primitive array pointer, no copy is happening in OpenJDK.
+      auto primitiveArray =
```
jmazanec15, line 152: This seems risky to me. From https://docs.oracle.com/javase/8/docs/technotes/guides/jni/spec/functions.html#GetPrimitiveArrayCritical_ReleasePrimitiveArrayCritical:
> If possible, the VM returns a pointer to the primitive array; otherwise, a copy is made.

What happens if this primitiveArray is a copy? OpenSearch can run potentially with JDKs other than OpenJDK...

Should, instead, we try to use `MemorySegment` somehow instead?

jni/src/faiss_index_service.cpp
```diff
@@ -86,15 +82,15 @@ jlong IndexService::initIndex(
...
+    if (threadCount != 0) {
         omp_set_num_threads(threadCount);
     }
 
     // Add extra parameters that cant be configured with the index factory
     SetExtraParameters<faiss::Index, faiss::IndexIVF, faiss::IndexHNSW>(jniUtil, env, parameters, index.get());
 
     // Check that the index does not need to be trained
-    if(!index->is_trained) {
+    if (!index->is_trained) [[unlikely]] {
```
jmazanec15, line 93: Will we see benefit of this? Also, does this require c++ >= 20?

jni/src/faiss_wrapper.cpp
```diff
@@ -170,39 +184,41 @@ void knn_jni::faiss_wrapper::InsertToIndex(knn_jni::JNIUtilInterface * jniUtil,
...
-    std::string indexPathCpp(jniUtil->ConvertJavaStringToCppString(env, indexPathJ));
+    // IndexOutput wrapper.
+    knn_jni::stream::NativeEngineIndexOutputMediator mediator {jniUtil, env, output};
+    knn_jni::stream::FaissOpenSearchIOWriter writer {&mediator};
 
-    // Create index
-    indexService->writeIndex(indexPathCpp, index_ptr);
+    // Create index.
+    indexService->writeIndex(&writer, index_ptr);
+    mediator.flush();
```
jmazanec15, line 199: Shouldnt mediator.flush() be called from writeIndex?

jni/src/nmslib_wrapper.cpp
```diff
@@ -317,25 +323,16 @@ jobjectArray knn_jni::nmslib_wrapper::QueryIndex(knn_jni::JNIUtilInterface *jniU
...
-    }
-    throw;
+  if (queryEfSearch == -1) {
+    query.reset(new similarity::KNNQuery<float>(*(indexWrapper->space), queryObject.get(), kJ));
+  } else {
+    query.reset(new similarity::HNSWQuery<float>(*(indexWrapper->space), queryObject.get(), kJ, queryEfSearch));
   }
-  delete query;
+
+  indexWrapper->index->Search(query.get());
```
jmazanec15, line 334: Do we need to touch search code in this change?

jmazanec15 (commented):
Finished first path - overall looks pretty good - have a few questions

jni/include/faiss_stream_support.h
```diff
@@ -56,6 +56,36 @@ class FaissOpenSearchIOReader final : public faiss::IOReader {
...
+    const auto writeBytes = size * nitems;
+    if (writeBytes > 0) {
+      mediator->writeBytes((uint8_t *) ptr, writeBytes);
+    }
+    return nitems;
+  }
+
+  // return a file number that can be memory-mapped
+  int filedescriptor() final {
+    throw std::runtime_error("filedescriptor() is not supported in FaissOpenSearchIOWriter.");
```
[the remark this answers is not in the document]
0ctopus13prime: Yes, it is reserved API for mmap use case in Faiss.

jni/include/jni_util.h
```diff
@@ -200,13 +204,30 @@ namespace knn_jni {
...
         void * GetPrimitiveArrayCritical(JNIEnv * env, jarray array, jboolean *isCopy) final;
         void ReleasePrimitiveArrayCritical(JNIEnv * env, jarray array, void *carray, jint mode) final;
 
     private:
         std::unordered_map<std::string, jclass> cachedClasses;
         std::unordered_map<std::string, jmethodID> cachedMethods;
-    };
+    };  // class JNIUtil
+
+    struct JNIReleaseElementsRelease {
```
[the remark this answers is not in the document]
0ctopus13prime: Sure! will do in the next rev.

jni/include/native_engines_stream_support.h
0ctopus13prime: I thought it would be better than otherwise!
Caching the values to avoid calling methods several times, and initialize 0 to `nextWriteIndex`.
The mediator will be allocated in stack area, having few bytes and inlining them should be better. Also it will give more readability!

jni/include/native_engines_stream_support.h
0ctopus13prime: sure, I understand your concern, but it's very likely that the most JDKs out there will return their internal pointer rather than returning a copied version. 
I think we can trade-off this to serve most frequent cases with just one copy. If we introduce an intermediate buffer like MemorySegment or native ByteBuffer, then we need two copies. bytes -> intermediate -> byte[] for IndexInput.

But I think I missed to pass `0` instead of `JNI_COMMIT` to ensure releasing the acquired pointer after receiving a copied one. (`0` will be ignored otherwise, so we can just pass `0`)
```
mode	actions
0	copy back the content and free the elems buffer
JNI_COMMIT	copy back the content but do not free the elems buffer
JNI_ABORT	free the buffer without copying back the possible changes
```

What do you think?

Solution 1 (Preferred).
Check whether the returned is copied, then pass `jint mode` accordingly.
Pros :
  - Will work most cases. Modern JDK will return a primitive pointer.
Cons:
  - In case JDK not supporting returning a primitive pointer, it will involve several memory allocations.


Solution 2
Have an intermediate buffer then convert it to `byte[]` for IndexOutput.
Pros : 
  - Will cover all the JDK implementations.
Cons :
  - Even when JDK supports a primitive array, we still need two copies.

jni/src/faiss_index_service.cpp
0ctopus13prime: It's C++11
Yeah, it's basically giving compiler hints so that it can reorder if branch.
I would say we won't lose anything by having it!

jni/src/faiss_wrapper.cpp
0ctopus13prime: right... that would be better!
will update in the next rev!

jni/src/nmslib_wrapper.cpp
0ctopus13prime: No!
I thought it would be a simple fix to have it in this PR.
If you're concern, then I can revert it back.

src/main/java/org/opensearch/knn/index/store/IndexOutputWithBuffer.java
```diff
@@ -0,0 +1,37 @@
...
+
+import org.apache.lucene.store.IndexOutput;
+
+import java.io.IOException;
+
+public class IndexOutputWithBuffer {
+    // Underlying `IndexOutput` obtained from Lucene's Directory.
+    private IndexOutput indexOutput;
+    // Write buffer. Native engine will copy bytes into this buffer.
+    private byte[] buffer = new byte[64 * 1024];
```
navneet1v, line 16: lets add a comment on why we are using this particular size as the buffer.

jni/include/native_engines_stream_support.h
```diff
@@ -119,6 +120,105 @@ class NativeEngineIndexInputMediator {
...
+
+  static jfieldID getBufferFieldId(JNIUtilInterface *jni_interface, JNIEnv *env) {
+    static jfieldID BUFFER_FIELD_ID =
+        jni_interface->GetFieldID(env, getIndexOutputWithBufferClass(jni_interface, env), "buffer", "[B");
+    return BUFFER_FIELD_ID;
+  }
+
+  void callWriteBytesInIndexOutput() {
+    auto jclazz = getIndexOutputWithBufferClass(jni_interface, env);
+    jvalue args {.i = nextWriteIndex};
```
navneet1v, line 203: what is this line doing? my question is mainly on the syntax.

jni/include/faiss_stream_support.h
```diff
@@ -56,6 +56,36 @@ class FaissOpenSearchIOReader final : public faiss::IOReader {
...
+ */
+class FaissOpenSearchIOWriter final : public faiss::IOWriter {
+ public:
+  explicit FaissOpenSearchIOWriter(NativeEngineIndexOutputMediator *_mediator)
+      : faiss::IOWriter(),
+        mediator(_mediator) {
+    name = "FaissOpenSearchIOWriter";
+  }
+
+  size_t operator()(const void *ptr, size_t size, size_t nitems) final {
```
navneet1v, line 71: please add docs on this function on what it is doing. mainly what is this *ptr?

jni/include/native_engines_stream_support.h
```diff
@@ -119,6 +120,105 @@ class NativeEngineIndexInputMediator {
...
+    while (left > 0) {
+      const auto writeBytes = std::min(bufferLength - nextWriteIndex, left);
+
+      // === Critical Section Start ===
+
+      // Get primitive array pointer, no copy is happening in OpenJDK.
+      auto primitiveArray =
+          (jbyte *) jni_interface->GetPrimitiveArrayCritical(env, bufferArray, nullptr);
+
+      // Copy Java bytes to C++ destination address.
```
navneet1v, line 155: I think this comment is incorrect. You are copying the c++ data(source) to java object(primitive array).
```
void* memcpy( void* dest, const void* src, std::size_t count );
```

src/main/java/org/opensearch/knn/index/store/IndexOutputWithBuffer.java
```diff
@@ -0,0 +1,37 @@
...
+    // This method will be called in JNI layer which precisely knows
+    // the amount of bytes need to be written.
+    public void writeBytes(int length) {
+        try {
+            // Delegate Lucene `indexOuptut` to write bytes.
+            indexOutput.writeBytes(buffer, 0, length);
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+    }
```
navneet1v, line 31: should we clean up the buffer once data is written?

jni/src/nmslib_wrapper.cpp
```diff
@@ -174,17 +172,25 @@ void knn_jni::nmslib_wrapper::CreateIndex(knn_jni::JNIUtilInterface *jniUtil, JN
...
-    index->SaveIndex(indexPathCpp);
 
-    for (auto &it : dataset) {
+    knn_jni::stream::NativeEngineIndexOutputMediator mediator {jniUtil, env, output};
+    knn_jni::stream::NmslibOpenSearchIOWriter writer {&mediator};
+
+    if (auto hnswFloatIndex = dynamic_cast<similarity::Hnsw<float> *>(index.get())) {
+      hnswFloatIndex->SaveIndexWithStream(writer);
+      mediator.flush();
+    } else {
```
navneet1v, line 182: why flush function cannot be combined with `SaveIndexWithStream` function?

jni/src/faiss_wrapper.cpp
navneet1v: +1 on Jack comment.

jni/src/faiss_wrapper.cpp
```diff
@@ -367,14 +390,14 @@ void knn_jni::faiss_wrapper::CreateByteIndexFromTemplate(knn_jni::JNIUtilInterfa
     jbyte * indexBytesJ = jniUtil->GetByteArrayElements(env, templateIndexJ, nullptr);
 
     faiss::VectorIOReader vectorIoReader;
+    vectorIoReader.data.reserve(indexBytesCount);
```
navneet1v, line 393: why we need this?

navneet1v (commented):
Please add java docs on all public functions and classes. both c++ and java.

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ## [Unreleased 2.x](https://github.com/opensearch-project/k-NN/compare/2.18...2.x)
 ### Features
 ### Enhancements
+- Introduced a writing layer (#2241)[https://github.com/opensearch-project/k-NN/pull/2241]
```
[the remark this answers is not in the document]
0ctopus13prime: will add more context into it!

src/main/java/org/opensearch/knn/index/store/IndexOutputWithBuffer.java
0ctopus13prime: sure, sounds good.

jni/include/native_engines_stream_support.h
0ctopus13prime: will add a comment in that line.
It's union initialization - https://en.cppreference.com/w/c/language/struct_initialization.
Since `writeBytes` has an integer arguments, I'm initializing jvalue as an integer, which `i` represents.

jni/include/faiss_stream_support.h
0ctopus13prime: sure, sounds good.

jni/include/native_engines_stream_support.h
0ctopus13prime: oh shoot!, yes, it's incorrect.
will update in the next rev.

jni/src/nmslib_wrapper.cpp
0ctopus13prime: There's no reason to do it!
I will update NMSLIB patch to do it inside in the end before returning.

jni/src/faiss_wrapper.cpp
0ctopus13prime: To avoid frequent malloc calls. It's an old trick.
Default implementation of vector uses doubling strategy when it reaches its max capacity, and given that we already know the size we are going to use, then we can just allocate the required memory without having to have it grew gradually up to the point.

jni/include/native_engines_stream_support.h
jmazanec15: I like solution 1 - to confirm, it will work for all jdks in general right?

Im wondering in future if we could leverage panama MemorySegment so that we can manage the buffer on the java side. Maybe for another review

jni/src/faiss_index_service.cpp
jmazanec15: Is there doc on this? Im seeing 20 here: https://en.cppreference.com/w/cpp/language/attributes/likely

jni/src/nmslib_wrapper.cpp
jmazanec15: Sure - Id say dont revert now. In future, especially for C++ layer changes though, I would try to stick to the scope.

0ctopus13prime: sure, for the next time, I will restrict PR to be within its own scope!

jni/include/native_engines_stream_support.h
0ctopus13prime: Hi Jack, 
Even with MemorySegment, it will need at least one copy.
Since it is native engine that uses IOWriter to write bytes, the given bytes from it locates at a different region in memory. So we still need to copy those bytes to the region that MemorySegment is pointing to. (Java does not provide a way for us to reset the internal pointer)

After copied bytes into MemorySegment, Java can directly access to via `get() -> byte` method, but given that the index size can grow giga byte scale, I think just copy bytes in bulk would be better than getting single byte then pass it.

0ctopus13prime: Also since it is bytes copying, there's no endianess restriction on it, it will be guaranteed that byte[] will be always identical to uint8_t[] what we are given.

Guarantee:
```
len(byte[]) == len(uint8_t[])
and for any i, byte[i] == uint8_t[i]
```

jni/src/faiss_index_service.cpp
0ctopus13prime: oh right. it's C++20. sorry for the confusion. :sweat_smile: 
will throw them away

jmazanec15:
rolling BWC is expected to fail due to pending version release - ignore.

@0ctopus13prime can you signoff the commit? DCO seems to be failing

jni/include/faiss_stream_support.h
```diff
@@ -56,6 +56,40 @@ class FaissOpenSearchIOReader final : public faiss::IOReader {
...
 
+/**
+ * A glue component inheriting IOWriter to delegate IO processing down to the given
+ * mediator. The mediator is expected to do write bytes via the provided Lucene's IndexOutput.
+ */
+class FaissOpenSearchIOWriter final : public faiss::IOWriter {
+ public:
+  explicit FaissOpenSearchIOWriter(NativeEngineIndexOutputMediator *_mediator)
+      : faiss::IOWriter(),
+        mediator(_mediator) {
```
Vikasht34, line 67: Can we do null check on mediator before proceeding.
```
if (!mediator) {
    throw std::invalid_argument("Mediator cannot be nullptr");
  }
```

jni/include/faiss_stream_support.h
```diff
@@ -56,6 +56,40 @@ class FaissOpenSearchIOReader final : public faiss::IOReader {
...
+    name = "FaissOpenSearchIOWriter";
+  }
+
+  size_t operator()(const void *ptr, size_t size, size_t nitems) final {
+    const auto writeBytes = size * nitems;
+    if (writeBytes > 0) {
+      mediator->writeBytes((uint8_t *) ptr, writeBytes);
```
Vikasht34, line 74: Any particular reason to not use static_cast instead of the C-style cast when converting the ptr to uint8_t *?
```
mediator->writeBytes(static_cast<const uint8_t *>(ptr), writeBytes);

```

jni/include/faiss_stream_support.h
```diff
@@ -56,6 +56,40 @@ class FaissOpenSearchIOReader final : public faiss::IOReader {
 };  // class FaissOpenSearchIOReader
 
 
+/**
+ * A glue component inheriting IOWriter to delegate IO processing down to the given
+ * mediator. The mediator is expected to do write bytes via the provided Lucene's IndexOutput.
+ */
+class FaissOpenSearchIOWriter final : public faiss::IOWriter {
+ public:
+  explicit FaissOpenSearchIOWriter(NativeEngineIndexOutputMediator *_mediator)
```
Vikasht34, line 65: If NativeEngineIndexOutputMediator is managed by this class (i.e., it should be deleted when FaissOpenSearchIOWriter is destroyed), Could we consider using a std::unique_ptr instead of a raw pointer to ensure proper cleanup?

jni/include/faiss_stream_support.h
```diff
@@ -56,6 +56,40 @@ class FaissOpenSearchIOReader final : public faiss::IOReader {
 };  // class FaissOpenSearchIOReader
 
 
+/**
+ * A glue component inheriting IOWriter to delegate IO processing down to the given
+ * mediator. The mediator is expected to do write bytes via the provided Lucene's IndexOutput.
+ */
+class FaissOpenSearchIOWriter final : public faiss::IOWriter {
```
Vikasht34, line 63: If FaissOpenSearchIOWriter ever needs to release resources or perform cleanup, could explicitly define a destructor?Even if the mediator is not owned, a comment can clarify this behavior.

jni/include/jni_util.h
```diff
@@ -200,13 +204,36 @@ namespace knn_jni {
...
+     * }};
+     */
+    struct JNIReleaseElements {
+      explicit JNIReleaseElements(std::function<void()> _release_func)
+          : release_func(std::move(_release_func)) {
+      }
+
+      ~JNIReleaseElements() {
+        try {
+          release_func();
```
Vikasht34, line 229: Can we ensure  release_func is a valid function before calling it to prevent unexpected behavior.

jni/include/jni_util.h
```diff
@@ -200,13 +204,36 @@ namespace knn_jni {
...
+     * Ex: JNIReleaseElements release_int_array_elements {[=](){
+     *   jniUtil->ReleaseIntArrayElements(env, idsJ, idsCpp, JNI_ABORT);
+     * }};
+     */
+    struct JNIReleaseElements {
+      explicit JNIReleaseElements(std::function<void()> _release_func)
+          : release_func(std::move(_release_func)) {
+      }
+
+      ~JNIReleaseElements() {
```
Vikasht34, line 227: The destructor should be marked noexcept to avoid any exceptions from propagating out, which is best practice for destructors.

jni/include/native_engines_stream_support.h
```diff
@@ -119,6 +121,110 @@ class NativeEngineIndexInputMediator {
 
 
 
+/**
+ * This class delegates the provided index output to do IO processing.
+ * In most cases, it is expected that IndexOutputWithBuffer was passed down to this,
+ * which eventually have Lucene's IndexOutput to write bytes.
+ */
+class NativeEngineIndexOutputMediator {
```
Vikasht34, line 129: Please add clearer comments to describe the purpose and functionality of each member and method. The initialization of jvalue args using the union syntax, for example, should have a clearer explanation.

jni/include/native_engines_stream_support.h
```diff
@@ -119,6 +121,110 @@ class NativeEngineIndexInputMediator {
...
+ public:
+  NativeEngineIndexOutputMediator(JNIUtilInterface *_jni_interface,
+                                  JNIEnv *_env,
+                                  jobject _indexOutput)
+      : jni_interface(_jni_interface),
+        env(_env),
+        indexOutput(_indexOutput),
+        bufferArray((jbyteArray) (_jni_interface->GetObjectField(_env,
```
Vikasht34, line 137: What is your thought of using like this static_cast<jbyteArray>

jni/include/native_engines_stream_support.h
```diff
@@ -119,6 +121,110 @@ class NativeEngineIndexInputMediator {
...
 
 
+/**
+ * This class delegates the provided index output to do IO processing.
+ * In most cases, it is expected that IndexOutputWithBuffer was passed down to this,
+ * which eventually have Lucene's IndexOutput to write bytes.
+ */
+class NativeEngineIndexOutputMediator {
+ public:
+  NativeEngineIndexOutputMediator(JNIUtilInterface *_jni_interface,
```
Vikasht34, line 131: Could we improve readability and safety by validating input parameters. Check for null pointers for jni_interface, env, and indexOutput to prevent potential issues.

jni/include/native_engines_stream_support.h
```diff
@@ -119,6 +121,110 @@ class NativeEngineIndexInputMediator {
...
+      std::memcpy(primitiveArray + nextWriteIndex, source, writeBytes);
+
+      // Release the acquired primitive array pointer.
+      // 0 tells JVM to copy back the content, and to free the pointer. It will be ignored if we acquired an internal
+      // primitive array pointer instead of a copied version.
+      // From JNI docs:
+      // Mode 0 : copy back the content and free the elems buffer
+      // The mode argument provides information on how the array buffer should be released. mode has no effect if elems
+      // is not a copy of the elements in array.
+      jni_interface->ReleasePrimitiveArrayCritical(env, bufferArray, primitiveArray, 0);
```
Vikasht34, line 166: Could we use  RAII (Resource Acquisition Is Initialization) to manage critical sections helps to ensure that resources are released even if an exception occurs. We can create a helper class to manage the JNI array critical section safely. any thought?

jni/include/native_engines_stream_support.h
```diff
@@ -119,6 +121,110 @@ class NativeEngineIndexInputMediator {
...
+        bufferLength(jni_interface->GetJavaBytesArrayLength(env, bufferArray)),
+        nextWriteIndex() {
+  }
+
+  void writeBytes(uint8_t *source, size_t nbytes) {
+    auto left = nbytes;
+    while (left > 0) {
+      const auto writeBytes = std::min(bufferLength - nextWriteIndex, left);
+
+      // === Critical Section Start ===
```
Vikasht34, line 150: Can we RAII for critical section part

jni/include/native_engines_stream_support.h
Vikasht34, line 129: If any member variables need to manage their own memory, can we consider using smart pointers instead of raw pointers.

jni/include/nmslib_stream_support.h
```diff
@@ -44,6 +44,27 @@ class NmslibOpenSearchIOReader final : public similarity::NmslibIOReader {
 };  // class NmslibOpenSearchIOReader
 
 
+class NmslibOpenSearchIOWriter final : public similarity::NmslibIOWriter {
+ public:
+  explicit NmslibOpenSearchIOWriter(NativeEngineIndexOutputMediator *_mediator)
+      : similarity::NmslibIOWriter(),
+        mediator(_mediator) {
```
Vikasht34, line 51: What if mediator is null here , how we are handling this?

0ctopus13prime: Mediator will be passed from wrapper, why it can be null?

jni/include/native_engines_stream_support.h
0ctopus13prime: No member variables are allocated in memory at the moment by design. 
we don't need to use smart pointers!
Personally, I prefer to avoid using smart pointers as possible

jni/include/native_engines_stream_support.h
0ctopus13prime: why it's needed?
in the critical section, no methods are throwing exceptions

jni/include/native_engines_stream_support.h
0ctopus13prime: No exceptions will be thrown by design!
all the methods in the critical sections are marked with `nothrow`.

jni/include/native_engines_stream_support.h
0ctopus13prime: em.. I can, but is it needed? :sweat_smile: 
I've added unit tests to cover this, and there will be a single path to pass pointer which is created in stack area.
It is guaranteed non-null for all parameters including JNIUtilInterface.
but, if you insist, then will update in the next rev!

Any thoughts?

jni/include/jni_util.h
0ctopus13prime: will update in the next rev

jni/include/faiss_stream_support.h
0ctopus13prime: `FaissOpenSearchIOWriter` does not have a resource to be released!
It just has a raw pointer

jni/include/faiss_stream_support.h
0ctopus13prime: We can avoid heap (I'm referring the heap in OS memory layout, not java) memory by creating the object in stack area.
I chose this way in purpose to avoid calling malloc, as glic + jmalloc has an internal locking mechanism which probably have an impact on performance.

jni/include/faiss_stream_support.h
0ctopus13prime: will update in next rev.

jni/include/faiss_stream_support.h
0ctopus13prime: Mediator will be passed from wrapper, and it is not null.
but yeah, will update in the next rev

jni/include/native_engines_stream_support.h
Vikasht34: Make sense , if there is no expection thrown we indeed don't need RAII , but can gurantee that this code will not be extended in future and no expection ever be thrown ..I was just trying to be cautious here !!!

jni/include/native_engines_stream_support.h
Vikasht34: Make sense , if there is no expection thrown we indeed don't need RAII , but can gurantee that this code will not be extended in future and no expection ever be thrown ..I was just trying to be cautious here !!!

jni/include/nmslib_stream_support.h
Vikasht34: I see and are we making sure that does wraper always gurantee that it won't pass null

jni/include/faiss_stream_support.h
Vikasht34: Make Sense

jni/include/faiss_stream_support.h
Vikasht34: Sure

jni/include/native_engines_stream_support.h
Vikasht34: Will leave up to you !!

jni/include/native_engines_stream_support.h
Vikasht34: If this becuase of Heap vs Stack and due to performance hit ..I am ok !!

jni/include/jni_util.h
0ctopus13prime: I don't think that is the best practice.
In c++, destructors are implicitly `noexcept(true)`. actually, even if we throw an exception there, the mark will still be `noexcept(true)`.

jni/include/native_engines_stream_support.h
0ctopus13prime: hmm.. yeah, it's always better to have a clearer explanation attach to it...
but not sure what is missing.
could you pinpoint what is missing form the comment?? I think it provide a general view of the class

jni/include/native_engines_stream_support.h
0ctopus13prime: yeah, I can do that, but it should be fine with C99 casting right for this case?
it's tooo long haha

jni/include/jni_util.h
Vikasht34: I see from c++11 they made destructors implicit with noexcept(true), ..I was not aware of that !! we don't need this then .

jni/include/native_engines_stream_support.h
Vikasht34: Yeah one instance was I was not aware of  jvalue args using the union syntax, if we could see something helps understand on special syntax , would be better

0ctopus13prime:
@Vikasht34 
Hi vikash, I added nullptr check for each parameters. Please take a look at when you're free.

```
  explicit FaissOpenSearchIOReader(NativeEngineIndexInputMediator *_mediator)
      : faiss::IOReader(),
        mediator(_mediator) {
        mediator(knn_jni::util::ParameterCheck::require_non_null(_mediator, "mediator")) {
    name = "FaissOpenSearchIOReader";
  }



struct ParameterCheck {
  template<typename PtrType>
  static PtrType *require_non_null(PtrType *ptr, const char *parameter_name) {
    if (ptr == nullptr) {
      throw std::invalid_argument(std::string("Parameter [") + parameter_name + "] should not be null.");
    }
    return ptr;
  }
```

jmazanec15 (approved):

Vikasht34 (approved):

jni/src/faiss_wrapper.cpp
navneet1v: I understand the part of `vector uses doubling strategy when it reaches its max capacity`, may I was not clear on the question, so it was mainly trying to ask here is where is this `vectorIoReader.data` being used basically what it is for? 

So what I can see from here is, we are creating an indexWithTemplate(so indices like IndexIVF and PQ based indices is being created here). We are first putting all the data in the temporary buffer of `vectorIoReader` and then creating the index. I also see that we are doing the same thing where we are ingesting the same vectors via `faiss::IndexIDMap idMap`. This does seems weird to me. 

@jmazanec15  any thoughts here?

0ctopus13prime:
## Benchmark Environment
- OpenSearch Version : 2.18 (backporting this change to 2.xx branch)
- Primary shards : 3
- No replica
- Data node : r7gd.4xlarge (16 vCPU, 128 memory)
- Storage : SSD, EBS, 5000 IOPS
- JVM Heap : -Xmx32g -Xms32g

0ctopus13prime:
@navneet1v 
Will paste this to the RFC issue as well!

## Faiss benchmark results conclusion

From the results, it's expected that total cumulative indexing time will be increased up to 2% (67.67min -> 68.98min), thus it means that the bulk indexing throughput can be decreased down to 1.4% (7861 -> 7744). 


## Faiss benchmark details
<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:////Users/kdooyong/Library/Group%20Containers/UBF8T346G9.Office/TemporaryItems/msohtmlclip/clip.htm">
<link rel=File-List
href="file:////Users/kdooyong/Library/Group%20Containers/UBF8T346G9.Office/TemporaryItems/msohtmlclip/clip_filelist.xml">
<style>

</style>
</head>

<body link="#0563C1" vlink="#954F72">


Metric | Task | Baseline | Candidate | Diff | Unit
-- | -- | -- | -- | -- | --
Cumulative indexing time   of primary shards |   | 67.6766 | 68.9868 | 0.01935972 | min
Min cumulative indexing time   across primary shards |   | 0 | 0 | 0 | min
Median cumulative indexing time across primary shards |   | 10.92455 | 11.3572 | 0.039603462 | min
Max cumulative indexing time   across primary shards |   | 22.9782 | 23.4515 | 0.020597784 | min
Cumulative indexing throttle time   of primary shards |   | 0 | 0 | 0 | min
Min cumulative indexing throttle time across primary shards |   | 0 | 0 | 0 | min
Median cumulative indexing throttle time   across primary shards |   | 0 | 0 | 0 | min
Max cumulative indexing throttle time across primary shards |   | 0 | 0 | 0 | min
Cumulative merge time   of primary shards |   | 455.27 | 463.033 | 0.01705142 | min
Cumulative merge count   of primary shards |   | 628 | 631 | 0.00477707 |  
Min cumulative merge time   across primary shards |   | 0 | 0 | 0 | min
Median cumulative merge time   across primary shards |   | 74.92695 | 75.6642 | 0.009839584 | min
Max cumulative merge time   across primary shards |   | 154.315 | 158.499 | 0.027113372 | min
Cumulative merge throttle time   of primary shards |   | 25.78985 | 22.7165 | -0.119168975 | min
Min cumulative merge throttle time across primary shards |   | 0 | 0 | 0 | min
Median cumulative merge throttle time across primary shards |   | 3.979025 | 3.18752 | -0.198919333 | min
Max cumulative merge throttle time across primary shards |   | 9.53849 | 8.29422 | -0.130447272 | min
Cumulative refresh time   of primary shards |   | 0.8357415 | 0.8457 | 0.011915766 | min
Cumulative refresh count   of primary shards |   | 475.5 | 473 | -0.005257624 |  
Min cumulative refresh time   across primary shards |   | 0 | 0 | 0 | min
Median cumulative refresh time across primary shards |   | 0.134925 | 0.132867 | -0.015252918 | min
Max cumulative refresh time   across primary shards |   | 0.2874665 | 0.302833 | 0.053454924 | min
Cumulative flush time   of primary shards |   | 15.55735 | 15.5443 | -0.000838832 | min
Cumulative flush count   of primary shards |   | 350 | 349 | -0.002857143 |  
Min cumulative flush time   across primary shards |   | 0 | 0 | 0 | min
Median cumulative flush time   across primary shards |   | 2.5186 | 2.5161 | -0.000992615 | min
Max cumulative flush time   across primary shards |   | 5.39728 | 5.36828 | -0.005373077 | min
Total   Young Gen GC time |   | 1.469 | 1.706 | 0.161334241 | s
Total   Young Gen GC count |   | 160 | 174 | 0.0875 |  
Total Old Gen GC time |   | 0 | 0 | 0 | s
Total Old Gen GC count |   | 0 | 0 | 0 |  
Store size |   | 340.746 | 340.744 | -5.86947E-06 | GB
Translog size |   | 1.43983E-06 | 1.20E-06 | -0.168176688 | GB
Heap   used for segments |   | 0 | 0 | 0 | MB
Heap   used for doc values |   | 0 | 0 | 0 | MB
Heap used for terms |   | 0 | 0 | 0 | MB
Heap used for norms |   | 0 | 0 | 0 | MB
Heap used for points |   | 0 | 0 | 0 | MB
Heap used   for stored fields |   | 0 | 0 | 0 | MB
Segment count |   | 5.5 | 5 | -0.090909091 |  
Min Throughput | custom-vector-bulk | 6584.245 | 5835.71 | -0.113685776 | docs/s
Mean Throughput | custom-vector-bulk | 7861.24 | 7744.98 | -0.014789015 | docs/s
Median Throughput | custom-vector-bulk | 7596.51 | 7518.42 | -0.010279721 | docs/s
Max Throughput | custom-vector-bulk | 10233.015 | 9612.79 | -0.060610192 | docs/s
50th   percentile latency | custom-vector-bulk | 72.44355 | 73.3195 | 0.012091484 | ms
90th   percentile latency | custom-vector-bulk | 151.5965 | 155.167 | 0.023552655 | ms
99th   percentile latency | custom-vector-bulk | 260.645 | 257.376 | -0.012541963 | ms
99.9th   percentile latency | custom-vector-bulk | 378.015 | 381.875 | 0.010211235 | ms
99.99th   percentile latency | custom-vector-bulk | 608.632 | 932.898 | 0.532778428 | ms
100th   percentile latency | custom-vector-bulk | 1080.8535 | 3556.3 | 2.290270143 | ms
50th   percentile service time | custom-vector-bulk | 72.44355 | 73.3195 | 0.012091484 | ms
90th   percentile service time | custom-vector-bulk | 151.5965 | 155.167 | 0.023552655 | ms
99th   percentile service time | custom-vector-bulk | 260.645 | 257.376 | -0.012541963 | ms
99.9th   percentile service time | custom-vector-bulk | 378.015 | 381.875 | 0.010211235 | ms
99.99th   percentile service time | custom-vector-bulk | 608.632 | 932.898 | 0.532778428 | ms
100th   percentile service time | custom-vector-bulk | 1080.8535 | 3556.3 | 2.290270143 | ms
error rate | custom-vector-bulk | 0 | 0 | 0 | %
Min Throughput | force-merge-segments | 0 | 0 | 0 | ops/s
Mean Throughput | force-merge-segments | 0 | 0 | 0 | ops/s
Median Throughput | force-merge-segments | 0 | 0 | 0 | ops/s
Max Throughput | force-merge-segments | 0 | 0 | 0 | ops/s
100th   percentile latency | force-merge-segments | 11015600 | 1.13E+07 | 0.026789281 | ms
100th   percentile service time | force-merge-segments | 11015600 | 1.13E+07 | 0.026789281 | ms
error rate | force-merge-segments | 0 | 0 | 0 | %
Min Throughput | warmup-indices | 0.02 | 0.02 | 0 | ops/s
Mean Throughput | warmup-indices | 0.02 | 0.02 | 0 | ops/s
Median Throughput | warmup-indices | 0.02 | 0.02 | 0 | ops/s
Max Throughput | warmup-indices | 0.02 | 0.02 | 0 | ops/s
100th   percentile latency | warmup-indices | 46760.8 | 54720 | 0.170210946 | ms
100th   percentile service time | warmup-indices | 46760.8 | 54720 | 0.170210946 | ms
error rate | warmup-indices | 0 | 0 | 0 | %
Min Throughput | prod-queries | 2.08 | 1.61 | -0.225961538 | ops/s
Mean Throughput | prod-queries | 8.765 | 10.09 | 0.151169424 | ops/s
Median Throughput | prod-queries | 8.765 | 4.43 | -0.494580719 | ops/s
Max Throughput | prod-queries | 15.445 | 24.24 | 0.569439948 | ops/s
50th   percentile latency | prod-queries | 9.361765 | 10.2061 | 0.090189724 | ms
90th   percentile latency | prod-queries | 11.2586 | 12.3034 | 0.092800171 | ms
99th   percentile latency | prod-queries | 482.748 | 557.695 | 0.155250773 | ms
100th   percentile latency | prod-queries | 503.6475 | 621.016 | 0.233036995 | ms
50th   percentile service time | prod-queries | 9.361765 | 10.2061 | 0.090189724 | ms
90th   percentile service time | prod-queries | 11.2586 | 12.3034 | 0.092800171 | ms
99th   percentile service time | prod-queries | 482.748 | 557.695 | 0.155250773 | ms
100th   percentile service time | prod-queries | 503.6475 | 621.016 | 0.233036995 | ms
error rate | prod-queries | 0 | 0 | 0 | %
Mean recall@k | prod-queries | 0.34 | 0.35 | 0.029411765 |  
Mean recall@1 | prod-queries | 0.495 | 0.4 | -0.191919192 |  



</body>

</html>

0ctopus13prime:
## NMSLIB benchmark conclusion
Unlike Faiss, it is expected there will be a slight improvement in indexing related metrics in NMSLIB.
Cumulative indexing time decreased 5.4% (from 70.26min -> 66.5min), mean bulk indexing throughput has been increased 1.5% (from 7797 docs/sec to 7921 docs/sec)

## Benchmark details
<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:////Users/kdooyong/Library/Group%20Containers/UBF8T346G9.Office/TemporaryItems/msohtmlclip/clip.htm">
<link rel=File-List
href="file:////Users/kdooyong/Library/Group%20Containers/UBF8T346G9.Office/TemporaryItems/msohtmlclip/clip_filelist.xml">
<style>

</style>
</head>

<body link="#0563C1" vlink="#954F72">


Metric | Task | Baseline-Value | Candidate-Value | Diff | Unit
-- | -- | -- | -- | -- | --
Cumulative indexing time   of primary shards |   | 70.2665 | 66.4061 | -0.054939409 | min
Min cumulative indexing time   across primary shards |   | 0 | 0 | 0 | min
Median cumulative indexing time across primary shards |   | 11.1663 | 10.4298 | -0.06595739 | min
Max cumulative indexing time   across primary shards |   | 24.5895 | 23.3949 | -0.048581712 | min
Cumulative indexing throttle time   of primary shards |   | 0 | 0 | 0 | min
Min cumulative indexing throttle time across primary shards |   | 0 | 0 | 0 | min
Median cumulative indexing throttle time   across primary shards |   | 0 | 0 | 0 | min
Max cumulative indexing throttle time across primary shards |   | 0 | 0 | 0 | min
Cumulative merge time   of primary shards |   | 547.03 | 546.824 | -0.000376579 | min
Cumulative merge count   of primary shards |   | 650 | 617 | -0.050769231 |  
Min cumulative merge time   across primary shards |   | 0 | 0 | 0 | min
Median cumulative merge time   across primary shards |   | 88.9109 | 89.2281 | 0.003567617 | min
Max cumulative merge time   across primary shards |   | 188.095 | 186.289 | -0.009601531 | min
Cumulative merge throttle time   of primary shards |   | 18.9961 | 20.5987 | 0.084364685 | min
Min cumulative merge throttle time across primary shards |   | 0 | 0 | 0 | min
Median cumulative merge throttle time across primary shards |   | 2.98061 | 3.06296 | 0.027628573 | min
Max cumulative merge throttle time across primary shards |   | 6.81933 | 7.71148 | 0.130826635 | min
Cumulative refresh time   of primary shards |   | 0.785367 | 0.735133 | -0.063962453 | min
Cumulative refresh count   of primary shards |   | 484 | 477 | -0.01446281 |  
Min cumulative refresh time   across primary shards |   | 0 | 0 | 0 | min
Median cumulative refresh time across primary shards |   | 0.127833 | 0.110533 | -0.135332817 | min
Max cumulative refresh time   across primary shards |   | 0.273 | 0.276367 | 0.012333333 | min
Cumulative flush time   of primary shards |   | 15.2446 | 14.0613 | -0.077620928 | min
Cumulative flush count   of primary shards |   | 363 | 355 | -0.022038567 |  
Min cumulative flush time   across primary shards |   | 0 | 0 | 0 | min
Median cumulative flush time   across primary shards |   | 2.37743 | 2.25187 | -0.052813332 | min
Max cumulative flush time   across primary shards |   | 5.28568 | 4.94943 | -0.063615278 | min
Total   Young Gen GC time |   | 2.127 | 2.052 | -0.035260931 | s
Total   Young Gen GC count |   | 179 | 176 | -0.016759777 |  
Total Old Gen GC time |   | 0 | 0 | 0 | s
Total Old Gen GC count |   | 0 | 0 | 0 |  
Store size |   | 340.825 | 340.825 | 0 | GB
Translog size |   | 1.20E-06 | 1.20E-06 | 0 | GB
Heap   used for segments |   | 0 | 0 | 0 | MB
Heap   used for doc values |   | 0 | 0 | 0 | MB
Heap used for terms |   | 0 | 0 | 0 | MB
Heap used for norms |   | 0 | 0 | 0 | MB
Heap used for points |   | 0 | 0 | 0 | MB
Heap used   for stored fields |   | 0 | 0 | 0 | MB
Segment count |   | 5 | 5 | 0 |  
Min Throughput | custom-vector-bulk | 4198.96 | 5499.58 | 0.309748128 | docs/s
Mean Throughput | custom-vector-bulk | 7797.42 | 7921.59 | 0.015924498 | docs/s
Median Throughput | custom-vector-bulk | 7560.44 | 7625.18 | 0.008562994 | docs/s
Max Throughput | custom-vector-bulk | 9657.7 | 9963.47 | 0.031660747 | docs/s
50th   percentile latency | custom-vector-bulk | 68.3138 | 68.0887 | -0.003295088 | ms
90th   percentile latency | custom-vector-bulk | 154.12 | 153.953 | -0.001083571 | ms
99th   percentile latency | custom-vector-bulk | 258.562 | 258.702 | 0.000541456 | ms
99.9th   percentile latency | custom-vector-bulk | 370.548 | 387.22 | 0.044992821 | ms
99.99th   percentile latency | custom-vector-bulk | 506.498 | 582.752 | 0.150551434 | ms
100th   percentile latency | custom-vector-bulk | 2082.6 | 1156.62 | -0.444626909 | ms
50th   percentile service time | custom-vector-bulk | 68.3138 | 68.0887 | -0.003295088 | ms
90th   percentile service time | custom-vector-bulk | 154.12 | 153.953 | -0.001083571 | ms
99th   percentile service time | custom-vector-bulk | 258.562 | 258.702 | 0.000541456 | ms
99.9th   percentile service time | custom-vector-bulk | 370.548 | 387.22 | 0.044992821 | ms
99.99th   percentile service time | custom-vector-bulk | 506.498 | 582.752 | 0.150551434 | ms
100th   percentile service time | custom-vector-bulk | 2082.6 | 1156.62 | -0.444626909 | ms
error rate | custom-vector-bulk | 0 | 0 | 0 | %
Min Throughput | force-merge-segments | 0 | 0 | 0 | ops/s
Mean Throughput | force-merge-segments | 0 | 0 | 0 | ops/s
Median Throughput | force-merge-segments | 0 | 0 | 0 | ops/s
Max Throughput | force-merge-segments | 0 | 0 | 0 | ops/s
100th   percentile latency | force-merge-segments | 1.50E+07 | 1.47E+07 | -0.024653036 | ms
100th   percentile service time | force-merge-segments | 1.50E+07 | 1.47E+07 | -0.024653036 | ms
error rate | force-merge-segments | 0 | 0 | 0 | %
Min Throughput | warmup-indices | 0.02 | 0.02 | 0 | ops/s
Mean Throughput | warmup-indices | 0.02 | 0.02 | 0 | ops/s
Median Throughput | warmup-indices | 0.02 | 0.02 | 0 | ops/s
Max Throughput | warmup-indices | 0.02 | 0.02 | 0 | ops/s
100th   percentile latency | warmup-indices | 47770.6 | 48938.4 | 0.024445998 | ms
100th   percentile service time | warmup-indices | 47770.6 | 48938.4 | 0.024445998 | ms
error rate | warmup-indices | 0 | 0 | 0 | %
Min Throughput | prod-queries | 1.68 | 1.75 | 0.041666667 | ops/s
Mean Throughput | prod-queries | 7.31 | 10.71 | 0.465116279 | ops/s
Median Throughput | prod-queries | 7.31 | 2.39 | -0.673050616 | ops/s
Max Throughput | prod-queries | 12.93 | 27.99 | 1.164733179 | ops/s
50th   percentile latency | prod-queries | 10.2375 | 9.81192 | -0.041570696 | ms
90th   percentile latency | prod-queries | 12.6484 | 12.481 | -0.013234876 | ms
99th   percentile latency | prod-queries | 491.371 | 515.637 | 0.049384274 | ms
100th   percentile latency | prod-queries | 592.629 | 569.577 | -0.03889786 | ms
50th   percentile service time | prod-queries | 10.2375 | 9.81192 | -0.041570696 | ms
90th   percentile service time | prod-queries | 12.6484 | 12.481 | -0.013234876 | ms
99th   percentile service time | prod-queries | 491.371 | 515.637 | 0.049384274 | ms
100th   percentile service time | prod-queries | 592.629 | 569.577 | -0.03889786 | ms
error rate | prod-queries | 0 | 0 | 0 | %
Mean recall@k | prod-queries | 0.51 | 0.53 | 0.039215686 |  
Mean recall@1 | prod-queries | 0.7 | 0.79 | 0.128571429 |  



</body>

</html>

0ctopus13prime:
@navneet1v 
Hi Navneet!
Could you review the benchmark results? You can find the benchmark environment in the above comment.
Overall, 2% increase in indexing time, -1.5% bulk indexing throughput is expected for FAISS, and the metrics will be the same for NMSLIB.

Thank you!

jni/src/faiss_wrapper.cpp
navneet1v: Thanks @jmazanec15 and @0ctopus13prime for the details. I got my answer. Didn't want to hold off on this PR. Approving the PR.

navneet1v (approved):

Issue #2215, opened 2024-10-16
[Enhancement] Reducing double vector reading and distance calculation for Disk based Filtered Vector Search 


navneet1v:
## Oppurtunity
Currently in plugin, whenever we are doing disk based vector search we first do a ANN search on quantized index and then do the rescoring using full precision vectors. ref: https://github.com/opensearch-project/k-NN/blob/7cf45c8cfd4219912d67becb731f8a7af61410c0/src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java#L64-L75

But this flow of first doing ANN search is not always true when efficient filters are present. For efficient filters we first see if exact search can be done or not. If exact search can be done then we do exact search rather than ANN search. ref: https://github.com/opensearch-project/k-NN/blob/7cf45c8cfd4219912d67becb731f8a7af61410c0/src/main/java/org/opensearch/knn/index/query/KNNWeight.java#L132-L156

For disk based index, this exact search during filters first fetches the full precision vectors from disk, do binary quantization on them. ref: https://github.com/opensearch-project/k-NN/blob/7cf45c8cfd4219912d67becb731f8a7af61410c0/src/main/java/org/opensearch/knn/index/query/iterators/VectorIdsKNNIterator.java#L91-L101 The same vectors is also read when we do rescoring. Hence I think we can actually remove this double reading of vectors and during the first pass itself we should calculate the correct scores since we already have vectors with us.

We would have to some changes in [reduceToTopK](https://github.com/opensearch-project/k-NN/blob/7cf45c8cfd4219912d67becb731f8a7af61410c0/src/main/java/org/opensearch/knn/index/query/ResultUtil.java#L33) function as with the above code scores of documents from different segments are not using same space_type. But I think that is still an easy problem to solve where we can just avoid the docs from the segments where hamming distance is not used. 

We would need to do some benchmarks to see the performance differences. But I think in case of constraint envs, this can really help in reducing disk reads, and not necessary the distance computations.

frejonb:
Maybe related to this. We ran some latency benchmarks on the different compression levels on 2.18:

![image (12)](https://github.com/user-attachments/assets/4227aa9d-5a8e-4e0c-9608-e10801cd49e4)

Upon inspecting the segments, we saw multiple ones with less than 15k documents. Upon setting `index.knn.advanced.approximate_threshold=0` for >=8x compression we recover the expected latencies:


[...]

PR #2271, opened 2024-11-14
Refactor scoring to map leaf reader context with results

VijayanB:
### Description
We map order of results to order of segments, and finally rely on that order to build top docs. Refactor method to use map.Entry to map leaf reader context with results from those leaves.
This is required when we want to split segments based on approx search or exact search to reduce rescoring twice by exact search

### Related Issues
Prerequisite for  #2215 
### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java
+++ b/src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java
@@ -28,6 +28,7 @@ import org.opensearch.knn.index.query.ResultUtil;
 import org.opensearch.knn.index.query.rescore.RescoreContext;
 
 import java.io.IOException;
+import java.util.AbstractMap;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Comparator;
@@ -55,7 +56,7 @@ public class NativeEngineKnnVectorQuery extends Query {
         final IndexReader reader = indexSearcher.getIndexReader();
         final KNNWeight knnWeight = (KNNWeight) knnQuery.createWeight(indexSearcher, scoreMode, 1);
         List<LeafReaderContext> leafReaderContexts = reader.leaves();
-        List<Map<Integer, Float>> perLeafResults;
+        List<Map.Entry<LeafReaderContext, Map<Integer, Float>>> perLeafResults;
         RescoreContext rescoreContext = knnQuery.getRescoreContext();
         final int finalK = knnQuery.getK();
         if (rescoreContext == null) {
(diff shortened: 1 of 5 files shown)


heemin32:
Do you have a follow up code showing why this is needed? Can we always get the relation between leaf reader and the result based on their index in the list?

VijayanB:
> Do you have a follow up code showing why this is needed? Can't we always get the relation between leaf reader and the result based on their index in the list?

Sure. https://github.com/opensearch-project/k-NN/pull/2273 . This is draft PR.  I want to test recall before making it ready.

heemin32:
Thanks. I think this refactoring should be accompanied with the final PR instead of pushing it separately.

heemin32:
The challenge is that it's difficult to justify this refactoring without the actual PR that requires it. There's always a possibility that the implementation in the subsequent PR may change, in which case this refactoring might end up providing no real benefit.

VijayanB:
> The challenge is that it's difficult to justify this refactoring without the actual PR that requires it. There's always a possibility that the implementation in the subsequent PR may change, in which case this refactoring might end up providing no real benefit.

Fair enough. How about this? I will backport to 2.x only if both PRs are merged. Will that work?

heemin32:
I am against merging it in main. Doesn't matter we backport it to 2.x or not. The change is not big so I think it should be okay to have this with following PR together.

VijayanB:
> I am against merging it in main. Doesn't matter we backport it to 2.x or not. The change is not big so I think it should be okay to have this with following PR together.

I don't see a reason why it is absolute necessary that this feature cannot be broken into two PR with predefined scope, where each does two different thing. Can you give reason why you are **against** breaking into two PRs by pointing out how it breaks either build or existing feature?

heemin32:
> > I am against merging it in main. Doesn't matter we backport it to 2.x or not. The change is not big so I think it should be okay to have this with following PR together.
> 
> I don't see a reason why it is absolute necessary that this feature cannot be broken into two PR with predefined scope, where each does two different thing. Can you give reason why you are **against** breaking into two PRs by pointing out how it breaks either build or existing feature?

Thank you for your effort in keeping the PR small to make the review process easier—I really appreciate that. My concern is that the value of this PR depends heavily on the approval of the subsequent PR, as it primarily supports those changes. If the following PR undergoes modifications, this one may also need to be adjusted accordingly. By the way, I didn't say it will breaks either build or existing feature.

src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java
```diff
@@ -87,32 +89,29 @@ public Weight createWeight(IndexSearcher indexSearcher, ScoreMode scoreMode, flo
...
-        List<Callable<Map<Integer, Float>>> tasks = new ArrayList<>(leafReaderContexts.size());
+        List<Callable<Map.Entry<LeafReaderContext, Map<Integer, Float>>>> tasks = new ArrayList<>(leafReaderContexts.size());
         for (LeafReaderContext leafReaderContext : leafReaderContexts) {
             tasks.add(() -> searchLeaf(leafReaderContext, knnWeight, k));
         }
         return indexSearcher.getTaskExecutor().invokeAll(tasks);
     }
 
-    private List<Map<Integer, Float>> doRescore(
+    private List<Map.Entry<LeafReaderContext, Map<Integer, Float>>> doRescore(
```
shatejas, line 105: This data structure is getting complex, can we look into simplifying it in favor of readability?

Issue #2266, opened 2024-11-12
[RFC]Reduce Disk Usage By Reusing NativeEngine Files


luyuncheng:
### Description

before ISSUE #1572 AND MR #1571, we found that we reuse docValues field like `KNNVectorFieldData` and do synthetic logic in _source field. and it can save about 1/3 disk usage.  Also @jmazanec15 mentioned a great method to implements storeFields as  https://github.com/opensearch-project/k-NN/pull/1571#issuecomment-2438764949 says. 

**this RFC,**  i proposal a new method to reduce disk usage, that we can read nativeEngines files to create DocValues. so we can save the disk for skip write `flatFieldVectorsWriter` or `BinaryDocValues` 

i read the Faiss code: `faiss/impl/index_write.cpp` it show that faiss HNSW32,Flat file structure like followings:

```

|-typeIDMap   -|-id_header-|
|-typeHnsw    -|-hnsw_header-|-hnswGraph-|
|-typeStorage -|-storage_Header-|-storageVector-|
|-idmap_vector-|-FOOTER_MAGIC+CHECKSUM-|

```

i implements a `FaissEngineFlatVectorValues` which read `_0_2011_target_field.faissc` files directly and wrap a `DocIdSetIterator` instead of using `FlatVectorsReader`. at POC code , it shows that, **we can cut almost 50% disk usage for skip write flatVectors** also without write flatVectors, write performance do a little optimize

in the next: 

-  at 1st step i would support all faiss format as reader, also write flatVectors.  #2267
-  at 2nd step, i see #2175 we would not create xxx.faiss file, we also need flatVectors when skipBuildingVector. i think we can do create either nativeFile or flatVectors, not create both.

navneet1v:
> at POC code , it shows that, **we can cut almost 50% disk usage for skip write flatVectors** also without write flatVectors, write performance do a little optimize

This is an interesting gain. I am wondering when you say 50% gain in disk space it will happen only in case when source is not enabled for vectors. Cutting down the flat vectors and just reading it via Faiss index has been discussed couple of times. My only concern with this is will reading flat vectors from Faiss file be as efficient as reading the flat vectors from .vec file? 

Also, did we explore this option where we don't store/serialize flat vectors in Faiss and use the .vec file instead. It can also help this feature: https://github.com/opensearch-project/k-NN/issues/1693

@0ctopus13prime , @jmazanec15

luyuncheng:
[...]

PR #2273, opened 2024-11-14
Split segment by search type

VijayanB:
### Description
For exact search, it is not required to perform
qunatization during rescore with oversamples.
However, to avoid normalization between segments from
approx search and exact search, we will first identify
segments that needs approxsearch and will perform oversamples
and, at end, after rescore, we will add scores from segments that
will perform exact search.

### Related Issues
#2215 

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/query/KNNWeight.java
+++ b/src/main/java/org/opensearch/knn/index/query/KNNWeight.java
@@ -153,6 +153,26 @@ public class KNNWeight extends Weight {
         return docIdsToScoreMap;
     }
 
+    /**
+     * For given {@link LeafReaderContext}, this api will return will KNNWeight perform exact search or not
+     * always. This decision is based on two properties, 1) if there are no native engine files in segments,
+     * exact search will always be performed, 2) if number of docs after filter is less than 'k'
+     * @param context
+     * @return
+     * @throws IOException
+     */
+    public boolean isExactSearchPreferred(LeafReaderContext context) throws IOException {
+        final BitSet filterBitSet = getFilteredDocsBitSet(context);
+        int cardinality = filterBitSet.cardinality();
+        if (isFilteredExactSearchPreferred(cardinality)) {
+            return true;
+        }
+        if (isMissingNativeEngineFiles(context)) {
+            return true;
+        }
+        return false;
+    }
+
     private BitSet getFilteredDocsBitSet(final LeafReaderContext ctx) throws IOException {
         if (this.filterWeight == null) {
             return new FixedBitSet(0);
--- a/src/main/java/org/opensearch/knn/index/query/ResultUtil.java
+++ b/src/main/java/org/opensearch/knn/index/query/ResultUtil.java
@@ -5,6 +5,7 @@
 
 package org.opensearch.knn.index.query;
 
+import org.apache.lucene.index.LeafReaderContext;
 import org.apache.lucene.search.DocIdSetIterator;
 import org.apache.lucene.search.ScoreDoc;
 import org.apache.lucene.search.TopDocs;
@@ -30,14 +31,15 @@ public final class ResultUtil {
      * @param perLeafResults Results from the list
      * @param k the number of results across all leaf results to return
      */
-    public static void reduceToTopK(List<Map<Integer, Float>> perLeafResults, int k) {
+    public static void reduceToTopK(List<Map.Entry<LeafReaderContext, Map<Integer, Float>>> perLeafResults, int k) {
         // Iterate over all scores to get min competitive score
         PriorityQueue<Float> topKMinQueue = new PriorityQueue<>(k);
 
         int count = 0;
-        for (Map<Integer, Float> perLeafResult : perLeafResults) {
-            count += perLeafResult.size();
-            for (Float score : perLeafResult.values()) {
+        for (Map.Entry<LeafReaderContext, Map<Integer, Float>> perLeafResult : perLeafResults) {
+            Map<Integer, Float> docIdScoreMap = perLeafResult.getValue();
+            count += docIdScoreMap.size();
+            for (Float score : docIdScoreMap.values()) {
                 if (topKMinQueue.size() < k) {
                     topKMinQueue.add(score);
                 } else if (topKMinQueue.peek() != null && score > topKMinQueue.peek()) {
@@ -54,7 +56,7 @@ public final class ResultUtil {
 
         // Reduce the results based on min competitive score
         float minScore = topKMinQueue.peek() == null ? -Float.MAX_VALUE : topKMinQueue.peek();
-        perLeafResults.forEach(results -> results.entrySet().removeIf(entry -> entry.getValue() < minScore));
+        perLeafResults.forEach(results -> results.getValue().entrySet().removeIf(entry -> entry.getValue() < minScore));
     }
 
     /**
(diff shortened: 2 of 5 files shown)


src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java
```diff
@@ -64,20 +65,33 @@ public Weight createWeight(IndexSearcher indexSearcher, ScoreMode scoreMode, flo
...
+                }
+            }
+            perLeafResults = doSearch(indexSearcher, approxSearchSegments, knnWeight, firstPassK);
             if (isShardLevelRescoringEnabled == true) {
                 ResultUtil.reduceToTopK(perLeafResults, firstPassK);
             }
 
             StopWatch stopWatch = new StopWatch().start();
-            perLeafResults = doRescore(indexSearcher, leafReaderContexts, knnWeight, perLeafResults, finalK);
+            perLeafResults = doRescore(indexSearcher, knnWeight, perLeafResults, finalK);
```
heemin32, line 84: Previously, we did exact search on finalK. After this change, we still does exact search on finalK. Could you tell me how will this improve the latency?

VijayanB: doSearch can call either ApproxSearch or Exact Search based on conditions like whether engine files exists or not, number of docs after filter is less than k. In those cases, we will quantize query vector, and every vector from segments, and, then perform distance computation using Hamming distance for firstPassK. With this approach, we only call doSearch for those segments which we know will always call approxsearch, and, for other segments we will call exact search without quantization with finalK.  The optimization is at  https://github.com/opensearch-project/k-NN/pull/2273/files#diff-9cfe412357ba56b3ef216427d491fc653535686a760e8ba19ea1aa00fc0e0338R68-R78

heemin32: Are you assuming that an exact search on full precision vectors will be faster than an exact search with quantized vectors due to the slower quantization process? It would be interesting to see the benchmark results for this.

If that’s the case, an alternative could be to retrieve quantized values directly from the Faiss file instead of performing on-the-fly quantization.

VijayanB: Yes, exact search on full precision for k is lesser than, exact search on quantization for first pass K + rescore matched docs on full precision . The linked GitHub issues actually shows how performance got impacted 10x when there are segments with no faiss engine files. In my POC, I saw improvements but recall was poor because of using order as link between results and leaf reader context. I am rerunning experiments with my change to collect metrics with latency and recall

heemin32: There is one case where we are running exact search; when the returned result is less than k. Are we going to handle that case as well?

VijayanB: I believe this happens with filter, if so, yes, it was already taken care

heemin32: No. It happens regardless there is filter or not. https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/query/KNNWeight.java#L149

VijayanB: No. It can happen either when there are no engine files or after filter that number of matched documents is less than k. We only decided to call doSearch if we know that it will call Approx Search API. For other segments we will directly call Exact Search, this PR is about that only

jmazanec15:
Should we skip quantization on indexing if we are not using it here then?

navneet1v:
> Should we skip quantization on indexing if we are not using it here then?

+1. I think it is valid point. We should do that too. But @VijayanB there were couple of more ideas that we were discussing on how to fix this issue. Did you put some thoughts on that like why we should split the segments by search type?

src/main/java/org/opensearch/knn/index/query/ResultUtil.java
```diff
@@ -30,14 +31,15 @@ public final class ResultUtil {
      * @param perLeafResults Results from the list
      * @param k the number of results across all leaf results to return
      */
-    public static void reduceToTopK(List<Map<Integer, Float>> perLeafResults, int k) {
+    public static void reduceToTopK(List<Map.Entry<LeafReaderContext, Map<Integer, Float>>> perLeafResults, int k) {
```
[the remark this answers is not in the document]
heemin32: Would `PerLeafResults` work as a class name? It feels more generic and versatile compared to including `FirstPass`. Perhaps `firstPassResults` could be used as a variable name instead?

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -153,6 +153,26 @@ public Map<Integer, Float> searchLeaf(LeafReaderContext context, int k) throws I
...
+    /**
+     * For given {@link LeafReaderContext}, this api will return will KNNWeight perform exact search or not
+     * always. This decision is based on two properties, 1) if there are no native engine files in segments,
+     * exact search will always be performed, 2) if number of docs after filter is less than 'k'
+     * @param context
+     * @return
+     * @throws IOException
+     */
+    public boolean isExactSearchPreferred(LeafReaderContext context) throws IOException {
+        final BitSet filterBitSet = getFilteredDocsBitSet(context);
```
shatejas, line 165: For each ANN leaf, getting the filter Bitset is now happening twice. Once for this check, and then for getting the actual [filter bitset](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/query/KNNWeight.java#L129). 

Considering worst case scenario, This goes through [filterWeight.scorer](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/query/KNNWeight.java#L164) twice and then creating a bitset which involves a [linear loop](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/query/KNNWeight.java#L184)

How confident are we that this won't impact latencies for filtering cases? 

Can we avoid this duplicate? One way is to do only engine empty check here. and then pass the bitset in [searchLeaf](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/query/KNNWeight.java#L128) and do exact Search  context

Let me know if I am missing something

src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java
```diff
@@ -121,12 +132,35 @@ private List<Map<Integer, Float>> doRescore(
...
                     .build();
-                return knnWeight.exactSearch(leafReaderContext, exactSearcherContext);
+                final Map<Integer, Float> docIdScoreMap = knnWeight.exactSearch(entry.getKey(), exactSearcherContext);
+                return new AbstractMap.SimpleEntry<>(entry.getKey(), docIdScoreMap);
             });
         }
         return indexSearcher.getTaskExecutor().invokeAll(rescoreTasks);
     }
 
+    private List<Map.Entry<LeafReaderContext, Map<Integer, Float>>> doExactSearch(
```
shatejas, line 142: doRescore and this method are more or less similar, any possiblity that we can reuse something here?

src/main/java/org/opensearch/knn/index/query/ResultUtil.java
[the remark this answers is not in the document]
shatejas: +1

heemin32:
> > Should we skip quantization on indexing if we are not using it here then?
> 
> +1. I think it is valid point. We should do that too. But @VijayanB there were couple of more ideas that we were discussing on how to fix this issue. Did you put some thoughts on that like why we should split the segments by search type?

Another option could be to read the quantized values directly from the native engine file, as discussed here: https://github.com/opensearch-project/k-NN/issues/2266. This approach would also address cases where the search results are fewer than `k`, allowing us to fall back to an exact search.

Issue #114, opened 2021-10-03
[faiss] KNNMethodContext and MethodComponentContext  need to be streamable for train api


jmazanec15:
The training api's request will take a k-NN method definition as a parameter. Because this request may be passed around to different nodes in the cluster, it needs to implement stream interfaces.

PR #2283, opened 2024-11-20
Add Support for Multi Values in innerHit for Nested k-NN Fields in Lucene and FAISS

### LINKED ISSUE #2249: [RFC] Multiple inner hits for nested field
## Update
We are going to introduce a new parameter, `expand_nested_docs` in knn query. The default value will be false.
```
GET /my-knn-index-1/_search
{
  "query": {
    "nested": {
      "path": "nested_field",
      "query": {
        "knn": {
          "nested_field.my_vector": {
            "vector": [1,1,1],
            "k": 2,
            "expand_nested_docs": true (false when not specified)
          }
        }
      },
      "inner_hits": {}
    }
  }
}
```
This parameter retrieves all nested documents of the top-k parent documents. Since the default score mode for nested field queries is `avg`, enabling this parameter will alter the parent document scores and their order in the search results. To maintain the current scoring, users can set the `score_mode` to `max`.  

The primary purpose of this change is to ensure consistency between the main query and the `inner_hit` query, eliminating the risk of score discrepancies between the two. Additionally, for use cases where multi `inner_hits` are not required, queries with `expand_nested_docs` set to `false` will have reduced latency. This approach also offers the added benefit of allowing users to experiment with alternative `score_mode` options, such as `avg` or `sum`, beyond the `max`. However, it’s important to note that using avg or sum does not guarantee the returned documents are true top-k by those modes; the top-k is initially determined by max and then reordered using avg or sum among those documents.

## Overview

Following the launch of multi-vector support in k-NN nested fields, search results with inner hits currently return only the single highest-scoring nested document per parent document. However, many users prefer to see the scores of all nested documents within inner hits, allowing them to display all relevant nested documents rather than just the top-scoring one.

## Related GitHub Issue
* https://github.com/opensearch-project/k-NN/issues/2113
* https://github.com/opensearch-project/k-NN/issues/1768

## Work Items
- [x] POC for Lucene
- [x] Adding new field in QueryShardContext
- [x] Implementation for faiss
- [x] Implementation for lucene (Maybe after https://github.com/opensearch-project/k-NN/issues/2115 is completed?)

## Current state

Let’s try to understand the issue with example. First we are ingesting a document with two nested documents each of which has vector [1, 1, 1], and vector [2, 2, 2]. 

```
PUT /my-knn-index-1/_doc/1
{"nested_field":[{"my_vector":[1,1,1]}, {"my_vector":[2,2,2]}]}
```

Then, we execute the search with inner_hits parameter.

```
GET /my-knn-index-1/_search
{
  "query": {
    "nested": {
      "path": "nested_field",
      "query": {
        "knn": {
          "nested_field.my_vector": {
            "vector": [1,1,1],
            "k": 2
          }
        }
      },
      "inner_hits": {}
    }
  }
}
```

The search result returns only single nested document with vector [1, 1, 1]
```
{
    "_index": "my-knn-index-1",
    "_id": "1",
    "_score": 1.0,
    "_source": {
        "nested_field": [
        {
            "my_vector": [
            1,
            1,
            1
            ]
        },
        {
            "my_vector": [
            2,
            2,
            2
            ]
        }
        ]
    },
    "inner_hits": {
        "nested_field": {
        "hits": {
            "total": {
            "value": 1,
            "relation": "eq"
            },
            "max_score": 1.0,
            "hits": [
            {
                "_index": "my-knn-index-1",
                "_id": "1",
                "_nested": {
                "field": "nested_field",
                "offset": 0
                },
                "_score": 1.0,
                "_source": {
                "my_vector": [
                    1,
                    1,
                    1
                ]
                }
            }
            ]
        }
      }
    }
}
```
## Desired state

We want to have all nested documents inside inner_hits block for each returned parent documents.
```
"inner_hits": {
    "nested_field": {
    "hits": {
        "total": {
        "value": 2,
        "relation": "eq"
        },
        "max_score": 1.0,
        "hits": [
        {
            "_index": "my-knn-index-1",
            "_id": "1",
            "_nested": {
            "field": "nested_field",
            "offset": 0
            },
            "_score": 1.0,
            "_source": {
            "my_vector": [
                1,
                1,
                1
            ]
            }
        },
       {
            "_index": "my-knn-index-1",
            "_id": "1",
            "_nested": {
            "field": "nested_field",
            "offset": 1
            },
            "_score": 0.25,
            "_source": {
            "my_vector": [
                2,
                2,
                2
            ]
            }
        }
        ]
    }
   }
}
```

## Class diagram
Screenshot 2024-11-05 at 1 06 48 PM

### For Faiss

For faiss, we need to migrate the entire query to NativeEngineKnnVectorQuery from KNNQuery. Currently, we use NativeEngineKnnVectorQuery only when the rescoring context exist.

Inside createWeight method of NativeEngineKnnVectorQuery, if it is nested field, we do another exact search with filtering on topK parentIds to retrieve all child documents with its score.

```
public class NativeEngineKnnVectorQuery extends Query {

    private final KNNQuery knnQuery;

    // Update on existing method
    @Override
    public Weight createWeight(IndexSearcher indexSearcher, ScoreMode scoreMode, float boost) throws IOException {
        final IndexReader reader = indexSearcher.getIndexReader();
        final KNNWeight knnWeight = (KNNWeight) knnQuery.createWeight(indexSearcher, scoreMode, 1);
        List<Map<Integer, Float>> perLeafResults;
        RescoreContext rescoreContext = knnQuery.getRescoreContext();
        final int finalK = knnQuery.getK();
        if (rescoreContext == null) {
            perLeafResults = doSearch(indexSearcher, leafReaderContexts, knnWeight, finalK);
        } else {
            boolean isShardLevelRescoringEnabled = KNNSettings.isShardLevelRescoringEnabledForDiskBasedVector(knnQuery.getIndexName());
            int dimension = knnQuery.getQueryVector().length;
            int firstPassK = rescoreContext.getFirstPassK(finalK, isShardLevelRescoringEnabled, dimension);
            perLeafResults = doSearch(indexSearcher, leafReaderContexts, knnWeight, firstPassK);
            if (isShardLevelRescoringEnabled == true) {
                ResultUtil.reduceToTopK(perLeafResults, firstPassK);
            }
    
            perLeafResults = doRescore(indexSearcher, leafReaderContexts, knnWeight, perLeafResults, finalK);
            long rescoreTime = stopWatch.stop().totalTime().millis();
            log.debug("Rescoring results took {} ms. oversampled k:{}, segments:{}", rescoreTime, firstPassK, leafReaderContexts.size());
        }
        ResultUtil.reduceToTopK(perLeafResults, finalK);
        **// if it is innerHit query
            // get parents doc ids from the topK results
            // do exact search using the parent doc ids as filtered ids
            // return the results of the exact search** 
        TopDocs[] topDocs = new TopDocs[perLeafResults.size()];
        for (int i = 0; i < perLeafResults.size(); i++) {
            topDocs[i] = ResultUtil.resultMapToTopDocs(perLeafResults.get(i), leafReaderContexts.get(i).docBase);
        }
    
        TopDocs topK = TopDocs.merge(knnQuery.getK(), topDocs);
        if (topK.scoreDocs.length == 0) {
            return new MatchNoDocsQuery().createWeight(indexSearcher, scoreMode, boost);
        }
        return createDocAndScoreQuery(reader, topK).createWeight(indexSearcher, scoreMode, boost);
    }
}
```

### For Lucene

For lucene, we will create our own wrapper query around lucene knn query as @navneet1v suggested in this GitHub issue. https://github.com/opensearch-project/k-NN/issues/2115
The rest of the code logic will be similar to NativeEngineKnnVectorQuery where we do another exact search on topK parentIds to retrieve all child documents with its score. 
```
// This is a new class need to be added
public class LuceneEngineKNNQuery extends Query {
    Query luceneQuery; // this can be KnnFloatVectorQuery, KnnByteVectorQuery, DiversifyingChildrenByteKnnVectorQuery etc.

        @Override
        // rewrite the lucene query
        Query docAndScoreQuery = searcher.rewrite(luceneQuery);
    
        Weight weight = docAndScoreQuery.createWeight(searcher, scoreMode, boost);
        // get scorer for each segments
        for each leafReaderContext
            List<Scorer> scorers = weight.scorer(leafReaderContext);
        
        // get perLeafResults from scorers
        List<Map<Integer, Float>> perLeafResults;
        perLeafResults = getPerLeafResults(scorers);
        
        // reduce the result to topK
        ResultUtil.reduceToTopK(perLeafResults, firstPassK);
        
        // if it is innerHit query
            // get parents doc ids from the topK results
            // do exact search using the parent doc ids as filtered ids
            // return the results of the exact search
    }
}
```

### Separating nested field query with innerHit query

Currently, KNNQueryBuilder.doToQuery(QueryShardContext) is called twice when innerHit is requested: one for nested field query and another for inner hit query. We want to add new field in QueryShardContext to distinguish if this is for nested field query or inner hit query so that we can retrieve all child documents only for inner hit query. The reason of not retrieving all child documents for nested field query is to avoid a degradation on latency.  This is a two way door where we can support retrieving all child documents for nested field query if we want to support different scoring mode other than max. https://github.com/opensearch-project/k-NN/issues/1743

heemin32:
### Description
This PR introduces support for returning all nested fields with their scores inside innerHit for nested k-NN fields, applicable to both Lucene and FAISS engines.

The implementation involves executing a search request across all segments and collecting results at the shard level, similar to the approach used in disk-based k-NN searches. After reducing the results to the top k, we retrieve all sibling documents associated with these results. Using the IDs of the retrieved sibling documents as filtered document IDs, we perform another exact search to score them comprehensively.

Here are additional explanations for the changes made:
1. Added JsonPath as a dependency exclusively for integration testing, using version 2.8.0. 2.9.0 has an dependency conflict issue with SLF4J.
2. Adopted a composite approach in NestedKnnVectorInnerHitQuery.java to enable code reuse between byte vectors and float vectors.
3. Replaced the use of BitSet with DocIdSetIterator for filteredDocId to eliminate the overhead of converting from an iterator to a BitSet and back to an iterator.

### Related Issues
Resolves #2249

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/query/ExactSearcher.java
+++ b/src/main/java/org/opensearch/knn/index/query/ExactSearcher.java
@@ -17,7 +17,6 @@ import org.apache.lucene.index.SegmentReader;
 import org.apache.lucene.search.DocIdSetIterator;
 import org.apache.lucene.search.HitQueue;
 import org.apache.lucene.search.ScoreDoc;
-import org.apache.lucene.util.BitSet;
 import org.opensearch.common.lucene.Lucene;
 import org.opensearch.knn.common.FieldInfoExtractor;
 import org.opensearch.knn.index.SpaceType;
@@ -68,8 +67,8 @@ public class ExactSearcher {
         if (exactSearcherContext.getKnnQuery().getRadius() != null) {
             return doRadialSearch(leafReaderContext, exactSearcherContext, iterator);
         }
-        if (exactSearcherContext.getMatchedDocs() != null
-            && exactSearcherContext.getMatchedDocs().cardinality() <= exactSearcherContext.getK()) {
+        if (exactSearcherContext.getMatchedDocsIterator() != null
+            && exactSearcherContext.numberOfMatchedDocs <= exactSearcherContext.getK()) {
             return scoreAllDocs(iterator);
         }
         return searchTopCandidates(iterator, exactSearcherContext.getK(), Predicates.alwaysTrue());
@@ -155,7 +154,7 @@ public class ExactSearcher {
 
     private KNNIterator getKNNIterator(LeafReaderContext leafReaderContext, ExactSearcherContext exactSearcherContext) throws IOException {
         final KNNQuery knnQuery = exactSearcherContext.getKnnQuery();
-        final BitSet matchedDocs = exactSearcherContext.getMatchedDocs();
+        final DocIdSetIterator matchedDocs = exactSearcherContext.getMatchedDocsIterator();
         final SegmentReader reader = Lucene.segmentReader(leafReaderContext.reader());
         final FieldInfo fieldInfo = FieldInfoExtractor.getFieldInfo(reader, knnQuery.getField());
         if (fieldInfo == null) {
@@ -245,7 +244,8 @@ public class ExactSearcher {
          */
         boolean useQuantizedVectorsForSearch;
         int k;
-        BitSet matchedDocs;
+        DocIdSetIterator matchedDocsIterator;
+        long numberOfMatchedDocs;
         KNNQuery knnQuery;
         /**
          * whether the matchedDocs contains parent ids or child ids. This is relevant in the case of
--- a/src/main/java/org/opensearch/knn/index/query/KNNQueryFactory.java
+++ b/src/main/java/org/opensearch/knn/index/query/KNNQueryFactory.java
@@ -10,27 +10,28 @@ import org.apache.lucene.search.KnnByteVectorQuery;
 import org.apache.lucene.search.KnnFloatVectorQuery;
 import org.apache.lucene.search.Query;
 import org.apache.lucene.search.join.BitSetProducer;
-import org.apache.lucene.search.join.DiversifyingChildrenByteKnnVectorQuery;
-import org.apache.lucene.search.join.DiversifyingChildrenFloatKnnVectorQuery;
 import org.opensearch.index.query.QueryShardContext;
 import org.opensearch.knn.index.VectorDataType;
 import org.opensearch.knn.index.engine.KNNEngine;
+import org.opensearch.knn.index.query.common.QueryUtils;
+import org.opensearch.knn.index.query.lucenelib.NestedKnnVectorQueryFactory;
 import org.opensearch.knn.index.query.nativelib.NativeEngineKnnVectorQuery;
 import org.opensearch.knn.index.query.rescore.RescoreContext;
 
 import java.util.Locale;
 import java.util.Map;
 
+import static org.opensearch.knn.common.KNNConstants.EXPAND_NESTED;
 import static org.opensearch.knn.common.KNNConstants.METHOD_PARAMETER_EF_SEARCH;
 import static org.opensearch.knn.common.KNNConstants.VECTOR_DATA_TYPE_FIELD;
 import static org.opensearch.knn.index.VectorDataType.SUPPORTED_VECTOR_DATA_TYPES;
+import static org.opensearch.knn.index.engine.KNNEngine.ENGINES_SUPPORTING_MULTI_VECTORS;
 
 /**
  * Creates the Lucene k-NN queries
  */
 @Log4j2
 public class KNNQueryFactory extends BaseQueryFactory {
-
     /**
      * Creates a Lucene query for a particular engine.
      * @param createQueryRequest request object that has all required fields to construct the query
@@ -48,11 +49,23 @@ public class KNNQueryFactory extends BaseQueryFactory {
         final Query filterQuery = getFilterQuery(createQueryRequest);
         final Map<String, ?> methodParameters = createQueryRequest.getMethodParameters();
         final RescoreContext rescoreContext = createQueryRequest.getRescoreContext().orElse(null);
-
+        final KNNEngine knnEngine = createQueryRequest.getKnnEngine();
+        final boolean expandNested = createQueryRequest.isExpandNested();
         BitSetProducer parentFilter = null;
         if (createQueryRequest.getContext().isPresent()) {
             QueryShardContext context = createQueryRequest.getContext().get();
             parentFilter = context.getParentFilter();
+        }
+
+        if (parentFilter == null && expandNested) {
+            throw new IllegalArgumentException(
+                String.format(
+                    Locale.ROOT,
+                    "Invalid value provided for the [%s] field. [%s] is only supported with a nested field.",
+                    EXPAND_NESTED,
+                    EXPAND_NESTED
+                )
+            );
         }
 
         if (KNNEngine.getEnginesThatCreateCustomSegmentFiles().contains(createQueryRequest.getKnnEngine())) {
@@ -95,7 +108,16 @@ public class KNNQueryFactory extends BaseQueryFactory {
                         .rescoreContext(rescoreContext)
                         .build();
             }
-            return createQueryRequest.getRescoreContext().isPresent() ? new NativeEngineKnnVectorQuery(knnQuery) : knnQuery;
+
+            if (createQueryRequest.getRescoreContext().isPresent()) {
+                return new NativeEngineKnnVectorQuery(knnQuery, QueryUtils.INSTANCE, expandNested);
+            }
+
+            if (ENGINES_SUPPORTING_MULTI_VECTORS.contains(knnEngine) && expandNested) {
+                return new NativeEngineKnnVectorQuery(knnQuery, QueryUtils.INSTANCE, expandNested);
+            }
+
+            return knnQuery;
         }
 
         Integer requestEfSearch = null;
@@ -106,9 +128,9 @@ public class KNNQueryFactory extends BaseQueryFactory {
         log.debug(String.format("Creating Lucene k-NN query for index: %s \"\", field: %s \"\", k: %d", indexName, fieldName, k));
         switch (vectorDataType) {
             case BYTE:
-                return getKnnByteVectorQuery(fieldName, byteVector, luceneK, filterQuery, parentFilter);
+                return getKnnByteVectorQuery(fieldName, byteVector, luceneK, filterQuery, parentFilter, expandNested);
             case FLOAT:
-                return getKnnFloatVectorQuery(fieldName, vector, luceneK, filterQuery, parentFilter);
+                return getKnnFloatVectorQuery(fieldName, vector, luceneK, filterQuery, parentFilter, expandNested);
             default:
                 throw new IllegalArgumentException(
                     String.format(
@@ -131,38 +153,56 @@ public class KNNQueryFactory extends BaseQueryFactory {
     }
 
     /**
-     * If parentFilter is not null, it is a nested query. Therefore, we return {@link DiversifyingChildrenByteKnnVectorQuery}
-     * which will dedupe search result per parent so that we can get k parent results at the end.
+     * If parentFilter is not null, it is a nested query. Therefore, we delegate creation of query to {@link NestedKnnVectorQueryFactory}
+     * which will create query to dedupe search result per parent so that we can get k parent results at the end.
      */
     private static Query getKnnByteVectorQuery(
         final String fieldName,
         final byte[] byteVector,
         final int k,
         final Query filterQuery,
-        final BitSetProducer parentFilter
+        final BitSetProducer parentFilter,
+        final boolean expandNested
     ) {
         if (parentFilter == null) {
+            assert expandNested == false : "expandNested is allowed to be true only for nested fields.";
             return new KnnByteVectorQuery(fieldName, byteVector, k, filterQuery);
         } else {
-            return new DiversifyingChildrenByteKnnVectorQuery(fieldName, byteVector, filterQuery, k, parentFilter);
+            return NestedKnnVectorQueryFactory.createNestedKnnVectorQuery(
+                fieldName,
+                byteVector,
+                k,
+                filterQuery,
+                parentFilter,
+                expandNested
+            );
         }
     }
 
     /**
-     * If parentFilter is not null, it is a nested query. Therefore, we return {@link DiversifyingChildrenFloatKnnVectorQuery}
-     * which will dedupe search result per parent so that we can get k parent results at the end.
+     * If parentFilter is not null, it is a nested query. Therefore, we delegate creation of query to {@link NestedKnnVectorQueryFactory}
+     * which will create query to dedupe search result per parent so that we can get k parent results at the end.
      */
     private static Query getKnnFloatVectorQuery(
         final String fieldName,
         final float[] floatVector,
         final int k,
         final Query filterQuery,
-        final BitSetProducer parentFilter
+        final BitSetProducer parentFilter,
+        final boolean expandNested
     ) {
         if (parentFilter == null) {
+            assert expandNested == false : "expandNested is allowed to be true only for nested fields.";
             return new KnnFloatVectorQuery(fieldName, floatVector, k, filterQuery);
         } else {
-            return new DiversifyingChildrenFloatKnnVectorQuery(fieldName, floatVector, filterQuery, k, parentFilter);
+            return NestedKnnVectorQueryFactory.createNestedKnnVectorQuery(
+                fieldName,
+                floatVector,
+                k,
+                filterQuery,
+                parentFilter,
+                expandNested
+            );
         }
     }
 }
--- a/src/main/java/org/opensearch/knn/index/query/KNNWeight.java
+++ b/src/main/java/org/opensearch/knn/index/query/KNNWeight.java
@@ -6,6 +6,7 @@
 package org.opensearch.knn.index.query;
 
 import com.google.common.annotations.VisibleForTesting;
+import lombok.Getter;
 import lombok.extern.log4j.Log4j2;
 import org.apache.lucene.index.FieldInfo;
 import org.apache.lucene.index.LeafReaderContext;
@@ -66,6 +67,7 @@ public class KNNWeight extends Weight {
     private final float boost;
 
     private final NativeMemoryCacheManager nativeMemoryCacheManager;
+    @Getter
     private final Weight filterWeight;
     private final ExactSearcher exactSearcher;
 
@@ -109,7 +111,7 @@ public class KNNWeight extends Weight {
 
     @Override
     public Scorer scorer(LeafReaderContext context) throws IOException {
-        final Map<Integer, Float> docIdToScoreMap = searchLeaf(context, knnQuery.getK());
+        final Map<Integer, Float> docIdToScoreMap = searchLeaf(context, knnQuery.getK()).getResult();
         if (docIdToScoreMap.isEmpty()) {
             return KNNScorer.emptyScorer(this);
         }
@@ -125,14 +127,14 @@ public class KNNWeight extends Weight {
      * @param k Number of results to return
      * @return A Map of docId to scores for top k results
      */
-    public Map<Integer, Float> searchLeaf(LeafReaderContext context, int k) throws IOException {
+    public PerLeafResult searchLeaf(LeafReaderContext context, int k) throws IOException {
         final BitSet filterBitSet = getFilteredDocsBitSet(context);
         int cardinality = filterBitSet.cardinality();
         // We don't need to go to JNI layer if no documents are found which satisfy the filters
         // We should give this condition a deeper look that where it should be placed. For now I feel this is a good
         // place,
         if (filterWeight != null && cardinality == 0) {
-            return Collections.emptyMap();
+            return PerLeafResult.EMPTY_RESULT;
         }
         /*
          * The idea for this optimization is to get K results, we need to at least look at K vectors in the HNSW graph
@@ -140,17 +142,19 @@ public class KNNWeight extends Weight {
          * This improves the recall.
          */
         if (isFilteredExactSearchPreferred(cardinality)) {
-            return doExactSearch(context, filterBitSet, k);
+            Map<Integer, Float> result = doExactSearch(context, new BitSetIterator(filterBitSet, cardinality), cardinality, k);
+            return new PerLeafResult(filterWeight == null ? null : filterBitSet, result);
         }
         Map<Integer, Float> docIdsToScoreMap = doANNSearch(context, filterBitSet, cardinality, k);
         // See whether we have to perform exact search based on approx search results
         // This is required if there are no native engine files or if approximate search returned
         // results less than K, though we have more than k filtered docs
         if (isExactSearchRequire(context, cardinality, docIdsToScoreMap.size())) {
-            final BitSet docs = filterWeight != null ? filterBitSet : null;
-            return doExactSearch(context, docs, k);
+            final BitSetIterator docs = filterWeight != null ? new BitSetIterator(filterBitSet, cardinality) : null;
+            Map<Integer, Float> result = doExactSearch(context, docs, cardinality, k);
+            return new PerLeafResult(filterWeight == null ? null : filterBitSet, result);
         }
-        return docIdsToScoreMap;
+        return new PerLeafResult(filterWeight == null ? null : filterBitSet, docIdsToScoreMap);
     }
 
     private BitSet getFilteredDocsBitSet(final LeafReaderContext ctx) throws IOException {
@@ -205,17 +209,21 @@ public class KNNWeight extends Weight {
         return intArray;
     }
 
-    private Map<Integer, Float> doExactSearch(final LeafReaderContext context, final BitSet acceptedDocs, int k) throws IOException {
+    private Map<Integer, Float> doExactSearch(
+        final LeafReaderContext context,
+        final DocIdSetIterator acceptedDocs,
+        final long numberOfAcceptedDocs,
+        int k
+    ) throws IOException {
         final ExactSearcherContextBuilder exactSearcherContextBuilder = ExactSearcher.ExactSearcherContext.builder()
             .isParentHits(true)
             .k(k)
             // setting to true, so that if quantization details are present we want to do search on the quantized
             // vectors as this flow is used in first pass of search.
             .useQuantizedVectorsForSearch(true)
-            .knnQuery(knnQuery);
-        if (acceptedDocs != null) {
-            exactSearcherContextBuilder.matchedDocs(acceptedDocs);
-        }
+            .knnQuery(knnQuery)
+            .matchedDocsIterator(acceptedDocs)
+            .numberOfMatchedDocs(numberOfAcceptedDocs);
         return exactSearch(context, exactSearcherContextBuilder.build());
     }
 
--- a/src/main/java/org/opensearch/knn/index/query/PerLeafResult.java
+++ b/src/main/java/org/opensearch/knn/index/query/PerLeafResult.java
@@ -1,0 +1,28 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.knn.index.query;
+
+import lombok.Getter;
+import lombok.Setter;
+import org.apache.lucene.util.Bits;
+import org.opensearch.common.Nullable;
+
+import java.util.Collections;
+import java.util.Map;
+
+@Getter
+public class PerLeafResult {
+    public static final PerLeafResult EMPTY_RESULT = new PerLeafResult(new Bits.MatchNoBits(0), Collections.emptyMap());
+    @Nullable
+    private final Bits filterBits;
+    @Setter
+    private Map<Integer, Float> result;
+
+    public PerLeafResult(final Bits filterBits, final Map<Integer, Float> result) {
+        this.filterBits = filterBits == null ? new Bits.MatchAllBits(0) : filterBits;
+        this.result = result;
+    }
+}
--- a/src/main/java/org/opensearch/knn/index/query/ResultUtil.java
+++ b/src/main/java/org/opensearch/knn/index/query/ResultUtil.java
@@ -9,7 +9,6 @@ import org.apache.lucene.search.DocIdSetIterator;
 import org.apache.lucene.search.ScoreDoc;
 import org.apache.lucene.search.TopDocs;
 import org.apache.lucene.search.TotalHits;
-import org.apache.lucene.util.BitSet;
 import org.apache.lucene.util.DocIdSetBuilder;
 
 import java.io.IOException;
@@ -30,14 +29,14 @@ public final class ResultUtil {
      * @param perLeafResults Results from the list
      * @param k the number of results across all leaf results to return
      */
-    public static void reduceToTopK(List<Map<Integer, Float>> perLeafResults, int k) {
+    public static void reduceToTopK(List<PerLeafResult> perLeafResults, int k) {
         // Iterate over all scores to get min competitive score
         PriorityQueue<Float> topKMinQueue = new PriorityQueue<>(k);
 
         int count = 0;
-        for (Map<Integer, Float> perLeafResult : perLeafResults) {
-            count += perLeafResult.size();
-            for (Float score : perLeafResult.values()) {
+        for (PerLeafResult perLeafResult : perLeafResults) {
+            count += perLeafResult.getResult().size();
+            for (Float score : perLeafResult.getResult().values()) {
                 if (topKMinQueue.size() < k) {
                     topKMinQueue.add(score);
                 } else if (topKMinQueue.peek() != null && score > topKMinQueue.peek()) {
@@ -54,23 +53,22 @@ public final class ResultUtil {
 
         // Reduce the results based on min competitive score
         float minScore = topKMinQueue.peek() == null ? -Float.MAX_VALUE : topKMinQueue.peek();
-        perLeafResults.forEach(results -> results.entrySet().removeIf(entry -> entry.getValue() < minScore));
+        perLeafResults.forEach(results -> results.getResult().entrySet().removeIf(entry -> entry.getValue() < minScore));
     }
 
     /**
-     * Convert map to bit set, if resultMap is empty or null then returns an Optional. Returning an optional here to
-     * ensure that the caller is aware that BitSet may not be present
+     * Convert map of docs to doc id set iterator
      *
      * @param resultMap Map of results
-     * @return BitSet of results; null is returned if the result map is empty
+     * @return Doc id set iterator
      * @throws IOException If an error occurs during the search.
      */
-    public static BitSet resultMapToMatchBitSet(Map<Integer, Float> resultMap) throws IOException {
-        if (resultMap == null || resultMap.isEmpty()) {
-            return null;
+    public static DocIdSetIterator resultMapToDocIds(Map<Integer, Float> resultMap) throws IOException {
+        if (resultMap.isEmpty()) {
+            return DocIdSetIterator.empty();
         }
         final int maxDoc = Collections.max(resultMap.keySet()) + 1;
-        return BitSet.of(resultMapToDocIds(resultMap, maxDoc), maxDoc);
+        return resultMapToDocIds(resultMap, maxDoc);
     }
 
     /**
--- a/src/main/java/org/opensearch/knn/index/query/common/DocAndScoreQuery.java
+++ b/src/main/java/org/opensearch/knn/index/query/common/DocAndScoreQuery.java
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-package org.opensearch.knn.index.query.nativelib;
+package org.opensearch.knn.index.query.common;
 
 import org.apache.lucene.index.LeafReaderContext;
 import org.apache.lucene.search.DocIdSetIterator;
@@ -32,7 +32,7 @@ final class DocAndScoreQuery extends Query {
     private final int[] segmentStarts;
     private final Object contextIdentity;
 
-    DocAndScoreQuery(int k, int[] docs, float[] scores, int[] segmentStarts, Object contextIdentity) {
+    public DocAndScoreQuery(int k, int[] docs, float[] scores, int[] segmentStarts, Object contextIdentity) {
         this.k = k;
         this.docs = docs;
         this.scores = scores;
--- a/src/main/java/org/opensearch/knn/index/query/common/QueryUtils.java
+++ b/src/main/java/org/opensearch/knn/index/query/common/QueryUtils.java
@@ -1,0 +1,173 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.knn.index.query.common;
+
+import org.apache.lucene.index.IndexReader;
+import org.apache.lucene.index.LeafReaderContext;
+import org.apache.lucene.search.DocIdSetIterator;
+import org.apache.lucene.search.FilteredDocIdSetIterator;
+import org.apache.lucene.search.IndexSearcher;
+import org.apache.lucene.search.Query;
+import org.apache.lucene.search.Scorer;
+import org.apache.lucene.search.TopDocs;
+import org.apache.lucene.search.Weight;
+import org.apache.lucene.search.join.BitSetProducer;
+import org.apache.lucene.util.BitSet;
+import org.apache.lucene.util.BitSetIterator;
+import org.apache.lucene.util.Bits;
+import org.opensearch.knn.index.query.iterators.GroupedNestedDocIdSetIterator;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.Callable;
+
+/**
+ * This class contains utility methods that help customize the search results
+ */
+public class QueryUtils {
+    public static QueryUtils INSTANCE = new QueryUtils();
+
+    /**
+     * Returns a query that represents the specified TopDocs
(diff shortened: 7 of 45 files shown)


src/main/java/org/opensearch/knn/index/query/ExactSearcher.java
```diff
@@ -60,11 +60,13 @@ public class ExactSearcher {
...
         KNNIterator iterator = getKNNIterator(leafReaderContext, exactSearcherContext);
+        if (iterator == null) {
+            return Collections.emptyMap();
+        }
         if (exactSearcherContext.getKnnQuery().getRadius() != null) {
             return doRadialSearch(leafReaderContext, exactSearcherContext, iterator);
         }
-        if (exactSearcherContext.getMatchedDocs() != null
-            && exactSearcherContext.getMatchedDocs().cardinality() <= exactSearcherContext.getK()) {
+        if (exactSearcherContext.getMatchedDocs() != null && exactSearcherContext.getMatchedDocs().cost() <= exactSearcherContext.getK()) {
```
navneet1v, line 69: the `cost()` not necessary always give the right cost. It depends on the underline implementation of the iterator. So what Implementation of iterator will be present under the hood and how correctly does `cost()` function gives the number of docs.

src/main/java/org/opensearch/knn/index/query/ResultUtil.java
```diff
@@ -58,19 +57,18 @@ public static void reduceToTopK(List<Map<Integer, Float>> perLeafResults, int k)
...
-     * @return BitSet of results; null is returned if the result map is empty
+     * @return Doc id set iterator
      * @throws IOException If an error occurs during the search.
      */
-    public static BitSet resultMapToMatchBitSet(Map<Integer, Float> resultMap) throws IOException {
-        if (resultMap == null || resultMap.isEmpty()) {
-            return null;
+    public static DocIdSetIterator resultMapToDocIds(Map<Integer, Float> resultMap) throws IOException {
+        if (resultMap.isEmpty()) {
+            return DocIdSetIterator.empty();
```
navneet1v, line 68: I think we would like to keep this as null in return to ensure that caller can exit out of the code early.

navneet1v (commented):
I am reviewing the lucene lib classes as of now and it will take some time for me get through them but publishing the comments here. to kick start the discussion on some of the other comments in the code

src/main/java/org/opensearch/knn/index/query/ExactSearcher.java
heemin32: That's a good point. I focused primarily on the one that I passed here (GroupedNestedDocIdSetIterator) to ensure the cost calculation isn't an approximation.

When I reviewed the code, it seems the main difference between scoreAllDocs and searchTopCandidates is whether a heap is used to maintain the top-k results. Would it make sense to always call searchTopCandidates, considering the latency difference might be negligible?

Alternatively, I can introduce a new class that holds the iterator along with the accurate cost value, if that would be preferable.

src/main/java/org/opensearch/knn/index/query/common/QueryUtils.java
```diff
@@ -0,0 +1,170 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.Callable;
+
+/**
+ * This class contains utility methods that help customize the search results
+ */
+public class QueryUtils {
```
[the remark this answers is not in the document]
heemin32: Yes. Static is not unit test friendly.

src/main/java/org/opensearch/knn/index/query/common/DocAndScoreQuery.java
```diff
@@ -24,15 +24,15 @@
 /**
  * This is the same as {@link org.apache.lucene.search.AbstractKnnVectorQuery.DocAndScoreQuery}
  */
-final class DocAndScoreQuery extends Query {
+final public class DocAndScoreQuery extends Query {
```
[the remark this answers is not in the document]
heemin32: We need this class in both in lucene and native query.

src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java
```diff
@@ -113,14 +169,9 @@ private List<Map<Integer, Float>> doRescore(
             LeafReaderContext leafReaderContext = leafReaderContexts.get(i);
             int finalI = i;
             rescoreTasks.add(() -> {
-                final BitSet convertedBitSet = ResultUtil.resultMapToMatchBitSet(perLeafResults.get(finalI));
-                // if there is no docIds to re-score from a segment we should return early to ensure that we are not
-                // wasting any computation
-                if (convertedBitSet == null) {
-                    return Collections.emptyMap();
-                }
+                DocIdSetIterator matchedDocs = ResultUtil.resultMapToDocIds(perLeafResults.get(finalI));
```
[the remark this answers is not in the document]
heemin32: Make sense.

src/main/java/org/opensearch/knn/index/query/ResultUtil.java
heemin32: We can exit by checking if the cost is zero or not. How about it instead of null?

src/main/java/org/opensearch/knn/index/query/KNNQueryFactory.java
```diff
@@ -95,7 +102,14 @@ public static Query create(CreateQueryRequest createQueryRequest) {
...
             }
-            return createQueryRequest.getRescoreContext().isPresent() ? new NativeEngineKnnVectorQuery(knnQuery) : knnQuery;
+
+            if (createQueryRequest.getRescoreContext().isPresent()) {
+                return new NativeEngineKnnVectorQuery(knnQuery, QUERY_UTILS, isInnerHitQuery);
+            } else if (ENGINES_SUPPORTING_MULTI_VECTORS.contains(knnEngine) && isInnerHitQuery) {
+                return new NativeEngineKnnVectorQuery(knnQuery, QUERY_UTILS, isInnerHitQuery);
+            } else {
+                return knnQuery;
+            }
```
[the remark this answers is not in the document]
shatejas: It was to isolate disk based vector search as a precaution, we can always call NativeQuery. If we do that we should consider if we need KNNQuery and KNNWeight classes

src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java
```diff
@@ -76,16 +77,71 @@ public Weight createWeight(IndexSearcher indexSearcher, ScoreMode scoreMode, flo
...
+        List<LeafReaderContext> leafReaderContexts,
+        KNNWeight knnWeight,
+        List<Map<Integer, Float>> perLeafResults
+    ) throws IOException {
+        List<Callable<Map<Integer, Float>>> nestedQueryTasks = new ArrayList<>(leafReaderContexts.size());
+        for (int i = 0; i < perLeafResults.size(); i++) {
+            LeafReaderContext leafReaderContext = leafReaderContexts.get(i);
+            int finalI = i;
+            nestedQueryTasks.add(() -> {
+                Bits queryFilter = queryUtils.createBits(leafReaderContext, knnWeight.getFilterWeight());
```
shatejas, line 121: Lets create the Bits for filter in NativeEngineKnnVectorQuery, there is another case for this where the [bits](https://github.com/opensearch-project/k-NN/pull/2273#discussion_r1850774606) are needed, it makes sense to compute once and pass it around

src/main/java/org/opensearch/knn/index/query/common/DocAndScoreQuery.java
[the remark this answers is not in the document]
shatejas: Why can't the lucene query be in the same package?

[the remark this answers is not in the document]
heemin32: No preference. I can put them in the same package if that is what we want.

src/main/java/org/opensearch/knn/index/query/lucenelib/NestedKnnVectorInnerHitQuery.java
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    private TopDocs[] retrieveAll(
+        final IndexSearcher indexSearcher,
+        final List<LeafReaderContext> leafReaderContexts,
+        final List<Map<Integer, Float>> perLeafResults
+    ) throws IOException {
+        // Construct query
+        List<Callable<TopDocs>> nestedQueryTasks = new ArrayList<>(leafReaderContexts.size());
+        Weight filterWeight = getFilterWeight(indexSearcher);
```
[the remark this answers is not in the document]
heemin32: FilterWeight in `AbstractKnnVectorQuery` class need to be stored in variable and it should be accessible from child class. Then, we can reuse it.

src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java
heemin32: We can refactor when it is really needed. If the PR you referenced is merged first, I will try to reuse the Bits.

src/main/java/org/opensearch/knn/index/query/lucenelib/NestedKnnVectorInnerHitQuery.java
```diff
@@ -0,0 +1,149 @@
...
+
+    @Override
+    public Weight createWeight(IndexSearcher searcher, ScoreMode scoreMode, float boost) throws IOException {
+        Query docAndScoreQuery = internalNestedKnnVectorQuery.knnRewrite(searcher);
+        Weight weight = docAndScoreQuery.createWeight(searcher, scoreMode, boost);
+        IndexReader reader = searcher.getIndexReader();
+        List<LeafReaderContext> leafReaderContexts = reader.leaves();
+        List<Map<Integer, Float>> perLeafResults;
+        perLeafResults = queryUtils.doSearch(searcher, leafReaderContexts, weight);
+        ResultUtil.reduceToTopK(perLeafResults, internalNestedKnnVectorQuery.getK());
```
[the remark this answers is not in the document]
heemin32: Removed the line.

src/main/java/org/opensearch/knn/index/query/ExactSearcher.java
navneet1v: > When I reviewed the code, it seems the main difference between scoreAllDocs and searchTopCandidates is whether a heap is used to maintain the top-k results. Would it make sense to always call searchTopCandidates, considering the latency difference might be negligible?

The difference between both of them in 1 case when heap is used we can have more than k docs and in case no heap, then we don't need heap. The latency difference will not be negligible when done enough number of times. This is the reason why the 2 functions were created because having a heaps add extra latency.

src/main/java/org/opensearch/knn/index/query/ExactSearcher.java
```diff
@@ -233,7 +238,7 @@ public static class ExactSearcherContext {
          */
         boolean useQuantizedVectorsForSearch;
         int k;
-        BitSet matchedDocs;
+        DocIdSetIterator matchedDocs;
```
navneet1v, line 241: probably rename this to `matchedDocsIterator`.

src/main/java/org/opensearch/knn/index/query/ExactSearcher.java
navneet1v: > Alternatively, I can introduce a new class that holds the iterator along with the accurate cost value, if that would be preferable.

If you already have the cost of iterator then I would recommend passing it along in this ExactSearcherContext class, to ensure we find the cost of these iterators/bitset only once.

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -140,14 +142,14 @@ public Map<Integer, Float> searchLeaf(LeafReaderContext context, int k) throws I
...
-            return doExactSearch(context, filterBitSet, k);
+            return doExactSearch(context, new BitSetIterator(filterBitSet, filterBitSet.cardinality()), k);
         }
         Map<Integer, Float> docIdsToScoreMap = doANNSearch(context, filterBitSet, cardinality, k);
         // See whether we have to perform exact search based on approx search results
         // This is required if there are no native engine files or if approximate search returned
         // results less than K, though we have more than k filtered docs
         if (isExactSearchRequire(context, cardinality, docIdsToScoreMap.size())) {
-            final BitSet docs = filterWeight != null ? filterBitSet : null;
+            final BitSetIterator docs = filterWeight != null ? new BitSetIterator(filterBitSet, filterBitSet.cardinality()) : null;
```
navneet1v, line 152: use the `cardinality` variable retrieved in line 132. Multiple calls to cardinality adds extra latency in the flow.

src/main/java/org/opensearch/knn/index/query/common/QueryUtils.java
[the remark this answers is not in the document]
navneet1v: if that is case then lets have QueryUilts class expose its object rather than we creating it in some other class to ensure that this class have a singleton behavior.

src/main/java/org/opensearch/knn/index/query/iterators/GroupedNestedDocIdSetIterator.java
```diff
@@ -0,0 +1,100 @@
...
+        if (currentDocId == NO_MORE_DOCS) {
+            return currentDocId;
+        }
+
+        return slowAdvance(i);
+    }
+
+    @Override
+    public long cost() {
+        if (cost == null) {
```
navneet1v, line 74: do you think it will be better to go with `cost == -1` and use primitives rather than `cost == null`

src/main/java/org/opensearch/knn/index/query/iterators/GroupedNestedDocIdSetIterator.java
```diff
@@ -0,0 +1,100 @@
...
+ *
+ * The {@link #docIds} should include only a single nested document ID per parent document. Otherwise, the nested documents
+ * of that parent document will be iterated multiple times.
+ *
+ * The order of document IDs in {@link #docIds} does not need to be sorted.
+
+ */
+public class GroupedNestedDocIdSetIterator extends DocIdSetIterator {
+    private final BitSet parentBitSet;
+    private final ArrayList<Integer> docIds;
```
navneet1v, line 26: why not `List<Integer>`?

src/main/java/org/opensearch/knn/index/query/common/QueryUtils.java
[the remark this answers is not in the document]
heemin32: Ack

src/main/java/org/opensearch/knn/index/query/ResultUtil.java
heemin32: If we return `null`, it could lead to a risk of NPE wherever we use the iterator. There's another option to avoid the overhead of `cost()` call in this cases. I'll switch to using a static instance and remove the use of `cost()`. Let me know if you still prefer returning `null` after my change.

src/main/java/org/opensearch/knn/index/query/ExactSearcher.java
heemin32: Ack

src/main/java/org/opensearch/knn/index/query/iterators/GroupedNestedDocIdSetIterator.java
heemin32: Other form of list might cause performance degradation because we use `docIds.get(currentIndex)` a lot.

src/main/java/org/opensearch/knn/index/query/ResultUtil.java
heemin32: Hm. Cannot reuse same iterator. Let me switch to null.

src/main/java/org/opensearch/knn/index/query/common/DocAndScoreQuery.java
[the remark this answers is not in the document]
heemin32: @shatejas Please suggest ways we can restructure the package to improve it.

src/main/java/org/opensearch/knn/index/query/KNNQueryFactory.java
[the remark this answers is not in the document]
navneet1v: @shatejas at-least for this PR, I would like us to track and simplify this logic and then make be take another PR for removing the `KNNQuery`. Atleast for now I think if we want to remove the `KNNQuery` it will be a big refactor which is completely out of scope for this PR. Open for suggestions here.

src/main/java/org/opensearch/knn/index/query/ResultUtil.java
navneet1v: > Cannot reuse same iterator. I think it is better for caller to check if result map is empty or not before calling this method instead 

Then whoever is passing the `resultMap` then needs to know that empty check on the map should be done outside. I still think we should go with `null` returns and ensure things are added in java docs. I won't suggest validating `resultMap` as empty in the caller as the `resultMapToDocIds` is a public function, so it needs to handle the resultMap being empty case.

src/main/java/org/opensearch/knn/index/query/KNNQueryFactory.java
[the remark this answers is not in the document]
heemin32: I can create a separate PR for the change if needed, making it easier to revert in case of any unforeseen issues.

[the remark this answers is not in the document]
navneet1v: when you say a separate PR you mean for the simplification of this condition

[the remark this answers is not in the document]
heemin32: Yes

src/main/java/org/opensearch/knn/index/query/ResultUtil.java
jmazanec15: Looking at Lucene, DocIdSetIterator.empty() is used pretty frequently, so I think its okay to use that.

src/main/java/org/opensearch/knn/index/query/lucenelib/NestedKnnVectorQueryFactory.java
```diff
@@ -0,0 +1,77 @@
...
+
+/**
+ * A class to create a nested knn vector query for lucene
+ */
+public class NestedKnnVectorQueryFactory {
+    /**
+     * Create a query for k-NN nested field.
+     *
+     * The query is generated two times when inner_hits() parameter exist in the request.
+     * For inner hit, we return all filtered nested field documents belongs to the final result of parent documents.
```
jmazanec15, line 22: Ive been thinking this over and it does not seem correct. By passing inner_hits, we are altering the behavior of the query. One of the reasons for inner_hits is to understand why certain results from a query are returned (https://opensearch.org/docs/latest/search-plugins/searching-data/inner-hits/#benefits-of-using-inner_hits). By making this the behavior, we are altering the logic and thus compromising the integrity of the feature from a debugging perspective - you could get different results depending on if you pass "inner_hits" or not.

That being said, I think we can expand the other children docs but just mark the scores as null to signal that the results exist, but are not matches. See result of https://opensearch.org/docs/latest/search-plugins/searching-data/inner-hits/#example-inner_hits-parameters-with-a-parent-join-relationship in 3. In this approach, users can do what they want with all of the hits, but does not alter the behavior.

src/main/java/org/opensearch/knn/index/query/ResultUtil.java
shatejas: Would prefer empty over null if there is no performance impact. Its more defensive then passing null around. 

Considering  https://github.com/opensearch-project/k-NN/pull/2283#discussion_r1859878598, there doesn't seem to be a perf impact, I would recommend empty

navneet1v: > Would prefer empty over null if there is no performance impact. Its more defensive then passing null around.

There will be performance impact because in non empty case `.cost()` function don't give you `O(1)` runtime.

src/main/java/org/opensearch/knn/index/query/common/DocAndScoreQuery.java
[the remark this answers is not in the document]
shatejas: For this one I wouldn't want DocsAndScoreQuery to be exposed ideally, I would recommend if there is a way we can have Queries that use it in one package.

As far as structure of the query package goes, it needs a hard look to what fits where, I can come take a stab at it

src/main/java/org/opensearch/knn/index/query/ResultUtil.java
shatejas: I see, performance takes over in that case. if this is forcing us to have cost() and there is a clear impact for worst case scenarios, empty should be rethought

heemin32: As of now, there is no place where we have to call `.cost()` because we are using empty instead of null.

src/main/java/org/opensearch/knn/index/query/lucenelib/NestedKnnVectorQueryFactory.java
jmazanec15: > The main query results will remain unchanged regardless of whether `inner_hits` is included.

Will scores differ?

> Having scores is what users expect. Without scores, `inner_hits` adds limited value.

Yes, but users also expect that the `inner_hits` show how the parent score got formed. So, it is misleading if the result 1 has score = child_A_Score, but child_B_Score > child_A_score. The question would then be why did the parent have a score lower than its greatest child.

jmazanec15: If users want scores of all children, we should introduce some parameter like expand children or add something in the rescore section on children. I dont think we should do it by default unless we change the baseline functionality of the main query behavior

src/main/java/org/opensearch/knn/index/query/common/DocAndScoreQuery.java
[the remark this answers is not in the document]
heemin32: Revert it back to as package scope.

src/main/java/org/opensearch/knn/index/query/lucenelib/NestedKnnVectorQueryFactory.java
heemin32: > Will scores differ?
Score of main query will remain same.

> Yes, but users also expect that the inner_hits show how the parent score got formed. So, it is misleading if the result 1 has score = child_A_Score, but child_B_Score > child_A_score. The question would then be why did the parent have a score lower than its greatest child.

That is valid point and we will document it why it happens. 

> If users want scores of all children, we should introduce some parameter like expand children or add something in the rescore section on children. I dont think we should do it by default unless we change the baseline functionality of the main query behavior

That's a valid point, and I like the idea. Perhaps we could introduce a new parameter in the KNN query to specify whether to expand children or not, ensuring consistent behavior between `inner_hits` and non-`inner_hits` queries? This way, they can also utilize the `score_mode` if they want without breaking backward compatibility.

```
{
  "query": {
    "nested": {
      "path": "nested_field",
      "query": {
        "knn": {
          "nested_field.my_vector": {
            "vector": [
              6,
              6,
              6
            ],
            "k": 2,
            "rescore_with_nested_docs": true
          }
        }
      }
    }
  }
}
```

src/main/java/org/opensearch/knn/index/query/lucenelib/InternalNestedKnnVectorQuery.java
```diff
@@ -0,0 +1,63 @@
...
+     * @return
+     * @throws IOException
+     */
+    TopDocs knnExactSearch(final LeafReaderContext leafReaderContext, final DocIdSetIterator iterator) throws IOException;
+
+    /**
+     * Return a field name
+     * @return field name
+     */
+    String getField();
```
kotwanikunal, line 44: I think I am missing some context here - but both the implementations are missing the `getField()`, `getQuery()`, `getK()` definitions. Are they derived from some other class in the hierarchy?

heemin32: There is `@Getter` lombok annotation.

heemin32:
Rolling upgrade is failing because min version to support new parameter is 2.19 but 2.19(2.x) does not have this code change yet.

src/test/java/org/opensearch/knn/integ/ExpandNestedDocsIT.java
```diff
@@ -0,0 +1,333 @@
...
+        assertEquals(3, docIdToOffsets.keySet().size());
+        assertEquals(2, docIdToOffsets.get(String.valueOf(3)).size());
+        assertEquals(1, docIdToOffsets.get(String.valueOf(4)).size());
+        assertEquals(1, docIdToOffsets.get(String.valueOf(5)).size());
+
+        assertTrue(docIdToOffsets.get(String.valueOf(4)).contains(1));
+        assertTrue(docIdToOffsets.get(String.valueOf(5)).contains(0));
+    }
+
+    private Float[] createVector() {
```
jmazanec15, line 165: Do we need to reimplement this?

jmazanec15 (commented):
@heemin32 Can we add bwc test?

heemin32:
> @heemin32 Can we add bwc test?

What kind of bwc test do you think we need? For the new parameter, it is already covered with existing BWC test.

src/test/java/org/opensearch/knn/integ/ExpandNestedDocsIT.java
heemin32: Why do you think this need to be reimplemented?

jmazanec15:
@heemin32 Oh youre right. Forgot we were not changing defaults. We should be good then

src/test/java/org/opensearch/knn/integ/ExpandNestedDocsIT.java
jmazanec15: I meant we have similar functions that already exist

heemin32: Could you tell me where we have similar function?
For my case, it needs to be data type agnostic which means the value should be byte range and should be able to handle the dimension for both binary and non binary case.

jmazanec15 (dismissed):

navneet1v (commented):
Some code style related comments. Also, should we add a BWC test for the new parameter that we have added in the query?

src/main/java/org/opensearch/knn/index/query/PerLeafResult.java
```diff
@@ -0,0 +1,26 @@
...
+
+import lombok.Getter;
+import lombok.Setter;
+import org.apache.lucene.util.Bits;
+
+import java.util.Collections;
+import java.util.Map;
+
+@Getter
+public class PerLeafResult {
```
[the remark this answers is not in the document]
heemin32: No need to. Since I only defined a constructor with arguments, creating an instance without any arguments is not permitted.

src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java
```diff
@@ -76,101 +77,131 @@ public Weight createWeight(IndexSearcher indexSearcher, ScoreMode scoreMode, flo
             log.debug("Rescoring results took {} ms. oversampled k:{}, segments:{}", rescoreTime, firstPassK, leafReaderContexts.size());
         }
         ResultUtil.reduceToTopK(perLeafResults, finalK);
+
+        if (expandNestedDocs) {
+            perLeafResults = retrieveAll(indexSearcher, leafReaderContexts, knnWeight, perLeafResults, rescoreContext == null);
+        }
```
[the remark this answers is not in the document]
heemin32: This code is not exclusively for rescoring; it will be executed even when there is no rescore context. Refer to the changes in `KNNQueryFactory` at line #114.
```
            } else if (ENGINES_SUPPORTING_MULTI_VECTORS.contains(knnEngine) && expandNested) {
                return new NativeEngineKnnVectorQuery(knnQuery, QueryUtils.INSTANCE, expandNested);
```

src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java
```diff
@@ -76,101 +77,131 @@ public Weight createWeight(IndexSearcher indexSearcher, ScoreMode scoreMode, flo
...
+    ) throws IOException {
+        List<Callable<PerLeafResult>> nestedQueryTasks = new ArrayList<>(leafReaderContexts.size());
+        for (int i = 0; i < perLeafResults.size(); i++) {
+            LeafReaderContext leafReaderContext = leafReaderContexts.get(i);
+            int finalI = i;
+            nestedQueryTasks.add(() -> {
+                PerLeafResult perLeafeResult = perLeafResults.get(finalI);
+                if (perLeafeResult.getResult().isEmpty()) {
+                    return PerLeafResult.EMPTY_RESULT;
+                }
```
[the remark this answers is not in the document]
heemin32: We could, but it wouldn't change anything since they refer to the same instance.

src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java
```diff
@@ -76,101 +77,131 @@ public Weight createWeight(IndexSearcher indexSearcher, ScoreMode scoreMode, flo
...
             rescoreTasks.add(() -> {
-                final BitSet convertedBitSet = ResultUtil.resultMapToMatchBitSet(perLeafResults.get(finalI));
-                // if there is no docIds to re-score from a segment we should return early to ensure that we are not
-                // wasting any computation
-                if (convertedBitSet == null) {
-                    return Collections.emptyMap();
+                PerLeafResult perLeafeResult = perLeafResults.get(finalI);
+                if (perLeafeResult.getResult().isEmpty()) {
+                    return PerLeafResult.EMPTY_RESULT;
                 }
```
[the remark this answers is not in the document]
heemin32: I thought this is easier to read. There is no difference as they are same instance at the end. Let me just return the result.

src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java
```diff
@@ -76,101 +77,131 @@ public Weight createWeight(IndexSearcher indexSearcher, ScoreMode scoreMode, flo
...
+        TopDocs topK = TopDocs.merge(getTotalTopDoc(topDocs), topDocs);
+
         if (topK.scoreDocs.length == 0) {
             return new MatchNoDocsQuery().createWeight(indexSearcher, scoreMode, boost);
         }
-        return createDocAndScoreQuery(reader, topK).createWeight(indexSearcher, scoreMode, boost);
+        return queryUtils.createDocAndScoreQuery(reader, topK).createWeight(indexSearcher, scoreMode, boost);
+    }
+
+    private int getTotalTopDoc(TopDocs[] topDocs) {
```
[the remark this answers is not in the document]
heemin32: The maximum number of documents a shard can have is Integer.MAX_VALUE. Therefore the max value the method will return will be Integer.MAX_VALUE. I will add some doc.

src/main/java/org/opensearch/knn/index/query/parser/KNNQueryBuilderParser.java
```diff
@@ -128,6 +132,10 @@ public static KNNQueryBuilder.Builder streamInput(StreamInput in, Function<Strin
             builder.rescoreContext(RescoreParser.streamInput(in));
         }
 
+        if (minClusterVersionCheck.apply(EXPAND_NESTED)) {
+            builder.expandNested(in.readBoolean());
+        }
```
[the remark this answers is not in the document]
heemin32: I will do that, but I assume it won't block this PR? I can follow up there after creating an issue in neural search plugin.

navneet1v (dismissed):
Code looks good to me. Just check this thing, since we are not returning all the child documents of the parent docs, will this results into same behavior where if 1 parent child docs are better than other parent child docs, will Opensearch returns just 1 parent doc to customer or it will return 2 parent docs to customers.

jmazanec15 (approved):

heemin32:
> Code looks good to me. Just check this thing, since we are not returning all the child documents of the parent docs, will this results into same behavior where if 1 parent child docs are better than other parent child docs, will Opensearch returns just 1 parent doc to customer or it will return 2 parent docs to customers.

Even when multiple nested documents are returned per parent document, they are joined back to the parent document, ensuring that the final parent document count remains unaffected. It has been confirmed that, in such cases, the result will still include 2 parent documents.

### Create Index With 2 shards
```
PUT /my-knn-index-1
{
  "settings": {
    "index": {
      "knn": true,
      "number_of_shards": 2
    }
  },
  "mappings": {
    "properties": {
      "nested_field": {
        "type": "nested",
        "properties": {
          "my_vector": {
            "type": "knn_vector",
            "dimension": 3,
            "space_type": "l2",
            "method": {
              "name": "hnsw",
              "engine": "faiss"
            }
          }
        }
      }
    }
  }
}
```

### Ingest 4 documents with 10 nested docs per each
```
PUT /_bulk?refresh=true
{ "index": { "_index": "my-knn-index-1", "_id": "1" } }
{"nested_field":[{"my_vector":[1,1,1]},{"my_vector":[1,1,1]},{"my_vector":[1,1,1]},{"my_vector":[1,1,1]},{"my_vector":[1,1,1]},{"my_vector":[1,1,1]},{"my_vector":[1,1,1]},{"my_vector":[1,1,1]},{"my_vector":[1,1,1]},{"my_vector":[1,1,1]}]}
{ "index": { "_index": "my-knn-index-1", "_id": "2" } }
{"nested_field":[{"my_vector":[10,10,10]},{"my_vector":[10,10,10]},{"my_vector":[10,10,10]},{"my_vector":[10,10,10]},{"my_vector":[10,10,10]},{"my_vector":[10,10,10]},{"my_vector":[10,10,10]},{"my_vector":[10,10,10]},{"my_vector":[10,10,10]},{"my_vector":[10,10,10]}]}
{ "index": { "_index": "my-knn-index-1", "_id": "3" } }
{"nested_field":[{"my_vector":[100,100,100]},{"my_vector":[100,100,100]},{"my_vector":[100,100,100]},{"my_vector":[100,100,100]},{"my_vector":[100,100,100]},{"my_vector":[100,100,100]},{"my_vector":[100,100,100]},{"my_vector":[100,100,100]},{"my_vector":[100,100,100]},{"my_vector":[100,100,100]}]}
{ "index": { "_index": "my-knn-index-1", "_id": "4" } }
{"nested_field":[{"my_vector":[1000,1000,1000]},{"my_vector":[1000,1000,1000]},{"my_vector":[1000,1000,1000]},{"my_vector":[1000,1000,1000]},{"my_vector":[1000,1000,1000]},{"my_vector":[1000,1000,1000]},{"my_vector":[1000,1000,1000]},{"my_vector":[1000,1000,1000]},{"my_vector":[1000,1000,1000]},{"my_vector":[1000,1000,1000]}]}
```

### Check documents are distributed across 2 shards
```
GET /_cat/shards/my-knn-index-1
my-knn-index-1 0 p STARTED    33 4.8kb 127.0.0.1 integTest-0
my-knn-index-1 0 r UNASSIGNED                    
my-knn-index-1 1 p STARTED    11 4.5kb 127.0.0.1 integTest-0
my-knn-index-1 1 r UNASSIGNED                    
```

### Search
```
GET /my-knn-index-1/_search
{
  "_source": false,
  "query": {
    "nested": {
      "path": "nested_field",
      "query": {
        "knn": {
          "nested_field.my_vector": {
            "vector": [
              10,
              10,
              10
            ],
            "k": 4,
            "expand_nested_docs": true
          }
        }
      },
      "score_mode": "max"
    }
  }
}
```
### Result
Confirmed that 4 result is returned properly.
```
{
  "took": 15,
  "timed_out": false,
  "_shards": {
    "total": 2,
    "successful": 2,
    "skipped": 0,
    "failed": 0
  },
  "hits": {
    "total": {
      "value": 4,
      "relation": "eq"
    },
    "max_score": 1.0,
    "hits": [
      {
        "_index": "my-knn-index-1",
        "_id": "2",
        "_score": 1.0
      },
      {
        "_index": "my-knn-index-1",
        "_id": "1",
        "_score": 0.0040983604
      },
      {
        "_index": "my-knn-index-1",
        "_id": "3",
        "_score": 4.115057E-5
      },
      {
        "_index": "my-knn-index-1",
        "_id": "4",
        "_score": 3.4010122E-7
      }
    ]
  }
}
```

navneet1v:
@heemin32 thanks for confirming. Do we have a similar IT test added? also were you able to figure out in the code where this translation of child to parent docs is happening and ensuring that are picking up the parent docs == size only.

heemin32:
> @heemin32 thanks for confirming. Do we have a similar IT test added? also were you able to figure out in the code where this translation of child to parent docs is happening and ensuring that are picking up the parent docs == size only.

Let me add one. There was a hidden bug that I missed as well. The translation of child to parent docs is happening in `OpenSearchToParentBlockJoinQuery` -> `ToParentBlockJoinQuery`.

navneet1v (approved):

jmazanec15 (approved):

heemin32:
The benchmark results do not indicate a significant difference in latency when "expand_nested_docs" is enabled.
### Setup
Screenshot 2025-01-08 at 9 48 52 AM

### Result
Screenshot 2025-01-08 at 9 53 57 AM

PR #2302, opened 2024-12-03
Fixed a bug that does not get a space type from index setting when its value is empty for compatibility.

0ctopus13prime:
### Description

### Related Issues

### Check List


0ctopus13prime:
PR for main branch - https://github.com/opensearch-project/k-NN/pull/2309

PR #2305, opened 2024-12-03
Adds in lazy execution for Lucene kNN queries

### LINKED ISSUE #2115: [FEATURE] Improving Lucene Engine Query Performance by reducing number of times a single Lucene k-NN query gets executed
## Description
Currently Lucene Engine KNN queries get executed during the [re-write phase](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java#L70) and not in the Weight class. On recent deep-dive we observed that rewrite function of a query can be called multiple times in the overall search flow. 

Please check this code trace on rewrite running before start of fetch phase. 
1. Transport action registering fetchphase: [https://github.com/opensearch-project/OpenSearch/blob/f67ed1796749376401c5cc617eff[…]n/java/org/opensearch/action/search/SearchTransportService.java](https://github.com/opensearch-project/OpenSearch/blob/f67ed1796749376401c5cc617eff2b7903325a9d/server/src/main/java/org/opensearch/action/search/SearchTransportService.java#L647-L661)
2. Search Context is built: [https://github.com/opensearch-project/OpenSearch/blob/8148e4d295397d4f5b50b85f4dc3[…]6/server/src/main/java/org/opensearch/search/SearchService.java](https://github.com/opensearch-project/OpenSearch/blob/8148e4d295397d4f5b50b85f4dc358ac70438fe6/server/src/main/java/org/opensearch/search/SearchService.java#L835)
3. In building the searchcontext we parse the source again: [https://github.com/opensearch-project/OpenSearch/blob/8148e4d295397d4f5b50b85f4dc3[…]6/server/src/main/java/org/opensearch/search/SearchService.java](https://github.com/opensearch-project/OpenSearch/blob/8148e4d295397d4f5b50b85f4dc358ac70438fe6/server/src/main/java/org/opensearch/search/SearchService.java#L1064) (this is already done in query phase)
4. During the create context we do the preProcess: [https://github.com/opensearch-project/OpenSearch/blob/8148e4d295397d4f5b50b85f4dc3[…]6/server/src/main/java/org/opensearch/search/SearchService.java](https://github.com/opensearch-project/OpenSearch/blob/8148e4d295397d4f5b50b85f4dc358ac70438fe6/server/src/main/java/org/opensearch/search/SearchService.java#L1076)
5. In preprocess we do context preprocess: [https://github.com/opensearch-project/OpenSearch/blob/4035db48c6963e46909f28ba8552[…]erver/src/main/java/org/opensearch/search/query/QueryPhase.java](https://github.com/opensearch-project/OpenSearch/blob/4035db48c6963e46909f28ba8552365c10c5762a/server/src/main/java/org/opensearch/search/query/QueryPhase.java#L127)
6. And then we do rewrite again: [https://github.com/opensearch-project/OpenSearch/blob/4f97fd3d5588f9be52bee37d2c51[…]r/src/main/java/org/opensearch/search/DefaultSearchContext.java](https://github.com/opensearch-project/OpenSearch/blob/4f97fd3d5588f9be52bee37d2c51c5397a033348/server/src/main/java/org/opensearch/search/DefaultSearchContext.java#L363-L369)

The same was observed in the flame graphs, where when we have more than 1 shard, during fetch phase the rewrite on the query is called again. This leads to running of the Lucene engine k-NN query more than 1 and adds latency.

**_Flame Graph_**
Lucene Query_2shards_1M_128D

Number of shards: 2
KNN engine: Lucene
Dataset: 1M 128D sift.
Tool used: OSB
Docker image: opensearchstaging/opensearch:2.17.0.10284 
Heap: 16GB
RAM: 64GB
Cores: 16
Search JFR: [search.jfr.zip](https://github.com/user-attachments/files/17072724/search.jfr.zip)

### Why we need Query and its rewrite in fetch phase
A quick scan of the opensearch core code in fetch phase I found use cases that might require to run the query rewrite again and then use it during fetch phase. Below are the references where query which was rewritten in the DefaultSearchContext during FetchPhase is added to [FetchPhaseSearchContext](https://github.com/opensearch-project/OpenSearch/blob/af2e109310d0d07496fc04d8b92b4be06b7f9518/server/src/main/java/org/opensearch/search/fetch/FetchContext.java#L114-L116) and was used.
[Explain Sub phase](https://github.com/opensearch-project/OpenSearch/blob/bc3c606938c7188b5b96c4036a11126a7159d2b0/server/src/main/java/org/opensearch/search/fetch/subphase/ExplainPhase.java#L64): This is used to provide the explanation why a particular result is in the query.
[PercolateQuery Highlighting](https://github.com/opensearch-project/OpenSearch/blob/c687116724575d8b068494db80bb6d5845f4f1de/modules/percolator/src/main/java/org/opensearch/percolator/PercolatorHighlightSubFetchPhase.java#L73): No idea what this query type is but it does use some visitor pattern of the Query(Lucene interface) to do something.
[Inner Hits](https://github.com/opensearch-project/OpenSearch/blob/af2e109310d0d07496fc04d8b92b4be06b7f9518/server/src/main/java/org/opensearch/search/fetch/subphase/InnerHitsContext.java#L131-L138): This is used to get the child hits when there is a parent child relationship between fields. Not sure about this use case as it is actually doing some more funky logic on query. This needs more deep-dive

### Solution 1
One solution we can explore is by wrapping all the Lucene queries in another QueryClause lets say LuceneEngineKNNQuery and have a class member in this class which actual Lucene query. Now when createWeight is called we can first rewrite the query and then call the scorer on top of it. This will ensure that Lucene k-NN query is executed only once.

#### Sample Code:
```
public class LuceneEngineKNNQuery extends Query {

    Query luceneQuery; // this can be KnnFloatVectorQuery, KnnByteVectorQuery, DiversifyingChildrenByteKnnVectorQuery etc.
    
    @Override
        // rewrite the lucene query
        Query docAndScoreQuery = searcher.rewrite(luceneQuery);
        // now return the weight class
        return docAndScoreQuery.createWeight(searcher, scoreMode, boost);
    }

    
    @Override
    public String toString(String field) {
        return "";
    }

    
    @Override
    public void visit(QueryVisitor visitor) {

    }

    
    @Override
    public boolean equals(Object obj) {
        return false;
    }

    
    @Override
    public int hashCode() {
        return 0;
    }
}
```

### Solution 2
Another solution we can implement by caching the SearchContext at a shard level, and when fetch phase is executed we use the same SearchContext so that we don't need to rewrite the queries.
Another approach can be we can defer the rewrite and make it lazy so that only Fetch Pre-Processors that needs re-write should do the rewrite, and once it is done by a Fetch Processor then none of the other will need to run the re-write again as the query has changed now.

## Pros and Cons
Both solution has its own pros and cons. Like
1. Solution 1, solves the problem without making drastic changes in core, but may create problem if Lucene query execution changes from rewrite.
2. Solution 2, Implementing a shard level SearchContext Cache may be tricky as it can increase the heap and what could be a blast radius is unknown. But this solution will ensure that rewrite is happening once for the query in the whole search request at Opensearch level.

kotwanikunal:
### Description
- Optimizes Lucene kNN query execution path in case of rewrites
- The caveat here is that the rewrites would be deferred until weight creation. Any underlying changes to Lucene `rewrite` method would potentially be skipped with this execution mechanism. 

### Setup and Dataset
- Opensearch docker container on a M1 Macbook Pro - Image ID: [e1d9151cce32](https://hub.docker.com/layers/opensearchproject/opensearch/latest/images/sha256-e1d9151cce329062fdb68e3bd74072df9ab8b97ce5e803cb53576269cb02c0a6?context=explore)
- Single node, default config
- Load Gen: OSB vectorsearch workload running on local node hosting the docker container
- Dataset: sift-128-euclidean.hdf5 (Params: https://github.com/opensearch-project/opensearch-benchmark-workloads/blob/main/vectorsearch/params/faiss-sift-128-l2.json)
- Num docs: 1M, num shards: 2

### Flame Graph

Default 2.18 code
2 18-default-code

Patched 2.18 code
2 18-patched-code

### Benchmarks
 2.18 Default Code:
 |                                                         Metric |           Task |      Value |   Unit |
|---------------------------------------------------------------:|---------------:|-----------:|-------:|
|                                                 Min Throughput |   prod-queries |      20.93 |  ops/s |
|                                                Mean Throughput |   prod-queries |      63.23 |  ops/s |
|                                              Median Throughput |   prod-queries |      69.54 |  ops/s |
|                                                 Max Throughput |   prod-queries |      75.17 |  ops/s |
|                                        50th percentile latency |   prod-queries |    10.6464 |     ms |
|                                        90th percentile latency |   prod-queries |    13.7257 |     ms |
|                                        99th percentile latency |   prod-queries |    21.3663 |     ms |
|                                      99.9th percentile latency |   prod-queries |     71.429 |     ms |
|                                       100th percentile latency |   prod-queries |    269.424 |     ms |
|                                   50th percentile service time |   prod-queries |    10.6464 |     ms |
|                                   90th percentile service time |   prod-queries |    13.7257 |     ms |
|                                   99th percentile service time |   prod-queries |    21.3663 |     ms |
|                                 99.9th percentile service time |   prod-queries |     71.429 |     ms |
|                                  100th percentile service time |   prod-queries |    269.424 |     ms |
|                                                     error rate |   prod-queries |          0 |      % |
|                                                  Mean recall@k |   prod-queries |       0.97 |        |
|                                                  Mean recall@1 |   prod-queries |       0.98 |        |
 
 
Patched new 2.18 code:

|                                                         Metric |           Task |      Value |   Unit |
|---------------------------------------------------------------:|---------------:|-----------:|-------:| 
|                                                 Min Throughput |   prod-queries |      75.97 |  ops/s |
|                                                Mean Throughput |   prod-queries |      79.29 |  ops/s |
|                                              Median Throughput |   prod-queries |      78.68 |  ops/s |
|                                                 Max Throughput |   prod-queries |      83.04 |  ops/s |
|                                        50th percentile latency |   prod-queries |    8.66315 |     ms |
|                                        90th percentile latency |   prod-queries |    11.4286 |     ms |
|                                        99th percentile latency |   prod-queries |    21.4076 |     ms |
|                                      99.9th percentile latency |   prod-queries |    34.8355 |     ms |
|                                       100th percentile latency |   prod-queries |     53.219 |     ms |
|                                   50th percentile service time |   prod-queries |    8.66315 |     ms |
|                                   90th percentile service time |   prod-queries |    11.4286 |     ms |
|                                   99th percentile service time |   prod-queries |    21.4076 |     ms |
|                                 99.9th percentile service time |   prod-queries |    34.8355 |     ms |
|                                  100th percentile service time |   prod-queries |     53.219 |     ms |
|                                                     error rate |   prod-queries |          0 |      % |
|                                                  Mean recall@k |   prod-queries |       0.97 |        |
|                                                  Mean recall@1 |   prod-queries |       0.98 |        |

### JFRs
[JFR-patched-default.zip](https://github.com/user-attachments/files/18013154/JFR-patched-default.zip)

### Related Issues
Resolves #2115 

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/query/lucene/LuceneEngineKnnVectorQuery.java
+++ b/src/main/java/org/opensearch/knn/index/query/lucene/LuceneEngineKnnVectorQuery.java
@@ -1,0 +1,69 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.knn.index.query.lucene;
+
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+import lombok.extern.log4j.Log4j2;
+import org.apache.lucene.search.IndexSearcher;
+import org.apache.lucene.search.Query;
+import org.apache.lucene.search.QueryVisitor;
+import org.apache.lucene.search.ScoreMode;
+import org.apache.lucene.search.Weight;
+
+import java.io.IOException;
+
+/**
+ * LuceneEngineKnnVectorQuery is a wrapper around a vector queries for the Lucene engine.
+ * This enables us to defer rewrites until weight creation to optimize repeated execution
+ * of Lucene based k-NN queries.
+ */
+@AllArgsConstructor
+@Log4j2
+public class LuceneEngineKnnVectorQuery extends Query {
+    @Getter
+    private final Query luceneQuery;
+
+    /*
+      Prevents repeated rewrites of the query for the Lucene engine.
+    */
+    @Override
+    public Query rewrite(IndexSearcher indexSearcher) {
+        return this;
+    }
+
+    /*
+       Rewrites the query just before weight creation.
+     */
+    @Override
+    public Weight createWeight(IndexSearcher searcher, ScoreMode scoreMode, float boost) throws IOException {
+        Query rewrittenQuery = luceneQuery.rewrite(searcher);
+        return rewrittenQuery.createWeight(searcher, scoreMode, boost);
+    }
+
+    @Override
+    public String toString(String s) {
+        return luceneQuery.toString();
+    }
+
+    @Override
+    public void visit(QueryVisitor queryVisitor) {
+        queryVisitor.visitLeaf(this);
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        LuceneEngineKnnVectorQuery otherQuery = (LuceneEngineKnnVectorQuery) o;
+        return luceneQuery.equals(otherQuery.luceneQuery);
+    }
+
+    @Override
+    public int hashCode() {
+        return luceneQuery.hashCode();
+    }
+}
--- a/src/test/java/org/opensearch/knn/index/query/KNNQueryBuilderTests.java
+++ b/src/test/java/org/opensearch/knn/index/query/KNNQueryBuilderTests.java
@@ -8,7 +8,6 @@ package org.opensearch.knn.index.query;
 import com.google.common.collect.ImmutableMap;
 import lombok.SneakyThrows;
 import org.apache.lucene.search.FloatVectorSimilarityQuery;
-import org.apache.lucene.search.KnnFloatVectorQuery;
 import org.apache.lucene.search.MatchNoDocsQuery;
 import org.apache.lucene.search.Query;
 import org.junit.Before;
@@ -33,6 +32,7 @@ import org.opensearch.knn.index.engine.qframe.QuantizationConfig;
 import org.opensearch.knn.index.mapper.KNNMappingConfig;
 import org.opensearch.knn.index.mapper.KNNVectorFieldType;
 import org.opensearch.knn.index.mapper.Mode;
+import org.opensearch.knn.index.query.lucene.LuceneEngineKnnVectorQuery;
 import org.opensearch.knn.index.query.rescore.RescoreContext;
 import org.opensearch.knn.index.util.KNNClusterUtil;
 import org.opensearch.knn.index.engine.KNNMethodContext;
@@ -512,7 +512,7 @@ public class KNNQueryBuilderTests extends KNNTestCase {
 
         // Then
         assertNotNull(query);
-        assertTrue(query.getClass().isAssignableFrom(KnnFloatVectorQuery.class));
+        assertTrue(query.getClass().isAssignableFrom(LuceneEngineKnnVectorQuery.class));
     }
 
     @SneakyThrows
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Enhancements
 - Introduced a writing layer in native engines where relies on the writing interface to process IO. (#2241)[https://github.com/opensearch-project/k-NN/pull/2241]
 - Allow method parameter override for training based indices (#2290) https://github.com/opensearch-project/k-NN/pull/2290]
+- Optimizes lucene query execution to prevent unnecessary rewrites (#2305)[https://github.com/opensearch-project/k-NN/pull/2305]
 ### Bug Fixes
 * Fixing the bug when a segment has no vector field present for disk based vector search (#2282)[https://github.com/opensearch-project/k-NN/pull/2282]
 ### Infrastructure
--- a/src/main/java/org/opensearch/knn/index/query/KNNQueryFactory.java
+++ b/src/main/java/org/opensearch/knn/index/query/KNNQueryFactory.java
@@ -15,6 +15,7 @@ import org.opensearch.knn.index.VectorDataType;
 import org.opensearch.knn.index.engine.KNNEngine;
 import org.opensearch.knn.index.query.common.QueryUtils;
 import org.opensearch.knn.index.query.lucenelib.NestedKnnVectorQueryFactory;
+import org.opensearch.knn.index.query.lucene.LuceneEngineKnnVectorQuery;
 import org.opensearch.knn.index.query.nativelib.NativeEngineKnnVectorQuery;
 import org.opensearch.knn.index.query.rescore.RescoreContext;
 
@@ -128,9 +129,13 @@ public class KNNQueryFactory extends BaseQueryFactory {
         log.debug(String.format("Creating Lucene k-NN query for index: %s \"\", field: %s \"\", k: %d", indexName, fieldName, k));
         switch (vectorDataType) {
             case BYTE:
-                return getKnnByteVectorQuery(fieldName, byteVector, luceneK, filterQuery, parentFilter, expandNested);
+                return new LuceneEngineKnnVectorQuery(
+                    getKnnByteVectorQuery(fieldName, byteVector, luceneK, filterQuery, parentFilter, expandNested)
+                );
             case FLOAT:
-                return getKnnFloatVectorQuery(fieldName, vector, luceneK, filterQuery, parentFilter, expandNested);
+                return new LuceneEngineKnnVectorQuery(
+                    getKnnFloatVectorQuery(fieldName, vector, luceneK, filterQuery, parentFilter, expandNested)
+                );
             default:
                 throw new IllegalArgumentException(
                     String.format(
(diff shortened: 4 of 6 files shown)


navneet1v:
@kotwanikunal can you fix the CIs. Change looks good to me as it is minial in terms of changing the code and getting the best performance out of the system

kotwanikunal:
> @kotwanikunal can you fix the CIs. Change looks good to me as it is minial in terms of changing the code and getting the best performance out of the system

@navneet1v I have updated the description with updated flame graphs and benchmarking numbers.
There are some fixes for the CI - I'll fix them shortly.

navneet1v:
@kotwanikunal thanks. This p99 seems interesting. Rest all latency have decreased as expected.

navneet1v:
@kotwanikunal can you please add the details like:
1. Setup.
2. Dataset used
for the benchmarks

kotwanikunal:
> @kotwanikunal can you please add the details like:
> 
> 1. Setup.
> 2. Dataset used
>    for the benchmarks

Added to the description.

navneet1v (approved):
overall code looks good to me. One question I have is why we are not changing the `DiversifyingChildrenByteKnnVectorQuery` and `DiversifyingChildrenFloatKnnVectorQuery` in this PR? Is there some other PR that is taking care of this?

@kotwanikunal

navneet1v:
> overall code looks good to me. One question I have is why we are not changing the `DiversifyingChildrenByteKnnVectorQuery` and `DiversifyingChildrenFloatKnnVectorQuery` in this PR? Is there some other PR that is taking care of this?
> 
> @kotwanikunal

NVM, I see this PR: https://github.com/opensearch-project/k-NN/pull/2283/files#diff-d6bb9cc296becc98be3ac4bddc74c2b1520d09ff8b7d77467b3a765df17c4c11 is handling that case.

src/test/java/org/opensearch/knn/index/query/KNNQueryBuilderTests.java
```diff
@@ -512,7 +512,7 @@ public void testDoToQuery_KnnQueryWithFilter_Lucene() throws Exception {
 
         // Then
         assertNotNull(query);
-        assertTrue(query.getClass().isAssignableFrom(KnnFloatVectorQuery.class));
+        assertTrue(query.getClass().isAssignableFrom(LuceneEngineKnnVectorQuery.class));
```
martin-gaievski, line 515: is there a way to check the type of the underlying lucene query, looks like this is what we're asserting in today's version? How about adding a package private getter for the lucene query, or a public method that returns the type of the query class.

src/main/java/org/opensearch/knn/index/query/lucene/LuceneEngineKnnVectorQuery.java
```diff
@@ -0,0 +1,67 @@
...
+       Rewrites the query just before weight creation.
+     */
+    @Override
+    public Weight createWeight(IndexSearcher searcher, ScoreMode scoreMode, float boost) throws IOException {
+        Query rewrittenQuery = luceneQuery.rewrite(searcher);
+        return rewrittenQuery.createWeight(searcher, scoreMode, boost);
+    }
+
+    @Override
+    public String toString(String s) {
```
martin-gaievski, line 46: some of methods from java Object can be generated by Lombok annotations, e.g. @ToString and @@EqualsAndHashCode

martin-gaievski (commented):
Overall code looks good, added few minor comments. 
Did you see impact of efficient filtering with Lucene, can you run a benchmark to check if there is an impact?

src/main/java/org/opensearch/knn/index/query/lucene/LuceneEngineKnnVectorQuery.java
navneet1v: For query class we should just use the delegation approach here rather than using lombok. That would be my recommendation.

martin-gaievski: ah, makes sense, delegating here makes more sense

martin-gaievski (approved):
Please address my comments in follow up PRs

navneet1v:
Conflicts fixed. Will wait for CI to complete

navneet1v (approved):

navneet1v:
The BWC test: [k-NN Rolling-Upgrade BWC Tests (21, ubuntu-latest, 2.19.0-SNAPSHOT, 3.0.0-SNAPSHOT)](https://github.com/opensearch-project/k-NN/actions/runs/12284150324/job/34279260841?pr=2305#logs) will fixed once we have code for innerhits merges and added in 2.19 snapshot.

navneet1v (dismissed):

martin-gaievski (approved):

PR #36, opened 2021-06-03
Add Google tests for unit testing jni

jmazanec15:
### Description
This PR adds google tests for testing the nmslib_wrapper and faiss_wrapper. Additionally, I made some minor stylistic changes to the jni to help testing and clean up.  The primary purpose of these tests is to be able to identify potential memory leaks.

In order to mock the JNIUtilInterface, this PR adds MockJNIUtil class built off of gmock. In the constructor of this class, the default behavior is defined. This can be overriden inside individual test cases.

While the tests cover each library wrapper function, they do not yet exercise all possible functionality. In order to keep the scope of this PR focused, I will leave expansion to a future PR.

In order to build tests:
```
// This will produce the test executable in bin/jni_test
cd jni
cmake .
make
```

To run the tests:
```
// Run all tests
bin/jni_test

// Run certain tests
bin/jni_test --gtest_filter=<Test filter name (can include *)>

// Run tests a certain number of times
bin/jni_test --gtest_repeat=X
```

In order to test memory checks, I ran the tests with valgrind. Here are some of the results:
```
>> valgrind bin/jni_test
...
==28779== LEAK SUMMARY:
==28779==    definitely lost: 0 bytes in 0 blocks
==28779==    indirectly lost: 0 bytes in 0 blocks
==28779==      possibly lost: 336 bytes in 1 blocks
==28779==    still reachable: 124,880 bytes in 10 blocks

>> valgrind bin/jni_test --gtest_repeat=4
...
==28790== LEAK SUMMARY:
==28790==    definitely lost: 0 bytes in 0 blocks
==28790==    indirectly lost: 0 bytes in 0 blocks
==28790==      possibly lost: 336 bytes in 1 blocks
==28790==    still reachable: 124,880 bytes in 10 blocks

>> valgrind bin/jni_test --gtest_filter=Nmslib*
...
==28865== LEAK SUMMARY:
==28865==    definitely lost: 0 bytes in 0 blocks
==28865==    indirectly lost: 0 bytes in 0 blocks
==28865==      possibly lost: 0 bytes in 0 blocks
==28865==    still reachable: 122,888 bytes in 7 blocks

>> valgrind bin/jni_test --gtest_filter=NmslibInit*
...
==28875== LEAK SUMMARY:
==28875==    definitely lost: 0 bytes in 0 blocks
==28875==    indirectly lost: 0 bytes in 0 blocks
==28875==      possibly lost: 0 bytes in 0 blocks
[...]

PR #2308, opened 2024-12-03
Add thread to periodically perform pending cache maintenance

### LINKED ISSUE #2239: Perform cache maintenance in a separate thread
## Description

For the NativeMemoryCacheManager, we allow users to enable an "expirary" time setting, so that cache entries that arent touched for some duration, get marked as expired and are then evicted.

However, the problem is that the guava cache does not maintain a separate thread that can perform this maintenance. Instead it will typically perform this maintenance on writes and sometimes on reads. See https://github.com/google/guava/wiki/CachesExplained#timed-eviction. Thus, if you are not touching the cache for awhile, entries may be expired but not get evicted. Thus, resources are not freed.

The Guava team recommends creating a separate thread that will call cache.cleanUp() to explicitly perform maintenance. This will ensure that the time-based evictions happen as one would expect.

owenhalpert:
### Description
Adds a ScheduledExecutor class that, for each cache, takes in a Runnable with a call to `cleanUp` and periodically executes according to the values of `KNN_CACHE_ITEM_EXPIRY_TIME_MINUTES` and `QUANTIZATION_STATE_CACHE_EXPIRY_TIME_MINUTES`. This will perform any pending maintenance (such as evicting expired entries) which was previously only performed when the cache was accessed. The maintenance thread is created whenever a NativeMemoryCache or QuantizationStateCache is instantiated or rebuilt and can be shut down with either class's `close` method. Relevant logic for cleanup was added to some testing base classes.

### Related Issues
Resolves #2239 

### Check List


kotwanikunal:
@owenhalpert The changes look good in general. What I'd be interested in is testing under load for a resource constrained system - can you verify if this adds to the latency or impact the performance in any way?

We did implement a force evict before writes with #2015. Can you also enable this feature flag and run the above tests to ensure it behaves well?

src/main/java/org/opensearch/knn/index/CacheMaintainer.java
```diff
@@ -0,0 +1,42 @@
...
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Performs periodic maintenance for a Guava cache. The Guava cache is implemented in a way that maintenance operations (such as evicting expired
+ * entries) will only occur when the cache is accessed. See {@see <a href="https://github.com/google/guava/wiki/cachesexplained#timed-eviction"> Guava Cache Guide</a>}
+ * for more details. Thus, to perform any pending maintenance, the cleanUp method will be called periodically from a CacheMaintainer instance.
+ */
+public class CacheMaintainer<K, V> implements Closeable {
+    private final Cache<K, V> cache;
```
kotwanikunal, line 21: You can also avoid maintaining the Cache object reference here by using a functional interface. That would also get rid of the generification of this class.

Simply pass and store the runnable reference instead of the cache as `new CacheMaintainer(() -> cache.cleanUp());` 

Possibly also move this class to the util package and call it a `ScheduledExecutor` with Runnable ref and interval as parameters.

```code
public class ScheduledExecutor implements Closeable {

...

  public ScheduledExecutor(Runnable reference, long scheduleMillis) {
  ...
  }

  ...
}
```

owenhalpert: @kotwanikunal what do you think about creating the executor and calling scheduleAtFixedRate within each cache class instead of creating a new ScheduledExecutor class?

owenhalpert:
@kotwanikunal I've completed the benchmarking on a single node cluster limited to 3GB of memory. 

Before benchmarking my code, I validated the CacheMaintainer was actually running by inspecting the OpenSearch process status in the Docker container and found that after about a minute, the `RssAnon` value would decrease by about 0.5gb without any interference on my part, signaling the maintainer successfully cleaned up the expired entries. This is aligned with what I saw on the standard maintenance on the clean code (which I triggered manually by accessing the cache). 

Results below:

> Performance Summary for Resource-Constrained Testing (3GB Memory Limit)
> 
> **Run 1: Clean 2.18 code**
> 
> * Indexing:
>     * p50 latency: 16.96 ms
>     * p90 latency: 32.47 ms
> * Search:
>     * p50 latency: 330.48 ms
>     * p90 latency: 437.78 ms
> 
> **Run 2: Clean 2.18 code, Force evict ON**
> 
> * Indexing:
>     * p50 latency: 17.27 ms
>     * p90 latency 33.82 ms
> * Search:
>     * p50 latency 337.43 ms
>     * p90 latency 428.70 ms
> 
> **Run 3: PR changes added, Force evict ON:**
> 
> * Indexing:
>     * p50 latency: 16.89 ms
>     * p90 latency: 32.24 ms
> * Search:
>     * p50 latency: 341.38 ms
>     * p90 latency: 414.06 ms
> 
> **Run 4: PR changes added, Force evict OFF:**
> 
> * Indexing:
>     * p50 latency: 17.43 ms
>     * p90 latency: 34.63 ms
> * Search:
>     * p50 latency: 346.83 ms
>     * p90 latency: 430.04 ms

This suggests there is **no significant impact on latency** with my code changes. I've included the full results of these test runs here:

https://gist.github.com/owenhalpert/05ad4f5ae9577f717f2c59f2039d52e4

src/main/java/org/opensearch/knn/quantization/models/quantizationState/QuantizationStateCache.java
```diff
@@ -71,6 +78,8 @@ private void buildCache() {
             )
             .removalListener(this::onRemoval)
             .build();
+
+        this.cacheMaintainer = new ScheduledExecutor(() -> cache.cleanUp(), 60 * 1000);
```
0ctopus13prime, line 82: Could you consider to use `KNN_CACHE_ITEM_EXPIRY_TIME_MINUTES`?
By using the setting value, we can expect that max retention time after time expiration will be `KNN_CACHE_ITEM_EXPIRY_TIME_MINUTES` minutes, but average retention time would be half of it.

src/main/java/org/opensearch/knn/index/util/ScheduledExecutor.java
```diff
@@ -0,0 +1,34 @@
...
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Executes a task periodically
+ */
+public class ScheduledExecutor implements Closeable {
+    final ScheduledExecutorService executor;
+    public final Runnable task;
```
0ctopus13prime, line 18: Let's hide this and make a getter.

src/main/java/org/opensearch/knn/index/memory/NativeMemoryCacheManager.java
```diff
@@ -99,6 +105,7 @@ private void initialize(NativeMemoryCacheManagerDto nativeMemoryCacheDTO) {
 
         if (nativeMemoryCacheDTO.isExpirationLimited()) {
             cacheBuilder.expireAfterAccess(nativeMemoryCacheDTO.getExpiryTimeInMin(), TimeUnit.MINUTES);
+            this.cacheMaintainer = new ScheduledExecutor(() -> cache.cleanUp(), 60 * 1000);
```
0ctopus13prime, line 108: Could you wrap `cache.cleanUp` with try-catch? 
A task throwing an exception will never be rescheduled.

src/main/java/org/opensearch/knn/quantization/models/quantizationState/QuantizationStateCache.java
0ctopus13prime: Also could you wrap `cache.cleanUp` with try-catch to make sure not throwing an exception?
Once it throws an exception, it will never be rescheduled.

src/main/java/org/opensearch/knn/quantization/models/quantizationState/QuantizationStateCacheManager.java
```diff
@@ -79,4 +79,8 @@ public void setMaxCacheSizeInKB(long maxCacheSizeInKB) {
     public void clear() {
         QuantizationStateCache.getInstance().clear();
     }
+
+    public void close() throws IOException {
```
jmazanec15, line 83: Implement closeable?

src/main/java/org/opensearch/knn/index/memory/NativeMemoryCacheManager.java
```diff
@@ -99,6 +105,16 @@ private void initialize(NativeMemoryCacheManagerDto nativeMemoryCacheDTO) {
 
         if (nativeMemoryCacheDTO.isExpirationLimited()) {
             cacheBuilder.expireAfterAccess(nativeMemoryCacheDTO.getExpiryTimeInMin(), TimeUnit.MINUTES);
+            Runnable cleanUp = () -> {
+                try {
+                    cache.cleanUp();
+                } catch (Exception e) {
+                    logger.error("Error cleaning up cache", e);
```
jmazanec15, line 112: Why swallow exception here?

jmazanec15 (commented):
Minor comments overall looks good

src/main/java/org/opensearch/knn/index/memory/NativeMemoryCacheManager.java
owenhalpert: Any exceptions from Guava cache operations would otherwise halt our scheduled executor. If an exception occurs here, it would be from Guava's cache manager internals rather than our logic, so per [Dooyong's suggestion](https://github.com/opensearch-project/k-NN/pull/2308#discussion_r1893549051) above we can log it and continue scheduling cleanup tasks.

This way we can ensure the cache maintenance keeps running even if individual cleanup attempts fail, and we can monitor Guava errors in the logs.

src/main/java/org/opensearch/knn/index/util/ScheduledExecutor.java
```diff
@@ -0,0 +1,46 @@
...
+    @Getter
+    private final Runnable task;
+
+    /**
+     * @param task task to be completed
+     * @param scheduleMillis time in milliseconds to wait before executing the task again
+     */
+    public ScheduledExecutor(Runnable task, long scheduleMillis) {
+        this.task = task;
+        this.executor = Executors.newSingleThreadScheduledExecutor();
```
shatejas, line 29: This seems to be creating a new thread for each instance, I wanted to understand the thought process behind this.

I understand that this is used for cache so the threads won't grow exponentially, but since its in util package and the name is generic - any misuse of this can create unbounded number of threads with each instance of this. Can we be a bit more defensive here?

The suggestion here is to refactor and make sure there are fixed number of threads for cache cleanup across caches

owenhalpert: Thanks, I updated ScheduledExecutor to accept an external executor and modified each cache to create and provide their own single threaded executor.

src/main/java/org/opensearch/knn/index/util/ScheduledExecutor.java
```diff
@@ -0,0 +1,45 @@
...
+    final ScheduledExecutorService executor;
+    @Getter
+    private final Runnable task;
+
+    /**
+     * @param task task to be completed
+     * @param scheduleMillis time in milliseconds to wait before executing the task again
+     */
+    public ScheduledExecutor(ScheduledExecutorService executor, Runnable task, long scheduleMillis) {
+        this.task = task;
```
kotwanikunal, line 27: @owenhalpert You can move the duplicated error handling here instead.

```

Runnable task // <- might throw exceptions


Runnable runnableWrapperWithErrorHandling = () -> {
            try {
                task.run();
            }  catch (Exception e) {
                logger.error("Error executing scheduled task", e);
            }
          
        };
   
scheduledExecutorService.scheduleAtFixedRate(runnableWrapperWithErrorHandling, 0, scheduleMillis, TimeUnit.MILLISECONDS);

Comment (shatejas) [reply src/main/java/org/opensearch/knn/index/util/ScheduledExecutor.java:27]: @kotwanikunal considering this is a generic class, it seems a bit risky. Thoughts?

File: src/main/java/org/opensearch/knn/index/util/ScheduledExecutor.java [hunk 9]
```diff
@@ -0,0 +1,45 @@
...
+        this.executor = executor;
+        executor.scheduleAtFixedRate(task, 0, scheduleMillis, TimeUnit.MILLISECONDS);
+    }
+
+    @Override
+    public void close() {
+        executor.shutdown();
+        try {
+            if (!executor.awaitTermination(60, TimeUnit.SECONDS)) {
+                executor.shutdownNow();
```
shatejas, line 37: With executor instantiation moving out of this class, shutting down the executor here can be risky in case of shared threadpools. Should the responsibility shift outside of this class?

kotwanikunal: I was thinking of this too and at that point it made me wonder if we really need this class at that point. 
I guess we can instead utilize the individual cache implementations and move the cleanup + schedule into individual classes. There would be some duplication but this class is not as useful as it stands currently.

src/main/java/org/opensearch/knn/index/util/ScheduledExecutor.java
kotwanikunal: Ahh. Yes. I agree. WDYT of https://github.com/opensearch-project/k-NN/pull/2308#discussion_r1901235519? 
We get rid of this problem in that case

src/main/java/org/opensearch/knn/index/memory/NativeMemoryCacheManager.java
```diff
@@ -16,27 +16,25 @@
...
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.Map;
 import java.util.Optional;
-import java.util.concurrent.ConcurrentLinkedDeque;
-import java.util.concurrent.ExecutionException;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
-import java.util.concurrent.TimeUnit;
+import java.util.concurrent.*;
```
shatejas, line 37: Please undo this

src/main/java/org/opensearch/knn/index/memory/NativeMemoryCacheManager.java
```diff
@@ -469,4 +455,24 @@ private Float getSizeAsPercentage(long size) {
         }
         return 100 * size / (float) cbLimit;
     }
+
+    public void startMaintenance(Cache<String, NativeMemoryAllocation> cacheInstance) {
```
shatejas, line 459: why is this public?

src/main/java/org/opensearch/knn/index/memory/NativeMemoryCacheManager.java
```diff
@@ -469,4 +455,24 @@ private Float getSizeAsPercentage(long size) {
...
+            try {
+                cacheInstance.cleanUp();
+            } catch (Exception e) {
+                logger.error("Error cleaning up cache", e);
+            }
+        };
+
+        TimeValue interval = KNNSettings.state().getSettingValue(KNNSettings.KNN_CACHE_ITEM_EXPIRY_TIME_MINUTES);
+
+        if (threadPool != null) {
```
shatejas, line 474: We need to log an error in `else` if its null. I think throwing is an option - @kotwanikunal thoughts?

src/main/java/org/opensearch/knn/index/memory/NativeMemoryCacheManager.java
```diff
@@ -53,6 +52,9 @@ public class NativeMemoryCacheManager implements Closeable {
     private final ExecutorService executor;
     private AtomicBoolean cacheCapacityReached;
     private long maxWeight;
+    @Setter
+    private static ThreadPool threadPool;
+    private Cancellable maintenanceTask;
 
     NativeMemoryCacheManager() {
         this.executor = Executors.newSingleThreadExecutor();
```
shatejas, line 60: Do we still need this executor?

owenhalpert: This is the executor used in `rebuildCache`, unrelated to my changes

src/main/java/org/opensearch/knn/index/memory/NativeMemoryCacheManager.java
```diff
@@ -53,6 +56,9 @@ public class NativeMemoryCacheManager implements Closeable {
     private final ExecutorService executor;
     private AtomicBoolean cacheCapacityReached;
     private long maxWeight;
+    @Setter
+    private static ThreadPool threadPool;
```
kotwanikunal, line 60: nit: Move the static instance above after `NativeMemoryCacheManager INSTANCE`

src/main/java/org/opensearch/knn/index/memory/NativeMemoryCacheManager.java
kotwanikunal: I think logging makes more sense here since the operations are unaffected without the maintenance thread. 
We don't want the OS process to fail because of an auxiliary process.

Also, please move the runnable definition within the if block, given it would not be used otherwise.

src/main/java/org/opensearch/knn/quantization/models/quantizationState/QuantizationStateCache.java
```diff
@@ -27,14 +31,17 @@
...
 
     private static volatile QuantizationStateCache instance;
     private Cache<String, QuantizationState> cache;
     @Getter
     private long maxCacheSizeInKB;
     @Getter
     private Instant evictedDueToSizeAt;
+    private Cancellable maintenanceTask;
+    @Setter
+    private static ThreadPool threadPool;
```
kotwanikunal, line 44: nit: Please move this var next to the other static variable on #36

src/main/java/org/opensearch/knn/index/memory/NativeMemoryCacheManager.java
owenhalpert: I'll move the call to startMaintenance into a null check on `threadPool` because the entire method has no functionality otherwise.

owenhalpert:
The Github Actions were showing some failures in the Quantization State Cache unit tests, presumably because of concurrent test runs setting/shutting down the thread pool. To address this I added a thread pool that is created and terminated only within the two cache unit test classes to be used for those tests.

src/main/java/org/opensearch/knn/index/memory/NativeMemoryCacheManager.java
```diff
@@ -104,6 +110,9 @@ private void initialize(NativeMemoryCacheManagerDto nativeMemoryCacheDTO) {
         cacheCapacityReached = new AtomicBoolean(false);
         accessRecencyQueue = new ConcurrentLinkedDeque<>();
         cache = cacheBuilder.build();
+        if (threadPool != null) {
```
shatejas, line 113: Irrespective of where this is - we should have visibility when threapool is null, so this needs a warn in `else`.

shatejas (commented):
Looks good overall

shatejas (approved):

kotwanikunal (approved):
LGTM. Thanks @owenhalpert!

jmazanec15 (approved):

PR #2309, opened 2024-12-04
Fixed a bug that does not get a space type from index setting when it is empty for compatibility.

0ctopus13prime:
### Description

### Related Issues

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/engine/SpaceTypeResolver.java
+++ b/src/main/java/org/opensearch/knn/index/engine/SpaceTypeResolver.java
@@ -7,8 +7,10 @@ package org.opensearch.knn.index.engine;
 
 import org.apache.logging.log4j.util.Strings;
 import org.opensearch.index.mapper.MapperParsingException;
+import org.opensearch.knn.index.KNNSettings;
 import org.opensearch.knn.index.SpaceType;
 import org.opensearch.knn.index.VectorDataType;
+import org.opensearch.common.settings.Settings;
 
 import java.util.Locale;
 
@@ -25,24 +27,30 @@ public final class SpaceTypeResolver {
     private SpaceTypeResolver() {}
 
     /**
-     * Resolves space type from configuration details. It is guaranteed not to return either null or
-     * {@link SpaceType#UNDEFINED}
+     * Resolves space type from configuration details. It is guaranteed not to return null.
+     * When space is not in either method and top level, UNDEFINED will be returned.
+     * This can happen when it is defined at index level which is deprecated and no longer allowed in the future.
+     * In this case, UNDEFINED will be returned.
      *
-     * @param knnMethodContext Method context
-     * @param vectorDataType Vectordatatype
+     * @param knnMethodContext        Method context
      * @param topLevelSpaceTypeString Alternative top-level space type
      * @return {@link SpaceType} for the method
      */
     public SpaceType resolveSpaceType(
         final KNNMethodContext knnMethodContext,
-        final VectorDataType vectorDataType,
-        final String topLevelSpaceTypeString
+        final String topLevelSpaceTypeString,
+        final Settings indexSettings,
+        final VectorDataType vectorDataType
     ) {
         SpaceType methodSpaceType = getSpaceTypeFromMethodContext(knnMethodContext);
         SpaceType topLevelSpaceType = getSpaceTypeFromString(topLevelSpaceTypeString);
 
         if (isSpaceTypeConfigured(methodSpaceType) == false && isSpaceTypeConfigured(topLevelSpaceType) == false) {
-            return getSpaceTypeFromVectorDataType(vectorDataType);
+            final String spaceType = indexSettings.get(KNNSettings.INDEX_KNN_SPACE_TYPE.getKey());
+            if (spaceType == null) {
+                return getDefaultSpaceType(vectorDataType);
+            }
+            return SpaceType.getSpace(spaceType);
         }
 
         if (isSpaceTypeConfigured(methodSpaceType) == false) {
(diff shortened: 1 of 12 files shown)


src/main/java/org/opensearch/knn/index/engine/SpaceTypeResolver.java
```diff
@@ -93,4 +86,15 @@ private SpaceType getSpaceTypeFromString(final String spaceType) {
     private boolean isSpaceTypeConfigured(final SpaceType spaceType) {
         return spaceType != null && spaceType != SpaceType.UNDEFINED;
     }
+
+    public SpaceType pickDefaultSpaceTypeWhenEmpty(SpaceType resolvedSpaceType, VectorDataType vectorDataType) {
```
navneet1v, line 90: please add docs on the public function.

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java
```diff
@@ -387,14 +387,17 @@ public Mapper.Builder<?> parse(String name, Map<String, Object> node, ParserCont
...
                     builder.topLevelSpaceType.get()
                 );
-                setSpaceType(builder.originalParameters.getKnnMethodContext(), resolvedSpaceType);
+                setSpaceType(
+                    builder.originalParameters.getKnnMethodContext(),
+                    SpaceTypeResolver.INSTANCE.pickDefaultSpaceTypeWhenEmpty(resolvedSpaceType, builder.vectorDataType.get())
+                );
                 validateSpaceType(builder);
-                resolveKNNMethodComponents(builder, parserContext, resolvedSpaceType);
+                resolveKNNMethodComponents(builder, parserContext, resolvedSpaceType, builder.vectorDataType.get());
```
navneet1v, line 400: we should add some comments explaining what is happening to ensure that we understand why these changes are made.

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java
```diff
@@ -520,7 +524,7 @@ private void resolveKNNMethodComponents(
                 builder.originalParameters.getResolvedKnnMethodContext(),
                 builder.knnMethodConfigContext,
                 false,
-                resolvedSpaceType
+                SpaceTypeResolver.INSTANCE.pickDefaultSpaceTypeWhenEmpty(resolvedSpaceType, vectorDataType)
```
navneet1v, line 527: same as above

navneet1v (commented):
Please add ITs for this change. so that we don't get this bug in future.

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java
0ctopus13prime: sounds good, will update in the rev.

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java
0ctopus13prime: sounds good, will update in the rev.

0ctopus13prime:
@heemin32 
Hi Heemin, could you validate the change in `testPainlessScript_whenNonBinary_thenException`? 
It was using script against to a disabled KNN index.

heemin32:
> @heemin32 Hi Heemin, could you validate the change in `testPainlessScript_whenNonBinary_thenException`? It was using script against to a disabled KNN index.

painless script is supposed to work with disabled KNN index

0ctopus13prime:
Seems like several IT cases are failing.
Let me change `SpaceTypeResolver.INSTANCE.resolveSpaceType` to make it fallback to index settings.
With this change, legacy logic will remain as compatible.

0ctopus13prime:
## Only Difference
Affected changes by this would be for 2.17+ version to visit index settings to get space_type when method + top level don't have space_type.
Baseline is picking up the default space_type based on vector data type for such case. This will be changed to visit index settings to see whether it has `space_type`. After that it will fallback to the default `space_type` based on its vector data type.

## Table Columns Explanation
- Actual Space Type : The final `space_type` engine would get in the end.
- Method : `space_type` at method blob of mapping.
- Top Level : `space_type` at top level mapping blob.
- Index : `space_type` at index level.

For example, below table describes that the engine got `X` as `space_type` when `X` was configured in method blob.
| Version | Actual Space Type | Method | Top Level | Index |
| 2.Y (Y >= 17) | X | null | X | Dont Care |

## ASIS

| Version | Actual Space Type | Method | Top Level | Index |
| 2.Z (Z < 17) | cos | Dont Care | Dont Care | cos |
| 2.Y (Y >= 17) | X | X | Dont Care | Dont Care |
| 2.Y (Y >= 17) | X | null | X | Dont Care |
| 2.Y (Y >= 17) | default based on data type | null | null | X |


## TOBE
| Version | Actual Space Type | Method | Top Level | Index |
| 2.Z (Z < 17) | cos | Dont Care | Dont Care | cos |
| 2.Y (Y >= 17) | X | X | Dont Care | Dont Care |
| 2.Y (Y >= 17) | X | null | X | Dont Care |
| 2.Y (Y >= 17) | X | null | null | X |
| 2.Y (Y >= 17) | default based on data type | null | null | null |

src/main/java/org/opensearch/knn/plugin/rest/RestTrainModelHandler.java
```diff
@@ -151,6 +151,10 @@ && ensureSpaceTypeNotSet(topLevelSpaceType)) {
             vectorDataType = VectorDataType.DEFAULT;
         }
 
+        if (knnMethodContext.getSpaceType() == SpaceType.UNDEFINED && topLevelSpaceType == SpaceType.UNDEFINED) {
```
0ctopus13prime, line 154: @jmazanec15 
Hi Jack, could you confirm it is safe that when both method and top level don't have space_type, then let it pick up the default one based on the vector data type?
This will be a bit different to `KNNVectorFieldMapper`, where it falls back to index settings to get space_type, while RestTrainModelHandler does not check index settings.

0ctopus13prime:
BWC is failing:
will continue investigate the root cause.

```
Cannot set modelId or method parameters when index.knn setting is false
```

src/main/java/org/opensearch/knn/plugin/rest/RestTrainModelHandler.java
0ctopus13prime: Train handler does not have nothing to do with index setting.
Safe to skip that part.

0ctopus13prime:
BWC is failing : 
```
»  Likely root cause: org.opensearch.common.ValidationException: Validation Failed: 1: "hnsw" with "faiss" configuration does not support space type: "linf".;
```

0ctopus13prime:
Raised new PR - https://github.com/opensearch-project/k-NN/pull/2374

PR #2320, opened 2024-12-11
Add check to directly use ANN Search when filters match all docs.

### LINKED ISSUE #1387: [Latency Improvement Oppurtunity] Improve filter search performance when filter results is matching all the documents in the segment
### Description
Currently when filtered vector search is run, we run the filters to find out the relevant filter ids and see if we can do ANN search or exact search for each segment. The identified filterIds are passed to ANN search so that we can do filter while searching.

But in case when lets say filter matched all the docIds in the segment, we can avoid few things:
1. Converting the filterIds bitmap to filterIds array.
2. Validating should we do ANN search or Exact search.

2, is very very cheap(as it has bunch of if else check) but 1 can add latency(exact latency is not identified as of now) if the segment is large enough because we need to [iterate over the bitmap](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/query/KNNWeight.java#L156-L189) to find the docids and pass it till the C++ layer. There are some conversion that happen in c++ layer ref: https://github.com/opensearch-project/k-NN/blob/main/jni/src/faiss_wrapper.cpp#L465-L490.

### Solution

We can avoid all this computation by adding a simple check to see if max doc and filter bitmap size is same. If both of them are same, we can just to simple ANN search.

weiwang118:
### Description
This change aims to resolve this github issue: https://github.com/opensearch-project/k-NN/issues/1387
by adding a simple check in searchLeaf method so that if the filters match all doc IDs in the segment (Max Doc == size of FilterBitSet), then we directly do ANN search and avoid other unnecessary steps. This will help improve the search performance.

### Related Issues
Resolves #1387 

### Check List


src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -127,7 +127,14 @@ public Scorer scorer(LeafReaderContext context) throws IOException {
      */
     public Map<Integer, Float> searchLeaf(LeafReaderContext context, int k) throws IOException {
         final BitSet filterBitSet = getFilteredDocsBitSet(context);
+        final int maxDoc = context.reader().maxDoc();
         int cardinality = filterBitSet.cardinality();
+        /*
+        * If filters match all docs in this segment, then there is no need to do any extra step
+        * and should directly do ANN Search*/
+        if (cardinality ==  maxDoc){
+            return doANNSearch(context, filterBitSet, cardinality, k);
```
heemin32, line 136:
```suggestion
            return doANNSearch(context, new FixedBitSet(0), 0, k);
```

Also, the code should be run after
```
if (filterWeight != null && cardinality == 0) {
            return Collections.emptyMap();
}
```

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -127,7 +127,14 @@ public Scorer scorer(LeafReaderContext context) throws IOException {
      */
     public Map<Integer, Float> searchLeaf(LeafReaderContext context, int k) throws IOException {
         final BitSet filterBitSet = getFilteredDocsBitSet(context);
+        final int maxDoc = context.reader().maxDoc();
         int cardinality = filterBitSet.cardinality();
+        /*
+        * If filters match all docs in this segment, then there is no need to do any extra step
+        * and should directly do ANN Search*/
+        if (cardinality ==  maxDoc){
```
heemin32, line 135:
```suggestion
        if (filterWeight != null && cardinality ==  maxDoc){
```

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -145,6 +146,12 @@ public PerLeafResult searchLeaf(LeafReaderContext context, int k) throws IOExcep
             Map<Integer, Float> result = doExactSearch(context, new BitSetIterator(filterBitSet, cardinality), cardinality, k);
             return new PerLeafResult(filterWeight == null ? null : filterBitSet, result);
         }
+        /*
+         * If filters match all docs in this segment, then there is no need to do any extra step
+         * and should directly do ANN Search*/
+        if (filterWeight != null && cardinality == maxDoc) {
+            return new PerLeafResult(filterBitSet, doANNSearch(context, filterBitSet, cardinality, k));
+        }
```
VijayanB, line 154: 1. What if there are no engine files? I think you have to verify that doANNSearch results are not empty to make sure that ANN Search is success. 
2. Also can you please add Integration test?
3. if cardinality == maxDoc, why can't we pass null instead of filterBitSet?

weiwang118: Already update the pull request and address the comments.

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -145,6 +146,12 @@ public PerLeafResult searchLeaf(LeafReaderContext context, int k) throws IOExcep
             Map<Integer, Float> result = doExactSearch(context, new BitSetIterator(filterBitSet, cardinality), cardinality, k);
             return new PerLeafResult(filterWeight == null ? null : filterBitSet, result);
         }
+        /*
+         * If filters match all docs in this segment, then there is no need to do any extra step
+         * and should directly do ANN Search*/
+        if (filterWeight != null && cardinality == maxDoc) {
+            return new PerLeafResult(new FixedBitSet(0), doANNSearch(context, new FixedBitSet(0), 0, k));
```
0ctopus13prime, line 153: I see lots of `new FixedBitSet(0)` in KNNWeight. 
Could you factor it out with a meaningful name as a static variable then pass it down to the downstream?

```
private static final FixedBitSet ... = new FixedBitSet(0);
```

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -145,6 +146,12 @@ public PerLeafResult searchLeaf(LeafReaderContext context, int k) throws IOExcep
             Map<Integer, Float> result = doExactSearch(context, new BitSetIterator(filterBitSet, cardinality), cardinality, k);
             return new PerLeafResult(filterWeight == null ? null : filterBitSet, result);
         }
+        /*
+         * If filters match all docs in this segment, then there is no need to do any extra step
```
0ctopus13prime, line 150: Could you give more context to the comment?? Would like to see why there's no need for extra steps for all doc matches case, and what passing `new FixedBitSet(0)` means. Basically, we want to save a bitset look up cost when it's not required as possible.

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -145,7 +146,16 @@ public PerLeafResult searchLeaf(LeafReaderContext context, int k) throws IOExcep
...
             return new PerLeafResult(filterWeight == null ? null : filterBitSet, result);
         }
-        Map<Integer, Float> docIdsToScoreMap = doANNSearch(context, filterBitSet, cardinality, k);
+        final Map<Integer, Float> docIdsToScoreMap;
+        /*
+         * If filters match all docs in this segment, then null should be passed as filterBitSet
+         * so that it will not do a bitset look up in bottom search layer.
+         */
+        if (filterWeight != null && cardinality == maxDoc) {
+            docIdsToScoreMap = doANNSearch(context, null, 0, k);
```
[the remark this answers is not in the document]
shatejas: @navneet1v `FixedBitSet(0)` seems to be artificially manipulating the filter bitset and cardinality of it. With code changes its easy to miss that and can lead to future bugs. Would prefer keeping it this way, let me know your thoughts on this. 

@weiwang118 as far as the code pointed out by @navneet1v, can we please handle that in `FilterIdsSelector.getFilterIdSelector` that seems like the right place, you will probably have to pass in max docs there

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -145,7 +146,16 @@ public PerLeafResult searchLeaf(LeafReaderContext context, int k) throws IOExcep
             Map<Integer, Float> result = doExactSearch(context, new BitSetIterator(filterBitSet, cardinality), cardinality, k);
             return new PerLeafResult(filterWeight == null ? null : filterBitSet, result);
         }
-        Map<Integer, Float> docIdsToScoreMap = doANNSearch(context, filterBitSet, cardinality, k);
+        final Map<Integer, Float> docIdsToScoreMap;
+        /*
+         * If filters match all docs in this segment, then null should be passed as filterBitSet
+         * so that it will not do a bitset look up in bottom search layer.
+         */
+        if (filterWeight != null && cardinality == maxDoc) {
```
shatejas, line 154: nit: Instead refactor this to reduce verbosity
```
final FilterIdBitSet annFilter = filterWeight != null && cardinality == maxDoc ? null : filterBitSet;
final Map<Integer, Float> docIdsToScoreMap = doANNSearch(context, filterBitSet, cardinality, k);
```

src/test/java/org/opensearch/knn/integ/FilteredSearchANNSearchIT.java
```diff
@@ -0,0 +1,75 @@
...
+            .build()
+            .getQueryString();
+        Response response = searchKNNIndex(INDEX_NAME, query, expectResultSize);
+        String entity = EntityUtils.toString(response.getEntity());
+        List<String> docIds = parseIds(entity);
+        assertEquals(expectResultSize, docIds.size());
+        assertEquals(expectResultSize, parseTotalSearchHits(entity));
+    }
+
+    private void createKnnIndex(final String indexName, final String fieldName, final int dimension, final KNNEngine knnEngine)
```
shatejas, line 59: nit: can you check if this already exists in the base class?

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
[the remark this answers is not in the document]
weiwang118: Got it~ Will do.

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -145,7 +146,14 @@ public PerLeafResult searchLeaf(LeafReaderContext context, int k) throws IOExcep
             Map<Integer, Float> result = doExactSearch(context, new BitSetIterator(filterBitSet, cardinality), cardinality, k);
             return new PerLeafResult(filterWeight == null ? null : filterBitSet, result);
         }
-        Map<Integer, Float> docIdsToScoreMap = doANNSearch(context, filterBitSet, cardinality, k);
+
+        /*
+         * If filters match all docs in this segment, then null should be passed as filterBitSet
+         * so that it will not do a bitset look up in bottom search layer.
+         */
+        final BitSet annFilter = filterWeight != null && cardinality == maxDoc ? null : filterBitSet;
```
VijayanB, line 154:
```suggestion
        final BitSet annFilter = (filterWeight != null && cardinality == maxDoc) ? null : filterBitSet;
```

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -154,7 +162,7 @@ public PerLeafResult searchLeaf(LeafReaderContext context, int k) throws IOExcep
             Map<Integer, Float> result = doExactSearch(context, docs, cardinality, k);
             return new PerLeafResult(filterWeight == null ? null : filterBitSet, result);
         }
-        return new PerLeafResult(filterWeight == null ? null : filterBitSet, docIdsToScoreMap);
+        return new PerLeafResult((filterWeight == null || cardinality == maxDoc) ? null : filterBitSet, docIdsToScoreMap);
```
VijayanB, line 165: Why do we have to make this change here? I believe updating filter to null is only required for AnnSearch. is there a reason why we have to update the return result as well?

VijayanB (approved):
LGTM. Thanks @weiwang118 for making changes and addressing comments

shatejas (approved):

PR #2331, opened 2024-12-12
Add expand_nested_docs Parameter support to NMSLIB engine

heemin32:
### Description
Add support for the `expand_nested_docs` parameter in the nmslib engine. As nmslib does not support multi-vector functionality, this feature may have limited usefulness but poses no downside to including it.

### Related Issues
N/A

### Check List
- [] New functionality has been documented.


navneet1v:
> As nmslib does not support multi-vector functionality, this feature may have limited usefulness but poses no downside to including it.

Can you explain a bit more on this?

heemin32:
> > As nmslib does not support multi-vector functionality, this feature may have limited usefulness but poses no downside to including it.
> 
> Can you explain a bit more on this?

For nmslib, searches are performed at the nested document level without deduplicating per parent documents As a result, multiple inner hits can occur as long as the nested documents belong to the top-K results. In this context, the demand for this functionality may be lower compared to other engines.

For instance, consider an extreme scenario where there are two documents, each containing 10 nested documents. If a query is executed with k=5, it is possible that all five closest vectors come from just one of the two documents. In this case, expanding nested documents would include five additional nested documents with scores lower than the minimum score of the already retrieved nested documents.

src/main/java/org/opensearch/knn/index/engine/KNNEngine.java
```diff
@@ -34,7 +34,6 @@ public enum KNNEngine implements KNNLibrary {
     private static final Set<KNNEngine> CUSTOM_SEGMENT_FILE_ENGINES = ImmutableSet.of(KNNEngine.NMSLIB, KNNEngine.FAISS);
     private static final Set<KNNEngine> ENGINES_SUPPORTING_FILTERS = ImmutableSet.of(KNNEngine.LUCENE, KNNEngine.FAISS);
     public static final Set<KNNEngine> ENGINES_SUPPORTING_RADIAL_SEARCH = ImmutableSet.of(KNNEngine.LUCENE, KNNEngine.FAISS);
-    public static final Set<KNNEngine> ENGINES_SUPPORTING_MULTI_VECTORS = ImmutableSet.of(KNNEngine.LUCENE, KNNEngine.FAISS);
```
navneet1v, line 37: didn't we have a UT where we will be checking this?

navneet1v (commented):
Please add/update the IT of expand nested docs variable for nmslib engine too.

src/main/java/org/opensearch/knn/index/engine/KNNEngine.java
heemin32: Obviously, we didn't...

heemin32:
> Please add/update the IT of expand nested docs variable for nmslib engine too.

Added.

src/test/java/org/opensearch/knn/integ/ExpandNestedDocsIT.java
```diff
@@ -73,7 +74,12 @@ public class ExpandNestedDocsIT extends KNNRestTestCase {
     @After
     @SneakyThrows
     public final void cleanUp() {
-        deleteKNNIndex(INDEX_NAME);
+        try {
+            deleteKNNIndex(INDEX_NAME);
+        } catch (ResponseException e) {
+            // Index not found exception is acceptable because some test cases do not create an index
+            assert e.getMessage().contains("index_not_found_exception");
+        }
```
navneet1v, line 82: Not sure why we are adding this. There is already a function: `wipeAllODFEIndices` in base ODFERestTestCase that is cleaning up all the indices. So why we need this?

navneet1v: Ref: https://github.com/opensearch-project/k-NN/blob/f3d7d49a3dfe5c16a6d71dee04224fc095430255/src/testFixtures/java/org/opensearch/knn/ODFERestTestCase.java#L136-L138

and If I am not wrong in general we don't delete indices at the end. It gets deleted after every test, by the base functions.

heemin32: You are right. I just saw previous test case where we manually delete index at the end and forgot this part. Will remote it.

jmazanec15 (approved):

PR #2332, opened 2024-12-12
Update knn index default to always do approximate search in test

### LINKED ISSUE #2329: [BUG] Integ test and futuer BWC test will be running only with exact search
After introducing this setting where we don't build a graph when the number of document per segment is less than [15,000](https://github.com/opensearch-project/k-NN/blob/2b9a741a7cfb8c9a55c7412a6e19ea699b7a921d/src/main/java/org/opensearch/knn/index/KNNSettings.java#L101), all existing integ test and bwc test after 2.19 might be running only in exact search mode but not approximate search mode.

We will fail to catch issue related with ANN search through the integ and bwc test.

VijayanB:
### Description
Update knn index default to always do approximate search in test cases

### Related Issues
Resolves #2329 

### Check List

### DIFF
--- a/src/test/java/org/opensearch/knn/index/KNNESSettingsTestIT.java
+++ b/src/test/java/org/opensearch/knn/index/KNNESSettingsTestIT.java
@@ -106,7 +106,11 @@ public class KNNESSettingsTestIT extends KNNRestTestCase {
     }
 
     public void testUpdateIndexSetting() throws IOException {
-        Settings settings = Settings.builder().put("index.knn", true).put(KNNSettings.KNN_ALGO_PARAM_EF_SEARCH, 512).build();
+        Settings settings = Settings.builder()
+            .put("index.knn", true)
+            .put(KNNSettings.KNN_ALGO_PARAM_EF_SEARCH, 512)
+            .put(KNNSettings.INDEX_KNN_ADVANCED_APPROXIMATE_THRESHOLD, 0)
+            .build();
         createKnnIndex(INDEX_NAME, settings, createKnnIndexMapping(FIELD_NAME, 2));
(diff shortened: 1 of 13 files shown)


navneet1v (approved):

heemin32:
There might be test cases which use custom setting instead of default setting. We need to update those cases as well.

VijayanB:
> There might be test cases which use custom setting instead of default setting. We need to update those cases as well.

like what? Can you share the test that you are referring to?

src/test/java/org/opensearch/knn/index/KNNESSettingsTestIT.java
```diff
@@ -106,7 +106,11 @@ public void testCreateIndexWithInvalidSpaceType() throws IOException {
     }
 
     public void testUpdateIndexSetting() throws IOException {
-        Settings settings = Settings.builder().put("index.knn", true).put(KNNSettings.KNN_ALGO_PARAM_EF_SEARCH, 512).build();
+        Settings settings = Settings.builder()
+            .put("index.knn", true)
+            .put(KNNSettings.KNN_ALGO_PARAM_EF_SEARCH, 512)
+            .put(KNNSettings.INDEX_KNN_ADVANCED_APPROXIMATE_THRESHOLD, 0)
```
navneet1v, line 112: for cases like this I think we should just move the code to use `getDefaultIndexSettings` so that in future we don't need to make the changes.

navneet1v (approved):
Overall looks good to me. Just 1 minor comment

VijayanB:
> There might be test cases which use custom setting instead of default setting. We need to update those cases as well.
@heemin32 I realized what you mean by custom mapping. I searched for index.knn is set true and updated those index setting with approximate threshold parameter

heemin32 (approved):

PR #2333, opened 2024-12-12
Switch type of expandNested from boolean to Boolean

heemin32:
### Description
Change the type of the `expandNested` variable from `boolean` to `Boolean`. This modification allows the neural search to pass the `expandNestedDocs` parameter as-is (`null`, `true`, or `false`), instead of converting `null` or `false` to `false` and `true` to `true`.
https://github.com/opensearch-project/neural-search/blob/main/src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java#L315

### Related Issues
N/A

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/query/BaseQueryFactory.java
+++ b/src/main/java/org/opensearch/knn/index/query/BaseQueryFactory.java
@@ -50,7 +50,7 @@ public abstract class BaseQueryFactory {
         private QueryBuilder filter;
         private QueryShardContext context;
         private RescoreContext rescoreContext;
-        private boolean expandNested;
+        private Boolean expandNested;
 
         public Optional<QueryBuilder> getFilter() {
             return Optional.ofNullable(filter);
@@ -62,6 +62,10 @@ public abstract class BaseQueryFactory {
 
         public Optional<RescoreContext> getRescoreContext() {
             return Optional.ofNullable(rescoreContext);
+        }
+
+        public Optional<Boolean> getExpandNested() {
+            return Optional.ofNullable(expandNested);
         }
     }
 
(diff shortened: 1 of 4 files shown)


heemin32:
Rolling upgrade is failing because the query data type is changed. However, it will pass once this change is back ported to 2.x branch.

martin-gaievski (commented):
is there a test that is checking scenario similar to one in neural search (`expandNested` is not set)?

heemin32:
> is there a test that is checking scenario similar to one in neural search (`expandNested` is not set)?

Yes. All the existing tests have query without expandNested being set.

martin-gaievski (approved):
Looks good to me, thank you!

src/main/java/org/opensearch/knn/index/query/BaseQueryFactory.java
```diff
@@ -63,6 +63,10 @@ public Optional<QueryShardContext> getContext() {
         public Optional<RescoreContext> getRescoreContext() {
             return Optional.ofNullable(rescoreContext);
         }
+
+        public Optional<Boolean> getExpandNested() {
+            return Optional.ofNullable(expandNested);
+        }
```
navneet1v, line 69: should we add the default value directly here rather than in `src/main/java/org/opensearch/knn/index/query/KNNQueryFactory.java` ?

navneet1v (commented):
Please add IT and UT for the null value too to ensure everything is working correctly

heemin32:
> Please add IT and UT for the null value too to ensure everything is working correctly.
All existing nested query is actually using null value here. So they are already covered.

src/main/java/org/opensearch/knn/index/query/BaseQueryFactory.java
heemin32: I will let the factory to decide on default value to be consist with other values. For example, rescoreContext's default value is null but we are still wrapping it as Optional instead of just returning null.

navneet1v (approved):

PR #2345, opened 2024-12-19
Concurrency optimization for native graph loading

### LINKED ISSUE #2265: [FEATURE] Concurrency optimizations with native memory graph loading and force eviction
- With the introduction of Lucene compatible loading layer within `NativeMemoryLoadStrategy` - the `IndexLoadStrategy.load()` takes care of loading the graph file into the native memory cache using `IndexInput`
- With force eviction, the [synchronized block](https://github.com/opensearch-project/k-NN/blob/e5599aa151cf43700d0ea327ad536cc106e5bb61/src/main/java/org/opensearch/knn/index/memory/NativeMemoryCacheManager.java#L331-L351) contains the logic for loading the entry into[ the cache as well as the memory ](https://github.com/opensearch-project/k-NN/blob/e5599aa151cf43700d0ea327ad536cc106e5bb61/src/main/java/org/opensearch/knn/index/memory/NativeMemoryCacheManager.java#L347)
- The synchronized nature of the block to deal with cache sizing causes a premature bottleneck with the memory load, especially in the case of concurrent segment search where multiple threads are forced to be synchronized for graph load operations 

- An ideal solution here would be to ensure that the preload to memory and any preliminary operations (for eg download segments in case of remote store or searchable snapshots or checksumming) can be performed outside of the synchronized block to allow for better parallelism
- A suggested approach would look like adding in a new API `ensureLoadable`  to the `NativeMemoryEntryContext` which will make sure  that the graph file is accessible, and ready to be loaded into memory once space is available

N/A

N/A

Gankris96:
### Description
Fixes https://github.com/opensearch-project/k-NN/issues/2265 

Refactors the graph load into a 2 step approach detailed here - https://github.com/opensearch-project/k-NN/issues/2265#issuecomment-2518306522 

This will help to move out the opening of indexInput file outside of the synchronized block so that the graphfile can be downloaded in parallel even if the graph load and createIndexAllocation are inside synchronized block.

### Related Issues
Resolves #2265 

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/memory/NativeMemoryCacheManager.java
+++ b/src/main/java/org/opensearch/knn/index/memory/NativeMemoryCacheManager.java
@@ -35,12 +35,14 @@ import java.util.HashMap;
 import java.util.Iterator;
 import java.util.Map;
 import java.util.Optional;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentLinkedDeque;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.locks.ReentrantLock;
 
 /**
  * Manages native memory allocations made by JNI.
@@ -56,6 +58,7 @@ public class NativeMemoryCacheManager implements Closeable {
 
     private Cache<String, NativeMemoryAllocation> cache;
     private Deque<String> accessRecencyQueue;
+    private final ConcurrentHashMap<String, ReentrantLock> indexLocks = new ConcurrentHashMap<>();
     private final ExecutorService executor;
     private AtomicBoolean cacheCapacityReached;
     private long maxWeight;
@@ -307,6 +310,55 @@ public class NativeMemoryCacheManager implements Closeable {
     }
 
     /**
+     * Opens a vector index with proper locking mechanism to ensure thread safety.
+     * The method uses a ReentrantLock to synchronize access to the index file and
+     * cleans up the lock when no other threads are waiting.
+     *
+     * @param key the unique identifier for the index
+     * @param nativeMemoryEntryContext the context containing vector index information
+     */
+    private void open(String key, NativeMemoryEntryContext nativeMemoryEntryContext) {
+        ReentrantLock indexFileLock = indexLocks.computeIfAbsent(key, k -> new ReentrantLock());
+        try {
+            indexFileLock.lock();
+            nativeMemoryEntryContext.open();
+        } finally {
+            indexFileLock.unlock();
+            if (!indexFileLock.hasQueuedThreads()) {
+                indexLocks.remove(key, indexFileLock);
+            }
+        }
+    }
+
+    /**
+     * Retrieves an entry from the cache and updates its access recency if found.
+     * This method combines cache access with recency queue management to maintain
+     * the least recently used (LRU) order of cached entries.
+     *
+     * @param key the unique identifier for the cached entry
+     * @return the cached NativeMemoryAllocation if present, null otherwise
+     */
+    private NativeMemoryAllocation getFromCacheAndUpdateRecency(String key) {
+        NativeMemoryAllocation result = cache.getIfPresent(key);
+        if (result != null) {
+            updateAccessRecency(key);
+        }
+        return result;
+    }
+
+    /**
+     * Updates the access recency of a cached entry by moving it to the end of the queue.
+     * This method maintains the least recently used (LRU) order by removing the entry
+     * from its current position and adding it to the end of the queue.
+     *
+     * @param key the unique identifier for the cached entry whose recency needs to be updated
+     */
+    private void updateAccessRecency(String key) {
+        accessRecencyQueue.remove(key);
+        accessRecencyQueue.addLast(key);
+    }
+
+    /**
      * Retrieves NativeMemoryAllocation associated with the nativeMemoryEntryContext.
      *
      * @param nativeMemoryEntryContext Context from which to get NativeMemoryAllocation
@@ -338,7 +390,6 @@ public class NativeMemoryCacheManager implements Closeable {
             // In case of a cache miss, least recently accessed entries are evicted in a blocking manner
             // before the new entry can be added to the cache.
             String key = nativeMemoryEntryContext.getKey();
-            NativeMemoryAllocation result = cache.getIfPresent(key);
 
             // Cache Hit
             // In case of a cache hit, moving the item to the end of the recency queue adds
@@ -346,15 +397,21 @@ public class NativeMemoryCacheManager implements Closeable {
             // as lightweight as possible. Multiple approaches and their outcomes were documented
             // before moving forward with the current solution.
             // The details are outlined here: https://github.com/opensearch-project/k-NN/pull/2015#issuecomment-2327064680
+            NativeMemoryAllocation result = getFromCacheAndUpdateRecency(key);
             if (result != null) {
-                accessRecencyQueue.remove(key);
-                accessRecencyQueue.addLast(key);
                 return result;
             }
 
             // Cache Miss
             // Evict before put
+            // open the graph file before proceeding to load the graph into memory
+            open(key, nativeMemoryEntryContext);
             synchronized (this) {
+                // recheck if another thread already loaded this entry into the cache
+                result = getFromCacheAndUpdateRecency(key);
+                if (result != null) {
+                    return result;
+                }
                 if (getCacheSizeInKilobytes() + nativeMemoryEntryContext.calculateSizeInKB() >= maxWeight) {
                     Iterator<String> lruIterator = accessRecencyQueue.iterator();
                     while (lruIterator.hasNext()
@@ -376,7 +433,12 @@ public class NativeMemoryCacheManager implements Closeable {
                 return result;
             }
         } else {
-            return cache.get(nativeMemoryEntryContext.getKey(), nativeMemoryEntryContext::load);
+            // open graphFile before load
+            try (nativeMemoryEntryContext) {
+                String key = nativeMemoryEntryContext.getKey();
+                open(key, nativeMemoryEntryContext);
+                return cache.get(key, nativeMemoryEntryContext::load);
+            }
         }
     }
 
--- a/src/main/java/org/opensearch/knn/index/memory/NativeMemoryEntryContext.java
+++ b/src/main/java/org/opensearch/knn/index/memory/NativeMemoryEntryContext.java
@@ -12,12 +12,16 @@
 package org.opensearch.knn.index.memory;
 
 import lombok.Getter;
+import lombok.extern.log4j.Log4j2;
 import org.apache.lucene.store.Directory;
+import org.apache.lucene.store.IOContext;
+import org.apache.lucene.store.IndexInput;
 import org.opensearch.cluster.service.ClusterService;
 import org.opensearch.common.Nullable;
 import org.opensearch.knn.index.codec.util.NativeMemoryCacheKeyHelper;
 import org.opensearch.knn.index.engine.qframe.QuantizationConfig;
 import org.opensearch.knn.index.VectorDataType;
+import org.opensearch.knn.index.store.IndexInputWithBuffer;
 
 import java.io.IOException;
 import java.util.Map;
@@ -26,7 +30,7 @@ import java.util.UUID;
 /**
  * Encapsulates all information needed to load a component into native memory.
  */
-public abstract class NativeMemoryEntryContext<T extends NativeMemoryAllocation> {
+public abstract class NativeMemoryEntryContext<T extends NativeMemoryAllocation> implements AutoCloseable {
 
     protected final String key;
 
@@ -56,12 +60,26 @@ public abstract class NativeMemoryEntryContext<T extends NativeMemoryAllocation>
     public abstract Integer calculateSizeInKB();
 
     /**
+     * Opens the graph file by opening the corresponding indexInput so
+     * that it is available for graph loading
+     */
+
+    public void open() {}
+
+    /**
+     * Provides the capability to close the closable objects in the {@link NativeMemoryEntryContext}
+     */
+    @Override
+    public void close() {}
+
+    /**
      * Loads entry into memory.
      *
      * @return NativeMemoryAllocation associated with NativeMemoryEntryContext
      */
     public abstract T load() throws IOException;
 
+    @Log4j2
     public static class IndexEntryContext extends NativeMemoryEntryContext<NativeMemoryAllocation.IndexAllocation> {
 
         @Getter
@@ -74,6 +92,17 @@ public abstract class NativeMemoryEntryContext<T extends NativeMemoryAllocation>
         @Nullable
         @Getter
         private final String modelId;
+
+        @Getter
+        private boolean indexGraphFileOpened = false;
+        @Getter
+        private int indexSizeKb;
+
+        @Getter
+        private IndexInput readStream;
+
+        @Getter
+        IndexInputWithBuffer indexInputWithBuffer;
 
         /**
          * Constructor
@@ -132,8 +161,59 @@ public abstract class NativeMemoryEntryContext<T extends NativeMemoryAllocation>
         }
 
         @Override
+        public void open() {
+            // if graph file is already opened for index, do nothing
+            if (isIndexGraphFileOpened()) {
+                return;
+            }
+            // Extract vector file name from the given cache key.
+            // Ex: _0_165_my_field.faiss@1vaqiupVUwvkXAG4Qc/RPg==
+            final String cacheKey = this.getKey();
+            final String vectorFileName = NativeMemoryCacheKeyHelper.extractVectorIndexFileName(cacheKey);
+            if (vectorFileName == null) {
+                throw new IllegalStateException(
+                    "Invalid cache key was given. The key [" + cacheKey + "] does not contain the corresponding vector file name."
+                );
+            }
+
+            // Prepare for opening index input from directory.
+            final Directory directory = this.getDirectory();
+
+            // Try to open an index input then pass it down to native engine for loading an index.
+            try {
+                indexSizeKb = Math.toIntExact(directory.fileLength(vectorFileName) / 1024);
+                readStream = directory.openInput(vectorFileName, IOContext.READONCE);
+                readStream.seek(0);
+                indexInputWithBuffer = new IndexInputWithBuffer(readStream);
+                indexGraphFileOpened = true;
+                log.debug("[KNN] NativeMemoryCacheManager open successful");
+            } catch (IOException e) {
+                throw new RuntimeException("Failed to open the index " + openSearchIndexName);
+            }
+        }
+
+        @Override
         public NativeMemoryAllocation.IndexAllocation load() throws IOException {
+            if (!isIndexGraphFileOpened()) {
+                throw new IllegalStateException("Index graph file is not open");
+            }
             return indexLoadStrategy.load(this);
+        }
+
+        // close the indexInput
+        @Override
+        public void close() {
+            if (readStream != null) {
+                try {
+                    readStream.close();
+                    indexGraphFileOpened = false;
+                } catch (IOException e) {
+                    throw new RuntimeException(
+                        "Exception while closing the indexInput index [" + openSearchIndexName + "] for loading the graph file.",
+                        e
+                    );
+                }
+            }
         }
     }
 
--- a/src/main/java/org/opensearch/knn/index/memory/NativeMemoryLoadStrategy.java
+++ b/src/main/java/org/opensearch/knn/index/memory/NativeMemoryLoadStrategy.java
@@ -13,12 +13,9 @@ package org.opensearch.knn.index.memory;
 
 import lombok.extern.log4j.Log4j2;
 import org.apache.lucene.store.Directory;
-import org.apache.lucene.store.IOContext;
-import org.apache.lucene.store.IndexInput;
 import org.opensearch.core.action.ActionListener;
 import org.opensearch.knn.index.codec.util.NativeMemoryCacheKeyHelper;
 import org.opensearch.knn.index.engine.qframe.QuantizationConfig;
-import org.opensearch.knn.index.store.IndexInputWithBuffer;
 import org.opensearch.knn.index.util.IndexUtil;
 import org.opensearch.knn.jni.JNIService;
 import org.opensearch.knn.index.engine.KNNEngine;
@@ -88,10 +85,16 @@ public interface NativeMemoryLoadStrategy<T extends NativeMemoryAllocation, U ex
             final int indexSizeKb = Math.toIntExact(directory.fileLength(vectorFileName) / 1024);
 
             // Try to open an index input then pass it down to native engine for loading an index.
-            try (IndexInput readStream = directory.openInput(vectorFileName, IOContext.READONCE)) {
-                final IndexInputWithBuffer indexInputWithBuffer = new IndexInputWithBuffer(readStream);
-                final long indexAddress = JNIService.loadIndex(indexInputWithBuffer, indexEntryContext.getParameters(), knnEngine);
-
+            // open in NativeMemoryEntryContext takes care of opening the indexInput file
+            if (!indexEntryContext.isIndexGraphFileOpened()) {
+                throw new IllegalStateException("Index [" + indexEntryContext.getOpenSearchIndexName() + "] is not preloaded");
+            }
+            try (indexEntryContext) {
+                final long indexAddress = JNIService.loadIndex(
+                    indexEntryContext.indexInputWithBuffer,
+                    indexEntryContext.getParameters(),
+                    knnEngine
+                );
                 return createIndexAllocation(indexEntryContext, knnEngine, indexAddress, indexSizeKb, vectorFileName);
             }
         }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 * [Remote Vector Index Build] Introduce Remote Native Index Build feature flag, settings, and initial skeleton [#2525](https://github.com/opensearch-project/k-NN/pull/2525)
 * [Remote Vector Index Build] Implement vector data upload and vector data size threshold setting [#2550](https://github.com/opensearch-project/k-NN/pull/2550)
 * [Remote Vector Index Build] Implement data download and IndexOutput write functionality [#2554](https://github.com/opensearch-project/k-NN/pull/2554)
+* Add concurrency optimizations with native memory graph loading and force eviction (#2265) [https://github.com/opensearch-project/k-NN/pull/2345]
 ### Enhancements
 * Introduce node level circuit breakers for k-NN [#2509](https://github.com/opensearch-project/k-NN/pull/2509)
 ### Bug Fixes
--- a/src/test/java/org/opensearch/knn/index/memory/NativeMemoryCacheManagerTests.java
+++ b/src/test/java/org/opensearch/knn/index/memory/NativeMemoryCacheManagerTests.java
@@ -12,10 +12,22 @@
 package org.opensearch.knn.index.memory;
 
 import com.google.common.cache.CacheStats;
+import org.apache.lucene.store.Directory;
+import org.apache.lucene.store.IndexInput;
+import org.junit.After;
 import org.junit.Before;
+import lombok.SneakyThrows;
+import org.junit.Test;
+import org.mockito.Mock;
+import org.mockito.MockitoAnnotations;
 import org.opensearch.action.admin.cluster.settings.ClusterUpdateSettingsRequest;
+import org.opensearch.cluster.service.ClusterService;
+import org.opensearch.common.settings.ClusterSettings;
+import org.opensearch.common.settings.Setting;
 import org.opensearch.common.settings.Settings;
+import org.opensearch.knn.TestUtils;
 import org.opensearch.knn.common.exception.OutOfNativeMemoryException;
+import org.opensearch.knn.common.featureflags.KNNFeatureFlags;
 import org.opensearch.knn.index.KNNSettings;
 import org.opensearch.knn.index.VectorDataType;
 import org.opensearch.knn.plugin.KNNPlugin;
@@ -28,10 +40,24 @@ import java.io.IOException;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Map;
+import java.util.HashSet;
+import java.util.ArrayList;
+import java.util.Set;
+import java.util.List;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.stream.Collectors;
 
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
+import static org.mockito.Mockito.doReturn;
+import static org.mockito.Mockito.doAnswer;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.any;
+import static org.mockito.Mockito.spy;
 import static org.opensearch.knn.index.memory.NativeMemoryCacheManager.GRAPH_COUNT;
 import static org.opensearch.knn.plugin.stats.StatNames.GRAPH_MEMORY_USAGE;
 
@@ -39,10 +65,36 @@ public class NativeMemoryCacheManagerTests extends OpenSearchSingleNodeTestCase 
 
     private ThreadPool threadPool;
 
+    @Mock
+    protected ClusterService clusterService;
+    @Mock
+    protected ClusterSettings clusterSettings;
+
+    protected AutoCloseable openMocks;
+
     @Before
-    public void setThreadPool() {
+    public void setUp() throws Exception {
+        super.setUp();
+        openMocks = MockitoAnnotations.openMocks(this);
+        clusterService = mock(ClusterService.class);
+        Set<Setting<?>> defaultClusterSettings = new HashSet<>(ClusterSettings.BUILT_IN_CLUSTER_SETTINGS);
+        defaultClusterSettings.addAll(
+            KNNSettings.state()
+                .getSettings()
+                .stream()
+                .filter(s -> s.getProperties().contains(Setting.Property.NodeScope))
+                .collect(Collectors.toList())
+        );
+        KNNSettings.state().setClusterService(clusterService);
+        when(clusterService.getClusterSettings()).thenReturn(new ClusterSettings(Settings.EMPTY, defaultClusterSettings));
         threadPool = new ThreadPool(Settings.builder().put("node.name", "NativeMemoryCacheManagerTests").build());
         NativeMemoryCacheManager.setThreadPool(threadPool);
+    }
+
+    @After
+    public void shutdown() throws Exception {
+        terminate(threadPool);
+        tearDown();
     }
 
     @Override
@@ -53,7 +105,6 @@ public class NativeMemoryCacheManagerTests extends OpenSearchSingleNodeTestCase 
         clusterUpdateSettingsRequest.persistentSettings(circuitBreakerSettings);
         client().admin().cluster().updateSettings(clusterUpdateSettingsRequest).get();
         NativeMemoryCacheManager.getInstance().close();
-        terminate(threadPool);
         super.tearDown();
     }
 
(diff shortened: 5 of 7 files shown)


navneet1v:
Please add an entry in the changelog.

src/main/java/org/opensearch/knn/index/memory/NativeMemoryEntryContext.java
```diff
@@ -131,10 +159,55 @@ public Integer calculateSizeInKB() {
...
+            if (!isPreloaded()) {
+                preload();
+            }
             return indexLoadStrategy.load(this);
         }
+
+        // close the indexInput
+        @Override
+        public void close() {
+            if (readStream != null) {
```
0ctopus13prime, line 200: Could you set `preload=false` in here?

src/main/java/org/opensearch/knn/index/memory/NativeMemoryEntryContext.java
```diff
@@ -131,10 +159,55 @@ public Integer calculateSizeInKB() {
...
+            }
+
+            // Prepare for opening index input from directory.
+            final Directory directory = this.getDirectory();
+
+            // Try to open an index input then pass it down to native engine for loading an index.
+            try {
+                indexSizeKb = Math.toIntExact(directory.fileLength(vectorFileName) / 1024);
+                readStream = directory.openInput(vectorFileName, IOContext.READONCE);
+                readStream.seek(0);
```
0ctopus13prime, line 181: Curious, why we need this?

src/main/java/org/opensearch/knn/index/memory/NativeMemoryEntryContext.java
```diff
@@ -55,6 +59,18 @@ public String getKey() {
      */
     public abstract Integer calculateSizeInKB();
 
+    /**
+     * Preloads the entry by opening the indexInput
+     */
+
+    public abstract void preload();
```
0ctopus13prime, line 66: 1. Can we give a meaningful name? What it does is actually doing a preparation before 'loading' bytes into memory. e.g. opening a stream

2. Can we make it have an empty body?
Then let those subclasses need this to add their implementation.

```
void preload() {
}
```

src/main/java/org/opensearch/knn/index/memory/NativeMemoryCacheManager.java
```diff
@@ -328,6 +328,8 @@ public NativeMemoryAllocation get(NativeMemoryEntryContext<?> nativeMemoryEntryC
 
             // Cache Miss
             // Evict before put
+            // preload the graph file before proceeding to load the graph into memory
```
0ctopus13prime, line 331: It's not preloading the graph file, isn't it? 🤔 
Opening a stream != loading the graph file, I believe

src/main/java/org/opensearch/knn/index/memory/NativeMemoryEntryContext.java
```diff
@@ -75,6 +91,18 @@ public static class IndexEntryContext extends NativeMemoryEntryContext<NativeMem
         @Getter
         private final String modelId;
 
+        @Setter
+        @Getter
```
0ctopus13prime, line 95: Could you remove @Setter here?
This variable should be modified from outside other than testing.
Instead of having a public scope getter, please find another way to change it in testings.

0ctopus13prime:
HI @Gankris96, thank you for the PR. 
I see this clearly will bring benefit for the cases where multiple threads are competing each other.
But just curious, after this fix, how much performance gain do you see?

Gankris96:
> HI @Gankris96, thank you for the PR. I see this clearly will bring benefit for the cases where multiple threads are competing each other. But just curious, after this fix, how much performance gain do you see?

Hi @0ctopus13prime  - yes i am working on getting the benchmarking numbers. Primarily trying to test it on remote store backed index to see the performance gains.

navneet1v:
@Gankris96 please fix the failing CIs.

src/main/java/org/opensearch/knn/index/memory/NativeMemoryEntryContext.java
Gankris96: Yes if there is indexInput classes with lazy-loading or we need to download from an external source. For example, in searchable snapshots case the data might reside in s3 which is downloaded at query time. https://github.com/opensearch-project/k-NN/issues/2265 this issue has some more details on why this would be needed

src/main/java/org/opensearch/knn/index/memory/NativeMemoryEntryContext.java
```diff
@@ -131,10 +157,56 @@ public Integer calculateSizeInKB() {
...
+                indexInputOpened = true;
+            } catch (IOException e) {
+                throw new RuntimeException("Failed to openIndexInput the index " + openSearchIndexName);
+            }
+        }
+
         @Override
         public NativeMemoryAllocation.IndexAllocation load() throws IOException {
+            if (!isIndexInputOpened()) {
+                openIndexInput();
```
shatejas, line 190: Why not throw here? Or let indexLoadStrategy throw?

My concern is with any future code changes - the optimization is not evident and can be thought of as redundant as its already done above. Throwing here with the right message makes it obvious that opening should be done before calling load

Gankris96: I was being a little conservative with doing the `openIndexInput` call here - you're right, maybe throwing instead of trying to open again would be better.

kotwanikunal: +1 on throwing here. If we haven't loaded it until now - we should be explicit in reporting the issue rather than losing the optimization given that load() would be called within the sync block.

src/main/java/org/opensearch/knn/index/memory/NativeMemoryEntryContext.java
```diff
@@ -55,6 +58,18 @@ public String getKey() {
      */
     public abstract Integer calculateSizeInKB();
 
+    /**
+     * Opens the indexInput file so that it is available for graph loading
+     */
+
+    public void openIndexInput() {}
```
kotwanikunal, line 65: I was wondering if we should expose `indexInput` as a part of the interface here. We do not have `indexInput` usages in other entry contexts - we could instead use `loadGraphFile()` or `openGraphFile()`
WDYT?

Gankris96: we could do that but `loadGraphFile` probably would be misleading and can be confused with the `load` method. 
I guess `openGraphFile` makes more sense?

kotwanikunal (approved):
LGTM. Thanks @Gankris96!

src/main/java/org/opensearch/knn/index/memory/NativeMemoryCacheManager.java
```diff
@@ -328,6 +328,9 @@ public NativeMemoryAllocation get(NativeMemoryEntryContext<?> nativeMemoryEntryC
 
             // Cache Miss
             // Evict before put
+            // open the graph file before proceeding to load the graph into memory
+            nativeMemoryEntryContext.openGraphFile();
+            logger.debug("[KNN] NativeMemoryCacheManager openGraphFile successful");
```
navneet1v, line 333: shouldn't this log is better suited to be un the openGraphFile Function?

src/main/java/org/opensearch/knn/index/memory/NativeMemoryEntryContext.java
```diff
@@ -55,6 +58,19 @@ public String getKey() {
      */
     public abstract Integer calculateSizeInKB();
 
+    /**
+     * Opens the graph file by opening the corresponding indexInput so
+     * that it is available for graph loading
+     */
+
+    public void openGraphFile() {}
```
navneet1v, line 66: since this is a base class with different implementations which does not necessarily have graph file or graph data structures, should we rename the function to `openVectorIndex()`.

src/main/java/org/opensearch/knn/index/memory/NativeMemoryEntryContext.java
```diff
@@ -192,6 +269,11 @@ public Integer calculateSizeInKB() {
             return size;
         }
 
+        @Override
+        public void openGraphFile() {
```
navneet1v, line 273: We should throw UnsupportedOperation exception from this function as we are not implementing this.

Gankris96:
# concurrency_benchmarking

Created 2 indices - 

```
sh-4.2$ curl "localhost:9200/_cat/indices"
yellow open target_index  L561sVIRSEWw47bXNx9GQQ 1 1 1000000 0 1.4gb 1.4gb
yellow open target_index2 VEC59SA_Snmup7vX-QTZgA 1 1 1000000 0 1.3gb 1.3gb
sh-4.2$ 
```
Modified my benchmarking to run on 2 indices instead of 1; just to show that there will be more contention while trying to load the graphs

Now run benchmarking with regular warmup and then query 

Index settings - 

```
sh-4.2$ curl "localhost:9200/*/_settings?pretty"
{
  "target_index" : {
    "settings" : {
      "index" : {
        "replication" : {
          "type" : "DOCUMENT"
        },
        "number_of_shards" : "1",
        "provided_name" : "target_index",
        "knn" : "true",
        "creation_date" : "1735945963770",
        "number_of_replicas" : "1",
        "uuid" : "zpIFNh1cTbKtnd2hMaHFFw",
        "version" : {
          "created" : "137217827"
        }
      }
    }
  },
  "target_index2" : {
    "settings" : {
      "index" : {
        "replication" : {
          "type" : "DOCUMENT"
        },
        "knn.advanced" : {
          "approximate_threshold" : "20"
        },
        "number_of_shards" : "1",
        "provided_name" : "target_index2",
        "knn" : "true",
        "creation_date" : "1735945431355",
        "number_of_replicas" : "1",
        "uuid" : "cN8siSW-RAmCV3FJ1YQbKA",
        "version" : {
          "created" : "137217827"
        }
      }
    }
  }
}
sh-4.2$ 
sh-4.2$ 
```

Benchmarking params

```
{
    "target_index_name": "target_index,target_index2",
    "target_field_name": "target_field",
    "target_index_body": "indices/faiss-index.json",
    "target_index_primary_shards": 1,
    "target_index_dimension": 128,
    "target_index_space_type": "l2",
    
    "target_index_bulk_size": 100,
    "target_index_bulk_index_data_set_format": "hdf5",
    "target_index_bulk_index_data_set_path": "sift-128-euclidean.hdf5",
    "target_index_bulk_indexing_clients": 10,
    
    "target_index_max_num_segments": 1,
    "target_index_force_merge_timeout": 300,
    "hnsw_ef_search": 100,
    "hnsw_ef_construction": 100,

    "query_k": 100,
    "query_body": {
         "docvalue_fields" : ["_id"],
         "stored_fields" : "_none_"
    },

    "query_data_set_format": "hdf5",
    "query_data_set_path":"sift-128-euclidean.hdf5",
    "query_count": 100,
    "search_clients":5
  }

```

clear cache before starting the tests - 

```

gramadur@c889f3be2320 knn_scripts % curl -X POST "localhost:9200/_plugins/_knn/clear_cache/target_index"
{"_shards":{"total":2,"successful":1,"failed":0}}%                                                                                                                              gramadur@c889f3be2320 knn_scripts % curl -X POST "localhost:9200/_plugins/_knn/clear_cache/target_index2"
{"_shards":{"total":2,"successful":1,"failed":0}}%                                                                                                                              gramadur@c889f3be2320 knn_scripts % 

```

## 1. Regular index testing

### 1a. Regular index testing without fix


With warmup queries

```
|                                   50th percentile service time |   prod-queries |       23.22 |     ms |
|                                   90th percentile service time |   prod-queries |     61.1095 |     ms |
|                                   99th percentile service time |   prod-queries |     378.385 |     ms |
|                                  100th percentile service time |   prod-queries |     378.405 |     ms |
```


---

Clear cache testing - after each search request, clear cache api is run so. That's why we only get p100

```
|                                  100th percentile service time |  prod-query-1 |     3069.56 |     ms |
|                                  100th percentile service time |  prod-query-2 |     2761.68 |     ms |
|                                  100th percentile service time |  prod-query-3 |     1722.23 |     ms |
|                                  100th percentile service time |  prod-query-4 |     1502.61 |     ms |
|                                  100th percentile service time |  prod-query-5 |     1691.39 |     ms |
|                                  100th percentile service time |  prod-query-6 |     1705.31 |     ms |
|                                  100th percentile service time |  prod-query-7 |     2861.13 |     ms |
|                                  100th percentile service time |  prod-query-8 |      1019.7 |     ms |
|                                  100th percentile service time |  prod-query-9 |     1194.15 |     ms |
```

--- 

### 1b. Regular index testing with fix

First clear cache - 

```
gramadur@c889f3be2320 knn_scripts % curl -X POST "localhost:9200/_plugins/_knn/clear_cache/target_index"
{"_shards":{"total":2,"successful":1,"failed":0}}%                                                                                                                              gramadur@c889f3be2320 knn_scripts % curl -X POST "localhost:9200/_plugins/_knn/clear_cache/target_index2"
{"_shards":{"total":2,"successful":1,"failed":0}}%                                                                                                                              gramadur@c889f3be2320 knn_scripts % 
gramadur@c889f3be2320 knn_scripts % 

```
Next - run the regular testing with index warmup 


With fix regular search -

```
|                                   50th percentile service time |   prod-queries |     17.1917 |     ms |
|                                   90th percentile service time |   prod-queries |     32.2796 |     ms |
|                                   99th percentile service time |   prod-queries |     285.952 |     ms |
|                                  100th percentile service time |   prod-queries |     287.998 |     ms |

```


With fix clear cache - 

```
|                                  100th percentile service time |   prod-query-1 |     2041.92 |     ms |
|                                  100th percentile service time |   prod-query-2 |     1222.25 |     ms |
|                                  100th percentile service time |   prod-query-3 |     878.898 |     ms |
|                                  100th percentile service time |   prod-query-4 |     864.431 |     ms |
|                                  100th percentile service time |   prod-query-5 |     860.176 |     ms |
|                                  100th percentile service time |   prod-query-6 |     872.603 |     ms |
|                                  100th percentile service time |   prod-query-7 |      1028.3 |     ms |
|                                  100th percentile service time |   prod-query-8 |     1131.18 |     ms |
|                                  100th percentile service time |   prod-query-9 |     883.215 |     ms |
|                                  100th percentile service time |  prod-query-10 |     850.829 |     ms |
```

---

## 2. Remote-store index testing

Setup a fs based remote-store and created remote-store indices 

```
sh-4.2$ curl "localhost:9200/target*/_settings?pretty"
{
  "target_index" : {
    "settings" : {
      "index" : {
        "replication" : {
          "type" : "SEGMENT"
        },
        "knn.advanced" : {
          "approximate_threshold" : "20"
        },
        "number_of_shards" : "1",
        "remote_store" : {
          "translog" : {
            "repository" : "local-repo"
          },
          "enabled" : "true",
          "segment" : {
            "repository" : "local-repo"
          }
        },
        "provided_name" : "target_index",
        "knn" : "true",
        "creation_date" : "1736203092702",
        "number_of_replicas" : "1",
        "uuid" : "01BL4u79TYSSitkUCktcVg",
        "version" : {
          "created" : "137217827"
        }
      }
    }
  },
  "target_index2" : {
    "settings" : {
      "index" : {
        "replication" : {
          "type" : "SEGMENT"
        },
        "number_of_shards" : "1",
        "remote_store" : {
          "translog" : {
            "repository" : "local-repo"
          },
          "enabled" : "true",
          "segment" : {
            "repository" : "local-repo"
          }
        },
        "provided_name" : "target_index2",
        "knn" : "true",
        "creation_date" : "1736203842991",
        "number_of_replicas" : "1",
        "uuid" : "GuFw20DbQ067yZRtIy34ow",
        "version" : {
          "created" : "137217827"
        }
      }
    }
  }
}
sh-4.2$ 
```


### 2a. remote store index without fix 
Without fix queries with warmup -

```
|                                   50th percentile service time |   prod-queries |     22.3057 |     ms |
|                                   90th percentile service time |   prod-queries |     86.1489 |     ms |
|                                   99th percentile service time |   prod-queries |     708.677 |     ms |
|                                  100th percentile service time |   prod-queries |     709.391 |     ms |

```


Without fix queries clear cache each time - 

```
|                                  100th percentile service time |   prod-query-1 |    2381.68 |     ms |
|                                  100th percentile service time |   prod-query-2 |    933.662 |     ms |
|                                  100th percentile service time |   prod-query-3 |    984.446 |     ms |
|                                  100th percentile service time |   prod-query-4 |    977.129 |     ms |
|                                  100th percentile service time |   prod-query-5 |    877.563 |     ms |
|                                  100th percentile service time |   prod-query-6 |    984.302 |     ms |
|                                  100th percentile service time |   prod-query-7 |    1082.44 |     ms |
|                                  100th percentile service time |   prod-query-8 |    1118.12 |     ms |
|                                  100th percentile service time |   prod-query-9 |    921.837 |     ms |
|                                  100th percentile service time |  prod-query-10 |     902.73 |     ms |
```


### 2b. remote store index with fix

With fix remote index with warmup 

```
|                                   50th percentile service time |   prod-queries |     18.5364 |     ms |
|                                   90th percentile service time |   prod-queries |     42.2512 |     ms |
|                                   99th percentile service time |   prod-queries |     583.201 |     ms |
|                                  100th percentile service time |   prod-queries |     583.968 |     ms |

```

With fix remote index with clearing cache - 

```
|                                  100th percentile service time |   prod-query-1 |     886.207 |     ms |
|                                  100th percentile service time |   prod-query-2 |     945.632 |     ms |
|                                  100th percentile service time |   prod-query-3 |     951.203 |     ms |
|                                  100th percentile service time |   prod-query-4 |     1037.99 |     ms |
|                                  100th percentile service time |   prod-query-5 |     982.428 |     ms |
|                                  100th percentile service time |   prod-query-6 |      1116.4 |     ms |
|                                  100th percentile service time |   prod-query-7 |     1152.06 |     ms |
|                                  100th percentile service time |   prod-query-8 |     985.593 |     ms |
|                                  100th percentile service time |   prod-query-9 |     949.892 |     ms |
|                                  100th percentile service time |  prod-query-10 |     1006.75 |     ms |
```

Gankris96:
@0ctopus13prime Have added the benchmarking numbers above and here is the summary -

1. Regular index testing with warmup
```
|                  Percentile | Without Fix | With Fix | Delta (ms) | Improvement |
|----------------------------|-------------|----------|------------|-------------|
|                       50th |     23.2200 | 17.1917 |    -6.0283 |     -26.0% |
|                       90th |     61.1095 | 32.2796 |   -28.8299 |     -47.2% |
|                       99th |    378.3850 | 285.9520|   -92.4330 |     -24.4% |
|                      100th |    378.4050 | 287.9980|   -90.4070 |     -23.9% |
```
2. Regular index testing with clear cache after each query 
```
| Query #   | Without Fix | With Fix  | Delta (ms) | Improvement |
|-----------|-------------|-----------|------------|-------------|
| Query 1   |    3069.56 |  2041.92  | -1027.64   |    -33.5%   |
| Query 2   |    2761.68 |  1222.25  | -1539.43   |    -55.7%   |
| Query 3   |    1722.23 |   878.90  |  -843.33   |    -49.0%   |
| Query 4   |    1502.61 |   864.43  |  -638.18   |    -42.5%   |
| Query 5   |    1691.39 |   860.18  |  -831.21   |    -49.1%   |
| Query 6   |    1705.31 |   872.60  |  -832.71   |    -48.8%   |
| Query 7   |    2861.13 |  1028.30  | -1832.83   |    -64.1%   |
| Query 8   |    1019.70 |  1131.18  |  +111.48   |    +10.9%   |
| Query 9   |    1194.15 |   883.22  |  -310.94   |    -26.0%   |
| Query 10* |        N/A |   850.83  |      N/A   |      N/A    |


Key Statistics:
- Average improvement (excluding Query 10): -39.8%
- Average time without fix: 1947.53 ms
- Average time with fix: 1063.38 ms
```

--- 
3. Remote store index with warmup 
```
|                  Percentile | Without Fix | With Fix | Delta (ms) | Improvement |
|----------------------------|-------------|----------|------------|-------------|
|                       50th |     22.3057 | 18.5364  |    -3.7693 |     -16.9% |
|                       90th |     86.1489 | 42.2512  |   -43.8977 |     -51.0% |
|                       99th |    708.6770 | 583.2010 |  -125.4760 |     -17.7% |
|                      100th |    709.3910 | 583.9680 |  -125.4230 |     -17.7% |
```
4. Remote store index with clear cache after each - 
```
| Query #   | Without Fix | With Fix  | Delta (ms) | Improvement |
|-----------|-------------|-----------|------------|-------------|
| Query 1   |    2381.68 |   886.21  | -1495.47   |    -62.8%   |
| Query 2   |     933.66 |   945.63  |   +11.97   |     +1.3%   |
| Query 3   |     984.45 |   951.20  |   -33.25   |     -3.4%   |
| Query 4   |     977.13 |  1037.99  |   +60.86   |     +6.2%   |
| Query 5   |     877.56 |   982.43  |  +104.87   |    +11.9%   |
| Query 6   |     984.30 |  1116.40  |  +132.10   |    +13.4%   |
| Query 7   |    1082.44 |  1152.06  |   +69.62   |     +6.4%   |
| Query 8   |    1118.12 |   985.59  |  -132.53   |    -11.9%   |
| Query 9   |     921.84 |   949.89  |   +28.05   |     +3.0%   |
| Query 10  |     902.73 |  1006.75  |  +104.02   |    +11.5%   |

Key Statistics:
- Average time without fix: 1116.39 ms
- Average time with fix: 1001.41 ms
- Overall average improvement: -2.4%
```

Overall we dont see much improvement in the remote-store case but with a more constrained environment (larger graphs) and more concurrent searches, we should see even more improvement.

@0ctopus13prime , @kotwanikunal , @shatejas  let me know your thoughts

navneet1v:
Was the remote index hosted on like S3? because what I read is it is FS store and not S3.

src/main/java/org/opensearch/knn/index/memory/NativeMemoryEntryContext.java
```diff
@@ -131,10 +159,55 @@ public Integer calculateSizeInKB() {
...
+
+            // Prepare for opening index input from directory.
+            final Directory directory = this.getDirectory();
+
+            // Try to open an index input then pass it down to native engine for loading an index.
+            try {
+                indexSizeKb = Math.toIntExact(directory.fileLength(vectorFileName) / 1024);
+                readStream = directory.openInput(vectorFileName, IOContext.READONCE);
+                readStream.seek(0);
+                indexInputWithBuffer = new IndexInputWithBuffer(readStream);
```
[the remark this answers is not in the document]
Gankris96: jni layer loadIndex method takes in indexInputWithBuffer as a parameter. https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/jni/JNIService.java#L196

src/main/java/org/opensearch/knn/index/memory/NativeMemoryLoadStrategy.java
```diff
@@ -88,10 +85,16 @@ public NativeMemoryAllocation.IndexAllocation load(NativeMemoryEntryContext.Inde
...
-                final IndexInputWithBuffer indexInputWithBuffer = new IndexInputWithBuffer(readStream);
-                final long indexAddress = JNIService.loadIndex(indexInputWithBuffer, indexEntryContext.getParameters(), knnEngine);
-
+            // openGraphFile takes care of opening the indexInput file
+            if (!indexEntryContext.isIndexGraphFileOpened()) {
+                throw new IllegalStateException("Index [" + indexEntryContext.getOpenSearchIndexName() + "] is not preloaded");
+            }
+            try (indexEntryContext) {
+                final long indexAddress = JNIService.loadIndex(
+                    indexEntryContext.indexInputWithBuffer,
```
Gankris96, line 94: This is where we use the indexInputWithBuffer. Did you have something else in mind @navneet1v  ?

src/main/java/org/opensearch/knn/index/memory/NativeMemoryEntryContext.java
Gankris96: Makes sense, will update the name,

src/main/java/org/opensearch/knn/index/memory/NativeMemoryCacheManager.java
Gankris96: Yeah thats true. There is no logger object in that class and i didn't want to initialize it for just 1 debug log line.

Gankris96:
> Was the remote index hosted on like S3? because what I read is it is FS store and not S3. Also did we enable the concurrent segment search here? @Gankris96

You are right, Remote store index is hosted on fs directory and s3. However from remote store pov it is the same for both these cases. What i mean is even if it is hosted on s3, the data will reside locally as well, it is just backed up by s3. In case of fs directory, the data resides in the fs directory path we provide. 

Good point on concurrent search, I didn't enable it. Let me run one round with concurrent search enabled and get some benchmarking numbers

src/main/java/org/opensearch/knn/index/memory/NativeMemoryEntryContext.java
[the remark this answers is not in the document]
navneet1v: got it. I think I am good here.

src/main/java/org/opensearch/knn/index/memory/NativeMemoryLoadStrategy.java
navneet1v: I think its fine. I was thinking something else. We are good here.

src/main/java/org/opensearch/knn/index/memory/NativeMemoryCacheManager.java
navneet1v: > There is no logger object in that class and i didn't want to initialize it for just 1 debug log line. Do you still think its ok to leave it as is or should we move it to the NativeMemoryEntryContext class ?

I think you can just add the @Log4j2 annotation on top of the class and move the log line in that class. Moving the line to the class is better. But its not a make or break thing for me. So take your call.

Gankris96:
Here is the benchmarking comparison with concurrent segment search enabled on remote-store index 

```
 Query #   | Without Fix | With Fix  | Delta (ms) | Improvement |
|-----------|-------------|-----------|------------|-------------|
| Query 1   |    1940.73 |  1899.53  |   -41.20   |     -2.1%   |
| Query 2   |     915.54 |   957.23  |   +41.69   |     +4.6%   |
| Query 3   |     883.05 |   874.13  |    -8.92   |     -1.0%   |
| Query 4   |     850.17 |   859.59  |    +9.42   |     +1.1%   |
| Query 5   |     936.52 |   748.44  |  -188.08   |    -20.1%   |
| Query 6   |     857.33 |   841.82  |   -15.51   |     -1.8%   |
| Query 7   |     870.06 |   773.51  |   -96.55   |    -11.1%   |
| Query 8   |     825.31 |   847.89  |   +22.58   |     +2.7%   |
| Query 9   |     841.98 |   897.34  |   +55.37   |     +6.6%   |
| Query 10  |     828.68 |   860.58  |   +31.90   |     +3.9%   |

Key Statistics:
- Average time without fix: 974.93 ms
- Average time with fix: 956.00 ms
```

--- 

Searchable snapshot indices restored_target_index and restored_target_index2
```
red    open target_index           01BL4u79TYSSitkUCktcVg 1 1                      
yellow open restored_target_index  un9f_mzTRQKAQWniE0bqoA 1 1 1000000 0 1.4gb 1.4gb
red    open target_index2          GuFw20DbQ067yZRtIy34ow 1 1                      
yellow open restored_target_index2 S8lTIANoT1ytZ4jVJsxgAQ 1 1 1000000 0 1.4gb 1.4gb
red    open million_songs          -rbMHk1QQF2U7WlcMgO9Dw 1 1                      


sh-4.2$ curl "localhost:9200/restored*/_settings?pretty"
{
  "restored_target_index" : {
    "settings" : {
      "index" : {
        "replication" : {
          "type" : "SEGMENT"
        },
        "knn.advanced" : {
          "approximate_threshold" : "20"
        },
        "number_of_shards" : "1",
        "remote_store" : {
          "translog" : {
            "repository" : "local-repo"
          },
          "enabled" : "true",
          "segment" : {
            "repository" : "local-repo"
          }
        },
        "provided_name" : "target_index",
        "searchable_snapshot" : {
          "index" : {
            "id" : "nNNDvxFeQ1qFJl4LjcrBkQ"
          },
          "shard_path_type" : "FIXED",
          "repository" : "my_backup",
          "snapshot_id" : {
            "name" : "snapshot_1",
            "uuid" : "SnNo_A97S_yoUhfDOn8qZA"
          }
        },
        "knn" : "true",
        "creation_date" : "1736203092702",
        "store" : {
          "type" : "remote_snapshot"
        },
        "number_of_replicas" : "1",
        "uuid" : "un9f_mzTRQKAQWniE0bqoA",
        "version" : {
          "created" : "137217827"
        }
      }
    }
  },
  "restored_target_index2" : {
    "settings" : {
      "index" : {
        "replication" : {
          "type" : "SEGMENT"
        },
        "number_of_shards" : "1",
        "remote_store" : {
          "translog" : {
            "repository" : "local-repo"
          },
          "enabled" : "true",
          "segment" : {
            "repository" : "local-repo"
          }
        },
        "provided_name" : "target_index2",
        "searchable_snapshot" : {
          "index" : {
            "id" : "4fkcDpVGSeuHpyKBlJEHWg"
          },
          "shard_path_type" : "FIXED",
          "repository" : "my_backup",
          "snapshot_id" : {
            "name" : "snapshot_2",
            "uuid" : "AAQ6KfAsTGmHKdvTxXUCTw"
          }
        },
        "knn" : "true",
        "creation_date" : "1736203842991",
        "store" : {
          "type" : "remote_snapshot"
        },
        "number_of_replicas" : "1",
        "uuid" : "S8lTIANoT1ytZ4jVJsxgAQ",
        "version" : {
          "created" : "137217827"
        }
      }
    }
  }
}
sh-4.2$ 
```

### Cluster settings 

```
{
    "persistent": {
        "cluster": {
            "blocks": {
                "create_index": "true"
            }
        },
        "knn": {
            "feature": {
                "cache": {
                    "force_evict": {
                        "enabled": "true"
                    }
                }
            }
        },
        "search": {
            "concurrent_segment_search": {
                "mode": "all"
            }
        }
    },
    "transient": {}
}
```

## Searchable snapshots - 
```
| Query #   | Without Fix | With Fix  | Delta (ms) | Improvement |
|-----------|-------------|-----------|------------|-------------|
| Query 1   |    3127.25 |   108.58  | -3018.67   |    -96.5%   |
| Query 2   |     119.83 |   101.21  |   -18.62   |    -15.5%   |
| Query 3   |     112.18 |    96.32  |   -15.86   |    -14.1%   |
| Query 4   |     125.90 |   101.54  |   -24.36   |    -19.3%   |
| Query 5   |     118.81 |   102.40  |   -16.41   |    -13.8%   |
| Query 6   |     122.63 |   105.12  |   -17.51   |    -14.3%   |
| Query 7   |     131.87 |   110.51  |   -21.36   |    -16.2%   |
| Query 8   |     112.92 |   107.89  |    -5.03   |     -4.5%   |
| Query 9   |     112.51 |   102.68  |    -9.83   |     -8.7%   |

Key Statistics:
- Average time without fix: 453.77 ms
- Average time with fix: 104.03 ms
- Overall average improvement: -77.1%

```

The searchable snapshot case shows improvement @navneet1v

src/main/java/org/opensearch/knn/index/memory/NativeMemoryEntryContext.java
```diff
@@ -278,6 +363,11 @@ public Integer calculateSizeInKB() {
             return size;
         }
 
+        @Override
+        public void openVectorIndex() {
+            return;
+        }
```
navneet1v, line 369: this function should also throw the `UnsupportedOperationException` just like the other class.

navneet1v:
> The searchable snapshot case shows improvement @navneet1v

Great this is what was expected.

src/main/java/org/opensearch/knn/index/memory/NativeMemoryCacheManager.java
```diff
@@ -350,7 +352,11 @@ public NativeMemoryAllocation get(NativeMemoryEntryContext<?> nativeMemoryEntryC
                 return result;
             }
         } else {
-            return cache.get(nativeMemoryEntryContext.getKey(), nativeMemoryEntryContext::load);
+            // open graphFile before load
+            try (nativeMemoryEntryContext) {
```
Vikasht34, line 356: There could be a case where Multiple threads trigger eviction and graph loading concurrently, leading to temporary spikes in memory usage. Can we think of using bounded concurrency for eviction and graph loading tasks with thread pools?

src/main/java/org/opensearch/knn/index/memory/NativeMemoryCacheManager.java
```diff
@@ -350,7 +352,11 @@ public NativeMemoryAllocation get(NativeMemoryEntryContext<?> nativeMemoryEntryC
...
+                nativeMemoryEntryContext.openVectorIndex();
```
Vikasht34, line 357: Can we avoid this case when graph is partially loaded or an error occurs during loading, which endup cache being  an inconsistent state . Can we ensure automaticity in graph loading and only put in cache if it is successful.

Vikasht34 (commented):
Could we make our unit and integ test more robust with code changes we are doing 

1. for cases isIndexGraphFileOpened() Ensure that openVectorIndex does nothing if the index graph file is already opened.
2.Verify that the method extracts the vector file name correctly and proceeds to load the index without errors.
3.Pass an invalid cache key that does not contain a vector file name and verify that the method throws an IllegalStateException with the correct error message.
4.Can we mock directory.openInput method to return a valid IndexInput and verify that readStream and indexInputWithBuffer are initialized correctly.
5. Can we veify   readStream.seek(0) is called successfully.

src/main/java/org/opensearch/knn/index/memory/NativeMemoryEntryContext.java
Gankris96: Since we are offloading the opening of vector index from the `load` method in loadStrategy, this call ends up calling the `openVectorIndex`  of the corresponding entryContext here - https://github.com/Gankris96/k-NN/blob/ecdb8fa6c23992bac07170cc54e347ed3f2bc85f/src/main/java/org/opensearch/knn/index/memory/NativeMemoryCacheManager.java#L357 

This means that we can't throw `UnsupportedOperationException` here. I can either leave the return here as a no-op for the non vector cases or i have to scope the openVectorIndex call in the above line i shared to only do so if `nativeMemoryEntryContext instanceof NativeMemoryEntryContext.IndexEntryContext`. This is needed because if `forceEviction` is not enabled, then we directly jump to the load call, but the graph file is not yet opened.

navneet1v (dismissed):

navneet1v:
@Gankris96 can you please fix the CIs

src/main/java/org/opensearch/knn/index/memory/NativeMemoryCacheManager.java
Gankris96: if there is error in graph loading then the entry will not be in cache. What would be the scenario where cache ends up in inconsistent state ?

src/main/java/org/opensearch/knn/index/memory/NativeMemoryCacheManager.java
Gankris96: will take it up in a separate issue

src/main/java/org/opensearch/knn/index/memory/NativeMemoryEntryContext.java
```diff
@@ -131,10 +160,61 @@ public Integer calculateSizeInKB() {
...
+            }
+
+            // Prepare for opening index input from directory.
+            final Directory directory = this.getDirectory();
+
+            // Try to open an index input then pass it down to native engine for loading an index.
+            try {
+                indexSizeKb = Math.toIntExact(directory.fileLength(vectorFileName) / 1024);
+                readStream = directory.openInput(vectorFileName, IOContext.READONCE);
+                readStream.seek(0);
```
Gankris96, line 186: there is another issue which can arise here that is with multiple threads trying to open the same indexInput file. Basically what can happen is that 2 threads can create 2 separate IndexEntryContext for the same index and then call `openVectorIndex()`. 
There will be contention between both the threads to open the indexInput and download the data. However, only 1 of them needs to do the download and other on can reuse the same indexInput and the other can just reuse it by checking if `isIndexGraphFileOpened()`. This will avoid unnecessary i/o. Also, we can go into a situation where we unnecessarily evict from cache 2 times for the same vectorFile because of this contention. 

To avoid this, we need to synchronize on the indexFile name within `NativeMemoryCacheManager`. If 2 threads try to call `openVectorIndex` on the same `key`, then one of the threads has to wait for the other. Basically we will synchronize on the cache key to prevent duplicate file descriptors being created and duplicate downloads from being triggered. 
I will update this pr with fix that addresses this

Gankris96:
@Vikasht34  @kotwanikunal @navneet1v  have updated with some additional locking around `openVectorIndex` and added some UTs around the same. Please take a look

src/main/java/org/opensearch/knn/index/memory/NativeMemoryCacheManager.java
```diff
@@ -328,6 +328,8 @@ public NativeMemoryAllocation get(NativeMemoryEntryContext<?> nativeMemoryEntryC
 
             // Cache Miss
             // Evict before put
+            // open the graph file before proceeding to load the graph into memory
+            nativeMemoryEntryContext.openVectorIndex();
```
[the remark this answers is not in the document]
kotwanikunal: Ganesh and I had a chat around this separately. To summarize, we realized that even though the cache would only hold a single entry, there can be cases where we try to open the index files in parallel if two search requests land up in here at the same time.
Thanks for the catch @Vikasht34!

src/main/java/org/opensearch/knn/index/memory/NativeMemoryCacheManager.java
kotwanikunal: This is a fair callout. I think we need to improve on our cache operations in general.
I think the problem we are going through right now is that the cache operations can be async in nature (cleanup, eviction) where as we use it as a 1:1 reference for the off heap memory in use.
We can create a tracking issue and deal with this separately.

src/main/java/org/opensearch/knn/index/memory/NativeMemoryCacheManager.java
kotwanikunal: @Gankris96 Can we wrap this call behind the same lock based logic above?
Just to make sure we do not open the same index files concurrently in two different threads?

Vikasht34 (dismissed):
Thanks for addressing the comments , Looks good to me.

src/main/java/org/opensearch/knn/index/memory/NativeMemoryCacheManager.java
Gankris96: Wrapping this within a lock still seems to fail some bwc search tests where we endup getting incorrect results. Even doing so would not really help coz we don't solve the eventual problem of multiple graph files getting loaded at the same time because the load is not synchronized anymore. 
This probably requires revisiting in a new separate issue where we refactor the whole cache strategy imo.

kotwanikunal: Please create an issue so that we can track it.

kotwanikunal (approved):

src/main/java/org/opensearch/knn/index/memory/NativeMemoryCacheManager.java
Gankris96: I guess the bwc failure was a different issue unrelated to this. I did add back the locking logic for this as well. It seems to work fine so we can keep this in.

Gankris96:
@navneet1v @jmazanec15  please take a look and approve if all looks good.

kotwanikunal (approved):

jmazanec15 (approved):

src/main/java/org/opensearch/knn/index/memory/NativeMemoryCacheManager.java
```diff
@@ -345,7 +348,22 @@ public NativeMemoryAllocation get(NativeMemoryEntryContext<?> nativeMemoryEntryC
 
             // Cache Miss
             // Evict before put
+            // open the graph file before proceeding to load the graph into memory
+            ReentrantLock indexFileLock = indexLocks.computeIfAbsent(key, k -> new ReentrantLock());
+            indexFileLock.lock();
+            nativeMemoryEntryContext.openVectorIndex();
+            indexFileLock.unlock();
```
shatejas, line 355: Can we please, have a private method `openIndex()` here so this is taken care as an when the code changes?

src/main/java/org/opensearch/knn/index/memory/NativeMemoryCacheManager.java
```diff
@@ -345,7 +348,22 @@ public NativeMemoryAllocation get(NativeMemoryEntryContext<?> nativeMemoryEntryC
...
+                indexLocks.remove(key, indexFileLock);
+            }
             synchronized (this) {
+                // recheck if another thread already loaded this entry into the cache
+                result = cache.getIfPresent(key);
+                if (result != null) {
+                    accessRecencyQueue.remove(key);
+                    accessRecencyQueue.addLast(key);
+                    return result;
+                }
```
shatejas, line 366: private method for this as well? There will be additional null check but everytime `get` returns accessRecency should be updated.

shatejas (commented):
Looks good overall. Some comments related to maintenance of code

navneet1v:
Overall looks good to me. I think I agree with @shatejas comments. Please resolve them so that we can ship this change

src/main/java/org/opensearch/knn/index/memory/NativeMemoryEntryContext.java
```diff
@@ -55,13 +59,27 @@ public String getKey() {
      */
     public abstract Integer calculateSizeInKB();
 
+    /**
+     * Opens the graph file by opening the corresponding indexInput so
+     * that it is available for graph loading
+     */
+
+    public void openVectorIndex() {}
```
jmazanec15, line 67: Can we just call this open? The index aspect only really applies to IndexEntryContext. There are other allocation types this might not make sense for

src/main/java/org/opensearch/knn/index/memory/NativeMemoryCacheManager.java
```diff
@@ -306,6 +309,55 @@ public CacheStats getCacheStats() {
...
 
+    /**
+     * Opens a vector index with proper locking mechanism to ensure thread safety.
+     * The method uses a ReentrantLock to synchronize access to the index file and
+     * cleans up the lock when no other threads are waiting.
+     *
+     * @param key the unique identifier for the index
+     * @param nativeMemoryEntryContext the context containing vector index information
+     */
+    private void openIndex(String key, NativeMemoryEntryContext nativeMemoryEntryContext) {
```
jmazanec15, line 320: Can we call open?

jmazanec15 (commented):
A few minor comments on naming

kotwanikunal (approved):

shatejas (approved):

PR #2348, opened 2024-12-21
Fixed a bug to prevent updating index.knn setting after index creation

### LINKED ISSUE #2334: [BUG]Block updation of knn.enabled flag.
**What is the bug?**
Today, user can update `index.knn` flag even after an index is created which should not be allowed. Users can close the index, update this setting and open the index back which is not an expected behaviour. We have seen some issues due to this recently as well during version upgrades to 2.17.
Also, user can today restore an existing index with `index.knn` setting as true as an index with `index.knn` setting as false. we need to block this setting from getting updated during snapshot restore as well.

1. Create a kNN enabled index.
2. Close the index.
3. Open the index.

knn.enabled setting should be final and should not be allowed to get updated.

This index setting should be made final, so that it cannot be updated even after closing the index.  
Also, during snapshot restore, there is a list of index settings for which we do not support updation. We should add this setting in that list.

anntians:
### Description
This change prevents customers to update index.knn setting after creating an index. Previously, there was a loophole to migrate a live k-nn index from hot to UW by closing the index, setting index.knn to false, opening the index, and migrating to UW. However, migrating k-nn indices to UW nodes is not supported. Thus, this PR updates index.knn to be FINAL and immutable after index creation.

### Related Issues
Resolves #2334 

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/KNNSettings.java
+++ b/src/main/java/org/opensearch/knn/index/KNNSettings.java
@@ -43,6 +43,7 @@ import static java.util.stream.Collectors.toUnmodifiableMap;
 import static org.opensearch.common.settings.Setting.Property.Dynamic;
 import static org.opensearch.common.settings.Setting.Property.IndexScope;
 import static org.opensearch.common.settings.Setting.Property.NodeScope;
+import static org.opensearch.common.settings.Setting.Property.Final;
 import static org.opensearch.common.unit.MemorySizeValue.parseBytesSizeValueOrHeapRatio;
 import static org.opensearch.core.common.unit.ByteSizeValue.parseBytesSizeValue;
 import static org.opensearch.knn.common.featureflags.KNNFeatureFlags.getFeatureFlags;
@@ -268,7 +269,7 @@ public class KNNSettings {
     /**
      * This setting identifies KNN index.
      */
-    public static final Setting<Boolean> IS_KNN_INDEX_SETTING = Setting.boolSetting(KNN_INDEX, false, IndexScope);
+    public static final Setting<Boolean> IS_KNN_INDEX_SETTING = Setting.boolSetting(KNN_INDEX, false, IndexScope, Final);
 
     /**
      * index_thread_quantity - the parameter specifies how many threads the nms library should use to create the graph.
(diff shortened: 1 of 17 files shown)


src/main/java/org/opensearch/knn/index/KNNSettings.java
```diff
@@ -268,7 +269,7 @@ public class KNNSettings {
     /**
      * This setting identifies KNN index.
      */
-    public static final Setting<Boolean> IS_KNN_INDEX_SETTING = Setting.boolSetting(KNN_INDEX, false, IndexScope);
```
jmazanec15, line 271: Can you check what behavior happens for an index created on an old version of OpenSearch? Is the setting immutable after upgrading? This could be checked by adding a test in https://github.com/opensearch-project/k-NN/tree/main/qa/restart-upgrade/src/test/java/org/opensearch/knn/bwc

kotwanikunal: +1. I think we also need to add in a rolling version upgrade test for this specific scenario to ensure upgrades are covered correctly.

anntians: Rolling version upgrade tests for checking immutability of the index.knn setting will be inconsistent. The reason is because in the mixed cluster state, some nodes in the cluster may not have been updated with the immutability code change, so depending on which node is hit in the rolling version upgrade test, the result can be success or failure.

So, should we just write tests for the case when cluster is fully upgraded, in which all nodes will have the updated immutability change. Or should we also include tests when the cluster is in mixed state, in which some nodes are updated and some are not, potentially leading to inconsistent test results.

kotwanikunal (approved):
LGTM.

anntians:
The commit [Add more detailed error messages for KNN model training](https://github.com/opensearch-project/k-NN/pull/2348/commits/9538aadbf3c3e7923a5bf5da7b00f92008906ecd) was pushed by accident. It was meant for another PR on another branch.

jmazanec15 (approved):

kotwanikunal:
@shatejas / @navneet1v / @Vikasht34  - Can you please review this one?

VijayanB (approved):

navneet1v (approved):

navneet1v:
@anntians can you please port this change for 2.x branch and create a new PR

anntians:
> @anntians can you please port this change for 2.x branch and create a new PR

Sure thing, linked is the backport PR: https://github.com/opensearch-project/k-NN/pull/2379

Issue #2319, opened 2024-12-11
[BUG] Different score translations for ann search and exact search


wdongyu:
**What is the bug?**
After introducing the settings `index.knn.advanced.approximate_threshold` in https://github.com/opensearch-project/k-NN/pull/2188, we may encounter such a scenario that two segments exist in one single shard. Suppose Segment_1 includes a graph and segment_2 does not, and they both include the same vector **X**.

When we conduct a search with a vector **Q**, we will get two different scores for the same vector **X**, because score translations for ann search and exact search are sightly different. For example, for cosine metric we have a score( `1 / (2 - cos(Q, X))`) in ann search, but get another score (`(1 + cos(Q, X)) / 2`) in exact search.

**How can one reproduce the bug?**
1.  Create a index with nmslib and cosine distance, set `index.knn.advanced.approximate_threshold` to -1, so that it **never** build a graph:
```
PUT test_nmslib_cosine
{
  "settings": {
    "index.knn": true,
    "index.knn.advanced.approximate_threshold": "-1",
    "index.number_of_shards": 1,
    "index.number_of_replicas": 0
  },
  "mappings": {
    "properties": {
      "target_field": {
        "type": "knn_vector",
        "dimension": 2,
        "method": {
          "engine": "nmslib",
          "space_type": "cosinesimil",
          "name": "hnsw"
        }
      }
    }
  }
}
```

2. Ingest a doc with vector[0.6, 0.8]:
```
POST test_nmslib_cosine/_doc/1?refresh
{
  "target_field": [0.6, 0.8] 
}
```

3. update `index.knn.advanced.approximate_threshold` to 1, so that it **always** build a graph:
```
PUT test_nmslib_cosine/_settings
{
  "index.knn.advanced.approximate_threshold": "1"
}
```

5. Ingest another doc with vector[0.6, 0.8]:
```
POST test_nmslib_cosine/_doc/2?refresh
{
  "target_field": [0.6, 0.8] 
}
```

6. Search the data:
```
POST test_nmslib_cosine/_search
{
  "query": {
    "knn": {
      "target_field": {
        "vector": [1, 2],
        "k": 10
      }
    }
  }
}
```

7. Get the result:
```
"hits": [
      {
        "_index": "test_nmslib_cosine",
        "_id": "1",
        "_score": 0.99193496,    // -> score = (1 + cos([1, 2], [0.6, 0.8])) / 2
        "_source": {
          "target_field": [
            0.6,
            0.8
[...]

PR #2357, opened 2024-12-26
Update cosine score translation for nmslib

VijayanB:
### Description
Nmslib has different score range than exact search. After introducing approximate threshold setting, now, both exact search and approx search can be executed as part of same knn search api. Hence, to keep it consitent, we are overriding the score translation for nmslib for approx search.

### Related Issues
#2319

### Check List


VijayanB:
Looks like flaky test failure since the same test is successful in Linux and Mac. Create GH issue
https://github.com/opensearch-project/k-NN/issues/2358

CHANGELOG.md
```diff
@@ -28,6 +28,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 * Allow validation for non knn index only after 2.17.0 (#2315)[https://github.com/opensearch-project/k-NN/pull/2315]
 * Release query vector memory after execution (#2346)[https://github.com/opensearch-project/k-NN/pull/2346]
 * Fix shard level rescoring disabled setting flag (#2352)[https://github.com/opensearch-project/k-NN/pull/2352]
+* Use one formula to calculate cosine similarity (#2357)[https://github.com/opensearch-project/k-NN/pull/2357]
```
jmazanec15, line 31: Probably enhancement instead of bug fix?

src/main/java/org/opensearch/knn/index/SpaceType.java
```diff
@@ -62,7 +62,10 @@ public float scoreToDistanceTranslation(float score) {
     COSINESIMIL("cosinesimil") {
         @Override
         public float scoreTranslation(float rawScore) {
-            return 1 / (1 + rawScore);
+            // To be consistent, we will be using same formula used by lucene as mentioned below
+            // https://github.com/apache/lucene/blob/0494c824e0ac8049b757582f60d085932a890800/lucene/core/src/java/org/apache/lucene/index/VectorSimilarityFunction.java#L73
+            // Here raw score = 1 - cosine(x,y), hence, formula will be updated to below to get final score as similar to Lucene.
```
jmazanec15, line 67: I think we need to make this explicit in the contract - maybe add to java doc API. Then, if an engine does something else, it needs to be overridden in the engine specific space translations (see the Faiss specific engine methods for more details).

jmazanec15 (commented):
thanks @VijayanB  - can you summarize impact of change in description. For instance, this only impacts nmslib, correct?

src/main/java/org/opensearch/knn/index/SpaceType.java
VijayanB: Updated

src/main/java/org/opensearch/knn/plugin/script/KNNScoringSpace.java
```diff
@@ -144,7 +144,12 @@ public CosineSimilarity(Object query, MappedFieldType fieldType) {
         protected BiFunction<float[], float[], Float> getScoringMethod(final float[] processedQuery) {
             SpaceType.COSINESIMIL.validateVector(processedQuery);
             float qVectorSquaredMagnitude = getVectorMagnitudeSquared(processedQuery);
-            return (float[] q, float[] v) -> 1 + KNNScoringUtil.cosinesimilOptimized(q, v, qVectorSquaredMagnitude);
+            // To be consistent, we will be using same formula used by lucene as mentioned below
+            // https://github.com/apache/lucene/blob/0494c824e0ac8049b757582f60d085932a890800/lucene/core/src/java/org/apache/lucene/index/VectorSimilarityFunction.java#L73
+            return (float[] q, float[] v) -> Math.max(
+                (1.0F + KNNScoringUtil.cosinesimilOptimized(q, v, qVectorSquaredMagnitude)) / 2.0F,
+                0.0F
+            );
```
navneet1v, line 152: will this create any BWC challenges, mainly in cases when there is a mix cluster. 

Also another thing:
shouldn't this code be 

```suggestion
            // To be consistent, we will be using same formula used by lucene as mentioned below
            // https://github.com/apache/lucene/blob/0494c824e0ac8049b757582f60d085932a890800/lucene/core/src/java/org/apache/lucene/index/VectorSimilarityFunction.java#L73
            return (float[] q, float[] v) -> Math.max(
                (2.0F - KNNScoringUtil.cosinesimilOptimized(q, v, qVectorSquaredMagnitude)) / 2.0F,
                0.0F
            );
```

isn't `KNNScoringUtil.cosinesimilOptimized(q, v, qVectorSquaredMagnitude)` == `rawScore` ?

src/test/java/org/opensearch/knn/index/NmslibIT.java
```diff
@@ -195,6 +195,64 @@ public void testEndToEnd() throws Exception {
...
+        List<KNNResult> knnResults = parseSearchResponse(responseBody, fieldName);
+        assertEquals(k, knnResults.size());
+
+        List<Float> actualScores = parseSearchResponseScore(responseBody, fieldName);
+
+        // both document should have identical score
+        assertEquals(actualScores.get(0), actualScores.get(1), 0.001);
+
+        // Delete index
+        deleteKNNIndex(indexName);
```
navneet1v, line 253: you don't need to delete the index. The base classes IT handles the deletion of indices

src/main/java/org/opensearch/knn/plugin/script/KNNScoringSpace.java
VijayanB: cosinesimilOptimized  = (dotProduct / (Math.sqrt(normalizedProduct)));

VijayanB: > will this create any BWC challenges, mainly in cases when there is a mix cluster.
I don't think it will affect the order of neighbors. I believe we should already have bwc test, let me check and confirm

navneet1v: > cosinesimilOptimized = (dotProduct / (Math.sqrt(normalizedProduct)));

Oh wait sorry my bad. This formula is after simplification. Can we please add the docs around how we have simplified it? Just for reference in future.

```math
1 - cos\theta = cosineSimil
```

```math
OS Score = (2 - cosineSimil)/2
```

```math
OSScore = (2 - (1 - cos\theta))/2

```

which simplifies to

```math
OSScore = (1 + cos\theta))/2

```

```math
cos\theta = cosinesimilOptimized
```

VijayanB: I see your point since this is not about approx search. Let me add version check.  Good Catch @navneet1v

VijayanB: updated

navneet1v (approved):
Will we be adding the BWC test for this change after this PR?

VijayanB:
> Will you be adding the BWC test for this change after this PR?

@navneet1v Yes. Will add it as next PR

VijayanB:
Rebased to resolve CHANGE.MD file conflict

navneet1v (approved):

jmazanec15 (approved):

PR #2365, opened 2025-01-03
Added the fix for NPE happening during merges when all vector fields docs are deleted in the segments getting merged

navneet1v:
### Description
Added the fix for NPE happening during merges when all vector fields docs are deleted in the segments getting merged.

### What is the bug:
Where there are a lot of deleted docs and we are trying to merge the segments, we are seeing the NPE exception coming as there were not docs containing vector and we are trying to iterate over vector values.

The issue is not always reproducible hence adding an IT is not possible here. Also with the changes add in 2.18 related to threshold based graph creation, the issue is now only limited to indices created <=2.16 version of Opensearch using BDV.

Attaching a script here(author: @jmazanec15 ) that helped in reproducing the issue. 
[merge-failure-2.sh.zip](https://github.com/user-attachments/files/18304323/merge-failure-2.sh.zip)

### Related Issues
NA

### Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Use one formula to calculate cosine similarity (#2357)[https://github.com/opensearch-project/k-NN/pull/2357]
 ### Bug Fixes
 * Fixing the bug when a segment has no vector field present for disk based vector search (#2282)[https://github.com/opensearch-project/k-NN/pull/2282]
+* Fix for NPE while merging segments after all the vector fields docs are deleted (#2365)[https://github.com/opensearch-project/k-NN/pull/2365]
 * Allow validation for non knn index only after 2.17.0 (#2315)[https://github.com/opensearch-project/k-NN/pull/2315]
 * Release query vector memory after execution (#2346)[https://github.com/opensearch-project/k-NN/pull/2346]
 * Fix shard level rescoring disabled setting flag (#2352)[https://github.com/opensearch-project/k-NN/pull/2352]
--- a/src/main/java/org/opensearch/knn/common/KNNVectorUtil.java
+++ b/src/main/java/org/opensearch/knn/common/KNNVectorUtil.java
@@ -7,9 +7,7 @@ package org.opensearch.knn.common;
 
 import lombok.AccessLevel;
 import lombok.NoArgsConstructor;
-import org.opensearch.knn.index.vectorvalues.KNNVectorValues;
 
-import java.io.IOException;
 import java.util.List;
 import java.util.Objects;
 
@@ -61,18 +59,5 @@ public class KNNVectorUtil {
             intArray[i] = integerList.get(i);
         }
         return intArray;
-    }
-
-    /**
-     * Iterates vector values once if it is not at start of the location,
-     * Intended to be done to make sure dimension and bytesPerVector are available
-     * @param vectorValues
-     * @throws IOException
-     */
-    public static void iterateVectorValuesOnce(final KNNVectorValues<?> vectorValues) throws IOException {
-        if (vectorValues.docId() == -1) {
-            vectorValues.nextDoc();
-            vectorValues.getVector();
-        }
     }
 }
--- a/src/main/java/org/opensearch/knn/index/codec/util/KNNCodecUtil.java
+++ b/src/main/java/org/opensearch/knn/index/codec/util/KNNCodecUtil.java
@@ -9,12 +9,15 @@ import lombok.NonNull;
 import org.apache.lucene.index.BinaryDocValues;
 import org.apache.lucene.index.FieldInfo;
 import org.apache.lucene.index.SegmentInfo;
+import org.apache.lucene.search.DocIdSetIterator;
 import org.opensearch.knn.common.FieldInfoExtractor;
 import org.opensearch.knn.common.KNNConstants;
 import org.opensearch.knn.index.VectorDataType;
 import org.opensearch.knn.index.codec.KNN80Codec.KNN80BinaryDocValues;
 import org.opensearch.knn.index.engine.KNNEngine;
+import org.opensearch.knn.index.vectorvalues.KNNVectorValues;
 
+import java.io.IOException;
 import java.util.Comparator;
 import java.util.List;
 import java.util.stream.Collectors;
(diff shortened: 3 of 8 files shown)


vibrantvarun:
Checks failing

CHANGELOG.md
```diff
@@ -26,6 +26,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Add check to directly use ANN Search when filters match all docs. (#2320)[https://github.com/opensearch-project/k-NN/pull/2320]
 ### Bug Fixes
 * Fixing the bug when a segment has no vector field present for disk based vector search (#2282)[https://github.com/opensearch-project/k-NN/pull/2282]
+* Added the fix for NPE happening during merges when all vector fields docs are deleted in the segments getting merged (#2365)[https://github.com/opensearch-project/k-NN/pull/2365]
```
vibrantvarun, line 29:
```suggestion
*  Fix for NPE while merging segments after all the vector fields docs are deleted (#2365)[https://github.com/opensearch-project/k-NN/pull/2365]
```

src/main/java/org/opensearch/knn/common/KNNVectorUtil.java
```diff
@@ -62,17 +60,4 @@ public static int[] intListToArray(final List<Integer> integerList) {
...
         return intArray;
     }
-
-    /**
-     * Iterates vector values once if it is not at start of the location,
-     * Intended to be done to make sure dimension and bytesPerVector are available
-     * @param vectorValues
-     * @throws IOException
-     */
-    public static void iterateVectorValuesOnce(final KNNVectorValues<?> vectorValues) throws IOException {
```
vibrantvarun, line 72: Why we have moved this to KNNCodecUtil?

navneet1v: i feel that was not the right place for this util. As the function was used only in codec so I moved it.

CHANGELOG.md
navneet1v: fixed

vibrantvarun (approved):
LGTM Thanks

navneet1v:
@jmazanec15 and @shatejas can you please take a look

jmazanec15 (approved):
LGTM

src/main/java/org/opensearch/knn/index/codec/util/KNNCodecUtil.java
```diff
@@ -116,6 +119,31 @@ public static String getNativeEngineFileFromFieldInfo(FieldInfo field, SegmentIn
...
+     * Positions the vectorValuesIterator to the first vector document ID if not already positioned there.
+     * This initialization is crucial for setting up vector dimensions and other properties in VectorValues.
+     *
+     * If the VectorValues contains no vector documents, the iterator will be positioned at
+     * {@link DocIdSetIterator#NO_MORE_DOCS}
+     *
+     * @param vectorValues {@link KNNVectorValues}
+     * @throws IOException if there is an error while accessing the vector values
+     */
+    public static void iterateVectorValuesOnce(final KNNVectorValues<?> vectorValues) throws IOException {
```
Vikasht34, line 132: This part of code will still have prone to issue, Can we do something like this and it's seems confusing as well , 



```
1. First thing Check if there is no more docs or we are at end of iteration 

if (vectorValues.docId() == DocIdSetIterator.NO_MORE_DOCS) {
        return;
    }
    
    2. if doc id is -1 then go next 
    // Move to the next document if not yet started
    if (vectorValues.docId() == -1) {
        vectorValues.nextDoc();
    }
    3.third thing is ensure we are not at end 

    if (vectorValues.docId() != DocIdSetIterator.NO_MORE_DOCS) {
        vectorValues.getVector();
    }
```

navneet1v: > This part of code will still have prone to issue, Can we do something like this and it's seems confusing as well ,

can you please elaborate on this point more?

Vikasht34: The Function name is iterateVectorValuesOnce , I assume that this function will be called only once for Vector Value .when we start iterating , which means initially we have DOCID = -1 , but the first line says if doc id != -1 , return which means without iterating we returned. since we have -1 intially. Is this correct understanding ?

if (vectorValues.docId() != -1) {
            return;
        }

navneet1v: > The Function name is iterateVectorValuesOnce , I assume that this function will be called only once for Vector Value .when we start iterating ,

No thats not the case. Since we cannot control how many times this function is called, I wanted to make this consistent even when it is called multiple times. May be a better name can be initVectorValues. WDYT? because at the end of the day it is just initing the vector values.

src/test/java/org/opensearch/knn/index/codec/util/KNNCodecUtilTests.java
```diff
@@ -46,4 +52,23 @@ public void testGetKNNEngines() {
...
+        // Give
+        final List<float[]> floatArray = List.of(new float[] { 1, 2 }, new float[] { 2, 3 });
+        final int dimension = floatArray.get(0).length;
+        final TestVectorValues.PreDefinedFloatVectorValues randomVectorValues = new TestVectorValues.PreDefinedFloatVectorValues(
+            floatArray
+        );
+        final KNNVectorValues<float[]> knnVectorValues = KNNVectorValuesFactory.getVectorValues(VectorDataType.FLOAT, randomVectorValues);
+
+        // When
+        iterateVectorValuesOnce(knnVectorValues);
```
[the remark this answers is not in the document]
navneet1v: sure will add.

src/main/java/org/opensearch/knn/index/codec/util/KNNCodecUtil.java
navneet1v: @shatejas I am thinking to change the name of the function to initVectorValues as that makes the function more readable. I know you have added this function sometime back I hope that fine here. and I am not missing any other context here.

src/test/java/org/opensearch/knn/index/codec/util/KNNCodecUtilTests.java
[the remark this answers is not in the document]
navneet1v: added the test

jmazanec15 (approved):

shatejas (approved):

PR #2374, opened 2025-01-09
Fixed a bug that does not get a space type from index setting when it…

0ctopus13prime:
### Description

### Related Issues

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/engine/SpaceTypeResolver.java
+++ b/src/main/java/org/opensearch/knn/index/engine/SpaceTypeResolver.java
@@ -6,7 +6,9 @@
 package org.opensearch.knn.index.engine;
 
 import org.apache.logging.log4j.util.Strings;
+import org.opensearch.common.settings.Settings;
 import org.opensearch.index.mapper.MapperParsingException;
+import org.opensearch.knn.index.KNNSettings;
 import org.opensearch.knn.index.SpaceType;
 import org.opensearch.knn.index.VectorDataType;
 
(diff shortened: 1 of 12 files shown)


src/main/java/org/opensearch/knn/index/engine/SpaceTypeResolver.java
```diff
@@ -25,24 +27,30 @@ public final class SpaceTypeResolver {
     private SpaceTypeResolver() {}
 
     /**
-     * Resolves space type from configuration details. It is guaranteed not to return either null or
-     * {@link SpaceType#UNDEFINED}
+     * Resolves space type from configuration details. It is guaranteed not to return null.
+     * When space is not in either method and top level, UNDEFINED will be returned.
+     * This can happen when it is defined at index level which is deprecated and no longer allowed in the future.
+     * In this case, UNDEFINED will be returned.
```
Vikasht34, line 33: I was just curious to know here , as Let's say if I have Old Index where it is defined at Index.knn.space_type so in this method , are you not trying to take that , if yes then this statement is (This can happen when it is defined at index level which is deprecated and no longer allowed in the future.
     * In this case, UNDEFINED will be returned.) needs modification right?

src/main/java/org/opensearch/knn/index/engine/SpaceTypeResolver.java
```diff
@@ -25,24 +27,30 @@ public final class SpaceTypeResolver {
...
-        final VectorDataType vectorDataType,
-        final String topLevelSpaceTypeString
+        final String topLevelSpaceTypeString,
+        final Settings indexSettings,
+        final VectorDataType vectorDataType
     ) {
         SpaceType methodSpaceType = getSpaceTypeFromMethodContext(knnMethodContext);
         SpaceType topLevelSpaceType = getSpaceTypeFromString(topLevelSpaceTypeString);
 
         if (isSpaceTypeConfigured(methodSpaceType) == false && isSpaceTypeConfigured(topLevelSpaceType) == false) {
```
Vikasht34, line 48: Can we add comments above this line on new bahviuour

Vikasht34 (approved):

src/main/java/org/opensearch/knn/index/engine/SpaceTypeResolver.java
0ctopus13prime: Sure will do

src/main/java/org/opensearch/knn/index/engine/SpaceTypeResolver.java
```diff
@@ -25,24 +27,30 @@ public final class SpaceTypeResolver {
...
-            return getSpaceTypeFromVectorDataType(vectorDataType);
+            final String spaceType = indexSettings.get(KNNSettings.INDEX_KNN_SPACE_TYPE.getKey());
```
shatejas, line 49: Isn't there a possibility of an NPE considering indexSettings value is null in `RestTrainModelHandler.java` line 170

0ctopus13prime: yes, it is...
will fix this in the next revision.

0ctopus13prime: technically, `topLevelSpaceType` will be guaranteed to have a value, control flow won't reach index settings.
But good to have a defence there

Vikasht34:
Looks good

Vikasht34 (approved):

0ctopus13prime:
@shatejas 
Could you approve this if it looks good on you?

shatejas (approved):

Issue #2242, opened 2024-10-31
Implement cosine for faiss engine


jmazanec15:
## Description

Cosine similarity is one of the more popular space types. faiss does not support it directly. Instead, they prefer to have data be normalized and then use the inner product (which is equivalent to cosine --> <u,v>/||u||||v|| = cos(theta) (https://github.com/facebookresearch/faiss/blob/2c961cc308ade8a85b3aa10a550728ce3387f625/README.md?plain=1#L11). We should figure out how to add cosine for faiss now that it is default (#2163)

We have a couple different options:
1. We could normalize data before serializing to flat vector values
2. We could just normalize when passing to faiss and then for exact scoring lazily compute norms and normalize
3. We could compute the norm and then use it to divide the dot product when scoring is needed or when vectors need to be normalized.

I prefer option 3 mainly because it will allow us to use use knn vector values as synthetic source (see #1571), but also let us be efficient on search.

We would need to investigate how best to do this. One simple way to do it would be to add one extra dimension and store the normalized value there.

luyuncheng:
we added an ingest pipeline to normalize the data for faiss engine which want to Cosine similarity

heemin32:
I prefer option 2. It is more simple as we keep everything as is and only pass normalized data to faiss engine. Could you tell why we need `exact scoring lazily compute norms and normalize` for option 2?

jmazanec15:
For option 2, for re-scoring (exact search), because the vectors are not normalized, we would need to compute norms, as we do now.

vamshin:
> we added an ingest pipeline to normalize the data for faiss engine who want to Cosine similarity

@luyuncheng  interesting. What are your thoughts on having this support in k-NN plugin VS Ingest pipeline approach?

heemin32:
> For option 2, for re-scoring (exact search), because the vectors are not normalized, we would need to compute norms, as we do now.

For re-scoring we can just use cosine distance calculation instead of (normalization + innerproduct)?

jmazanec15:
cosine distance calculation is implemented as normalization + innerproduct I believe

luyuncheng:
> > we added an ingest pipeline to normalize the data for faiss engine who want to Cosine similarity
> 
> @luyuncheng interesting. What are your thoughts on having this support in k-NN plugin VS Ingest pipeline approach?

@vamshin @jmazanec15 @heemin32 as @jmazanec15 says
 
[...]

PR #2376, opened 2025-01-09
Add cosine similarity support for faiss engine

VijayanB:
### Description
FAISS engine doesn't support cosine similarity natively. However we can use inner product to achieve the same, because, when vectors are normalized then inner product will be same as cosine similarity. Hence, before ingestion, normalize the input vector, and add it to faiss index with type as inner product, and, before search, normalize query vector if space type is cosine and engine is faiss.

Since we will be storing normalized vector in segments, we don't have to normalize whenever segments are merged. This will keep force merge time and search at competitive, provided we will face additional latency during indexing (one time where we normalize). To avoid this additional latency, customers can normalize their data set and create inner product.

This also adds support to radial search, for both max distance and min score.

### Related Issues
#2242 

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/engine/AbstractKNNMethod.java
+++ b/src/main/java/org/opensearch/knn/index/engine/AbstractKNNMethod.java
@@ -13,6 +13,8 @@ import org.opensearch.knn.index.VectorDataType;
 import org.opensearch.knn.index.mapper.PerDimensionProcessor;
 import org.opensearch.knn.index.mapper.PerDimensionValidator;
 import org.opensearch.knn.index.mapper.SpaceVectorValidator;
+import org.opensearch.knn.index.mapper.VectorTransformer;
+import org.opensearch.knn.index.mapper.VectorTransformerFactory;
 import org.opensearch.knn.index.mapper.VectorValidator;
 
 import java.util.ArrayList;
@@ -106,6 +108,10 @@ public abstract class AbstractKNNMethod implements KNNMethod {
         return PerDimensionProcessor.NOOP_PROCESSOR;
     }
 
+    protected VectorTransformer getVectorTransformer(KNNMethodContext knnMethodContext) {
+        return VectorTransformerFactory.getVectorTransformer(knnMethodContext);
+    }
+
     @Override
     public KNNLibraryIndexingContext getKNNLibraryIndexingContext(
         KNNMethodContext knnMethodContext,
@@ -116,7 +122,7 @@ public abstract class AbstractKNNMethod implements KNNMethod {
             knnMethodConfigContext
         );
         Map<String, Object> parameterMap = knnLibraryIndexingContext.getLibraryParameters();
-        parameterMap.put(KNNConstants.SPACE_TYPE, knnMethodContext.getSpaceType().getValue());
+        parameterMap.put(KNNConstants.SPACE_TYPE, convertUserToMethodSpaceType(knnMethodContext.getSpaceType()).getValue());
         parameterMap.put(KNNConstants.VECTOR_DATA_TYPE_FIELD, knnMethodConfigContext.getVectorDataType().getValue());
         return KNNLibraryIndexingContextImpl.builder()
             .quantizationConfig(knnLibraryIndexingContext.getQuantizationConfig())
@@ -124,11 +130,29 @@ public abstract class AbstractKNNMethod implements KNNMethod {
             .vectorValidator(doGetVectorValidator(knnMethodContext, knnMethodConfigContext))
             .perDimensionValidator(doGetPerDimensionValidator(knnMethodContext, knnMethodConfigContext))
             .perDimensionProcessor(doGetPerDimensionProcessor(knnMethodContext, knnMethodConfigContext))
+            .vectorTransformer(getVectorTransformer(knnMethodContext))
             .build();
     }
 
     @Override
     public KNNLibrarySearchContext getKNNLibrarySearchContext() {
         return knnLibrarySearchContext;
+    }
+
+    /**
+     * Converts user defined space type to method space type that is supported by library.
+     * The subclass can override this method and returns the appropriate space type that
+     * is supported by the library. This is required because, some libraries may not
+     * support all the space types supported by OpenSearch, however. this can be achieved by using compatible space type by the library.
+     * For example, faiss does not support cosine similarity. However, we can use inner product space type for cosine similarity after normalization.
+     * In this case, we can return the inner product space type for cosine similarity.
+     *
+     * @param spaceType The space type to check for compatibility
+     * @return The compatible space type for the given input, returns the same
+     *         space type if it's already compatible
+     * @see SpaceType
+     */
+    protected SpaceType convertUserToMethodSpaceType(SpaceType spaceType) {
+        return spaceType;
     }
 }
--- a/src/main/java/org/opensearch/knn/index/engine/AbstractKNNMethod.java
+++ b/src/main/java/org/opensearch/knn/index/engine/AbstractKNNMethod.java
@@ -108,8 +108,8 @@ public abstract class AbstractKNNMethod implements KNNMethod {
         return PerDimensionProcessor.NOOP_PROCESSOR;
     }
 
-    protected VectorTransformer getVectorTransformer(KNNMethodContext knnMethodContext) {
-        return VectorTransformerFactory.getVectorTransformer(knnMethodContext.getKnnEngine(), knnMethodContext.getSpaceType());
+    protected VectorTransformer getVectorTransformer(SpaceType spaceType) {
+        return VectorTransformerFactory.NOOP_VECTOR_TRANSFORMER;
     }
 
     @Override
@@ -130,7 +130,7 @@ public abstract class AbstractKNNMethod implements KNNMethod {
             .vectorValidator(doGetVectorValidator(knnMethodContext, knnMethodConfigContext))
             .perDimensionValidator(doGetPerDimensionValidator(knnMethodContext, knnMethodConfigContext))
             .perDimensionProcessor(doGetPerDimensionProcessor(knnMethodContext, knnMethodConfigContext))
-            .vectorTransformer(getVectorTransformer(knnMethodContext))
+            .vectorTransformer(getVectorTransformer(knnMethodContext.getSpaceType()))
             .build();
     }
 
--- a/src/main/java/org/opensearch/knn/index/engine/AbstractKNNMethod.java
+++ b/src/main/java/org/opensearch/knn/index/engine/AbstractKNNMethod.java
@@ -109,7 +109,7 @@ public abstract class AbstractKNNMethod implements KNNMethod {
     }
 
     protected VectorTransformer getVectorTransformer(KNNMethodContext knnMethodContext) {
-        return VectorTransformerFactory.getVectorTransformer(knnMethodContext);
+        return VectorTransformerFactory.getVectorTransformer(knnMethodContext.getKnnEngine(), knnMethodContext.getSpaceType());
     }
 
     @Override
--- a/src/main/java/org/opensearch/knn/index/engine/KNNLibraryIndexingContext.java
+++ b/src/main/java/org/opensearch/knn/index/engine/KNNLibraryIndexingContext.java
@@ -8,6 +8,7 @@ package org.opensearch.knn.index.engine;
 import org.opensearch.knn.index.engine.qframe.QuantizationConfig;
 import org.opensearch.knn.index.mapper.PerDimensionProcessor;
 import org.opensearch.knn.index.mapper.PerDimensionValidator;
+import org.opensearch.knn.index.mapper.VectorTransformer;
 import org.opensearch.knn.index.mapper.VectorValidator;
 
 import java.util.Map;
@@ -47,4 +48,12 @@ public interface KNNLibraryIndexingContext {
      * @return Get the per dimension processor
      */
     PerDimensionProcessor getPerDimensionProcessor();
+
+    /**
+     * Get the vector transformer that will be used to transform the vector before indexing.
+     * This will be applied at vector level once entire vector is parsed and validated.
+     *
+     * @return VectorTransformer
+     */
+    VectorTransformer getVectorTransformer();
 }
--- a/src/main/java/org/opensearch/knn/index/mapper/FlatVectorFieldMapper.java
+++ b/src/main/java/org/opensearch/knn/index/mapper/FlatVectorFieldMapper.java
@@ -109,4 +109,9 @@ public class FlatVectorFieldMapper extends KNNVectorFieldMapper {
     protected PerDimensionProcessor getPerDimensionProcessor() {
         return PerDimensionProcessor.NOOP_PROCESSOR;
     }
+
+    @Override
+    protected VectorTransformer getVectorTransformer() {
+        return VectorTransformer.NOOP_VECTOR_TRANSFORMER;
+    }
 }
--- a/src/main/java/org/opensearch/knn/index/mapper/FlatVectorFieldMapper.java
+++ b/src/main/java/org/opensearch/knn/index/mapper/FlatVectorFieldMapper.java
@@ -110,8 +110,4 @@ public class FlatVectorFieldMapper extends KNNVectorFieldMapper {
         return PerDimensionProcessor.NOOP_PROCESSOR;
     }
 
-    @Override
-    protected VectorTransformer getVectorTransformer() {
-        return VectorTransformer.NOOP_VECTOR_TRANSFORMER;
-    }
 }
--- a/src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java
+++ b/src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java
@@ -689,11 +689,21 @@ public abstract class KNNVectorFieldMapper extends ParametrizedFieldMapper {
     protected abstract PerDimensionProcessor getPerDimensionProcessor();
 
     /**
-     * Getter for vector transformer after vector parsing and validation
+     * Retrieves the vector transformer for the KNN vector field.
+     * This method provides access to the vector transformer instance that will be used
+     * for processing vectors in the KNN field. The transformer is responsible for any
+     * necessary vector transformations before indexing or searching.
+     * This implementation delegates to the VectorTransformerFactory to obtain
+     * the appropriate transformer instance. The returned transformer is typically
+     * stateless and thread-safe.
      *
-     * @return VectorTransformer
+     * @return VectorTransformer An instance of VectorTransformer that will be used
+     *         for vector transformations in this field
+     *
      */
-    protected abstract VectorTransformer getVectorTransformer();
+    protected VectorTransformer getVectorTransformer() {
+        return VectorTransformerFactory.getVectorTransformer();
+    }
 
     protected void parseCreateField(ParseContext context, int dimension, VectorDataType vectorDataType) throws IOException {
         validatePreparse();
--- a/src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java
+++ b/src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java
@@ -675,7 +675,7 @@ public abstract class KNNVectorFieldMapper extends ParametrizedFieldMapper {
     protected abstract VectorValidator getVectorValidator();
 
     /**
-     * Getter for per dimension validator during vector parsing
+     * Getter for per dimension validator during vector parsing, and before any transformation
      *
      * @return PerDimensionValidator
      */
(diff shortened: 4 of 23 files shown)


VijayanB:
Adding additional unit and integration test for radial search. Will mark it as ready and after adding few more tests

src/main/java/org/opensearch/knn/index/SpaceType.java
```diff
@@ -77,6 +77,11 @@ public float scoreTranslation(float rawScore) {
             return Math.max((2.0F - rawScore) / 2.0F, 0.0F);
         }
 
+        @Override
+        public float scoreToDistanceTranslation(float score) {
+            return score;
```
jmazanec15, line 82: Confused - Why is this correct?

src/main/java/org/opensearch/knn/index/engine/AbstractKNNMethod.java
```diff
@@ -116,19 +122,35 @@ public KNNLibraryIndexingContext getKNNLibraryIndexingContext(
...
 
     @Override
     public KNNLibrarySearchContext getKNNLibrarySearchContext() {
         return knnLibrarySearchContext;
     }
+
+    /**
+     * Gets the compatible space type for the given space type parameter.
+     * The subclass can override this method and returns the appropriate space type that
+     * is compatible with the library.
```
jmazanec15, line 145: Can you add context on why this is needed via an example?

src/main/java/org/opensearch/knn/index/engine/AbstractKNNMethod.java
```diff
@@ -116,19 +122,35 @@ public KNNLibraryIndexingContext getKNNLibraryIndexingContext(
...
+     *
+     * @param spaceType The space type to check for compatibility
+     * @return The compatible space type for the given input, returns the same
+     *         space type if it's already compatible
+     * @see SpaceType
+     */
+
```
jmazanec15, line 152: nit: extra line

src/main/java/org/opensearch/knn/index/engine/AbstractKNNMethod.java
```diff
@@ -116,19 +122,35 @@ public KNNLibraryIndexingContext getKNNLibraryIndexingContext(
...
+    protected SpaceType getCompatibleSpaceType(SpaceType spaceType) {
```
jmazanec15, line 153: Can we call something like convertUserToMethodSpaceType? The naming here doesnt seem to give it enough description.

src/main/java/org/opensearch/knn/index/engine/KNNLibraryIndexingContext.java
```diff
@@ -47,4 +48,6 @@ public interface KNNLibraryIndexingContext {
      * @return Get the per dimension processor
      */
     PerDimensionProcessor getPerDimensionProcessor();
+
+    VectorTransformer getVectorTransformer();
```
jmazanec15, line 52: javadoc?

src/main/java/org/opensearch/knn/index/mapper/ModelFieldMapper.java
```diff
@@ -192,6 +193,37 @@ protected PerDimensionProcessor getPerDimensionProcessor() {
...
+        if (vectorTransformer != null) {
+            return;
+        }
+        ModelMetadata modelMetadata = getModelMetadata(modelDao, modelId);
+
+        KNNMethodContext knnMethodContext = getKNNMethodContextFromModelMetadata(modelMetadata);
+        KNNMethodConfigContext knnMethodConfigContext = getKNNMethodConfigContextFromModelMetadata(modelMetadata);
+        // Need to handle BWC case
+        if (knnMethodContext == null || knnMethodConfigContext == null) {
+            vectorTransformer = VectorTransformerFactory.getVectorTransformer(modelMetadata.getKnnEngine(), modelMetadata.getSpaceType());
```
jmazanec15, line 218: Can we add log here?

src/main/java/org/opensearch/knn/index/mapper/VectorTransformer.java
```diff
@@ -0,0 +1,47 @@
...
+ * for specific use cases such as normalization, scaling, or other transformations.
+ *
+ * <p>This interface provides default implementations that pass through the original
+ * vector without modification. Implementing classes should override these methods
+ * to provide specific transformation logic.
+ */
+public interface VectorTransformer {
+
+    /**
+     * Transforms a float vector into a new vector of the same type.
```
jmazanec15, line 20: Will this be in place or always a copy?

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
```diff
@@ -611,7 +612,13 @@ private void updateQueryStats(VectorQueryType vectorQueryType) {
         }
     }
 
-    private float[] getVectorForCreatingQueryRequest(VectorDataType vectorDataType, KNNEngine knnEngine) {
+    private float[] getVectorForCreatingQueryRequest(VectorDataType vectorDataType, KNNEngine knnEngine, SpaceType spaceType) {
+
+        // Cosine similarity is supported as Inner product by FAISS by normalizing input vector, hence, we have to normalize
```
jmazanec15, line 617: Can we move this check out of this class? This class is already very crowded and I want to avoid adding more checks around engines. Instead, could we investigate either adding it to https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/engine/KNNLibrarySearchContext.java and/or adding a method in KNNVectorFieldType (https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldType.java#L85) that says "should normalize query" or, better yet, transformQuery?

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java
```diff
@@ -681,6 +681,8 @@ protected void validatePreparse() {
      */
     protected abstract PerDimensionProcessor getPerDimensionProcessor();
 
+    protected abstract VectorTransformer getVectorTransformer();
```
jmazanec15, line 684: Can you add java doc for this one? Also, can you update javadoc for getVectorValidator() to say that it is validated before any transform calls?

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java
```diff
@@ -700,7 +703,8 @@ protected void parseCreateField(ParseContext context, int dimension, VectorDataT
             }
             final float[] array = floatsArrayOptional.get();
             getVectorValidator().validateVector(array);
-            context.doc().addAll(getFieldsForFloatVector(array));
+            final float[] transformedArray = getVectorTransformer().transform(array);
```
jmazanec15, line 706: This will be called before the per-dimension processor too? What should contract around these 2 be? Im wondering if we even need the per-dimension or if we can wrap that in this new full vector transform.

jmazanec15 (commented):
Thanks @VijayanB - completed a first pass review

src/main/java/org/opensearch/knn/index/mapper/VectorTransformer.java
VijayanB: IMO, Transformer should generally return a new value rather than modifying the original value it receives, to preserve immutability.

src/main/java/org/opensearch/knn/index/engine/KNNLibraryIndexingContext.java
VijayanB: Ack

src/main/java/org/opensearch/knn/index/engine/AbstractKNNMethod.java
VijayanB: Ack

src/main/java/org/opensearch/knn/index/engine/AbstractKNNMethod.java
VijayanB: Ack

src/main/java/org/opensearch/knn/index/engine/AbstractKNNMethod.java
VijayanB: Ack

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java
VijayanB: Ack

src/main/java/org/opensearch/knn/index/SpaceType.java
VijayanB: updated

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
VijayanB: Moved it to KNNVectorFieldType

jmazanec15 (approved):

jmazanec15 (approved):

src/main/java/org/opensearch/knn/index/mapper/VectorTransformer.java
```diff
@@ -0,0 +1,43 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.opensearch.knn.index.mapper;
+
+/**
+ * Defines operations for transforming vectors in the k-NN search context.
+ * Implementations can modify vectors while preserving their dimensional properties
+ * for specific use cases such as normalization, scaling, or other transformations.
+ */
+public interface VectorTransformer {
```
Vikasht34, line 12: Please make this interface generic.

src/main/java/org/opensearch/knn/index/mapper/VectorTransformer.java
```diff
@@ -0,0 +1,43 @@
...
+    default void transform(final byte[] vector) {
+        if (vector == null) {
+            throw new IllegalArgumentException("Input vector cannot be null");
+        }
+    }
+
+    /**
+     * A no-operation transformer that returns vector values unchanged.
+     */
+    VectorTransformer NOOP_VECTOR_TRANSFORMER = new VectorTransformer() {
```
Vikasht34, line 41: Geting Instance of VectorTransformer is not job of Contract basesd Interface, It's responsibility of Factory, Please have factory to handle this.

src/main/java/org/opensearch/knn/index/mapper/VectorTransformerFactory.java
```diff
@@ -0,0 +1,78 @@
...
+import org.opensearch.knn.index.engine.KNNEngine;
+import org.opensearch.knn.index.engine.KNNMethodContext;
+import org.opensearch.knn.indices.ModelMetadata;
+
+/**
+ * Factory class responsible for creating appropriate vector transformers based on the KNN method context.
+ * This factory determines whether vectors need transformation based on the engine type and space type.
+ */
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+public final class VectorTransformerFactory {
```
Vikasht34, line 20: Please have one method which takes , engine and space type, how to get engine and space type should not be responsibility of Factory. Example below factory only acts on params becuase that's the deciding factor , how to create final object. I can not have mutiple methods from which I need to get param to get final Object

```
/**
     * Retrieves a quantizer instance based on the provided quantization parameters.
     *
     * @param params the quantization parameters used to determine the appropriate quantizer
     * @param <P>    the type of quantization parameters, extending {@link QuantizationParams}
     * @param <T>    the type of the input vector to be quantized
     * @param <R>    the type of the output after quantization
     * @return an instance of {@link Quantizer} corresponding to the provided parameters
     */
    public static <P extends QuantizationParams, T, R> Quantizer<T, R> getQuantizer(final P params) {
        if (params == null) {
            throw new IllegalArgumentException("Quantization parameters must not be null.");
        }
        // Lazy Registration instead of static block as class level;
        ensureRegistered();
        return (Quantizer<T, R>) QuantizerRegistry.getQuantizer(params);
    }

```

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldType.java
```diff
@@ -115,4 +119,37 @@ public RescoreContext resolveRescoreContext(RescoreContext userProvidedContext)
...
+     * @param vector The float array to be transformed in-place. Must not be null.
+     * @throws IllegalStateException if neither KNN method context nor Model ID is configured
+     *
+     * The transformation process follows this order:
+     * 1. If vector is not FLOAT type, no transformation is performed
+     * 2. Attempts to use KNN method context if present
+     * 3. Falls back to model ID if KNN method context is not available
+     * 4. Throws exception if neither configuration is present
+     */
+    public void transformQueryVector(float[] vector) {
```
Vikasht34, line 136: It's very un-consistent with the whole flow , in Indexing flow , I am tranforming through vectorTranformer inteface and to get VectorTransformer Interface , we are getting through factory. But here ..we are directly getting from factory and transforming it. 

That's why I told either we have inteface based transformer or completely remove it. we can not have two ways to do same thing and query and Indexing

src/main/java/org/opensearch/knn/index/mapper/MethodFieldMapper.java
```diff
@@ -196,4 +198,9 @@ protected PerDimensionValidator getPerDimensionValidator() {
     protected PerDimensionProcessor getPerDimensionProcessor() {
         return perDimensionProcessor;
     }
+
```
Vikasht34, line 201: Why this method if I can directly write , because all we want  a access to property of this class which is even getting initialised in constructor.
@getter
private final VectorTransformer vectorTransformer;

Vikasht34 (commented):
Overall I think we should try making this whole transformation process simpler , by simplifying the class and interfaces.

src/main/java/org/opensearch/knn/index/engine/AbstractKNNMethod.java
```diff
@@ -106,6 +108,10 @@ protected PerDimensionProcessor doGetPerDimensionProcessor(
         return PerDimensionProcessor.NOOP_PROCESSOR;
     }
 
+    protected VectorTransformer getVectorTransformer(KNNMethodContext knnMethodContext) {
```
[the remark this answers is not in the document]
VijayanB: AbstractKNNMethod builds KNNLibraryIndexingContext  that is why the method is added here

src/main/java/org/opensearch/knn/index/mapper/MethodFieldMapper.java
VijayanB: This is needed by super class.

src/main/java/org/opensearch/knn/index/engine/AbstractKNNMethod.java
```diff
@@ -116,19 +122,37 @@ public KNNLibraryIndexingContext getKNNLibraryIndexingContext(
...
+     * support all the space types supported by OpenSearch, however. this can be achieved by using compatible space type by the library.
+     * For example, faiss does not support cosine similarity. However, we can use inner product space type for cosine similarity after normalization.
+     * In this case, we can return the inner product space type for cosine similarity.
+     *
+     * @param spaceType The space type to check for compatibility
+     * @return The compatible space type for the given input, returns the same
+     *         space type if it's already compatible
+     * @see SpaceType
+     */
+    protected SpaceType convertUserToMethodSpaceType(SpaceType spaceType) {
```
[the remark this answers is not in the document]
VijayanB: Here it is used as delegate to get Space Type based on user defined space type. Not all of its subclass needs this, thats why it is no op at base class. Only subclass like Faiss which needs can provide non default implementation

src/main/java/org/opensearch/knn/index/mapper/FlatVectorFieldMapper.java
```diff
@@ -109,4 +109,9 @@ protected PerDimensionValidator getPerDimensionValidator() {
     protected PerDimensionProcessor getPerDimensionProcessor() {
         return PerDimensionProcessor.NOOP_PROCESSOR;
     }
+
+    @Override
+    protected VectorTransformer getVectorTransformer() {
```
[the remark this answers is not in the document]
VijayanB: Ack. i made this method not abstract and added default implementation to avoid overriding it every place

src/main/java/org/opensearch/knn/index/mapper/VectorTransformer.java
VijayanB: Can you add more context on what are you thinking about generic mean here?

src/main/java/org/opensearch/knn/index/mapper/VectorTransformer.java
VijayanB: Good point. I updated it.

src/main/java/org/opensearch/knn/index/mapper/VectorTransformerFactory.java
VijayanB: IMO, Either creating POJO or Container or Generic Map to put all values and creating one big method to parse this input is not scalable. Having multiple method with different input will provide greater visibility on the client on which method to use. Current i made all input params are final to avoid any side-effects. Given the current use of this class, i don't see it is complex enough to justify creating either builder or POJO for an input to this factory. I will add TODO to add builder if we see this class needs more input that what we have it now.

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldType.java
VijayanB: Unlike Indexing where we have faiss, lucene, and nmslib, Query doesn't have multiple implementation. So we don't need engine to provide which transformer to use, we can ask transformer factory to get the instance

Vikasht34: yeah but it's totally does not matter. ,we should have  a consistent way to get object for Transformeres , It should not be via direct factory for my query and via interface to Indexing ...

in this case also we should be aming for or  prefer going via either of one route ..either via Interface or via factory ...so have uniform implementation.

src/main/java/org/opensearch/knn/index/engine/AbstractKNNMethod.java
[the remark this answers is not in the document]
Vikasht34: We don't need a delegate , when information is already availble at child classes, if any child needs transformation , they do it then and there , and other class keeps it origianal , becuase this class does not own space type as class level property.


Unlike there is a point to have this method in this class , because this class owns spaces as class level property.
```
protected final Set<SpaceType> spaces;
    protected final KNNLibrarySearchContext knnLibrarySearchContext;
    @Override
    public boolean isSpaceTypeSupported(SpaceType space) {
        return spaces.contains(space);
    }
    ```

File: src/main/java/org/opensearch/knn/index/engine/AbstractKNNMethod.java (hunk 16 above)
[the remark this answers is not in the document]
Comment (Vikasht34) [reply src/main/java/org/opensearch/knn/index/engine/AbstractKNNMethod.java:111]: Simple ,

If the getTransformer method is not relevant to all child classes, it breaks the principle of abstraction. Abstract classes should only include methods that are meaningful to all subclasses.

I don't see this case here , and if I have to give common functionality I would give like this 

```
public VectorTransformer getTransformer(KNNMethodContext context) {
        return VectorTransformer.NOOP_VECTOR_TRANSFORMER; // Default no-op
    }
```

File: src/main/java/org/opensearch/knn/index/mapper/VectorTransformer.java (hunk 11 above)
Comment (Vikasht34) [reply src/main/java/org/opensearch/knn/index/mapper/VectorTransformer.java:12]:
```
public interface VectorTransformer<T> {
    default void transform(final T vector) {
        if (vector == null) {
            throw new IllegalArgumentException("Input vector cannot be null");
        }
    }
}

```

File: src/main/java/org/opensearch/knn/index/engine/AbstractKNNMethod.java (hunk 17 above)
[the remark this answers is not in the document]
Comment (VijayanB) [reply src/main/java/org/opensearch/knn/index/engine/AbstractKNNMethod.java:155]: This is not about transformation. This is about what the underlying library supports as space type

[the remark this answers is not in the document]
Comment (Vikasht34) [reply src/main/java/org/opensearch/knn/index/engine/AbstractKNNMethod.java:155]: I was giving an example , Since Spaces are class level property of base class ,it make sense here , If you want any manupilation on space , make space as class level property of abstract class rather than child classes.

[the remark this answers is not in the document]
Comment (VijayanB) [reply src/main/java/org/opensearch/knn/index/engine/AbstractKNNMethod.java:155]: Updated method parameter to use space type instead of context

File: src/main/java/org/opensearch/knn/index/engine/KNNLibraryIndexingContext.java [hunk 19]
```diff
@@ -47,4 +48,12 @@ public interface KNNLibraryIndexingContext {
...
      */
     PerDimensionProcessor getPerDimensionProcessor();
+
+    /**
+     * Get the vector transformer that will be used to transform the vector before indexing.
+     * This will be applied at vector level once entire vector is parsed and validated.
+     *
+     * @return VectorTransformer
+     */
+    VectorTransformer getVectorTransformer();
```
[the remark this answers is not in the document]
VijayanB: As discussed will have this method since implementation is not used by client, only interface is being used

Vikasht34 (approved):
Thanks for addressing comments. Looks good to me.

jmazanec15 (approved):

junqiu-lei (approved):

PR #2378, opened 2025-01-09
Added more detailed error messages for KNN model training

### LINKED ISSUE #2268: Improve error messaging/validation for faiss training
## Description
One consistent feedback we get around PQ and IVF is that there is limited visibility into the failure cases. Part of this is because the errors are thrown on the faiss side and we dont return stack traces in Rest response.

This makes it difficult to use PQ and IVF.

To make it better, we should explicitly detect and message the following cases:
[ ] For PQ, explicitly check in OpenSearch an invalid configuration where m does not divide dimension
[ ] For PQ/IVF, check the number of training points matches the minimum clustering criteria defined in faiss
[ ] If there is not enough memory, explicitly say that there is not enough memory.

Checking these 3 configurations will cover 90% of the training failures that occur. With the changes made in disk-based, it should be relatively easy to do this.

## Related Issues
#1332 
#593

anntians:
### Description
Previously, a consistent feedback we get around PQ and IVF is that there is limited visibility into the failure cases. Part of this is because the errors are thrown on the Faiss side and we don't return stack traces in Rest response. So, this makes it difficult to use PQ and IVF. Thus, this PR provides improved error messages by adding explicit checks for the most common errors:

[ ] For PQ, explicitly check in OpenSearch an invalid configuration where m does not divide dimension
[ ] For PQ/IVF, check the number of training points matches the minimum clustering criteria defined in faiss
[ ] If there is not enough memory, explicitly say that there is not enough memory.

Adding these 3 checks will cover 90% of the training failures that occur. 

### Related Issues
Resolves #2268 

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/engine/AbstractKNNMethod.java
+++ b/src/main/java/org/opensearch/knn/index/engine/AbstractKNNMethod.java
@@ -22,6 +22,12 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
+import java.util.function.Function;
+
+import static org.opensearch.knn.common.KNNConstants.ENCODER_PARAMETER_PQ_M;
+import static org.opensearch.knn.common.KNNConstants.METHOD_ENCODER_PARAMETER;
+import static org.opensearch.knn.common.KNNConstants.METHOD_PARAMETER_NLIST;
+import static org.opensearch.knn.common.KNNConstants.ENCODER_PARAMETER_PQ_CODE_SIZE;
 
 /**
  * Abstract class for KNN methods. This class provides the common functionality for all KNN methods.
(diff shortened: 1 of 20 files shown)


src/main/java/org/opensearch/knn/plugin/transport/TrainingModelRequest.java
```diff
@@ -283,6 +285,15 @@ public ActionRequestValidationException validate() {
             exception.addValidationError("Description exceeds limit of " + KNNConstants.MAX_MODEL_DESCRIPTION_LENGTH + " characters");
         }
 
+        // Check if ENCODER_PARAMETER_PQ_M is divisible by vector dimension
+        if (knnMethodContext.getMethodComponentContext().getParameters().containsKey(ENCODER_PARAMETER_PQ_M)
+            && knnMethodConfigContext.getDimension() % (Integer) knnMethodContext.getMethodComponentContext()
+                .getParameters()
+                .get(ENCODER_PARAMETER_PQ_M) != 0) {
```
martin-gaievski, line 292: I'm not sure if that parameter is always present or not, but if it's optional then this line can generate the runtime exception in case parameter is not present

anntians: I believe Java has short-circuit evaluation, so if containsKey(ENCODER_PARAMETER_PQ_M) returns false then the second expression will not be evaluated. So a runtime exception shouldn't be thrown.

src/main/java/org/opensearch/knn/index/engine/AbstractKNNMethod.java
```diff
@@ -108,6 +114,55 @@ protected PerDimensionProcessor doGetPerDimensionProcessor(
         return PerDimensionProcessor.NOOP_PROCESSOR;
     }
 
+    protected Function<TrainingConfigValidationInput, TrainingConfigValidationOutput> doGetTrainingConfigValidationSetup() {
```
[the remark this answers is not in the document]
anntians: Got it, will merge this PR and create a separate PR for followup. Thanks

navneet1v (approved):

PR #2380, opened 2025-01-10
[Backport 2.x] Fixed a bug to prevent updating index.knn setting after index creation

### LINKED ISSUE #2334: [BUG]Block updation of knn.enabled flag.

anntians:
### Description
This change prevents customers to update index.knn setting after creating an index. Previously, there was a loophole to migrate a live k-nn index from hot to UW by closing the index, setting index.knn to false, opening the index, and migrating to UW. However, migrating k-nn indices to UW nodes is not supported. Thus, this PR updates index.knn to be FINAL and immutable after index creation.

### Related Issues
Resolves #2334 

### Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 * Fixing the bug when a segment has no vector field present for disk based vector search (#2282)[https://github.com/opensearch-project/k-NN/pull/2282]
 * Fix for NPE while merging segments after all the vector fields docs are deleted (#2365)[https://github.com/opensearch-project/k-NN/pull/2365]
 * Allow validation for non knn index only after 2.17.0 (#2315)[https://github.com/opensearch-project/k-NN/pull/2315]
+* Fixing the bug to prevent updating the index.knn setting after index creation(#2348)[https://github.com/opensearch-project/k-NN/pull/2348]
 * Release query vector memory after execution (#2346)[https://github.com/opensearch-project/k-NN/pull/2346]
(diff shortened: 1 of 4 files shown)


kotwanikunal:
@anntians Can you please fix spotless?

```
You can use '--warning-mode all' to show the individual deprecation warnings and determine if they come from your own scripts or plugins.
Execution failed for task ':qa:restart-upgrade:spotlessJavaCheck'.
> The following files had format violations:
      src\test\java\org\opensearch\knn\bwc\IndexingIT.java
          @@ -6,7 +6,6 @@
           package�org.opensearch.knn.bwc;
           
           import�org.junit.Assert;
          -import�org.opensearch.client.Response;
           import�org.opensearch.client.ResponseException;
           import�org.opensearch.common.settings.Settings;
           import�org.opensearch.common.xcontent.XContentFactory;

kotwanikunal (approved):

jmazanec15 (approved):

navneet1v (approved):

PR #2403, opened 2025-01-17
Explain API for Exact/ANN/Radial/Disk based KNN search on Faiss

### LINKED ISSUE #875: [BUG] Explain API not compatible with k-NN queries
**What is the bug?**
The [Explain](https://opensearch.org/docs/2.6/api-reference/explain/) API is not compatible with k-NN queries and assumes they have a score of 1. Well, I don't think there is a useful explanation to be given for k-NN queries, but when you have multiple k-NN queries or k-NN queries combined with text match queries, knowing the score of each k-NN part individually can be helpful.

Just make a k-NN query and set the `explain` flag to `true`.

Just have the score of the k-NN part correctly as the `value` field in the explanation.

image

neetikasinghal:
### Description
Add support for explain for Exact/ANN/Radial/Disk/Filtering k-nn search. Score calculation explanation is currently added only for ANN search.
Proposal for explain is given here: https://github.com/opensearch-project/k-NN/issues/875#issuecomment-2611349466

### Related Issues
Resolves #875

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/SpaceType.java
+++ b/src/main/java/org/opensearch/knn/index/SpaceType.java
@@ -40,7 +40,7 @@ public enum SpaceType {
             throw new IllegalStateException("Unsupported method");
         }
     },
-    L2("l2") {
+    L2("l2", SpaceType.GENERIC_SCORE_TRANSLATION) {
         @Override
         public float scoreTranslation(float rawScore) {
             return 1 / (1 + rawScore);
@@ -59,7 +59,7 @@ public enum SpaceType {
             return 1 / score - 1;
         }
     },
-    COSINESIMIL("cosinesimil") {
+    COSINESIMIL("cosinesimil", "`Math.max((2.0F - rawScore) / 2.0F, 0.0F)`") {
         /**
          * Cosine similarity has range of [-1, 1] where -1 represents vectors are at diametrically opposite, and 1 is where
          * they are identical in direction and perfectly similar. In Lucene, scores have to be in the range of [0, Float.MAX_VALUE].
@@ -100,13 +100,13 @@ public enum SpaceType {
             }
         }
     },
-    L1("l1") {
+    L1("l1", SpaceType.GENERIC_SCORE_TRANSLATION) {
         @Override
         public float scoreTranslation(float rawScore) {
             return 1 / (1 + rawScore);
         }
     },
-    LINF("linf") {
+    LINF("linf", SpaceType.GENERIC_SCORE_TRANSLATION) {
         @Override
         public float scoreTranslation(float rawScore) {
             return 1 / (1 + rawScore);
@@ -130,11 +130,16 @@ public enum SpaceType {
         }
 
         @Override
+        public String explainScoreTranslation(float rawScore) {
+            return rawScore >= 0 ? GENERIC_SCORE_TRANSLATION : "`-rawScore + 1`";
+        }
+
+        @Override
         public KNNVectorSimilarityFunction getKnnVectorSimilarityFunction() {
             return KNNVectorSimilarityFunction.MAXIMUM_INNER_PRODUCT;
         }
     },
-    HAMMING("hamming") {
+    HAMMING("hamming", SpaceType.GENERIC_SCORE_TRANSLATION) {
         @Override
         public float scoreTranslation(float rawScore) {
             return 1 / (1 + rawScore);
@@ -169,13 +174,28 @@ public enum SpaceType {
         .collect(Collectors.toList())
         .toArray(new String[0]);
 
+    private static final String GENERIC_SCORE_TRANSLATION = "`1 / (1 + rawScore)`";
     private final String value;
+    private final String explanationFormula;
 
     SpaceType(String value) {
         this.value = value;
+        this.explanationFormula = null;
+    }
+
+    SpaceType(String value, String explanationFormula) {
+        this.value = value;
+        this.explanationFormula = explanationFormula;
     }
 
     public abstract float scoreTranslation(float rawScore);
+
+    public String explainScoreTranslation(float rawScore) {
+        if (explanationFormula != null) {
+            return explanationFormula;
+        }
+        throw new UnsupportedOperationException("explainScoreTranslation is not defined for this space type.");
+    }
 
     /**
      * Get KNNVectorSimilarityFunction that maps to this SpaceType
--- a/src/main/java/org/opensearch/knn/index/query/KNNWeight.java
+++ b/src/main/java/org/opensearch/knn/index/query/KNNWeight.java
@@ -38,6 +38,7 @@ import org.opensearch.knn.index.memory.NativeMemoryLoadStrategy;
 import org.opensearch.knn.index.engine.KNNEngine;
 import org.opensearch.knn.index.quantizationservice.QuantizationService;
 import org.opensearch.knn.index.query.ExactSearcher.ExactSearcherContext.ExactSearcherContextBuilder;
+import org.opensearch.knn.index.query.explain.KnnExplanation;
 import org.opensearch.knn.indices.ModelDao;
 import org.opensearch.knn.indices.ModelMetadata;
 import org.opensearch.knn.indices.ModelUtil;
@@ -77,6 +78,7 @@ public class KNNWeight extends Weight {
 
     private static ExactSearcher DEFAULT_EXACT_SEARCHER;
     private final QuantizationService quantizationService;
+    private final KnnExplanation knnExplanation;
 
     public KNNWeight(KNNQuery query, float boost) {
         super(query);
@@ -86,6 +88,7 @@ public class KNNWeight extends Weight {
         this.filterWeight = null;
         this.exactSearcher = DEFAULT_EXACT_SEARCHER;
         this.quantizationService = QuantizationService.getInstance();
+        this.knnExplanation = new KnnExplanation();
     }
 
     public KNNWeight(KNNQuery query, float boost, Weight filterWeight) {
@@ -96,6 +99,7 @@ public class KNNWeight extends Weight {
         this.filterWeight = filterWeight;
         this.exactSearcher = DEFAULT_EXACT_SEARCHER;
         this.quantizationService = QuantizationService.getInstance();
+        this.knnExplanation = new KnnExplanation();
     }
 
     public static void initialize(ModelDao modelDao) {
@@ -108,9 +112,148 @@ public class KNNWeight extends Weight {
         KNNWeight.DEFAULT_EXACT_SEARCHER = exactSearcher;
     }
 
+    @VisibleForTesting
+    KnnExplanation getKnnExplanation() {
+        return knnExplanation;
+    }
+
     @Override
+    // This method is called in case of Radial-Search
     public Explanation explain(LeafReaderContext context, int doc) {
-        return Explanation.match(1.0f, "No Explanation");
+        return explain(context, doc, 0);
+    }
+
+    // This method is called for ANN/Exact/Disk-based/Efficient-filtering search
+    public Explanation explain(LeafReaderContext context, int doc, float score) {
+        knnQuery.setExplain(true);
+        try {
+            final KNNScorer knnScorer = getOrCreateKnnScorer(context);
+            if (score == 0) {
+                score = getKnnScore(knnScorer, doc);
+            }
+        } catch (IOException e) {
+            throw new RuntimeException(String.format("Error while explaining KNN score for doc [%d], score [%f]", doc, score), e);
+        }
+        final String highLevelExplanation = getHighLevelExplanation();
+        final StringBuilder leafLevelExplanation = getLeafLevelExplanation(context);
+
+        final SegmentReader reader = Lucene.segmentReader(context.reader());
+        final FieldInfo fieldInfo = FieldInfoExtractor.getFieldInfo(reader, knnQuery.getField());
+        if (fieldInfo == null) {
+            return Explanation.match(score, highLevelExplanation, Explanation.match(score, leafLevelExplanation.toString()));
+        }
+        final SpaceType spaceType = FieldInfoExtractor.getSpaceType(modelDao, fieldInfo);
+        leafLevelExplanation.append(", spaceType = ").append(spaceType.getValue());
+
+        final Float rawScore = knnExplanation.getRawScore(doc);
+        Explanation rawScoreDetail = null;
+        if (rawScore != null && knnQuery.getRescoreContext() == null) {
+            leafLevelExplanation.append(" where score is computed as ")
+                .append(spaceType.explainScoreTranslation(rawScore))
+                .append(" from:");
+            rawScoreDetail = Explanation.match(
+                rawScore,
+                "rawScore, returned from " + FieldInfoExtractor.extractKNNEngine(fieldInfo) + " library"
+            );
+        }
+
+        return rawScoreDetail != null
+            ? Explanation.match(score, highLevelExplanation, Explanation.match(score, leafLevelExplanation.toString(), rawScoreDetail))
+            : Explanation.match(score, highLevelExplanation, Explanation.match(score, leafLevelExplanation.toString()));
+    }
+
+    private StringBuilder getLeafLevelExplanation(LeafReaderContext context) {
+        int filterThresholdValue = KNNSettings.getFilteredExactSearchThreshold(knnQuery.getIndexName());
+        int cardinality = knnExplanation.getCardinality();
+        final StringBuilder sb = new StringBuilder("the type of knn search executed at leaf was ");
+        if (filterWeight != null) {
+            if (isFilterIdCountLessThanK(cardinality)) {
+                sb.append(KNNConstants.EXACT_SEARCH)
+                    .append(" since filteredIds = ")
+                    .append(cardinality)
+                    .append(" is less than or equal to K = ")
+                    .append(knnQuery.getK());
+            } else if (isExactSearchThresholdSettingSet(filterThresholdValue) && (filterThresholdValue >= cardinality)) {
+                sb.append(KNNConstants.EXACT_SEARCH)
+                    .append(" since filtered threshold value = ")
+                    .append(filterThresholdValue)
+                    .append(" is greater than or equal to cardinality = ")
+                    .append(cardinality);
+            } else if (!isExactSearchThresholdSettingSet(filterThresholdValue) && isMDCGreaterThanFilterIdCnt(cardinality)) {
+                sb.append(KNNConstants.EXACT_SEARCH)
+                    .append(" since max distance computation = ")
+                    .append(KNNConstants.MAX_DISTANCE_COMPUTATIONS)
+                    .append(" is greater than or equal to cardinality = ")
+                    .append(cardinality);
+            }
+        }
+        final Integer annResult = knnExplanation.getAnnResult(context.id());
+        if (annResult != null && annResult == 0 && isMissingNativeEngineFiles(context)) {
+            sb.append(KNNConstants.EXACT_SEARCH).append(" since no native engine files are available");
+        }
+        if (annResult != null && isFilteredExactSearchRequireAfterANNSearch(cardinality, annResult)) {
+            sb.append(KNNConstants.EXACT_SEARCH)
+                .append(" since the number of documents returned are less than K = ")
+                .append(knnQuery.getK())
+                .append(" and there are more than K filtered Ids = ")
+                .append(cardinality);
+        }
+        if (annResult != null && annResult > 0 && !isFilteredExactSearchRequireAfterANNSearch(cardinality, annResult)) {
+            sb.append(KNNConstants.ANN_SEARCH);
+        }
+        sb.append(" with vectorDataType = ").append(knnQuery.getVectorDataType());
+        return sb;
+    }
+
+    private String getHighLevelExplanation() {
+        final StringBuilder sb = new StringBuilder("the type of knn search executed was ");
+        if (knnQuery.getRescoreContext() != null) {
+            sb.append(buildDiskBasedSearchExplanation());
+        } else if (knnQuery.getRadius() != null) {
+            sb.append(KNNConstants.RADIAL_SEARCH).append(" with the radius of ").append(knnQuery.getRadius());
+        } else {
+            sb.append(KNNConstants.ANN_SEARCH);
+        }
+        return sb.toString();
+    }
+
+    private String buildDiskBasedSearchExplanation() {
+        StringBuilder sb = new StringBuilder(KNNConstants.DISK_BASED_SEARCH);
+        boolean isShardLevelRescoringDisabled = KNNSettings.isShardLevelRescoringDisabledForDiskBasedVector(knnQuery.getIndexName());
+        if (!knnQuery.getRescoreContext().isRescoreEnabled()) {
+            isShardLevelRescoringDisabled = true;
+        }
+        int dimension = knnQuery.getQueryVector().length;
+        int firstPassK = knnQuery.getRescoreContext().getFirstPassK(knnQuery.getK(), isShardLevelRescoringDisabled, dimension);
+        sb.append(" and the first pass k was ")
+            .append(firstPassK)
+            .append(" with vector dimension of ")
+            .append(dimension)
+            .append(", over sampling factor of ")
+            .append(knnQuery.getRescoreContext().getOversampleFactor());
+        if (isShardLevelRescoringDisabled) {
+            sb.append(", shard level rescoring disabled");
+        } else {
+            sb.append(", shard level rescoring enabled");
+        }
+        return sb.toString();
+    }
+
+    private KNNScorer getOrCreateKnnScorer(LeafReaderContext context) throws IOException {
+        // First try to get the cached scorer
+        KNNScorer scorer = knnExplanation.getKnnScorer(context);
+
+        // If no cached scorer exists, create and cache a new one
+        if (scorer == null) {
+            scorer = (KNNScorer) scorer(context);
+            knnExplanation.addKnnScorer(context, scorer);
+        }
+
+        return scorer;
+    }
+
+    private float getKnnScore(KNNScorer knnScorer, int doc) throws IOException {
+        return (knnScorer.iterator().advance(doc) == doc) ? knnScorer.score() : 0;
     }
 
     @Override
@@ -161,6 +304,9 @@ public class KNNWeight extends Weight {
         if (filterWeight != null && cardinality == 0) {
             return PerLeafResult.EMPTY_RESULT;
         }
+        if (knnQuery.isExplain()) {
+            knnExplanation.setCardinality(cardinality);
+        }
         /*
          * The idea for this optimization is to get K results, we need to at least look at K vectors in the HNSW graph
          * . Hence, if filtered results are less than K and filter query is present we should shift to exact search.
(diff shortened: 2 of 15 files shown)


src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -106,7 +110,100 @@ static void initialize(ModelDao modelDao, ExactSearcher exactSearcher) {
 
     @Override
     public Explanation explain(LeafReaderContext context, int doc) {
```
Vikasht34, line 112: Could we improve readability of this function , I see this function is doing 

1. Building Score 
2. Build the main explanation
3. Build the detailed explanation for the leaf node.

Would we start with something simple 

```
public Explanation explain(LeafReaderContext context, int doc) {
   //Function to Calculate Score
    // Build the main explanation

      // Build the detailed explanation for the leaf node
    return Explanation.match(score, mainExplanation, Explanation.match(score, leafExplanation));
}
```
And Please break down other functions as well when needed.

Vikasht34 (commented):
We need to have Unit test for each case 
1. Disk-based search with valid rescore context.
2. Radial search.
3. ANN search (default case).
4. Shard-level rescoring enabled.
5. Shard-level rescoring disabled.
6.  Filter weight case where filtered IDs are less than k.
7. Filter threshold value greater than cardinality.
8. Missing native engine files.
9. Valid context with matching document and disk-based search.

And Please validate with Explaination Object.

neetikasinghal:
> We need to have Unit test for each case
> 
> 1. Disk-based search with valid rescore context.
> 2. Radial search.
> 3. ANN search (default case).
> 4. Shard-level rescoring enabled.
> 5. Shard-level rescoring disabled.
> 6. Filter weight case where filtered IDs are less than k.
> 7. Filter threshold value greater than cardinality.
> 8. Missing native engine files.
> 9. Valid context with matching document and disk-based search.
> 
> And Please validate with Explaination Object.

@Vikasht34 yes the tests are not yet added in here, hence its in a draft status. I will add the coverage with all the possible cases.

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
neetikasinghal: sure i will refactor this.

neetikasinghal:
@navneet1v / @Vikasht34 would you please review the changes?

navneet1v:
@neetikasinghal can you please add an entry in the change log

neetikasinghal:
> @neetikasinghal can you please add an entry in the change log

yup i generally add it towards the end of the review so that its easier to rebase with the latest changes.

src/main/java/org/opensearch/knn/index/SpaceType.java
```diff
@@ -46,6 +51,11 @@ public float scoreTranslation(float rawScore) {
             return 1 / (1 + rawScore);
         }
 
+        @Override
+        public String explainScoreTranslation(float rawScore) {
+            return "`1 / (1 + rawScore)`";
```
navneet1v, line 56: can we make this as private static final string.

src/main/java/org/opensearch/knn/index/SpaceType.java
```diff
@@ -77,6 +87,11 @@ public float scoreTranslation(float rawScore) {
             return Math.max((2.0F - rawScore) / 2.0F, 0.0F);
         }
 
+        @Override
+        public String explainScoreTranslation(float rawScore) {
+            return "`Math.max((2.0F - rawScore) / 2.0F, 0.0F)`";
```
navneet1v, line 92: same as above

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -105,8 +109,147 @@ static void initialize(ModelDao modelDao, ExactSearcher exactSearcher) {
...
+        if (knnScorer == null) {
+            float knnScore;
+            try {
+                knnScore = getKnnScore(context, doc);
+            } catch (IOException e) {
+                throw new RuntimeException("Error while getting KNN score during explanation", e);
+            }
+            if (score == 0) {
+                score = knnScore;
+            }
```
navneet1v, line 129: shouldn't this be the first check we do when we enter the if condition.

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -105,8 +109,147 @@ static void initialize(ModelDao modelDao, ExactSearcher exactSearcher) {
...
+        if (isShardLevelRescoringDisabled) {
+            sb.append(", shard level rescoring disabled");
+        } else {
+            sb.append(", shard level rescoring enabled");
+        }
+        return sb.toString();
+    }
+
+    private float getKnnScore(LeafReaderContext context, int doc) throws IOException {
+        KNNScorer knnScorer = (KNNScorer) scorer(context);
```
navneet1v, line 247: why we are computing the scorer for every doc again and again? In case of explain should we just cache scorer in the KNNExplainObject once the first explain call happens.

navneet1v (commented):
Looked at high level in the code, need to go through more in terms of explanation. But one thing I want to add is I am not seeing any ITs related to explain api. Can we please add them too

neetikasinghal:
> Looked at high level in the code, need to go through more in terms of explanation. But one thing I want to add is I am not seeing any ITs related to explain api. Can we please add them too

yup, that WIP as my setup for ITs was broken. I am able to fix that now, however the PR has the coverage for all the UTs.

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
neetikasinghal: thats a good point, will make changes

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
neetikasinghal: not really, knnScore should be computed for it to be assigned to the score.

Vikasht34 (commented):
Will continue on other files.

src/main/java/org/opensearch/knn/index/query/explain/KnnExplanation.java
```diff
@@ -0,0 +1,40 @@
...
+import org.opensearch.knn.index.query.KNNScorer;
+
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
+ * This class captures details around knn explain queries that is used
+ * by explain API to generate explanation for knn queries
+ */
+public class KnnExplanation {
```
Vikasht34, line 19: There are couple of things that we can improve on this class, 

1.  As of Now every client of this class has to get Maps and then put values so Maps are final but Modificable outside of this class. Ideally as  client,  I should not be knowing internal data structure of KnnExplanation Class ,  Client must to exposed to api like addAnnResult, AddRawScores , addKnnScores that's all , how we are adding these and which data structure to use , it's should be upto KNNExplanation class. 
2. Please make all maps Unmodifiable outside of this class. No Client should be able to Modify these maps except KNNExplanation.

src/main/java/org/opensearch/knn/index/query/common/DocAndScoreQuery.java
```diff
@@ -31,13 +32,15 @@ final class DocAndScoreQuery extends Query {
     private final float[] scores;
     private final int[] segmentStarts;
     private final Object contextIdentity;
+    private final KNNWeight knnWeight;
 
-    public DocAndScoreQuery(int k, int[] docs, float[] scores, int[] segmentStarts, Object contextIdentity) {
+    public DocAndScoreQuery(int k, int[] docs, float[] scores, int[] segmentStarts, Object contextIdentity, KNNWeight knnWeight) {
```
Vikasht34, line 37: Now whole explain method is there in KNNWeight class , DocAndScoreQuery has to take reference of this class , just to call , explain method of this class , Please abstract explain method from KnnWeight class so that in future if we have another type of query , we don't end up passing KnnWeight class everywhere.

src/main/java/org/opensearch/knn/index/query/explain/KnnExplanation.java
neetikasinghal: makes sense, will update it.

src/main/java/org/opensearch/knn/index/query/common/DocAndScoreQuery.java
neetikasinghal: i understand the caveat here of passing the KNNWeight class, but didnt really get your suggestion of abstracting out the explain method out of the KNNWeight class.

src/main/java/org/opensearch/knn/index/SpaceType.java
```diff
@@ -177,6 +215,8 @@ public KNNVectorSimilarityFunction getKnnVectorSimilarityFunction() {
 
     public abstract float scoreTranslation(float rawScore);
 
+    public abstract String explainScoreTranslation(float rawScore);
```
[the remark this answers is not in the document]
neetikasinghal: i can introduce introduce the static final variables to reduce code duplication, however the logic highlighted above of constructors wont work in case of inner product where the explain formula differs:
```
@Override
        public String explainScoreTranslation(float rawScore) {
            if (rawScore >= 0) {
                return "`1 / (1 + rawScore)`";
            }
            return "`-rawScore + 1`";
        }
```

Its dependent on raw score. Also, there could be future space type whose formula depends on the raw score, hence i thought of introducing this function instead. let me know your thoughts here.

src/main/java/org/opensearch/knn/index/query/common/DocAndScoreQuery.java
Vikasht34: Please move explain method to utlity class , which takes context , k and scorere and call that utility method from both KnnWieght and DocsAndScoREqUERY

src/main/java/org/opensearch/knn/index/SpaceType.java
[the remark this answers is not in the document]
Vikasht34: This is how Ineer product will work, if explainScoreTranslation is differnt then it can be overridden.

```
INNER_PRODUCT("innerproduct") {
        @Override
            return rawScore >= 0 ? (1 / (1 + rawScore)) : (-rawScore + 1);
        }

        @Override
        public String explainScoreTranslation(float rawScore) {
            return rawScore >= 0 ? "`1 / (1 + rawScore)`" : "`-rawScore + 1`";
        
```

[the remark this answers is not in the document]
Vikasht34:
```
public enum SpaceType {
    UNDEFINED("undefined") {
        @Override
        public float scoreTranslation(final float rawScore) {
            throw new IllegalStateException("Unsupported method");
        }

        @Override
        public String explainScoreTranslation(float rawScore) {
            throw new IllegalStateException("Unsupported method");
        }
    },

    L2("l2", "`1 / (1 + rawScore)`") {
        @Override
            return 1 / (1 + rawScore);
        }
    },

    COSINE("cosine", "`Math.max((2.0F - rawScore) / 2.0F, 0.0F)`") {
        @Override
            return Math.max((2.0F - rawScore) / 2.0F, 0.0F);
        }
    },

    INNER_PRODUCT("innerproduct") {
        @Override
            return rawScore >= 0 ? (1 / (1 + rawScore)) : (-rawScore + 1);
        }

        @Override
        public String explainScoreTranslation(float rawScore) {
            return rawScore >= 0 ? "`1 / (1 + rawScore)`" : "`-rawScore + 1`";
        }
    };

    private final String name;
    private final String explanationFormula;

    SpaceType(String name) {
        this.name = name;
        this.explanationFormula = null;
    }

    SpaceType(String name, String explanationFormula) {
        this.name = name;
        this.explanationFormula = explanationFormula;
    }

    public abstract float scoreTranslation(float rawScore);

    public String explainScoreTranslation(float rawScore) {
        if (explanationFormula != null) {
            return explanationFormula;
        }
        throw new UnsupportedOperationException("explainScoreTranslation not defined for this space type.");
    }
}

```

Let me know if it makes sense.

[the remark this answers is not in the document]
neetikasinghal: yes this makes sense, will update it.

src/main/java/org/opensearch/knn/index/query/common/DocAndScoreQuery.java
neetikasinghal: so KnnWeight has access to KnnQuery which is needed as part of explain function determining the explanation, knnQuery is not present in DocAndScoreQuery. Also, we are setting the explain flag as true as part of knnQuery to set some maps in the searchLeaf path which would not be set if the explain is set to false.

@Vikasht34 let me know if this makes sense or if u have a better way of doing it.

neetikasinghal:
pending items -
- rebase to the latest changes
- Add ITs

src/main/java/org/opensearch/knn/index/query/common/DocAndScoreQuery.java
Vikasht34: I think that's fine .

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -105,8 +109,153 @@ static void initialize(ModelDao modelDao, ExactSearcher exactSearcher) {
...
     public Explanation explain(LeafReaderContext context, int doc) {
-        return Explanation.match(1.0f, "No Explanation");
+        return explain(context, doc, 0, null);
+    }
+
+    // This method is called for ANN/Exact/Disk-based/Efficient-filtering search
+    public Explanation explain(LeafReaderContext context, int doc, float score, KNNScorer knnScorer) {
+        knnQuery.setExplain(true);
+        try {
+            knnScorer = getOrCreateKnnScorer(context, knnScorer);
```
martin-gaievski, line 121: is there a scenario when knnScorer is not null? If no, then you should not pass it to this method and always create new instance inside the method.

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -105,8 +109,153 @@ static void initialize(ModelDao modelDao, ExactSearcher exactSearcher) {
...
+        }
+        sb.append(" with vectorDataType = ").append(knnQuery.getVectorDataType());
+        return sb;
+    }
+
+    private SpaceType getSpaceType(FieldInfo fieldInfo) {
+        try {
+            return FieldInfoExtractor.getSpaceType(modelDao, fieldInfo);
+        } catch (IllegalArgumentException e) {
+            return knnQuery.getVectorDataType() == VectorDataType.BINARY ? SpaceType.DEFAULT_BINARY : SpaceType.DEFAULT;
```
martin-gaievski, line 207: why we are not signaling to user that system is in bad state, wouldn't this be misleading, I mean giving false data to the user?

src/main/java/org/opensearch/knn/index/SpaceType.java
```diff
@@ -169,14 +174,29 @@ public KNNVectorSimilarityFunction getKnnVectorSimilarityFunction() {
...
+        this.explanationFormula = explanationFormula;
     }
 
     public abstract float scoreTranslation(float rawScore);
 
+    public String explainScoreTranslation(float rawScore) {
+        if (explanationFormula != null) {
+            return explanationFormula;
+        }
+        throw new UnsupportedOperationException("explainScoreTranslation is not defined for this space type.");
```
martin-gaievski, line 197: this effectively means if object created with old constructor (formula is null) then this method will throw exception, is this desired behavior? Can we have some constant like "undefined" or similar, or if formula isn't defined the rets of explanation detail don't make sense as well?

src/main/java/org/opensearch/knn/index/query/explain/KnnExplanation.java
```diff
@@ -0,0 +1,56 @@
...
+
+    @Getter
+    private final Map<Object, Integer> annResultPerLeaf;
```
martin-gaievski, line 22: are you sure you want to expose raw map? clients will be able to modify them. You may want to consider returning `Collections.unmodifiableMap(your_map)` in getter method

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
neetikasinghal: makes sense, i think we dont need to handle this exception here, will remove this. It should be taken care of during index creation itself.

src/main/java/org/opensearch/knn/index/query/explain/KnnExplanation.java
neetikasinghal: good catch! removed the getter for this map as well as other maps in the class too.

src/main/java/org/opensearch/knn/index/SpaceType.java
neetikasinghal: @martin-gaievski The old constructor is called for two enums: `UNDEFINED` and `INNER_PRODUCT`. `INNER_PRODUCT` overrides the `explainScoreTranslation` method and for undefined its fine to throw an exception.
The callers of this function would know that the explainScoreTranslation is not defined for a specific space type. I think that should be fine, what do u think?

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
neetikasinghal: makes sense, removed the knnScorer being passed in the function.

neetikasinghal:
> pending items -
> 
> * rebase to the latest changes
> * Add ITs

completed the pending items and addressed all the comments.

@navneet1v @Vikasht34 @martin-gaievski can u please take another look at it?

Vikasht34 (approved):
LGTM , Thanks for addressing the suggestions !!

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased 2.x](https://github.com/opensearch-project/k-NN/compare/2.19...2.x)
 ### Features
+* [Explain API for Exact/ANN/Radial/Disk based KNN search on Faiss] Explain API for Exact/ANN/Radial/Disk based KNN search on Faiss [#2403] (https://github.com/opensearch-project/k-NN/pull/2403)
```
navneet1v, line 11:
```suggestion
* [Explain API Support] Added Explain API support for Exact/ANN/Radial/Disk based KNN search on Faiss Engine [#2403] (https://github.com/opensearch-project/k-NN/pull/2403)
```

src/main/java/org/opensearch/knn/index/SpaceType.java
```diff
@@ -169,14 +174,29 @@ public KNNVectorSimilarityFunction getKnnVectorSimilarityFunction() {
...
+        this.value = value;
+        this.explanationFormula = explanationFormula;
     }
 
     public abstract float scoreTranslation(float rawScore);
 
+    public String explainScoreTranslation(float rawScore) {
+        if (explanationFormula != null) {
+            return explanationFormula;
+        }
```
navneet1v, line 196: what is the use of rawScore parameter here when that parameter is not getting used at all?

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -108,9 +112,148 @@ static void initialize(ModelDao modelDao, ExactSearcher exactSearcher) {
...
+    // This method is called for ANN/Exact/Disk-based/Efficient-filtering search
+    public Explanation explain(LeafReaderContext context, int doc, float score) {
+        knnQuery.setExplain(true);
+        try {
+            final KNNScorer knnScorer = getOrCreateKnnScorer(context);
+            if (score == 0) {
+                score = getKnnScore(knnScorer, doc);
+            }
+        } catch (IOException e) {
+            throw new RuntimeException("Error while explaining KNN score", e);
```
navneet1v, line 135:
```suggestion
            throw new RuntimeException("Error while explaining KNN score for doc: " + doc + " score: " + score, e);
```

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -108,9 +112,148 @@ static void initialize(ModelDao modelDao, ExactSearcher exactSearcher) {
...
+    }
+
+    // This method is called for ANN/Exact/Disk-based/Efficient-filtering search
+    public Explanation explain(LeafReaderContext context, int doc, float score) {
+        knnQuery.setExplain(true);
+        try {
+            final KNNScorer knnScorer = getOrCreateKnnScorer(context);
+            if (score == 0) {
+                score = getKnnScore(knnScorer, doc);
+            }
```
navneet1v, line 133: why this condition and what will happen if score is not zero?

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -108,9 +112,148 @@ static void initialize(ModelDao modelDao, ExactSearcher exactSearcher) {
...
+        }
+        return sb.toString();
+    }
+
+    private String buildDiskBasedSearchExplanation() {
+        StringBuilder sb = new StringBuilder(KNNConstants.DISK_BASED_SEARCH);
+        boolean isShardLevelRescoringDisabled = KNNSettings.isShardLevelRescoringDisabledForDiskBasedVector(knnQuery.getIndexName());
+        if (!knnQuery.getRescoreContext().isRescoreEnabled()) {
+            isShardLevelRescoringDisabled = true;
+        }
```
navneet1v, line 225: why the disk based rescore logic coming in this query? this logic in explain should be part of NativeEngineKnnVectorQuery

src/test/java/org/opensearch/knn/index/ExplainIT.java
```diff
@@ -0,0 +1,333 @@
...
+import static org.opensearch.knn.common.KNNConstants.METHOD_HNSW;
+import static org.opensearch.knn.common.KNNConstants.METHOD_PARAMETER_SPACE_TYPE;
+import static org.opensearch.knn.common.KNNConstants.MODE_PARAMETER;
+import static org.opensearch.knn.common.KNNConstants.NAME;
+import static org.opensearch.knn.common.KNNConstants.RADIAL_SEARCH;
+
+public class ExplainIT extends KNNRestTestCase {
+
+    @SneakyThrows
+    public void testAnnSearch() {
```
navneet1v, line 46: Please rename all test functions to `testABC_whenDEF_thenXYZ()` format to ensure better readability.

src/test/java/org/opensearch/knn/index/ExplainIT.java
```diff
@@ -0,0 +1,333 @@
...
+            Map<String, Object> hitMap = (Map<String, Object>) hit;
+            Double score = (Double) hitMap.get("_score");
+            String explanation = hitMap.get("_explanation").toString();
+            assertNotNull(explanation);
+            for (String description : descriptions) {
+                assertTrue(explanation.contains(description));
+            }
+            assertTrue(explanation.contains(String.valueOf(score)));
+        });
+    }
```
navneet1v, line 291: this is not a strong assertion as these assertions are just validating that explanation is present. Please validate the contents of the explain api response too to ensure that correct strings are returned.

src/main/java/org/opensearch/knn/index/SpaceType.java
neetikasinghal: its used for a space type like `INNER_PRODUCT,` check: https://github.com/opensearch-project/k-NN/pull/2403/files#diff-c481eefa59727056eb79ce9614cbcbb8b84358cf4c8b4bc51eb2d089b8a60a1bR132

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
neetikasinghal: if the score is 0, it will calculate the score, however if score is passed from the caller, like in case of disk based search, there is no need to re-compute the score and we can rely on the score passed from the caller.

src/test/java/org/opensearch/knn/index/ExplainIT.java
neetikasinghal: the individual tests are passing all the description types that it is expecting in the explain response, please refer to individual tests to check what all we are asserting for.

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
neetikasinghal: NativeEngineKnnVectorQuery is not aware of any explain related details, it flows down to KNNWeight.

src/test/java/org/opensearch/knn/index/ExplainIT.java
neetikasinghal: i think the tests are named readable even now, i dont see a need to rename the tests unless there is a strong opinion on it.

Vikasht34 (approved):

neetikasinghal:
thanks @Vikasht34 for the approval. need to merge changes by 9m to make it to the cutoff for 3.0.0-beta.

would appreciate if i can get a second approval on this.

naveentatikonda (approved):

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
navneet1v: It would be better to provide this information as a Java doc.

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
navneet1v: I understand the part that we use KNNWeight for both first pass and second pass in disk based vector search. But the decision on wether to use the shard level or segment level rescoring is not controlled by KNNWeight. Hence this is just extra information which kNNWeight has to bear now due to explain API, which in my true sense is not a good design.

src/test/java/org/opensearch/knn/index/ExplainIT.java
navneet1v: Got it, I think I saw what you have done. It is not validating the full string but parts of the string. `assertTrue(explanation.contains(description));`. This is still not 100% validation. We should have at-least some test which checks the full string rather than partial string.

navneet1v:
@neetikasinghal see if you want to followup with another PR with some of the suggestion being provided as the PR is already merged.

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
neetikasinghal: the problem is NativeEngineKnnVectorQuery doesnt handle all the search use-cases, for Radial search, the logic is handled separately which also looks to be redundant. If we can have all the logic in `NativeEngineKnnVectorQuery`, i think we can move the explanation part to the `NativeEngineKnnVectorQuery` as well. 
KNNWeight is not bearing this extra information, its just one check that it has to do in case of disk-based search.
I think we should refactor radial search also to have the same code flow as `NativeEngineKnnVectorQuery`.

Let me know your thoughts and if u have any other suggestions.

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
neetikasinghal: sure i can add it

src/test/java/org/opensearch/knn/index/ExplainIT.java
neetikasinghal: ok i can rename them.

src/test/java/org/opensearch/knn/index/ExplainIT.java
neetikasinghal: i dont agree here as the exact language can change in future and the tests would start to break which should not be the case. the current tests are validating the required description which should not change in the future.

neetikasinghal:
> @neetikasinghal see if you want to followup with another PR with some of the suggestion being provided as the PR is already merged.

@navneet1v had to merge this in before to make it to 3.0, created another pr for some minor comments: https://github.com/opensearch-project/k-NN/pull/2669

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
neetikasinghal: created issue for this: https://github.com/opensearch-project/k-NN/issues/2670

PR #2405, opened 2025-01-17
Partial loading implementation for FAISS HNSW

0ctopus13prime:
### Description

OpenSearch KNN plugin supports three engines: NMSLIB, FAISS, and Lucene.
The first two native engines, NMSLIB and FAISS, require all vector-related data structures (such as HNSW graphs) to be loaded into memory for search operation.
For large workloads, this memory cost can quickly become substantial if quantization techniques are not applied.
Therefore, 'Partial Loading' must be enabled as an option in native engines to control the available memory for KNN search. The objective of partial loading is twofold:

To allow users to control the maximum memory available for KNN searching.
To enable native engines to partially load only the necessary data within the constraint.
If we look closely a HNSW graph mainly consist of below things:

Full precision 32 bit vectors.
Graph representations.
Metadata like dimensions, number of vectors, space type, headers etc.
From the above items, main memory is used by these full precision vectors 4 bytes * the number of vectors * the number of dimension.
The way FAISS stores these vectors is in a Flat Index and during serialization and deserialization these vectors are written and read to/from the file and put in the main memory which increases the memory consumption.

### Related Issues
#2401 

### Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased 3.0](https://github.com/opensearch-project/k-NN/compare/2.x...HEAD)
 ### Features
+* Added a boolean flag that enable user to use Lucene searcher on FAISS index. [#2405](https://github.com/opensearch-project/k-NN/pull/2405)
 ### Enhancements
 ### Bug Fixes
 ### Infrastructure
--- a/src/main/java/org/opensearch/knn/common/KNNConstants.java
+++ b/src/main/java/org/opensearch/knn/common/KNNConstants.java
@@ -120,6 +120,8 @@ public class KNNConstants {
     public static final List<String> FAISS_SQ_ENCODER_TYPES = List.of(FAISS_SQ_ENCODER_FP16);
     public static final String FAISS_SIGNED_BYTE_SQ = "SQ8_direct_signed";
     public static final String FAISS_SQ_CLIP = "clip";
+    public static final String USE_LUCENE_HNSW_SEARCHER = "use_lucene_searcher";
+    public static final boolean DEFAULT_USE_LUCENE_HNSW_SEARCHER = false;
 
     // Parameter defaults/limits
     public static final Integer ENCODER_PARAMETER_PQ_CODE_COUNT_DEFAULT = 1;
--- a/src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
+++ b/src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
@@ -21,9 +21,21 @@ import org.apache.lucene.search.KnnCollector;
 import org.apache.lucene.search.ScoreDoc;
 import org.apache.lucene.search.TopDocs;
 import org.apache.lucene.search.TotalHits;
+import org.apache.lucene.store.IOContext;
+import org.apache.lucene.store.IndexInput;
+import org.apache.lucene.store.ReadAdvice;
 import org.apache.lucene.util.Bits;
 import org.apache.lucene.util.IOUtils;
 import org.opensearch.common.UUIDs;
+import org.opensearch.common.xcontent.XContentHelper;
+import org.opensearch.core.common.bytes.BytesArray;
+import org.opensearch.core.xcontent.DeprecationHandler;
+import org.opensearch.core.xcontent.MediaTypeRegistry;
+import org.opensearch.core.xcontent.NamedXContentRegistry;
+import org.opensearch.core.xcontent.XContentParser;
+import org.opensearch.knn.common.KNNConstants;
+import org.opensearch.knn.index.codec.luceneonfaiss.FaissHNSWVectorReader;
+import org.opensearch.knn.index.codec.luceneonfaiss.LuceneOnFaissUtils;
 import org.opensearch.knn.index.codec.util.KNNCodecUtil;
 import org.opensearch.knn.index.codec.util.NativeMemoryCacheKeyHelper;
 import org.opensearch.knn.index.memory.NativeMemoryCacheManager;
@@ -32,6 +44,7 @@ import org.opensearch.knn.quantization.models.quantizationState.QuantizationStat
 import org.opensearch.knn.quantization.models.quantizationState.QuantizationStateCacheManager;
 import org.opensearch.knn.quantization.models.quantizationState.QuantizationStateReadConfig;
 
+import java.io.Closeable;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.HashMap;
@@ -48,12 +61,71 @@ public class NativeEngines990KnnVectorsReader extends KnnVectorsReader {
     private Map<String, String> quantizationStateCacheKeyPerField;
     private SegmentReadState segmentReadState;
     private final List<String> cacheKeys;
+    private Map<String, FaissHNSWVectorReader> faissHNSWVectorReaderMap;
 
     public NativeEngines990KnnVectorsReader(final SegmentReadState state, final FlatVectorsReader flatVectorsReader) {
         this.flatVectorsReader = flatVectorsReader;
         this.segmentReadState = state;
         this.cacheKeys = getVectorCacheKeysFromSegmentReaderState(state);
+        this.faissHNSWVectorReaderMap = new HashMap<>(2 * state.fieldInfos.size(), 0.6f);
         loadCacheKeyMap();
+        loadFaissIndexForLuceneSearcher(segmentReadState);
+    }
+
+    private void loadFaissIndexForLuceneSearcher(SegmentReadState state) {
+        for (FieldInfo fieldInfo : state.fieldInfos) {
+            // Ex: {"index_description":"HNSW16,Flat","spaceType":"l2","name":"hnsw","data_type":"float",
+            // "parameters":{"use_lucene_searcher":true,"ef_search":100,"ef_construction":100,"encoder":{"name":"flat","parameters":{}}}}
+            final String parametersString = fieldInfo.getAttribute(KNNConstants.PARAMETERS);
+            if (parametersString != null) {
+                try {
+                    try (
+                        XContentParser parser = XContentHelper.createParser(NamedXContentRegistry.EMPTY,
+                                                                            DeprecationHandler.THROW_UNSUPPORTED_OPERATION,
+                                                                            new BytesArray(parametersString),
+                                                                            MediaTypeRegistry.getDefaultMediaType()
+                        )
+                    ) {
+                        // Extract boolean flag
+                        final Map<String, Object> parameters = parser.map();
+                        final Object innerParameters = parameters.get(KNNConstants.PARAMETERS);
+                        if (!LuceneOnFaissUtils.isUseLuceneOnFaiss(innerParameters)) {
+                            continue;
+                        }
+
+                        // Acquire index file name
+                        final String faissIndexFile = KNNCodecUtil.getNativeEngineFileFromFieldInfo(fieldInfo, state.segmentInfo);
+                        if (faissIndexFile == null) {
+                            continue;
+                        }
+
+                        // Load faiss index with IndexInput
+                        final IndexInput indexInput = state.directory.openInput(faissIndexFile,
+                                                                                new IOContext(IOContext.Context.DEFAULT,
+                                                                                              null,
+                                                                                              null,
+                                                                                              ReadAdvice.RANDOM
+                                                                                )
+                        );
+
+                        try {
+                            final FaissHNSWVectorReader vectorReader = new FaissHNSWVectorReader(indexInput);
+                            faissHNSWVectorReaderMap.put(fieldInfo.getName(), vectorReader);
+                        } catch (Exception e) {
+                            // If something went bad, we close the stream and rethrow
+                            try {
+                                indexInput.close();
+                            } catch (Exception ioException) {
+                                // Ignore
+                            }
+                            throw e;
+                        }
+                    }
+                } catch (IOException e) {
+                    throw new RuntimeException(e);
+                }
+            }
+        }
     }
 
     /**
@@ -124,18 +196,27 @@ public class NativeEngines990KnnVectorsReader extends KnnVectorsReader {
             String cacheKey = quantizationStateCacheKeyPerField.get(field);
             FieldInfo fieldInfo = segmentReadState.fieldInfos.fieldInfo(field);
             QuantizationState quantizationState = QuantizationStateCacheManager.getInstance()
-                .getQuantizationState(
-                    new QuantizationStateReadConfig(
-                        segmentReadState,
-                        QuantizationService.getInstance().getQuantizationParams(fieldInfo),
-                        field,
-                        cacheKey
-                    )
-                );
+                .getQuantizationState(new QuantizationStateReadConfig(segmentReadState,
+                                                                      QuantizationService.getInstance().getQuantizationParams(fieldInfo),
+                                                                      field,
+                                                                      cacheKey
+                ));
             ((QuantizationConfigKNNCollector) knnCollector).setQuantizationState(quantizationState);
             return;
         }
-        throw new UnsupportedOperationException("Search functionality using codec is not supported with Native Engine Reader");
+
+        // Try with Lucene searcher
+        final FaissHNSWVectorReader vectorReader = faissHNSWVectorReaderMap.get(field);
+        if (vectorReader != null) {
+            try {
+                vectorReader.search(target, knnCollector, acceptDocs);
+            } catch (Exception e) {
+                // KDY
+                e.printStackTrace();
+            }
+        } else {
+            throw new UnsupportedOperationException("Search functionality using codec is not supported with Native Engine Reader");
+        }
     }
 
     /**
@@ -187,8 +268,10 @@ public class NativeEngines990KnnVectorsReader extends KnnVectorsReader {
         final NativeMemoryCacheManager nativeMemoryCacheManager = NativeMemoryCacheManager.getInstance();
         cacheKeys.forEach(nativeMemoryCacheManager::invalidate);
 
-        // Close a reader.
-        IOUtils.close(flatVectorsReader);
+        // Close all reader.
+        List<Closeable> readers = new ArrayList<>(faissHNSWVectorReaderMap.values());
+        readers.add(flatVectorsReader);
+        IOUtils.close(readers);
 
         // Clean up quantized state cache.
         if (quantizationStateCacheKeyPerField != null) {
--- a/src/main/java/org/opensearch/knn/index/codec/luceneonfaiss/FaissHNSW.java
+++ b/src/main/java/org/opensearch/knn/index/codec/luceneonfaiss/FaissHNSW.java
@@ -1,0 +1,92 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.knn.index.codec.luceneonfaiss;
+
+import lombok.Getter;
+import org.apache.lucene.store.IndexInput;
+
+import java.io.IOException;
+
+/**
+ * Ported implementation of the FAISS HNSW graph search algorithm.
+ * While it follows the same steps as the original FAISS implementation, differences in how the JVM and C++ handle floating-point
+ * calculations can lead to slight variations in results. However, such cases are very rare, and in most instances, the results are
+ * identical to FAISS. Even when there are ranking differences, they do not impact the precision or recall of the search.
+ * For more details, refer to the [FAISS HNSW implementation](
+ * <a href="https://github.com/facebookresearch/faiss/blob/main/faiss/impl/HNSW.h">...</a>).
+ */
+@Getter
+public class FaissHNSW {
+    // Cumulative number of neighbors per each level.
+    private int[] cumNumberNeighborPerLevel;
+    // Offset to be added to cumNumberNeighborPerLevel[level] to get the actual start offset of neighbor list.
+    private long[] offsets = null;
+    // Neighbor list storage.
+    private final Storage neighbors = new Storage();
+    // Entry point in HNSW graph
+    private int entryPoint;
+    // Maximum level of HNSW graph
+    private int maxLevel = -1;
+    // Default efSearch parameter. This determines the navigation queue size.
+    // More value, algorithm will more navigate candidates.
+    private int efSearch = 16;
+    // Total number of vectors stored in graph.
+    private long totalNumberOfVectors;
+
+    /**
+     * Partially loads the FAISS HNSW graph from the provided index input stream.
+     * The graph is divided into multiple sections, and this method marks the starting offset of each section then skip to the next
+     * section instead of loading the entire graph into memory. During the search, bytes will be accessed via {@link IndexInput}.
+     *
+     * @param input An input stream for a FAISS HNSW graph file, allowing access to the neighbor list and vector locations.
+     * @param totalNumberOfVectors The total number of vectors stored in the graph.
+     * @return {@link FaissHNSW}, a graph search structure that represents the FAISS HNSW graph
+     * @throws IOException
+     */
+    public static FaissHNSW load(IndexInput input, long totalNumberOfVectors) throws IOException {
+        // Total number of vectors
+        FaissHNSW faissHNSW = new FaissHNSW();
+        faissHNSW.totalNumberOfVectors = totalNumberOfVectors;
+
+        // We don't use `double[] assignProbas` for search. It is for index construction.
+        long size = input.readLong();
+        input.skipBytes(Double.BYTES * size);
+
+        // Accumulate number of neighbor per each level.
+        size = input.readLong();
+        faissHNSW.cumNumberNeighborPerLevel = new int[(int) size];
+        if (size > 0) {
+            input.readInts(faissHNSW.cumNumberNeighborPerLevel, 0, (int) size);
+        }
+
+        // We don't use `level`.
+        final Storage levels = new Storage();
+        levels.markSection(input, Integer.BYTES);
+
+        // Load `offsets` into memory.
+        size = input.readLong();
+        faissHNSW.offsets = new long[(int) size];
+        input.readLongs(faissHNSW.offsets, 0, faissHNSW.offsets.length);
+
+        // Mark neighbor list section.
+        faissHNSW.neighbors.markSection(input, Integer.BYTES);
+
+        // HNSW graph parameters
+        faissHNSW.entryPoint = input.readInt();
+
+        faissHNSW.maxLevel = input.readInt();
+
+        // We don't use this field. It's for index building.
+        final int efConstruction = input.readInt();
+
+        faissHNSW.efSearch = input.readInt();
+
+        // dummy read a deprecated field.
+        input.readInt();
+
+        return faissHNSW;
+    }
+}
--- a/src/main/java/org/opensearch/knn/index/codec/luceneonfaiss/FaissHNSWVectorReader.java
+++ b/src/main/java/org/opensearch/knn/index/codec/luceneonfaiss/FaissHNSWVectorReader.java
@@ -1,0 +1,86 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.knn.index.codec.luceneonfaiss;
+
+import org.apache.lucene.codecs.hnsw.FlatVectorScorerUtil;
+import org.apache.lucene.codecs.hnsw.FlatVectorsScorer;
+import org.apache.lucene.index.VectorEncoding;
+import org.apache.lucene.search.KnnCollector;
+import org.apache.lucene.store.IndexInput;
+import org.apache.lucene.util.Bits;
+import org.apache.lucene.util.IOSupplier;
+import org.apache.lucene.util.hnsw.HnswGraphSearcher;
+import org.apache.lucene.util.hnsw.OrdinalTranslatedKnnCollector;
+import org.apache.lucene.util.hnsw.RandomVectorScorer;
+
+import java.io.Closeable;
+import java.io.IOException;
+
+public class FaissHNSWVectorReader implements Closeable {
+    private static FlatVectorsScorer VECTOR_SCORER = FlatVectorScorerUtil.getLucene99FlatVectorsScorer();
+
+    private IndexInput indexInput;
+    private FaissIdMapIndex faissIdMapIndex;
+    private FaissIndexFlat faissIndexFlat;
+    private LuceneFaissHnswGraph faissHnswGraph;
+
+    public FaissHNSWVectorReader(IndexInput indexInput) throws IOException {
+        this.indexInput = indexInput;
+        faissIdMapIndex = (FaissIdMapIndex) FaissIndex.load(indexInput);
+        final FaissHNSWFlatIndex faissHNSWFlatIndex = faissIdMapIndex.getNestedIndex();
+        faissIndexFlat = faissHNSWFlatIndex.getStorage();
+        faissHnswGraph = new LuceneFaissHnswGraph(faissIdMapIndex.getNestedIndex(), indexInput);
+    }
+
+    public void search(float[] target, KnnCollector knnCollector, Bits acceptDocs) throws IOException {
+        search(VectorEncoding.FLOAT32,
+               () -> VECTOR_SCORER.getRandomVectorScorer(faissIndexFlat.getSimilarityFunction(),
+                                                         faissIdMapIndex.getFloatValues(indexInput),
+                                                         target
+               ),
+               knnCollector,
+               acceptDocs
+        );
+    }
+
+    private void search(
+        final VectorEncoding vectorEncoding,
+        final IOSupplier<RandomVectorScorer> scorerSupplier,
+        final KnnCollector knnCollector,
+        final Bits acceptDocs
+    ) throws IOException {
+        if (faissIndexFlat.getTotalNumberOfVectors() == 0 || knnCollector.k() == 0
+            || faissIndexFlat.getVectorEncoding() != vectorEncoding) {
+            return;
+        }
+
+        final RandomVectorScorer scorer = scorerSupplier.get();
+        final KnnCollector collector = new OrdinalTranslatedKnnCollector(knnCollector, scorer::ordToDoc);
+        final Bits acceptedOrds = scorer.getAcceptOrds(acceptDocs);
+
+        if (knnCollector.k() < scorer.maxOrd()) {
+            HnswGraphSearcher.search(scorer, collector, faissHnswGraph, acceptedOrds);
+        } else {
+            // if k is larger than the number of vectors, we can just iterate over all vectors
+            // and collect them.
+            for (int i = 0; i < scorer.maxOrd(); i++) {
+                if (acceptedOrds == null || acceptedOrds.get(i)) {
+                    if (!knnCollector.earlyTerminated()) {
+                        knnCollector.incVisitedCount(1);
+                        knnCollector.collect(scorer.ordToDoc(i), scorer.score(i));
+                    } else {
+                        break;
+                    }
+                }
+            }
+        }  // End if
+    }
+
+    @Override
+    public void close() throws IOException {
+        indexInput.close();
+    }
+}
--- a/src/main/java/org/opensearch/knn/index/codec/luceneonfaiss/FaissIdMapIndex.java
+++ b/src/main/java/org/opensearch/knn/index/codec/luceneonfaiss/FaissIdMapIndex.java
@@ -1,0 +1,173 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.knn.index.codec.luceneonfaiss;
+
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
+import org.apache.lucene.index.FloatVectorValues;
+import org.apache.lucene.store.IndexInput;
+import org.apache.lucene.util.Bits;
+
+import java.io.IOException;
+
+/**
+ * A FAISS index with an ID mapping that maps the internal vector ID to a logical ID, along with the actual vector index.
+ * It first delegates the vector search to its nested vector index, then transforms the vector ID into a logical index that is
+ * understandable by upstream components. This is particularly useful when not all Lucene documents are indexed with a vector field.
+ * For example, if 70% of the documents have a vector field and the remaining 30% do not, the FAISS vector index will still assign
+ * increasing and continuous vector IDs starting from 0.
+ * However, these IDs only cover the sparse 30% of Lucene documents, so an ID mapping is needed to convert the internal physical vector ID
+ * into the corresponding Lucene document ID.
+ * If the mapping is an identity mapping, where each `i` is mapped to itself, we omit storing it to save memory.
+ */
+public class FaissIdMapIndex extends FaissIndex {
+    public static final String IXMP = "IxMp";
+
+    @Getter
+    private FaissHNSWFlatIndex nestedIndex;
+    private long[] vectorIdToDocIdMapping;
+    private long oneVectorByteSize;
+
+    /**
+     * Partially load id mapping and its nested index to which vector searching will be delegated.
+     *
+     * @param input An input stream for a FAISS HNSW graph file, allowing access to the neighbor list and vector locations.
+     * @return {@link FaissIdMapIndex} instance consists of index hierarchy.
+     * @throws IOException
+     */
+    public static FaissIdMapIndex load(IndexInput input) throws IOException {
+        FaissIdMapIndex faissIdMapIndex = new FaissIdMapIndex();
+        readCommonHeader(input, faissIdMapIndex);
+        FaissIndex nestedIndex = FaissIndex.load(input);
+        if (nestedIndex instanceof FaissHNSWFlatIndex) {
+            faissIdMapIndex.nestedIndex = (FaissHNSWFlatIndex) nestedIndex;
+        } else {
+            throw new IllegalStateException("Invalid nested index. Expected FaissHNSWFlatIndex, but got " + nestedIndex.getIndexType());
+        }
+        faissIdMapIndex.oneVectorByteSize = faissIdMapIndex.nestedIndex.getStorage().getOneVectorByteSize();
+
+        // Load `idMap`
+        final long numElements = input.readLong();
+        long[] vectorIdToDocIdMapping = new long[(int) numElements];
+        input.readLongs(vectorIdToDocIdMapping, 0, vectorIdToDocIdMapping.length);
+
+        // If `idMap` is an identity function that maps `i` to `i`, then we don't need to keep it.
+        for (int i = 0; i < vectorIdToDocIdMapping.length; i++) {
+            if (vectorIdToDocIdMapping[i] != i) {
+                // Only keep it if it's not an identify mapping.
+                faissIdMapIndex.vectorIdToDocIdMapping = vectorIdToDocIdMapping;
+                break;
+            }
+        }
+
+        return faissIdMapIndex;
+    }
+
+    @Override
+    public String getIndexType() {
+        return IXMP;
+    }
+
+    public FloatVectorValues getFloatValues(IndexInput indexInput) throws IOException {
+        if (vectorIdToDocIdMapping == null) {
+            return denseFloatValues(indexInput);
+        }
+
+        return sparseFloatValues(indexInput);
+    }
+
+    private FloatVectorValues denseFloatValues(IndexInput indexInput) throws IOException {
+        final FaissIndexFlat indexFlat = nestedIndex.getStorage();
+        final Storage codes = indexFlat.getCodes();
+        final int dimension = getDimension();
+        final int totalNumVectors = (int) getTotalNumberOfVectors();
+
+        @RequiredArgsConstructor
+        class DenseFloatVectorValuesImpl extends FloatVectorValues {
+            final IndexInput data;
+            final float[] vector = new float[dimension];
+
+            @Override
+            public float[] vectorValue(int targetOrd) throws IOException {
+                data.seek(oneVectorByteSize * targetOrd);
+                data.readFloats(vector, 0, vector.length);
+                return vector;
+            }
+
+            @Override
+            public int dimension() {
+                return dimension;
+            }
+
+            @Override
+            public int size() {
+                return totalNumVectors;
+            }
+
+            @Override
+            public FloatVectorValues copy() throws IOException {
+                return new DenseFloatVectorValuesImpl(indexInput.slice("FaissIndexFlat", codes.baseOffset, codes.sectionSize));
+            }
+        }
+
+        return new DenseFloatVectorValuesImpl(indexInput.slice("FaissIndexFlat", codes.baseOffset, codes.sectionSize));
+    }
+
+    private FloatVectorValues sparseFloatValues(IndexInput indexInput) throws IOException {
+        final FaissIndexFlat indexFlat = nestedIndex.getStorage();
+        final Storage codes = indexFlat.getCodes();
+        final int dimension = getDimension();
+        final int totalNumVectors = (int) getTotalNumberOfVectors();
+
+        @RequiredArgsConstructor
+        class SparseFloatVectorValuesImpl extends FloatVectorValues {
+            final IndexInput data;
+            final float[] vector = new float[dimension];
+
+            @Override
+            public float[] vectorValue(int targetOrd) throws IOException {
+                data.seek(oneVectorByteSize * targetOrd);
+                data.readFloats(vector, 0, vector.length);
+                return vector;
+            }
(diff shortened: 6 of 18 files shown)


0ctopus13prime:
Please note that will make sure all `System.out` for debugging to be removed after finalized before merging.

src/main/java/org/opensearch/knn/partialloading/KdyPerfCheck.java
```diff
@@ -0,0 +1,271 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.knn.partialloading;
+
+public class KdyPerfCheck {
```
0ctopus13prime, line 8: This is temp class for tracking performance.
Will be removed before merging to main.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriter.java
```diff
@@ -106,7 +106,7 @@ public void flush(int maxDoc, final Sorter.DocMap sortMap) throws IOException {
             final QuantizationState quantizationState = train(field.getFieldInfo(), knnVectorValuesSupplier, totalLiveDocs);
             // Check only after quantization state writer finish writing its state, since it is required
             // even if there are no graph files in segment, which will be later used by exact search
-            if (shouldSkipBuildingVectorDataStructure(totalLiveDocs)) {
+            if (false /*TMP*/ && shouldSkipBuildingVectorDataStructure(totalLiveDocs)) {
```
0ctopus13prime, line 109: This is temp code. Will revert it back before merging.

0ctopus13prime:
# Partial Loading Code Review Breaks Down

# 1. Goal

This document provides a comprehensive overview of a big PR on partial loading to minimize the time required for reviewers to complete a review.


# 2. Scope

Design Document : [Partial Loading Low Level Design 2](https://quip-amazon.com/d21MA230mNTr)

**1. Supported Vector Types**

* Only `float32` vectors are supported initially.
* Binary and byte vector indices are not yet supported.

**2.. Supported Metrics**

* Dot product.
* Euclidean distance.

**3.. Filtered Query**

* Partial loading supports filtered queries.

**4.. Nested Vectors**

* Supported for scenarios where parent documents contain multiple vectors.
* Integer parent IDs are provided in `KNNWeight`.

**5.. Sparse Vector Documents**

* Supports cases where not all Lucene documents contain vectors.
* Handles indexing documents without vectors.





# 3. Break Downs

The PR can be divided into two main parts, with the search part further split into five subparts:

1. **Index Loading**
    1. Graceful resource cleanup.
2. **Searching**
    1. Basic framework.
    2. Normal case: No filtering, no parent IDs, and all documents have indexed vectors.
    3. Filtering:
        1. Filtered queries.
        2. Handling deletions.
    4. Handling parent IDs.
    5. Sparse vector documents.



# 4. [Part 1] Index partial loading

1. **[NativeMemoryLoadStrategy](https://github.com/opensearch-project/k-NN/pull/2405/files#diff-7ceef664bd5b61a0c6b137d637698547f3bd382700c73164a2af7eb11c4b3e69)**
    1. Fetches mapping configuration from settings to check if the current KNN field supports partial loading. If partial loading is disabled, it falls back to the default mode, loading everything into memory.
        1. Currently, retrieving this configuration from settings is not implemented and can be replaced with a placeholder for now.
2. **Partial Loading in FAISS**
    1. Source : `partialloading.faiss` package.
    2. `FaissIndex.partialLoad(InputStream input)` is the entry point for partially loading a FAISS index by reading bytes from the provided `InputStream`. The main idea is to mark starting offsets and load bytes on demand.
        1. `FaissIndex.partialLoad` is a Java port of a corresponding function in FAISS.
            1. Please refer to [FAISS C++ source code](https://github.com/facebookresearch/faiss/blob/3f3d18d2e7821356a8ad9105a236a6230a720528/faiss/impl/index_read.cpp#L529).
        2. Supported index types:
            1. `IxMp` - FaissIdMapIndex
                1. Containing a mapping that maps an internal vector id to Lucene document id.
            2. `IHNf` - FaissHNSWFlatIndex
                1. Contains FaissHNSW
            3. `IxF2` - FaissIndexFlat
                1. For Euclidian distance.
            4. `IxFI` - FaissIndexFlat
                1. For inner product distance.
3. **Resource Cleanup**
    1. `PartialLoadingContext` may hold a non-null `IndexInput` reference, which is passed to a search thread for vector searches (e.g., HNSW graph search).
    2. Graceful resource cleanup is managed in `NativeMemoryAllocation.IndexAllocation.close`, which invokes `PartialLoadingContext.close` to release the `IndexInput`.



# 5. [Part 2] Search

## 2.1. Partial Loading Basic Framework

1. The flow will reach at `KNNWeight.doANNSearch`.
2. Retrieves the configured partial loading mode from settings. [Not yet implemented]
3. If partial loading is disabled, it falls back to the default search using the C++ FAISS.
4. If partial loading is enabled:
    1. Obtains `PartialLoadingContext` from `IndexAllocation`.
    2. Retrieves the search strategy based on the partial loading mode.
        1. Currently, the only available strategy is `MemoryEfficientPartialLoadingSearchStrategy`, which accesses and loads bytes on demand without caching.
    3. Copies `IndexInput`.
    4. Extracts the `efSearch` value from the query.
    5. Calls `queryIndex` of the selected search strategy.
    6. Invokes `FaissIndex.search` to perform the search.
5. Sources
    1. KNNWeight
    2. MemoryEfficientPartialLoadingSearchStrategy
    3. PartialLodingContext
    4. FaissIndex
    5. FaissIdMapIndex → FaissHNSWFlatIndex → FaissIndexFlat



## 2.2. Normal Case — Happy Path

This is the straightforward case with no filtering IDs, parent IDs, and all documents having indexed vectors.

1. **FaissIdMapIndex**:
    1. Operates without a grouper or selector.
    2. Delegates the search directly to the nested index, `FaissHNSWFlatIndex`.
2. **FaissHNSWFlatIndex**:
    1. Creates a max-heap based on the distance metric.
    2. Passes the heap to `FaissHNSW` to initiate HNSW search.
3. **FaissHNSW**:
    1. Executes the HNSW search.



## 2.3. Having a Filtering

**With Filtering**:

* If filtering is applied, `filterIdsBitSet` will have a non-null value in `doANNSearch`.
    * Live bits (representing "live" documents) are included in the bitset only when a filter is specified in the query.

**No Integer List Conversion**:

* Unlike C++ FAISS, running a vector search in partial loading in the JVM does not require converting the bitset into an integer list.
* The search can directly use the bitset as provided.



## 2.4. Having Parent Ids

**Parent IDs Handling**:

* Parent IDs are passed down to `FaissIndex`.

**Conversion to BitSet**:

* The passed parent IDs are converted into a bitset.
    * Refer to the comments in `BitSetParentIdGrouper` for details.

**Grouper Creation**:

* A grouper is created to map child document IDs to their corresponding parent document IDs.

**Parent-Level BFS in HNSW**:

* During BFS in HNSW, the max heap based on distance considers only the parent IDs.
    * For implementation, see `GroupedDistanceMaxHeap`.
    * Example: Child IDs (1, 2, 3) with parent ID '4'. The max heap evaluates distances at the parent level only.
        * But, we keep tracking the max child per each parent id though.



## 2.5. Sparse Vector Documents

1. **Handling Sparse Vectors**:

* If some documents lack indexed vectors, `vectorIdToDocIdMapping` in `FaissIdMapIndex` will hold a non-null value.
    * Example: If only documents 1, 5, and 10 have vectors, the mapping will be:
        * 0 → 1
        * 1 → 5
        * 2 → 10

shatejas (commented):
Still working on this PR

- Should the cache key change [here](https://github.com/opensearch-project/k-NN/pull/2405/files#diff-b0eb527e944c912b803db37ed9dd3dfa7689864b77d7146a0d45d7b73138d7b8R317) for partial loading? This depends on how we are planning to launch it but might be a good idea to change the cache key to avoid any unpredictable behavior

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -388,10 +429,113 @@ private Map<Integer, Float> doANNSearch(
...
+        int[] parentIds,
+        SpaceType spaceType
+    ) throws IllegalAccessException, IOException {
+        KdyPerfCheck.start();
+
+        if (k <= 0) {
+            throw new UnsupportedOperationException("Partial loading does not support radius query with k=0");
+        }
+
+        final PartialLoadingSearchStrategy searchStrategy = partialLoadingContext.getPartialLoadingMode().createSearchStrategy();
```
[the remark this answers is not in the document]
0ctopus13prime: Like the idea, only thing bugging me is that it would be cumbersome to define common set of parameters as those required parameters are varied between partial loading and the baseline. 
It is natural as we're trying to abstract out two different mechanisms -- partial loading vs calling C++ FAISS.

1. Partial loading is fine with bitset, but baseline needs long array
2. Generally, partial loading needs more parameters that baseline does not need at all.

I'm open to it, but I have an impression that we're forcefully fitting two different mechanisms under one interface. 
Please share your thoughts.

src/main/java/org/opensearch/knn/partialloading/PartialLoadingContext.java
```diff
@@ -0,0 +1,43 @@
...
+import org.opensearch.knn.partialloading.faiss.FaissIndex;
+import org.opensearch.knn.partialloading.search.PartialLoadingMode;
+
+import java.io.Closeable;
+import java.io.IOException;
+
+@RequiredArgsConstructor
+@Getter
+public class PartialLoadingContext implements Closeable {
+    private final FaissIndex faissIndex;
```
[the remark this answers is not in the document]
0ctopus13prime: Partial loading context has required FAISS components for searching.
And FaissIndex a top level index that will recursively delegate search and get the search results.

Partial loading context will be defined in IndexAllocation after 'partial load' in native load strategy.
Then in KNNWeight, it retrieves the context to get the entry point for searching. I think we need to have it here.

src/main/java/org/opensearch/knn/partialloading/PartialLoadingContext.java
```diff
@@ -0,0 +1,43 @@
...
+    private final String vectorFileName;
+    private final PartialLoadingMode partialLoadingMode;
```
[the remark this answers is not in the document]
0ctopus13prime: When loading in native memory load strategy, it will retrieve the mode from mapping then put it in here.
Then in KNNWeight, after it acquired the context, it will creates partial load search strategy based on the mode without having to look up mapping.

src/main/java/org/opensearch/knn/partialloading/PartialLoadingContext.java
```diff
@@ -0,0 +1,43 @@
...
+    private IndexInput indexInput;
+
+    public synchronized IndexInput getIndexInput(Directory directory) throws IOException {
+        if (indexInput != null) {
+            return indexInput.clone();
+        }
+        indexInput = directory.openInput(vectorFileName, IOContext.RANDOM);
```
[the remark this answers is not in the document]
0ctopus13prime: 1. Sure, I think we should let partial load search strategy to handle the low level IOContext/ReadAdvice.
2. Sure, will update in the next revision to make it initialize in the constructor.

0ctopus13prime:
Will clean up commit messages before squash merge

0ctopus13prime:
## By passing cache manager
Partial loading design was inspired by Lucene, where `IndexInput` is used to fetch bytes during vector search.
The key difference merely lies in underlying file format, but both approaches share the same algorithm with `IndexInput`.

Since we rely on NRT reader, which periodically triggers `Codec` to create vector readers for newly published segments, there's no need for us to do lazy loading when query comes in.

If we could fully delegate this responsibility to NRT reader, maintenance will become much easier.
The idea is to move the partial loading logic into `Codec`, returning `PartialLoadingVectorReader`, which will handle vector search within `KNNWeight`. This eliminates the need to keep `partial loading` logic within the load strategy and index allocation object.

I'll assess the required efforts, and if feasible, will refactor in the next revision. Otherwise, will take an iterative approach in subsequent PRs.

cc @shatejas

Vikasht34 (commented):
I am not really conformable with the change we are doing to enable Partial Loading , from the change it fells like we are re-writing whole faiss search path again in Java. 

After looking at Faiss Library , Faiss has in built capability to partailly load the file the way IndexInput is doing for us using MMAP 


Faiss already supports Mmap based loading 

```
/*
 * Copyright (c) Meta Platforms, Inc. and affiliates.
 *
 * This source code is licensed under the MIT license found in the
 * LICENSE file in the root directory of this source tree.
 */

// I/O code for indexes

#ifndef FAISS_INDEX_IO_H
#define FAISS_INDEX_IO_H

#include <cstdio>
#include <string>
#include <typeinfo>
#include <vector>

/** I/O functions can read/write to a filename, a file handle or to an
 * object that abstracts the medium.
 *
 * The read functions return objects that should be deallocated with
 * delete. All references within these objectes are owned by the
 * object.
 */

namespace faiss {

struct Index;
struct IndexBinary;
struct VectorTransform;
struct ProductQuantizer;
struct IOReader;
struct IOWriter;
struct InvertedLists;

/// skip the storage for graph-based indexes
const int IO_FLAG_SKIP_STORAGE = 1;

void write_index(const Index* idx, const char* fname, int io_flags = 0);
void write_index(const Index* idx, FILE* f, int io_flags = 0);
void write_index(const Index* idx, IOWriter* writer, int io_flags = 0);

void write_index_binary(const IndexBinary* idx, const char* fname);
void write_index_binary(const IndexBinary* idx, FILE* f);
void write_index_binary(const IndexBinary* idx, IOWriter* writer);

// The read_index flags are implemented only for a subset of index types.
const int IO_FLAG_READ_ONLY = 2;
// strip directory component from ondisk filename, and assume it's in
// the same directory as the index file
const int IO_FLAG_ONDISK_SAME_DIR = 4;
// don't load IVF data to RAM, only list sizes
const int IO_FLAG_SKIP_IVF_DATA = 8;
// don't initialize precomputed table after loading
const int IO_FLAG_SKIP_PRECOMPUTE_TABLE = 16;
// don't compute the sdc table for PQ-based indices
// this will prevent distances from being computed
// between elements in the index. For indices like HNSWPQ,
// this will prevent graph building because sdc
// computations are required to construct the graph
const int IO_FLAG_PQ_SKIP_SDC_TABLE = 32;
// try to memmap data (useful to load an ArrayInvertedLists as an
// OnDiskInvertedLists)
const int IO_FLAG_MMAP = IO_FLAG_SKIP_IVF_DATA | 0x646f0000;

Index* read_index(const char* fname, int io_flags = 0);
Index* read_index(FILE* f, int io_flags = 0);
Index* read_index(IOReader* reader, int io_flags = 0);

IndexBinary* read_index_binary(const char* fname, int io_flags = 0);
IndexBinary* read_index_binary(FILE* f, int io_flags = 0);
IndexBinary* read_index_binary(IOReader* reader, int io_flags = 0);

void write_VectorTransform(const VectorTransform* vt, const char* fname);
void write_VectorTransform(const VectorTransform* vt, IOWriter* f);

VectorTransform* read_VectorTransform(const char* fname);
VectorTransform* read_VectorTransform(IOReader* f);

ProductQuantizer* read_ProductQuantizer(const char* fname);
ProductQuantizer* read_ProductQuantizer(IOReader* reader);

void write_ProductQuantizer(const ProductQuantizer* pq, const char* fname);
void write_ProductQuantizer(const ProductQuantizer* pq, IOWriter* f);

void write_InvertedLists(const InvertedLists* ils, IOWriter* f);
InvertedLists* read_InvertedLists(IOReader* reader, int io_flags = 0);

} // namespace faiss

#endif


```
In case if we need to make additinal change to make faiss work , it would be very minimal . But with this change , we are going to have really good challenging with maintaince.

0ctopus13prime:
> I am not really conformable with the change we are doing to enable Partial Loading , from the change it fells like we are re-writing whole faiss search path again in Java. I know we must have discussed during design review but did not anticipate the magnitude of change....

Thank you @Vikasht34 for taking it a look.
1. But MMap is not enough. We want to run the search on S3 as well, and it is not feasible for that case. 
2. Also, your presented code is to give an option for user to be able to use MMap to load 'everything' into memory instead of calling `fread`. Therefore, with the apis you shared does not solve the problem we want to solve -- Running vector search on FAISS in memory constraint environment anyway.

And agree with the maintenance costs. But on the flip side, with this, user could run FAISS in memory constraints environment which solves the current FAISS limitation that requires memory heavy, expensive hardware.
It is not an exaggeration saying that user could run 500GB FAISS index in 32GB instance which is not even possible at the moment. 
Please look at the benefit user can take from this.

src/main/java/org/opensearch/knn/partialloading/faiss/hnsw/FaissHNSW.java
```diff
@@ -0,0 +1,325 @@
...
+        final IdAndDistance nearest = new IdAndDistance(entryPoint, distanceComputer.compute(entryPoint));
+        final IntArray neighborList = new IntArray(16);
+        for (int level = maxLevel; level >= 1; --level) {
+            greedyUpdateNearest(neighborListInput, distanceComputer, level, nearest, neighborList);
+        }
+        KdyPerfCheck.hnswGreedyDone();
+        KdyPerfCheck.setEntryPointAfterGreedy(nearest.id, nearest.distance);
+
+        // Start exhaustive search at the bottom layer.
+        final SparseFixedBitSet visited = new SparseFixedBitSet(Math.toIntExact(totalNumberOfVectors));
```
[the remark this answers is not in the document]
0ctopus13prime: Hmm... after having a quick calculation, memory space should be the same for all no?
For N elements, the memory space required is still O(N) for all I believe. 🤔
Actually it is O(N / 64) for SparseFixedBitSet, but still.

src/main/java/org/opensearch/knn/partialloading/faiss/hnsw/FaissHNSW.java
```diff
@@ -0,0 +1,325 @@
...
+                candidates.insertWithOverflow(neighborId, dist);
+                // System.out.println("neighborId=" + neighborId + ", dist=" + dist);
+                if (!hasSelector || matchDocSelector.test(neighborId)) {
+                    resultMaxHeap.insertWithOverflow(neighborId, dist);
+                }
+            }
+        }
+    }
+
+    private static class IntArray {
```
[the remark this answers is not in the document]
0ctopus13prime: It is commonly used pattern in Lucene and I could not see GC pressure burden or extra latency in the benchmark.
In text searching - https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/SegmentTermsEnumFrame.java#L192-L194

Actually, I keep getting much faster latency than Lucene right now..

And it will stop once it reaches the maximum neighbor list size after few trials (I guess 3-4 times?)
So you're right, will add extra costs, but I think its impact should be minimal. (and benchmark result is also showing that)

Vikasht34:
> 1. But MMap is not enough. We want to run the search on S3 as well, and it is not feasible for that case.

Let's have one more discussion among us. 

> > I am not really conformable with the change we are doing to enable Partial Loading , from the change it fells like we are re-writing whole faiss search path again in Java. I know we must have discussed during design review but did not anticipate the magnitude of change....
> 
> Thank you @Vikasht34 for taking it a look.
> 
> 1. But MMap is not enough. We want to run the search on S3 as well, and it is not feasible for that case.
> 2. Also, your presented code is to give an option for user to be able to use MMap to load 'everything' into memory instead of calling `fread`. Therefore, with the apis you shared does not solve the problem we want to solve -- Running vector search on FAISS in memory constraint environment anyway.
> 
> And agree with the maintenance costs. But on the flip side, with this, user could run FAISS in memory constraints environment which solves the current FAISS limitation that requires memory heavy, expensive hardware. It is not an exaggeration saying that user could run 500GB FAISS index in 32GB instance which is not even possible at the moment. Please look at the benefit user can take from this.


Let's Discuss this in detail , I think we need to revisit again some part of it.

0ctopus13prime:
In the new revision, I added two parts. But please note that testings are still on the way, as they can be affected by the result of discussion we have in this PR. But core unit tests will be shipped shortly.

1.. Added `use_lucene_searcher` boolean flag. This will make engine to force to use Lucene searcher on FAISS index. 
This is reflecting the discussion we had to use Lucene HNSW graph searcher on FAISS index. It is disabled by default, meaning make FAISS engine to load everything into bytes, and let C++ code to perform vector search. 
Once it is enabled, it will be treated like semi-lucene while underlying index would be FAISS for HNSW graph but core algorithm will be Lucene. We can risk the subtle different output with the same input on FAISS.

2.. Logic for loading HNSW graph from FAISS index. In here, it will try to create LuceneFaissHnswGraph which is an implementation of HnswGraph holding graph index from FAISS. At the same time, VectorScorer will load vector from FAISS index then compute the distance using Lucene's similarity function.

Please note that the current scope for this PR is aiming at:
1. Faiss engine.
2. HNSW index type
3. L2, Inner product metric types are only supported for now.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
```diff
@@ -124,18 +196,27 @@ public void search(String field, float[] target, KnnCollector knnCollector, Bits
...
         }
-        throw new UnsupportedOperationException("Search functionality using codec is not supported with Native Engine Reader");
+
+        // Try with Lucene searcher
+        final FaissHNSWVectorReader vectorReader = faissHNSWVectorReaderMap.get(field);
+        if (vectorReader != null) {
+            try {
+                vectorReader.search(target, knnCollector, acceptDocs);
+            } catch (Exception e) {
+                // KDY
```
0ctopus13prime, line 214: This will be removed before finalizing.

0ctopus13prime:
Will continuously post the benchmark results. Below is the acquired result of cohere-1m dataset
Conclusion : FAISS shows 120% faster than Lucene or Lucene on FAISS. 

# Results
1. FAISS : 
Min Throughput,prod-queries,183.85,ops/s
Mean Throughput,prod-queries,203.44,ops/s
Median Throughput,prod-queries,204.80,ops/s
Max Throughput,prod-queries,206.10,ops/s
50th percentile latency,prod-queries,3.754240504349582,ms
90th percentile latency,prod-queries,4.312251597002614,ms
99th percentile latency,prod-queries,4.800578189897352,ms
99.9th percentile latency,prod-queries,5.713345205411337,ms
99.99th percentile latency,prod-queries,8.915240254473746,ms
100th percentile latency,prod-queries,40.61461999663152,ms

2. FAISS + Use Lucene Searcher : 
Min Throughput,prod-queries,151.49,ops/s
Mean Throughput,prod-queries,170.19,ops/s
Median Throughput,prod-queries,171.07,ops/s
Max Throughput,prod-queries,172.54,ops/s
50th percentile latency,prod-queries,4.654030504752882,ms
90th percentile latency,prod-queries,5.472216008638497,ms
99th percentile latency,prod-queries,6.232268399908208,ms
99.9th percentile latency,prod-queries,6.879154362148264,ms
99.99th percentile latency,prod-queries,8.669991978490653,ms
100th percentile latency,prod-queries,43.74621399620082,ms

3. Lucene
Min Throughput,prod-queries,143.24,ops/s
Mean Throughput,prod-queries,169,ops/s
Median Throughput,prod-queries,170.25,ops/s
Max Throughput,prod-queries,173.4,ops/s
50th percentile latency,prod-queries,4.6506,ms
90th percentile latency,prod-queries,5.36926,ms
99th percentile latency,prod-queries,6.10365,ms
99.9th percentile latency,prod-queries,6.96111,ms
99.99th percentile latency,prod-queries,40.5339,ms
100th percentile latency,prod-queries,68.4304,ms

# Set up
- Instance type : c6a.32xlarge
- Recall : `Mean recall@k,prod-queries,0.93, Mean recall@1,prod-queries,0.99`
- JVM args : `-Xms100G -Xmx100G`

src/main/java/org/opensearch/knn/index/query/KNNQueryFactory.java
```diff
@@ -121,7 +121,6 @@ public static Query create(CreateQueryRequest createQueryRequest) {
             requestEfSearch = (Integer) methodParameters.get(METHOD_PARAMETER_EF_SEARCH);
         }
         int luceneK = requestEfSearch == null ? k : Math.max(k, requestEfSearch);
-        log.debug("Creating Lucene k-NN query for index: {}, field:{}, k: {}", indexName, fieldName, k);
```
0ctopus13prime, line 124: Oops, will revert this.

src/main/java/org/opensearch/knn/index/query/KNNQueryFactory.java
```diff
@@ -70,7 +70,7 @@ public static Query create(CreateQueryRequest createQueryRequest) {
             );
         }
 
-        if (KNNEngine.getEnginesThatCreateCustomSegmentFiles().contains(createQueryRequest.getKnnEngine())) {
+        if (!forceUseLuceneSearcher && KNNEngine.getEnginesThatCreateCustomSegmentFiles().contains(createQueryRequest.getKnnEngine())) {
```
kotwanikunal, line 73: nit: the standard convention across OpenSearch is `if(forceUseLuceneSearcher == false && ..)`

src/main/java/org/opensearch/knn/index/engine/faiss/FaissHNSWMethod.java
```diff
@@ -105,6 +107,14 @@ private static MethodComponent initMethodComponent() {
                 )
             )
             .addParameter(METHOD_ENCODER_PARAMETER, initEncoderParameter())
+            .addParameter(
+                USE_LUCENE_HNSW_SEARCHER,
```
kotwanikunal, line 111: Are we exposing this to the user as a query parameter?

src/main/java/org/opensearch/knn/index/codec/luceneonfaiss/FaissHNSW.java
```diff
@@ -0,0 +1,92 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.knn.index.codec.luceneonfaiss;
```
kotwanikunal, line 6: Shouldn't it be `faissonlucene`?

src/main/java/org/opensearch/knn/index/codec/luceneonfaiss/FaissHNSW.java
```diff
@@ -0,0 +1,92 @@
...
+/**
+ * Ported implementation of the FAISS HNSW graph search algorithm.
+ * While it follows the same steps as the original FAISS implementation, differences in how the JVM and C++ handle floating-point
+ * calculations can lead to slight variations in results. However, such cases are very rare, and in most instances, the results are
+ * identical to FAISS. Even when there are ranking differences, they do not impact the precision or recall of the search.
+ * For more details, refer to the [FAISS HNSW implementation](
+ * <a href="https://github.com/facebookresearch/faiss/blob/main/faiss/impl/HNSW.h">...</a>).
+ */
+@Getter
+public class FaissHNSW {
```
kotwanikunal, line 22: If your intention is to make this feature experiemental - please add in the following - 

```
/**
..
..
 * @opensearch.experimental
*/

@ExperimentalApi
@Getter
public class FaissHNSW {
...
```

src/main/java/org/opensearch/knn/index/codec/luceneonfaiss/FaissHNSWVectorReader.java
```diff
@@ -0,0 +1,86 @@
...
+import org.apache.lucene.util.IOSupplier;
+import org.apache.lucene.util.hnsw.HnswGraphSearcher;
+import org.apache.lucene.util.hnsw.OrdinalTranslatedKnnCollector;
+import org.apache.lucene.util.hnsw.RandomVectorScorer;
+
+import java.io.Closeable;
+import java.io.IOException;
+
+public class FaissHNSWVectorReader implements Closeable {
+    private static FlatVectorsScorer VECTOR_SCORER = FlatVectorScorerUtil.getLucene99FlatVectorsScorer();
```
navneet1v, line 23: we should pass this scorer from the caller of this class rathe than creating it here.

src/main/java/org/opensearch/knn/index/codec/luceneonfaiss/FaissIndexFlatIP.java
```diff
@@ -0,0 +1,16 @@
...
+package org.opensearch.knn.index.codec.luceneonfaiss;
+
+import org.apache.lucene.index.VectorSimilarityFunction;
+
+public class FaissIndexFlatIP extends FaissIndexFlat {
+
+    @Override
+    public VectorSimilarityFunction getSimilarityFunction() {
+        return VectorSimilarityFunction.DOT_PRODUCT;
+    }
```
navneet1v, line 15: from lucene side we should use MAX_INNER_PRODUCT

src/main/java/org/opensearch/knn/index/codec/luceneonfaiss/LuceneOnFaissUtils.java
```diff
@@ -0,0 +1,25 @@
...
+
+package org.opensearch.knn.index.codec.luceneonfaiss;
+
+import org.opensearch.knn.common.KNNConstants;
+
+import java.util.Map;
+
+public final class LuceneOnFaissUtils {
+    private LuceneOnFaissUtils() {
+    }
```
navneet1v, line 14: use @UtlityClass as an annotation from Lombok

navneet1v (commented):
passed through the code, I think rather than merging this change in main we should cut a feature branch develop on that. 

@0ctopus13prime the code has reduced pretty much but I feel some of the things in the interfaces can be modified and made more restrictive. Also the way we are enabling the lucene based HNSW search for faiss engine doesn't seem correct. For first change lets use IndexSettings.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
```diff
@@ -124,18 +196,27 @@ public void search(String field, float[] target, KnnCollector knnCollector, Bits
...
+                e.printStackTrace();
```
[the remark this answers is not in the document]
0ctopus13prime: It's temp code. (marked by KDY)
Will remove it before merging.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
```diff
@@ -48,12 +61,71 @@ public class NativeEngines990KnnVectorsReader extends KnnVectorsReader {
...
+                            final FaissHNSWVectorReader vectorReader = new FaissHNSWVectorReader(indexInput);
+                            faissHNSWVectorReaderMap.put(fieldInfo.getName(), vectorReader);
+                        } catch (Exception e) {
+                            // If something went bad, we close the stream and rethrow
+                            try {
+                                indexInput.close();
+                            } catch (Exception ioException) {
+                                // Ignore
+                            }
+                            throw e;
```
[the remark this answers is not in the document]
0ctopus13prime: I don't think we can use IOUtils here, as we want to suppress any exceptions being thrown.

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased 3.0](https://github.com/opensearch-project/k-NN/compare/2.x...HEAD)
 ### Features
+* Added a boolean flag that enable user to use Lucene searcher on FAISS index. [#2405](https://github.com/opensearch-project/k-NN/pull/2405)
```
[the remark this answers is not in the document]
0ctopus13prime: may I ask why?

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
```diff
@@ -48,12 +61,71 @@ public class NativeEngines990KnnVectorsReader extends KnnVectorsReader {
...
+        this.faissHNSWVectorReaderMap = new HashMap<>(2 * state.fieldInfos.size(), 0.6f);
         loadCacheKeyMap();
+        loadFaissIndexForLuceneSearcher(segmentReadState);
+    }
+
+    private void loadFaissIndexForLuceneSearcher(SegmentReadState state) {
+        for (FieldInfo fieldInfo : state.fieldInfos) {
+            // Ex: {"index_description":"HNSW16,Flat","spaceType":"l2","name":"hnsw","data_type":"float",
+            // "parameters":{"use_lucene_searcher":true,"ef_search":100,"ef_construction":100,"encoder":{"name":"flat","parameters":{}}}}
+            final String parametersString = fieldInfo.getAttribute(KNNConstants.PARAMETERS);
```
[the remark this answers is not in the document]
0ctopus13prime: but why is that? putting the config at field level will give more flexibility.

src/main/java/org/opensearch/knn/index/codec/luceneonfaiss/FaissHNSWVectorReader.java
0ctopus13prime: it's a static scorer that is initialized at static scope.
In which, it will pick up Panama if it can be used, otherwise Default will be used.
So it is actually about static scorer not meant to be used for runtime.

src/main/java/org/opensearch/knn/index/codec/luceneonfaiss/FaissIndexFlatIP.java
0ctopus13prime: oops, yeah, correct.
Nice catch, did not applied the change in this PR

src/main/java/org/opensearch/knn/index/codec/luceneonfaiss/FaissIndexFlatL2.java
```diff
@@ -0,0 +1,15 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.knn.index.codec.luceneonfaiss;
+
+import org.apache.lucene.index.VectorSimilarityFunction;
+
+public class FaissIndexFlatL2 extends FaissIndexFlat {
```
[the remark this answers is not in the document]
0ctopus13prime: Prefer to have it public for unit testings. 
Otherwise I had to use reflection to create a new one.

src/main/java/org/opensearch/knn/index/codec/luceneonfaiss/LuceneFaissHnswGraph.java
```diff
@@ -0,0 +1,105 @@
...
+        final long end = o + faissHnsw.getCumNumberNeighborPerLevel()[level + 1];
+        loadNeighborIdList(begin, end);
+    }
+
+    private void loadNeighborIdList(final long begin, final long end) {
+        // Make sure we have sufficient space for neighbor list
+        final long maxLength = end - begin;
+        if (neighborIdList == null || neighborIdList.length < maxLength) {
+            neighborIdList = new int[(int) (maxLength * 1.5)];
+        }
```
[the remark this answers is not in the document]
0ctopus13prime: It's not exactly equal to the number of neighbors. it's size is `maxLength * 1.5`.

src/main/java/org/opensearch/knn/index/codec/luceneonfaiss/FaissIdMapIndex.java
```diff
@@ -0,0 +1,173 @@
...
+
+        return faissIdMapIndex;
+    }
+
+    @Override
+    public String getIndexType() {
+        return IXMP;
+    }
+
+    public FloatVectorValues getFloatValues(IndexInput indexInput) throws IOException {
```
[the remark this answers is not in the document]
0ctopus13prime: This is an input to VectorScorer. This is an intermediate accessor to be provided to Lucene searcher.

0ctopus13prime:
> Looking at the code, Im just not sure how this will be maintainble/extendable. It seems like if either lucene or faiss is upgraded or we make changes to the calling client code, this could break (potentially silently). Also, every feature added in faiss, we will have to add into the java code.

Hi @jmazanec15, thank you for taking it a look.

Wondering, what might be break 'in silence' in the future? File format will be preserved in FAISS (I got a confirmation from the team), and any interface changes will break the build explicitly so I don't think that will happen in silence. And an interface change will break other features in OpenSearch anyway, I don' think that will be unique for this PR

And not every feature needs to be added to Java code. This will be all for float vectors. Features we want is already in Lucene. What we need to provide to Lucene is to the way of accessing vectors from faiss index, that would be all!

src/main/java/org/opensearch/knn/common/KNNConstants.java
```diff
@@ -120,6 +120,8 @@ public class KNNConstants {
     public static final List<String> FAISS_SQ_ENCODER_TYPES = List.of(FAISS_SQ_ENCODER_FP16);
     public static final String FAISS_SIGNED_BYTE_SQ = "SQ8_direct_signed";
     public static final String FAISS_SQ_CLIP = "clip";
+    public static final String USE_LUCENE_HNSW_SEARCHER = "use_lucene_searcher";
```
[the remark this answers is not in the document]
0ctopus13prime: We can discuss it more, but do you have suggestion?
This PR is exactly delegate Lucene searcher on FAISS index. Having clear name will give user better understanding.
But I'm open to it.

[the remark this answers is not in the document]
jmazanec15: My issue with it is most users are not going to be familiar with the behavior characteristics of lucene vs faiss. So, this name wont give any indication as to what change in behavior they might expect. So, I think we need to go with something else.

A couple ones I can think of:
- mode: "on-disk" vs "in-memory"
- memory_optimized: true
- partial_load: true
- lazy_load: true

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
```diff
@@ -48,12 +61,71 @@ public class NativeEngines990KnnVectorsReader extends KnnVectorsReader {
...
     public NativeEngines990KnnVectorsReader(final SegmentReadState state, final FlatVectorsReader flatVectorsReader) {
         this.flatVectorsReader = flatVectorsReader;
         this.segmentReadState = state;
         this.cacheKeys = getVectorCacheKeysFromSegmentReaderState(state);
+        this.faissHNSWVectorReaderMap = new HashMap<>(2 * state.fieldInfos.size(), 0.6f);
         loadCacheKeyMap();
+        loadFaissIndexForLuceneSearcher(segmentReadState);
+    }
+
+    private void loadFaissIndexForLuceneSearcher(SegmentReadState state) {
```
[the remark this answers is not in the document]
0ctopus13prime: sure, we can do it. but I'm having a difficulty of imagining how it works, could you give me more details?

src/main/java/org/opensearch/knn/index/codec/luceneonfaiss/FaissHNSW.java
0ctopus13prime: So the idea is to apply lucene search on 'FAISS'.

src/main/java/org/opensearch/knn/index/query/KNNQueryFactory.java
0ctopus13prime: thank you, will update in the next rev.

src/main/java/org/opensearch/knn/index/engine/faiss/FaissHNSWMethod.java
0ctopus13prime: Will put it at index level config

src/main/java/org/opensearch/knn/index/codec/luceneonfaiss/FaissHNSW.java
0ctopus13prime: So the idea is to apply 'lucene' on 'Faiss'.

0ctopus13prime:
@navneet1v 
The above perf benchmark is to see the core performance of Lucene HNSW on FAISS index after loaded everything into memory using MMapDirectory. It at least gives us the upper bound of performance in ideal world. 

Working on running benchmark under memory constraints environment to see where it shines. 
X variables would be physical free space, and Y axis will be latency (or qps).

This will give us more insight for us to decide on which environment this feature can bring benefits.

src/main/java/org/opensearch/knn/common/KNNConstants.java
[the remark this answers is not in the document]
0ctopus13prime: yeah, fair points. Thank you for the suggestion. Let's discuss more about it.

0ctopus13prime:
Hi all, once gathered feedbacks then will merge into another branch in origin first.
Then will work on adding unit tests + IT tests then merge it to main branch.

In this PR, hoping to align on the name of flags, structure wise topics. 
In the second PR, will add basic unit tests + IT tests.
In the last PR, will raise a PR to make the bool flag updatable. Ex. enabled -> disabled or vice versa.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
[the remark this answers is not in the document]
jmazanec15: One idea is just hve the engine return a vectorSearch similar to the FaissHNSWVectorReader

src/main/java/org/opensearch/knn/index/query/KNNQueryFactory.java
navneet1v, line 73: This logic should not be part of this PR. Wether we have to use Lucene provided query or we should use the NativeEngine query is a completely separate thing. If we add this logic right now it will bi-furcate the query when different settings are set. 

My recommendation would be to have a separate GH issue talking about should we use Lucene based queries for our native engines or not. Having this logic as part of this PR or the partial loading effort just increase the scope of the PR.

Just to iterate again, I am not against using the Lucene query clauses its just having partially onboarded to Lucene query clauses(for faiss engine) without a proper plan make me nervous to have this change.

navneet1v: I found another usecase where we cannot use Lucene Query, so for all the old indices < 2.17 version which use DcoValues are using KNNVectorsFormat. So as they are not using KNNVectorsFormat, search using new KNNVectorReaders will not work. My recommendation here would be to do the Search using readers but from the KNNWeight.java by ensuring the the field is using the new KNNVectorsFormat. 

cc: @jmazanec15 , @shatejas , @0ctopus13prime , @Vikasht34

0ctopus13prime: @navneet1v 
Hi Naveet, hope to understand the second comment more, but why those indices < 2.17 matter? 🤔 
Isn't it required for user to upgrade the version to 3.0.0 to have this feature?
As long as OpenSearch knn plugin is using codec in 3.0.0, I don't see any risk here.

navneet1v: So if a cluster upgrade from 2.16 to 2.19 and then to 3.0. The index will remain same. It will still keep on using DocValues. and now as we have migrated to 3.0 user can set the setting which is getting created in this PR.

This is one usecase I was able to think in terms of BWC. There can be many more, hence my recommendation is to not go with Lucene query in this PR. Move the logic to use Reader based query in KNNWeight class. and tackle the problem of using Lucene query in separate GH issue.

navneet1v: Just thought of another use-case which will break here, since we are just checking the index setting, and not seeing whether this index is faiss HNSW or not. If the index is IVF PQ then the search will not work.

0ctopus13prime: Can we can limit this feature exclusively to FAISS float vector on FAISS 3.0 index??

It's my bad that I did not include those defensive logics you mentioned in this PR, but we can start small with validation logic, then gradually expand its territory. Will add validation logic in next PR.

Reusing Lucene query will significantly minimize the implementation effort. Basically for dense case, sparse case, nested query, filter query, it will give us the results with zero development cost! 
Don't get me wrong, I'm open to refactor this, but can we limit the scope in this first PR and focus to integrate Lucene's HNSW graph searcher on FAISS index first? 
After merged this PR, I'm more than happy to discuss how we can expand the scope to accommodate other use cases.

navneet1v: > Can we can limit this feature exclusively to FAISS float vector on FAISS 3.0 index??
> 
> It's my bad that I did not include those defensive logics you mentioned in this PR, but we can start small with validation logic, then gradually expand its territory. Will add validation logic in next PR.
> 
> Reusing Lucene query will significantly minimize the implementation effort. Basically for dense case, sparse case, nested query, filter query, it will give us the results with zero development cost! Don't get me wrong, I'm open to refactor this, but can we limit the scope in this first PR and focus to integrate Lucene's HNSW graph searcher on FAISS index first? After merged this PR, I'm more than happy to discuss how we can expand the scope to accommodate other use cases.

See to get to a stage to add these extra checks you have to call the MapperService and then bunch index version checks. This will create a whole lot of problem and latencies if you still keep this is the QueryFactory. Hence my suggestion was move this 1 layer down. If you do this check in `KNNQuery.createWeight()` all the checks can be added there as you have all the fieldInfo already available, engine info and what not. FYI, then disk based vector search will be supported once you add support for BinaryFaiss Index. 


IMO, this actually far better than what classes of Lucene you are trying to use.

src/main/java/org/opensearch/knn/index/codec/luceneonfaiss/FaissIndex.java
```diff
@@ -0,0 +1,84 @@
...
+@Getter
+public abstract class FaissIndex {
+    // Vector dimension
+    private int dimension;
+    // Total number of vectors saved within this index.
+    private long totalNumberOfVectors;
+    // Space type used to index vectors in this index.
+    private SpaceType spaceType;
+
+    public static FaissIndex load(IndexInput input) throws IOException {
```
shatejas, line 28: Please add some documentation and context here with a clear stopping condition for the recursion documented. It can be really hard to understand and debug recursive behavior without context

src/main/java/org/opensearch/knn/partialloading/storage/Storage.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+package org.opensearch.knn.partialloading.storage;
+
+import lombok.Getter;
+import org.apache.lucene.store.IndexInput;
+
+import java.io.IOException;
+
+public class Storage {
```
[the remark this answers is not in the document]
0ctopus13prime: will change it in the next rev.

src/main/java/org/opensearch/knn/index/codec/luceneonfaiss/FaissIndex.java
0ctopus13prime: sure, will add it in the next revision.

0ctopus13prime:
Hi all, will split this PR into multiple small PRs.
Please refer to https://github.com/opensearch-project/k-NN/issues/2401#issuecomment-2699777824 for more details

## 3. PR Plans
1. High level interfaces.
In this PR, I will add foundational interfaces at the Codec level. For now, there will be only one implementation: FAISS.

2. Top level FAISS index with id mapping - IxMp section.
In this PR, will add a part deserializing IxMp section holding id mapping table.
After shipped, it will provide sparse/dense FloatValues and ByteValues.

3. FAISS HNSW
This PR introduces deserialization of the HNSW graph, including vector IDs and the neighbor list. Once the PR is merged, Lucene’s HNSW graph searcher will be able to navigate the FAISS HNSW graph. However, it will not yet support vector search, as the flat layers have not been added.

4. Float vector + HNSW
This PR introduces deserialization of the float flat vector section in the FAISS index, such as IxF2 or IxFI.
After shipped, Lucene HNSW graph searcher will be able to perform vector search on FAISS index where data type is float.

5. Byte vector + HNSW
This PR introduces deserialization of the byte flat vector section in the FAISS index, such as IxSQ
After shipped, Lucene HNSW graph searcher will be able to perform vector search on FAISS index where data type is byte.

6. FP16 Support
This PR introduces a custom RandomVectorScorer implementation that loads Float16 vectors and computes similarity with the query vector. Once merged, FAISS HNSW with FP16 encoding can be served using Lucene’s HNSW graph searcher.

7. Complete Faiss Loading Part at Codec Level
This PR introduces functionality to fully load a FAISS index within a segment using IndexInput. Once merged, VectorReader will be able to search the FAISS index when memory_optimized_search is enabled.

8. Index Settings Update
This PR introduces a setting update, ensuring that updates are allowed only for the FAISS HNSW method. It also ensures proper deallocation of off-heap memory. Once merged, users will be able to update the setting to enable Lucene’s HNSW graph searcher to perform vector search on a FAISS index.

9. Add unit tests.

10. Add integration testings.

11. Merge to main branch.

PR #2406, opened 2025-01-18
Separate QuantizationStateCache teardown from KNNTestCase

owenhalpert:
### Description
`QuantizationStateCacheManager.getInstance()` will create an instance of a QuantizationStateCache if one is not already present. Since the cache is not created by default (like the NativeMemoryCache is), the call to `QuantizationStateCacheManager.getInstance().close()` in KNNTestCase will wastefully create instances in many cases. There are very few test classes in which a QuantizationStateCache is created, so it will be better to move this logic somewhere more specific.

Importantly, a side effect of the current flow is [KNNWeightTests](https://github.com/opensearch-project/k-NN/blob/main/src/test/java/org/opensearch/knn/index/query/KNNWeightTests.java#L135) mocks KNNSettings but doesn't mock the QS cache settings, so when the cache is created due to the call above, a NullPointerException is thrown and all the tests in the class fail when running in isolation:
`java.lang.NullPointerException: Cannot invoke "org.opensearch.core.common.unit.ByteSizeValue.getKb()" because the return value of "org.opensearch.knn.index.KNNSettings.getSettingValue(String)" is null`

This error doesn't show up when running the full suite of tests, presumably because the default setting is applied elsewhere. 

### Check List
- [NA] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
- [NA] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).

### DIFF
--- a/src/main/java/org/opensearch/knn/index/memory/NativeMemoryCacheManager.java
+++ b/src/main/java/org/opensearch/knn/index/memory/NativeMemoryCacheManager.java
@@ -116,7 +116,9 @@ public class NativeMemoryCacheManager implements Closeable {
         if (threadPool != null && !threadPool.scheduler().isShutdown()) {
             startMaintenance(cache);
         } else {
-            logger.warn("ThreadPool is null or scheduler is terminated during NativeMemoryCacheManager initialization. Maintenance will not start.");
+            logger.warn(
+                "ThreadPool is null or scheduler is terminated during NativeMemoryCacheManager initialization. Maintenance will not start."
+            );
         }
     }
 
--- a/src/main/java/org/opensearch/knn/index/memory/NativeMemoryCacheManager.java
+++ b/src/main/java/org/opensearch/knn/index/memory/NativeMemoryCacheManager.java
@@ -116,7 +116,7 @@ public class NativeMemoryCacheManager implements Closeable {
         if (threadPool != null) {
             startMaintenance(cache);
         } else {
-            logger.warn("ThreadPool is null or during NativeMemoryCacheManager initialization. Maintenance will not start.");
+            logger.warn("ThreadPool is null during NativeMemoryCacheManager initialization. Maintenance will not start.");
         }
     }
 
--- a/src/main/java/org/opensearch/knn/index/memory/NativeMemoryCacheManager.java
+++ b/src/main/java/org/opensearch/knn/index/memory/NativeMemoryCacheManager.java
@@ -113,12 +113,10 @@ public class NativeMemoryCacheManager implements Closeable {
         accessRecencyQueue = new ConcurrentLinkedDeque<>();
         cache = cacheBuilder.build();
 
-        if (threadPool != null && !threadPool.scheduler().isShutdown()) {
+        if (threadPool != null) {
             startMaintenance(cache);
         } else {
-            logger.warn(
-                "ThreadPool is null or scheduler is terminated during NativeMemoryCacheManager initialization. Maintenance will not start."
-            );
+            logger.warn("ThreadPool is null or during NativeMemoryCacheManager initialization. Maintenance will not start.");
         }
     }
 
--- a/src/main/java/org/opensearch/knn/index/memory/NativeMemoryCacheManager.java
+++ b/src/main/java/org/opensearch/knn/index/memory/NativeMemoryCacheManager.java
@@ -113,10 +113,10 @@ public class NativeMemoryCacheManager implements Closeable {
         accessRecencyQueue = new ConcurrentLinkedDeque<>();
         cache = cacheBuilder.build();
 
-        if (threadPool != null) {
+        if (threadPool != null && !threadPool.scheduler().isShutdown()) {
             startMaintenance(cache);
         } else {
-            logger.warn("ThreadPool is null during NativeMemoryCacheManager initialization. Maintenance will not start.");
+            logger.warn("ThreadPool is null or scheduler is terminated during NativeMemoryCacheManager initialization. Maintenance will not start.");
         }
     }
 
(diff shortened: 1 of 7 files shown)


src/test/java/org/opensearch/knn/quantization/models/quantizationState/QuantizationStateCacheTests.java
```diff
@@ -42,8 +42,9 @@ public void setThreadPool() {
     }
 
     @After
-    public void terminateThreadPool() {
+    public void stopMaintenance() throws IOException {
         terminate(threadPool);
+        QuantizationStateCache.getInstance().close();
```
owenhalpert, line 47: This call to close (which solely cancels the maintenance) isn't necessary because of [how the RescheduledRunnable is implemented](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/threadpool/Scheduler.java#L230-L271). When we terminate the threadpool, the next task is rejected and the task is canceled anyway. 

I thought I'd include it to be clear, but open to thoughts.

src/main/java/org/opensearch/knn/index/memory/NativeMemoryCacheManager.java
```diff
@@ -113,10 +113,12 @@ private void initialize(NativeMemoryCacheManagerDto nativeMemoryCacheDTO) {
         accessRecencyQueue = new ConcurrentLinkedDeque<>();
         cache = cacheBuilder.build();
 
-        if (threadPool != null) {
+        if (threadPool != null && !threadPool.scheduler().isShutdown()) {
```
owenhalpert, line 116: Scheduling maintenance (using `scheduleWithFixedDelay`) on a threadpool that has already been shutdown will cause RejectedExecutionExceptions. The threadpool is set during initialization to be the OpenSearch maintenance threadpool, so the variable should always be set and functioning. Should the threadpool be shutdown, the `threadPool` variable may still be set so this extra check is more safe.

src/test/java/org/opensearch/knn/index/memory/NativeMemoryCacheManagerTests.java
```diff
@@ -54,6 +54,7 @@ public void tearDown() throws Exception {
         client().admin().cluster().updateSettings(clusterUpdateSettingsRequest).get();
         NativeMemoryCacheManager.getInstance().close();
         terminate(threadPool);
+        NativeMemoryCacheManager.setThreadPool(null);
```
kotwanikunal, line 57: nit: I don't think we need this explicitly.

src/test/java/org/opensearch/knn/quantization/models/quantizationState/QuantizationStateCacheTests.java
```diff
@@ -42,8 +42,10 @@ public void setThreadPool() {
     }
 
     @After
-    public void terminateThreadPool() {
+    public void stopMaintenance() throws IOException {
```
kotwanikunal, line 45: nit: you can use cleanup() or tearDown() as a generic method name for test class cleanup.

src/test/java/org/opensearch/knn/quantization/models/quantizationState/QuantizationStateCacheTests.java
kotwanikunal: We can have it for code clarity. My concern is the `setThreadPool(null)` pattern. 
I know it exists solely for the tests but the threadpool termination and close() should ideally take care of all the cleanup.

Going one step further - I think the `setThreadPool` should be on a final variable with a non-null check.

src/test/java/org/opensearch/knn/index/memory/NativeMemoryCacheManagerTests.java
Gankris96: I guess the close of NativeMemoryCacheManager should have this. Because that is where the `executor.shutdown()` is called. The same place should be used to cleanup all tasks. I see the `maintenanceTask.cancel()` is called. Post that we should clear out the threadpool; no?

owenhalpert: @Gankris96 the only consideration there is that the threadpool is set only once in KNNPlugin where we have access to pass in OpenSearch's threadpool. Once close() is called and that is set to null, we lose that reference unless `KNNPlugin.createComponents` is called again. I would need to investigate the impact of getting rid of that reference fully.

src/test/java/org/opensearch/knn/quantization/models/quantizationState/QuantizationStateCacheTests.java
Gankris96: I guess NativeMemoryCacheManager will always exist because it is created as part of plugin initialization. Thats why the `NativeMemoryCacheManager.getInstance().close()` works but`QuantizationStateCacheManager.getInstance().close();` doesn't

Why cant we just have something like `public boolean quantizationCacheEnabled;` as part of `KNNTestCase.java`
and inside the `resetState()` have a check like - 

```
// Clean up the cache
        NativeMemoryCacheManager.getInstance().invalidateAll();
        NativeMemoryCacheManager.getInstance().close();
        if (quantizationCacheEnabled) {
            QuantizationStateCacheManager.getInstance().close();
        }
```

Only in the tests where we actually have QuantizationStateCacheManager created, we can go set this value to true

src/test/java/org/opensearch/knn/index/memory/NativeMemoryCacheManagerTests.java
Gankris96: I think the problem is that the threadpool is getting shutdown in the tests before you clear out the QSCache object. That is what causes the `threadpool != null` check to succeed but the `mockNode.shutDown()` terminates the threadpool.

PR #2429, opened 2025-01-23
K-NN Plugin Upgrade with Lucene 10.0.1

Vikasht34:
### Description
As per Lucene 10.0.1 upgrade,  There are lots of breaking changes. This PR consumes Lucene 10.0.1 as part of Opensearch 3.0.0 snapshot and Fixes all breaking changes.

Note :- This  PR Does not aim to refactor any part of code, It's just making compatible with Lucene 10.0.1

### Resolves this issues
https://github.com/opensearch-project/k-NN/issues/2434
https://github.com/opensearch-project/k-NN/issues/2433

### Check List


martin-gaievski:
it looks like `opensearch-knn-test-fixtures` needs to consume latest corre main with lucene 10

```
/home/runner/work/k-NN/k-NN/src/main/java/org/opensearch/knn/index/codec/KNNCodecVersion.java:16: error: package org.apache.lucene.backward_codecs.lucene912 does not exist
import org.apache.lucene.backward_codecs.lucene912.Lucene912Codec;
```

Vikasht34:
> it looks like `opensearch-knn-test-fixtures` needs to consume latest corre main with lucene 10
> 
> ```
> /home/runner/work/k-NN/k-NN/src/main/java/org/opensearch/knn/index/codec/KNNCodecVersion.java:16: error: package org.apache.lucene.backward_codecs.lucene912 does not exist
> import org.apache.lucene.backward_codecs.lucene912.Lucene912Codec;
> ```


Yes ...Agree ...We need to change to alpha !! lots of things started breaking ..This Pr would be updated with all latest changes.

.github/workflows/backwards_compatibility_tests_workflow.yml
```diff
@@ -35,7 +35,7 @@ jobs:
       matrix:
         java: [ 21 ]
         os: [ubuntu-latest]
-        bwc_version : [ "2.0.1", "2.1.0", "2.2.1", "2.3.0", "2.4.1", "2.5.0", "2.6.0", "2.7.0", "2.8.0", "2.9.0", "2.10.0", "2.11.0", "2.12.0", "2.13.0", "2.14.0", "2.15.0", "2.16.0", "2.17.0", "2.18.0", "2.19.0-SNAPSHOT"]
+        bwc_version : [ "2.0.1", "2.1.0", "2.2.1", "2.3.0", "2.4.1", "2.5.0", "2.6.0", "2.7.0", "2.8.0", "2.9.0", "2.10.0", "2.11.0", "2.12.0", "2.13.0", "2.14.0", "2.15.0", "2.16.0", "2.17.0", "2.18.0","2.19.0-SNAPSHOT", "2.20.0-SNAPSHOT"]
```
Vikasht34, line 38: This I had just put for Passing the Builds and BWC , will revert back to 2.19 once 2.19 is released.

src/main/java/org/opensearch/knn/index/query/common/DocAndScoreQuery.java
```diff
@@ -62,92 +61,104 @@ public int count(LeafReaderContext context) {
...
-                        }
-                        return docs[upTo] - context.docBase;
                     }
 
                     @Override
-                    public int docID() {
-                        return docIdNoShadow();
+                    public long cost() {
+                        // Estimate the cost of the scoring operation, if applicable.
+                        return DocIdSetIterator.NO_MORE_DOCS;
```
navneet1v, line 161: same as above

src/main/java/org/opensearch/knn/index/vectorvalues/KNNVectorValuesIterator.java
```diff
@@ -71,18 +71,28 @@ public interface KNNVectorValuesIterator {
...
-        DocIdsIteratorValues(@NonNull final DocIdSetIterator docIdSetIterator) {
-            validateIteratorType(docIdSetIterator);
+        private final DocIdSetIterator docIdSetIterator;
+        private KnnVectorValues knnVectorValues = null; // Added reference to KnnVectorValues
+        @Getter
+        @Setter
+        private int lastOrd = -1;
+        @Getter
+        @Setter
+        private Object lastAccessedVector = null;
```
navneet1v, line 81: why we need setters and getters here?

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
```diff
@@ -199,14 +199,6 @@ public void close() throws IOException {
         }
     }
 
-    /**
-     * Return the memory usage of this object in bytes. Negative values are illegal.
-     */
-    @Override
-    public long ramBytesUsed() {
-        return flatVectorsReader.ramBytesUsed();
-    }
```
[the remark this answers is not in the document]
Vikasht34: This function has been removed from Reader , so no need to override and it's not getting used anywhere. Lucene now have new Interface 

```
public interface Accountable {

  /** Return the memory usage of this object in bytes. Negative values are illegal. */
  long ramBytesUsed();

  /**
   * Returns nested resources of this class. The result should be a point-in-time snapshot (to avoid
   * race conditions).
   *
   * @see Accountables
   */
  default Collection<Accountable> getChildResources() {
    return Collections.emptyList();
  }

  /** An accountable that always returns 0 */
  Accountable NULL_ACCOUNTABLE = () -> 0;
}
```

and this is wrapped in RAMUsageEstimator.

src/main/java/org/opensearch/knn/index/vectorvalues/KNNVectorValuesIterator.java
Vikasht34: There is a change in MergeByteValues , that we can not use getValue(ord) more than once , becuase they check if lastord != ord +1 then fails , so we need this to track this before calling mergeByteValues in Extract method in Stragetey.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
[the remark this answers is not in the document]
navneet1v: will we implement this interface somewhere now? or that it will be taken in next PR.

src/main/java/org/opensearch/knn/index/KNNVectorScriptDocValues.java
```diff
@@ -81,12 +80,13 @@ public float[] get(int i) {
      * @return A KNNVectorScriptDocValues object based on the type of the values.
      * @throws IllegalArgumentException If the type of values is unsupported.
      */
-    public static KNNVectorScriptDocValues create(DocIdSetIterator values, String fieldName, VectorDataType vectorDataType) {
+    public static KNNVectorScriptDocValues create(Object values, String fieldName, VectorDataType vectorDataType) {
```
navneet1v, line 83: can we avoid passing the `Object` here? Can we use KNNVectorValues here.

src/main/java/org/opensearch/knn/index/vectorvalues/KNNVectorValuesIterator.java
navneet1v, line 81: same as the old comment, lets try to avoid Object data type. Please have a GH issue on this and should be taken as part of next PR.

src/main/java/org/opensearch/knn/index/vectorvalues/KNNVectorValuesIterator.java
```diff
@@ -71,18 +71,28 @@ public interface KNNVectorValuesIterator {
...
-        protected DocIdSetIterator docIdSetIterator;
-        private static final List<Function<DocIdSetIterator, Boolean>> VALID_ITERATOR_INSTANCE = List.of(
-            (itr) -> itr instanceof BinaryDocValues,
-            (itr) -> itr instanceof FloatVectorValues,
-            (itr) -> itr instanceof ByteVectorValues
-        );
-
-        DocIdsIteratorValues(@NonNull final DocIdSetIterator docIdSetIterator) {
-            validateIteratorType(docIdSetIterator);
+        private final DocIdSetIterator docIdSetIterator;
```
navneet1v, line 74: I think this class will also need a bit of refactoring, for now to unblock main this is fine. Lets have a GH issue for this to fix the abstraction. Because this class otherwise is not going to be maintainable in future

navneet1v (approved):
Approving the code. Lets have the follow up GH issues and PR on the refactoring. Since k-NN plugin is dependency for other plugins I am approving this PR so that we can unblock other plugin main branches

PR #2430, opened 2025-01-23
Remove skip building graph check for quantization use case

VijayanB:
### Description
For quantization indices, we don't have to apply building graph check since it is already faster, this is now only applied for fp32/16 indices and where threshold is configured.

### Related Issues

### Check List


VijayanB:
I am raising directly to 2.x since i am blocked because of lucene codec issue. Will backport this to main once it is fixed.

VijayanB:
Currently quantization state is not null only for on-disk mode. We don't support in-memory quantization at this moment. Hence, this check is good enough. However if we want to check for on-disk, we have to pass this as parameter and add it to check.

@jmazanec15 @shatejas @Vikasht34 @navneet1v  What do you think?

jmazanec15:
In general, the "mode" is meant to select defaults for other parameters, so it should not be directly used to disable and enable certain features. Instead, it may resolve to a parameter that then makes this decision.

Anyway, in the PR, it appears that you actually disable if quantization is used. I think in general something like this is okay, but might be misleading/confusing to users who set the setting anticipating that graph builds will be skipped.

That being said, is this just a concern of performance, or is it a functional bug? If it is performance related only, Id recommend using mode to resolve a per-field value for skipping graph build. Then, users could override if need be. This would be consistent with current approach around mode.

shatejas (commented):
Do we need to update documentation around this?

jmazanec15 (approved):

VijayanB:
> Do we need to update documentation around this?

Yes, i will update that this parameter will be no-hop for lucene, and on-disk

PR #2435, opened 2025-01-24
Fix race condition bug in NativeMemoryCacheManager

### LINKED ISSUE #2262: [BUG] Race condition with approximate query leading to Index has already been closed exception.
**What is the bug?**
There is a race condition in the current code path for the native memory cache manager. An entry from the NativeMemoryCacheManager can be evicted before being read for the query, resulting in an "Index has already been closed" exception.

The issue stems from the current code flow, which acquires a read lock after the entry is loaded into the cache. This disjoint between loading and locking operations creates a window where the cache can evict the entry before the read lock is acquired.

Code References
Load to cache: [KNNWeight.java#L286-L302](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/query/KNNWeight.java#L286-L302)
Read Lock: [KNNWeight.java#L313](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/query/KNNWeight.java#L313)
Exception: [KNNWeight.java#L316-L318](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/query/KNNWeight.java#L316-L318)

The error is reproducible with force eviction turned on along with running a OSB vectorsearch benchmark with 5+ clients

- Query to go through

 - OS: Linux AL2
 - Version OS 2.17
 - Plugins - kNN

owenhalpert:
### Description
Previously, the window between loading an item from its nativeMemoryContext and acquiring a read lock allowed for the item to be cleaned or evicted before being read for the query. The goal here is to synchronize the load and incRef() operations to avoid this race condition.

We were seeing the following exception when running an OSB vectorsearch workload with 5+ search clients:
`[ERROR] search_phase_execution_exception ({'error': {'root_cause': [{'type': 'illegal_state_exception', 'reason': "IndexAllocation-Reference is already closed can't increment refCount current count [0]"}`

With these changes, we are able to consistently execute the entire benchmark successfully with an error rate of 0%

### Related Issues
Resolves #2262 

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/memory/NativeMemoryCacheManager.java
+++ b/src/main/java/org/opensearch/knn/index/memory/NativeMemoryCacheManager.java
@@ -333,10 +333,19 @@ public class NativeMemoryCacheManager implements Closeable {
         }
 
         if (KNNFeatureFlags.isForceEvictCacheEnabled()) {
-            logger.info("NEWEST CHANGES: Synchronized force evict path, testing new race condition fix");
+            // Utilizes a force eviction mechanism to free up memory before the entry can be added to the cache
+            // In case of a cache hit, the operation just updates the locally maintained recency list
+            // In case of a cache miss, least recently accessed entries are evicted in a blocking manner
+            // before the new entry can be added to the cache.
             String key = nativeMemoryEntryContext.getKey();
             NativeMemoryAllocation result = cache.getIfPresent(key);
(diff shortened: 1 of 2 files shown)


naveentatikonda:
@owenhalpert are there any updates on this PR ?

PR #2443, opened 2025-01-25
[Backport Main] Added more detailed error messages for KNN model training

### LINKED ISSUE #2268: Improve error messaging/validation for faiss training

anntians:
### Description
Previously, a consistent feedback we get around PQ and IVF is that there is limited visibility into the failure cases. Part of this is because the errors are thrown on the Faiss side and we don't return stack traces in Rest response. So, this makes it difficult to use PQ and IVF. Thus, this PR provides improved error messages by adding explicit checks for the most common errors:

[ ] For PQ, explicitly check in OpenSearch an invalid configuration where m does not divide dimension
[ ] For PQ/IVF, check the number of training points matches the minimum clustering criteria defined in faiss
[ ] If there is not enough memory, explicitly say that there is not enough memory.

Adding these 3 checks will cover 90% of the training failures that occur.

### Related Issues
Resolves #2268 

### Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,15 +21,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ## [Unreleased 2.x](https://github.com/opensearch-project/k-NN/compare/2.19...2.x)
 ### Features
 ### Enhancements
-- Introduced a writing layer in native engines where relies on the writing interface to process IO. (#2241)[https://github.com/opensearch-project/k-NN/pull/2241]
-- Allow method parameter override for training based indices (#2290) https://github.com/opensearch-project/k-NN/pull/2290]
-- Optimizes lucene query execution to prevent unnecessary rewrites (#2305)[https://github.com/opensearch-project/k-NN/pull/2305]
 - Added more detailed error messages for KNN model training (#2378)[https://github.com/opensearch-project/k-NN/pull/2378]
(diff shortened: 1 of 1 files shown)


navneet1v:
@naveentatikonda do you know why the bwc test is failing here? I see you were fixing some tests. 

Does this PR require pulling changes from main for the fix?

naveentatikonda:
> @naveentatikonda do you know why the bwc test is failing here? I see you were fixing some tests.
> 
> Does this PR require pulling changes from main for the fix?

I was just fixing some bug related to infrastructure, which faiss library needs to be built based on the CPU flags. Here, the failure seems to be related to some tests. I triggered rerun to see if it is due to flakiness .

jmazanec15 (approved):

naveentatikonda (approved):

Issue #2310, opened 2024-12-04
[BUG] NPE with Faiss Byte Vectors when filtering


naveentatikonda:
**What is the bug?**
Running into NullPointerException when we are [computing](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/query/KNNWeight.java#L425-L427) `P * Dimension <= Max_distance_computations` for faiss efficient filters using byte vectors. But, it is working fine for all other scenarios.

```
{
  "error": {
    "root_cause": [
      {
        "type": "null_pointer_exception",
        "reason": "Cannot read the array length because the return value of \"org.opensearch.knn.index.query.KNNQuery.getByteQueryVector()\" is null"
      }
    ],
    "type": "search_phase_execution_exception",
    "reason": "all shards failed",
    "phase": "query",
    "grouped": true,
    "failed_shards": [
      {
        "shard": 0,
        "index": "development-20241204111110-3de6471c",
        "node": "5ry1X70gTgyIOlMItg6TzA",
        "reason": {
          "type": "null_pointer_exception",
          "reason": "Cannot read the array length because the return value of \"org.opensearch.knn.index.query.KNNQuery.getByteQueryVector()\" is null"
        }
      }
    ],
    "caused_by": {
      "type": "null_pointer_exception",
      "reason": "Cannot read the array length because the return value of \"org.opensearch.knn.index.query.KNNQuery.getByteQueryVector()\" is null",
      "caused_by": {
        "type": "null_pointer_exception",
        "reason": "Cannot read the array length because the return value of \"org.opensearch.knn.index.query.KNNQuery.getByteQueryVector()\" is null"
      }
    }
  },
  "status": 500
}
```

**Proposed Solution**
Here, we are trying to get byteQueryVector instead of queryVector. We need to get queryVector here because for byte datatype with Faiss engine we will do [byte range validation and still store it as a normal query vector](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java#L499-L513) to pass it later to Faiss ScalarQuantizer. The Faiss SQ expects the query vector as an fp32 vector.

The fix with integ test in my feature branch - https://github.com/naveentatikonda/k-NN/commit/bbeb3de10923f325a95399b7c4c5ed269764d373

Issue #2447, opened 2025-01-27
[Refactoring] Refactor query logic to handle different vector datatypes


naveentatikonda:
### Description

Refactor query logic to handle different vector datatypes, to avoid branching, running into bugs like #2310 and to improve the code quality.

PR #2448, opened 2025-01-27
Fix Faiss byte vector efficient filter bug

naveentatikonda:
### Description
Resolves a bug which is running into NullPointerException when we are [computing](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/query/KNNWeight.java#L425-L427) P * Dimension <= Max_distance_computations for faiss efficient filters using byte vectors.

This is a short term solution, created a separate github issue #2447 to refactor and handle different vector datatypes without branching, to improve the code quality and avoid running into bugs like these.  

### Related Issues
Resolves https://github.com/opensearch-project/k-NN/issues/2310

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/query/KNNWeight.java
+++ b/src/main/java/org/opensearch/knn/index/query/KNNWeight.java
@@ -45,6 +45,7 @@ import java.io.IOException;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.concurrent.ExecutionException;
 import java.util.stream.Collectors;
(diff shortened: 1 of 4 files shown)


ryanbogan (dismissed):

Vikasht34 (approved):

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -438,9 +438,9 @@ private boolean isFilteredExactSearchPreferred(final int filterIdsCount) {
          * TODO we can have a different MAX_DISTANCE_COMPUTATIONS for binary index as computation cost for binary index
          * is cheaper than computation cost for non binary vector
          */
-        return KNNConstants.MAX_DISTANCE_COMPUTATIONS >= filterIdsCount * (knnQuery.getVectorDataType() == VectorDataType.FLOAT
-            ? knnQuery.getQueryVector().length
-            : knnQuery.getByteQueryVector().length);
+        return KNNConstants.MAX_DISTANCE_COMPUTATIONS >= filterIdsCount * (knnQuery.getVectorDataType() == VectorDataType.BINARY
+            ? knnQuery.getByteQueryVector().length
+            : knnQuery.getQueryVector().length);
```
VijayanB, line 443: can we split into individual VectorDataType check to avoid else part always assumes it is non byte?

VijayanB:
Can we also cover benchmark to verify that whether one threshold is good enough or do we need different threshold per vector data type?

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
naveentatikonda: updated

naveentatikonda:
> Can we also cover benchmark to verify that whether one threshold is good enough or do we need different threshold per vector data type as long term fix?

What threshold are you referring to ?

VijayanB:
> > Can we also cover benchmark to verify that whether one threshold is good enough or do we need different threshold per vector data type as long term fix?
> 
> What threshold are you referring to ?

I think it is added as TODO just before your code change

VijayanB (approved):

martin-gaievski (approved):
looks good, thank you

Issue #2377, opened 2025-01-09
[RFC] Derived Source for Vectors


jmazanec15:
## Introduction

This is an RFC that presents a proposal for removing knn_vector from "_source" field without loss of OpenSearch functionality that "_source" enables. "_source" in this context refers to the per document field in OpenSearch that stores the original source provided by the user as a StoredField in lucene. See [SourceFieldMapper](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/mapper/SourceFieldMapper.java#L287) for more details.

This is a followup for https://github.com/opensearch-project/k-NN/pull/1571 and https://github.com/opensearch-project/k-NN/issues/1572.

## Problem

Currently, vectors for native indices are stored in 3 places by default

1. _source stored field. Vectors along with the reset of json body of the document are stored (i.e. .fdt)
2. Native library files — ANN structure and vectors are stored (i.e. .hnsw)
3. FlatVectorsFormat format — Basically doc values for vectors (i.e. .vec)

In an experiment with 10k 128-dimensional vectors, the size break down of these files was:
Total Index Size | 24.3 mb
-- | --
HNSW files | 5.91 mb
Doc values | 3.8 mb
Source | 14.6 mb

With BEST_COMPRESSION codec:
Total Index Size | 18.3 mb
-- | --
HNSW files | 5.91 mb
Doc values | 3.75 mb
Source | 8.64 mb

From this, we can see that 47%-60% of the storage is going towards the _source storage. For more details, see: https://github.com/opensearch-project/OpenSearch/issues/6356. Worse yet, for our disk based feature with our quantized vectors (vectors that get compressed in a lossy fashion), the native lib files will get smaller than the FlatVectorsFormat file, so the _source will take up an even larger percentage of the storage.

For a typical user, **they should not need to get the source vector from OpenSearch**. Thus, storing the vectors in _source poses significant problems for users with minimal benefits:

1. Users have to pay to store data they do not really need or use. This issue gets even more pronounced for disk-based vector search, where memory is no longer the bottleneck. Users end up having to provision their clusters based on storage capacity.
[...]

PR #2449, opened 2025-01-27
Introduce derived vector source via stored fields

jmazanec15:
### Description
This PR introduces derived source for flat vector field mapper via the approach outlined in #2377.

First, for some quick background for reviewers: In OpenSearch, the source refers to a per-document StoredField (key = "_source") that stores the json representation of the document. This field is not searchable - instead, it used during the fetch phase in order to return fields of documents that matched the search request to end users. For instance, a user may search an index using a knn search over a field called passage_embedding, but they really only want to get back the field passage_text. During fetch, the source field of the k-Nearest Neighbors would be fetched and the passage_text field parsed out and sent back. In addition to this, _source may used to reindex documents and or update and delete by queries.

The goal of this PR/feature is to remove the vectors from this _source field on disk, but inject them back from other data formats into the _source when needed.

To implement this, from a high level, we introduce a custom StoredFieldsFormat. On write, this StoredFieldsFormat removes the vector fields from source. On read, it puts them back if necessary. 

In its current state, this PR supports:
1. Flat (non object) vector fields
2. Single level nested vector fields (still cleaning up a bit but it passed tests in DerivedSourceIT
3. Arbitrarily nested object fields

Here are the following todo items:
- [x] Nested support. I initially had some support for nesting but this needs to be rethought
- [x] Integrate optimization for skipping source reconstruction (see https://github.com/opensearch-project/OpenSearch/pull/17080)
- [x] Introduce feature flag settings
- [x] Generic object field types
- [x] Confirm flat mappings work as expected.

Some leftovers after this PR. Leaving out now because feature is experimental for 2.19 and properly isolated behind index setting
- [ ] Increase test coverage - need to add uTs 
- [ ] Performance testing - need to show performance of reindexing and updates. 
- [ ] Look into optimization

### Related Issues
#2377 

### Check List


Vikasht34 (commented):
Took the first Pass , A really good shoutout to 

1. Clean Interfaces .
2. Good use of abstraction on Visitor and clean injection.
3.Code has good modularity .
4. Function interface and supplier.

shatejas (commented):
Reviewed the reader

src/main/java/org/opensearch/knn/index/codec/KNN9120Codec/KNN9120Codec.java
```diff
@@ -58,4 +64,14 @@ public CompoundFormat compoundFormat() {
     public KnnVectorsFormat knnVectorsFormat() {
         return perFieldKnnVectorsFormat;
     }
+
+    @Override
+    public StoredFieldsFormat storedFieldsFormat() {
+        DerivedSourceReadersSupplier derivedSourceReadersSupplier = new DerivedSourceReadersSupplier(
+            (segmentReadState) -> knnVectorsFormat().fieldsReader(segmentReadState),
```
shatejas, line 71: nit: If this is for Lucene engine, you will also open index inputs for graph files, might be a bit unnecessary. you might be able to do 
```
() -> new Lucene99FlatVectorsFormat(FlatVectorScorerUtil.getLucene99FlatVectorsScorer()).fieldsReader
```

The only caveat is that will changing lucene versions it might be hard to keep a trace

src/main/java/org/opensearch/knn/index/codec/KNN9120Codec/DerivedSourceStoredFieldsReader.java
```diff
@@ -0,0 +1,74 @@
...
+        if (shouldInject && isVisitorNeedFields) {
+            delegate.document(docId, new DerivedSourceStoredFieldVisitor(storedFieldVisitor, docId, derivedSourceVectorInjector));
+            return;
+        }
+        delegate.document(docId, storedFieldVisitor);
+    }
+
+    @Override
+    public StoredFieldsReader clone() {
+        return new DerivedSourceStoredFieldsReader(delegate.clone(), derivedSourceVectorInjector);
```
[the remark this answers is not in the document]
jmazanec15: This class is supposed to heavily delegate. So, when a certain method is called, we want to call the delegate's method. Hence, I believe this is correct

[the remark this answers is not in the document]
jmazanec15: similar example: https://github.com/apache/lucene/blob/branch_9_12/lucene/core/src/java/org/apache/lucene/index/SortingCodecReader.java#L442-L464

src/main/java/org/opensearch/knn/index/codec/KNN9120Codec/KNN9120Codec.java
jmazanec15: I tried this change and ended up getting:

src/main/java/org/opensearch/knn/index/codec/KNN9120Codec/DerivedSourceStoredFieldsReader.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    /**
+     * For merging, we need to tell the derived source stored fields reader to skip injecting the source. Otherwise,
+     * on merge we will end up just writing the source to disk
+     *
+     * @param storedFieldsReader stored fields reader to wrap
+     * @return wrapped stored fields reader
+     */
+    public static StoredFieldsReader wrapForMerge(StoredFieldsReader storedFieldsReader) {
```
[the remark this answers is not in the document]
jmazanec15: This was a good call - fixed

src/main/java/org/opensearch/knn/index/codec/KNN9120Codec/DerivedSourceStoredFieldsReader.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    @Override
+    public void checkIntegrity() throws IOException {
+        delegate.checkIntegrity();
+    }
+
+    @Override
+    public void close() throws IOException {
+        delegate.close();
```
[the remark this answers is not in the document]
jmazanec15: I implemented derivedSourceVectorInjector.close() - please have a look @shatejas

src/main/java/org/opensearch/knn/index/codec/derivedsource/RootPerFieldDerivedVectorInjector.java
```diff
@@ -0,0 +1,46 @@
...
+            derivedSourceReaders.getDocValuesProducer(),
+            derivedSourceReaders.getKnnVectorsReader()
+        );
+    }
+
+    @Override
+    public void inject(int docId, Map<String, Object> sourceAsMap) throws IOException {
+        KNNVectorValues<?> vectorValues = vectorValuesSupplier.get();
+        if (vectorValues.docId() == docId || vectorValues.advance(docId) == docId) {
+            sourceAsMap.put(fieldInfo.name, vectorValues.getVector());
```
shatejas, line 43: Do we need conditional clone vector here?

src/main/java/org/opensearch/knn/index/codec/KNN9120Codec/KNN9120Codec.java
shatejas: Makes sense, Thanks for trying. Maybe added a TODO for now

shatejas: > Will there be any penalty incurred for opening indexInput but not reading from it?

If its already open, then no since its mmapped. But otherwise it will. Was just trying to avoid it

src/main/java/org/opensearch/knn/index/codec/KNN9120Codec/DerivedSourceStoredFieldsReader.java
```diff
@@ -0,0 +1,74 @@
...
+        // If the visitor has explicitly indicated it does not need the fields, we should not inject them
+        boolean isVisitorNeedFields = true;
+        if (storedFieldVisitor instanceof FieldsVisitor) {
+            isVisitorNeedFields = derivedSourceVectorInjector.shouldInject(
+                ((FieldsVisitor) storedFieldVisitor).includes(),
+                ((FieldsVisitor) storedFieldVisitor).excludes()
+            );
+        }
+        if (shouldInject && isVisitorNeedFields) {
+            delegate.document(docId, new DerivedSourceStoredFieldVisitor(storedFieldVisitor, docId, derivedSourceVectorInjector));
```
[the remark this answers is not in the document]
shatejas: Maybe a GH issue would help track it

src/main/java/org/opensearch/knn/index/codec/derivedsource/RootPerFieldDerivedVectorInjector.java
jmazanec15: I think in general, I shouldnt need to clone the vector because it get serialized (and copied) later. Im going to remove it from above.

src/main/java/org/opensearch/knn/index/codec/KNN9120Codec/DerivedSourceStoredFieldsReader.java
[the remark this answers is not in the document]
shatejas: Looks perfect! Thanks!

src/main/java/org/opensearch/knn/index/codec/derivedsource/RootPerFieldDerivedVectorInjector.java
jmazanec15: @navneet1v can you double check my reasoning here?

src/main/java/org/opensearch/knn/index/codec/KNN9120Codec/DerivedSourceStoredFieldsReader.java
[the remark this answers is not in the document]
jmazanec15: yeah will note and create follow up issue after this completes.

src/main/java/org/opensearch/knn/index/codec/derivedsource/RootPerFieldDerivedVectorInjector.java
jmazanec15: Discussed with @navneet1v - need to clone

src/main/java/org/opensearch/knn/index/codec/KNN9120Codec/DerivedSourceStoredFieldsWriter.java
```diff
@@ -0,0 +1,117 @@
...
+    @Override
+    public int merge(MergeState mergeState) throws IOException {
+        return delegate.merge(mergeState);
+    }
+
+    @Override
+    public void writeField(FieldInfo fieldInfo, BytesRef bytesRef) throws IOException {
+        // Parse out the vectors from the source
+        if (Objects.equals(fieldInfo.name, SourceFieldMapper.NAME) && !vectorFieldTypes.isEmpty()) {
+            Tuple<? extends MediaType, Map<String, Object>> mapTuple = XContentHelper.convertToMap(
```
navneet1v, line 75: lets have some timing logs here. I think this is going to be an expensive operation for converting bytes to JSON

src/main/java/org/opensearch/knn/index/codec/KNN9120Codec/DerivedSourceStoredFieldsWriter.java
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    @Override
+    public void close() throws IOException {
+        delegate.close();
+    }
+
+    @Override
+    public long ramBytesUsed() {
+        return delegate.ramBytesUsed();
```
navneet1v, line 115: Please add the bytes used by the member variables of this class in ramBytes

src/main/java/org/opensearch/knn/index/codec/derivedsource/DerivedSourceVectorInjector.java
```diff
@@ -0,0 +1,132 @@
...
+        Map<String, Object> sourceAsMap = new HashMap<>(mapTuple.v2());
+
+        // For each vector field, add in the source. The per field injectors are responsible for skipping if
+        // the field is not present.
+        for (PerFieldDerivedVectorInjector vectorInjector : perFieldDerivedVectorInjectors) {
+            vectorInjector.inject(docId, sourceAsMap);
+        }
+
+        // At this point, we can serialize the modified source map
+        BytesStreamOutput bStream = new BytesStreamOutput(1024);
```
navneet1v, line 90: any reason for having this constant of 1024?

src/main/java/org/opensearch/knn/index/codec/derivedsource/NestedPerFieldDerivedVectorInjector.java
```diff
@@ -0,0 +1,276 @@
...
+            injectObject(parentDocId, sourceAsMap);
+            return;
+        }
+
+        // Setup the iterator. Return if no parent
+        String childFieldName = ParentChildHelper.getChildField(childFieldInfo.name);
+        String parentFieldName = ParentChildHelper.getParentField(childFieldInfo.name);
+        if (parentFieldName == null) {
+            return;
+        }
```
navneet1v, line 52: what will be the case when parentFieldName will be null?

src/main/java/org/opensearch/knn/index/codec/derivedsource/NestedPerFieldDerivedVectorInjector.java
```diff
@@ -0,0 +1,276 @@
...
+        } else if (fieldInfo.getVectorDimension() != 0 && derivedSourceReaders.getKnnVectorsReader() != null) { // the field indexes vectors
+            switch (fieldInfo.getVectorEncoding()) {
+                case FLOAT32:
+                    iterator = derivedSourceReaders.getKnnVectorsReader().getFloatVectorValues(fieldInfo.name);
+                    break;
+                case BYTE:
+                    iterator = derivedSourceReaders.getKnnVectorsReader().getByteVectorValues(fieldInfo.name);
+                    break;
+            }
+        } else if (fieldInfo.getDocValuesType() != DocValuesType.NONE && derivedSourceReaders.getDocValuesProducer() != null) { // the field
```
navneet1v, line 224: why we would have 3 types of field indexes? when this is just vector field.

src/main/java/org/opensearch/knn/index/codec/KNN9120Codec/DerivedSourceStoredFieldsFormat.java
```diff
@@ -0,0 +1,80 @@
...
+    private final DerivedSourceReadersSupplier derivedSourceReadersSupplier;
+    // IMPORTANT Do not rely on this for the reader, it will be null if SPI is used
+    private final MapperService mapperService;
+
+    @Override
+    public StoredFieldsReader fieldsReader(Directory directory, SegmentInfo segmentInfo, FieldInfos fieldInfos, IOContext ioContext)
+        throws IOException {
+        List<FieldInfo> derivedVectorFields = new ArrayList<>();
+        for (FieldInfo fieldInfo : fieldInfos) {
+            if (DERIVED_VECTOR_FIELD_ATTRIBUTE_TRUE_VALUE.equals(fieldInfo.attributes().get(DERIVED_VECTOR_FIELD_ATTRIBUTE_KEY))) {
```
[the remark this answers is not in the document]
jmazanec15: Sure - do you remember if we have a generic check on fieldInfo to see if its knn? I guess it would be checking the knn attribute

src/main/java/org/opensearch/knn/index/codec/KNN9120Codec/DerivedSourceStoredFieldsFormat.java
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    @Override
+    public StoredFieldsWriter fieldsWriter(Directory directory, SegmentInfo segmentInfo, IOContext ioContext) throws IOException {
+        StoredFieldsWriter delegateWriter = delegate.fieldsWriter(directory, segmentInfo, ioContext);
+        if (mapperService != null && KNNSettings.isKNNDerivedSourceEnabled(mapperService.getIndexSettings().getSettings())) {
+            List<String> vectorFieldTypes = new ArrayList<>();
+            for (MappedFieldType fieldType : mapperService.fieldTypes()) {
+                if (fieldType instanceof KNNVectorFieldType) {
+                    vectorFieldTypes.add(fieldType.name());
```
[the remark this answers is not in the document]
jmazanec15: my.nested.path.vector

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java
```diff
@@ -363,7 +370,8 @@ public Mapper.Builder<?> parse(String name, Map<String, Object> node, ParserCont
                 modelDaoSupplier.get(),
                 parserContext.indexVersionCreated(),
                 null,
-                null
+                null,
+                KNNSettings.isKNNDerivedSourceEnabled(parserContext.getSettings())
```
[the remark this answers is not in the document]
jmazanec15: I believe this will be false for all older versions. Will validate manually by de-registering the setting, creating an index, and then checking this value.

src/main/java/org/opensearch/knn/index/mapper/LuceneFieldMapper.java
```diff
@@ -117,6 +128,12 @@ private LuceneFieldMapper(
             this.vectorFieldType = null;
         }
 
+        if (isDerivedSourceEnabled) {
+            this.fieldType = new FieldType(this.fieldType);
```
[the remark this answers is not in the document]
jmazanec15: It gets frozen in Lucene - https://github.com/opensearch-project/k-NN/blob/135f7ebad4805cce2588aeefb5928d4e022c6a70/src/main/java/org/opensearch/knn/index/VectorDataType.java#L45. So, we need to copy and add. Ill add a comment.

src/main/java/org/opensearch/knn/index/vectorvalues/KNNVectorValuesFactory.java
```diff
@@ -72,6 +74,37 @@ public static <T> KNNVectorValues<T> getVectorValues(final FieldInfo fieldInfo,
         return getVectorValues(FieldInfoExtractor.extractVectorDataType(fieldInfo), vectorValuesIterator);
     }
 
+    /**
+     * Returns a {@link KNNVectorValues} for the given {@link FieldInfo} and {@link LeafReader}
```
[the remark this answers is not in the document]
jmazanec15: Will fix

src/main/java/org/opensearch/knn/index/vectorvalues/KNNVectorValuesFactory.java
```diff
@@ -72,6 +74,37 @@ public static <T> KNNVectorValues<T> getVectorValues(final FieldInfo fieldInfo,
...
+     *
+     * @param fieldInfo {@link FieldInfo}
+     * @param docValuesProducer {@link DocValuesProducer}
+     * @param knnVectorsReader {@link KnnVectorsReader}
+     * @return {@link KNNVectorValues}
+     */
+    public static <T> KNNVectorValues<T> getVectorValues(
+        final FieldInfo fieldInfo,
+        final DocValuesProducer docValuesProducer,
```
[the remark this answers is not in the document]
jmazanec15: Did we completely get rid of ability to use docValuesProducer for vectors? I thought it could still be possible.

src/main/java/org/opensearch/knn/index/codec/derivedsource/NestedPerFieldDerivedVectorInjector.java
jmazanec15: We shouldnt get here, but adding just as a check

src/main/java/org/opensearch/knn/index/codec/derivedsource/DerivedSourceVectorInjector.java
jmazanec15: Let me link code in core where this was copied from.

src/main/java/org/opensearch/knn/index/codec/KNN9120Codec/DerivedSourceStoredFieldsWriter.java
jmazanec15: Sure - let me see try to do that.

src/main/java/org/opensearch/knn/index/codec/KNN9120Codec/DerivedSourceStoredFieldsReader.java
```diff
@@ -0,0 +1,128 @@
...
+
+    /**
+     * For merging, we need to tell the derived source stored fields reader to skip injecting the source. Otherwise,
+     * on merge we will end up just writing the source to disk
+     *
+     * @return Merged instance that wont inject by default
+     */
+    @Override
+    public StoredFieldsReader getMergeInstance() {
+        try {
```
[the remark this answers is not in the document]
jmazanec15: The issue is that for merging, the reader is used. So, if we use this, it will add the vectors back into source

src/main/java/org/opensearch/knn/index/codec/derivedsource/PerFieldDerivedVectorInjectorFactory.java
```diff
@@ -0,0 +1,36 @@
...
+
+package org.opensearch.knn.index.codec.derivedsource;
+
+import org.apache.lucene.index.FieldInfo;
+import org.apache.lucene.index.SegmentReadState;
+
+/**
+ * Factory for creating {@link PerFieldDerivedVectorInjector} instances.
+ */
+public class PerFieldDerivedVectorInjectorFactory {
```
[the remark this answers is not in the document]
jmazanec15: will change

src/main/java/org/opensearch/knn/index/codec/KNN9120Codec/DerivedSourceStoredFieldsFormat.java
```diff
@@ -0,0 +1,80 @@
...
+import static org.opensearch.knn.common.KNNConstants.DERIVED_VECTOR_FIELD_ATTRIBUTE_KEY;
+import static org.opensearch.knn.common.KNNConstants.DERIVED_VECTOR_FIELD_ATTRIBUTE_TRUE_VALUE;
+
+@AllArgsConstructor
+public class DerivedSourceStoredFieldsFormat extends StoredFieldsFormat {
+
+    private final StoredFieldsFormat delegate;
+    private final DerivedSourceReadersSupplier derivedSourceReadersSupplier;
+    // IMPORTANT Do not rely on this for the reader, it will be null if SPI is used
+    private final MapperService mapperService;
```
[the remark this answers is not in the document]
jmazanec15: sure

build.gradle
```diff
@@ -390,6 +390,7 @@ integTest {
     systemProperty("https", is_https)
     systemProperty("user", user)
     systemProperty("password", password)
+    systemProperty("test.exhaustive", System.getProperty("test.exhaustive"))
```
jmazanec15, line 393: Added this flag in order to provide options for more exhaustive testing. In this PR, I put the index size checks wrapped in it to avoid flakiness. @navneet1v

src/main/java/org/opensearch/knn/index/KNNSettings.java
```diff
@@ -581,6 +593,14 @@ public static boolean isFaissAVX2Disabled() {
         }
     }
 
+    /**
+     * check this index enabled/disabled derived source
+     * @param settings Settings
+     */
+    public static boolean isKNNDerivedSourceEnabled(Settings settings) {
```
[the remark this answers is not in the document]
jmazanec15: Will take FAR

src/main/java/org/opensearch/knn/index/codec/KNN9120Codec/DerivedSourceStoredFieldsFormat.java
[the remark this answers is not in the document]
jmazanec15: Im going to take this check for another review

src/main/java/org/opensearch/knn/index/codec/KNN9120Codec/DerivedSourceStoredFieldsWriter.java
jmazanec15: Will add in another review

build.gradle
navneet1v: make sense. I think this is a good idea for now.

src/main/java/org/opensearch/knn/index/codec/KNN9120Codec/DerivedSourceStoredFieldsWriter.java
jmazanec15: Will add in follow up. Only other member is list of strings which are fields which will be on order of bytes to low kbs

src/test/java/org/opensearch/knn/integ/DerivedSourceIT.java
```diff
@@ -0,0 +1,1252 @@
...
+    }
+
+    @SneakyThrows
+    private void testMerging(List<IndexConfigContext> indexConfigContexts) {
+        IndexConfigContext derivedSourceEnabledContext = indexConfigContexts.get(0);
+        IndexConfigContext derivedSourceDisabledContext = indexConfigContexts.get(1);
+        String originalIndexNameDerivedSourceEnabled = derivedSourceEnabledContext.indexName;
+        String originalIndexNameDerivedSourceDisabled = derivedSourceDisabledContext.indexName;
+        forceMergeKnnIndex(originalIndexNameDerivedSourceEnabled, 10);
+        forceMergeKnnIndex(originalIndexNameDerivedSourceDisabled, 10);
```
navneet1v, line 880: should go here with 1 segment? with 10 segment and 50 docs I am not even sure if merges are happening in the backend.

navneet1v: nvm. This is already taken care later in the code.

src/main/java/org/opensearch/knn/index/mapper/LuceneFieldMapper.java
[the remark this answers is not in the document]
navneet1v: oh yeah my bad. For lucene we freeze the field early.

src/main/java/org/opensearch/knn/index/vectorvalues/KNNVectorValuesFactory.java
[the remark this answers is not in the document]
navneet1v: since the DV is used for older indices and this setting is never going to be enabled for new indices so we don't need this.

src/main/java/org/opensearch/knn/index/codec/derivedsource/NestedPerFieldDerivedVectorInjector.java
jmazanec15: This is not necessarily the vector field. This is for a field in one child before adding back the vector. We are trying to figure out which docId that child maps to. To do it, I get the first field on/after the offset that contains that field. 

So, this is basically field exists - https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/FieldExistsQuery.java

src/main/java/org/opensearch/knn/common/KNNConstants.java
```diff
@@ -162,4 +162,8 @@ public class KNNConstants {
 
     public static final String MODE_PARAMETER = "mode";
     public static final String COMPRESSION_LEVEL_PARAMETER = "compression_level";
+
+    public static final String DERIVED_VECTOR_FIELD_ATTRIBUTE_KEY = "knn-derived-source-enabled";
+    public static final String DERIVED_VECTOR_FIELD_ATTRIBUTE_TRUE_VALUE = "true";
+    public static final String DERIVED_VECTOR_FIELD_ATTRIBUTE_FALSE_VALUE = "false";
```
navneet1v, line 168: [nit-pick] this constant is not used.

src/main/java/org/opensearch/knn/index/codec/KNN9120Codec/DerivedSourceStoredFieldsReader.java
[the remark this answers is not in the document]
jmazanec15: I think covered in abvove comment

src/main/java/org/opensearch/knn/index/mapper/FlatVectorFieldMapper.java
```diff
@@ -73,13 +79,17 @@ private FlatVectorFieldMapper(
...
-            originalMappingParameters
+            originalMappingParameters,
+            isDerivedSourceEnabled
         );
         // setting it explicitly false here to ensure that when flatmapper is used Lucene based Vector field is not created.
         this.useLuceneBasedVectorField = false;
         this.perDimensionValidator = selectPerDimensionValidator(vectorDataType);
         this.fieldType = new FieldType(KNNVectorFieldMapper.Defaults.FIELD_TYPE);
         this.fieldType.setDocValuesType(DocValuesType.BINARY);
+        if (isDerivedSourceEnabled) {
```
VijayanB, line 90: Are we adding only if it is true? is that intentional? if so why do we have DERIVED_VECTOR_FIELD_ATTRIBUTE_FALSE_VALUE ?

src/main/java/org/opensearch/knn/index/vectorvalues/KNNVectorValuesFactory.java
```diff
@@ -72,6 +74,37 @@ public static <T> KNNVectorValues<T> getVectorValues(final FieldInfo fieldInfo,
...
+                docIdSetIterator = knnVectorsReader.getByteVectorValues(fieldInfo.getName());
+            } else if (fieldInfo.getVectorEncoding() == VectorEncoding.FLOAT32) {
+                docIdSetIterator = knnVectorsReader.getFloatVectorValues(fieldInfo.getName());
+            } else {
+                throw new IllegalArgumentException("Invalid Vector encoding provided, hence cannot return VectorValues");
+            }
+        } else if (docValuesProducer != null) {
+            docIdSetIterator = docValuesProducer.getBinary(fieldInfo);
+        } else {
+            throw new IllegalArgumentException("Field does not have vector values and DocValues");
```
VijayanB, line 102: nit: Field should have either vector values or DocValues

src/main/java/org/opensearch/knn/index/mapper/FlatVectorFieldMapper.java
jmazanec15: Good point. Im going to remove FALSE in another PR.

src/main/java/org/opensearch/knn/index/vectorvalues/KNNVectorValuesFactory.java
jmazanec15: Will update in another PR

src/main/java/org/opensearch/knn/common/KNNConstants.java
jmazanec15: @VijayanB mentioned this. Will remove in another review

navneet1v (approved):

shatejas:
Looks good overall

shatejas (approved):

src/main/java/org/opensearch/knn/index/codec/derivedsource/NestedPerFieldDerivedVectorInjector.java
```diff
@@ -0,0 +1,276 @@
...
+        int lowestDocIdForFieldWithParentAsOffset = getLowestDocIdForField(childFieldInfo.name, parentDocId);
+        if (lowestDocIdForFieldWithParentAsOffset == parentDocId) {
+            injectObject(parentDocId, sourceAsMap);
+            return;
+        }
+
+        // Setup the iterator. Return if no parent
+        String childFieldName = ParentChildHelper.getChildField(childFieldInfo.name);
+        String parentFieldName = ParentChildHelper.getParentField(childFieldInfo.name);
+        if (parentFieldName == null) {
```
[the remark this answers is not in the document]
jmazanec15: being defensive

PR #2472, opened 2025-01-30
Switch away from getMergeInstance for derived source

jmazanec15:
### Description
I was running the DerivedSourceIts for the derived source feature in a multi-node setting with replicas and was noticing that getMergeInstance was being called outside of merge when reindex was happening. I couldnt figure out the stack trace, but decided to revert back to wrapping the readers during merge. This fixed it.

It might be getting called in the SequentialStoredFieldsReader in this hack: 
* https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/search/fetch/FetchPhase.java#L154-L157
* https://github.com/opensearch-project/OpenSearch/blob/2847695ad1fcc04c88b96c8bab0bfdf694fa05dc/server/src/main/java/org/opensearch/common/lucene/index/SequentialStoredFieldsLeafReader.java#L71

### Related Issues

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/codec/KNN9120Codec/DerivedSourceStoredFieldsWriter.java
+++ b/src/main/java/org/opensearch/knn/index/codec/KNN9120Codec/DerivedSourceStoredFieldsWriter.java
@@ -65,6 +65,10 @@ public class DerivedSourceStoredFieldsWriter extends StoredFieldsWriter {
 
     @Override
     public int merge(MergeState mergeState) throws IOException {
+        // We have to wrap these here to avoid storing the vectors during merge
+        for (int i = 0; i < mergeState.storedFieldsReaders.length; i++) {
+            mergeState.storedFieldsReaders[i] = DerivedSourceStoredFieldsReader.wrapForMerge(mergeState.storedFieldsReaders[i]);
+        }
         return delegate.merge(mergeState);
     }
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/opensearch/knn/index/codec/KNN9120Codec/DerivedSourceStoredFieldsWriter.java
```diff
@@ -65,6 +65,10 @@ public void writeField(FieldInfo info, DataInput value, int length) throws IOExc
 
     @Override
     public int merge(MergeState mergeState) throws IOException {
+        // We have to wrap these here to avoid storing the vectors during merge
+        for (int i = 0; i < mergeState.storedFieldsReaders.length; i++) {
```
vamshin, line 69: Do we need null check on mergeState?

shatejas (approved):
Looks good, I don't see unit tests. Please iterate on it in separate PR

I do think the opensearch core shouldn't be using it in a hacky way. It's violating the interface contract. It would be good to raise this issue so the pattern is not repeated at other places.

Approving for now

src/main/java/org/opensearch/knn/index/codec/KNN9120Codec/DerivedSourceStoredFieldsWriter.java
navneet1v: the iteration is based on the segments, we don't need to do a null check here.

navneet1v (approved):

Issue #1581, opened 2024-03-27
Investigate concurrent merges from Lucene for HNSW


jmazanec15:
## Description

In Lucene, it is possible to create an HNSW graph merger that will use multiple threads (https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/util/hnsw/ConcurrentHnswMerger.java) to build the graph. This is similar to our current index thread qty setting. We should investigate adding this into OpenSearch.

In order to do so, for the format, we need to pass in an ExecutorService to do the merge, which we do not do now (https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsFormat.java#L167-L169).

navneet1v:
@jmazanec15 and @vamshin I think this is a simple issue which we should get done, not sure why have not done by now. I am willing to pick this up.

navneet1v:
The changes for this feature is merged and based on the nighly run I can see the performance of the Force merge has improved. Nightly runs uses index_thread_qty as 2. and now with this change we can see the force merge time has reduced by 50%.

### 10M 768D dataset

<img width="712" alt="Image" src="https://github.com/user-attachments/assets/ff5a8f9c-a76f-405c-b497-e4aa3a5f83e9" />


### 1M 768D Dataset

<img width="716" alt="Image" src="https://github.com/user-attachments/assets/fba6a6bb-ebc7-4519-b1ea-81b4b86ee426" />

navneet1v:
Doc PR: https://github.com/opensearch-project/documentation-website/pull/9152

PR #2480, opened 2025-01-31
Enabled concurrent graph creation for Lucene engine with index thread qty settings

navneet1v:
### Description
Enabled concurrent graph creation for Lucene engine with index thread qty settings

### Related Issues
Resolves https://github.com/opensearch-project/k-NN/issues/1581

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/codec/KNN9120Codec/KNN9120PerFieldKnnVectorsFormat.java
+++ b/src/main/java/org/opensearch/knn/index/codec/KNN9120Codec/KNN9120PerFieldKnnVectorsFormat.java
@@ -7,18 +7,22 @@ package org.opensearch.knn.index.codec.KNN9120Codec;
 
 import org.apache.lucene.codecs.lucene99.Lucene99HnswScalarQuantizedVectorsFormat;
 import org.apache.lucene.codecs.lucene99.Lucene99HnswVectorsFormat;
+import org.opensearch.common.collect.Tuple;
 import org.opensearch.index.mapper.MapperService;
+import org.opensearch.knn.index.KNNSettings;
 import org.opensearch.knn.index.SpaceType;
 import org.opensearch.knn.index.codec.BasePerFieldKnnVectorsFormat;
 import org.opensearch.knn.index.engine.KNNEngine;
 
 import java.util.Optional;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 
 /**
  * Class provides per field format implementation for Lucene Knn vector type
  */
 public class KNN9120PerFieldKnnVectorsFormat extends BasePerFieldKnnVectorsFormat {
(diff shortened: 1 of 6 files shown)


src/main/java/org/opensearch/knn/index/codec/KNN9120Codec/KNN9120PerFieldKnnVectorsFormat.java
```diff
@@ -41,23 +43,31 @@ public KNN9120PerFieldKnnVectorsFormat(final Optional<MapperService> mapperServi
...
+    @Override
     public int getMaxDimensions(String fieldName) {
         return KNNEngine.getMaxDimensionByEngine(KNNEngine.LUCENE);
     }
+
+    private static int getMergeThreadCount() {
+        return KNNSettings.getIndexThreadQty();
+    }
+
+    private static ExecutorService getMergeExecutorService() {
```
navneet1v, line 70: these functions needs to be static as they are getting called from the super function call of the constructor.

src/main/java/org/opensearch/knn/index/codec/KNN9120Codec/KNN9120PerFieldKnnVectorsFormat.java
```diff
@@ -41,23 +43,31 @@ public KNN9120PerFieldKnnVectorsFormat(final Optional<MapperService> mapperServi
...
+        return Executors.newFixedThreadPool(getMergeThreadCount());
```
kotwanikunal, line 71: This can be done later but I think it would make sense to initialize all our executors through a single provider.
We have no control over this reference from other parts of the code with local init.

kotwanikunal (approved):

Vikasht34 (approved):

src/main/java/org/opensearch/knn/index/codec/KNN9120Codec/KNN9120PerFieldKnnVectorsFormat.java
navneet1v: Since this executor is more of internal and will be present per segment, I was trying to avoid keeping a global reference for this. Another thing is it closely mimics Faiss behavior where we use OMP threads to parallelize the HNSW builds.

naveentatikonda (approved):

shatejas (approved):

navneet1v:
Merging the change in 2.x branch since 2.19 is cut off. and I will work with release owner to see if we can put this change in 2.19

navneet1v:
The changes for this feature is merged and based on the nighly run I can see the performance of the Force merge has improved. Nightly runs uses index_thread_qty as 2. and now with this change we can see the force merge time has reduced by 50%.

### 10M 768D dataset

Image


### 1M 768D Dataset

Image

navneet1v:
Doc PR: https://github.com/opensearch-project/documentation-website/pull/9152

PR #2495, opened 2025-02-06
Fix warn logs for engine deprecation

kotwanikunal:
### Description
- Fixes log flooding for engine deprecation notice

### Related Issues
N/A

### Check List
- [ ] ~New functionality includes testing.~
- [ ] ~New functionality has been documented.~

### DIFF
--- a/src/main/java/org/opensearch/knn/index/engine/EngineResolver.java
+++ b/src/main/java/org/opensearch/knn/index/engine/EngineResolver.java
@@ -5,14 +5,19 @@
 
 package org.opensearch.knn.index.engine;
 
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 import org.opensearch.knn.index.mapper.CompressionLevel;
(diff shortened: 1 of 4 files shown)


src/main/java/org/opensearch/knn/index/engine/EngineResolver.java
```diff
@@ -30,33 +35,39 @@ public KNNEngine resolveEngine(
         KNNMethodContext knnMethodContext,
         boolean requiresTraining
     ) {
+        KNNEngine knnEngine = KNNEngine.FAISS;
```
jmazanec15, line 38: Lets keep return statements as is and just add a function before return that logs if deprecated. Im worried that we might be altering resolution logic with this

kotwanikunal: I wanted to avoid that pattern, but sure. Let me work that through.

navneet1v (approved):

Issue #2386, opened 2025-01-14
[FEATURE] Integrate Jvector engine as another vector engine of choice


sam-herman:
**Is your feature request related to a problem?**
Currently k-NN plugin supports 3 engines, Nmslib, Faiss and Lucene.
In this change I would like to integrate JVector as another engine of choice.
There are a number of reasons to doing so:
1. _**Disk ANN**_ - JVector is capable to perform search without loading the entire index into RAM.
2. _**Compatibility**_ - JVector is compatible with Cassandra. Which allows to more easily transfer vector encoded data from Cassandra to OpenSearch and vice versa.

**What solution would you like?**
Introduce JVector into K-NN plugin as another supported engine.

**What alternatives have you considered?**
NA

jmazanec15:
Thanks @sam-herman. Jvector looks very cool and would be a good addition. I think to add this, we would need clean/formalize engine abstractions/interfaces to make vector engine's truly extendable. We have done this to some degree, but definitely need to go further.

jmazanec15:
There arent any formal plans, but a trail of issues:
1. https://github.com/opensearch-project/k-NN/issues/1328
2. https://github.com/opensearch-project/k-NN/issues/1540
3. https://github.com/opensearch-project/k-NN/issues/1467
4. https://github.com/opensearch-project/k-NN/issues/1347

I think long term goal is to provide interface for new engines and clean separation between core k-NN plugin and different engines. We started doing that in 2.17 with https://github.com/opensearch-project/k-NN/tree/main/src/main/java/org/opensearch/knn/index/engine. Theres still work todo, but idea is to get away from having branches on engines.

sam-herman:
Thanks for sharing those.
For JVector the immediate straightforward path I see is to add it as an engine that extends the `JVMLibrary` interface same as Lucene KnnEngine does. Currently I don't see any substantial difference between the two in terms of plugin interface requirements, therefore my suggestion is to add JVector with the existing abstractions and this should be orthogonal to future refactoring.

sam-herman:
@navneet1v @jmazanec15 I added more bullet points in the description to the "added value" category. I'm hoping to add the output of a few benchmarks to support those pretty soon. But hopefully for the time being it can provide some qualitative measure.
Prior to adding the benchmarks I will provide a benchmark plan for your review, however if you have any suggestions in the time being that you think can help let me know.

navneet1v:
[...]

PR #2505, opened 2025-02-07
2386/integrate jvector knn engine

sam-herman:
### Description
This change integrates jVector as a JVM based vector format into KNN plugin.
jVector introduces DiskANN search that is purely implemented in Java and doesn't require native dependencies.

### Related Issues
Resolves https://github.com/opensearch-project/k-NN/issues/2386

### Check List
- [x ] New functionality includes testing.
- [ x] New functionality has been documented.
- [x ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
- [x ] Commits are signed per the DCO using `--signoff`.
- [x ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).


navneet1v:
@sam-herman I think based on our discussion we were thinking to move the JVector not part of the original build and code but to be vended out as a separate engine whose code will be residing in a separate module. What is the reason of choosing this approach?

sam-herman:
> @sam-herman I think based on our discussion we were thinking to move the JVector not part of the original build and code but to be vended out as a separate engine whose code will be residing in a separate module and also will be a custom build process.
> 
> The main reason for choosing this approach was:
> 
> 1. Adding engine directly in k-NN plugin is not scaleable.
> 
> What is the reason of choosing this approach? Are putting this as a draft PR to provide a context on interfaces we need to change?

Hi @navneet1v totally agree with the scalability issues mentioned, if we keep adding random libraries it will get bloated. I think we also mentioned that we have two options:
1. Provide benchmarks to show that it has some added value.
2. With the absence of benchmark to make a point of added value, we'll make it part of a "extras" modules folder.

I am working on getting those benchmarks out next week, so how about we treat it as a draft for now to review the correctness of the integration?
Once the benchmarks are out we can decide to refactor to different module or keep as default?
Moreover, I made some sample files that I wanted the reviewer to consider as suggestions for abstraction that would make the project more concise. So I think it's fair to say we can treat it as a draft at the moment to review the approach.

navneet1v:
> So I think it's fair to say we can treat it as a draft at the moment to review the approach.

Thanks for the confirmation.

build.gradle
```diff
@@ -331,7 +335,7 @@ task windowsPatches(type:Exec) {
 task cmakeJniLib(type:Exec) {
     workingDir 'jni'
     def args = []
-    args.add("cmake")
+    args.add("/opt/homebrew/bin/cmake")
```
sam-herman, line 338: I will fix this one

demo.sh
sam-herman, line 1: Will move this to resources or as part of a `README.md` for jVector demonstration

src/main/java/org/opensearch/knn/index/codec/WrapperCodecForKNNPlugin.java
sam-herman, line 1: This file is a not used right now, but it illustrates a suggestion for a more concise abstraction that will also allow to select compound format based on engine.
It also more consistent with how suppliers and constructors are passed in OpenSearch core and avoids back and forth referenced between class to enum and back by being self contained.

build.gradle
```diff
@@ -316,6 +316,10 @@ dependencies {
     }
     testFixturesImplementation "org.opensearch:common-utils:${version}"
     implementation 'com.github.oshi:oshi-core:6.4.13'
+
+    implementation 'io.github.jbellis:jvector:4.0.0-beta.2-SNAPSHOT'
+    implementation 'org.agrona:agrona:1.20.0'
```
jmazanec15, line 321: Where is this used? I didnt see in the codec

src/main/java/org/opensearch/knn/index/codec/jvector/JVectorCompoundFormat.java
```diff
@@ -0,0 +1,182 @@
...
+import java.io.IOException;
+import java.nio.file.Path;
+import java.util.Collections;
+import java.util.Map;
+import java.util.Set;
+
+// TODO: This needs to be moved under the same package name as the Lucene internal package name for {@link Lucene90CompoundReader}
+// this way the internal package constants can be accessed directly and we can avoid duplicating them.
+@Log4j2
+public class JVectorCompoundFormat extends CompoundFormat {
```
jmazanec15, line 28: Can you remind me if this is still needed after our discuss? I remember that @navneet1v mentioned on slack that we dont need this. It seems like this is the only reason we'd need to create a custom codec for jvector as opposed to just the knnvectorsformat. So, Id like to get rid of it if possible.

jmazanec15 (commented):
Do we need updates to all of the gradle wrapper?

build.gradle
tjake: should be runtimeOnly

sam-herman: It's a transient dependency of jvector, I had to include it because it doesn't resolve automatically. Can make it a runtimeOnly because it's not explicitly used.

src/main/java/org/opensearch/knn/index/codec/jvector/JVectorCompoundFormat.java
sam-herman: It can potentially be removed after `RandomAccessReader` for jVector is delegated to `IndexInput`.

Issue #2223, opened 2024-10-21
[Enhancement] Having a reference to the state in KNN990 vector reader, then construct cache keys to invalidate in close method.


0ctopus13prime:
## Description
In NativeEngine990KnnVectorsReader's constructor, we construct cache keys for invalidating with the given segment reader state.
But this is unnecessary, we can always construct with field infos and segment info when closing.

```
    private final List<String> cacheKeys;

    public NativeEngines990KnnVectorsReader(final SegmentReadState state, final FlatVectorsReader flatVectorsReader) {
        this.flatVectorsReader = flatVectorsReader;
        this.segmentReadState = state;
        this.cacheKeys = getVectorCacheKeysFromSegmentReaderState(state); <---------- This!
        loadCacheKeyMap();
    }

    @Override
    public void close() throws IOException {
        // Clean up allocated vector indices resources from cache.
        final NativeMemoryCacheManager nativeMemoryCacheManager = NativeMemoryCacheManager.getInstance();
        cacheKeys.forEach(nativeMemoryCacheManager::invalidate); <------- This!
        ...
```

Not for sure, but once I tried to change it to have a reference in constructor and build a cache key in `close` method, I bumped into this exceptions during testing.

```
Suppressed: java.lang.AssertionError: should not be called by a cluster state applier. reason [the applied cluster state is not yet available]
»               at org.opensearch.cluster.service.ClusterApplierService.assertNotCalledFromClusterStateApplier(ClusterApplierService.java:446) ~[opensearch-3.0.0-SNAPSHOT.jar:3.0.0-SNAPSHOT]
»               at org.opensearch.cluster.service.ClusterApplierService.state(ClusterApplierService.java:230) ~[opensearch-3.0.0-SNAPSHOT.jar:3.0.0-SNAPSHOT]
»               at org.opensearch.cluster.service.ClusterService.state(ClusterService.java:183) ~[opensearch-3.0.0-SNAPSHOT.jar:3.0.0-SNAPSHOT]
»               at org.opensearch.knn.indices.ModelDao$OpenSearchKNNModelDao.getMetadata(ModelDao.java:457) ~[?:?]
»               at org.opensearch.knn.indices.ModelUtil.getModelMetadata(ModelUtil.java:52) ~[?:?]
»               at org.opensearch.knn.common.FieldInfoExtractor.extractKNNEngine(FieldInfoExtractor.java:41) ~[?:?]
[...]

PR #2509, opened 2025-02-07
Introduce node level circuit breaker settings for k-NN

### LINKED ISSUE #2263: [FEATURE] Introduce node level circuit breaker settings for k-NN
The existing k-NN plugin uses a cluster-level circuit breaker to prevent excessive memory consumption. While effective, this approach may not be optimal for all cluster configurations, especially in heterogeneous environments where nodes have varying capacities. It lacks fine-grained control over memory usage on individual nodes.

Implement node-level circuit breakers for the k-NN plugin, allowing memory limits to be set and enforced on a per-node basis.
The enhancement will refactor the circuit breaker system in OpenSearch to support differentiated limits based on node attributes. 

The approach involves defining circuit breaker limits at the cluster level, but with distinct values for different node types. Nodes would be categorized using attributes such as "node.attr.type" set to values like "big" or "small" in their opensearch.yml configuration.
For example - 
```
PUT _cluster/settings
{
  "persistent": {
    "plugins.knn.circuit_breaker.limit.big": "70%",
    "plugins.knn.circuit_breaker.limit.small": "40%"
  }
}
```
    

The k-NN plugin would then apply the appropriate limit based on each node's attributes. This method leverages existing OpenSearch configuration mechanisms, allows for centralized management, and provides the necessary flexibility for mixed-capability clusters. It maintains backwards compatibility by falling back to a default or existing cluster-wide setting for nodes without specified attributes.

Implementation would require modifying the k-NN plugin to read node attributes, select the corresponding circuit breaker limit, and apply it in the circuit breaker logic. This solution offers a balance of granular control and ease of management, tailoring resource allocation to node capabilities while keeping configuration centralized and straightforward.

- Moving to per node configuration within opensearch.yml as overrides

markwu-sde:
### Description

KNN plugin currently uses a cluster-wide circuit breaker. This doesn't work as well as it could when nodes have different memory capacities.

### Solution
Added node-specific circuit breaker limits using node attributes. This opens up flexibility for heterogenous circuit breaker limits.

### Usage

1. Set node attribute in `opensearch.yml`:
```yaml
node.attr.knn_cb_tier: "integ"
```

2. Configure limit for that tier:
```json
PUT /_cluster/settings
{
  "persistent": {
    "knn.memory.circuit_breaker.limit.integ": "75%"
  }
}
```

### Implementation
- Uses OpenSearch's `groupSetting` for dynamic limit configurations
- Cache initially uses cluster-level default
- Updates to node-specific limit when node attributes are available
- Falls back to cluster default if no node-specific limit exists

### Testing

Modified KNNCircuitBreakerIT integration tests to include node-level CB.

Used OSB benchmarking to run a modified low load test with/without the node level circuit breaker. 

- Added `node.attr.knn_cb_tier = 'integ'` to build.gradle
- Ran 5-query search-only vector benchmark after indexing 
- Verified memory usage and circuit breaker behavior

Initial state (no limits set):
```json
{
  "graph_memory_usage_percentage": 1.35,
  "graph_memory_usage": 110056
}
```

After setting node limit to 500000kb:
```json

curl -XPUT "http://localhost:9200/_cluster/settings" -H 'Content-Type: application/json' -d'
{
  "persistent" : {
    "knn.memory.circuit_breaker.limit.integ" : "500000kb"
  }
}
'

Results:

{
  "graph_memory_usage_percentage": 22.01,
  "graph_memory_usage": 110056,
  "circuit_breaker_triggered": false
}
```

Setting cluster limit to 5kb didn't affect node with specific limit which confirms proper override behavior.

### Related Issues
Resolves #2263 

### Check List


src/main/java/org/opensearch/knn/index/KNNSettings.java
```diff
@@ -564,8 +590,121 @@ public static boolean isCircuitBreakerTriggered() {
...
+     * this will return the limit configured for the "high" tier.
+     *
+     * @return String representing the circuit breaker limit for this node's tier,
+     *         null if either:
+     *         - Node has no knn_cb_tier attribute
+     *         - No configuration exists for the node's tier
+     */
+    private String getKnnCircuitBreakerLimitForNode(Settings settings) {
+        // Get this node's circuit breaker tier attribute
+        String tierAttribute = clusterService.localNode().getAttributes().get(KNN_CIRCUIT_BREAKER_TIER);
```
markwu-sde, line 652: Running into an issue very similar to #2223 for the build failures

```
java.lang.AssertionError: should not be called by a cluster state applier. reason [the applied cluster state is not yet available]
»       at org.opensearch.cluster.service.ClusterApplierService.assertNotCalledFromClusterStateApplier(ClusterApplierService.java:443)
»       at org.opensearch.cluster.service.ClusterApplierService.state(ClusterApplierService.java:229)
»       at org.opensearch.cluster.service.ClusterService.state(ClusterService.java:182)
»       at org.opensearch.cluster.service.ClusterService.localNode(ClusterService.java:166)
»       at org.opensearch.knn.index.KNNSettings.getKnnCircuitBreakerLimitForNode(KNNSettings.java:652)

```

Since node attributes are not dynamically configurable currently we can cache the value on node initialization when we refresh the cache. This would prevent the update thread from trying to access the attributes via the cluster state.

src/main/java/org/opensearch/knn/index/KNNCircuitBreaker.java
```diff
@@ -106,5 +109,26 @@ public void initialize(ThreadPool threadPool, ClusterService clusterService, Cli
             }
         };
         this.threadPool.scheduleWithFixedDelay(runnable, TimeValue.timeValueSeconds(CB_TIME_INTERVAL), ThreadPool.Names.GENERIC);
+
+        // Update when node is fully joined
+        clusterService.addLifecycleListener(new LifecycleListener() {
```
markwu-sde, line 114: Node attributes aren't available during the time the cache manager computes its size on initialization. With the additional dependency on node attributes to the circuit breaker we may need to check if the circuit breaker needs to be updated after the node has bootstrapped. 

I only see 1 example of a listener being attached in [here](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/plugin/KNNPlugin.java#L213) but attached the listener on KnnCircuitBreaker initialization due needing to refresh the cache to recompute the size and caching the attribute value. If there's any feedback on this that'd be appreciated.

src/main/java/org/opensearch/knn/index/KNNSettings.java
```diff
@@ -385,17 +388,42 @@ public class KNNSettings {
...
+             */
             put(
-                KNN_MEMORY_CIRCUIT_BREAKER_LIMIT,
+                KNN_MEMORY_CIRCUIT_BREAKER_LIMIT_PREFIX,
+                Setting.groupSetting(KNNSettings.KNN_MEMORY_CIRCUIT_BREAKER_LIMIT_PREFIX, settings -> {
+                    settings.keySet()
+                        .forEach(
+                            (limit) -> parseknnMemoryCircuitBreakerValue(
+                                settings.get(limit),
+                                KNNSettings.KNN_MEMORY_CIRCUIT_BREAKER_CLUSTER_LIMIT
```
markwu-sde, line 404: Ran into an issue where not having the validation being performed in `parseKnnMemoryCircuitBreakerValue` specifically here would result in dropping the cluster on a uncaught exception if the same method was hit in line 633 with a invalid knn cb value. There may be some exception handling happening earlier in the stack that allows an exception to be caught gracefully here doesn't exist later on but to maintain the same behavior we validate the given values before trying to update the cache.

src/main/java/org/opensearch/knn/index/KNNCircuitBreaker.java
kotwanikunal: I like this approach. I think it makes sense.

kotwanikunal (approved):
LGTM.

kotwanikunal:
@jmazanec15 WDYT?

src/main/java/org/opensearch/knn/index/KNNCircuitBreaker.java
```diff
@@ -106,5 +109,26 @@ public void initialize(ThreadPool threadPool, ClusterService clusterService, Cli
...
+            public void afterStart() {
+                // Attempt to fetch a cb tier from node attributes and cache the result.
+                // Get this node's circuit breaker tier attribute
+                Optional<String> tierAttribute = Optional.ofNullable(
+                    clusterService.localNode().getAttributes().get(KNN_CIRCUIT_BREAKER_TIER)
+                );
+                if (tierAttribute.isPresent()) {
+                    // Only rebuild the cache if the attribute was present
+                    logger.info(
+                        "[KNN] Node specific circuit breaker " + tierAttribute.get() + " classification found. Rebuilding the cache."
```
naveentatikonda, line 125:
```suggestion
                        "[KNN] Node specific circuit breaker {} classification found. Rebuilding the cache.", tierAttribute.get());
```

ryanbogan:
I believe we have an item in the KNNStats API that indicates when the circuit breaker is triggered.  Do we need to modify this at all to accommodate the new node-level circuit breaker and maybe give more details about which circuit breaker is triggered?

markwu-sde:
> I believe we have an item in the KNNStats API that indicates when the circuit breaker is triggered. Do we need to modify this at all to accommodate the new node-level circuit breaker and maybe give more details about which circuit breaker is triggered?

The KNNStats API is reliant on a flag set [here](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/memory/NativeMemoryCacheManager.java#L455) in the cache whenever an eviction happens based off of a capacity constraint. There can only be 1 circuit breaker value active at any given moment and the circuit breaker flag will continue to operate as expected here when the cache gets updated with a node level limit and so I don't think we're breaking any of our existing contracts here.

The other pending item revolves around allowing more visibility into letting customers know about their current circuit breaker limit and usage. While we don't have an method to extract the currently used circuit breaker limit we can use the existing memory metrics to extrapolate if necessary such as graph memory usage percentage which is already available. However that is something we can consider maybe in the future to add a circuit breaker limit value API to give customers more flexibility

src/main/java/org/opensearch/knn/index/KNNCircuitBreaker.java
```diff
@@ -106,5 +109,26 @@ public void initialize(ThreadPool threadPool, ClusterService clusterService, Cli
             }
         };
         this.threadPool.scheduleWithFixedDelay(runnable, TimeValue.timeValueSeconds(CB_TIME_INTERVAL), ThreadPool.Names.GENERIC);
+
+        // Update when node is fully joined
+        clusterService.addLifecycleListener(new LifecycleListener() {
+            @Override
+            public void afterStart() {
```
Vikasht34, line 116: Do we really node attribute (knn_cb_tier) on node start 
Instead, this should fetch attributes dynamically whenever needed.

src/main/java/org/opensearch/knn/index/memory/NativeMemoryCacheManager.java
```diff
@@ -127,7 +136,7 @@ public synchronized void rebuildCache() {
         rebuildCache(
```
Vikasht34, line 136: The entire cache is rebuilt every time a node joins, even if the circuit breaker limit remains unchanged. Can we lazily do this that updates only if the limit actually changes.

src/main/java/org/opensearch/knn/index/KNNSettings.java
```diff
@@ -564,8 +592,106 @@ public static boolean isCircuitBreakerTriggered() {
...
+     * Evaluates both node-specific and cluster-level updates in the updated settings,
+     * maintaining proper precedence:
+     * 1. Node-tier specific limit from updates (if available)
+     * 2. Appropriate fallback value based on node's current configuration
+     *
+     * @param updatedCbLimits Settings object containing pending circuit breaker updates
+     * @return ByteSizeValue representing the new circuit breaker limit to apply,
+     *         or null if no applicable updates found
+     */
+    private ByteSizeValue getUpdatedCircuitBreakerLimit(Settings updatedCbLimits) {
```
Vikasht34, line 661: updates circuit breaker settings every time the cluster settings change, should we only update when the node-specific tier is modified.

src/main/java/org/opensearch/knn/index/KNNCircuitBreaker.java
markwu-sde: Initially I had the attribute fetched dynamically to compute the circuit breaker value whenever we need it. The circuit breaker value is really only used by the CacheManager and I had [mentioned](https://github.com/opensearch-project/k-NN/pull/2509#discussion_r1953410790) earlier that on cache initialization we don't have access to node attributes yet and so there were 2 thoughts that I had with this approach -

1. Not having fetching this attribute on node start means that the cache will continue to maintain the cluster level size until the cache is rebuilt again with a updated value from the attribute. Not doing this on node start would mean this process would have to be happening during some sort of dynamic update of the cache (through a settings update or something else) in which we'll have to pass in the cache and have it rebuild. My idea here is that we set the correct cache size as soon as the node starts to ensure that there's no discrepancy between what users are expecting and what the cache size actually is. Since the circuit breaker here already interacts with the NativeCacheManager I thought it'd be the logical place to insert

2. Another issue is related to the comment I [posted](https://github.com/opensearch-project/k-NN/pull/2509#discussion_r1951918237). I want to avoid having to reference cluster state configurations during different parts of the lifecycle and just have the value persist through the entirety. 

Operating under the assumption that node attributes are considered to be statically defined as of now fetching the value on node start seems to be easier here but also open to suggestions on how we can go about doing this dynamically.

src/main/java/org/opensearch/knn/index/memory/NativeMemoryCacheManager.java
markwu-sde: From what I understand here the cache gets rebuilt in 3 distinct places within the code:

1. On [initialization](https://github.com/opensearch-project/k-NN/blob/45ecb5bf360b36c702e4562e73f5413bac5d60d6/src/main/java/org/opensearch/knn/index/memory/NativeMemoryCacheManager.java#L85) 
2. On a dynamic settings [update](https://github.com/opensearch-project/k-NN/blob/45ecb5bf360b36c702e4562e73f5413bac5d60d6/src/main/java/org/opensearch/knn/index/KNNSettings.java#L449)
3. On indexModule

For option 1, when a node joins I assumed this to be a part of the initialization. In that case my thought was that we need to recompute the value of the circuit breaker limit since each the attributes belonging to that node may correspond to a different circuit breaker tier. 

As for option 2, existing behavior within k-nn dictates that whenever a dynamic setting gets updated the cache gets rebuilt. This applies to all settings that are dynamically configured. The configuration system within k-nn will throw away no-ops in the updated settings and so each call to this option will guarantee to have one of the dynamically configurations to have changed. A good callout here is that the cache is being rebuilt every time one of these settings have changed (regardless of if the limit has changed) but I see this as modifying existing k-nn behavior which I'd like to minimize outside the scope of my PR. The only change I am making here is modifying how we compute the weight of the cache in the event where it needs to be rebuilt. 

The behavior with option 3 doesn't have a relation with our change I believe.

src/main/java/org/opensearch/knn/index/KNNSettings.java
markwu-sde: I believe I had addressed this in my comment [here](https://github.com/opensearch-project/k-NN/pull/2509/files/e5f652b3039473ea81cdf91b94d42dc2398e65ad#r1958853616). The cache gets rebuilt every time the a dynamic cluster setting changes so I don't see an option to lazily update the based off what setting got updated. 

If we look at the current code the weight is already getting [recomputed](https://github.com/opensearch-project/k-NN/blob/45ecb5bf360b36c702e4562e73f5413bac5d60d6/src/main/java/org/opensearch/knn/index/KNNSettings.java#L435) every time any dynamic setting gets refreshed. What I'm doing here is simplifying that logic to recompute the max weight to take into account node level attributes.

jmazanec15 (approved):
LGTM thanks @markwu-sde  - make sure to update docs

Vikasht34 (approved):
Thanks for seeting up call and go through this PR. Looks good to me.

markwu-sde:
> LGTM thanks @markwu-sde - make sure to update docs

https://github.com/opensearch-project/documentation-website/issues/9307

PR for documentation

Issue #2465, opened 2025-01-29
Remote Vector Index Build Component -- Repository Integration Low Level Design


jed326:
## Overview

This is the low level design follow-up to https://github.com/opensearch-project/k-NN/issues/2392. Specifically, the following are covered:
1. Control knobs for configuring and enabling the feature
2. Metrics
3. Detailed analysis of repository upload/download

## User Experience

First, we define the user experience for how a user can configure and control the remote vector build feature.

### Vector Repository Configuration

We will expose a cluster settings for users to indicate the name of the (registered) repository they would like to use as the vector repository. A user must register the repository on their own.

```
knn.remote_build.vector_repo (String)
```

It is difficult to get a reference to the RepositoryService from KNNSettings, so for now we will not validate that the repository specified here is registered. We will leave this for https://github.com/opensearch-project/k-NN/issues/2464.

From https://github.com/opensearch-project/k-NN/issues/2392 we outlined that we will use the `vectors` path outside of the indices path to avoid collision with any snapshots, so a user will still be able to use this repository for snapshots if they wish (although we will recommend them not to).

There are many knobs exposed by the specific repository implementations that can be used for performance tuning. For example, repository-s3 allows configuring the chunk size, buffer size, etc (https://opensearch.org/docs/latest/api-reference/snapshots/create-repository/#s3-repository). We will also perform performance benchmarks and performance tuning suggestions in https://github.com/opensearch-project/k-NN/issues/2464.

## Feature Controls

The various settings with which a user can enable/disable this feature are listed below.

1. Feature flag in KNNFeatureFlag. This will be internal only and removed before GA.
```
knn.feature.remote_build.enabled (boolean)
```

2. Index Setting to enable/disable the feature. We will explore a field level setting as a part of https://github.com/opensearch-project/k-NN/issues/2464.

```
index.knn.remote_build.enabled (boolean)
```

3. Index threshold settings to control the minimum vector data size required to use the remote vector build feature. For example, a segment with 1k 5 dimension vectors would be roughly the same size as a segment with 5 1k dimension vectors.

```
index.knn.remote_build.min_segment_size (int)
```

## Visibility

[...]

Issue #2391, opened 2025-01-14
[Meta] Remote Vector Index Build Component in OpenSearch Vector Engine


jed326:
## Description
This is a meta issue to track the work needed to design and implement the component described by the previous RFC:
- https://github.com/opensearch-project/k-NN/issues/2294

## Background

Following content is copied from previous RFCs for reference
- https://github.com/opensearch-project/k-NN/issues/2293
- https://github.com/opensearch-project/k-NN/issues/2294

### Components Definition

1. Vector Index Build Component: This is new component that is being proposed as part of this design. It  will be responsible for building a vector index for a segment and then notifying Vector Engine that index is ready to download. This index build component can have both GPU and CPU machines or any customer hardware to build the vector index.
2. Opensearch Vector Engine(k-NN plugin): This nothing but k-NN plugin which is responsible for providing vector related capabilities in Opensearch.
3. Object Store/Intermediate Vector Store: This an intermediate storage component which will temporarily store the vectors and index for different components. The store will ensure that it has right deletion polices to remove these vectors. This is not the same store where we store the segments for Remote store feature in Opensearch. This will be a separate store.

### Assumptions

Below are some of the assumptions taken while designing the integration of Vector Engine with remote Index Build Service.

1. This design assumes that there is an IndexBuildService hosted at a endpoint where Vector Engine can submit the request of create Index.
2. Vector engine has all the details on how to connect to object store to stream vectors and download index.
3. Add what will be provided from customer
4. The building of an index via a remote endpoint will supported with indices that uses NativeKNNVectorsFormat and with old indices which uses DocValuesFormat(old indices format due to the limitations on access of different index/mapping related attributes at the codec level)

### Roles and Responsibilities

1. Given the details of the object store, upload vectors data to the store and download the index once it is created.
2. Given any endpoint for Index build service call the remote endpoint to create the vector index per segment.
3. Have intelligent logic to take a decision when to use Index build service endpoint vs Local compute to build the index.
4. Once the index is built download the index from remote store and put it along with other segment related files.

### High Level Designs:
- [x] #2392
- [x] #2393

### POC Implementations:
- [x] Repository Service Integration POCs: https://github.com/opensearch-project/k-NN/issues/2465#issuecomment-2641328426
- [x] POC - Remote Vector Build Service Client

### Low Level Designs:
- [x] https://github.com/opensearch-project/k-NN/issues/2465
- [x] https://github.com/opensearch-project/k-NN/issues/2518

[...]

PR #2525, opened 2025-02-13
[Remote Vector Index Build] Introduce RemoteNativeIndexBuildStrategy skeleton

jed326:
### Description
First PR for https://github.com/opensearch-project/k-NN/issues/2465

In order to review changes incrementally, this PR is scoped down to only the following:
- [ ] Feature Flag
- [ ] Index Setting
- [ ] Repository name cluster setting
- [ ] RemoteIndexBuilder skeleton with repository service and index settings wired through the Codec

I am keeping the vector upload changes in a separate follow-up PR as that will deserve it's own in-depth discussion.

The key part of this PR is that we need to pass a `Supplier` for `KNNVectorValues` through to the `NativeIndexBuildStrategy` as we need to open up multiple `InputStream`s on multiple `KNNVectorValues` in order to support uploading blobs in parallel. The actual implementation of this parallel upload will come in the next PR, however I am laying the groundwork for that in this skeleton implementation.

### Related Issues
Relates: https://github.com/opensearch-project/k-NN/issues/2465

### Check List


src/main/java/org/opensearch/knn/index/codec/KNN10010Codec/KNN10010PerFieldKnnVectorsFormat.java
```diff
@@ -0,0 +1,120 @@
...
+import org.opensearch.knn.index.remote.RemoteIndexBuilder;
+
+import java.util.Optional;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+
+/**
+ * Class provides per field format implementation for Lucene Knn vector type
+ */
+public class KNN10010PerFieldKnnVectorsFormat extends BasePerFieldKnnVectorsFormat {
```
navneet1v, line 30: do we really need this class? It seems like a copy of the old class to me.

src/main/java/org/opensearch/knn/index/codec/KNN10010Codec/NativeEngines10010KnnVectorsWriter.java
```diff
@@ -0,0 +1,338 @@
...
+import java.util.function.Supplier;
+
+import static org.opensearch.knn.common.FieldInfoExtractor.extractVectorDataType;
+import static org.opensearch.knn.index.vectorvalues.KNNVectorValuesFactory.getVectorValues;
+
+/**
+ * A KNNVectorsWriter class for writing the vector data strcutures and flat vectors for Native Engines.
+ */
+@Log4j2
+public class NativeEngines10010KnnVectorsWriter extends KnnVectorsWriter {
```
navneet1v, line 51: @jed326 if we are writing this new Writer class then we should delete the old class. Or we should make change in the old Writer class.

src/main/java/org/opensearch/knn/index/codec/KNN10010Codec/KNN10010PerFieldKnnVectorsFormat.java
jed326: I think we need the`10010` Codec classes for BWC because otherwise an index created on OpenSearch 2.19 would start trying to build the indices remotely after cluster version upgrade. Same for the `NativeEngines10010KnnVectorsWriter` as well, we only want to use that writer on the OpenSearch 3.0+ codec versions.

jed326:
Thanks @navneet1v , I also agree that a new writer isn't completely necessary as the underlying formats are not changing. Moreover, the remote build should be format agnostic anyways, so I've refactored `NativeIndexWriter` into an interface to be used either locally or remotely when performing flush or merge.

src/main/java/org/opensearch/knn/index/KNNSettings.java
```diff
@@ -371,6 +373,15 @@ public class KNNSettings {
...
     );
 
+    public static final Setting<Boolean> KNN_INDEX_REMOTE_VECTOR_BUILD_SETTING = Setting.boolSetting(
+        KNN_INDEX_REMOTE_VECTOR_BUILD,
+        false,
+        Dynamic,
+        IndexScope
+    );
+
+    public static final Setting<String> KNN_REMOTE_VECTOR_REPO_SETTING = Setting.simpleString(KNN_REMOTE_VECTOR_REPO, Dynamic, NodeScope);
```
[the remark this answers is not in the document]
kotwanikunal: I think I answered my question. I saw the implementation for `getRepository()`

src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexWriter.java
```diff
@@ -5,69 +5,78 @@
...
 
 /**
- * Writes KNN Index for a field in a segment. This is intended to be used for native engines
+ * Interface for writing a KNN index field in a segment. This is intended to be used for native engines.
  */
-@AllArgsConstructor
-@Log4j2
-public class NativeIndexWriter {
-    private static final Long CRC32_CHECKSUM_SANITY = 0xFFFFFFFF00000000L;
+public interface NativeIndexWriter {
```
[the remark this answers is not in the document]
jed326: Thanks @kotwanikunal , regarding class marking I think that's a concept specific to core as [1] it's only mentioned in the core repo and [2] there are no other tags in the k-nn plugin: https://github.com/opensearch-project/OpenSearch/blob/fc81a906da88d99c7ea09ffd9fec92ff81925cf6/CONTRIBUTING.md?plain=1#L174-L177

Regarding consumers:
1. I don't see any entry point for a custom `NativeIndexWriter` to be provided by a consumer so I'm not really sure how a consumer could instantiate it, and we've extracted all the public methods into the interface.
2. We're targeting 3.0 for this change, so I'm fine with taking the potentially breaking change especially as it's a much cleaner implementation

src/main/java/org/opensearch/knn/index/codec/nativeindex/RemoteNativeIndexWriter.java
```diff
@@ -0,0 +1,45 @@
...
+
+    @Override
+    public void flushIndex(KNNVectorValues<?> knnVectorValues, int totalLiveDocs) throws IOException {
+        fallbackWriter.flushIndex(knnVectorValues, totalLiveDocs);
+    }
+
+    @Override
+    public void mergeIndex(KNNVectorValues<?> knnVectorValues, int totalLiveDocs) throws IOException {
+        fallbackWriter.mergeIndex(knnVectorValues, totalLiveDocs);
+    }
```
navneet1v, line 44: I think we are just using fallback writer here for tests to succeed. Please add TODO that in upcoming PR this needs to be fixed.

src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexWriter.java
```diff
@@ -5,69 +5,78 @@
...
+     * @return {@link RemoteNativeIndexWriter} if {@link RemoteIndexBuilder} is available configured and configured properly. Uses {@link LocalNativeIndexWriter} as fallback.
+     */
+    static NativeIndexWriter getWriter(
+        final FieldInfo fieldInfo,
+        final SegmentWriteState segmentWriteState,
+        final QuantizationState quantizationState,
+        final RemoteIndexBuilder remoteIndexBuilder,
+        final Supplier<KNNVectorValues<?>> knnVectorValuesSupplier
+    ) {
+        if (remoteIndexBuilder != null && remoteIndexBuilder.shouldBuildIndexRemotely()) {
```
navneet1v, line 61: I think this logic will evolve with number of docs logic to ensure that not for every request we are going to remote service for index builds. Lets add this TODO to ensure that we are not missing it in next iteration and also clear for other reviewers.

src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexWriter.java
```diff
@@ -5,69 +5,78 @@
...
+    /**
+     * Merges kNN index
+     * @param knnVectorValues
+     * @throws IOException
+     */
+    void mergeIndex(KNNVectorValues<?> knnVectorValues, int totalLiveDocs) throws IOException;
+
+    /**
+     *
+     * @param fieldInfo
```
navneet1v, line 47: add the docs

src/main/java/org/opensearch/knn/plugin/KNNPlugin.java
```diff
@@ -281,10 +282,14 @@ public Optional<EngineFactory> getEngineFactory(IndexSettings indexSettings) {
...
 
+    /**
+     * Get KNN codec service.
+     * We create {@link RemoteIndexBuilder} per-index in order to consume the {@link IndexSettings} from the {@link org.apache.lucene.codecs.KnnVectorsWriter} implementation.
+     */
     @Override
     public Optional<CodecServiceFactory> getCustomCodecServiceFactory(IndexSettings indexSettings) {
         if (indexSettings.getValue(KNNSettings.IS_KNN_INDEX_SETTING)) {
-            return Optional.of(KNNCodecService::new);
+            return Optional.of((config) -> new KNNCodecService(config, new RemoteIndexBuilder(repositoriesServiceSupplier, indexSettings)));
```
navneet1v, line 292: I like this idea of passing index settings during codec creation to RemoteIndex builder rather than just using index name and getting index settings from cluster state from any part of the code. 

This is good thinking.

src/test/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriterFlushTests.java
```diff
@@ -695,8 +703,9 @@ public void testFlush_whenQuantizationIsProvided_whenBuildGraphDatStructureThres
                     throw new RuntimeException(e);
                 }
 
-                nativeIndexWriterMockedStatic.when(() -> NativeIndexWriter.getWriter(fieldInfo, segmentWriteState, quantizationState))
-                    .thenReturn(nativeIndexWriter);
+                nativeIndexWriterMockedStatic.when(
+                    () -> NativeIndexWriter.getWriter(eq(fieldInfo), eq(segmentWriteState), eq(quantizationState), any(), any())
```
navneet1v, line 707: can we avoid doing `any()` in the mocking, have proper mock objects.

src/main/java/org/opensearch/knn/common/featureflags/KNNFeatureFlags.java
```diff
@@ -35,8 +36,15 @@ public class KNNFeatureFlags {
         Dynamic
     );
 
+    public static final Setting<Boolean> KNN_REMOTE_VECTOR_BUILD_SETTING = Setting.boolSetting(
```
navneet1v, line 39: Please add IT where where this setting is enabled and the tests are passing.

src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexWriter.java
jed326: Yes we are tracking this in https://github.com/opensearch-project/k-NN/issues/2391, will add a TODO as well

src/main/java/org/opensearch/knn/index/codec/nativeindex/RemoteNativeIndexWriter.java
jed326: I think I can just add the control flow here, no need to wait for the next PR

src/test/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriterFlushTests.java
jed326: Good call, will make this change.

jed326: I'm actually not totally sure how to match the `Supplier<KNNVectorValues<?>>` argument to `getWriter()` now without an `any()`. From what I can tell I don't think you can, so I would need to pull the supplier creation into it's own method in `NativeEngines990KnnVectorsWriter` to mock properly. WDYT?

jed326: I'm going to stick with the `any()` for now, I will figure out how to properly mock this in order to write the tests for the following PR as we will need to test the `RemoteIndexBuilder` functionalities there too.

src/main/java/org/opensearch/knn/index/codec/KNN10010Codec/KNN10010PerFieldKnnVectorsFormat.java
shatejas: Can't we simply check the index version for remote builds?

src/main/java/org/opensearch/knn/index/remote/RemoteIndexBuilder.java
```diff
@@ -0,0 +1,161 @@
...
+
+        stopWatch = new StopWatch().start();
+        submitVectorBuild();
+        time_in_millis = stopWatch.stop().totalTime().millis();
+        log.debug("Submit vector build took {} ms for vector field [{}]", time_in_millis, fieldInfo.getName());
+
+        stopWatch = new StopWatch().start();
+        awaitVectorBuild();
+        time_in_millis = stopWatch.stop().totalTime().millis();
+        log.debug("Await vector build took {} ms for vector field [{}]", time_in_millis, fieldInfo.getName());
```
shatejas, line 114: Reusing the stopwatch adds to the total time, can you double check and make sure this debug log is not misleading?

src/main/java/org/opensearch/knn/index/codec/KNNCodecService.java
```diff
@@ -5,21 +5,24 @@
...
 /**
  * KNNCodecService to inject the right KNNCodec version
  */
 public class KNNCodecService extends CodecService {
 
     private final MapperService mapperService;
+    private final RemoteIndexBuilder remoteIndexBuilder;
 
-    public KNNCodecService(CodecServiceConfig codecServiceConfig) {
+    public KNNCodecService(CodecServiceConfig codecServiceConfig, RemoteIndexBuilder remoteIndexBuilder) {
```
shatejas, line 22: I am a bit hesitant for the entire codec should have access to remote index builder, Ideally the access be limited to knn writer implementation. Any restrictions on `KNNCodecService` only hold the respository service supplier instance and then pass it to the writer?

src/main/java/org/opensearch/knn/index/codec/nativeindex/RemoteNativeIndexWriter.java
```diff
@@ -0,0 +1,63 @@
...
+    ) {
+        this.fieldInfo = fieldInfo;
+        this.segmentWriteState = segmentWriteState;
+        this.fallbackWriter = fallbackWriter;
+        this.remoteIndexBuilder = remoteIndexBuilder;
+        this.knnVectorValuesSupplier = knnVectorValuesSupplier;
+    }
+
+    @Override
+    public void flushIndex(KNNVectorValues<?> knnVectorValues, int totalLiveDocs) throws IOException {
```
shatejas, line 45: I do think the responsibilities of submitting the vector builds and polling should be this class's responsibility, it will help keep the flexibility of the entire logic of trainsitioning into different states as the logic evolves. I think  `RemoteIndexBuilder` should be renamed and should have an isolated responsibility of managing uploads and downloads.

thoughts?

src/main/java/org/opensearch/knn/index/codec/nativeindex/RemoteNativeIndexWriter.java
```diff
@@ -0,0 +1,63 @@
...
+        try {
+            remoteIndexBuilder.buildIndexRemotely(fieldInfo, knnVectorValuesSupplier, totalLiveDocs, segmentWriteState);
+        } catch (Exception e) {
+            log.warn("Failed to flush index remotely", e);
+            fallbackWriter.flushIndex(knnVectorValues, totalLiveDocs);
```
shatejas, line 50: cancel API call before this?

src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexWriter.java
```diff
@@ -5,69 +5,84 @@
...
-    private final NativeIndexBuildStrategy indexBuilder;
-    @Nullable
-    private final QuantizationState quantizationState;
+    /**
+     * flushes the index
+     *
+     * @param knnVectorValues
+     * @throws IOException
+     */
+    void flushIndex(KNNVectorValues<?> knnVectorValues, int totalLiveDocs) throws IOException;
```
shatejas, line 36: I like the abstraction!

src/main/java/org/opensearch/knn/index/codec/nativeindex/LocalNativeIndexWriter.java
```diff
@@ -0,0 +1,260 @@
...
+import static org.opensearch.knn.index.codec.util.KNNCodecUtil.buildEngineFileName;
+import static org.opensearch.knn.index.engine.faiss.Faiss.FAISS_BINARY_INDEX_DESCRIPTION_PREFIX;
+
+/**
+ * Writes KNN Index for a field in a segment. This is intended to be used for native engines. This class uses the local CPU path for building segments.
+ * See {@link RemoteNativeIndexWriter} for remote vector index build path.
+ */
+@AllArgsConstructor
+@Log4j2
+public class LocalNativeIndexWriter implements NativeIndexWriter {
```
shatejas, line 56: I am hoping nothing changed in this, do highlight if there are changes and I can take a deeper look

src/main/java/org/opensearch/knn/index/codec/nativeindex/RemoteNativeIndexWriter.java
```diff
@@ -0,0 +1,63 @@
...
+        this.segmentWriteState = segmentWriteState;
+        this.fallbackWriter = fallbackWriter;
+        this.remoteIndexBuilder = remoteIndexBuilder;
+        this.knnVectorValuesSupplier = knnVectorValuesSupplier;
+    }
+
+    @Override
+    public void flushIndex(KNNVectorValues<?> knnVectorValues, int totalLiveDocs) throws IOException {
+        try {
+            remoteIndexBuilder.buildIndexRemotely(fieldInfo, knnVectorValuesSupplier, totalLiveDocs, segmentWriteState);
```
shatejas, line 47: Its missing flush and merge stats

src/main/java/org/opensearch/knn/index/remote/RemoteIndexBuilder.java
jed326: I'm only re-using the variable, a new instance is being created each time:
```
```

src/main/java/org/opensearch/knn/index/codec/KNNCodecService.java
jed326: >  Any restrictions on KNNCodecService only hold the respository service supplier instance and then pass it to the writer?

This is actually how I started in my first POC, but I had 2 main reasons for this. First, I also want to pass in the indexSettings so that I can check within the `RemoteIndexBuilder` per-index if the feature should actually be used. Second, in the future we may want to support GPU acceleration for search case in a similar fashion, so at that time we could re-use the same `RemoteIndexBuilder` (probably renamed) to perform the similar orchestration.

>  Ideally the access be limited to knn writer implementation.

I don't think I fully understand this -- if I were to pass just the repository service supplier down and only use it in the knn writer implementation isn't that the same concern?

src/main/java/org/opensearch/knn/index/codec/nativeindex/RemoteNativeIndexWriter.java
jed326: My thinking here was that the `RemoteIndexBuilder` would be responsible for all parts of "building" the index, so that includes uploading, polling, downloading, and then the `RemoteNativeIndexWriter` would be responsible for actually writing the index to the `IndexOutput`.

src/main/java/org/opensearch/knn/index/codec/nativeindex/RemoteNativeIndexWriter.java
jed326: We'll finalize on all functionality related aspects in a future PR, but yes this most likely would be the place.

src/test/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriterFlushTests.java
jed326: Agreed, there's no functionality to be tested yet in this PR though so it would be hard to verify if any captor/matcher semantic is even working so I will follow-up with that on the next PR.

src/main/java/org/opensearch/knn/index/codec/nativeindex/LocalNativeIndexWriter.java
jed326: Yep no changes, this was done by IntelliJ's "extract as interface" functionality

src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexWriter.java
shatejas: I  took a look at the code and I was wondering why we need this on NativeIndexWriter level? We already have [NativeIndexBuildStrategy](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexBuildStrategy.java) and RemoteBuildStrategy should just be an extension of that, Any reason we didn't use it?

jed326: Good question. So the main problem is that I need the `knnVectorValuesSupplier` itself so that I can create multiple `InputStreams` in order to perform parallel uploads as the knnVectorValues iterator itself is sequential. However, the `BuildIndexParams` takes a concrete `KNNVectorValues<?>`, so I would need to refactor this class to instead take the supplier and that is a huge refactor.

Beyond that, I would still need to pass the `RemoteIndexBuilder` (or the repository service supplier + index settings into the `NativeIndexWriter` anyways to determine if/when I should use the `RemoteBuildStrategy`.

I actually do agree that long-term, refactoring this into a `RemoteBuildStrategy` may be a good idea, but I really want to avoid taking on large scale refactors in the middle of feature development.

src/main/java/org/opensearch/knn/index/codec/KNNCodecService.java
shatejas: > First, I also want to pass in the indexSettings so that I can check within the RemoteIndexBuilder per-index if the feature should actually be used

Why not pass in the IndexSettings separately rather than wrapping up in remoteIndexBuilder?

> Second, in the future we may want to support GPU acceleration for search case in a similar fashion, so at that time we could re-use the same RemoteIndexBuilder (probably renamed) to perform the similar orchestration

IndexBuilding shouldn't have search logic, we can always have a separate class for search to have single responsibility

> f I were to pass just the repository service supplier down and only use it in the knn writer implementation isn't that the same concern?

Not really, because no other part of code has access to buildIndex and cannot misuse it

src/main/java/org/opensearch/knn/index/codec/nativeindex/RemoteNativeIndexWriter.java
shatejas: See https://github.com/opensearch-project/k-NN/pull/2525#discussion_r1956772899 I think the implementation of build strategy should be responsible

src/main/java/org/opensearch/knn/index/codec/KNNCodecService.java
jed326: > Why not pass in the IndexSettings separately rather than wrapping up in remoteIndexBuilder?

First, the two are meant to be used together so it makes sense to logically bundle them together. Second, and more pragmatically, the `knnCodecSupplier` is a `TriFunction` currently and if we need to pass the repository supplier + settings then we need 4 arguments and there's no "QuadFunction" available. We could implement this functional interface of course but ideally that should go in opensearch core.

> IndexBuilding shouldn't have search logic, we can always have a separate class for search to have single responsibility

Agreed, that's why I said we would probably rename it if we do see the need for the functionality of using repository service + settings in the future.

> Not really, because no other part of code has access to buildIndex and cannot misuse it

But you could make the same argument that if repositoryservice is only meant for use during index build then it could be misused still?

src/main/java/org/opensearch/knn/index/remote/RemoteIndexBuilder.java
shatejas: Missed it, Thanks for clarifying

jed326:
@shatejas I've reworked this skeleton in the form of `RemoteIndexBuildStrategy implements NativeIndexBuildStrategy`. Javadoc is still needed in a few places, but this PR is ready for review on the overall approach. Thanks!

src/test/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriterFlushTests.java
```diff
@@ -287,35 +284,31 @@ public void testFlush_WithQuantization() {
...
+                        verify(nativeIndexWriter).flushIndex(expectedVectorValuesSuppliers.get(i), vectorsPerField.get(i).size());
                     }
                 } catch (Exception e) {
                     throw new RuntimeException(e);
                 }
             });
             final Long expectedTimesGetVectorValuesIsCalled = vectorsPerField.stream().filter(Predicate.not(Map::isEmpty)).count();
             knnVectorValuesFactoryMockedStatic.verify(
-                () -> KNNVectorValuesFactory.getVectorValues(any(VectorDataType.class), any(DocsWithFieldSet.class), any()),
-                times(Math.toIntExact(expectedTimesGetVectorValuesIsCalled) * 2)
```
jed326, line 302: In quantization case `getVectorValues` is called twice, however now we are retrieving the supplier itself via `getVectorValuesSupplier`, which we only need to do once then we are passing the supplier through the rest of the flow.

src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexWriter.java
```diff
@@ -304,19 +311,37 @@ private void recordRefreshStats() {
...
         final KNNEngine knnEngine = extractKNNEngine(fieldInfo);
         boolean isTemplate = fieldInfo.attributes().containsKey(MODEL_ID);
         boolean iterative = !isTemplate && KNNEngine.FAISS == knnEngine;
         NativeIndexBuildStrategy strategy = iterative
             ? MemOptimizedNativeIndexBuildStrategy.getInstance()
             : DefaultIndexBuildStrategy.getInstance();
-        return new NativeIndexWriter(state, fieldInfo, strategy, quantizationState);
+
+        // TODO: Add threshold checks for if/when we should use remote index build strategy
+        if (knnEngine == KNNEngine.FAISS
```
jmazanec15, line 333: Engine checks should be abstracted into engine interface. We want to avoid hard checks like this

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
```diff
@@ -0,0 +1,170 @@
...
+ */
+@Log4j2
+@ExperimentalApi
+public class RemoteIndexBuildStrategy implements NativeIndexBuildStrategy {
+
+    private final Supplier<RepositoriesService> repositoriesServiceSupplier;
+    private final NativeIndexBuildStrategy fallbackStrategy;
+    private static final String VECTOR_BLOB_FILE_EXTENSION = ".knnvec";
+    private static final String DOC_ID_FILE_EXTENSION = ".knndid";
+    private static final String GRAPH_FILE_EXTENSION = ".faiss";
```
jmazanec15, line 42: This is present in the engine interface

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
```diff
@@ -0,0 +1,170 @@
...
+     */
+    private BlobStoreRepository getRepository() throws RepositoryMissingException {
+        RepositoriesService repositoriesService = repositoriesServiceSupplier.get();
+        assert repositoriesService != null;
+        String vectorRepo = KNNSettings.state().getSettingValue(KNN_REMOTE_VECTOR_REPO_SETTING.getKey());
+        if (vectorRepo == null || vectorRepo.isEmpty()) {
+            throw new RepositoryMissingException("Vector repository " + KNN_REMOTE_VECTOR_REPO_SETTING.getKey() + " is not registered");
+        }
+        final Repository repository = repositoriesService.repository(vectorRepo);
+        assert repository instanceof BlobStoreRepository : "Repository should be instance of BlobStoreRepository";
```
jmazanec15, line 126: How sure are we this check will succeed?

src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexWriter.java
jed326: Do you mean we should add a method to `KNNEngine` to do something like `supportsRemoteBuild`? This same check is happening on L327 above so I figured it was fine here too.

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
jed326: This exact same check is being used everywhere else in core: https://github.com/search?q=repo%3Aopensearch-project%2FOpenSearch%20%22assert%20repository%20instanceof%20BlobStoreRepository%20%22&type=code

`BlobStoreRepository` is the base implementation (abstract class) of `Repository` interface

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
jed326: Removed in the latest revision

Vikasht34 (approved):
@shatejas already had made some good suggestions on PR , I see those being addressed. I am good with approach and class hierarchy.

src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexWriter.java
jed326: Moved this into an attribute check of the faiss engine instead in the latest revision, thanks!

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriter.java
```diff
@@ -54,15 +55,29 @@ public class NativeEngines990KnnVectorsWriter extends KnnVectorsWriter {
     private final List<NativeEngineFieldVectorsWriter<?>> fields = new ArrayList<>();
     private boolean finished;
     private final Integer approximateThreshold;
+    private final Supplier<RepositoriesService> repositoriesServiceSupplier;
```
jmazanec15, line 58: Instead of taking a hard dependency on the RepositoriesService, can we build an abstraction, like RemoteIndexBuilder, that hides these details from the IndexWriter class?

src/main/java/org/opensearch/knn/index/KNNSettings.java
```diff
@@ -371,6 +373,15 @@ public class KNNSettings {
         NodeScope
     );
 
+    public static final Setting<Boolean> KNN_INDEX_REMOTE_VECTOR_BUILD_SETTING = Setting.boolSetting(
```
jmazanec15, line 376: Can we add javadocs on public settings for future maintainability?

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
```diff
@@ -0,0 +1,169 @@
...
+            stopWatch = new StopWatch().start();
+            awaitVectorBuild();
+            time_in_millis = stopWatch.stop().totalTime().millis();
+            log.debug("Await vector build took {} ms for vector field [{}]", time_in_millis, indexInfo.getFieldName());
+
+            stopWatch = new StopWatch().start();
+            readFromRepository();
+            time_in_millis = stopWatch.stop().totalTime().millis();
+            log.debug("Repository read took {} ms for vector field [{}]", time_in_millis, indexInfo.getFieldName());
+        } catch (Exception e) {
```
jmazanec15, line 104: Why are we catching everything?

jmazanec15 (commented):
Can we hide repositories service dependency inside some RemoteIndexBuilder abstraction instead of taking direct dependency on it in codec classes? 

Also, I think this should go into a feature branch, especially now that main is a release branch. I think overall structure is good, but its still WIP and doesnt have tests.

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
jed326: The idea is we want to fall back to the CPU path on any failures, we will have more precise exception handling in future PRs such as specific upload/download failures, as well as the whole flow with using the remote index build client.

src/main/java/org/opensearch/knn/index/codec/nativeindex/model/BuildIndexParams.java
```diff
@@ -31,6 +33,7 @@ public class BuildIndexParams {
      */
     @Nullable
     QuantizationState quantizationState;
-    KNNVectorValues<?> vectorValues;
+    Supplier<KNNVectorValues<?>> knnVectorValuesSupplier;
```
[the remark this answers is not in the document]
jed326: Copying this from the PR description:

> The key part of this PR is that we need to pass a Supplier for KNNVectorValues through to the NativeIndexBuildStrategy as we need to open up multiple InputStreams on multiple KNNVectorValues in order to support uploading blobs in parallel. The actual implementation of this parallel upload will come in the next PR, however I am laying the groundwork for that in this skeleton implementation.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriter.java
jed326: One approach I previously considered was in `NativeEngines990KnnVectorsFormat` itself we can do the feature checks and then create an instance of `RemoteIndexBuildStrategy` to pass to the `NativeEngines990KnnVectorsWriter`. However, it seemed better to me to not instantiate any of the remote index build related classes unless they were actually needed, so I went with this approach of passing the `repositoriesServiceSupplier` through to the index writer.

jed326:
Thanks @jmazanec15 

> Also, I think this should go into a feature branch, especially now that main is a release branch. I think overall structure is good, but its still WIP and doesnt have tests.

I do have a feature branch in my fork that I have been developing on (https://github.com/jed326/k-NN/commits/remote-vector-staging-2.19/), but I strongly believe we need to start merging these changes into `main` incrementally otherwise it feels like we are just kicking the can down the road on these high level discussions.

In terms of testing I will add some coverage to ensure the fallback mechanism is working. I was thinking these weren't needed for now since all the functional testing would come in subsequent PRs that include the functionality itself.

jmazanec15:
@navneet1v @shatejas @Vikasht34 What do you guys think on main vs. feature branch?

Vikasht34:
> @navneet1v @shatejas @Vikasht34 What do you guys think on main vs. feature branch?

Feature Branch

navneet1v:
> @navneet1v @shatejas @Vikasht34 What do you guys think on main vs. feature branch?

if tests are there and code is buildable then we should go with main branch. There is no point in feature branch.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriter.java
jed326: Introduced a `NativeIndexBuildStrategyFactory` class in the latest revision

jed326:
> @navneet1v @shatejas @Vikasht34 What do you guys think on main vs. feature branch?

I would strongly prefer to not use a feature branch. I think this PR itself is a prime example of why it's important to get a wide range of opinions on changes like this that include a decent amount of refactoring and I think this PR would not have gotten the same amount of feedback if it were directed at a feature branch.

For testing, in the latest revision I have added some randomization to the base test class to randomly enable/disable the new settings and feature flag to ensure the fallback mechanisms are working correctly.

jmazanec15:
Okay, lets add some tests and we can develop on main, assuming proper feature flag.

jed326:
Added a new `RemoteIndexBuildStrategyTests.java` as well to test the fallback mechanism. I think that + the settings updates in the `KNNRestTestCase` base class should give us good coverage on this change, but please let me know if there's any other coverage you think is missing, thanks!

jmazanec15 (approved):
LGTM

Vikasht34 (approved):
Thanks for making all the changes as suggested !! LGTM

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
```diff
@@ -0,0 +1,169 @@
...
+    /**
+     * @return whether to use the remote build feature
+     */
+    public static boolean shouldBuildIndexRemotely(IndexSettings indexSettings) {
+        String vectorRepo = KNNSettings.state().getSettingValue(KNN_REMOTE_VECTOR_REPO_SETTING.getKey());
+        return KNNFeatureFlags.isKNNRemoteVectorBuildEnabled()
+            && indexSettings.getValue(KNN_INDEX_REMOTE_VECTOR_BUILD_SETTING)
+            && vectorRepo != null
+            && !vectorRepo.isEmpty();
+    }
```
navneet1v, line 62: This logic needs to be involved bit more where the doc count needs to be added, say if number of docs in segment < 10K we should go with local build as compare to remote build. This number 10K can be tuned, or we can make this logic more sophisticated.

So please add that check in the next PR.

PR #2529, opened 2025-02-14
[Enhancement] Make Merge in nativeEngine can Abort

luyuncheng:
### Description
When there is a scenarios: 
1. There is A Merge Task On `Node1#Index1#Shard1`(long time running)
2. After merge task started, begin relocating from `Node1#Index1#Shard1` TO `Node2#Index1#Shard1`
3. At the finalize step, source need do closeShard, but the merge task would take a long time, stack as following shows.
4. The clusterApplierService would wait for about N minutes(long time running), and mark the node stale, and master let node1 left because of this.
 
```
opensearch[datanode1][clusterApplierService#updateTask][T#1]" #41 daemon prio=5 os_prio=0 cpu=5183.70ms elapsed=93132.85s tid=0x00007f3f392509d0 nid=0x101 in Object.wait()  [0x00007f3f6ddfb000]
   java.lang.Thread.State: TIMED_WAITING (on object monitor)
	at java.lang.Object.wait(java.base@17.0.2/Native Method)
	- waiting on <no object reference available>
	at org.apache.lucene.index.IndexWriter.doWait(IndexWriter.java:5410)
	- locked <0x0000001022b0abe8> (a org.apache.lucene.index.IndexWriter)
	at org.apache.lucene.index.IndexWriter.abortMerges(IndexWriter.java:2721)
	- locked <0x0000001022b0abe8> (a org.apache.lucene.index.IndexWriter)
	at org.apache.lucene.index.IndexWriter.rollbackInternalNoCommit(IndexWriter.java:2469)
	- locked <0x0000001022b0abe8> (a org.apache.lucene.index.IndexWriter)
	at org.apache.lucene.index.IndexWriter.rollbackInternal(IndexWriter.java:2449)
	- locked <0x0000001022bae6d0> (a java.lang.Object)
	at org.apache.lucene.index.IndexWriter.rollback(IndexWriter.java:2441)
	at org.opensearch.index.engine.InternalEngine.closeNoLock(InternalEngine.java:2370)
	at org.opensearch.index.engine.Engine.close(Engine.java:2000)
	at org.opensearch.index.engine.Engine.flushAndClose(Engine.java:1987)
	at org.opensearch.index.shard.IndexShard.close(IndexShard.java:1907)
	- locked <0x0000001022b07ea0> (a java.lang.Object)
	at org.opensearch.index.IndexService.closeShard(IndexService.java:623)
	at org.opensearch.index.IndexService.removeShard(IndexService.java:599)
	- locked <0x0000001022a976a8> (a org.opensearch.index.IndexService)
	at org.opensearch.index.IndexService.close(IndexService.java:374)
	- locked <0x0000001022a976a8> (a org.opensearch.index.IndexService)
	at org.opensearch.indices.IndicesService.removeIndex(IndicesService.java:993)
	at org.opensearch.indices.cluster.IndicesClusterStateService.removeIndices(IndicesClusterStateService.java:446)
	at org.opensearch.indices.cluster.IndicesClusterStateService.applyClusterState(IndicesClusterStateService.java:287)
	- locked <0x000000100b7da520> (a org.opensearch.indices.cluster.IndicesClusterStateService)
	at org.opensearch.cluster.service.ClusterApplierService.callClusterStateAppliers(ClusterApplierService.java:606)
	at org.opensearch.cluster.service.ClusterApplierService.callClusterStateAppliers(ClusterApplierService.java:593)
``` 
### Proposal

i think we can introduce abort mechanism for long time merge task meanwhile close shard called.

i think we can introduce `KNNMergeHelper` class to check if merge aborted. and when build the graph, we can reuse faiss::InterruptCallback which is interrupt callback mechanism to check whether aborted or not

BUT `ConcurrentMergeScheduler#MergeThread` is a internal class, we can not call this directly. it throws `org.apache.lucene.index.ConcurrentMergeScheduler$MergeThread is in unnamed module of loader 'app'`

we can added this static method into OpenSearch Core like `OneMergeHelper`

### Related Issues

### Check List

### DIFF
--- a/jni/src/org_opensearch_knn_jni_FaissService.cpp
+++ b/jni/src/org_opensearch_knn_jni_FaissService.cpp
@@ -481,3 +481,23 @@ JNIEXPORT jobjectArray JNICALL Java_org_opensearch_knn_jni_FaissService_rangeSea
     }
     return nullptr;
 }
+
+JNIEXPORT void JNICALL Java_org_opensearch_knn_jni_FaissService_setMergeInterruptCallback(JNIEnv * env, jclass cls)
+{
+    try {
+        faiss::InterruptCallback::instance.reset(
+            new knn_jni::faiss_wrapper::OpenSearchMergeInterruptCallback(&jniUtil, env)
+        );
+    } catch (...) {
+        jniUtil.CatchCppExceptionAndThrowJava(env);
+    }
+}
+
+JNIEXPORT void JNICALL Java_org_opensearch_knn_jni_FaissService_unsetMergeInterruptCallback(JNIEnv * env, jclass cls)
+{
+    try {
+        faiss::InterruptCallback::instance.get()->clear_instance();
+    } catch (...) {
+        jniUtil.CatchCppExceptionAndThrowJava(env);
+    }
+}
--- a/src/main/java/org/apache/lucene/index/KNNMergeHelper.java
+++ b/src/main/java/org/apache/lucene/index/KNNMergeHelper.java
@@ -1,0 +1,18 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.apache.lucene.index;
+
+public class KNNMergeHelper {
+
+    private KNNMergeHelper() {}
+    public static boolean isMergeAborted() {
+        Thread mergeThread = Thread.currentThread();
+        if (mergeThread instanceof ConcurrentMergeScheduler.MergeThread) {
+            return ((ConcurrentMergeScheduler.MergeThread) mergeThread).merge.isAborted();
+        }
+        return false;
+    }
+}
--- a/src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexWriter.java
+++ b/src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexWriter.java
@@ -9,6 +9,7 @@ import lombok.AllArgsConstructor;
 import lombok.extern.log4j.Log4j2;
 import org.apache.lucene.codecs.CodecUtil;
 import org.apache.lucene.index.FieldInfo;
+import org.apache.lucene.index.KNNMergeHelper;
 import org.apache.lucene.index.SegmentWriteState;
 import org.apache.lucene.store.IndexOutput;
 import org.opensearch.common.Nullable;
@@ -46,6 +47,8 @@ import static org.opensearch.knn.common.KNNConstants.PARAMETERS;
 import static org.opensearch.knn.index.codec.util.KNNCodecUtil.initializeVectorValues;
 import static org.opensearch.knn.index.codec.util.KNNCodecUtil.buildEngineFileName;
 import static org.opensearch.knn.index.engine.faiss.Faiss.FAISS_BINARY_INDEX_DESCRIPTION_PREFIX;
+import static org.opensearch.knn.jni.JNIService.setMergeInterruptCallback;
+import static org.opensearch.knn.jni.JNIService.unsetMergeInterruptCallback;
 
 /**
  * Writes KNN Index for a field in a segment. This is intended to be used for native engines
@@ -119,9 +122,18 @@ public class NativeIndexWriter {
         }
 
         long bytesPerVector = knnVectorValues.bytesPerVector();
-        startMergeStats(totalLiveDocs, bytesPerVector);
-        buildAndWriteIndex(knnVectorValues, totalLiveDocs);
-        endMergeStats(totalLiveDocs, bytesPerVector);
+        final KNNEngine knnEngine = extractKNNEngine(fieldInfo);
+        setMergeInterruptCallback(knnEngine);
+        try {
+            startMergeStats(totalLiveDocs, bytesPerVector);
+            buildAndWriteIndex(knnVectorValues, totalLiveDocs);
+            endMergeStats(totalLiveDocs, bytesPerVector);
+        } catch (Exception ex) {
+            //TODO handle
+            log.debug("Merge may abort {}",KNNMergeHelper.isMergeAborted());
+        } finally {
+            unsetMergeInterruptCallback(knnEngine);
+        }
     }
 
     private void buildAndWriteIndex(final KNNVectorValues<?> knnVectorValues, int totalLiveDocs) throws IOException {
--- a/src/main/java/org/opensearch/knn/jni/FaissService.java
+++ b/src/main/java/org/opensearch/knn/jni/FaissService.java
@@ -463,4 +463,8 @@ class FaissService {
         int indexMaxResultWindow,
         int[] parentIds
     );
+
+    public static native void setMergeInterruptCallback();
+
+    public static native void unsetMergeInterruptCallback();
 }
--- a/jni/include/faiss_index_service.h
+++ b/jni/include/faiss_index_service.h
@@ -17,6 +17,7 @@
 #include <jni.h>
 #include "faiss/MetricType.h"
 #include "faiss/impl/io.h"
+#include "faiss/impl/AuxIndexStructures.h"
 #include "jni_util.h"
 #include "faiss_methods.h"
 #include "faiss_stream_support.h"
@@ -195,6 +196,19 @@ public:
     void allocIndex(faiss::Index * index, size_t dim, size_t numVectors) final;
 };  // class ByteIndexService
 
+struct OpenSearchMergeInterruptCallback : faiss::InterruptCallback {
+
+    OpenSearchMergeInterruptCallback(JNIUtil *jniUtil, JNIEnv *env) {
+        mergeHelperClass = jniUtil->FindClass(env,"org/apache/lucene/index/KNNMergeHelper");
+        isAbortedMethod = jniUtil->FindMethod(env, "org/apache/lucene/index/KNNMergeHelper", "isMergeAborted");
+    }
+    bool want_interrupt () override {
+        return (bool) jenv->CallStaticBooleanMethod(mergeHelperClass, isAbortedMethod);
+    }
+    JNIEnv *jenv;
+    jclass mergeHelperClass;
+    jmethodID isAbortedMethod;
+};
 }
 }
 
--- a/jni/include/org_opensearch_knn_jni_FaissService.h
+++ b/jni/include/org_opensearch_knn_jni_FaissService.h
@@ -268,6 +268,20 @@ JNIEXPORT jobjectArray JNICALL Java_org_opensearch_knn_jni_FaissService_rangeSea
 JNIEXPORT jobjectArray JNICALL Java_org_opensearch_knn_jni_FaissService_rangeSearchIndex
   (JNIEnv *, jclass, jlong, jfloatArray, jfloat, jobject, jint, jintArray);
 
+/*
+ * Class:     org_opensearch_knn_jni_FaissService
+ * Method:    setMergeInterruptCallback
+ * Signature: ()V
+ */
+JNIEXPORT void JNICALL Java_org_opensearch_knn_jni_FaissService_setMergeInterruptCallback(JNIEnv * env, jclass cls);
+
+/*
+ * Class:     org_opensearch_knn_jni_FaissService
+ * Method:    setMergeInterruptCallback
+ * Signature: ()V
+ */
+JNIEXPORT void JNICALL Java_org_opensearch_knn_jni_FaissService_unsetMergeInterruptCallback(JNIEnv * env, jclass cls);
+
 #ifdef __cplusplus
 }
 #endif
--- a/jni/src/jni_util.cpp
+++ b/jni/src/jni_util.cpp
@@ -65,6 +65,11 @@ void knn_jni::JNIUtil::Initialize(JNIEnv *env) {
     this->cachedClasses["org/opensearch/knn/index/query/KNNQueryResult"] = (jclass) env->NewGlobalRef(tempLocalClassRef);
     this->cachedMethods["org/opensearch/knn/index/query/KNNQueryResult:<init>"] = env->GetMethodID(tempLocalClassRef, "<init>", "(IF)V");
     env->DeleteLocalRef(tempLocalClassRef);
+
+    tempLocalClassRef = env->FindClass("org/apache/lucene/index/KNNMergeHelper");
+    this->cachedClasses["org/apache/lucene/index/KNNMergeHelper"] = (jclass) env->NewGlobalRef(tempLocalClassRef);
+    this->cachedMethods["org/apache/lucene/index/KNNMergeHelper:isMergeAborted"] = env->GetStaticMethodID(tempLocalClassRef, "isMergeAborted", "()Z");
+    env->DeleteLocalRef(tempLocalClassRef);
 }
 
 void knn_jni::JNIUtil::Uninitialize(JNIEnv* env) {
--- a/src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80DocValuesConsumer.java
+++ b/src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80DocValuesConsumer.java
@@ -6,6 +6,8 @@
 package org.opensearch.knn.index.codec.KNN80Codec;
 
 import lombok.extern.log4j.Log4j2;
+import org.apache.lucene.index.ConcurrentMergeScheduler;
+import org.apache.lucene.index.KNNMergeHelper;
 import org.opensearch.common.StopWatch;
 import org.opensearch.knn.index.VectorDataType;
 import org.opensearch.knn.index.engine.KNNEngine;
--- a/src/main/java/org/opensearch/knn/jni/JNIService.java
+++ b/src/main/java/org/opensearch/knn/jni/JNIService.java
@@ -476,4 +476,17 @@ public class JNIService {
         }
         throw new IllegalArgumentException(String.format(Locale.ROOT, "RadiusQueryIndex not supported for provided engine"));
     }
+
+    public static void setMergeInterruptCallback(KNNEngine knnEngine) {
+
+        if (KNNEngine.FAISS == knnEngine) {
+            FaissService.setMergeInterruptCallback();
+        }
+    }
+
+    public static void unsetMergeInterruptCallback(KNNEngine knnEngine) {
+        if (KNNEngine.FAISS == knnEngine) {
+            FaissService.unsetMergeInterruptCallback();
+        }
+    }
 }


jmazanec15:
@luyuncheng This is interesting - but can the interrupt callback with faiss be per graph or would it be for all graphs? In other words, would it cancel all graph builds happening on the node instead of just the one for the closed shard.

navneet1v:
@luyuncheng thanks for creating the GH issue. and I think we ourselves have seen this problem in couple of places(ref: https://github.com/opensearch-project/OpenSearch/issues/14828, @kotwanikunal created this), and seeing a solution around this problem is really great. I looked through the code and I want to know how this code is even working? Because as per my understanding of the code you are checking if merge is aborted if somehow write fails and then eating up that exception.


Is this PR only to see if the merge is aborted and write to directory fails how we can handle the errors in case shard is not present on the node because it moved?

luyuncheng:
> This is interesting - but can the interrupt callback with faiss be per graph or would it be for all graphs? In other words, would it cancel all graph builds happening on the node instead of just the one for the closed shard.

@jmazanec15  In Sample Code, https://github.com/opensearch-project/k-NN/blob/32e151a0e41b99688ebe219911734bdf58e05f3d/src/main/java/org/apache/lucene/index/KNNMergeHelper.java#L11-L16 which need include in OpenSearch repo, it shows that it only checked for the `current merge thread` like OS code https://github.com/opensearch-project/OpenSearch/blob/99a9a81da366173b0c2b963b26ea92e15ef34547/server/src/main/java/org/apache/lucene/index/OneMergeHelper.java#L65-L69

jmazanec15:
@luyuncheng right, but in the faiss InterruptCallback: https://github.com/facebookresearch/faiss/blob/657c563604c774461aed0394ae99210713145e03/faiss/impl/AuxIndexStructures.h#L135-L162, it operates globally. So, setting the instance in one thread, will set it in another thread.

So, for instance, here is the interrupt check in HNSW during graph build: https://github.com/facebookresearch/faiss/blob/657c563604c774461aed0394ae99210713145e03/faiss/IndexHNSW.cpp#L178-L180. If the interrupt gets set and another shard is building a graph on the node, wont this shard's graph build fail too?

luyuncheng:
> it operates globally. So, setting the instance in one thread, will set it in another thread.

@jmazanec15  exactly right, `AuxIndexStructures` is an singleton struct. 

every different thread call `want_interrupt` have different return because current thread context is different because we override `want_interrupt`.  also https://github.com/facebookresearch/faiss/blob/657c563604c774461aed0394ae99210713145e03/faiss/IndexHNSW.cpp#L139 `interrupt` is a local variable, 

so every graph build would go into check `if (InterruptCallback::is_interrupted())` but only the current merge thread which aborted would return true.

also, multi thread would get into lock area which would reduce graph build performance, but `check_period` would help reduce the impacts.

jmazanec15:
Oh I see (https://github.com/facebookresearch/faiss/blob/657c563604c774461aed0394ae99210713145e03/faiss/impl/AuxIndexStructures.cpp#L223-L229). That makes sense. 

Let me take look at this PR. This would also help address: https://github.com/opensearch-project/OpenSearch/issues/8590.

src/main/java/org/opensearch/knn/jni/FaissService.java
```diff
@@ -463,4 +463,8 @@ public static native KNNQueryResult[] rangeSearchIndex(
         int indexMaxResultWindow,
         int[] parentIds
     );
+
+    public static native void setMergeInterruptCallback();
```
jmazanec15, line 467: Can we remove these and just have one interrupt callback registered at the time of library initialization? I dont think these need to be in the interface.

src/main/java/org/apache/lucene/index/KNNMergeHelper.java
```diff
@@ -0,0 +1,18 @@
...
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.apache.lucene.index;
+
+public class KNNMergeHelper {
+
+    private KNNMergeHelper() {}
+    public static boolean isMergeAborted() {
```
jmazanec15, line 11: Can we generalize this to include shard closed events? See this comment: https://github.com/opensearch-project/OpenSearch/issues/8590#issuecomment-2590059735. Im not sure if its possible or not to look up that kind of information. @shwetathareja do you have any ideas around here?

src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexWriter.java
```diff
@@ -119,9 +122,18 @@ public void mergeIndex(final KNNVectorValues<?> knnVectorValues, int totalLiveDo
...
-        buildAndWriteIndex(knnVectorValues, totalLiveDocs);
-        endMergeStats(totalLiveDocs, bytesPerVector);
+        final KNNEngine knnEngine = extractKNNEngine(fieldInfo);
+        setMergeInterruptCallback(knnEngine);
+        try {
+            startMergeStats(totalLiveDocs, bytesPerVector);
+            buildAndWriteIndex(knnVectorValues, totalLiveDocs);
+            endMergeStats(totalLiveDocs, bytesPerVector);
+        } catch (Exception ex) {
+            //TODO handle
```
jmazanec15, line 132: Do we need to catch this?

src/main/java/org/opensearch/knn/jni/FaissService.java
jmazanec15: I just prefer not to do this when its a global, static callback.

luyuncheng:
> @luyuncheng are you still working on this?

Yes, Sorry for the late. i am trying to find a way bypass the `KNNMergeHelper` for protect calling into OpenSearch. and trying to find a way for testing

luyuncheng:
@jmazanec15  i found a way using reflect, it would get Accessable into `ConcurrentMergeScheduler.MergeThread.class` like following code:
https://github.com/opensearch-project/k-NN/blob/952d3fb6a6ebccdcdab12a37b59a5d6eff9df697/src/main/java/org/apache/lucene/index/KNNMergeHelper.java#L15-L32
and call into reflect as following which using to get `org.apache.lucene.index.MergePolicy.OneMerge` in `org.apache.lucene.index.ConcurrentMergeScheduler.MergeThread`
 https://github.com/opensearch-project/k-NN/blob/952d3fb6a6ebccdcdab12a37b59a5d6eff9df697/src/main/java/org/apache/lucene/index/LucenePackagePrivateCaller.java#L16-L29

so we can test it without modified OpenSearch Core. 
Also i added tests in `KNN80DocValuesConsumerTests.java`

src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexWriter.java
luyuncheng: DONE

src/main/java/org/apache/lucene/index/KNNMergeHelper.java
luyuncheng: Add Reflect function

src/main/java/org/opensearch/knn/jni/FaissService.java
luyuncheng: @jmazanec15  how about put it into `FaissService#initLibrary`  with global init

navneet1v:
Hi @luyuncheng 
Sorry for delay in response. I looked into this PR and I think the solution looks pretty awesome to me. What would it take to make this code ready for merge?

luyuncheng:
> Hi @luyuncheng Sorry for delay in response. I looked into this PR and I think the solution looks pretty awesome to me. What would it take to make this code ready for merge?

@navneet1v , i have talked with @jmazanec15  but he had some concern about it.  and i would like to continue make this pr in progress.

shatejas:
@luyuncheng would love to see this in 3.5 release, are you actively working on this?

luyuncheng:
> would love to see this in 3.5 release, are you actively working on this?

@shatejas  yes, i will update codes ASAP

jni/src/org_opensearch_knn_jni_FaissService.cpp
```diff
@@ -90,7 +92,17 @@ JNIEXPORT void JNICALL Java_org_opensearch_knn_jni_FaissService_insertToIndex(JN
         std::unique_ptr<knn_jni::faiss_wrapper::FaissMethods> faissMethods(new knn_jni::faiss_wrapper::FaissMethods());
         knn_jni::faiss_wrapper::IndexService indexService(std::move(faissMethods));
         knn_jni::faiss_wrapper::InsertToIndex(&jniUtil, env, idsJ, vectorsAddressJ, dimJ, indexAddress, threadCount, &indexService);
-    } catch (...) {
+    }
+    catch (const faiss::FaissException& e) {
+        std::string errormsg = e.msg;
+        std::size_t found = errormsg.find("computation interrupted");
```
kotwanikunal, line 99: Is there a better way than checking for the error message regex? 
My guess is tests will catch the issue during upgrades but if there is a specific exception we can catch - it will be more robust.

src/main/java/org/apache/lucene/index/LucenePackagePrivateCaller.java
```diff
@@ -0,0 +1,54 @@
...
+public class LucenePackagePrivateCaller {
+
+    public static Object callPrivateFieldWithMethod(Class<?> clz, String fieldName, String methodName, Object called) {
+        return AccessController.doPrivileged((PrivilegedAction<Object>) () -> {
+            try {
+                Field field = clz.getDeclaredField(fieldName);
+                field.setAccessible(true);
+                return callMethod(field.getType(), methodName, null, field.get(called), null);
+            } catch (Exception e) {
+                log.error("callPrivateFieldWithMethod", e);
```
kotwanikunal: Please update all the log statements in this class to be more usable.

Something along the lines of 
```
log.error("Unable to access private field with methods", e)
```

src/main/java/org/apache/lucene/index/KNNMergeHelper.java
```diff
@@ -0,0 +1,33 @@
...
+            try {
+                Object mergeObject = LucenePackagePrivateCaller.callPrivateFieldWithMethod(
+                    ConcurrentMergeScheduler.MergeThread.class,
+                    "merge",
+                    "isAborted",
+                    mergeThread
+                );
+                return ((Boolean) mergeObject).booleanValue();
+            } catch (RuntimeException e) {
+                return false;
```
kotwanikunal: Please log exceptions here.

src/main/java/org/apache/lucene/index/LucenePackagePrivateCaller.java
```diff
@@ -0,0 +1,54 @@
...
+
+import lombok.extern.log4j.Log4j2;
+
+import java.lang.reflect.Field;
+import java.lang.reflect.Method;
+import java.security.AccessController;
+import java.security.PrivilegedAction;
+
+@Log4j2
+public class LucenePackagePrivateCaller {
```
kotwanikunal: The privileged access here is unnecessary. 
I added a simpler reflection mechanism as a comment within KNNMergeHelper. Please take a look

src/main/java/org/apache/lucene/index/KNNMergeHelper.java
```diff
@@ -0,0 +1,33 @@
...
+
+package org.apache.lucene.index;
+
+import lombok.extern.log4j.Log4j2;
+
+@Log4j2
+public class KNNMergeHelper {
+
+    private KNNMergeHelper() {}
+
```
kotwanikunal: Given the accessors are package private ([here](https://github.com/apache/lucene/blob/branch_10_2/lucene/core/src/java/org/apache/lucene/index/ConcurrentMergeScheduler.java#L704-L706)) - we can simplify the reflection approach and the`LucenePackagePrivate` class can be skipped with  - 

```
private static final Field MERGE_FIELD;

static {
   try {
       MERGE_FIELD = ConcurrentMergeScheduler.MergeThread.class.getDeclaredField("merge");
       MERGE_FIELD.setAccessible(true);
   } catch (NoSuchFieldException e) {
       throw new RuntimeException("Failed to initialize", e);
   }
}

public static boolean isMergeAborted() {
   Thread t = Thread.currentThread();
   if (t instanceof ConcurrentMergeScheduler.MergeThread) {
       try {
           MergePolicy.OneMerge merge = (MergePolicy.OneMerge) MERGE_FIELD.get(t);
           return merge.isAborted();
       } catch (IllegalAccessException e) {
           return false;
       }
   }
   return false;
}
```

luyuncheng: LGTM Thx

jni/src/org_opensearch_knn_jni_FaissService.cpp
luyuncheng: i prefer to catch specific exception name, but in https://github.com/facebookresearch/faiss/blob/32a3912e4303f5f92b8060b2428d20314cc95159/faiss/IndexHNSW.cpp#L187-L189 faiss only throw `FaissException` 
```
            if (interrupt) {
                FAISS_THROW_MSG("computation interrupted");
            }
```

so i have the only way to do is check the error message. if there is any good ideas pls let me know

src/main/java/org/apache/lucene/index/KNNMergeHelper.java
luyuncheng: LGTM

src/main/java/org/opensearch/knn/jni/FaissService.java
```diff
@@ -26,12 +28,23 @@
...
     static {
         KNNLibraryLoader.loadFaissLibrary();
         initLibrary();
         KNNEngine.FAISS.setInitialized(true);
+
+        try {
+            KNNMergeHelper.class.getMethod("isMergeAborted");
+            KNNMergeHelper.isMergeAborted();
+            setMergeInterruptCallback();
+            testMergeInterruptCallback();
```
shatejas: Whats the intent behind this testMergeInterruptCallback()? We are ignoring the error right? 

Can't we rely on ITs somehow

luyuncheng: > Can't we rely on ITs somehow

LGTM, i'll move it into integration tests

luyuncheng:
> One of the solution is to override InterruptCallback#check to be no-op, since indexing is using [InterrupCallback#is_interrupted](https://github.com/facebookresearch/faiss/blob/5616caad4cea4e8326cdef90aa177bd618af8531/faiss/IndexHNSW.cpp#L180). The caveat is that if check is used somewhere in indexing that case will be missed.

@shatejas  yes, correctly, we need  override `InterruptCallback#check` thanks for your valuable ideas !

shatejas (approved):

src/main/java/org/apache/lucene/index/MergeAbortChecker.java
```diff
@@ -0,0 +1,76 @@
...
+    private static final String MERGE_FIELD_NAME = "merge";
+
+    static {
+        try {
+            MERGE_FIELD = ConcurrentMergeScheduler.MergeThread.class.getDeclaredField(MERGE_FIELD_NAME);
+            MERGE_FIELD.setAccessible(true);
+        } catch (NoSuchFieldException e) {
+            throw new RuntimeException("Failed to initialize MergeAbortChecker: merge field not found", e);
+        }
+    }
```
navneet1v, line 42: The exception throwing from piece of code can lead to boot time failures if ConcurrentMergeScheduler.MergeThread.class.MERGE_FIELD_NAME variable is not found. The reason for not finding this variable can be change in name of the variable etc. Can we log the error rather than throwing exception during the boot of OpenSearch? Also, handle it properly during in `isMergeAborted` function?

src/main/java/org/apache/lucene/index/MergeAbortChecker.java
```diff
@@ -0,0 +1,76 @@
...
+
+    /**
+     * Private constructor to prevent instantiation of utility class.
+     */
+    private MergeAbortChecker() {}
```
navneet1v: nit pick: use Lombok @NoArgsConstructor(access = PRIVATE)

src/main/java/org/apache/lucene/index/MergeAbortChecker.java
```diff
@@ -0,0 +1,76 @@
...
+
+    /**
+     * Checks if the current thread is a merge thread and if its merge operation has been aborted.
+     *
+     * <p>This method uses reflection to access the private merge field from
+     * {@link ConcurrentMergeScheduler.MergeThread} and checks if the associated
```
navneet1v, line 48: Question for my understanding: What other type of threads we can encounter here?

src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexWriter.java
```diff
@@ -121,9 +122,17 @@ public void mergeIndex(final Supplier<KNNVectorValues<?>> knnVectorValuesSupplie
...
+        try {
+            startMergeStats(totalLiveDocs, bytesPerVector);
+            buildAndWriteIndex(knnVectorValuesSupplier, totalLiveDocs, false);
+            endMergeStats(totalLiveDocs, bytesPerVector);
+        } catch (IndexBuildAbortedException ex) {
+            log.warn("Merge Aborted", ex);
+            throw new MergePolicy.MergeAbortedException("KNN Merge aborted.");
+        } catch (Exception ex) {
+            log.error("Merge exception {}", ex.getMessage());
+        }
```
navneet1v, line 135: lets not eat the exception trace here and improve the error message. Below is a suggestion. See if we can add more details for better debugging. 

```
log.error("Merge exception happened for field {}", fieldInfo.name, ex);
```

src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexWriter.java
```diff
@@ -121,9 +122,17 @@ public void mergeIndex(final Supplier<KNNVectorValues<?>> knnVectorValuesSupplie
...
-        startMergeStats(totalLiveDocs, bytesPerVector);
-        buildAndWriteIndex(knnVectorValuesSupplier, totalLiveDocs, false);
-        endMergeStats(totalLiveDocs, bytesPerVector);
+
+        try {
+            startMergeStats(totalLiveDocs, bytesPerVector);
+            buildAndWriteIndex(knnVectorValuesSupplier, totalLiveDocs, false);
+            endMergeStats(totalLiveDocs, bytesPerVector);
+        } catch (IndexBuildAbortedException ex) {
+            log.warn("Merge Aborted", ex);
```
navneet1v:
```suggestion
            log.warn("Merge Aborted for field {}",fieldInfo.name,  ex);
```

src/main/java/org/opensearch/knn/jni/FaissService.java
```diff
@@ -26,12 +28,22 @@
...
+@Log4j2
 class FaissService {
 
     static {
         KNNLibraryLoader.loadFaissLibrary();
         initLibrary();
         KNNEngine.FAISS.setInitialized(true);
+
+        try {
+            MergeAbortChecker.class.getMethod("isMergeAborted");
```
navneet1v: why we need this?

src/main/java/org/opensearch/knn/jni/FaissService.java
```diff
@@ -26,12 +28,22 @@
...
+            MergeAbortChecker.isMergeAborted();
+            setMergeInterruptCallback();
+        } catch (Exception e) {
+            // Ignore merge abort callback
+            log.error("Init Faiss Service Error" + e);
```
navneet1v:
```suggestion
            log.warn("Unable to add the mergeAbortChecker during Faiss Initialization", e);
```

navneet1v (commented):
Hi @luyuncheng 
thanks for raising the PR and following up on this. I have added some comments which is mainly on the robustness and debugability of the code. Sorry I wasn't able to review it last week since I was on leave. Hopefully we can close out this PR soon. I am pretty excited for this feature.

src/main/java/org/apache/lucene/index/MergeAbortChecker.java
luyuncheng: i am not sure about when set to @@NoArgsConstructor for Inherited classes, here i want it to be static singleton

src/main/java/org/apache/lucene/index/MergeAbortChecker.java
luyuncheng: yes, when refresh thread check for abort.

src/main/java/org/opensearch/knn/jni/FaissService.java
luyuncheng: LGTM

src/main/java/org/opensearch/knn/jni/FaissService.java
luyuncheng: LGTM, remove it

src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexWriter.java
luyuncheng: LGTM

src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexWriter.java
luyuncheng: LGTM

src/main/java/org/apache/lucene/index/MergeAbortChecker.java
luyuncheng: LGTM

shatejas (approved):

Vikasht34 (approved):
LGTM!!

navneet1v:
test coverage is low and I think its causing [codecov/patch](https://github.com/opensearch-project/k-NN/pull/2529/checks?check_run_id=62951597494) to fail

PR #2533, opened 2025-02-18
Fix derived source for binary and byte vectors

jmazanec15:
### Description
For binary and byte vectors, for derived source, we were not formatting them before adding them back to the source. Thus, they were binary strings in the source. This change fixes this formatting to format them as ints before adding back.

### Related Issues

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/codec/derivedsource/AbstractPerFieldDerivedVectorInjector.java
+++ b/src/main/java/org/opensearch/knn/index/codec/derivedsource/AbstractPerFieldDerivedVectorInjector.java
@@ -1,0 +1,39 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.knn.index.codec.derivedsource;
+
+import lombok.extern.log4j.Log4j2;
+import org.apache.lucene.index.FieldInfo;
+import org.apache.lucene.util.BytesRef;
+import org.opensearch.knn.common.FieldInfoExtractor;
+import org.opensearch.knn.index.VectorDataType;
+import org.opensearch.knn.index.mapper.KNNVectorFieldMapperUtil;
+import org.opensearch.knn.index.vectorvalues.KNNVectorValues;
+
+import java.io.IOException;
+
+@Log4j2
+abstract class AbstractPerFieldDerivedVectorInjector implements PerFieldDerivedVectorInjector {
+    /**
+     * Utility method for formatting the vector values based on the vector data type. KNNVectorValues must be advanced
+     * to the correct position.
+     *
+     * @param fieldInfo fieldinfo for the vector field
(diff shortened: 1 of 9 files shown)


Vikasht34 (approved):
LGTM

src/main/java/org/opensearch/knn/index/codec/derivedsource/AbstractPerFieldDerivedVectorInjector.java
```diff
@@ -0,0 +1,39 @@
...
+     *
+     * @param fieldInfo fieldinfo for the vector field
+     * @param vectorValues vector values of the field. getVector or getConditionalVector should return expected vector.
+     * @return vector formatted based on the vector data type
+     * @throws IOException if unable to deserialize stored vector
+     */
+    protected Object formatVector(FieldInfo fieldInfo, KNNVectorValues<?> vectorValues) throws IOException {
+        Object vectorValue = vectorValues.getVector();
+        // If the vector value is a byte[], we must deserialize
+        if (vectorValue instanceof byte[]) {
```
navneet1v, line 32: can we use the datatype of the field here, rather than instance of check on byte[].

jmazanec15: We need a byte[] in order to deserialize, so this check is required. In terms of displaying, deserializeStoredVector takes the vectorDataType, so we can be sure that it will format it properly.

src/main/java/org/opensearch/knn/index/codec/derivedsource/AbstractPerFieldDerivedVectorInjector.java
```diff
@@ -0,0 +1,39 @@
...
+            BytesRef vectorBytesRef = new BytesRef((byte[]) vectorValue);
+            VectorDataType vectorDataType = FieldInfoExtractor.extractVectorDataType(fieldInfo);
+            return KNNVectorFieldMapperUtil.deserializeStoredVector(vectorBytesRef, vectorDataType);
```
navneet1v, line 35: Can you elaborate why we need to do this? I am trying to understand this like why we need it, since we already have byte[]

jmazanec15: Sure, this is what the IT I added looks like before:
```
  2> java.lang.AssertionError: Docs do not match: 1 expected:<{test_vector=[115, -43, 26, -69, -40, -100, -72, 25, 111, 14, -5, 104, -110, -7, 77, 104]}> but was:<{test_vector=c9Uau9icuBlvDvtokvlNaA==}>
```

Basically, source is expected to be an int array, but because we are adding a byte array, it gets serialized as a byte string

navneet1v (approved):

Issue #1467, opened 2024-02-07
[FEATURE]


nknize:
**Is your feature request related to a problem?**
Core OpenSearch does not support Vector types as a first class field. The correlation engine has a `CorrelationVectorFieldMapper` that uses Lucene's `KNNFloatVectorField`. 

**What solution would you like?**
A discussion around making `vector` field type as a first class citizen in core. We've discussed this before in "person" but I don't know if we have an issue around it. I don't think there's a reason to not have Lucene vector fields and HNSW backed KNN search as a core feature and leverage the OpenSearch kNN plugin as an optional accelerator using alternative native options like FAISS or nmslib? 

**What alternatives have you considered?**
Leave as is if there is a compelling reason to keep this base Lucene capability integration in a separate downstream plugin.

**Do you have any additional context?**

heemin32:
If we only move lucene vector field to core, there would be two different APIs: one for lucene engine and another for native engines.

nknize:
> If we only move lucene vector field to core, there would be two different APIs: one for lucene engine and another for native engines.

Why?

heemin32:
> > If we only move lucene vector field to core, there would be two different APIs: one for lucene engine and another for native engines.
> 
> Why?

I thought the knn plugin API applies to both lucene and native engine but they are only for native engines. So, the previous comment is not correct. If lucene ever support model training feature for other algorithm in the future, then there will be two different APIs.

nknize:
> I thought the knn plugin API applies to both lucene and native engine but they are only for native engines.

++ sounds good. So then I propose a `vector` field type in core and have the k-NN plugin extend the FieldMapper to add the native field type and queries needed to implement fiass, nmslib, others...

nknize:
If there's no opposition then I'll move this issue to core because it'll essentially be a no-op for kNN. kNN can decide in the future whether to simplify the vector field type API.

heemin32:
The index setting and mapping is shared between lucene engine and native engine currently. The `engine` parameter determine if it is lucene engine or native engine. Not sure we can keep the same API format after moving the lucene engine into core.
 
```
PUT my-knn-index-1
{
  "settings": {
    "index": {
      "knn": true,
      "knn.algo_param.ef_search": 100
    }
  },
  "mappings": {
    "properties": {
        "my_vector1": {
          "type": "knn_vector",
          "dimension": 2,
          "method": {
            "name": "hnsw",
            "space_type": "l2",
[...]

PR #2547, opened 2025-02-20
Support custom codecs via stored fields.

### LINKED ISSUE #2414: [FEATURE] Using KNN codec in conjunction with custom codec
One of our customers uses the following index settings, which combines the KNN codec with the zstd codec.
```
"settings": {
    "index.knn": true,
    "index.codec": "zstd"
 }
```
Everything works fine when the shard is running all the time. However, when the index is recovered after a server reboot, the following exception is thrown.
```
java.lang.IllegalStateException: missing value for Lucene90StoredFieldsFormat.mode for segment: _3
        at org.apache.lucene.codecs.lucene90.Lucene90StoredFieldsFormat.fieldsReader(Lucene90StoredFieldsFormat.java:133) ~[lucene-core-9.7.0.jar:9.7.0 ccf4b198ec328095d45d2746189dc8ca633e8bcf - 2023-06-21 11:48:16]
        at org.apache.lucene.index.SegmentCoreReaders.<init>(SegmentCoreReaders.java:138) ~[lucene-core-9.7.0.jar:9.7.0 ccf4b198ec328095d45d2746189dc8ca633e8bcf - 2023-06-21 11:48:16]
        at org.apache.lucene.index.SegmentReader.<init>(SegmentReader.java:92) ~[lucene-core-9.7.0.jar:9.7.0 ccf4b198ec328095d45d2746189dc8ca633e8bcf - 2023-06-21 11:48:16]
        at org.apache.lucene.index.ReadersAndUpdates.getReader(ReadersAndUpdates.java:180) ~[lucene-core-9.7.0.jar:9.7.0 ccf4b198ec328095d45d2746189dc8ca633e8bcf - 2023-06-21 11:48:16]
        at org.apache.lucene.index.ReadersAndUpdates.getReadOnlyClone(ReadersAndUpdates.java:222) ~[lucene-core-9.7.0.jar:9.7.0 ccf4b198ec328095d45d2746189dc8ca633e8bcf - 2023-06-21 11:48:16]
        at org.apache.lucene.index.IndexWriter.lambda$getReader$0(IndexWriter.java:536) ~[lucene-core-9.7.0.jar:9.7.0 ccf4b198ec328095d45d2746189dc8ca633e8bcf - 2023-06-21 11:48:16]
        at org.apache.lucene.index.StandardDirectoryReader.open(StandardDirectoryReader.java:138) ~[lucene-core-9.7.0.jar:9.7.0 ccf4b198ec328095d45d2746189dc8ca633e8bcf - 2023-06-21 11:48:16]
        at org.apache.lucene.index.IndexWriter.getReader(IndexWriter.java:598) ~[lucene-core-9.7.0.jar:9.7.0 ccf4b198ec328095d45d2746189dc8ca633e8bcf - 2023-06-21 11:48:16]
        at org.apache.lucene.index.DirectoryReader.open(DirectoryReader.java:112) ~[lucene-core-9.7.0.jar:9.7.0 ccf4b198ec328095d45d2746189dc8ca633e8bcf - 2023-06-21 11:48:16]
        at org.apache.lucene.index.DirectoryReader.open(DirectoryReader.java:91) ~[lucene-core-9.7.0.jar:9.7.0 ccf4b198ec328095d45d2746189dc8ca633e8bcf - 2023-06-21 11:48:16]
        at org.opensearch.index.engine.InternalEngine.createReaderManager(InternalEngine.java:656) ~[opensearch-2.9.0.jar:2.9.0]
        at org.opensearch.index.engine.InternalEngine.<init>(InternalEngine.java:322) ~[opensearch-2.9.0.jar:2.9.0]
        at org.opensearch.index.engine.InternalEngine.<init>(InternalEngine.java:223) ~[opensearch-2.9.0.jar:2.9.0]
        at org.opensearch.index.engine.InternalEngineFactory.newReadWriteEngine(InternalEngineFactory.java:43) ~[opensearch-2.9.0.jar:2.9.0]
        at org.opensearch.index.shard.IndexShard.innerOpenEngineAndTranslog(IndexShard.java:2362) ~[opensearch-2.9.0.jar:2.9.0]
        at org.opensearch.index.shard.IndexShard.innerOpenEngineAndTranslog(IndexShard.java:2324) ~[opensearch-2.9.0.jar:2.9.0]
        at org.opensearch.index.shard.IndexShard.openEngineAndRecoverFromTranslog(IndexShard.java:2294) ~[opensearch-2.9.0.jar:2.9.0]
        at org.opensearch.index.shard.StoreRecovery.internalRecoverFromStore(StoreRecovery.java:630) ~[opensearch-2.9.0.jar:2.9.0]
        at org.opensearch.index.shard.StoreRecovery.lambda$recoverFromStore$0(StoreRecovery.java:115) ~[opensearch-2.9.0.jar:2.9.0]
        at org.opensearch.action.ActionListener.completeWith(ActionListener.java:342) ~[opensearch-2.9.0.jar:2.9.0]
        ... 8 more
```

This is because when indices are created, we use the parametric constructor to initialize the KNN codec, so we can get the correct delegate code. 

https://github.com/opensearch-project/k-NN/blob/f5cf255286239dbf48742951f38ac68d80e5f7cb/src/main/java/org/opensearch/knn/index/codec/KNNCodecService.java#L32-L34

And when indices are recovering, we initialize the KNN codec using a parameterless constructor, so the delegate codec we get will always be the default one, which doesn't match the codec used for writing, and thus throws an exception.

https://github.com/opensearch-project/k-NN/blob/f5cf255286239dbf48742951f38ac68d80e5f7cb/src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNN990Codec.java#L29-L31

We can put the delegate codec name to the segment info attribute during [writing `SegmentInfo` data](https://github.com/apache/lucene/blob/fb48d0da4991906c0a6598388c1ac29f3ace15a9/lucene/core/src/java/org/apache/lucene/codecs/SegmentInfoFormat.java#L53). And when getting stored fields reader, we first read the delegate codec name from the `SegmentInfo` and then use the correct delegate codec to generate the stored fields reader.
```
private final StoredFieldsFormat storedFieldsFormat = new StoredFieldsFormat() {
        @Override
        public StoredFieldsReader fieldsReader(Directory dir, SegmentInfo si, FieldInfos fn, IOContext context) throws IOException {
            String name = si.getAttribute(DELEGATE_KEY);
            if (name == null) {
                return delegate.storedFieldsFormat().fieldsReader(dir, si, fn, context);
            }
            return Codec.forName(name).storedFieldsFormat().fieldsReader(dir, si, fn, context);
        }
}
```
Note: since we can't get a delegate for `SegmentInfoFormat`, we must use the default delegate codec's `SegmentInfoFormat`.

If we don't want the user to use a custom codec, then we can check the `index.codec` configuration when fetching the codec service factory and throw an exception if it's not `default` or `best_compression`.  Instead of having unintended behavior during runtime.

https://github.com/opensearch-project/k-NN/blob/f5cf255286239dbf48742951f38ac68d80e5f7cb/src/main/java/org/opensearch/knn/plugin/KNNPlugin.java#L285-L290

jmazanec15:
### Description
Following up on discussion in #2414. @bugmakerrrrrr had idea to serialize info about delegate as segment attribute. Something similar is done in https://github.com/opensearch-project/custom-codecs/blob/2.19.0.0/src/main/java/org/opensearch/index/codec/customcodecs/Lucene912CustomStoredFieldsFormat.java#L95-L100.

One short coming with this approach is that it only works for stored fields formats. We could generalize to write in segmentinfoformat. However, with that, I think we would need to create a custom format for all formats, which seems a bit heavy. So, just focusing on segmentinfoformat for now.

Still need to write some tests, but wanted to see what people thought early.

### Related Issues
Resolves #2414 

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/codec/KNN10010Codec/KNN10010Codec.java
+++ b/src/main/java/org/opensearch/knn/index/codec/KNN10010Codec/KNN10010Codec.java
@@ -104,6 +104,11 @@ public class KNN10010Codec extends FilterCodec {
             }
             return null;
         }));
-        return new DerivedSourceStoredFieldsFormat(delegate.storedFieldsFormat(), derivedSourceReadersSupplier, mapperService);
+        return new DerivedSourceStoredFieldsFormat(
+            delegate.storedFieldsFormat(),
+            derivedSourceReadersSupplier,
+            mapperService,
+            delegate.getName()
+        );
     }
 }
--- a/src/main/java/org/opensearch/knn/index/codec/KNN9120Codec/DerivedSourceStoredFieldsFormat.java
+++ b/src/main/java/org/opensearch/knn/index/codec/KNN9120Codec/DerivedSourceStoredFieldsFormat.java
@@ -6,6 +6,7 @@
 package org.opensearch.knn.index.codec.KNN9120Codec;
 
 import lombok.AllArgsConstructor;
+import org.apache.lucene.codecs.Codec;
 import org.apache.lucene.codecs.StoredFieldsFormat;
 import org.apache.lucene.codecs.StoredFieldsReader;
 import org.apache.lucene.codecs.StoredFieldsWriter;
@@ -32,15 +33,26 @@ import static org.opensearch.knn.common.KNNConstants.DERIVED_VECTOR_FIELD_ATTRIB
 @AllArgsConstructor
 public class DerivedSourceStoredFieldsFormat extends StoredFieldsFormat {
 
+    private static final String DELEGATE_CODEC_KEY = "knn_delegate_codec";
+
     private final StoredFieldsFormat delegate;
     private final DerivedSourceReadersSupplier derivedSourceReadersSupplier;
     // IMPORTANT Do not rely on this for the reader, it will be null if SPI is used
     @Nullable
     private final MapperService mapperService;
+    // IMPORTANT Do not rely on this for the reader, it will be null if SPI is used
+    @Nullable
+    private final String delegateCodecName;
 
     @Override
     public StoredFieldsReader fieldsReader(Directory directory, SegmentInfo segmentInfo, FieldInfos fieldInfos, IOContext ioContext)
         throws IOException {
+        StoredFieldsFormat delegateFromWriting = delegate;
+        if (segmentInfo.getAttribute(DELEGATE_CODEC_KEY) != null) {
+            String delegateCodecName = segmentInfo.getAttribute(DELEGATE_CODEC_KEY);
+            delegateFromWriting = Codec.forName(delegateCodecName).storedFieldsFormat();
+        }
+
         List<FieldInfo> derivedVectorFields = null;
         for (FieldInfo fieldInfo : fieldInfos) {
             if (DERIVED_VECTOR_FIELD_ATTRIBUTE_TRUE_VALUE.equals(fieldInfo.attributes().get(DERIVED_VECTOR_FIELD_ATTRIBUTE_KEY))) {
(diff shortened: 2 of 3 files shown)


navneet1v:
@jmazanec15 and @bugmakerrrrrr this looks like a very small change, but I want us to think little bit on the scenario when k-NN doesn't provide its own codec. This I am envisioning will happen once we move KNNVector field to core: https://github.com/opensearch-project/k-NN/issues/1467. Once the KNNVectorField move to core the codec will also move there and if Core is giving the codec then I am little bit concerned if this approach will acceptable in core? and also how the whole thing will play out?

jmazanec15:
As long as we have a custom codec, we need to persist delegate information in order to properly create delegate on case of loading from SPI. I dont think there is a better way.

If there is a per field vector format extension point, we will still need the custom stored fields format to support derived source, so we will still need the custom codec. Further, I think for all of 3.x we will need the custom codec anyway to support upgrades from 2.x.

bugmakerrrrrr:
> As long as we have a custom codec, we need to persist delegate information in order to properly create delegate on case of loading from SPI. I dont think there is a better way.
> 
> If there is a per field vector format extension point, we will still need the custom stored fields format to support derived source, so we will still need the custom codec. Further, I think for all of 3.x we will need the custom codec anyway to support upgrades from 2.x.

+1

Issue #2518, opened 2025-02-11
Remote Vector Index Build Component -- Client Low Level Design


owenhalpert:
## Overview

The Remote Vector Service Client will handle all communication with the Remote Vector Index Build Service. Its main responsibilities, as outlined below, are to call the remote build service to begin index construction once the vector files have been uploaded, periodically monitor the build requests, and pass the completed index build path along to the index writer after construction is complete, handling any errors along the way.
* * *

#### Http Client

The HttpClient we will start with is the ApacheHttpClient. Holistically speaking, the Apache client is well-supported, well-documented, and widely used. Apache offers a synchronous client, which matches the overall logic of segment merge/flush, where the thread is dedicated to the task until completion. An asynchronous API in this case would mean the API returns a future/promise of the result and doesn’t block the calling thread. The server-side implementation for an asynchronous API adds a lot more complexity. 

ApacheHttpClient is already a dependency from OpenSearch Core through the AWS SDK, so we can transitively use the dependency from OpenSearch Core instead of adding it to k-NN, leaving versioning logic to Core. The included HttpClient implementations are already used by the [`opensearch/ml-commons`](https://github.com/opensearch-project/ml-commons/blob/06d17424d65de18b189b7c259325128b09f3a9ec/ml-algorithms/src/main/java/org/opensearch/ml/engine/httpclient/MLHttpClientFactory.java#L26-L38) package, and in Core OpenSearch by the [S3 Service](https://github.com/opensearch-project/OpenSearch/blob/main/plugins/repository-s3/src/main/java/org/opensearch/repositories/s3/S3Service.java#L258), [AWS KMS](https://github.com/opensearch-project/OpenSearch/blob/main/plugins/crypto-kms/src/main/java/org/opensearch/crypto/kms/KmsService.java#L88), and more. The[`ml-connectors`](https://github.com/opensearch-project/ml-commons/blob/06d17424d65de18b189b7c259325128b09f3a9ec/ml-algorithms/src/main/java/org/opensearch/ml/engine/httpclient/MLHttpClientFactory.java#L26-L38) use case is to create connectors to outside services (e.g. OpenAI). By following the same patterns and utilizing its established dependencies and protocols, we stay consistent with current standards and are aligned on proven security practices. 

[...]

Issue #2393, opened 2025-01-14
Remote Vector Index Build Component — Remote Vector Service Client


jed326:
See https://github.com/opensearch-project/k-NN/issues/2391 for background information

## Overview
Following up on the RFCs, this is the first part of the low-level design for the Vector Index Build Component. The Vector Index Build Component is a logical component we further split into 2 subcomponents and their respective responsibilities:

1. Object Store I/O Component
    1. Upload flat vectors to Object Store
    2. Download graph file from object store
2. Remote Vector Service Client Component
    1. Signal to Remote Vector Index Build Service to begin graph construction after vector files have been uploaded
    2. Receive a signal from Remote Vector Index Build Service to begin graph file download after graph construction is completed

This document contains the low level design for [2] Remote Vector Service Client Component, covering the design for the client the vector engine will use to interact with the remote vector build service as well the workflows associated with using this client. 

## Tenets

The key tenets of the client are straightforward:

1. Signal to the remote vector service that vector blob upload is complete and ready for graph construction
2. Receive a signal from the remote vector service that graph construction is complete and graph download is ready
3. Handle failures so as to not fail the merge/flush operation

High level overview:

![Image](https://github.com/user-attachments/assets/3f3591f6-b39c-407d-9a28-dae4d612faaf)

### 1. [Recommended] REST requests with polling

In this approach we submit the graph build request via REST request and then use REST requests to poll the vector build service for completion status.

Pros:

* Simplest implementation 
* Vector build service itself could still implement a queue to receive the build requests

Cons:

* Higher request count to remote vector build service, need to configure polling interval and smart retry logic
* A state machine would be required to keep track of the build progress

### 2. Persistent connection (gRPC, websockets)

In this approach we open a persistent connection between the OpenSearch cluster and the remote vector build service and keep the connection open until the graph construction is complete.
Pros:

* Lower request count to remote vector build service

Cons:

* Need to build robust reconnect logic. How do we handle the case where vector build completes or fails during a disconnect?
* vector build service is multi-tenant so we may be bottlenecked on the number of concurrent persistent connections to the remote vector build service. Specifically any request queueing may lead to unexpected outcomes.
* If we design a stateless system then it would be difficult to retry only specific actions, for example retrying only the graph upload part if that were to fail
* There is very little data transfer happening through this client

### 3. REST callback

In this approach we submit a graph build request via REST request and expose a REST callback endpoint for the remote vector build service to notify on when the graph construction is complete
Pros:

[...]

PR #2548, opened 2025-02-20
Introduce RemoteIndexClient

owenhalpert:
### Description
First PR for #2518 

This PR introduces the initial implementation of the Remote Index client, which is awaiting implementation on the following:
- [Programmatic build request construction](https://github.com/owenhalpert/k-NN/blob/57290c18e2eafcc5eba4d48574da173befa2282a/src/main/java/org/opensearch/knn/index/remote/RemoteIndexClient.java#L177)
- Metric collection
- Complete unit testing

### Related Issues
#2391 Meta Issue
#2393 HLD by @jed326 
#2518 My LLD

### Check List

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -321,7 +321,11 @@ dependencies {
     api "net.java.dev.jna:jna-platform:5.13.0"
     // OpenSearch core is using slf4j 1.7.36. Therefore, we cannot change the version here.
     implementation 'org.slf4j:slf4j-api:1.7.36'
-
+    api "org.apache.httpcomponents.client5:httpclient5:${versions.httpclient5}"
+    api "org.apache.httpcomponents.core5:httpcore5:${versions.httpcore5}"
+    api "org.apache.httpcomponents.core5:httpcore5-h2:${versions.httpcore5}"
+    api "com.fasterxml.jackson.core:jackson-databind:${versions.jackson_databind}"
+    api "com.fasterxml.jackson.core:jackson-annotations:${versions.jackson}"
     zipArchive group: 'org.opensearch.plugin', name:'opensearch-security', version: "${opensearch_build}"
 }
 
--- a/src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
+++ b/src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
@@ -11,10 +11,13 @@ import org.apache.lucene.index.SegmentWriteState;
 import org.opensearch.common.StopWatch;
 import org.opensearch.common.annotation.ExperimentalApi;
 import org.opensearch.index.IndexSettings;
+import org.opensearch.knn.common.KNNConstants;
 import org.opensearch.knn.common.featureflags.KNNFeatureFlags;
 import org.opensearch.knn.index.KNNSettings;
 import org.opensearch.knn.index.codec.nativeindex.NativeIndexBuildStrategy;
 import org.opensearch.knn.index.codec.nativeindex.model.BuildIndexParams;
+import org.opensearch.knn.index.remote.RemoteBuildRequest;
+import org.opensearch.knn.index.remote.RemoteIndexClient;
 import org.opensearch.knn.index.vectorvalues.KNNVectorValues;
 import org.opensearch.repositories.RepositoriesService;
 import org.opensearch.repositories.Repository;
@@ -22,10 +25,13 @@ import org.opensearch.repositories.RepositoryMissingException;
 import org.opensearch.repositories.blobstore.BlobStoreRepository;
 
 import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
 import java.util.function.Supplier;
 
 import static org.opensearch.knn.index.KNNSettings.KNN_INDEX_REMOTE_VECTOR_BUILD_SETTING;
 import static org.opensearch.knn.index.KNNSettings.KNN_REMOTE_VECTOR_REPO_SETTING;
+import static org.opensearch.knn.index.KNNSettings.state;
 
 /**
  * This class orchestrates building vector indices. It handles uploading data to a repository, submitting a remote
@@ -54,7 +60,7 @@ public class RemoteIndexBuildStrategy implements NativeIndexBuildStrategy {
      * @return whether to use the remote build feature
      */
     public static boolean shouldBuildIndexRemotely(IndexSettings indexSettings) {
-        String vectorRepo = KNNSettings.state().getSettingValue(KNN_REMOTE_VECTOR_REPO_SETTING.getKey());
+        String vectorRepo = state().getSettingValue(KNN_REMOTE_VECTOR_REPO_SETTING.getKey());
         return KNNFeatureFlags.isKNNRemoteVectorBuildEnabled()
             && indexSettings.getValue(KNN_INDEX_REMOTE_VECTOR_BUILD_SETTING)
             && vectorRepo != null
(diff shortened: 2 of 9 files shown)


src/main/java/org/opensearch/knn/index/remote/RemoteIndexClient.java
```diff
@@ -0,0 +1,248 @@
...
+ * InterruptedExceptions will cause a fallback to local CPU build.
+ */
+@Log4j2
+public class RemoteIndexClient {
+    private static RemoteIndexClient INSTANCE;
+    private volatile CloseableHttpClient httpClient;
+
+    public static final int MAX_RETRIES = 1; // 2 total attempts
+    public static final long BASE_DELAY_MS = 100;
+    public static final long INITIAL_DELAY_MS = 10;
```
owenhalpert, line 47: These are currently set low for testing, though we haven't thought on strong defaults for these anyway.

build.gradle
```diff
@@ -321,7 +321,11 @@ dependencies {
     api "net.java.dev.jna:jna-platform:5.13.0"
     // OpenSearch core is using slf4j 1.7.36. Therefore, we cannot change the version here.
     implementation 'org.slf4j:slf4j-api:1.7.36'
-
+    api 'org.apache.httpcomponents.client5:httpclient5:5.4.1'
+    api 'org.apache.httpcomponents.core5:httpcore5:5.3.2'
+    api 'org.apache.httpcomponents.core5:httpcore5-h2:5.3.2'
```
jed326, line 326: Don't hardcode these, this should also be available in versions as it's probably used by other plugins

src/main/java/org/opensearch/knn/index/remote/RemoteIndexClient.java
```diff
@@ -0,0 +1,250 @@
...
+    * @throws InterruptedException on a timeout or terminal failure from the server to trigger local CPU fallback
+    */
+    @SuppressWarnings("BusyWait")
+    public String awaitVectorBuild(String jobId) throws InterruptedException, IOException, URISyntaxException {
+        long startTime = System.currentTimeMillis();
+        long timeout = ((TimeValue) (KNNSettings.state().getSettingValue(KNNSettings.KNN_REMOTE_BUILD_SERVICE_TIMEOUT))).getMillis();
+        long pollInterval = ((TimeValue) (KNNSettings.state().getSettingValue(KNNSettings.KNN_REMOTE_BUILD_SERVICE_POLL_INTERVAL)))
+            .getMillis();
+
+        Thread.sleep(INITIAL_DELAY_MS);
```
jed326, line 127: I need to look into this some more, but I'm not sure this is the best implementation. Off the top of my head I am thinking of some sort of job scheduler semantic with a latch and we can do latch.await() here or something like that. Also it seems a little dangerous to do this in a Singleton class.

jed326: Also, shouldn't `INITIAL_DELAY_MS` just be `pollInterval`? or maybe some function of it.

src/main/java/org/opensearch/knn/index/remote/RemoteIndexClient.java
```diff
@@ -0,0 +1,250 @@
...
+        }
+        throw new InterruptedException("Build timed out, falling back to CPU build.");
+    }
+
+    /**
+     * Helper method to directly get the status response for a given job ID
+     * @param jobId to check
+     * @return HttpExecuteResponse for the status request
+     */
+    public String getBuildStatus(String jobId) throws IOException, URISyntaxException {
```
jed326, line 163: I think a lot of these methods can be private or protected

src/main/java/org/opensearch/knn/index/remote/RemoteIndexClient.java
owenhalpert: I don't see the two as related. I was thinking of the `INITIAL_DELAY_MS` as a buffer to prevent the client from sending any status requests when we know for sure that the build is still in progress. For example finding a lower bound `t` on the GPU build or calculating this based on workload size, then waiting `0.5t` so to not send any unnecessary status checks.

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
```diff
@@ -88,12 +89,12 @@ public void buildAndWriteIndex(BuildIndexParams indexInfo) throws IOException {
             log.debug("Repository write took {} ms for vector field [{}]", time_in_millis, indexInfo.getFieldName());
 
             stopWatch = new StopWatch().start();
-            submitVectorBuild();
+            String jobId = RemoteIndexClient.getInstance().submitVectorBuild();
```
jed326, line 92: I wonder if it makes sense to build the requests out here and then pass the request itself into `RemoteIndexClient`, else you may need to pass a lot of other parameters through too.

src/main/java/org/opensearch/knn/index/remote/RemoteIndexClient.java
```diff
@@ -0,0 +1,250 @@
...
+    /**
+     * Initialize the httpClient to be used
+     * @return The HTTP Client
+     */
+    private CloseableHttpClient createHttpClient() {
+        return HttpClients.custom()
+            .setRetryStrategy(new RemoteIndexClientRetryStrategy())
+            .setConnectionBackoffStrategy(new DefaultBackoffStrategy())
+            .build();
+    }
```
jed326, line 78: We would need to recreate this client if/when any of the retry strategy settings change right? Let's add a TODO for that

src/main/java/org/opensearch/knn/index/remote/RemoteIndexClient.java
```diff
@@ -0,0 +1,250 @@
...
+    private volatile CloseableHttpClient httpClient;
+
+    public static final int MAX_RETRIES = 1; // 2 total attempts
+    public static final long BASE_DELAY_MS = 1000;
+    public static final long INITIAL_DELAY_MS = 1000;
+
+    private static final ObjectMapper objectMapper = new ObjectMapper();
+
+    // To be used for round-robin task assignment to know which endpoint accepted the given job.
+    private final Map<String, String> jobEndpoints = new ConcurrentHashMap<>();
```
jed326, line 52: I'm not sure about this, there's no need for job A to have access to this and be able to look up the endpoints of all the other jobs, that seems like it's unnecessarily leaking information.

src/main/java/org/opensearch/knn/index/remote/RemoteIndexClient.java
```diff
@@ -0,0 +1,250 @@
...
+                }
+                return EntityUtils.toString(body.getEntity());
+            });
+            if (response != null) {
+                String jobId = getValueFromResponse(response, "job_id");
+                jobEndpoints.put(jobId, endpoints.get(i));
+                return jobId;
+            }
+            i++;
+        }
```
jed326, line 110: Let's abstract out the round robin aspect into a `getEndpoint` method so we can test it more specifically.

Also since `i` is initialized within this method, won't we always start with endpoint `0`? It seems to me if we want to perform round robin across requests then this class needs to track the the list of endpoints in some sort of circular queue, otherwise it's basically just a list of backup endpoints.

src/main/java/org/opensearch/knn/index/remote/RemoteIndexClient.java
jed326: Thinking on it some more I think `Thread.sleep` is fine, but let's make sure we have tests for it across concurrent jobs.

For the other point, the polling interval itself should also be a function of `t` else it's also going to send unnecessary status checks.

jed326 (commented):
Thanks @owenhalpert!

At a high level I think we should separate out the client changes into the following individual PRs:

1. Client skeleton + build request, including polling logic + tests, failure handling
2. Polling + failure handling
3. Metrics

I think there is a lot more testing that needs to be written for all of these and I think it will be much easier to focus on each component individually.

src/main/java/org/opensearch/knn/index/remote/RemoteIndexClient.java
owenhalpert: I tried to implement it such that the 5 settings (endpoint, timeout, poll interval, username, password) were fetched when needed so they could be dynamically updated. We don't have a user setting for `MAX_RETRY` but you are right that the client would need to be rebuilt if we needed to update the retry interval, max retries, or retriable status codes.

src/main/java/org/opensearch/knn/index/remote/RemoteIndexClient.java
owenhalpert: I think we could pass the working endpoint back to RemoteIndexBuildStrategy (along with the job ID) and have it `awaitVectorBuild(acceptedEndpoint, jobID)` and eliminate the need for this map.

Issue #2392, opened 2025-01-14
Remote Vector Index Build Component — Object Store Upload/Download


jed326:
See https://github.com/opensearch-project/k-NN/issues/2391 for background information

## Overview

Following up on the RFCs, this is the first part of the low-level design for the Vector Index Build Component. The Vector Index Build Component is a logical component we further split into 2 subcomponents and their respective responsibilities:

1. Object Store I/O Component
    1. Upload flat vectors to Object Store
    2. Download graph file from object store
2. Remote Vector Service Client Component
    1. Signal to Remote Vector Index Build Service to begin graph construction after vector files have been uploaded
    2. Receive a signal from Remote Vector Index Build Service to begin graph file download after graph construction is completed

This document contains the low level design for [1] Object Store I/O Component, covering how we can upload vectors and download graph files from a remote object store, as well as how we can configure the object store. The low level design for the remote vector service client is in a separate issue.

## Alternatives Considered

The specific problem we are addressing in this design is [1] how to upload vectors to a remote object store from the vector engine and [2] how to download a graph file from a remote object store to the vector engine.

For discussion on high level architectural alternatives see: #2293 

### 1. [Recommended] Integrate Repository Service with Vector Engine

This approach involves consuming the RepositoriesService in the k-NN plugin, which will then use the BlobContainer interface to read/write blobs to the remote repository.

Pros:

* Uses existing repository interface to interact with remote object store, which is a well tested interface that has been around for a long time
* Out of the box support for all existing supported repositories, so no specific implementation for S3, GCP, Azure is needed
* This use case is well tested by remote store, where we write data to repositories not in snapshot format
* Encryption comes “for free” as it already exists for snapshot and remote store repositories

Cons:

* Will be harder to make any vector upload performance improvements specific to a certain object store implementation
* Added complexity in bootstrapping vector repository

#### 2. Custom object store upload/download client

Instead of using the existing interfaces, we build our own custom blob interface for uploading vectors and download graph files. Additionally we will also need to implement all of the object store clients.

Pros:

* Complete control over client implementation / usage making it easy to make any needed performance improvements 

Cons:

* We will need to keep separate object store clients in the k-NN plugin, and we would be responsible for implementing all of the object store clients.
* Encryption at rest will require special handling
* We are duplicating much of the same functionality the repository interface already provides within the k-NN plugin

## Repository Service Integration

[...]

PR #2550, opened 2025-02-21
[Remote Vector Index Build] Add vector data upload implementation to RemoteIndexBuildStrategy

jed326:
### Description
This PR includes implementation + testing for upload vector and doc id blobs to a remote repository with the `RemoteIndexBuildStrategy`. This PR also includes a new `knn.remote_index_build.size_threshold` setting to control the size threshold above which the remote index build service will be used.

---
#### Vector Upload Details
For a detailed explanation on the vector uploads, please see the `Parallel Blob Upload` section from the LLD: https://github.com/opensearch-project/k-NN/issues/2465

At a high level the flow is as follows (Currently only repository-s3 supports parallel blob uploads via this interface, so I will reference S3 directly in the following):
1. KNN Plugin provides an InputStream supplier to asyncBlobUpload (See: `RemoteIndexBuildStrategy#getTransferPartStreamSupplier`)
2. Repository S3 determines the number of parts (`N`) to split the blob into based on repository settings. By default we use 16mb parts.
3. Repository uses the supplier from [1] to create `N` `VectorValueInputStreams` each representing a only a specific part of the underlying `KNNVectorValues`.
4. Repository S3 performs a multi-part upload request which each stream as an upload part. The number of parallel parts depends on the upload priority and threadpool configurations through the repository settings.

Because `KNNVectorValues` is an iterator, we need to create `N` instances of it, otherwise the iterating through all of the vectors is still done sequentially which will greatly reduce the throughput. `VectorValueInputStreams` takes in a constructor parameter to position the head of the input stream at a specific byte offset in the `KNNVectorValues`, including offsets within a given vector. For example, for doc id `X` the vector could be partially read by InputStream part 1 and partially in InputStream part 2.

---

This PR DOES NOT include:
* Checksum implementation, which will come as a separate follow-up PR to focus on the parallel upload approach here
* Prefix pattern optimizations (see: https://github.com/opensearch-project/OpenSearch/issues/12567). This will be re-evaluated as a follow-up

For reference, this is the blob path being used:
```
<BASE_PATH>/<Index UUID>"_vectors"/<uuid>"_"<field name>"_"<segment name".knnvec/knndid
```

### Related Issues

### Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased 3.0](https://github.com/opensearch-project/k-NN/compare/2.x...HEAD)
 ### Features
+* [Remote Vector Index Build] Introduce Remote Native Index Build feature flag, settings, and initial skeleton [#2525](https://github.com/opensearch-project/k-NN/pull/2525)
+* [Remote Vector Index Build] Implement vector data upload and vector data size threshold setting [#2550](https://github.com/opensearch-project/k-NN/pull/2550)
 ### Enhancements
 ### Bug Fixes
 ### Infrastructure
--- a/src/main/java/org/opensearch/knn/index/KNNSettings.java
+++ b/src/main/java/org/opensearch/knn/index/KNNSettings.java
@@ -11,7 +11,6 @@ import org.apache.logging.log4j.Logger;
 import org.opensearch.OpenSearchParseException;
 import org.opensearch.action.admin.cluster.settings.ClusterUpdateSettingsRequest;
 import org.opensearch.action.admin.cluster.settings.ClusterUpdateSettingsResponse;
-import org.opensearch.transport.client.Client;
 import org.opensearch.cluster.metadata.IndexMetadata;
 import org.opensearch.cluster.service.ClusterService;
 import org.opensearch.common.Booleans;
@@ -28,6 +27,7 @@ import org.opensearch.knn.index.util.IndexHyperParametersUtil;
 import org.opensearch.knn.quantization.models.quantizationState.QuantizationStateCacheManager;
 import org.opensearch.monitor.jvm.JvmInfo;
 import org.opensearch.monitor.os.OsProbe;
+import org.opensearch.transport.client.Client;
 
 import java.security.InvalidParameterException;
 import java.util.Arrays;
@@ -96,6 +96,7 @@ public class KNNSettings {
     public static final String KNN_DERIVED_SOURCE_ENABLED = "index.knn.derived_source.enabled";
     public static final String KNN_INDEX_REMOTE_VECTOR_BUILD = "index.knn.remote_index_build.enabled";
     public static final String KNN_REMOTE_VECTOR_REPO = "knn.remote_index_build.vector_repo";
+    public static final String KNN_INDEX_REMOTE_VECTOR_BUILD_THRESHOLD = "index.knn.remote_index_build.size_threshold";
 
     /**
      * Default setting values
@@ -126,6 +127,8 @@ public class KNNSettings {
                                                                                              // 10% of the JVM heap
     public static final Integer KNN_DEFAULT_QUANTIZATION_STATE_CACHE_EXPIRY_TIME_MINUTES = 60;
     public static final boolean KNN_DISK_VECTOR_SHARD_LEVEL_RESCORING_DISABLED_VALUE = false;
+    // TODO: Tune this default value based on benchmarking
+    public static final ByteSizeValue KNN_INDEX_REMOTE_VECTOR_BUILD_THRESHOLD_DEFAULT_VALUE = new ByteSizeValue(50, ByteSizeUnit.MB);
 
     /**
      * Settings Definition
@@ -389,6 +392,15 @@ public class KNNSettings {
     public static final Setting<String> KNN_REMOTE_VECTOR_REPO_SETTING = Setting.simpleString(KNN_REMOTE_VECTOR_REPO, Dynamic, NodeScope);
 
     /**
+     * Index level setting which indicates the size threshold above which remote vector builds will be enabled.
+     */
+    public static final Setting<ByteSizeValue> KNN_INDEX_REMOTE_VECTOR_BUILD_THRESHOLD_SETTING = Setting.byteSizeSetting(
+        KNN_INDEX_REMOTE_VECTOR_BUILD_THRESHOLD,
+        KNN_INDEX_REMOTE_VECTOR_BUILD_THRESHOLD_DEFAULT_VALUE,
+        Dynamic,
+        IndexScope
+    );
+    /**
      * Dynamic settings
      */
     public static Map<String, Setting<?>> dynamicCacheSettings = new HashMap<String, Setting<?>>() {
@@ -550,6 +562,10 @@ public class KNNSettings {
             return KNN_REMOTE_VECTOR_REPO_SETTING;
         }
 
+        if (KNN_INDEX_REMOTE_VECTOR_BUILD_THRESHOLD.equals(key)) {
+            return KNN_INDEX_REMOTE_VECTOR_BUILD_THRESHOLD_SETTING;
+        }
+
         throw new IllegalArgumentException("Cannot find setting by key [" + key + "]");
     }
 
@@ -577,7 +593,8 @@ public class KNNSettings {
             KNN_DISK_VECTOR_SHARD_LEVEL_RESCORING_DISABLED_SETTING,
             KNN_DERIVED_SOURCE_ENABLED_SETTING,
             KNN_INDEX_REMOTE_VECTOR_BUILD_SETTING,
-            KNN_REMOTE_VECTOR_REPO_SETTING
+            KNN_REMOTE_VECTOR_REPO_SETTING,
+            KNN_INDEX_REMOTE_VECTOR_BUILD_THRESHOLD_SETTING
         );
         return Stream.concat(settings.stream(), Stream.concat(getFeatureFlags().stream(), dynamicCacheSettings.values().stream()))
             .collect(Collectors.toList());
--- a/src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexBuildStrategyFactory.java
+++ b/src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexBuildStrategyFactory.java
@@ -7,14 +7,18 @@ package org.opensearch.knn.index.codec.nativeindex;
 
 import org.apache.lucene.index.FieldInfo;
 import org.opensearch.index.IndexSettings;
+import org.opensearch.knn.common.featureflags.KNNFeatureFlags;
 import org.opensearch.knn.index.codec.nativeindex.remote.RemoteIndexBuildStrategy;
 import org.opensearch.knn.index.engine.KNNEngine;
+import org.opensearch.knn.index.vectorvalues.KNNVectorValues;
 import org.opensearch.repositories.RepositoriesService;
 
+import java.io.IOException;
 import java.util.function.Supplier;
 
 import static org.opensearch.knn.common.FieldInfoExtractor.extractKNNEngine;
 import static org.opensearch.knn.common.KNNConstants.MODEL_ID;
+import static org.opensearch.knn.index.codec.util.KNNCodecUtil.initializeVectorValues;
 
 /**
  * Creates the {@link NativeIndexBuildStrategy}
@@ -34,11 +38,18 @@ public final class NativeIndexBuildStrategyFactory {
     }
 
     /**
-     * Creates or returns the desired {@link NativeIndexBuildStrategy} implementation. Intended to be used by {@link NativeIndexWriter}
-     * @param fieldInfo
-     * @return
+     * @param fieldInfo             Field related attributes/info
+     * @param totalLiveDocs         Number of documents with the vector field. This values comes from {@link org.opensearch.knn.index.codec.KNN990Codec.NativeEngines990KnnVectorsWriter#flush}
+     *                              and {@link org.opensearch.knn.index.codec.KNN990Codec.NativeEngines990KnnVectorsWriter#mergeOneField}
+     * @param knnVectorValues       An instance of {@link KNNVectorValues} which is used to evaluate the size threshold KNN_REMOTE_VECTOR_BUILD_THRESHOLD
+     * @return                      The {@link NativeIndexBuildStrategy} to be used. Intended to be used by {@link NativeIndexWriter}
+     * @throws IOException
      */
-    public NativeIndexBuildStrategy getBuildStrategy(final FieldInfo fieldInfo) {
+    public NativeIndexBuildStrategy getBuildStrategy(
+        final FieldInfo fieldInfo,
+        final int totalLiveDocs,
+        final KNNVectorValues<?> knnVectorValues
+    ) throws IOException {
         final KNNEngine knnEngine = extractKNNEngine(fieldInfo);
         boolean isTemplate = fieldInfo.attributes().containsKey(MODEL_ID);
         boolean iterative = !isTemplate && KNNEngine.FAISS == knnEngine;
@@ -47,11 +58,15 @@ public final class NativeIndexBuildStrategyFactory {
             ? MemOptimizedNativeIndexBuildStrategy.getInstance()
             : DefaultIndexBuildStrategy.getInstance();
 
-        if (repositoriesServiceSupplier != null
+        initializeVectorValues(knnVectorValues);
+        long vectorBlobLength = ((long) knnVectorValues.bytesPerVector()) * totalLiveDocs;
+
+        if (KNNFeatureFlags.isKNNRemoteVectorBuildEnabled()
+            && repositoriesServiceSupplier != null
             && indexSettings != null
             && knnEngine.supportsRemoteIndexBuild()
-            && RemoteIndexBuildStrategy.shouldBuildIndexRemotely(indexSettings)) {
-            return new RemoteIndexBuildStrategy(repositoriesServiceSupplier, strategy);
+            && RemoteIndexBuildStrategy.shouldBuildIndexRemotely(indexSettings, vectorBlobLength)) {
+            return new RemoteIndexBuildStrategy(repositoriesServiceSupplier, strategy, indexSettings);
         } else {
             return strategy;
         }
--- a/src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/DefaultVectorRepositoryAccessor.java
+++ b/src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/DefaultVectorRepositoryAccessor.java
@@ -1,0 +1,213 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.knn.index.codec.nativeindex.remote;
+
+import lombok.AllArgsConstructor;
+import lombok.extern.log4j.Log4j2;
+import org.opensearch.action.LatchedActionListener;
+import org.opensearch.common.CheckedTriFunction;
+import org.opensearch.common.StreamContext;
+import org.opensearch.common.blobstore.AsyncMultiStreamBlobContainer;
+import org.opensearch.common.blobstore.BlobContainer;
+import org.opensearch.common.blobstore.BlobPath;
+import org.opensearch.common.blobstore.stream.write.WriteContext;
+import org.opensearch.common.blobstore.stream.write.WritePriority;
+import org.opensearch.common.io.InputStreamContainer;
+import org.opensearch.core.action.ActionListener;
+import org.opensearch.index.IndexSettings;
+import org.opensearch.knn.index.VectorDataType;
+import org.opensearch.knn.index.vectorvalues.KNNVectorValues;
+import org.opensearch.repositories.blobstore.BlobStoreRepository;
+
+import java.io.BufferedInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Supplier;
+
+import static org.opensearch.knn.index.codec.nativeindex.remote.RemoteIndexBuildStrategy.DOC_ID_FILE_EXTENSION;
+import static org.opensearch.knn.index.codec.nativeindex.remote.RemoteIndexBuildStrategy.VECTORS_PATH;
+import static org.opensearch.knn.index.codec.nativeindex.remote.RemoteIndexBuildStrategy.VECTOR_BLOB_FILE_EXTENSION;
+import static org.opensearch.knn.index.codec.util.KNNCodecUtil.initializeVectorValues;
+
+@Log4j2
+@AllArgsConstructor
+public class DefaultVectorRepositoryAccessor implements VectorRepositoryAccessor {
+    private final BlobStoreRepository repository;
+    private final IndexSettings indexSettings;
+
+    /**
+     * If the repository implements {@link AsyncMultiStreamBlobContainer}, then parallel uploads will be used. Parallel uploads are backed by a {@link WriteContext}, for which we have a custom
+     * {@link org.opensearch.common.blobstore.stream.write.StreamContextSupplier} implementation.
+     *
+     * @see DefaultVectorRepositoryAccessor#getStreamContext
+     * @see DefaultVectorRepositoryAccessor#getTransferPartStreamSupplier
+     *
+     * @param blobName                  Base name of the blobs we are writing, excluding file extensions
+     * @param totalLiveDocs             Number of documents we are processing. This is used to compute the size of the blob we are writing
+     * @param vectorDataType            Data type of the vector (FLOAT, BYTE, BINARY)
+     * @param knnVectorValuesSupplier   Supplier for {@link KNNVectorValues}
+     * @throws IOException
+     * @throws InterruptedException
+     */
+    @Override
+    public void writeToRepository(
+        String blobName,
+        int totalLiveDocs,
+        VectorDataType vectorDataType,
+        Supplier<KNNVectorValues<?>> knnVectorValuesSupplier
+    ) throws IOException, InterruptedException {
+        assert repository != null;
+        // Get the blob container based on blobName and the repo base path. This is where the blobs will be written to.
+        BlobPath path = repository.basePath().add(indexSettings.getUUID() + VECTORS_PATH);
+        BlobContainer blobContainer = repository.blobStore().blobContainer(path);
+
+        KNNVectorValues<?> knnVectorValues = knnVectorValuesSupplier.get();
+        initializeVectorValues(knnVectorValues);
+        long vectorBlobLength = (long) knnVectorValues.bytesPerVector() * totalLiveDocs;
+
+        if (blobContainer instanceof AsyncMultiStreamBlobContainer) {
+            // First initiate vectors upload
+            log.debug("Repository {} Supports Parallel Blob Upload", repository);
+            // WriteContext is the main entry point into asyncBlobUpload. It stores all of our upload configurations, analogous to
+            // BuildIndexParams
+            WriteContext writeContext = new WriteContext.Builder().fileName(blobName + VECTOR_BLOB_FILE_EXTENSION)
+                .streamContextSupplier((partSize) -> getStreamContext(partSize, vectorBlobLength, knnVectorValuesSupplier, vectorDataType))
+                .fileSize(vectorBlobLength)
+                .failIfAlreadyExists(true)
+                .writePriority(WritePriority.NORMAL)
+                // TODO: Checksum implementations -- It is difficult to calculate a checksum on the knnVectorValues as
+                // there is no underlying file upon which we can create the checksum. We should be able to create a
+                // checksum still by iterating through once, however this will be an expensive operation.
+                .uploadFinalizer((bool) -> {})
+                .doRemoteDataIntegrityCheck(false)
+                .expectedChecksum(null)
+                .build();
+
+            AtomicReference<Exception> exception = new AtomicReference<>();
+            final CountDownLatch latch = new CountDownLatch(1);
+            ((AsyncMultiStreamBlobContainer) blobContainer).asyncBlobUpload(
+                writeContext,
+                new LatchedActionListener<>(new ActionListener<>() {
+                    @Override
+                    public void onResponse(Void unused) {
+                        log.debug(
+                            "Parallel vector upload succeeded for blob {} with size {}",
+                            blobName + VECTOR_BLOB_FILE_EXTENSION,
+                            vectorBlobLength
+                        );
+                    }
+
+                    @Override
+                    public void onFailure(Exception e) {
+                        log.error(
+                            "Parallel vector upload failed for blob {} with size {}",
+                            blobName + VECTOR_BLOB_FILE_EXTENSION,
+                            vectorBlobLength,
+                            e
+                        );
+                        exception.set(e);
+                    }
+                }, latch)
+            );
+
+            // Then upload doc id blob before waiting on vector uploads
+            // TODO: We wrap with a BufferedInputStream to support retries. We can tune this buffer size to optimize performance.
+            // Note: We do not use the parallel upload API here as the doc id blob will be much smaller than the vector blob
+            writeDocIds(knnVectorValuesSupplier.get(), vectorBlobLength, totalLiveDocs, blobName, blobContainer);
+            latch.await();
+            if (exception.get() != null) {
+                throw new IOException(exception.get());
+            }
+        } else {
+            log.debug("Repository {} Does Not Support Parallel Blob Upload", repository);
+            // Write Vectors
+            InputStream vectorStream = new BufferedInputStream(new VectorValuesInputStream(knnVectorValuesSupplier.get(), vectorDataType));
+            log.debug("Writing {} bytes for {} docs to {}", vectorBlobLength, totalLiveDocs, blobName + VECTOR_BLOB_FILE_EXTENSION);
+            blobContainer.writeBlob(blobName + VECTOR_BLOB_FILE_EXTENSION, vectorStream, vectorBlobLength, true);
+            // Then write doc ids
+            writeDocIds(knnVectorValuesSupplier.get(), vectorBlobLength, totalLiveDocs, blobName, blobContainer);
+        }
+    }
+
+    /**
+     * Helper method for uploading doc ids to repository, as it's re-used in both parallel and sequential upload cases
+     * @param knnVectorValues
+     * @param vectorBlobLength
+     * @param totalLiveDocs
+     * @param blobName
+     * @param blobContainer
+     * @throws IOException
+     */
+    private void writeDocIds(
+        KNNVectorValues<?> knnVectorValues,
+        long vectorBlobLength,
+        long totalLiveDocs,
+        String blobName,
+        BlobContainer blobContainer
+    ) throws IOException {
+        InputStream docStream = new BufferedInputStream(new DocIdInputStream(knnVectorValues));
+        log.debug(
+            "Writing {} bytes for {} docs ids to {}",
+            vectorBlobLength,
+            totalLiveDocs * Integer.BYTES,
+            blobName + DOC_ID_FILE_EXTENSION
+        );
+        blobContainer.writeBlob(blobName + DOC_ID_FILE_EXTENSION, docStream, totalLiveDocs * Integer.BYTES, true);
+    }
+
+    /**
+     * Returns a {@link org.opensearch.common.StreamContext}. Intended to be invoked as a {@link org.opensearch.common.blobstore.stream.write.StreamContextSupplier},
+     * which takes the partSize determined by the repository implementation and calculates the number of parts as well as handles the last part of the stream.
+     *
+     * @see DefaultVectorRepositoryAccessor#getTransferPartStreamSupplier
+     *
+     * @param partSize                  Size of each InputStream to be uploaded in parallel. Provided by repository implementation
+     * @param vectorBlobLength          Total size of the vectors across all InputStreams
+     * @param knnVectorValuesSupplier   Supplier for {@link KNNVectorValues}
+     * @param vectorDataType            Data type of the vector (FLOAT, BYTE, BINARY)
+     * @return a {@link org.opensearch.common.StreamContext} with a function that will create {@link InputStream}s of {@param partSize}
+     */
+    private StreamContext getStreamContext(
+        long partSize,
+        long vectorBlobLength,
+        Supplier<KNNVectorValues<?>> knnVectorValuesSupplier,
+        VectorDataType vectorDataType
+    ) {
+        long lastPartSize = (vectorBlobLength % partSize) != 0 ? vectorBlobLength % partSize : partSize;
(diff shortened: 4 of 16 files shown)


jed326:
@jmazanec15 @shatejas @Vikasht34 @navneet1v could you please help review? thanks!

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/VectorValuesInputStream.java
```diff
@@ -0,0 +1,199 @@
...
+
+        if (!currentBuffer.hasRemaining()) {
+            advanceAndReloadBuffer();
+            if (currentBuffer == null) {
+                return -1;
+            }
+        }
+
+        bytesRemaining--;
+        return currentBuffer.get() & 0xFF;
```
navneet1v, line 92: why we need a `&` here?

navneet1v (commented):
@jed326 can you please add the flow of functions used by the Repository service for InputStream. Since there are some override functions its hard to follow which function is called when. 

It will also be good if you can add a packge-info.java explaining the flow and usage of the InputStream in more details.

navneet1v:
@jed326 lets avoid using skip-change log tag and update changelogs

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
```diff
@@ -126,23 +169,121 @@ private BlobStoreRepository getRepository() throws RepositoryMissingException {
...
+        String blobName,
         int totalLiveDocs,
-        SegmentWriteState segmentWriteState
+        VectorDataType vectorDataType,
+        Supplier<KNNVectorValues<?>> knnVectorValuesSupplier
     ) throws IOException, InterruptedException {
-        throw new NotImplementedException();
+        BlobPath path = getRepository().basePath().add(indexSettings.getUUID() + VECTORS_PATH);
+        BlobContainer blobContainer = getRepository().blobStore().blobContainer(path);
+
```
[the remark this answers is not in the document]
jed326: Added more detailed comments to all of the private methods here, let me know if you think there is still a gap

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/VectorValuesInputStream.java
jed326: We technically don't (added a comment explaining), however I think it's safer to keep it here. IntelliJ itself also gives you a warning if you don't do the conversion and all the base `InputStream` implementations do the same.

I also added some tests to verify that `read() and `read(byte[] b, int offset, int length)` read the same data over a given stream.

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased 3.0](https://github.com/opensearch-project/k-NN/compare/2.x...HEAD)
 ### Features
+* [Remote Vector Index Build] Introduce Remote Native Index Build feature flag, settings, and initial skeleton [#2525](https://github.com/opensearch-project/k-NN/pull/2525)
```
jmazanec15, line 9: Why 2?

src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexBuildStrategyFactory.java
```diff
@@ -38,7 +42,11 @@ public NativeIndexBuildStrategyFactory(Supplier<RepositoriesService> repositorie
      * @param fieldInfo
      * @return
      */
-    public NativeIndexBuildStrategy getBuildStrategy(final FieldInfo fieldInfo) {
+    public NativeIndexBuildStrategy getBuildStrategy(
+        final FieldInfo fieldInfo,
+        final int totalLiveDocs,
```
jmazanec15, line 47: This has potential to be inaccurate if many of the docs are missing the vector field, right?

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/DocIdInputStream.java
```diff
@@ -0,0 +1,97 @@
...
+ * {@link InputStream} implementation of doc ids backed by {@link KNNVectorValues} rather than any file. Intended for use by {@link RemoteIndexBuildStrategy}
+ */
+@Log4j2
+class DocIdInputStream extends InputStream {
+    private final KNNVectorValues<?> knnVectorValues;
+    // Doc ids are 4 byte integers, byte read() only returns a single byte, so we will need to track the byte position within a doc id.
+    // For simplicity, and to maintain the byte ordering, we use a buffer with size of 1 int.
+    private ByteBuffer currentBuffer;
+
+    public DocIdInputStream(KNNVectorValues<?> knnVectorValues) throws IOException {
```
jmazanec15, line 29: javadoc

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
```diff
@@ -127,22 +172,157 @@ private BlobStoreRepository getRepository() throws RepositoryMissingException {
...
+                        log.debug(
+                            "Parallel upload succeeded for blob {} with size {}",
+                            blobName + VECTOR_BLOB_FILE_EXTENSION,
+                            vectorBlobLength
+                        );
+                    }
+
+                    @Override
+                    public void onFailure(Exception e) {
+                        log.debug(
```
jmazanec15, line 236: Should this be error level?

src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexBuildStrategyFactory.java
jed326: This `totalLiveDocs` value comes from `NativeEngines990KnnVectorsWriter#flush` or `NativeEngines990KnnVectorsWriter#merge`, so it's already only a count of docs with the vector field. Will add this in code comments. 

https://github.com/opensearch-project/k-NN/blob/c7ac05c8822d02b2fe93aebefec5788e73c27e25/src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriter.java#L128

https://github.com/opensearch-project/k-NN/blob/c7ac05c8822d02b2fe93aebefec5788e73c27e25/src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriter.java#L172

CHANGELOG.md
jed326: I think we should keep accurate changelogs for this feature, so added the missing changelog from #2525 here as well. I can remove it if you don't think this is the right thing to do though.

jmazanec15: no all good

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/DocIdInputStream.java
```diff
@@ -0,0 +1,103 @@
...
+            return -1;
+        }
+
+        int available = currentBuffer.remaining();
+        if (available <= 0) {
+            advanceAndReloadBuffer();
+            if (currentBuffer == null) {
+                return -1;
+            }
+            available = currentBuffer.remaining();
```
jmazanec15, line 72: this will always be positive?

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
```diff
@@ -37,28 +53,57 @@ public class RemoteIndexBuildStrategy implements NativeIndexBuildStrategy {
...
     private final NativeIndexBuildStrategy fallbackStrategy;
+    private final IndexSettings indexSettings;
+
     private static final String VECTOR_BLOB_FILE_EXTENSION = ".knnvec";
     private static final String DOC_ID_FILE_EXTENSION = ".knndid";
+    private static final String VECTORS_PATH = "_vectors";
 
     /**
      * Public constructor
      *
```
jmazanec15, line 64: Fix javadoc

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
```diff
@@ -37,28 +53,57 @@ public class RemoteIndexBuildStrategy implements NativeIndexBuildStrategy {
...
+        NativeIndexBuildStrategy fallbackStrategy,
+        IndexSettings indexSettings
+    ) {
         this.repositoriesServiceSupplier = repositoriesServiceSupplier;
         this.fallbackStrategy = fallbackStrategy;
+        this.indexSettings = indexSettings;
     }
 
     /**
      * @return whether to use the remote build feature
```
jmazanec15, line 78: fix javadoc

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
```diff
@@ -37,28 +53,57 @@ public class RemoteIndexBuildStrategy implements NativeIndexBuildStrategy {
...
      */
-    public static boolean shouldBuildIndexRemotely(IndexSettings indexSettings) {
+    public static boolean shouldBuildIndexRemotely(IndexSettings indexSettings, long vectorBlobLength) {
```
jmazanec15, line 80: Can we null check index settings and return false?

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
```diff
@@ -73,16 +118,18 @@ public static boolean shouldBuildIndexRemotely(IndexSettings indexSettings) {
      */
     @Override
     public void buildAndWriteIndex(BuildIndexParams indexInfo) throws IOException {
-        // TODO: Metrics Collection
         StopWatch stopWatch;
         long time_in_millis;
         try {
             stopWatch = new StopWatch().start();
+            // We create a new time based UUID per file in order to avoid conflicts across shards. It is also very difficult to get the
```
jmazanec15, line 125: What makes this time based?

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
```diff
@@ -127,22 +174,163 @@ private BlobStoreRepository getRepository() throws RepositoryMissingException {
...
+                .build();
+
+            AtomicReference<Exception> exception = new AtomicReference<>();
+            final CountDownLatch latch = new CountDownLatch(1);
+            ((AsyncMultiStreamBlobContainer) blobContainer).asyncBlobUpload(
+                writeContext,
+                new LatchedActionListener<>(new ActionListener<>() {
+                    @Override
+                    public void onResponse(Void unused) {
+                        log.debug(
```
jmazanec15, line 231: Im thinking this can be info as well - I dont think this will happen that frequently and will be good to know.

src/testFixtures/java/org/opensearch/knn/KNNRestTestCase.java
```diff
@@ -178,6 +178,7 @@ public void setupRemoteIndexBuildSettings() throws Exception {
         if (randomBoolean() && isRemoteIndexBuildSupported(getBWCVersion())) {
             updateClusterSettings(KNNFeatureFlags.KNN_REMOTE_VECTOR_BUILD_SETTING.getKey(), true);
             updateClusterSettings(KNNSettings.KNN_REMOTE_VECTOR_REPO, "integ-test-repo");
+            updateClusterSettings(KNNSettings.KNN_REMOTE_VECTOR_BUILD_THRESHOLD, "0mb");
```
jmazanec15, line 181: Why change this?

jmazanec15 (commented):
overall - looks good to me. A few minor comments, but nothing blocking

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/DocIdInputStream.java
jed326: Yes, this is the implementation:
```
    /**
     * Returns the number of elements between the current position and the
     * limit.
     *
     * @return  The number of elements remaining in this buffer
     */
    public final int remaining() {
        int rem = limit - position;
        return rem > 0 ? rem : 0;
    }
```

`advanceAndReloadBuffer()` above this will also always set the buffer position to 0.

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
jed326: I was thinking not needed since it's already checked by the calling class, but just to be safe we can add it

src/testFixtures/java/org/opensearch/knn/KNNRestTestCase.java
jed326: This is to ensure the rest tests randomly go down the remote vector build path, as otherwise the `RemoteIndexBuildStrategy` won't be used. I'm tracking more comprehensive e2e testing as a part of https://github.com/opensearch-project/k-NN/issues/2553, we definitely will not remove the feature flag without that completed.

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
jed326: This is the implementation: https://github.com/opensearch-project/OpenSearch/blob/7e2d2437a14580e985440c140f06680dc4e3cd81/libs/common/src/main/java/org/opensearch/common/UUIDs.java#L47-L51

src/main/java/org/opensearch/knn/index/KNNSettings.java
```diff
@@ -388,6 +391,15 @@ public class KNNSettings {
      */
     public static final Setting<String> KNN_REMOTE_VECTOR_REPO_SETTING = Setting.simpleString(KNN_REMOTE_VECTOR_REPO, Dynamic, NodeScope);
 
+    /**
+     * Cluster level setting which indicates the size threshold above which remote vector builds will be enabled.
+     */
+    public static final Setting<ByteSizeValue> KNN_REMOTE_VECTOR_BUILD_THRESHOLD_SETTING = Setting.byteSizeSetting(
```
navneet1v, line 397: we should have this at the index level rather than cluster level.

jed326: I don't have a strong opinion on this, but why do you think index is preferable to cluster? It's not clear to me the use case where you would want to have different values for different indices

jed326: @jmazanec15 @shatejas any thoughts on this?

jmazanec15: Id say index would be better. Itd be easier in the very least for testing

jed326: Fair enough, made this an index setting in the latest revision.

jmazanec15 (dismissed):

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
```diff
@@ -37,28 +54,64 @@ public class RemoteIndexBuildStrategy implements NativeIndexBuildStrategy {
...
+     * @param vectorBlobLength      The size of the vector blob, used to determine if the size threshold is met
+     * @return true if remote index build should be used, else false
      */
-    public static boolean shouldBuildIndexRemotely(IndexSettings indexSettings) {
+    public static boolean shouldBuildIndexRemotely(IndexSettings indexSettings, long vectorBlobLength) {
+        if (indexSettings == null) {
+            return false;
+        }
+
+        // If setting is not enabled, return false
```
shatejas, line 90: Can we throw in a unit test for this? it has quite a bit of ifs

shatejas (commented):
Looks good overall. A few things related to maintenance and UTs

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
```diff
@@ -127,22 +182,163 @@ private BlobStoreRepository getRepository() throws RepositoryMissingException {
...
+                .build();
+
+            AtomicReference<Exception> exception = new AtomicReference<>();
+            final CountDownLatch latch = new CountDownLatch(1);
+            ((AsyncMultiStreamBlobContainer) blobContainer).asyncBlobUpload(
+                writeContext,
+                new LatchedActionListener<>(new ActionListener<>() {
+                    @Override
+                    public void onResponse(Void unused) {
+                        log.info(
```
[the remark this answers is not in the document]
jed326: This was requested by @jmazanec15 -- https://github.com/opensearch-project/k-NN/pull/2550#discussion_r1970227679

src/test/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategyTests.java
```diff
@@ -16,52 +40,229 @@
...
-            .parameters(Map.of("index", "param"))
-            .knnVectorValuesSupplier(() -> knnVectorValues)
-            .totalLiveDocs((int) knnVectorValues.totalLiveDocs())
-            .build();
+        objectUnderTest.writeToRepository("test_blob", 100, VectorDataType.FLOAT, knnVectorValuesSupplier);
+
+        assertTrue(asyncUpload.get());
+        assertTrue(upload.get());
+        verify(mockBlobStore).blobContainer(any());
+        verify(mockRepository).basePath();
```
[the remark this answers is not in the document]
jed326: That should be verified by the `assertTrue(upload.get());` call which is happening on the `writeBlob`. Let me convert `testContainer` into a Mockito spy though, think that will be a lot cleaner.

src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexBuildStrategyFactory.java
navneet1v: @jmazanec15 that bug of wrong counting the docs was removed and liveDocs are counted once by iterating over vector values.

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
```diff
@@ -127,22 +182,163 @@ private BlobStoreRepository getRepository() throws RepositoryMissingException {
...
+     * @param knnVectorValuesSupplier       Supplier for {@link KNNVectorValues}
+     * @param vectorDataType                Data type of the vector (FLOAT, BYTE, BINARY)
+     * @return a function with which the repository implementation will use to create {@link VectorValuesInputStream}s of specific sizes and start positions.
+     */
+    private CheckedTriFunction<Integer, Long, Long, InputStreamContainer, IOException> getTransferPartStreamSupplier(
+        Supplier<KNNVectorValues<?>> knnVectorValuesSupplier,
+        VectorDataType vectorDataType
+    ) {
+        return ((partNo, size, position) -> {
+            log.info("Creating InputStream for partNo: {}, size: {}, position: {}", partNo, size, position);
```
navneet1v, line 333: should this be info? I feel this is debug log

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
```diff
@@ -127,22 +182,163 @@ private BlobStoreRepository getRepository() throws RepositoryMissingException {
...
+        BlobPath path = getRepository().basePath().add(indexSettings.getUUID() + VECTORS_PATH);
+        BlobContainer blobContainer = getRepository().blobStore().blobContainer(path);
+
+        KNNVectorValues<?> knnVectorValues = knnVectorValuesSupplier.get();
+        initializeVectorValues(knnVectorValues);
+        long vectorBlobLength = (long) knnVectorValues.bytesPerVector() * totalLiveDocs;
+
+        if (blobContainer instanceof AsyncMultiStreamBlobContainer) {
+            // First initiate vectors upload
+            log.debug("Repository {} Supports Parallel Blob Upload", getRepository());
```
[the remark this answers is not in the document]
navneet1v: +1

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
```diff
@@ -127,22 +182,163 @@ private BlobStoreRepository getRepository() throws RepositoryMissingException {
...
+     * @param totalLiveDocs             Number of documents we are processing. This is used to compute the size of the blob we are writing
+     * @param vectorDataType            Data type of the vector (FLOAT, BYTE, BINARY)
+     * @param knnVectorValuesSupplier   Supplier for {@link KNNVectorValues}
      * @throws IOException
      * @throws InterruptedException
      */
-    private void writeToRepository(
-        String fieldName,
-        Supplier<KNNVectorValues<?>> knnVectorValuesSupplier,
+    @VisibleForTesting
```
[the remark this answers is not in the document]
navneet1v: Yeah I think this is a good suggestion. I should have mentioned the same when I first reviewed the code: https://github.com/opensearch-project/k-NN/pull/2550#discussion_r1965109028

navneet1v (commented):
Overall code looks good to me. Please fix @shatejas comments specially related to creating repo/dataacess class .

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
```diff
@@ -127,22 +182,163 @@ private BlobStoreRepository getRepository() throws RepositoryMissingException {
...
+        String blobName,
         int totalLiveDocs,
-        SegmentWriteState segmentWriteState
+        VectorDataType vectorDataType,
+        Supplier<KNNVectorValues<?>> knnVectorValuesSupplier
     ) throws IOException, InterruptedException {
-        throw new NotImplementedException();
+        // Get the blob container based on blobName and the repo base path. This is where the blobs will be written to.
+        BlobPath path = getRepository().basePath().add(indexSettings.getUUID() + VECTORS_PATH);
+        BlobContainer blobContainer = getRepository().blobStore().blobContainer(path);
```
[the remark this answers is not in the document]
jed326: `getRepository()` NPE is handled by itself, then the subsequent calls in the chain are handled by the implementations coming from core. Any NPE would cause us to fall back to the fallback strategy anyways though.

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/VectorValuesInputStream.java
```diff
@@ -0,0 +1,208 @@
...
+     * @return
+     * @throws IOException
+     */
+    private void setPosition(long n) throws IOException {
+        if (currentBuffer.position() != 0) {
+            throw new UnsupportedOperationException("setPosition is only supported from the start of a vector");
+        }
+
+        long bytesSkipped = 0;
+        int vectorsToSkip = (int) (n / bytesPerVector);
```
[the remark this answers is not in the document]
jed326: No, `n` will not necessarily be a whole number of vectors, as `n` is a value computed by the repository implementation without context of the input stream type. This method will correctly set the position to the remainder of bytes in the current buffer after skipping the correct number of vectors.

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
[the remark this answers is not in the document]
jed326: Added a `VectorRepositoryAccessor` interface with `DefaultVectorRepositoryAccessor` implementation

src/test/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategyTests.java
```diff
@@ -16,52 +40,229 @@
...
         );
-        final KNNVectorValues<byte[]> knnVectorValues = KNNVectorValuesFactory.getVectorValues(VectorDataType.FLOAT, randomVectorValues);
+        objectUnderTest.buildAndWriteIndex(buildIndexParams);
+        assertTrue(fallback.get());
+    }
 
+    /**
+     * Test that whenever an AsyncMultiStreamBlobContainer is used, both asyncBlobUpload and writeBlob are invoked once and only once
+     */
+    public void testRepositoryInteraction() throws IOException, InterruptedException {
```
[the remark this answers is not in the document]
jed326: Added in latest revision

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
jed326: Added in latest revision

jed326:
Thanks @navneet1v @shatejas @jmazanec15, I've added a new `VectorRepositoryAccessor` interface with a default implementation of `DefaultVectorRepositoryAccessor` to handle all of the `KNNVectorValues` -> `InputStream` conversions. Please take another look whenever you get a chance, thanks!

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/DefaultVectorRepositoryAccessor.java
```diff
@@ -0,0 +1,213 @@
...
+     * @param vectorDataType            Data type of the vector (FLOAT, BYTE, BINARY)
+     * @return a {@link org.opensearch.common.StreamContext} with a function that will create {@link InputStream}s of {@param partSize}
+     */
+    private StreamContext getStreamContext(
+        long partSize,
+        long vectorBlobLength,
+        Supplier<KNNVectorValues<?>> knnVectorValuesSupplier,
+        VectorDataType vectorDataType
+    ) {
+        long lastPartSize = (vectorBlobLength % partSize) != 0 ? vectorBlobLength % partSize : partSize;
```
Vikasht34, line 181: is not this would be simple with this 

```
int numberOfParts = (int) ((vectorBlobLength + partSize - 1) / partSize);

Ensures the last part is always counted properly.
```

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/DocIdInputStream.java
```diff
@@ -0,0 +1,103 @@
...
+     * Use to represent the doc ids of a {@link KNNVectorValues} as an {@link InputStream}. Expected to be used only with {@link org.opensearch.common.blobstore.BlobContainer#writeBlob}.
+     * @param knnVectorValues
+     * @throws IOException
+     * @see VectorValuesInputStream
+     */
+    public DocIdInputStream(KNNVectorValues<?> knnVectorValues) throws IOException {
+        this.currentBuffer = ByteBuffer.allocate(Integer.BYTES).order(ByteOrder.LITTLE_ENDIAN);
+        this.knnVectorValues = knnVectorValues;
+        initializeVectorValues(this.knnVectorValues);
+        reloadBuffer();
```
Vikasht34, line 39: Could we  call advanceAndReloadBuffer here as well, that will ensure checks.

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/DocIdInputStream.java
```diff
@@ -0,0 +1,103 @@
...
+            }
+        }
+
+        // Unsigned byte conversion is not technically needed as we are using a ByteBuffer, however we perform this operation still just in
+        // case.
+        return currentBuffer.get() & 0xFF;
+    }
+
+    @Override
+    public int read(byte[] b, int off, int len) throws IOException {
```
Vikasht34, line 61: is the idea is to keep calling read till we get 4 bytes

or we can simplify 

```
while (bytesRead < len && (buffer.hasRemaining() || RealodDocId())) {
            int toRead = Math.min(buffer.remaining(), len - bytesRead);
            buffer.get(b, off + bytesRead, toRead);
            bytesRead += toRead;
        }
        return (bytesRead == 0) ? -1 : bytesRead;
```

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/DocIdInputStream.java
```diff
@@ -0,0 +1,103 @@
...
+        int bytesToRead = Math.min(available, len);
+        currentBuffer.get(b, off, bytesToRead);
+        return bytesToRead;
+    }
+
+    /**
+     * Advances to the next doc, and then refills the buffer with the new doc.
+     * @throws IOException
+     */
+    private void advanceAndReloadBuffer() throws IOException {
```
Vikasht34, line 84: We don;t need if else here , it can be 

```
int docId = knnVectorValues.nextDoc();
            endOfStream = true;
            return false;
        }

        buffer.clear();
        buffer.putInt(docId);
        buffer.flip();
```

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/DocIdInputStream.java
```diff
@@ -0,0 +1,103 @@
...
+        }
+
+        // Unsigned byte conversion is not technically needed as we are using a ByteBuffer, however we perform this operation still just in
+        // case.
+        return currentBuffer.get() & 0xFF;
+    }
+
+    @Override
+    public int read(byte[] b, int off, int len) throws IOException {
+        if (currentBuffer == null) {
```
Vikasht34, line 62: nit :- Can we have somthing boolean instread of making CurrentBuffer null , like end of stream.

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/VectorValuesInputStream.java
```diff
@@ -0,0 +1,208 @@
...
+import static org.opensearch.knn.index.VectorDataType.BINARY;
+import static org.opensearch.knn.index.VectorDataType.BYTE;
+import static org.opensearch.knn.index.VectorDataType.FLOAT;
+import static org.opensearch.knn.index.codec.util.KNNCodecUtil.initializeVectorValues;
+
+/**
+ * {@link InputStream} implementation backed by {@link KNNVectorValues} rather than any file. Intended for use by {@link RemoteIndexBuildStrategy}
+ */
+@Log4j2
+class VectorValuesInputStream extends InputStream {
```
Vikasht34, line 30: I have similar comments on this file like DocIdInputStream, I guess We can simplify this whole class.

src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexBuildStrategyFactory.java
```diff
@@ -34,11 +38,18 @@ public NativeIndexBuildStrategyFactory(Supplier<RepositoriesService> repositorie
...
-     * @return
+     * @param fieldInfo             Field related attributes/info
+     * @param totalLiveDocs         Number of documents with the vector field. This values comes from {@link org.opensearch.knn.index.codec.KNN990Codec.NativeEngines990KnnVectorsWriter#flush}
+     *                              and {@link org.opensearch.knn.index.codec.KNN990Codec.NativeEngines990KnnVectorsWriter#mergeOneField}
+     * @param knnVectorValues       An instance of {@link KNNVectorValues} which is used to evaluate the size threshold KNN_REMOTE_VECTOR_BUILD_THRESHOLD
+     * @return                      The {@link NativeIndexBuildStrategy} to be used. Intended to be used by {@link NativeIndexWriter}
+     * @throws IOException
      */
-    public NativeIndexBuildStrategy getBuildStrategy(final FieldInfo fieldInfo) {
+    public NativeIndexBuildStrategy getBuildStrategy(
```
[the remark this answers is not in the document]
jed326: We need the additional parameters in order to do the threshold checks for `shouldBuildIndexRemotely`. Otherwise we cannot do the threshold checks until `buildAndWriteIndex`, which kind of defeats the point of the factory.

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/DefaultVectorRepositoryAccessor.java
```diff
@@ -0,0 +1,213 @@
...
+     *
+     * @param blobName                  Base name of the blobs we are writing, excluding file extensions
+     * @param totalLiveDocs             Number of documents we are processing. This is used to compute the size of the blob we are writing
+     * @param vectorDataType            Data type of the vector (FLOAT, BYTE, BINARY)
+     * @param knnVectorValuesSupplier   Supplier for {@link KNNVectorValues}
+     * @throws IOException
+     * @throws InterruptedException
+     */
+    @Override
+    public void writeToRepository(
```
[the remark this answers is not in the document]
Vikasht34: the provided code does not stop you to doing that..same code I wrote differently

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/DefaultVectorRepositoryAccessor.java
```diff
@@ -0,0 +1,213 @@
...
+     * @throws IOException
+     */
+    private void writeDocIds(
+        KNNVectorValues<?> knnVectorValues,
+        long vectorBlobLength,
+        long totalLiveDocs,
+        String blobName,
+        BlobContainer blobContainer
+    ) throws IOException {
+        InputStream docStream = new BufferedInputStream(new DocIdInputStream(knnVectorValues));
```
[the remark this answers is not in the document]
jed326: Repository implementation requires an `InputStream` that supports mark/reset in order to support upload retries as the stream needs to be able to be replayed from a specific point.

Rather than maintaining our own buffer and implementing this functionality internally within `DocIdInputStream`, we just use the `BufferedInputStream` to wrap it. The async upload path does this `BufferedInputStream` wrapping internally.

We have a future follow-up task to tune the size of this buffer.

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/DocIdInputStream.java
jed326: I think it's better to explicitly call initialize + reload instead of advanceAndReload so we don't depend on advanceAndReload to initialize the knnVectorValues. That behavior doesn't seem very intuitive to me.

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/DefaultVectorRepositoryAccessor.java
jed326: You need both `lastPartSize` and `numberOfParts` regardless, so you would need to do the `lastPartSize` calculation in both cases. I think the existing code reads more intuitively.

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/DocIdInputStream.java
jed326: Not sure I understand. The current implementation of this method would read at most 4 bytes. If >4 bytes are required, we leave it to the `InputStream`'s `readNBytes()` implementation to call this method multiple times. I think it is much simpler to isolate the functionalities that way.

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/DefaultVectorRepositoryAccessor.java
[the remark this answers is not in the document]
jed326: We can only do latch.await (or completableFuture.whenComplete) in the if case though, wouldn't you need to add another if statement to make the code you wrote work?

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/DocIdInputStream.java
jed326: I'm not against using a boolean, but I think it's important to make currentBuffer null to enforce we do not read extra bytes out of the buffer. Then an NPE here would mean a fatal/unrecoverable error in the remote upload process so we want to fail and use the fallbackStrategy. Without setting currentBuffer to null I think we're at risk of reading extra bytes.

shatejas (approved):
Looks good no major concerns as such

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/DefaultVectorRepositoryAccessor.java
```diff
@@ -0,0 +1,213 @@
...
+                .writePriority(WritePriority.NORMAL)
+                // TODO: Checksum implementations -- It is difficult to calculate a checksum on the knnVectorValues as
+                // there is no underlying file upon which we can create the checksum. We should be able to create a
+                // checksum still by iterating through once, however this will be an expensive operation.
+                .uploadFinalizer((bool) -> {})
+                .doRemoteDataIntegrityCheck(false)
+                .expectedChecksum(null)
+                .build();
+
+            AtomicReference<Exception> exception = new AtomicReference<>();
```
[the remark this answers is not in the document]
jed326: I looked into this some more, I think the existing implementation is still preferable to CompletableFuture. The reason is `asyncBlobUpload` does not throw an exception on upload failures, instead it propagates it to whatever ActionListener is passed into it. 

If we want to do this with CompletableFuture without an AtomicReference then we would need to throw the exception within the ActionListener, which is not correct behavior as the listener is not intended to interrupt execution flows.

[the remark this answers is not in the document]
Vikasht34: This is how it should work.

```
CompletableFuture<Void> vectorUploadFuture;
    
    if (blobContainer instanceof AsyncMultiStreamBlobContainer) {
        log.info("Using parallel upload for repository: {}", repository);
        vectorUploadFuture = CompletableFuture.runAsync(() -> {
            try {
                uploadVectorsAsync(blobContainer, blobName, vectorBlobLength, vectorDataType, knnVectorValuesSupplier);
            } catch (IOException e) {
                throw new RuntimeException("Vector upload failed", e);
            }
        });
    } else {
        log.info("Using sequential upload for repository: {}", repository);
        vectorUploadFuture = CompletableFuture.runAsync(() -> {
            try {
                uploadVectorsSequentially(blobContainer, blobName, vectorBlobLength, vectorDataType, knnVectorValues);
            } catch (IOException e) {
                throw new RuntimeException("Sequential vector upload failed", e);
            }
        });
    }

    CompletableFuture<Void> docIdUploadFuture = CompletableFuture.runAsync(() -> {
        try {
            writeDocIds(knnVectorValues, totalLiveDocs, blobName, blobContainer);
        } catch (IOException e) {
            throw new RuntimeException("Doc ID upload failed", e);
        }
    });

    // Wait for both tasks to complete
    CompletableFuture.allOf(vectorUploadFuture, docIdUploadFuture).join();
```

Vikasht34 (approved):
LGTM

jmazanec15 (approved):

Issue #2464, opened 2025-01-29
Remote Vector Index Build Component -- Future Work


jed326:
This issue is to track future work items for the remote vector index build work including both "fast follow" and "deprioritized" items.

owenhalpert:
LTI: 
- Client retry configuration enhancements
  - Support configuration on retryable codes, max retry count, base delay (for retries and backoffs)

PR #2554, opened 2025-02-24
[Remote Vector Index Build] Add download + indexOuput#write implementation to RemoteIndexBuildStrategy

jed326:
### Description
As a follow-up to https://github.com/opensearch-project/k-NN/pull/2550, this PR includes the implementation + testing for downloading a graph file from the remote repository and then writing to an `IndexOutput`.

The key change here is that in order to manage memory consumption, we use a buffer of size 64kb, the same as used in the `IndexOutputWithBuffer` class, to buffer the graph file from the remote repository into the `IndexOutput`, else we will go OOM on the download of larger graph files.

For now, we are only using sequential download for the graph file. Parallel download of the graph file has several difficulties associated with it, both on the download side as well as on the writing to `IndexOutput` side. For more details on these, see: https://github.com/opensearch-project/k-NN/issues/2464. We take on parallel blob download as a future improvement.

### Related Issues

### Check List


src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
```diff
@@ -98,7 +113,9 @@ public void buildAndWriteIndex(BuildIndexParams indexInfo) throws IOException {
             log.debug("Await vector build took {} ms for vector field [{}]", time_in_millis, indexInfo.getFieldName());
 
             stopWatch = new StopWatch().start();
-            readFromRepository();
+            // TODO: This blob will be retrieved from the remote vector build service status response
+            String blobName = UUIDs.base64UUID() + "_" + indexInfo.getFieldName() + "_" + indexInfo.getSegmentWriteState().segmentInfo.name;
```
navneet1v, line 117: not sure why this TODO is here?

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
```diff
@@ -163,7 +188,27 @@ private void awaitVectorBuild() {
...
-        throw new NotImplementedException();
+    @VisibleForTesting
+    void readFromRepository(String blobName, IndexOutput indexOutput) throws IOException {
+        BlobContainer blobContainer = getBlobContainer();
+        // TODO: We are using the sequential download API as multi-part parallel download is difficult for us to implement today and
+        // requires some changes in core. For more details, see: https://github.com/opensearch-project/k-NN/issues/2464
+        InputStream graphStream = blobContainer.readBlob(blobName);
+
+        // Allocate buffer of 64KB, same as used for CPU builds, see: IndexOutputWithBuffer
+        int CHUNK_SIZE = 64 * 1024;
```
navneet1v, line 199: We should move this to a constant in `IndexOutputWithBuffer` class and then reuse the value here

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
```diff
@@ -163,7 +188,27 @@ private void awaitVectorBuild() {
...
+        int bytesRead = 0;
+        // InputStream uses -1 indicates there are no more bytes to be read
+        while (bytesRead != -1) {
+            // Try to read CHUNK_SIZE into the buffer. The actual amount read may be less.
+            bytesRead = graphStream.read(buffer, 0, CHUNK_SIZE);
+            assert bytesRead <= CHUNK_SIZE;
+            // However many bytes we read, write it to the IndexOutput if != -1
+            if (bytesRead != -1) {
+                indexOutput.writeBytes(buffer, 0, bytesRead);
+            }
```
navneet1v, line 211: I think this whole logic can be moved inside the `IndexOutputWithBuffer` class rather than we keeping it here.

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
jed326: I actually specifically did not do that because I think we will want to tune this chunk size further based on benchmarking. Logically, this buffer size should not be related to the one in `IndexOutputWithBuffer` as this one is for downloading from remote repo while the other one is for copying from native memory.

Most likely requesting 64kb at a time from the S3 Input Stream is not optimal, I just picked this as the starting point since it's what we're doing for local memory copy. Let me add a TODO here with that info. I'm also tracking all the different buffer sizes / knobs we need to further optimize via benchmarking here: https://github.com/opensearch-project/k-NN/issues/2464

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
jed326: I was going to wait on the `awaitVectorBuild()` implementation to see how to best propagate this value, but I can just add that as a return value now and let that PR make any changes it needs to.

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
jed326: Originally I was thinking that due to https://github.com/opensearch-project/k-NN/pull/2554#discussion_r1968422535 I wanted to keep these separate as the buffer sizes would most likely diverge. I think even if the buffer sizes diverge though, keeping all this logic in the same `IndexOutputWithBuffer` would be better so I will make this change.

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
navneet1v: why would `awaitVectorBuild` will give you this value?

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
navneet1v: Got it make sense. But I would still like to make this as a constant.

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
jed326: Agreed! Made this change in the latest revision

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
jed326: We will get the file path to download from via the status API, so some part of this will still depend on the client implementation. I added a basic implementation in this PR assuming we get a `BlobPath` back from the client.

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
```diff
@@ -93,12 +100,12 @@ public void buildAndWriteIndex(BuildIndexParams indexInfo) throws IOException {
             log.debug("Submit vector build took {} ms for vector field [{}]", time_in_millis, indexInfo.getFieldName());
 
             stopWatch = new StopWatch().start();
-            awaitVectorBuild();
+            BlobPath downloadPath = awaitVectorBuild();
```
jed326, line 103: Assume that `awaitVectorBuild` returns a `BlobPath` constructed from the response the remote index build client gave. This may change in future PRs, see: https://github.com/opensearch-project/k-NN/pull/2548

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
```diff
@@ -155,15 +161,22 @@ private void submitVectorBuild() {
...
      * Read constructed vector file from remote repository and write to IndexOutput
      */
-    private void readFromRepository() {
-        throw new NotImplementedException();
+    @VisibleForTesting
+    void readFromRepository(BlobPath downloadPath, IndexOutputWithBuffer indexOutputWithBuffer) throws IOException {
+        BlobContainer blobContainer = getRepository().blobStore().blobContainer(downloadPath.parent());
+        // TODO: We are using the sequential download API as multi-part parallel download is difficult for us to implement today and
+        // requires some changes in core. For more details, see: https://github.com/opensearch-project/k-NN/issues/2464
+        String fileName = downloadPath.toArray()[downloadPath.toArray().length - 1];
```
navneet1v, line 178: this seems like a place that can cause NPE, can we ensure that length checks and other things to ensure that this line is not throwing NPEs

src/main/java/org/opensearch/knn/index/store/IndexOutputWithBuffer.java
```diff
@@ -33,6 +40,43 @@ public void writeBytes(int length) {
...
+     * bytes downloaded from the repository, so it may be more performant to use a larger buffer.
+     * We do not change the size of the existing buffer in case a fallback to the existing build strategy is needed.
+     * TODO: Tune the size of the buffer used by RemoteIndexBuildStrategy based on benchmarking
+     *
+     * @param inputStream       The stream from which we are reading bytes to write
+     * @param outputBuffer      The buffer used to buffer bytes
+     * @throws IOException
+     * @see IndexOutputWithBuffer#writeFromStreamWithBuffer(InputStream)
+     */
+    public void writeFromStreamWithBuffer(InputStream inputStream, byte[] outputBuffer) throws IOException {
```
navneet1v, line 66: why we need to make both functions public?

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
jed326: The point is that I'm trying to not take the file path resolution as part of this PR as whether we get it as a `BlobPath` or `String` or any other format is still tbd by the client implementation. Let me add some assertions / checks here for now though.

jed326: So `BlobPath` is not supposed to be used to represent a specific file, only a given directory. So in latest revision I'm going to take in a String and do the `BlobPath` creation manually.

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
```diff
@@ -155,15 +164,38 @@ private void submitVectorBuild() {
...
-    private void readFromRepository() {
-        throw new NotImplementedException();
+    @VisibleForTesting
+    void readFromRepository(String path, IndexOutputWithBuffer indexOutputWithBuffer) throws IOException {
+        if (path == null || path.isEmpty()) {
+            throw new IllegalArgumentException("download path is null or empty");
+        }
+        Path downloadPath = Paths.get(path);
+        String fileName = downloadPath.getFileName().toString();
+        if (!fileName.endsWith(KNNEngine.FAISS.getExtension())) {
```
rchitale7, line 183: Can we make this generic? Instead of .faiss, why not '.knnindex'?

jed326: This is a good point, but rather than making it generic like `*.knnindex` I think it would be better to just extract the extension from the knn engine here instead of hard coding faiss. And on the remote index build service side we wouldn't have to rename any graph files before uploading.

navneet1v (dismissed):

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/DefaultVectorRepositoryAccessor.java
```diff
@@ -210,4 +215,30 @@ private CheckedTriFunction<Integer, Long, Long, InputStreamContainer, IOExceptio
             return new InputStreamContainer(vectorValuesInputStream, size, position);
         });
     }
+
+    @VisibleForTesting
+    public void readFromRepository(String path, IndexOutputWithBuffer indexOutputWithBuffer) throws IOException {
```
navneet1v, line 220: should this be public if you are adding @VisibleForTesting

jed326: Ah that's an artifact from rebasing. Since we're implementing an interface now, it has to be public. I will make the `VectorRepositoryAccessor` package private though.

navneet1v (approved):

src/main/java/org/opensearch/knn/index/store/IndexOutputWithBuffer.java
```diff
@@ -33,6 +40,43 @@ public void writeBytes(int length) {
...
 
+    /**
+     * Writes to the {@link IndexOutput} by buffering bytes into the existing buffer in this class.
+     *
+     * @param inputStream       The stream from which we are reading bytes to write
+     * @throws IOException
+     * @see IndexOutputWithBuffer#writeFromStreamWithBuffer(InputStream, byte[])
+     */
+    public void writeFromStreamWithBuffer(InputStream inputStream) throws IOException {
+        writeFromStreamWithBuffer(inputStream, this.buffer);
```
shatejas, line 51: nit: Not sure why we need to wrap the method again

shatejas (approved):

src/main/java/org/opensearch/knn/index/store/IndexOutputWithBuffer.java
jed326: I can remove this if you think it would be better, but the idea is that the buffer we use for downloading graph files for remote index builds is not necessarily the same buffer as the one we use to buffer bytes out of native memory.

As a part of https://github.com/opensearch-project/k-NN/issues/2464 we will tune this buffer size (as well as the other ones used in the upload path), but for now we just use the same buffer as a starting point.

jed326: See previous discussion here: https://github.com/opensearch-project/k-NN/pull/2554#discussion_r1968383213

PR #2567, opened 2025-02-27
Remove usage of cluster level setting for circuit breaker

### LINKED ISSUE #1607: Remove cluster setting for native memory circuit breaker
## Description

Right now, as part of our k-NN circuit breaker and memory management service, we update a cluster setting [knn.circuit_breaker.triggered](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/KNNSettings.java#L222-L227) that is then used to block operations such as ingest ([ref](https://github.com/opensearch-project/k-NN/blob/653e7eb49769a60813a680813ec8df341742883d/src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java#L626-L632)). The setting allows there to be one value across the whole cluster - so, if one node circuit breaker is tripped, all ingestion is blocked.

I dont think this coordination across the cluster is necessary. It ends up making the circuit breaking logic more complex than it needs to be. It is inconsistent with OpenSearch circuit breaking, which only takes into account the node whose circuit breaker is tripped.

## Proposal
Im proposing as a first step of #1582, we remove (or deprecate) this setting and make the circuit breaking logic not require cluster synchronization.

jmazanec15:
### Description
Simplification of circuit breaker management. Before, we were periodically updating the circuit breaker cluster setting to have a global circuit breaker. However, this is inconvenient and not actually useful.

This change removes that logic and only trips based on node level circuit breaker. For knn stats, in order to fetch the cluster level circuit breaker, a node stats transport call with just the cache_capacity_reached stat is made to check all of the nodes. This isnt super efficient but its made for stats calls so its not on the critical path.

Overall, this greatly simplfiies the memory management code. However, it does mean that the cluster setting "knn.circuit_breaker.triggered" will no longer have an effect. But, I think for 3.0 this is okay. For users interested in this info, they can get it from the stats API.

### Related Issues
Resolves #1607 

### Check List


src/main/java/org/opensearch/knn/index/KNNCircuitBreaker.java
```diff
@@ -44,68 +32,34 @@ public static synchronized KNNCircuitBreaker getInstance() {
...
-                } catch (Exception e) {
-                    logger.error("[KNN] Exception getting stats: " + e);
+                // Unset capacityReached flag if currentSizeBytes is less than circuitBreakerUnsetSizeBytes
+                if (currentSizeKiloBytes <= circuitBreakerUnsetSizeKiloBytes) {
+                    setTripped(false);
                 }
             }
         };
-        this.threadPool.scheduleWithFixedDelay(runnable, TimeValue.timeValueSeconds(CB_TIME_INTERVAL), ThreadPool.Names.GENERIC);
+        threadPool.scheduleWithFixedDelay(runnable, TimeValue.timeValueSeconds(CB_TIME_INTERVAL), ThreadPool.Names.GENERIC);
```
jmazanec15, line 54: I could go even a step further an move this to the native cache manager to simplfiy things more.

src/main/java/org/opensearch/knn/plugin/stats/suppliers/KNNClusterLevelCircuitBreakerSupplier.java
```diff
@@ -0,0 +1,34 @@
...
+
+@AllArgsConstructor
+public class KNNClusterLevelCircuitBreakerSupplier implements Supplier<Boolean> {
+
+    private final Client client;
+
+    @Override
+    public Boolean get() {
+        try {
+            KNNCircuitBreakerTrippedResponse knnCircuitBreakerTrippedResponse = client.execute(
```
jmazanec15, line 25: TODO: This is going to have to somehow be made to be async, which will complicate things.

jmazanec15: I made this change. Its a bit messy - will clean up a bit.

src/main/java/org/opensearch/knn/index/KNNCircuitBreaker.java
jmazanec15: I made this change - I think its fairly straightforward and completely removes the potential circular dependency.

src/main/java/org/opensearch/knn/plugin/stats/CircuitBreakerStat.java
```diff
@@ -0,0 +1,51 @@
...
+
+import org.opensearch.core.action.ActionListener;
+import org.opensearch.knn.plugin.transport.KNNCircuitBreakerTrippedAction;
+import org.opensearch.knn.plugin.transport.KNNCircuitBreakerTrippedRequest;
+import org.opensearch.transport.client.Client;
+
+import java.util.Map;
+import java.util.function.Function;
+
+public class CircuitBreakerStat extends KNNStat<Boolean> {
```
jmazanec15, line 16: TODO: In mixed cluster state, this should fall back to reading from the (deprecated) index setting.

kotwanikunal:
@jmazanec15  - I like the overall idea of getting rid of the cluster level breaker and relying on the node level breakers.
The concern I have here is - we still trip up the entire cluster instead of just the node. What do you think of making it act up per node? 
It heads in the direction of rejecting per node requests based on the current node load, and we can make the coordinator rely on retrying on another node instead. 

It will get us closer to the backpressure constructs: https://opensearch.org/blog/shard-indexing-backpressure-in-opensearch/

jmazanec15:
> The concern I have here is - we still trip up the entire cluster instead of just the node. What do you think of making it act up per node?

@kotwanikunal Why do you say that? We only trip up per node.

src/main/java/org/opensearch/knn/index/memory/NativeMemoryCacheManager.java
```diff
@@ -500,4 +486,21 @@ private void startMaintenance(Cache<String, NativeMemoryAllocation> cacheInstanc
 
         maintenanceTask = threadPool.scheduleWithFixedDelay(cleanUp, interval, ThreadPool.Names.MANAGEMENT);
     }
+
+    private void circuitBreakerUpdater() {
+        Runnable runnable = () -> {
```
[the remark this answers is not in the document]
jmazanec15: This code was actually in the KNNCircuitBreaker class: https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/KNNCircuitBreaker.java#L60-L109. I moved it to the cache manager because it removes the circular dependency between the 2. However, the logic remains consistent

src/main/java/org/opensearch/knn/plugin/stats/CircuitBreakerStat.java
jmazanec15: I fixed this in latest revision

kotwanikunal:
> > The concern I have here is - we still trip up the entire cluster instead of just the node. What do you think of making it act up per node?
> 
> @kotwanikunal Why do you say that? We only trip up per node.



> > The concern I have here is - we still trip up the entire cluster instead of just the node. What do you think of making it act up per node?
> 
> @kotwanikunal Why do you say that? We only trip up per node.

`CircuitBreakerStat` is still marked as a cluster level stat. I was wondering if we could get away from maintaining it at the cluster level all together.

jmazanec15:
@kotwanikunal Oh I see. Yeah I figured thatd be a breaking change and so couldnt remove it without deprecation. Hence, the stats update.

kotwanikunal:
> @kotwanikunal Oh I see. Yeah I figured thatd be a breaking change and so couldnt remove it without deprecation. Hence, the stats update. I might be mistaken

I think it makes sense. We missed the deprecation train for this one. We have to wait until 4.0 now 😢 
Looks good to me overall.

src/main/java/org/opensearch/knn/index/KNNSettings.java
```diff
@@ -317,7 +313,8 @@ public class KNNSettings {
         KNN_CIRCUIT_BREAKER_TRIGGERED,
         false,
         NodeScope,
-        Dynamic
+        Dynamic,
+        Setting.Property.Deprecated
```
shatejas, line 317: Out of curiosity - we can't delete this for 3.0?

jmazanec15: No it has to be deprecated for upgrade purposes. We didnt deprecate it in time.

src/main/java/org/opensearch/knn/plugin/stats/KNNStatFetchContext.java
```diff
@@ -0,0 +1,28 @@
...
+package org.opensearch.knn.plugin.stats;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Additional context needed for fetching KNN stats.
+ */
+public class KNNStatFetchContext {
+    private final Map<String, Map<String, Object>> contexts;
```
shatejas, line 15: Can you please add documentation on the context - an example would be helpful on this nested map

src/main/java/org/opensearch/knn/plugin/stats/KNNStat.java
```diff
@@ -31,7 +40,11 @@ public KNNStat(Boolean clusterLevel, Supplier<T> supplier) {
      * @return boolean that is true if the stat is clusterLevel; false otherwise
      */
     public Boolean isClusterLevel() {
-        return clusterLevel;
+        return isClusterLevel;
+    }
+
+    public ActionListener<Void> setupContext(KNNStatFetchContext knnStatFetchContext, ActionListener<Void> actionListener) {
```
shatejas, line 46: nit: documentation please

src/main/java/org/opensearch/knn/plugin/stats/KNNStatFetchContext.java
jmazanec15: I got rid of this class. Basically, each stat has the chance to collect additional information via async cluster calls in order to support collecting cluster information for the stats api.

jmazanec15: I switched this up to be less extendabile. Instead, cluster stats can rely on node stats. This will cap transport calls to 2.

jmazanec15:
@kotwanikunal @shatejas Can you take a look again when you get the chance? The stats are similar except instead of a new transport action, I just reuse the stats transport action and make a single node stats call before making the complete stats call..

jmazanec15:
> @jmazanec15 can you add the output of how new stats will look like? Also, if a user has to know a CB is triggered or not, how user will identify it?

There wont be any change to the stats api interface. The only change is how we set the cluster circuit_breaker_triggered stat. Before, we were just reading from the setting. Now, we are sending out a request to all nodes to get the cache_capacity_reached stat and if any are true, then we return true. 

> With this new logic should I assume that if CB is triggered for a node all the indexing requests on that node will fail correct? if the earlier statement is true then what will happen if a node has a replica so will the replica ingestion will fail only?

Yes, if node has CB tripped (instead of cluster) we will fail the request. This change is https://github.com/opensearch-project/k-NN/pull/2567/files#diff-9dad8c6bb50a281c98d7cee29de3491e73a561f530d5e0a029d7deb860cf1698. It gets called here: https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java#L690. So, we would be unable to parse the index request, so replica would fail.

src/main/java/org/opensearch/knn/plugin/transport/KNNStatsTransportAction.java
```diff
@@ -62,6 +70,40 @@ public KNNStatsTransportAction(
...
+        if (dependentStats.stream().anyMatch(stat -> knnStats.getClusterStats().containsKey(stat))) {
+            throw new IllegalStateException("Cluster stats cannot depend on other cluster stats.");
+        }
+
+        if (dependentStats.isEmpty() == false) {
+            KNNStatsRequest knnStatsRequest = new KNNStatsRequest();
+            // Add the stats makes sure that we dont recurse infinitely.
+            dependentStats.forEach(knnStatsRequest::addStat);
+            client.execute(KNNStatsAction.INSTANCE, knnStatsRequest, ActionListener.wrap(knnStatsResponse -> {
+                knnNodeStatAggregation = new KNNNodeStatAggregation(knnStatsResponse.getNodes());
```
shatejas, line 101: Is this action singleton or a new action object is created per request? My concern is will this aggregation object remain the same with two threads have the request in parallel?

navneet1v:
> Yes, if node has CB tripped (instead of cluster) we will fail the request. This change is https://github.com/opensearch-project/k-NN/pull/2567/files#diff-9dad8c6bb50a281c98d7cee29de3491e73a561f530d5e0a029d7deb860cf1698. It gets called here: https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java#L690. So, we would be unable to parse the index request, so replica would fail.

Thanks @jmazanec15 for providing the info. I am bit concerned now, since if replicas fails then indexing is done partially and other cases can come. How we are thinking about those cases?

jmazanec15:
> Thanks @jmazanec15 for providing the info. I am bit concerned now, since if replicas fails then indexing is done partially and other cases can come. How we are thinking about those cases?

This will be consistent to if the jvm circuit breaker for a node is tripped and it cannot send the request to the replica node. I dont think it will fail indexing on the primary.

src/main/java/org/opensearch/knn/plugin/transport/KNNStatsTransportAction.java
jmazanec15: Good catch. Will associate it with the request.

jmazanec15: fixed

kotwanikunal:
> > Thanks @jmazanec15 for providing the info. I am bit concerned now, since if replicas fails then indexing is done partially and other cases can come. How we are thinking about those cases?
> 
> This will be consistent to if the jvm circuit breaker for a node is tripped and it cannot send the request to the replica node. I dont think it will fail indexing on the primary.

It should be consistent irrespective of where it is triggered. Jack's changes focus on changing the tracking mechanism and not the overall behavior from what I can think of.

Today, if a circuit breaker is triggered before the replica shard indexes - we would fall into the same scenario.

jmazanec15:
Yes, so if CB is tripped during indexing on replica action, they will retry the operation (https://github.com/opensearch-project/OpenSearch/blob/f6d6aa61e5039e4c6143cc25a71c3e448572dd33/server/src/main/java/org/opensearch/action/support/replication/ReplicationOperation.java#L329), but other than that, it will fail the action in the same way. So this would be consistent with how core handles CB.

jmazanec15:
We could also change the exception KnnCircuitBreakerException to CircuitBreakingException so that expected retries happen. This is done here for snapshot: https://github.com/opensearch-project/OpenSearch/blob/f6d6aa61e5039e4c6143cc25a71c3e448572dd33/server/src/main/java/org/opensearch/action/admin/cluster/snapshots/status/TransportSnapshotsStatusAction.java#L710

navneet1v:
@jmazanec15 in the stats api response will we now be seeing node level CB value and also the top level CB value?

jmazanec15:
> @jmazanec15 thanks for the deep-dive. Can we make the behavior as close as possible to text based indices CB exceptions.

Sure, the problem is that any exception thrown in the mapper gets wrapped: `internalParseDocument(mapping, metadataFieldsMappers, context, parser);.` So its not going to be able to retry easily even if we throw the CBException. That being said, Im wondering if we should (re)move this check on the circuit breaker. We could put it in the onIndexModule.addIndexOperationListener##preIndex. This might be better because it actually would be used before the index operation occurs: https://github.com/opensearch-project/OpenSearch/blob/2e4cc8c6e12f0e5fdfe9274da0126e81f95f59b3/server/src/main/java/org/opensearch/index/shard/IndexShard.java#L1199. 

> @jmazanec15 in the stats api response will we now be seeing node level CB value and also the top level CB value? can you paste a response from your testing

Sure there is no change in this/.

```
curl-os GET /_cat/indices?v
health status index         uuid                   pri rep docs.count docs.deleted store.size pri.store.size
yellow open   target_index2 G-Nig9ziTw2Qmgke3s0QjQ   1   1          1            0      4.3kb          4.3kb
7cf34dd6bba9:k-NN-1 jmazane$
7cf34dd6bba9:k-NN-1 jmazane$ curl-os GET /_plugins/_knn/stats?pretty
{
  "_nodes" : {
    "total" : 3,
    "successful" : 3,
    "failed" : 0
  },
  "cluster_name" : "integTest",
  "circuit_breaker_triggered" : false,
  "model_index_status" : null,
  "nodes" : {
    "J5C9PgClRcW7PV3GfVNCGA" : {
      "max_distance_query_with_filter_requests" : 0,
      "graph_memory_usage_percentage" : 0.0,
      "graph_query_requests" : 0,
      "graph_memory_usage" : 0,
      "cache_capacity_reached" : false,
      "load_success_count" : 0,
      "training_memory_usage" : 0,
      "indices_in_cache" : { },
      "script_query_errors" : 0,
      "hit_count" : 0,
      "knn_query_requests" : 0,
      "total_load_time" : 0,
      "miss_count" : 0,
      "min_score_query_requests" : 0,
      "knn_query_with_filter_requests" : 0,
      "training_memory_usage_percentage" : 0.0,
      "max_distance_query_requests" : 0,
      "lucene_initialized" : false,
      "graph_index_requests" : 0,
      "faiss_initialized" : false,
      "load_exception_count" : 0,
      "training_errors" : 0,
      "min_score_query_with_filter_requests" : 0,
      "eviction_count" : 0,
      "nmslib_initialized" : false,
      "script_compilations" : 0,
      "script_query_requests" : 0,
      "graph_stats" : {
        "merge" : {
          "current" : 0,
          "total" : 0,
          "total_time_in_millis" : 0,
          "current_docs" : 0,
          "total_docs" : 0,
          "total_size_in_bytes" : 0,
          "current_size_in_bytes" : 0
        },
        "refresh" : {
          "total" : 0,
          "total_time_in_millis" : 0
        }
      },
      "graph_query_errors" : 0,
      "indexing_from_model_degraded" : false,
      "graph_index_errors" : 0,
      "training_requests" : 0,
      "script_compilation_errors" : 0
    },
    "WYIjPvCcQHKvQGoMFEATow" : {
      "max_distance_query_with_filter_requests" : 0,
      "graph_memory_usage_percentage" : 0.0,
      "graph_query_requests" : 0,
      "graph_memory_usage" : 0,
      "cache_capacity_reached" : false,
      "load_success_count" : 0,
      "training_memory_usage" : 0,
      "indices_in_cache" : { },
      "script_query_errors" : 0,
      "hit_count" : 0,
      "knn_query_requests" : 0,
      "total_load_time" : 0,
      "miss_count" : 0,
      "min_score_query_requests" : 0,
      "knn_query_with_filter_requests" : 0,
      "training_memory_usage_percentage" : 0.0,
      "max_distance_query_requests" : 0,
      "lucene_initialized" : false,
      "graph_index_requests" : 0,
      "faiss_initialized" : false,
      "load_exception_count" : 0,
      "training_errors" : 0,
      "min_score_query_with_filter_requests" : 0,
      "eviction_count" : 0,
      "nmslib_initialized" : false,
      "script_compilations" : 0,
      "script_query_requests" : 0,
      "graph_stats" : {
        "refresh" : {
          "total_time_in_millis" : 0,
          "total" : 0
        },
        "merge" : {
          "current" : 0,
          "total" : 0,
          "total_time_in_millis" : 0,
          "current_docs" : 0,
          "total_docs" : 0,
          "total_size_in_bytes" : 0,
          "current_size_in_bytes" : 0
        }
      },
      "graph_query_errors" : 0,
      "indexing_from_model_degraded" : false,
      "graph_index_errors" : 0,
      "training_requests" : 0,
      "script_compilation_errors" : 0
    },
    "XZKc9zROR_OHbSU7KlWYKQ" : {
      "max_distance_query_with_filter_requests" : 0,
      "graph_memory_usage_percentage" : 0.0,
      "graph_query_requests" : 0,
      "graph_memory_usage" : 0,
      "cache_capacity_reached" : false,
      "load_success_count" : 0,
      "training_memory_usage" : 0,
      "indices_in_cache" : { },
      "script_query_errors" : 0,
      "hit_count" : 0,
      "knn_query_requests" : 0,
      "total_load_time" : 0,
      "miss_count" : 0,
      "min_score_query_requests" : 0,
      "knn_query_with_filter_requests" : 0,
      "training_memory_usage_percentage" : 0.0,
      "max_distance_query_requests" : 0,
      "lucene_initialized" : false,
      "graph_index_requests" : 0,
      "faiss_initialized" : false,
      "load_exception_count" : 0,
      "training_errors" : 0,
      "min_score_query_with_filter_requests" : 0,
      "eviction_count" : 0,
      "nmslib_initialized" : false,
      "script_compilations" : 0,
      "script_query_requests" : 0,
      "graph_stats" : {
        "refresh" : {
          "total_time_in_millis" : 0,
          "total" : 0
        },
        "merge" : {
          "current" : 0,
          "total" : 0,
          "total_time_in_millis" : 0,
          "current_docs" : 0,
          "total_docs" : 0,
          "total_size_in_bytes" : 0,
          "current_size_in_bytes" : 0
        }
      },
      "graph_query_errors" : 0,
      "indexing_from_model_degraded" : false,
      "graph_index_errors" : 0,
      "training_requests" : 0,
      "script_compilation_errors" : 0
    }
  }
}
```

navneet1v:
> Sure, the problem is that any exception thrown in the mapper gets wrapped: `internalParseDocument(mapping, metadataFieldsMappers, context, parser);.` So its not going to be able to retry easily even if we throw the CBException. That being said, Im wondering if we should (re)move this check on the circuit breaker. We could put it in the onIndexModule.addIndexOperationListener##preIndex. This might be better because it actually would be used before the index operation occurs: https://github.com/opensearch-project/OpenSearch/blob/2e4cc8c6e12f0e5fdfe9274da0126e81f95f59b3/server/src/main/java/org/opensearch/index/shard/IndexShard.java#L1199.

Yes I think we should do this.

navneet1v:
> Sure there is no change in this/.

Will it be better to have the node level CB info too. Since now the CB is moved from top level to node level. Thoughts?

jmazanec15:
I discussed with @shatejas and thinking about this more, I think that we should have the circuit breaker established at the node level as opposed to cluster level. However, I think that we should change the behavior to reject query requests when it is tripped as opposed to blocking index parsing requests.

The problem with blocking the index parsing requests is that we do not actually integrate indexing with the native memory management. So, we are blocking requests that may not even use off heap memory. Additionally, because of where we fail the request (in the mapper) users do not even get the proper exception, they get a Mapper Parse Exception.

That being said, in the future, I think we should predict off heap memory by indexing and trip the circuit breaker with it. We may or may not choose to fail the request. Additionally, we can publish this info cluster wide via cluster state, which can be used to make routing decisions.

@navneet1v @kotwanikunal What do you guys think? Is there any concern with blocking search requests when the CB is tripped?

PR #2575, opened 2025-03-04
Improve Streaming Compatibility Issue for MethodComponetContext and Remove OpenDistro URL

Vikasht34:
### Description
Improve Streaming Compatibility Issue for MethodComponetContext and Remove OpenDistro URL

### Related Issues

### Check List


navneet1v (approved):

src/main/java/org/opensearch/knn/index/engine/MethodComponentContext.java
```diff
@@ -414,7 +425,10 @@ public void writeTo(StreamOutput out) throws IOException {
         // do not write if parameters is null. Make sure this is in sync with the fellow read method. For more
         // information, refer to https://github.com/opensearch-project/k-NN/issues/353.
         if (this.parameters != null) {
+            out.writeBoolean(true);
```
jmazanec15, line 428: We need to add same version check here as well.

navneet1v: @jmazanec15 can you share why you think we need a version check here? I also thought about while reviewing the PR. Since this in write path and new code will only be deployed in 3.0 upgraded nodes, then why we would still need the version checks?

jmazanec15: Right, in mixed cluster state, we will still communicate with old nodes. So its best to mirror logic of read

jmazanec15 (approved):
LGTM - thanks @Vikasht34

naveentatikonda (approved):

PR #2560, opened 2025-02-25
[Remote Vector Index Build] Introduce RemoteIndexClient skeleton and Build Request construction

owenhalpert:
### Description
First PR for #2518 (Client Skeleton, Build Request)

This PR introduces the initial implementation of the Remote Index client.
In this PR, I have implemented:
- Client interface + HTTP implementation skeleton, showing overview of the client creation, code flow, etc
  - Updated code flow in RemoteIndexBuildStrategy based on finalized method parameters/return types
- Programmatic build request construction
- All client-related settings
- Extendable RemoteBuildRequest interface and class
- Client retry strategy
- Unit tests

In the next PR's, I will implement:
- Polling mechanism
- Metric collection
- Complete unit testing

### Related Issues
#2391 Meta Issue
#2393 HLD by @jed326 
#2518 My LLD

### Check List
- [X] New functionality includes testing.
~~- [ ] New functionality has been documented.~~
~~- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).~~
- [X] Commits are signed per the DCO using `--signoff`.
~~- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).~~

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/k-NN/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #2576, opened 2025-03-04
[Remote Index Client] Vector client polling mechanism

owenhalpert:
### Description
This PR is a followup to #2560. This PR adds the polling mechanism for the client to poll the server.

This PR does not include:
- Metric collection
which will go in the next PR.

### Related Issues
#2560 PR 1
#2518 LLD
#2391 Meta issue

### Check List


src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/DefaultVectorRepositoryAccessor.java
```diff
@@ -65,18 +58,13 @@ public void writeToRepository(
         VectorDataType vectorDataType,
         Supplier<KNNVectorValues<?>> knnVectorValuesSupplier
     ) throws IOException, InterruptedException {
-        assert repository != null;
```
jed326, line 68: Let's keep the assertion,
```
assert blobContainer !- null;
```

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
```diff
@@ -126,19 +131,24 @@ public void buildAndWriteIndex(BuildIndexParams indexInfo) throws IOException {
             log.debug("Repository write took {} ms for vector field [{}]", time_in_millis, indexInfo.getFieldName());
 
             RemoteIndexClient client = RemoteIndexClientFactory.getRemoteIndexClient();
-            RemoteBuildRequest request = new RemoteBuildRequest(indexSettings, indexInfo, getRepository().getMetadata(), blobName);
+            RemoteBuildRequest request = new RemoteBuildRequest(
+                indexSettings,
+                indexInfo,
+                repository.getMetadata(),
+                blobPath.buildAsString() + blobName
```
jed326, line 138: Does this need a another delimiter still? Maybe we should just do `blobPath.add(blobName).buildAsString()`.

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
```diff
@@ -165,4 +173,13 @@ private BlobStoreRepository getRepository() throws RepositoryMissingException {
...
     }
+
+    /**
+     * @param blobStoreRepository {@link BlobStoreRepository} containing the blob container
+     * @return {@link BlobContainer} referencing the location for vector upload and graph download
+     * @throws RepositoryMissingException if repository is not registered or if {@link KNNSettings#KNN_REMOTE_VECTOR_REPO_SETTING} is not set
+     */
+    private BlobContainer getBlobContainer(BlobStoreRepository blobStoreRepository, BlobPath path) throws RepositoryMissingException {
+        return blobStoreRepository.blobStore().blobContainer(path);
+    }
```
jed326, line 184: Given this is only used in 1 spot maybe we don't actually need this helper method.

src/main/java/org/opensearch/knn/index/remote/KNNRemoteConstants.java
```diff
@@ -20,8 +20,24 @@ public class KNNRemoteConstants {
...
     // HTTP implementation
+    public static final String BASIC_PREFIX = "Basic ";
     public static final String BUILD_ENDPOINT = "/_build";
     public static final String STATUS_ENDPOINT = "/_status";
+
+    // Status response keys
+    public static final String TASK_STATUS = "task_status";
+    public static final String RUNNING_INDEX_BUILD = "RUNNING_INDEX_BUILD";
+    public static final String COMPLETED_INDEX_BUILD = "COMPLETED_INDEX_BUILD";
+    public static final String FAILED_INDEX_BUILD = "FAILED_INDEX_BUILD";
```
jed326, line 39: Wondering if this should be an Enum instead

src/main/java/org/opensearch/knn/index/remote/KNNRemoteConstants.java
```diff
@@ -20,8 +20,24 @@ public class KNNRemoteConstants {
...
+    public static final String FILE_NAME = "file_name";
+    public static final String ERROR_MESSAGE = "error_message";
+    public static final String MOCK_FILE_NAME = "graph.faiss";
```
jed326, line 42: Keep test info out of src

src/main/java/org/opensearch/knn/index/remote/RemoteIndexPoller.java
```diff
@@ -0,0 +1,61 @@
...
+import org.opensearch.common.unit.TimeValue;
+import org.opensearch.knn.index.KNNSettings;
+
+import java.io.IOException;
+
+import static org.opensearch.knn.index.remote.KNNRemoteConstants.COMPLETED_INDEX_BUILD;
+import static org.opensearch.knn.index.remote.KNNRemoteConstants.FAILED_INDEX_BUILD;
+import static org.opensearch.knn.index.remote.KNNRemoteConstants.RUNNING_INDEX_BUILD;
+
+public class RemoteIndexPoller {
```
jed326, line 17: This can probably be package private

jed326:
Thanks @owenhalpert, the overall approach looks good to me, just needs some cleanup in a few places.

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
owenhalpert: Resolved offline — the current approach is fine.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 * [Remote Vector Index Build] Implement vector data upload and vector data size threshold setting [#2550](https://github.com/opensearch-project/k-NN/pull/2550)
 * [Remote Vector Index Build] Implement data download and IndexOutput write functionality [#2554](https://github.com/opensearch-project/k-NN/pull/2554)
 * [Remote Vector Index Build] Introduce Client Skeleton + basic Build Request implementation [#2560](https://github.com/opensearch-project/k-NN/pull/2560)
+* [Remote Vector Index Build] Client polling mechanism
```
kotwanikunal, line 13: nit: Missing link for the change

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/DefaultVectorRepositoryAccessor.java
```diff
@@ -65,18 +58,14 @@ public void writeToRepository(
...
-        // Get the blob container based on blobName and the repo base path. This is where the blobs will be written to.
-        BlobPath path = repository.basePath().add(indexSettings.getUUID() + VECTORS_PATH);
-        BlobContainer blobContainer = repository.blobStore().blobContainer(path);
-
+        assert blobContainer != null;
```
kotwanikunal, line 61: assserts do not work in prod. Can we do null checks instead?

kotwanikunal:
On a high level - looking at the changes - I am not sure if we should add in all the client/HTTP constructs directly into the plugin. The job status constants, client configuration seems like a better fit for a client package. 
This adds in a direct dependency between the build service and the kNN plugin - which I feel can be abstracted out.

navneet1v:
> On a high level - looking at the changes - I am not sure if we should add in all the client/HTTP constructs directly into the plugin. The job status constants, client configuration seems like a better fit for a client package. This adds in a direct dependency between the build service and the kNN plugin - which I feel can be abstracted out.

@kotwanikunal , since remote index build service is not an hosted service like S3 which vends out a HTTP client. Also, we don't want to maintain another repo as the maintainence overhead will be high. Hence the client code is better suited to be in k-NN repo only. If you see right now, the code is abstracted in a different java package. Now if you think we should move the java package to a separate module then that is something which can be easily achieved in upcoming iterations. Let us know your thoughts.

CHANGELOG.md
owenhalpert: Thanks

src/main/java/org/opensearch/knn/index/remote/KNNRemoteConstants.java
owenhalpert: Resolved offline — if these evolve into having some common functionalities beyond their string representations we can explore this.

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/DefaultVectorRepositoryAccessor.java
jed326: All the logic of getting the blob container comes from OpenSearch core so the assertion is just a sanity check on that

src/main/java/org/opensearch/knn/index/engine/faiss/Faiss.java
```diff
@@ -121,9 +127,31 @@ public ResolvedMethodContext resolveMethod(
...
+        while (parser.nextToken() != XContentParser.Token.END_OBJECT) {
+            if (parser.currentToken() == XContentParser.Token.FIELD_NAME) {
+                String fieldName = parser.currentName();
+                parser.nextToken();
+                if (NAME.equals(fieldName)) {
+                    return parser.text();
+                }
+            }
+        }
+        return null;
```
jed326, line 154: I think it was mentioned before, but can we track in a GitHub issue somewhere to improve this parsing? I think that's a generic problem in k-NN and not specific to remote index build too.

src/test/java/org/opensearch/knn/index/engine/KNNEngineTests.java
```diff
@@ -5,17 +5,32 @@
...
 import java.util.Arrays;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.stream.Collectors;
 
+import static org.mockito.Mockito.when;
+import static org.opensearch.knn.common.KNNConstants.*;
+import static org.opensearch.knn.index.SpaceType.*;
+import static org.opensearch.knn.index.VectorDataType.*;
```
jed326, line 25: Don't use `*` imports

src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexBuildStrategyFactory.java
```diff
@@ -64,7 +64,7 @@ public NativeIndexBuildStrategy getBuildStrategy(
         if (KNNFeatureFlags.isKNNRemoteVectorBuildEnabled()
             && repositoriesServiceSupplier != null
             && indexSettings != null
-            && knnEngine.supportsRemoteIndexBuild()
+            && knnEngine.supportsRemoteIndexBuild(fieldInfo.attributes())
```
navneet1v, line 67: why cannot we just pass the whole fieldInfo Object, rather than passing attributes.

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/DefaultVectorRepositoryAccessor.java
navneet1v: @jed326 if assertion fails will we fallback to CPU based index builds? and will there be a useful exception which can tell what happened? If no, please add these information

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/DefaultVectorRepositoryAccessor.java
```diff
@@ -216,25 +205,15 @@ private CheckedTriFunction<Integer, Long, Long, InputStreamContainer, IOExceptio
     }
 
     @Override
-    public void readFromRepository(String path, IndexOutputWithBuffer indexOutputWithBuffer) throws IOException {
-        if (path == null || path.isEmpty()) {
+    public void readFromRepository(String fileName, IndexOutputWithBuffer indexOutputWithBuffer) throws IOException {
+        if (fileName == null || fileName.isEmpty()) {
```
navneet1v, line 209: use StringUtils.isEmpty function here.

src/main/java/org/opensearch/knn/index/engine/faiss/Faiss.java
navneet1v: +1

src/main/java/org/opensearch/knn/index/engine/faiss/FaissHNSWMethod.java
```diff
@@ -187,4 +193,58 @@ private static int getMFromIndexDescription(String indexDescription) {
...
+    /**
+     * Return whether this engine/method supports remote build.
+     * @param attributes
+     * @return true if remote build is supported, false otherwise
+     * @throws IOException
+     */
+    static boolean supportsRemoteIndexBuild(Map<String, String> attributes) throws IOException {
+        String parametersJson = attributes.get("parameters");
+        String encoderName = getEncoderName(parametersJson);
+        return "flat".equals(encoderName);
```
navneet1v, line 206: use KNNConstants.ENCODER_FLAT constant here.

src/main/java/org/opensearch/knn/index/remote/RemoteBuildStatusResponse.java
```diff
@@ -0,0 +1,65 @@
...
+                    throw new IOException("Invalid response format, unknown field: " + currentFieldName);
+                }
+            }
+        }
+        if (StringUtils.isBlank(builder.taskStatus)) {
+            throw new IOException("Invalid response format, missing " + TASK_STATUS);
+        }
+        if (COMPLETED_INDEX_BUILD.equals(builder.taskStatus) && StringUtils.isBlank(builder.fileName)) {
+            throw new IOException("Invalid response format, missing " + FILE_NAME + " for completed status");
+        }
```
navneet1v, line 62: should this validation be part of fromXContent() ? I feel we should move this validation out of this function and put it at the place where we are validating the status and other attributes.

src/main/java/org/opensearch/knn/index/remote/RemoteIndexPoller.java
```diff
@@ -0,0 +1,61 @@
...
+    @SuppressWarnings("BusyWait")
+    RemoteBuildStatusResponse pollRemoteEndpoint(RemoteBuildResponse remoteBuildResponse) throws InterruptedException, IOException {
+        long startTime = System.currentTimeMillis();
+        long timeout = ((TimeValue) KNNSettings.state().getSettingValue(KNNSettings.KNN_REMOTE_BUILD_CLIENT_TIMEOUT)).getMillis();
+        long pollInterval = ((TimeValue) (KNNSettings.state().getSettingValue(KNNSettings.KNN_REMOTE_BUILD_CLIENT_POLL_INTERVAL)))
+            .getMillis();
+
+        // Initial delay to allow build service to process the job and store the ID before getting its status.
+        // TODO tune default based on benchmarking
+        Thread.sleep(pollInterval * 3);
```
navneet1v, line 41: what is this magic number 3? better to move this as a constant.

src/main/java/org/opensearch/knn/index/remote/RemoteIndexHTTPClient.java
```diff
@@ -124,13 +124,45 @@ private HttpPost getHttpPost(String jsonRequest) {
...
+     * @param remoteBuildResponse containing job_id from the server response used to track the job
+     * @return RemoteBuildStatusResponse containing the path to the completed index
+     */
     @Override
-    public RemoteStatusResponse awaitVectorBuild(RemoteBuildResponse remoteBuildResponse) {
-        throw new NotImplementedException();
+    public RemoteBuildStatusResponse awaitVectorBuild(RemoteBuildResponse remoteBuildResponse) throws InterruptedException, IOException {
+        RemoteIndexPoller remoteIndexPoller = new RemoteIndexPoller(this);
+        return remoteIndexPoller.pollRemoteEndpoint(remoteBuildResponse);
+    }
```
navneet1v, line 135: when I mentioned that Polling function should not be part of client I mean the Poller class should call the `awaitVectorBuild` api from outside. What is being done here is exactly reversed. This is the reason why I still think the api name should be `getIndexBuildStatus` rather than `awaitVectorBuild`. Because polling is one way to wait for the IndexBuild to complete. 

I think `RemoteIndexPoller -> getIndexBuildStatus()` rather than `awaitVectorBuild() -> RemoteIndexPoller()`

src/main/java/org/opensearch/knn/index/remote/RemoteIndexPoller.java
```diff
@@ -0,0 +1,61 @@
...
+                    String errorMessage = remoteBuildStatusResponse.getErrorMessage();
+                    if (errorMessage != null) {
+                        throw new InterruptedException("Index build failed: " + errorMessage);
+                    }
+                    throw new InterruptedException("Index build failed without an error message.");
+                case RUNNING_INDEX_BUILD:
+                    Thread.sleep(pollInterval);
+            }
+        }
+        throw new InterruptedException("Build timed out, falling back to CPU build.");
```
navneet1v, line 59: please add details on how much time was the timeout and for how long the code waited before giving up in this exception.

src/main/java/org/opensearch/knn/index/remote/RemoteIndexHTTPClient.java
```diff
@@ -124,13 +124,45 @@ private HttpPost getHttpPost(String jsonRequest) {
...
-    * @return RemoteStatusResponse containing the path to the completed index
-    */
+     * Await the completion of the index build using a {@link RemoteIndexPoller}.
+     * @param remoteBuildResponse containing job_id from the server response used to track the job
+     * @return RemoteBuildStatusResponse containing the path to the completed index
+     */
     @Override
-    public RemoteStatusResponse awaitVectorBuild(RemoteBuildResponse remoteBuildResponse) {
-        throw new NotImplementedException();
+    public RemoteBuildStatusResponse awaitVectorBuild(RemoteBuildResponse remoteBuildResponse) throws InterruptedException, IOException {
```
navneet1v, line 132: awaitVectorBuild should accept a input object RemoteBuildStatusRequest rather than `RemoteBuildResponse`

navneet1v (changes requested):

src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexBuildStrategyFactory.java
owenhalpert: In my previous PR I used BuildIndexParams.parameters instead of the whole BuildIndexParams in a similar way, because of feedback from @Vikasht34 who suggested I pass only the minimum necessary data between these classes.

src/main/java/org/opensearch/knn/index/remote/RemoteBuildStatusResponse.java
owenhalpert: I can move this to the poller class since it's a little bit more specific, but the other checks I think should stay in the parser (I think erroring on unexpected response formats, like missing a status, makes sense for a parser).

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/DefaultVectorRepositoryAccessor.java
jed326: My point is there is not any way for `blobContainer` to be `null`,  and none of the methods in the calling path to retrieve `blobContainer` throw any checked exceptions. The only thing that can throw an exception is `getRepository`, and creating a blobContainer on top of a repository reference is more or less just setting the file path to be written/read to. This assertion just helps make sure in tests if we do try to test/mock this method we are properly configuring things.

src/main/java/org/opensearch/knn/index/remote/RemoteIndexHTTPClient.java
owenhalpert: Got it. Thanks. The new flow will be that the build strategy gets the Waiter implementation from a factory method. In this case it's a Poller. The Poller uses client.getBuildStatus and does the polling all on its own. 

The client is specifically responsible for getting and sending requests — the Waiter will use the client to do so when needed.

src/main/java/org/opensearch/knn/index/remote/RemoteIndexHTTPClient.java
```diff
@@ -124,13 +124,34 @@ private HttpPost getHttpPost(String jsonRequest) {
...
-    @Override
-    public RemoteStatusResponse awaitVectorBuild(RemoteBuildResponse remoteBuildResponse) {
-        throw new NotImplementedException();
+     * Helper method to directly get the status response for a given build
+     * @param remoteBuildStatusRequest containing job ID to check
+     * @return The entire response for the status request
+     */
+    public RemoteBuildStatusResponse getBuildStatus(RemoteBuildStatusRequest remoteBuildStatusRequest) throws IOException {
+        String jobId = remoteBuildStatusRequest.getJobId();
+        HttpGet request = new HttpGet(endpoint + STATUS_ENDPOINT + "/" + jobId);
```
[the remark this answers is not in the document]
jed326: Yeah thinking about it some more I'm good with path parameter, since the jobid is going to uniquely identify our resource (the build task) over it's lifetime.

jed326 (approved):
LGTM, thanks @owenhalpert

src/main/java/org/opensearch/knn/index/engine/faiss/Faiss.java
owenhalpert: Will get this out tomorrow

src/main/java/org/opensearch/knn/index/engine/faiss/Faiss.java
```diff
@@ -121,9 +128,42 @@ public ResolvedMethodContext resolveMethod(
...
+     * Get method name from a {@link FieldInfo} formatted attributes map
+     * Example:
+     * {
+     *     "index_description": "HNSW12,Flat",
+     *     "spaceType": "l2",
+     *     "name": "hnsw",
+     *     ...
+     * }
+     */
+    private String getMethodName(String parametersJson) throws IOException {
```
navneet1v, line 152: Lets avoid IOException in method signature, it is just making the whole interfaces ugly. Lets catch the exception in the functions and then see if we want throw a proper runtime exception or we want to return some defaults.

navneet1v:
@owenhalpert please check why CIs are failing. Also, I think code is in pretty good shape and close to be getting shipped.

src/main/java/org/opensearch/knn/index/engine/faiss/FaissHNSWMethod.java
```diff
@@ -187,4 +195,82 @@ private static int getMFromIndexDescription(String indexDescription) {
...
+     *         }
+     *     }
+     *     --------------------
+     * }
+     *
+     * @param parametersJson json string of parameters (inner parameter map above)
+     * @return encoder name or null if not found
+     * @throws IOException if the json string is not valid
+     */
+    private static String getEncoderName(String parametersJson) throws IOException {
```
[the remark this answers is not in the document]
jmazanec15: Hmmm thinking about this a bit more. Do we only need the encoder on write? If so, I think itd be safest to retrieve it from the KnnVectorFieldType via mapper service (https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldType.java). We should expose it as a method in https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/engine/KNNLibraryIndexingContext.java which will be available via the mapped field type.

If we need it on read, then we have to parse it out of this. But, from my understanding, we shouldnt need it on read.

src/main/java/org/opensearch/knn/index/remote/RemoteIndexPoller.java
```diff
@@ -0,0 +1,87 @@
...
+     * @throws InterruptedException if the thread is interrupted while polling
+     * @throws IOException if an I/O error occurs
+     */
+    @SuppressWarnings("BusyWait")
+    public RemoteBuildStatusResponse awaitVectorBuild(RemoteBuildStatusRequest remoteBuildStatusRequest) throws InterruptedException,
+        IOException {
+        long startTime = System.nanoTime();
+        long timeout = KNNSettings.getRemoteBuildClientTimeout().getNanos();
+        // Thread.sleep expects millis
+        long pollInterval = KNNSettings.getRemoteBuildClientPollInterval().getMillis();
```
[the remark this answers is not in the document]
jed326: I was thinking about this too -- I lean towards no though for 3 main reasons:
1. I think it's safer / more stable to have a merge/flush task be immutable wrt settings, similar to how we are doing with the endpoint
2. It would be good to not make so may `getClusterSettings()` calls in a loop (although we could instead cache the value with a settings update consumer)
3. If we want to support cancel functionality we should properly scope that out instead of overloading it into the timeout settings

src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexBuildStrategyFactory.java
Vikasht34: As Me and Navneet has discussed , Please keep Map as it is , Rather than passing FieldInfo

src/main/java/org/opensearch/knn/index/remote/RemoteBuildStatusRequest.java
```diff
@@ -0,0 +1,20 @@
...
+import lombok.Getter;
+
+/**
+ * Request object to extract and wrap necessary parameters for the `/_status` API.
+ */
+@Getter
+public class RemoteBuildStatusRequest {
+    private final String jobId;
+
+    public RemoteBuildStatusRequest(RemoteBuildResponse remoteBuildResponse) {
```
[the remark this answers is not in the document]
Vikasht34: Why Status Build Request work based on RemoteBuildResponse rather than it has to work with minimum JobId which is enough for get status.?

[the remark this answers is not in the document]
owenhalpert: I think this keeps the design more extensible if the request/response were to contain something in addition to or instead of the job ID (e.g. tenantId).

src/main/java/org/opensearch/knn/index/engine/faiss/Faiss.java
owenhalpert: Good point. I think we can return null for the parsing methods on exceptions so the supportsRemoteIndexBuild just returns false.

src/main/java/org/opensearch/knn/index/remote/RemoteIndexPoller.java
```diff
@@ -0,0 +1,87 @@
...
+import static org.opensearch.knn.index.remote.KNNRemoteConstants.FILE_NAME;
+import static org.opensearch.knn.index.remote.KNNRemoteConstants.RUNNING_INDEX_BUILD;
+import static org.opensearch.knn.index.remote.KNNRemoteConstants.TASK_STATUS;
+
+/**
+ * Implementation of a {@link RemoteIndexWaiter} that awaits the vector build by polling.
+ */
+class RemoteIndexPoller implements RemoteIndexWaiter {
+    // The poller waits KNN_REMOTE_BUILD_CLIENT_POLL_INTERVAL * INITIAL_DELAY_FACTOR before sending the first status request
+    private static final int INITIAL_DELAY_FACTOR = 3;
```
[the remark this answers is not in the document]
owenhalpert: Going with 5

src/main/java/org/opensearch/knn/index/remote/RemoteBuildStatusRequest.java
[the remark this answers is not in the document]
Vikasht34: My Question is why constructor is taking RemoteBuildResponse , and then derive JOBID from here ? this is not flexibility , this is tight coupling with un related Response from some otheer API ...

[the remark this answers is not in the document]
Vikasht34: Why RemoteBuildStatusRequest is tied to RemoteBuildResponse , both are totally unrelated !!

src/main/java/org/opensearch/knn/index/KNNSettings.java
```diff
@@ -138,7 +138,7 @@ public class KNNSettings {
 
     // TODO: Tune these default values based on benchmarking
     public static final Integer KNN_DEFAULT_REMOTE_BUILD_CLIENT_TIMEOUT_MINUTES = 60;
-    public static final Integer KNN_DEFAULT_REMOTE_BUILD_CLIENT_POLL_INTERVAL_SECONDS = 30;
+    public static final Integer KNN_DEFAULT_REMOTE_BUILD_CLIENT_POLL_INTERVAL_SECONDS = 5;
```
Vikasht34, line 141: so If I have 3 vector Field with Dimesion 1024, 314 and 1536, are we going to have same defaults for all  and why?

src/main/java/org/opensearch/knn/index/remote/RemoteBuildStatusRequest.java
```diff
@@ -0,0 +1,25 @@
...
+ */
+@Getter
+@AllArgsConstructor
+public class RemoteBuildStatusRequest {
+    private final String jobId;
+
+    /**
+     * Creates a status request from a build response using its job ID.
+     */
+    public static RemoteBuildStatusRequest build(RemoteBuildResponse remoteBuildResponse) {
```
Vikasht34, line 22: This whole function is not required at all , it makes RemoteBuildStatusRequest and RemoteBuildResponse  coupled , and just 


@Getter
@AllArgsConstructor
@Builder
public class RemoteBuildStatusRequest {
    private final String jobId;
} Works fine

src/main/java/org/opensearch/knn/index/remote/RemoteBuildStatusRequest.java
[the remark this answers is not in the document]
Vikasht34: This is perfect !!

src/main/java/org/opensearch/knn/index/remote/RemoteIndexPoller.java
```diff
@@ -0,0 +1,87 @@
...
+import static org.opensearch.knn.index.remote.KNNRemoteConstants.COMPLETED_INDEX_BUILD;
+import static org.opensearch.knn.index.remote.KNNRemoteConstants.FAILED_INDEX_BUILD;
+import static org.opensearch.knn.index.remote.KNNRemoteConstants.FILE_NAME;
+import static org.opensearch.knn.index.remote.KNNRemoteConstants.RUNNING_INDEX_BUILD;
+import static org.opensearch.knn.index.remote.KNNRemoteConstants.TASK_STATUS;
+
+/**
+ * Implementation of a {@link RemoteIndexWaiter} that awaits the vector build by polling.
+ */
+class RemoteIndexPoller implements RemoteIndexWaiter {
```
Vikasht34, line 23: I rewrote the entier class to make polling more intellizent with Jitter , distributed safe with exponential backoff but at the same time it respects user setting values . We can adustst MIN_POLL_INTERVAL_MS and max ones ..Let me know how it looks @navneet1v 

```
import java.io.IOException;
import java.time.Duration;
import java.util.Objects;
import java.util.Random;
import org.apache.logging.log4j.LogManager;
import org.apache.logging.log4j.Logger;

class RemoteIndexPoller implements RemoteIndexWaiter {
    private static final Logger logger = LogManager.getLogger(RemoteIndexPoller.class);

    private static final int INITIAL_DELAY_FACTOR = 3;
    private static final int MAX_BACKOFF_MULTIPLIER = 8;
    private static final int MIN_POLL_INTERVAL_MS = 500;
    private static final int MAX_POLL_INTERVAL_MS = 30_000;

    private static final double JITTER_LOWER = 0.8;
    private static final double JITTER_UPPER = 1.2;

    private final RemoteIndexClient client;
    private final Random random = new Random();

    RemoteIndexPoller(RemoteIndexClient client) {
        this.client = Objects.requireNonNull(client, "RemoteIndexClient cannot be null");
    }

    @SuppressWarnings("BusyWait")
    public RemoteBuildStatusResponse awaitVectorBuild(RemoteBuildStatusRequest request) throws InterruptedException, IOException {
        long startTime = System.nanoTime();
        long timeout = KNNSettings.getRemoteBuildClientTimeout().getNanos();
        long pollInterval = KNNSettings.getRemoteBuildClientPollInterval().getMillis() * INITIAL_DELAY_FACTOR;

        Thread.sleep(applyJitter(pollInterval)); // Initial delay

        for (int attempt = 1; System.nanoTime() - startTime < timeout; attempt++) {
            RemoteBuildStatusResponse response = client.getBuildStatus(request);
            String taskStatus = Objects.requireNonNullElse(response.getTaskStatus(), "").trim();

            logger.debug("Polling attempt {}: Task status = {}", attempt, taskStatus);

            switch (taskStatus) {
                case COMPLETED_INDEX_BUILD -> {
                    String fileName = Objects.requireNonNullElse(response.getFileName(), "").trim();
                    if (fileName.isEmpty()) {
                        throw new IOException("Invalid response: Missing file name for completed status.");
                    }
                    logger.info("Remote index build completed successfully in {} seconds.",
                        Duration.ofNanos(System.nanoTime() - startTime).toSeconds());
                    return response;
                }
                case FAILED_INDEX_BUILD -> throw new InterruptedException(
                    "Remote index build failed after " + Duration.ofNanos(System.nanoTime() - startTime).toMinutes() +
                        " minutes. Error: " + response.getErrorMessage());

                case RUNNING_INDEX_BUILD -> {
                    pollInterval = Math.min(pollInterval * 2, MAX_POLL_INTERVAL_MS);
                    pollInterval = Math.max(pollInterval, MIN_POLL_INTERVAL_MS);
                    if (timeout - (System.nanoTime() - startTime) < pollInterval * 2) {
                        pollInterval = MIN_POLL_INTERVAL_MS; // Ensure last-minute checks
                    }
                    Thread.sleep(applyJitter(pollInterval));
                }
                default -> throw new IOException("Invalid task status: " + taskStatus);
            }
        }

        throw new InterruptedException("Remote index build timed out after " +
            Duration.ofNanos(System.nanoTime() - startTime).toMinutes() + " minutes.");
    }

    private long applyJitter(long baseInterval) {
        return (long) (baseInterval * (JITTER_LOWER + (random.nextDouble() * (JITTER_UPPER - JITTER_LOWER))));
    }
}

```

jed326: Thanks @Vikasht34. For the jitter aspect I can agree that would be generically good and I think would be fine to pull into this.

For exponential backoff I don't think we should use that logic for the poller for now and I think this is something we need to revisit based on benchmarking. We want to strike a balance between calling the remote index build service too often and waiting too long after the graph build has completed to get the status. For example, let's take a 1 minute poll interval. After polling 2 times (so we've waited 2 minutes), the build job is still just as likely to finish in 2min + 1ms as 2min + 2min, and I think it's better to over-index against the second case.

Based on benchmarking we should be able to establish p50/90/99 values on how long the remote index build takes for a given segment size and then we can set the polling interval as some function of that value. If the p50/p90/p99 is very narrow then exponential backoff probably wouldn't be necessary, but if it's a very wide range we can revisit this.

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
```diff
@@ -52,11 +59,13 @@ public class RemoteIndexBuildStrategy implements NativeIndexBuildStrategy {
...
         Supplier<RepositoriesService> repositoriesServiceSupplier,
         NativeIndexBuildStrategy fallbackStrategy,
-        IndexSettings indexSettings
+        IndexSettings indexSettings,
+        KNNMethodContext knnMethodContext
     ) {
         this.repositoriesServiceSupplier = repositoriesServiceSupplier;
         this.fallbackStrategy = fallbackStrategy;
         this.indexSettings = indexSettings;
+        this.knnMethodContext = knnMethodContext;
```
jed326, line 68: Can you update javadoc as well?

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
```diff
@@ -126,19 +139,28 @@ public void buildAndWriteIndex(BuildIndexParams indexInfo) throws IOException {
...
+                blobPath.buildAsString(),
+                knnMethodContext
+            );
             stopWatch = new StopWatch().start();
-            RemoteBuildResponse remoteBuildResponse = client.submitVectorBuild(request);
+            RemoteBuildResponse remoteBuildResponse = client.submitVectorBuild(buildRequest);
             time_in_millis = stopWatch.stop().totalTime().millis();
             log.debug("Submit vector build took {} ms for vector field [{}]", time_in_millis, indexInfo.getFieldName());
 
+            String jobId = remoteBuildResponse.getJobId();
```
jed326, line 154: Don't need this local variable, unless we are planning on doing validation on it

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
```diff
@@ -126,19 +139,28 @@ public void buildAndWriteIndex(BuildIndexParams indexInfo) throws IOException {
             log.debug("Repository write took {} ms for vector field [{}]", time_in_millis, indexInfo.getFieldName());
 
             RemoteIndexClient client = RemoteIndexClientFactory.getRemoteIndexClient();
-            RemoteBuildRequest request = new RemoteBuildRequest(indexSettings, indexInfo, getRepository().getMetadata(), blobName);
+            RemoteBuildRequest buildRequest = new RemoteBuildRequest(
+                indexSettings,
+                indexInfo,
+                repository.getMetadata(),
+                blobPath.buildAsString(),
```
jed326, line 146: blobPath will refer to the directory only, don't we need to pass in blobName as well for the vector/doc id file name? I think I may have missed this in the last PR.

src/main/java/org/opensearch/knn/index/remote/RemoteBuildRequest.java
```diff
@@ -81,14 +82,18 @@ public RemoteBuildRequest(
...
         this.repositoryType = repositoryType;
         this.containerName = containerName;
-        this.vectorPath = blobName + VECTOR_BLOB_FILE_EXTENSION;
-        this.docIdPath = blobName + DOC_ID_FILE_EXTENSION;
+        // remove the / automatically appended by BlobPath.buildAsString()
+        if (fullPath.endsWith("/")) {
+            fullPath = fullPath.substring(0, fullPath.length() - 1);
+        }
+        this.vectorPath = fullPath + VECTOR_BLOB_FILE_EXTENSION;
+        this.docIdPath = fullPath + DOC_ID_FILE_EXTENSION;
```
jed326, line 90: See comment in `RemoteIndexBuildStrategy` on this, I think we also need to pass in blobName still. Please add a test for this as well.

src/test/java/org/opensearch/knn/index/remote/RemoteBuildRequestTests.java
```diff
@@ -84,22 +109,74 @@ public void testBuildRequest() {
             assertEquals(2, request.getDimension());
 
             String expectedJson = "{"
```
jed326, line 111: optional: I think have a comment with the expect raw JSON here could be useful

src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexBuildStrategyFactory.java
```diff
@@ -27,6 +29,8 @@ public final class NativeIndexBuildStrategyFactory {
 
     private final Supplier<RepositoriesService> repositoriesServiceSupplier;
     private final IndexSettings indexSettings;
+    @Setter
+    private KNNMethodContext knnMethodContext;
```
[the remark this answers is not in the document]
owenhalpert: Thanks, you caught this before the CIs did! I looked into this more because IntelliJ warns against Optional fields, so I followed a suggestion to make a private getter that returns an Optional instead of making the field itself optional. Then I'm still checking `.isPresent()` like normal.

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
```diff
@@ -126,19 +140,29 @@ public void buildAndWriteIndex(BuildIndexParams indexInfo) throws IOException {
             log.debug("Repository write took {} ms for vector field [{}]", time_in_millis, indexInfo.getFieldName());
 
             RemoteIndexClient client = RemoteIndexClientFactory.getRemoteIndexClient();
-            RemoteBuildRequest request = new RemoteBuildRequest(indexSettings, indexInfo, getRepository().getMetadata(), blobName);
+            RemoteBuildRequest buildRequest = new RemoteBuildRequest(
+                indexSettings,
+                indexInfo,
+                repository.getMetadata(),
+                blobPath.add(blobName).buildAsString(),
```
jed326, line 147:
```suggestion
                blobPath.buildAsString() + blobName,
```
1. `BlobPath` class is only intended to represent directories, not the file itself (hence the ending `/`), so let's not mis-use it in this way.
2. You don't need the logic to get rid of the ending `/` this way

owenhalpert: Nice! Thanks, fixed.

jed326 (approved):
LGTM

src/main/java/org/opensearch/knn/index/engine/faiss/FaissHNSWMethod.java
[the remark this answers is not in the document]
owenhalpert: Thanks for the suggestion Jack, I made these changes to fetch the parameters through the MethodComponentContext

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
```diff
@@ -155,7 +157,11 @@ public void buildAndWriteIndex(BuildIndexParams indexInfo) throws IOException {
             RemoteBuildStatusRequest remoteBuildStatusRequest = RemoteBuildStatusRequest.builder()
                 .jobId(remoteBuildResponse.getJobId())
                 .build();
-            RemoteIndexWaiter waiter = RemoteIndexWaiterFactory.getRemoteIndexWaiter(client);
+            RemoteIndexWaiter waiter = RemoteIndexWaiterFactory.getRemoteIndexWaiter(
+                client,
+                getRemoteBuildClientTimeout().getNanos(),
+                getRemoteBuildClientPollInterval().getMillis()
```
owenhalpert, line 163: I don't love that we are passing in the poll interval here because not all Waiter implementations will need it. If we add new implementations, of course the Factory can just choose to not use these parameters, but I'm not sure of a better way to handle this besides having the factory in k-NN (or passing the whole Settings).

navneet1v: The question arises if RemoteIndexWaiter is part of the RemoteIndexBuildClient or not. I think it is outside of the client. So it is okay to access the settings directly.

src/main/java/org/opensearch/knn/index/remote/RemoteIndexPoller.java
```diff
@@ -0,0 +1,98 @@
...
+        throw new InterruptedException(
+            String.format(
+                "Remote index build timed out after %d minutes, timeout is set to %d minutes. Falling back to CPU build",
+                waitedDuration.toMinutesPart(),
+                timeoutDuration.toMinutesPart()
+            )
+        );
+    }
+
+    private void sleepWithJitter(long baseInterval) throws InterruptedException {
```
navneet1v, line 94: Lets add in the comments in terms of how the jitter will look like. So lets say first if its 5sec, then next value will be X, and then next will Y. It will help understand the code better.

navneet1v (dismissed):
Overall code looks good to me. Just 2 minor comments. Thanks @Vikasht34 for suggesting the jitter logic for polling.

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
owenhalpert: That's a good point. One reason why it might be good to expose the polling and timeout settings to the BuildStrategy would be in the future if we were to want to intelligently set them based on estimations from the workload size. Since we don't have that functionality yet, we can just fetch these values in the Poller and keep the Poller and the Waiter in the k-NN package.

src/main/java/org/opensearch/knn/index/remote/RemoteIndexPoller.java
owenhalpert: Done, thx

Vikasht34 (approved):
LGTM

navneet1v (approved):
Approving this PR and as a next step @owenhalpert can you please take this PR: https://github.com/opensearch-project/k-NN/pull/2588 and adapt it to the module level. Also please create a GH issue to move the client to remote-index-build repo.

PR #2578, opened 2025-03-05
Remove integrity checking TODO and leave up to the vendor implementation

jed326:
### Description

In this PR we are making the decision to leave integrity checking up to the individual vendor SDK and repository implementations. For example, in https://github.com/opensearch-project/OpenSearch/pull/17396 the aws s3 SDK was bumped to automatically support integrity checking.

### Integrity Checking Overview
We are making the decision to leave integrity checking to the vendor implementation in all cases. This section will give a brief overview of how this is done for S3. 

When using the aws java sdk and repository-s3, during object upload a CRC32 checksum is automatically computed by the SDK, and validated + persisted to the object store. During object download, if a checksum is present in the object store, the SDK will automatically validate the downloaded object against the checksum found in the object store.

As a future improvement, we can also manually do this integrity checking if the vendor implementation indicates that integrity checking is not supported. For example, remote store does this like so: https://github.com/opensearch-project/OpenSearch/blob/3ea0a31f6f613ddd5b7bde0053195d5b212c813d/server/src/main/java/org/opensearch/common/blobstore/transfer/RemoteTransferContainer.java#L213-L222.

Tracking the future improvement here: https://github.com/opensearch-project/k-NN/issues/2579

### Other

We have separate testing in `KnnVectorValuesInputStreamTests` that provides coverage on if the InputStream is correctly reading the `KNNVectorValues`, and we do not need to separately compute a checksum on this.

### Related Issues

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/DefaultVectorRepositoryAccessor.java
+++ b/src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/DefaultVectorRepositoryAccessor.java
@@ -74,50 +74,36 @@ public class DefaultVectorRepositoryAccessor implements VectorRepositoryAccessor
         initializeVectorValues(knnVectorValues);
         long vectorBlobLength = (long) knnVectorValues.bytesPerVector() * totalLiveDocs;
 
-        if (blobContainer instanceof AsyncMultiStreamBlobContainer) {
+        if (blobContainer instanceof AsyncMultiStreamBlobContainer asyncBlobContainer) {
             // First initiate vectors upload
             log.debug("Repository {} Supports Parallel Blob Upload", repository);
             // WriteContext is the main entry point into asyncBlobUpload. It stores all of our upload configurations, analogous to
             // BuildIndexParams
-            WriteContext writeContext = new WriteContext.Builder().fileName(blobName + VECTOR_BLOB_FILE_EXTENSION)
-                .streamContextSupplier((partSize) -> getStreamContext(partSize, vectorBlobLength, knnVectorValuesSupplier, vectorDataType))
-                .fileSize(vectorBlobLength)
-                .failIfAlreadyExists(true)
-                .writePriority(WritePriority.NORMAL)
-                // TODO: Checksum implementations -- It is difficult to calculate a checksum on the knnVectorValues as
-                // there is no underlying file upon which we can create the checksum. We should be able to create a
-                // checksum still by iterating through once, however this will be an expensive operation.
-                .uploadFinalizer((bool) -> {})
-                .doRemoteDataIntegrityCheck(false)
-                .expectedChecksum(null)
-                .build();
+            WriteContext writeContext = createWriteContext(blobName, vectorBlobLength, knnVectorValuesSupplier, vectorDataType);
 
             AtomicReference<Exception> exception = new AtomicReference<>();
             final CountDownLatch latch = new CountDownLatch(1);
-            ((AsyncMultiStreamBlobContainer) blobContainer).asyncBlobUpload(
-                writeContext,
-                new LatchedActionListener<>(new ActionListener<>() {
-                    @Override
-                    public void onResponse(Void unused) {
-                        log.debug(
-                            "Parallel vector upload succeeded for blob {} with size {}",
-                            blobName + VECTOR_BLOB_FILE_EXTENSION,
-                            vectorBlobLength
-                        );
-                    }
+            asyncBlobContainer.asyncBlobUpload(writeContext, new LatchedActionListener<>(new ActionListener<>() {
+                @Override
+                public void onResponse(Void unused) {
+                    log.debug(
+                        "Parallel vector upload succeeded for blob {} with size {}",
+                        blobName + VECTOR_BLOB_FILE_EXTENSION,
+                        vectorBlobLength
+                    );
+                }
 
-                    @Override
-                    public void onFailure(Exception e) {
-                        log.error(
-                            "Parallel vector upload failed for blob {} with size {}",
-                            blobName + VECTOR_BLOB_FILE_EXTENSION,
-                            vectorBlobLength,
-                            e
-                        );
-                        exception.set(e);
-                    }
-                }, latch)
-            );
+                @Override
+                public void onFailure(Exception e) {
+                    log.error(
+                        "Parallel vector upload failed for blob {} with size {}",
+                        blobName + VECTOR_BLOB_FILE_EXTENSION,
+                        vectorBlobLength,
+                        e
+                    );
+                    exception.set(e);
+                }
+            }, latch));
 
             // Then upload doc id blob before waiting on vector uploads
             // TODO: We wrap with a BufferedInputStream to support retries. We can tune this buffer size to optimize performance.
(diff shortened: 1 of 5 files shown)


owenhalpert (approved):

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/DefaultVectorRepositoryAccessor.java
```diff
@@ -215,6 +210,61 @@ private CheckedTriFunction<Integer, Long, Long, InputStreamContainer, IOExceptio
...
+     *
+     * @param knnVectorValues
+     * @param vectorDataType
+     * @return
+     * @throws IOException
+     */
+    @VisibleForTesting
+    long getExpectedChecksum(KNNVectorValues<?> knnVectorValues, VectorDataType vectorDataType) throws IOException {
+        initializeVectorValues(knnVectorValues);
+        CheckedInputStream checkedStream = new CheckedInputStream(
```
Gankris96, line 256: We are opening an inputStream here but not closing it right ? 
can we have try with resources here ?

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/DefaultVectorRepositoryAccessor.java
```diff
@@ -215,6 +210,61 @@ private CheckedTriFunction<Integer, Long, Long, InputStreamContainer, IOExceptio
...
+            new VectorValuesInputStream(knnVectorValues, vectorDataType),
+            new CRC32()
+        );
+        // VectorValuesInputStream#read only reads 1 vector max at a time, so no point making this buffer any larger than that
+        int bufferSize = knnVectorValues.bytesPerVector();
+        final byte[] buffer = new byte[bufferSize];
+        while (checkedStream.read(buffer, 0, bufferSize) != -1) {
```
Gankris96, line 263: im assuming this alters the checksum and thats the purpose of this empty while loop ? 
can you add some comments to make this clear ?

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/DefaultVectorRepositoryAccessor.java
jed326: Yes, this stream is not backed by any files but instead the `KNNVectorValues`, so there is nothing to close per se. This is a good point though I should still implement the close functionality where exceptions are thrown after `close()` is called.

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/DefaultVectorRepositoryAccessor.java
jed326: Yeah, the `CheckedInputStream#read` javadoc explains this, but I can add the same here too.

Gankris96 (approved):

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/DefaultVectorRepositoryAccessor.java
```diff
@@ -215,6 +216,65 @@ private CheckedTriFunction<Integer, Long, Long, InputStreamContainer, IOExceptio
...
+        boolean supportsIntegrityCheck
+    ) throws IOException {
+        return new WriteContext.Builder().fileName(blobName + VECTOR_BLOB_FILE_EXTENSION)
+            .streamContextSupplier((partSize) -> getStreamContext(partSize, vectorBlobLength, knnVectorValuesSupplier, vectorDataType))
+            .fileSize(vectorBlobLength)
+            .failIfAlreadyExists(true)
+            .writePriority(WritePriority.NORMAL)
+            .doRemoteDataIntegrityCheck(supportsIntegrityCheck)
+            .uploadFinalizer((bool) -> {})
+            .expectedChecksum(supportsIntegrityCheck ? getExpectedChecksum(knnVectorValuesSupplier.get(), vectorDataType) : null)
```
navneet1v, line 243: is this checksum getting written with the BlobContainer file? If yes why we need to add it? since doing checksum will means that we need to read the full vector file and compute the checksum.

jed326: 1. Whether or not this checksum gets written to the `blobContainer` is determined by the vendor implementation. S3 specifically does do this, but that is not guaranteed by the existing repository interface contract.
2. If we do not provide a checksum, then the value that is stored is whatever value is calculated by the SDK as it uploads the `InputStream`s. However, since we are doing multi-part uploads, this does not guard against many issues with reading the `KNNVectorValues` as multiple parts. One example is if the parts were read out of order (part C of the `KNNVectorValues` was uploaded as part D).

I'm actually not totally against saying [2] is covered by unit tests so we do not need to compute our own checksum like this as like you said it does mean we are reading the full vector file an extra time. However it does provide an extra layer of safety wrt file upload integrity.

jed326:
So it turns out the java SDK version used by core currently does not perform the checksum by default for s3. I've opened an issue in core to update this, https://github.com/opensearch-project/OpenSearch/issues/17524, I think we can pause on this PR pending discussion there.

Generally speaking I do think it's better to offload the responsibility of checksum computation to the vendor SDK, or at least to the vendor repository implementation.

shatejas (commented):
This seems to be computing checksum and writing it, Do we need to checkIntrgrity on downloads or is the underlying implementation (blobContainer.read) makes sure of the integrity when read happens?

shatejas:
> So it turns out the java SDK version used by core currently does not perform the checksum by default for s3. I've opened an issue in core to update this, [opensearch-project/OpenSearch#17524](https://github.com/opensearch-project/OpenSearch/issues/17524), I think we can pause on this PR pending discussion there.
> 
> Generally speaking I do think it's better to offload the responsibility of checksum computation to the vendor SDK, or at least to the vendor repository implementation.

So do we need to check integrity till

jed326:
Actually, it turns out that core bumped the aws sdk to >2.30 less than 1 day ago: https://github.com/opensearch-project/OpenSearch/pull/17396

After pulling in the latest changes on my local, I can see that even the doc id blob, which we are not manually computing a checksum for, has a checksum computed for it automatically by the aws SDK. With that, I am inclined to not perform our own manual checksum computations which involve reading the vector blob twice. This way, we leave it to the vendor SDK implementation to [1] compute and persist the checksum on upload and [2] verify the checksum on download.

Will update this PR accordingly.

jmazanec15 (approved):

shatejas (approved):

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/DocIdInputStream.java
```diff
@@ -41,6 +43,7 @@ public DocIdInputStream(KNNVectorValues<?> knnVectorValues) throws IOException {
 
     @Override
     public int read() throws IOException {
+        checkClosed();
```
navneet1v, line 46: why we need this check?

jed326: This is a safety check as we are not always able to use this stream in a `try-with-resources` manner when we are relinquishing control of it to the repository implementation.

jed326: This is how many other streams are implemented, for example: https://github.com/opensearch-project/OpenSearch/blob/73453718a1c85565a3f7e4309d6fa83bf9a30522/plugins/repository-s3/src/main/java/org/opensearch/repositories/s3/S3RetryingInputStream.java#L157-L160

navneet1v (approved):

Issue #2568, opened 2025-02-27
Move query building responsibility to KnnVectorFieldType and engine


jmazanec15:
## Description

Right now, our KNN query building contains a significant number of branches depending on the algo, engine and type (radial vs. knn) and with/without model. Any engine specific feature leads to more branching in the KNNQueryBuilder, KNNQuery or KNNWeight

Ideally, we should provide the parsed context to the engine and it should take care of building the query.

That being said, I have seen a convention in OpenSearch where the MappedFieldType can produce the query: See [TextFieldMapper](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/mapper/TextFieldMapper.java#L634) for example.

The KNNQueryBuilder would parse the context and then decide if it is radial or knn and then pass it on to the knnVectorFieldType with new methods:
```
Query knnQuery(KNNQueryParseContext)

Query radialQuery(RadialQueryParseContext)
```

From here, the mapped field type could delegate to the engine to validate and create the given query. This would provide for much better encapsulation and less branching.

## Related Issues
#1914

PR #2581, opened 2025-03-05
[LuceneOnFaiss - Part1] Added building blocks for memory optimized search.

0ctopus13prime:
### Description
[1/11] This is the first PR to establish the building blocks for memory optimized search.
At the moment, FAISS is the only engine that supports memory optimized search where loading vectors in demand fashion.

Note that this will be merged into the feature branch first. Unit tests + IT tests will be covered in PR-9 and PR-10.

### Related Issues

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
+++ b/src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
@@ -11,6 +11,7 @@
 
 package org.opensearch.knn.index.codec.KNN990Codec;
 
+import lombok.extern.slf4j.Slf4j;
 import org.apache.lucene.codecs.KnnVectorsReader;
 import org.apache.lucene.codecs.hnsw.FlatVectorsReader;
 import org.apache.lucene.index.ByteVectorValues;
@@ -22,12 +23,16 @@ import org.apache.lucene.search.ScoreDoc;
 import org.apache.lucene.search.TopDocs;
 import org.apache.lucene.search.TotalHits;
 import org.apache.lucene.util.Bits;
+import org.apache.lucene.util.IOSupplier;
 import org.apache.lucene.util.IOUtils;
 import org.opensearch.common.UUIDs;
 import org.opensearch.knn.index.codec.util.KNNCodecUtil;
 import org.opensearch.knn.index.codec.util.NativeMemoryCacheKeyHelper;
+import org.opensearch.knn.index.engine.KNNEngine;
 import org.opensearch.knn.index.memory.NativeMemoryCacheManager;
 import org.opensearch.knn.index.quantizationservice.QuantizationService;
+import org.opensearch.knn.memoryoptsearch.VectorSearcher;
+import org.opensearch.knn.memoryoptsearch.VectorSearcherFactory;
 import org.opensearch.knn.quantization.models.quantizationState.QuantizationState;
 import org.opensearch.knn.quantization.models.quantizationState.QuantizationStateCacheManager;
 import org.opensearch.knn.quantization.models.quantizationState.QuantizationStateReadConfig;
@@ -37,23 +42,89 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
+
+import static org.opensearch.knn.common.KNNConstants.KNN_ENGINE;
+import static org.opensearch.knn.index.mapper.KNNVectorFieldMapper.KNN_FIELD;
 
 /**
  * Vectors reader class for reading the flat vectors for native engines. The class provides methods for iterating
  * over the vectors and retrieving their values.
  */
+@Slf4j
 public class NativeEngines990KnnVectorsReader extends KnnVectorsReader {
+    private static final int RESERVE_TWICE_SPACE = 2;
+    private static final float SUFFICIENT_LOAD_FACTOR = 0.6f;
 
     private final FlatVectorsReader flatVectorsReader;
     private Map<String, String> quantizationStateCacheKeyPerField;
     private SegmentReadState segmentReadState;
     private final List<String> cacheKeys;
+    private Map<String, VectorSearcher> vectorSearchers;
 
     public NativeEngines990KnnVectorsReader(final SegmentReadState state, final FlatVectorsReader flatVectorsReader) {
         this.flatVectorsReader = flatVectorsReader;
         this.segmentReadState = state;
         this.cacheKeys = getVectorCacheKeysFromSegmentReaderState(state);
+        this.vectorSearchers = new HashMap<>(RESERVE_TWICE_SPACE * segmentReadState.fieldInfos.size(), SUFFICIENT_LOAD_FACTOR);
         loadCacheKeyMap();
+
+        //
+        // TMP(KDY) : Dynamic update will be covered in part-7. Please refer to
+        // https://github.com/opensearch-project/k-NN/issues/2401#issuecomment-2699777824
+        //
+        final boolean isMemoryOptimizedSearchEnabled = false;
+        if (isMemoryOptimizedSearchEnabled) {
+            loadMemoryOptimizedSearcher();
+        }
+    }
+
+    private IOSupplier<VectorSearcher> getIndexFileNameIfMemoryOptimizedSearchSupported(final FieldInfo fieldInfo) {
+        // Skip non-knn fields.
+        final Map<String, String> attributes = fieldInfo.attributes();
+        if (attributes == null || attributes.containsKey(KNN_FIELD) == false) {
+            return null;
+        }
+
+        // Get engine
+        final String engineName = attributes.getOrDefault(KNN_ENGINE, KNNEngine.DEFAULT.getName());
+        final KNNEngine knnEngine = KNNEngine.getEngine(engineName);
+
+        // Get memory optimized searcher from engine
+        final VectorSearcherFactory searcherFactory = knnEngine.getVectorSearcherFactory();
+        if (searcherFactory == null) {
+            // It's not supported
+            return null;
+        }
+
+        // Start creating searcher
+        final String fileName = KNNCodecUtil.getNativeEngineFileFromFieldInfo(fieldInfo, segmentReadState.segmentInfo);
+        if (fileName != null) {
+            return () -> searcherFactory.createVectorSearcher(segmentReadState.directory, fileName);
+        }
+
+        // Not supported
+        return null;
+    }
+
+    private void loadMemoryOptimizedSearcher() {
+        try {
+            for (FieldInfo fieldInfo : segmentReadState.fieldInfos) {
+                final IOSupplier<VectorSearcher> searcherSupplier = getIndexFileNameIfMemoryOptimizedSearchSupported(fieldInfo);
+                if (searcherSupplier != null) {
+                    final VectorSearcher searcher = Objects.requireNonNull(searcherSupplier.get());
+                    vectorSearchers.put(fieldInfo.getName(), searcher);
+                }
+            }
+        } catch (Exception e) {
+            // Close opened searchers first, then suppress
+            try {
+                IOUtils.closeWhileHandlingException(vectorSearchers.values());
+            } catch (Exception closeException) {
+                log.error(closeException.getMessage(), closeException);
+            }
+            throw new RuntimeException(e);
+        }
     }
 
     /**
@@ -135,6 +206,14 @@ public class NativeEngines990KnnVectorsReader extends KnnVectorsReader {
             ((QuantizationConfigKNNCollector) knnCollector).setQuantizationState(quantizationState);
             return;
         }
+
+        // Try with memory optimized searcher
+        final VectorSearcher memoryOptimizedSearcher = vectorSearchers.get(field);
+        if (memoryOptimizedSearcher != null) {
+            memoryOptimizedSearcher.search(target, knnCollector, acceptDocs);
+            return;
+        }
+
         throw new UnsupportedOperationException("Search functionality using codec is not supported with Native Engine Reader");
     }
 
@@ -197,6 +276,9 @@ public class NativeEngines990KnnVectorsReader extends KnnVectorsReader {
                 quantizationStateCacheManager.evict(cacheKey);
             }
         }
+
+        // TODO(KDY)
+        // Close all memory optimized searchers.
     }
 
     private void loadCacheKeyMap() {
--- a/src/main/java/org/opensearch/knn/index/engine/KNNEngine.java
+++ b/src/main/java/org/opensearch/knn/index/engine/KNNEngine.java
@@ -8,6 +8,7 @@ package org.opensearch.knn.index.engine;
 import com.google.common.collect.ImmutableSet;
 import org.opensearch.common.ValidationException;
 import org.opensearch.knn.index.SpaceType;
+import org.opensearch.knn.memoryoptsearch.VectorSearcherFactory;
 import org.opensearch.knn.index.engine.faiss.Faiss;
 import org.opensearch.knn.index.engine.lucene.Lucene;
 import org.opensearch.knn.index.engine.nmslib.Nmslib;
@@ -215,5 +216,10 @@ public enum KNNEngine implements KNNLibrary {
     @Override
     public boolean supportsRemoteIndexBuild() {
         return knnLibrary.supportsRemoteIndexBuild();
+    }
+
+    @Override
+    public VectorSearcherFactory getVectorSearcherFactory() {
+        return knnLibrary.getVectorSearcherFactory();
     }
 }
--- a/src/main/java/org/opensearch/knn/index/engine/KNNLibrary.java
+++ b/src/main/java/org/opensearch/knn/index/engine/KNNLibrary.java
@@ -7,6 +7,7 @@ package org.opensearch.knn.index.engine;
 
 import org.opensearch.common.ValidationException;
 import org.opensearch.knn.index.SpaceType;
+import org.opensearch.knn.memoryoptsearch.VectorSearcherFactory;
 
 import java.util.Collections;
 import java.util.List;
@@ -140,11 +141,21 @@ public interface KNNLibrary extends MethodResolver {
         return Collections.emptyList();
     }
 
-    /**
+    /*
      * Returns whether or not the engine implementation supports remote index build
      * @return true if remote index build is supported, false otherwise
      */
     default boolean supportsRemoteIndexBuild() {
         return false;
+    }
+
+    /**
+     * Create a new vector searcher factory that compatible with on Lucene search API.
+     * @return New searcher factory that returns {@link org.opensearch.knn.memoryoptsearch.VectorSearcher}
+     *         If it is not supported, it should return null.
+     *         But, if it is supported, the factory shall not return null searcher.
+     */
+    default VectorSearcherFactory getVectorSearcherFactory() {
+        return null;
     }
 }
--- a/src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissMemoryOptimizedSearcherFactory.java
+++ b/src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissMemoryOptimizedSearcherFactory.java
@@ -1,0 +1,32 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.knn.memoryoptsearch.faiss;
+
+import org.apache.lucene.store.Directory;
+import org.apache.lucene.store.IOContext;
+import org.apache.lucene.store.IndexInput;
+import org.apache.lucene.store.ReadAdvice;
+import org.opensearch.knn.memoryoptsearch.VectorSearcher;
+import org.opensearch.knn.memoryoptsearch.VectorSearcherFactory;
+
+import java.io.IOException;
+
+/**
+ * This factory returns {@link VectorSearcher} that performs vector search directly on FAISS index.
+ * Note that we pass `RANDOM` as advice to prevent the underlying storage from performing read-ahead. Since vector search naturally accesses
+ * random vector locations, read-ahead does not improve performance. By passing the `RANDOM` context, we explicitly indicate that
+ * this searcher will access vectors randomly.
+ */
+public class FaissMemoryOptimizedSearcherFactory implements VectorSearcherFactory {
+    @Override
+    public VectorSearcher createVectorSearcher(final Directory directory, final String fileName) throws IOException {
+        final IndexInput indexInput = directory.openInput(
+            fileName,
+            new IOContext(IOContext.Context.DEFAULT, null, null, ReadAdvice.RANDOM)
+        );
+        return new FaissMemoryOptimizedSearcher(indexInput);
+    }
+}
--- a/src/main/java/org/opensearch/knn/index/engine/faiss/Faiss.java
+++ b/src/main/java/org/opensearch/knn/index/engine/faiss/Faiss.java
@@ -8,12 +8,14 @@ package org.opensearch.knn.index.engine.faiss;
 import com.google.common.collect.ImmutableMap;
 import org.opensearch.knn.common.KNNConstants;
 import org.opensearch.knn.index.SpaceType;
+import org.opensearch.knn.memoryoptsearch.VectorSearcherFactory;
 import org.opensearch.knn.index.engine.KNNMethod;
 import org.opensearch.knn.index.engine.KNNMethodConfigContext;
 import org.opensearch.knn.index.engine.KNNMethodContext;
 import org.opensearch.knn.index.engine.MethodResolver;
 import org.opensearch.knn.index.engine.NativeLibrary;
 import org.opensearch.knn.index.engine.ResolvedMethodContext;
+import org.opensearch.knn.memoryoptsearch.faiss.FaissMemoryOptimizedSearcherFactory;
 
 import java.util.Map;
 import java.util.function.Function;
@@ -122,5 +124,10 @@ public class Faiss extends NativeLibrary {
     @Override
     public boolean supportsRemoteIndexBuild() {
         return true;
+    }
+
+    @Override
+    public VectorSearcherFactory getVectorSearcherFactory() {
+        return new FaissMemoryOptimizedSearcherFactory();
     }
 }
--- a/src/main/java/org/opensearch/knn/memoryoptsearch/VectorSearcher.java
+++ b/src/main/java/org/opensearch/knn/memoryoptsearch/VectorSearcher.java
@@ -1,0 +1,54 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.knn.memoryoptsearch;
+
+import org.apache.lucene.search.KnnCollector;
+import org.apache.lucene.util.Bits;
+
+import java.io.Closeable;
+import java.io.IOException;
+
+/**
+ * This searcher performs vector search on non-Lucene index, for example FAISS index.
+ * Two search APIs will be compatible with Lucene, taking {@link KnnCollector} and {@link Bits}.
+ * In its implementation, it must collect top vectors that is similar to the given query. Make sure to transform the result to similarity
+ * value if internally calculates distance between.
+ */
+public interface VectorSearcher extends Closeable {
+    /**
+     * Return the k nearest neighbor documents as determined by comparison of their vector values for
+     * this field, to the given vector, by the field's similarity function. The score of each document
+     * is derived from the vector similarity in a way that ensures scores are positive and that a
+     * larger score corresponds to a higher ranking.
+     *
+     * <p>The search is allowed to be approximate, meaning the results are not guaranteed to be the
+     * true k closest neighbors. For large values of k (for example when k is close to the total
+     * number of documents), the search may also retrieve fewer than k documents.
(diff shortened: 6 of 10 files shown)


src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissMemoryOptimizedSearcherFactory.java
```diff
@@ -0,0 +1,23 @@
...
+import org.opensearch.knn.memoryoptsearch.MemoryOptimizedSearcher;
+import org.opensearch.knn.memoryoptsearch.MemoryOptimizedSearcherFactory;
+
+import java.io.IOException;
+
+public class FaissMemoryOptimizedSearcherFactory implements MemoryOptimizedSearcherFactory {
+    @Override
+    public MemoryOptimizedSearcher createMemoryOptimizedSearcher(Parameters parameters) throws IOException {
+        final IndexInput indexInput = parameters.getDirectory()
+            .openInput(parameters.getFileName(), new IOContext(IOContext.Context.DEFAULT, null, null, ReadAdvice.RANDOM_PRELOAD));
```
navneet1v, line 20: for all the IOContext parameters can we add java doc on this line why we are setting the specific values.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
```diff
@@ -48,12 +56,73 @@ public class NativeEngines990KnnVectorsReader extends KnnVectorsReader {
...
+                    // Engine file was not found.
+                    continue;
+                }
+
+                // Load memory optimized searcher then put it into the map.
+                final MemoryOptimizedSearcherFactory.Parameters parameters = new MemoryOptimizedSearcherFactory.Parameters(
+                    segmentReadState.directory,
+                    fileName
+                );
+                final MemoryOptimizedSearcher searcher = searcherFactory.get().createMemoryOptimizedSearcher(parameters);
```
shatejas, line 111: This should get a searcher, memory optimized or not should be an implementation detail

src/main/java/org/opensearch/knn/memoryoptsearch/MemoryOptimizedSearcherFactory.java
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * Factory to create {@link MemoryOptimizedSearcher}.
+ * Provided parameters will have {@link Directory} and a file name where implementation can rely on it to open an input stream.
+ */
+public interface MemoryOptimizedSearcherFactory {
+    @AllArgsConstructor
+    @Getter
+    class Parameters {
+        private Directory directory;
+        private String fileName;
```
navneet1v, line 23: lets add @NonNull annotation on these parameters. Also will it be better use @Value and @Builder annotations rather than @AllArgs and @Getter

src/main/java/org/opensearch/knn/memoryoptsearch/MemoryOptimizedSearcher.java
```diff
@@ -0,0 +1,27 @@
...
+ * The main focus of the pressure is limited to JVM, and it would be better if it could take a leverage of OS cache memory to improve
+ * performance internally.
+ * Although its main focus is on memory pressure, but it does not mean that we sacrifice performance over memory consumption.
+ * If anything, this searcher must be balanced between two factors, seeking the best performance while minimizing memory pressure.
+ * This is the goal of this searcher.
+ */
+public interface MemoryOptimizedSearcher extends Closeable {
+    void search(float[] target, KnnCollector knnCollector, Bits acceptDocs) throws IOException;
+
+    void search(byte[] target, KnnCollector knnCollector, Bits acceptDocs) throws IOException;
```
navneet1v, line 26: lets add Java doc on these functions.

src/main/java/org/opensearch/knn/index/engine/KNNLibrary.java
```diff
@@ -140,11 +142,19 @@ default List<String> mmapFileExtensions() {
...
     default boolean supportsRemoteIndexBuild() {
         return false;
     }
+
+    /**
+     * Create a new memory optimized searcher factory.
+     * @return New searcher factory that performs KNN search with optimized memory management.
+     */
+    default Optional<MemoryOptimizedSearcherFactory> getMemoryOptimizedSearcherFactory() {
+        return Optional.empty();
```
navneet1v, line 158: [Suggestion]
I would suggest removing the Optional from the api signature and using the `null` as return type.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
```diff
@@ -48,12 +56,73 @@ public class NativeEngines990KnnVectorsReader extends KnnVectorsReader {
     private Map<String, String> quantizationStateCacheKeyPerField;
     private SegmentReadState segmentReadState;
     private final List<String> cacheKeys;
+    private Map<String, MemoryOptimizedSearcher> memoryOptimizedSearchers;
 
     public NativeEngines990KnnVectorsReader(final SegmentReadState state, final FlatVectorsReader flatVectorsReader) {
         this.flatVectorsReader = flatVectorsReader;
         this.segmentReadState = state;
         this.cacheKeys = getVectorCacheKeysFromSegmentReaderState(state);
+        this.memoryOptimizedSearchers = new HashMap<>(2 * segmentReadState.fieldInfos.size(), 0.6f);
```
navneet1v, line 65: don't we need to close these memoryOptimizedSearchers in the close function of this class?

navneet1v:
@0ctopus13prime please fix the CIs too

src/main/java/org/opensearch/knn/memoryoptsearch/MemoryOptimizedSearcherFactory.java
```diff
@@ -0,0 +1,34 @@
...
+import java.io.IOException;
+
+/**
+ * Factory to create {@link MemoryOptimizedSearcher}.
+ * Provided parameters will have {@link Directory} and a file name where implementation can rely on it to open an input stream.
+ */
+public interface MemoryOptimizedSearcherFactory {
+    @AllArgsConstructor
+    @Getter
+    class Parameters {
```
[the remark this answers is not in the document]
0ctopus13prime: will update in the next revision.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
```diff
@@ -48,12 +56,73 @@ public class NativeEngines990KnnVectorsReader extends KnnVectorsReader {
...
+        }
+    }
+
+    public void loadMemoryOptimizedSearcher() {
+        final List<Closeable> closeableSearchers = new ArrayList<>();
+
+        try {
+            for (FieldInfo fieldInfo : segmentReadState.fieldInfos) {
+                // Skip non-knn fields.
+                if (fieldInfo.attributes().containsKey(KNN_FIELD) == false) {
```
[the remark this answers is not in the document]
0ctopus13prime: will update in the next rev.

src/main/java/org/opensearch/knn/index/engine/KNNEngine.java
```diff
@@ -216,4 +218,9 @@ public ResolvedMethodContext resolveMethod(
     public boolean supportsRemoteIndexBuild() {
         return knnLibrary.supportsRemoteIndexBuild();
     }
+
+    @Override
+    public Optional<MemoryOptimizedSearcherFactory> getMemoryOptimizedSearcherFactory() {
```
shatejas, line 223: I am not really sure we should keep adding implementation details to the enum. Shouldn't a simple engine check inside the factory take care of this?

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
```diff
@@ -48,12 +56,73 @@ public class NativeEngines990KnnVectorsReader extends KnnVectorsReader {
...
+                }
+
+                // Load memory optimized searcher then put it into the map.
+                final MemoryOptimizedSearcherFactory.Parameters parameters = new MemoryOptimizedSearcherFactory.Parameters(
+                    segmentReadState.directory,
+                    fileName
+                );
+                final MemoryOptimizedSearcher searcher = searcherFactory.get().createMemoryOptimizedSearcher(parameters);
+                if (searcher != null) {
+                    closeableSearchers.add(searcher);
```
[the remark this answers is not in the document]
0ctopus13prime: yeah, we can do that.

src/main/java/org/opensearch/knn/memoryoptsearch/MemoryOptimizedSearcher.java
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * Memory optimized searcher that performs vector search with the best efforts to minimize memory pressure.
+ * Its implementation should only focus on optimizing memory allocations, and must not increase memory pressure in JVM.
+ * The main focus of the pressure is limited to JVM, and it would be better if it could take a leverage of OS cache memory to improve
+ * performance internally.
+ * Although its main focus is on memory pressure, but it does not mean that we sacrifice performance over memory consumption.
+ * If anything, this searcher must be balanced between two factors, seeking the best performance while minimizing memory pressure.
+ * This is the goal of this searcher.
+ */
+public interface MemoryOptimizedSearcher extends Closeable {
```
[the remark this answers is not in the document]
0ctopus13prime: Quite a bit long history:

1. This project begins as a partial loading.
2. FAISS is the first target, and the project goal is to have FAISS engine to rely on IndexInput to load vector during search.
3. The purpose of doing this is to alleviate memory pressure, as FAISS loads every bytes into memory.
4. 'MemoryOptimized' is coming from there. It is not covered in this PR, but we will have `memory_optimized_search` bool flag at index setting.
5. Another reason to have it as a prefix is to make it general, hiding 'Lucene'.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
0ctopus13prime: Why it should be hided in details?
The broader theme of this change focuses on 'MemoryOptimizedSearch', implicitly implying that FAISS won't load every bytes into memory, instead it will load bytes it needs during search.
I think this should be coming upfront.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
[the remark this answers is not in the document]
0ctopus13prime: will update in the next rev.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
```diff
@@ -48,12 +56,73 @@ public class NativeEngines990KnnVectorsReader extends KnnVectorsReader {
...
+                    memoryOptimizedSearchers.put(fieldInfo.getName(), searcher);
+                }
+            }
+        } catch (Exception e) {
+            // Close opened searchers first, then suppress
+            try {
+                IOUtils.closeWhileHandlingException(closeableSearchers);
+            } catch (Exception closeException) {
+                e.addSuppressed(closeException);
```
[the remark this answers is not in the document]
0ctopus13prime: yeah, will update in the next rev.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
```diff
@@ -135,6 +204,14 @@ public void search(String field, float[] target, KnnCollector knnCollector, Bits
...
             return;
         }
+
+        // Try with memory optimized searcher
+        final MemoryOptimizedSearcher memoryOptimizedSearcher = memoryOptimizedSearchers.get(field);
+        if (memoryOptimizedSearcher != null) {
+            memoryOptimizedSearcher.search(target, knnCollector, acceptDocs);
+            return;
+        }
+
```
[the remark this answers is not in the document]
0ctopus13prime: So, `swalloing null searcher` VS `explicitly throwing an exception`.

I think we can just use `memoryOptimizedSearcher` without nullity check.
Because, since the non-null memoryOptimizedSearcherFactory indicates that it supports the mode. 

Can I instead enforce the factory to return non-null searcher?

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissMemoryOptimizedSearcherFactory.java
0ctopus13prime: sure, will do!

src/main/java/org/opensearch/knn/memoryoptsearch/MemoryOptimizedSearcherFactory.java
0ctopus13prime: will update in the next rev.

src/main/java/org/opensearch/knn/memoryoptsearch/MemoryOptimizedSearcher.java
0ctopus13prime: will update in the next rev.

src/main/java/org/opensearch/knn/index/engine/KNNLibrary.java
0ctopus13prime: sure, will update in the next rev.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
0ctopus13prime: oops, did not mentioned in this PR, but that will be covered in Part-7 (Complete Faiss Loading Part at Codec Level)
There's topic to discuss MemoryOptimizedSearcher at codec level, `close` implementation will depending on the result of the discussion.

src/main/java/org/opensearch/knn/index/engine/KNNEngine.java
0ctopus13prime: I think each Engine either supports or does not support `memory_optimized_search` mode. 
So far, FAISS is the single engine that supports the mode but it can be expanded to other future engines.
From that perspective, it is natural to put `factory-getter` method in Engine.

What is your concern?

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissMemoryOptimizedSearcherFactory.java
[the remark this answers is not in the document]
0ctopus13prime: oh, yes you're right. Just typed that one from recommendations in my IDE.
Will update in the next rev.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
```diff
@@ -48,12 +56,73 @@ public class NativeEngines990KnnVectorsReader extends KnnVectorsReader {
...
+                    // It's not supported
+                    continue;
+                }
+
+                // Start creating searcher
+                final String fileName = KNNCodecUtil.getNativeEngineFileFromFieldInfo(fieldInfo, segmentReadState.segmentInfo);
+                if (fileName == null) {
+                    // Engine file was not found.
+                    continue;
+                }
```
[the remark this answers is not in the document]
0ctopus13prime: Will make the validation method returning `fileName: String`. 
if it returns null fileName:String, then engine file does not exist and skip to the next one.

navneet1v:
> > @0ctopus13prime please fix the CIs too
> 
> Hmm.. this PR should not affect anything at all 😅 Let me check, suspecting a flaky test

I also think so, but its always better to see build CIs passing and also DCO checks. This will ensure that branch is healthy

src/main/java/org/opensearch/knn/memoryoptsearch/MemoryOptimizedSearcher.java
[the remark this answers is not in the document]
navneet1v: +1 on this.

src/main/java/org/opensearch/knn/index/engine/KNNEngine.java
jmazanec15: It will be per engine. Ideally, it would be included in something like https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/engine/KNNLibrarySearchContext.java. This can be retrieved via https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/engine/KNNLibrary.java#L121.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
```diff
@@ -37,23 +42,89 @@
...
+        if (attributes == null || attributes.containsKey(KNN_FIELD) == false) {
+            return null;
+        }
+
+        // Get engine
+        final String engineName = attributes.getOrDefault(KNN_ENGINE, KNNEngine.DEFAULT.getName());
+        final KNNEngine knnEngine = KNNEngine.getEngine(engineName);
+
+        // Get memory optimized searcher from engine
+        final MemoryOptimizedSearcherFactory searcherFactory = knnEngine.getMemoryOptimizedSearcherFactory();
```
0ctopus13prime, line 94: Hi @shatejas 
Hope to keep `MemoryOptimized` as a theme here if you're ok with it. Also I want to force each engine to return a searcher that won't consume much memory needed. If it is not possible, then engine can return null instance.

Please share your thoughts.

0ctopus13prime:
Please note that once get verbal approval on the implementation, then will ship unit tests in this PR.
Thanks.

src/main/java/org/opensearch/knn/memoryoptsearch/MemoryOptimizedSearcher.java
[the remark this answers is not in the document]
shatejas: I have the context - The idea here is we should eventually tie the doANNSearch logic from KNNWeight in the reader - and then the factory should decide whether to pick Perf optimized or memory optimized. That will clean up the logic on KNNWeight side of things making it consistent and all it has to do call the `reader.search` (similar to lucene). And reader decides the right implementation to pick -this standardizes the interface eliminating the if else complications opening up a door to not have two separate queries like native engine and lucene queries


Having a MemoryOptimized prefix limits it to partial loading implementation. The method signatures  in no way requires any memory optimized specific params and we should keep it that way

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
shatejas: Its basic [Liskov Substitution principle](https://en.wikipedia.org/wiki/Liskov_substitution_principle) couple with [open-closed principle](https://en.wikipedia.org/wiki/Open%E2%80%93closed_principle). your reader should be able to take any implementation of searcher and run it - the behavior doesn't change but the implementation can change. couple that with a factory and its always open to an extention for various implementations but closed to any sort of modification

src/main/java/org/opensearch/knn/index/engine/KNNEngine.java
shatejas: My problem with these enums is that its turning into giant static configuration database, along with a workaround to have dependency injection. Its a maintenance concern- I haven't seen such a coding pattern and I am not sure its the right way to go

src/main/java/org/opensearch/knn/memoryoptsearch/MemoryOptimizedSearcher.java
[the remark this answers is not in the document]
0ctopus13prime: @shatejas 
I think we can still make it happen while keep the prefix.
As you said, the factory will eventually decide to which searcher to pick, and when `memory_optimized_search` is on, it can return `MemoryOptimizedSearcher`. If it needs to prioritize performance, then probably it will return `PerformanceOptimizedSearcher`. 

Don't get me wrong, I totally agree with your long term view, and it is clear the benefit we could get from cleaning up the logic from KNNWeight.

We can take an iterative approach here, make it first return `MemoryOptimizedSearcher`, then when the time comes to clean up logic from KNNWeight, we can change the signature to return a general searcher while hiding the implementation behind.

The reason is, so far we don't have a clear plan for that, but the logic I'm trying to add is something like below:

```
if (memory optimized search is on) {
  Searcher searcher = factory.get();

  versus

  MemoryOptimizedSearcher searcher = factory.get();
}
```

The other reason is that we might end up taking a entirely different direction. For example, keep the logic in KNNWeigh and do some magic somewhere. 

While refactoring cost is not expensive, just merely changing the signature and hide it behind, can we keep this as it is first then refactor later if you're fine with it?

Thank you.

src/main/java/org/opensearch/knn/index/engine/KNNEngine.java
jmazanec15: > Shouldn't a simple engine check inside the factory take care of this?

I dont think we should branch throughout the code based on engine. This will make engine extendability very difficult and also will created complex branching in the code - this has happened to some degree based in the query.

The purpose of KNNLibrarySearchContext was to let the given engine tell how we are supposed to search. This isnt totally complete, but the direction to kind of further improve this is in #2568. On the indexing/reader setup side, we would need the engines to return a PerFieldKnnVectorFormat that can be added via https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/codec/KNNCodecService.java#L43.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
0ctopus13prime: I think this design is not the type system. It's 'has-a' relationship between Reader and the searcher, I don't think the principle can be applied to exactly.

> your reader should be able to take any implementation of searcher and run it..
I don' think it applies to the change I want to add it here. 
Reader should fallback to memory optimized searcher which will be enabled ONLY when `memory_optimized_search` is enabled.

What we want to do is, when `memory_optimized_search` is enabled, then corresponding searcher that specializes in using less memory to achieve vector search should be used as a fallback for native engine -- Faiss for example.

Do we have a concrete plan to move searcher factor in VectorReader?

src/main/java/org/opensearch/knn/memoryoptsearch/MemoryOptimizedSearcher.java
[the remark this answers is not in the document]
0ctopus13prime: Discussed in offline, will change it to have a general name but keep the prefix at FAISS level.
In the long term, we can move search logic from KNNWeight to the searcher, then make searcher factory return it.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
```diff
@@ -37,23 +42,89 @@
...
+import java.util.Objects;
+
+import static org.opensearch.knn.common.KNNConstants.KNN_ENGINE;
+import static org.opensearch.knn.index.mapper.KNNVectorFieldMapper.KNN_FIELD;
 
 /**
  * Vectors reader class for reading the flat vectors for native engines. The class provides methods for iterating
  * over the vectors and retrieving their values.
  */
+@Slf4j
```
shatejas, line 54: Why not log4j?

shatejas (approved):

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
0ctopus13prime: Slf4j is a general logging framework without tied dependency over a specific logger framework like Log4j, Logback etc. It's a facade for different logging frameworks. When there's a log framework upgrade or change, it won't need any changes

navneet1v: let just use what we are using in the plugin to avoid conflicts for future.

0ctopus13prime: why it conflicts? opensearch core is using slf4j with log4j

0ctopus13prime: sounds good, will update in the next rev.
before raising a new PR, could you share your thoughts on the code?
If you leave comments, I will factor them into the next PR.

src/main/java/org/opensearch/knn/index/engine/KNNEngine.java
navneet1v: @0ctopus13prime merging the thread from here https://github.com/opensearch-project/k-NN/pull/2581#discussion_r1982393038 too. and also adding my thoughts from the discussion happening on this thread.

@jmazanec15, the idea of engine providing a PerFieldKnnVectorFormat based on the MappedfieldType is a step in right direction. But once we have the KNNVectorFormat which is engine specific then searcher should not come from KNNEngine. It should be job of Reader to find the right searcher for the algorithm.

We moving searcher currently in KNNEngine, I somehow feel a decision we are taking too early. Should we take this decision once we move to PerFieldKnnVectorFormat being returned via KNNEngine?

0ctopus13prime: If we could align on having a dedicated FAISS VectorFormat, then will update accordingly in the next rev.
I will add a branch in `BasePerFieldKnnVectorsFormat` to return FaissVectorFormat returning a reader where it performs vector search on FAISS index for knn fields using FAISS engine.

```
// In BasePerFieldKnnVectorsFormat
if (engine == FAISS) {
  return new FaissKnnVectorsFormat(...);
}
```

jmazanec15: My concern is that the engine/method will determine if the optimized memory searcher is supported and how its supported. In the current code, Im not sure Im seeing where the branching based on method logic is implemented - for instance, where would we say HNSW fp16 is not supported or IVF is not supported (I might just be missing it). But I think this can get fairly complex, and is why I think that it would be better to go through the engine that determines what functionality is supported.

So we could just add a method in KNNLibrarySearchContext like "getMemoryOptimizedSearcherFactory()". Then, we can retrieve it via `knnEngine.getLibrarySearchContext(methodName)` (we might need to take additional info like library params like what @owenhalpert is working on). With this, we keep the codec somewhat engine agnostic.

Vikasht34: > My concern is that the engine/method will determine if the optimized memory searcher is supported and how its supported. In the current code, Im not sure Im seeing where the branching based on method logic is implemented - for instance, where would we say HNSW fp16 is not supported or IVF is not supported (I might just be missing it). But I think this can get fairly complex, and is why I think that it would be better to go through the engine that determines what functionality is supported.
> 
> So we could just add a method in KNNLibrarySearchContext like "getMemoryOptimizedSearcherFactory()". Then, we can retrieve it via `knnEngine.getLibrarySearchContext(methodName)` (we might need to take additional info like library params like what @owenhalpert is working on). With this, we keep the codec somewhat engine agnostic.

This seems reasonable to me !!

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
Vikasht34: I think we should start using Slf4j and clean code by replacing log4j to Slf4j !! There are couple of benefits we have 1. We are aligning to core and Lucene , Lucene also uses Slf4j , and in future either if core replaces that with any other inline framework !! We get for free !!

navneet1v: in that case it should a be part of a separate GH and not scope of this PR.

0ctopus13prime:
@jmazanec15 
For those index types that don't support memory optimized searching, FaissIndex.load (which will be added in the next PR) will throw an exception. In vector reader, it catches it then moving on to the next field.
And in the index settings update event listener, it will make sure to allow it to be turned on ONLY IF all KNN fields have FAISS HNSW. Also the flag check in KnnQueryBuilder will make sure other index types but HNSW will be fallbacked to NativeQuery which will use native shared library to continue the search.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
0ctopus13prime: Will update the annotation in the next rev! :)

jmazanec15 (approved):

PR #2583, opened 2025-03-05
Refactor derived source for nested field types

jmazanec15:
### Description
PR refactors derived source feature to be more robust and maintainable. In particular, it is intended to improve nested field support.

The following changes are introduced in this PR
1. Enablement of derived source is moved from a fieldInfo attribute to a segment info attribute. The reasoning for this is that segmentinfo attributes allow us to use the mapper service to impose conditions based on the overall index mapping on whether or not to enable the feature for a particular field.
2. Attaching the nested field lineage for each field we are injecting as a segment info attribute. This allows us to avoid querying and simplifies handling of nested and object field overlap.
3. Move out all low level lucene utilities into a helper class. This will make the code more readable and testable.
4. Simplification of the nested field iterator. This makes code easier to read
5. More testing was added. A new iT as well as several uTs. The code is restructured so that it will be easier to test via uTs
6. Modifications to insertion into nested doc algorithm (see below)

### Insertion into nested algorithm
The assumption is that each mapped document will have at least one field value in it. With this, the following strategy will be followed to match a given map in an array to a document ID:

1. For each field info, check if the document contains the field. There must be atleast one, or else the empty map wouldve been parsed out
2. Starting at the given offset, find the next document id that contains the same parent as the field being injected. This must be the doc ID that maps to this source

### Related Issues

### Check List


src/main/java/org/opensearch/knn/index/codec/derivedsource/DerivedSourceLuceneHelper.java
```diff
@@ -0,0 +1,238 @@
...
+    /**
+     * Get the lowest docId for a field that is greater than (or equal to) the offset. This method is implemented in a
+     * very similar way as checking if a field exists.
+     *
+     * @param fieldToMatch field to find the lowest docId for
+     * @param offset       offset to start searching from (inclusive)
+     * @return lowest docId for the field that is greater than the offset. Returns {@link DocIdSetIterator#NO_MORE_DOCS} if doc cannot be found
+     * @throws IOException if there is an issue reading from the formats
+     */
+    public int getFirstDocWhereFieldExists(String fieldToMatch, int offset) throws IOException {
```
msfroh, line 113: Are we going to want to use this multiple times in a single request? Would it make sense to return the DocIdSetIterator instead?

jmazanec15: We will make this call multiple times, but for the field will not be consistent. But we could reuse if we map it to a particular field.

navneet1v:
> 1. Enablement of derived source is moved from a fieldInfo attribute to a segment info attribute. The reasoning for this is that segmentinfo attributes allow us to use the mapper service to impose conditions based on the overall index mapping on whether or not to enable the feature for a particular field.

@jmazanec15 will this cause any BWC issue? Can we have a BWC test for this.

0ctopus13prime:
# Goal

This paper presents an approach to preserving the structure in `_source` for nested cases with an empty array. A Depth First Search (DFS) algorithm is then applied to inject nested vector values when reading the stored `_source` field.



# _source structure for nested case

## User Input

```
{
  "f1": [
     {
       "f2": [
         {
           "vector_field": [...]
         }
       ],
       "vector_field": [...]
     } 
  ],
  "vector_field": [...]
}
```


From the above, OpenSearch internally maps each field using the following full path name:

```
f1.f2.vector_field
f1.vector_field
vector_field
```



## Lucene Documents

With the given input, the expected Lucene document representation is shown below.
Note that Document-2 contains the `_source` field since it is the top-level document.

```
Document-0 : {
  "f1.f2.vector_field": [...]
}

Document-1 : {
  "f1.vector_field": [...]
}

Document-2 : {
  "vector_field": [...],
  "_source": """
  {
    "f1": [
        {
        "f2": [
            {
            "vector_field": [...]
            }
        ],
        "vector_field": [...]
        } 
    ],
    "vector_field": [...]
    }
  """
}
```




## Ideas

Instead of storing the full vector string in `_source`, we can save an empty vector string `[]`. During reading, we can intercept the request and replace the empty vector string with its actual representative.
As a result, the `_source` of Document-2 would have an empty vector string, as shown below:

```
Document-2 : {
  "vector_field": [...],
  "_source": """
  {
    "f1": [
        {
        "f2": [
            {
            "vector_field": []
            }
        ],
        "vector_field": []
        } 
    ],
    "vector_field": []
    }
  """
}
```



## Running DFS algorithm in recovering.

In the `NestedPerFieldDerivedVectorInjector.inject` method [[Link](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/codec/derivedsource/NestedPerFieldDerivedVectorInjector.java#L39)], the parent document ID is already known. From the example above, the given parent ID would be 2 (e.g., Document-2). Additionally, by the time the method is called, the full path of the nested vector field is also known. The field name could be one of the following:

1. f1.f2.vector_field
2. f1.vector_field
3. vector_field

This allows us to prepare a document ID iterator before the method is called. e.g. We know the fullpath child field name, we can get PostingsEnum (which is a subtype of DocIdIterator) with it.

The idea is to run a Depth First Search (DFS) algorithm to traverse all fields. Whenever the traversal reaches the target nested vector field, the iterator advances to retrieve the internal Lucene document ID, replacing the empty vector array string with the actual vector data.


```

// Get the field name
public NestedPerFieldDerivedVectorInjector(FieldInfo childFieldInfo, ...) {
    this.nestedVectorFieldName = ...;  // Ex: f1.f2.vector_field
    this.childDocIdIterator = ...;  // We can get it from PostingsEnum
    this.dfsDepth = ...;  // Ex: it would be 2 (0 is the base) for f1.f2.vector_field
    
    // initializing other parts.
    ...
}

// Run DFS altorithm then replace empty array string
@Override
public void inject(int parentDocId, Map<String, Object> sourceAsMap) throws IOException {
    runDfsAndReplace(sourceAsMap, "", 0);
}

private void runDfsAndReplace(Map<String, Object> map, String prefix, int depth) {
    if (depth > dfsDepth) {
        // We can stop here.
        // Ex: given the field name was f1.f2.vector_field
        //     then, we won't need to go deeper than depth 2 (ex: f1.f2.f3.f4...)
        return;
    }

   Set<String> keys = get keys from `map`.
   for (String key : keys) {
       Object value = map.get(key);
       // e.g. vector_field or f1.f2.vector_field
       String fullPathKey = prefix.isEmpty() ? key : prefix + "." + key;
        
       if (value instanceof List) {
           // It should be List[Map]
           // Ex: "key": [
           //   {...}, {...}
           // ]
           List listValue = (List) list;
           for (Object subMap : listValue) {
               runDfsAndReplace(subMap, fullPathKey, depth + 1);
           }
           continue;
       }
        
       if (value instanceof Map) {
           runDfsAndReplace((Map) value, fullPathKey, depth + 1);
           continue;
       }
        
       if (depth == dfsDepth && fullPathKey.equals(this.nestedVectorFieldName)) {
           // Found it!
           // Note : we must make sure it's a mutable map first,
           //        but will skip it to make the example short.
           
           // Advance to the next child document 
           int childDocId = iterator.nextDoc();
           map.put(key, getVectorAndToString(childDocId));
       }
   }
}

```

# 

## Simulation

Input :

```
{
  "f1": [
     {
       "f2": [
         {
           "vector_field": [...]
         },
         {
           "vector_field": [...]
         }
       ],
       "vector_field": [...],
     },
     {
       "f2": [
         {
           "vector_field": [...]
         },
         {
           "vector_field": [...]
         }
       ],
       "vector_field": [...],
     }
  ],
  "vector_field": [...]
}

```


Lucene Documents :

```
Document-0 : {
  "f1.f2.vector_field": [...]
}

Document-1 : {
  "f1.f2.vector_field": [...]
}

Document-2 : {
  "f1.vector_field": [...]
}

Document-3 : {
  "f1.f2.vector_field": [...]
}

Document-4 : {
  "f1.f2.vector_field": [...]
}

Document-5 : {
  "f1.vector_field": [...]
}

Document-6 : {
  "vector_field": [...],
  "_source": """
{
  "f1": [
     {
       "f2": [
         {
           "vector_field": [] <----- we only store an empty array
         },
         {
           "vector_field": []
         }
       ],
       "vector_field": [],
     },
     {
       "f2": [
         {
           "vector_field": []
         },
         {
           "vector_field": []
         }
       ],
       "vector_field": [],
     }
  ],
  "vector_field": []
}
  """
}

```


NestedPerFieldDerivedVectorInjector with `f1.f2.vector_field`’s Visit order

1. f1.f2.vector_field
    1. We found the field! After calling `nextDoc`, we would get 0.
2. f1.f2.vector_field
    1. We found the field! After calling `nextDoc`, we would get 1.
3. f1.vector_field
4. f1.f2.vector_field
    1. We found the field! After calling `nextDoc`, we would get 3.
5. f1.f2.vector_field
    1. We found the field! After calling `nextDoc`, we would get 4.
6. f1.vector_field
7. vector_field



Transformation Result

```
{
  "f1": [
     {
       "f2": [
         {
           "vector_field": [...]
         },
         {
           "vector_field": [...]
         }
       ],
       "vector_field": [...],
     },
     {
       "f2": [
         {
           "vector_field": [...]
         },
         {
           "vector_field": [...]
         }
       ],
       "vector_field": [...],
     }
  ],
  "vector_field": [...]
}

```



## Pros / Cons

Pros

1. The algorithm can handle any nested structure.
2. It is relatively simple to implement.


Cons

1. We rely on a specific internal data type hidden behind `Map`. 
    Currently, we assume the value type is one of the following: `Map`, `List`, or any scalar data type (e.g., `String`, `Float`, `Boolean`, etc.). However, in the future, the core team may introduce a **List-like** data type as a value. If that happens, we would need to update the logic to properly navigate and process its contents.
2. Ordering issue.
    This algorithm assumes that the key ordering in the map matches the order in the original JSON.
    For example, given the JSON below, I expect the algorithm to visit the first object with the name `"name-1"`, followed by the second object with the name `"name-2"`. However, since `List` does not guarantee order, it may return the first object as `"name-2"` instead of `"name-1"`, leading to inconsistencies.

```
"field": [
  {
    "name": "name-1",
    "vector_field": [...]
  },
  {
    "name": "name-2",
    "vector_field": [...]
  }
]
```

navneet1v:
@0ctopus13prime this is an interesting approach and I won't lie that I haven't think about it. For me another concern comes with multiple nested fields is getting the vector from .vec file. Since the nested docs ordering and ingestion is controlled by OpenSearch and that logic is pretty fragile IMO.

jmazanec15:
@0ctopus13prime Thanks - I think this is a better approach. Im trying it out and got a version to pass the existing iTs. 

>  Currently, we assume the value type is one of the following: Map, List, or any scalar data type (e.g., String, Float, Boolean, etc.). However, in the future, the core team may introduce a List-like data type as a value. If that happens, we would need to update the logic to properly navigate and process its contents.

Not sure I understand this point, but I think will also be there for filter approach. Also, looks like what is done in XContentMapValues. I think we can put code in core there for stability.

> However, since List does not guarantee order, it may return the first object as "name-2" instead of "name-1", leading to inconsistencies.

From my understanding, lists do guarantee ordering. I dont think this is a problem.

> For me another concern comes with multiple nested fields is getting the vector from .vec file. Since the nested docs ordering and ingestion is controlled by OpenSearch and that logic is pretty fragile IMO.

Ordering is determined here: https://github.com/opensearch-project/OpenSearch/blob/127501789334d6deb19d206bf76d8475a9e27c54/server/src/main/java/org/opensearch/index/mapper/ParseContext.java#L530-L555. It is something we'd have to rely on and is a con to the overall codec approach. However, it would be same thing for the filtered approach.

jmazanec15:
Going to move towards masking approach. Going to close PR and open up new ones.

PR #1135, opened 2023-09-14
Upgrade bytebuddy version to match OpenSearch core

ryanbogan:
### Description
Upgrades bytebuddy version to match OpenSearch core
 
### Check List
- [X] New functionality includes testing.
  - [X] All tests pass
- [X] New functionality has been documented.
  - [X] New functionality has javadoc added
- [X] Commits are signed as per the DCO using --signoff 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/k-NN/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #2585, opened 2025-03-06
Add filter function to KNNQueryBuilder

chloewqg:
### Description
Add filter function to KNNQueryBuilder with unit tests and integration tests

### Related Issues
Resolves #2585
Related: [#282](https://github.com/opensearch-project/neural-search/issues/282), [#1135](https://github.com/opensearch-project/neural-search/issues/1135)

### Check List
- [x] Public documentation issue/PR [created](https://github.com/opensearch-project/neural-search/issues/282).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -314,7 +314,10 @@ dependencies {
     // Excluding slf4j here since json-path is only used for testing, and logging failures in this context are acceptable.
     testFixturesImplementation('com.jayway.jsonpath:json-path:2.9.0') {
         exclude group: 'org.slf4j', module: 'slf4j-api'
+        exclude group: 'net.minidev', module: 'json-smart'
     }
+    // Explicitly include a safe version of json-smart for test fixtures
+    testFixturesImplementation group: 'net.minidev', name: 'json-smart', version: "${versions.json_smart}"
     testFixturesImplementation "org.opensearch:common-utils:${version}"
     implementation 'com.github.oshi:oshi-core:6.4.13'
     api "net.java.dev.jna:jna:5.13.0"


src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
```diff
@@ -288,6 +288,23 @@ private void validate() {
         }
     }
 
+    /**
+     * Add a filter to Neural Query Builder
+     * @param filterToBeAdded filter to be added
+     * @return return itself with underlying filter combined with passed in filter
+     */
+    public QueryBuilder filter(QueryBuilder filterToBeAdded) {
```
shatejas, line 296: There is a `Builder` created to be able to set the filter - why is this needed?

chloewqg: This is for the issue here https://github.com/opensearch-project/neural-search/issues/1135 where we want to add common filter from HybridQueryBuilder to all queries below, including KNNQueryBuilder. So it means adding additional filter to previous filter.

shatejas: Modifying an existing query builder is not a great idea, the builder is created to make sure its immutable in a way and every parameter is validated together rather than individually - you need to create a copy of it in neural side to support the use case

```
KNNQueryBuilder initialqueryBuidler = //whatever it is

KNNQueryBuilder filterQueryBuilder = KNNQueryBuilder.builder().filter(filterTobeAdded).k(initialQuery.k()) and so on before calling .build()
```

heemin32: @shatejas, This is newly added method from OpenSearch core in QueryBuilder interface.
https://github.com/opensearch-project/OpenSearch/pull/17409/files#diff-cc69a4975dd7181beb8d1124f87bc6e3714618b2fa218d307be8a2e0db4b9651


@chloewqg We should add `@Override` annotation.

shatejas: In that case, we still need to go through the builder. We need to create a new instance and return to make sure all current and future validations are applied

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
```diff
@@ -381,6 +381,24 @@ public void doXContent(XContentBuilder builder, Params params) throws IOExceptio
...
+        if (validateFilterParams(filterToBeAdded) == false) {
+            return this;
+        }
+        if (filter == null) {
+            filter = filterToBeAdded;
+            return this;
+        }
+        filter = filter.filter(filterToBeAdded);
+        return this;
```
shatejas, line 399: Please refer https://github.com/opensearch-project/k-NN/pull/2585#discussion_r1987634062 and go through the builder here. Do not manipulate the existing filter, it increases the chances of side effects and bugs

```
return KNNQueryBuilder.builder().filter(newFilter)....build()//copy everything from existing
```

chloewqg: Updated in new commit

shatejas (approved):

CHANGELOG.md
```diff
@@ -6,6 +6,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased 3.0](https://github.com/opensearch-project/k-NN/compare/2.x...HEAD)
 ### Features
+* [Remote Vector Index Build] Introduce Remote Native Index Build feature flag, settings, and initial skeleton [#2525](https://github.com/opensearch-project/k-NN/pull/2525)
```
heemin32, line 9: Please fix this changelog file

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
```diff
@@ -381,6 +381,35 @@ public void doXContent(XContentBuilder builder, Params params) throws IOExceptio
...
+     * @return return itself with underlying filter combined with passed in filter
+     */
+    @Override
+    public QueryBuilder filter(QueryBuilder filterToBeAdded) {
+        if (validateFilterParams(filterToBeAdded) == false) {
+            return this;
+        }
+
+        if (this.filter != null) {
+            filterToBeAdded = filter.filter(filterToBeAdded);
```
shatejas, line 396: I looked at this again, and this looks suspicious. Do not manipulate the parameter passed in the function - it will have unknown side effects if its used after this function is called. create a local variable and pass it to the builder

heemin32: I agree that using a separate variable is preferable.

By the way this code won't cause any side effects since it only changes the reference of the local variable, leaving the passed variable outside the method unaffected.

PR #253, opened 2022-01-08
Changing serialization for knn vector from single array object to collection of floats

martin-gaievski:
### Description
Changing logic of k-NN vector serialization/deserialization in order to decrease memory taken by the index. Existing serialization stores vector as Java array, which has 27 bytes of overhead for each individual array (as per [stream protocol](https://docs.oracle.com/javase/7/docs/platform/serialization/spec/protocol.html)). In newly implemented logic vector is serialized as a collection of individual numbers. During deserialization we can restore this collection to an array assuming it length from number of bytes in byte stream. 

Logic has been added in a backward compatibility manner for deserialization. Initially we read first 27 bytes of the byte stream and identify if this is a serialized array based on the same [stream protocol](https://docs.oracle.com/javase/7/docs/platform/serialization/spec/protocol.html) grammar. Depending on result of the check we apply old logic (deserialize as a single array object) or a new logic (deserialize as collection of individual floats). 

Performance stayed roughly the same time-wise but we see an improvement in memory usage. Below are results of 2 tests comparing existing and new approaches. For testing we have used the [benchmark tool](https://github.com/opensearch-project/k-NN/tree/main/benchmarks/perf-tool) from main k-NN branch with two data sets Fashion-MNIST and SIFT from [here](https://github.com/erikbern/ann-benchmarks).

Diff (changed - original) for SIFT dataset:
```
{
  "test_took": -37602.98385410011,
  "train_model_took_total": -717.2553265995593,
  "create_index_took_total": 58.110593299596985,
  "ingest_took_total": 248.3000000000029,
  "ingest_took_p50": -0.5999999999999943,
  "ingest_took_p90": -1.799999999999983,
  "ingest_took_p99": -0.6000000000000227,
  "refresh_index_store_kb_total": -14444.896679687488,
  "refresh_index_took_total": -152.2593566997159,
  "force_merge_took_total": -119.27976410053702,
  "query_took_total": -36920.59999999998,
  "query_took_p50": -0.6999999999999993,
  "query_took_p90": -1.0999999999999979,
  "query_took_p99": -2.6999999999999993,
  "query_memory_kb_total": -0.8000000000029104,
  "query_recall@K_total": 0.004822199999999999,
  "query_recall@1_total": -0.00017999999999984695
}
```

Diff (changed - original) for Fashion-MNIST dataset:
```
{
  "test_took": -1796.963270999724,
  "create_index_took_total": -195.58473960000993,
  "ingest_took_total": 158.40000000000146,
[...]

PR #2587, opened 2025-03-06
Removes Array based vector serialization

jmazanec15:
### Description
Previously, we had serialized vectors via an array method. This was inefficient and deprecated and switched in #253 and released in 1.3.0. With that, we no longer serialized new segment data via the array based serializer. Now that it is 3.0, because the oldest index that can be upgraded is from 1.3.0 (https://opensearch.org/docs/latest/install-and-configure/upgrade-opensearch/index/#compatibility), we no longer need to handle array based serialization. So, we can remove all of it.

### Related Issues

### Check List


martin-gaievski (approved):
Looks good to me, thanks Jack

naveentatikonda (approved):

PR #2590, opened 2025-03-10
Added IxMp section loading logic from FAISS index.

0ctopus13prime:
### Description
[2/11] This PR adds a loading logic of FaissIdMapIndex section in FAISS section.
This section contains a mapping table to maps an internal vector id to external Lucene document id.
There are two cases
1. Dense case : All documents have at least one KNN field, which has exactly one vector. Therefore, internal vector id is corresponding to Lucene document id.
2. Sparse case : Not all documents having vectors. Or parent-child case. In which case, FaissIdMapIndex will have a mapping to maps an internal vector id to its parent Lucene document id.

### Related Issues

### Check List


0ctopus13prime:
Will continuously add unit tests to this PR.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissIdMapIndex.java
```diff
@@ -0,0 +1,244 @@
...
+     * Partially load id mapping and its nested index to which vector searching will be delegated.
+     *
+     * @param input An input stream for a FAISS HNSW graph file, allowing access to the neighbor list and vector locations.
+     * @return {@link FaissIdMapIndex} instance consists of index hierarchy.
+     * @throws IOException
+     */
+    public static FaissIdMapIndex load(IndexInput input) throws IOException {
+        FaissIdMapIndex faissIdMapIndex = new FaissIdMapIndex();
+        readCommonHeader(input, faissIdMapIndex);
+        final FaissIndex nestedIndex = FaissIndex.load(input);
```
shatejas, line 49: Is recursion still needed for future cases? At least the way the code is currently setup, adding two extra lines here would void recursion which is a good tradeoff. Recursion is often hard to debug if there is a stopping condition missed so want to see if we can avoid it

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissIdMapIndex.java
```diff
@@ -0,0 +1,244 @@
...
+    }
+
+    /**
+     * Partially load id mapping and its nested index to which vector searching will be delegated.
+     *
+     * @param input An input stream for a FAISS HNSW graph file, allowing access to the neighbor list and vector locations.
+     * @return {@link FaissIdMapIndex} instance consists of index hierarchy.
+     * @throws IOException
+     */
+    public static FaissIdMapIndex load(IndexInput input) throws IOException {
```
shatejas, line 46: Was wondering if load can be abstract? this will make faissIndex abstract as well - we want to enforce load on each type of index, making it static doesn't seem to do it. Let me know your thoughts on it

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
```diff
@@ -51,7 +51,7 @@
  * Vectors reader class for reading the flat vectors for native engines. The class provides methods for iterating
  * over the vectors and retrieving their values.
  */
-@Slf4j
+@Log4j2
```
shatejas, line 54: Thanks for this

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissIdMapIndex.java
0ctopus13prime: sure, we can do it.

```
String indexType = ...  // Get index type first
FaissIndex index = supplier.get(indexType);  // Create an instance based on the acquired indexType.
index.load(input);
return index;
```

How does it look?

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissIdMapIndex.java
0ctopus13prime: yes, I think so. Conceptually, it is a nested index, index1 can have index2, in turn, index2 can have another index as a nested index type.

For example, IxMp will require IHNf index which needs IxF2, flat float vector index.
After loaded IHNf index, IxMp will load a mapping table.

```
-- IxMp
   ----- IHNf
       ---- IxF2
       ---- IxF2
   ----- IHNf 
-- IxMp loads a mapping table.  
```

I think this is how sections in FAISS index are organized so that there will be a top level index at the top.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissIdMapIndex.java
shatejas: On surface it looks good

0ctopus13prime: Will update in the next rev.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissIdMapIndex.java
```diff
@@ -0,0 +1,242 @@
...
+    /**
+     * Partially load id mapping and its nested index to which vector searching will be delegated.
+     *
+     * @param input An input stream for a FAISS HNSW graph file, allowing access to the neighbor list and vector locations.
+     * @throws IOException
+     */
+    @Override
+    protected void doLoad(IndexInput input) throws IOException {
+        FaissIdMapIndex faissIdMapIndex = new FaissIdMapIndex();
+        readCommonHeader(input, faissIdMapIndex);
```
shatejas, line 48: I am not sure but with the change, it should be
```suggestion
      readCommonHeader(input, this);
```

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissIdMapIndex.java
```diff
@@ -0,0 +1,242 @@
...
+/**
+ * A FAISS index with an ID mapping that maps the internal vector ID to a logical ID, along with the actual vector index.
+ * It first delegates the vector search to its nested vector index, then transforms the vector ID into a logical index that is
+ * understandable by upstream components. This is particularly useful when not all Lucene documents are indexed with a vector field.
+ * For example, if 70% of the documents have a vector field and the remaining 30% do not, the FAISS vector index will still assign
+ * increasing and continuous vector IDs starting from 0.
+ * However, these IDs only cover the sparse 30% of Lucene documents, so an ID mapping is needed to convert the internal physical vector ID
+ * into the corresponding Lucene document ID.
+ * If the mapping is an identity mapping, where each `i` is mapped to itself, we omit storing it to save memory.
+ */
```
jmazanec15, line 27: For reference serialization, can you add links to code in faiss? This will make validation easier.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissIdMapIndex.java
```diff
@@ -0,0 +1,242 @@
...
+            public ByteVectorValues copy() throws IOException {
+                return new SparseByteVectorValuesImpl(vectorValues.copy());
+            }
+        }
+
+        return new SparseByteVectorValuesImpl(vectorValues);
+    }
+
+    /**
+     * For sparse or nested cases, {@link ByteVectorValues} needs to be wrapped to correctly map an internal vector ID to a
```
jmazanec15, line 179: What do you mean by nested case?

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissIndex.java
```diff
@@ -0,0 +1,118 @@
...
+import org.apache.lucene.index.FloatVectorValues;
+import org.apache.lucene.index.VectorEncoding;
+import org.apache.lucene.store.IndexInput;
+import org.opensearch.knn.index.KNNVectorSimilarityFunction;
+import org.opensearch.knn.index.SpaceType;
+
+import java.io.IOException;
+
+/**
+ * A FAISS index file consists of multiple logical sections, each beginning with four bytes indicating an index type. A section may contain
```
jmazanec15, line 19: Add links to faiss code

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/IndexTypeToFaissIndexMapping.java
```diff
@@ -0,0 +1,34 @@
...
+
+import lombok.experimental.UtilityClass;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.function.Supplier;
+
+@UtilityClass
+public class IndexTypeToFaissIndexMapping {
```
jmazanec15, line 16: Add comments

jmazanec15 (commented):
A few minor comments, but overall looks good. Would be good to add tests

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissIdMapIndex.java
0ctopus13prime: sure will do in the next rev.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissIdMapIndex.java
0ctopus13prime: Nested case where a parent child have N child documents.
And each child document has a vector field within it. It is for nested query.
In that case, this mapping table will have a parent doc id rather than child doc id.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissIndex.java
0ctopus13prime: will do in the next rev.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/IndexTypeToFaissIndexMapping.java
0ctopus13prime: will do the next rev.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissIdMapIndex.java
0ctopus13prime: will do the next rev.

jmazanec15 (approved):
LGTM

shatejas (approved):

PR #2599, opened 2025-03-12
Add filter function to KNNQueryBuilder with unit tests and integration tests

chloewqg:
### Description
Add filter function to KNNQueryBuilder with unit tests and integration tests

### Related Issues

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
+++ b/src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
@@ -381,6 +381,46 @@ public class KNNQueryBuilder extends AbstractQueryBuilder<KNNQueryBuilder> imple
         KNNQueryBuilderParser.toXContent(builder, params, this);
     }
 
+    /**
+     * Add a filter to Neural Query Builder
+     * @param filterToBeAdded fiXlter to be added
+     * @return return itself with underlying filter combined with passed in filter
+     */
+    @Override
+    public QueryBuilder filter(QueryBuilder filterToBeAdded) {
+        if (validateFilterParams(filterToBeAdded) == false) {
+            return this;
+        }
+
+        if (this.filter == null) {
+            return KNNQueryBuilder.builder()
+                .fieldName(fieldName)
+                .vector(vector)
+                .k(k)
+                .maxDistance(maxDistance)
+                .minScore(minScore)
+                .methodParameters(methodParameters)
+                .filter(filterToBeAdded)
+                .ignoreUnmapped(ignoreUnmapped)
+                .rescoreContext(rescoreContext)
+                .expandNested(expandNested)
+                .build();
+        }
+
+        return KNNQueryBuilder.builder()
+            .fieldName(fieldName)
+            .vector(vector)
+            .k(k)
+            .maxDistance(maxDistance)
+            .minScore(minScore)
+            .methodParameters(methodParameters)
+            .filter(filter.filter(filterToBeAdded))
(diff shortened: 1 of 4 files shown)


chloewqg:
Republish the pull request here: @heemin32 @shatejas

shatejas (approved):

heemin32 (approved):

heemin32 (approved):

junqiu-lei (approved):

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
```diff
@@ -381,6 +381,46 @@ public void doXContent(XContentBuilder builder, Params params) throws IOExceptio
         KNNQueryBuilderParser.toXContent(builder, params, this);
     }
 
+    /**
+     * Add a filter to Neural Query Builder
```
bzhangam, line 385: Typo? It should be KNNQueryBuilder.

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
```diff
@@ -381,6 +381,46 @@ public void doXContent(XContentBuilder builder, Params params) throws IOExceptio
...
+     * @return return itself with underlying filter combined with passed in filter
+     */
+    @Override
+    public QueryBuilder filter(QueryBuilder filterToBeAdded) {
+        if (validateFilterParams(filterToBeAdded) == false) {
+            return this;
+        }
+
+        if (this.filter == null) {
+            return KNNQueryBuilder.builder()
```
bzhangam, line 396: Why w need to create a new KNNQueryBuilder? Can't we simply modify the filter and return this?

heemin32: There is no actual difference at the moment. However, since query validation logic is currently handled within the build method, adding a new validation related to filters in the future could be bypassed if the filter is modified directly.

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
chloewqg: Will publish a new pr fixing this.

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
bzhangam: ic. But should we modify it directly for now? If in future we do have a use case that we need to validate the filter whenever it's changed we should mark it as final to enforce that we only can modify it through the builder so that the validation will not be bypassed? I think this is a two way door and probably it's not a good idea to implement it this way for a use case that we don't have yet.

heemin32: Setting it as `final` still allows us to modify the filter; it only prevents switching the instance. This approach promotes safer code by reducing the risk of potential bugs in the future. If we might need to follow this pattern later anyway, there's no downside to adopting it now.

chloewqg: Hi Bo, thanks for your comments. According to Tejas's [comments](https://github.com/opensearch-project/k-NN/pull/2585#discussion_r1984080747), modifying an existing query builder is not a great idea, the builder is created to make sure its immutable in a way and every parameter is validated together rather than individually. I think Tejas's point makes sense that the query builder were not supposed to be changed after creation.

PR #2603, opened 2025-03-14
[Remote Index Client] Move remote client to a separate module

### LINKED ISSUE #2602: [Remote Vector Index Build] Move Remote Client to Separate Repository
PRs #2560 and #2576 saw the creation of the RemoteIndexClient in the k-NN plugin. The client as it stands is tightly coupled with k-NN by way of setting retrievals and other k-NN method calls. In order to reduce these dependencies, work has started in #2588 to move the HTTP client into a separate module that would only depend on `opensearch-core`.

At this point, it makes sense to go one step further and move the client to the new separate repository for the remote build service, at https://github.com/opensearch-project/remote-vector-index-builder. This will reduce the dependencies even further, allow for better modularity, independent versioning, and potential reuse by other plugins.

The first priority will be to finish the independent module work started by @navneet1v in #2588. This work of moving the client to the new repository will be done on the side.

owenhalpert:
### Description
This PR moves the RemoteIndexClient and associated classes to a separate module within k-NN to reduce dependencies between the two. This is to support an eventual move to a separate repo (https://github.com/opensearch-project/remote-vector-index-builder). 

At a high level, all the client classes (including request and response objects) are moved to a `remote-index-build-client` module. The RemoteIndexWaiter/Poller implementation are left in k-NN as they represent concrete usages of the client but not client implementations themself.

For the testing, most of the unit tests were testing full functionality with k-NN objects being passed into the methods. Without the k-NN dependency, this is no longer possible, so the unit tests in the client are more lightweight unit tests and some of the more involved tests with k-NN constructs are moved to classes like `RemoteIndexBuildStrategy`, where they create a client for the tests (but don't otherwise share any constants or strong dependencies that the plugin itself doesn't use).

A small change was made to the RemoteIndexHTTPClient constructor to support dependency injection for testing and to make the static initialization logic and holder pattern a little more readable.

Thanks to @navneet1v for the early work on this refactoring.

### Related Issues
Resolves #2602 

### Check List


remote-index-build-client/src/test/java/org/opensearch/remoteindexbuild/client/RemoteIndexHTTPClientTests.java
```diff
@@ -0,0 +1,148 @@
...
+        when(metadata.settings()).thenReturn(repoSettings);
+        return metadata;
+    }
+
+    public static IndexSettings createTestIndexSettings() {
+        IndexSettings mockIndexSettings = mock(IndexSettings.class);
+        Settings indexSettingsSettings = Settings.builder().put(ClusterName.CLUSTER_NAME_SETTING.getKey(), TEST_CLUSTER).build();
+        when(mockIndexSettings.getSettings()).thenReturn(indexSettingsSettings);
+        return mockIndexSettings;
+    }
```
navneet1v, line 143: none of this is needed in this client test class. I think we can remove it.

remote-index-build-client/src/test/java/org/opensearch/remoteindexbuild/client/RemoteIndexHTTPClientTests.java
```diff
@@ -0,0 +1,148 @@
...
+
+    protected AutoCloseable openMocks;
+
+    @Before
+    public void setup() {
+        openMocks = MockitoAnnotations.openMocks(this);
+        clusterService = mock(ClusterService.class);
+        Set<Setting<?>> defaultClusterSettings = new HashSet<>(ClusterSettings.BUILT_IN_CLUSTER_SETTINGS);
+        when(clusterService.getClusterSettings()).thenReturn(new ClusterSettings(Settings.EMPTY, defaultClusterSettings));
+    }
```
navneet1v, line 68: I don't think we need these in the tests now.

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
```diff
@@ -188,4 +196,48 @@ private BlobStoreRepository getRepository() throws RepositoryMissingException {
...
+    /**
+     * Constructor for RemoteBuildRequest.
+     *
+     * @param indexSettings IndexSettings object
+     * @param indexInfo BuildIndexParams object
+     * @param repositoryMetadata RepositoryMetadata object
+     * @param fullPath Full blob path + file name representing location of the vectors/doc IDs (excludes repository-specific prefix)
+     * @throws IOException if an I/O error occurs
+     */
+    public static RemoteBuildRequest buildRemoteBuildRequest(
```
navneet1v, line 208: why this is public?

navneet1v:
Please fix all the github CIs.

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
owenhalpert: I've made it package-private for testing. If we think it's necessary, I can work on rewriting the test to access it via `buildAndWriteIndex` and keep it private.

navneet1v (approved):

Vikasht34 (approved):
LGTM !! Thanks for moving this .

navneet1v:
waiting for CIs to complete before merging the code to main branch

PR #2608, opened 2025-03-17
[LuceneOnFaiss Part-7] Enable memory optimized searching in VectorReader for FAISS engine.

0ctopus13prime:
### Description
This PR enables memory optimized search in VectorReader for FAISS HNSW graph index.
Internally, it will delegate Lucene's HNSW graph searcher to perform vector search on FAISS index.

NOTE : Will add unit tests shortly.

### Related Issues

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
+++ b/src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
@@ -22,6 +22,7 @@ import org.apache.lucene.search.KnnCollector;
 import org.apache.lucene.search.ScoreDoc;
 import org.apache.lucene.search.TopDocs;
 import org.apache.lucene.search.TotalHits;
+import org.apache.lucene.store.IOContext;
 import org.apache.lucene.util.Bits;
 import org.apache.lucene.util.IOSupplier;
 import org.apache.lucene.util.IOUtils;
@@ -37,12 +38,12 @@ import org.opensearch.knn.quantization.models.quantizationState.QuantizationStat
 import org.opensearch.knn.quantization.models.quantizationState.QuantizationStateCacheManager;
 import org.opensearch.knn.quantization.models.quantizationState.QuantizationStateReadConfig;
 
+import java.io.Closeable;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Objects;
 
 import static org.opensearch.knn.common.KNNConstants.KNN_ENGINE;
 import static org.opensearch.knn.index.mapper.KNNVectorFieldMapper.KNN_FIELD;
@@ -60,21 +61,19 @@ public class NativeEngines990KnnVectorsReader extends KnnVectorsReader {
     private Map<String, String> quantizationStateCacheKeyPerField;
     private SegmentReadState segmentReadState;
     private final List<String> cacheKeys;
-    private Map<String, VectorSearcher> vectorSearchers;
+    private volatile Map<String, VectorSearcher> vectorSearchers;
 
     public NativeEngines990KnnVectorsReader(final SegmentReadState state, final FlatVectorsReader flatVectorsReader) {
         this.flatVectorsReader = flatVectorsReader;
         this.segmentReadState = state;
         this.cacheKeys = getVectorCacheKeysFromSegmentReaderState(state);
-        this.vectorSearchers = new HashMap<>(RESERVE_TWICE_SPACE * segmentReadState.fieldInfos.size(), SUFFICIENT_LOAD_FACTOR);
+        this.vectorSearchers = null;
+
         loadCacheKeyMap();
 
-        //
-        // TMP(KDY) : Dynamic update will be covered in part-7. Please refer to
-        // https://github.com/opensearch-project/k-NN/issues/2401#issuecomment-2699777824
-        //
-        final boolean isMemoryOptimizedSearchEnabled = false;
-        if (isMemoryOptimizedSearchEnabled) {
+        // Memory optimized searcher will be ONLY used for searching, we don't need this during merging.
+        // TODO(KDY) : Enable this based on setting flag value. e.g. index.knn.memory_optimized_search = True
+        if (state.context.context() != IOContext.Context.MERGE) {
             loadMemoryOptimizedSearcher();
         }
     }
@@ -107,19 +106,34 @@ public class NativeEngines990KnnVectorsReader extends KnnVectorsReader {
         return null;
     }
 
-    private void loadMemoryOptimizedSearcher() {
+    private synchronized void loadMemoryOptimizedSearcher() {
+        if (vectorSearchers != null) {
+            return;
+        }
+
+        final Map<String, VectorSearcher> vectorSearcherPerField = new HashMap<>(
+            RESERVE_TWICE_SPACE * segmentReadState.fieldInfos.size(),
+            SUFFICIENT_LOAD_FACTOR
+        );
+
         try {
             for (FieldInfo fieldInfo : segmentReadState.fieldInfos) {
                 final IOSupplier<VectorSearcher> searcherSupplier = getIndexFileNameIfMemoryOptimizedSearchSupported(fieldInfo);
                 if (searcherSupplier != null) {
-                    final VectorSearcher searcher = Objects.requireNonNull(searcherSupplier.get());
-                    vectorSearchers.put(fieldInfo.getName(), searcher);
+                    final VectorSearcher searcher = searcherSupplier.get();
+                    if (searcher != null) {
+                        // It's supported. There can be a case where a certain index type underlying is not yet supported while KNNEngine
+                        // itself supports memory optimized searching.
+                        vectorSearcherPerField.put(fieldInfo.getName(), searcher);
+                    }
                 }
             }
+
+            vectorSearchers = vectorSearcherPerField;
         } catch (Exception e) {
             // Close opened searchers first, then suppress
             try {
-                IOUtils.closeWhileHandlingException(vectorSearchers.values());
+                IOUtils.closeWhileHandlingException(vectorSearcherPerField.values());
             } catch (Exception closeException) {
                 log.error(closeException.getMessage(), closeException);
             }
@@ -207,10 +221,7 @@ public class NativeEngines990KnnVectorsReader extends KnnVectorsReader {
             return;
         }
 
-        // Try with memory optimized searcher
-        final VectorSearcher memoryOptimizedSearcher = vectorSearchers.get(field);
-        if (memoryOptimizedSearcher != null) {
-            memoryOptimizedSearcher.search(target, knnCollector, acceptDocs);
+        if (trySearchWithMemoryOptimizedSearch(field, target, knnCollector, acceptDocs, true)) {
             return;
         }
 
@@ -244,6 +255,10 @@ public class NativeEngines990KnnVectorsReader extends KnnVectorsReader {
      */
     @Override
     public void search(String field, byte[] target, KnnCollector knnCollector, Bits acceptDocs) throws IOException {
+        if (trySearchWithMemoryOptimizedSearch(field, target, knnCollector, acceptDocs, false)) {
+            return;
+        }
+
         throw new UnsupportedOperationException("Search functionality using codec is not supported with Native Engine Reader");
     }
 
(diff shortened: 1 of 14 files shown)


src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissMemoryOptimizedSearcher.java
```diff
@@ -16,32 +23,91 @@
...
 
     @Override
     public void search(float[] target, KnnCollector knnCollector, Bits acceptDocs) throws IOException {
-        // TODO(KDY) : This will be covered in part-7 (Complete Faiss Loading Part at Codec Level)
-        throw new UnsupportedOperationException("Not implemented yet");
+        search(
+            VectorEncoding.FLOAT32,
+            () -> VECTOR_SCORER.getRandomVectorScorer(
+                faissIndex.getVectorSimilarityFunction().getVectorSimilarityFunction(),
+                faissIndex.getFloatValues(indexInput),
```
jmazanec15, line 53: Are we opening up the flatVectorReaders as well already? Can we avoid one of the two?

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
```diff
@@ -207,6 +217,9 @@ public void search(String field, float[] target, KnnCollector knnCollector, Bits
             return;
         }
 
+        // Lazy load vector searcher if not loaded.
+        loadMemoryOptimizedSearcherIfNotLoaded();
```
jmazanec15, line 221: Do we need to do this lazily?

shatejas: +1

Will there be any latency impact here? the idea all initialization done as soon as the reader is opened after refresh or merge

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
```diff
@@ -107,23 +100,40 @@ private IOSupplier<VectorSearcher> getIndexFileNameIfMemoryOptimizedSearchSuppor
...
+
+        synchronized (this) {
+            if (vectorSearchers != null) {
+                return;
             }
-        } catch (Exception e) {
-            // Close opened searchers first, then suppress
+
+            final Map<String, VectorSearcher> vectorSearcherPerField = new HashMap<>(
+                RESERVE_TWICE_SPACE * segmentReadState.fieldInfos.size(),
```
shatejas, line 114: Why reserve more though? At this point the segment is flushed and there will be no more vector fields per segment than fielsInfos.size I think?

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissMemoryOptimizedSearcher.java
```diff
@@ -16,32 +23,91 @@
...
+        final KnnCollector collector = new OrdinalTranslatedKnnCollector(knnCollector, scorer::ordToDoc);
+        final Bits acceptedOrds = scorer.getAcceptOrds(acceptDocs);
+
+        if (knnCollector.k() < scorer.maxOrd()) {
+            // Do ANN search with Lucene's HNSW graph searcher.
+            HnswGraphSearcher.search(scorer, collector, faissHnswGraph, acceptedOrds);
+        } else {
+            // If k is larger than the number of vectors, we can just iterate over all vectors
+            // and collect them.
+            for (int i = 0; i < scorer.maxOrd(); i++) {
```
shatejas, line 101: nit: this should eventually intergrate with exact searcher to benefit from all its future optimizations. Let me know your thoughts on it, maybe add a TODO if you are aligned

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissMemoryOptimizedSearcher.java
```diff
@@ -16,32 +23,91 @@
...
         indexInput.close();
     }
+
+    private void search(
+        final VectorEncoding vectorEncoding,
+        final IOSupplier<RandomVectorScorer> scorerSupplier,
+        final KnnCollector knnCollector,
+        final Bits acceptDocs
+    ) throws IOException {
+        if (faissIndex.getTotalNumberOfVectors() == 0 || knnCollector.k() == 0 || faissIndex.getVectorEncoding() != vectorEncoding) {
```
shatejas, line 86: > faissIndex.getVectorEncoding() != vectorEncoding

is this an error case?

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissMemoryOptimizedSearcher.java
```diff
@@ -16,32 +23,91 @@
...
+                target
+            ),
+            knnCollector,
+            acceptDocs
+        );
     }
 
     @Override
     public void close() throws IOException {
         indexInput.close();
```
shatejas, line 77: Ideally we should be closing the index input where we opened it, this is purely from maintenance perspective, it can be hard to track down whether the resource is closed or not if its not done in the same class where its opened

shatejas (commented):
Overall looks good, a few minor things

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
0ctopus13prime: During the search, this table will be used to look up VectorSearcher for every query. 
So having larger space to approximate perfect hash for the better performance.
Asymptotic space is still O(N), won't bring impact on memory pressure.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissMemoryOptimizedSearcher.java
0ctopus13prime: sure, will raise as a future PR.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissMemoryOptimizedSearcher.java
0ctopus13prime: I think we can see this as FaissMemoryOptimizedSearcher has the ownership of IndexInput. 
Likewise passing RValue in C++. `Ex: Foo(std::move(ifstream))`
It is natural to close the resource within the class as the class's life cycle is the same as IndexInput.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
0ctopus13prime: sure, will update in the next rev.
When I raised this revision, underlying assumption is the setting would be dynamic.
But, as the setting will be static, we can load based on the flag value.
Thank you.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissMemoryOptimizedSearcher.java
0ctopus13prime: On second thought, yes it is an error case.
This indicates that the mapping is configured as BYTE, but underlying index is indexed as FLOAT or vice versa.

0ctopus13prime: Will make it throw an exception likewise we're doing it in FAISS C++

0ctopus13prime:
## Testing covered
1. Dense case : Where all docs have vectors.
2. Sparse case : Only portion of docs have vectors.
3. Dense nested case : All parent docs have vectors.
4. Sparse nested case : Portions of parent docs have vectors

Index type : Float index

Byte index + FP16 will be covered in part6 PR.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
```diff
@@ -107,19 +106,34 @@ private IOSupplier<VectorSearcher> getIndexFileNameIfMemoryOptimizedSearchSuppor
         return null;
     }
 
-    private void loadMemoryOptimizedSearcher() {
+    private synchronized void loadMemoryOptimizedSearcher() {
```
jmazanec15, line 109: nit: do we need sync?

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
```diff
@@ -276,9 +299,33 @@ public void close() throws IOException {
...
+
+    private boolean trySearchWithMemoryOptimizedSearch(
+        String field,
+        Object target,
+        KnnCollector knnCollector,
+        Bits acceptDocs,
+        boolean isFloatVector
+    ) throws IOException {
+        if (vectorSearchers == null) {
+            // Null `vectorSearchers` indicates that this reader was instantiated during merge.
```
jmazanec15, line 312: Can you elaborate on this case? Is this possible if we override getMergeInstance() method in this class? Are there any links in lucene that explain this?

See https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/codecs/KnnVectorsReader.java#L117-L125.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
0ctopus13prime: I think we should go with lazy-loading fashion for memory optimized searcher.
It is tricky to get index name in VectorReacer on codec level. Also it has SPI issue, having a dependency to KNNSettings which relies on cluster state. 
Since we can identify the faiss index file name from field attribute at the moment, when search comes (this indicates that memory-optimized-search is enabled), and the file hasn't been partially loaded, then try to load it.

Therefore, `synchronized` will be needed to guard against several search threads.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
0ctopus13prime: This will be replaced to lazy-loading in part-8. https://github.com/opensearch-project/k-NN/pull/2616
Since we can identify the faiss index file name from field attribute at the moment, when search comes (this indicates that memory-optimized-search is enabled), and the file hasn't been partially loaded, then try to load it.

No nullity check will be needed.

jmazanec15 (approved):

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissMemoryOptimizedSearcher.java
shatejas: So one of the simplest thing you can do is open the index input outside of the factory and pass the index input as a parameter. This will force the responsibility of opening the index input in the reader making it responsible for its lifecycle
```
public VectorSearcher createVectorSearcher(final Directory directory, final IndexInput faissSegmentIndexInput) throws IOException
```

src/test/java/org/opensearch/knn/memoryoptsearch/FaissMemoryOptimizedSearcherTests.java
```diff
@@ -0,0 +1,621 @@
...
+    private static final String TARGET_FIELD = "target_field";
+    private static final String FLOAT_HNSW_INDEX_DESCRIPTION = "HNSW16,Flat";
+    private static final String BYTE_HNSW_INDEX_DESCRIPTION = "HNSW16,SQ8_direct_signed";
+    private static final int DIMENSIONS = 128;
+    private static final int MIN_VALUE = -1000000;
+    private static final int MAX_VALUE = 1000000;
+    private static final int TOTAL_NUM_DOCS_IN_SEGMENT = 1000;
+    private static final int NUM_CHILD_DOCS = 5;
+
+    public void testFloatIndexType() {
```
shatejas, line 86: This is an excellent exhaustive test. Thanks for this!

shatejas (approved):

PR #2609, opened 2025-03-17
[LuceneOnFaiss] Apply monotonic increasing integer encoding to FAISS HNSW and IdMapIndex.

0ctopus13prime:
### Description
This PR introduces a monotonic integer sequence encoding scheme for doc IDs in IdMapIndex and offsets in FAISS HNSW.
Previously, both were stored as long[], consuming O(N) memory where N is the total number of vectors. For example, with a 10M dataset, they occupied 228MB (76MB + 152MB). Using monotonic encoding, this can be reduced to a few hundred KB. Lucene already uses this encoding scheme [Link](https://github.com/apache/lucene/blob/9472dcad0ebfadf2ae6460793682635880ac48eb/lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsReader.java#L402C13-L402C17)

The key difference is that Lucene applies encoding during segment creation, whereas we do it at load time. However, encoding is extremely fast, on my local setup it takes about 3.2 seconds for 100M integers.
A warm-up API could further optimize this by preloading data before search, though this hasn't been raised yet, but it is part of future PR.

### Related Issues

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissHnswGraph.java
+++ b/src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissHnswGraph.java
@@ -11,6 +11,7 @@ import org.apache.lucene.util.hnsw.HnswGraph;
 
 import java.io.IOException;
 import java.util.NoSuchElementException;
+import java.util.Objects;
 
 import static org.apache.lucene.search.DocIdSetIterator.NO_MORE_DOCS;
 
@@ -18,7 +19,7 @@ import static org.apache.lucene.search.DocIdSetIterator.NO_MORE_DOCS;
  * This graph implements Lucene's HNSW graph interface using the FAISS HNSW graph. Conceptually, both libraries represent the graph
  * similarly, maintaining a list of neighbor IDs. This implementation acts as a bridge, enabling Lucene's HNSW graph searcher to perform
  * vector searches on a FAISS index.
- *
+ * <p>
  * NOTE: This is not thread safe. It should be created every time in {@link KnnVectorsReader}.search likewise
  * <a href="https://github.com/apache/lucene/blob/92290a0201458152c9e03d199f38f2e8a479f045/lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsReader.java#L467">OffHeapHnswGraph</a>
  * in Lucene.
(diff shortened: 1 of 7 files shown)


src/main/java/org/opensearch/knn/memoryoptsearch/faiss/MonotonicIntegerSequenceEncoder.java
```diff
@@ -0,0 +1,95 @@
...
+import org.apache.lucene.store.IndexInput;
+import org.apache.lucene.util.IOUtils;
+import org.apache.lucene.util.packed.DirectMonotonicReader;
+import org.apache.lucene.util.packed.DirectMonotonicWriter;
+import org.opensearch.common.lucene.store.ByteArrayIndexInput;
+
+import java.io.IOException;
+
+@UtilityClass
+public class MonotonicIntegerSequenceEncoder {
```
jmazanec15, line 20: Are there no utilities in OpenSearch or Lucene that do this already?

0ctopus13prime: no, I could not find any useful. 
This has three goals
1. Make sure the value is within Integer value range : `Math.toIntExact`
2. For identical mapping (e.g. i -> i), it will return null as we don't actually need it when i will always be mapped to i.

I think I need to add more comments on this, but this is doing what tailored for partial loading case.

0ctopus13prime:
@shatejas @jmazanec15 Could you approve this PR if it looks good on you? :) 
Thank you

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissHnswGraph.java
```diff
@@ -45,7 +45,7 @@ public FaissHnswGraph(final FaissHNSW faissHNSW, final IndexInput indexInput) {
     @Override
     public void seek(int level, int internalVectorId) {
         // Get a relative starting offset of neighbor list at `level`.
-        long o = faissHnsw.getOffsets()[internalVectorId];
+        long o = faissHnsw.getOffsetsReader().get(internalVectorId);
```
shatejas, line 48: offsetreader can be null here - it might be rare or not possible but its better to handle the case since `MonotonicIntegerSequenceEncoder.encode` returns null in isIdentical case

0ctopus13prime: Offsets never can be null, as it won't be identical mapping. 
But, will add its validation in FaissHnswGraph's constructor.

jmazanec15 (approved):

PR #2610, opened 2025-03-17
Add VectorDataType check to NativeIndexBuildStrategy

jed326:
### Description
Check for supported `VectorDataType` in `NativeIndexBuildStrategy`

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/engine/faiss/Faiss.java
+++ b/src/main/java/org/opensearch/knn/index/engine/faiss/Faiss.java
@@ -8,6 +8,7 @@ package org.opensearch.knn.index.engine.faiss;
 import com.google.common.collect.ImmutableMap;
 import org.opensearch.knn.common.KNNConstants;
 import org.opensearch.knn.index.SpaceType;
+import org.opensearch.knn.index.VectorDataType;
 import org.opensearch.knn.index.engine.KNNMethod;
 import org.opensearch.knn.index.engine.KNNMethodConfigContext;
 import org.opensearch.knn.index.engine.KNNMethodContext;
@@ -125,10 +126,10 @@ public class Faiss extends NativeLibrary {
      * Use the method name to route the check to the specific method class
      */
     @Override
-    public boolean supportsRemoteIndexBuild(MethodComponentContext methodComponentContext) {
+    public boolean supportsRemoteIndexBuild(MethodComponentContext methodComponentContext, VectorDataType vectorDataType) {
         if (METHOD_HNSW.equals(methodComponentContext.getName())) {
             if (methodComponentContext.getParameters() != null) {
-                return FaissHNSWMethod.supportsRemoteIndexBuild(methodComponentContext.getParameters());
+                return FaissHNSWMethod.supportsRemoteIndexBuild(methodComponentContext.getParameters(), vectorDataType);
             }
         }
         return false;
--- a/src/main/java/org/opensearch/knn/index/engine/faiss/FaissHNSWMethod.java
+++ b/src/main/java/org/opensearch/knn/index/engine/faiss/FaissHNSWMethod.java
@@ -49,6 +49,8 @@ public class FaissHNSWMethod extends AbstractFaissMethod {
         VectorDataType.BYTE
     );
 
+    private static final Set<VectorDataType> SUPPORTED_REMOTE_INDEX_DATA_TYPES = ImmutableSet.of(VectorDataType.FLOAT);
+
     public final static List<SpaceType> SUPPORTED_SPACES = Arrays.asList(
         SpaceType.UNDEFINED,
         SpaceType.HAMMING,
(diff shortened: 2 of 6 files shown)


navneet1v (approved):

src/main/java/org/opensearch/knn/index/engine/faiss/Faiss.java
```diff
@@ -125,7 +126,10 @@ public ResolvedMethodContext resolveMethod(
      * Use the method name to route the check to the specific method class
      */
     @Override
-    public boolean supportsRemoteIndexBuild(MethodComponentContext methodComponentContext) {
+    public boolean supportsRemoteIndexBuild(MethodComponentContext methodComponentContext, VectorDataType vectorDataType) {
+        if (vectorDataType != VectorDataType.FLOAT) {
+            return false;
+        }
```
owenhalpert, line 132: This is nitpicky, but if we're taking the extra step to make this an engine-specific check, should we by the same logic make it an engine/method combo specific check and pass this along to FaissHNSW in 135? The question to answer that would be is it plausible that we would ever support (using examples here) FaissIVF with binary but not FaissHNSW? As of now this method symbolically is routing the check to the correct method class which checks the compatibility, so having this here could be out of place.

jed326: > is it plausible that we would ever support (using examples here) FaissIVF with binary but not FaissHNSW? 

That's a good question. Historically, yes FaissHNSW supported binary vectors before FaissIVF and whether or not one supports a given data type should be independent from the other. Let me rework this a little and do something like this: https://github.com/opensearch-project/k-NN/blob/495ec329cb260a9f76c3974366d7e2c72f9924ec/src/main/java/org/opensearch/knn/index/engine/faiss/FaissHNSWMethod.java#L46-L50

src/main/java/org/opensearch/knn/index/engine/faiss/FaissHNSWMethod.java
```diff
@@ -172,9 +174,13 @@ public static RemoteIndexParameters createRemoteIndexingParameters(KNNMethodCont
     /**
      * Return whether this engine/method supports remote build, currently by checking the encoder to ensure FP32.
```
owenhalpert, line 175: Can you update the Javadoc? You can just change it to: 
> Return whether this engine/method supports remote build.

if you think thats better.

owenhalpert:
LGTM thanks Jay

owenhalpert (approved):

src/main/java/org/opensearch/knn/index/engine/faiss/FaissHNSWMethod.java
```diff
@@ -170,11 +172,16 @@ public static RemoteIndexParameters createRemoteIndexingParameters(KNNMethodCont
...
+     * @param vectorDataType        {@link VectorDataType}
+     * @return                      true if the method parameters + vector data type combination is supported for remote index build
      */
-    static boolean supportsRemoteIndexBuild(Map<String, Object> methodParameters) {
-        String encoderName = ((MethodComponentContext) methodParameters.get("encoder")).getName();
-        return ENCODER_FLAT.equals(encoderName);
+    static boolean supportsRemoteIndexBuild(Map<String, Object> methodParameters, VectorDataType vectorDataType) {
+        if (SUPPORTED_REMOTE_INDEX_DATA_TYPES.contains(vectorDataType)) {
+            String encoderName = ((MethodComponentContext) methodParameters.get("encoder")).getName();
+            return ENCODER_FLAT.equals(encoderName);
```
epalaz, line 182: Nit: Encoder names can follow a similar pattern to data type, supported types. Something like `SUPPORTED_REMOTE_INDEX_ENCODERS`

src/main/java/org/opensearch/knn/index/engine/faiss/FaissHNSWMethod.java
```diff
@@ -170,11 +172,16 @@ public static RemoteIndexParameters createRemoteIndexingParameters(KNNMethodCont
...
+     * @param methodParameters      Map of method parameters including encoder information
+     * @param vectorDataType        {@link VectorDataType}
+     * @return                      true if the method parameters + vector data type combination is supported for remote index build
      */
-    static boolean supportsRemoteIndexBuild(Map<String, Object> methodParameters) {
-        String encoderName = ((MethodComponentContext) methodParameters.get("encoder")).getName();
-        return ENCODER_FLAT.equals(encoderName);
+    static boolean supportsRemoteIndexBuild(Map<String, Object> methodParameters, VectorDataType vectorDataType) {
+        if (SUPPORTED_REMOTE_INDEX_DATA_TYPES.contains(vectorDataType)) {
+            String encoderName = ((MethodComponentContext) methodParameters.get("encoder")).getName();
```
epalaz, line 181: Nit: this is not your change but since you are touching this. Looks like `METHOD_ENCODER_PARAMETER` is the project wide constant for "encoder"

epalaz (approved):
Left some nits but LGTM!

src/main/java/org/opensearch/knn/index/engine/faiss/FaissHNSWMethod.java
owenhalpert: Good catch, @jed326 I think we should include this if you don't mind.

src/main/java/org/opensearch/knn/index/engine/faiss/FaissHNSWMethod.java
jed326: I'm going to leave this open for now as there might be additional parsing that needs to happen for other encoder types, we can revisit this when we need to add more encoders in the future.

jmazanec15 (approved):

PR #2615, opened 2025-03-19
[Remote Vector Index Build] Add metric collection

owenhalpert:
### Description
Adds metric collection for various metrics related to the Remote Vector Index Build service.

From @jed326's original PR:

> New metrics:
> ```
>       "remote_vector_index_build_stats" : {
>         "repository_stats" : {
>           "write_time_in_millis" : 15,
>           "read_success_count" : 0,
>           "read_failure_count" : 0,
>           "write_success_count" : 1,
>           "write_failure_count" : 0,
>           "read_time_in_millis" : 0
>         },
>         "client_stats" : {
>           "status_request_failure_count" : 0,
>           "status_request_success_count" : 0,
>           "index_build_success_count" : 0,
>           "index_build_failure_count" : 0,
>           "build_request_failure_count" : 1,
>           "build_request_success_count" : 0,
>           "waiting_time_in_ms" : 0
>         },
>         "build_stats" : {
>           "remote_index_build_time_in_millis" : 23,
>           "remote_index_build_current_size" : 0,
>           "remote_index_build_current_operations" : 0
>         }
>       },
> ```
> Complementary to these, existing repository implementations may also collect their own metrics. For example, repository-s3 collects metrics like so:
> Existing repo metrics:
> ```
>      "repositories" : [
>         {
>           "repository_name" : "vector-repo",
>           "repository_type" : "s3",
>           "repository_location" : {
>             "bucket" : "gpu-acceleration",
>             "base_path" : "vectors"
>           },
>           "request_time_in_millis" : {
>             "ListObjects" : 49,
>             "PutMultipartObject" : 0,
>             "PutObject" : 199,
>             "GetObject" : 38,
>             "DeleteObjects" : 51
>           },
>           "request_retry_count_total" : {
>             "ListObjects" : 0,
>             "PutMultipartObject" : 0,
>             "PutObject" : 0,
>             "GetObject" : 0,
>             "DeleteObjects" : 0
>           },
>           "request_failures_total" : {
>             "ListObjects" : 0,
>             "PutMultipartObject" : 0,
>             "PutObject" : 0,
>             "GetObject" : 0,
>             "DeleteObjects" : 0
>           },
>           "generic_stats" : {
>             "LowPriorityPermitsUtilization" : 0,
>             "NormalPriorityPermitsUtilization" : 0,
>             "LowPriorityQUtilization" : 0,
>             "NormalPriorityQUtilization" : 0
>           },
>           "request_success_total" : {
>             "ListObjects" : 1,
>             "PutMultipartObject" : 0,
>             "PutObject" : 6,
>             "GetObject" : 3,
>             "DeleteObjects" : 1
>           }
>         }
>       ],
> ```

### Related Issues
Meta #2391

### Check List


src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
```diff
@@ -120,63 +134,154 @@ public static boolean shouldBuildIndexRemotely(IndexSettings indexSettings, long
...
             vectorRepositoryAccessor.writeToRepository(
-                blobName,
+                repositoryContext.blobName,
                 indexInfo.getTotalLiveDocs(),
                 indexInfo.getVectorDataType(),
                 indexInfo.getKnnVectorValuesSupplier()
             );
             time_in_millis = stopWatch.stop().totalTime().millis();
+            WRITE_SUCCESS_COUNT.increment();
+            WRITE_TIME.incrementBy(time_in_millis);
```
jed326, line 191: Should we use the same `startXStats` pattern here? I think I like having the metrics separated from the business logic.

src/main/java/org/opensearch/knn/plugin/stats/KNNStats.java
```diff
@@ -218,4 +229,70 @@ private Map<String, Map<String, Object>> createGraphStatsMap() {
...
+        repoStatsMap.put(KNNRemoteIndexBuildValue.WRITE_FAILURE_COUNT.getName(), KNNRemoteIndexBuildValue.WRITE_FAILURE_COUNT.getValue());
+        repoStatsMap.put(KNNRemoteIndexBuildValue.WRITE_TIME.getName(), KNNRemoteIndexBuildValue.WRITE_TIME.getValue());
+        repoStatsMap.put(KNNRemoteIndexBuildValue.READ_SUCCESS_COUNT.getName(), KNNRemoteIndexBuildValue.READ_SUCCESS_COUNT.getValue());
+        repoStatsMap.put(KNNRemoteIndexBuildValue.READ_FAILURE_COUNT.getName(), KNNRemoteIndexBuildValue.READ_FAILURE_COUNT.getValue());
+        repoStatsMap.put(KNNRemoteIndexBuildValue.READ_TIME.getName(), KNNRemoteIndexBuildValue.READ_TIME.getValue());
+
+        Map<String, Object> buildStatsMap = new HashMap<>();
+        buildStatsMap.put(
+            KNNRemoteIndexBuildValue.REMOTE_INDEX_BUILD_CURRENT_OPERATIONS.getName(),
+            KNNRemoteIndexBuildValue.REMOTE_INDEX_BUILD_CURRENT_OPERATIONS.getValue()
```
jed326, line 276: I'm fine with taking it up as a follow-up PR, but I think we should figure out a way to separate remote flush vs local flush and remote merge vs local merge

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
owenhalpert: Check out the helper methods I made for each phase — I don't love all of these in the `BuildStrategy` class, but it is convenient for access to some objects in the class

src/main/java/org/opensearch/knn/plugin/stats/KNNStats.java
owenhalpert: I added this logic, let me know what you think of the approach

src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexWriter.java
```diff
@@ -153,7 +155,11 @@ private void buildAndWriteIndex(final Supplier<KNNVectorValues<?>> knnVectorValu
                 totalLiveDocs,
                 knnVectorValuesSupplier.get()
             );
-            indexBuilder.buildAndWriteIndex(nativeIndexParams);
+            if (indexBuilder instanceof RemoteIndexBuildStrategy) {
+                indexBuilder.buildAndWriteIndex(nativeIndexParams, isFlush);
+            } else {
+                indexBuilder.buildAndWriteIndex(nativeIndexParams);
+            }
```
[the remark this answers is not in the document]
owenhalpert: Good call!

src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexBuildStrategy.java
```diff
@@ -14,5 +14,21 @@
...
     void buildAndWriteIndex(BuildIndexParams indexInfo) throws IOException;
+
+    /**
+     * Build and write index with flush flag. Default implementation calls the non-flag version.
+     * RemoteIndexBuildStrategy will override this to use the flag.
+     *
+     * @param indexInfo {@link BuildIndexParams} containing information about the index to be built
+     * @param isFlush Flag indicating if operation is from flush (vs merge)
+     * @throws IOException if an error occurs during the build process
+     */
```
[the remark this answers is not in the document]
shatejas: There doesn't seem to be a need for this method - why can't we add `context` in `BuildIndexParams`? the context can be either merge or flush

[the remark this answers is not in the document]
owenhalpert: We can do that, good idea.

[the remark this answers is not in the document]
owenhalpert: @shatejas I added a boolean `isFlush` field in `BuildIndexParams`. I considered having this be a string of either `merge` or `flush` but I felt that keeping it as a boolean named `isFlush` is more clear for its usage.

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
```diff
@@ -116,67 +133,149 @@ public static boolean shouldBuildIndexRemotely(IndexSettings indexSettings, long
...
             );
-            stopWatch = new StopWatch().start();
-            final RemoteBuildResponse remoteBuildResponse = client.submitVectorBuild(buildRequest);
-            time_in_millis = stopWatch.stop().totalTime().millis();
-            log.debug("Submit vector build took {} ms for vector field [{}]", time_in_millis, indexInfo.getFieldName());
+            remoteBuildResponse = client.submitVectorBuild(buildRequest);
+            recordBuildRequestSuccess(stopWatch.stop().totalTime().millis(), indexInfo.getFieldName());
+            return remoteBuildResponse;
+        } catch (IOException e) {
+            recordBuildRequestFailure(stopWatch.stop().totalTime().millis(), indexInfo.getFieldName(), e);
```
shatejas, line 225: We need to do these in finally block. Any other exception and you are not capturing failure. another approach is to catch a generic approach which is generally not recommended. I see the pattern at other places as well. 
```
bool success = false;
try {

success = true;

......
} finally {
    recordBuildRequestMetrics(success, stopWatch.millis);
}
```

I see you are passing in the exception and its just used for logging. Please log in the catch block so you don't have to pass the error

owenhalpert: Thanks. Can you take a look at the new flow, with the metric collection in the `finally` blocks?

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
```diff
@@ -43,6 +42,24 @@
...
+import static org.opensearch.knn.plugin.stats.KNNRemoteIndexBuildValue.REMOTE_INDEX_BUILD_CURRENT_FLUSH_OPERATIONS;
+import static org.opensearch.knn.plugin.stats.KNNRemoteIndexBuildValue.REMOTE_INDEX_BUILD_CURRENT_FLUSH_SIZE;
+import static org.opensearch.knn.plugin.stats.KNNRemoteIndexBuildValue.REMOTE_INDEX_BUILD_CURRENT_MERGE_OPERATIONS;
+import static org.opensearch.knn.plugin.stats.KNNRemoteIndexBuildValue.REMOTE_INDEX_BUILD_CURRENT_MERGE_SIZE;
+import static org.opensearch.knn.plugin.stats.KNNRemoteIndexBuildValue.REMOTE_INDEX_BUILD_FLUSH_TIME;
+import static org.opensearch.knn.plugin.stats.KNNRemoteIndexBuildValue.REMOTE_INDEX_BUILD_MERGE_TIME;
+import static org.opensearch.knn.plugin.stats.KNNRemoteIndexBuildValue.WAITING_TIME;
+import static org.opensearch.knn.plugin.stats.KNNRemoteIndexBuildValue.WRITE_FAILURE_COUNT;
+import static org.opensearch.knn.plugin.stats.KNNRemoteIndexBuildValue.WRITE_SUCCESS_COUNT;
+import static org.opensearch.knn.plugin.stats.KNNRemoteIndexBuildValue.WRITE_TIME;
```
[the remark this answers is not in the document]
owenhalpert: What if we had a RemoteBuildMetrics class? And removed any stopwatch/metric incrementing from this class? Phase-specific Stopwatch starts would be replaced with metrics.startWritePhase(); and then we would call `metrics.recordWritePhaseMetrics(boolean success, String fieldName)` as normal but the class would keep track of its own stopwatches and just do `writePhaseStopwatch.stop()`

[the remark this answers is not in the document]
owenhalpert: Do you think this refactoring should go in this PR?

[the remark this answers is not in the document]
owenhalpert: @jed326 can you take a look at how I refactored the metric logic out to the `RemoteIndexBuildMetrics` class?

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
```diff
@@ -122,62 +122,148 @@ public static boolean shouldBuildIndexRemotely(IndexSettings indexSettings, long
...
      */
     @Override
     public void buildAndWriteIndex(BuildIndexParams indexInfo) throws IOException {
-        StopWatch stopWatch;
-        long time_in_millis;
+        boolean isFlush = indexInfo.isFlush();
+        KNNVectorValues<?> knnVectorValues = indexInfo.getKnnVectorValuesSupplier().get();
+        initializeVectorValues(knnVectorValues);
+
+        this.metrics = new RemoteIndexBuildMetrics(
```
jed326, line 134: Should we initialize this in the constructor as well?

jed326: Alternatively, we could create this per-method as an `AutoCloseable` to make sure we stop the metrics collection on any errors. I think in the constructor is better thuogh.

jed326 (commented):
I'm a big fan of this `RemoteIndexBuildMetrics` abstraction. Had a small comment on it

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
owenhalpert: I agree putting it in the constructor makes the most sense. I had to do some small refactoring to pass in the `indexInfo` object needed to create the metrics, which I decided was a better alternative than making a generic Metrics object and then setting its parameters later in `buildAndWriteIndex`.

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
```diff
@@ -57,25 +56,37 @@ public class RemoteIndexBuildStrategy implements NativeIndexBuildStrategy {
...
         this.fallbackStrategy = fallbackStrategy;
         this.indexSettings = indexSettings;
         this.knnLibraryIndexingContext = knnLibraryIndexingContext;
+        boolean isFlush = indexInfo.isFlush();
+        KNNVectorValues<?> knnVectorValues = indexInfo.getKnnVectorValuesSupplier().get();
+        initializeVectorValues(knnVectorValues);
+        this.metrics = new RemoteIndexBuildMetrics(
+            (long) indexInfo.getTotalLiveDocs() * knnVectorValues.bytesPerVector(),
+            isFlush,
+            indexInfo.getFieldName()
```
jed326, line 88: Do these need to get passed into the constructor, or can we pass them in during build? I'm not sure if we really need the additional `BuildIndexParams` parameter here.

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
owenhalpert: Edited it to keep the indexInfo as a later parameter to avoid this refactoring

jed326 (approved):

src/main/java/org/opensearch/knn/plugin/transport/KNNStatsRequest.java
```diff
@@ -60,10 +61,15 @@ public KNNStatsRequest(String... nodeIds) {
     }
 
     /**
-     * Add all stats to be retrieved
+     * Add all stats to be retrieved, omitting Remote Build metrics if disabled.
      */
     public void all() {
-        statsToBeRetrieved.addAll(validStats);
+        for (String stat : validStats) {
```
jmazanec15, line 67: Can we set this with valid stats instead of in all? https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/plugin/transport/KNNStatsRequest.java#L35

owenhalpert: Thanks, good point. I made a method `getValidStats` that will add the right stats when this request is created. Also added a test to ensure the stats aren't returned when the feature flag is disabled.

jmazanec15 (dismissed):

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
```diff
@@ -122,62 +124,138 @@ public static boolean shouldBuildIndexRemotely(IndexSettings indexSettings, long
...
-        long time_in_millis;
+        metrics.startRemoteIndexBuildMetrics(indexInfo);
+        try {
+            RepositoryContext repositoryContext = getRepositoryContext(indexInfo);
+
+            // 1. Write required data to repository
+            writeToRepository(repositoryContext, indexInfo);
+
+            // 2. Trigger remote index build
+            RemoteIndexClient client = RemoteIndexClientFactory.getRemoteIndexClient(KNNSettings.getRemoteBuildServiceEndpoint());
```
shatejas, line 140: This client seems to be creating a new object fro each request which is not ideal. The core client is static so we are good there but seems unnecessary to create a new object every request.

shatejas (approved):

shatejas (approved):

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
```diff
@@ -197,6 +275,26 @@ private BlobStoreRepository getRepository() throws RepositoryMissingException {
...
+    }
+
+    /**
+     * Helper method to get repository context. Generates a unique UUID for the blobName so should only be used once.
+     */
+    private RepositoryContext getRepositoryContext(BuildIndexParams indexInfo) {
+        BlobStoreRepository repository = getRepository();
+        BlobPath blobPath = repository.basePath().add(indexSettings.getUUID() + VECTORS_PATH);
+        String blobName = UUIDs.base64UUID() + "_" + indexInfo.getFieldName() + "_" + indexInfo.getSegmentWriteState().segmentInfo.name;
+        VectorRepositoryAccessor vectorRepositoryAccessor = new DefaultVectorRepositoryAccessor(
```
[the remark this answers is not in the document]
owenhalpert: Thanks for pointing this out — good call. I will work on optimizing the object creation in this class in a followup PR

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
owenhalpert: See https://github.com/opensearch-project/k-NN/pull/2615#discussion_r2029502981

jmazanec15 (approved):

PR #2616, opened 2025-03-19
[LuceneOnFaiss Part-8] Added `index.knn.memory_optimized_search` index setting.

0ctopus13prime:
### Description
This PR adds `index.knn.memory_optimized_search` index settings.
Overall, this brings two major changes listed below along with the new setting definition:

1. In codec level, it relies on MapperService to retrieve `index.knn.memory_optimized_search` bool flag. Which then be used to determine whether to initialize partial loading on KNN fields. 
But this is not always working, during recovery MapperService can be null, therefore adding lazy-loading logic in both constructor and search method. For 99%, it will be loaded in the constructor, so it won't affect search p99 latency as by the time the search method is called, MemoryOptimizedSearch should be loaded already.

2. Branching in KNNQueryBuilder
This PR will make each knn field type decide whether memory-optimized-search is supported or not.
Currently only Float HNSW is supported, therefore any binary HNSW graphs will return `false` to indicate the feature is not supported.
If target field in search is supported for memory-optimized-search, control will fallback to Lucene query to proceed vector search. In which, Lucene's HNSW graph searcher will perform KNN search + Radius search on FAISS index.

### Related Issues

RFC: https://github.com/opensearch-project/k-NN/issues/2401

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/KNNSettings.java
+++ b/src/main/java/org/opensearch/knn/index/KNNSettings.java
@@ -5,12 +5,14 @@
 
 package org.opensearch.knn.index;
 
+import lombok.NonNull;
 import lombok.extern.log4j.Log4j2;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.opensearch.OpenSearchParseException;
 import org.opensearch.action.admin.cluster.settings.ClusterUpdateSettingsRequest;
 import org.opensearch.action.admin.cluster.settings.ClusterUpdateSettingsResponse;
+import org.opensearch.knn.index.engine.MemoryOptimizedSearchSupportSpec;
 import org.opensearch.transport.client.Client;
 import org.opensearch.cluster.metadata.IndexMetadata;
 import org.opensearch.cluster.service.ClusterService;
@@ -96,6 +98,11 @@ public class KNNSettings {
     public static final String KNN_DERIVED_SOURCE_ENABLED = "index.knn.derived_source.enabled";
     public static final String KNN_INDEX_REMOTE_VECTOR_BUILD = "index.knn.remote_index_build.enabled";
     public static final String KNN_REMOTE_VECTOR_REPO = "knn.remote_index_build.vector_repo";
+    /**
+     * For more details on supported engines, refer to {@link MemoryOptimizedSearchSupportSpec}
+     */
+    public static final String MEMORY_OPTIMIZED_KNN_SEARCH_MODE = "index.knn.memory_optimized_search";
+    public static final boolean DEFAULT_MEMORY_OPTIMIZED_KNN_SEARCH_MODE = false;
 
     /**
      * Default setting values
@@ -121,9 +128,9 @@ public class KNNSettings {
 
     public static final Integer ADVANCED_FILTERED_EXACT_SEARCH_THRESHOLD_DEFAULT_VALUE = -1;
     public static final Integer KNN_DEFAULT_QUANTIZATION_STATE_CACHE_SIZE_LIMIT_PERCENTAGE = 5; // By default, set aside 5% of the JVM for
-                                                                                                // the limit
+    // the limit
     public static final Integer KNN_MAX_QUANTIZATION_STATE_CACHE_SIZE_LIMIT_PERCENTAGE = 10; // Quantization state cache limit cannot exceed
-                                                                                             // 10% of the JVM heap
+    // 10% of the JVM heap
     public static final Integer KNN_DEFAULT_QUANTIZATION_STATE_CACHE_EXPIRY_TIME_MINUTES = 60;
     public static final boolean KNN_DISK_VECTOR_SHARD_LEVEL_RESCORING_DISABLED_VALUE = false;
 
@@ -284,6 +291,12 @@ public class KNNSettings {
      * This setting identifies KNN index.
      */
     public static final Setting<Boolean> IS_KNN_INDEX_SETTING = Setting.boolSetting(KNN_INDEX, false, IndexScope, Final);
+
+    public static final Setting<Boolean> MEMORY_OPTIMIZED_KNN_SEARCH_MODE_SETTING = Setting.boolSetting(
+        MEMORY_OPTIMIZED_KNN_SEARCH_MODE,
+        false,
+        IndexScope
+    );
 
     /**
      * index_thread_quantity - the parameter specifies how many threads the nms library should use to create the graph.
@@ -577,7 +590,8 @@ public class KNNSettings {
             KNN_DISK_VECTOR_SHARD_LEVEL_RESCORING_DISABLED_SETTING,
             KNN_DERIVED_SOURCE_ENABLED_SETTING,
             KNN_INDEX_REMOTE_VECTOR_BUILD_SETTING,
-            KNN_REMOTE_VECTOR_REPO_SETTING
+            KNN_REMOTE_VECTOR_REPO_SETTING,
+            MEMORY_OPTIMIZED_KNN_SEARCH_MODE_SETTING
         );
         return Stream.concat(settings.stream(), Stream.concat(getFeatureFlags().stream(), dynamicCacheSettings.values().stream()))
             .collect(Collectors.toList());
@@ -734,12 +748,25 @@ public class KNNSettings {
             );
     }
 
+    /**
+     * Return whether memory optimized search enabled for the given index.
+     *
+     * @param indexName The name of target index to test whether if it is on.
+     * @return True if memory optimized search is enabled, otherwise False.
+     */
+    public static boolean isMemoryOptimizedKnnSearchModeEnabled(@NonNull final String indexName) {
+        return KNNSettings.state().clusterService.state()
+            .getMetadata()
+            .index(indexName)
+            .getSettings()
+            .getAsBoolean(MEMORY_OPTIMIZED_KNN_SEARCH_MODE, DEFAULT_MEMORY_OPTIMIZED_KNN_SEARCH_MODE);
+    }
+
     public void setClusterService(ClusterService clusterService) {
         this.clusterService = clusterService;
     }
 
     static class SpaceTypeValidator implements Setting.Validator<String> {
-
         @Override
         public void validate(String value) {
             try {
--- a/src/main/java/org/opensearch/knn/index/codec/BasePerFieldKnnVectorsFormat.java
+++ b/src/main/java/org/opensearch/knn/index/codec/BasePerFieldKnnVectorsFormat.java
@@ -155,11 +155,12 @@ public abstract class BasePerFieldKnnVectorsFormat extends PerFieldKnnVectorsFor
     private NativeEngines990KnnVectorsFormat nativeEngineVectorsFormat() {
         // mapperService is already checked for null or valid instance type at caller, hence we don't need
         // addition isPresent check here.
-        int approximateThreshold = getApproximateThresholdValue();
+        final int approximateThreshold = getApproximateThresholdValue();
         return new NativeEngines990KnnVectorsFormat(
             new Lucene99FlatVectorsFormat(FlatVectorScorerUtil.getLucene99FlatVectorsScorer()),
             approximateThreshold,
-            nativeIndexBuildStrategyFactory
+            nativeIndexBuildStrategyFactory,
+            mapperService
         );
     }
 
--- a/src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsFormat.java
+++ b/src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsFormat.java
@@ -11,6 +11,7 @@
 
 package org.opensearch.knn.index.codec.KNN990Codec;
 
+import lombok.extern.log4j.Log4j2;
 import org.apache.lucene.codecs.KnnVectorsFormat;
 import org.apache.lucene.codecs.KnnVectorsReader;
 import org.apache.lucene.codecs.KnnVectorsWriter;
@@ -19,22 +20,31 @@ import org.apache.lucene.codecs.hnsw.FlatVectorsFormat;
 import org.apache.lucene.codecs.lucene99.Lucene99FlatVectorsFormat;
 import org.apache.lucene.index.SegmentReadState;
 import org.apache.lucene.index.SegmentWriteState;
+import org.opensearch.common.settings.Settings;
+import org.opensearch.index.IndexSettings;
+import org.opensearch.index.mapper.MapperService;
 import org.opensearch.knn.index.KNNSettings;
 import org.opensearch.knn.index.codec.nativeindex.NativeIndexBuildStrategyFactory;
 import org.opensearch.knn.index.engine.KNNEngine;
 
 import java.io.IOException;
+import java.util.Optional;
+
+import static org.opensearch.knn.index.KNNSettings.DEFAULT_MEMORY_OPTIMIZED_KNN_SEARCH_MODE;
+import static org.opensearch.knn.index.KNNSettings.MEMORY_OPTIMIZED_KNN_SEARCH_MODE;
 
 /**
  * This is a Vector format that will be used for Native engines like Faiss and Nmslib for reading and writing vector
  * related data structures.
  */
+@Log4j2
 public class NativeEngines990KnnVectorsFormat extends KnnVectorsFormat {
     /** The format for storing, reading, merging vectors on disk */
     private static FlatVectorsFormat flatVectorsFormat;
     private static final String FORMAT_NAME = "NativeEngines990KnnVectorsFormat";
     private static int approximateThreshold;
     private final NativeIndexBuildStrategyFactory nativeIndexBuildStrategyFactory;
+    private final boolean memoryOptimizedSearchEnabled;
 
     public NativeEngines990KnnVectorsFormat() {
         this(new Lucene99FlatVectorsFormat(new DefaultFlatVectorScorer()));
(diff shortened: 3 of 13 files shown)


src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
```diff
@@ -69,11 +70,8 @@ public NativeEngines990KnnVectorsReader(final SegmentReadState state, final Flat
         this.vectorSearchers = new HashMap<>(RESERVE_TWICE_SPACE * segmentReadState.fieldInfos.size(), SUFFICIENT_LOAD_FACTOR);
         loadCacheKeyMap();
 
-        //
-        // TMP(KDY) : Dynamic update will be covered in part-7. Please refer to
-        // https://github.com/opensearch-project/k-NN/issues/2401#issuecomment-2699777824
-        //
-        final boolean isMemoryOptimizedSearchEnabled = false;
+        final String indexName = KNNCodecUtil.extractIndexNameFromFieldInfos(state.fieldInfos, log);
+        final boolean isMemoryOptimizedSearchEnabled = indexName != null && KNNSettings.isMemoryOptimizedKnnSearchModeEnabled(indexName);
```
navneet1v, line 74: Having an index name is not a good design choice here. Since index name is the construct of OpenSearch and serializing this information on disk with segment information is just going to create a lot of issues in future. I won't suggest doing this and lets try to find out better way for this. 

looking at the use-case it seems like we need this information at start of the segment initialization to read faiss file into memory. I would suggest either doing this via warm up api or take the hit of first time load of file during first search query.

navneet1v: Another question while running this `loadMemoryOptimizedSearcher` are you reading full file and keeping some pointers in memory?

navneet1v (changes requested):

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
0ctopus13prime: Hi @navneet1v 
So far, I could not find a way to get an index name in codec level. I'm not strongly leaning forward, but alternatively we could always partial load FAISS index.
Or we could do the lazy search. I think lazy loading does make sense in this case.

For the restored case, I don't think it will cause the issue as it's the field mapper puts the index name into segment when index is being created (including dynamic template). And those persisted info will never be changed. 

Q : Another question while running this loadMemoryOptimizedSearcher are you reading full file and keeping some pointers in memory?
No, it will read few meta infos then skip to the next section. 
For example, HNSW graph, it will mark the starting offsets and read few meta info, entry pointer etc, then skip to the Flat vector section. In there, it also mark the starting offset then return. There will be no full-reading.

navneet1v: > For the restored case, I don't think it will cause the issue as it's the field mapper puts the index name into segment when index is being created (including dynamic template). And those persisted info will never be changed.

No, field mapper doesn't put the index name in the segment when an index is created. It only during adding of documents. FYI when you restore an index from snapshot, documents are not ingested again. The same segment files are directly copied.

0ctopus13prime: Ah.. I see... thank you for the correction.
Then will switch to the lazy loading! I think that will reduce the amount of lines as well.

0ctopus13prime:
Part-7 https://github.com/opensearch-project/k-NN/pull/2608 was merged.
Will rebase and re-raise a new revision.
In the new revision, it will have unit tests.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
0ctopus13prime: Will try a hybrid approach to get IndexSettings sourced into the reader + Lazy loading.

0ctopus13prime:
Hi @jmazanec15 ,  

In this revision, I've included a hybrid approach inspired by @shatejas 's suggestion.  

Previously, I added lazy loading in search because accessing the `index.knn.memory_optimized_search` setting in the constructor seemed tricky. Retrieving this value requires knowing the index name, and more importantly, during recovery, the codec is loaded via Lucene's SPI, where OpenSearch's internal framework isn't available. That was the main reason for switching to lazy loading within search.  

However, this scenario is rare. In most cases, OpenSearch's internal framework (e.g., `IndexSettings`) is available. Additionally, since we're trying to compress `long[]` on the fly during loading in #2609 , relying solely on lazy loading in search could negatively impact p99, as it might take several seconds for larger datasets.  

With this hybrid approach, we attempt to initialize in the constructor. If `IndexSettings` is available, we fetch the boolean value; otherwise, we do nothing. This way, most searches will find the table already initialized, no need for lazy loading. Only searches on a just-recovered index will require it.  

Let me know if you have any major concerns. I think this approach not only mitigates potential p99 issues but also avoids the Lucene SPI limitation.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
```diff
@@ -350,4 +287,76 @@ private static List<String> getVectorCacheKeysFromSegmentReaderState(SegmentRead
...
+                    IOUtils.closeWhileHandlingException(vectorSearcherPerField.values());
+                } catch (Exception closeException) {
+                    log.error(closeException.getMessage(), closeException);
+                }
+                throw new RuntimeException(e);
+            }
+        }
+    }
+
+    private IOSupplier<VectorSearcher> getIndexFileNameIfMemoryOptimizedSearchSupported(final FieldInfo fieldInfo) {
```
jmazanec15, line 335: nit: name of this method looks wrong - I would expect it to return a file name not a vector searcher. Can you fix?

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
```diff
@@ -350,4 +287,76 @@ private static List<String> getVectorCacheKeysFromSegmentReaderState(SegmentRead
...
+        // Skip non-knn fields.
+        final Map<String, String> attributes = fieldInfo.attributes();
+        if (attributes == null || attributes.containsKey(KNN_FIELD) == false) {
+            return null;
+        }
+
+        // Get engine
+        final String engineName = attributes.getOrDefault(KNN_ENGINE, KNNEngine.DEFAULT.getName());
```
jmazanec15, line 343: I think this will break with models - do we skip this feature if its a model based index?

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/MemoryOptimizedSearchSupportSpec.java
```diff
@@ -0,0 +1,100 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.knn.memoryoptsearch.faiss;
```
jmazanec15, line 6: This seems like it should not be in the faiss package if its generic?

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/MemoryOptimizedSearchSupportSpec.java
```diff
@@ -0,0 +1,100 @@
...
+                return true;
+            }
+
+            // We don't support non-FAISS engine
+            if (engine != KNNEngine.FAISS) {
+                return false;
+            }
+
+            // We only support HNSW method.
+            final MethodComponentContext methodComponentContext = methodContext.getMethodComponentContext();
```
jmazanec15, line 64: IMO, this logic needs to go in engine package. The purpose of the engine package is to avoid messy branching like this and keep engine specific logic in one location. The goal is to make engine extensible and I dont think this goes in the right direction.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
0ctopus13prime: hoping to learn more, why it will break for models based index? If the resulting index is what memory_optimized_search supports, then it will be loaded.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/MemoryOptimizedSearchSupportSpec.java
0ctopus13prime: sure, will move it to src/main/java/org/opensearch/knn/memoryoptsearch pacakge

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/MemoryOptimizedSearchSupportSpec.java
0ctopus13prime: will update it in next rev

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
jmazanec15: Concern is that engine name is not stored in field attributes.

src/main/java/org/opensearch/knn/index/codec/BasePerFieldKnnVectorsFormat.java
```diff
@@ -155,14 +158,30 @@ public KnnVectorsFormat getKnnVectorsFormatForField(final String field) {
...
             new Lucene99FlatVectorsFormat(FlatVectorScorerUtil.getLucene99FlatVectorsScorer()),
             approximateThreshold,
-            nativeIndexBuildStrategyFactory
+            nativeIndexBuildStrategyFactory,
+            memoryOptimizedSearchSupported
         );
     }
 
+    private boolean isMemoryOptimizedSearchSupported() {
+        if (nativeIndexBuildStrategyFactory != null && nativeIndexBuildStrategyFactory.getIndexSettings() != null) {
```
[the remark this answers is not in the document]
shatejas: Alternatively you can use mapperservice to get index settings. https://github.com/opensearch-project/k-NN/pull/2616/files#diff-60cad16fb7fef0302c7ca18486b1be7a366fd53f49511ae4674042c775edd9c8R188

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsFormat.java
```diff
@@ -85,7 +88,7 @@ public KnnVectorsWriter fieldsWriter(final SegmentWriteState state) throws IOExc
      */
     @Override
     public KnnVectorsReader fieldsReader(final SegmentReadState state) throws IOException {
-        return new NativeEngines990KnnVectorsReader(state, flatVectorsFormat.fieldsReader(state));
+        return new NativeEngines990KnnVectorsReader(state, flatVectorsFormat.fieldsReader(state), memoryOptimizedSearchEnabled);
```
shatejas, line 91: nit: We might be able to pass index settings all the way to the constructor to avoid parameter explosion related to index settings in future

src/main/java/org/opensearch/knn/index/KNNSettings.java
```diff
@@ -96,6 +97,8 @@ public class KNNSettings {
     public static final String KNN_DERIVED_SOURCE_ENABLED = "index.knn.derived_source.enabled";
     public static final String KNN_INDEX_REMOTE_VECTOR_BUILD = "index.knn.remote_index_build.enabled";
     public static final String KNN_REMOTE_VECTOR_REPO = "knn.remote_index_build.vector_repo";
+    public static final String MEMORY_OPTIMIZED_KNN_SEARCH_MODE = "index.knn.memory_optimized_search";
```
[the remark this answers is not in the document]
0ctopus13prime: will update in the next rev

src/main/java/org/opensearch/knn/index/codec/BasePerFieldKnnVectorsFormat.java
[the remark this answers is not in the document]
0ctopus13prime: yeah, mapper service looks good.
will update in the next rev

src/main/java/org/opensearch/knn/index/codec/BasePerFieldKnnVectorsFormat.java
```diff
@@ -155,14 +158,30 @@ public KnnVectorsFormat getKnnVectorsFormatForField(final String field) {
...
+            final Settings settings = nativeIndexBuildStrategyFactory.getIndexSettings().getSettings();
+            if (settings != null) {
+                try {
+                    return settings.getAsBoolean(MEMORY_OPTIMIZED_KNN_SEARCH_MODE, DEFAULT_MEMORY_OPTIMIZED_KNN_SEARCH_MODE);
```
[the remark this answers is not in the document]
0ctopus13prime: sure, will see that works.
will update in the next rev

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
0ctopus13prime: thank you, will fix it

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
```diff
@@ -350,4 +291,95 @@ private static List<String> getVectorCacheKeysFromSegmentReaderState(SegmentRead
...
+                return null;
+            }
+            knnEngine = modelMetadata.getKnnEngine();
+        }
+
+        // If knnEngine is still null, then try to get it with KNN_ENGINE
+        if (knnEngine == null) {
+            final String engineName = attributes.getOrDefault(KNN_ENGINE, KNNEngine.DEFAULT.getName());
+            knnEngine = KNNEngine.getEngine(engineName);
+        }
```
shatejas, line 362: Can we use [FieldInfoExtractor](https://github.com/opensearch-project/k-NN/blob/294b4c7daf8fa498e66d96f63ac569080aa9d719/src/main/java/org/opensearch/knn/common/FieldInfoExtractor.java#L42)?

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
```diff
@@ -621,8 +638,15 @@ private float[] getVectorForCreatingQueryRequest(VectorDataType vectorDataType,
...
-        if (VectorDataType.BINARY == vectorDataType || (VectorDataType.BYTE == vectorDataType && KNNEngine.LUCENE == knnEngine)) {
+    private byte[] getVectorForCreatingQueryRequest(
+        VectorDataType vectorDataType,
+        KNNEngine knnEngine,
+        byte[] byteVector,
+        boolean memoryOptimizedSearchSupported
+    ) {
+
+        if (VectorDataType.BINARY == vectorDataType
+            || (VectorDataType.BYTE == vectorDataType && isUsingLuceneQuery(knnEngine, memoryOptimizedSearchSupported))) {
```
shatejas, line 649: why is `memoryOptimizedSearchSupported` being considered here?

src/test/java/org/opensearch/knn/index/query/KNNQueryBuilderTests.java
```diff
@@ -645,6 +651,77 @@ public void testDoToQuery_whenknnQueryWithFilterAndNmsLibEngine_thenException()
         expectThrows(IllegalArgumentException.class, () -> knnQueryBuilder.doToQuery(mockQueryShardContext));
     }
 
+    public void testDoToQuery_whenMemoryOptimizedSearchIsEnabled() {
+        for (final boolean memoryOptimizedSearchEnabled : new boolean[] { true, false }) {
+            for (final boolean memoryOptimizedSearchSupportedInField : new boolean[] { true, false }) {
+                for (final VectorDataType vectorDataType : Arrays.asList(VectorDataType.FLOAT, VectorDataType.BYTE)) {
```
shatejas, line 657: nit: This is going to be complicated to debug when it fails. I would prefer a parametrized test or a separate test for which combo calling a private test method so its clear what case failed

shatejas (commented):
Looks good overall - couple of comments

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
0ctopus13prime: Since query is always `float[]` for FAISS engine and `byte[]` is always null. This led unexpected result when memory optimized search is enabled for FAISS Byte index. Hence we need consider this case to have non-null `byte[] to pass `MemoryOptimizedSearcher`.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
0ctopus13prime: will update in the next rev.

src/test/java/org/opensearch/knn/index/query/KNNQueryBuilderTests.java
0ctopus13prime: will update in the next rev

jmazanec15 (approved):

shatejas (approved):

Issue #2553, opened 2025-02-23
[META] [Remote Vector Index Build] Integration Testing Support


jed326:
Meta: https://github.com/opensearch-project/k-NN/issues/2391

The remote index build feature takes a dependency on an external component to perform the vector index build operation. This makes it so that spinning up an OpenSearch cluster by itself is not sufficient to run integration tests and we also need a build service running in order to run all of the tests.

Once a remote index build service is available in https://github.com/opensearch-project/remote-vector-index-builder, we would like to be able to spin up both an OpenSearch cluster as well as a remote index build service within the same github action in order to run tests.

### Tasks:
- [x] Add a basic sanity check test against https://github.com/opensearch-project/remote-vector-index-builder: https://github.com/opensearch-project/k-NN/pull/2620
- [ ] Update developer guide on how to set up these tests to run manually
- [ ] Update IT to pull image from opensearch-staging instead of personal dockerhub
- [ ] Resolve test based issues in https://github.com/opensearch-project/remote-vector-index-builder/issues
    - [x] https://github.com/opensearch-project/remote-vector-index-builder/issues/30
    - [x] https://github.com/opensearch-project/remote-vector-index-builder/issues/31
    - [x] https://github.com/opensearch-project/remote-vector-index-builder/issues/32
    - [x] https://github.com/opensearch-project/remote-vector-index-builder/issues/33
    - [x] https://github.com/opensearch-project/remote-vector-index-builder/issues/34
    - [x] https://github.com/opensearch-project/remote-vector-index-builder/issues/35
    - [x] https://github.com/opensearch-project/remote-vector-index-builder/issues/40
- [ ] Once test issues are resolved, converge the tests to run **ALL** tests against remote vector index builder
- [ ] Run k-NN test with remote index build service as a part of new version release

anntians:
I would like to work on this task. Will provide more details about the github action for IT as well as decisions made. Thanks

PR #2620, opened 2025-03-19
Add github action to run ITs against remote index builder

anntians:
### Description
Add github action to run ITs against remote index builder

For sanity check we're running remote index builder against 3 fp_32 test cases (for all spaceTypes: L2, IP, Cosine):
1. End to end radial search test
  - Indexing -> refresh -> radial search -> force merge -> radial search
2. Filtering search test
  - Indexing -> refresh -> force merge -> filtered search
3. KNN search test
  - Indexing -> refresh -> force merge -> KNN unfiltered search 

### Related Issues
Resolves first task of https://github.com/opensearch-project/k-NN/issues/2553 to add a basic sanity check test against https://github.com/opensearch-project/remote-vector-index-builder

### Check List


build.gradle
```diff
@@ -445,6 +469,62 @@ integTest {
     systemProperty propertyKeys.breaker.useRealMemory, getBreakerSetting()
 }
 
+task integTestRemoteIndexBuild(type: RestIntegTestTask) {
```
navneet1v, line 472: Please try to reuse/refactor the other gradle tasks to move the common things at 1 place rather than copying everything.

build.gradle
```diff
@@ -45,6 +45,8 @@ buildscript {
             opensearch_build += "-SNAPSHOT"
         }
         opensearch_no_snapshot = opensearch_build.replace("-SNAPSHOT","")
+        repo_s3_resource_folder = "build/resource/repository-s3"
+        repo_s3_download_url = "https://ci.opensearch.org/ci/dbc/distribution-build-opensearch/3.0.0-alpha1/latest/linux/arm64/tar/builds/opensearch/core-plugins/repository-s3-3.0.0-alpha1.zip"
```
navneet1v, line 49: Please use relative links to build this URL. This includes versions of OpenSearch and also the distribution URL. Take some reference from here: https://github.com/opensearch-project/k-NN/blob/main/qa/build.gradle#L67-L74

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ Add filter function to KNNQueryBuilder with unit tests and integration tests [#2
 ### Bug Fixes
 * Fixing bug to prevent NullPointerException while doing PUT mappings [#2556](https://github.com/opensearch-project/k-NN/issues/2556)
 ### Infrastructure
+Add github action to run ITs against remote index builder [2620](https://github.com/opensearch-project/k-NN/pull/2620)
```
navneet1v, line 16:
```suggestion
* Add github action to run ITs against remote index builder [2620](https://github.com/opensearch-project/k-NN/pull/2620)
```

.github/workflows/remote_index_build.yml
```diff
@@ -0,0 +1,141 @@
...
+          sudo yum install gcc-gfortran -y
+
+      - name: Initial cleanup
+        run: |
+          docker ps -aq | xargs -r docker rm -f
+          docker system prune -af --volumes
+
+      - name: Pull Remote Index Build Docker Image from Docker Hub
+        run: |
+          docker pull rchitale7/remote-index-build-service:api
```
kotwanikunal, line 78: Can we move this to the opensearch-project repository?

anntians: We will eventually move this image to the opensearch-project dockerhub, but the remote index builder code hasn't been merged to main yet, so we are using this image for testing purposes on this branch

src/test/java/org/opensearch/knn/index/RemoteBuildIT.java
```diff
@@ -0,0 +1,355 @@
...
+    protected Settings buildKNNIndexSettingsRemoteBuild(int approximateThreshold) {
+        Settings.Builder builder = Settings.builder()
+            .put("number_of_shards", 1)
+            .put("number_of_replicas", 0)
+            .put(KNN_INDEX, true)
+            .put(INDEX_KNN_ADVANCED_APPROXIMATE_THRESHOLD, approximateThreshold);
+
+        final String remoteBuild = System.getProperty("test.remoteBuild", null);
+
+        if (isRemoteIndexBuildSupported(getBWCVersion()) && remoteBuild != null) {
```
jed326, line 349: We shouldn't need this if we are only running this in the remote build task right? We should exclude this class from the default integ test

jed326 (commented):
Thanks @anntians. Are we planning on pointing this PR at `main`?

.github/workflows/remote_index_build.yml
```diff
@@ -0,0 +1,141 @@
...
+      - 'jni/**'
+      - '.github/workflows/CI.yml'
+  pull_request:
+    branches:
+      - "*"
+      - "feature/**"
+    paths:
+      - 'build.gradle'
+      - 'settings.gradle'
+      - 'src/**'
```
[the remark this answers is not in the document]
anntians: The BWC GA https://github.com/opensearch-project/k-NN/blob/main/.github/workflows/backwards_compatibility_tests_workflow.yml#L8-L16 also triggers on a wide range of paths. I think remote build GA should follow the same pattern because remote build related changes/classes are pretty scattered throughout the code, so we wouldn't want to miss triggering this GA on changes.

Thoughts? @navneet1v

anntians:
@jed326  Yes, since the new GA is passing, we could point this PR to main. But will need to update the branches of the GA to
```
      - "*"
      - "feature/**"
```

jed326:
> @jed326 Yes, since the new GA is passing, we could point this PR to main. But will need to update the branches of the GA to
> 
> ```
>       - "*"
>       - "feature/**"
> ```

So what's the plan for running CI on PRs to `main` then?

anntians:
Currently, the same branches and paths will trigger the GA for both push and PRs https://github.com/anntians/k-NN/blob/feature/github-ci-remote-index-build/.github/workflows/remote_index_build.yml#L16-L28

jed326:
My question is, what's the plan for getting `.github/workflows/remote_index_build.yml` merged to `main? Are you merging it to the feature branch first, and then you'll merge it to main later? Or do you want to just do that now.

.github/workflows/remote_index_build.yml
```diff
@@ -0,0 +1,140 @@
...
+      - 'build-tools/**'
+      - 'buildSrc/**'
+      - 'gradle/**'
+      - 'jni/**'
+      - '.github/workflows/remote_index_build.yml'
+  pull_request:
+    branches:
+      - "*"
+      - "feature/**"
```
jed326, line 19: If you're pointing this at `main` then you don't need this right?

.github/workflows/remote_index_build.yml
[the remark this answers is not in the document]
jed326: I'm inclined to scope it down for now and make it run for all PRs later, wdyt?

.github/workflows/remote_index_build.yml
```diff
@@ -0,0 +1,140 @@
...
+            echo "No success logs found. Full logs:"
+            docker logs remote-index-builder-container
+            exit 1
+          fi
+
+      - name: Final cleanup
+        if: always()
+        run: |
+          docker ps -aq | xargs -r docker rm -f
+          docker system prune -af --volumes
```
peterzhuamazon, line 139:
```suggestion
          docker ps -aq | xargs -r docker rm -f
          docker system prune -af --volumes
          docker logout
          rm -rf ${{ github.workspace }}/*
```

.github/workflows/remote_index_build.yml
```diff
@@ -0,0 +1,140 @@
...
+          distribution: 'temurin'
+
+      - name: Install dependencies on linux
+        run: |
+          sudo yum install gcc g++ -y
+          sudo yum install openblas openblas-devel -y
+          sudo yum install -y zlib
+          sudo yum install -y zlib-devel
+          sudo yum install -y cmake
+          sudo yum install gcc-gfortran -y
```
peterzhuamazon, line 69: Note that this is not a self-cleaned runner so if you sure this is the only thing you need for installation, you stick with it or else it is very hard to cleanup if you keep installing more.

build.gradle
```diff
@@ -17,6 +17,7 @@ buildscript {
         // explicitly as 'opensearch.version' property, for instance opensearch.version=2.0.0-rc1-SNAPSHOT
         opensearch_version = System.getProperty("opensearch.version", "3.0.0-alpha1-SNAPSHOT")
         version_qualifier = System.getProperty("build.version_qualifier", "alpha1")
+        arch = System.getProperty("arch", "arm64")
```
peterzhuamazon, line 20: Why do we need to set arch to arm64 hardcoded.

peterzhuamazon (commented):
I can only comment on a few things I have knowledge of.
Please check.

Thanks.

.github/workflows/remote_index_build.yml
```diff
@@ -0,0 +1,140 @@
...
+      - 'buildSrc/**'
+      - 'gradle/**'
+      - 'jni/**'
+      - '.github/workflows/remote_index_build.yml'
+
+jobs:
+  Remote-Index-Build-IT-Tests:
+    strategy:
+      matrix:
+        java: [21, 23]
```
peterzhuamazon, line 34: Leave this to 21 only now as 21+23 consumed too much runners and main is still built with 21 now.

Keep 23 for other workflows for compatibility purposes.

THanks.

.github/workflows/remote_index_build.yml
anntians: If we only want this GA to run when we have PRs targeting `main` then we won't need to specify branches here and default behavior is to only run GA when PR into `main`. We can limit the branches to just `main` for now, but eventually we would want to use the current configuration so that the GA runs for PRs merging into any branch. Similar to BWC GA: https://github.com/opensearch-project/k-NN/blob/main/.github/workflows/backwards_compatibility_tests_workflow.yml#L4-L6

.github/workflows/remote_index_build.yml
anntians: I tried following the BWC GA with 
```
sudo apt-get install libopenblas-dev gfortran -y
```
at https://github.com/opensearch-project/k-NN/blob/main/.github/workflows/backwards_compatibility_tests_workflow.yml#L73. But I get error
```
sudo: apt-get: command not found
```
since I don't think the AMI used to create the GPU runners come with `apt` installed. 

So that's why I'm using the commands I ran when setting up my testing `g5.4xlarge EC2 instance, with Deep Learning Base OSS Nvidia Driver GPU AMI (Amazon Linux 2023) AMI`, which works

anntians:
@jed326 Sure we can scope branches down to `main` and paths to `build.gradle`, `src/**`, and `.github/workflows/remote_index_build.yml`?

jed326:
How about all of these places for now?
- https://github.com/opensearch-project/k-NN/tree/main/remote-index-build-client/src
- https://github.com/opensearch-project/k-NN/tree/main/src/test/java/org/opensearch/knn/index/remote
- https://github.com/opensearch-project/k-NN/tree/main/src/main/java/org/opensearch/knn/index/remote
- https://github.com/opensearch-project/k-NN/tree/main/src/main/java/org/opensearch/knn/index/codec/nativeindex/remote
- https://github.com/opensearch-project/k-NN/tree/main/src/test/java/org/opensearch/knn/index/codec/nativeindex/remote

I'm open to input on this from some other kNN maintainers though

.github/workflows/remote_index_build.yml
jed326: Doesn't `*` already match `feature/**` though? Shouldn't it be `main` and `feature/**` to run on `main` and all `feature` branches?

navneet1v:
> How about all of these places for now?
> 
> * https://github.com/opensearch-project/k-NN/tree/main/remote-index-build-client
> * https://github.com/opensearch-project/k-NN/tree/main/src/test/java/org/opensearch/knn/index/remote
> * https://github.com/opensearch-project/k-NN/tree/main/src/main/java/org/opensearch/knn/index/remote
> * https://github.com/opensearch-project/k-NN/tree/main/src/main/java/org/opensearch/knn/index/codec/nativeindex/remote
> * https://github.com/opensearch-project/k-NN/tree/main/src/test/java/org/opensearch/knn/index/codec/nativeindex/remote
> 
> I'm open to input on this from some other kNN maintainers though

I think this is more specific. We should just go with whatever is there in the other CIs. Keeping the runs for specific package path can lead to other changes making tests fail. So I won't suggest this.

build.gradle
anntians: We need to set a default value here as I don't believe the system properties like `opensearch_version` and `arch` are set when we run `integTestRemoteIndexBuild` gradle task. Both `arm64` and `x64` distributions work, so I defaulted to `arm64`

.github/workflows/remote_index_build.yml
```diff
@@ -0,0 +1,144 @@
+name: Build and Test k-NN using Remote Index Builder
+on:
+  schedule:
+    - cron: '0 0 * * *'  # every night
```
jed326, line 4: Does this run on every branch every night?

anntians: No, scheduled workflows run on the latest commit on the default or base branch. So will only run every night on main branch.

jed326 (approved):

build.gradle
navneet1v: @anntians why we need this line? this seems counterintutive to have in the build.gradle file

build.gradle
```diff
@@ -45,6 +46,9 @@ buildscript {
             opensearch_build += "-SNAPSHOT"
         }
         opensearch_no_snapshot = opensearch_build.replace("-SNAPSHOT","")
+        opensearch_version_no_snapshot = opensearch_version.replace("-SNAPSHOT","")
```
navneet1v, line 49: we already have `opensearch_no_snapshot`, why we need `opensearch_version_no_snapshot`

scripts/build.sh
```diff
@@ -129,7 +129,7 @@ fi
 
 # Build k-NN lib and plugin through gradle tasks
 cd $work_dir
-./gradlew build --no-daemon --refresh-dependencies -x integTest -x test -Dopensearch.version=$VERSION -Dbuild.snapshot=$SNAPSHOT -Dbuild.version_qualifier=$QUALIFIER -Dbuild.lib.commit_patches=false
+./gradlew build --no-daemon --refresh-dependencies -x integTest -x test -Dopensearch.version=$VERSION -Darch=$ARCHITECTURE -Dbuild.snapshot=$SNAPSHOT -Dbuild.version_qualifier=$QUALIFIER -Dbuild.lib.commit_patches=false
```
navneet1v, line 132: why we need `arch` during the build?

src/test/java/org/opensearch/knn/index/RemoteBuildIT.java
```diff
@@ -0,0 +1,347 @@
...
+        assert testIndexVectors != null;
+        assert testQueries != null;
+        testData = new TestUtils.TestData(testIndexVectors.getPath(), testQueries.getPath());
+    }
+
+    @Before
+    public void setupAdditionalRemoteIndexBuildSettings() throws Exception {
+        updateClusterSettings(KNNFeatureFlags.KNN_REMOTE_VECTOR_BUILD_SETTING.getKey(), true);
+        updateClusterSettings(KNNSettings.KNN_REMOTE_VECTOR_REPO, "integ-test-repo");
+        updateClusterSettings(KNNSettings.KNN_REMOTE_BUILD_SERVICE_ENDPOINT, "http://0.0.0.0:80");
```
navneet1v, line 81: can we update this setting via `integTestRemoteIndexBuild`

src/test/java/org/opensearch/knn/index/RemoteBuildIT.java
```diff
@@ -0,0 +1,347 @@
...
+
+        // Delete index
+        deleteKNNIndex(INDEX_NAME);
+    }
+
+    @SneakyThrows
+    public void testHNSWSQFP16_whenGraphThresholdIsMetDuringMerge_thenCreateGraph() {
+        final String indexName = "test-index-hnsw-sqfp16";
+        final String fieldName = "test-field-hnsw-sqfp16";
+        final SpaceType[] spaceTypes = { SpaceType.L2, SpaceType.INNER_PRODUCT };
```
navneet1v, line 150: for fp16, we should be using the local index build. Why for fp16 we are going to remote index build?

src/test/java/org/opensearch/knn/index/RemoteBuildIT.java
```diff
@@ -0,0 +1,347 @@
...
+import static org.opensearch.knn.common.KNNConstants.METHOD_PARAMETER_EF_CONSTRUCTION;
+import static org.opensearch.knn.common.KNNConstants.METHOD_PARAMETER_M;
+import static org.opensearch.knn.common.KNNConstants.METHOD_PARAMETER_SPACE_TYPE;
+import static org.opensearch.knn.common.KNNConstants.MIN_SCORE;
+import static org.opensearch.knn.common.KNNConstants.NAME;
+import static org.opensearch.knn.common.KNNConstants.PARAMETERS;
+import static org.opensearch.knn.index.KNNSettings.*;
+import static org.opensearch.knn.index.KNNSettings.KNN_INDEX_REMOTE_VECTOR_BUILD_THRESHOLD;
+
+public class RemoteBuildIT extends KNNRestTestCase {
```
navneet1v, line 62: Where is the test an index of fp32 getting created via remote index build and then getting searched?

build.gradle
anntians: `opensearch_no_snapshot` has an extra `.0`, so it looks like `s3-3.0.0.0-beta1` instead of `s3-3.0.0-beta1`. And the s3 download url only works for `s3-3.0.0-beta1`

src/test/java/org/opensearch/knn/index/RemoteBuildIT.java
anntians: There is a way to set cluster settings in `build.gradle`, but I believe each IT cleans up cluster settings after they're run, so we need to set cluster settings before each test is run

build.gradle
peterzhuamazon: native plugin aka repository-s3 or similar is from core, unlike individual repo plugins that follow 3 digit from core, instead of 4digits.

src/test/java/org/opensearch/knn/index/RemoteBuildIT.java
```diff
@@ -0,0 +1,337 @@
...
+    public void setupAdditionalRemoteIndexBuildSettings() throws Exception {
+        updateClusterSettings(KNNFeatureFlags.KNN_REMOTE_VECTOR_BUILD_SETTING.getKey(), true);
+        updateClusterSettings(KNNSettings.KNN_REMOTE_VECTOR_REPO, "integ-test-repo");
+        updateClusterSettings(KNNSettings.KNN_REMOTE_BUILD_SERVICE_ENDPOINT, "http://0.0.0.0:80");
+        updateClusterSettings(KNNSettings.KNN_REMOTE_BUILD_CLIENT_POLL_INTERVAL, 0);
+        setupRepository("integ-test-repo");
+    }
+
+    @SneakyThrows
+    public void testEndToEnd_whenDoRadiusSearch_whenDistanceThreshold_whenMethodIsHNSWFlat_thenSucceed() {
```
navneet1v, line 83: can we have the test here for all the different SpaceTypes. L2, IP and Consine.

src/test/java/org/opensearch/knn/index/RemoteBuildIT.java
```diff
@@ -141,11 +157,11 @@ public void testEndToEnd_whenDoRadiusSearch_whenDistanceThreshold_whenMethodIsHN
 
     @SneakyThrows
     public void testHNSW_whenGraphThresholdIsMetDuringMerge_thenCreateGraph() {
+        if (spaceType == SpaceType.COSINESIMIL) {
```
anntians, line 160: This test was copied from FaissIT for the purpose of this sanity check PR. However, the test fails for `SpaceType.COSINESIMIL`, so for the sanity check we will be skipping this case for now. The purpose of the sanity check is to ensure basic indexing and search functionality, and the `COSINESIMIL` is covered by the other test above.

jed326: 1. Please use `Assume.assumeFalse`
2. Why does this fail for `COSINESIMIL`? If it's covered by the test above then why even have this test at all?

jed326 (approved):

jed326 (approved):

navneet1v (approved):

.github/workflows/remote_index_build.yml
```diff
@@ -0,0 +1,144 @@
...
+      - name: Run Docker container
+        run: |
+          docker run --rm -d --name remote-index-builder-container --gpus all -p 80:80 -e INTEGRATION_TESTS=TRUE -e AWS_ACCESS_KEY_ID=${{ env.AWS_ACCESS_KEY_ID }} -e AWS_SECRET_ACCESS_KEY=${{ env.AWS_SECRET_ACCESS_KEY }} -e AWS_SESSION_TOKEN=${{ env.AWS_SESSION_TOKEN}} rchitale7/remote-index-build-service:api
+          sleep 5
+
+      - name: Run tests
+        run: |
+          if lscpu  | grep -i avx512f | grep -i avx512cd | grep -i avx512vl | grep -i avx512dq | grep -i avx512bw
+          then
+            if lscpu | grep -q "GenuineIntel" && lscpu | grep -i avx512_fp16 | grep -i avx512_bf16 | grep -i avx512_vpopcntdq
```
jmazanec15, line 110: Why do we need to worry about architecture here? I didnt think remoteIndexBuild should care about this

anntians: I believe checks for architecture is for the KNN side for the case if the underlying hardware runners change/update. For instance, the current runners don't support `avx512` so we need to run with `-Davx512.enabled=false -Davx512_spr.enabled=false` to build avx2 version of Faiss library for KNN to run.

The GPU remote builder doesn't change with different architecture of the runners.

This architecture check is also present in other existing GAs like https://github.com/opensearch-project/k-NN/blob/main/.github/workflows/CI.yml#L77 and https://github.com/opensearch-project/k-NN/blob/main/.github/workflows/backwards_compatibility_tests_workflow.yml#L106

jmazanec15: I see - makes sense.

jmazanec15 (approved):

PR #2621, opened 2025-03-20
[LuceneOnFaiss CAGRA] Added CAGRA index partial loading logic.

0ctopus13prime:
### Description
This PR add CAGRA index support in partial loading.
CAGRA index is built with GPU in parallel then copied to CPU index then flushed into a file.
This change has the logic to decode CAGRA index section (IHNc) and load HNSW graph.
See : https://github.com/facebookresearch/faiss/blob/main/faiss/impl/index_read.cpp#L966

### Related Issues

### Check List


0ctopus13prime:
Will add unit tests thorough with binary file.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissHNSWCagra.java
```diff
@@ -0,0 +1,94 @@
...
+        return flatVectors.getVectorEncoding();
+    }
+
+    @Override
+    public FloatVectorValues getFloatValues(IndexInput indexInput) throws IOException {
+        return flatVectors.getFloatValues(indexInput);
+    }
+
+    @Override
+    public ByteVectorValues getByteValues(IndexInput indexInput) throws IOException {
```
navneet1v, line 91: Byte vectors are not currently supported with cagra HNSW

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissHNSWCagra.java
```diff
@@ -0,0 +1,94 @@
...
+import org.apache.lucene.store.IndexInput;
+
+import java.io.IOException;
+
+/**
+ * This class represents FAISS CAGRA index which is built with GPU.
+ * The index has the same structure with a plain HNSW index built with CPU.
+ * But in addition to HNSW graph, it contains special CAGRA meta info such as `numBaseLevelSearchEntryPoint`.
+ */
+public class FaissHNSWCagra extends FaissIndex {
```
navneet1v, line 21: For search we will need a separate path. Will that be in a seperate PR?

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissHNSWCagra.java
0ctopus13prime: That decision will be made in underlying storage format - https://github.com/opensearch-project/k-NN/blob/lucene-on-faiss/src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissIndexFloatFlat.java#L117

So far, only float type is being supported, and an exception will be thrown when it was called.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissHNSWCagra.java
0ctopus13prime: no, it will not be the separate path as CAGRA is HNSW graph. We can pass it to Lucene graph searcher still.
I did not added the part in this revision, as this one showcase the idea of decoding, but that will be very simple. Just passing the HNSW graph to Lucene, that's all!

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissHNSWCagra.java
navneet1v: got it make sense.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissHNSWCagra.java
navneet1v: Just FYI, the HNSW graph doesn't have the entry points. It gets created randomly for cagra here: https://github.com/facebookresearch/faiss/blob/main/faiss/IndexHNSW.cpp#L919-L968

Please check the above. Its a small lift but please validate. May be Lucene HNSW graph searcher might be creating those entry points itself. Just check once.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/cagra/FaissCagraHNSW.java
```diff
@@ -0,0 +1,42 @@
...
+     *
+     * See : <a href="https://github.com/facebookresearch/faiss/blob/main/faiss/IndexHNSW.cpp#L956">...</a>
+     * @throws IOException
+     */
+    public void load(IndexInput input, long totalNumberOfVectors) throws IOException {
+        super.load(input, totalNumberOfVectors);
+
+        // We pick a random entry point
+        // See : https://github.com/facebookresearch/faiss/blob/main/faiss/IndexHNSW.cpp#L945
+        entryPoint = ThreadLocalRandom.current().nextInt((int) totalNumberOfVectors);
```
navneet1v, line 37: can pick a bunch of points and then find the closet of all those points and use that as the entry point?

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissHNSWCagra.java
0ctopus13prime: Yeah, I made it randomly pick entry point. 
Also found `maxLevel` will always have -1 which is a default value. 
So I fixed it that. Added unit tests to verified it can perform both exhaustive and ANN search

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/cagra/FaissCagraHNSW.java
0ctopus13prime: yeah, we can, but can we have a separate GH issue to optimize the entry point?
Not only to that, we could do some more experiments to make entry point optimal.

0ctopus13prime: Tracking in here - https://github.com/opensearch-project/k-NN/issues/2401#issuecomment-2744589698

0ctopus13prime:
@navneet1v 
Hi Navneet, could you approve the PR if it looks good on you?
Thank you.

Vikasht34 (approved):

navneet1v (approved):

PR #2623, opened 2025-03-20
Update merge policy floor_setting from 2MB to 16MB

finnroblin:
### Description

Sets `merge.policy.floor_segment : 16mb` for knn indices to create more merges and have fewer segments after indexing. The `floor_segment` setting in Lucene [TieredMergePolicy](https://www.google.com/url?sa=t&source=web&rct=j&opi=89978449&url=https://lucene.apache.org/core/9_3_0/core/org/apache/lucene/index/TieredMergePolicy.html&ved=2ahUKEwjOvaG8vJmMAxWzOTQIHeU6HmEQFnoECBIQAQ&usg=AOvVaw1ltvLgMA8pT0r5h2D3tR1m) dictates the "size" of a segment to the merge logic. That is, the merge logic [uses](https://github.com/apache/lucene/blob/3752dba9c51d817363c76e9a388cd68dcba5c4a0/lucene/core/src/java/org/apache/lucene/index/TieredMergePolicy.java#L405) `max(floor_segment, actualSegmentSize)` when calculating merges. The [OpenSearch default](https://github.com/opensearch-project/OpenSearch/blob/47d3655d8e86e44ab8ea4b58079d09fa0439db8c/server/src/main/java/org/opensearch/index/TieredMergePolicyProvider.java#L161) is 2mb. The [Lucene default](https://github.com/apache/lucene/commit/f28a8ae29c22aa1ebce51ced5f18525b444312c7) is 16mb since 10.2. This PR is against k-NN repo and not core because there are likely cases where a 2mb default is smarter for non-vector search 😄.

This [blog post](https://blog.mikemccandless.com/2011/02/visualizing-lucenes-segment-merges.html) by a Lucene maintainer explains and visualizes Lucene merges. The [TieredMergePolicy javadoc](https://github.com/apache/lucene/blob/3752dba9c51d817363c76e9a388cd68dcba5c4a0/lucene/core/src/java/org/apache/lucene/index/TieredMergePolicy.java#L31) also describes how the merge works. The upshot is that this setting change will lead to fewer segments after indexing, which will improve search performance. 

Benchmarks on the 10M 768d cohere dataset with 3x r6g.4xl instances [show](https://github.com/opensearch-project/k-NN/pull/2623#issuecomment-2749056885) 5-7% improvements in query throughput and common-case latency, larger 20+% improvements in tail latency, ~4% decrease in indexing time, but ~27% increase in median merge time. The original merge floor policy results in 212 segments and this new merge floor policy results in 154 segments. The ~27% decrease in segments explains the ~27% increase in merging time.

We can see that these setting listeners worked by querying the `_settings` endpoint. The response includes ` "merge":{"policy":{"floor_segment":"16mb"}}` for knn indices and `policy.floor_segment":"2097152b"` for non-knn indices. See the full [`_settings` response for knn index here](https://github.com/user-attachments/files/19436530/output-of-_settings-knn-index.txt) and the full [`_settings` response for non-knn index here](https://github.com/user-attachments/files/19436531/output-of-_settings-normal-index.txt).

### Check List


output-of-_settings-knn-index.txt
```diff
@@ -0,0 +1,32 @@
+curl -XPUT $ENDPOINT/test-index -H 'Content-Type: application/json' -d'
```
finnroblin, line 1: I'll remove this file after I get some feedback. The point of the file is to show the settings difference after adding this change for knn and non-knn indices. The merge policy is updated here:
"merge":{"policy":{"floor_segment":"16mb"}}

src/main/java/org/opensearch/knn/plugin/KNNIndexSettingProvider.java
```diff
@@ -0,0 +1,32 @@
...
+    private static boolean isKNNIndex(Settings settings) {
+        return settings.hasValue("index.knn") && settings.getAsBoolean("index.knn", true);
+    }
+
+    /**
+     * Returns additional index settings for k-NN index. In particular, we set the index.merge.policy.floor_segment = 16MB.
+     * This change is in line with Lucene 10.2 default and will lead to fewer segments (more merges), improving search performance.
+     */
+    @Override
+    public Settings getAdditionalIndexSettings(String indexName, boolean isDataStreamIndex, Settings templateAndRequestSettings) {
```
finnroblin, line 25: This overridden method hooks into OpenSearch [here](https://github.com/opensearch-project/OpenSearch/blob/8ee5eebc3728a19487c0604d9828f92770fcd7ec/server/src/main/java/org/opensearch/node/Node.java#L1001).

finnroblin:
Currently running cohere-10M benchmark.

src/test/java/org/opensearch/knn/index/SegmentSizeFloorMergePolicyIT.java
```diff
@@ -0,0 +1,56 @@
...
+        String mapping = XContentFactory.jsonBuilder()
+            .startObject()
+            .startObject("properties")
+            .startObject(FIELD_NAME)
+            .field("type", "knn_vector")
+            .field("dimension", 3)
+            .startObject(KNN_METHOD)
+            .field(NAME, METHOD_HNSW)
+            .field(METHOD_PARAMETER_SPACE_TYPE, SpaceType.L2)
+            .field(KNN_ENGINE, KNNEngine.FAISS.getName())
```
finnroblin, line 30: Should be Lucene engine in the test case (although this test is just changing the value, so engine being faiss won't make a difference). I'll change in next revision.

jmazanec15: Why does engine matter?

src/main/java/org/opensearch/knn/plugin/KNNIndexSettingProvider.java
```diff
@@ -0,0 +1,32 @@
...
+        if (isKNNIndex(templateAndRequestSettings)) {
+            return Settings.builder().put("index.merge.policy.floor_segment", KNN_DEFAULT_FLOOR_SEGMENT_VALUE).build();
```
jmazanec15, line 27: Can we depend on https://github.com/opensearch-project/OpenSearch/blob/1acba95906877249aa7beed68b212483849d3fe9/server/src/main/java/org/opensearch/index/TieredMergePolicyProvider.java#L161-L166 instead of hard coding string?

src/main/java/org/opensearch/knn/plugin/KNNIndexSettingProvider.java
shatejas, line 25: Lets add a unit test for this

finnroblin:
Here are benchmark results for 10M 768D dataset:

## Setup: 

instance types

- 3x r6g.4xl
- 6 primary shards
- 1 replica/shard
- deployed via opensearch-cluster-cdk
- osb client: m5.2xlarge w 8vcpu

benchmark setup
- same workload as [lucene   10M](https://github.com/opensearch-project/opensearch-benchmark-workloads/blob/main/vectorsearch/params/corpus/10million/lucene-cohere-768-dp.json)   but with no force merge
- tasks:   delete-target-index,create-target-index,custom-vector-bulk,refresh-target-index,prod-queries
- 1st prod-queries run discarded due to lucene cold start behavior. 
- The 2nd prod-queries run with the changed policy was an outlier and also discarded.


## results 
- Old policy: 212 segments, new policy: 154 segments
- Main takeaways: **5-7% improvements in query throughput and common-case latency**, larger 20+% improvements in tail latency, **~27% increase in median merge time**, ~4% decrease in indexing time (I don't think the indexing metric includes the merge time).
image

## full document

[lucene-floor-size-analysis.xlsx](https://github.com/user-attachments/files/19435861/lucene-floor-size-analysis.xlsx)

src/test/java/org/opensearch/knn/index/SegmentSizeFloorMergePolicyIT.java
finnroblin: It doesn't, you're right. I had the incorrect understanding that this merge policy is specific to Lucene engine, but it will also change the faiss merging behavior. I've left in the faiss engine setting.

src/main/java/org/opensearch/knn/plugin/KNNIndexSettingProvider.java
finnroblin: I added a few in [KNNIndexSettingProviderTests.java](https://github.com/opensearch-project/k-NN/pull/2623/files#diff-e230fd0dc2c51bd3193ea1d404bfa83dca1e44a552a6e40c2befe672aa4a8503). But the actual changed index setting (not just the result of getAdditionalindexSetting) is tested in an IT, [egmentSizeFloorMergePolicyIT.java](https://github.com/opensearch-project/k-NN/pull/2623/files#diff-481e58c91b6f76cab254fac96ea28015d9151783a3802a1c0d8dabbcec2b26a5).

output-of-_settings-knn-index.txt
finnroblin: Sure, moved to attachments in description with a brief summary.

src/test/java/org/opensearch/knn/index/SegmentSizeFloorMergePolicyIT.java
```diff
@@ -0,0 +1,56 @@
...
+            .endObject()
+            .toString();
+
+        createKnnIndex(knnIndexName, mapping);
+
+        // Create a non-KNN index
+        String regularIndexName = "test-regular-floor-segment";
+        createIndex(regularIndexName, Settings.EMPTY);
+
+        String knnFloorSegment = getIndexSettingByName(knnIndexName, "index.merge.policy.floor_segment", true);
```
kotwanikunal, line 43: nit: Use the setting key here `org.opensearch.index.TieredMergePolicyProvider.INDEX_MERGE_POLICY_FLOOR_SEGMENT_SETTING`

kotwanikunal (approved):

shatejas (dismissed):

finnroblin:
Want to call out that OS core just opened a PR for this change: https://github.com/opensearch-project/OpenSearch/pull/17699.

We can merge in this listener + setting update on the k-NN side to see the performance impact on nightly benchmarks, but we should probably revert the change once it's merged into core since it's unnecessary overhead to overwrite an equivalent setting value.

kotwanikunal:
Closing this out since core merged this in with #17687

Issue #2622, opened 2025-03-20
[FEATURE] Introduce Vector Sampling in K-NN


markwu-sde:
**Problem**

Today we provide index configurations comprising of index settings and mappings. With a wide range of algorithms we have multiple options like compression levels, hyper parameters, space type, encoders etc. All of these options allow for us to optimize for better performance, lower cost, etc. However they may not be applicable for every use case and in many instances depend on the how the vector space actually looks like (dimensionality, clustering, dense areas, etc.).

As of now we don’t have a lot of insight into what the actual vector space looks like in the index and so we aren't able derive insights to make more accurate and concrete recommendations related to recall issues or optimizations. 

**Solution**

The solution introduces sampling capabilities at both ingestion and query time to provide insights into the vector space. During document indexing, vectors are intercepted at flush time and asynchronously sampled, with statistical computations performed on this subset and stored alongside segment data in the Lucene directory.

Both ingestion and query flows feed into a global level API (like the Stats API) that aggregates sampling data across segments, providing metrics about vector distributions, dimension-level statistics, and search patterns. 

A high level sequence diagram of the design is found below:

![Image](https://github.com/user-attachments/assets/4d8ee7d6-fe34-48b6-a011-4e8b62a0058a)

User will be able to query their profiled statistics like so:

```
curl -X GET "localhost:9200/_plugins/_knn/sampling/my_index1/stats?pretty"
```

Result:

```
{
  "my_index1": {
    "vector_sample_count": 50000,
    "last_updated": "2025-03-20T10:15:30Z",
    "field_stats": {
      "product_vector": {
        "dimension_stats": [
          {
            "min": -0.9876,
            "max": 0.9954,
            "mean": 0.0234,
            "variance": 0.2567
          },
          // ... (stats for other dimensions)
        ],
        "distance_distribution": {
          "min_distance": 0.0123,
          "max_distance": 1.9876,
          "mean_distance": 0.7654,
          "percentiles": {
            "p50": 0.7123,
            "p75": 1.0234,
            "p90": 1.3456,
            "p99": 1.7890
          }
        },
        "sparsity": 0.1234,
        "clustering_coefficient": 0.7890
      }
    }
  }
}
```

Users should be ideally be able able to trigger the analysis "on-demand" for debugging:

```
curl -X POST "localhost:9200/_plugins/_knn/sampling/my_index1/_analyze?pretty" -H 'Content-Type: application/json' -d'
{
  "field": "product_vector",
[...]

PR #2624, opened 2025-03-22
Adding basic generic vector profiler implementation and tests.

oaganesh:
### Description
Basic generic implementation of a vector profiler that will produce a list of sampling statistics such as mean, standard deviation, and variance. Utility class for performing vector calculations and profiling operations on collections of float arrays representing vectors. This class utilizes the singleton pattern to maintain consistent state across operations within OpenSearch.

Utilizing [The Apache Commons Mathematics Library](https://commons.apache.org/proper/commons-math/userguide/stat.html) for generating vector statistics per dimension.

VectorProfiler provides functionality to:
 - Calculate statistical measures (mean, variance, standard deviation) on vector collections
 - Sample vectors from larger collections for efficient processing
 - Track and record vectors during read and write operations
 - Save vector statistics to disk for later analysis

Example input:
1) Creating the index
```
curl -X PUT "http://localhost:9200/my-index" -H 'Content-Type: application/json' -d '     
{
  "settings": {                          
    "index": {
      "knn": true
    }
  },
  "mappings": {
    "properties": {
      "my_vector_field": {
        "type": "knn_vector",
        "dimension": 4
      }
    }
  }
}'
```

2) Adding vectors to the index
```
curl -X POST "http://localhost:9200/my-index/_doc" -H 'Content-Type: application/json' -d'
{
  "my_vector_field": [0.1, 2.2, 3.3, 0.4]
}'
```

Observed example output:
```
[2025-03-28T16:10:27,867][INFO ][o.o.k.p.SegmentProfilerState] [integTest-0] Starting vector profiling with dimension: 4
[2025-03-28T16:10:27,887][INFO ][o.o.k.p.SegmentProfilerState] [integTest-0] Vector profiling completed - processed 1 vectors with 4 dimensions
[2025-03-28T16:10:27,887][INFO ][o.o.k.p.SegmentProfilerState] [integTest-0] Dimension 0 stats: mean=0.10000000149011612, std=0.0, min=0.10000000149011612, max=0.10000000149011612
[2025-03-28T16:10:27,888][INFO ][o.o.k.p.SegmentProfilerState] [integTest-0] Dimension 1 stats: mean=2.200000047683716, std=0.0, min=2.200000047683716, max=2.200000047683716
[2025-03-28T16:10:27,888][INFO ][o.o.k.p.SegmentProfilerState] [integTest-0] Dimension 2 stats: mean=3.299999952316284, std=0.0, min=3.299999952316284, max=3.299999952316284
[2025-03-28T16:10:27,888][INFO ][o.o.k.p.SegmentProfilerState] [integTest-0] Dimension 3 stats: mean=0.4000000059604645, std=0.0, min=0.4000000059604645, max=0.4000000059604645

```

### Related Issues
Implements #2622 

### Check List
- [ ✔️ ] New functionality includes testing.
- [ ✔️ ] New functionality has been documented.
- [ ✔️ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
- [✔️  ] Commits are signed per the DCO using `--signoff`.
- [ ✔️ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).

For more information on following Developer Certificate of Origin and signing off your commits, please check [✔️ ](https://github.com/opensearch-project/k-NN/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).


src/main/java/org/opensearch/knn/index/codec/KNN990Codec/StatisticalOperators.java
```diff
@@ -0,0 +1,59 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.knn.index.codec.KNN990Codec;
```
markwu-sde, line 6: Let's not put this into the codec folder. We should move everything to our own separate folder within knn.

oaganesh: Should I create a separate "sampler" or "profiler" folder within "org/opensearch/knn/index" for example?

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/VectorProfiler.java
markwu-sde, line 1: Good start overall. I think our previous conversations may have led to some confusion. I think the purpose of this PR is to exclude any of the lucene persistence details (SegmentWriteState, segments, directories, etc) and just expose a entry point that we can build on top of in the indexing and query path. Not saying that those are not important, but I believe that the persistence/output of the sampling is another conversation on it's own and deserves it's own PR.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/VectorProfiler.java
```diff
@@ -0,0 +1,492 @@
...
+    private static VectorProfiler INSTANCE;
+
+    // Default maximum number of vectors to sample per segment
+    private static final int DEFAULT_MAX_SAMPLE_SIZE = 100;
+
+    // Max number of vector elements to print in stats files
+    private static final int MAX_VECTOR_ELEMENTS_TO_PRINT = 10;
+
+    // Maps to store segment context information
+    private static final ConcurrentHashMap<String, String> SEGMENT_BASE_NAMES = new ConcurrentHashMap<>();
```
markwu-sde, line 52: Can you give more context on why we need these in-memory data structures? The generic form of the VectorProfiler class should in some Collection/Iterator of Vectors and performs a set of computations. Introducing lucene concepts here should be done in my opinion in the PR's where we integrate them into k-nn paths (read, write, etc)

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/VectorProfiler.java
```diff
@@ -0,0 +1,492 @@
...
+        if (vectors == null || vectors.isEmpty()) {
+            throw new IllegalArgumentException("Vectors collection cannot be null or empty");
+        }
+
+        try {
+            float[] firstVector = vectors.iterator().next();
+            int dim = firstVector.length;
+
+            float[] result = new float[dim];
+            Arrays.fill(result, 0);
```
markwu-sde, line 112: Do we need this?

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/VectorProfiler.java
```diff
@@ -0,0 +1,492 @@
...
+                    throw new IllegalArgumentException("All vectors must have same dimension");
+                }
+                for (int i = 0; i < dim; i++) {
+                    result[i] = computation.apply(result[i], vec[i])[0];
+                }
+            }
+
+            return computation.apply(result, vectors.size());
+            // return result;
+        } catch (Exception e) {
```
markwu-sde, line 125: Try to avoid catching generic exceptions.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/VectorProfiler.java
```diff
@@ -0,0 +1,492 @@
...
+     * This allows tracking statistics of vectors that are used during query operations.
+     *
+     * @param <T> Type parameter extending Computation interface
+     * @param segBaseName Base name of the segment
+     * @param segSuffix Suffix of the segment
+     * @param directoryPath Path to the directory where the segment is stored
+     * @param vectors Collection of vectors to record
+     * @param computation The computation to be performed on the vectors
+     */
+    public static <T extends Computation> void recordReadTimeVectors(
```
markwu-sde, line 142: So from I understand the sampling of the vectors between the ingestion (indexing) flow and the query (read) flow will be slightly different. On the ingestion flow with our current implementation all the vectors for that particular segment are sent to be sampled during `flush` while for the query side we'll probably need to do some sort of dynamic window type persistence.

Both paths will need a slightly different approach but since our plan is to have them in separate PR's can we leave the segment level implementation of this to those PRs? It will make it easier for us to look at each case separately instead of having to deduce what we're doing with the segments here.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/VectorProfiler.java
```diff
@@ -0,0 +1,492 @@
...
+     */
+    public static void saveVectorStats(SegmentWriteState segmentWriteState, Collection<float[]> vectors) throws IOException {
+        if (vectors == null || vectors.isEmpty()) {
+            log.debug("No vectors to save statistics for");
+            return;
+        }
+
+        try {
+            // Calculate all statistics
+            float[] meanVector = calculateVector(vectors, StatisticalOperators.MEAN);
```
markwu-sde, line 264: We don't want to explicitly include these operations in the method. Imagine a scenario where we have 100+ metrics that we need to compute - this would congest the content of what we're trying to do by a large factor and make refactoring difficult.  At the top level of the vector profiler we can include some sort of dynamic collection of computations that we have to execute that can be configurable via a setting and then read from there.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/VectorProfiler.java
```diff
@@ -0,0 +1,492 @@
...
+            float[] varianceVector = calculateVector(vectors, StatisticalOperators.VARIANCE);
+            float[] stdDevVector = calculateVector(vectors, StatisticalOperators.STANDARD_DEVIATION);
```
markwu-sde, line 266: ``calculateVector`` doesn't include any sampling? In the case that this is the entry point we would want to sample the vectors before we compute. We can reuse the sampling implementation in the plugin to do this.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/VectorProfiler.java
```diff
@@ -0,0 +1,492 @@
...
+
+            String statsFileName = IndexFileNames.segmentFileName(
```
markwu-sde, line 268: Let's move anything related to writing the results to a file/persisting it to the PR's related to indexing and querying unless you feel strongly about including them together. This would improve readability and keep the PR's manageable in terms of turnaround time.

src/main/java/org/opensearch/knn/profiler/VectorProfiler.java
```diff
@@ -0,0 +1,140 @@
...
+
+        if (vectors == null || vectors.isEmpty()) {
+            throw new IllegalArgumentException("Vectors collection cannot be null or empty");
+        }
+
+        float[] firstVector = vectors.iterator().next();
+        int dim = firstVector.length;
+        int numVectors = vectors.size();
+
+        // Create rotated matrix where each row represents a dimension
```
markwu-sde, line 88: Rotate may be an incorrect term - the technical concept of this is a transposition (flipping x and y)

src/main/java/org/opensearch/knn/profiler/VectorProfiler.java
```diff
@@ -0,0 +1,146 @@
...
+        float[] results = new float[aggregators.size()];
+        for (int dim = 0; dim < aggregators.size(); dim++) {
+            StatisticalSummary stats = aggregators.get(dim).getAggregateStatistics();
+            results[dim] = (float) computation.compute(stats);
+        }
+        return results;
+    }
+
+    private Collection<float[]> sampleVectors(Collection<float[]> vectors, int sampleSize) {
+        Sampler sampler = SamplingFactory.getSampler(SamplerType.RESERVOIR);
```
markwu-sde, line 127: I think there's room for improvement with how we sample our vectors and reduce memory footprint here. 

Given that we're not going to be on the hot path of indexing/query would it make sense to skip the sampling for now? We don't store the vectors in memory (due to SummaryStatistics) and the operation would be done concurrently.

src/main/java/org/opensearch/knn/profiler/VectorProfiler.java
```diff
@@ -0,0 +1,146 @@
...
+        }
+
+        // Update statistics for each dimension
+        for (int dim = 0; dim < dimensions; dim++) {
+            aggregators.get(dim).addSegmentStatistics(dimensionValues.get(dim));
+        }
+    }
+
+    private float[] generateSampledDimensionVectors(List<DimensionStatisticAggregator> aggregators, Computation computation) {
+        float[] results = new float[aggregators.size()];
```
markwu-sde, line 118: Do we still need this?

src/main/java/org/opensearch/knn/profiler/VectorProfiler.java
```diff
@@ -0,0 +1,146 @@
...
+        // Group values by dimension
+        for (float[] vector : vectors) {
+            for (int dim = 0; dim < Math.min(dimensions, vector.length); dim++) {
+                dimensionValues.get(dim).add(vector[dim]);
+            }
+        }
+
+        // Update statistics for each dimension
+        for (int dim = 0; dim < dimensions; dim++) {
+            aggregators.get(dim).addSegmentStatistics(dimensionValues.get(dim));
```
markwu-sde, line 113: I understand that we're trying to transpose the matrix before adding it to the ``DimensionStatisticAggregator``. 

I think we can avoid this by creating a segment item in the ``DimensionStatisticAggregator`` mapping as a one time operation to ensure that we can get the summary statistics. Then we can pass in the segment id to add the value to here and so we'd only have to iterate through the collection once instead of having to do multiple passes.

src/main/java/org/opensearch/knn/profiler/VectorProfiler.java
```diff
@@ -0,0 +1,146 @@
...
+        return new ArrayList<>(registeredComputations);
+    }
+
+    public Map<Computation, float[]> sampleAndCompute(String fieldName, Collection<float[]> vectors, int... sampleSize) {
+        validateVectors(vectors);
+
+        Collection<float[]> sampledVectors = sampleVectors(vectors, sampleSize.length > 0 ? sampleSize[0] : DEFAULT_SAMPLE_SIZE);
+
+        // Initialize dimension aggregators for the field if not exists
+        if (!fieldToDimensionStats.containsKey(fieldName)) {
```
markwu-sde, line 68: For readability can we extract this to another method?

src/main/java/org/opensearch/knn/profiler/VectorProfiler.java
```diff
@@ -0,0 +1,146 @@
...
+
+    public void unregisterComputation(Computation computation) {
+        registeredComputations.remove(computation);
+    }
+
+    public List<Computation> getRegisteredComputations() {
+        return new ArrayList<>(registeredComputations);
+    }
+
+    public Map<Computation, float[]> sampleAndCompute(String fieldName, Collection<float[]> vectors, int... sampleSize) {
```
markwu-sde, line 62: NIT: Final whenever possible

src/main/java/org/opensearch/knn/profiler/VectorProfiler.java
```diff
@@ -0,0 +1,146 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.HashMap;
+
+@Log4j2
+public class VectorProfiler {
+    private static VectorProfiler INSTANCE;
+    private static final int DEFAULT_SAMPLE_SIZE = 1000;
+    private final Map<String, List<DimensionStatisticAggregator>> fieldToDimensionStats;
+    private List<Computation> registeredComputations;
```
markwu-sde, line 25: We may not need this anymore because we will compute everything using the ``DimensionStatisticAggregator``

src/main/java/org/opensearch/knn/profiler/VectorProfiler.java
```diff
@@ -0,0 +1,146 @@
...
+
+    public VectorProfiler() {
```
markwu-sde, line 27: Singletons have [private](https://www.geeksforgeeks.org/singleton-design-pattern/) constructors.

src/main/java/org/opensearch/knn/profiler/DimensionStatisticAggregator.java
```diff
@@ -0,0 +1,46 @@
...
+        }
+
+        segmentToSummaryMapping.put(segmentId, segmentStats);
+    }
+
+    public StatisticalSummary getAggregateStatistics() {
+        return aggregateStats;
+    }
+
+    public Map<String, SummaryStatistics> getSegmentStatistics() {
```
markwu-sde, line 43: Can we also add a method to retrieve a SummaryStatistic for a particular segment?

src/main/java/org/opensearch/knn/profiler/DimensionStatisticAggregator.java
```diff
@@ -0,0 +1,46 @@
...
+    private final int dimension;
+
+    public DimensionStatisticAggregator(int dimension) {
+        this.dimension = dimension;
+        this.aggregateStats = new AggregateSummaryStatistics();
+        this.segmentToSummaryMapping = new HashMap<>();
+    }
+
+    public void addSegmentStatistics(Collection<Float> values) {
+        String segmentId = UUID.randomUUID().toString();
```
markwu-sde, line 29: Can we add a TODO saying that we'll attach the segmentId in another PR and why we're using a uuid for now?

src/main/java/org/opensearch/knn/profiler/DimensionStatisticAggregator.java
```diff
@@ -0,0 +1,46 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.UUID;
+
+public class DimensionStatisticAggregator {
+    private final AggregateSummaryStatistics aggregateStats;
+    private final Map<String, SummaryStatistics> segmentToSummaryMapping;
+    private final int dimension;
+
+    public DimensionStatisticAggregator(int dimension) {
```
markwu-sde, line 22: NIT: final variables whenever possible

src/main/java/org/opensearch/knn/profiler/DimensionStatisticAggregator.java
```diff
@@ -0,0 +1,46 @@
...
+
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.UUID;
+
+public class DimensionStatisticAggregator {
+    private final AggregateSummaryStatistics aggregateStats;
+    private final Map<String, SummaryStatistics> segmentToSummaryMapping;
+    private final int dimension;
```
markwu-sde, line 20: Is this corresponding to the # of dimensions or the id of the particular dimension?

src/main/java/org/opensearch/knn/profiler/DimensionStatisticAggregator.java
```diff
@@ -0,0 +1,46 @@
...
+import org.apache.commons.math3.stat.descriptive.AggregateSummaryStatistics;
+import org.apache.commons.math3.stat.descriptive.SummaryStatistics;
+import org.apache.commons.math3.stat.descriptive.StatisticalSummary;
+
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.UUID;
+
+public class DimensionStatisticAggregator {
```
markwu-sde, line 17: Can we add Javadoc comments here?

src/main/java/org/opensearch/knn/profiler/Computation.java
```diff
@@ -0,0 +1,24 @@
...
+
+import org.apache.commons.math3.stat.descriptive.StatisticalSummary;
+
+/**
+ * Interface defining computation operations for k-Nearest Neighbor (KNN) calculations.
+ * This interface provides methods to perform numerical computations on floating point values
+ * used in KNN algorithm implementations.
+ */
+
+public interface Computation {
```
markwu-sde, line 16: We don't need the computation interface here. We can leave the selection of computations at the API level since we now have the ``StatisticalSummaries``.

markwu-sde:
``Tests with failures:
org.opensearch.knn.profiler.VectorProfilerTests.testDifferentDimensions``
 
 Some of the tests are failing can we check?

src/main/java/org/opensearch/knn/profiler/StatisticalOperators.java
```diff
@@ -0,0 +1,54 @@
...
+
+package org.opensearch.knn.profiler;
+
+/**
+ * Enum for statistical operators used in KNN vector computation.
+ */
+public enum StatisticalOperators implements Computation {
+    MEAN {
+        @Override
+        public float[] compute(float[] perDimensionVector) {
```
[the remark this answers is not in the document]
oaganesh: Utilizing SummaryStatistics and AggregateSummaryStatistics from library. Components are not stored in memory.

src/main/java/org/opensearch/knn/profiler/VectorProfiler.java
oaganesh: Yes, we can remove the sampling

src/main/java/org/opensearch/knn/profiler/VectorProfiler.java
oaganesh: No, removed to simplify the VectorProfiler

src/main/java/org/opensearch/knn/profiler/VectorProfiler.java
```diff
@@ -0,0 +1,146 @@
...
+            fieldToDimensionStats.put(fieldName, dimensionAggregators);
+        }
+
+        // Update statistics for each dimension
+        List<DimensionStatisticAggregator> aggregators = fieldToDimensionStats.get(fieldName);
+        updateDimensionStatistics(aggregators, sampledVectors);
+
+        // Compute results
+        Map<Computation, float[]> results = new HashMap<>();
+        for (Computation computation : registeredComputations) {
```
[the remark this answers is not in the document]
oaganesh: No, removed the usecase for the registeredComputations

src/main/java/org/opensearch/knn/profiler/VectorProfiler.java
oaganesh: Created a initializeDimensionAggregators() method for this implementation

src/main/java/org/opensearch/knn/profiler/DimensionStatisticAggregator.java
oaganesh: Created a getSegmentStatistic() method

src/main/java/org/opensearch/knn/profiler/DimensionStatisticAggregator.java
oaganesh: The ID of the particular dimension; added comments and renamed for better visibility

src/main/java/org/opensearch/knn/profiler/VectorProfiler.java
```diff
@@ -0,0 +1,83 @@
...
+
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.ArrayList;
+import java.util.UUID;
+
+@Log4j2
+public class VectorProfiler {
```
markwu-sde, line 17: Can we have javadocs for this class?

src/main/java/org/opensearch/knn/profiler/VectorProfiler.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    public static synchronized VectorProfiler getInstance() {
+        if (INSTANCE == null) {
+            INSTANCE = new VectorProfiler();
+        }
+        return INSTANCE;
+    }
+
+    public static synchronized void setInstance(VectorProfiler instance) {
```
markwu-sde, line 32: Let's not have a setInstance here for singletons - for testing we can provide setters for the content within the class

src/main/java/org/opensearch/knn/profiler/VectorProfiler.java
```diff
@@ -0,0 +1,83 @@
...
+        INSTANCE = instance;
+    }
+
+    public void setFieldToDimensionStats(Map<String, List<DimensionStatisticAggregator>> stats) {
+        fieldToDimensionStats.clear();
```
markwu-sde, line 37: NIT: No need to clear here - if we want to set this just for testing we can call get() and then clear it ourselves in the test.

src/main/java/org/opensearch/knn/profiler/VectorProfiler.java
```diff
@@ -0,0 +1,83 @@
...
+        fieldToDimensionStats.putAll(stats);
+    }
+
+    public void processVectors(final String fieldName, final Collection<float[]> vectors) {
+        validateVectors(vectors);
+
+        float[] firstVector = vectors.iterator().next();
```
markwu-sde, line 44: Calling ``.next()`` in the iterator implementation would return the first element but it wouldn't be processed in the sampling as it's just a pointer.

We don't have to determine the dimension here - there should be a way to retrieve it from the field metadata during our write/read paths. We can probably add it as a parameter and leverage this.

src/main/java/org/opensearch/knn/profiler/VectorProfiler.java
```diff
@@ -0,0 +1,83 @@
...
+        }
+    }
+
+    public List<DimensionStatisticAggregator> getFieldStatistics(final String fieldName) {
+        return fieldToDimensionStats.get(fieldName);
+    }
+
+    private void validateVectors(final Collection<float[]> vectors) {
+        if (vectors == null || vectors.isEmpty()) {
+            throw new IllegalArgumentException("Vectors collection cannot be null or empty");
```
markwu-sde, line 80: No need to throw an exception here - we can just log and exit out of the profiler.

src/main/java/org/opensearch/knn/profiler/VectorProfiler.java
```diff
@@ -0,0 +1,90 @@
...
+    public void processVectors(final String fieldName, final Collection<float[]> vectors, final int dimensions) {
+        if (vectors == null || vectors.isEmpty()) {
+            log.warn("No vectors to process for field: {}", fieldName);
+            fieldToDimensionStats.remove(fieldName);
+            return;
+        }
+
+        fieldToDimensionStats.remove(fieldName);
+
+        initializeDimensionAggregators(fieldName, dimensions);
```
markwu-sde, line 57: If we've seen the field before, there's no need to create a new list of DA. Reuse the old one and add a segment entry to it.

src/main/java/org/opensearch/knn/profiler/VectorProfiler.java
```diff
@@ -0,0 +1,90 @@
...
+     * @param dimensions The number of dimensions in the vectors
+     */
+    public void processVectors(final String fieldName, final Collection<float[]> vectors, final int dimensions) {
+        if (vectors == null || vectors.isEmpty()) {
+            log.warn("No vectors to process for field: {}", fieldName);
+            fieldToDimensionStats.remove(fieldName);
+            return;
+        }
+
+        fieldToDimensionStats.remove(fieldName);
```
markwu-sde, line 55: We shouldn't have this here -> this will remove the stats for previous segments added.

src/main/java/org/opensearch/knn/profiler/VectorProfiler.java
```diff
@@ -0,0 +1,90 @@
...
+     * Process a collection of vectors for a specific field and dimension size.
+     *
+     * @param fieldName The name of the field containing the vectors
+     * @param vectors The collection of vectors to process
+     * @param dimensions The number of dimensions in the vectors
+     */
+    public void processVectors(final String fieldName, final Collection<float[]> vectors, final int dimensions) {
+        if (vectors == null || vectors.isEmpty()) {
+            log.warn("No vectors to process for field: {}", fieldName);
+            fieldToDimensionStats.remove(fieldName);
```
markwu-sde, line 51: We shouldn't have this here -> this will remove the stats for previous segments added.

src/test/java/org/opensearch/knn/profiler/VectorProfilerTests.java
```diff
@@ -0,0 +1,123 @@
...
+public class VectorProfilerTests extends OpenSearchTestCase {
+
+    private AutoCloseable mocks;
+    private VectorProfiler profiler;
+    private List<float[]> testVectors;
+    private static final String TEST_FIELD = "test_field";
+    private static final int TEST_DIMENSIONS = 3;
+
+    @Before
+    public void setUp() throws Exception {
```
markwu-sde, line 28: If we want to clear the map of fieldToDimensionStats, we can provide an accessor to VectorProfiler and clear the map within the test here.`getFieldToDimensionStats.clear()`

markwu-sde:
Thanks for the revisions. LGTM - @jmazanec15 can you take a look?

src/main/java/org/opensearch/knn/plugin/KNNPlugin.java
```diff
@@ -219,7 +219,6 @@ public Collection<Object> createComponents(
         KNNQueryBuilder.initialize(ModelDao.OpenSearchKNNModelDao.getInstance());
         KNNWeight.initialize(ModelDao.OpenSearchKNNModelDao.getInstance());
         TrainingModelRequest.initialize(ModelDao.OpenSearchKNNModelDao.getInstance(), clusterService);
-
```
jmazanec15, line 222: nit: revert this so it doesnt show up in diff.

CHANGELOG.md
```diff
@@ -20,6 +20,8 @@ Add filter function to KNNQueryBuilder with unit tests and integration tests [#2
 
 ## [Unreleased 2.x](https://github.com/opensearch-project/k-NN/compare/2.19...2.x)
 ### Features
+* [Vector Profiler] Adding generic vector profiler implementation to sample vector statistics and output towards the API (#2622)
```
jmazanec15, line 23: nit: check the format in change log on PR link

src/main/java/org/opensearch/knn/profiler/DimensionStatisticAggregator.java
```diff
@@ -0,0 +1,75 @@
...
+import org.apache.commons.math3.stat.descriptive.SummaryStatistics;
+import org.apache.commons.math3.stat.descriptive.StatisticalSummary;
+
+import java.util.Map;
+import java.util.HashMap;
+
+/**
+ * Aggregates statistics for a specific dimension across multiple segments.
+ * This class is used to collect and analyze statistical data for a particular dimension
+ * across multiple segments of a vector.
```
jmazanec15, line 17: nit: segments of an index?

src/main/java/org/opensearch/knn/profiler/DimensionStatisticAggregator.java
```diff
@@ -0,0 +1,75 @@
...
+            segmentId,
+            k -> aggregateStats.createContributingStatistics()
+        );
+        segmentStats.addValue(value);
+    }
+
+    /**
+     * @return the aggregate statistics across all segments
+     */
+    public StatisticalSummary getAggregateStatistics() {
```
jmazanec15, line 48: nit: use lombok getter for retrieving member variables

src/main/java/org/opensearch/knn/profiler/VectorProfiler.java
```diff
@@ -0,0 +1,97 @@
...
+
+    public static synchronized VectorProfiler getInstance() {
+        if (INSTANCE == null) {
+            INSTANCE = new VectorProfiler();
+        }
+        return INSTANCE;
+    }
+
+    /**
+     * Process a collection of vectors for a specific field and dimension size.
```
jmazanec15, line 42: nit: should this be profile? What does process mean in this context?

src/main/java/org/opensearch/knn/profiler/VectorProfiler.java
```diff
@@ -0,0 +1,97 @@
...
+     */
+    public List<DimensionStatisticAggregator> getFieldStatistics(final String fieldName) {
+        return fieldToDimensionStats.get(fieldName);
+    }
+
+    /**
+     * Retrieve the entire map of field to dimension statistics.
+     * @return fieldToDimensionStats
+     */
+    Map<String, List<DimensionStatisticAggregator>> getFieldToDimensionStats() {
```
jmazanec15, line 94: lombok getter

src/test/java/org/opensearch/knn/profiler/VectorProfilerTests.java
```diff
@@ -0,0 +1,123 @@
...
+import org.junit.Test;
+import org.mockito.MockitoAnnotations;
+import org.opensearch.test.OpenSearchTestCase;
+
+import java.util.Arrays;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+public class VectorProfilerTests extends OpenSearchTestCase {
```
jmazanec15, line 19: Does the profiler need to be singleton? This is one of the problems with going singleton route. It can complicate the testing. Instead, we can create via the KNNPLugin createComponents and pass into the codec/required transport actions from there.

jmazanec15 (commented):
Overall I think its a good start! I think we might want to make a few changes based on how its going to be used. 

Should a vector profiler be tied to the lifecycle of the NativeEngineWriter instance of a global singleton? See https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriter.java#L109. This is where per segment values are added and quantization samples/trained. If we are going to serialize the information in the segment, we will need access here.

Then, in order to aggregate at the segment, shard and index level, we follow path of warmup/clear cache in the KNNIndexShard. See https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/codecs/KnnVectorsReader.java#L86-L87 - we would need to access this info somehow through a custom KnnCollector.

src/main/java/org/opensearch/knn/profiler/VectorProfiler.java
oaganesh: True, I have changed it to profileVectors() for consistency.

src/main/java/org/opensearch/knn/profiler/DimensionStatisticAggregator.java
jmazanec15: Meant this:  
> \* This class is used to collect and analyze statistical data for a particular dimension
> \* across multiple segments of a ~vector~ index.

src/main/java/org/opensearch/knn/profiler/SegmentProfilerState.java
```diff
@@ -0,0 +1,110 @@
...
+            for (int j = 0; j < firstVector.length; j++) {
+                statistics.get(j).addValue(firstVector[j]);
+            }
+
+            // Process remaining vectors
+            int vectorCount = 1;
+            while (vectorValues.nextDoc() != NO_MORE_DOCS) {
+                vectorCount++;
+                float[] vector = (float[]) vectorValues.getVector();
+                for (int j = 0; j < vector.length; j++) {
```
markwu-sde, line 85: NIT: This is the same as line 76. If we need to reuse the code can we create a method for it

src/main/java/org/opensearch/knn/profiler/SegmentProfilerState.java
```diff
@@ -0,0 +1,110 @@
...
+                    i,
+                    stats.getMean(),
+                    stats.getStandardDeviation(),
+                    stats.getMin(),
+                    stats.getMax()
+                );
+            }
+
+            return new SegmentProfilerState(statistics);
+        } catch (ClassCastException e) {
```
markwu-sde, line 104: I'm curious to know more about the `ClassCastException` would be thrown, especially in what cases will we expect this. Have we debugged this?

src/main/java/org/opensearch/knn/profiler/SegmentProfilerState.java
```diff
@@ -0,0 +1,110 @@
...
+
+        // Return empty state if no documents are present
+        if (vectorValues.docId() == NO_MORE_DOCS) {
+            log.info("No vectors to profile");
+            return new SegmentProfilerState(statistics);
+        }
+
+        try {
+            // Process the first vector to determine dimensions
+            float[] firstVector = (float[]) vectorValues.getVector();
```
markwu-sde, line 67: `KNNVectorValues<?> vectorValues` has a dimension property. Can we use that instead?

src/test/java/org/opensearch/knn/profiler/SegmentProfilerStateTests.java
```diff
@@ -0,0 +1,114 @@
...
+import org.junit.Before;
+import org.opensearch.knn.index.vectorvalues.KNNVectorValues;
+import org.opensearch.test.OpenSearchTestCase;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.function.Supplier;
+
+import static org.mockito.Mockito.*;
```
markwu-sde, line 19: NIT: we shouldn't use wildcard imports

src/main/java/org/opensearch/knn/profiler/SegmentProfilerState.java
oaganesh: We can expect this to be thrown if the KNNVectorValues implementation changes and starts returning a different vector type. This is demonstrated when vectorValues.getVector() returns a type other than float[]. For example in line 67 when type casting to a float[] we can observe a unit test from "NativeEngines990KnnVectorsFormatTests" that throws the following error:
```
java.lang.ClassCastException: class [B cannot be cast to class [F ([B and [F are in module java.base of loader 'bootstrap')
```

markwu-sde: Can you elaborate on returning a different vector type? 

My point is here that our code changes are leading to some casting exceptions not in our class. We shouldn't be catching the exception unless we understand why it's giving the us the error and if we expect this error to happen or not.

It looks like it's returning a byte array instead of a float array. Can we figure out why it's doing that? Do we support byte[] inputs as well as floats?

src/main/java/org/opensearch/knn/profiler/SegmentProfilerState.java
```diff
@@ -0,0 +1,126 @@
...
+            }
+
+            processVectors(firstVector, statistics);
+
+            // Process remaining vectors
+            int vectorCount = 1;
+            while (vectorValues.nextDoc() != NO_MORE_DOCS) {
+                vectorCount++;
+                float[] vector = (float[]) vectorValues.getVector();
+                processVectors(vector, statistics);
```
markwu-sde, line 83: I think we can move the for loop back in here if we don't need to explicitly get the first vector. What do you think?

src/main/java/org/opensearch/knn/profiler/SegmentProfilerState.java
oaganesh: Refactored to process vectors in one sequence

src/main/java/org/opensearch/knn/profiler/SegmentProfilerState.java
oaganesh: Since we are adding float[] and byte[] vector types under the processVectors() method, I believe it would be better to keep the method in the loop for now.

src/main/java/org/opensearch/knn/profiler/SegmentProfilerState.java
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    /**
+     * Processes a byte vector by updating the statistical summaries for each dimension
+     * @param vector
+     * @param statistics
+     */
+    private static void processByteVector(byte[] vector, List<SummaryStatistics> statistics) {
+        for (int j = 0; j < vector.length; j++) {
+            statistics.get(j).addValue(vector[j] & 0xFF);
```
markwu-sde, line 119: If I recall correctly adding a Byte value to a double (which summary statistics require) would yield no lossy conversion since byte is 8 and double is 64 bits. Can you explain what what we're doing here with `& 0xFF`? I'm a little confused.

src/main/java/org/opensearch/knn/profiler/SegmentProfilerState.java
```diff
@@ -0,0 +1,141 @@
...
+            statistics.get(j).addValue(vector[j]);
+        }
+    }
+
+    /**
+     * Processes a byte vector by updating the statistical summaries for each dimension
+     * @param vector
+     * @param statistics
+     */
+    private static void processByteVector(byte[] vector, List<SummaryStatistics> statistics) {
```
markwu-sde, line 117: Same thing - does this need to be static?

src/main/java/org/opensearch/knn/profiler/SegmentProfilerState.java
```diff
@@ -0,0 +1,141 @@
...
+            throw new IllegalArgumentException("Unsupported vector type.");
+        }
+    }
+
+    /**
+     * Processes a float vector by updating the statistical summaries for each dimension
+     * @param vector
+     * @param statistics
+     */
+    private static void processFloatVector(float[] vector, List<SummaryStatistics> statistics) {
```
markwu-sde, line 106: Same here - do we need to call this from a static context?

src/main/java/org/opensearch/knn/profiler/SegmentProfilerState.java
```diff
@@ -0,0 +1,141 @@
...
+     * @param vector
+     * @param statistics
+     */
+    private static void processVectors(Object vector, List<SummaryStatistics> statistics) {
+        if (vector instanceof float[]) {
+            processFloatVector((float[]) vector, statistics);
+        } else if (vector instanceof byte[]) {
+            processByteVector((byte[]) vector, statistics);
+        } else {
+            throw new IllegalArgumentException("Unsupported vector type.");
```
markwu-sde, line 97: Should we throw an exception here if we aren't able to profile a vector of a unsupported type? This is a profiler that wouldn't be in the hot path - would it be better to log and move on?

src/main/java/org/opensearch/knn/profiler/SegmentProfilerState.java
```diff
@@ -0,0 +1,141 @@
...
+
+        return new SegmentProfilerState(statistics);
+    }
+
+    /**
+     * Helper method to process a vector and update statistics
+     * @param vector
+     * @param statistics
+     */
+    private static void processVectors(Object vector, List<SummaryStatistics> statistics) {
```
markwu-sde, line 91: Is there a reason for this method to be static? Are we calling this from a static context?

src/main/java/org/opensearch/knn/profiler/SegmentProfilerState.java
```diff
@@ -0,0 +1,141 @@
...
+        // Initialize statistics collectors for each dimension
+        for (int i = 0; i < dimension; i++) {
+            statistics.add(new SummaryStatistics());
+        }
+
+        // Process all vectors
+        int vectorCount = 0;
+        for (int doc = vectorValues.docId(); doc != NO_MORE_DOCS; doc = vectorValues.nextDoc()) {
+            vectorCount++;
+            Object vector = vectorValues.getVector();
```
markwu-sde, line 76: Do we have to cast this to an Object? Can we pass whatever type (I think `T`) to the process vectors and then cast it when we need it? 

There's a difference between generic `T` and `Object`. Which one do you think would be better?

src/main/java/org/opensearch/knn/profiler/SegmentProfilerState.java
```diff
@@ -0,0 +1,141 @@
...
+        log.info("Starting vector profiling with dimension: {}", dimension);
+
+        // Initialize statistics collectors for each dimension
+        for (int i = 0; i < dimension; i++) {
+            statistics.add(new SummaryStatistics());
+        }
+
+        // Process all vectors
+        int vectorCount = 0;
+        for (int doc = vectorValues.docId(); doc != NO_MORE_DOCS; doc = vectorValues.nextDoc()) {
```
markwu-sde, line 74: NIT: the vectorValues iterator will increment automatically once ``nextDoc()`` is called. Would it be clear. What is the purpose of ``int doc`` doing here?

src/main/java/org/opensearch/knn/profiler/SegmentProfilerState.java
markwu-sde: Synced with @oaganesh and this will be invoked from a static context

src/main/java/org/opensearch/knn/profiler/SegmentProfilerState.java
markwu-sde: Synced with @oaganesh and this will be invoked from a static context

src/main/java/org/opensearch/knn/profiler/SegmentProfilerState.java
oaganesh: The doc variable is keeping track of the current document ID being processed. While nextDoc() does advance the iterator, storing the document ID in doc allows us to know which document we're currently processing.

src/main/java/org/opensearch/knn/profiler/SegmentProfilerState.java
oaganesh: Added generic support for in processVectors() method since getVector already returns a T type.

src/main/java/org/opensearch/knn/profiler/SegmentProfilerState.java
oaganesh: Currently, it's static because it's called from the static profileVectors method. If the class were more instance-oriented, this could be changed to non-static.

Since profileVectors is static and processVectors is used in a static context such as in line 77 I have made these methods static.

src/main/java/org/opensearch/knn/profiler/SegmentProfilerState.java
oaganesh: Understood, creating log to catch issues

src/main/java/org/opensearch/knn/profiler/SegmentProfilerState.java
oaganesh: When I was trying to look into how to use the byte conversion I saw that we could do the 0xFF conversion which masks the signed byte to get its unsigned value.

As a signed byte, it's interpreted as -1, after & 0xFF, it's interpreted as 255

[+] https://stackoverflow.com/questions/11380062/what-does-value-0xff-do-in-java

@jmazanec15 do we do anything to the byte vector to process it where we need it to maintain it's signed value?

jmazanec15: This came up recently with @0ctopus13prime . I think this is fine. See https://github.com/opensearch-project/k-NN/blob/00addee01cbefb0612c3c52213271b979b98d8b6/src/main/java/org/opensearch/knn/memoryoptsearch/faiss/reconstruct/Faiss8BitsDirectSignedReconstructor.java#L16-L20.

jmazanec15 (approved):
LGTM thanks @oaganesh !

markwu-sde (approved):
LGTM

Issue #2586, opened 2025-03-06
Improve index parameter retrieval (`method` map in the index mappings)


owenhalpert:
The most recent Remote Build PR (#2560) introduced a change to retrieve the index parameters the client needs to send to the build service, but fetching the parameters whether from `FieldInfo` or `BuildIndexParams` is very fragile:
https://github.com/opensearch-project/k-NN/blob/fbb2d1b072a0d9d29683d1b7fb1cf17e8dcc02f5/src/main/java/org/opensearch/knn/index/engine/faiss/FaissHNSWMethod.java#L163-L187

As a Map<String, Object>, the `parameters` attribute of a BuildIndexParams object needs casting and a lot of unsafe `get`'s to parse the map.

As a Map<String, String>, `FieldInfo.attributes` is even more difficult because we need to use a JSON parser to parse the parameters map as a long string. 

Ideally, these parameters could be fetched in a safer way, or if not, the parsing mechanism can be improved to be less fragile and cumbersome.

owenhalpert:
BuildIndexParams:
```
BuildIndexParams(
fieldName=location, 
knnEngine=FAISS, 
indexOutputWithBuffer={indexOutput=FSIndexOutput(path="/"), len(buffer)=65536}, 
vectorDataType=FLOAT, 
parameters={name=hnsw, data_type=float, index_description=HNSW2,Flat, spaceType=l2, parameters={ef_search=100, ef_construction=94, encoder={name=flat, parameters={}}}, indexThreadQty=1}, 
quantizationState=null, 
knnVectorValuesSupplier=org.opensearch.knn.index.vectorvalues.KNNVectorValuesFactory, 
totalLiveDocs=5, 
segmentWriteState=org.apache.lucene.index.SegmentWriteState)
```

FieldInfo.attributes:
```
"knn_field" -> "true"
"PerFieldKnnVectorsFormat.suffix" -> "0"
"spaceType" -> "l2"
"engine" -> "faiss"
"data_type" -> "float"
"PerFieldKnnVectorsFormat.format" -> "NativeEngines990KnnVectorsFormat"
"dimension" -> "2"
"parameters" -> "{"index_description":"HNSW12,Flat","spaceType":"l2","name":"hnsw","data_type":"float","parameters":{"ef_search":24,"ef_construction":28,"encoder":{"name":"flat","parameters":{}}}}"
```

owenhalpert:
[...]

PR #2627, opened 2025-03-25
Add space type conversion to support Faiss cosine similarity

owenhalpert:
### Description
For the Faiss engine to support the `COSINESIMIL` space type, the `getKNNLibraryIndexingContext` method converts it to `INNER_PRODUCT` in the `KNNLibraryIndexingContextImpl`. This PR adds the same conversion to the remote build parameter fetching. 

The overall improvement of the parameter fetching is tracked in #2586, which may evolve to include migrating the entire parameter fetching to use `KNNLibraryIndexingContext` instead of just a `KNNMethodContext`. It wouldn't have much of an effect now, because the rest of the parameters are taken directly from the `KNNMethodContext` anyways ([ref](https://github.com/opensearch-project/k-NN/blob/e55e49b249feced18f3b39535969ea1cc5ad1798/src/main/java/org/opensearch/knn/index/engine/MethodComponent.java#L325)), but switching to `KNNLibraryIndexingContext` might better support any more future changes like this space type conversion.

### Related Issues
Resolves [#33](https://github.com/opensearch-project/remote-vector-index-builder/issues/33)

### Check List


navneet1v (approved):

owenhalpert:
Now that we are using the `LibraryIndexingContext` we can get the vector data type from there instead of [indexInfo](https://github.com/opensearch-project/k-NN/pull/2627/files#diff-26c0755811a046205de6e64f079dcb8cd415c6f751b40f8af83df87a328b63c3R225) what do you think @jmazanec15?

src/test/java/org/opensearch/knn/index/engine/faiss/FaissTests.java
```diff
@@ -346,7 +346,6 @@ public void testMethodAsMapBuilder() throws IOException {
 
         Map<String, Object> expectedParametersMap = new HashMap<>(methodComponentContext.getParameters());
         expectedParametersMap.put(parameter3, defaultValue3);
-        expectedParametersMap.remove(parameter1);
```
owenhalpert, line 349: Because of the change made in `MethodAsMapBuilder`, the maps won't differ here (we are no longer removing a parameter like `parameter1`, only putting it in the `index_description`.

src/main/java/org/opensearch/knn/index/engine/faiss/FaissHNSWMethod.java
```diff
@@ -154,34 +160,55 @@ protected Function<TrainingConfigValidationInput, TrainingConfigValidationOutput
...
+    @SuppressWarnings("unchecked")
+    static boolean supportsRemoteIndexBuild(Map<String, Object> parameters) {
+        try {
+            Map<String, Object> innerMap = (Map<String, Object>) parameters.get(PARAMETERS);
+            Map<String, Object> encoderMap = (Map<String, Object>) innerMap.get(METHOD_ENCODER_PARAMETER);
+            String dataType = getStringFromMap(parameters, VECTOR_DATA_TYPE_FIELD);
+            String encoder = getStringFromMap(encoderMap, NAME);
+            return SUPPORTED_REMOTE_INDEX_DATA_TYPES.contains(VectorDataType.get(dataType)) && ENCODER_FLAT.equals(encoder);
+        } catch (IllegalArgumentException e) {
+            log.error("Unrecognized indexing parameters in KNNLibraryIndexingContext");
```
navneet1v, line 209:
```suggestion
            log.error("Unrecognized indexing parameters in KNNLibraryIndexingContext ", e);
```

Lets also log what was the indexing parameters

src/main/java/org/opensearch/knn/index/engine/faiss/FaissHNSWMethod.java
```diff
@@ -197,4 +224,15 @@ private static Integer getIntegerFromMap(Map<String, Object> map, String key) {
...
+
+    /**
+     * Safely retrieve a String from {@code map} using {@code key}
+     */
+    private static String getStringFromMap(Map<String, Object> map, String key) throws IllegalArgumentException {
+        Object value = map.get(key);
+        if (value instanceof String) {
+            return (String) value;
+        }
+        throw new IllegalArgumentException("Could not parse value for key: " + key + map);
```
navneet1v, line 236:
```suggestion
        throw new IllegalArgumentException("Could not parse value for key: " + key + " map: " map);
```

jmazanec15 (dismissed):
LGTM thanks @owenhalpert!

src/main/java/org/opensearch/knn/index/engine/faiss/FaissHNSWMethod.java
owenhalpert: Oh thanks, I had that in for a test run. Do you think we should include the map? It would look something like this:
`Could not parse value for key: false_key map: {index_description=HNSW12,Flat, spaceType=innerproduct, name=hnsw, data_type=float, parameters={ef_search=24, ef_construction=28, m=12, encoder={name=flat, parameters={}}}}`
So just a bit long.

jmazanec15 (approved):

navneet1v (approved):

PR #2633, opened 2025-03-28
Draft: #2536 Remove custom code for writing indexes outside of compound file…

### LINKED ISSUE #2536: Move away from custom compoundformat
## Description

Recently, for native libraries, we introduced a change to interact with the files via indexinput and indexoutput. With this, we should be able to remove our custom compoundformat in our codec (see https://github.com/opensearch-project/k-NN/pull/2185).

However, when removing it, we get an error like:
```
Caused by: org.apache.lucene.index.CorruptIndexException: compound sub-files must have a valid codec header and footer: codec header mismatch: actual header=1232620912 vs expected header=1071082519 (resource=BufferedChecksumIndexInput(MemorySegmentIndexInput(path="/Users/jmazane/workspace/Opensearch/DockerRunner/k-NN-1/build/testclusters/integTest-0/data/nodes/0/indices/6zG12XzjQLaWyWiAL_OFaQ/0/index/_0_165_test_nested.test_vector.faiss")))
        at org.apache.lucene.codecs.CodecUtil.verifyAndCopyIndexHeader(CodecUtil.java:287) ~[lucene-core-10.1.0.jar:10.1.0 884954006de769dc43b811267230d625886e6515 - 2024-12-17 16:15:44]
```

This is because for the native index files we write a footer but no header: https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexWriter.java#L141-L150. See CompoundFormat interface: https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/codecs/CompoundFormat.java#L41-L45.

We should get rid of the CompoundFormat so we can move towards just extending the PerFieldVectorFormat. To do this, we need to write the header.

From a bwc perspective, for old codecs, we will need to keep around the old CompoundFormat. But we should be able to remove on new codecs.

epalaz:
…s. Implement a NativeIndexReader that can read these indexes from compound files and provide the IndexInput for loading it into the cache

### Description
Deprecating the use of custom compound codec for native index files. Previously, custom codec  [KNN80CompoundFormat](https://github.com/opensearch-project/k-NN/blob/f3f4767c987ea39d903400b3153f9901892a7673/src/main/java/org/opensearch/knn/index/codec/KNN80Codec/KNN80CompoundFormat.java#L25), would take compound files out of the compound file list and write them as standalone files with extensions like `.faissc`, `.hnswc`.

With this CR, we are removing this custom logic and write native indexes into compound files like all other index files. With this, extensions like `.faissc` for compound files are removed. All index files have the same name/index even if they are written with an index where compound files are enabled or not.

In order to enable this, a header is added to all native index files written in NativeIndexWriter since all subfiles in compound file requires a header. A new component NativeIndexReader is implemented to read IndexInput from either compound file or segment directory based on whether compound files are enabled. This component slices the input to remove the header and footer so only native binary index data is loaded into the native index cache.

### Backwards Compatibility
In order to ensure backwards compatibility, a check for Lucene version that is used for the segment is implemented. Since with 3.0 we are moving to Lucene version 10, we assume any index that is written with Lucene version older than 10, will have the older custom codec.

For those, we fallback to existing read logic of reading the custom file and skip header and footer removal.  

### Related Issues
Resolves #2536 

### Check List
- [In progress for full coverage] New functionality includes testing.
- [In progress] New functionality has been documented.


src/main/java/org/opensearch/knn/index/codec/util/KNNCodecUtil.java
```diff
@@ -28,6 +32,10 @@ public class KNNCodecUtil {
     // Floats are 4 bytes in size
     public static final int FLOAT_BYTE_SIZE = 4;
 
+    public static boolean isSegmentUsingLegacyIndexCompoundCodec(SegmentInfo segmentInfo) {
+        return !segmentInfo.getVersion().onOrAfter(Version.LUCENE_10_0_0);
```
0ctopus13prime, line 36: SUPER NIT : could you follow the convention?
`segmeintInfo.getVersion().onOrAfter(..) == false`

src/main/java/org/opensearch/knn/index/memory/NativeMemoryEntryContext.java
```diff
@@ -206,6 +202,7 @@ public void close() {
             if (readStream != null) {
                 try {
                     readStream.close();
```
0ctopus13prime, line 204: Could you use IOUtils.close here?

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriter.java
```diff
@@ -42,7 +42,7 @@
 import static org.opensearch.knn.index.vectorvalues.KNNVectorValuesFactory.getVectorValuesSupplier;
 
 /**
- * A KNNVectorsWriter class for writing the vector data strcutures and flat vectors for Native Engines.
+ * A KNNVectorsWriter class for writing the vector data structures and flat vectors for Native Engines.
```
0ctopus13prime, line 45: lol hawk eye

src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexReader.java
```diff
@@ -0,0 +1,92 @@
...
+                NativeIndexWriter.NATIVE_INDEX_CODEC_NAME,
+                NativeIndexWriter.NATIVE_INDEX_CODEC_MIN_VERSION,
+                NativeIndexWriter.NATIVE_INDEX_CODEC_CURRENT_VERSION,
+                segmentInfo.getId(),
+                NativeIndexWriter.NATIVE_INDEX_CODEC_SUFFIX
+            );
+            long headerLength = CodecUtil.indexHeaderLength(
+                NativeIndexWriter.NATIVE_INDEX_CODEC_NAME, NativeIndexWriter.NATIVE_INDEX_CODEC_SUFFIX
+            );
+            log.info("[KNN] returning sliced input");
```
0ctopus13prime, line 58: Could you fix the log message? we are returning both source IndexInput and a sliced input derived from it.

src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexInput.java
```diff
@@ -0,0 +1,64 @@
...
+
+import org.apache.lucene.store.IndexInput;
+
+/**
+ * NativeIndexInput allows us to pass down single IndexInput for drived index inputs that are cloned from sourc index
+ * inputs through IndexInput.slice method. When cloned/sliced index input is closed, source should be closed as well.
+ * This index input extension allows passing both index inputs down to consumer so when close is called, both
+ * index inputs are closed. See NativeIndexReader for usage of this class.
+ */
+public class NativeIndexInput extends IndexInput {
```
0ctopus13prime, line 18: Is there an use case calling `clone()`?

src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexReader.java
```diff
@@ -0,0 +1,92 @@
...
+import lombok.extern.log4j.Log4j2;
+
+import org.apache.lucene.codecs.CodecUtil;
+import org.apache.lucene.index.SegmentInfo;
+import org.apache.lucene.store.Directory;
+import org.apache.lucene.store.IOContext;
+import org.apache.lucene.store.IndexInput;
+
+@Log4j2
+public class NativeIndexReader implements Closeable {
```
0ctopus13prime, line 22: Could you give brief comments on what this class does?

src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexReader.java
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    public long calculateIndexSize(String indexFileName) throws IOException {
+        if (isSegmentUsingLegacyIndexCompoundCodec(segmentInfo)) {
+            return directory.fileLength(indexFileName);
+        }
+
+        // If it is non-legacy compound file, calculate size from file size - header - footer
+        // Cache will hold the index file bytes after the header hence the size calculation
+        long compoundFileLength = directory.fileLength(indexFileName);
```
0ctopus13prime, line 77: NIT : hmm... this is somewhat misleading? it should be indexFileLength?

src/main/java/org/opensearch/knn/index/codec/util/KNNCodecUtil.java
```diff
@@ -72,25 +80,38 @@ public static long getTotalLiveDocsCount(final BinaryDocValues binaryDocValues)
...
      * Get Engine Files from segment with specific fieldName and engine extension
      *
-     * @param extension Engine extension comes from {@link KNNEngine#getExtension()}}
+     * @param engine object comes from {@link KNNEngine} contains utilities for getting engine file extensions
      * @param fieldName Filed for knn field
      * @param segmentInfo {@link SegmentInfo} One Segment info to use for compute.
      * @return List of engine files
      */
-    public static List<String> getEngineFiles(String extension, String fieldName, SegmentInfo segmentInfo) {
+    public static List<String> getEngineFiles(KNNEngine engine, String fieldName, SegmentInfo segmentInfo) throws IOException {
```
0ctopus13prime, line 88: oh wait, this will be called for every query. 
With this change, then for the compound format, it will involve all sorts of opening a file + decoding per each query which will likely impact search latency.
I think we need to come up with plan to avoid this costs.

- Decoding compound meta info : https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/codecs/lucene90/Lucene90CompoundReader.java#L60
- Query engine files during query : https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/query/KNNWeight.java#L321

src/test/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexReaderTests.java
```diff
@@ -0,0 +1,9 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.knn.index.codec.nativeindex;
+
+public class NativeIndexReaderTests {
```
0ctopus13prime, line 8: haha what is this?

0ctopus13prime (commented):
I have one concern wrt search performance degradation when compound format is enabled. Could you address it?

src/main/java/org/opensearch/knn/index/KNNIndexShard.java
```diff
@@ -89,17 +92,18 @@ public String getIndexName() {
      */
     public void warmup() throws IOException {
         log.info("[KNN] Warming up index: [{}]", getIndexName());
-        final Directory directory = indexShard.store().directory();
         try (Engine.Searcher searcher = indexShard.acquireSearcher("knn-warmup")) {
+            log.info("[KNN] entered searcher");
             getAllEngineFileContexts(searcher.getIndexReader()).forEach((engineFileContext) -> {
+                log.info("[KNN] Engine file context: [{}]", engineFileContext);
```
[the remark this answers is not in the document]
epalaz: I will remove these new info logs, I was using them for tracking some cluster logs.

src/test/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexReaderTests.java
epalaz: Oh yeah, wanted to get general logic verified before finalizing all the tests. I will polish this before actual PR.

src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexReader.java
epalaz: that is true, will update.

src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexInput.java
```diff
@@ -0,0 +1,64 @@
...
+        this.indexInput.seek(l);
+    }
+
+    @Override
+    public long length() {
+        return this.indexInput.length();
+    }
+
+    @Override
+    public IndexInput slice(String s, long l, long l1) throws IOException {
```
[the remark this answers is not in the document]
epalaz: For sure, I will update these on final draft. If we can get away without closing the source IndexInput, we won't need this class anyways.

src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexInput.java
```diff
@@ -0,0 +1,64 @@
...
+
+    public NativeIndexInput(final String name, final IndexInput sliceSourceIndexInput, final IndexInput indexInput) {
+        super(name);
+        this.sliceSourceIndexInput = sliceSourceIndexInput;
+        this.indexInput = indexInput;
+    }
+
+    @Override
+    public void close() throws IOException {
+        this.indexInput.close();
```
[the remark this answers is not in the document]
epalaz: I was having issues with KNNWeightTests which extend LuceneTestCase when it is not closed. It uses `MockDirectoryWrapper` to verify if all files are closed. Without closing the source, tests were failing with file being left opened.

src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexInput.java
```diff
@@ -0,0 +1,64 @@
...
+        return this.indexInput.slice(s, l, l1);
+    }
+
+    @Override
+    public byte readByte() throws IOException {
+        return this.indexInput.readByte();
+    }
+
+    @Override
+    public void readBytes(byte[] bytes, int i, int i1) throws IOException {
```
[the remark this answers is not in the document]
epalaz: For sure, I will update these on final draft. If we can get away without closing the source IndexInput, we won't need this class anyways.

src/main/java/org/opensearch/knn/index/codec/util/KNNCodecUtil.java
epalaz: Great point! Let me see if there is an alternative.

src/main/java/org/opensearch/knn/index/memory/NativeMemoryEntryContext.java
epalaz: What would be the benefit here? null safety?

src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexReader.java
epalaz: This log, I will delete.

src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexInput.java
epalaz: Not in our use case but in terms of Lucene jargon, they call input streams created from cloning/slicing `clones` of the original.

0ctopus13prime: Let's add clone method for completeness here. 
It will give subtle error if relying on the default `clone()`.

src/main/java/org/opensearch/knn/index/memory/NativeMemoryEntryContext.java
0ctopus13prime: `IOUtils.close(readStream, nativeIndexReader);` 

This one will ensure the best efforts to clean up resources even it hits IOException in the middle.

src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexInput.java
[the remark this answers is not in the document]
0ctopus13prime: That's from opensearch. Could you find an example closing sliced IndexInput in Lucene?
Sliced IndexInput is a view, and it does not have to be closed. (Cloned index input is a whole different story though)

Not sure why MockDirectoryWrapper recognizes sliced IndexInput as a separate stream to track. Let me reproduce from my end

src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexReader.java
```diff
@@ -0,0 +1,92 @@
...
+        this.segmentInfo = segmentInfo;
+        if (!segmentInfo.getUseCompoundFile() || isSegmentUsingLegacyIndexCompoundCodec(segmentInfo)) {
+            directory = segmentInfo.dir;
+        } else {
+            // compound directory opens the .cfs file that needs to be closed with directory
+            directory = segmentInfo.getCodec().compoundFormat().getCompoundReader(segmentInfo.dir, segmentInfo);
+        }
+    }
+
+    public static NativeIndexReader getReader(SegmentInfo segmentInfo) throws IOException {
```
shatejas, line 37: I am not sure if we want this reader. this deviates from the construct of a lucene reader which encapsulated in a codec. Shouldn't this logic be encapsulated in **KNN80CompoundReader**? Is there a way we can do it?

Moreover this again circles back to the point of performance @0ctopus13prime bought up. While currently this is being cached it might still impact tail latencies

src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexInput.java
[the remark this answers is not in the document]
epalaz: @shatejas yes, that is the approach I will take in case we don't need it. I am considering some other options as well.

src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexReader.java
shatejas: - Readers and writers are supposed to be inside the codec - so lets not deviate from the convention and we need to rename as it can get really confusing.
- The problem with this approach is that its opening a new compound reader with each request, which underneath opens index input which will assign new virtual address spaces every time. It will eventually run out of addresses spaces with enough load and crash the node making the cluster red. So even if the latencies are okay we might not be able to use this

I looked around and bit and there doesn't seem a way that you can access compound reader which is cached in `SegmentReader`. From what it looks like, Lucene expects that the knnvector reader should be used and a compound directory is passed to it in segment read state. The reader, as a part of opening index input, should check the header and put index input in the right location to start reading the graph file. The links below should help you verify it 
- [SegmentCoreReader caching compound format ](https://github.com/apache/lucene/blob/2201913fe07090f9cfc2f6c6e7c87f979d5f4288/lucene/core/src/java/org/apache/lucene/index/SegmentCoreReaders.java#L83)
- [Segment read state with the right directory (cfs or otherwise)](https://github.com/apache/lucene/blob/2201913fe07090f9cfc2f6c6e7c87f979d5f4288/lucene/core/src/java/org/apache/lucene/index/SegmentCoreReaders.java#L94)
- [Caching of knnvector reader in segment reader](https://github.com/apache/lucene/blob/2201913fe07090f9cfc2f6c6e7c87f979d5f4288/lucene/core/src/java/org/apache/lucene/index/SegmentCoreReaders.java#L137)
- [Lucenes implementation of reader where it checks header](https://github.com/apache/lucene/blob/2201913fe07090f9cfc2f6c6e7c87f979d5f4288/lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsReader.java#L149) 


The right way would be to move doANNSearch in a VectorSearcher which is used in Native99VectorReader

PR #2634, opened 2025-03-28
Add checks to block derived source feature

### LINKED ISSUE #2377: [RFC] Derived Source for Vectors

jmazanec15:
### Description
Adds checks to block derived fields under certain cases. This will prevent source derivation from happening. To accomplish this, I had to create a light wrapper around KnnFloatVectorField and KnnByteVectorField so that the index operation listener can get some of this information from the parsed document.

The checks that are being added are:
1. Is the source enabled
2. Is there a composite index present
3. Is derived source setting enabled
4. If a vector has the copyTo variable (this is broken right now in knn but added for safety in future - see #2636 )
5. Node to node segrep - there are some issues around translog where this will not work and so needs to be blocked.

### Related Issues
Resolves #2377 

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/DerivedKnnByteVectorField.java
+++ b/src/main/java/org/opensearch/knn/index/DerivedKnnByteVectorField.java
@@ -1,0 +1,32 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.knn.index;
+
+import lombok.Getter;
+import org.apache.lucene.document.FieldType;
+import org.apache.lucene.document.KnnByteVectorField;
+
+/**
+ * Wrapper around {@link KnnByteVectorField} so that {@link org.opensearch.knn.index.codec.derivedsource.DerivedSourceIndexOperationListener}
+ * can understand if it needs to modify the source in the translog
+ */
+
+public class DerivedKnnByteVectorField extends KnnByteVectorField {
+
+    @Getter
+    private final boolean isDerivedEnabled;
+
+    /**
+     *
+     * @param name Name of the field
+     * @param vector vector for the field
+     * @param isDerivedEnabled boolean to indicate if derived source is enabled
+     */
+    public DerivedKnnByteVectorField(String name, byte[] vector, FieldType fieldType, boolean isDerivedEnabled) {
+        super(name, vector, fieldType);
+        this.isDerivedEnabled = isDerivedEnabled;
+    }
+}
--- a/src/main/java/org/opensearch/knn/index/codec/KNN10010Codec/KNN10010DerivedSourceStoredFieldsFormat.java
+++ b/src/main/java/org/opensearch/knn/index/codec/KNN10010Codec/KNN10010DerivedSourceStoredFieldsFormat.java
@@ -17,11 +17,11 @@ import org.apache.lucene.store.IOContext;
 import org.opensearch.common.Nullable;
 import org.opensearch.index.mapper.MappedFieldType;
 import org.opensearch.index.mapper.MapperService;
-import org.opensearch.knn.index.KNNSettings;
 import org.opensearch.knn.index.codec.derivedsource.DerivedFieldInfo;
 import org.opensearch.knn.index.codec.derivedsource.DerivedSourceReadersSupplier;
 import org.opensearch.knn.index.codec.derivedsource.DerivedSourceSegmentAttributeParser;
 import org.opensearch.knn.index.mapper.KNNVectorFieldType;
+import org.opensearch.knn.index.util.IndexUtil;
 
 import java.io.IOException;
 import java.util.ArrayList;
@@ -66,14 +66,17 @@ public class KNN10010DerivedSourceStoredFieldsFormat extends StoredFieldsFormat 
     @Override
     public StoredFieldsWriter fieldsWriter(Directory directory, SegmentInfo segmentInfo, IOContext ioContext) throws IOException {
         StoredFieldsWriter delegateWriter = delegate.fieldsWriter(directory, segmentInfo, ioContext);
-        if (mapperService == null || KNNSettings.isKNNDerivedSourceEnabled(mapperService.getIndexSettings().getSettings()) == false) {
+        if (IndexUtil.isDerivedEnabledForIndex(mapperService) == false) {
             return delegateWriter;
         }
-
         List<String> vectorFieldTypes = new ArrayList<>();
         List<String> nestedVectorFieldTypes = new ArrayList<>();
         for (MappedFieldType fieldType : mapperService.fieldTypes()) {
-            if (fieldType instanceof KNNVectorFieldType) {
+            if (fieldType instanceof KNNVectorFieldType knnVectorFieldType) {
+                if (IndexUtil.isDerivedEnabledForField(knnVectorFieldType, mapperService) == false) {
+                    continue;
+                }
+
                 boolean isNested = mapperService.documentMapper().mappers().getNestedScope(fieldType.name()) != null;
                 if (isNested) {
                     nestedVectorFieldTypes.add(fieldType.name());
--- a/src/main/java/org/opensearch/knn/index/codec/derivedsource/DerivedSourceIndexOperationListener.java
+++ b/src/main/java/org/opensearch/knn/index/codec/derivedsource/DerivedSourceIndexOperationListener.java
@@ -6,8 +6,6 @@
 package org.opensearch.knn.index.codec.derivedsource;
 
 import lombok.extern.log4j.Log4j2;
-import org.apache.lucene.document.KnnByteVectorField;
-import org.apache.lucene.document.KnnFloatVectorField;
 import org.apache.lucene.index.IndexableField;
 import org.apache.lucene.util.BytesRef;
 import org.opensearch.common.collect.Tuple;
@@ -22,6 +20,8 @@ import org.opensearch.core.xcontent.XContentBuilder;
 import org.opensearch.index.engine.Engine;
 import org.opensearch.index.mapper.ParseContext;
 import org.opensearch.index.shard.IndexingOperationListener;
+import org.opensearch.knn.index.DerivedKnnByteVectorField;
+import org.opensearch.knn.index.DerivedKnnFloatVectorField;
 import org.opensearch.knn.index.VectorDataType;
 import org.opensearch.knn.index.mapper.KNNVectorFieldMapperUtil;
 
@@ -51,12 +51,18 @@ public class DerivedSourceIndexOperationListener implements IndexingOperationLis
         if (isRecoverySourceEnabled(operation)) {
             return operation;
         }
+
+        Function<Map<String, Object>, Map<String, Object>> injectTransformer = createInjectTransformer(operation);
+        if (injectTransformer == null) {
+            return operation;
+        }
+
         Tuple<? extends MediaType, Map<String, Object>> originalSource = XContentHelper.convertToMap(
             operation.parsedDoc().source(),
             true,
             operation.parsedDoc().getMediaType()
         );
-        Map<String, Object> derivedSource = createInjectTransformer(operation).apply(originalSource.v2());
+        Map<String, Object> derivedSource = injectTransformer.apply(originalSource.v2());
 
         try (BytesStreamOutput bStream = new BytesStreamOutput();) {
             XContentBuilder builder = MediaTypeRegistry.contentBuilder(originalSource.v1(), bStream).map(derivedSource);
@@ -75,17 +81,22 @@ public class DerivedSourceIndexOperationListener implements IndexingOperationLis
         for (ParseContext.Document document : operation.parsedDoc().docs()) {
             for (Iterator<IndexableField> it = document.iterator(); it.hasNext();) {
                 IndexableField indexableField = it.next();
-                if (indexableField instanceof KnnFloatVectorField knnVectorFieldType) {
+                if (indexableField instanceof DerivedKnnFloatVectorField knnVectorFieldType && knnVectorFieldType.isDerivedEnabled()) {
                     injectedVectors.computeIfAbsent(indexableField.name(), k -> new ArrayList<>())
                         .add(formatVector(VectorDataType.FLOAT, knnVectorFieldType.vectorValue()));
                 }
 
-                if (indexableField instanceof KnnByteVectorField knnByteVectorField) {
+                if (indexableField instanceof DerivedKnnByteVectorField knnByteVectorField && knnByteVectorField.isDerivedEnabled()) {
                     injectedVectors.computeIfAbsent(indexableField.name(), k -> new ArrayList<>())
                         .add(formatVector(VectorDataType.BYTE, knnByteVectorField.vectorValue()));
                 }
             }
         }
+
+        if (injectedVectors.isEmpty()) {
+            return null;
+        }
+
         Map<String, Function<Object, Object>> injectTransformers = new HashMap<>();
         for (Map.Entry<String, List<Object>> entry : injectedVectors.entrySet()) {
             Iterator<Object> iterator = entry.getValue().iterator();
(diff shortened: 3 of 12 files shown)


src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java
```diff
@@ -779,6 +781,46 @@ protected void parseCreateField(ParseContext context, int dimension, VectorDataT
...
     }
 
+    private boolean isDerivedEnabled(ParseContext parseContext) {
+        if (isDerivedSourceEnabled == null) {
+            isDerivedSourceEnabled = isDerivedEnabledForIndex(parseContext) && isDerivedEnabledForField();
+        }
+        return isDerivedSourceEnabled;
+    }
+
+    private boolean isDerivedEnabledForIndex(ParseContext parseContext) {
```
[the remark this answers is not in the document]
jmazanec15: Let me see if I can refactor to share.

src/main/java/org/opensearch/knn/index/codec/KNN10010Codec/KNN10010DerivedSourceStoredFieldsFormat.java
```diff
@@ -96,4 +100,37 @@ public StoredFieldsWriter fieldsWriter(Directory directory, SegmentInfo segmentI
...
+    private boolean isEnabledForIndex() {
+        if (mapperService == null) {
+            return false;
+        }
+        // If there are excluded fields, we will not derive any vector fields
+        if (mapperService.documentMapper().sourceMapper().enabled() == false) {
+            return false;
+        }
+        // Skipping composite indices for now
+        if (mapperService.isCompositeIndexPresent() == true) {
```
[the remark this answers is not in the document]
jmazanec15: sure - will add

src/main/java/org/opensearch/knn/index/codec/derivedsource/DerivedSourceIndexOperationListener.java
```diff
@@ -75,17 +81,23 @@ private Function<Map<String, Object>, Map<String, Object>> createInjectTransform
         for (ParseContext.Document document : operation.parsedDoc().docs()) {
             for (Iterator<IndexableField> it = document.iterator(); it.hasNext();) {
                 IndexableField indexableField = it.next();
-                if (indexableField instanceof KnnFloatVectorField knnVectorFieldType) {
+                if (indexableField instanceof DerivedKnnFloatVectorField knnVectorFieldType
+                    && knnVectorFieldType.isDerivedSourceEnabled()) {
```
shatejas, line 85: Whats the reason to check derivedSource here, are there any backward compatibility concerns? Injecting the corrected vector for all vector cases keeps the translog consistent and the source consistent right?

src/main/java/org/opensearch/knn/index/mapper/DerivedSourceFieldType.java
```diff
@@ -0,0 +1,18 @@
...
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.knn.index.mapper;
+
+/**
+ * This interface is used to indicate if derived source is enabled for a field type
+ */
+public interface DerivedSourceFieldType {
```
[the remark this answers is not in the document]
jmazanec15: Yeah youre probably right. I dont think it is necessary.

src/main/java/org/opensearch/knn/index/codec/derivedsource/DerivedSourceIndexOperationListener.java
jmazanec15: Yes the purpose of injecting is to ensure that the translog is consistent during replica replay. By injecting here, we keep it consistent (except for the possibility that the key ordering gets out of order, but this shouldnt happen because the maps are ordered).

The old version of the feature is broken, so its not really breaking bwc.

src/main/java/org/opensearch/knn/index/DerivedKnnByteVectorField.java
```diff
@@ -0,0 +1,32 @@
...
+import lombok.Getter;
+import org.apache.lucene.document.FieldType;
+import org.apache.lucene.document.KnnByteVectorField;
+
+/**
+ * Wrapper around {@link KnnByteVectorField} so that {@link org.opensearch.knn.index.codec.derivedsource.DerivedSourceIndexOperationListener}
+ * can understand if it needs to modify the source in the translog
+ */
+
+public class DerivedKnnByteVectorField extends KnnByteVectorField {
```
Vikasht34, line 17: Do we need these two classes .

DerivedKnnByteVectorField and DerivedKnnFloatVectorField

Here is my suggestion , let me know if it's fesible.

Can we use IndexingOperationListener to Use Mapping Metadata

Use the MapperService to check if a field has derived_source_enabled in its KNNVectorFieldType.

We need to inject MapperService into DerivedSourceIndexOperationListener, which is not currently done.

Then, instead of checking instanceof DerivedKnnFloatVectorField, you can just check:

```
mapperService.documentMapper().mappers().getMapper(fieldName) instanceof KNNVectorFieldMapper

```

jmazanec15: Thats an interesting idea. The reason I went with this approach, howeverm is that I couldnt figure out a way to inject mapper service into the index operation listener from the IndexModule.

src/main/java/org/opensearch/knn/index/codec/derivedsource/DerivedSourceIndexOperationListener.java
shatejas: Was wondering if we should rely on index version in that case. We don't want to rely on feature support, any other feature can run into this in the future. From 3.0 we keep it consistent irrespective of the feature - thoughts? and is it possible?

jmazanec15: There is a benefit for uniformly keeping translog/source in same format regardless of feature enabled or disabled. However the downsides are:
1. There are some features in OpenSearch which will not parse the source (segrep node-to-node). So, in order to continue supporting segrep node to node, we would need to skip modifying the source, which then makes it feature dependent again.
2. This would lead to a performance regression when derived source is not enabled. This may cause problems for users.

For those 2 reasons, I think its best to feature gate it.

shatejas (approved):

Vikasht34 (approved):

PR #2637, opened 2025-04-01
Optimize first child lookup for nested docs

jmazanec15:
### Description
Optimization for derived source. The only changed source file is `src/main/java/org/opensearch/knn/index/codec/derivedsource/DerivedSourceLuceneHelper.java`. The rest are tests. 

For nested documents, in order to initialize the iterators properly, we need to find the first child document of the parent document. To do this, we can scan the numeric doc values for the _primary_term field, which is only present for parent documents. In the code now, we start this scan at 0. This requires reading all of the numeric doc values. This code introduces an optimization where instead of starting at 0, it first tries 150 documents before the parentdoc, then 1500, then 6000, then tries from 0.

Along with this, I updated some testing logic.

### Related Issues

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/codec/derivedsource/DerivedSourceLuceneHelper.java
+++ b/src/main/java/org/opensearch/knn/index/codec/derivedsource/DerivedSourceLuceneHelper.java
@@ -5,6 +5,7 @@
 
 package org.opensearch.knn.index.codec.derivedsource;
 
+import com.google.common.annotations.VisibleForTesting;
 import lombok.RequiredArgsConstructor;
 import org.apache.lucene.index.FieldInfo;
 import org.apache.lucene.index.NumericDocValues;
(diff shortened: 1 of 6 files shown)


src/main/java/org/opensearch/knn/index/codec/derivedsource/DerivedSourceLuceneHelper.java
```diff
@@ -36,16 +50,39 @@ public int getFirstChild(int parentDocId) throws IOException {
         if (parentDocId == 0) {
             return NO_MORE_DOCS;
         }
+        int lastStartingPoint = -1;
+        int currentStartingPoint;
```
Vikasht34, line 54: Should be inside loop.

Vikasht34 (approved):
Just one nit :- Overall Looks good.

shatejas (approved):

Vikasht34 (approved):

Issue #2511, opened 2025-02-10
[BUG] usage of nested knn query and efficient filter with nested query is not correct


wdongyu:
**What is the bug?**
From https://github.com/opensearch-project/k-NN/issues/1356, when we need to execute a nested knn query and an efficient filter with a nested field, we are advised to use query like this:

```
{
  "query": {
    "nested": {
      "path": "test_nested",
      "query": {
        "knn": {
          "test_nested.test_vector": {
            "vector": [
              5
            ],
            "k": 24,
            "filter": {
              "nested": {
                "path": "test_nested",
                "query": {
                  "term": {
                    "test_nested.parking": "false"
                  }
                }
              }
            }
          }
        }
      },
      "inner_hits": {}
    }
  }
}'
```

But indeed, it will return wrong number of hits. When I use the following query, it works as expected:
```
{
  "query": {
    "nested": {
      "path": "test_nested",
      "query": {
        "knn": {
          "test_nested.test_vector": {
            "vector": [
              5
            ],
            "k": 24,
            "filter": {
              "term": {
                "test_nested.parking": "false"
              }
            }
          }
        }
      },
      "inner_hits": {}
    }
  }
}
```

**How can one reproduce the bug?**
1. Base on testcase [AdvancedFilteringUseCasesIT.testFiltering_whenNestedKNNAndFilterFieldWithNestedQueries_thenSuccess](https://github.com/opensearch-project/k-NN/blob/2.18.0.0/src/test/java/org/opensearch/knn/index/AdvancedFilteringUseCasesIT.java#L105), we first increase k to 100 as we need to match more nested knn docs in all(NUM_DOCS=50) parent docs:
```
private static final int k = 100;
```
2. [In final validation](https://github.com/opensearch-project/k-NN/blob/2.18.0.0/src/test/java/org/opensearch/knn/index/AdvancedFilteringUseCasesIT.java#L450), change totalSearchHits to NUM_DOCS/2 = 25, as there are half docs that meet the requirements:
```
Assert.assertEquals("For engine " + engine + ", totalSearchHits: ", NUM_DOCS / 2, parseTotalSearchHits(response));
if (KNNEngine.getEngine(engine) == KNNEngine.FAISS) {
    ...
    Assert.assertEquals("For engine " + engine + ", totalSearchHits with ANN search :", NUM_DOCS / 2, parseTotalSearchHits(response));
    ...
}
```
3. Run the test and it fail, but other tests in the same IT pass:
```
[...]

PR #2641, opened 2025-04-02
[BUGFIX] FIX nested vector query at efficient filter scenarios

luyuncheng:
### Description
#2511 there is an bug for efficient filter at nested vector query. i think the main idea is when vector is nested. OpenSearch would go into `NestedQueryBuilder#doToQuery` at https://github.com/opensearch-project/OpenSearch/blob/a586a6202853c15b8aa295550469b0dff418d588/server/src/main/java/org/opensearch/index/query/NestedQueryBuilder.java#L302-L332

so when the vector is nested, and efficient filter is also nested, it would get into this twice, and changed parentFilter which make efficient filter get wrong state as following shows:

```
        ObjectMapper objectMapper = context.nestedScope().getObjectMapper();
        if (objectMapper == null) {
            parentFilter = context.bitsetFilter(Queries.newNonNestedFilter());
        } else {
            /**second nested go into here, so nested with wrong parent filter**/
            parentFilter = context.bitsetFilter(objectMapper.nestedTypeFilter());
        }
```

#Proposal

1. When nested vector query with nested filter query, we need preserve the nestedLevelStack, and then parse efficient nested query. this would be correctly
2. When nested vector query with bool nested filter query, also need parse the boolean query tree, and rewrite it.

### Related Issues
#2511 

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/query/BaseQueryFactory.java
+++ b/src/main/java/org/opensearch/knn/index/query/BaseQueryFactory.java
@@ -104,9 +104,11 @@ public abstract class BaseQueryFactory {
         // preserve nestedStack
         Deque<ObjectMapper> nestedLevelStack = new LinkedList<>();
         ObjectMapper objectMapper = null;
-        while ((objectMapper = queryShardContext.nestedScope().getObjectMapper()) != null) {
-            nestedLevelStack.push(objectMapper);
-            queryShardContext.nestedScope().previousLevel();
+        if (queryShardContext.nestedScope() != null) {
+            while ((objectMapper = queryShardContext.nestedScope().getObjectMapper()) != null) {
+                nestedLevelStack.push(objectMapper);
+                queryShardContext.nestedScope().previousLevel();
+            }
         }
 
         final Query filterQuery;


src/test/java/org/opensearch/knn/index/AdvancedFilteringUseCasesIT.java
```diff
@@ -58,7 +58,7 @@ public class AdvancedFilteringUseCasesIT extends KNNRestTestCase {
 
     private static final String TERM_FIELD = "term";
 
-    private static final int k = 20;
```
jmazanec15, line 61: Can we add a test that validates this fix works?

src/main/java/org/opensearch/knn/index/query/BaseQueryFactory.java
```diff
@@ -90,20 +100,71 @@ protected static Query getFilterQuery(BaseQueryFactory.CreateQueryRequest create
                 createQueryRequest.getFieldName()
             )
         );
+
+        // preserve nestedStack
+        Deque<ObjectMapper> nestedLevelStack = new LinkedList<>();
+        ObjectMapper objectMapper = null;
+        while ((objectMapper = queryShardContext.nestedScope().getObjectMapper()) != null) {
+            nestedLevelStack.push(objectMapper);
+            queryShardContext.nestedScope().previousLevel();
```
jmazanec15, line 109: Can we call this in line 107 to set the object mapper?

src/main/java/org/opensearch/knn/index/query/BaseQueryFactory.java
```diff
@@ -90,20 +100,71 @@ protected static Query getFilterQuery(BaseQueryFactory.CreateQueryRequest create
...
+
         final Query filterQuery;
         try {
             filterQuery = createQueryRequest.getFilter().get().toQuery(queryShardContext);
         } catch (IOException e) {
             throw new RuntimeException("Cannot create query with filter", e);
+        } finally {
+            while ((objectMapper = nestedLevelStack.peek()) != null) {
+                queryShardContext.nestedScope().nextLevel(objectMapper);
+                nestedLevelStack.pop();
```
jmazanec15, line 120: Can we call this instead of peek in while loop?

jmazanec15:
Thanks @luyuncheng. Also, @heemin32 do you have any concerns with this - I know you recently added functionality around nested queries.

> When nested vector query with nested filter query, we need preserve the nestedLevelStack, and then parse efficient nested query. this would be correctly

This makes sense to me.

> When nested vector query with bool nested filter query, also need parse the boolean query tree, and rewrite it.

Can you expand upon this?

heemin32:
> Also, @heemin32 do you have any concerns with this - I know you recently added functionality around nested queries.

As long as we have full coverage for various cases in testing, I am fine. 
The efficient filtering is added by @navneet1v so I don't have much context around it tbh.

luyuncheng:
>> When nested vector query with bool nested filter query, also need parse the boolean query tree, and rewrite it.

>Can you expand upon this?

@jmazanec15 like the following query with boolean query: we need change parent field term query to parent to child also 
 the nested path `test_nested.parking` query which have the same nested docID

```
{
  "query": {
    "nested": {
      "path": "test_nested",
      "query": {
        "knn": {
          "test_nested.test_vector": {
            "vector": [
              5
            ],
            "k": 24,
            "filter": {
              "bool": {
                "must": [
                  {
                    "term": {
                      "parent_field": "xxx"
                    }
                  },
                  {
                    "nested": {
                      "path": "test_nested",
                      "query": {
                        "term": {
                          "test_nested.parking": "false"
                        }
                      }
                    }
                  }
                ]
              }
            }
          }
        },
        "inner_hits": {}
      }
    }
  }
}
```
so i added https://github.com/opensearch-project/k-NN/blob/00adf469c865309ecb65410d4ecdb1e122a009dc/src/main/java/org/opensearch/knn/index/query/BaseQueryFactory.java#L131-L141

src/test/java/org/opensearch/knn/index/AdvancedFilteringUseCasesIT.java
luyuncheng: in #2511 it shows how to reproduce the bug, so i changed k to `NUM_DOCS / 2` as https://github.com/opensearch-project/k-NN/blob/00adf469c865309ecb65410d4ecdb1e122a009dc/src/test/java/org/opensearch/knn/index/AdvancedFilteringUseCasesIT.java#L459-L463

src/main/java/org/opensearch/knn/index/query/BaseQueryFactory.java
luyuncheng: No, because `NestedScope#levelStack` is a `Deque` object. when call `previousLevel` directly in while condition, it would throw `NoSuchElementException` when deque is empty.

src/main/java/org/opensearch/knn/index/query/BaseQueryFactory.java
luyuncheng: same as https://github.com/opensearch-project/k-NN/pull/2641#discussion_r2039669066

jmazanec15 (approved):
LGTM - thanks @luyuncheng  - can you rebase/fix CI?

navneet1v:
@luyuncheng change looks good to me and thanks for catching the issue. Once all CIs are successful I can approve the PR.

luyuncheng:
@jmazanec15 @navneet1v  Thanks for the review, all CIs are passed

navneet1v (approved):

jmazanec15 (approved):

jmazanec15:
Thanks @luyuncheng !

PR #2646, opened 2025-04-03
Combine method and lucene mappers to EngineFieldMapper

### LINKED ISSUE #2631: Merge MethodFieldMapper and LuceneFieldMapper into EngineFieldMapper
## Description

Currently, we have different field mappers for lucene and faiss/nmslib - the MethodFieldMapper and the LuceneFieldMapper. As a part of #2605, I think we should merge these into EngineFieldMapper. With that, we would have 3 types of vector field mappers
1. EngineFieldMapper - creating a knn index configuration with full engine specifications
2. ModelFieldMapper - creating a knn index from a pre-trained model
3. FlatFieldMapper - creating a knn index that doesnt do any special ANN indexing

LuceneFieldMapper and MethodFieldMapper have a lot of duplicate code. This will make it easier to manage and build engine extension points.

### LINKED ISSUE #2640: Remove doc values for lucene engine
## Description

I added store size in dashboards in https://benchmarks.opensearch.org/. Surprisingly, the size of the lucene engine is same as faiss, despite faiss storing the vecs in the .vec file and the .faiss file, and lucene not. 

The reasoning for this is that we add a doc values field for lucene as well as a vector field. This happens in the LuceneFieldMapper. The reasoning behind this was initially to support scripting, but that is no longer needed, because we can script with the .vec files now.

Thus, we should disable the doc values fields for the lucene engine fields. Further, to prevent inconsistencies like this in the future, we should get rid of LuceneFieldMapper and refactor MethodFieldMapper into an EngineFieldMapper that includes lucene. Related #2631.

kotwanikunal:
### Description
- Merges FieldMappers for Native engines and Lucene into a single EngineFieldMapper
- Change the default value for docValues 

### Related Issues
Resolves #2640
Resolves #2631 

### Check List


src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java
```diff
@@ -325,7 +304,7 @@ public KNNVectorFieldMapper build(BuilderContext context) {
                 copyToBuilder,
                 ignoreMalformed,
                 stored.getValue(),
-                hasDocValues.getValue(),
+                hasDocValues.isConfigured() ? hasDocValues.getValue() : false,
```
kotwanikunal, line 307: Not changing the default but adding in a check for explicit value setting on the mapping. 
@jmazanec15 WDYT?

jmazanec15: I would say just set it to false - is there a bwc concern?

src/main/java/org/opensearch/knn/index/mapper/EngineFieldMapper.java
```diff
@@ -0,0 +1,285 @@
...
+
+        // Common Attributes
+        this.perDimensionProcessor = knnLibraryIndexingContext.getPerDimensionProcessor();
+        this.perDimensionValidator = knnLibraryIndexingContext.getPerDimensionValidator();
+        this.vectorValidator = knnLibraryIndexingContext.getVectorValidator();
+    }
+
+    @Override
+    protected List<Field> getFieldsForFloatVector(final float[] array, boolean isDerivedSourceEnabled) {
+        if (this.isLuceneEngine) {
```
jmazanec15, line 226: Do we need to branch on this still?

src/main/java/org/opensearch/knn/index/mapper/EngineFieldMapper.java
```diff
@@ -0,0 +1,285 @@
...
+        if (isLuceneEngine) {
+            return super.getVectorTransformer();
+        }
+        return vectorTransformer;
+    }
+
+    @Override
+    void updateEngineStats() {
+        if (this.isLuceneEngine) {
+            KNNEngine.LUCENE.setInitialized(true);
```
jmazanec15, line 282: Can we just initialize all of them?

kotwanikunal: Do you mean based on the mapper engine or just generally set all to true?

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java
kotwanikunal: I was chewing a bit on that - thinking of BwC in general. From the code flow, it does not look like LuceneFieldMapper and MethodFieldMapper are serialized directly.
We use fromXContent and toXContent - which specifies all the initialized values.
I dont think we will break BwC with it.

Changing the default would entail changing defaults for FlatVectorField and ModelFieldMapper as well.

With this approach - I only change it for the general engine mappers.

src/main/java/org/opensearch/knn/index/mapper/EngineFieldMapper.java
```diff
@@ -0,0 +1,285 @@
...
+                }
+
+                @Override
+                public Version getIndexCreatedVersion() {
+                    return knnMethodConfigContext.getVersionCreated();
+                }
+
+                @Override
+                public QuantizationConfig getQuantizationConfig() {
+                    return isLuceneEngine ? QuantizationConfig.EMPTY : libraryContext.getQuantizationConfig();
```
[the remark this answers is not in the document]
kotwanikunal: Since we are overriding the method - the defaults have to be returned explicitly. Let me clean it up to `if != null -> return config`

src/main/java/org/opensearch/knn/index/mapper/EngineFieldMapper.java
kotwanikunal: We do have differences in Lucene fields structure around docValues and derivedSource - 

Lucene 
```
final List<Field> fields = new ArrayList<>();
fields.add(new DerivedKnnFloatVectorField(name(), array, fieldType, isDerivedSourceEnabled));
if (hasDocValues && vectorFieldType != null) {  <- Difference here
    fields.add(new VectorField(name(), array, vectorFieldType));
}
if (stored) {
    fields.add(createStoredFieldForFloatVector(name(), array));
}
return fields;
```


Method/Native
```

 final List<Field> fields = new ArrayList<>();
  fields.add(createVectorField(array, isDerivedEnabled));
  if (this.stored) {
      fields.add(createStoredFieldForFloatVector(name(), array));
  }
  return fields;

private Field createVectorField(float[] vectorValue, boolean isDerivedEnabled) {
  if (useLuceneBasedVectorField) {  <- Difference here
      return new DerivedKnnFloatVectorField(name(), vectorValue, fieldType, isDerivedEnabled);
  }
  return new VectorField(name(), vectorValue, fieldType);
}
```

kotwanikunal: We can streamline this with a followup PR.

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java
kotwanikunal: Updated the default to false except for flat fields.

src/main/java/org/opensearch/knn/index/mapper/EngineFieldMapper.java
jmazanec15: Like engine.setInitialized to true

src/main/java/org/opensearch/knn/index/mapper/EngineFieldMapper.java
jmazanec15: Sure - lets take in next PR

src/main/java/org/opensearch/knn/index/mapper/EngineFieldMapper.java
```diff
@@ -0,0 +1,287 @@
...
+                throw new RuntimeException(String.format("Unable to create KNNVectorFieldMapper: %s", ioe), ioe);
+            }
+
+            if (useLuceneBasedVectorField) {
+                int adjustedDimension = mappedFieldType.vectorDataType == VectorDataType.BINARY
+                    ? knnMappingConfig.getDimension() / 8
+                    : knnMappingConfig.getDimension();
+                final VectorEncoding encoding = mappedFieldType.vectorDataType == VectorDataType.FLOAT
+                    ? VectorEncoding.FLOAT32
+                    : VectorEncoding.BYTE;
```
shatejas, line 206: nit: I think this can be encapsulated in vectorDataType

src/main/java/org/opensearch/knn/index/mapper/EngineFieldMapper.java
```diff
@@ -0,0 +1,287 @@
...
+        KNNLibraryIndexingContext knnLibraryIndexingContext = resolvedKnnMethodContext.getKnnEngine()
+            .getKNNLibraryIndexingContext(resolvedKnnMethodContext, knnMethodConfigContext);
+
+        // LuceneFieldMapper attributes
+        if (this.isLuceneEngine) {
+            this.fieldType = vectorDataType.createKnnVectorFieldType(knnMappingConfig.getDimension(), knnVectorSimilarityFunction);
+
+            if (this.hasDocValues) {
+                this.vectorFieldType = buildDocValuesFieldType(resolvedKnnMethodContext.getKnnEngine());
+            } else {
```
shatejas, line 171: Is this kept because user can still have `doc_values = true` in the mapping? Or is it some mapping recovery case for bwc?

shatejas (dismissed):
Looks good overall

src/main/java/org/opensearch/knn/index/mapper/EngineFieldMapper.java
kotwanikunal: Updated.

src/main/java/org/opensearch/knn/index/mapper/EngineFieldMapper.java
kotwanikunal: I can take this up as a part of the cleanup/logic consolidation for Lucene/MethodField mappers.

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java
```diff
@@ -330,6 +309,23 @@ public KNNVectorFieldMapper build(BuilderContext context) {
...
+         * defaults to false when not explicitly configured. This reduces storage
+         * overhead and improves indexing performance for k-NN vector fields.
+         * Changing the default value breaks BwC for existing indices on a cluster.
+         *
+         * Behavior matrix:
+         * - Index < 3.0.0: Uses original default value
+         * - Index >= 3.0.0, docValues not configured: Sets to false
+         * - Any version, docValues explicitly configured: Respects configured value
+         */
+        private void updateDocValueDefaults() {
```
kotwanikunal, line 323: An alternative for this approach is to not use the default OpenSearch parameter `Parameter.docValuesParam` and override it. This is because we need to add in a version check for BwC. Sample implementation - 

```
protected final Parameter<Boolean> hasDocValues = new Parameter<Boolean>("doc_values", false, () -> false, (n, c, o) -> {
            if (o == null) {
                // Check the index version and set appropriate default
                Version indexVersion = c.indexVersionCreated();
                if (indexVersion.onOrAfter(Version.V_3_0_0)) {
                    return false;
                }
                return true;
            }
            return XContentMapValues.nodeBooleanValue(o);
        }, m -> toType(m).hasDocValues);
```

This is cleaner but will lead to deviations from the core code.

jmazanec15: What core code? I think this looks good

kotwanikunal: The core code has the pre defined `Parmeter.docValuesParam` which keeps the constants/types configured for us, rather than us having to redefine it - given "doc_values" is a core construct.

On a further dive, this is not feasible. The key thing that I realized is default values are not serialized and we need to ensure we pick the right default value for our cases.
The current code changes that I implemented add in support for contextual default value picking based on the version of the index as well as the branching logic for the different mappers we support. This will ensure that we only default it to `false` for indexes created on or after 3.0 and keeps the `true` value for FlatFieldMapper.

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java
```diff
@@ -330,6 +329,23 @@ public KNNVectorFieldMapper build(BuilderContext context) {
...
+            // if (indexCreatedVersion.onOrAfter(Version.V_3_0_0) && hasDocValues.isConfigured() == false) {
+            // hasDocValues.setValue(false);
+            // }
```
jmazanec15, line 346: Remove commented code?

jmazanec15 (commented):
Overall looks good - some minor comments

src/test/java/org/opensearch/knn/index/FaissIT.java
```diff
@@ -1387,6 +1390,7 @@ private void createIndexFromModelAndIngestDocuments(String indexName, String mod
             .startObject(FIELD_NAME)
             .field("type", "knn_vector")
             .field("model_id", modelId)
+            .field("doc_values", false)
```
jmazanec15, line 1393: Why do we need this?

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java
kotwanikunal: Oops. 
Updating this.

shatejas (dismissed):

src/main/java/org/opensearch/knn/index/mapper/EngineFieldMapper.java
```diff
@@ -0,0 +1,286 @@
...
+        String name,
+        KNNVectorFieldType mappedFieldType,
+        MultiFields multiFields,
+        CopyTo copyTo,
+        Explicit<Boolean> ignoreMalformed,
+        boolean stored,
+        boolean hasDocValues,
+        KNNMethodConfigContext knnMethodConfigContext,
+        OriginalMappingParameters originalMappingParameters,
+        boolean isLuceneEngine
```
navneet1v, line 141: this defeats the purpose of having a single FieldMapper. If we are not getting rid of this boolean flag, then I would suggest lets not have a single EngineFieldMapper class. Because this boolean flag will create bifercations in the code, and then tomorrow if we bring in a new engine then will we will be creating a new boolean?

navneet1v (changes requested):

src/main/java/org/opensearch/knn/index/mapper/EngineFieldMapper.java
kotwanikunal: I have that as a followup task. The initial commit here just lays out the structure and default values.

navneet1v (approved):
Approving this PR as @kotwanikunal promised to improve the code for EngineFieldMapper and we need to merge this change for 3.0 release to ensure that data redundancy is removed for Lucene Engine which is coming via doc_values.

Issue #2649, opened 2025-04-07
[BUG] Invalid scoring during exact search when memory optimized search is enabled.


0ctopus13prime:
**What is the bug?**
When memory optimized search is enabled, a filtering is provided in a query, it will evaluate whether to fallback to exact search to preserve high recall.
During the exact search, Lucene will load `FloatVectorValues` and `ByteVectorValues` from configured VectorReader, which is NativeEngines990KnnVectorsReader at the moment - [Code](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java#L107)

Currently, regardless of the space type in mapping, the default space type will be saved in `.vec` file - [Code](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/mapper/MethodFieldMapper.java#L180)

We should change it to find the best matching similarity function with the best effort so that it will pick up the correct similarity function during exact search when memory optimized is enabled.
Hamming space type is the only FAISS exclusive one that does not support the corresponding Lucene similarity function. But this space type is not the scope of M1 milestone of LuceneOnFaiss project.

Also, this change will not bring side effect to the case where the mode is turned off, where delegating vector search to C++ layer, as [ExactSearcher](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/query/ExactSearcher.java#L155) has a separate scoring phase, and it is relying on scorer extracted from Codec at the moment.

**How can one reproduce the bug?**
1. Index 1000 vectors, with 128 dimensions and inner product as a space type.
2. Turn on memory optimized search.
3. Query with a filter that filters 1% of docs.
4. It will likely to fallback to exact search.
5. Check the recall.

**What is the expected behavior?**
Scorer should be loaded according to the configured space type in mapping, which is using EUCLIDIAN always. For example, for inner product space type, then maximum inner product similarity function should be used.

**What is your host/environment?**
 - OS: All
 - Version : 3.0
 - Plugins : KNN

**Do you have any screenshots?**
N/A

**Do you have any additional context?**
N/A

0ctopus13prime:
Turns out, we can't just find best matching Lucene vector similarity function in field type.
It will fail during version upgrade where it has Euclidian similarity function but it tries to merge to a new field info having inner product. 
Will try to find another way.

0ctopus13prime:
[...]

PR #2650, opened 2025-04-07
[Bug fix LuceneOnFaiss] Having field info to have a corresponding Lucene vector similarity function.

0ctopus13prime:
### Description
When memory optimized search is enabled, if a filtering is provided in a query, it will evaluate whether to fallback to exact search to preserve high recall.
During the exact search, Lucene will load FloatVectorValues and ByteVectorValues from configured VectorReader, which is NativeEngines990KnnVectorsReader at the moment - [Code](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java#L107)

Currently, regardless of the space type in mapping, the default space type will be saved in .vec file - [Code](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/mapper/MethodFieldMapper.java#L180)

We should change it to find the best matching similarity function with the best effort so that it will pick up the correct similarity function during exact search when memory optimized is enabled.
Hamming space type is the only FAISS exclusive one that does not support the corresponding Lucene similarity function. But this space type is not the scope of M1 milestone of LuceneOnFaiss project.

Also, this change will not bring side effect to the case where the mode is turned off, where delegating vector search to C++ layer, as [ExactSearcher](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/query/ExactSearcher.java#L155) has a separate scoring phase, and it is relying on scorer extracted from Codec at the moment.

FYI : LuceneOnFaiss RFC : https://github.com/opensearch-project/k-NN/issues/2401

### Related Issues
https://github.com/opensearch-project/k-NN/issues/2649

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/mapper/EngineFieldMapper.java
+++ b/src/main/java/org/opensearch/knn/index/mapper/EngineFieldMapper.java
@@ -9,6 +9,7 @@ import org.apache.lucene.document.Field;
 import org.apache.lucene.document.FieldType;
 import org.apache.lucene.index.DocValuesType;
 import org.apache.lucene.index.VectorEncoding;
+import org.apache.lucene.index.VectorSimilarityFunction;
 import org.opensearch.Version;
 import org.opensearch.common.Explicit;
 import org.opensearch.common.xcontent.XContentFactory;
@@ -204,11 +205,10 @@ public class EngineFieldMapper extends KNNVectorFieldMapper {
                 final VectorEncoding encoding = mappedFieldType.vectorDataType == VectorDataType.FLOAT
                     ? VectorEncoding.FLOAT32
                     : VectorEncoding.BYTE;
-                fieldType.setVectorAttributes(
-                    adjustedDimension,
-                    encoding,
-                    SpaceType.DEFAULT.getKnnVectorSimilarityFunction().getVectorSimilarityFunction()
+                final VectorSimilarityFunction similarityFunction = findBestMatchingVectorSimilarityFunction(
+                    resolvedKnnMethodContext.getSpaceType()
                 );
+                fieldType.setVectorAttributes(adjustedDimension, encoding, similarityFunction);
             } else {
                 fieldType.setDocValuesType(DocValuesType.BINARY);
             }
@@ -221,6 +221,23 @@ public class EngineFieldMapper extends KNNVectorFieldMapper {
         this.perDimensionProcessor = knnLibraryIndexingContext.getPerDimensionProcessor();
         this.perDimensionValidator = knnLibraryIndexingContext.getPerDimensionValidator();
         this.vectorValidator = knnLibraryIndexingContext.getVectorValidator();
+    }
+
+    private VectorSimilarityFunction findBestMatchingVectorSimilarityFunction(final SpaceType spaceType) {
+        if (indexCreatedVersion.onOrAfter(Version.V_3_0_0)) {
+            // We need to find the best matching similarity function and not just save DEFAULT space type after 3.0.
+            // This is required for memory optimized search where utilizing .vec file to retrieve vectors.
+            // During the retrieval, it will locate similarity function from the meta info. Without this best effort, always the default
+            // similarity function will be used even when other space type is configured in a mapping.
+            // However, for keeping the backward compatibility, we only apply this to indices created after 3.0+.
+            try {
+                return spaceType.getKnnVectorSimilarityFunction().getVectorSimilarityFunction();
+            } catch (Exception e) {
+                // ignore
+            }
+        }
+
+        return SpaceType.DEFAULT.getKnnVectorSimilarityFunction().getVectorSimilarityFunction();
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


jmazanec15:
Why not update https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/mapper/MethodFieldMapper.java#L180 to be accurate? SpaceType.getKnnVectorSimilarityFunction.getVectorSimilarityFunction should work? 

Spacetype can be gotten from:
```
originalMappingParameters.getResolvedKnnMethodContext().getSpaceType()
```

We would need to ensure bwc

0ctopus13prime:
> Why not update https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/mapper/MethodFieldMapper.java#L180 to be accurate? SpaceType.getKnnVectorSimilarityFunction.getVectorSimilarityFunction should work?
> 
> Spacetype can be gotten from:
> 
> ```
> originalMappingParameters.getResolvedKnnMethodContext().getSpaceType()
> ```
> 
> We would need to ensure bwc

Yeah, that was my 1st trial.  But BWC will be breaking.

For example, 2.x had a segment having EUCLIDIAN while INNER_PRODUCT is the actual space_type.
This will conflict during merging, since two segments have the different space_type (similarity function).
I could not find a way to handle this transparently without reindexing, if you know a graceful way to resolve it, please let me know! Ideally, it should be the quickest way to do it.

navneet1v (approved):

navneet1v:
change looks good to me. Please fix all the CIs

jmazanec15:
@0ctopus13prime Can we branch on
```
return knnMethodConfigContext.getVersionCreated();
```

We do this else where as well. See in KnnVectorFieldMapper:
```
                if (indexCreatedVersion.onOrAfter(Version.V_3_0_0) && hasDocValues.isConfigured() == false) {
                    hasDocValues = Parameter.docValuesParam(m -> toType(m).hasDocValues, true);
                }
```

0ctopus13prime:
I think CI failures are related to other factors:
```
Suite: Test class org.opensearch.knn.bwc.ClearCacheIT
  2> java.lang.NoClassDefFoundError: org/opensearch/secure_sm/SecureSM
        at org.opensearch.test.OpenSearchTestCase.<clinit>(OpenSearchTestCase.java:331)
        at java.base/java.lang.Class.forName0(Native Method)
        at java.base/java.lang.Class.forName(Class.java:534)
        at java.base/java.lang.Class.forName(Class.java:513)
        at com.carrotsearch.randomizedtesting.RandomizedRunner$2.run(RandomizedRunner.java:623)
```

0ctopus13prime:
Thanks to @jmazanec15 suggestion, we found more simpler way to fix the bug.
We can add if-else branch in mapper to save the best matching vector similarity function in meta info for indices created after 3.0+.

src/main/java/org/opensearch/knn/index/mapper/EngineFieldMapper.java
```diff
@@ -223,6 +223,23 @@ private EngineFieldMapper(
...
+    private VectorSimilarityFunction findBestMatchingVectorSimilarityFunction(final SpaceType spaceType) {
+        if (indexCreatedVersion.onOrAfter(Version.V_3_0_0)) {
+            // We need to find the best matching similarity function and not just save DEFAULT space type after 3.0.
+            // This is required for memory optimized search where utilizing .vec file to retrieve vectors.
+            // During the retrieval, it will locate similarity function from the meta info. Without this best effort, always the default
+            // similarity function will be used even when other space type is configured in a mapping.
+            // However, for keeping the backward compatibility, we only apply this to indices created after 3.0+.
+            try {
+                return spaceType.getKnnVectorSimilarityFunction().getVectorSimilarityFunction();
+            } catch (Exception e) {
```
jmazanec15, line 235: Why not catch the unsupported operation exception?

shatejas (approved):

src/main/java/org/opensearch/knn/index/mapper/EngineFieldMapper.java
0ctopus13prime: Hamming space type is the only space that is throwing `IllegalStateException` at the moment.
I can replace Exception to `IllegalStateException`, but `getVectorSimilarityFunction()` can actually throw arbitrary exceptions in there. So I've decided to have it Exception instead of enumerating all possible exceptions to be thrown.

jmazanec15 (approved):

src/main/java/org/opensearch/knn/index/mapper/EngineFieldMapper.java
navneet1v: Lets not ignore this exception. Please throw this exception out.

navneet1v: Just to provide some context why Default space type was used:
@jmazanec15 and @0ctopus13prime BTW this is not only about Hamming distance. For older indices with Nmslib engine a user may have set L1/linf as space type. Hence, default space type was chosen when we moved to Lucene based vector field for our native engines.

0ctopus13prime: I think we should suppress the exception and fallback to the default one.
The purpose of this check is to test whether it could find the best matching similarity function in Lucene.
And Hamming is just throwing IllegalArgumentException when try to get vector similarity function. 
If we just rethrow everything we got, then user cannot create an index with hamming distance.
I understand that snatching all exception then suppress might be uncomfortable, I think this is the way to achieve the goal while minimizing maintenance efforts. One day, a FAISS unique space type can be added where it throws UnsupportedOperationException.

PR #2494, opened 2025-02-05
Fix bytes offset bug and duplicate readers and add uTs for derived source

jmazanec15:
### Description
Fixes a bug in the derived source writer where we are reading the entire bytes array from the bytes ref instead of just the offset+length.

This was the error:
```
[2025-02-04T23:43:54,004][WARN ][o.o.i.e.Engine           ] [test] [target_index][0] failed engine [already closed by tragic event on the index writer]
org.opensearch.core.compress.NotXContentException: Compressor detection can only be called on some xcontent bytes or compressed xcontent bytes
	at org.opensearch.core.compress.CompressorRegistry.compressor(CompressorRegistry.java:75) ~[opensearch-core-2.19.0.jar:2.19.0]
	at org.opensearch.common.xcontent.XContentHelper.convertToMap(XContentHelper.java:166) ~[opensearch-2.19.0.jar:2.19.0]
	at org.opensearch.knn.index.codec.KNN9120Codec.DerivedSourceStoredFieldsWriter.writeField(DerivedSourceStoredFieldsWriter.java:81) ~[?:?]
	at org.apache.lucene.index.StoredFieldsConsumer.writeField(StoredFieldsConsumer.java:80) ~[lucene-core-9.12.1.jar:9.12.1 7a97a05a239d6fb9f1f347aa09bfa52e875be092 - 2024-12-09 16:47:48]
	at org.apache.lucene.index.IndexingChain.processField(IndexingChain.java:754) ~[lucene-core-9.12.1.jar:9.12.1 7a97a05a239d6fb9f1f347aa09bfa52e875be092 - 2024-12-09 16:47:48]
	at org.apache.lucene.index.IndexingChain.processDocument(IndexingChain.java:609) ~[lucene-core-9.12.1.jar:9.12.1 7a97a05a239d6fb9f1f347aa09bfa52e875be092 - 2024-12-09 16:47:48]
	at org.apache.lucene.index.DocumentsWriterPerThread.updateDocuments(DocumentsWriterPerThread.java:263) ~[lucene-core-9.12.1.jar:9.12.1 7a97a05a239d6fb9f1f347aa09bfa52e875be092 - 2024-12-09 16:47:48]
	at org.apache.lucene.index.DocumentsWriter.updateDocuments(DocumentsWriter.java:425) ~[lucene-core-9.12.1.jar:9.12.1 7a97a05a239d6fb9f1f347aa09bfa52e875be092 - 2024-12-09 16:47:48]
[...]

PR #2652, opened 2025-04-08
Prevent derived source from open reader per transform

jmazanec15:
### Description
Fixes a bug that was already fixed in #2494 but was then reverted by accident in a refactor. It makes it so that instead of opening up readers for each transform request, it opens up once per reader.

This was caught in a 100M perf test I was running. Im thinking we should maybe upgrade our nightlies to use this to prevent in future.

### Related Issues

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/codec/backward_codecs/KNN9120Codec/KNN9120DerivedSourceReaders.java
+++ b/src/main/java/org/opensearch/knn/index/codec/backward_codecs/KNN9120Codec/KNN9120DerivedSourceReaders.java
@@ -11,11 +11,13 @@ import org.apache.lucene.codecs.DocValuesProducer;
 import org.apache.lucene.codecs.FieldsProducer;
 import org.apache.lucene.codecs.KnnVectorsReader;
 import org.apache.lucene.codecs.NormsProducer;
+import org.apache.lucene.store.AlreadyClosedException;
 import org.apache.lucene.util.IOUtils;
 import org.opensearch.common.Nullable;
 
 import java.io.Closeable;
 import java.io.IOException;
+import java.util.concurrent.atomic.AtomicInteger;
 
 /**
  * Class holds the readers necessary to implement derived source. Important to note that if a segment does not have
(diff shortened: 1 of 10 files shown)


src/main/java/org/opensearch/knn/index/codec/backward_codecs/KNN9120Codec/KNN9120DerivedSourceReaders.java
```diff
@@ -33,8 +33,24 @@ public class KNN9120DerivedSourceReaders implements Closeable {
     @Nullable
     private final NormsProducer normsProducer;
 
+    private boolean isCloned;
+
     @Override
     public void close() throws IOException {
-        IOUtils.close(knnVectorsReader, docValuesProducer, fieldsProducer, normsProducer);
+        if (isCloned == false) {
```
0ctopus13prime, line 40: Is the case where the root derived source readers that gives cloned one is closed, but the cloned one is still being used? Would it be better to switch it to ref count based approach?

shatejas: - While clone is generally used for deep copy by convention, I have only seen shallow copy with index input, so closing a clone would actually have side effects as it will close underlying resources which are shared. Considering readers are already refcounted we should be good.
- Moreover this clone is used for merge and the readers themselves are not being cloned. Since we cannot have `getMergeInstance` with source we should explicitly rename it to cloneWithMerge so its not used in any other way. Moreover for readers we should do `new DerivedSourceReaders(vectorReader.getMergeInstance, and so on)`

@0ctopus13prime @jmazanec15 thoughts?

jmazanec15: I think that makes sense to me @shatejas

0ctopus13prime: Em.. why it has side effects? Rather, I think it becomes more safer, likewise `std::shared_ptr<T>` for instance.
It will never close underlying resources if other is still using it. 
Also, I believe readers (e.g. 4 readers in this source readers class) are not ref counted? They are directly created from format -- [Code](https://github.com/opensearch-project/k-NN/blob/f028757556573ac4d665ae8efc6fe3e6b02380e5/src/main/java/org/opensearch/knn/index/codec/backward_codecs/KNN9120Codec/KNN9120Codec.java#L77)

My only concern is the case where the root `KNN9120DerivedSourceReaders` having `isClone=true` is closed first, but somewhere a cloned one is still being used. In this case, it will hit already closed exception. 
FYI Lucene always clones stored field readers in SegmentReader - [Code](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/index/SegmentReader.java#L264). 

Of course, we can 100% make sure it won't happen, then why not, we can go for it.

jmazanec15: I see. Let me just add some reference counting to be safe. Ill copy some code from SegmentCoreReaders.

0ctopus13prime (dismissed):

0ctopus13prime (dismissed):

shatejas (dismissed):

shatejas (approved):

src/main/java/org/opensearch/knn/index/codec/backward_codecs/KNN9120Codec/KNN9120DerivedSourceReaders.java
0ctopus13prime: Got an official answer for this in this thread - https://lists.apache.org/thread/d14b5bftl50dmycq08kq76jwb1kk4wts

```
The general contract is that only the original Index input needs to be closed. Slices and clones don't need a close, they are just views on the original input.

All clones or slices always have a noop close method? So it won't hurt to call close.
```

Therefore, the case where the root reader closed but other cloned readers are still being used will not happen.
We don't need ref counting based approach.

Thank you!

0ctopus13prime (approved):

reta:
@jmazanec15 https://github.com/opensearch-project/k-NN/pull/2657 to fix the builds, ty

PR #2653, opened 2025-04-08
Adding main segment implementation for API and indexing.

oaganesh:
### Description
Adding implementation for profiling and statistical analysis of KNN vector segments in OpenSearch. Provides functionality to collect, process, and store statistical information about vector dimensions across different shards and segments.

Example input:

Creating the index
```
curl -X PUT "http://localhost:9200/my-index" -H 'Content-Type: application/json' -d '     
{
  "settings": {                          
    "index": {
      "knn": true
    }
  },
  "mappings": {
    "properties": {
      "my_vector_field": {
        "type": "knn_vector",
        "dimension": 4
      }
    }
  }
}'
```
Adding vectors to the index
```
curl -X POST "http://localhost:9200/my-index/_doc" -H 'Content-Type: application/json' -d'
{
  "my_vector_field": [1.0, 2.0, 3.0, 4.0]
}'
```

```
curl -X POST "http://localhost:9200/my-index/_doc" -H 'Content-Type: application/json' -d'
{
  "my_vector_field": [1.5, 2.0, 3.0, 4.0]
}'
```

Flushing the index
```
curl -X POST "http://localhost:9200/my-index/_flush"
```

Observed example output:
```
curl -X GET "localhost:9200/_plugins/_knn/sampling/my-index/stats?pretty"
{
  "index_summary" : {
    "doc_count" : 2,
    "size_in_bytes" : 5763,
    "timestamp" : "2025-04-07T21:38:31.814948Z"
  },
  "vector_stats" : {
    "sample_size" : 2,
    "summary_stats" : {
      "dimensions" : [
        {
          "dimension" : 0,
          "count" : 2,
          "min" : 1.0,
          "max" : 1.5,
          "sum" : 2.5,
          "mean" : 1.25,
          "standardDeviation" : 0.3536,
          "variance" : 0.125
        },
        {
          "dimension" : 1,
          "count" : 2,
          "min" : 2.0,
          "max" : 2.0,
          "sum" : 4.0,
          "mean" : 2.0,
          "standardDeviation" : 0.0,
          "variance" : 0.0
        },
        {
          "dimension" : 2,
          "count" : 2,
          "min" : 3.0,
          "max" : 3.0,
          "sum" : 6.0,
          "mean" : 3.0,
          "standardDeviation" : 0.0,
          "variance" : 0.0
        },
        {
          "dimension" : 3,
          "count" : 2,
          "min" : 4.0,
          "max" : 4.0,
          "sum" : 8.0,
          "mean" : 4.0,
          "standardDeviation" : 0.0,
          "variance" : 0.0
        }
      ]
    }
  }
}

```

### Related Issues
Implements #2687 

### Check List
- [ ✔️ ] New functionality includes testing.
- [ ✔️ ] New functionality has been documented.
- [ ✔️] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
- [ ✔️] Commits are signed per the DCO using `--signoff`.
- [ ✔️] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).

For more information on following Developer Certificate of Origin and signing off your commits, please check [✔️](https://github.com/opensearch-project/k-NN/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).


src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriter.java
```diff
@@ -108,7 +108,14 @@ public void flush(int maxDoc, final Sorter.DocMap sortMap) throws IOException {
                 field.getVectors()
             );
             final QuantizationState quantizationState = train(field.getFieldInfo(), knnVectorValuesSupplier, totalLiveDocs);
-            SegmentProfilerState.profileVectors(knnVectorValuesSupplier);
+            try {
+                log.info("Starting vector profiling for field: {}", fieldInfo.getName());
+                SegmentProfilerState.profileVectors(knnVectorValuesSupplier, segmentWriteState, fieldInfo.getName());
+                log.info("Completed vector profiling for field: {}", fieldInfo.getName());
+            } catch (Exception e) {
```
markwu-sde, line 115: 1. Should we catch the exception in our SegmentProfileState class instead? We want to keep the business logic here relatively untouched.

src/main/java/org/opensearch/knn/plugin/KNNPlugin.java
```diff
@@ -172,6 +173,8 @@ public class KNNPlugin extends Plugin
 
     private KNNStats knnStats;
     private ClusterService clusterService;
+    // private IndicesService indicesService;
```
markwu-sde, line 176: uncomment or remove this line. Why do we need IndicesService here? It's a pretty heavy object to be using inside ``KnnPlugin``

src/main/java/org/opensearch/knn/plugin/KNNPlugin.java
```diff
@@ -172,6 +173,8 @@ public class KNNPlugin extends Plugin
...
+    private Environment environment;
```
markwu-sde, line 177: Do we need Environment here? What information are we trying to get here?

src/main/java/org/opensearch/knn/plugin/KNNPlugin.java
```diff
@@ -202,7 +205,9 @@ public Collection<Object> createComponents(
         Supplier<RepositoriesService> repositoriesServiceSupplier
     ) {
         this.clusterService = clusterService;
+        // this.indicesService = client.getInstanceFromNode(IndicesService.class);
         this.repositoriesServiceSupplier = repositoriesServiceSupplier;
+        this.environment = environment;
```
markwu-sde, line 210: Why do we need the environment here?

src/main/java/org/opensearch/knn/plugin/KNNPlugin.java
```diff
@@ -249,6 +254,14 @@ public List<RestHandler> getRestHandlers(
             clusterService,
             indexNameExpressionResolver
         );
+
+        RestKNNSamplingStatsHandler restKNNSamplingStatsHandler = new RestKNNSamplingStatsHandler(
+            settings,
+            clusterService,
+            indexNameExpressionResolver,
+            this.environment
+            // indicesService
```
markwu-sde, line 263: Remove comment.

src/main/java/org/opensearch/knn/plugin/rest/RestKNNSamplingStatsHandler.java
```diff
@@ -0,0 +1,159 @@
...
+    private final Environment environment;
+
+    /**
+     * Constructor for the REST handler
+     * @param settings OpenSearch settings
+     * @param clusterService Service for cluster operations
+     * @param indexNameExpressionResolver Resolver for index names
+     * @param environment OpenSearch environment configuration
+     */
+    public RestKNNSamplingStatsHandler(
```
markwu-sde, line 49: NIT: use lombok for @AllArgsConstructor

src/main/java/org/opensearch/knn/plugin/rest/RestKNNSamplingStatsHandler.java
```diff
@@ -0,0 +1,159 @@
...
+
+    /**
+     * Prepares and processes the REST request
+     * @param request The incoming REST request
+     * @param client Node client for executing requests
+     * @return RestChannelConsumer to handle the response
+     */
+    @Override
+    protected RestChannelConsumer prepareRequest(RestRequest request, NodeClient client) throws IOException {
+        String indexName = request.param("index");
```
markwu-sde, line 87: 1. Let's not hardcode parameter strings here.

2. An OS index can consist of multiple vector fields so getting just the index could return multiple fields. Do we need to query by field?

src/main/java/org/opensearch/knn/profiler/SegmentProfilerState.java
```diff
@@ -8,133 +8,439 @@
...
+     * Stores raw statistics data to disk for later retrieval.
+     * Called during vector indexing/processing.
+     * @param outputFile Path to output file
+     * @param statistics List of statistics to write
+     * @param fieldName Name of the field being processed
+     * @param vectorCount Total number of vectors
      */
-    public static SegmentProfilerState profileVectors(final Supplier<KNNVectorValues<?>> knnVectorValuesSupplier) throws IOException {
-        KNNVectorValues<?> vectorValues = knnVectorValuesSupplier.get();
+    private static void writeStatsToFile(Path outputFile, List<SummaryStatistics> statistics, String fieldName, int vectorCount)
```
markwu-sde, line 74: NIT: final whenever possible

src/main/java/org/opensearch/knn/profiler/SegmentProfilerState.java
```diff
@@ -8,133 +8,439 @@
...
+        throws IOException {
+        // Create parent directories if they don't exist
+        Files.createDirectories(outputFile.getParent());
```
markwu-sde, line 77: Based off our conversation with @jmazanec15 and what we've discussed we may want to revisit how we're writing to the segment directory. Can we re-use the implementation used by the ``QuantizationStateWriter`` to persist our state?

src/main/java/org/opensearch/knn/profiler/SegmentProfilerState.java
```diff
@@ -8,133 +8,439 @@
...
+
+        try (XContentBuilder jsonBuilder = XContentFactory.jsonBuilder()) {
```
markwu-sde, line 79: I'm a little confused here... can we leverage some sort of ``ObjectMapper`` to simply write the SummaryStatistic object to the segment? That way we can avoid the XContentBuilder since we already have the object we need to serialize. 

If we need to serialize additional metadata not included in SummaryStatistic we can also wrap it around a wrapper object and serialize that.

src/main/java/org/opensearch/knn/profiler/SegmentProfilerState.java
```diff
@@ -8,133 +8,439 @@
...
+                .field("timestamp", ISO_FORMATTER.format(Instant.now()))
+                .field("fieldName", fieldName)
+                .field("vectorCount", vectorCount)
+                .field("dimension", statistics.size())
+                .startArray("dimensions");
+            // Add statistics for each dimension
+            for (int i = 0; i < statistics.size(); i++) {
+                SummaryStatistics stats = statistics.get(i);
+                jsonBuilder.startObject()
+                    .field("dimension", i)
```
markwu-sde, line 93: We shouldn't hardcode strings if possible in business logic.

src/main/java/org/opensearch/knn/profiler/SegmentProfilerState.java
```diff
@@ -8,133 +8,439 @@
...
+    }
 
+    /**
+     * Profiles vectors in a segment and collects statistical information
+     * @param knnVectorValuesSupplier Supplier for vector values
+     * @param segmentWriteState State of the segment being written
+     * @param fieldName Name of the field being processed
+     * @return SegmentProfilerState containing collected statistics
+     */
+    public static SegmentProfilerState profileVectors(
```
markwu-sde, line 121: ``profileVectors`` here returns a ``SegmentProfilerState`` but I'm not seeing it used anywhere in the PR. Are we using the method signature anywhere? Should we change it?

src/main/java/org/opensearch/knn/profiler/SegmentProfilerState.java
```diff
@@ -8,133 +8,439 @@
...
+        // Get vector values from the supplier
+        KNNVectorValues<?> vectorValues = knnVectorValuesSupplier.get();
         if (vectorValues == null) {
-            log.info("No vector values available");
             return new SegmentProfilerState(new ArrayList<>());
         }
-
-        // Initialize vector values
+        // Initialize new profiler state and vector values
+        SegmentProfilerState profilerState = new SegmentProfilerState(new ArrayList<>());
```
markwu-sde, line 132: Do we need to create the ``SegmentProfilerState`` here? Can we just return it if possible (assuming we need to return it)

src/main/java/org/opensearch/knn/profiler/SegmentProfilerState.java
```diff
@@ -8,133 +8,439 @@
...
-        logDimensionStatistics(statistics, dimension);
+        // Generate filename and write statistics to file
+        String statsFileName = IndexFileNames.segmentFileName(
+            segmentWriteState.segmentInfo.name,
+            segmentWriteState.segmentSuffix,
+            VECTOR_STATS_EXTENSION
+        );
+
+        Directory directory = getUnderlyingDirectory(segmentWriteState.directory);
+        Path statsFile = ((FSDirectory) directory).getDirectory().resolve(statsFileName);
```
markwu-sde, line 158: Let's revisit this approach of getting the directory. Can we reuse what we've done for the Quantization sampling?

src/main/java/org/opensearch/knn/profiler/SegmentProfilerState.java
```diff
@@ -8,133 +8,439 @@
...
+    /**
+     * Determines the path to a shard's index directory
+     * @param shard Shard statistics
+     * @param environment OpenSearch environment
+     * @return Path to the shard's index directory
+     */
+    private static Path getShardIndexPath(ShardStats shard, Environment environment) {
+        int shardId = shard.getShardRouting().shardId().getId();
+        String indexUUID = shard.getShardRouting().shardId().getIndex().getUUID();
+        return environment.dataFiles()[0].resolve("nodes")
```
markwu-sde, line 248: I don't think we should be reliant on a hard-coded pathing to get the file that we'd like to get. This is brittle and can break if the directory path structure changes.

src/main/java/org/opensearch/knn/profiler/SegmentProfilerState.java
```diff
@@ -8,133 +8,439 @@
...
+        });
+    }
+
+    /**
+     * Merges statistics from source into target
+     * @param target Target statistics list
+     * @param source Source statistics list
+     */
+    static void mergeStatistics(List<SummaryStatistics> target, List<SummaryStatistics> source) {
+        if (target.isEmpty()) {
```
markwu-sde, line 279: I'm confused with how we're merging the statistics here. Why are we only transferring the min/max? What's happening with source and target? 

SummaryStatistics can be combined using aggregate summary statistics: https://commons.apache.org/proper/commons-math/commons-math-docs/apidocs/org/apache/commons/math4/legacy/stat/descriptive/AggregateSummaryStatistics.html

Is there any way we can leverage this?

src/main/java/org/opensearch/knn/profiler/SegmentProfilerState.java
```diff
@@ -8,133 +8,439 @@
...
+            for (SummaryStatistics sourceStat : source) {
+                SummaryStatistics newStat = new SummaryStatistics();
+                newStat.addValue(sourceStat.getMin());
+                if (sourceStat.getN() > 1) {
```
markwu-sde, line 283: NIT: don't use hardcoded values.

src/main/java/org/opensearch/knn/profiler/SegmentProfilerState.java
oaganesh: Having some issues reusing the quantization as it appears to have issues when reading the file. Kept current implementation for now.

src/main/java/org/opensearch/knn/profiler/SegmentProfilerState.java
oaganesh: Added variables for string values

src/main/java/org/opensearch/knn/profiler/SegmentProfilerState.java
oaganesh: Faced some difficulty with objectmapper to help serialize. Keeping XContentBuilder approach for now to accomodate for bigger changes. Can change more if necessary.

src/main/java/org/opensearch/knn/profiler/SegmentProfilerState.java
oaganesh: Added changes

src/main/java/org/opensearch/knn/profiler/SegmentProfilerState.java
oaganesh: Changed

src/main/java/org/opensearch/knn/plugin/rest/RestKNNSamplingStatsHandler.java
oaganesh: Changed

src/main/java/org/opensearch/knn/plugin/rest/RestKNNSamplingStatsHandler.java
```diff
@@ -0,0 +1,159 @@
...
+     * @param environment OpenSearch environment configuration
+     */
+    public RestKNNSamplingStatsHandler(
+        Settings settings,
+        ClusterService clusterService,
+        IndexNameExpressionResolver indexNameExpressionResolver,
+        Environment environment
+        // IndicesService indicesService
+    ) {
+        // this.indicesService = indicesService;
```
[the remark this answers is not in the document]
oaganesh: Changed

src/main/java/org/opensearch/knn/plugin/rest/RestKNNSamplingStatsHandler.java
oaganesh: Changed

src/main/java/org/opensearch/knn/plugin/rest/RestKNNSamplingStatsHandler.java
```diff
@@ -0,0 +1,159 @@
...
+import java.util.List;
+
+import static org.opensearch.rest.RestRequest.Method.GET;
+
+@Log4j2
+/**
+ * Rest handler for sampling stats endpoint
+ */
+public class RestKNNSamplingStatsHandler extends BaseRestHandler {
+    // private final IndicesService indicesService;
```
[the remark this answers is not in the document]
oaganesh: Changed

src/main/java/org/opensearch/knn/plugin/KNNPlugin.java
oaganesh: Changed

src/main/java/org/opensearch/knn/plugin/KNNPlugin.java
oaganesh: It's used to access the data directory paths and node-specific items. It's needed to pass to the RestKNNSamplingStatsHandler to locate and read statistics files stored on disk. The environment provides access to the node's data paths where vector statistics files are stored under the indices directory structure.

src/main/java/org/opensearch/knn/plugin/KNNPlugin.java
oaganesh: It's used to access the data directory paths and node-specific items. It's needed to pass to the RestKNNSamplingStatsHandler to locate and read statistics files stored on disk. The environment provides access to the node's data paths where vector statistics files are stored under the indices directory structure.

src/main/java/org/opensearch/knn/plugin/KNNPlugin.java
oaganesh: Changed

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriter.java
oaganesh: Changed

src/main/java/org/opensearch/knn/index/query/SegmentLevelQuantizationUtil.java
```diff
@@ -57,4 +75,92 @@ static QuantizationState getQuantizationState(final LeafReader leafReader, Strin
...
+        for (int i = 0; i < statistics.size(); i++) {
+            Collection<StatisticalSummary> statCollection = Collections.singletonList(statistics.get(i));
+            AggregateSummaryStatistics aggregator = new AggregateSummaryStatistics();
+            aggregator.aggregate(statCollection);
+
+            StatisticalSummary summary = aggregator.getSummary();
+            SummaryStatistics aggregatedStats = new SummaryStatistics();
+
+            if (summary.getN() > 0) {
+                aggregatedStats.addValue(summary.getMin());
```
markwu-sde, line 149: I'm a little confused on what we're trying to do here. 

``aggregatedStats.addValue`` adds a value to the stat dataset which it then can compute the required stats(min, max, etc).  

I'm assuming the intention of this method is to take in a list of all the statistics we have collected for each segment and aggregating them. I don't think it makes sense to feed the computed stats into a new SummaryStatistics to reprocess as a aggregation. They would be treated as values within the dataset when they're not.

Let's say you have 2 segments with these raw values:
Segment 1: [1, 2, 3]
Segment 2: [4, 5, 6]

Their individual statistics would be:
Segment 1: min=1, max=3, mean=2
Segment 2: min=4, max=6, mean=5

If you feed these statistics into a new SummaryStatistics like:
aggregatedStats.addValue(segment1Stats.getMin());  // adds 1
aggregatedStats.addValue(segment1Stats.getMax());  // adds 3 
aggregatedStats.addValue(segment1Stats.getMean()); // adds 2
aggregatedStats.addValue(segment2Stats.getMin());  // adds 4
aggregatedStats.addValue(segment2Stats.getMax());  // adds 6
aggregatedStats.addValue(segment2Stats.getMean()); // adds 5

You'd end up with a dataset of [1,3,2,4,6,5] which would give you:
min=1, max=6, mean=3.5

But the actual statistics across all raw values [1,2,3,4,5,6] should be:
min=1, max=6, mean=3.5

While in this case the min/max/mean happen to match, this is just coincidental. The computed variance and other statistics would be incorrect since we're treating summary statistics as if they were raw data points. We need to properly aggregate the underlying statistics instead.

src/main/java/org/opensearch/knn/index/KNNIndexShard.java
```diff
@@ -9,17 +9,25 @@
 import lombok.AllArgsConstructor;
 import lombok.Getter;
 import lombok.extern.log4j.Log4j2;
-import org.apache.lucene.index.FieldInfo;
+import org.apache.commons.math3.stat.descriptive.SummaryStatistics;
+import org.apache.lucene.codecs.CodecUtil;
+import org.apache.lucene.index.SegmentInfo;
+import org.apache.lucene.index.SegmentCommitInfo;
```
markwu-sde, line 15: Make sure you clean up any unused imports.

src/main/java/org/opensearch/knn/index/KNNIndexShard.java
```diff
@@ -83,14 +100,117 @@ public String getIndexName() {
     }
 
     /**
-     * Load all of the k-NN segments for this shard into the cache.
```
markwu-sde, line 86: NIT: Does this comment need to be changed?

src/main/java/org/opensearch/knn/index/KNNIndexShard.java
```diff
@@ -83,14 +100,117 @@ public String getIndexName() {
...
+                        String fieldName = fieldInfo.name;
+                        int fieldNumber = fieldInfo.number;
+                        log.info("[KNN Stats] Processing field: {} (number: {}) in segment: {}", fieldName, fieldNumber, segmentName);
+
+                        // Look for quantization state file
+                        String[] files = segmentReader.directory().listAll();
+                        log.debug("[KNN Stats] Files in segment {}: {}", segmentName, String.join(", ", files));
+
+                        for (String file : files) {
+                            if (file.startsWith(segmentName) && file.endsWith(KNNConstants.QUANTIZATION_STATE_FILE_SUFFIX)) {
```
markwu-sde, line 131: Why do we need this? Can we just use this to get the file?

https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNN990QuantizationStateReader.java#L116

src/main/java/org/opensearch/knn/index/KNNIndexShard.java
```diff
@@ -83,14 +100,117 @@ public String getIndexName() {
...
+                                log.info("[KNN Stats] Found quantization state file: {} for field: {}", file, fieldName);
+
+                                try {
+                                    // Extract the suffix correctly
+                                    String suffix = file.substring(
+                                        file.indexOf("NativeEngines990KnnVectorsFormat"),
+                                        file.length() - KNNConstants.QUANTIZATION_STATE_FILE_SUFFIX.length()
+                                    );
+
+                                    SegmentReadState segmentReadState = new SegmentReadState(
```
markwu-sde, line 141: We shouldn't try to construct a `SegmentReadState` - this is a lucene object that lucene provides but we shouldn't generate by ourselves. 

But in general is there really a need to do all of this in `KNNIndexShard`? From the `KNNIndexShard` if we are able to access the searcher then we should be able to use this `public void search(String field, float[] target, KnnCollector knnCollector, Bits acceptDocs)` in `NativeEngines990KnnVectorsReader` and getting the SegmentProfilerState there.

src/main/java/org/opensearch/knn/index/KNNIndexShard.java
```diff
@@ -83,14 +100,117 @@ public String getIndexName() {
...
+                                        segmentReader.directory(),
+                                        segmentReader.getSegmentInfo().info,
+                                        segmentReader.getFieldInfos(),
+                                        IOContext.DEFAULT,
+                                        suffix
+                                    );
+
+                                    QuantizationStateReadConfig readConfig = new QuantizationStateReadConfig(
```
markwu-sde, line 149: The `QuantizationStateReadConfig` has items that we don't need like cache key and `QuantizationParams`. Should we use our own here?

src/main/java/org/opensearch/knn/index/KNNIndexShard.java
```diff
@@ -235,4 +358,181 @@ static class EngineFileContext {
...
+        try (Engine.Searcher searcher = indexShard.acquireSearcher("knn-stats-collection")) {
+            for (LeafReaderContext leafContext : searcher.getIndexReader().leaves()) {
+                SegmentReader segmentReader = Lucene.segmentReader(leafContext.reader());
+                String segmentName = segmentReader.getSegmentName();
+                log.info("[KNN Debug] Processing segment: {}", segmentName);
+
+                // Get directory info for this segment
+                Directory directory = segmentReader.directory();
+                if (directory instanceof FSDirectory) {
+                    Path dirPath = ((FSDirectory) directory).getDirectory();
```
markwu-sde, line 380: I'm a little confused on why we are looking into the directory here?

As discussed we have access to a searcher which we have inserted our `NativeEngines990KnnVectorsReader` which we pass into Lucene. When we call 'search' and pass in a custom KNNCollector of our type we can retrieve the segment profiler state. 

Can we pass in the LeafReader similar to what we do here? This should give us our SegmentProfilerState

src/main/java/org/opensearch/knn/index/KNNIndexShard.java
```diff
@@ -235,4 +358,181 @@ static class EngineFileContext {
...
+                                // Use the string constant directly to avoid reference issues
+                                CodecUtil.checkHeader(input, "NativeEngines990KnnVectorsFormatQSData", 0, 0);
+                                log.info("[KNN Debug] Successfully verified header for file: {}", file);
+                            } catch (Exception e) {
+                                log.error("[KNN Debug] Invalid header in file {}: {}", file, e.getMessage());
+                                continue;
+                            }
+
+                            // Read footer to find index position
+                            long fileLength = input.length();
```
markwu-sde, line 418: Why is this logic in KNNIndexShard? This logic is used to retrieve the `SegmentProfileState` in `KNN990QuantizationStateReader. Can we do only the aggregation here instead?

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNN990QuantizationStateReader.java
```diff
@@ -95,7 +98,7 @@ public static QuantizationState read(QuantizationStateReadConfig readConfig) thr
     }
 
     @VisibleForTesting
-    static int getNumFields(IndexInput input) throws IOException {
+    public static int getNumFields(IndexInput input) throws IOException {
```
markwu-sde, line 101: Let's be careful when we are making an existing method public - we shouldn't do this just to mold to our code. We should be looking for ways to achieve our change without modifying existing code as much as possible.

src/main/java/org/opensearch/knn/plugin/KNNPlugin.java
markwu-sde: We don't need this - we can directly follow what's being used in the `QuantizationState`

src/main/java/org/opensearch/knn/plugin/KNNPlugin.java
markwu-sde: We don't need this - we can directly follow what's being used in the `QuantizationState`

src/main/java/org/opensearch/knn/profiler/SegmentProfilerState.java
```diff
@@ -1,140 +1,534 @@
...
+
+            if (dimensions <= 0 || dimensions > 10000) {
+                throw new IOException("Invalid dimension count: " + dimensions);
+            }
+
+            List<SummaryStatistics> statistics = new ArrayList<>(dimensions);
+
+            for (int i = 0; i < dimensions; i++) {
+                SummaryStatistics stats = new SummaryStatistics();
+                long n = dis.readLong();
```
markwu-sde, line 497: Similar to what I had mentioned in a previous object - this is not deserializing a object. In your code path here we are re-inserting the outputs back into a ``SummaryStatistics`` object. 

Reposting a TLDR of what I had mentioned.

Let's say you have 2 segments with these raw values:
Segment 1: [1, 2, 3] Segment 2: [4, 5, 6]

Their individual statistics would be:
Segment 1: min=1, max=3, mean=2 Segment 2: min=4, max=6, mean=5

If you feed these statistics into a new SummaryStatistics like:
aggregatedStats.addValue(segment1Stats.getMin()); // adds 1 aggregatedStats.addValue(segment1Stats.getMax()); // adds 3 aggregatedStats.addValue(segment1Stats.getMean()); // adds 2 aggregatedStats.addValue(segment2Stats.getMin()); // adds 4 aggregatedStats.addValue(segment2Stats.getMax()); // adds 6 aggregatedStats.addValue(segment2Stats.getMean()); // adds 5

The computed variance and other statistics would be incorrect since we're treating summary statistics as if they were raw data points. We need to properly aggregate the underlying statistics instead.

We need to serialize and deserialize the actual object itself.

src/test/java/org/opensearch/knn/profiler/SegmentProfilerStateTests.java
```diff
@@ -5,161 +5,125 @@
...
-    private Supplier<KNNVectorValues<?>> mockSupplier;
-
-    @Before
-    public void setUp() throws Exception {
-        super.setUp();
-        mockVectorValues = (KNNVectorValues<Object>) mock(KNNVectorValues.class);
-        mockSupplier = () -> mockVectorValues;
-    }
-
-    public void testConstructor() {
```
markwu-sde, line 34: No need to test constructors.

markwu-sde (changes requested):
Overall I think our approach is a little too overcomplicated given that we already have a similar implementation available. Attached below is a reference PoC PR that I had drafted for what I think could fit better:

https://github.com/opensearch-project/k-NN/pull/2667

https://github.com/opensearch-project/k-NN/pull/2667/commits/b9557923bad308629576430f4e58ba9379134b41

Feel free to cross-reference it with my comments when making the revision.

src/main/java/org/opensearch/knn/profiler/SegmentProfilerState.java
```diff
@@ -1,140 +1,534 @@
...
      */
-    private static void processByteVector(byte[] vector, List<SummaryStatistics> statistics) {
-        for (int j = 0; j < vector.length; j++) {
-            statistics.get(j).addValue(vector[j] & 0xFF);
+    public byte[] toByteArray() throws IOException {
+        try (ByteArrayOutputStream baos = new ByteArrayOutputStream(); DataOutputStream dos = new DataOutputStream(baos)) {
+
+            dos.writeInt(statistics.size());
+
+            for (SummaryStatistics stats : statistics) {
```
markwu-sde, line 460: I would suggest looking into the serializable interface: https://docs.oracle.com/javase/8/docs/api/java/io/Serializable.html

We may be able to leverage this to directly serialize/deserialize

src/main/java/org/opensearch/knn/profiler/RestKNNProfileHandler.java
```diff
@@ -0,0 +1,101 @@
...
+import java.util.List;
+
+import static org.opensearch.knn.index.KNNSettings.KNN_INDEX;
+import static org.opensearch.action.support.IndicesOptions.strictExpandOpen;
+
+/**
+ * RestHandler for k-NN index profile API. API provides the ability for a user to get statistical information
+ * about vector dimensions in specific indices.
+ */
+public class RestKNNProfileHandler extends BaseRestHandler {
```
markwu-sde, line 38: Will need more time to look into the API implementation for this PR since it's not too closely coupled with our profiling implementation

src/main/java/org/opensearch/knn/plugin/transport/KNNProfileAction.java
```diff
@@ -0,0 +1,20 @@
...
+ */
+
+package org.opensearch.knn.plugin.transport;
+
+import org.opensearch.action.ActionType;
+
+/**
+ * Action for profiling KNN vectors in an index
+ */
+public class KNNProfileAction extends ActionType<KNNProfileResponse> {
```
markwu-sde, line 13: I know we synced offline on this but is it possible to move the API implementation to another PR? I don't think there's

src/main/java/org/opensearch/knn/index/query/SegmentProfilerUtil.java
```diff
@@ -0,0 +1,37 @@
...
+     * Gets the segment profile state for a given field
+     * @param leafReader The leaf reader to query
+     * @param fieldName The field name to profile
+     * @return The segment profiler state
+     * @throws IOException If there's an error reading the segment
+     */
+    public static SegmentProfilerState getSegmentProfileState(final LeafReader leafReader, String fieldName) throws IOException {
+        final SegmentProfileKNNCollector tempCollector = new SegmentProfileKNNCollector();
+        leafReader.searchNearestVectors(fieldName, new float[0], tempCollector, null);
+        if (tempCollector.getSegmentProfilerState() == null) {
```
markwu-sde, line 32: Do we want to throw an exception if we're unable to get the SegmentProfilerState? What if the user has the feature disabled?

src/main/java/org/opensearch/knn/index/KNNIndexShard.java
```diff
@@ -82,6 +87,83 @@ public String getIndexName() {
...
+                    .map(state -> state.getStatistics().get(dimensionId))
+                    .collect(Collectors.toList());
+
+                shardVectorProfile.add(AggregateSummaryStatistics.aggregate(transposed));
+            }
+
+            // Log the results for each dimension
+            for (int i = 0; i < shardVectorProfile.size(); i++) {
+                StatisticalSummaryValues stats = shardVectorProfile.get(i);
+                log.info(
```
markwu-sde, line 140: I've mentioned this before but would it just be easier to use `toString` here given we're not really focused on a specific output?

src/main/java/org/opensearch/knn/index/KNNIndexShard.java
```diff
@@ -82,6 +87,83 @@ public String getIndexName() {
...
+
+        try (Engine.Searcher searcher = indexShard.acquireSearcher("knn-profile")) {
+            List<SegmentProfilerState> segmentLevelProfilerStates = new ArrayList<>();
+
+            log.info("[KNN] Beginning profiling for field: {} in shard: {}", fieldName, indexShard.shardId());
+
+            // For each leaf, collect the profile
+            searcher.getIndexReader().leaves().forEach(leaf -> {
+                try {
+                    log.info("[KNN] Processing leaf reader for segment: {}", leaf.reader());
```
markwu-sde, line 108: Should we avoid clutter in the logs by reducing these info lines? This doesn't add any value here when debugging as we already have exception handling

src/main/java/org/opensearch/knn/index/KNNIndexShard.java
```diff
@@ -82,6 +87,83 @@ public String getIndexName() {
...
     }
 
+    /**
+     * Profile the vector fields in this shard and return statistical information.
+     *
+     * @param fieldName The name of the vector field to profile
+     * @return List of statistical summaries for each dimension
+     */
+    // TODO: Write unit tests to ensure that the segment statistic aggregation is correct.
+    public List<StatisticalSummaryValues> profile(String fieldName) {
```
markwu-sde, line 97: NIT: final

src/main/java/org/opensearch/knn/index/KNNIndexShard.java
```diff
@@ -82,6 +87,83 @@ public String getIndexName() {
...
+                    stats.getSum(),
+                    stats.getVariance(),
+                    Math.sqrt(stats.getVariance())
+                );
+            }
+
+            log.info("[KNN] Profiling completed for field: {} in shard: {}", fieldName, indexShard.shardId());
+        } catch (Exception e) {
+            log.error(
+                "[KNN] Critical error during profiling for field: {} in shard: {}: {}",
```
markwu-sde, line 156: What does `Critical` mean in this sense? Error messages should be as succinct and objective as possible.

PR #2659, opened 2025-04-11
Add support to run all ITs with remote index builder

anntians:
### Description
Add support to run all ITs with remote index builder. This PR updates the `remote_index_build` GA to run all ITs against remote index builder (`opensearchstaging` docker image). The developer guide is also updated on steps to run ITs with remote index builder feature. 

The PR adds new `@ExpectRemoteBuildValidation` annotation to mark ITs for remote build checks via metrics. For tests marked with `@ExpectRemoteBuildValidation`, there is an `@After` method `verifyRemoteIndexBuild` that checks the metric `INDEX_BUILD_SUCCESS_COUNT` has been incremented. Through discussion, metrics is best option we have so far to verify remote index build in ITs.

### Related Issues
Partially resolves https://github.com/opensearch-project/k-NN/issues/2553 
- Update developer guide on how to set up these tests to run manually
- Update IT to pull image from opensearch-staging instead of personal dockerhub
- Once test issues are resolved, converge the tests to run ALL tests against remote vector index builder

### Check List


src/test/java/org/opensearch/knn/index/KNNMapperSearcherIT.java
```diff
@@ -49,6 +49,9 @@ private void addTestData() throws Exception {
 
         Float[] f4 = { 3.0f, 3.0f };
         addKnnDoc(INDEX_NAME, "4", FIELD_NAME, f4);
+
```
anntians, line 52: Ingest 5 instead of 4 docs because remote index builder fails for doc_count 1,3,4. More details https://github.com/opensearch-project/remote-vector-index-builder/issues/30

src/test/java/org/opensearch/knn/integ/FilteredSearchANNSearchIT.java
```diff
@@ -28,8 +28,8 @@ public void testFilteredSearchWithFaissHnsw_whenFiltersMatchAllDocs_thenReturnCo
         final String filterValue = "red";
         createKnnIndex(INDEX_NAME, getKNNDefaultIndexSettings(), createKnnIndexMapping(FIELD_NAME, 3, METHOD_HNSW, FAISS_NAME));
 
-        // ingest 4 vector docs into the index with the same field {"color": "red"}
-        for (int i = 0; i < 4; i++) {
+        // ingest 5 vector docs into the index with the same field {"color": "red"}
```
anntians, line 31: Ingest 5 instead of 4 docs because remote index builder fails for doc_count 1,3,4. More details https://github.com/opensearch-project/remote-vector-index-builder/issues/31

src/test/java/org/opensearch/knn/index/KNNCircuitBreakerIT.java
```diff
@@ -100,18 +102,27 @@ private void testNodeLevelCircuitBreaker() throws Exception {
...
-        createKnnIndex(indexName1, settings, createKnnIndexMapping(FIELD_NAME, 2));
-        createKnnIndex(indexName2, settings, createKnnIndexMapping(FIELD_NAME, 2));
+        // through testing, 7 is minimum number of docs to trip circuit breaker at 1kb, using local index build
+        int docsInIndex = 7;
+
+        final String remoteBuild = System.getProperty("test.remoteBuild", null);
+        if (isRemoteIndexBuildSupported(getBWCVersion()) && remoteBuild != null) {
+            settings.put(KNN_INDEX_REMOTE_VECTOR_BUILD, true);
+            settings.put(KNN_INDEX_REMOTE_VECTOR_BUILD_THRESHOLD, "0kb");
+            // through testing, 10 is minimum number of docs to trip circuit breaker at 1kb, using remote index build
```
anntians, line 118: It seems that remote index build creates smaller graph files, so we need to ingest more docs in the indices to trip the circuit breaker. More details https://github.com/opensearch-project/remote-vector-index-builder/issues/40

DEVELOPER_GUIDE.md
```diff
@@ -346,6 +347,35 @@ In case remote cluster is secured it's possible to pass username and password wi
 ./gradlew :integTestRemote -Dtests.rest.cluster=localhost:9200 -Dtests.cluster=localhost:9200 -Dtests.clustername="integTest-0" -Dhttps=true -Duser=admin -Dpassword=<admin-password>
 ```
 
+### Run Integration Tests With Remote Index Builder Feature
```
rchitale7, line 350: Should mention somewhere that we need a machine with GPUs to run the image and the ITs. And add a link to here: https://github.com/opensearch-project/remote-vector-index-builder/blob/main/DEVELOPER_GUIDE.md#provisioning-an-instance-for-development

src/testFixtures/java/org/opensearch/knn/DerivedSourceUtils.java
```diff
@@ -28,58 +28,83 @@
...
+        if (knnRestTestCase.isRemoteIndexBuildSupported(knnRestTestCase.getBWCVersion()) && remoteBuild != null) {
+            derived_enabled_settings_builder.put(KNN_INDEX_REMOTE_VECTOR_BUILD, true);
+            derived_enabled_settings_builder.put(KNN_INDEX_REMOTE_VECTOR_BUILD_THRESHOLD, "0kb");
+
+            derived_enabled_with_segrep_settings_builder.put(KNN_INDEX_REMOTE_VECTOR_BUILD, true);
+            derived_enabled_with_segrep_settings_builder.put(KNN_INDEX_REMOTE_VECTOR_BUILD_THRESHOLD, "0kb");
+
+            derived_disabled_settings_builder.put(KNN_INDEX_REMOTE_VECTOR_BUILD, true);
+            derived_disabled_settings_builder.put(KNN_INDEX_REMOTE_VECTOR_BUILD_THRESHOLD, "0kb");
+        }
```
jed326, line 101: Same thing here, if we can move this to the base class that would avoid these changes

.github/workflows/remote_index_build.yml
```diff
@@ -110,23 +110,23 @@ jobs:
...
-            ./gradlew :integTestRemoteIndexBuild -Ds3.enabled=true -Dtest.remoteBuild=s3.localStack -Dtest.bucket=remote-index-build-bucket -Dtest.base_path=vectors -Daccess_key=${{ env.AWS_ACCESS_KEY_ID }} -Dsecret_key=${{ env.AWS_SECRET_ACCESS_KEY }} -Dsession_token=${{ env.AWS_SESSION_TOKEN}} -Dtests.class=org.opensearch.knn.index.RemoteBuildIT -Davx2.enabled=false -Davx512.enabled=false -Davx512_spr.enabled=false -Dnproc.count=`nproc`
+            ./gradlew :integTestRemoteIndexBuild -Ds3.enabled=true -Dtest.remoteBuild=s3.localStack -Dtest.bucket=remote-index-build-bucket -Dtest.base_path=vectors -Daccess_key=${{ env.AWS_ACCESS_KEY_ID }} -Dsecret_key=${{ env.AWS_SECRET_ACCESS_KEY }} -Dsession_token=${{ env.AWS_SESSION_TOKEN}} -Davx2.enabled=false -Davx512.enabled=false -Davx512_spr.enabled=false -Dnproc.count=`nproc`
           fi
 
       - name: Verify Remote Index Builder logs
         run: |
-          if docker logs remote-index-builder-container 2>&1 | grep -q "INFO - Index built successfully!"; then
+          if docker logs remote-index-builder-container 2>&1 | grep -q "completed with status: JobStatus.COMPLETED"; then
```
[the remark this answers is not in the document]
navneet1v: +1

DEVELOPER_GUIDE.md
navneet1v: +1

src/test/java/org/opensearch/knn/integ/KNNScriptScoringIT.java
```diff
@@ -948,6 +954,8 @@ private void createIndexAndAssertScriptScore(
                 ExceptionsHelper.catchAsRuntimeException(() -> addKnnDoc(INDEX_NAME, k, (v != null) ? FIELD_NAME : "dummy", vector));
             });
 
+            forceMergeKnnIndex(INDEX_NAME);
```
anntians, line 957: ITs in this class were not consistently triggering remote index build, adding `forceMergeKnnIndex` ensures remote build is triggered consistently

src/test/java/org/opensearch/knn/index/AdvancedFilteringUseCasesIT.java
```diff
@@ -104,6 +104,7 @@ public class AdvancedFilteringUseCasesIT extends KNNRestTestCase {
      */
     @SneakyThrows
     public void testFiltering_whenNestedKNNAndFilterFieldWithNestedQueries_thenSuccess() {
+        setExpectRemoteBuild(true);
```
jed326, line 107: Thinking on this some more, I don't really like this implementation for a few reasons
1. It's really easy for future tests to miss adding this
2. We're making a change to every single test here (is it less changes if we default this to true, and `setExpectRemoteBuild(false);` instead?
3. Whether or not a test uses remote build path is still somewhat arbitrary

Curious to hear what @navneet1v or any other maintainers have to say on this though

anntians: That makes sense. Following up from our discussion: 

1. Perhaps we can choose a subset of test classes that will explicitly check remote build with `setExpectRemoteBuild`, such as `FaissIT` or filtered search test class. 
2. Then for other test classes we can go with a best effort approach and just check whether the entire test class triggered remote build or not, instead of individual tests. 

Or we can just have `1.` and not `2.`

Thoughts? @navneet1v

navneet1v: I thought this was one of the AI for this task to identify what is the best way to enable the tests which ensures that the feature doesn't break.

anntians: Currently, we are already using a separate CI https://github.com/opensearch-project/k-NN/blob/main/.github/workflows/remote_index_build.yml to run ITs with remote index builder using the `integTestRemoteIndexBuild` gradle task, and we pass settings in the command line to enable remote build.

Full command:
```
./gradlew :integTestRemoteIndexBuild -Ds3.enabled=true -Dtest.remoteBuild=s3.localStack -Dtest.bucket=remote-index-build-bucket -Dtest.base_path=vectors -Daccess_key=${{ env.AWS_ACCESS_KEY_ID }} -Dsecret_key=${{ env.AWS_SECRET_ACCESS_KEY }} -Dsession_token=${{ env.AWS_SESSION_TOKEN}}
```

I think the concern here is how we ensure `all the tests run and be successful`. In the latest revision we tried adding an `@After` method to check metrics on whether remote build worked for all tests that do trigger remote build, but @jed326 pointed out a few concerns above. So, should we just narrow it down to a few crucial test classes to check for remote build, to minimize changes to many test files?

DEVELOPER_GUIDE.md
```diff
@@ -346,6 +347,64 @@ In case remote cluster is secured it's possible to pass username and password wi
...
+The steps below should be run from a GPU machine. Reference link below to provision a GPU instance
+
+https://github.com/opensearch-project/remote-vector-index-builder/blob/main/DEVELOPER_GUIDE.md#provisioning-an-instance-for-development
+
+To run integration tests using the remote index builder feature, first create an S3 bucket `<bucket_name>` in an AWS account. Then run below to setup remote index builder
```
jed326, line 355: Given that your CI itself is using localstack, I think you can mention using localstack as an alternative here.

DEVELOPER_GUIDE.md
```diff
@@ -346,6 +347,64 @@ In case remote cluster is secured it's possible to pass username and password wi
...
+export AWS_SECRET_ACCESS_KEY=
+export AWS_SESSION_TOKEN=
+
+// 2. Run integration tests against remote index builder
+./gradlew :integTestRemoteIndexBuild -Ds3.enabled=true -Dtest.remoteBuild=s3 -Dtest.bucket=<bucket_name> -Dtest.base_path=vectors -Daccess_key=${AWS_ACCESS_KEY_ID} -Dsecret_key=${AWS_SECRET_ACCESS_KEY} -Dsession_token=${AWS_SESSION_TOKEN}
+```
+
+#### Verify Remote Index Build in Integration Tests
+Currently we have a subset of integration tests in which we explicitly verify remote index build is triggered. The integration tests below contain tests
+with `setExpectRemoteBuild`, which sets a flag for the `@After` method `verifyRemoteIndexBuild` in `KNNRestTestCase` to verify remote build was triggered.
```
jed326, line 384: Since we want to do remote path validation on just a subset of the tests, can't we use a base class with this logic instead of adding `setExpectRemoteBuild` to each test?

src/test/java/org/opensearch/knn/index/AdvancedFilteringUseCasesIT.java
```diff
@@ -170,6 +171,7 @@ public void testFiltering_whenNestedKNNAndFilterFieldWithNestedQueries_thenSucce
      */
     @SneakyThrows
     public void testFiltering_whenNestedKNNAndFilterFieldWithNoNestedContextInFilterQuery_thenSuccess() {
+        setExpectRemoteBuild(true);
```
jed326, line 174: IMO this would be cleaner as a custom annotation instead of adding a method call to the top of each test

src/test/java/org/opensearch/knn/index/AdvancedFilteringUseCasesIT.java
```diff
@@ -103,6 +103,7 @@ public class AdvancedFilteringUseCasesIT extends KNNRestTestCase {
      * }
      */
     @SneakyThrows
+    @ExpectRemoteBuildValidation
```
jed326, line 106: nit: If all the methods in a test class support it you should be able to annotate at the class level. I'm fine if that's not in this PR though.

src/test/java/org/opensearch/knn/integ/ExpandNestedDocsIT.java
```diff
@@ -168,7 +168,10 @@ public void testExpandNestedDocs_whenFilteredOnNestedFieldDoc_thenReturnFiltered
     }
 
     @SneakyThrows
+    @ExpectRemoteBuildValidation
     public void testExpandNestedDocs_whenMultiShards_thenReturnCorrectResult() {
+        if (description.equals("Faiss with float format and in memory mode")) {
+        }
```
jed326, line 174: This doesn't look right

src/testFixtures/java/org/opensearch/knn/KNNRestTestCase.java
```diff
@@ -177,13 +191,80 @@ public void cleanUpCache() throws Exception {
...
-    public void setupRemoteIndexBuildSettings() throws Exception {
+    public void setupRemoteIndexBuildSettingsTestFallback() throws Exception {
         if (randomBoolean() && isRemoteIndexBuildSupported(getBWCVersion())) {
             updateClusterSettings(KNNFeatureFlags.KNN_REMOTE_VECTOR_BUILD_SETTING.getKey(), true);
             updateClusterSettings(KNNSettings.KNN_REMOTE_VECTOR_REPO, "integ-test-repo");
         }
     }
 
+    @Before
+    public void setupRemoteIndexBuildSettings() throws Exception {
```
jed326, line 202: Why do we have both of these?

Rajrahane:
@anntians can you refer this conversation and discuss runner cleanup for faiss-base and localstack images with @peterzhuamazon
 
https://github.com/opensearch-project/remote-vector-index-builder/pull/58#issuecomment-2822387635

src/testFixtures/java/org/opensearch/knn/KNNRestTestCase.java
anntians: `setupRemoteIndexBuildSettingsTestFallback` is used to test fallback mechanism when remote build is not enabled. And `setupRemoteIndexBuildSettings` sets up config when remote build is enabled.

I could combine these two methods into one method, but will still keep the 
```
if (randomBoolean() && isRemoteIndexBuildSupported(getBWCVersion()))
```
case to test fallback mechanism when remote build is not enabled.

jed326: Yeah I think having these 2 `@Before` methods is very confusing, please combine them into one. Seems like it should just be one if condition on `remoteBuild != null`?

src/test/java/org/opensearch/knn/index/AdvancedFilteringUseCasesIT.java
anntians: I don't think class annotations automatically applies to the test methods inside the class in Java

src/test/java/org/opensearch/knn/integ/KNNScriptScoringIT.java
```diff
@@ -795,7 +795,7 @@ private Map<String, KNNResult> createDataset(
         final Map<String, KNNResult> dataset = new HashMap<>(dense ? numDocsWithField : numDocsWithField * 3);
         int id = 0;
         for (int i = 0; i < numDocsWithField; i++) {
-            final int dummyDocs = dense ? 0 : randomIntBetween(2, 5);
+            final int dummyDocs = dense ? 0 : 5;
```
anntians, line 798: The ITs in this class are flaky in triggering remote build, most likely due to randomness in how many docs to ingest

src/test/java/org/opensearch/knn/integ/ExpandNestedDocsIT.java
```diff
@@ -168,9 +185,9 @@ public void testExpandNestedDocs_whenFilteredOnNestedFieldDoc_thenReturnFiltered
     }
 
     @SneakyThrows
```
anntians, line 187: Annotations can't be scoped down to a parameter set for ITs that are parameterized, as they are applied to the entire test method. In this case, only the parameter set `Faiss with float format and in memory mode` triggers remote build in this test. To only mark this parameter set for remote build check, I'm setting the `EXPECT_REMOTE_BUILD` flag in `KNNRestTestCase`.

The issue is that now we have two ways of marking a test for remote build check: annotations and `EXPECT_REMOTE_BUILD` (for parameterized tests). This might get confusing in future.

The alternative is to just match ITs against hard-coded list of test names to check for remote build, but that can get messy in future as well.

src/test/java/org/opensearch/knn/index/AdvancedFilteringUseCasesIT.java
jed326: Yes you have to implement that logic, we can take it as a follow-up.

src/testFixtures/java/org/opensearch/knn/KNNRestTestCase.java
```diff
@@ -173,17 +187,85 @@ public void cleanUpCache() throws Exception {
...
+            // working correctly.
+            updateClusterSettings(KNNFeatureFlags.KNN_REMOTE_VECTOR_BUILD_SETTING.getKey(), true);
+            updateClusterSettings(KNNSettings.KNN_REMOTE_VECTOR_REPO, "integ-test-repo");
+        }
+    }
+
+    @Retention(RetentionPolicy.RUNTIME)
+    @Target(ElementType.METHOD)
+    public @interface ExpectRemoteBuildValidation {
+    }
```
navneet1v, line 211: Loved this. Thanks @anntians for contributing the annotation this should really help ease out the development.

src/testFixtures/java/org/opensearch/knn/KNNRestTestCase.java
```diff
@@ -173,17 +187,85 @@ public void cleanUpCache() throws Exception {
...
+            // Set up cluster settings for remote index build feature. We do this for all tests to ensure the fallback mechanisms are
+            // working correctly.
+            updateClusterSettings(KNNFeatureFlags.KNN_REMOTE_VECTOR_BUILD_SETTING.getKey(), true);
+            updateClusterSettings(KNNSettings.KNN_REMOTE_VECTOR_REPO, "integ-test-repo");
+        }
+    }
+
+    @Retention(RetentionPolicy.RUNTIME)
+    @Target(ElementType.METHOD)
+    public @interface ExpectRemoteBuildValidation {
```
navneet1v, line 210: can move this annotation to a separate file?

navneet1v (approved):

.github/workflows/remote_index_build.yml
```diff
@@ -73,7 +73,10 @@ jobs:
       - name: Initial cleanup
         run: |
           docker ps -aq | xargs -r docker rm -f
-          docker system prune -af --volumes
+          docker container prune -f
```
anntians, line 76: Update docker prune command following https://github.com/opensearch-project/remote-vector-index-builder/pull/58#issuecomment-2822387635.

navneet1v (approved):

naveentatikonda (approved):

PR #2663, opened 2025-04-16
Fix concurrency bug to share non-thread safe graph structure.

0ctopus13prime:
### Description
This fixes concurrency bug that shares non-threading safe graph structure among search threads.

### Related Issues
N/A

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissMemoryOptimizedSearcher.java
+++ b/src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissMemoryOptimizedSearcher.java
@@ -26,14 +26,13 @@ public class FaissMemoryOptimizedSearcher implements VectorSearcher {
     private static final FlatVectorsScorer VECTOR_SCORER = FlatVectorScorerUtil.getLucene99FlatVectorsScorer();
 
     private final IndexInput indexInput;
-    private FaissIndex faissIndex;
-    private FaissHnswGraph faissHnswGraph;
+    private final FaissIndex faissIndex;
+    private final FaissHNSW hnsw;
 
     public FaissMemoryOptimizedSearcher(IndexInput indexInput) throws IOException {
         this.indexInput = indexInput;
         this.faissIndex = FaissIndex.load(indexInput);
-        final FaissHNSW hnsw = extractFaissHnsw(faissIndex);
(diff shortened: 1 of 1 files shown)


shatejas (approved):

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissMemoryOptimizedSearcher.java
```diff
@@ -26,14 +26,13 @@ public class FaissMemoryOptimizedSearcher implements VectorSearcher {
...
-    private FaissHnswGraph faissHnswGraph;
+    private final FaissIndex faissIndex;
+    private final FaissHNSW hnsw;
 
     public FaissMemoryOptimizedSearcher(IndexInput indexInput) throws IOException {
         this.indexInput = indexInput;
         this.faissIndex = FaissIndex.load(indexInput);
-        final FaissHNSW hnsw = extractFaissHnsw(faissIndex);
-        this.faissHnswGraph = new FaissHnswGraph(hnsw, indexInput);
+        this.hnsw = extractFaissHnsw(faissIndex);
```
Vikasht34, line 35: Should we do something like to ensure high throughput 

```
private final ThreadLocal<FaissHnswGraph> threadLocalGraph = ThreadLocal.withInitial(() -> new FaissHnswGraph(hnsw, indexInput));

```

Vikasht34 (approved):

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissMemoryOptimizedSearcher.java
0ctopus13prime: This was my initial thought and switched to create the instance per request as it have few pointers + int[32] at the worst case, does not look like it will impact search latency that much.

I just tested out, and verified there's no degradation with current approach!
Will resolve this, if you have concerns please let me know!

navneet1v: @0ctopus13prime can you please share the benchmarks with and without this change for better understanding of the impact?

navneet1v:
Can we see if we can add a unit test to validate this behavior? If this is impacting performance we should have some UT to ensure this change doesn’t revert

PR #2664, opened 2025-04-16
Switch derived to default and fix bug in byte derived

jmazanec15:
### Description
Enables derived source by default. Benchmarks in https://benchmarks.opensearch.org show significant improvements, so we should turn on by default. 

Lucene latencies:
image
Faiss store size:
image

In particular, this change will help configurations that leverage disk in some component.

### Related Issues

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/common/FieldInfoExtractor.java
+++ b/src/main/java/org/opensearch/knn/common/FieldInfoExtractor.java
@@ -9,6 +9,7 @@ import lombok.experimental.UtilityClass;
 import org.apache.commons.lang.StringUtils;
 import org.apache.lucene.index.FieldInfo;
 import org.apache.lucene.index.LeafReader;
+import org.apache.lucene.index.VectorEncoding;
 import org.opensearch.common.Nullable;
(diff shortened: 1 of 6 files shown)


shatejas (approved):

Vikasht34 (dismissed):

Vikasht34 (approved):

naveentatikonda (approved):

src/main/java/org/opensearch/knn/common/FieldInfoExtractor.java
```diff
@@ -60,6 +61,10 @@ public static VectorDataType extractVectorDataType(final FieldInfo fieldInfo) {
             if (modelMetadata != null) {
                 VectorDataType vectorDataType = modelMetadata.getVectorDataType();
                 vectorDataTypeString = vectorDataType == null ? null : vectorDataType.getValue();
+            } else if (fieldInfo.hasVectorValues()) {
```
shatejas, line 64: I am surprised this didn't cause any issues before - probably wasn't being use in any lucene engine code paths. Thanks for fixing it!

Can we update the unit tests as a follow up?

Issue #2665, opened 2025-04-17
[BUG] Quantization cache limit is in kb but objects are weights computed with bytes


jmazanec15:
It seems we are setting weight limit with KB but weight objects with Byte ([https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/[…]antization/models/quantizationState/QuantizationStateCache.java](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/quantization/models/quantizationState/QuantizationStateCache.java#L69-L71)). As a consequence, instead of limit being 5% of jvm it will be 0.005%. So for 32 GB heap, this will be a limit of 1.6 MB. 

I dont think we hit this frequently unless we really have high scale. Not sure what impact is, but at the very least, it pollutes the logs.

0ctopus13prime:
PR is raised : https://github.com/opensearch-project/k-NN/pull/2666

0ctopus13prime:
# Root cause analysis
The bug is the result of the two combinations listed below:
1. Multiple threads are trying to open input stream, which internally mmap the file - Code
As a result, there are several mappings exists for a one single file.

2. We are passing IOContext.DEFAULT when opening a quantization state config - Code
* Because of this, internally Lucene uses a shared arena to mmap the file - Code
* But by the time Native KNN Vector reader tries to open the quantization config, other files are already mmaped by the directory it is using. (Other files refer to physical files belong to the same segment. .vec file for example)
  * This already had created a shared Arena.
* The shared area is a nothing but a wrapper having an arena instance obtained from Java API with a reference count.
* Therefore, by the time a quantization config is loaded, the reference count of the shared arena is already greater than 0 -> This is important.
* When IndexInput closes after finished loading quantization config, it calls close.
* BUT! Calling close merely decrease the reference count within the shared area which is not zero.
* Therefore, even there's no query is incoming, we are still seeing dangling mapped file.


# Reproduce
```
import org.apache.lucene.store.IOContext;
import org.apache.lucene.store.IndexInput;
import org.apache.lucene.store.MMapDirectory;

import java.io.IOException;
import java.nio.file.Path;

public class KdyTest {
    public static void main(final String... args) throws IOException {
        // Print PID
        long pid = ProcessHandle.current().pid();
        System.out.println("PID: " + pid);

        // Define Lucene directory
        String dirPath = "/home/ec2-user/cache-bug";
[...]

PR #2666, opened 2025-04-17
Fix quantization cache bugs.

0ctopus13prime:
### Description
This PR fixes two bugs:
1. Scale issue : We are using KB as the limit as oppose to use bytes. As a result, even the value occupies few bytes, it is interpreted as KB leading to frequent cache eviction.
2. Non thread safe : Since Check-then-update logic is not properly guarded, leading to multiple threads are loading the meta info from the disk and update the value, which is unnecessary though.

### Related Issues
https://github.com/opensearch-project/k-NN/issues/2665

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/quantization/models/quantizationState/QuantizationStateCacheManager.java
+++ b/src/main/java/org/opensearch/knn/quantization/models/quantizationState/QuantizationStateCacheManager.java
@@ -42,24 +42,12 @@ public final class QuantizationStateCacheManager implements Closeable {
      * @return The associated QuantizationState
      */
     public QuantizationState getQuantizationState(QuantizationStateReadConfig quantizationStateReadConfig) throws IOException {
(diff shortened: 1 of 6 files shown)


0ctopus13prime:
Working on adding unit tests at the moment.
Will update shortly.

src/main/java/org/opensearch/knn/quantization/models/quantizationState/QuantizationStateCacheManager.java
```diff
@@ -1,3 +1,4 @@
+
```
jmazanec15, line 1: remove

0ctopus13prime: oops, sure

jmazanec15 (approved):
LGTM - thanks @0ctopus13prime great find!

Vikasht34 (approved):
LGTM !! Thanks @0ctopus13prime

PR #2675, opened 2025-04-24
[LuceneOnFaiss M2, PR1] Added binary flat format in Faiss index for memory optimized search.

0ctopus13prime:
### Description
This PR is to support binary flat format in memory optimized search. 
With 8x, 16x, 32x compression in `on_disk` mode, Faiss will compress the vector.
As a result of compression, each vector will have `codeSize` amount of bytes in flat vector. 

For example, 100 dimension float vectors will be compressed down to 50 bytes with 8x compression.
e.g. (4 * 100) / 8 = 50 bytes. 

### Related Issues

### Check List


src/main/java/org/opensearch/knn/memoryoptsearch/faiss/binary/FaissBinaryIndex.java
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    /**
+     * Read common header for binary format.
+     * FYI Faiss - <a href="https://github.com/facebookresearch/faiss/blob/main/faiss/impl/index_read.cpp#L1284">...</a>
+     *
+     * @param inputStream Input stream reading bytes from Faiss index file.
+     * @throws IOException
+     */
+    public void readCommonBinaryHeader(final IndexInput inputStream) throws IOException {
```
shatejas, line 36: Can't you just override `readCommonHeader`?

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/binary/FaissIndexBinaryFlat.java
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    @Override
+    public VectorEncoding getVectorEncoding() {
+        return VectorEncoding.BYTE;
+    }
+
+    @Override
+    public FloatVectorValues getFloatValues(IndexInput indexInput) throws IOException {
+        throw new UnsupportedOperationException();
```
jmazanec15, line 60: nit: Can you add message?

src/test/java/org/opensearch/knn/memoryoptsearch/FaissIndexBinaryFlatTests.java
```diff
@@ -0,0 +1,103 @@
...
+        IndexInput indexInput = loadFlatBinaryVectors();
+        final FaissIndex faissIndex = FaissIndexFloatFlat.load(indexInput);
+
+        // Prepare a new input stream
+        indexInput = loadFlatBinaryVectors();
+
+        // Validate it
+        final ByteVectorValues values = faissIndex.getByteValues(indexInput);
+        assertEquals(BINARY_DIMENSION, values.dimension());
+        assertEquals(NUM_VECTORS, values.size());
```
shatejas, line 74: nit: do we want to pick a few random vectorValues and test if its the same?

jmazanec15 (commented):
looks good - just a few nits

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/binary/FaissIndexBinaryFlat.java
0ctopus13prime: sure, will do in the next rev

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/binary/FaissBinaryIndex.java
0ctopus13prime: Unfortunately, we cannot reuse it.
They are little bit different.

Common header fields

| Element no  | Common header fields  | Binary common header fields |
| ------------- | ------------- | ------------- |
| 1. | Dimension, int32  |  Dimension, int32  |
| 2. | Total number of vectors, int64  | Code size, int32 |
| 3. | Dummy field 1, int64  | Total number of vectors, int64  |
| 4. | Dummy field 2, int64  | Is trained?, bool  |
| 5. | Is trained?, bool  | Metric type, int32  |
| 6. | Metric type, int32  | -  |

src/test/java/org/opensearch/knn/memoryoptsearch/FaissIndexBinaryFlatTests.java
0ctopus13prime: oops, I thought I had included the random vector access testings, but I had not added it.
Will update in the next rev.

Vikasht34 (approved):
LGTM !!

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/binary/FaissIndexBinaryFlat.java
```diff
@@ -0,0 +1,96 @@
...
+ * 2. A quantized flat vector section
+ * <p>
+ * The quantized vector section contains a list of vectors, each of size `codeSize`.
+ * For example, applying 8x compression to a 100-dimensional float vector results in
+ * a `codeSize` of 50 bytes per vector, calculated as (4 * 100) / 8.
+ * <p>
+ * Note: Quantized vectors should be compared using Hamming distance only.
+ * See <a href="https://github.com/facebookresearch/faiss/blob/main/faiss/IndexBinaryFlat.h">...</a> for more details.
+ */
+public class FaissIndexBinaryFlat extends FaissBinaryIndex {
```
navneet1v, line 31: Will there ever be another implementation of BinaryIndex apart from Flat?

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/binary/FaissIndexBinaryFlat.java
```diff
@@ -0,0 +1,96 @@
...
+import org.opensearch.knn.memoryoptsearch.faiss.FaissSection;
+
+import java.io.IOException;
+
+/**
+ * Binary flat vector format for a Faiss index.
+ * <p>
+ * The format consists of two parts:
+ * 1. A binary header
+ * 2. A quantized flat vector section
```
navneet1v, line 22: Is this really quantized or this is just the binary vectors represented as bytes

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/binary/FaissIndexBinaryFlat.java
```diff
@@ -0,0 +1,96 @@
...
+ * <p>
+ * The quantized vector section contains a list of vectors, each of size `codeSize`.
+ * For example, applying 8x compression to a 100-dimensional float vector results in
+ * a `codeSize` of 50 bytes per vector, calculated as (4 * 100) / 8.
+ * <p>
```
navneet1v, line 27: Am I missing something here? We are taking about quantized vectors here, but we do have the support for binary vectors with Faiss i.e. a binary faiss index is not only storing the quantized fp32 vectors but also standard binary vectors provided to k-NN plugin. 

Please correct me if meaning of quantized vectors is something else here.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/binary/FaissIndexBinaryFlat.java
0ctopus13prime: It will be byte quantized vectors having size of `codeSize`.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/binary/FaissIndexBinaryFlat.java
0ctopus13prime: Yeah, that's correct. We do support binary vectors as well from user to input. The description is there to give a context of `codeSize` along with the quantization framework we provide.

jmazanec15 (approved):

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/binary/FaissIndexBinaryFlat.java
navneet1v: I think that is redundant, we should remove it.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/binary/FaissIndexBinaryFlat.java
navneet1v: I think quantization words here is creating confusion. I think we should simplify the docs

0ctopus13prime: Hmm.. we can add another explanation to the comment as well for the future reference.
Looks like your comment is about why pure binary case is not included in the comment right?

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/binary/FaissIndexBinaryFlat.java
0ctopus13prime: Can we just keep it?? I believe it will be helpful for the future debugging since without it, it will be only me know what does `codeSize` really mean in our quantization framework.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/binary/FaissBinaryIndex.java
shatejas: I understand they are different, I suggested overriding it not reusing it. FaissIndex already had readHeader, Ideally we would want to make that abstract and enforce the behavior on various indices but I am not sure if its currently possible. So override the readHeader method from faissIndex instead of creating a new signature

0ctopus13prime: Ah, that makes sense. Will change it in the next PR (https://github.com/opensearch-project/k-NN/pull/2676)

PR #2676, opened 2025-04-24
[LuceneOnFaiss M2, PR2] Added binary HNSW graph support in Faiss index for memory optimized search

0ctopus13prime:
### Description
This PR is to support binary HNSW in memory optimized search.
The binary HNSW consists of two parts:
1. HNSW
2. Binary flat vectors

In the change, added `FaissBinaryHnswIndex` along with an util class having a common loading functions that being used when loading both binary index and non-binary index.

### Related Issues
N/A

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissIndexLoadUtils.java
+++ b/src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissIndexLoadUtils.java
@@ -1,0 +1,48 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.knn.memoryoptsearch.faiss;
+
+import lombok.experimental.UtilityClass;
+import org.apache.lucene.store.IndexInput;
+import org.opensearch.knn.memoryoptsearch.faiss.binary.FaissBinaryIndex;
+
+import java.io.IOException;
+
+/**
+ * Util class being used during partial load each section in Faiss index file.
+ * Refer to {@link FaissIndex#doLoad(IndexInput)} for more details regarding how it is being used.
+ */
+@UtilityClass
+public class FaissIndexLoadUtils {
+    /**
+     * Each section of Faiss index must start four leading characters indicating its index type.
+     * This is being used to read four bytes, then convert them to String.
+     *
+     * @param input IndexInput reading bytes from Faiss index.
+     * @return A string of index type.
+     * @throws IOException
+     */
+    public static String readIndexType(final IndexInput input) throws IOException {
+        final byte[] fourBytes = new byte[4];
+        input.readBytes(fourBytes, 0, fourBytes.length);
+        return new String(fourBytes);
+    }
+
+    /**
+     * Util function to convert {@link FaissIndex} to {@link FaissBinaryIndex}.
+     * If it fails to cast the given index instance, it will throw {@link IllegalArgumentException}.
+     *
+     * @param index Target index instance to be casted to {@link FaissBinaryIndex}.
+     * @return Casted {@link FaissBinaryIndex}.
+     */
+    public static FaissBinaryIndex toBinaryIndex(final FaissIndex index) {
+        if (index instanceof FaissBinaryIndex binaryIndex) {
+            return binaryIndex;
+        }
+
+        throw new IllegalArgumentException("Failed to convert [" + index.getIndexType() + "] to FaissBinaryIndex.");
+    }
+}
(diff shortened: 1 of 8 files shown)


0ctopus13prime:
Testing is failing as it requires to partial load `IBxF` section through FaissIndexBinaryFlat, which is covered in [PR1](https://github.com/opensearch-project/k-NN/pull/2675).
Once PR1 is merged, it will pass the testing.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/binary/FaissBinaryHnswIndex.java
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    @Override
+    public VectorEncoding getVectorEncoding() {
+        return storage.getVectorEncoding();
+    }
+
+    @Override
+    public FloatVectorValues getFloatValues(IndexInput indexInput) throws IOException {
+        return storage.getFloatValues(indexInput);
```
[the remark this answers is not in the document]
0ctopus13prime: will replace it to throw an exception in the next rev.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissIndexLoadUtils.java
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    /**
+     * Util function to convert {@link FaissIndex} to {@link FaissBinaryIndex}.
+     * If it fails to cast the given index instance, it will throw {@link IllegalArgumentException}.
+     *
+     * @param index Target index instance to be casted to {@link FaissBinaryIndex}.
+     * @return Casted {@link FaissBinaryIndex}.
+     */
+    public static FaissBinaryIndex toBinaryIndex(final FaissIndex index) {
```
jmazanec15, line 41: So this is different from faiss model where index and index_binary are separate types, right?

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/binary/FaissBinaryHnswIndex.java
```diff
@@ -0,0 +1,71 @@
...
+     * @param input
+     * @throws IOException
+     */
+    @Override
+    protected void doLoad(IndexInput input) throws IOException {
+        // Read common binary index header
+        readCommonHeader(input);
+
+        // Partial load HNSW graph
+        faissHnsw.load(input, getTotalNumberOfVectors());
```
jmazanec15, line 51: Why doesnt this return something?

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissIndexLoadUtils.java
0ctopus13prime: Yes, in Java there's no equivalent way to implement [CRTP](https://en.cppreference.com/w/cpp/language/crtp) (curiously recurring template pattern) that Faiss is using for IdMap - [Link](https://github.com/facebookresearch/faiss/blob/main/faiss/IndexIDMap.h#L21)
To faithfully follow Faiss, then we will need to have two identical IdMap classes for each FaissIndex and FaissBinaryIndex. So I made FaissBinaryIndex inherit FaissIndex to solve CRTP. And I think it does make sense as FaissBinaryIndex can be thought of a sub-type of FaissIndex.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/binary/FaissBinaryHnswIndex.java
0ctopus13prime: I don't think I'm following, could you share more context?
`load` method tells FaissHnsw to load via the given `IndexInput` to fill its member fields, hence it has a `void load(...)` signature.

jmazanec15 (approved):

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/binary/FaissBinaryIndex.java
```diff
@@ -33,7 +33,7 @@ public FaissBinaryIndex(final String indexType) {
      * @param inputStream Input stream reading bytes from Faiss index file.
      * @throws IOException
      */
-    public void readCommonBinaryHeader(final IndexInput inputStream) throws IOException {
+    protected void readCommonHeader(final IndexInput inputStream) throws IOException {
```
shatejas, line 36: nit:
```
@Override
```

shatejas (approved):

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/binary/FaissBinaryIndex.java
0ctopus13prime: Will update in the next PR!

PR #2678, opened 2025-04-24
[LuceneOnFaiss M2, PR4] Support hamming distance in memory optimized search.

0ctopus13prime:
### Description
This PR is to add a hamming distance calculation for memory optimized search.
Also it looses constraints in `MemoryOptimizedSearchSupportSpec` to allow memory optimized search for disk mode. 
But note that PQ is still not in the scope, so it will reject PQ based field type having model id.

### Related Issues
N/A

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/common/KNNConstants.java
+++ b/src/main/java/org/opensearch/knn/common/KNNConstants.java
@@ -112,6 +112,7 @@ public class KNNConstants {
     public static final String METHOD_PARAMETER_NPROBES = "nprobes";
     public static final String ENCODER_FLAT = "flat";
     public static final String ENCODER_PQ = "pq";
+    public static final String ENCODER_BINARY = "binary";
     public static final String ENCODER_PARAMETER_PQ_M = "m";
     public static final String ENCODER_PARAMETER_PQ_CODE_SIZE = "code_size";
     public static final String FAISS_HNSW_DESCRIPTION = "HNSW";
--- a/src/main/java/org/opensearch/knn/index/engine/MemoryOptimizedSearchSupportSpec.java
+++ b/src/main/java/org/opensearch/knn/index/engine/MemoryOptimizedSearchSupportSpec.java
@@ -5,15 +5,13 @@
 
 package org.opensearch.knn.index.engine;
 
-import org.opensearch.knn.index.SpaceType;
-import org.opensearch.knn.index.VectorDataType;
-import org.opensearch.knn.index.engine.qframe.QuantizationConfig;
 import org.opensearch.knn.memoryoptsearch.VectorSearcher;
 
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 
+import static org.opensearch.knn.common.KNNConstants.ENCODER_BINARY;
 import static org.opensearch.knn.common.KNNConstants.ENCODER_FLAT;
 import static org.opensearch.knn.common.KNNConstants.ENCODER_SQ;
 import static org.opensearch.knn.common.KNNConstants.METHOD_ENCODER_PARAMETER;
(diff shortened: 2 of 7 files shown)


src/main/java/org/opensearch/knn/common/KNNConstants.java
```diff
@@ -112,6 +112,7 @@ public class KNNConstants {
     public static final String METHOD_PARAMETER_NPROBES = "nprobes";
     public static final String ENCODER_FLAT = "flat";
     public static final String ENCODER_PQ = "pq";
+    public static final String ENCODER_BINARY = "binary";
```
jmazanec15, line 115: This is available in QFrameBitEncoder. Can we define in one place?

src/main/java/org/opensearch/knn/index/engine/MemoryOptimizedSearchSupportSpec.java
```diff
@@ -63,29 +61,16 @@ public static boolean supported(
...
-                return false;
-            }
-
-            // L2 or Inner product are supported.
-            if (methodContext.getSpaceType() != SpaceType.L2 && methodContext.getSpaceType() != SpaceType.INNER_PRODUCT) {
-                return false;
-            }
-
-            // We only support Flat and SQ encoder for HNSW.
+            // We only support Flat, SQ and binary encoder for HNSW.
```
jmazanec15, line 64: Have we tested with disk mode yet? Do we know if page cache evictions of graph caused by re-scoring will hamper performance?

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FlatVectorsScorerProvider.java
```diff
@@ -0,0 +1,76 @@
...
+                    + knnVectorValues.getClass().getSimpleName()
+            );
+        }
+
+        @Override
+        public RandomVectorScorerSupplier getRandomVectorScorerSupplier(
+            VectorSimilarityFunction vectorSimilarityFunction,
+            KnnVectorValues knnVectorValues
+        ) {
+            throw new UnsupportedOperationException();
```
jmazanec15, line 64: Exception message?

src/main/java/org/opensearch/knn/common/KNNConstants.java
0ctopus13prime: sure, will update in the next rev

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FlatVectorsScorerProvider.java
0ctopus13prime: will update in the next rev

src/main/java/org/opensearch/knn/index/engine/MemoryOptimizedSearchSupportSpec.java
0ctopus13prime: I've personally tested disk mode with this, and after all PRs got merged into mainline, will do the deep dive then share in GH.

shatejas (approved):
Looks good

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FlatVectorsScorerProvider.java
```diff
@@ -0,0 +1,76 @@
...
+
+    public static FlatVectorsScorer getFlatVectorsScorer(final KNNVectorSimilarityFunction similarityFunction) {
+        if (similarityFunction == KNNVectorSimilarityFunction.HAMMING) {
+            return HAMMING_VECTOR_SCORER;
+        }
+
+        return DELEGATE_VECTOR_SCORER;
+    }
+
+    private static class HammingFlatVectorsScorer implements FlatVectorsScorer {
```
shatejas, line 33: Should we be contributing to lucene with this scorer? With knn leveraging more and more lucene code we should start contributing rather than having wrappers around it 

@jmazanec15 @0ctopus13prime Thoughts?

0ctopus13prime: yeah, that sounds good. 
Internally it ends up using VectorUtil anyway. Will make a small PR to see what Lucene folks think.
Thank you.

jmazanec15 (approved):

PR #2687, opened 2025-04-30
Adding serializer and api implementation for segment profiler state

oaganesh:
### Description
Adding implementation for profiling and statistical analysis of KNN vector segments in OpenSearch with compression when benchmarking. Provides functionality to collect, process, and store statistical information about vector dimensions across different shards and segments.

### Related Issues
Implements #2622 

### Check List
- [✔️  ] New functionality includes testing.
- [✔️ ] New functionality has been documented.
- [✔️ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
- [✔️ ] Commits are signed per the DCO using `--signoff`.
- [✔️ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).

For more information on following Developer Certificate of Origin and signing off your commits, please check [✔️](https://github.com/opensearch-project/k-NN/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

Example input:
```
curl -XPUT "http://localhost:9200/target_index" -H 'Content-Type: application/json' -d'{      
  "settings": {
    "index": {
      "knn": true,
      "number_of_shards": 1,
      "number_of_replicas": 0
    }
  },                                                                                            
  "mappings": {
    "properties": {
      "target_field": {
        "type": "knn_vector",
        "dimension": 128,
        "space_type": "l2",
        "compression_level": "16x",
        "method": {
          "name": "hnsw",
          "space_type": "l2",
          "engine": "faiss",
          "parameters": {
            "ef_construction": 100,
            "m": 16
          }
        }
      }
    }
  }
}'

```

```
curl -XPOST "http://localhost:9200/target_index/_doc" -H 'Content-Type: application/json' -d'{
  "target_field": [0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1.0, 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1.0, 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1.0, 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1.0, 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1.0, 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1.0, 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1.0, 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1.0, 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1.0, 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1.0, 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1.0, 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 1.0, 1.2, 1.3, 1.4, 1.5, 1.6, 1.7, 1.8, 1.9, 2.0]
}'
```

```
curl -X GET "localhost:9200/_plugins/_knn/warmup/target_index?pretty" 
```

```
curl -X GET "localhost:9200/_plugins/_knn/profile/target_index?pretty"
```

Example output:

```
  "total_shards" : 1,
  "successful_shards" : 1,
  "failed_shards" : 0,
  "profile_results" : {
    "target_index" : {
      "dimensions" : [
        {
          "min" : 1.3,
          "max" : 1.9,
          "variance" : 0.0,
          "mean" : 3.8,
          "count" : 5,
          "std_deviation" : 0.0,
          "sum" : 1.9,
          "dimension" : 0
        },
        {
          "min" : 0.2,
          "max" : 1.89
          "variance" : 0.2,
          "mean" : 3.78,
          "count" : 5,
          "std_deviation" : 0.3,
          "sum" : 1.89,
          "dimension" : 1
        }
      ]
    }
  }
}

```


CHANGELOG.md
```diff
@@ -14,6 +14,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased 2.x](https://github.com/opensearch-project/k-NN/compare/2.19...2.x)
 ### Features
+* [Vector Profiler] Adding basic generic vector profiler implementation and tests. [#2624](https://github.com/opensearch-project/k-NN/pull/2624)
+* [Vector Profiler] Updating serializer, compression, and api. [#2687](https://github.com/opensearch-project/k-NN/pull/2687)
```
markwu-sde, line 18: Should we be a little more specific on what we're updating? 

Also AFAIK we're not updating anything related to how compression is done or any API (but we're introducing one). Should we reword this?

src/main/java/org/opensearch/knn/index/KNNIndexShard.java
```diff
@@ -82,6 +87,103 @@ public String getIndexName() {
...
+            if (segmentLevelProfilerStates.isEmpty()) {
+                log.info("[KNN] No segment profiles were collected for field: {} in shard: {}", fieldName, indexShard.shardId());
+                return shardVectorProfile; // Return empty list
+            }
+
+            log.info("[KNN] Collected {} segment profiles", segmentLevelProfilerStates.size());
+
+            // Get dimension and validate all segments have the same dimension
+            int dimension = segmentLevelProfilerStates.get(0).getDimension();
+            boolean dimensionsMatch = segmentLevelProfilerStates.stream().allMatch(state -> state.getDimension() == dimension);
```
markwu-sde, line 128: Would there ever be a case where the dimensions won't match accordingly here? Do we need to check for this?

src/main/java/org/opensearch/knn/index/KNNIndexShard.java
```diff
@@ -82,6 +87,103 @@ public String getIndexName() {
...
+
+            if (!dimensionsMatch) {
+                log.error("[KNN] Inconsistent dimensions found across segments");
+                return shardVectorProfile; // Return empty list
```
markwu-sde, line 132: Nit: you can return List.of()

src/main/java/org/opensearch/knn/index/KNNIndexShard.java
```diff
@@ -82,6 +87,103 @@ public String getIndexName() {
...
+            log.info("[KNN] Vector dimension: {}", dimension);
+
+            // Transpose our list to aggregate per dimension
+            for (int i = 0; i < dimension; i++) {
+                final int dimensionId = i;
+                List<SummaryStatistics> transposed = new ArrayList<>();
+
+                for (SegmentProfilerState state : segmentLevelProfilerStates) {
+                    List<SummaryStatistics> stateStats = state.getStatistics();
+                    if (dimensionId < stateStats.size()) {
```
markwu-sde, line 144: Again - if the dimension is constant here do we really need this check? it will always be within bounds.

src/main/java/org/opensearch/knn/index/KNNIndexShard.java
```diff
@@ -82,6 +87,103 @@ public String getIndexName() {
...
+                        SummaryStatistics stat = stateStats.get(dimensionId);
+                        if (stat != null) {
```
markwu-sde, line 146: When can this be possible?

src/main/java/org/opensearch/knn/index/KNNIndexShard.java
```diff
@@ -82,6 +87,103 @@ public String getIndexName() {
...
+                    stats.getSum(),
+                    stats.getVariance(),
+                    Math.sqrt(stats.getVariance())
+                );
+            }
+
+            log.info("[KNN] Profiling completed for field: {} in shard: {}", fieldName, indexShard.shardId());
+        } catch (Exception e) {
+            log.error(
+                "[KNN] Critical error during profiling for field: {} in shard: {}: {}",
```
markwu-sde, line 176: Critical sounds a bit too extreme here given we're not in the hot path.

src/main/java/org/opensearch/knn/index/KNNIndexShard.java
```diff
@@ -235,4 +337,13 @@ static class EngineFileContext {
...
         private final SegmentInfo segmentInfo;
     }
+
+    /**
+     * Profile the vector fields in this shard with default field name.
+     *
+     * @return List of statistical summaries for each dimension
+     */
+    public List<StatisticalSummaryValues> profile() {
+        return profile("target_field");
```
markwu-sde, line 347: We shouldn't hard code a field name here. I understand we're doing this from a benchmarking OSB perspective but if we want customers to be able to leverage this feature we need to provide the ability to take in a field/index name as input.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNN990QuantizationStateReader.java
```diff
@@ -54,7 +54,7 @@ public static QuantizationState read(QuantizationStateReadConfig readConfig) thr
         String quantizationStateFileName = getQuantizationStateFileName(segmentReadState);
         int fieldNumber = segmentReadState.fieldInfos.fieldInfo(field).getFieldNumber();
 
-        try (IndexInput input = segmentReadState.directory.openInput(quantizationStateFileName, IOContext.READONCE)) {
+        try (IndexInput input = segmentReadState.directory.openInput(quantizationStateFileName, IOContext.DEFAULT)) {
```
markwu-sde, line 57: Could you give some context around this? Why are we changing from READONCE to DEFAULT?

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriter.java
```diff
@@ -107,9 +108,10 @@ public void flush(int maxDoc, final Sorter.DocMap sortMap) throws IOException {
                 field.getVectors()
             );
             final QuantizationState quantizationState = train(field.getFieldInfo(), knnVectorValuesSupplier, totalLiveDocs);
+            profile(field.getFieldInfo(), knnVectorValuesSupplier, totalLiveDocs);
             // should skip graph building only for non quantization use case and if threshold is met
             if (quantizationState == null && shouldSkipBuildingVectorDataStructure(totalLiveDocs)) {
-                log.debug(
+                log.info(
```
markwu-sde, line 114: NIT: use log.debug wherever possible

src/main/java/org/opensearch/knn/index/query/SegmentLevelQuantizationUtil.java
```diff
@@ -57,4 +59,23 @@ static QuantizationState getQuantizationState(final LeafReader leafReader, Strin
         }
         return tempCollector.getQuantizationState();
     }
+
+    /**
+     * A utility function to get {@link SegmentProfilerState} for a given segment and field.
+     * This needs to public as we are accessing this on a transport action
+     * TODO: move this out of this Util class and into another one.
```
markwu-sde, line 66: Same here. Are we going to act on this TODO?

src/main/java/org/opensearch/knn/index/query/SegmentLevelQuantizationUtil.java
```diff
@@ -57,4 +59,23 @@ static QuantizationState getQuantizationState(final LeafReader leafReader, Strin
...
+     * @param leafReader {@link LeafReader}
+     * @param fieldName {@link String}
+     * @return {@link SegmentProfilerState}
+     * @throws IOException exception during reading the {@link SegmentProfilerState}
+     */
+    public static SegmentProfilerState getSegmentProfileState(final LeafReader leafReader, String fieldName) throws IOException {
```
markwu-sde, line 72: We have the util function below that is a exact replica of this method. Do we still need this?

src/main/java/org/opensearch/knn/plugin/transport/KNNProfileResponse.java
```diff
@@ -0,0 +1,172 @@
...
+        out.writeVInt(shardResults.size());
+        for (KNNProfileShardResult result : shardResults) {
+            result.writeTo(out);
+        }
+    }
+
+    /**
+     * Get aggregated dimension statistics by index
+     */
+    private Map<String, Map<Integer, Map<String, Object>>> getAggregatedStats() {
```
markwu-sde, line 69: I'm assuming you want to get the aggregation at the field level here across all shards. 

1. Shouldn't this be by field/index instead of just index?
2. NIT: `getAggregatedStats` can be more specific because there are now multiple layers of aggregation

src/main/java/org/opensearch/knn/plugin/transport/KNNProfileResponse.java
```diff
@@ -0,0 +1,172 @@
...
+
+        for (KNNProfileShardResult result : shardResults) {
+            String indexName = result.getShardId().getIndexName();
+            List<StatisticalSummaryValues> stats = result.getDimensionStats();
+
+            if (stats == null || stats.isEmpty()) {
+                continue;
+            }
+
+            Map<Integer, Map<String, Object>> dimensions = indexDimensions.computeIfAbsent(indexName, k -> new HashMap<>());
```
markwu-sde, line 80: So the mapping that you're trying to do is:

Map<Index, Map<Dimension ID, Map<String,Object>>.

In lucene and specifically within k-nn a index can have multiple fields. Not all fields are of type `knn_vector`

src/main/java/org/opensearch/knn/plugin/transport/KNNProfileResponse.java
```diff
@@ -0,0 +1,172 @@
...
+
+            for (int i = 0; i < stats.size(); i++) {
+                StatisticalSummaryValues stat = stats.get(i);
```
markwu-sde, line 83: Please correct me if I'm wrong, but I think the crux of your situation here by looking at `StatisticalSummaryValues` is that the library we're using doesn't provide us with a aggregation technique for that object. 

If that is the case, how about we collect our SegmentProfileStates and put them in the shard response? That way we can craft both the shard/field level aggregation as we see fit since we have access to those values.

src/main/java/org/opensearch/knn/plugin/transport/KNNProfileResponse.java
```diff
@@ -0,0 +1,172 @@
...
+                if (stat == null) {
+                    continue;
+                }
+
+                int finalI = i;
+                Map<String, Object> dimension = dimensions.computeIfAbsent(i, k -> {
+                    Map<String, Object> newDim = new HashMap<>();
+                    newDim.put("dimension", finalI);
```
markwu-sde, line 91: NIT: static imports when possible

src/main/java/org/opensearch/knn/plugin/transport/KNNProfileShardResult.java
```diff
@@ -0,0 +1,87 @@
...
+        int size = in.readVInt();
+        this.dimensionStats = new ArrayList<>(size);
+        for (int i = 0; i < size; i++) {
+            double mean = in.readDouble();
+            double variance = in.readDouble();
+            long n = in.readVLong();
+            double min = in.readDouble();
+            double max = in.readDouble();
+            double sum = in.readDouble();
+            this.dimensionStats.add(new StatisticalSummaryValues(mean, variance, n, min, max, sum));
```
markwu-sde, line 54: If we're deserialize it here again would it be worthwhile to just save the deserialization at the profile API layer?

What kind of responses do we need to deserialize here?

src/main/java/org/opensearch/knn/plugin/transport/KNNWarmupTransportAction.java
```diff
@@ -89,7 +89,9 @@ protected EmptyResult shardOperation(KNNWarmupRequest request, ShardRouting shar
         KNNIndexShard knnIndexShard = new KNNIndexShard(
             indicesService.indexServiceSafe(shardRouting.shardId().getIndex()).getShard(shardRouting.shardId().id())
         );
+
         knnIndexShard.warmup();
+        knnIndexShard.profile();
```
markwu-sde, line 94: Do we need this in WarmupTransportAction?

src/main/java/org/opensearch/knn/profiler/SegmentProfilerState.java
```diff
@@ -0,0 +1,196 @@
...
+    /**
+     * Converts the SegmentProfilerState to a byte array for serialization
+     * @return
+     */
+    public byte[] toByteArray() {
+        try (BytesStreamOutput out = new BytesStreamOutput()) {
+            out.writeVInt(dimension);
+            out.writeVInt(statistics.size());
+            for (SummaryStatistics stat : statistics) {
+                out.writeDouble(stat.getMean());
```
markwu-sde, line 156: We shouldn't be serializing/deserializing the values but rather the object itself.

`SummaryStatistics` already implements the serializable interface so we should be able to convert it directly to bytes without needing to retrieve it's properties.

src/main/java/org/opensearch/knn/profiler/SegmentProfilerState.java
```diff
@@ -0,0 +1,196 @@
...
+     */
+    public static SegmentProfilerState fromBytes(byte[] bytes) {
+        try (StreamInput input = StreamInput.wrap(bytes)) {
+            int dimension = input.readVInt();
+            int statsSize = input.readVInt();
+            List<SummaryStatistics> statistics = new ArrayList<>(statsSize);
+
+            for (int i = 0; i < statsSize; i++) {
+                SummaryStatistics stat = new SummaryStatistics();
+                stat.addValue(input.readDouble());
```
markwu-sde, line 182: I think we've previously had a [discussion](https://github.com/opensearch-project/k-NN/pull/2653#discussion_r2049674320) on the deserialization aspect of this - the same issue is present here as well. 

Let's make sure to modify this as this would yield incorrect results for our profiler.

markwu-sde (commented):
One over-arching comment here is that we should have unit/integration tests for all our code. Make sure to include them in the PR so that we can test our functionality such as making sure the aggregation at the shard level is what we're expecting.

src/main/java/org/opensearch/knn/plugin/transport/KNNProfileResponse.java
```diff
@@ -0,0 +1,172 @@
...
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.HashMap;
+
+/**
+ * Response for KNN profile request
+ */
+@Log4j2
+public class KNNProfileResponse extends BroadcastResponse implements ToXContentObject {
```
markwu-sde, line 28: Have not looked too closely into these since I do have some comments on the serialization/deserialization component.

src/main/java/org/opensearch/knn/index/KNNIndexShard.java
oaganesh: Reverting back to previous way of implementing KNNIndexShard as current way is less robust.

src/main/java/org/opensearch/knn/index/KNNIndexShard.java
oaganesh: Understood, reverting back to previous way of implementing KNNIndexShard as current way is less robust.

src/main/java/org/opensearch/knn/index/KNNIndexShard.java
oaganesh: Removed specific field input

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNN990QuantizationStateReader.java
oaganesh: Mostly did it for consistency when accessing quantization state files multiple times.
https://lucene.apache.org/core/9_12_1/core/org/apache/lucene/store/IOContext.html?is-external=true

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriter.java
```diff
@@ -241,6 +248,23 @@ private QuantizationState train(
...
 
+    private SegmentProfilerState profile(
+        final FieldInfo fieldInfo,
+        final Supplier<KNNVectorValues<?>> knnVectorValuesSupplier,
+        final int totalLiveDocs
+    ) throws IOException {
+
+        SegmentProfilerState segmentProfilerState = null;
+        if (totalLiveDocs > 0) {
+            // TODO:Refactor to another init
```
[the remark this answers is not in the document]
oaganesh: Refactors NativeEnginesVectorsWriter class to encompass segmentStateWriter and initSegmentStateWriterIfNecessary()

src/main/java/org/opensearch/knn/index/query/SegmentLevelQuantizationUtil.java
oaganesh: The method has already been refactors into it's own class for under org.opensearch.knn.index.query. Need to remove the existing code.

src/main/java/org/opensearch/knn/profiler/SegmentProfilerState.java
oaganesh: Yes, this was just used to help test for compression by creating a new instance. Reverting back to pre-existing method for serializing and deserializing.

src/main/java/org/opensearch/knn/plugin/transport/KNNProfileResponse.java
```diff
@@ -0,0 +1,166 @@
...
+                        .field("variance", variance)
+                        .endObject();
+                }
+                builder.endArray().endObject();
+            }
+
+            builder.endObject();
+        }
+        builder.endObject();
+
```
markwu-sde, line 154: What about at the cluster level with every segment aggregated together? This output currently does this per shard right? Can we combine every `SummaryStatistics` in the cluster together?

src/main/java/org/opensearch/knn/plugin/transport/KNNProfileResponse.java
```diff
@@ -0,0 +1,166 @@
...
+                    double max = Double.NEGATIVE_INFINITY;
+                    double sumOfValues = 0.0;
+                    double sumOfSquares = 0.0;
+
+                    // Combine statistics properly
+                    for (SegmentProfilerState state : shardProfileResult.segmentProfilerStateList) {
+                        if (dim < state.getStatistics().size()) {
+                            SummaryStatistics stats = state.getStatistics().get(dim);
+
+                            totalCount += stats.getN();
```
markwu-sde, line 124: Should we leverage a existing library for aggregation?

1. This doesn't give an correct aggregation outside of min max and sum. 
2.  https://commons.apache.org/proper/commons-math/commons-math-docs/apidocs/src-html/org/apache/commons/math4/legacy/stat/descriptive/AggregateSummaryStatistics.html#line.305 - there's a library provided to us that aggregates a `SummaryStatistics` object for us. Can we use that instead of trying to combine it on our own?

src/main/java/org/opensearch/knn/plugin/transport/KNNProfileResponse.java
```diff
@@ -0,0 +1,166 @@
...
+            .field("successful_shards", getSuccessfulShards())
+            .field("failed_shards", getFailedShards())
+            .startObject("shard_profiles");
+
+        // Add shard profile results
+        for (KNNIndexShardProfileResult shardProfileResult : shardProfileResults) {
+            builder.startObject(shardProfileResult.shardId);
+
+            // Individual segment statistics
+            builder.startArray("segments");
```
markwu-sde, line 72: Can you provide an example of output in the Javadoc of what this would look like?

src/main/java/org/opensearch/knn/plugin/transport/KNNProfileResponse.java
```diff
@@ -0,0 +1,166 @@
...
+        List<DefaultShardOperationFailedException> shardFailures
+    ) {
+        super(totalShards, successfulShards, failedShards, shardFailures);
+
+        this.shardProfileResults = shardProfileResults;
+    }
+
+    @Override
+    public void writeTo(StreamOutput streamOutput) throws IOException {
+        super.writeTo(streamOutput);
```
markwu-sde, line 50: Let's see if we use this override for anything. If we don't we can throw an exception here to make sure we don't stray off the intended path

src/main/java/org/opensearch/knn/plugin/transport/KNNIndexShardProfileResult.java
```diff
@@ -0,0 +1,50 @@
...
+@AllArgsConstructor
+public class KNNIndexShardProfileResult implements Writeable {
+    List<SegmentProfilerState> segmentProfilerStateList;
+    String shardId;
+
+    /**
+     * Constructor for reading from StreamInput
+     */
+    public KNNIndexShardProfileResult(StreamInput streamInput) throws IOException {
+        this.shardId = streamInput.readString();
```
markwu-sde, line 27: Do we ever use this constructor?

src/main/java/org/opensearch/knn/plugin/transport/KNNProfileResponse.java
oaganesh: Added cluster level stats as well.

src/main/java/org/opensearch/knn/plugin/transport/KNNProfileResponse.java
oaganesh: Updated to use AggregateSummaryStatistics

src/main/java/org/opensearch/knn/plugin/transport/KNNProfileResponse.java
oaganesh: Updated javadocs

src/main/java/org/opensearch/knn/plugin/transport/KNNIndexShardProfileResult.java
oaganesh: No, it was used previously in KNNProfileResponse, but when removing the logic and the above constructor I was able to validate the output was still working.

CHANGELOG.md
```diff
@@ -14,6 +14,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased 2.x](https://github.com/opensearch-project/k-NN/compare/2.19...2.x)
 ### Features
+* [Vector Profiler] Adding basic generic vector profiler implementation and tests. [#2624](https://github.com/opensearch-project/k-NN/pull/2624)
```
markwu-sde, line 17: Why are we adding these changes? Weren't they already merged?

src/main/java/org/opensearch/knn/profiler/SegmentProfilerState.java
```diff
@@ -0,0 +1,174 @@
...
+
+import static org.apache.lucene.search.DocIdSetIterator.NO_MORE_DOCS;
+
+/**
+ * SegmentProfilerState is responsible for analyzing and profiling vector data within segments.
+ * This class calculates statistical measurements for each dimension of the vectors in a segment.
+ */
+@Log4j2
+@AllArgsConstructor
+public class SegmentProfilerState implements Serializable {
```
markwu-sde, line 33: Why is this file being shown as new? Isn't this [already](https://github.com/opensearch-project/k-NN/blob/feature/vector-profile/src/main/java/org/opensearch/knn/profiler/SegmentProfilerState.java) in the branch?

CHANGELOG.md
oaganesh: Yes, updated branch to adjust with changes already implemented in feature branch

src/main/java/org/opensearch/knn/profiler/SegmentProfilerState.java
oaganesh: Updated branch to adjust with changes already implemented in feature branch

src/main/java/org/opensearch/knn/plugin/transport/KNNProfileRequest.java
```diff
@@ -0,0 +1,54 @@
...
+import org.opensearch.action.support.broadcast.BroadcastRequest;
+import org.opensearch.core.common.io.stream.StreamInput;
+import org.opensearch.core.common.io.stream.StreamOutput;
+
+import java.io.IOException;
+
+@Getter
+public class KNNProfileRequest extends BroadcastRequest<KNNProfileRequest> {
+
+    private String index;
```
jmazanec15, line 19: Broadcast request should already have the indices: https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/action/support/broadcast/BroadcastRequest.java#L52.

src/main/java/org/opensearch/knn/plugin/transport/KNNProfileRequest.java
```diff
@@ -0,0 +1,54 @@
...
+    private String field;
```
jmazanec15, line 20: nit: lets make these final.

src/main/java/org/opensearch/knn/plugin/transport/KNNProfileRequest.java
```diff
@@ -0,0 +1,54 @@
...
+     */
+    public KNNProfileRequest(StreamInput in) throws IOException {
+        super(in);
+        index = in.readString();
+        field = in.readString();
+    }
+
+    @Override
+    public ActionRequestValidationException validate() {
+        return null;
```
jmazanec15, line 45: nit: no need to override

src/main/java/org/opensearch/knn/plugin/transport/KNNProfileResponse.java
```diff
@@ -0,0 +1,260 @@
...
+ *         "sum": 1032.45,
+ *         "variance": 0.099
+ *       }
+ *       // ... additional dimensions
+ *     ]
+ *   },
+ *   "failures": []
+ * }
+ */
+public class KNNProfileResponse extends BroadcastResponse implements ToXContentObject {
```
jmazanec15, line 99: nit: BroadcastResponse already implements ToXContentObject - do we need to do it again? https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/action/support/broadcast/BroadcastResponse.java

src/main/java/org/opensearch/knn/plugin/transport/KNNProfileResponse.java
```diff
@@ -0,0 +1,260 @@
...
+
+    List<KNNIndexShardProfileResult> shardProfileResults;
```
jmazanec15, line 101: nit: private final?

src/main/java/org/opensearch/knn/plugin/transport/KNNProfileResponse.java
```diff
@@ -0,0 +1,260 @@
...
+        List<DefaultShardOperationFailedException> shardFailures
+    ) {
+        super(totalShards, successfulShards, failedShards, shardFailures);
+
+        this.shardProfileResults = shardProfileResults;
+    }
+
+    @Override
+    public void writeTo(StreamOutput streamOutput) throws IOException {
+        throw new UnsupportedOperationException("This method is not available");
```
jmazanec15, line 123: Should we implement this?

src/main/java/org/opensearch/knn/plugin/transport/KNNProfileResponse.java
```diff
@@ -0,0 +1,260 @@
...
+    }
+
+    @Override
+    public XContentBuilder toXContent(XContentBuilder builder, Params params) throws IOException {
```
jmazanec15, line 127: I dont think we need to implement this method. See https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/action/support/broadcast/BroadcastResponse.java#L171-L183. We should just implement https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/action/support/broadcast/BroadcastResponse.java#L183.

src/main/java/org/opensearch/knn/plugin/transport/KNNProfileResponse.java
```diff
@@ -0,0 +1,260 @@
...
+            for (SegmentProfilerState state : shardProfileResult.segmentProfilerStateList) {
+                builder.startObject()
+                    .field("segment_id", state.getSegmentId())
+                    .field("dimension", state.getDimension())
+                    .startArray("vector_statistics");
+
+                for (int i = 0; i < state.getStatistics().size(); i++) {
+                    SummaryStatistics stats = state.getStatistics().get(i);
+                    builder.startObject()
+                        .field("dimension_index", i)
```
jmazanec15, line 150: Do we need this field?

src/main/java/org/opensearch/knn/plugin/transport/KNNProfileResponse.java
```diff
@@ -0,0 +1,260 @@
...
+                        .startObject("statistics")
+                        .field("count", stats.getN())
+                        .field("min", stats.getMin())
+                        .field("max", stats.getMax())
+                        .field("sum", stats.getSum())
+                        .field("mean", stats.getMean())
```
jmazanec15, line 156: Lets break the logic in this method into separate methods. It will be easier to modifu in future.

src/main/java/org/opensearch/knn/plugin/transport/KNNProfileResponse.java
```diff
@@ -0,0 +1,260 @@
...
+                            dimensionStats.add(state.getStatistics().get(dim));
+                        }
+                    }
+                }
+
+                StatisticalSummaryValues aggregatedStats = AggregateSummaryStatistics.aggregate(dimensionStats);
+
+                builder.startObject()
+                    .field("dimension_id", dim)
+                    .field("count", aggregatedStats.getN())
```
jmazanec15, line 236: lets put string literals in final static variables at top of class.

src/main/java/org/opensearch/knn/plugin/transport/KNNProfileTransportAction.java
```diff
@@ -0,0 +1,105 @@
...
+
+import java.io.IOException;
+import java.util.List;
+
+public class KNNProfileTransportAction extends TransportBroadcastByNodeAction<
+    KNNProfileRequest,
+    KNNProfileResponse,
+    KNNIndexShardProfileResult> {
+
+    public static Logger logger = LogManager.getLogger(KNNProfileTransportAction.class);
```
jmazanec15, line 35: replace this with `@Log4j2` annotation

src/main/java/org/opensearch/knn/plugin/transport/KNNWarmupTransportAction.java
```diff
@@ -89,6 +89,7 @@ protected EmptyResult shardOperation(KNNWarmupRequest request, ShardRouting shar
         KNNIndexShard knnIndexShard = new KNNIndexShard(
             indicesService.indexServiceSafe(shardRouting.shardId().getIndex()).getShard(shardRouting.shardId().id())
         );
+
```
jmazanec15, line 92: nit: remove for git history

src/main/java/org/opensearch/knn/profiler/SegmentProfileKNNCollector.java
```diff
@@ -0,0 +1,60 @@
...
+package org.opensearch.knn.profiler;
+
+import lombok.Getter;
+import lombok.Setter;
+import org.apache.lucene.search.KnnCollector;
+import org.apache.lucene.search.TopDocs;
+
+@Setter
+@Getter
+public class SegmentProfileKNNCollector implements KnnCollector {
```
jmazanec15, line 15: Add comment on top of this class.

src/test/java/org/opensearch/knn/plugin/transport/KNNProfileTransportActionIT.java
```diff
@@ -0,0 +1,191 @@
...
+
+    @SneakyThrows
+    protected Response executeProfileRequest(String indexName, String fieldName) throws IOException {
+        Request request = new Request("GET", "/_plugins/_knn/profile/" + indexName + "/" + fieldName);
+        Response response = client().performRequest(request);
+        assertEquals(RestStatus.OK, RestStatus.fromCode(response.getStatusLine().getStatusCode()));
+        return response;
+    }
+
+    private void validateProfileResponse(Response response, int dimension) throws IOException, ParseException {
```
jmazanec15, line 154: Itd be good to update this to ensure expected statistics are being computed. For instance, you can generate the vectors from a normalized distribution (i.e. 0-1) and confirm that the mean values is computed as close to 0.5.

src/main/java/org/opensearch/knn/plugin/transport/KNNIndexShardProfileResult.java
```diff
@@ -0,0 +1,34 @@
...
+import lombok.AllArgsConstructor;
+import org.opensearch.core.common.io.stream.StreamOutput;
+import org.opensearch.core.common.io.stream.Writeable;
+import org.opensearch.knn.profiler.SegmentProfilerState;
+
+import java.io.IOException;
+import java.util.List;
+
+@AllArgsConstructor
+public class KNNIndexShardProfileResult implements Writeable {
```
jmazanec15, line 17: Do we also need deserialization logic with this?

jmazanec15 (commented):
A few comments. I think overall structure is looking good. I think itd be good to enhance tests to ensure that we are validating functionality. Seems you are already working on that!

src/test/java/org/opensearch/knn/plugin/transport/KNNProfileTransportActionIT.java
markwu-sde: +1 agree with this. 

Make sure to validate that the aggregation is what'd we expect. Since merge is out of scope for this this can probably be done be explicitly creating the 2 segments and calling the profile API to make sure the aggregate is what we expect.

oaganesh: Updated tests

src/main/java/org/opensearch/knn/plugin/transport/KNNIndexShardProfileResult.java
oaganesh: I think it may be useful especially when compressing vectors.

PR #2693, opened 2025-05-02
[Remote Vector Index Build] fix: end remote build metrics before falling back to CPU, log exceptions

owenhalpert:
### Description
Because it is in the `finally` block, the call to `endRemoteBuildMetrics` is happening after the fallback strategy so the failure time is inaccurate. 

Additionally, Exceptions in the remote build process are caught and rethrown to let each step give a failure message:
https://github.com/opensearch-project/k-NN/blob/76d6662a5e80c44a924c23491cc7e99715cb5c43/src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java#L251-L253

We should add the exception itself to these debug logs to add more insight into why the failure occurred (for example, adding this to the `submitBuild` step will give the HTTP status code instead of just saying "Submit vector build failed". This is already the case for `awaitVectorBuild`.

### Related Issues
N/A

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
+++ b/src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
@@ -146,20 +146,20 @@ public class RemoteIndexBuildStrategy implements NativeIndexBuildStrategy {
 
             // 4. Download index file and write to indexOutput
             readFromRepository(indexInfo, repositoryContext, remoteBuildStatusResponse);
-
             success = true;
+            return;
         } catch (Exception e) {
-            fallbackStrategy.buildAndWriteIndex(indexInfo);
+            log.error("Failed to build index remotely: " + indexInfo, e);
         } finally {
             metrics.endRemoteIndexBuildMetrics(success);
         }
+        fallbackStrategy.buildAndWriteIndex(indexInfo);
     }
 
     /**
      * Writes the required vector and doc ID data to the repository
      */
-    private void writeToRepository(RepositoryContext repositoryContext, BuildIndexParams indexInfo) throws IOException,
-        InterruptedException {
+    private void writeToRepository(RepositoryContext repositoryContext, BuildIndexParams indexInfo) {
         VectorRepositoryAccessor vectorRepositoryAccessor = repositoryContext.vectorRepositoryAccessor;
         boolean success = false;
         metrics.startRepositoryWriteMetrics();
@@ -172,8 +172,7 @@ public class RemoteIndexBuildStrategy implements NativeIndexBuildStrategy {
             );
             success = true;
         } catch (InterruptedException | IOException e) {
-            log.debug("Repository write failed for vector field [{}]", indexInfo.getFieldName());
-            throw e;
+            throw new RuntimeException(String.format("Repository write failed for vector field [%s]", indexInfo.getFieldName()), e);
         } finally {
             metrics.endRepositoryWriteMetrics(success);
         }
@@ -183,8 +182,7 @@ public class RemoteIndexBuildStrategy implements NativeIndexBuildStrategy {
      * Submits a remote build request to the remote index build service
      * @return RemoteBuildResponse containing the response from the remote service
      */
-    private RemoteBuildResponse submitBuild(RepositoryContext repositoryContext, BuildIndexParams indexInfo, RemoteIndexClient client)
-        throws IOException {
+    private RemoteBuildResponse submitBuild(RepositoryContext repositoryContext, BuildIndexParams indexInfo, RemoteIndexClient client) {
         final RemoteBuildResponse remoteBuildResponse;
         boolean success = false;
         metrics.startBuildRequestMetrics();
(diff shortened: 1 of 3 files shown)


jed326 (approved):

navneet1v:
Can we directly raise pr against 3.0 branch since all CIs are failing

navneet1v (approved):

navneet1v:
> @navneet1v The CI's still seem to be failing after changing the base branch to 3.0 and trying twice

this is the reason I am saying lets directly port the changes to 3.0

navneet1v (approved):

navneet1v:
approving the changes as the CI failures are not related to the changes.

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
```diff
@@ -147,11 +146,10 @@ public void buildAndWriteIndex(BuildIndexParams indexInfo) throws IOException {
...
             readFromRepository(indexInfo, repositoryContext, remoteBuildStatusResponse);
 
-            success = true;
+            metrics.endRemoteIndexBuildMetrics(true);
         } catch (Exception e) {
+            metrics.endRemoteIndexBuildMetrics(false);
             fallbackStrategy.buildAndWriteIndex(indexInfo);
-        } finally {
-            metrics.endRemoteIndexBuildMetrics(success);
         }
```
shatejas, line 153: Ideally executing logic in catch block isn't considered a good practice. There are exceptions to this but this doesn't seem to be one. Instead we need to pull out the logic from catch
```java
try {
 ....//logic 
 success = true;
 return;
} catch(Exception e) {
    log.warn("Falling back to CPU build", e);
} finally {
    metrics.endRemoteIndexBuildMetrics(success);
}

  fallbackStrategy.buildAndWriteIndex(indexInfo);
}


```

Vikasht34: +1 to this ...

owenhalpert: Makes sense, thanks for the help.

shatejas (dismissed):

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
```diff
@@ -146,13 +146,14 @@ public void buildAndWriteIndex(BuildIndexParams indexInfo) throws IOException {
 
             // 4. Download index file and write to indexOutput
             readFromRepository(indexInfo, repositoryContext, remoteBuildStatusResponse);
-
             success = true;
+            return;
         } catch (Exception e) {
-            fallbackStrategy.buildAndWriteIndex(indexInfo);
+            log.warn("Failed to build index remotely", e);
```
navneet1v, line 152: exception message here is incomplete. It doesn't provide any information about for what parameters index build failed. Lets log all of them to understand what failed.

Another thing is why this is warm log? This should be a error log.

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
```diff
@@ -250,7 +251,7 @@ private void readFromRepository(
             );
             success = true;
         } catch (Exception e) {
-            log.debug("Repository read failed for vector field [{}]", indexInfo.getFieldName());
+            log.debug("Repository read failed for vector field [{}]", indexInfo.getFieldName(), e);
```
navneet1v, line 254: same as above.

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
```diff
@@ -200,7 +201,7 @@ private RemoteBuildResponse submitBuild(RepositoryContext repositoryContext, Bui
             success = true;
             return remoteBuildResponse;
         } catch (IOException e) {
-            log.debug("Submit vector build failed for vector field [{}]", indexInfo.getFieldName());
+            log.debug("Submit vector build failed for vector field [{}]", indexInfo.getFieldName(), e);
             throw e;
```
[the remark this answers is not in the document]
owenhalpert: This makes sense, with the current code you can see the exception when it is rethrown to the caller method and have these logs at the debug level, but it's probably best to just have it all logged once. I'll change this.

owenhalpert:
Thanks @navneet1v for the comments, I changed the exception flow a bit. From the user's perspective:

If a remote index build fails, by default, they will see a single log at the error level: "Failed to build index remotely" with the full stack trace of the exception. This exception will contain anything from an S3 IOException to an incorrect parameter in the build request. If they change their logs to the **debug** level, they will see the StopWatch results from each phase, and upon failure they will see exactly how long the remote build took before failing (in addition to the aforementioned stack trace).

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
navneet1v: It is always better to have those parameters. I would suggest logging them. So lets add it.

navneet1v (dismissed):

navneet1v:
@owenhalpert please check this: https://github.com/opensearch-project/k-NN/actions/runs/14868668558/job/41751731764?pr=2693

owenhalpert:
Seems unrelated, rerunning

navneet1v (approved):

shatejas (approved):

PR #2702, opened 2025-05-09
[BUGFIX] Block derived source enable if index.knn is false

jmazanec15:
### Description
This is a minor experience bug fix. Basically, for derived source, we need the knn codec to be used, so we need the index.knn setting to be set to true.

Adds setting dependency on derived source setting so that if index.knn is set to false, then index creation will be blocked.

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/KNNSettings.java
+++ b/src/main/java/org/opensearch/knn/index/KNNSettings.java
@@ -41,11 +41,13 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.Set;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 import static java.util.stream.Collectors.toUnmodifiableMap;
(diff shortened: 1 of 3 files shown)


src/main/java/org/opensearch/knn/index/KNNSettings.java
```diff
@@ -270,13 +272,31 @@ public class KNNSettings {
...
-        false,
+        (s) -> Boolean.toString(false),
+        (b) -> Booleans.parseBooleanStrict(b, false),
         IndexScope,
         Final,
         UnmodifiableOnRestore
-    );
+    ) {
+        @Override
+        public Set<SettingDependency> getSettingsDependencies(String key) {
```
shatejas, line 284: Quick question: we have getadditionalSettings which makes sure that the setting is applied only for knn index. Why do we need a check again?

shatejas: never mind, I think thats for the default case, it won't take care of misconfigured case

shatejas (approved):

Vikasht34 (approved):

PR #2704, opened 2025-05-13
Apply mask operation in preindex to optimize derived source

jmazanec15:
### Description
Optimization to reduce CPU cycles/JVM memory usage by applying the mask operation in the preindex listener and then resetting the source field mapper's source bytes. 

Running a 10M single node benchmark, I found that map deserialization was consuming a lot of cpu and memory in both the preindex listener and the codec writer. In the preindex listener, we can take advantage of the already deserialized map and apply the mask so that the codec writer deserializes a much smaller map. The experiment shows a significant drop in resource utilization after this optimization is made.

Id like to avoid deserializing completely in the codec writer. However, Im not confident that writeField will always be called after preindex. Thus, I am not removing it in this PR to be on the safe side. Open to opinions on this though

### Baseline Experiments

**CPU - [PreIndex Listener](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/codec/derivedsource/DerivedSourceIndexOperationListener.java#L60-L64)**
image

**CPU - [Codec Writer](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/codec/KNN10010Codec/KNN10010DerivedSourceStoredFieldsWriter.java#L111-L115)**
image

**Memory - [PreIndex Listener](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/codec/derivedsource/DerivedSourceIndexOperationListener.java#L60-L64)**
image

**Memory - [Codec Writer](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/codec/KNN10010Codec/KNN10010DerivedSourceStoredFieldsWriter.java#L111-L115)**
image

### Test Experiments

**CPU - PreIndex Listener**
image

**CPU - Codec Writer**
image

**Memory - PreIndex Listener**
image

**Memory - Codec Writer**
image

### Related Issues

### Check List


src/main/java/org/opensearch/knn/index/codec/derivedsource/DerivedSourceIndexOperationListener.java
```diff
@@ -52,29 +55,45 @@ public Engine.Index preIndex(ShardId shardId, Engine.Index operation) {
...
-            XContentBuilder builder = MediaTypeRegistry.contentBuilder(originalSource.v1(), bStream).map(derivedSource);
+            XContentBuilder builder = MediaTypeRegistry.contentBuilder(originalSource.v1(), bStream).map(cleanVectorSource);
             builder.close();
             operation.parsedDoc().setSource(bStream.bytes(), XContentType.valueOf(originalSource.v1().subtype().toUpperCase(Locale.ROOT)));
         } catch (IOException e) {
             throw new RuntimeException(e);
         }
+
+        IndexableField field = operation.parsedDoc().rootDoc().getField(SourceFieldMapper.CONTENT_TYPE);
+        if (field == null || field.storedValue() == null) {
```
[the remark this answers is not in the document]
jmazanec15: Right, I think this check might be redundant, but adding it to be on safer side.

src/main/java/org/opensearch/knn/index/codec/derivedsource/DerivedSourceIndexOperationListener.java
```diff
@@ -52,29 +55,45 @@ public Engine.Index preIndex(ShardId shardId, Engine.Index operation) {
...
         try (BytesStreamOutput bStream = new BytesStreamOutput();) {
-            XContentBuilder builder = MediaTypeRegistry.contentBuilder(originalSource.v1(), bStream).map(derivedSource);
+            XContentBuilder builder = MediaTypeRegistry.contentBuilder(originalSource.v1(), bStream).map(cleanVectorSource);
             builder.close();
             operation.parsedDoc().setSource(bStream.bytes(), XContentType.valueOf(originalSource.v1().subtype().toUpperCase(Locale.ROOT)));
         } catch (IOException e) {
             throw new RuntimeException(e);
         }
+
+        IndexableField field = operation.parsedDoc().rootDoc().getField(SourceFieldMapper.CONTENT_TYPE);
```
shatejas, line 79: I think there should be a comment here explaining why the mask is applied here instead of codec. Its not obvious from the code flow if someone doesn't have the context.

src/main/java/org/opensearch/knn/index/codec/KNN10010Codec/KNN10010DerivedSourceStoredFieldsWriter.java
```diff
@@ -35,7 +35,7 @@ public class KNN10010DerivedSourceStoredFieldsWriter extends StoredFieldsWriter
     private final Function<Map<String, Object>, Map<String, Object>> vectorMask;
 
     // Keeping the mask as small as possible.
-    private final static Byte MASK = 0x1;
+    public final static Byte MASK = 0x1;
```
shatejas, line 38: We also need a comment in codec writer as to why we are trying to deserialize and apply the mask again, I know this will be a lighter operation in most cases so also mention that in the comment

shatejas (commented):
Looks good, just needs comments

Vikasht34 (approved):

src/main/java/org/opensearch/knn/index/codec/KNN10010Codec/KNN10010DerivedSourceStoredFieldsWriter.java
jmazanec15: Right will add

src/main/java/org/opensearch/knn/index/codec/derivedsource/DerivedSourceIndexOperationListener.java
jmazanec15: yeah good point

src/main/java/org/opensearch/knn/index/codec/KNN10010Codec/KNN10010DerivedSourceStoredFieldsWriter.java
jmazanec15: updated

src/main/java/org/opensearch/knn/index/codec/derivedsource/DerivedSourceIndexOperationListener.java
jmazanec15: updated

Vikasht34 (approved):

shatejas (approved):

0ctopus13prime (approved):

Issue #2714, opened 2025-05-20
[FEATURE][RFC] Asymmetric Distance Calculation and Random Rotation to Boost Recall for On-Disk Vector Search.


finnroblin:
**Is your feature request related to a problem?**
Disk-based vector search was added in OpenSearch 2.17. This feature uses binary quantization to decrease the memory footprint of vector search while keeping recall high. While disk-based vector search is successful at keeping recall high for some datasets, other data sets have recall less than 0.9 out of the box. We can use the two techniques discussed in this RFC, Asymmetric Distance Computation (ADC) and Random  Rotation (RR), to increase the recall on some of these datasets.
 
In order to maintain high recall, disk-based search uses two phases. The first phase searches a binary quantized index for an oversampled number of neighbors. The second phase rescores and reduces these oversampled neighbors using exact search.

![Image](https://github.com/user-attachments/assets/824cf0f7-a507-4146-9daa-f8999791a259)

Currently, the first phase quantizes each 4-byte floating point entry in a vector into a single bit, then computes its distance to neighbors using hamming distance.

While this approach achieves high recall on some datasets, it does not perform well on others. 

We can see that for some datasets, the recall for on-disk (32x compression) is above 0.9 out of the box:<br>
Data set | Recall@k
-- | --
minillm-msmarco | 0.90
mpnet-msmarco | 0.92
mxbai-msmarco | 0.95
snowflake-msmarco | 0.92

However, the success of this approach is dataset dependent. There are data sets that are well below 0.9 out of the box:
| Data set | Recall@k |
|-----------|-----------|
| sift | 0.39 |
| glove | 0.42 |
| gist | 0.12 |
| cohere-v2-wiki | 0.57 |
| cohere-v3-bioasq | 0.71 |
| tasb-msmarco | 0.87 |
| e5small-msmarco | 0.80 |
| clip-flickr-image-queries | 0.81 |
| cohere-v2-dbpedia | 0.76 |

In general, recall can be improved by increasing number of documents we rescore (via increasing `oversample_factor`). However, because rescoring has a random access pattern to disk, more rescoring leads to slower latencies. So recall is improved, but the recall/search latency tradeoff is not.

**What solution would you like?**

Reduce the impact of quantization errors through combining two separate optimizations:
1. Asymmetric Distance Computations (ADC)
2. Random Rotation (RR)

[...]

PR #2718, opened 2025-05-23
[ADC+RR] Random rotation feature

finnroblin:
### Description
Part of #2714 to add random rotation to k-NN for binary quantized indices. 

Random rotation can improve recall on some datasets by smoothing uneven variance across dimensions. See the [issue](https://github.com/opensearch-project/k-NN/issues/2714) for more details and benchmarks. 

This PR implements random rotation with an orthonormal rotation matrix. Each entry is sampled from a standard Gaussian variable and then the matrix is made orthonormal through Gram-Schmidt. The rotation process could be sped up with the panama API but that is out of scope for this PR. 

Random rotation can be enabled through an encoder-level index mapping:
```
PUT vector-index
{
  "settings" : {
    "index": {
      "knn": true
    }
  },
  "mappings": {
    "properties": {
      "vector_field": {
        "type": "knn_vector",
        "dimension": 8,
        "method": {
            "name": "hnsw",
            "engine": "faiss",
            "space_type": "l2",
            "parameters": {
              "encoder": {
                "name": "binary",
                "parameters": {
                  "bits": 1,
                  // NEW MAPPING PARAMETER
		  "enable_random_rotation": true
                }
              }
            }
        }
      }
    }
  }
}
```

Since the focus of the overlying issue is on binary quantization, random rotation is implemented only for binary indices. The rotation matrix is handled by the `ScalarQuantization` part of the quantization framework. 

- As described above, `QFrameBitEncoder` takes the new random rotation setting.
- `TrainingRequest` handles `isEnableRandomRotation`.
- `OneBitScalarQuantizationState`  and `MultiBitScalarQuantizationState` store the matrix if it exists. 
- Matrix generation is handled within the `QuantizationHelper.calculateQuantizationState` method. Once a matrix is generated and written to quantization state it does not change. 
- Rotation itself is handled by `XBitScalarQuantizer.quantize()`. If rotation is enabled then vectors are rotated before they are quantized. The thresholds used for quantization are (as a side effect) also rotated.
- The `aboveThresholdMeans`/`belowThresholdMeans` in scalar quantization states will be used in a follow-up PR to transform the vector for ADC (see the issue for more information).

### Related Issues
https://github.com/opensearch-project/k-NN/issues/2714

### Check List
- [ ] New functionality has been documented. [Will track documentation through RFC #2714].
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md). [No API changes]
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose). [Will track documentation through RFC #2714].

## Coauthors
Vikash Tiwari (@Vikasht34), Tal Wagner (@talwagner)


src/main/java/org/opensearch/knn/common/KNNConstants.java
```diff
@@ -84,6 +84,9 @@ public class KNNConstants {
     public static final String RADIAL_SEARCH_KEY = "radial_search";
     public static final String MODEL_VERSION = "model_version";
     public static final String QUANTIZATION_STATE_FILE_SUFFIX = "osknnqstate";
+    public static final int QUANTIZATION_RANDOM_ROTATION_DEFAULT_SEED = 1212121212; // used to seed the RNG for reproducability in unit
```
jmazanec15, line 87: Just to confirm, it will always be same rotation matrix for a given dimension size?

src/main/java/org/opensearch/knn/index/engine/faiss/QFrameBitEncoder.java
```diff
@@ -41,7 +43,8 @@ public class QFrameBitEncoder implements Encoder {
      *   "encoder": {
      *     "name": "binary",
      *     "parameters": {
-     *       "bits": 2
+     *       "bits": 2,
+     *       "enable_random_rotation": true,
```
jmazanec15, line 47: I think just call it random_rotation. enable_random_rotation seems too long

src/main/java/org/opensearch/knn/index/engine/qframe/QuantizationConfigParser.java
```diff
@@ -34,7 +35,17 @@ public static String toCsv(QuantizationConfig quantizationConfig) {
             return "";
         }
 
-        return TYPE_NAME + SEPARATOR + BINARY_TYPE + "," + BIT_COUNT_NAME + SEPARATOR + quantizationConfig.getQuantizationType().getId();
+        return TYPE_NAME
```
jmazanec15, line 38: Are we handling backwards compatibility here?

src/main/java/org/opensearch/knn/quantization/models/quantizationParams/ScalarQuantizationParams.java
```diff
@@ -25,6 +26,11 @@
 @EqualsAndHashCode
 public class ScalarQuantizationParams implements QuantizationParams {
     private ScalarQuantizationType sqType;
+    private boolean isEnableRandomRotation = QFrameBitEncoder.DEFAULT_ENABLE_RANDOM_ROTATION;
```
jmazanec15, line 29: nit: final? Also, do we use this somewhere?

src/main/java/org/opensearch/knn/quantization/models/quantizationState/MultiBitScalarQuantizationState.java
```diff
@@ -73,12 +112,31 @@ public void writeTo(StreamOutput out) throws IOException {
      * @throws IOException if an I/O error occurs during deserialization.
      */
     public MultiBitScalarQuantizationState(StreamInput in) throws IOException {
-        int version = in.readVInt(); // Read the version
+        int version = in.readVInt();
```
jmazanec15, line 115: Can we add back comments?

src/main/java/org/opensearch/knn/quantization/models/quantizationState/MultiBitScalarQuantizationState.java
```diff
@@ -73,12 +112,31 @@ public void writeTo(StreamOutput out) throws IOException {
...
+
+        int rows = in.readVInt();
         this.thresholds = new float[rows][];
         for (int i = 0; i < rows; i++) {
-            this.thresholds[i] = in.readFloatArray(); // Read each row as a float array
+            this.thresholds[i] = in.readFloatArray();
+        }
+
+        if (Version.fromId(version).onOrAfter(Version.V_3_1_0)) {
+            FloatArrayWrapper[] wrappedBelowMeans = in.readOptionalArray(FloatArrayWrapper::new, FloatArrayWrapper[]::new);
```
jmazanec15, line 125: Are we using below or above means for random rotation? If not, could we remove and add them in adc change?

src/main/java/org/opensearch/knn/quantization/models/quantizationState/MultiBitScalarQuantizationState.java
```diff
@@ -175,6 +233,39 @@ public long ramBytesUsed() {
...
+
+        public FloatArrayWrapper(float[] array) {
+            this.array = array;
+        }
+
+        public FloatArrayWrapper(StreamInput in) throws IOException {
+            this.array = in.readFloatArray();
+        }
+
+        public float[] getArray() {
```
jmazanec15, line 262: nit: lombok getter

src/main/java/org/opensearch/knn/quantization/models/quantizationState/MultiBitScalarQuantizationState.java
```diff
@@ -175,6 +233,39 @@ public long ramBytesUsed() {
...
+                size += RamUsageEstimator.sizeOf(row);
+            }
+        }
         return size;
     }
+
+    private static class FloatArrayWrapper implements Writeable {
+        private final float[] array;
+
+        public FloatArrayWrapper(float[] array) {
```
jmazanec15, line 254: nit: all arg ctor

src/main/java/org/opensearch/knn/quantization/models/quantizationState/OneBitScalarQuantizationState.java
```diff
@@ -139,6 +211,41 @@ public long ramBytesUsed() {
...
+            size += RamUsageEstimator.shallowSizeOf(rotationMatrix);
+            // Add size of each row array
+            for (float[] row : rotationMatrix) {
+                size += RamUsageEstimator.sizeOf(row);
+            }
+        }
         return size;
     }
+
+    private class FloatArrayWrapper implements Writeable {
```
jmazanec15, line 230: Can we put in a util since its used in multiple places?

src/main/java/org/opensearch/knn/quantization/models/requests/TrainingRequest.java
```diff
@@ -30,4 +42,6 @@ public abstract class TrainingRequest<T> {
      * @return the vector corresponding to the specified document ID.
      */
     public abstract T getVectorAtThePosition(int position) throws IOException;
+
+    public abstract void resetVectorValues();
```
jmazanec15, line 46: javadoc

src/main/java/org/opensearch/knn/quantization/quantizer/QuantizerHelper.java
```diff
@@ -5,104 +5,377 @@
...
- * extraction. This class is designed to be used with various quantizer implementations that require
- * consistent handling of training requests and sampled indices.
+ * Utility class for calculating quantization state information for both
+ * OneBit and MultiBit scalar quantizers. Handles computing thresholds,
+ * below/above mean statistics, L2/L1 ratios, and rotation matrix application.
  */
+@Log4j2
 @UtilityClass
 class QuantizerHelper {
+    // ========================= ONE BIT ========================= //
```
jmazanec15, line 26: nit: remove

src/main/java/org/opensearch/knn/quantization/quantizer/QuantizerHelper.java
```diff
@@ -5,104 +5,377 @@
...
 
 import java.io.IOException;
 
 /**
- * Utility class providing common methods for quantizer operations, such as parameter validation and
- * extraction. This class is designed to be used with various quantizer implementations that require
- * consistent handling of training requests and sampled indices.
+ * Utility class for calculating quantization state information for both
+ * OneBit and MultiBit scalar quantizers. Handles computing thresholds,
+ * below/above mean statistics, L2/L1 ratios, and rotation matrix application.
```
jmazanec15, line 21: do we need l2/l1 ratios?

src/main/java/org/opensearch/knn/quantization/quantizer/QuantizerHelper.java
```diff
@@ -5,104 +5,377 @@
...
-        int totalSamples = sampledIndices.length;
-        float[] mean = null;
-        int lastIndex = 0;
+    static OneBitScalarQuantizationState calculateQuantizationState(
+        TrainingRequest<float[]> trainingRequest,
+        int[] sampledIndices,
+        ScalarQuantizationParams quantizationParams
+    ) throws IOException {
+        validateSampledIndices(sampledIndices);
+        int first_vec_idx = sampledIndices[0];
```
jmazanec15, line 43: are we able to re-use more code between this and multi?

src/main/java/org/opensearch/knn/quantization/quantizer/RandomGaussianRotation.java
```diff
@@ -0,0 +1,91 @@
...
+
+import java.util.Random;
+
+import static org.opensearch.knn.common.KNNConstants.QUANTIZATION_RANDOM_ROTATION_DEFAULT_SEED;
+
+@UtilityClass
+public class RandomGaussianRotation {
+
+    /**
+     * Generates a random rotation matrix using Gaussian distribution and orthogonalization.
```
jmazanec15, line 18: Maybe add comment about preservation of dot products and l2 distances when this is applied. Also add comment about seeding

src/main/java/org/opensearch/knn/quantization/quantizer/RandomGaussianRotation.java
```diff
@@ -0,0 +1,91 @@
...
+    /**
+     * Applies a rotation to a vector using the provided rotation matrix.
+     *
+     * @param vector The input vector to be rotated.
+     * @param rotationMatrix The rotation matrix.
+     * @return The rotated vector.
+     */
+    public float[] applyRotation(float[] vector, float[][] rotationMatrix) {
+        int dimensions = vector.length;
+        float[] rotatedVector = new float[dimensions];
```
jmazanec15, line 80: can this be in place instead of allocating a new vector?

src/main/java/org/opensearch/knn/training/FloatTrainingDataConsumer.java
```diff
@@ -98,6 +98,12 @@ private List<byte[]> quantizeVectors(List<?> vectors) throws IOException {
             public float[] getVectorAtThePosition(int position) {
                 return ArrayUtils.toPrimitive((Float[]) vectors.get(position));
             }
+
+            @Override
+            public void resetVectorValues() {
+                // No-op
```
jmazanec15, line 104: Why no-op?

src/main/java/org/opensearch/knn/common/KNNConstants.java
finnroblin: That's the current implementation but I'm open to changing it. I don't think there's anything special about this particular seed -- Yonatan's experiments used this particular number so I moved it into the code.

The fixed seed gives reproducible results since the matrix is the same across multiple ingestions of the same data. I'm worried about a case where several otherwise equivalent benchmark runs produce different recall numbers due to different random matrices.

But as you pointed out it won't be truly random since the rotation matrices will be the same when we fix dimension size.

I think this is okay. Each gaussian in the matrix is still independent when we fix the seed. By central limit theorem we'll still get the property of random rotations that variance is smoothed across dimensions.

src/main/java/org/opensearch/knn/quantization/models/quantizationParams/ScalarQuantizationParams.java
finnroblin: Good catch, I moved the `isEnableRandomRotation` to training request and didn't delete. I don't think scalarquantizationparams needs to store this and we can keep solely in training request and in the quantization state. Will remove.

src/main/java/org/opensearch/knn/training/FloatTrainingDataConsumer.java
finnroblin: Since `getVectorAtThePosition` doesn't change the state of the vectors. `resetVectorValues()` is in the interface since for `KNNVectorQuantizationTrainingRequest` we need `resetVectorValues()` to reset the `knnVectorValuesSupplier` iterator state.

I'll add a clarifying comment here.

src/main/java/org/opensearch/knn/quantization/quantizer/RandomGaussianRotation.java
finnroblin: Do you mean make the method `void` and modify the vector in place? We need to make a vector copy either way or else the matrix-vector multiplication will break. At that point might as well return the modified vector and have one fewer side effect/no need to worry about java "modifying a reference" semantics.

src/main/java/org/opensearch/knn/quantization/models/quantizationParams/ScalarQuantizationParams.java
finnroblin: Took another look and `isEnableRandomRotation` is necessary for `QuantizationService::train` when that method constructs the training request.

If I make this field final I'll have to drop the `@AllArgsConstructor` and force the `@NoArgsConstructor`. What do you think?

finnroblin:
> > I'm seeing a drop in recall for product quantization with these changes (testRecall_whenFaissHNSWPQFP32_thenRecallAbove50percent). Investigating whether erroneous trigger of random rotation is the cause.
> 
> It Should be for PQ use case right , why PQ Use call BQ Random rotation? Strange

Yeah, I'm digging into it right now. None of my rotation breakpoints are being triggered for PQ. 

> > I'm seeing a drop in recall for product quantization with these changes (testRecall_whenFaissHNSWPQFP32_thenRecallAbove50percent). Investigating whether erroneous trigger of random rotation is the cause.
> 
> It Should be for PQ use case right , why PQ Use call BQ Random rotation? Strange

Yes, my hunch is there's some side effect to the new `random_rotation` boolean in the FloatTrainingDataConsumer (even if it's always false for PQ and the vectors are never rotated). Currently debugging

finnroblin:
@Vikasht34 That one recall tests was failing since I changed the dimension in the RecallIT to 128 from 32 while testing and forgot to change back. Fixed now and verified that random rotation param is always off for HNSW PQ + IVF PQ methods.

finnroblin:
The integration test `org.opensearch.knn.index.SegmentReplicationIT.testSearchOnReplicas_whenIndexHasDeletedDocs_thenSuccess` is failing. It doesn't relate to this PR and it passes when I modify `thread.sleep` duration from `5000` to `1500`. I suspect the test is flaky.

The error is:
```
SegmentReplicationIT > testSearchOnReplicas_whenIndexHasDeletedDocs_thenSuccess FAILED
    java.lang.AssertionError: expected:<5> but was:<4>
        at __randomizedtesting.SeedInfo.seed([6C7B80E82ACDA81D:D01CF5F37EA2C41B]:0)
        at org.junit.Assert.fail(Assert.java:89)
        at org.junit.Assert.failNotEquals(Assert.java:835)
        at org.junit.Assert.assertEquals(Assert.java:647)
        at org.junit.Assert.assertEquals(Assert.java:633)
        at org.opensearch.knn.index.SegmentReplicationIT.testSearchOnReplicas_whenIndexHasDeletedDocs_thenSuccess(SegmentReplicationIT.java:70)
```

I am assuming the test is flaky since modifying the `Thread.sleep` duration changes the behavior of the test. Locally I was able to get the test to pass by setting `Thread.sleep(1000)` instead of `Thread.sleep(5000)`. 

If I increase the sleep to 20 seconds then I get a different assertion error (3 docs in result as opposed to 4):
```
SegmentReplicationIT > testSearchOnReplicas_whenIndexHasDeletedDocs_thenSuccess FAILED
    java.lang.AssertionError: expected:<5> but was:<3>
        at __randomizedtesting.SeedInfo.seed([2C9EF4B6DE51F802:90F981AD8A3E9404]:0)
        at org.junit.Assert.fail(Assert.java:89)
        at org.junit.Assert.failNotEquals(Assert.java:835)
        at org.junit.Assert.assertEquals(Assert.java:647)
        at org.junit.Assert.assertEquals(Assert.java:633)
        at org.opensearch.knn.index.SegmentReplicationIT.testSearchOnReplicas_whenIndexHasDeletedDocs_thenSuccess(SegmentReplicationIT.java:70)
```

src/main/java/org/opensearch/knn/index/engine/faiss/QFrameBitEncoder.java
```diff
@@ -41,7 +43,8 @@ public class QFrameBitEncoder implements Encoder {
      *   "encoder": {
      *     "name": "binary",
      *     "parameters": {
-     *       "bits": 2
+     *       "bits": 2,
+     *       "random_rotation": true,
```
jmazanec15, line 49: make sure to take an AI to add this to the documentation.

src/main/java/org/opensearch/knn/index/engine/qframe/QuantizationConfigParser.java
```diff
@@ -33,8 +35,19 @@ public static String toCsv(QuantizationConfig quantizationConfig) {
...
+        String result = TYPE_NAME
+            + SEPARATOR
+            + BINARY_TYPE
+            + ","
+            + BIT_COUNT_NAME
+            + SEPARATOR
+            + quantizationConfig.getQuantizationType().getId();
 
-        return TYPE_NAME + SEPARATOR + BINARY_TYPE + "," + BIT_COUNT_NAME + SEPARATOR + quantizationConfig.getQuantizationType().getId();
+        if (Version.CURRENT.onOrAfter(Version.V_3_1_0)) {
```
finnroblin, line 46: Should use Lucene version for consistency with `fromCsv` method. Will change in next revision.

src/main/java/org/opensearch/knn/quantization/quantizer/RandomGaussianRotation.java
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    /**
+     * Applies a rotation to a vector using the provided rotation matrix.
+     *
+     * @param vector The input vector to be rotated. The input vector is not modified.
+     * @param rotationMatrix The rotation matrix.
+     * @return The copy of the original vector but rotated.
+     */
+    public float[] applyRotation(float[] vector, float[][] rotationMatrix) {
```
jmazanec15, line 86: Are we able to reuse anything from https://github.com/apache/lucene/blob/9afcfdb0fa3bb2f5c09243f9734adf42e0a71694/lucene/core/src/java24/org/apache/lucene/internal/vectorization/PanamaVectorUtilSupport.java#L93 in order to optimize this code? 

Also, is it possible to batch? i.e.
```
public float[][] applyRotationBatch(float[][] vectors, float[][] rotationMatrix) {
```

src/main/java/org/opensearch/knn/index/engine/qframe/QuantizationConfigParser.java
Vikasht34: this seems also fine

src/main/java/org/opensearch/knn/quantization/quantizer/QuantizerHelper.java
```diff
@@ -5,104 +5,187 @@
...
- * Utility class providing common methods for quantizer operations, such as parameter validation and
- * extraction. This class is designed to be used with various quantizer implementations that require
- * consistent handling of training requests and sampled indices.
+ * Utility class for calculating quantization state information for both
+ * OneBit and MultiBit scalar quantizers. Handles computing thresholds,
+ * below/above mean statistics, and rotation matrix application.
  */
 @UtilityClass
 class QuantizerHelper {
+    private static final int ONE_BIT_BITS_PER_COORDINATE = 1;
```
finnroblin: variable name

src/main/java/org/opensearch/knn/quantization/quantizer/RandomGaussianRotation.java
finnroblin: I think batching will be challenging. We'd have to take a larger initiative to batch calls to `quantize`, which is probably a good idea but out of scope for this PR. I added calls to Lucene VectorUtils where applicable to take advantage of its SIMD capabilities.

Vikasht34 (dismissed):

jmazanec15 (dismissed):

jmazanec15 (approved):

Issue #2619, opened 2025-03-19
[BUG] NativeMemoryCacheManager Force Evict Race Condition


owenhalpert:
#### Problem Introduction

When running OpenSearch Benchmark’s Vectorsearch workload on a local constrained cluster, the following exception is constantly thrown in the search phase:

`IndexAllocation-Reference is already closed can't increment refCount current count [0]`

This exception comes from a call to `allocation.incRef` in KNNWeight, after the allocation is returned from `NativeMemoryCacheManager.get()` . This `incRef` call is intending to increment the ref counter and ensure the allocation isn’t closed by another thread.

Code pointer: https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/query/KNNWeight.java#L330 

The same exception is thrown even if the `incRef` call is moved into the NativeMemory get() method, as shown in [this PR,](https://github.com/opensearch-project/k-NN/pull/2435/files) which was originally intending to fix a similar race condition.

#### Potential Cause

Adding some debug statements reveals this sequence:

* Thread A (in the force evict flow) starts to evict entry `foo` from the cache, and closes its allocation
* Thread B calls get() on `foo` and its allocation
* Thread B gets a cache hit
* Thread B errors when trying to increment the refCount

This suggests the force evict logic is not properly synchronizing with the cache state. Even after an allocation is closed, subsequent `get()` calls are returning the closed allocation as a cache hit instead of a cache miss. 

#### Further Testing

The error was consistently showing up on a single node local Docker cluster using the sample OpenSearch `docker-compose.yml` with off-heap memory constrained to 3GB. The benchmarks used the `sift-128-euclidean.hdf5`  dataset with 1M vectors and **6 search clients.** The error was most consistently appearing when running the `no-train-test-only` workflow and then separately triggering runs of `search-only` .

**Varying Search Clients**
With 3.5GB of total container memory, I was able to replicate the error with as few as **4** search clients.
**Varying Off-Heap Memory**
With 6 search clients, the starting point of **3GB** of off-heap memory seemed to be the maximum to still throw the exception. 

In conclusion, a constrained system of about 3GB of off-heap memory and below, and 4 search clients and above, should be able to replicate the bug. Note that once memory is decreased below 3GB, node drops happen often, showing `Cannot connect to host localhost:9200`.

#### Possible Solutions

[...]

PR #2728, opened 2025-06-03
[BUGFIX] Fix RefCount and ClearCache in some race conditions

luyuncheng:
Fix RefCount and ClearCache in some race conditions

### Description
in #2619 @owenhalpert describe a bugs: `IndexAllocation-Reference is already closed can't increment refCount current count [0]`,  i met the same case at scenarios:  at the time with hight throughput search, doing clear cache.

also it would also cause memory leak because it would waiting the writeLock.
image

i analysis the race condition as following shows:

1. Time1: Search Thread, do `KNNWeight#doANNSearch` and get indexAllocation at https://github.com/opensearch-project/k-NN/blob/52dc64b6311a71e174a6ba7a6c6e904f2e2d378f/src/main/java/org/opensearch/knn/index/query/KNNWeight.java#L486-L501
2. Time2: Clear Thread, clear cache triggered, and google Cache do async clear cache, get into https://github.com/opensearch-project/k-NN/blob/52dc64b6311a71e174a6ba7a6c6e904f2e2d378f/src/main/java/org/opensearch/knn/index/memory/NativeMemoryCacheManager.java#L526-L529
3. Time3: Clear Thread, checked refCount to zero, and ready to do closeInternal in NativeMemory Allocation.
4. Time4: Search Thread, got readLock at https://github.com/opensearch-project/k-NN/blob/52dc64b6311a71e174a6ba7a6c6e904f2e2d378f/src/main/java/org/opensearch/knn/index/query/KNNWeight.java#L513-L514, and trying to do incRef but throw exception: `IndexAllocation-Reference is already closed can't increment refCount current count [0]`
5. Time5: Search Thread, throw exception without unlock readLock
6. Time6: Clear Thread, waiting for `writeLock()` which is a `dead lock`. like my pictures shows.

### Related Issues
#2619

### Proposal

i see #2435, but i think it can not fix the bugs. my proposal just catch the throw, and release the lock. it would cause one query throw exception, and also it is reasonable, after next query, it would be reload into memory.

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/query/KNNWeight.java
+++ b/src/main/java/org/opensearch/knn/index/query/KNNWeight.java
@@ -21,7 +21,6 @@ import org.apache.lucene.util.BitSet;
 import org.apache.lucene.util.BitSetIterator;
 import org.apache.lucene.util.Bits;
 import org.apache.lucene.util.FixedBitSet;
-import org.opensearch.OpenSearchException;
 import org.opensearch.common.Nullable;
 import org.opensearch.common.StopWatch;
 import org.opensearch.common.lucene.Lucene;
@@ -516,7 +515,8 @@ public class KNNWeight extends Weight {
             indexAllocation.incRef();
         } catch (IllegalStateException e) {
             indexAllocation.readUnlock();
-            throw new OpenSearchException("failed to create knn search when clear cache");
+            log.error("[KNN] Exception when allocation getting evicted: " + e);
+            throw new RuntimeException("Failed to do kNN search when vector data structures getting evicted ");
         }
         try {
             if (indexAllocation.isClosed()) {
--- a/src/main/java/org/opensearch/knn/index/query/KNNWeight.java
+++ b/src/main/java/org/opensearch/knn/index/query/KNNWeight.java
@@ -44,6 +44,7 @@ import org.opensearch.knn.indices.ModelMetadata;
 import org.opensearch.knn.indices.ModelUtil;
 import org.opensearch.knn.jni.JNIService;
 import org.opensearch.knn.plugin.stats.KNNCounter;
+import org.opensearch.search.SearchException;
 
 import java.io.IOException;
 import java.util.Arrays;
@@ -515,7 +516,7 @@ public class KNNWeight extends Weight {
             indexAllocation.incRef();
         } catch (IllegalStateException e) {
             indexAllocation.readUnlock();
-            throw new SearchException(shardTarget, "failed to create knn search when clear cache");
+            throw new OpenSearchParseException("failed to create knn search when clear cache");
         }
         try {
             if (indexAllocation.isClosed()) {
--- a/src/main/java/org/opensearch/knn/index/query/KNNWeight.java
+++ b/src/main/java/org/opensearch/knn/index/query/KNNWeight.java
@@ -21,7 +21,7 @@ import org.apache.lucene.util.BitSet;
 import org.apache.lucene.util.BitSetIterator;
 import org.apache.lucene.util.Bits;
 import org.apache.lucene.util.FixedBitSet;
-import org.opensearch.OpenSearchParseException;
+import org.opensearch.OpenSearchException;
 import org.opensearch.common.Nullable;
 import org.opensearch.common.StopWatch;
 import org.opensearch.common.lucene.Lucene;
@@ -516,7 +516,7 @@ public class KNNWeight extends Weight {
             indexAllocation.incRef();
         } catch (IllegalStateException e) {
             indexAllocation.readUnlock();
-            throw new OpenSearchParseException("failed to create knn search when clear cache");
+            throw new OpenSearchException("failed to create knn search when clear cache");
         }
         try {
             if (indexAllocation.isClosed()) {
--- a/src/main/java/org/opensearch/knn/index/query/KNNWeight.java
+++ b/src/main/java/org/opensearch/knn/index/query/KNNWeight.java
@@ -511,7 +511,12 @@ public class KNNWeight extends Weight {
         FilterIdsSelector.FilterIdsSelectorType filterType = filterIdsSelector.getFilterType();
         // Now that we have the allocation, we need to readLock it
         indexAllocation.readLock();
-        indexAllocation.incRef();
+        try {
+            indexAllocation.incRef();
+        } catch (IllegalStateException e) {
+            indexAllocation.readUnlock();
+            throw new RuntimeException(e.getMessage() + " this is usually caused by clear cache");
+        }
         try {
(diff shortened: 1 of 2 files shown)


src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -511,7 +511,12 @@ private Map<Integer, Float> doANNSearch(
         FilterIdsSelector.FilterIdsSelectorType filterType = filterIdsSelector.getFilterType();
         // Now that we have the allocation, we need to readLock it
         indexAllocation.readLock();
-        indexAllocation.incRef();
+        try {
+            indexAllocation.incRef();
+        } catch (IllegalStateException e) {
+            indexAllocation.readUnlock();
+            throw new RuntimeException(e.getMessage() + " this is usually caused by clear cache");
```
navneet1v: will this exception be thrown to the user? or we will wrap it in some other exception? I just want to make sure that we are not exposing internal details/logic to end user.

luyuncheng: @navneet1v 
how about with exception: 

```
throw new OpenSearchException( "failed to create knn search when clear cache");
```

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -511,7 +512,12 @@ private Map<Integer, Float> doANNSearch(
         FilterIdsSelector.FilterIdsSelectorType filterType = filterIdsSelector.getFilterType();
         // Now that we have the allocation, we need to readLock it
         indexAllocation.readLock();
-        indexAllocation.incRef();
+        try {
+            indexAllocation.incRef();
+        } catch (IllegalStateException e) {
+            indexAllocation.readUnlock();
+            throw new OpenSearchException("failed to create knn search when clear cache");
```
navneet1v:
```suggestion
            throw new OpenSearchException("failed to do kNN search when vector data structures getting evicted ");
```

@kotwanikunal , @jmazanec15 any thoughts on the exception message?

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -511,7 +512,12 @@ private Map<Integer, Float> doANNSearch(
         FilterIdsSelector.FilterIdsSelectorType filterType = filterIdsSelector.getFilterType();
         // Now that we have the allocation, we need to readLock it
         indexAllocation.readLock();
-        indexAllocation.incRef();
+        try {
+            indexAllocation.incRef();
+        } catch (IllegalStateException e) {
```
navneet1v, line 516: also we should log the exception as an error.

jmazanec15: makes sense to me.

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
kotwanikunal: I think a simple "failed to perform kNN search due to high throughput on the node" would suffice. I feel we should not expose internal details   

```suggestion
throw new OpenSearchException("failed to perform kNN search due to high throughput on the node ");
```

OR
```suggestion
 throw new OpenSearchException("failed to perform kNN search due to resource constraints on the node");
```

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
kotwanikunal: +1. We should get visibility into the actual stack given we are not attaching it as a part of the raised exception.

kotwanikunal:
The changes look good overall apart from the mentioned comments.
The other thing that I noticed when I was working through this code was the use of `lock` and `incRef` - I think we could potentially implemented the RefCounted entry in a way where we can lock internally. 

Having two mechanisms to control the behavior of the same object can lead to this conditions in other scenarios v/s abstracting it out into a `incRef` call would just make it simpler to hold the object reference.

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
luyuncheng: @kotwanikunal  @navneet1v  i prefer to

```
 throw new OpenSearchException("failed to do kNN search when vector data structures getting evicted ");
```

because this happens when do clear cache but not comes from throughput or  resource constraints.

navneet1v: > @kotwanikunal @navneet1v i prefer to
> 
> ```
>  throw new OpenSearchException("failed to do kNN search when vector data structures getting evicted ");
> ```
> 
> because this happens when do clear cache but not comes from throughput or resource constraints.

Lets got with this.

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
luyuncheng: LGTM

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
luyuncheng: LGTM，updated at latest commits

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -511,7 +511,13 @@ private Map<Integer, Float> doANNSearch(
...
+            indexAllocation.readUnlock();
+            log.error("[KNN] Exception when allocation getting evicted: " + e);
```
navneet1v:
```suggestion
            log.error("[KNN] Exception when allocation getting evicted: ", e);
```

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -511,7 +511,13 @@ private Map<Integer, Float> doANNSearch(
         FilterIdsSelector.FilterIdsSelectorType filterType = filterIdsSelector.getFilterType();
         // Now that we have the allocation, we need to readLock it
         indexAllocation.readLock();
-        indexAllocation.incRef();
+        try {
+            indexAllocation.incRef();
+        } catch (IllegalStateException e) {
+            indexAllocation.readUnlock();
+            log.error("[KNN] Exception when allocation getting evicted: ", e);
+            throw new RuntimeException("Failed to do kNN search when vector data structures getting evicted ");
```
navneet1v:
```suggestion
            throw new RuntimeException("Failed to do kNN search when vector data structures getting evicted ", e);
```

kotwanikunal (approved):

navneet1v (approved):

Issue #2726, opened 2025-06-02
[BUG] Flaky Test org.opensearch.knn.index.SegmentReplicationIT.testSearchOnReplicas_whenIndexHasDeletedDocs_thenSuccess


finnroblin:
**What is the bug?**
In k-NN mainline integTestRemoteIndexBuild the test org.opensearch.knn.index.SegmentReplicationIT.testSearchOnReplicas_whenIndexHasDeletedDocs_thenSuccess fails. The failing assertion is `assertEquals(docsInIndex - deleteDocs, knnResults.size());`.

**How can one reproduce the bug?**
1. Setup cluster (g5.2xl) to run remote index build integration tests (see [dev guide](https://github.com/opensearch-project/k-NN/blob/main/DEVELOPER_GUIDE.md#s3-bucket-in-localstack) "S3 Bucket in LocalStack" instructions).
2. Checkout k-NN main branch
3. Run the following test
```
./gradlew :integTestRemoteIndexBuild --tests "org.opensearch.knn.index.SegmentReplicationIT.testSearchOnReplicas_whenIndexHasDeletedDocs_thenSuccess" -Ds3.enabled=true -Dtest.remoteBuild=s3.localStack -Dtest.bucket=bucket -Dtest.base_path=vectors -Daccess_key=${AWS_ACCESS_KEY_ID} -Dsecret_key=${AWS_SECRET_ACCESS_KEY} -Dsession_token=${AWS_SESSION_TOKEN}
```
4. Should see error like the following:
```
SegmentReplicationIT > testSearchOnReplicas_whenIndexHasDeletedDocs_thenSuccess FAILED
    java.lang.AssertionError: expected:<5> but was:<4>
        at __randomizedtesting.SeedInfo.seed([6C7B80E82ACDA81D:D01CF5F37EA2C41B]:0)
        at org.junit.Assert.fail(Assert.java:89)
        at org.junit.Assert.failNotEquals(Assert.java:835)
        at org.junit.Assert.assertEquals(Assert.java:647)
        at org.junit.Assert.assertEquals(Assert.java:633)
        at org.opensearch.knn.index.SegmentReplicationIT.testSearchOnReplicas_whenIndexHasDeletedDocs_thenSuccess(SegmentReplicationIT.java:70)
```
[...]

PR #2734, opened 2025-06-05
Add upper bound setting for remote vector index build + GA settings changes

### LINKED ISSUE #2732: [FEATURE] Add upper limit to remote vector index build size threshold
In addition to the lower bound `size_threshold` we have today, we should also have an upper bound size threshold. This is useful as we are still ultimately constrained on size by the GPU fleet instance memory, so there is an upper bound on segment size for which GPU workers will not have sufficient memory. In these scenarios, kNN should fail fast instead of uploading the segments to the remote object store. We can have this value default to unset, and allow users to configure it based on their GPU fleet implementation.

jed326:
### Description
Depending on the implementation of the remote vector index build service being used, there may be an upper limit on the segment size that the GPU instances being used can handle. For segments about this size, the GPU instances will run OOM, so we should fail fast before uploading the vector to S3. This PR introduces a new cluster setting which specifies the upper bound of the vector size.

This PR also makes an attempt at updating the CHANGELOG the be more in line with core/other plugins. Ref:
- https://github.com/opensearch-project/OpenSearch/blob/main/CHANGELOG.md
- https://github.com/opensearch-project/neural-search/blob/main/CHANGELOG.md

Docs issue for settings name change + GA: https://github.com/opensearch-project/documentation-website/issues/10053

Also, muting the flakey test from https://github.com/opensearch-project/k-NN/issues/2726 in this PR

### Related Issues
Resolves #2732

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/KNNSettings.java
+++ b/src/main/java/org/opensearch/knn/index/KNNSettings.java
@@ -5,16 +5,14 @@
 
 package org.opensearch.knn.index;
 
-import lombok.Setter;
 import lombok.NonNull;
+import lombok.Setter;
 import lombok.extern.log4j.Log4j2;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.opensearch.OpenSearchParseException;
 import org.opensearch.action.admin.cluster.settings.ClusterUpdateSettingsRequest;
 import org.opensearch.action.admin.cluster.settings.ClusterUpdateSettingsResponse;
-import org.opensearch.knn.index.engine.MemoryOptimizedSearchSupportSpec;
-import org.opensearch.transport.client.Client;
 import org.opensearch.cluster.metadata.IndexMetadata;
 import org.opensearch.cluster.service.ClusterService;
 import org.opensearch.common.Booleans;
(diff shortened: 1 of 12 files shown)


navneet1v (dismissed):
Looks good to me

src/main/java/org/opensearch/knn/index/KNNSettings.java
```diff
@@ -102,6 +102,7 @@ public class KNNSettings {
     public static final String KNN_INDEX_REMOTE_VECTOR_BUILD = "index.knn.remote_index_build.enabled";
     public static final String KNN_REMOTE_VECTOR_REPO = "knn.remote_index_build.vector_repo";
     public static final String KNN_INDEX_REMOTE_VECTOR_BUILD_THRESHOLD = "index.knn.remote_index_build.size_threshold";
+    public static final String KNN_REMOTE_VECTOR_BUILD_UPPER_BOUND = "knn.remote_index_build.size_upper_bound";
```
jmazanec15: should we call limit instead of upper bound? Just looking at other settings and we use limit

jed326: Yeah that makes sense to me, will make the change. I was also thinking about changing `index.knn.remote_index_build.size_threshold` to `index.knn.remote_index_build.size_lower_bound` to make it clear these are the opposites of each other, but since one is an index setting and the other is a cluster setting have the different names makes more sense to me.

jmazanec15: Oh I see. Hmmm what about: `knn.remote_index_build.size.max` and `knn.remote_index_build.size.min`. Im not sure I understand it completely, so if you have a strong vision for it, lets go with that

jed326: I think the only thing that I don't like about that is that the min size is an index setting, as we want to allow users to be able to tune the GPU acceleration on a per-index basis. However, the max size is a cluster setting, because it's based on some technical limitation in the user's hardware and would never be different across indices. So to make that distinction I think having the setting names not completely match is a find. Wdyt?

src/testFixtures/java/org/opensearch/knn/KNNRestTestCase.java
```diff
@@ -2448,7 +2448,7 @@ protected boolean isRemoteIndexBuildSupported(final Optional<String> bwcVersion)
             versionString = versionString.substring(0, versionString.length() - 9);
         }
         final Version version = Version.fromString(versionString);
-        return version.onOrAfter(Version.V_3_0_0);
+        return version.onOrAfter(Version.V_3_1_0);
```
jed326, line 2451: Need this as we are introducing breaking changes in this PR related to remote index build (changing settings names)

navneet1v (approved):

jmazanec15 (dismissed):
LGTM

jmazanec15 (approved):

PR #2727, opened 2025-06-02
Refactor Knn Search Results to use TopDocs

VijayanB:
### Description
With this change all search leaf will return TopDocs instead of map and later converted to TopDocs at last step. This will help us align with other search interfaces where results are always returned as TopDocs.

### Check List
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
- [x] Commits are signed per the DCO using `--signoff`.
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/k-NN/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #2735, opened 2025-06-06
Integrate LuceneOnFaiss logic into KNNWeight.

0ctopus13prime:
### Description
Integrate LuceneOnFaiss (e.g., memory-optimized search) into KNNWeight.
Prior to this PR, KNNWeight relied on Lucene queries for vector search operations, including radius and exact search. As a result, several Faiss-specific options were not honored when memory_optimized_search was enabled, even though the underlying Faiss index remained unchanged. For example, setting `index.knn.advanced.approximate_threshold` to -1 is used to skip HNSW graph construction in Faiss. However, since the Lucene query in LuceneOnFaiss was unaware that the index file was missing, it would throw an exception during search, which suppose to give a valid response when the option is off. This and similar options could lead to inconsistent or unexpected behavior when memory_optimized_search was suddenly enabled. (Filtering option, for example) This change removes the dependency on Lucene queries within LuceneOnFaiss for memory-optimized search, allowing a transparent switch to the optimized search algorithm while preserving the intended  and consistent user experience.

### Primary Change
1. Memory optimized search support check logic : Now it supports binary + disk mode compressions (32x, 16x and 8x)
2. Layered KNNWeight : KNNWeight has a common logic and have two subclasses -- DefaultKNNWeight and MemoryOptimizedKNNWeight. KNNWeight applies the common logic for filtering or falling back to exact search etc.
3. QueryFactory, QueryBuilder : LuceneOnFaiss does not use Lucene query no more.
4. Support nested case : Custom KnnCollector, Parent id grouper are added.

### Integration Test
In the previous milestone, I added a testing framework there to test possible combinations with a mapping.
Those combinations include 1. Sparse or Dense 2. Filtering or no filtering 3. Nested or non-nested 4. Vector data type

1. Pure binary case where user ingested quantized vectors to OpenSearch. In this case, no rescoring will be applied.
2. 32x compression
3. 16x compression
4. 8x compression

### Related Issues
N/A

### Check List


src/main/java/org/opensearch/knn/index/query/memoryoptsearch/MemoryOptimizedKNNWeight.java
```diff
@@ -0,0 +1,209 @@
...
+            addExplainIfRequired(results, knnEngine, spaceType);
+
+            return Arrays.stream(results).collect(Collectors.toMap(KNNQueryResult::getId, KNNQueryResult::getScore));
+        }
+
+        log.debug("[KNN] Query yielded 0 results");
+        return Collections.emptyMap();
+    }
+
+    private static class TargetVector {
```
shatejas: This can easily add up for JVM memory usage under load, I would keep it byte of float array and not create a wrapper object. A better option would be to have overloaded `queryIndex`

src/main/java/org/opensearch/knn/index/query/memoryoptsearch/MemoryOptimizedKNNWeight.java
```diff
@@ -0,0 +1,209 @@
...
+        } else {
+            reader.getVectorReader().search(knnQuery.getField(), targetVector.asByteVector(), knnCollector, bitSet);
+        }
+
+        // Make results to return
+        final TopDocs topDocs = knnCollector.topDocs();
+        return makeResults(topDocs, knnEngine, spaceType);
+    }
+
+    private Map<Integer, Float> makeResults(final TopDocs topDocs, KNNEngine knnEngine, SpaceType spaceType) {
```
shatejas, line 192: The conversion might not be needed with https://github.com/opensearch-project/k-NN/pull/2727

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -420,7 +385,7 @@ private Map<Integer, Float> doExactSearch(
         return exactSearch(context, exactSearcherContextBuilder.build());
     }
 
-    private Map<Integer, Float> doANNSearch(
+    private Map<Integer, Float> approximateSearch(
         final SegmentReader reader,
         final LeafReaderContext context,
```
shatejas: nit: Seems safe to move `final SegmentReader reader = Lucene.segmentReader(context.reader());` inside of this method

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -440,7 +405,7 @@ private Map<Integer, Float> doANNSearch(
 
         // Check if a modelId exists. If so, the space type and engine will need to be picked up from the model's
         // metadata.
-        String modelId = fieldInfo.getAttribute(MODEL_ID);
+        final String modelId = fieldInfo.getAttribute(MODEL_ID);
         if (modelId != null) {
```
shatejas, line 425: nit: might be a good idea to clean up the `if else` and move extraction to FieldInfoExtractor where you can

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -474,99 +439,45 @@ private Map<Integer, Float> doANNSearch(
             return Collections.emptyMap();
```
shatejas, line 453: nit: seems unnecessary to quantize the vector before this (maybe move 428-434 after this)

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -381,27 +367,6 @@ protected boolean match(int doc) {
         return BitSet.of(filterIterator, maxDoc);
```
shatejas, line 383: for memory optimized search its safe to consider deleted docs in this, currently its not doing it

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -474,99 +439,45 @@ private Map<Integer, Float> doANNSearch(
...
+        final LeafReaderContext context,
+        final SegmentReader reader,
+        final FieldInfo fieldInfo,
+        final SpaceType spaceType,
+        final KNNEngine knnEngine,
+        final VectorDataType vectorDataType,
+        final byte[] quantizedVector,
+        final String modelId,
+        final BitSet filterIdsBitSet,
+        final int cardinality,
```
shatejas, line 509: Seems like cardinality can be derived if required, will help reduce the parameters

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -474,99 +439,45 @@ private Map<Integer, Float> doANNSearch(
             return Collections.emptyMap();
         }
 
-        final String vectorIndexFileName = engineFiles.get(0);
-        final String cacheKey = NativeMemoryCacheKeyHelper.constructCacheKey(vectorIndexFileName, reader.getSegmentInfo().info);
-
-        final KNNQueryResult[] results;
-        KNNCounter.GRAPH_QUERY_REQUESTS.increment();
+        final Map<Integer, Float> results = doANNSearch(
+            engineFiles.get(0),
```
shatejas: is engine file needed for MemoryOptimizedSearch? It seems easy enough to extract the file name in DefaultKNNWeight, can we push it inside?

src/main/java/org/opensearch/knn/index/query/memoryoptsearch/MemoryOptimizedKNNWeight.java
```diff
@@ -0,0 +1,209 @@
...
+                    targetVector = new TargetVector(target);
+                } else {
+                    if (knnQuery.getVectorDataType() == VectorDataType.BYTE) {
+                        targetVector = new TargetVector(knnQuery.getByteQueryVector());
+                    } else {
+                        targetVector = new TargetVector(knnQuery.getQueryVector());
+                    }
+                }
+
+                return queryIndex(targetVector, cardinality, cardinality + 1, context, filterIdsBitSet, reader, knnEngine, spaceType);
```
shatejas:
```suggestion

                // Quantization case
                if (quantizedTargetVector != null) {
                        if(quantizationService.getVectorDataTypeForTransfer(fieldInfo) == VectorDataType.BINARY)) {
                             return queryIndex(quantizedTargetVector, cardinality, cardinality + 1, context, filterIdsBitSet, reader, knnEngine, spaceType);
                        }
                        // Should never occur, safety if ever any other quantization is added 
                        throw new IllegalStateException();
                }
                
                if (knnQuery.getVectorDataType() == VectorDataType.BINARY || knnQuery.getVectorDataType() == VectorDataType.BYTE) {
                     // when data_type is set byte or binary
                      return queryIndex(knnQuery.getByteVector(), cardinality, cardinality + 1, context, filterIdsBitSet, reader, knnEngine, spaceType);
                }
              
                // fallback to float
                return queryIndex(knnQuery.getFloatVector(), cardinality, cardinality + 1, context, filterIdsBitSet, reader, knnEngine, spaceType); 
```

src/main/java/org/opensearch/knn/index/query/memoryoptsearch/MemoryOptimizedKNNWeight.java
```diff
@@ -0,0 +1,209 @@
...
+            } else {
+                final TargetVector targetVector;
+                if (knnQuery.getVectorDataType() == VectorDataType.BYTE) {
+                    targetVector = new TargetVector(knnQuery.getByteQueryVector());
```
shatejas: Is this a valid case? is radial supported with byte vectors?

src/main/java/org/opensearch/knn/index/query/memoryoptsearch/MemoryOptimizedKNNWeight.java
```diff
@@ -0,0 +1,209 @@
...
+ * where the entire data is loaded into off-heap memory.
+ */
+@Log4j2
+public class MemoryOptimizedKNNWeight extends KNNWeight {
+    private final KnnCollectorManager knnCollectorManager;
+
+    public MemoryOptimizedKNNWeight(KNNQuery query, float boost, final Weight filterWeight, IndexSearcher searcher, int k) {
+        super(query, boost, filterWeight);
+
+        if (k > 0) {
```
shatejas, line 51: Not sure why the collector managers are in constructor, this could be in the core logic considering a new instance of weight is being created with each request

src/main/java/org/opensearch/knn/index/query/memoryoptsearch/MemoryOptimizedKNNWeight.java
```diff
@@ -0,0 +1,209 @@
...
+import static org.opensearch.knn.plugin.stats.KNNCounter.GRAPH_QUERY_ERRORS;
+
+/**
+ * Calculates query weights and builds query scorers.
+ * Internally, it relies on memory-optimized search logic to perform vector search,
+ * which sets it apart from {@link org.opensearch.knn.index.query.DefaultKNNWeight},
+ * where the entire data is loaded into off-heap memory.
+ */
+@Log4j2
+public class MemoryOptimizedKNNWeight extends KNNWeight {
```
shatejas, line 45: Did you consider separating out radial and knn weights? Ideally it should have been on the query level but worth considering to see how it turns out. It will essentially push the logic in KNNQuery to have the right weight but might simplify the if else conditions. let me know your thoughts

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
0ctopus13prime: sorry not following your comment, but mem optimized search is not doing what?

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
0ctopus13prime: will update in the next rev

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
0ctopus13prime: sure, will update in the next rev.

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
0ctopus13prime: sure, will update in the next rev.

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
0ctopus13prime: will update in the next rev

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
0ctopus13prime: it is true, but I'm not seeing a simple way we can derive the value in subclasses.
Let me try my best with this.

src/main/java/org/opensearch/knn/index/query/memoryoptsearch/MemoryOptimizedKNNWeight.java
0ctopus13prime: will update in the next rev

src/main/java/org/opensearch/knn/index/query/memoryoptsearch/MemoryOptimizedKNNWeight.java
0ctopus13prime: yeah, I can move the manager initialization logic within `doANNSearch`.
Will update in the next rev

0ctopus13prime: KNNWeight returns `true` for `isCacheable`, so LRUCache in IndexSearch might cache KNNWeight, better to not to have the manager as a member field.

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
shatejas: So currently if there is no filter, it simply ignores liveDocs (deleted docs) https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/query/KNNWeight.java#L354

For JNI case there are some latency concerns since bitset needs to be converted into an array so we haven't added it. But memoryOptimized search can directly use `BitSet` without converting to array so there are no latency concerns. This will improve recall when there are a lot of deleted docs in the segment. 

Ideally I would want to have it for JNI as well but that will need additional benchmarking so maybe just override the method for memoryoptimized?

src/main/java/org/opensearch/knn/index/query/memoryoptsearch/MemoryOptimizedKNNWeight.java
0ctopus13prime: Yeah, having a separate query would be ideal and I also think it would require a bit of effort as well.
The critical thing needs for consideration is to make sure user would get the same expected result regardless of whether memory_optimized_search is enabled. For example, user must get 100% recall when`index.knn.advanced.filtered_exact_search_threshold` option is set as a pretty big value so that the most number of filtered docs is less than the threshold making it fallback to exact search. 
And I found there are edge cases where only applied to KNN search while not applied to RNN search. (e.g. only applicable when knnQuery.radius() == null) 

I strongly agree with you on this, but hoping to have it as a mid term goal as its scope will be pretty big if we include it in this PR. We first need to factor out the common logic being applied to ANN and RNN at the moment, then define a separate class for memory_opt_search and native_search. So overall, at least 3 classes will be required.

src/main/java/org/opensearch/knn/index/query/memoryoptsearch/MemoryOptimizedKNNWeight.java
0ctopus13prime: no, I just checked that only Float is being supported although it would work for LuceneOnFaiss.
Will reject if a byte query was given for RNN search. I also found that when mem_opt_srch is disabled, if byte query was given for RNN search, NPE will be thrown. Better to add a validation in query builder.

src/main/java/org/opensearch/knn/index/query/memoryoptsearch/MemoryOptimizedKNNWeight.java
0ctopus13prime: yeah, synced with Vijay. And this PR will be blocked until his PR gets merged. 
But I needed this for running Integ testing :)

src/main/java/org/opensearch/knn/index/query/memoryoptsearch/MemoryOptimizedKNNWeight.java
0ctopus13prime: But expected memory required for this class would be roughly 20 bytes (object header + 4 bytes for Object and 1 byte for boolean + possible padding for alignment). I don't think it will add up JVM memory usage. 
Also I'm not seeing JVM GC metric regression in my end at least, leaning forward to keep it unless you have a strong opinion on this.

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
0ctopus13prime: I think we should consider put more weigh on the result consistency regarding this. Although there's a subtle floating calculation difference between C++ and Java side, so user might get slightly different result, but making LOF (LuceneOnFaiss) use liveDocs might give different result altogether. I guess once have mem_opt_search to use liveDocs, it can impact recall hugely for some cases as you mentioned. As HNSW graph does not work very well with the case where there are vast number of docs deleted, making MemoryOptimizedKNNWeight to use liveDocs can lead inconsistent results to user (e.g. turn on mem_opt_search I'm getting low recall suddenly!). FYI, Lucene fallbacks to exact search whenever its internal vector visiting counts exceeds cardinality (e.g. the number of non-deleted docs) due to the nature. 

Also when both filterWeight and liveDocs are given, since Lucene creates a combined bitset, so there will be same cost regarding to this case.

src/main/java/org/opensearch/knn/index/engine/MemoryOptimizedSearchSupportSpec.java
```diff
@@ -61,16 +68,22 @@ public static boolean supported(final Optional<KNNMethodContext> methodContextOp
                 return false;
             }
 
-            // We only support Flat, SQ and binary encoder for HNSW.
+            // Check whether it is a supported quantization.
+            if (quantizationConfig != null && quantizationConfig.getQuantizationType() != null) {
+                if (isSupportedQuantization(quantizationConfig) == false) {
```
jmazanec15, line 73: Why nest this if here?

src/main/java/org/opensearch/knn/index/query/DefaultKNNWeight.java
```diff
@@ -0,0 +1,161 @@
...
+        final byte[] quantizedVector,
+        final String modelId,
+        final BitSet filterIdsBitSet,
+        final int cardinality,
+        final int k
+    ) throws IOException {
+
+        final String cacheKey = NativeMemoryCacheKeyHelper.constructCacheKey(vectorIndexFileName, reader.getSegmentInfo().info);
+
+        KNNCounter.GRAPH_QUERY_REQUESTS.increment();
```
jmazanec15: Are we able to keep this in the abstract class?

src/main/java/org/opensearch/knn/index/query/DefaultKNNWeight.java
```diff
@@ -0,0 +1,161 @@
...
+            }
+        } catch (Exception e) {
+            GRAPH_QUERY_ERRORS.increment();
+            throw new RuntimeException(e);
+        } finally {
+            indexAllocation.readUnlock();
+            indexAllocation.decRef();
+        }
+
+        addExplainIfRequired(results, knnEngine, spaceType);
```
jmazanec15, line 162: Same as above - can we move this to parent class?

src/main/java/org/opensearch/knn/index/query/KNNQuery.java
```diff
@@ -189,10 +191,13 @@ public Weight createWeight(IndexSearcher searcher, ScoreMode scoreMode, float bo
             );
         }
 
-        if (filterWeight != null) {
-            return new KNNWeight(this, boost, filterWeight);
+        if (isMemoryOptimizedSearch) {
+            // Using memory optimized search logic on index.
+            return new MemoryOptimizedKNNWeight(this, boost, filterWeight, searcher, k);
+        } else {
```
jmazanec15: nit: can we remove else block and just return?

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -30,72 +30,51 @@
...
 
 import static org.opensearch.knn.common.KNNConstants.KNN_ENGINE;
 import static org.opensearch.knn.common.KNNConstants.MODEL_ID;
 import static org.opensearch.knn.common.KNNConstants.SPACE_TYPE;
 import static org.opensearch.knn.common.KNNConstants.VECTOR_DATA_TYPE_FIELD;
-import static org.opensearch.knn.index.util.IndexUtil.getParametersAtLoading;
-import static org.opensearch.knn.plugin.stats.KNNCounter.GRAPH_QUERY_ERRORS;
 
 /**
  * Calculate query weights and build query scorers.
```
jmazanec15: Can we update this to discuss strategy on why this is abstract

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -474,99 +439,45 @@ private Map<Integer, Float> doANNSearch(
...
-        if (results.length == 0) {
+        if (results.isEmpty()) {
             log.debug("[KNN] Query yielded 0 results");
             return Collections.emptyMap();
         }
+
+        return results;
+    }
+
+    abstract protected Map<Integer, Float> doANNSearch(
```
jmazanec15, line 499: Can we provide comment on this for extenders?

src/main/java/org/opensearch/knn/index/query/memoryoptsearch/GroupedScoreMinHeap.java
```diff
@@ -0,0 +1,214 @@
...
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * A min-heap based on score, designed with a grouper and bounded capacity. This structure is tailored for scenarios where a document
+ * has multiple child documents, each associated with a vector field. The heap tracks the minimum competitive vector found so far for each
+ * parent document relative to a query vector. Specifically, it maintains the child vector that is closest to the query vector among
+ * all children within the same parent document. The conversion from child IDs to group IDs is handled by the provided grouper.
+ * <p>
```
jmazanec15: Do we need this tag?

src/main/java/org/opensearch/knn/index/query/memoryoptsearch/GroupedScoreMinHeap.java
```diff
@@ -0,0 +1,214 @@
...
+ *
+ * For example, if the heap already contains a group ID `13` with a score of `0.5`, and a new child with a score of `1.2` is added,
+ * the heap updates the score for group ID 13 from 0.5 to 1.2 while maintaining the heap's contract. If the group ID is
+ * encountered for the first time, a new entry `[group ID, 1.2]` is added.
+ */
+public class GroupedScoreMinHeap {
```
jmazanec15: Is there a structure in Lucene we can re-use instead of implementing this from scratch?

src/main/java/org/opensearch/knn/index/query/memoryoptsearch/RadiusVectorSimilarityCollector.java
```diff
@@ -0,0 +1,74 @@
...
+import org.apache.lucene.search.ScoreDoc;
+import org.apache.lucene.search.TopDocs;
+import org.apache.lucene.search.TotalHits;
+import org.apache.lucene.search.knn.KnnSearchStrategy;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Clone of Lucene's VectorSimilarityCollector, which cannot be used directly due to its package-private visibility.
```
jmazanec15, line 18: nit: add link to class

src/main/java/org/opensearch/knn/index/query/memoryoptsearch/MemoryOptimizedSearchTopKnnCollectorManager.java
```diff
@@ -0,0 +1,70 @@
...
+import java.io.IOException;
+
+/**
+ * KnnCollector manager by wrapping {@link AbstractKnnCollector} with {@link MultiLeafKnnCollector}
+ * when multiple segments are present.
+ * <p>
+ * Internally, it maintains a global min-heap to track the lowest competitive similarity found so far.
+ * This can accelerate convergence in each search thread when concurrent search is enabled.
+ */
+public class MemoryOptimizedSearchTopKnnCollectorManager implements KnnCollectorManager {
```
jmazanec15: Why do we need a custom version of this?

src/main/java/org/opensearch/knn/index/query/memoryoptsearch/MemoryOptimizedSearchTopKnnCollectorManager.java
```diff
@@ -0,0 +1,70 @@
...
+import org.apache.lucene.search.knn.MultiLeafKnnCollector;
+import org.apache.lucene.util.hnsw.BlockingFloatHeap;
+import org.opensearch.common.CheckedTriFunction;
+
+import java.io.IOException;
+
+/**
+ * KnnCollector manager by wrapping {@link AbstractKnnCollector} with {@link MultiLeafKnnCollector}
+ * when multiple segments are present.
+ * <p>
```
jmazanec15: nit: remove

src/main/java/org/opensearch/knn/index/query/memoryoptsearch/MemoryOptimizedKNNWeight.java
shatejas: @navneet1v Can you verify this? we have seen additional memory being used when a new object was created per query, one of the main reason currently its a single reference object for quantization which is reused and cleared. 

Lets verify that there is no extra JVM usage if we want to use this?

src/main/java/org/opensearch/knn/index/query/memoryoptsearch/MemoryOptimizedKNNWeight.java
shatejas: > KNNWeight returns true for isCacheable, so LRUCache in IndexSearch might cache KNNWeight, better to not to have the manager as a member field.

The collector manager doesn't seem to be threadsafe, it has a global queue which maintains state and I am not sure if that queue is getting cleared anywhere. Can we verify that it gives right results with caching? I am not sure how caching works so don't know much about it.

We need to verify two cases:
- results of one query are not affected by the other
- If there are same two queries back to back - the results are not affected

src/main/java/org/opensearch/knn/index/query/memoryoptsearch/MemoryOptimizedKNNWeight.java
shatejas: > it as a mid term goal as its scope will be pretty big if we include it in this PR

My suggestion here is not to have a separate `Query` but to have a separate class for `Weight` .

This will make it easier for us to have a separate query in the future and simplifies the current code. Its an additional step that we can take to move closer to separate queries  as separate queries will need separate weight. The Query still remains as it is. Might be worth a try

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
shatejas: Deleted docs is a pain point currently, and consistent results is a fair point. I would get thoughts on this as its not a big a change.

@jmazanec15 @navneet1v @vamshin

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
shatejas: What I meant was the methods can simply do `filterIdsBitSet.cardinality()` let me know if I am missing something, If its for perf reasons I think we can keep it

src/main/java/org/opensearch/knn/index/query/memoryoptsearch/MemoryOptimizedKNNWeight.java
0ctopus13prime: I will move the collector initialization logic under search method, then the weight will be cacheable.
Will update in the next rev

src/main/java/org/opensearch/knn/index/query/memoryoptsearch/MemoryOptimizedKNNWeight.java
0ctopus13prime: sure, let me try to have a separate weight for each. 
So overall, there will be three KNNWeight subclasses
1. Default
2. ANN for mem_opt_srch
3. Radius search for mem_opt_srch

src/main/java/org/opensearch/knn/index/engine/MemoryOptimizedSearchSupportSpec.java
0ctopus13prime: So Naveen is working on supporting Faiss 4x at the moment, and I added the checking logic to explicitly limit 2x, 8x and 16x beforehand.

src/main/java/org/opensearch/knn/index/query/DefaultKNNWeight.java
0ctopus13prime: sure, will update in the next rev

src/main/java/org/opensearch/knn/index/query/DefaultKNNWeight.java
0ctopus13prime: will update in the next rev

src/main/java/org/opensearch/knn/index/query/KNNQuery.java
0ctopus13prime: will update in the next rev

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
0ctopus13prime: will update in the next rev

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
0ctopus13prime: will update in the next rev

src/main/java/org/opensearch/knn/index/query/memoryoptsearch/GroupedScoreMinHeap.java
0ctopus13prime: yeah, my intellij lint plugin added it. Adding <p> tag, it will add a new line in HTML instead showing the sentence in one line I believe.

src/main/java/org/opensearch/knn/index/query/memoryoptsearch/GroupedScoreMinHeap.java
0ctopus13prime: Ah! I found this one - DiversifyingNearestChildrenKnnCollector
Will test it to see whether we're getting the expected result

src/main/java/org/opensearch/knn/index/query/memoryoptsearch/RadiusVectorSimilarityCollector.java
0ctopus13prime: I've tried to add a link there, but since it is a package scoped class, it can't import the class leading to build failure due to invalid link in Javadoc.

src/main/java/org/opensearch/knn/index/query/memoryoptsearch/MemoryOptimizedSearchTopKnnCollectorManager.java
0ctopus13prime: To provide a custom collector for nested case, but if we can reuse what's existing in Lucene for nested case, this can be gone. Let me try my best in the next rev

src/main/java/org/opensearch/knn/index/query/memoryoptsearch/MemoryOptimizedSearchTopKnnCollectorManager.java
0ctopus13prime: I think it is the recommended practice according to this doc - https://www.oracle.com/technical-resources/articles/java/javadoc-tool.html 

> If you have more than one paragraph in the doc comment, separate the paragraphs with a <p> paragraph tag, as shown.

My lint did the work 😛

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
0ctopus13prime: Not a big code change, but it can be big change in the result user might get. 
But if we could align on this, I can add changes in the next rev

jmazanec15: I feel that maybe we provide consistency in the first go around and optimize in future. I think this feature is a pretty massive change so keeping things consistent initially will help work out any issues

src/main/java/org/opensearch/knn/index/query/memoryoptsearch/MemoryOptimizedKNNWeight.java
0ctopus13prime: sure, but so far from the FP16 benchmark (not yet posted in this PR), I don't see a meaningful increase in JVM heap metrics. 
I think getting a faiss index file name would allocate more memory if we don't pass index file name to DefaultKNNWeight haha 😛

src/main/java/org/opensearch/knn/index/query/memoryoptsearch/MemoryOptimizedSearchTopKnnCollectorManager.java
jmazanec15: oh interesting didnt know about that

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
navneet1v: @shatejas 
My take on considering the deleted docs for Faiss C++ with optimal performance is a dream. I would love to see when that feature can see the light of the day.

Now on this specific PR, I feel we should aim for consistency first from logic standpoint. So not passing the live docs currently would be my preference.

@0ctopus13prime one case you should think of is, if we are not passing live docs to LOF, and given that LOF uses the global queue collector and early termination will this cause a significant issue if some one turns on the memory_optimized_flag? This is kind of hard to gauge and predict. But I see theoretically it is high possible, because we get only topK docs. 

FYI our current C++ implementation of not considering the liveDocs also leads to very funny behaviors. Hence I think long term we should just fix the liveDocs for Faiss C++ and at same time add the live docs thing in LOF too.

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
0ctopus13prime: Unfortunately, approximateSearch method that I added does not accept `filterIdsBitSet`, rather it accepts `annFilter` which is created as below:

```
 final BitSet annFilter = (filterWeight != null && cardinality == maxDoc) ? null : filterBitSet;
```

Then we need another logic to derive cardinality from it like below:
```
if (filterIdsBitSet == null) {
    cardinality <- maxDocs
} else {
    cardinality <- filterIdsBitSet.cardinality()
}
```

We got rid of one parameter, but got one additional if-else statement in DefaultKNNWeight. Since MemoryOptimizedSearchKNNWeight does not use cardinality, it seems there's no way to put it in util class...

Do you know a better way than this?

src/main/java/org/opensearch/knn/index/query/memoryoptsearch/MemoryOptimizedKNNWeight.java
navneet1v: @shatejas @0ctopus13prime I would also lean towards not having a new wrapper vector object per query in this case. Looking at your code I see we are are just adding a wrapper over query vector and then also doing a if/else check if vector is float or not.

One approach I can suggest:
You can actually just pass vector as an object in queryIndex function and then use VectorDataType to do the if else check to know what underline JNI function needs to be called. This should be similar to what you are doing right now.

Thoughts on JVM pressure:
Now coming back on the JVM usage, I would suggest not having creating extra classes per query since this may impact high throughput cases in future when this object life extends beyond this small piece of code.

Is there any other strong reason for having this new class or cases which I am not able to see. Would love to understand that.

0ctopus13prime: @shatejas @navneet1v 
Will update in the next rev, thank you for the comments

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
0ctopus13prime: Ah.. could not find a simple way to do this. We need to assign three variables `KNNEngine`, `SpaceType` and `VectorDataType`. Also we need to pass modelId to DefaultKNNWeight. 
Will keep it as it is, but if you have a better way to do it, then will add it in the next revision of next revision.

src/main/java/org/opensearch/knn/index/query/DefaultKNNWeight.java
0ctopus13prime: Ah.. I tried this, but I think it is better to put it as is.
In DefaultKNNWeight, it first gets KNNQueryResult[] containing Faiss's distances. Then it logs raw scores it obtained from Faiss into `knnExplanations`. After this, it transforms the array into Map<Integer, Float> then returns it, which depends on whether it is quantized or not.

LuceneOnFaiss does not need this additional transformation. So if we put `addExplainIfRequired` logic, then we will lost raw scoring explanations as DefaultKNNWeight will return the transformed results.

src/main/java/org/opensearch/knn/index/query/memoryoptsearch/MemoryOptimizedSearchTopKnnCollectorManager.java
0ctopus13prime: By reusing `DiversifyingNearestChildrenKnnCollectorManager`, we can get rid of this.

jmazanec15 (approved):

shatejas (approved):

PR #1829, opened 2024-07-15
[feature to main branch] Add binary format support with HNSW method in Faiss Engine

heemin32:
### Description
https://github.com/opensearch-project/k-NN/pull/1815
https://github.com/opensearch-project/k-NN/pull/1781
 
### Issues Resolved
N/A
 
### Check List
- [X] New functionality includes testing.
  - [X] All tests pass
- [X] New functionality has been documented.
  - [X] New functionality has javadoc added
- [X] Commits are signed as per the DCO using --signoff 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/k-NN/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #2736, opened 2025-06-06
Add engine as a top level mapping parameter

### LINKED ISSUE #2534: Make engine top level field mapping parameter
## Description

Currently, the engine is included in the method subsection of the knn_vector mapping:
```
{
  "mappings": {
    "properties": {
      "my_vector": {
        "type": "knn_vector",
        "dimension": 4,
        "method": {
          "name": "hnsw",
          "engine": "faiss",
          "parameters": {}
        }
      }
    }
  }
}
```

The reasoning for this was to include ann index structure details in the method definition.

However, Im wondering if we should make engine a top-level mapping parameter, as we did with the space_type a couple releases ago:
```
{
  "mappings": {
    "properties": {
      "my_vector": {
        "type": "knn_vector",
        "dimension": 4,
        "engine": "faiss",
        "method": {
          "name": "hnsw",
          "parameters": {}
        }
      }
    }
  }
}
```

This will simplify the handling of the method and its parameters, so that it just includes the algorithm and its parameters. This will also make it much easier to add an engine extension point as well.

kaushikau:
### Description
The engine parameter is currently located in the method section of the knn_vector mapping. Moving it up to be a top-level mapping field parameter would make the method subsection simpler for the user, as it would now only contain the algorithm and its parameters. The user can also then set the engine without defining the method section when creating a vector field.

This PR introduces engine as a top-level field parameter, similar to how space_type was moved in a previous release. It is an optional parameter, and the user can set engine in either the method or at the top-level. 

### Changes Made

* “engine” added as a top-level mapping parameter
* new KNNEngine type “UNDEFINED” (similar to in SpaceType) that is used when user does not set an engine

### Testing

Both integration testing and unit testing were performed to test the new field. Unit tests were added for EngineResolver and KNNVectorFieldMapper. IT was added for the new top level engine parameter. An example of one of the integration tests is below:

```
{
    "settings": {
        "index": {
            "knn": "true"
        }
    },
    "mappings": {
        "properties": {
            "vector1": {
                "type": "knn_vector",
                "dimension": 4,
                "space_type": "l2",
                "engine": "lucene"
            }
        }
    }
}
```
This resulted in the index successfully being created with the “lucene” engine set.

### Related Issues
Resolves #2534

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/common/KNNConstants.java
+++ b/src/main/java/org/opensearch/knn/common/KNNConstants.java
@@ -39,6 +39,7 @@ public class KNNConstants {
     public static final String METHOD_PARAMETER_SPACE_TYPE = "space_type"; // used for mapping parameter
     // used for defining toplevel parameter
     public static final String TOP_LEVEL_PARAMETER_SPACE_TYPE = METHOD_PARAMETER_SPACE_TYPE;
+    public static final String TOP_LEVEL_PARAMETER_ENGINE = KNN_ENGINE;
     public static final String COMPOUND_EXTENSION = "c";
     public static final String MODEL = "model";
     public static final String MODELS = "models";
@@ -80,6 +81,7 @@ public class KNNConstants {
     public static final VectorDataType DEFAULT_VECTOR_DATA_TYPE_FIELD = VectorDataType.FLOAT;
     public static final String MINIMAL_MODE_AND_COMPRESSION_FEATURE = "mode_and_compression_feature";
     public static final String TOP_LEVEL_SPACE_TYPE_FEATURE = "top_level_space_type_feature";
+    public static final String TOP_LEVEL_ENGINE_FEATURE = "top_level_engine_feature";
 
     public static final String RADIAL_SEARCH_KEY = "radial_search";
     public static final String MODEL_VERSION = "model_version";
--- a/src/main/java/org/opensearch/knn/common/KNNConstants.java
+++ b/src/main/java/org/opensearch/knn/common/KNNConstants.java
@@ -81,7 +81,6 @@ public class KNNConstants {
     public static final VectorDataType DEFAULT_VECTOR_DATA_TYPE_FIELD = VectorDataType.FLOAT;
     public static final String MINIMAL_MODE_AND_COMPRESSION_FEATURE = "mode_and_compression_feature";
     public static final String TOP_LEVEL_SPACE_TYPE_FEATURE = "top_level_space_type_feature";
-    public static final String TOP_LEVEL_ENGINE_FEATURE = "top_level_engine_feature";
 
     public static final String RADIAL_SEARCH_KEY = "radial_search";
     public static final String MODEL_VERSION = "model_version";
--- a/src/main/java/org/opensearch/knn/common/KNNConstants.java
+++ b/src/main/java/org/opensearch/knn/common/KNNConstants.java
@@ -81,6 +81,7 @@ public class KNNConstants {
     public static final VectorDataType DEFAULT_VECTOR_DATA_TYPE_FIELD = VectorDataType.FLOAT;
     public static final String MINIMAL_MODE_AND_COMPRESSION_FEATURE = "mode_and_compression_feature";
     public static final String TOP_LEVEL_SPACE_TYPE_FEATURE = "top_level_space_type_feature";
+    public static final String TOP_LEVEL_ENGINE_FEATURE = "top_level_engine_feature";
 
     public static final String RADIAL_SEARCH_KEY = "radial_search";
     public static final String MODEL_VERSION = "model_version";
--- a/src/main/java/org/opensearch/knn/index/engine/EngineResolver.java
+++ b/src/main/java/org/opensearch/knn/index/engine/EngineResolver.java
@@ -8,9 +8,14 @@ package org.opensearch.knn.index.engine;
 import com.google.common.annotations.VisibleForTesting;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.apache.logging.log4j.util.Strings;
 import org.opensearch.Version;
+import org.opensearch.index.mapper.MapperParsingException;
+import org.opensearch.knn.index.SpaceType;
 import org.opensearch.knn.index.mapper.CompressionLevel;
 import org.opensearch.knn.index.mapper.Mode;
+
+import java.util.Locale;
 
 import static org.opensearch.knn.index.engine.KNNEngine.DEPRECATED_ENGINES;
 
@@ -45,6 +50,25 @@ public final class EngineResolver {
         Version version
     ) {
         return logAndReturnEngine(resolveKNNEngine(knnMethodConfigContext, knnMethodContext, requiresTraining, version));
+    }
+
+    /**
+     * Resolves engine from configuration details. It is guaranteed not to return null.
+     * When engine is not in either method and top level, DEFAULT will be returned.
+     *
+     * @param knnMethodConfigContext configuration context
+     * @param knnMethodContext KNNMethodContext
+     * @param toplevelEngineString Alternative top-level engine
+     * @return {@link SpaceType} for the method
+     */
+    public KNNEngine resolveEngine(
+            KNNMethodConfigContext knnMethodConfigContext,
+            KNNMethodContext knnMethodContext,
+            String toplevelEngineString,
+            boolean requiresTraining,
+            Version version
+    ) {
+        return logAndReturnEngine(resolveKNNEngine(knnMethodConfigContext, knnMethodContext, toplevelEngineString, requiresTraining, version));
     }
 
     /**
@@ -92,6 +116,76 @@ public final class EngineResolver {
         return KNNEngine.FAISS;
     }
 
+    /**
+     * Based on the provided {@link Mode} and {@link CompressionLevel}, resolve to a {@link KNNEngine}.
+     *
+     * @param knnMethodConfigContext configuration context
+     * @param knnMethodContext KNNMethodContext
+     * @param topLevelEngineString Alternative top-level engine
+     * @param requiresTraining whether config requires training
+     * @param version opensearch index version
+     * @return {@link KNNEngine}
+     */
+    private KNNEngine resolveKNNEngine(
+            KNNMethodConfigContext knnMethodConfigContext,
+            KNNMethodContext knnMethodContext,
+            String topLevelEngineString,
+            boolean requiresTraining,
+            Version version
+    ) {
+        KNNEngine methodEngine = getEngineTypeFromMethodContext(knnMethodContext);
+        KNNEngine topLevelEngine = getEngineFromString(topLevelEngineString);
+
+        // user configured method engine
+        if (isEngineConfigured(topLevelEngine) == false && isEngineConfigured(methodEngine) != false) {
+            return knnMethodContext.getKnnEngine();
+        }
+
+        // user configured top level engine
+        if (isEngineConfigured(methodEngine) == false && isEngineConfigured(topLevelEngine) != false) {
+            return KNNEngine.getEngine(topLevelEngineString);
+        }
+
+        if (topLevelEngine != KNNEngine.UNDEFINED && topLevelEngine == methodEngine && methodEngine != KNNEngine.UNDEFINED) {
+            // both engines are same
+            return topLevelEngine;
+        } else if (topLevelEngine != methodEngine) {
+            // engines are different
+            throw new MapperParsingException(
+                    String.format(
+                            Locale.ROOT,
+                            "Cannot specify conflicting engines: \"[%s]\" \"[%s]\"",
+                            methodEngine.getName(),
+                            topLevelEngine.getName()
+                    )
+            );
+        }
+
+        // Handle training case
+        if (requiresTraining) {
+            // Faiss is the only engine that supports training, so we default to faiss here for now
+            return KNNEngine.FAISS;
+        }
+
+        Mode mode = knnMethodConfigContext.getMode();
+        CompressionLevel compressionLevel = knnMethodConfigContext.getCompressionLevel();
+
+        // If both mode and compression are not specified, we can just default
+        if (Mode.isConfigured(mode) == false && CompressionLevel.isConfigured(compressionLevel) == false) {
+            return KNNEngine.DEFAULT;
+        }
+
+        if (compressionLevel == CompressionLevel.x4) {
+            // Lucene is only engine that supports 4x - so we have to default to it here.
+            return KNNEngine.LUCENE;
+        }
+        if (CompressionLevel.isConfigured(compressionLevel) == false || compressionLevel == CompressionLevel.x1) {
+            // For 1x or no compression, we need to default to faiss if mode is provided and use nmslib otherwise based on version check
+            return resolveEngineForX1OrNoCompression(mode, version);
+        }
+        return KNNEngine.FAISS;
+    }
+
     private boolean hasUserConfiguredEngine(KNNMethodContext knnMethodContext) {
         return knnMethodContext != null && knnMethodContext.isEngineConfigured();
     }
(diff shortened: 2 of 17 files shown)


src/main/java/org/opensearch/knn/common/KNNConstants.java
```diff
@@ -80,6 +81,7 @@ public class KNNConstants {
     public static final VectorDataType DEFAULT_VECTOR_DATA_TYPE_FIELD = VectorDataType.FLOAT;
     public static final String MINIMAL_MODE_AND_COMPRESSION_FEATURE = "mode_and_compression_feature";
     public static final String TOP_LEVEL_SPACE_TYPE_FEATURE = "top_level_space_type_feature";
+    public static final String TOP_LEVEL_ENGINE_FEATURE = "top_level_engine_feature";
```
rchitale7, line 84: is this parameter used?

src/main/java/org/opensearch/knn/index/engine/EngineResolver.java
```diff
@@ -47,6 +52,25 @@ public KNNEngine resolveEngine(
...
+    /**
+     * Resolves engine from configuration details. It is guaranteed not to return null.
+     * When engine is not in either method and top level, DEFAULT will be returned.
+     *
+     * @param knnMethodConfigContext configuration context
+     * @param knnMethodContext KNNMethodContext
+     * @param toplevelEngineString Alternative top-level engine
+     * @return {@link SpaceType} for the method
+     */
+    public KNNEngine resolveEngine(
```
rchitale7, line 53: can we remove the function definition of `resolveEngine` on line 46? Seems like we're replacing it with this one

src/main/java/org/opensearch/knn/index/engine/EngineResolver.java
```diff
@@ -92,6 +116,76 @@ private KNNEngine resolveKNNEngine(
...
+     * Based on the provided {@link Mode} and {@link CompressionLevel}, resolve to a {@link KNNEngine}.
+     *
+     * @param knnMethodConfigContext configuration context
+     * @param knnMethodContext KNNMethodContext
+     * @param topLevelEngineString Alternative top-level engine
+     * @param requiresTraining whether config requires training
+     * @param version opensearch index version
+     * @return {@link KNNEngine}
+     */
+    private KNNEngine resolveKNNEngine(
```
rchitale7, line 75: can we remove the function definition of resolveKNNEngine on line 83? Seems like we're replacing it with this one

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java
```diff
@@ -193,6 +192,15 @@ public static class Builder extends ParametrizedFieldMapper.Builder {
             SpaceType.UNDEFINED.getValue()
         ).setValidator(SpaceType::getSpace);
 
+        // A top level engine field.
+        protected final Parameter<String> topLevelEngine = Parameter.stringParam(
```
shatejas, line 197: We probably don't want a new parameter, we might just want to reuse the existing one and parse it on top level

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java
```diff
@@ -51,9 +51,8 @@
 import org.opensearch.knn.index.engine.SpaceTypeResolver;
 import org.opensearch.knn.index.util.IndexUtil;
 import org.opensearch.knn.indices.ModelDao;
-import static org.opensearch.knn.common.KNNConstants.DEFAULT_VECTOR_DATA_TYPE_FIELD;
-import static org.opensearch.knn.common.KNNConstants.KNN_METHOD;
-import static org.opensearch.knn.common.KNNConstants.VECTOR_DATA_TYPE_FIELD;
+
+import static org.opensearch.knn.common.KNNConstants.*;
```
jmazanec15: avoid wildcard import

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java
jmazanec15: I think we would want a new parameter.

src/test/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapperTests.java
```diff
@@ -71,24 +71,7 @@
...
-import static org.opensearch.knn.common.KNNConstants.METHOD_PARAMETER_EF_CONSTRUCTION;
-import static org.opensearch.knn.common.KNNConstants.METHOD_PARAMETER_M;
-import static org.opensearch.knn.common.KNNConstants.METHOD_PARAMETER_SPACE_TYPE;
-import static org.opensearch.knn.common.KNNConstants.MODEL_ID;
-import static org.opensearch.knn.common.KNNConstants.MODE_PARAMETER;
-import static org.opensearch.knn.common.KNNConstants.NAME;
-import static org.opensearch.knn.common.KNNConstants.NMSLIB_NAME;
-import static org.opensearch.knn.common.KNNConstants.PARAMETERS;
-import static org.opensearch.knn.common.KNNConstants.VECTOR_DATA_TYPE_FIELD;
+import static org.opensearch.knn.common.KNNConstants.*;
```
jmazanec15: avoid wildcard import

jmazanec15:
Overall I think this looks pretty good - thanks @kaushikau . A few things to do before checking in:
1. Rebase
2. Add some integ tests
3. Add some backwards compatibility tests

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java
shatejas: @jmazanec15 why is there a need? it adds to the confusion for the user IMO. I am assuming here that there is no difference between an engine in the method vs on the top level

src/main/java/org/opensearch/knn/index/engine/KNNEngine.java
```diff
@@ -31,7 +31,8 @@ public enum KNNEngine implements KNNLibrary {
     @Deprecated(since = "2.19.0", forRemoval = true)
     NMSLIB(NMSLIB_NAME, Nmslib.INSTANCE, Version.V_3_0_0),
     FAISS(FAISS_NAME, Faiss.INSTANCE),
-    LUCENE(LUCENE_NAME, Lucene.INSTANCE);
+    LUCENE(LUCENE_NAME, Lucene.INSTANCE),
+    UNDEFINED("undefined");
```
shatejas, line 35: I am not in favor of adding an UNDEFINED engine when its not needed. We should be easily be able to avoid this with null checks right? Whats forcing us to define a new enum?

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java
jmazanec15: Im not sure how else we would support it while making it be at the top level.

shatejas: I think `TOP_LEVEL_PARAMETER_ENGINE` constant threw me off, I thought we are creating a "top_level_engine" in the mapping. But I traced it to having `engine` so we are good

src/main/java/org/opensearch/knn/index/engine/KNNEngine.java
kaushikau: There is a non-null condition in KNNMethodContext, so for the case when the user sets a top-level engine but not a method engine, using null would not work. I was also following the structure of how space type was added as a top-level parameter, which uses an UNDEFINED enum for when top-level or method space type is not set by the user.

src/main/java/org/opensearch/knn/common/KNNConstants.java
kaushikau: yes, used for minimal version check

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java
```diff
@@ -452,6 +462,21 @@ private void validateSpaceType(KNNVectorFieldMapper.Builder builder) {
...
+                KNNEngine knnMethodContextEngine = knnMethodContext.getKnnEngine();
+                KNNEngine topLevelEngine = KNNEngine.getEngine(builder.originalParameters.getTopLevelEngine());
+                if (topLevelEngine != KNNEngine.UNDEFINED
+                    && knnMethodContextEngine != KNNEngine.UNDEFINED
+                    && topLevelEngine != knnMethodContextEngine) {
+                    throw new MapperParsingException(
+                        "Engine in method and top level engine should be same or one of them should be defined."
+                    );
+                }
+            }
```
shatejas: I am a bit confused on why this is needed? There is a similar validation happening in resolveKNNEngine

src/main/java/org/opensearch/knn/index/engine/EngineResolver.java
```diff
@@ -25,46 +30,81 @@ public final class EngineResolver {
...
-        // Check user configuration first
-        if (hasUserConfiguredEngine(knnMethodContext)) {
-            return knnMethodContext.getKnnEngine();
+        KNNEngine methodEngine = getEngineTypeFromMethodContext(knnMethodContext);
+        KNNEngine topLevelEngine = getEngineFromString(topLevelEngineString);
+
+        // user configured method engine
+        if (isEngineConfigured(topLevelEngine) == false && hasUserConfiguredEngine(knnMethodContext)) {
+            return methodEngine;
+        }
```
shatejas: This is a bit misleading - if topLevelEngine is set, that also means its a user configured engine. I think we should drop the misleading utility methods

I think we need to simplify this logic a bit
```
resolveAndValidateUserConfiguredEngine() {
     if (knnMethodContext.isUserConfiguredEngine() && topLevelEngineString != null) {
            topLevelEngine = validateTopLevel() //throw MapperParsingexception if its invalid
            if (methodEngine == topLevelEngine) {
                  return topLevelEngine
            }  
            throw new MapperParsingException(conflicting engines); 
     }

    if (knnMethodContext.isUserConfiguredEngine()) {
           return  methodEngine;
    }

    if (topLevelEngine != null) {
         return validateTopLevel() 
   }
   return default/ null;
}

resolveEngine() {
     Engine resolvedEngine = resolveAndValidateUserConfiguredEngine()
     // do all other validations
}

```

src/main/java/org/opensearch/knn/index/engine/EngineResolver.java
```diff
@@ -76,11 +116,6 @@ private KNNEngine resolveKNNEngine(
         Mode mode = knnMethodConfigContext.getMode();
         CompressionLevel compressionLevel = knnMethodConfigContext.getCompressionLevel();
 
-        // If both mode and compression are not specified, we can just default
```
shatejas: why was this moved?

src/main/java/org/opensearch/knn/index/engine/EngineResolver.java
```diff
@@ -25,46 +30,81 @@ public final class EngineResolver {
...
+                String.format(
+                    Locale.ROOT,
+                    "Cannot specify conflicting engines: \"[%s]\" \"[%s]\"",
+                    methodEngine.getName(),
+                    topLevelEngine.getName()
+                )
+            );
         }
 
         // Handle training case
```
shatejas, line 92: So till now the assumption was that knnmethod will be null for cases after this. Is it safe to make that assumption for toplevel engine? for instance can user have compression as x4 and topLevelEngine == faiss? in that case is this validation taken care somewhere else?

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java
kaushikau: You're right, I had this validation before I added the same one in resolveKNNEngine; will remove.

src/main/java/org/opensearch/knn/index/engine/EngineResolver.java
kaushikau: Moved back to original place.

src/main/java/org/opensearch/knn/index/engine/EngineResolver.java
kaushikau: Added a method for top level engine to take care of this validation!

src/test/java/org/opensearch/knn/index/engine/EngineResolverTests.java
```diff
@@ -6,11 +6,19 @@
...
+import org.opensearch.index.mapper.MapperParsingException;
 import org.opensearch.knn.KNNTestCase;
 import org.opensearch.knn.index.SpaceType;
 import org.opensearch.knn.index.mapper.CompressionLevel;
 import org.opensearch.knn.index.mapper.Mode;
 
+import java.io.IOException;
+import java.util.Map;
+
+import static org.opensearch.knn.common.KNNConstants.*;
```
jmazanec15: avoid wildcard imports

src/main/java/org/opensearch/knn/index/util/IndexUtil.java
```diff
@@ -66,6 +66,7 @@ public class IndexUtil {
     private static final Version MINIMAL_TOP_LEVEL_SPACE_TYPE_FEATURE = Version.V_2_17_0;
     private static final Version MINIMAL_SUPPORTED_VERSION_FOR_MODEL_VERSION = Version.V_2_17_0;
     private static final Version MINIMAL_EXPAND_NESTED_FEATURE = Version.V_2_19_0;
+    private static final Version MINIMAL_TOP_LEVEL_ENGINE_FEATURE = Version.V_3_0_0;
```
jmazanec15: 3.2

src/test/java/org/opensearch/knn/integ/TopLevelEngineParameterIT.java
```diff
@@ -0,0 +1,146 @@
...
+            .startObject(KNN_METHOD)
+            .field(NAME, METHOD_HNSW)
+            .field(KNN_ENGINE, KNNEngine.FAISS.getName())
+            .endObject()
+            .endObject()
+            .endObject()
+            .endObject();
+
+        String mapping = builder.toString();
+        createKnnIndex(INDEX_NAME, mapping);
```
jmazanec15, line 177: Can we assert the response code to make sure its valid in these tests?

kaushikau: Done!

jmazanec15 (approved):
LGTM - make sure to create documentation update for 3.2with this new parameter and deprecating the old one.

navneet1v:
@jmazanec15 is there a reason we need a UNDEFINED Engine? similar to space type? what was the reason for having the UNDEFINED space type.

jmazanec15:
> @jmazanec15 is there a reason we need a UNDEFINED Engine? similar to space type? what was the reason for having the UNDEFINED space type.

This was done for binary I believe. Related to https://github.com/opensearch-project/k-NN/pull/1829. IIRC correctly, it was in order to properly set default space type based on what data type was being used. L2 for float, hamming for bit.

jmazanec15:
@kaushikau  can you rebase?

src/test/java/org/opensearch/knn/integ/TopLevelEngineParameterIT.java
```diff
@@ -0,0 +1,204 @@
...
+import static org.opensearch.knn.common.KNNConstants.NAME;
+
+public class TopLevelEngineParameterIT extends KNNRestTestCase {
+    private final static float[] TEST_VECTOR = new float[] { 1.0f, 2.0f };
+    private final static int DIMENSION = 2;
+    private final static int K = 1;
+    private static final String INDEX_NAME = "top-level-engine-index";
+
+    @SneakyThrows
+    public void testBaseCase() {
```
shatejas, line 41: Add a test with compression as well

shatejas (commented):
This looks good just a small integ test and we are good to push

shatejas (approved):

vamshin (approved):

PR #2751, opened 2025-06-13
[Bug Fix] Warm-up and support on_disk + 1x compression.

0ctopus13prime:
### Description
This PR resolves three issues listed below:

**1. Warm up API with memory optimized search.**
With field-level support for memory_optimized_search in the on_disk mode with 1x compression, we need to modify the warmup API to selectively load memory-optimized searchers and off-heap indices based on each field's configuration.
We should avoid paying for the high cost of loading the graph when vector search is performed internally using LuceneOnFaiss which will make engine need twice bigger memory usage at the worst case. (e.g. one for off-heap loading and one for LuceneOnFaiss)

**2. Support OnDisk + 1x compression** 
With this change, the on_disk mode with 1x compression will use LuceneOnFaiss by default. This behavior will be limited to the Faiss engine when using HNSW as the indexing method.

**3. Make LuceneOnFaiss lazy loading**
We are using [KNN9120PerFieldKnnVectorsFormat](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/codec/KNN9120Codec/KNN9120PerFieldKnnVectorsFormat.java) as the KNN vector format. During initialization, before search begins, SegmentReader acquires a KnnVectorReader via this format and delegates KNN search operations to it.

I had assumed that [BasePerFieldKnnVectorsFormat::getKnnVectorsFormatForField](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/codec/BasePerFieldKnnVectorsFormat.java#L82C29-L82C56) would be called to provide the required format, but that turned out not to be the case. The method [getKnnVectorsFormatForField](https://github.com/apache/lucene/blob/686ef91e15f6be22aa6d2d55335d0e5160636c6a/lucene/core/src/java/org/apache/lucene/codecs/perfield/PerFieldKnnVectorsFormat.java#L101) is only invoked during the write path and is never called during search.

> Returns the numeric vector format that should be used for writing new segments of <code>field</code>

During reading, the KnnVectorsFormat is created via [KnnVectorsFormat.forName](https://github.com/apache/lucene/blob/686ef91e15f6be22aa6d2d55335d0e5160636c6a/lucene/core/src/java/org/apache/lucene/codecs/perfield/PerFieldKnnVectorsFormat.java#L223). As a result, the MapperService is always null, which causes the initialization of the MemoryOptimizedSearcher table to be skipped.

This table is instead initialized at search time. The search thread first determines whether to use a memory-optimized searcher based on the settings and field type, then calls KnnVectorReader::search. At that point, it tries to retrieve the searcher from the table. If the table is empty, a lazy loading step is triggered. [[Code]](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java#L166)

Therefore, it appears that KnnVectorReader currently has no way to determine which fields have memory-optimized search enabled. Hence, this PR removes the redundant passing of MapperService to the format, which will be always null anyway.

### Related Issues
N/A

### Check List


src/test/java/org/opensearch/knn/memoryoptsearch/MOSFaissFloatIndexIT.java
```diff
@@ -34,7 +34,8 @@ public void testNestedFloatIndexWithIP() {
     }
 
     public void testWhenNoIndexBuilt() {
-        doTestNonNestedIndex(VectorDataType.FLOAT, EMPTY_PARAMS, true, SpaceType.L2, NO_BUILD_HNSW);
-        doTestNonNestedIndex(VectorDataType.FLOAT, EMPTY_PARAMS, false, SpaceType.L2, NO_BUILD_HNSW);
+        // TODO : Fix this
```
0ctopus13prime, line 37: Found a bug in LuceneOnFaiss for radial search.
Will raise a separate PR for this.

src/test/java/org/opensearch/knn/memoryoptsearch/MOSFaissFP16IndexIT.java
```diff
@@ -38,7 +38,8 @@ public void testNestedFloatIndexWithIP() {
     }
 
     public void testWhenNoIndexBuilt() {
-        doTestNonNestedIndex(VectorDataType.FLOAT, METHOD_PARAMETERS, true, SpaceType.L2, NO_BUILD_HNSW);
-        doTestNonNestedIndex(VectorDataType.FLOAT, METHOD_PARAMETERS, false, SpaceType.L2, NO_BUILD_HNSW);
+        // TODO : Fix
```
0ctopus13prime, line 41: Found a bug in LuceneOnFaiss for radial search.
Will raise a separate PR for this.

src/main/java/org/opensearch/knn/index/engine/MemoryOptimizedSearchSupportSpec.java
```diff
@@ -28,6 +33,20 @@
 public class MemoryOptimizedSearchSupportSpec {
     private static final Set<String> SUPPORTED_HNSW_ENCODING = Set.of(ENCODER_FLAT, ENCODER_SQ, ENCODER_BINARY);
 
+    public static boolean supported(final KNNVectorFieldType fieldType, final String indexName) {
```
jmazanec15: javadoc - does this mean its supported or enabled?

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
```diff
@@ -363,4 +370,20 @@ private IOSupplier<VectorSearcher> getVectorSearcherSupplier(final FieldInfo fie
         // Not supported
         return null;
     }
+
+    public static class VectorSearcherHolder {
```
jmazanec15, line 380: Can you add comments?

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
```diff
@@ -446,8 +446,7 @@ protected Query doToQuery(QueryShardContext context) {
 
         VectorQueryType vectorQueryType = getVectorQueryType(k, maxDistance, minScore);
         final String indexName = context.index().getName();
-        final boolean memoryOptimizedSearchSupported = knnVectorFieldType.isMemoryOptimizedSearchSupported()
-            && KNNSettings.isMemoryOptimizedKnnSearchModeEnabled(indexName);
+        final boolean memoryOptimizedSearchSupported = MemoryOptimizedSearchSupportSpec.supported(knnVectorFieldType, indexName);
```
jmazanec15: Should this be called memoryOptimizedSearchEnabled instead of supported?

This value will determine which one to user right?

src/main/java/org/opensearch/knn/index/engine/MemoryOptimizedSearchSupportSpec.java
0ctopus13prime: it meant 'can memory optimized searching be applied to this field type?'
Sure, will update in the next rev

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
0ctopus13prime: will update in the next rev

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
0ctopus13prime: yeah, you're correct.
I think 'enabled' is giving more precise meaning.
will update in the next rev

jmazanec15 (approved):

Vikasht34 (approved):

PR #2758, opened 2025-06-17
[Bug Fix] Fix Cagra HNSW graph populates random entry points at the beginning of search.

0ctopus13prime:
### Description
Cagra HNSW graph first populates randomly selected entry points then do the search.
Which is not included in LuceneOnFaiss. This PR has added fix to populate random entry points at the beginning of search.

### Related Issues
N/A

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissMemoryOptimizedSearcher.java
+++ b/src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissMemoryOptimizedSearcher.java
@@ -8,7 +8,9 @@ package org.opensearch.knn.memoryoptsearch.faiss;
 import org.apache.lucene.codecs.hnsw.FlatVectorsScorer;
 import org.apache.lucene.index.VectorEncoding;
 import org.apache.lucene.index.VectorSimilarityFunction;
+import org.apache.lucene.search.DocIdSetIterator;
 import org.apache.lucene.search.KnnCollector;
+import org.apache.lucene.search.knn.KnnSearchStrategy;
 import org.apache.lucene.store.IndexInput;
 import org.apache.lucene.util.Bits;
 import org.apache.lucene.util.IOSupplier;
@@ -17,8 +19,10 @@ import org.apache.lucene.util.hnsw.OrdinalTranslatedKnnCollector;
 import org.apache.lucene.util.hnsw.RandomVectorScorer;
 import org.opensearch.knn.index.KNNVectorSimilarityFunction;
 import org.opensearch.knn.memoryoptsearch.VectorSearcher;
+import org.opensearch.knn.memoryoptsearch.faiss.cagra.FaissCagraHNSW;
 
 import java.io.IOException;
+import java.util.concurrent.ThreadLocalRandom;
 
 /**
  * This searcher directly reads FAISS index file via the provided {@link IndexInput} then perform vector search on it.
@@ -109,7 +113,7 @@ public class FaissMemoryOptimizedSearcher implements VectorSearcher {
 
         // Set up required components for vector search
         final RandomVectorScorer scorer = scorerSupplier.get();
-        final KnnCollector collector = new OrdinalTranslatedKnnCollector(knnCollector, scorer::ordToDoc);
+        final KnnCollector collector = createKnnCollector(knnCollector, scorer);
         final Bits acceptedOrds = scorer.getAcceptOrds(acceptDocs);
 
         if (knnCollector.k() < scorer.maxOrd()) {
@@ -133,5 +137,75 @@ public class FaissMemoryOptimizedSearcher implements VectorSearcher {
 
     private IndexInput getSlicedIndexInput() throws IOException {
         return indexInput.slice("FaissMemoryOptimizedSearcher", 0, fileSize);
+    }
+
+    private KnnCollector createKnnCollector(final KnnCollector knnCollector, final RandomVectorScorer scorer) {
+        final KnnCollector ordinalTranslatedKnnCollector = new OrdinalTranslatedKnnCollector(knnCollector, scorer::ordToDoc);
+
+        if (hnsw instanceof FaissCagraHNSW cagraHNSW) {
+            return new KnnCollector.Decorator(ordinalTranslatedKnnCollector) {
+                @Override
+                public KnnSearchStrategy getSearchStrategy() {
+                    return new RandomEntryPointsKnnSearchStrategy(
+                        cagraHNSW.getNumBaseLevelSearchEntryPoints(),
+                        cagraHNSW.getTotalNumberOfVectors(),
+                        knnCollector.getSearchStrategy()
+                    );
+                }
+            };
+        } else {
+            return ordinalTranslatedKnnCollector;
+        }
+    }
+
+    /**
+     * Knn search strategy having a doc-id-iterator returning random document ids.
+     * This is not designed for general purpose, it is particularly designed for populating random document ids for Cagra index.
+     * Note that doc-id-iterator returns a random ids in `nextDoc` method without sorting, and might return duplicated ids.
+     */
+    static class RandomEntryPointsKnnSearchStrategy extends KnnSearchStrategy.Seeded {
+        public RandomEntryPointsKnnSearchStrategy(
+            final int numberOfEntryPoints,
+            final long totalNumberOfVectors,
+            final KnnSearchStrategy originalStrategy
+        ) {
+            super(
+                generateRandomEntryPoints(numberOfEntryPoints, Math.toIntExact(totalNumberOfVectors)),
+                numberOfEntryPoints,
+                originalStrategy
+            );
+        }
+
+        private static DocIdSetIterator generateRandomEntryPoints(final int numberOfEntryPoints, int totalNumberOfVectors) {
+            return new DocIdSetIterator() {
+                int numPopulatedVectors = 0;
+
+                @Override
+                public int docID() {
+                    throw new UnsupportedOperationException("DISI in RandomEntryPointsKnnSearchStrategy does not support docID()");
+                }
+
+                @Override
+                public int nextDoc() {
+                    if (numPopulatedVectors < numberOfEntryPoints) {
+                        ++numPopulatedVectors;
+                        // It is fine to populate the same doc ids here, the same vectors will not be visited more than once with bitset.
+                        return ThreadLocalRandom.current().nextInt(totalNumberOfVectors);
+                    }
+
+                    return NO_MORE_DOCS;
+                }
+
+                @Override
+                public int advance(int targetDoc) {
+                    throw new UnsupportedOperationException("DISI in RandomEntryPointsKnnSearchStrategy does not support advance(int)");
+                }
+
+                @Override
+                public long cost() {
+                    throw new UnsupportedOperationException("DISI in RandomEntryPointsKnnSearchStrategy does not support cost()");
+                }
+            };
+        }
     }
 }
--- a/src/main/java/org/opensearch/knn/memoryoptsearch/faiss/cagra/FaissCagraHNSW.java
+++ b/src/main/java/org/opensearch/knn/memoryoptsearch/faiss/cagra/FaissCagraHNSW.java
@@ -5,11 +5,12 @@
 
 package org.opensearch.knn.memoryoptsearch.faiss.cagra;
 
+import lombok.Getter;
+import lombok.Setter;
 import org.apache.lucene.store.IndexInput;
 import org.opensearch.knn.memoryoptsearch.faiss.FaissHNSW;
 
 import java.io.IOException;
-import java.util.concurrent.ThreadLocalRandom;
 
 /**
  * This represents CAGRA HNSW.
@@ -17,6 +18,10 @@ import java.util.concurrent.ThreadLocalRandom;
  * The only difference are entry point and the number of layers, as searches in CAGRA HNSW are always conducted on the bottom layer.
  */
 public class FaissCagraHNSW extends FaissHNSW {
+    @Setter
+    @Getter
+    private int numBaseLevelSearchEntryPoints;
+
     /**
      * Partial load the CAGRA HNSW graph.
      * During a search on CAGRA HNSW, the entry point is randomly selected, and the search runs on the bottom graph.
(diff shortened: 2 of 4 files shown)


src/main/java/org/opensearch/knn/memoryoptsearch/faiss/cagra/RandomEntryPointsKnnSearchStrategy.java
```diff
@@ -0,0 +1,49 @@
...
+import org.apache.lucene.search.DocIdSetIterator;
+import org.apache.lucene.search.knn.KnnSearchStrategy;
+
+import java.util.concurrent.ThreadLocalRandom;
+
+/**
+ * This strategy randomly generates entry points for HNSW search.
+ * For more details, refer to <a href="https://github.com/facebookresearch/faiss/blob/main/faiss/IndexHNSW.cpp#L919">...</a>
+ */
+public class RandomEntryPointsKnnSearchStrategy extends KnnSearchStrategy.Seeded {
```
navneet1v: can we make this class package private?

navneet1v:
> Cagra HNSW graph first populates randomly selected entry points then do the search.

Based on the code which I can see here for Cagra: https://github.com/facebookresearch/faiss/blob/main/faiss/IndexHNSW.cpp#L950-L984

I don't think so they actually send all the random ids as entry points. Only 1 which is closet to query among the randomly generated vectors is sent as the entry point. 

I looked in the PR too, I think we are sending all the randomly generated vectors as entry points, any reason for that? @0ctopus13prime . Please correct me if my understanding of the PR is incorrect.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/cagra/RandomEntryPointsKnnSearchStrategy.java
```diff
@@ -0,0 +1,49 @@
...
+        return new DocIdSetIterator() {
+            @Override
+            public int docID() {
+                throw new UnsupportedOperationException("DISI in RandomEntryPointsKnnSearchStrategy does not support docID()");
+            }
+
+            @Override
+            public int nextDoc() {
+                return ThreadLocalRandom.current().nextInt(totalNumberOfVectors);
+            }
```
navneet1v: Correct me if I am wrong, the entry points order doesn't matter here ?

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/cagra/SearchStrategyOverrideKnnCollector.java
```diff
@@ -0,0 +1,69 @@
...
+import org.apache.lucene.search.TopDocs;
+import org.apache.lucene.search.knn.KnnSearchStrategy;
+
+/**
+ * Overriding {@link KnnSearchStrategy} to the given {@link KnnCollector}.
+ * Except for the search strategy, it delegates the given collector to return the value.
+ */
+@AllArgsConstructor
+public class SearchStrategyOverrideKnnCollector implements KnnCollector {
+    @NonNull
```
navneet1v: why we need this class?

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/cagra/RandomEntryPointsKnnSearchStrategy.java
0ctopus13prime: will update in the next rev

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/cagra/RandomEntryPointsKnnSearchStrategy.java
0ctopus13prime: yes, we send entry points then as `int[]` then it will run [this](https://github.com/facebookresearch/faiss/blob/main/faiss/IndexHNSW.cpp#L919) logic in [here](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/util/hnsw/HnswGraphSearcher.java#L252-L264).

so the order of entry points really does not matter

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/cagra/SearchStrategyOverrideKnnCollector.java
0ctopus13prime: To support Cagra, we have to do 2 things
1. Provide entry points (e.g. `int[]`)
2. Make HNSW searcher to run greedy BFS at the bottom graph.

And this logic is controlled by [KnnSearchStrategy::Seeded](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/knn/KnnSearchStrategy.java#L107). 

Therefore, in FaissMemoryOptimizedSearcher, if HNSW we have is Cagra, then wrapping the given KnnCollector to return [KnnSearchStrategy::Seeded](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/knn/KnnSearchStrategy.java#L107) with `SearchStrategyOverrideKnnCollector` to activate 2 things listed above.

0ctopus13prime:
> Based on the code which I can see here for Cagra: https://github.com/facebookresearch/faiss/blob/main/faiss/IndexHNSW.cpp#L950-L984
> 
> I don't think so they actually send all the random ids as entry points. Only 1 which is closet to query among the randomly generated vectors is sent as the entry point.
> 
> I looked in the PR too, I think we are sending all the randomly generated vectors as entry points, any reason for that? @0ctopus13prime . Please correct me if my understanding of the PR is incorrect.

They are doing the same thing fundamentally, just that Faiss first narrows down the best candidate from the randomly chosen vectors upfront then continue the greedy BFS at the bottom graph in `search_level_0`, while Lucene does two things in one method in [searchLevel](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/util/hnsw/HnswGraphSearcher.java#L252-L264). Therefore, from our side, passing random vector ids should do the same work as Faiss.

One subtle difference is that, as you mentioned, Lucene puts all random vectors into heap expecting little luck, as it might give more optimal vectors to help the algorithm converge quickly. However, from the perspective of running the greedy BFS algorithm, it will always lead convergence regardless of passing only one best candidate versus passing all of them. (Lucene first evaluates the best candidate from randomly picked up vectors in [here](https://github.com/apache/lucene/blame/main/lucene/core/src/java/org/apache/lucene/util/hnsw/HnswGraphSearcher.java#L273))

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/cagra/RandomEntryPointsKnnSearchStrategy.java
0ctopus13prime: Ah.. I tried, but `FaissMemoryOptimizedSearcher` is using `RandomEntryPointsKnnSearchStrategy`.
`FaissMemoryOptimizedSearcher` is in `org.opensearch.knn.memoryoptsearch.faiss` package and `RandomEntryPointsKnnSearchStrategy` is in `org.opensearch.knn.memoryoptsearch.faiss.cagra` package.

I can move `RandomEntryPointsKnnSearchStrategy` to `org.opensearch.knn.memoryoptsearch.faiss`, but the strategy is only being used in Cagra context.... 🫠

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/cagra/SearchStrategyOverrideKnnCollector.java
0ctopus13prime: - [[Code](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/util/hnsw/HnswGraphSearcher.java#L124-L128)] If search strategy is Seeded, then using SeededHnswGraphSearcher
- [[Code](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/util/hnsw/SeededHnswGraphSearcher.java#L76)] Providing random vectors
- [[Code](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/util/hnsw/AbstractHnswGraphSearcher.java#L82)] Start search at the bottom

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/cagra/SearchStrategyOverrideKnnCollector.java
```diff
@@ -0,0 +1,69 @@
...
+import org.apache.lucene.search.KnnCollector;
+import org.apache.lucene.search.TopDocs;
+import org.apache.lucene.search.knn.KnnSearchStrategy;
+
+/**
+ * Overriding {@link KnnSearchStrategy} to the given {@link KnnCollector}.
+ * Except for the search strategy, it delegates the given collector to return the value.
+ */
+@AllArgsConstructor
+public class SearchStrategyOverrideKnnCollector implements KnnCollector {
```
shatejas: Instead of maintaining a public class can you please do this.
```

if (hnsw instanceof FaissCagraHNSW cagraHNSW) {
    return new KnnCollector.Decorator(new OrdinalTranslatedKnnCollector(knnCollector, scorer::ordToDoc)) {
                    @Override
                    public KnnSearchStrategy getSearchStrategy() {
                        return new RandomEntryPointsKnnSearchStrategy(
                                 cagraHNSW.getNumBaseLevelSearchEntryPoints(),
                                 cagraHNSW.getTotalNumberOfVectors(),
                                 knnCollector.getSearchStrategy()
                );
                    }
                };
}
```

 Lets see if there is any other use case before we make this into a public class

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/cagra/RandomEntryPointsKnnSearchStrategy.java
```diff
@@ -0,0 +1,49 @@
...
+    private static DocIdSetIterator generateRandomEntryPoints(final int totalNumberOfVectors) {
+        return new DocIdSetIterator() {
+            @Override
+            public int docID() {
+                throw new UnsupportedOperationException("DISI in RandomEntryPointsKnnSearchStrategy does not support docID()");
+            }
+
+            @Override
+            public int nextDoc() {
+                return ThreadLocalRandom.current().nextInt(totalNumberOfVectors);
```
shatejas: what if this returns the same entry point?

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/cagra/RandomEntryPointsKnnSearchStrategy.java
```diff
@@ -0,0 +1,49 @@
...
+public class RandomEntryPointsKnnSearchStrategy extends KnnSearchStrategy.Seeded {
+    public RandomEntryPointsKnnSearchStrategy(
+        final int numberOfEntryPoints,
+        final long totalNumberOfVectors,
+        final KnnSearchStrategy originalStrategy
+    ) {
+        super(generateRandomEntryPoints(Math.toIntExact(totalNumberOfVectors)), numberOfEntryPoints, originalStrategy);
+    }
+
+    private static DocIdSetIterator generateRandomEntryPoints(final int totalNumberOfVectors) {
```
shatejas: Considering public class, this docIdSetIterator has a risk of an infinite loop since NO_MORE_DOCS is never returned. There are 2 options that I can think of

Limit docIdSetIterator to return only iterate  `numberOfEntryPoints` and then return NO_MORE_DOCS or make it an inner class

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/cagra/RandomEntryPointsKnnSearchStrategy.java
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * This strategy randomly generates entry points for HNSW search.
+ * For more details, refer to <a href="https://github.com/facebookresearch/faiss/blob/main/faiss/IndexHNSW.cpp#L919">...</a>
+ */
+public class RandomEntryPointsKnnSearchStrategy extends KnnSearchStrategy.Seeded {
+    public RandomEntryPointsKnnSearchStrategy(
+        final int numberOfEntryPoints,
+        final long totalNumberOfVectors,
+        final KnnSearchStrategy originalStrategy
```
shatejas: We should force `KNNSearchStrategy.Hnsw` or have an assert

https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/util/hnsw/HnswGraphSearcher.java#L101-L103

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/cagra/SearchStrategyOverrideKnnCollector.java
0ctopus13prime: will update in the next rev

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/cagra/RandomEntryPointsKnnSearchStrategy.java
0ctopus13prime: it is acceptable to have same entry point values. 
even we had a same value, the same vector will not be visited more than once due to the bitset

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/cagra/RandomEntryPointsKnnSearchStrategy.java
0ctopus13prime: will update in the next reva

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/cagra/RandomEntryPointsKnnSearchStrategy.java
shatejas: Okay, its not obvious please add a comment

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/cagra/RandomEntryPointsKnnSearchStrategy.java
0ctopus13prime: It should be fine to have an arbitrary strategy. As the line you mentioned is to determine the inner HNSW graph (which can be optimal for filtering case), and what we are trying to do is to have the top level HNSW graph to pass the populated vectors down to the inner graph. 

Therefore, we dont need to force any type of strategy as what we care is the top level HNSW graph.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/cagra/RandomEntryPointsKnnSearchStrategy.java
0ctopus13prime: sure!

shatejas (approved):

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/cagra/SearchStrategyOverrideKnnCollector.java
navneet1v: if we want to keep the class, then please add CagraPrefix to ensure that this is for Cagra Index.

navneet1v: Is this class getting used? I am not able to find the usage of this class in src file. am I missing it?

shatejas: The file is probably not needed anymore

0ctopus13prime: thank you for the hawk eye. 🐦 
Will remove it.

jmazanec15 (approved):

shatejas (approved):

navneet1v:
My github had some issues and it caused the comments to be published later on. Sorry for that. I can see the PR is already merged

Issue #2368, opened 2025-01-06
[Proposal] Improved experience for exact search 


harishbhakuni:
### Background:

Today, if user wants to perform exact search on a given data set, he has to:

* Either use [score scripting](https://opensearch.org/docs/latest/search-plugins/knn/knn-score-script/). Exact search with score scripting is also helpful if user wants to use a different space type other than the one used for the index.
* Or, set index.knn index setting as false during index creation, which when done will allow users to run all the kNN queries as exact search. (if this setting is set as false, no graph files will be created for the index).

#### Problems with current approach:

1. Unnecessary vector data structure creation:
    1. If user have more than one knn_vector field. vector related data structures will be still created for all the vector fields irrespective of whether user want to perform ANN search on all of them or not. 
    2. This consumes a lot of memory.
    3. related issue: https://github.com/opensearch-project/k-NN/issues/2270
2. Usage of score script for exact search:
    1. For doing exact search on a given field if index.knn setting is enabled, user needs to use score script query.
    2. This involves script compilation overhead during query execution. 
    3. Also usually there are security concerns associated with using scripts during search. Although, there are no known security risks with the knn_score script provided by plugin as it only provides a predefined way of executing exact search. user cannot use custom logic in the script.
    4. related issue: https://github.com/opensearch-project/k-NN/issues/1079

#### Solving problem (1): Unnecessary vector data structure creation
**Approach 1 (Preferred):**

* Use a new parameter in the index mapping called searchMode which can have exact/ann value and can be used to determine which search should be done with a default value of ann.

**Approach 2:**

* Introduce a field level parameter approximate_search_min_vectors which is the number of vectors a segment should have before creating specialized data structures for approximate search. 
* If set to -1, vector data structures will not be created and exact search will be used for the field. However, if set to 0, data structures will always get created.
* This is similar to the currently available index setting index.knn.advanced.approximate_threshold. The new parameter takes precedence over this existing index setting.
* Something like this:
```
PUT /test-index
        {
          "settings": {
            "index": {
              "knn": true
            }
          },
          "mappings": {
            "properties": {
              "vect_field_1": {         // vector data structures will always build.
                "type": "knn_vector",   // ANN will be used for the field.
                "dimension": 3,
                "approximate_search_min_vectors":0,
                ...
              },
              "vect_field_2": {         // vector data structures will never build.
                "type": "knn_vector",   // Exact search will be used for the field.
                "dimension": 3,
[...]

PR #2768, opened 2025-06-20
Add `index` parameter to remove unnecessary vector data structure creation

kaushikau:
### Description
If a user has multiple vector fields in a KNN index currently, and wants to perform exact search on certain fields, graph files will still be created for all vector fields, leading to high memory consumption. The creation of a mapping field-level parameter `index` will allow users to not create graphs on a per-field basis. 

In OpenSearch, `index` is already supported for a number of field types, such as text, numeric, boolean, and keyword. With this implementation, this parameter would now be supported for the k-NN vector field type as well.

Note: if a user explicitly sets `index.knn.advanced.approximate_threshold` to not be equal to 0, that will take precedence over `index`.

Creating test index with index = true and index = false vector fields:
```
{
"settings": {
    "index": {
        "knn": "true"
     }
},
"mappings": {
    "properties": {
        "my_vector1": {
            "type": "knn_vector",
            "dimension": 2,
            "index": "false"
        },
        "my_vector2": {
            "type": "knn_vector",
            "dimension": 2,
            "index": "true",
            "method": {
                "name": "hnsw",
                "engine": "lucene"
            }
        }
    }
}
}'
```

Result of performing exact search on the first vector field:
```
{
  "took" : 681,
  "timed_out" : false,
  "_shards" : {
    "total" : 1,
    "successful" : 1,
    "skipped" : 0,
    "failed" : 0
  },
  "hits" : {
    "total" : {
      "value" : 10,
      "relation" : "eq"
    },
    "max_score" : 0.21929827,
    "hits" : [
      {
        "_index" : "test-index",
        "_id" : "2",
        "_score" : 0.21929827,
        "_source" : {
          "my_vector1" : [
            2.5,
            3.5
          ],
          "my_vector2" : [
            8.1,
            2.9
          ]
        }
      },
      {
        "_index" : "test-index",
        "_id" : "3",
        "_score" : 0.18656717,
        "_source" : {
          "my_vector1" : [
            3.5,
            4.5
          ],
          "my_vector2" : [
            1.9,
            3.0
          ]
        }
      }
    ]
  }
```

Result of ANN search on the second vector field:
```
{
  "took" : 19,
  "timed_out" : false,
  "_shards" : {
    "total" : 1,
    "successful" : 1,
    "skipped" : 0,
    "failed" : 0
  },
  "hits" : {
    "total" : {
      "value" : 2,
      "relation" : "eq"
    },
    "max_score" : 0.5952381,
    "hits" : [
      {
        "_index" : "test-index",
        "_id" : "3",
        "_score" : 0.5952381,
        "_source" : {
          "my_vector1" : [
            3.5,
            4.5
          ],
          "my_vector2" : [
            1.9,
            3.0
          ]
        }
      },
      {
        "_index" : "test-index",
        "_id" : "9",
        "_score" : 0.45248863,
        "_source" : {
          "my_vector1" : [
            4.6,
            8.3
          ],
          "my_vector2" : [
            1.1,
            2.1
          ]
        }
      }
    ]
  }
```

### Related Issues
Resolves Problem 1 of #2368 

### Check List


src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNN990FlatVectorsFormat.java
```diff
@@ -0,0 +1,66 @@
...
+import org.apache.lucene.index.SegmentWriteState;
+
+import java.io.IOException;
+
+/**
+ * This is a Vector format that will be used for flat vectors that do not create graph data structures.
+ */
+@Log4j2
+public class KNN990FlatVectorsFormat extends KnnVectorsFormat{
+    private static Lucene99FlatVectorsFormat lucene99FlatVectorsFormat;
```
rchitale7: can we make this a generic `FlatVectorsFormat`

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNN990FlatVectorsFormat.java
```diff
@@ -0,0 +1,66 @@
...
+    private static final String FORMAT_NAME = "KNN990FlatVectorsFormat";
+
+    public KNN990FlatVectorsFormat() {
```
rchitale7, line 27: do we use this constructor anywhere?

src/main/java/org/opensearch/knn/index/codec/BasePerFieldKnnVectorsFormat.java
```diff
@@ -97,6 +99,14 @@ public KnnVectorsFormat getKnnVectorsFormatForField(final String field) {
             )
         ).fieldType(field);
 
+        // check if exact search is enabled
+        String searchMode = mappedFieldType.getSearchMode();
+        if (searchMode != null && searchMode.equals(EXACT_SEARCH_KEY)) {
```
rchitale7: if this happens before we check for the engine to use, I am wondering why not re-use the existing `NativeEngines990KnnVectorsFormat` class instead of creating a new class specifically for flat vectors?

src/main/java/org/opensearch/knn/index/mapper/EngineFieldMapper.java
```diff
@@ -32,12 +32,7 @@
 import java.util.Map;
 import java.util.Optional;
 
-import static org.opensearch.knn.common.KNNConstants.DIMENSION;
-import static org.opensearch.knn.common.KNNConstants.KNN_ENGINE;
-import static org.opensearch.knn.common.KNNConstants.PARAMETERS;
-import static org.opensearch.knn.common.KNNConstants.QFRAMEWORK_CONFIG;
-import static org.opensearch.knn.common.KNNConstants.SPACE_TYPE;
-import static org.opensearch.knn.common.KNNConstants.VECTOR_DATA_TYPE_FIELD;
+import static org.opensearch.knn.common.KNNConstants.*;
```
rchitale7: lets avoid wildcard imports, they make the code harder to understand

src/main/java/org/opensearch/knn/index/mapper/EngineFieldMapper.java
```diff
@@ -112,7 +107,8 @@ public QuantizationConfig getQuantizationConfig() {
                 public KNNLibraryIndexingContext getKnnLibraryIndexingContext() {
                     return libraryContext;
                 }
-            }
+            },
+            originalMappingParameters.getSearchMode()
```
rchitale7: is there a reason why we cant use the `KnnMappingConfig` object for this parameter? Seems like that would be preferable, since then we dont have to change the signature of `KNNVectorFieldType`

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java
```diff
@@ -291,7 +300,9 @@ public KNNVectorFieldMapper build(BuilderContext context) {
 
             // return FlatVectorFieldMapper only for indices that are created on or after 2.17.0, for others, use
             // EngineFieldMapper to maintain backwards compatibility
-            if (originalParameters.getResolvedKnnMethodContext() == null && indexCreatedVersion.onOrAfter(Version.V_2_17_0)) {
+            // adding check to see if exact search is enabled (only supported for indices created on or after 3.0.0)
+            if ((originalParameters.getResolvedKnnMethodContext() == null && indexCreatedVersion.onOrAfter(Version.V_2_17_0))
+                || (indexCreatedVersion.onOrAfter(Version.V_3_0_0) && originalParameters.getSearchMode() == KNNConstants.EXACT_SEARCH_KEY)) {
```
rchitale7: I don't think you need this. 3.0 comes after 2.17, and I thought this code path gets triggered regardless of the value of EXACT_SEARCH_KEY

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java
```diff
@@ -399,8 +410,9 @@ public Mapper.Builder<?> parse(String name, Map<String, Object> node, ParserCont
                 );
             }
 
-            // Check for flat configuration and validate only if index is created after 2.17
-            if (isKNNDisabled(parserContext.getSettings()) && parserContext.indexVersionCreated().onOrAfter(Version.V_2_17_0)) {
+            // Check for flat configuration and validate only if index is created after 2.17 or if search mode is exact and index is created after 3.0
+            if (isKNNDisabled(parserContext.getSettings()) && parserContext.indexVersionCreated().onOrAfter(Version.V_2_17_0)
```
rchitale7, line 414: I don't think you need this. 3.0 comes after 2.17, and I thought this code path gets triggered regardless of the value of EXACT_SEARCH_KEY

src/main/java/org/opensearch/knn/index/codec/BasePerFieldKnnVectorsFormat.java
kaushikau: fixed to use `NativeEngines990KnnVectorsFormat` with `approximateThreshold` explicitly set to -1

src/main/java/org/opensearch/knn/index/mapper/EngineFieldMapper.java
kaushikau: fixed to use `KnnMappingConfig` instead of `KNNVectorFieldType`! Made the default to be "ann".

navneet1v:
@jmazanec15 Here is what I think:

First I believe we should not provide something like search_mode parameter in index mappings, can we use already present parameter like "index" for a field, if true then you create graphs and if false then don't create graphs. 

search_mode is more of a search time parameter which should be part of query and not index mappings. 

On making parameters modifiable I think if vector data structures are not created then how opening and closing of the index will help?

jmazanec15:
Makes sense. I think we might need to think this over a bit and write-out the api structure to make sure its consistent.

jmazanec15:
Sure @rchitale7! Can we change the PR to point to the feature branch?

src/main/java/org/opensearch/knn/index/codec/BasePerFieldKnnVectorsFormat.java
```diff
@@ -97,6 +98,16 @@ public KnnVectorsFormat getKnnVectorsFormatForField(final String field) {
             )
         ).fieldType(field);
 
+        // check if exact search is enabled
+        String searchMode = mappedFieldType.getKnnMappingConfig().getSearchMode();
+        // setting approximateThreshold to -1 so graphs don't get built
+        if (searchMode != null && searchMode.equals(EXACT_SEARCH_KEY)) {
+            return new NativeEngines990KnnVectorsFormat(
+                new Lucene99FlatVectorsFormat(FlatVectorScorerUtil.getLucene99FlatVectorsScorer()),
+                -1
```
jmazanec15: lets use a constant variable instead of just -1.

src/main/java/org/opensearch/knn/index/mapper/FlatVectorFieldMapper.java
```diff
@@ -80,6 +82,9 @@ private FlatVectorFieldMapper(
         this.perDimensionValidator = selectPerDimensionValidator(vectorDataType);
         this.fieldType = new FieldType(KNNVectorFieldMapper.Defaults.FIELD_TYPE);
         this.fieldType.setDocValuesType(DocValuesType.BINARY);
+        if (originalMappingParameters.getSearchMode() != null) {
+            this.fieldType.putAttribute(SEARCH_MODE, originalMappingParameters.getSearchMode());
```
jmazanec15: Does this need to be an attribute? I think we can probably get away with just using the mapper service from the codec on write and then the mappedfieldtype in the query on read. In general, I like to avoid serializing additional attributes

src/test/java/org/opensearch/knn/integ/SearchModeParameterIT.java
```diff
@@ -0,0 +1,247 @@
...
+        Float[] vector1 = { 1.0f, 2.0f };
+        Float[] vector2 = { 3.0f, 6.0f };
+        addKnnDoc(INDEX_NAME, "0", Arrays.asList("my_vector1", "my_vector2"), Arrays.asList(vector1, vector2));
+
+        // lucene field
+        validateKNNSearch(INDEX_NAME, "my_vector1", DIMENSION, 1, K);
+        validateKNNScriptScoreSearch(INDEX_NAME, "my_vector1", DIMENSION, 1, K, TEST_SPACE_TYPE_L2);
+
+        // exact field
+        ResponseException ex = expectThrows(ResponseException.class, () -> validateKNNSearch(INDEX_NAME, "my_vector2", DIMENSION, 1, K));
```
jmazanec15: Shouldnt the user be able to use the knn query, but it just runs an exact search?

kaushikau: The query support for exact search is actually the second part of my project; this issue is only targeting the unnecessary graph creation. Once I finish part 2, yes, the user should be able to use the knn query for exact search!

src/main/java/org/opensearch/knn/index/codec/BasePerFieldKnnVectorsFormat.java
kaushikau: fixed!

src/main/java/org/opensearch/knn/index/mapper/FlatVectorFieldMapper.java
kaushikau: fixed, took out attributes

src/main/java/org/opensearch/knn/index/codec/BasePerFieldKnnVectorsFormat.java
kaushikau: Went back to using new flat vectors format because `NativeEngines990KnnVectorsReader` does not work for Lucene queries

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNN990FlatVectorsFormat.java
```diff
@@ -0,0 +1,61 @@
...
+import org.apache.lucene.codecs.lucene99.Lucene99FlatVectorsWriter;
+import org.apache.lucene.index.SegmentReadState;
+import org.apache.lucene.index.SegmentWriteState;
+import java.io.IOException;
+
+/**
+ * This is a Vector format that will be used only for flat vectors, when a field is set to index = false.
+ */
+@Log4j2
+public class KNN990FlatVectorsFormat extends FlatVectorsFormat {
```
jmazanec15, line 23: Do we need this class? Instead, can we just use Lucene99FlatVectorsFormat directly?

kaushikau: I tried this at the beginning but we actually can't: `Lucene99FlatVectorsFormat` is not added to the org.apache.lucene.codecs.KnnVectorsFormat file in the resources folder for Lucene, and that results in the "An SPI class of type org.apache.lucene.codecs.KnnVectorsFormat with name 'Lucene99FlatVectorsFormat' does not exist" error. By creating a wrapper class in the KNN plugin, we can add the new format to the org.apache.lucene.codecs.KNNVectorsFormat file and not get the error.

jmazanec15 (commented):
> Note: if a user explicitly sets index.knn.advanced.approximate_threshold to not be equal to 0, that will take precedence over index.

IMO, the mapping parameter should take precedence over the index setting because the mapping parameter is at the field level. Why would it be the other way around?

kaushikau:
> > Note: if a user explicitly sets index.knn.advanced.approximate_threshold to not be equal to 0, that will take precedence over index.
> 
> IMO, the mapping parameter should take precedence over the index setting because the mapping parameter is at the field level. Why would it be the other way around?

I understand why we'd want to give precedence to the index parameter since it's at field-level. However, in the case where, for example, a user explicitly sets `approximate_threshold` to -1, and does not set `index` anywhere, graphs will still be built since we're giving `index` precedence. In that case, wouldn't the user not be getting the expected outcome?

src/main/java/org/opensearch/knn/index/codec/BasePerFieldKnnVectorsFormat.java
```diff
@@ -97,6 +98,15 @@ public KnnVectorsFormat getKnnVectorsFormatForField(final String field) {
             )
         ).fieldType(field);
 
+        // check if index field level parameter is set to false and index.knn.approximate.advanced_threshold is equal to 0 (the default)
+        // since precedence
+        // given to that setting if set -> no graphs built
+        boolean indexed = mappedFieldType.getKnnMappingConfig().isIndexed();
+        if (indexed == false
+            && mapperService.get().getIndexSettings().getValue(KNNSettings.INDEX_KNN_ADVANCED_APPROXIMATE_THRESHOLD_SETTING) == 0) {
```
0ctopus13prime: `mapperService` can be null though it is rare, but still can happen.
Have we discussed over what to do when `mapperService` is null?

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNN990FlatVectorsFormat.java
```diff
@@ -0,0 +1,63 @@
...
+ * This format is necessary for making sure that graph files are not created when not needed, and is a wrapper around the
+ * Lucene99FlatVectorsFormat since that class is not included in the "org.apache.lucene.codecs.KnnVectorsFormat" resources file.
+ */
+@Log4j2
+public class KNN990FlatVectorsFormat extends FlatVectorsFormat {
+    private static FlatVectorsFormat flatVectorsFormat;
+    private static final String FORMAT_NAME = "KNN990FlatVectorsFormat";
+
+    public KNN990FlatVectorsFormat() {
+        this(new Lucene99FlatVectorsFormat(new DefaultFlatVectorScorer()));
```
0ctopus13prime: We should avoid using DefaultFlatVectorScorer as it will lose Panama optimization.
Could you use FlatVectorScorerUtil.getLucene99FlatVectorsScorer()?

0ctopus13prime: Also we don't support pure binary case where user ingests already quantized binary vector with hamming distance??
Hamming is not supported in Lucene, there's no corresponding FlataVectorsScorer for now.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNN990FlatVectorsFormat.java
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    public KNN990FlatVectorsFormat(final FlatVectorsFormat flatVectorsFormat) {
+        super(FORMAT_NAME);
+        KNN990FlatVectorsFormat.flatVectorsFormat = flatVectorsFormat;
```
0ctopus13prime: Could you add 'volatile' keyword for `flatVectorsFormat`? curious, how many times this ctor will be called at most? Will the parameter `flatVectorsFormat` be always the same? 
Also it is static, could you consider pascal format? ex: FLAT_VECTORS_FORMAT

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNN990FlatVectorsFormat.java
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    /**
+     * Returns a {@link FlatVectorsWriter} to write the vectors to the index.
+     *
+     * @param state {@link SegmentWriteState}
+     */
+    @Override
+    public FlatVectorsWriter fieldsWriter(final SegmentWriteState state) throws IOException {
+        return new Lucene99FlatVectorsWriter(state, new DefaultFlatVectorScorer());
```
0ctopus13prime: We should avoid using DefaultFlatVectorScorer as it will lose Panama optimization.
Could you use FlatVectorScorerUtil.getLucene99FlatVectorsScorer()?

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNN990FlatVectorsFormat.java
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    /**
+     * Returns a {@link FlatVectorsReader} to read the vectors from the index.
+     *
+     * @param state {@link SegmentReadState}
+     */
+    @Override
+    public FlatVectorsReader fieldsReader(final SegmentReadState state) throws IOException {
+        return new Lucene99FlatVectorsReader(state, new DefaultFlatVectorScorer());
```
0ctopus13prime: We should avoid using DefaultFlatVectorScorer as it will lose Panama optimization.
Could you use FlatVectorScorerUtil.getLucene99FlatVectorsScorer()?

src/main/java/org/opensearch/knn/index/query/KNNQueryFactory.java
```diff
@@ -71,7 +79,9 @@ public static Query create(CreateQueryRequest createQueryRequest) {
             );
         }
 
-        if (KNNEngine.getEnginesThatCreateCustomSegmentFiles().contains(createQueryRequest.getKnnEngine())) {
+        // if index = false and graphs are not built, use KNNQuery since engine would not matter
+        if (KNNEngine.getEnginesThatCreateCustomSegmentFiles().contains(createQueryRequest.getKnnEngine())
+            || (approxThreshold == 0 && mappedFieldType.getKnnMappingConfig().isIndexed() == false)) {
```
0ctopus13prime: Ah.. do we need this?
Please correct me if I'm missing sth, but it will fallback to exact search when failed to find `.faiss` file anyway.

src/main/java/org/opensearch/knn/index/codec/BasePerFieldKnnVectorsFormat.java
0ctopus13prime: Hmm.. on a second thought.. do we need this?
User explicitly marked `indexed` as false, why we would check the threshold setting again?

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java
```diff
@@ -435,6 +442,26 @@ public Mapper.Builder<?> parse(String name, Map<String, Object> node, ParserCont
             return builder;
         }
 
+        private void validateIndexParameter(KNNVectorFieldMapper.Builder builder) {
+            boolean indexedParam = builder.originalParameters.isIndexed();
+            // ignoring indexedParam if index before 3.0.0 (will always be set to true)
+            if (indexedParam == false && !builder.indexCreatedVersion.onOrAfter(Version.V_3_0_0)) {
```
rchitale7, line 448: @jmazanec15 the `indexCreatedVersion` check is needed for cluster upgrade from 2.x to 3.x, right? Just want to make sure we're handling backwards compatibility correctly.

jmazanec15: What are we trying to do here? If the index is created before 3.2, we should set the parameter as true by default.

src/main/java/org/opensearch/knn/index/query/KNNQueryFactory.java
kaushikau: This check is to make sure to use KNNQuery for queries on lucene indices that have no graphs created, because currently, we can't default to exact search with LuceneEngineKnnVectorQuery

kaushikau:
> > > > Note: if a user explicitly sets index.knn.advanced.approximate_threshold to not be equal to 0, that will take precedence over index.
> > > 
> > > 
> > > IMO, the mapping parameter should take precedence over the index setting because the mapping parameter is at the field level. Why would it be the other way around?
> > 
> > 
> > I understand why we'd want to give precedence to the index parameter since it's at field-level. However, in the case where, for example, a user explicitly sets `approximate_threshold` to -1, and does not set `index` anywhere, graphs will still be built since we're giving `index` precedence (`index` is set to true as default for BWC). In that case, wouldn't the user not be getting the expected outcome?
> 
> I see index mapping parameter as basically saying whether to use approximate index build logic or not. Thus, if its set to true, it should use the approximate build logic. Then, the approx_threshold setting is a part of the approximate build logic. So, if its -1, even if index=true, it still should not build the graphs. That being said, I dont imagine that after `index` mapping parameter is released, users will have the need to set approx_threshold=-1.

I see, this makes sense now. I modified so that index can override approx_threshold if it is false. If it is true, it uses approx_threshold to determine build.

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java
rchitale7: Yeah this code looks like it defaults the `index` field parameter to true if the index was created before 3.0. I was not sure if `index` field parameter should be enabled for 3.x versions or just for >= 3.2.

jmazanec15: I see. Maybe we can fix this version in a rebase. I think it makes sense to only enable for >= 3.2. It wont matter too much because the mapping params are not dynamic

0ctopus13prime (approved):

PR #2773, opened 2025-06-23
Don't fall back to CPU on terminal exceptions

### LINKED ISSUE #2766: [BUG] Terminate Index build request if Failed to write index in segment
## Description
Currently, if a k-NN index graph build fails at segment level for GPU based index acceleration, we always fallback to to CPU. But there are some hard failures like IndexInput is closed or index is deleted. We fallback to CPU and then again hit same hard failure before failing the shard. 

Ideally in case of hard failures we should throw the exception out and shouldn't fall back to CPU.

BTW in the CPU flow, if we are not able to write the index in the segment we completely fail the shard. So it should be safe to do so.

The current behavior is just not optimal it doesn't impact consistency. It just waste the compute.

owenhalpert:
### Description
From the issue:

> Currently, if a k-NN index graph build fails at segment level for GPU based index acceleration, we always fallback to to CPU. But there are some hard failures like IndexInput is closed or index is deleted. We fallback to CPU and then again hit same hard failure before failing the shard. Ideally in case of hard failures we should throw the exception out and shouldn't fall back to CPU.

This PR introduces a new exception to signal to the remote build process that no fall back strategy should be invoked, instead the exception should be rethrown to the caller to terminate the build altogether.

### Related Issues
Resolves #2766 

### Check List
~- [] New functionality has been documented.~

### DIFF
--- a/src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
+++ b/src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
@@ -5,6 +5,7 @@
 
 package org.opensearch.knn.index.codec.nativeindex.remote;
 
+import com.google.common.annotations.VisibleForTesting;
 import lombok.extern.log4j.Log4j2;
 import org.opensearch.cluster.ClusterName;
 import org.opensearch.cluster.metadata.RepositoryMetadata;
@@ -13,6 +14,7 @@ import org.opensearch.common.annotation.ExperimentalApi;
 import org.opensearch.common.blobstore.BlobPath;
 import org.opensearch.core.common.unit.ByteSizeValue;
 import org.opensearch.index.IndexSettings;
+import org.opensearch.knn.common.exception.TerminalIOException;
 import org.opensearch.knn.index.KNNSettings;
 import org.opensearch.knn.index.codec.nativeindex.NativeIndexBuildStrategy;
 import org.opensearch.knn.index.codec.nativeindex.model.BuildIndexParams;
@@ -137,9 +139,11 @@ public class RemoteIndexBuildStrategy implements NativeIndexBuildStrategy {
      *      2. Triggers index build
      *      3. Awaits on vector build to complete
      *      4. Downloads index file and writes to indexOutput
+     * In most cases, a failure in this process will trigger a fall back to the designated {@link NativeIndexBuildStrategy}.
+     * However, if a {@link TerminalIOException} is thrown, it will be caught and rethrown to terminate the request.
      *
      * @param indexInfo {@link BuildIndexParams} containing information about the index to be built
-     * @throws IOException if an error occurs during the build process
+     * @throws IOException if a terminal error occurs during the build process
      */
     @Override
     public void buildAndWriteIndex(BuildIndexParams indexInfo) throws IOException {
@@ -162,6 +166,8 @@ public class RemoteIndexBuildStrategy implements NativeIndexBuildStrategy {
             readFromRepository(indexInfo, repositoryContext, remoteBuildStatusResponse);
             success = true;
             return;
+        } catch (TerminalIOException e) {
+            throw e;
         } catch (Exception e) {
             log.error("Failed to build index remotely: " + indexInfo, e);
         } finally {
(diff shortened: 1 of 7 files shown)


src/main/java/org/opensearch/knn/index/store/IndexOutputWithBuffer.java
```diff
@@ -72,7 +73,11 @@ private void writeFromStreamWithBuffer(InputStream inputStream, byte[] outputBuf
             assert bytesRead <= outputBuffer.length;
             // However many bytes we read, write it to the IndexOutput if != -1
             if (bytesRead != -1) {
-                indexOutput.writeBytes(outputBuffer, 0, bytesRead);
+                try {
+                    indexOutput.writeBytes(outputBuffer, 0, bytesRead);
+                } catch (IOException e) {
+                    throw new RemoteIndexBuildStrategy.TerminalIOException("Failed to write to indexOutput", e);
```
jed326: So `writeFromStreamWithBuffer` is actually used by both the native and remote index build strategies. I'm a little worried about catching and rethrowing like this in the native case, so I'm wondering if we should just split the native and remote use cases into 2 different methods?

owenhalpert: It looks like both paths do share the `writeBytes` method but this `writeFromStreamWithBuffer` method is only in the remote path. So, it should be fine to keep this here.

Vikasht34: So here we are catching IO exception and then wrapping with terminal exception,  why ? Why can't we we terminate the RemoteBuild in case of IO exception? Unless/ until we have way to differentiate between out of these IO Execptions , theses are my termianal states , Which I don't see here.

owenhalpert: There are many places in the code, throughout the entire remote build process, where an IOException can be thrown that is not a terminal failure. In these cases we need to fallback. Example:

https://github.com/opensearch-project/k-NN/blob/5aec8550f34bb55f57eec895fd10c0a1ddc43ae6/src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/DefaultVectorRepositoryAccessor.java#L230-L243

The `readFromRepository` method has potential IOExceptions being thrown on both of the last two lines:

```
InputStream graphStream = blobContainer.readBlob(fileName);
indexOutputWithBuffer.writeFromStreamWithBuffer(graphStream, INDEX_DOWNLOAD_BUFFER_SIZE);
```

We do not want to cancel the build if the S3 operation `readBlob` fails. We do want to cancel if the `writeFromStreamWithBuffer` fails on the write to indexOutput, because of the statement at the top of this PR.

owenhalpert:
> @owenhalpert So we never gonna fall back to CPU ...because IndexInput will be closed on all the cases. Is my Understanding correct.

No, the index input will not be closed in every case. That would be considered a hard failure on the CPU build, so instead of trying the CPU build and failing again, we can save the compute by failing early in the GPU path for these specific terminal error scenarios like the index input being closed or the index deleted.

src/main/java/org/opensearch/knn/index/store/IndexOutputWithBuffer.java
Vikasht34: Makes Sense !!

Vikasht34 (dismissed):

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
```diff
@@ -349,4 +357,15 @@ static RemoteBuildRequest buildRemoteBuildRequest(
             .indexParameters(indexInfo.getKnnEngine().createRemoteIndexingParameters(parameters))
             .build();
     }
+
+    /**
+     * Custom exception class to distinguish between terminal IOExceptions (in the case of writing to index output,
+     * for example) and non-terminal (e.g. reading from S3). See {@link DefaultVectorRepositoryAccessor#readFromRepository}.
+     * Useful for determining whether to fall back to CPU build or terminate build altogether.
+     */
+    public static class TerminalIOException extends IOException {
```
navneet1v: [Minor] we should move this to a separate file, inner classes are good if you keep the scope limited to just same file. but since this exception is being used at multiple places better to have a separate file for it

navneet1v (approved):

navneet1v (approved):

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
owenhalpert: Thanks Navneet, I like that

PR #2781, opened 2025-07-02
Support ADC in Lucene-on-faiss initial commit

### LINKED ISSUE #2714: [FEATURE][RFC] Asymmetric Distance Calculation and Random Rotation to Boost Recall for On-Disk Vector Search.

finnroblin:
### Description
Draft implementation to integrate asymmetric distance calculation (ADC) with lucene-on-faiss.

This implementation uses the following low-level approach:
- Within the `NativeEngines990KnnVectorsReader` see from the `FieldInfo` if adc is enabled or not. We need to check here so we can obtain the original `SpaceType`, as this information is lost for binary indices when it is changed to hamming.
- Add new split within `trySearchWithMemoryOptimizedSearch` on if adc is enabled or not.
- Add `isAdc` parameter to `createVectorSearcher` method.
- Add new method `searchWithAdc`. The reason for a new method is that preexisting scorer supplier logic depends on the query and document vector types matching (`float/float` or `byte/byte`), whereas ADC is `float/byte`. 
- Create new ADC scorer, leveraging java logic originally implemented for ADC exact search.

### Related Issues
Resolves #2714 

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
+++ b/src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
@@ -29,6 +29,8 @@ import org.apache.lucene.util.IOSupplier;
 import org.apache.lucene.util.IOUtils;
 import org.opensearch.common.UUIDs;
 import org.opensearch.knn.common.FieldInfoExtractor;
+import org.opensearch.knn.common.KNNConstants;
+import org.opensearch.knn.index.SpaceType;
 import org.opensearch.knn.index.codec.util.KNNCodecUtil;
 import org.opensearch.knn.index.codec.util.NativeMemoryCacheKeyHelper;
 import org.opensearch.knn.index.engine.KNNEngine;
@@ -36,9 +38,12 @@ import org.opensearch.knn.index.memory.NativeMemoryCacheManager;
 import org.opensearch.knn.index.quantizationservice.QuantizationService;
 import org.opensearch.knn.memoryoptsearch.VectorSearcher;
 import org.opensearch.knn.memoryoptsearch.VectorSearcherFactory;
+import org.opensearch.knn.memoryoptsearch.faiss.FaissMemoryOptimizedSearcher;
 import org.opensearch.knn.quantization.models.quantizationState.QuantizationState;
 import org.opensearch.knn.quantization.models.quantizationState.QuantizationStateCacheManager;
 import org.opensearch.knn.quantization.models.quantizationState.QuantizationStateReadConfig;
+import org.opensearch.knn.index.engine.qframe.QuantizationConfig;
+import org.opensearch.knn.index.engine.qframe.QuantizationConfigParser;
 
 import java.io.Closeable;
 import java.io.IOException;
@@ -63,11 +68,23 @@ public class NativeEngines990KnnVectorsReader extends KnnVectorsReader {
     private final SegmentReadState segmentReadState;
     private final List<String> cacheKeys;
     private volatile Map<String, VectorSearcherHolder> vectorSearchers;
+    private final Map<String, Boolean> adcEnabledCache;
+
+    private static class AdcConfig {
+        final boolean isEnabled;
+        final SpaceType spaceType;
+
+        AdcConfig(boolean isEnabled, SpaceType spaceType) {
+            this.isEnabled = isEnabled;
+            this.spaceType = spaceType;
+        }
+    }
 
     public NativeEngines990KnnVectorsReader(final SegmentReadState state, final FlatVectorsReader flatVectorsReader) {
         this.flatVectorsReader = flatVectorsReader;
         this.segmentReadState = state;
         this.cacheKeys = getVectorCacheKeysFromSegmentReaderState(state);
+        this.adcEnabledCache = new HashMap<>();
 
         loadCacheKeyMap();
         fillVectorSearcherTable();
@@ -152,8 +169,10 @@ public class NativeEngines990KnnVectorsReader extends KnnVectorsReader {
             ((QuantizationConfigKNNCollector) knnCollector).setQuantizationState(quantizationState);
             return;
         }
-
-        if (trySearchWithMemoryOptimizedSearch(field, target, knnCollector, acceptDocs, true)) {
+        // get adc and spaceType from FieldInfo using caching mechanism to avoid repeated work. It should only be calculated
+        // once per Reader construction. There is no cache invalidation logic since adc is a static mapping parameter.
+        AdcConfig adcConfig = getAdcConfig(field);
+        if (trySearchWithMemoryOptimizedSearch(field, target, knnCollector, acceptDocs, true, adcConfig.isEnabled, adcConfig.spaceType)) {
             return;
         }
 
@@ -187,7 +206,8 @@ public class NativeEngines990KnnVectorsReader extends KnnVectorsReader {
      */
     @Override
     public void search(String field, byte[] target, KnnCollector knnCollector, Bits acceptDocs) throws IOException {
-        if (trySearchWithMemoryOptimizedSearch(field, target, knnCollector, acceptDocs, false)) {
+        // searching with byte vector is not supported by ADC.
+        if (trySearchWithMemoryOptimizedSearch(field, target, knnCollector, acceptDocs, false, false, null)) {
             return;
         }
 
@@ -240,12 +260,24 @@ public class NativeEngines990KnnVectorsReader extends KnnVectorsReader {
         final Object target,
         final KnnCollector knnCollector,
         final Bits acceptDocs,
-        final boolean isFloatVector
+        final boolean isFloatVector,
+        final boolean isAdc,
+        final SpaceType spaceType
     ) throws IOException {
         // Try with memory optimized searcher
-        final VectorSearcher memoryOptimizedSearcher = loadMemoryOptimizedSearcherIfRequired(field);
+        final VectorSearcher memoryOptimizedSearcher = loadMemoryOptimizedSearcherIfRequired(field, isAdc);
 
         if (memoryOptimizedSearcher != null) {
+            if (isAdc) {
+                return trySearchWithMemoryOptimizedSearchAndADC(
+                    memoryOptimizedSearcher,
+                    (float[]) target,
+                    knnCollector,
+                    acceptDocs,
+                    spaceType
+                );
+            }
+
             if (isFloatVector) {
                 memoryOptimizedSearcher.search((float[]) target, knnCollector, acceptDocs);
             } else {
@@ -254,6 +286,24 @@ public class NativeEngines990KnnVectorsReader extends KnnVectorsReader {
             return true;
         }
         return false;
+    }
+
+    private boolean trySearchWithMemoryOptimizedSearchAndADC(
+        final VectorSearcher memoryOptimizedSearcher,
+        float[] target,
+        final KnnCollector knnCollector,
+        final Bits acceptDocs,
+        final SpaceType spaceType
+    ) throws IOException {
+        // from NativeEngines990KnnVectorsReader if searching with Byte. Should never hit the below path.
+        if (spaceType == null) return false;
+
+        if (memoryOptimizedSearcher instanceof FaissMemoryOptimizedSearcher faissMemoryOptimizedSearcher) {
+            faissMemoryOptimizedSearcher.searchWithAdc(target, knnCollector, acceptDocs, spaceType);
+            return true;
+        }
+        return false;
+
     }
 
     private void loadCacheKeyMap() {
@@ -270,7 +320,8 @@ public class NativeEngines990KnnVectorsReader extends KnnVectorsReader {
         vectorSearchers = new HashMap<>(RESERVE_TWICE_SPACE * segmentReadState.fieldInfos.size(), SUFFICIENT_LOAD_FACTOR);
 
         for (final FieldInfo fieldInfo : segmentReadState.fieldInfos) {
-            final IOSupplier<VectorSearcher> searcherIOSupplier = getVectorSearcherSupplier(fieldInfo);
+            // TODO: verify that adc does not change this calculation/break anything
+            final IOSupplier<VectorSearcher> searcherIOSupplier = getVectorSearcherSupplier(fieldInfo, false);
             if (searcherIOSupplier != null) {
                 // This field type is supported
                 vectorSearchers.put(fieldInfo.getName(), new VectorSearcherHolder());
@@ -293,7 +344,7 @@ public class NativeEngines990KnnVectorsReader extends KnnVectorsReader {
         return cacheKeys;
     }
 
-    private VectorSearcher loadMemoryOptimizedSearcherIfRequired(final String fieldName) {
+    private VectorSearcher loadMemoryOptimizedSearcherIfRequired(final String fieldName, boolean isAdc) {
         final VectorSearcherHolder searcherHolder = vectorSearchers.get(fieldName);
         if (searcherHolder == null) {
             // This is not KNN field or unsupported field.
@@ -314,7 +365,7 @@ public class NativeEngines990KnnVectorsReader extends KnnVectorsReader {
             try {
                 final FieldInfo fieldInfo = segmentReadState.fieldInfos.fieldInfo(fieldName);
                 if (fieldInfo != null) {
-                    final IOSupplier<VectorSearcher> searcherSupplier = getVectorSearcherSupplier(fieldInfo);
+                    final IOSupplier<VectorSearcher> searcherSupplier = getVectorSearcherSupplier(fieldInfo, isAdc);
                     if (searcherSupplier != null) {
                         searcher = searcherSupplier.get();
                         if (searcher != null) {
@@ -339,13 +390,12 @@ public class NativeEngines990KnnVectorsReader extends KnnVectorsReader {
         }
     }
 
-    private IOSupplier<VectorSearcher> getVectorSearcherSupplier(final FieldInfo fieldInfo) {
+    private IOSupplier<VectorSearcher> getVectorSearcherSupplier(final FieldInfo fieldInfo, boolean isAdc) {
         // Skip non-knn fields.
         final Map<String, String> attributes = fieldInfo.attributes();
         if (attributes == null || attributes.containsKey(KNN_FIELD) == false) {
             return null;
         }
-
         // Try to get KNN engine from fieldInfo.
         final KNNEngine knnEngine = FieldInfoExtractor.extractKNNEngine(fieldInfo);
 
@@ -364,11 +414,29 @@ public class NativeEngines990KnnVectorsReader extends KnnVectorsReader {
         // Start creating searcher
         final String fileName = KNNCodecUtil.getNativeEngineFileFromFieldInfo(fieldInfo, segmentReadState.segmentInfo);
         if (fileName != null) {
-            return () -> searcherFactory.createVectorSearcher(segmentReadState.directory, fileName);
+            return () -> searcherFactory.createVectorSearcher(segmentReadState.directory, fileName, isAdc);
         }
 
         // Not supported
         return null;
+    }
+
+    private AdcConfig getAdcConfig(String field) {
+        boolean isAdcEnabled = adcEnabledCache.computeIfAbsent(field, f -> {
+            FieldInfo fieldInfo = segmentReadState.fieldInfos.fieldInfo(f);
+            String qframeConfig = fieldInfo.getAttribute(KNNConstants.QFRAMEWORK_CONFIG);
+            if (qframeConfig != null) {
+                QuantizationConfig config = QuantizationConfigParser.fromCsv(qframeConfig, segmentReadState.segmentInfo.getVersion());
+                return config.isEnableADC();
+            }
+            return false;
+        });
+
+        SpaceType spaceType = isAdcEnabled
+            ? SpaceType.getSpace(segmentReadState.fieldInfos.fieldInfo(field).getAttribute(KNNConstants.SPACE_TYPE))
+            : null;
+
+        return new AdcConfig(isAdcEnabled, spaceType);
     }
 
     /**
--- a/src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
+++ b/src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
@@ -29,8 +29,6 @@ import org.apache.lucene.util.IOSupplier;
 import org.apache.lucene.util.IOUtils;
 import org.opensearch.common.UUIDs;
 import org.opensearch.knn.common.FieldInfoExtractor;
-import org.opensearch.knn.common.KNNConstants;
-import org.opensearch.knn.index.SpaceType;
 import org.opensearch.knn.index.codec.util.KNNCodecUtil;
 import org.opensearch.knn.index.codec.util.NativeMemoryCacheKeyHelper;
 import org.opensearch.knn.index.engine.KNNEngine;
@@ -38,12 +36,9 @@ import org.opensearch.knn.index.memory.NativeMemoryCacheManager;
 import org.opensearch.knn.index.quantizationservice.QuantizationService;
 import org.opensearch.knn.memoryoptsearch.VectorSearcher;
 import org.opensearch.knn.memoryoptsearch.VectorSearcherFactory;
-import org.opensearch.knn.memoryoptsearch.faiss.FaissMemoryOptimizedSearcher;
 import org.opensearch.knn.quantization.models.quantizationState.QuantizationState;
 import org.opensearch.knn.quantization.models.quantizationState.QuantizationStateCacheManager;
 import org.opensearch.knn.quantization.models.quantizationState.QuantizationStateReadConfig;
-import org.opensearch.knn.index.engine.qframe.QuantizationConfig;
-import org.opensearch.knn.index.engine.qframe.QuantizationConfigParser;
 
 import java.io.Closeable;
 import java.io.IOException;
@@ -68,23 +63,11 @@ public class NativeEngines990KnnVectorsReader extends KnnVectorsReader {
     private final SegmentReadState segmentReadState;
     private final List<String> cacheKeys;
     private volatile Map<String, VectorSearcherHolder> vectorSearchers;
-    private final Map<String, Boolean> adcEnabledCache;
-
-    private static class AdcConfig {
-        final boolean isEnabled;
-        final SpaceType spaceType;
-
-        AdcConfig(boolean isEnabled, SpaceType spaceType) {
-            this.isEnabled = isEnabled;
-            this.spaceType = spaceType;
-        }
-    }
 
     public NativeEngines990KnnVectorsReader(final SegmentReadState state, final FlatVectorsReader flatVectorsReader) {
         this.flatVectorsReader = flatVectorsReader;
         this.segmentReadState = state;
         this.cacheKeys = getVectorCacheKeysFromSegmentReaderState(state);
-        this.adcEnabledCache = new HashMap<>();
 
         loadCacheKeyMap();
         fillVectorSearcherTable();
@@ -169,10 +152,7 @@ public class NativeEngines990KnnVectorsReader extends KnnVectorsReader {
             ((QuantizationConfigKNNCollector) knnCollector).setQuantizationState(quantizationState);
             return;
         }
-        // get adc and spaceType from FieldInfo using caching mechanism to avoid repeated work. It should only be calculated
-        // once per Reader construction. There is no cache invalidation logic since adc is a static mapping parameter.
-        AdcConfig adcConfig = getAdcConfig(field);
-        if (trySearchWithMemoryOptimizedSearch(field, target, knnCollector, acceptDocs, true, adcConfig.isEnabled, adcConfig.spaceType)) {
+        if (trySearchWithMemoryOptimizedSearch(field, target, knnCollector, acceptDocs, true)) {
             return;
         }
 
@@ -207,7 +187,7 @@ public class NativeEngines990KnnVectorsReader extends KnnVectorsReader {
     @Override
     public void search(String field, byte[] target, KnnCollector knnCollector, Bits acceptDocs) throws IOException {
         // searching with byte vector is not supported by ADC.
-        if (trySearchWithMemoryOptimizedSearch(field, target, knnCollector, acceptDocs, false, false, null)) {
+        if (trySearchWithMemoryOptimizedSearch(field, target, knnCollector, acceptDocs, false)) {
             return;
         }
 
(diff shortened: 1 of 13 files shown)


finnroblin:
Will add changelog entry to next revision.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
```diff
@@ -63,11 +68,23 @@ public class NativeEngines990KnnVectorsReader extends KnnVectorsReader {
     private final SegmentReadState segmentReadState;
     private final List<String> cacheKeys;
     private volatile Map<String, VectorSearcherHolder> vectorSearchers;
+    private final Map<String, Boolean> adcEnabledCache;
+
```
0ctopus13prime: Super NIT : Could you use `@RequiredArgsConstructor`

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
```diff
@@ -364,13 +414,31 @@ private IOSupplier<VectorSearcher> getVectorSearcherSupplier(final FieldInfo fie
...
+            return () -> searcherFactory.createVectorSearcher(segmentReadState.directory, fileName, isAdc);
         }
 
         // Not supported
         return null;
     }
 
+    private AdcConfig getAdcConfig(String field) {
+        boolean isAdcEnabled = adcEnabledCache.computeIfAbsent(field, f -> {
+            FieldInfo fieldInfo = segmentReadState.fieldInfos.fieldInfo(f);
```
0ctopus13prime: Do we need to wrap this with try-catch and return false whenever an exception to be thrown?

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
```diff
@@ -63,11 +68,23 @@ public class NativeEngines990KnnVectorsReader extends KnnVectorsReader {
     private final SegmentReadState segmentReadState;
     private final List<String> cacheKeys;
     private volatile Map<String, VectorSearcherHolder> vectorSearchers;
+    private final Map<String, Boolean> adcEnabledCache;
+
+    private static class AdcConfig {
```
shatejas: nit: create a record?

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
```diff
@@ -63,11 +68,23 @@ public class NativeEngines990KnnVectorsReader extends KnnVectorsReader {
...
+            this.isEnabled = isEnabled;
+            this.spaceType = spaceType;
+        }
+    }
 
     public NativeEngines990KnnVectorsReader(final SegmentReadState state, final FlatVectorsReader flatVectorsReader) {
         this.flatVectorsReader = flatVectorsReader;
         this.segmentReadState = state;
         this.cacheKeys = getVectorCacheKeysFromSegmentReaderState(state);
+        this.adcEnabledCache = new HashMap<>();
```
shatejas: - Curious why is the cache not initialized here and is done at search runtime
- Another question would be why not cache space type as well?

finnroblin: > - Curious why is the cache not initialized here and is done at search runtime

I thought getting the exact vector field names would be complicated here. I could also initialize the cache when the vector searcher is instantiated which I think is your suggestion. Let me see if that approach is possible.

>Another question would be why not cache space type as well?

We also cache spacetype. I'll rename adcEnabledCache to something more descriptive to convey that space type is also cached.

finnroblin:
Thanks @shatejas for the review. I understand the concern around breaking the FlatVectorScorer interface with the `searchWithAdc` method. 

I implemented a different approach in my latest commit that instantiates a `ADCFlatVectorsScorer` through the `FaissMemoryOptimizedSearcherFactory.createVectorSearcher` method. The factory is passed `FieldInfo` and then possibly creates an ADC scorer based on the quantization state. I believe that this is relatively efficient... since the vector searcher already makes a costly `Directory.openInput` call, it is not too much extra overhead to also grab adc info from quantization state. My hope is this change also addresses your comments about caching.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
finnroblin: Ack, went with a different approach where this info is stored in the scorer as per Tejas's suggestion.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
finnroblin: Ack, went with a different approach where this info is stored in the scorer as per Tejas's suggestion.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
finnroblin: Moved to searcher.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FlatVectorsScorerProvider.java
```diff
@@ -30,6 +32,78 @@ public static FlatVectorsScorer getFlatVectorsScorer(final KNNVectorSimilarityFu
...
+        }
+
+        @Override
+        public RandomVectorScorer getRandomVectorScorer(
+            VectorSimilarityFunction vectorSimilarityFunction,
+            KnnVectorValues knnVectorValues,
+            float[] target
+        ) {
+            if (knnVectorValues instanceof ByteVectorValues byteVectorValues) {
+                if (spaceType == SpaceType.HAMMING) {
```
shatejas: nit: switch case will be cleaner

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FlatVectorsScorerProvider.java
```diff
@@ -30,6 +32,78 @@ public static FlatVectorsScorer getFlatVectorsScorer(final KNNVectorSimilarityFu
...
+                            );
+                        }
+                    };
+                } else {
+                    throw new IllegalArgumentException("Unsupported space type: " + spaceType);
+                }
+            }
+
+            throw new IllegalArgumentException(
+                "Expected " + ByteVectorValues.class.getSimpleName() + " for ADC scorer, got " + knnVectorValues.getClass().getSimpleName()
```
shatejas: nit: check this upfront for cleaner code
```
if (!(knnvectorvalue instanceof byte) {
     throw
}

// rest of the logic
```

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissMemoryOptimizedSearcher.java
```diff
@@ -34,18 +35,27 @@ public class FaissMemoryOptimizedSearcher implements VectorSearcher {
...
             vectorSimilarityFunction = knnVectorSimilarityFunction.getVectorSimilarityFunction();
         } else {
             vectorSimilarityFunction = null;
         }
+        // spaceType is only used if ADC is enabled.
+        this.isAdc = isAdc;
+        this.spaceType = spaceType;
+        this.flatVectorsScorer = isAdc
+            ? new FlatVectorsScorerProvider.ADCFlatVectorsScorer(knnVectorSimilarityFunction, spaceType)
+            : FlatVectorsScorerProvider.getFlatVectorsScorer(knnVectorSimilarityFunction);
```
shatejas: Pass `isADC` to the provider. The purpose of the utility class is to provide the right scorer so pass in all the context it needs as a parameter

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissMemoryOptimizedSearcher.java
```diff
@@ -34,18 +35,27 @@ public class FaissMemoryOptimizedSearcher implements VectorSearcher {
...
-        this.flatVectorsScorer = FlatVectorsScorerProvider.getFlatVectorsScorer(knnVectorSimilarityFunction);
+
         if (knnVectorSimilarityFunction != KNNVectorSimilarityFunction.HAMMING) {
             vectorSimilarityFunction = knnVectorSimilarityFunction.getVectorSimilarityFunction();
         } else {
             vectorSimilarityFunction = null;
         }
+        // spaceType is only used if ADC is enabled.
+        this.isAdc = isAdc;
+        this.spaceType = spaceType;
```
shatejas: Is this member variable still needed here?

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissMemoryOptimizedSearcherFactory.java
```diff
@@ -23,15 +29,21 @@
...
         final IndexInput indexInput = directory.openInput(
             fileName,
             new IOContext(IOContext.Context.DEFAULT, null, null, ReadAdvice.RANDOM)
         );
 
         try {
+            // Extract ADC info from fieldInfo
+            final QuantizationConfig quantizationConfig = FieldInfoExtractor.extractQuantizationConfig(fieldInfo, Version.LATEST);
+            final boolean isAdc = quantizationConfig.isEnableADC();
+            final SpaceType spaceType = isAdc ? SpaceType.getSpace(fieldInfo.getAttribute(KNNConstants.SPACE_TYPE)) : null;
```
shatejas: nit: I would say delay this until FAISSSearcher, Factories should only do additional stuff if it helps it make a decision

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissMemoryOptimizedSearcher.java
finnroblin: No, will remove spaceType.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FlatVectorsScorerProvider.java
```diff
@@ -14,22 +14,131 @@
...
+                throw new IllegalArgumentException(
+                    "Expected "
+                        + ByteVectorValues.class.getSimpleName()
+                        + " for ADC scorer, got "
+                        + knnVectorValues.getClass().getSimpleName()
+                );
+            }
+
+            return switch (spaceType) {
+                case HAMMING -> throw new IllegalArgumentException("hamming distance unsupported by ADC");
```
shatejas: nit: probably not needed, let it go to default and change the message in default to "spaceType not supported for ADC"

shatejas (commented):
Thanks for all the changes, this is looking good. waiting for the CIs to pass before approving

0ctopus13prime (approved):

Issue #2286, opened 2024-11-25
[FEATURE] Additional components in profile breakdown for KNN query to increase visibility


shatejas:
**Is your feature request related to a problem?**

KNNQuery execution has many moving parts. It intelligently selects between exact knn search and approximate search, it rescores when there is oversampling factor present and it also executes filter query if any. With varying number of use cases, its difficult to figure out time taken by each or a combination of these individually to optimize the query or debug latencies. The request is to add time taken be each of these components when `profile = true` which increases visibility

**What solution would you like?**
I propose two new additional components in query break down `exact_knn_search` and `ann_search` along with the support to have visibility on filter query in knn in the current breakdown tree. Here is the sample response

```
"profile": {
		"shards": [
			{
				"id": "[lChSC1BdRFWt9dfBJoO3UA][hotels-hnsw][0]",
				"inbound_network_time_in_millis": 0,
				"outbound_network_time_in_millis": 0,
				"searches": [
					{
						"query": [
							{
								"type": "KNNQuery",
								"description": "",
								"time_in_nanos": 20503209,
								"breakdown": {
									"set_min_competitive_score_count": 0,
									"exact_knn_search": 5436542,
									"match_count": 0,
									"shallow_advance_count": 0,
									"next_doc": 7584,
									"score_count": 3,
									"compute_max_score_count": 0,
									"advance": 0,
									"advance_count": 0,
									"score": 6083,
									"shallow_advance": 0,
									"create_weight_count": 1,
									"build_scorer": 11771833,
									"ann_search": 0,
									"set_min_competitive_score": 0,
									"match": 0,
									"next_doc_count": 4,
									"compute_max_score": 0,
									"build_scorer_count": 2,
									"create_weight": 3281167,
									"ann_search_count": 0,
									"exact_knn_search_count": 1
								},
								"children": [
									{
										"type": "BooleanQuery",
										"description": "+IndexOrDocValuesQuery(indexQuery=rating:[8 TO 10], dvQuery=rating:[8 TO 10])",
										"time_in_nanos": 4047005,
										"breakdown": {
											"set_min_competitive_score_count": 0,
											"exact_knn_search": 0,
											"match_count": 0,
											"shallow_advance_count": 0,
											"next_doc": 28545,
											"score_count": 3,
											"compute_max_score_count": 0,
											"advance": 0,
											"advance_count": 0,
											"score": 3876,
											"shallow_advance": 0,
											"create_weight_count": 0,
[...]

PR #2785, opened 2025-07-09
Add KNN Profiling Info to Core Profiler

neuenfeldttj:
### Description
Adds Exact Search and ANN Search timing info to the profile breakdown for each `KNNQuery`, `NativeEngineKnnVectorQuery` or `LuceneEngineKnnVectorQuery`. **Currently, the core repo has a PR to implement a plugin extension point (https://github.com/opensearch-project/OpenSearch/pull/18656). Until then, `KNNPlugin` is still a WIP so please ignore.** 

Timers are found in the enums `KNNQueryTimingType`, `LuceneEngineKnnTimingType`, `NativeEngineTimingType`. In addition, the filter cardinality (KNNQuery, Native query) and num_nested_docs (Native query) stats are included in the profile breakdown. Feel free to suggest changes or more timers/other stats to the profiler!

### Related Issues
Resolves https://github.com/opensearch-project/OpenSearch/pull/17146
Resolves https://github.com/opensearch-project/k-NN/issues/2286

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/query/KNNQuery.java
+++ b/src/main/java/org/opensearch/knn/index/query/KNNQuery.java
@@ -24,6 +24,9 @@ import org.opensearch.common.StopWatch;
 import org.opensearch.knn.index.VectorDataType;
 import org.opensearch.knn.index.query.memoryoptsearch.MemoryOptimizedKNNWeight;
 import org.opensearch.knn.index.query.rescore.RescoreContext;
+import org.opensearch.knn.profile.query.KNNMetrics;
+import org.opensearch.search.profile.ContextualProfileBreakdown;
+import org.opensearch.search.profile.Profilers;
 
 import java.io.IOException;
 import java.util.Arrays;
@@ -175,6 +178,11 @@ public class KNNQuery extends Query {
      */
     @Override
     public Weight createWeight(IndexSearcher searcher, ScoreMode scoreMode, float boost) throws IOException {
+        Profilers profilers = KNNMetrics.getProfilers();
+        ContextualProfileBreakdown profile = null;
+        if (profilers != null) {
+            profile = profilers.getCurrentQueryProfiler().getTopBreakdown();
+        }
         StopWatch stopWatch = null;
         if (log.isDebugEnabled()) {
             stopWatch = new StopWatch().start();
@@ -193,11 +201,11 @@ public class KNNQuery extends Query {
 
         if (isMemoryOptimizedSearch) {
             // Using memory optimized search logic on index.
-            return new MemoryOptimizedKNNWeight(this, boost, filterWeight, searcher, k);
+            return new MemoryOptimizedKNNWeight(this, boost, filterWeight, searcher, k, profile);
         }
 
         // Using native library to perform search on index.
-        return new DefaultKNNWeight(this, boost, filterWeight);
+        return new DefaultKNNWeight(this, boost, filterWeight, profile);
     }
 
     private Weight getFilterWeight(IndexSearcher searcher) throws IOException {
--- a/src/main/java/org/opensearch/knn/index/query/KNNQueryFactory.java
+++ b/src/main/java/org/opensearch/knn/index/query/KNNQueryFactory.java
@@ -7,14 +7,14 @@ package org.opensearch.knn.index.query;
 
 import lombok.NonNull;
 import lombok.extern.log4j.Log4j2;
-import org.apache.lucene.search.KnnByteVectorQuery;
-import org.apache.lucene.search.KnnFloatVectorQuery;
 import org.apache.lucene.search.Query;
 import org.apache.lucene.search.join.BitSetProducer;
 import org.opensearch.index.query.QueryShardContext;
 import org.opensearch.knn.index.VectorDataType;
 import org.opensearch.knn.index.engine.KNNEngine;
 import org.opensearch.knn.index.query.common.QueryUtils;
+import org.opensearch.knn.index.query.lucene.ProfileKnnByteVectorQuery;
+import org.opensearch.knn.index.query.lucene.ProfileKnnFloatVectorQuery;
 import org.opensearch.knn.index.query.lucenelib.NestedKnnVectorQueryFactory;
 import org.opensearch.knn.index.query.lucene.LuceneEngineKnnVectorQuery;
 import org.opensearch.knn.index.query.nativelib.NativeEngineKnnVectorQuery;
@@ -179,8 +179,8 @@ public class KNNQueryFactory extends BaseQueryFactory {
         if (parentFilter == null) {
             assert expandNested == false : "expandNested is allowed to be true only for nested fields.";
             return vectorDataType == VectorDataType.FLOAT
-                ? new KnnFloatVectorQuery(fieldName, floatQueryVector, k, filterQuery)
-                : new KnnByteVectorQuery(fieldName, byteQueryVector, k, filterQuery);
+                ? new ProfileKnnFloatVectorQuery(fieldName, floatQueryVector, k, filterQuery)
+                : new ProfileKnnByteVectorQuery(fieldName, byteQueryVector, k, filterQuery);
         }
         // If parentFilter is not null, it is a nested query. Therefore, we delegate creation of query to {@link
         // NestedKnnVectorQueryFactory}
--- a/src/main/java/org/opensearch/knn/index/query/KNNWeight.java
+++ b/src/main/java/org/opensearch/knn/index/query/KNNWeight.java
@@ -41,6 +41,12 @@ import org.opensearch.knn.indices.ModelDao;
 import org.opensearch.knn.indices.ModelMetadata;
 import org.opensearch.knn.indices.ModelUtil;
 import org.opensearch.knn.plugin.stats.KNNCounter;
+import org.opensearch.knn.profile.LongMetric;
+import org.opensearch.knn.profile.query.KNNMetrics;
+import org.opensearch.knn.profile.query.KNNQueryTimingType;
+import org.opensearch.search.profile.AbstractProfileBreakdown;
+import org.opensearch.search.profile.Timer;
+import org.opensearch.search.profile.ContextualProfileBreakdown;
 
 import java.io.IOException;
 import java.util.Arrays;
@@ -85,11 +91,13 @@ public abstract class KNNWeight extends Weight {
     protected final QuantizationService quantizationService;
     private final KnnExplanation knnExplanation;
 
-    public KNNWeight(KNNQuery query, float boost) {
-        this(query, boost, null);
+    private final ContextualProfileBreakdown profile;
+
+    public KNNWeight(KNNQuery query, float boost, ContextualProfileBreakdown profile) {
+        this(query, boost, null, profile);
     }
 
-    public KNNWeight(KNNQuery query, float boost, Weight filterWeight) {
+    public KNNWeight(KNNQuery query, float boost, Weight filterWeight, ContextualProfileBreakdown profile) {
         super(query);
         this.knnQuery = query;
         this.boost = boost;
@@ -97,6 +105,7 @@ public abstract class KNNWeight extends Weight {
         this.exactSearcher = DEFAULT_EXACT_SEARCHER;
         this.quantizationService = QuantizationService.getInstance();
         this.knnExplanation = new KnnExplanation();
+        this.profile = profile;
     }
 
     public static void initialize(ModelDao modelDao) {
@@ -297,11 +306,27 @@ public abstract class KNNWeight extends Weight {
         final String segmentName = reader.getSegmentName();
 
         StopWatch stopWatch = startStopWatch();
-        final BitSet filterBitSet = getFilteredDocsBitSet(context);
+        final BitSet filterBitSet;
+        if (profile != null) {
+            Timer filterTimer = (Timer) profile.context(context).getMetric(KNNQueryTimingType.BITSET_CREATION.toString());
+            filterTimer.start();
+            try {
+                filterBitSet = getFilteredDocsBitSet(context);
+            } finally {
+                filterTimer.stop();
+            }
+        } else {
+            filterBitSet = getFilteredDocsBitSet(context);
+        }
         stopStopWatchAndLog(stopWatch, "FilterBitSet creation", segmentName);
 
         final int maxDoc = context.reader().maxDoc();
         int cardinality = filterBitSet.cardinality();
+        if (profile != null) {
+            AbstractProfileBreakdown pb = profile.context(context);
+            LongMetric cardMetric = (LongMetric) pb.getMetric(KNNMetrics.CARDINALITY);
+            cardMetric.setValue((long) cardinality);
+        }
         // We don't need to go to JNI layer if no documents are found which satisfy the filters
         // We should give this condition a deeper look that where it should be placed. For now I feel this is a good
         // place,
@@ -318,7 +343,18 @@ public abstract class KNNWeight extends Weight {
          * This improves the recall.
          */
         if (isFilteredExactSearchPreferred(cardinality)) {
-            TopDocs result = doExactSearch(context, new BitSetIterator(filterBitSet, cardinality), cardinality, k);
+            TopDocs result;
+            if (profile != null) {
+                Timer timer = (Timer) profile.context(context).getMetric(KNNQueryTimingType.EXACT_SEARCH_AFTER_FILTER.toString());
+                timer.start();
+                try {
+                    result = doExactSearch(context, new BitSetIterator(filterBitSet, cardinality), cardinality, k);
+                } finally {
+                    timer.stop();
+                }
+            } else {
+                result = doExactSearch(context, new BitSetIterator(filterBitSet, cardinality), cardinality, k);
+            }
             return new PerLeafResult(filterWeight == null ? null : filterBitSet, result);
         }
 
@@ -329,7 +365,18 @@ public abstract class KNNWeight extends Weight {
         final BitSet annFilter = (filterWeight != null && cardinality == maxDoc) ? null : filterBitSet;
 
         StopWatch annStopWatch = startStopWatch();
-        final TopDocs topDocs = approximateSearch(context, annFilter, cardinality, k);
+        final TopDocs topDocs;
+        if (profile != null) {
+            Timer annTimer = (Timer) profile.context(context).getMetric(KNNQueryTimingType.ANN_SEARCH.toString());
+            annTimer.start();
+            try {
+                topDocs = approximateSearch(context, annFilter, cardinality, k);
+            } finally {
+                annTimer.stop();
+            }
+        } else {
+            topDocs = approximateSearch(context, annFilter, cardinality, k);
+        }
         stopStopWatchAndLog(annStopWatch, "ANN search", segmentName);
         if (knnQuery.isExplain()) {
             knnExplanation.addLeafResult(context.id(), topDocs.scoreDocs.length);
@@ -339,7 +386,18 @@ public abstract class KNNWeight extends Weight {
         // results less than K, though we have more than k filtered docs
         if (isExactSearchRequire(context, cardinality, topDocs.scoreDocs.length)) {
             final BitSetIterator docs = filterWeight != null ? new BitSetIterator(filterBitSet, cardinality) : null;
-            TopDocs result = doExactSearch(context, docs, cardinality, k);
+            TopDocs result;
+            if (profile != null) {
+                Timer timer = (Timer) profile.context(context).getMetric(KNNQueryTimingType.EXACT_SEARCH_AFTER_ANN.toString());
+                timer.start();
+                try {
+                    result = doExactSearch(context, docs, cardinality, k);
+                } finally {
+                    timer.stop();
+                }
+            } else {
+                result = doExactSearch(context, docs, cardinality, k);
+            }
             return new PerLeafResult(filterWeight == null ? null : filterBitSet, result);
         }
         return new PerLeafResult(filterWeight == null ? null : filterBitSet, topDocs);
@@ -498,7 +556,6 @@ public abstract class KNNWeight extends Weight {
      * @param knnEngine Engine type configured for the target field.
      * @param vectorDataType Vector data type configured for the target field.
      * @param quantizedVector Quantized query vector if quantization is enabled for the target field. It can be null. Quantized query vector if quantization is enabled for the target field. It can be null. Quantized query vector if quantization is enabled for the target field. It can be null. Quantized query vector if quantization is enabled for the target field. It can be null. Quantized query vector if quantization is enabled for the target field. It can be null. Quantized query vector if quantization is enabled for the target field. It can be null.
-     * @param transformedVector Transformed query vector if ADC is enabled for the target field. It is null when ADC is disabled.
      * @param modelId Model id. It can be null if the index for searching was not derived from a trained index.
      * @param filterIdsBitSet Bit set for filtering a valid document for collecting.
      * @param cardinality Cardinality of filtering bit set. It will be the total number of documents if no filtering presents.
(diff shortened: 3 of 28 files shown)


src/main/java/org/opensearch/knn/plugin/KNNPlugin.java
```diff
@@ -185,6 +188,30 @@ public class KNNPlugin extends Plugin
         });
     }
 
+    @Override
+    public Optional<ProfileMetricsProvider> getQueryProfileMetricsProvider() {
+        return Optional.of(new ProfileMetricsProvider() {
+            @Override
+            public Collection<Supplier<ProfileMetric>> getProfileMetrics(SearchContext searchContext) {
+                KNNMetrics.setProfilers(searchContext.getProfilers());
```
shatejas: You are creating a static variable inside this for profilers, This can cause issues where you won't be returning the right set of profilers and breakdowns when there are multiple threads accessing this. For each request there is a new search context that is created. This means there is a new instance of Profilers object as well and so is the entire profile tree. With multiple threads profiling together it will cause race conditions and will result in wrong results. 

While the use case for profiling is debugging and visibility, and is generally used for one request, having that underlying assumption is risky as we don't know how user will use it. We should not have profilers in static context.

src/main/java/org/opensearch/knn/profile/LongMetric.java
```diff
@@ -0,0 +1,29 @@
...
+
+package org.opensearch.knn.profile;
+
+import lombok.Getter;
+import lombok.Setter;
+import org.opensearch.search.profile.ProfileMetric;
+
+import java.util.Map;
+
+public class LongMetric extends ProfileMetric {
```
shatejas, line 17: Should this be in core rather than plugin?

src/main/java/org/opensearch/knn/profile/LongMetric.java
```diff
@@ -0,0 +1,29 @@
...
+
+    @Getter
+    @Setter
```
shatejas, line 20: Remove the setter maybe, we would want to make this immutable and user should only be able to set it once

src/main/java/org/opensearch/knn/profile/query/LuceneEngineKnnTimingType.java
```diff
@@ -0,0 +1,23 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.knn.profile.query;
+
+import java.util.Locale;
+
+public enum LuceneEngineKnnTimingType {
```
shatejas: I don't think we need separate enums for each. Just one enum with all relevant information should work for all KNN queries

src/main/java/org/opensearch/knn/profile/query/LuceneEngineKnnTimingType.java
```diff
@@ -0,0 +1,23 @@
...
+    EXACT_SEARCH,
+    ANN_SEARCH,
+    EXPAND_NESTED_ANN,
+    BITSET_CREATION,
+    EXPAND_NESTED_EXACT,
+    INTERNAL_EXACT,
+    RESCORE;
```
shatejas: How are these 3 different from exact? why do we need segregation. The Timer  also holds the count information right? that should be enough indication for further deep dive. We don't want to overwhelm the user with internal details and segregation. same for EXPAND_NESTED_ANN and ANN

src/test/java/org/opensearch/knn/index/OpenSearchIT.java
```diff
@@ -35,18 +35,18 @@
...
+import org.opensearch.search.profile.query.QueryTimingType;
 
 import java.io.IOException;
 import java.net.URL;
-import java.util.Arrays;
-import java.util.List;
-import java.util.Locale;
-import java.util.Map;
-import java.util.TreeMap;
+import java.util.*;
```
shatejas: audit and revert all * imports please

shatejas (commented):
Paused the review for now till we address the static profiler concern, that will further dictate how we put metrics. I skimmed through it and I would want to see if there is a way to avoid creating a new query just for profiling as it has additional maintenance with each query being added. Moreover its wrapping it even when profiling is not needed which is completely unnecessary

src/main/java/org/opensearch/knn/profile/query/LuceneEngineKnnTimingType.java
neuenfeldttj: Yes, I can just simplify it to Exact and ANN search. I wasn't sure if a user would like to specifically see the segregation.

src/main/java/org/opensearch/knn/plugin/KNNPlugin.java
neuenfeldttj: I agree getting rid of static variables. Working on the core PR to allow for this.

src/main/java/org/opensearch/knn/profile/LongMetric.java
neuenfeldttj: I agree that this could be a final variable, but the problem arises in the execution because a `LongMetric` instance  is created by core and in `KNNWeight` is where value gets assigned.

src/main/java/org/opensearch/knn/profile/query/LuceneEngineKnnTimingType.java
neuenfeldttj: Simplified timers to just exact, ann, and bitset_creation.

src/main/java/org/opensearch/knn/index/query/RescoreKNNVectorQuery.java
```diff
@@ -86,8 +91,21 @@ public Weight createWeight(IndexSearcher searcher, ScoreMode scoreMode, float bo
...
         List<LeafReaderContext> leafReaderContexts = indexSearcher.getIndexReader().leaves();
         List<Callable<TopDocs>> rescoreTasks = new ArrayList<>(leafReaderContexts.size());
+        Profilers profilers = KNNMetrics.getProfilers();
         for (LeafReaderContext leafReaderContext : leafReaderContexts) {
-            rescoreTasks.add(() -> searchLeaf(exactSearcher, weight, k, leafReaderContext));
+            rescoreTasks.add(() -> {
+                if (profilers != null) {
+                    AbstractProfileBreakdown profile = profilers.getCurrentQueryProfiler().getTopBreakdown().context(leafReaderContext);
+                    Timer timer = profile.getTimer(LuceneEngineKnnTimingType.RESCORE);
+                    timer.start();
```
jmazanec15: should this be per segment or for all segments?

jmazanec15 (commented):
This is awesome and will be super helpful - thanks @neuenfeldttj  and @shatejas. I think time taken to load the graph might be a good one as well, although that might already be in the stats.

src/main/java/org/opensearch/knn/index/query/RescoreKNNVectorQuery.java
neuenfeldttj: The profiler is contextual -- it needs a leafReaderContext to figure out which breakdown to get. So I believe it's currently doing per segment.

neuenfeldttj:
@shatejas In regard to your comment about non-profiling queries still adding overhead with the wrapper classes, I agree that it would be nice to not add overhead but I looked at `KNNQueryBuilder` and since this happens before the query execution, there's no current way to determine if profiling is enabled.

src/main/java/org/opensearch/knn/plugin/KNNPlugin.java
```diff
@@ -185,6 +197,29 @@ public class KNNPlugin extends Plugin
...
+            ProfileDiversifyingChildrenByteKnnVectorQuery.class
+        );
+        return Optional.of((searchContext, query) -> {
+            if (knnQueryClasses.contains(query.getClass())) {
+                return KNNMetrics.getKNNQueryMetrics();
+            } else if (query instanceof NativeEngineKnnVectorQuery) {
+                return KNNMetrics.getNativeMetrics();
+            }
+            return List.of();
+        });
```
shatejas: These queries are mostly derived from certain queries so you want instance of checks instead of set contains. For instance you can cover a lot of these queries by simply checking `instance of AbstractKnnVectorQuery`. It will be more maintainable that way

src/main/java/org/opensearch/knn/index/query/lucene/ProfileKnnByteVectorQuery.java
```diff
@@ -0,0 +1,68 @@
...
+import org.apache.lucene.search.knn.KnnCollectorManager;
+import org.apache.lucene.util.Bits;
+import org.opensearch.knn.profile.query.KNNQueryTimingType;
+import org.opensearch.search.internal.ContextIndexSearcher;
+import org.opensearch.search.profile.Timer;
+import org.opensearch.search.profile.query.QueryProfiler;
+
+import java.io.IOException;
+
+public class ProfileKnnByteVectorQuery extends KnnByteVectorQuery {
```
shatejas, line 23: I am not sure if this is the right way to go about it. maintaining a new class for each query is a lot. One solution is to make a case for AbstractKNNVectorQuery to be public. That way it opens up a possibility to use composition to have profiler timings instead. 

We should try that in parallel, but for time being we should isolate it to only when the profile is needed. KNNQueryFactory should take in a flag or profiler to see if profile is needed and give the right query

src/main/java/org/opensearch/knn/index/query/RescoreKNNVectorQuery.java
```diff
@@ -86,8 +91,21 @@ public Weight createWeight(IndexSearcher searcher, ScoreMode scoreMode, float bo
     private TopDocs[] doRescore(final IndexSearcher indexSearcher, Weight weight) throws IOException {
         List<LeafReaderContext> leafReaderContexts = indexSearcher.getIndexReader().leaves();
         List<Callable<TopDocs>> rescoreTasks = new ArrayList<>(leafReaderContexts.size());
+        QueryProfiler profiler = ((ContextIndexSearcher) indexSearcher).getProfiler();
         for (LeafReaderContext leafReaderContext : leafReaderContexts) {
-            rescoreTasks.add(() -> searchLeaf(exactSearcher, weight, k, leafReaderContext));
+            rescoreTasks.add(() -> {
+                if (profiler != null) {
+                    AbstractProfileBreakdown profile = profiler.getProfileBreakdown(this).context(leafReaderContext);
+                    Timer timer = profile.getTimer(KNNQueryTimingType.EXACT_SEARCH);
```
shatejas: Why do you need this here, Should exact searcher be taking care of this? that way every query passing through exact searcher will have the profile

src/main/java/org/opensearch/knn/index/query/KNNQuery.java
```diff
@@ -175,6 +178,11 @@ public KNNQuery filterQuery(Query filterQuery) {
      */
     @Override
     public Weight createWeight(IndexSearcher searcher, ScoreMode scoreMode, float boost) throws IOException {
+        QueryProfiler profiler = ((ContextIndexSearcher) searcher).getProfiler();
+        ContextualProfileBreakdown profile = null;
+        if (profiler != null) {
+            profile = profiler.getProfileBreakdown(this);
+        }
         StopWatch stopWatch = null;
```
shatejas: How are we profiling filter?

shatejas:
> @shatejas In regard to your comment about non-profiling queries still adding overhead with the wrapper classes, I agree that it would be nice to not add overhead but I looked at `KNNQueryBuilder` and since this happens before the query execution, there's no current way to determine if profiling is enabled.

doToQuery takes in a `QueryShardContext` which has an IndexSearcher. We can check if that indexSearcher has profiler set. If it is, we can pass in the information to KNNFactory to generate the right profiling query. the idea is we are eliminating points of failure with making sure profiling is never done when not needed. Let me know if it works out

neuenfeldttj:
> > @shatejas In regard to your comment about non-profiling queries still adding overhead with the wrapper classes, I agree that it would be nice to not add overhead but I looked at `KNNQueryBuilder` and since this happens before the query execution, there's no current way to determine if profiling is enabled.
> 
> doToQuery takes in a `QueryShardContext` which has an IndexSearcher. We can check if that indexSearcher has profiler set. If it is, we can pass in the information to KNNFactory to generate the right profiling query. the idea is we are eliminating points of failure with making sure profiling is never done when not needed. Let me know if it works out

Yes I already ran a test and Profilers isn't set at that point since it's before query execution.

src/main/java/org/opensearch/knn/index/query/KNNQuery.java
neuenfeldttj: We are not, this is used to just pass into the KNNWeights. We could add a timer around filterWeight creation. Currently we just time the filterBitSet creation.

src/main/java/org/opensearch/knn/index/query/RescoreKNNVectorQuery.java
neuenfeldttj: Good point. Will fix

shatejas:
> Yes I already ran a test and Profilers isn't set at that point since it's before query execution.

I see, seems like its just a matter of ordering of if blocks and there isn't any context missing to create profilers before the `toQuery()` call. 

Can you check if its possible to move [if block on line 1556](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/search/SearchService.java#L1556) above the [if block of line 1384](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/search/SearchService.java#L1384)?

Since none of the existing code is already using profilers it also seems safe to do so.

src/main/java/org/opensearch/knn/index/query/KNNQuery.java
shatejas: Can you try something similar https://github.com/shatejas/k-NN/commit/66d7dae8a8d62e11081d1ff18e3e020e5f7c093e and see if it works for the changes you made

neuenfeldttj:
I haven't tested it but I don't think it will be since [line 1557](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/search/SearchService.java#L1557) depends on [line 1555](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/search/SearchService.java#L1555). And line 1555 needs the previous source information to determine whether concurrency should be enabled.

shatejas:
> I haven't tested it but I don't think it will be since [line 1557](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/search/SearchService.java#L1557) depends on [line 1555](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/search/SearchService.java#L1555). And line 1555 needs the previous source information to determine whether concurrency should be enabled.

Makes sense. A naive way would be to create a member variable in QueryShardContext and pass in the source.profile() value inside it

src/main/java/org/opensearch/knn/index/query/KNNQuery.java
neuenfeldttj: After some testing, I realized we don't need to profile the filter on the KNN side of things because core takes care of that for us. Every single query gets sent into `ContextIndexSearcher` where the profiling of the query happens. If we tried profiling here then it would duplicate the query profiling.

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -297,11 +306,27 @@ public PerLeafResult searchLeaf(LeafReaderContext context, int k) throws IOExcep
...
+            Timer filterTimer = profile.context(context).getTimer(KNNQueryTimingType.BITSET_CREATION);
+            filterTimer.start();
+            try {
+                filterBitSet = getFilteredDocsBitSet(context);
+            } finally {
+                filterTimer.stop();
+            }
+        } else {
+            filterBitSet = getFilteredDocsBitSet(context);
+        }
```
shatejas: Consider wrapping this inside a method to hide the verbosity of this

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -297,11 +306,27 @@ public PerLeafResult searchLeaf(LeafReaderContext context, int k) throws IOExcep
...
         stopStopWatchAndLog(stopWatch, "FilterBitSet creation", segmentName);
 
         final int maxDoc = context.reader().maxDoc();
         int cardinality = filterBitSet.cardinality();
+        if (profile != null) {
+            AbstractProfileBreakdown pb = profile.context(context);
+            LongMetric cardMetric = (LongMetric) pb.getMetric(KNNMetrics.CARDINALITY);
+            cardMetric.setValue((long) cardinality);
+        }
```
shatejas: Consider combining this with the if else for filterBitSet

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -329,7 +354,18 @@ public PerLeafResult searchLeaf(LeafReaderContext context, int k) throws IOExcep
...
+            Timer annTimer = (Timer) profile.context(context).getMetric(KNNQueryTimingType.ANN_SEARCH.toString());
+            annTimer.start();
+            try {
+                topDocs = approximateSearch(context, annFilter, cardinality, k);
+            } finally {
+                annTimer.stop();
+            }
+        } else {
+            topDocs = approximateSearch(context, annFilter, cardinality, k);
+        }
```
shatejas: Same here, consider wrapping it in a method

src/main/java/org/opensearch/knn/index/query/KNNQuery.java
```diff
@@ -175,6 +178,11 @@ public KNNQuery filterQuery(Query filterQuery) {
      */
     @Override
     public Weight createWeight(IndexSearcher searcher, ScoreMode scoreMode, float boost) throws IOException {
+        QueryProfiler profiler = ((ContextIndexSearcher) searcher).getProfiler();
+        ContextualProfileBreakdown profile = null;
+        if (profiler != null) {
+            profile = profiler.getProfileBreakdown(this);
+        }
```
shatejas: Don't think we should be skipping instance of checks for indexSearcher. create a method to get the profiler with instance of checks

```
    private QueryProfiler getProfiler(final IndexSearcher searcher) {
        if (searcher instanceof ContextIndexSearcher) {
            ContextIndexSearcher contextIndexSearcher = (ContextIndexSearcher) searcher;
            if (contextIndexSearcher.getProfiler() != null) {
                return contextIndexSearcher.getProfiler();
            }
        }
        return null;
    }
```

shatejas (commented):
For all native engine queries. Its introducing a lot of verbosity and branching. This is not ideal since this interferes with non-profile code path. To start with please audit and make sure the branching is wrapped inside a method as suggested. Also consider having utility methods where required

The ideal approach needs to be that we need to create an interface and use composition to create ProfileKNNWeight. This might need some thought and can take a while

src/main/java/org/opensearch/knn/index/query/KNNQuery.java
shatejas: In my understanding core will only profile outer queries, For instance it will add timing info for createWeight and scorer of KNNQuery but it won't do it for filter query inside the KNNQuery until its wrapped with ProfileQuery/ProfileWeight. 

So KNNQuery createWeight timing is a combination of filterWeight time. We need the breakdown of filter we need to wrap it with profile class. So in my understanding its not duplicate, it will be a breakdown. Maybe I am missing something. can you validate? Maybe upload a response json with the PR, it might help make things clear

neuenfeldttj:
Some notes:
- I added a sample output to the description up above. 
- Right now, there's not a good way to wrap Lucene queries unless `AbstractKnnVectorQuery` became public. 
- I will make an issue in core to change where `Profilers` is created so it's more optimal for plugins to know earlier if profiling is enabled.

src/main/java/org/opensearch/knn/index/query/RescoreKNNVectorQuery.java
```diff
@@ -86,13 +90,28 @@ public Weight createWeight(IndexSearcher searcher, ScoreMode scoreMode, float bo
     private TopDocs[] doRescore(final IndexSearcher indexSearcher, Weight weight) throws IOException {
         List<LeafReaderContext> leafReaderContexts = indexSearcher.getIndexReader().leaves();
         List<Callable<TopDocs>> rescoreTasks = new ArrayList<>(leafReaderContexts.size());
+        QueryProfiler profiler = KNNProfileUtil.getProfiler(indexSearcher);
```
markwu-sde, line 93: Should this have a null check here?

src/main/java/org/opensearch/knn/index/query/KNNQueryFactory.java
```diff
@@ -179,8 +179,8 @@ private static Query getKnnVectorQuery(
         if (parentFilter == null) {
             assert expandNested == false : "expandNested is allowed to be true only for nested fields.";
             return vectorDataType == VectorDataType.FLOAT
-                ? new KnnFloatVectorQuery(fieldName, floatQueryVector, k, filterQuery)
-                : new KnnByteVectorQuery(fieldName, byteQueryVector, k, filterQuery);
+                ? new ProfileKnnFloatVectorQuery(fieldName, floatQueryVector, k, filterQuery)
```
markwu-sde, line 182: This will be created regardless if the profiler is enabled or not?

src/main/java/org/opensearch/knn/index/query/RescoreKNNVectorQuery.java
```diff
@@ -86,13 +90,28 @@ public Weight createWeight(IndexSearcher searcher, ScoreMode scoreMode, float bo
...
+                    k,
+                    leafReaderContext,
+                    (ContextualProfileBreakdown) profiler.getProfileBreakdown(this)
+                )
+            );
         }
         return indexSearcher.getTaskExecutor().invokeAll(rescoreTasks).toArray(TopDocs[]::new);
     }
 
-    private TopDocs searchLeaf(ExactSearcher searcher, Weight weight, int k, LeafReaderContext leafReaderContext) throws IOException {
```
markwu-sde, line 95: We can probably keep this method here and add a override of `searchLeaf` that takes in a profiler. The original method would look something like this 

```
private TopDocs searchLeaf(ExactSearcher searcher, Weight weight, int k, LeafReaderContext leafReaderContext) throws IOException {
 return searchLeaf(searcher, weight, k, leafReaderContext, null)
}
```

Or if we don't want to pass in something null we can pass in a empty object.

src/main/java/org/opensearch/knn/index/query/common/QueryUtils.java
```diff
@@ -101,11 +104,18 @@ private int[] findSegmentStarts(final IndexReader reader, final int[] docs) {
     public List<Map<Integer, Float>> doSearch(
         final IndexSearcher indexSearcher,
         final List<LeafReaderContext> leafReaderContexts,
-        final Weight weight
+        final Weight weight,
+        ContextualProfileBreakdown profile
```
markwu-sde, line 108: Same thing here. Do we need to override these search paths to maintain the same flow for profiler not enabled?

neuenfeldttj: #2797 won't have this change but let me take a look for the feature branch

src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java
```diff
@@ -143,41 +155,59 @@ private List<PerLeafResult> retrieveAll(
         List<Callable<PerLeafResult>> nestedQueryTasks = new ArrayList<>(leafReaderContexts.size());
         for (int i = 0; i < perLeafResults.size(); i++) {
             LeafReaderContext leafReaderContext = leafReaderContexts.get(i);
+            QueryProfiler profiler = ((ContextIndexSearcher) indexSearcher).getProfiler();
             int finalI = i;
             nestedQueryTasks.add(() -> {
-                PerLeafResult perLeafResult = perLeafResults.get(finalI);
-                if (perLeafResult.getResult().scoreDocs.length == 0) {
-                    return perLeafResult;
+                if (profiler != null) {
```
markwu-sde, line 164: NIT: we don't have to check for the profiler being null here since we need to retrieve the PerLeafResult regardless.

```
                    PerLeafResult result = retrieveSingle(leafReaderContext, knnWeight, perLeafResults, useQuantizedVectors, finalI);
                    
                    if (profiler != null) {
                    
                    //Add the result to the metric
                    
                    }
                    
                    return result;
```

src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java
```diff
@@ -143,41 +155,59 @@ private List<PerLeafResult> retrieveAll(
...
-                    .build();
-                TopDocs rescoreResult = knnWeight.exactSearch(leafReaderContext, exactSearcherContext);
-                return new PerLeafResult(perLeafResult.getFilterBits(), rescoreResult);
+                return retrieveSingle(leafReaderContext, knnWeight, perLeafResults, useQuantizedVectors, finalI);
             });
         }
         return indexSearcher.getTaskExecutor().invokeAll(nestedQueryTasks);
     }
 
+    private PerLeafResult retrieveSingle(
```
markwu-sde, line 179: NIT: Javadoc for this and could be beneficial to make the method name more descriptive.

src/main/java/org/opensearch/knn/profile/LongMetric.java
markwu-sde: +1 if this metric is going to be used across plugins.

If this is a KNN specific metric we can provide it here.

markwu-sde (commented):
@neuenfeldttj I noticed that we are modifying the critical path here with the profiler implementations and let's make sure each time a profiler is used we have the right checks. Let's do extensive unit/IT tests to make sure that each case without a profiler works as expected (Exact, rescore, etc.)

Issue #1079, opened 2023-08-31
[FEATURE] Providing better experience for doing Exact Search without Script Score Query


navneet1v:
**Is your feature request related to a problem?**
Currently if a user has to do exact search, they need to use script query to the exact search. Ref: https://opensearch.org/docs/latest/search-plugins/knn/knn-score-script/ This is not intuitive and uses an extra hop, which is script compilation. Given that scripts are notorious and posses security concerns running in OpenSearch in multi tenant environments.

**What solution would you like?**
Solution is to provide the exact search feature in the k-nn query clause itself. Given that k-NN vectors are stored as doc values in the segment, during query execution code can easily iterate over these doc values of the segment to do the exact search. 

**What alternatives have you considered?**
NA

**Do you have any additional context?**
NA

navneet1v:
@jmazanec15 this is inline to the thought I had when creating this issue. Do we see any other alternative here? 

I see one, which is creating a new query clause but thats not a good option. So I never added it. So lets stick to this one unless some one has any better ideas. 

@vamshin @heemin32 any thoughts you have here.

navneet1v:
On thinking over this more I think, what we can do is we can start taking another option which is used for text fields too. We can use this field `index`(similar to `store`) in the field with value as false to indicate that whether to index vectors or not. Indexing vectors here means create KNN data structures or not. 

Example:
```
PUT /test-index
{
  "settings": {
    "index": {
      "knn": true
    }
  },
  "mappings": {
    "properties": {
      "my_vector1": {
        "type": "knn_vector",
        "dimension": 3,
        "index":false
      }
    }
  }
}
```

We can then pass this set the value as an attribute in VectorField, and read this value in PerFieldCodec to hit a new plain codec that just stores the vectors.

This approach will work for all the engines.

jmazanec15:
I think that makes sense. So this would fall back to brute force knn search?

navneet1v:
> I think that makes sense. So this would fall back to brute force knn search?

Yes that is correct.

jmazanec15:
@navneet1v do you think instead we can do "method": false? Right now, method specifies the structure for ANN search. So, if we set to false, it would make sense that we do not want to do ANN search.

navneet1v:
> @navneet1v do you think instead we can do "method": false? Right now, method specifies the structure for ANN search. So, if we set to false, it would make sense that we do not want to do ANN search.

[...]

PR #2787, opened 2025-07-10
[DRAFT] Adding support for exact search in KNN query clause

kaushikau:
### Description
Currently, if a user wants to perform exact search on a field with `index.knn` set to true, a score script query must be used. As mentioned in #2368 and #1079, score scripting is not intuitive and could pose security concerns. The addition of a query parameter `exactSearchSpaceType` gives users the option to override with exact search in the KNN query clause itself, without the use of a score script.

One benefit of score script was that a user could define a different space type from the one specified in the index, and this parameter will maintain that feature.

Example of query with `exactSearchSpaceType`:
```
{
    "size": 3, 
    "query": {
        "knn": {
            "my_vector1": {
                "vector": [2.1, 6.2],
                "k": 3,
                "exact_search_space_type": "l2"
             }
        }
    }
}
```

Note: Score scripting also allows for pre-filtering. That is not yet supported in this implementation, and needs to be added in the future.

### Related Issues
Resolves Problem 2 of #2368

### Check List


src/main/java/org/opensearch/knn/index/query/ExactSearcher.java
```diff
@@ -175,11 +178,35 @@ private KNNIterator getKNNIterator(LeafReaderContext leafReaderContext, ExactSea
             );
             return null;
         }
-        final VectorDataType vectorDataType = FieldInfoExtractor.extractVectorDataType(fieldInfo);
-        final SpaceType spaceType = FieldInfoExtractor.getSpaceType(modelDao, fieldInfo);
+        final VectorDataType vectorDataType;
+        if (exactSearcherContext.isLuceneExactSearch != null
```
rchitale7: do you need this check for null?

src/main/java/org/opensearch/knn/index/query/ExactSearcher.java
```diff
@@ -175,11 +178,35 @@ private KNNIterator getKNNIterator(LeafReaderContext leafReaderContext, ExactSea
...
+            resolvedSpaceType = SpaceType.getSpace(exactSearcherContext.getExactSearchSpaceType());
+        }
+        final SpaceType spaceType = resolvedSpaceType;
         boolean isNestedRequired = exactSearcherContext.getParentsFilter() != null;
 
         if (VectorDataType.BINARY == vectorDataType) {
+            if (exactSearcherContext.isLuceneExactSearch != null && exactSearcherContext.isLuceneExactSearch) {
+                KNNVectorValues<byte[]> vectorValues;
+                ByteVectorValues luceneVectorValues = leafReaderContext.reader().getByteVectorValues(exactSearcherContext.getField());
+                if (luceneVectorValues != null) {
```
rchitale7: why do we do this null check here? What happens if `luceneVectorValues` is null?

src/main/java/org/opensearch/knn/index/query/ExactSearcher.java
```diff
@@ -241,6 +281,21 @@ private KNNIterator getKNNIterator(LeafReaderContext leafReaderContext, ExactSea
             }
         }
 
+        if (exactSearcherContext.isLuceneExactSearch != null && exactSearcherContext.isLuceneExactSearch) {
```
rchitale7: This`if` block seems like it can be made generic, to support `BINARY` `BYTE` and `FLOAT`.

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
```diff
@@ -286,6 +298,15 @@ private void validate() {
                     );
                 }
             }
+
+            if (exactSearchSpaceType != null) {
+                List<String> allowedSpaceTypes = Arrays.asList("hamming", "hammingbit", "l1", "l2", "linf", "innerproduct", "cosinesimil");
```
rchitale7: Can we use the same space type validation that ann search uses? This seems fragile

src/main/java/org/opensearch/knn/index/query/lucene/LuceneEngineKnnVectorQuery.java
```diff
@@ -41,6 +51,77 @@ public Query rewrite(IndexSearcher indexSearcher) {
     @Override
     public Weight createWeight(IndexSearcher searcher, ScoreMode scoreMode, float boost) throws IOException {
         Query rewrittenQuery = luceneQuery.rewrite(searcher);
+        if (luceneQuery instanceof InternalKnnFloatVectorQuery || luceneQuery instanceof InternalKnnByteVectorQuery) {
+            return new Weight(this) {
```
rchitale7: Not sure I understand why we're returning a `Weight` object that looks like this.

src/main/java/org/opensearch/knn/index/query/ExactSearcher.java
kaushikau: removed

src/main/java/org/opensearch/knn/index/query/ExactSearcher.java
kaushikau: removed

src/main/java/org/opensearch/knn/index/query/lucene/LuceneEngineKnnVectorQuery.java
kaushikau: Refactored with new LuceneKNNWeight class instead

src/main/java/org/opensearch/knn/index/query/KNNQuery.java
```diff
@@ -98,6 +99,30 @@ public KNNQuery(
         this(field, null, byteQueryVector, k, indexName, filterQuery, parentsFilter, vectorDataType, rescoreContext);
     }
 
+    public KNNQuery(
```
shatejas, line 102: I think its time we create a builder to solve the problem of cascading constructors

src/main/java/org/opensearch/knn/index/query/lucene/LuceneEngineKnnVectorQuery.java
```diff
@@ -26,6 +27,8 @@
 public class LuceneEngineKnnVectorQuery extends Query {
     @Getter
     private final Query luceneQuery;
+    @Getter
+    private final String exactSearchSpaceType;
```
shatejas, line 31: I am not sure of this is the right way to determine if the query is for exact search. shouldn't this be a combination of mapping check and this?

shatejas (commented):
Instead of reusing the same query have we considered having a separate query. 

```json
{
  "size": 2,
  "query": {
    "exact_vector_search": { // OR exact_knn OR knn_exact
      "my_vector2": {
        "vector": [2, 3, 5, 6],
        "k": 2,
         "space_type": l2,
         "filter" : {
             bool { //filter query}
         }
      }
    }
  }
}
```

This gives a clear unambiguous way for users to be able to have exact search and might help simplify branching

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
```diff
@@ -440,6 +461,9 @@ protected Query doToQuery(QueryShardContext context) {
         KNNEngine knnEngine = queryConfigFromMapping.getKnnEngine();
         MethodComponentContext methodComponentContext = queryConfigFromMapping.getMethodComponentContext();
         SpaceType spaceType = queryConfigFromMapping.getSpaceType();
+        if (exactSearchSpaceType != null) {
+            spaceType = SpaceType.getSpace(exactSearchSpaceType);
+        }
```
shatejas, line 466: There should be a validation here
if (exactSearchSpaceType == null && exact search mapping is true) then throw

kaushikau:
> Instead of reusing the same query have we considered having a separate query.
> 
> ```json
> {
>   "size": 2,
>   "query": {
>     "exact_vector_search": { // OR exact_knn OR knn_exact
>       "my_vector2": {
>         "vector": [2, 3, 5, 6],
>         "k": 2,
>          "space_type": l2,
>          "filter" : {
>              bool { //filter query}
>          }
>       }
>     }
>   }
> }
> ```
> 
> This gives a clear unambiguous way for users to be able to have exact search and might help simplify branching

This alternative was mentioned in the original issue https://github.com/opensearch-project/k-NN/issues/1079#issuecomment-1701546935, but was said to not be a good idea, so I went with the design of having the exact search feature in the knn clause

src/main/java/org/opensearch/knn/index/query/KNNQuery.java
kaushikau: Would you suggest doing that in a separate PR or in this one?

kaushikau:
> > > Instead of reusing the same query have we considered having a separate query.
> > > ```json
> > > {
> > >   "size": 2,
> > >   "query": {
> > >     "exact_vector_search": { // OR exact_knn OR knn_exact
> > >       "my_vector2": {
> > >         "vector": [2, 3, 5, 6],
> > >         "k": 2,
> > >          "space_type": l2,
> > >          "filter" : {
> > >              bool { //filter query}
> > >          }
> > >       }
> > >     }
> > >   }
> > > }
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > This gives a clear unambiguous way for users to be able to have exact search and might help simplify branching
> > 
> > 
> > This alternative was mentioned in the original issue [#1079 (comment)](https://github.com/opensearch-project/k-NN/issues/1079#issuecomment-1701546935), but was said to not be a good idea, so I went with the design of having the exact search feature in the knn clause
> 
> Lets revisit it because there doesn't seem to be any reasons given as to why its not a good idea. A new query simplifies code and segregates the use cases giving users an unambiguous way to do exact search. Do list out the pros and cons either in the PR or issues so we can discuss

Sounds good. Here are some pros and cons for having a separate query for exact search:
Pros:
- Separates use cases → less ambiguous to do exact search for users
- Don’t need to add different support for Faiss and Lucene → new query can be engine-less
- Simplifies branching in code/cleaner code

Cons:
- Increased API for users
- If a user wants to change to ANN from exact search or vice versa, would need to use a completely different query
- Potential inconsistency of interface → radial search uses parameter for switching, but still uses knn query

Issue #2795, opened 2025-07-17
[RFC] Deduplicating Vectors from Flat Index Storage to Optimize Space


sobhu17:
# Deduplicating Flat Vectors

## **1. Abstract**

This document presents two solutions for deduplicating full-precision vectors in a FAISS index within OpenSearch. Currently, flat vectors are stored in both `.vec` and `.faiss` files, resulting in redundant data and increased resource usage.
With this optimization, the index size was reduced from 45 GB to 16 GB for FP16, from 59 GB to 32 GB for FP32 (FYI, Lucene has 30GB), 64% improvement. This deduplication also led to a 22% reduction in indexing throttle time (from 350 min to 272 min) and a 66% drop in total merge byte size (from 79.61 GB to 27.02 GB). Additionally, merge throttling time decreased by 66%, going from 33 min to 11 min. Memory efficiency also improved significantly, with JVM (off-heap) buffer memory dropping from 59.95 GB to 32.43 GB, 45% reduction.

The document starts with the problem statement, outlines benchmarking results, and provides a focused comparison of the two solutions, highlighting their implementation details, advantages, and trade-offs.

## **2. Scope**

The proposed solutions apply specifically to **FAISS HNSW indexing** and are **not applicable to IVF**.
This optimization targets cases where vectors are of type FP32, FP16, Byte and Pure Binary(case where user ingest already quantized bytes). 

## **3. Background**

OpenSearch’s k-NN plugin uses **FAISS with HNSW** for ANN(Approximate Nearest Neighbor Search) search. The figure below shows how vectors are currently indexed and stored in the system.

### **3.1. How We Are Storing Vectors**

When users send documents to OpenSearch, they are first temporarily buffered (Lucene’s [DocumentsWriterPerThread](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/index/DocumentsWriterPerThread.java)) then flushed into storage via [IndexOutput](https://lucene.apache.org/core/10_0_0/core/org/apache/lucene/store/IndexOutput.html).
[...]

PR #2798, opened 2025-07-18
Dedup - Add VectorReader to expose FloatVectorValues for native FAISS C++ layer.

sobhu17:
**Issue with the current vector storage:**

Currently we store vectors in two files (.vec and .faiss). The two vector sets stored in each file may differ where Faiss index file (.faiss) might contains quantized vectors for fast approximate search, while the Lucene .vec file  stores the full-precision vectors used during exact search.
For FP32, Byte, and Pure Binary cases, maintaining two copies of the same vectors in both .vec and .faiss files is redundant. Both approximate and exact search operations can run on a single set of vectors, making it possible to eliminate this duplication safely. 
Issue Link: https://github.com/opensearch-project/k-NN/issues/2795

**PR Description:**

This PR adds the Java-side support for exposing float vector values to the native FAISS C++ layer through JNI. It introduces the VectorReader class and updates key index loading flows to pass FloatVectorValues into the native layer. The complete implementation will be delivered in three separate PRs.

* PR1: Introduces Java-side support to expose FloatVectorValues via VectorReader, enabling FAISS to access flat vectors through JNI.
* PR2: Adds native integration to read float vectors from the JVM using the new VectorReader, removing the need to write flat vectors to .faiss.
* PR3: It will include bumping up FAISS version.

This is the first of three PRs.

**Key Changes**

* Added VectorReader.java to stream vectors from Lucene’s FloatVectorValues.
* Updated IndexInputWithBuffer to include FloatVectorValues and expose getFullPrecisionVectors().
* Modified NativeMemoryEntryContext and KNNIndexShard to pass FloatVectorValues into memory loading logic.
* This PR introduces Java-side hooks only.
* The impact to indexing — specifically, removing flat vector data from the .faiss file and relying solely on .vec , will take effect in PR2.


src/main/java/org/opensearch/knn/index/store/VectorReader.java
```diff
@@ -0,0 +1,38 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.knn.index.store;
+
+import org.apache.lucene.index.FloatVectorValues;
+import org.apache.lucene.index.KnnVectorValues.DocIndexIterator;
+import java.io.IOException;
+
+public class VectorReader {
```
shatejas, line 14: Why not use KNNVectorValues?

sobhu17: Your point about using `KnnVectorValues` is absolutely valid.
However, I chose to use FloatVectorValues because handling Java-side method complexity directly from C++ via JNI can become quite cumbersome, especially in terms of exception handling and type safety. To keep things clean and manageable, I decided to implement the logic on the Java side and expose a simple next() method that C++ can call.
I am planning to introduce separate methods like nextBinary(), and C++ will call the appropriate method based on the vector type that will make use of `KnnVectorValues`.

shatejas: From this PR, there doesn't seem to be a need for creating another wrapper. I am not sure of additional methods that are needed but it will be hard to suggest how to fit in with KNNVectorValues without more context. 

KNNVectorValues abstraction is intended to hide the source of vectors and its implementation, I would be hesitant to break that pattern unless there is an absolute need

> I am planning to introduce separate methods like nextBinary(), and C++ will call the appropriate method based on the vector type that will make use of KnnVectorValues

This seems to be another wrapper on KNNVectorValues. It should ideally be other way round where KNNVectorValues hides the details

sobhu17: Now I got your point, I will make sure that I will use `KnnVectorValues` to read vector values in my implementation.

sobhu17:
I have pushed all the changes for supporting vector deduplication. Some of the unit tests are currently failing after integrating the C++ side, which is expected, we're removing the flat vector section from the `.faiss` file, and a few tests assume vectors are still present there. I'll work on resolving the failing test cases. In the meantime, please start reviewing the PR.

navneet1v:
@0ctopus13prime can you please take a look.

Issue #2642, opened 2025-04-02
[RFC] Preload Graph For Prevent slow knn queries when nodes restart


luyuncheng:
# Introduction
We see when node restart, or shard rebalanced, or a large segment merged, it would cause query latency spike。

# Proposal
i see there is FEATURE request at #711 also have introduced  at #1946 so we can easily do `preload` jobs at here when doing `new KNN80DocValuesProducer` which occurs at open segments readers.

i also like https://github.com/opensearch-project/k-NN/issues/711#issuecomment-1674056746

> So, in this constructor, we can probably add the logic to preload graphs. We can make it configurable via index setting by passing index setting here: 

we can create settings maybe `index.knn.preload` likes `index.store.preload` settings to preload knn graph

jmazanec15:
Hi @luyuncheng  - I like the idea. The difficulty for the readers/producers is getting access to the setting if they are created via the no arg constructor with SPI. Do you have an idea on how to do this?

navneet1v:
Another problem I see with preloading is if the memory on the node is low, then it can fail the node bootup.

PR #2799, opened 2025-07-22
Index-level Setting for Graph Preloading

### LINKED ISSUE #711: [FEATURE] Prevent slow knn queries when nodes restart
(from https://forum.opensearch.org/t/any-best-practices-for-warming-up-indices-on-node-startup/12013)

The current behavior of knn queries is that they will first check if there are unloaded native library indices and load them to RAM before the query is performed. OpenSearch provides the warmup endpoint (/_plugins/_knn/warmup/index1,index2,index3) which makes it easier to find strategies that reduce the initial query latency.

We have an OpenSearch cluster with knn indices whose shards live in different nodes. Sometimes a node may go down, but by having multiple replicated shards, this results in no reduced availability. The issue appears when the node comes back online. 

What we see is the following:
1. A node containing a replica shard of a knn index goes down. All knn queries on this index execute without any extra latency or downtime, because all native library indices are already loaded
2. The node that was down comes back online, and now any knn query to the index will first try to load the native library indices of the node that just connected. This causes query latency and potential timeout errors if the shard is big.

We would like a strategy for removing the latency caused by the node coming back online. I'm not sure which solution is more practical but I could think of a couple of options:

1. Prevent the knn shards from being "seen" so that knn queries don't try to load the indices from the newly connected node. Meanwhile execute an initial warmup, and only then, allow the shard to be visible again (by visible here I mean whatever mechanism that knn queries use to realize that native library indices need loading).
2. Add option so that knn queries exclude replica shards where the native library indices are not loaded. But still send an asynchronous warmup operation so the missing shards can eventually be used.

I have tried running jobs that keep firing the warmup endpoint, as well as creating a script that runs a warmup operation immediately after the node starts. Neither of these solutions prevent slow queries, because by the time the script/job runs, the shard is already visible and causing slow queries/timeouts.

leoyu808:
### Description
Whenever a node restarts, a shard is rebalanced, or a large segment is merged, there is a large latency spike on the first query to the node due to graph files needing to be reloaded to native memory. This change creates a new index level setting: index.knn.warmup.enabled, which controls whether graph files are eagerly loaded into native memory as soon as a Lucene segment is opened.

### Related Issues
Implementation of #2642 which resolves #711.

### Check List


shatejas (commented):
This is more related to @navneet1v  concern https://github.com/opensearch-project/k-NN/issues/2642#issuecomment-2815863217. 

This seems to be changing the behavior to always warming up graphs for certain indices when the segment is created. Please correct me if I am wrong. From the issue it seemed like that we only wanted to do it on node drops and rebalancing

PR #2806, opened 2025-07-28
Optimizing Index Thread Quantity Defaults

### LINKED ISSUE #2747: [FEATURE] Hardware based index thread quantity defaults
Currently there is an existing setting `knn.algo_param.index_thread_qty` which controls the indexing rate on faiss. This decides the number of parallel threads that needs to be used during indexing for faiss. The default is set to 1 currently. Modern processors are multi-core and indexing speed can be increased by having multiple threads. 

A solution here would be to have defaults which are based on the type of instance rather than a static default of 1. This would increase the speed of indexing for processors which have more cores while keeping the current default as 1 for others

adityamachiroutu:
### Description
Currently, the index thread quantity default value is 1. This is generally okay, however for big machines, we can dynamically select a larger number of processors to use to make the indexing speed increase. As a preliminary solution this sets the quantity to 4 for large machines (32+ cores), and 1 for machines that are smaller machines (under 32 cores). 

Preliminary Benchmarking results show a consistent improvement in indexing time, with a smaller improvement in throughput. 

| Metric                                                     | Task                 | Original    | With Changes | Unit|
|------------------------------------------------------------|---------------------|-------------|--------------|------|
| Cumulative indexing time of primary shards                 |                     | 18.96618    | 15.02677     | min |
| Cumulative merge time of primary shards                    |                     | 10.59552    | 3.81717      | min |
| Median cumulative refresh time across primary shards       |                     | 0.45718     | 0.0831       | min |
| Mean Throughput                                           | custom-vector-bulk  | 8487.1      | 9817.53333   | docs/s |
| Median Throughput                                         | custom-vector-bulk  | 8449.67333  | 9726.87      | docs/s |

These results are a mean of 3 runs using the sift-128 dataset.

### Related Issues
Resolves #2747

### Check List


src/main/java/org/opensearch/knn/index/KNNSettings.java
```diff
@@ -1061,6 +1063,27 @@ public void onIndexModule(IndexModule module) {
...
     }
 
+    /**
+     * Finds the suggested number of indexing threads based on the number of available processors
+     *
+     * @return suggested number of indexing threads
+     */
+    public static int getHardwareDefaultIndexThreadQty() {
+        try {
+            return AccessController.doPrivileged((PrivilegedExceptionAction<Integer>) () -> {
```
kotwanikunal: @adityamachiroutu - Let's not do an explicit privileged access here. Also users can limit how many processors the OpenSearch process is restricted to.

Can we instead use https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/common/util/concurrent/OpenSearchExecutors.java#L119-L121?

Sample usage: https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/threadpool/ThreadPool.java#L250-L255

adityamachiroutu: Thanks for feedback, made the changes.

src/test/java/org/opensearch/knn/index/KNNSettingsTests.java
```diff
@@ -36,6 +37,7 @@
 import java.util.Map;
 import java.util.stream.Collectors;
 
+import static org.mockito.Mockito.*;
```
naveentatikonda: nit: don't use wildcards

src/main/java/org/opensearch/knn/index/KNNSettings.java
```diff
@@ -1061,6 +1062,25 @@ public void onIndexModule(IndexModule module) {
         });
     }
 
+    /**
+     * Finds the suggested number of indexing threads based on the number of available processors
+     *
+     * @return suggested number of indexing threads
+     */
+    public static int getHardwareDefaultIndexThreadQty() {
```
naveentatikonda: make it package-private as we aren't using it outside of this class except in tests

src/main/java/org/opensearch/knn/index/KNNSettings.java
```diff
@@ -1061,6 +1062,25 @@ public void onIndexModule(IndexModule module) {
...
     }
 
+    /**
+     * Finds the suggested number of indexing threads based on the number of available processors
+     *
+     * @return suggested number of indexing threads
+     */
+    static int getHardwareDefaultIndexThreadQty() {
+        try {
+            int availableProcessors = OpenSearchExecutors.allocatedProcessors(Settings.EMPTY);
```
kotwanikunal: I think we should still honor the processor limits. Can you try to pass through the settings in here?

naveentatikonda (dismissed):

adityamachiroutu:
failed test is flaky

naveentatikonda (approved):

src/main/java/org/opensearch/knn/index/KNNSettings.java
```diff
@@ -1056,6 +1062,25 @@ public void onIndexModule(IndexModule module) {
...
 
+    /**
+     * Finds the suggested number of indexing threads based on the number of available processors
+     *
+     * @return suggested number of indexing threads
+     */
+    static int getHardwareDefaultIndexThreadQty(final Settings settings) {
+        try {
+            int availableProcessors = OpenSearchExecutors.allocatedProcessors(settings);
+            if (availableProcessors >= 32) {
```
kotwanikunal: On another look - a nit - these values should be within KNNConstants

kotwanikunal (approved):

src/main/java/org/opensearch/knn/index/KNNSettings.java
adityamachiroutu: Made the change

shatejas:
@adityamachiroutu do we have the CPU utilization numbers for the experiments in the description.

Looks good otherwise

adityamachiroutu:
> @adityamachiroutu do we have the CPU utilization numbers for the experiments in the description.
> 
> Looks good otherwise

Hi Tejas. I did not record numbers specifically for CPU utilization, but looking back at my ec2 instance while running the benchmark, my changes had a 33% CPU utilization. 
Screenshot 2025-08-05 at 4 05 58 PM

kotwanikunal (approved):

src/main/java/org/opensearch/knn/index/KNNSettings.java
kotwanikunal: Thanks!

src/main/java/org/opensearch/knn/index/KNNSettings.java
```diff
@@ -1056,6 +1066,25 @@ public void onIndexModule(IndexModule module) {
...
+            if (availableProcessors >= INDEX_THREAD_QUANTITY_THRESHOLD) {
+                return INDEX_THREAD_QUANTITY_DEFAULT_LARGE;
+            } else {
+                return INDEX_THREAD_QUANTITY_DEFAULT_SMALL;
+            }
+        } catch (Exception e) {
+            logger.info("[KNN] Failed to determine available processors. Defaulting to 1. [{}]", e.getMessage(), e);
+            return 1;
+        }
+    }
```
navneet1v, line 1086: This logic is interesting. At-least what I remember was we will use 1/4 of the available processors with max as 4. But this logic is little different. I think what we are trying to say is if we have more than 32 processors on the machine then we will use 4 as merge threads. @naveentatikonda can we confirm here that this is what we want to do?

navneet1v (approved):

src/main/java/org/opensearch/knn/index/KNNSettings.java
navneet1v: Got my ans in description

PR #2807, opened 2025-07-28
Issue #459: Add Mustache and Expression scripting support for knn_vector fields

### LINKED ISSUE #459: Add support for Mustache and Expression scripting for knn_vector field
Currently, we support painless scripting with the `knn_vector` field type. OpenSearch also has 2 other scripting languages: [Mustache](https://github.com/opensearch-project/OpenSearch/tree/main/modules/lang-expression/src/main/java/org/opensearch/script/expression) and [Expression](https://github.com/opensearch-project/OpenSearch/tree/main/modules/lang-expression/src/main/java/org/opensearch/script/expression). This issue is to track adding support for `knn_vector` type with these scripting languages.

night-owl-1709:
## Description
  Implements issue #459 by adding Mustache template and Expression scripting support for knn_vector fields, complementing existing Painless scripting.

  ## Changes
  - Enhanced script engine with multi-language support and automatic type detection
  - Mustache templating with k-NN context variables and conditionals
  - Mathematical expression evaluation with k-NN variables and operations
  - Comprehensive test coverage with unit and integration tests
  - Full backward compatibility with existing k-NN and Painless scripts

  ## Testing
  - [x] Unit tests for all new components
  - [x] Integration tests for end-to-end functionality  
  - [x] Backward compatibility verified
  - [x] Code compiles and builds successfully

  ## Example Usage

  ### Mustache Template
  ```json
  {
    "script": {
      "source": "{{#has_vector}}1.0 / (1.0 + {{knn_l2_distance}}){{/has_vector}}{{^has_vector}}0.0{{/has_vector}}",
      "lang": "knn",
      "params": {"field": "my_vector", "query_value": [1.0, 2.0, 3.0]}
    }
  }

  Expression Script

  {
    "script": {
      "source": "1.0 / (1.0 + _knn_l2_distance)",
      "lang": "knn",
      "params": {"field": "my_vector", "query_value": [1.0, 2.0, 3.0]}
    }
  }

### Related Issues
Resolves #459

### Check List
- [Y ] New functionality includes testing.
- [ Y] New functionality has been documented.
- [ Y] Commits are signed per the DCO using `--signoff`.


jmazanec15:
@night-owl-1709 One question: for painless, we [extend the lang-painless plugin](https://github.com/opensearch-project/k-NN/blob/main/build.gradle#L318) and use SPI to extend it. Is that not possible with expression and mustache?

night-owl-1709:
Hi @jmazanec15  thanks for reviewing. 

I used factory based approach because of following issues:
  1. Context Separation: Mustache and Expression engines don't have access to k-NN vector operations by design
  2. Parameter Injection: We need to inject k-NN computed values as template/expression variables
  3. Document-level Processing: k-NN calculations need access to document vectors during execution
  4. Plugin Dependencies: Would require k-NN plugin to extend other plugins' SPIs

  Our Approach Benefits:
  - Self-contained within k-NN plugin
  - Direct access to k-NN utilities and vector operations
  - Backward compatible with existing k-NN scripts
  - Clear separation of concerns

src/main/java/org/opensearch/knn/plugin/KNNPlugin.java
```diff
@@ -380,7 +381,7 @@ public void onIndexModule(IndexModule indexModule) {
      */
     @Override
     public ScriptEngine getScriptEngine(Settings settings, Collection<ScriptContext<?>> contexts) {
-        return new KNNScoringScriptEngine();
```
jmazanec15, line 383: Should we delete this Engine?

night-owl-1709: fixed

src/main/java/org/opensearch/knn/plugin/KNNPlugin.java
```diff
@@ -380,7 +387,25 @@ public void onIndexModule(IndexModule indexModule) {
...
+    private KNNScoringScriptEngine scoringScriptEngine;
+
+    /**
+     * Get the k-NN enhanced script service that provides automatic Mustache support.
+     *
+     * This service automatically detects k-NN functions in Mustache templates and enhances them
+     * with k-NN context injection, enabling users to use standard "mustache" language with
+     * k-NN functions like {{knn.l2Squared}}.
+     */
+    public KNNScriptService getKNNScriptService() {
```
jmazanec15: Can we remove? I dont think this gets used

src/main/java/org/opensearch/knn/plugin/KNNPlugin.java
```diff
@@ -380,7 +387,25 @@ public void onIndexModule(IndexModule indexModule) {
...
-        return new KNNScoringScriptEngine();
+        // Return enhanced k-NN scoring engine that also supports Mustache templates
+        KNNScoringScriptEngine engine = new KNNScoringScriptEngine();
+        // We'll set the injector later in createComponents since ScriptService isn't available yet
+        this.scoringScriptEngine = engine;
+        return engine;
+    }
+
+    private KNNScriptService knnScriptService;
+    private KNNScoringScriptEngine scoringScriptEngine;
```
jmazanec15: Can we remove these 2 if above comment makes sense?

src/main/java/org/opensearch/knn/plugin/KNNPlugin.java
```diff
@@ -250,6 +251,12 @@ public Collection<Object> createComponents(
 
         clusterService.addListener(TrainingJobClusterStateListener.getInstance());
 
+        // Initialize k-NN enhanced script service for automatic Mustache support
+        knnScriptService = new KNNScriptService(scriptService);
```
jmazanec15: Can we remove this logic from create components and just have it all in getScriptEngine?

src/main/java/org/opensearch/knn/plugin/script/KNNScriptService.java
```diff
@@ -0,0 +1,90 @@
...
+        this.scriptService = scriptService;
+        this.knnInjector = new KNNMustacheContextInjector(scriptService);
+    }
+
+    /**
+     * Compile a script with automatic k-NN support detection.
+     * Enhances Mustache scripts containing k-NN functions, delegates others to standard compilation.
+     */
+    @SuppressWarnings("unchecked")
+    public <T> T compile(Script script, ScriptContext<T> context) {
```
jmazanec15: I dont believe this gets called anywhere.

src/main/java/org/opensearch/knn/plugin/script/KNNScriptService.java
```diff
@@ -0,0 +1,90 @@
...
+        } else if (context.equals(TemplateScript.CONTEXT)) {
+            TemplateScript.Factory factory = knnInjector.compileWithKNNSupport(script.getIdOrCode(), source, params);
+            return (T) factory;
+        } else {
+            return scriptService.compile(script, context);
+        }
+    }
+
+    /** Get the underlying ScriptService for direct access */
+    public ScriptService getScriptService() {
```
jmazanec15: I dont think this gets called anywhere

src/main/java/org/opensearch/knn/plugin/script/KNNScriptService.java
```diff
@@ -0,0 +1,90 @@
...
+        return scriptService;
+    }
+
+    /** Get the k-NN context injector for direct access */
+    public KNNMustacheContextInjector getKNNInjector() {
+        return knnInjector;
+    }
+
+    /** Check if a script requires k-NN processing */
+    public boolean requiresKNNProcessing(Script script) {
```
jmazanec15: I dont think this gets called anywhere

src/main/java/org/opensearch/knn/plugin/script/KNNScriptService.java
```diff
@@ -0,0 +1,90 @@
...
+        return "mustache".equals(script.getLang()) && knnInjector.requiresKNNProcessing(script.getIdOrCode());
+    }
+
+    /** Get supported k-NN functions */
+    public static java.util.Set<String> getSupportedKNNFunctions() {
```
jmazanec15: I dont think this gets called anywhere

src/main/java/org/opensearch/knn/plugin/script/KNNScriptService.java
```diff
@@ -0,0 +1,90 @@
...
+
+    private final ScriptService scriptService;
+    private final KNNMustacheContextInjector knnInjector;
+
+    public KNNScriptService(ScriptService scriptService) {
+        if (scriptService == null) {
+            throw new IllegalArgumentException("ScriptService cannot be null");
+        }
+        this.scriptService = scriptService;
+        this.knnInjector = new KNNMustacheContextInjector(scriptService);
```
jmazanec15: If this class just creates the knnInjector, do we need it as a class?

src/main/java/org/opensearch/knn/plugin/script/KNNScoringScriptEngine.java
```diff
@@ -9,42 +9,80 @@
...
     public static final String NAME = "knn";
     public static final String SCRIPT_SOURCE = "knn_score";
 
+    private KNNMustacheContextInjector knnInjector;
+
+    public KNNScoringScriptEngine() {
+        // Default constructor for backward compatibility
+    }
+
+    public KNNScoringScriptEngine(KNNMustacheContextInjector knnInjector) {
```
jmazanec15: Is this ever used?

src/main/java/org/opensearch/knn/plugin/script/KNNScoringScriptEngine.java
```diff
@@ -9,42 +9,80 @@
...
  */
 public class KNNScoringScriptEngine implements ScriptEngine {
 
     public static final String NAME = "knn";
     public static final String SCRIPT_SOURCE = "knn_score";
 
+    private KNNMustacheContextInjector knnInjector;
+
+    public KNNScoringScriptEngine() {
+        // Default constructor for backward compatibility
```
jmazanec15: I dont think we need to worry about backwards compatibility. Can you elaborate why we need to? If we can avoid having the setting below I think thatd be preferred

src/main/java/org/opensearch/knn/plugin/script/KNNMustacheContextInjector.java
```diff
@@ -0,0 +1,194 @@
...
+    public KNNMustacheContextInjector(ScriptService scriptService) {
+        if (scriptService == null) throw new IllegalArgumentException("ScriptService cannot be null");
+        this.scriptService = scriptService;
+    }
+
+    public boolean requiresKNNProcessing(String template) {
+        return template != null && KNN_FUNCTION_PATTERN.matcher(template).find();
+    }
+
+    public TemplateScript.Factory compileWithKNNSupport(String name, String template, Map<String, String> params) {
```
jmazanec15: do we need name arg? same for compileAsScoreScript

src/main/java/org/opensearch/knn/plugin/script/KNNMustacheContextInjector.java
```diff
@@ -0,0 +1,194 @@
...
+        if (!requiresKNNProcessing(template)) {
```
jmazanec15: We call this in the KNNScoringScriptEngine. Do we need to also perform check here?

src/main/java/org/opensearch/knn/plugin/script/KNNMustacheContextInjector.java
```diff
@@ -0,0 +1,194 @@
...
+import java.util.Map;
+import java.util.Set;
+import java.util.regex.Pattern;
+import java.util.regex.Matcher;
+
+/**
+ * SPI-like context injector for k-NN Mustache support.
+ * Automatically injects k-NN helper functions into Mustache template contexts.
+ */
+public class KNNMustacheContextInjector {
```
jmazanec15: Lets add extensive unit tests for this class.  Specifically, for the knn function pattern matching.

src/main/java/org/opensearch/knn/plugin/script/KNNMustacheContextInjector.java
```diff
@@ -0,0 +1,194 @@
...
+    }
+
+    public ScoreScript.Factory compileAsScoreScript(String name, String template, Map<String, String> params) {
+        if (!requiresKNNProcessing(template)) {
+            throw new IllegalArgumentException("Template does not contain k-NN functions: " + template);
+        }
+        return compileKNNScoreScript(template, asObjectMap(params));
+    }
+
+    public static Set<String> getSupportedFunctions() {
```
jmazanec15: Can we remove if we are not using?

build.gradle
```diff
@@ -340,6 +340,7 @@ dependencies {
     api "org.opensearch:opensearch:${opensearch_version}"
     api project(":remote-index-build-client")
     compileOnly "org.opensearch.plugin:opensearch-scripting-painless-spi:${versions.opensearch}"
+    compileOnly "com.github.spullara.mustache.java:compiler:0.9.14"
```
jmazanec15, line 353: Do we use this anywhere?

src/main/java/org/opensearch/knn/plugin/script/KNNScoringScriptEngine.java
```diff
@@ -9,42 +9,80 @@
...
+            ScoreScript.Factory factory = new KNNScoreScriptFactory();
+            return context.factoryClazz.cast(factory);
         }
-        // we use the script "source" as the script identifier
-        if (!SCRIPT_SOURCE.equals(code)) {
-            KNNCounter.SCRIPT_COMPILATION_ERRORS.increment();
-            throw new IllegalArgumentException("Unknown script name " + code);
+
+        // Handle k-NN enhanced Mustache templates (if injector is available)
+        if (knnInjector != null && knnInjector.requiresKNNProcessing(code)) {
```
jmazanec15: If knnInjector.requiresKNNProcessing(code)) is false, why should it throw an error?

src/main/java/org/opensearch/knn/plugin/script/KNNMustacheContextInjector.java
```diff
@@ -0,0 +1,194 @@
...
+                return factory.newInstance(enhanced);
+            };
+        } catch (Exception e) {
+            throw new IllegalArgumentException("Failed to compile k-NN Mustache template: " + e.getMessage(), e);
+        }
+    }
+
+    private ScoreScript.Factory compileKNNScoreScript(String template, Map<String, Object> params) {
+        try {
+            // For score scripts, bypass Mustache entirely and convert directly to Painless
```
jmazanec15: What do you mean by this point?

jmazanec15 (commented):
Just want to

src/main/java/org/opensearch/knn/plugin/script/KNNMustacheContextInjector.java
night-owl-1709: Added in KNNMustacheIT.java

jmazanec15: That files for iTs, can we create a KNNMustacheTests file for uTs?

src/main/java/org/opensearch/knn/plugin/script/KNNMustacheContextInjector.java
```diff
@@ -0,0 +1,182 @@
...
+        return compileKNNScoreScript(template, asObjectMap(params));
+    }
+
+    private Map<String, Object> asObjectMap(Map<String, String> params) {
+        return params == null ? new HashMap<>() : new HashMap<>(params);
+    }
+
+    private TemplateScript.Factory compileKNNTemplate(String template, Map<String, Object> params) {
+        try {
+            Map<String, String> functions = extractKNNFunctions(template);
```
jmazanec15: Do we need extractKNNFunctions? From logs, I see the following:
```
{{{knn.l2Squared}}=l2Squared}
{{{knn.cosineSimilarity}}=cosineSimilarity}
```

Can we just make this a static map of all of the functions? Then, we dont need the extract functions call.

src/main/java/org/opensearch/knn/plugin/script/KNNMustacheContextInjector.java
```diff
@@ -0,0 +1,182 @@
...
+        }
+        // Wrap with return statement if needed
+        String trimmed = result.trim();
+        if (!trimmed.contains("return") && !trimmed.contains(";") && !trimmed.contains("{")) {
+            result = "return " + trimmed + ";";
+        }
+        return result;
+    }
+
+    private String buildFunctionCall(String functionName, Map<String, Object> params) {
```
jmazanec15: do we need params arg here?

src/main/java/org/opensearch/knn/plugin/script/KNNMustacheContextInjector.java
```diff
@@ -0,0 +1,182 @@
...
+        return result;
+    }
+
+    private String convertTemplateToPainless(String template, Map<String, String> functions, Map<String, Object> params) {
+        String result = template;
+        // Replace k-NN function patterns directly with Painless calls
+        for (Map.Entry<String, String> entry : functions.entrySet()) {
+            result = result.replace(entry.getKey(), buildFunctionCall(entry.getValue(), params));
+        }
+        // Wrap with return statement if needed
```
jmazanec15: Can you elaborate on why we need this?

src/main/java/org/opensearch/knn/plugin/script/KNNMustacheContextInjector.java
```diff
@@ -0,0 +1,182 @@
...
+            Map<String, String> functions = extractKNNFunctions(template);
+            String markedTemplate = markKNNFunctions(template, functions);
+            TemplateScript.Factory factory = scriptService.compile(
+                new Script(ScriptType.INLINE, "mustache", markedTemplate, params),
+                TemplateScript.CONTEXT
+            );
+            return scriptParams -> {
+                Map<String, Object> enhanced = new HashMap<>(scriptParams);
+                enhanced.put("knn", new KNNTemplateHelpers());
+                return factory.newInstance(enhanced);
```
jmazanec15: For my understanding, whats purpose of the KNNTemplateHelpers? How does it end up getting used and who calls it?

src/test/java/org/opensearch/knn/plugin/script/KNNMustacheIT.java
```diff
@@ -0,0 +1,349 @@
...
+import org.apache.hc.core5.http.ParseException;
+import java.util.List;
+import java.util.Map;
+
+import static org.opensearch.common.xcontent.XContentFactory.jsonBuilder;
+
+/**
+ * Integration tests for k-NN with Mustache templating
+ */
+public class KNNMustacheIT extends KNNRestTestCase {
```
jmazanec15: Can we add a few more test cases? Divide by 0, multiple {{knn.<function>}} in the same source script, complex control flow (if/else) with mustache cases, default values, loops. 

Can we also have a test case where this is used as a search template?

Can we also validate that the score is coming back as expected (not just positive) in these scenarios?

src/main/java/org/opensearch/knn/plugin/KNNPlugin.java
```diff
@@ -380,6 +382,11 @@ public void onIndexModule(IndexModule indexModule) {
      */
     @Override
     public ScriptEngine getScriptEngine(Settings settings, Collection<ScriptContext<?>> contexts) {
+        // Check if this is specifically a template script request
+        if (contexts.size() == 1 && contexts.contains(TemplateScript.CONTEXT)) {
+            return new KNNMustacheEngine();
```
shatejas, line 387: I was wondering if plugin should reuse the existing implementation of [mustache engine](https://github.com/opensearch-project/OpenSearch/blob/main/modules/lang-mustache/src/main/java/org/opensearch/script/mustache/MustacheScriptEngine.java#L67). The plugin will probably need a dependency on lang-mustache module. @night-owl-1709 have you explored the approach? were there any problems with it?

night-owl-1709: I checked but we do not have lang-mustache support yet like we have for painless in opensearch.

shatejas: Not sure if I understand this but did you try adding lang-mustache here https://github.com/opensearch-project/k-NN/blob/main/build.gradle#L318

shatejas: So the implementation here is similar to https://github.com/opensearch-project/OpenSearch/blob/main/modules/lang-mustache/src/main/java/org/opensearch/script/mustache/MustacheScriptEngine.java#L67 with minor changes.
- This uses default factory compared to custom factory. The compile implementation of both is the same as custom factory inherits from default factory.
- There are some node metrics that are being added, if core implementation is used, I don't think its needed.

so instead of returning knn mustache engine we can return core implementation of it. 

I would prefer using core implementation to be consistent and less maintenance on the plugin. Let me know if I am missing something here

src/main/java/org/opensearch/knn/plugin/script/KNNMustacheEngine.java
```diff
@@ -0,0 +1,91 @@
...
+            super(params);
+            this.template = template;
+            this.params = params;
+        }
+
+        @Override
+        public String execute() {
+            StringWriter writer = new StringWriter();
+            try {
+                template.execute(writer, params);
```
shatejas, line 83: Does this need privileged access? https://github.com/opensearch-project/OpenSearch/blob/main/modules/lang-mustache/src/main/java/org/opensearch/script/mustache/MustacheScriptEngine.java#L67

Issue #2811, opened 2025-07-28
[FEATURE] [RFC] FP16 Support for Exact Search and Lucene ANN


ManasviGoyal:
## 1. Overview

This design introduces a new vector data type, `half_float`, to enable FP16 (half-precision floating-point) vector support for kNN.  The `half_float` type stores vector components as 16-bit IEEE 754 half-precision floats on disk, providing a 50% reduction in storage size compared to traditional 32-bit FP32 vectors. This compression is particularly beneficial for large-scale vector datasets, as it reduces disk I/O, improves cache efficiency, and lowers memory footprint  without incurring the significant accuracy loss typically seen with byte-level quantization techniques. The proposed solution is specifically targeted at exact search scenarios, configured with `index.knn = true` and `index.knn.advanced.approximate_threshold = -1`, as well as for use in Lucene’s approximate nearest neighbor search. 

---

## 2. Background

In vector search, two primary strategies exist for retrieving nearest neighbors: exact k-Nearest Neighbors (kNN) search and approximate nearest neighbor (ANN) search. Exact kNN search performs a brute-force comparison between the query vector and every vector stored in the index. This exhaustive approach guarantees the retrieval of the true nearest neighbors based on a defined distance metric. In contrast, ANN uses various optimization techniques (HNSW, IVF) to reduce search space and accelerate retrieval at the cost of some accuracy. ANN allows faster searches and better scalability, especially for large datasets. However, it consumes additional memory for storing vector data structures and may produce slightly inconsistent results across repeated queries. 

To support both exact and approximate search modes, Lucene introduces the [KNNVectorsFormat](https://github.com/apache/lucene/blob/7fc9fd36095b781cee818f974317e0faf06f7fc0/lucene/core/src/java/org/apache/lucene/codecs/KnnVectorsFormat.java) abstraction, which defines how vector data is stored, indexed, and retrieved. It handles the serialization of raw vectors, as well as the optional construction of graph-based structures like HNSW. 

* The storage occurs in two files per segment: the `.vec` file, which contains the flat vector data, and the `.vex` file, which contains information regarding vector graph connections.
* Depending on the [index.knn.advanced.approximate_threshold](https://docs.opensearch.org/docs/latest/vector-search/performance-tuning-indexing/) setting, it supports purely flat vectors storage for exact search or hybrid storage with graphs for ANN. 
[...]

Issue #2655, opened 2025-04-10
[FEATURE] fp16 quantization support for exact search


pflaquerre-figma:
[Exact k-NN](https://opensearch.org/docs/latest/vector-search/vector-search-techniques/knn-score-script/) (i.e. `index.knn = true` + `index.knn.advanced.approximate_threshold:-1` + using `knn_score` script at retrieval time) currently only allows two data types for vectors: `fp32` and `bytes`. It would be helpful if `fp16` were supported, as it would offer users a nice tradeoff between memory+shard size vs. impact on quality.

**What alternatives have you considered?**

1) Using faiss, which supports fp16 out of the box: this is a no go for us due to how approximate search interacts with our highly selective query filters. It also only applies to the vectors in the graph, rather than the underlying vectors stored in the docs.
2) Dropping down to `byte`: this requires heavier trade-offs on the quality side.
3) Stick with full floats and explore dimensionality reduction options outside of opensearch instead: that's what we're pursuing for now as an alternative. This is a lot more work, and is arguably orthogonal to quantization anyhow.

ManasviGoyal:
@yigithub Can you assign me this issue?

navneet1v:
@ManasviGoyal assigned the GH to you

PR #2814, opened 2025-07-29
Add custom format and C++ SIMD serializer for FP16

ManasviGoyal:
### Description
Second PR for #2811 (Custom Format and Serializer)

In this PR, I have implemented a custom format `KNN990HalfFloatFlatVectorsFormat` that uses `KNN990HalfFloatFlatVectorsWriter` to store vectors as FP16 (2 bytes) in the `.vec` file and `KNN990HalfFloatFlatVectorsReader` to load the vectors by converting FP16 bytes to FP32.  The FP32 <-> FP16 conversion is done using C++ SIMD instructions with separate implementations for ARM NEON and x86 machines.

In the next PRs, I will implement:
- FP16 implementation for Exact Search + tests for custom format
- FP16 implementation for Lucene ANN

### Related Issues
#2655 Feature Request
#2811 RFC

### Check List


jni/CMakeLists.txt
```diff
@@ -70,14 +70,23 @@ endif()
 # Set architecture specific variables
 if (${CMAKE_SYSTEM_PROCESSOR} STREQUAL aarch64)
     set(MACH_ARCH arm64)
+    set(COMMONS_FLAGS "-O3 -march=armv8.4-a+fp16 -fPIC")
 elseif(${CMAKE_SYSTEM_PROCESSOR} STREQUAL x86_64)
     set(MACH_ARCH x64)
+    set(COMMONS_FLAGS "-O3 -march=native -mf16c -fPIC")
```
finnroblin: @naveentatikonda Do you see any issues with these flags conflicting with the preexisting avx2/avx512/spr faiss build logic?

jni/src/commons.cpp
```diff
@@ -107,3 +114,140 @@ int knn_jni::commons::getIntegerMethodParameter(JNIEnv * env, knn_jni::JNIUtilIn
...
+
+// FP32 → FP16
+void knn_jni::commons::convertFP32ToFP16(knn_jni::JNIUtilInterface* jniUtil,
+                                         JNIEnv* env,
+                                         jfloatArray fp32Array,
+                                         jbyteArray fp16Array,
+                                         jint count) {
+    if (count <= 0) return;
+
+    jfloat* src_f32 = reinterpret_cast<jfloat*>(env->GetPrimitiveArrayCritical(fp32Array, nullptr));
```
finnroblin: Let's use `JNIUtil::GetPrimitiveArrayCritical` helper.

jni/src/commons.cpp
```diff
@@ -107,3 +114,140 @@ int knn_jni::commons::getIntegerMethodParameter(JNIEnv * env, knn_jni::JNIUtilIn
...
+    jbyte* dst_bytes = reinterpret_cast<jbyte*>(env->GetPrimitiveArrayCritical(fp16Array, nullptr));
+    const float* src = reinterpret_cast<const float*>(src_f32);
+    uint16_t* dst = reinterpret_cast<uint16_t*>(dst_bytes);
+
+    int i = 0;
+
+#if defined(__aarch64__)
```
finnroblin: Can we move each of these extension-specific implementations to a separate file? Like in this faiss code https://github.com/facebookresearch/faiss/blob/main/faiss/utils/distances_fused/distances_fused.cpp#L30-L39.

jni/src/commons.cpp
```diff
@@ -107,3 +114,140 @@ int knn_jni::commons::getIntegerMethodParameter(JNIEnv * env, knn_jni::JNIUtilIn
...
+    }
+  #elif defined(__AVX2__) && defined(__F16C__)
+    // x86 AVX2+F16C bulk 8-wide
+    for (; i + 8 <= count; i += 8) {
+        __m256 v = _mm256_loadu_ps(&src[i]);
+        __m128i h = _mm256_cvtps_ph(v, _MM_FROUND_TO_NEAREST_INT | _MM_FROUND_NO_EXC);
+        _mm_storeu_si128(reinterpret_cast<__m128i*>(&dst[i]), h);
+    }
+  #else
+    #error "x86_64 must support AVX512F or AVX2+F16C"
```
finnroblin: Can we fallback to a non-SIMD implementation in this case?

jni/src/commons.cpp
```diff
@@ -107,3 +114,140 @@ int knn_jni::commons::getIntegerMethodParameter(JNIEnv * env, knn_jni::JNIUtilIn
...
+    for (; i < count; ++i) {
+        __fp16 half = *reinterpret_cast<const __fp16*>(&src[i]);
+        float32x4_t fv = vcvt_f32_f16(vdup_n_f16(half));
+        dst[i] = vgetq_lane_f32(fv, 0);
+    }
+
+#elif defined(__x86_64__)
+  #if defined(__AVX512F__)
+    // x86 AVX-512 bulk 16-wide
+    for (; i + 16 <= count; i += 16) {
```
finnroblin: Is there a way/plan to add unit tests for the C++ implementation? That will give us more confidence about the correctness of the SIMD conversions. But I'm not sure how to integrate that into our existing unit testing framework, since not every machine running tests will have a supported processor.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNN990HalfFloatFlatVectorsWriter.java
```diff
@@ -0,0 +1,337 @@
...
+            );
+            CodecUtil.writeIndexHeader(
+                vectorData,
+                KNN990HalfFloatFlatVectorsFormat.VECTOR_DATA_CODEC_NAME,
+                KNN990HalfFloatFlatVectorsFormat.VERSION_CURRENT,
+                state.segmentInfo.getId(),
+                state.segmentSuffix
+            );
+        } catch (Throwable t) {
+            KNNIOUtils.closeWhileSuppressingExceptions(t, this);
```
finnroblin: Can we use Lucene IOUtils instead? Not sure if there's a reason to make a knn-specific ioutils class, but I might be missing something. Ref: https://github.com/opensearch-project/k-NN/blob/134a4107b572be9a44e4e10d74fe8e0c9f8c22ce/src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriter.java#L212

ManasviGoyal: Thank for the feedback @finnroblin!

I tried that but `closeWhileSuppressingExceptions` is not accessible via Lucene [IOUtils](https://github.com/apache/lucene/blob/b3f4011292c2e4ffc0e0471e830ccb813c72f44a/lucene/core/src/java/org/apache/lucene/util/IOUtils.java#L114). So the only way was to use a kNN specific IOUtils.

jni/src/commons.cpp
ManasviGoyal: I was thinking about it and we have 2 options here for the fallback - 
1. C++ : I can only think of bit manipulation but that needs proper testing to make sure the results are correct.
2. Java: Using `Float.floatToFloat16()` and `Float.float16ToFloat()`. 

Any suggestions?

jni/src/commons.cpp
ManasviGoyal: The only way I can think of is calling it via `JNICommons` in java and testing the conversions. I do have some edge case tests in a similar way in this PR.

finnroblin: I think it's okay to test from the Java side. Could you do some manual/sanity testing later on with avx2,avx512, and neon instances? Something like a r5 instance will have avx512 and any graviton/apple m chip machine will have neon. M4 instance has avx2 and not avx512. I don't think our current CI tests across different processor platforms, but we want to be sure that the float conversions work on all supported platforms. Just want to be careful since I think this is the first PR (exciting!) that uses SIMD intrinsics directly in k-NN.

jni/src/commons.cpp
finnroblin: I'd prefer the [java](https://docs.oracle.com/en/java/javase/24/docs/api/jdk.incubator.vector/jdk/incubator/vector/Float16.html) route since it will avoid jni overhead and (optimistically) the compiler will use simd instructions if available. But I think we can keep this error for now and then see after benchmarking how much worse Java is compared to these intrinsics. If it's a small enough penalty I'd rather use Java since it will be much less complex to test and maintain.

ManasviGoyal: @finnroblin During my initial POC for DocValuesFormat (which is not included here since it is on deprecation path), I had tested and benchmarked the Java FP16 approach and it was quite slow, that was the reason I switched to SIMD. According to the benchmark, Java was 2-3x slower than the SIMD intrinsics. 

That's why I am not quite sure if Java, although easier to maintain, would be a good fallback given the increase in latency or maybe this can be the tradeoff the users have to make if they do not have SIMD supported machines?

finnroblin: > I had tested and benchmarked the Java FP16 approach and it was quite slow, that was the reason I switched to SIMD. According to the benchmark, Java was 2-3x slower than the SIMD intrinsics.

Gotcha, thanks! Would you mind sharing the experiment details -- is it 3x slower for fp32->fp16 latency or some other opensearch metric? I'd think that the jit would be smart enough to call the right intrinsic for conversion, but that might not always be true. We might need to be super clear so that the compiler actually vectorizes like: 
```java
      for (int i = 0; i < in.length; i++) {
          out[i] = Float.floatToFloat16(in[i]); // out is short[], in is float[]
      }
```


>That's why I am not quite sure if Java, although easier to maintain, would be a good fallback given the increase in latency or maybe this can be the tradeoff the users have to make if they do not have SIMD supported machines?

If it's a true 2-3x higher fp32->fp16 conversion latency in java with vectorization enabled vs C++ then we should use C++ intrinsics imo and only support it for avx512/avx2/neon platforms (which probably covers most if not all the machines running opensearch k-NN). others might disagree with me on this.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNN990HalfFloatFlatVectorsReader.java
```diff
@@ -0,0 +1,384 @@
...
+import org.apache.lucene.util.RamUsageEstimator;
+import org.apache.lucene.util.hnsw.RandomVectorScorer;
+import org.opensearch.knn.index.codec.util.KNNVectorAsCollectionOfHalfFloatsSerializer;
+import org.opensearch.knn.index.codec.util.KNNIOUtils;
+
+/**
+ * A FlatVectorsReader that reads half-precision (2-byte) FP16 data from .vec files,
+ * decodes to float32 via the KNN serializer, and otherwise follows Lucene99 logic.
+ */
+public final class KNN990HalfFloatFlatVectorsReader extends FlatVectorsReader {
```
e-emoto: Are there tests that will be added for the HalfFloat Reader/Writer/Format?  Like [KNN990QuantizationStateWriterTests](https://github.com/opensearch-project/k-NN/blob/main/src/test/java/org/opensearch/knn/index/codec/KNN990Codec/KNN990QuantizationStateWriterTests.java) or [KNN990QuantizationStateReaderTests](https://github.com/opensearch-project/k-NN/blob/main/src/test/java/org/opensearch/knn/index/codec/KNN990Codec/KNN990QuantizationStateReaderTests.java)

ManasviGoyal: Yes! those will be in the next PR since this PR was already quite big.

jni/src/commons.cpp
finnroblin: Discussed offline, java was 2x slower for search latencies (so likely >>2x slower for actual float16->float conversions). I don't think it's worth using java approach then unless others have concerns with platform-specific intrinsics.

jni/include/commons.h
```diff
@@ -101,6 +101,27 @@ namespace knn_jni {
...
          **/
         int getIntegerMethodParameter(JNIEnv *, knn_jni::JNIUtilInterface *, std::unordered_map<std::string, jobject>, std::string, int);
+
+        /**
+         * Converts an array of FP32 values to FP16 values.
+         * @param fp32Array The input array of FP32 values.
+         * @param fp16Array The output array of FP16 values.
+         * @param count The number of elements in the arrays.
+         */
+        void convertFP32ToFP16(
```
0ctopus13prime: I don't think these encoding (fp16 -> fp32 or vice versa) should be placed in common.
could you create 'encoding.h' and put these in there?
you can give a better name to the header name.

jni/include/org_opensearch_knn_jni_JNICommons.h
```diff
@@ -66,6 +66,21 @@ JNIEXPORT void JNICALL Java_org_opensearch_knn_jni_JNICommons_freeBinaryVectorDa
 JNIEXPORT void JNICALL Java_org_opensearch_knn_jni_JNICommons_freeByteVectorData
 (JNIEnv *, jclass, jlong);
 
+/* Class:     org_opensearch_knn_jni_JNICommons
+ * Method:    convertFP16ToFP32
+ * Signature: ([F[BI)V
+ */
+JNIEXPORT void JNICALL Java_org_opensearch_knn_jni_JNICommons_convertFP32ToFP16
```
0ctopus13prime: I think JNICommon is for general JNI operation, like allocation/deallocation etc.
Could you create a dedicated file for encoding and decoding?
I have a strong hunch that encoding using native SIMD instructions will grow until Panama covers all cases.
For example, BF16.

jni/CMakeLists.txt
```diff
@@ -70,14 +70,23 @@ endif()
 # Set architecture specific variables
 if (${CMAKE_SYSTEM_PROCESSOR} STREQUAL aarch64)
     set(MACH_ARCH arm64)
+    set(COMMONS_FLAGS "-O3 -march=armv8.4-a+fp16 -fPIC")
```
0ctopus13prime: This will not work for ARM cpu that does not support the specific SIMD instruction.
We have to check whether `vld1q_f32` is available in processor's features set. CMAKE_SYSTEM_PROCESSOR this only tell ISA family, for individual feature, manual test is needed.

Ex:
```
check_cxx_source_compiles("
  #include <arm_neon.h>
  int main() {
    float data[4] = {0};
    float32x4_t v = vld1q_f32(data);
    (void)v;
    return 0;
  }
" HAVE_FP16_SIMD)

if(HAVE_FP16_SIMD)
  // Add -march=armv8.4-a+fp16 -fPIC
endif()
```

jni/CMakeLists.txt
0ctopus13prime: This too, we can't just give -march, it's not safe.

jni/src/commons.cpp
```diff
@@ -107,3 +114,140 @@ int knn_jni::commons::getIntegerMethodParameter(JNIEnv * env, knn_jni::JNIUtilIn
 
     return defaultValue;
 }
+
+// FP32 → FP16
+void knn_jni::commons::convertFP32ToFP16(knn_jni::JNIUtilInterface* jniUtil,
```
0ctopus13prime: Best practice to have SIMD code is to have one controller header + multiple cpp using dedicated SIMD instruction.
So the general include flow would look like KNN JNI header -> encoding.h -> default_encoding.h (e.g. without simd) OR x86_encoding.h OR arm_encoding.h.

In encoding.h, we should have something like below branching.

```
#if defined(ARM neon with FP16)
  #include <arm_encoding.h>
#elif defined(X86 AVX512 with FP16)
  #include <x86_encoding.h>
#else
  #include <default_encoding.h>
#endif
```

Lastly, we have 3 cpp files for each case.

jni/src/commons.cpp
0ctopus13prime: +1

jni/src/commons.cpp
```diff
@@ -107,3 +114,140 @@ int knn_jni::commons::getIntegerMethodParameter(JNIEnv * env, knn_jni::JNIUtilIn
...
+                                         JNIEnv* env,
+                                         jfloatArray fp32Array,
+                                         jbyteArray fp16Array,
+                                         jint count) {
+    if (count <= 0) return;
+
+    jfloat* src_f32 = reinterpret_cast<jfloat*>(jniUtil->GetPrimitiveArrayCritical(env, fp32Array, nullptr));
+    jbyte* dst_bytes = reinterpret_cast<jbyte*>(jniUtil->GetPrimitiveArrayCritical(env, fp16Array, nullptr));
+    const float* src = reinterpret_cast<const float*>(src_f32);
+    uint16_t* dst = reinterpret_cast<uint16_t*>(dst_bytes);
```
0ctopus13prime: We can't directly cast jbyte* to uint16_t* and we have to make sure alignment matches up between first.
```
if (((uintptr_t)jbyte % alignof(uint16_t)) == 0) {
    // safe to cast
}
```

jni/src/commons.cpp
```diff
@@ -107,3 +114,140 @@ int knn_jni::commons::getIntegerMethodParameter(JNIEnv * env, knn_jni::JNIUtilIn
...
+                                         jbyteArray fp16Array,
+                                         jfloatArray fp32Array,
+                                         jint count,
+                                         jint offset) {
+    if (count <= 0) return;
+
+    jfloat* dst_f32 = reinterpret_cast<jfloat*>(jniUtil->GetPrimitiveArrayCritical(env, fp32Array, nullptr));
+    jbyte* src_bytes = reinterpret_cast<jbyte*>(jniUtil->GetPrimitiveArrayCritical(env, fp16Array, nullptr));
+    float* dst = reinterpret_cast<float*>(dst_f32);
+    const uint16_t* src = reinterpret_cast<const uint16_t*>(src_bytes + offset);
```
0ctopus13prime: This too, we need to check alignment for safe up-casting.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNN990HalfFloatFlatVectorsFormat.java
```diff
@@ -0,0 +1,63 @@
...
+ */
+
+/**
+ * Custom FlatVectorsFormat implementation to support half-float vectors. This class is mostly identical to
+ * {@link org.apache.lucene.codecs.lucene99.Lucene99FlatVectorsFormat}, however we use the custom {@link KNN990HalfFloatFlatVectorsWriter}
+ * and {@link KNN990HalfFloatFlatVectorsReader} for storage and retrieval of half-float vectors.
+ */
+public final class KNN990HalfFloatFlatVectorsFormat extends FlatVectorsFormat {
+
+    static final String NAME = "KNN990HalfFloatFlatVectorsFormat";
```
0ctopus13prime: Could you get rid of this and use getClass().getSimpleName()?

jni/CMakeLists.txt
akashsha1: -march=native essentially implies that the code is not guaranteed to run on a machine older than the machine where its compiled. 

https://gcc.gnu.org/onlinedocs/gcc/x86-Options.html#x86-Options-1

akashsha1: the other flags seem fine, since the flags are contained to only commons.cpp. They do not modify the faiss build logic. 

I'd suggest test on some older and newer generation of machines, compared to the machine where they're built for compat.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNN990HalfFloatFlatVectorsReader.java
```diff
@@ -0,0 +1,386 @@
...
+        final FieldInfo info = fieldInfos.fieldInfo(field);
+        final FieldEntry entry;
+        if (info == null || (entry = fields.get(info.number)) == null) {
+            throw new IllegalArgumentException("field=\"" + field + "\" not found");
+        }
+        return entry;
+    }
+
+    @Override
+    public ByteVectorValues getByteVectorValues(String field) throws IOException {
```
0ctopus13prime: Could you give a java doc saying this will give raw bytes?
The returned bytes is not actually ByteVectorValues

jni/src/commons.cpp
ManasviGoyal: @finnroblin I manually ran the unit tests and performed exact/ann search on `cohere-1m` dataset on `m4.2xlarge` (avx2), `r5.2xlarge` (avx512) and `r6g.2xlarge` (neon) as you suggested. The conversions work fine on all these.

jni/src/commons.cpp
naveentatikonda: @ManasviGoyal we must support non-SIMD implementation though it is a lot slower. I don't think it is a good practice to break the functionality for non-SIMD users.

ManasviGoyal: Sure @naveentatikonda. I have already added a java fallback which will be triggered on non-SIMD based on the feedback. I will test it on a non-SIMD supported instance (found a old one that doesn't support simd) before pushing the changes. Thanks!

naveentatikonda: @ManasviGoyal You can use this as a reference to disable SIMD on an x86 instance type - https://github.com/opensearch-project/opensearch-build/issues/4386#issuecomment-1917854007

But, I don't think we can disable NEON on ARM instance

ManasviGoyal:
@0ctopus13prime @finnroblin @naveentatikonda I have resolved all the comments. 
- The code is now split into different files for arm, x86 and default (java fallback) cases. At compile time it is decided which files to build.
- Java fallback which is triggered when the required neon, avx512 or avx2 architecture not supported. 
- The unsafe flags are removed from `CMakeLists.txt` and only architecture specific flags are added.
- I manually ran the unit FP16 tests on `m4.2xlarge` (avx2), `r5.2xlarge` (avx512), `r6g.2xlarge` (neon) and `m1.small` (non-simd). The correct instructions are being triggered and all conversions work fine on all these.

jni/CMakeLists.txt
```diff
@@ -127,6 +128,50 @@ endif ()
...
+# ---------------------------------- SIMD ENCODING/DECODING ----------------------------------
+set(TARGET_LIB_SIMD opensearchknn_simd)
+
+# Architecture-specific SIMD implementation
+set(SIMD_ENCODING_SRCS
+    ${CMAKE_CURRENT_SOURCE_DIR}/src/org_opensearch_knn_jni_SIMDEncoding.cpp
+    ${CMAKE_CURRENT_SOURCE_DIR}/src/org_opensearch_knn_jni_SIMDDecoding.cpp
+)
+
+if ("${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "aarch64" AND KNN_HAVE_ARM_FP16)
```
0ctopus13prime: Generally, this branching should be made in cpp side. 
Since we're adding `-march` flag, I think we can add something like below in `encoding.cpp` file.
Then just add `encoding.cpp` to `SIMD_ENCODING_SRCS`.

```
// encoding_dispatch.cpp
#if defined(__x86_64__) || defined(_M_X64)
    #include "x86_encoding.cpp"
#elif defined(__aarch64__) || defined(_M_ARM64)
    #include "arm_encoding.cpp"
#else
    #error "Unsupported architecture"
#endif
```

jni/src/decoding/arm_decoding.cpp
```diff
@@ -0,0 +1,58 @@
...
+    return JNI_TRUE;
+}
+
+jboolean knn_jni::decoding::convertFP16ToFP32(knn_jni::JNIUtilInterface *jniUtil, JNIEnv* env, jbyteArray fp16Array, jfloatArray fp32Array, jint count, jint offset) {
+    if (count <= 0) return JNI_TRUE;
+    jfloat* dst_f32 = reinterpret_cast<jfloat*>(jniUtil->GetPrimitiveArrayCritical(env, fp32Array, nullptr));
+    jbyte* src_bytes = reinterpret_cast<jbyte*>(jniUtil->GetPrimitiveArrayCritical(env, fp16Array, nullptr));
+
+    if ((reinterpret_cast<uintptr_t>(src_bytes + offset) % alignof(uint16_t)) != 0) {
+        env->ReleasePrimitiveArrayCritical(fp16Array, src_bytes, JNI_ABORT);
```
0ctopus13prime: Could you consider using JNIReleaseElements?

```
    knn_jni::JNIReleaseElements release_java_string {[=](){
        jniUtil->ReleasePrimitiveArrayCritical(env, fp32Array, dst_f32, 0);
        jniUtil->ReleasePrimitiveArrayCritical(env, fp16Array, src_bytes, JNI_ABORT);
    }};
```

It will release the arrays in its destructor, then code can be much simpler and safe.

jni/src/decoding/x86_decoding.cpp
```diff
@@ -0,0 +1,69 @@
...
+    return JNI_TRUE;
+}
+
+jboolean knn_jni::decoding::convertFP16ToFP32(knn_jni::JNIUtilInterface *jniUtil, JNIEnv* env, jbyteArray fp16Array, jfloatArray fp32Array, jint count, jint offset) {
+    if (count <= 0) return JNI_TRUE;
+    jfloat* dst_f32 = reinterpret_cast<jfloat*>(jniUtil->GetPrimitiveArrayCritical(env, fp32Array, nullptr));
+    jbyte* src_bytes = reinterpret_cast<jbyte*>(jniUtil->GetPrimitiveArrayCritical(env, fp16Array, nullptr));
+
+    if ((reinterpret_cast<uintptr_t>(src_bytes + offset) % alignof(uint16_t)) != 0) {
+        env->ReleasePrimitiveArrayCritical(fp16Array, src_bytes, JNI_ABORT);
```
0ctopus13prime: Could you consider using JNIReleaseElements?

```
    knn_jni::JNIReleaseElements release_java_string {[=](){
        jniUtil->ReleasePrimitiveArrayCritical(env, fp32Array, dst_f32, 0);
        jniUtil->ReleasePrimitiveArrayCritical(env, fp16Array, src_bytes, JNI_ABORT);
    }};
```

It will release the arrays in its destructor, then code can be much simpler and safe.

jni/src/encoding/arm_encoding.cpp
```diff
@@ -0,0 +1,63 @@
...
+    return JNI_TRUE;
+}
+
+jboolean knn_jni::encoding::convertFP32ToFP16(knn_jni::JNIUtilInterface *jniUtil, JNIEnv* env, jfloatArray fp32Array, jbyteArray fp16Array, jint count) {
+    if (count <= 0) return JNI_TRUE;
+    jfloat* src_f32 = reinterpret_cast<jfloat*>(jniUtil->GetPrimitiveArrayCritical(env, fp32Array, nullptr));
+    jbyte* dst_bytes = reinterpret_cast<jbyte*>(jniUtil->GetPrimitiveArrayCritical(env, fp16Array, nullptr));
+
+    if ((reinterpret_cast<uintptr_t>(dst_bytes) % alignof(uint16_t)) != 0) {
+        env->ReleasePrimitiveArrayCritical(fp16Array, dst_bytes, 0);
```
0ctopus13prime: Could you consider using JNIReleaseElements?

```
    knn_jni::JNIReleaseElements release_java_string {[=](){
        jniUtil->ReleasePrimitiveArrayCritical(env, fp32Array, dst_f32, 0);
        jniUtil->ReleasePrimitiveArrayCritical(env, fp16Array, src_bytes, JNI_ABORT);
    }};
```

It will release the arrays in its destructor, then code can be much simpler and safe.

jni/src/encoding/default_encoding.cpp
```diff
@@ -0,0 +1,27 @@
...
+
+#include "jni_util.h"
+#include "encoding/default_encoding.h"
+
+jboolean knn_jni::encoding::isSIMDSupported() {
+    return JNI_FALSE;
+}
+
+jboolean knn_jni::encoding::convertFP32ToFP16(knn_jni::JNIUtilInterface *jniUtil, JNIEnv*, jfloatArray, jbyteArray, jint) {
+    std::cerr << "[KNN] Warning: convertFP32ToFP16 called on unsupported platform. Java fallback expected.\n";
```
0ctopus13prime: Not sure this is a good way for logging. The message might be swallowed and not always to be passed into Java's logger.
Could you consider take a logger and construct Java string then call `warn` method instead?

jni/src/encoding/x86_encoding.cpp
```diff
@@ -0,0 +1,75 @@
...
+    if (count <= 0) return JNI_TRUE;
+
+    // Obtain direct in-place access to the Java byte[] backing store for FP16 results.
+    // GetPrimitiveArrayCritical typically pins the Java heap object and returns its real address
+    // (no copy on most JVMs, since they support pinning).
+    // If pinning isn’t possible (rare on modern OpenJDK), it will fall back to copying,
+    // so always check for NULL.
+    // While pinned, GC compaction is disabled, so keep this critical region very short
+    // to avoid stalling other threads or garbage collection.
+    jfloat* src_f32 = reinterpret_cast<jfloat*>(jniUtil->GetPrimitiveArrayCritical(env, fp32Array, nullptr));
```
0ctopus13prime, line 37: Could you consider using JNIReleaseElements?

```
    knn_jni::JNIReleaseElements release_java_string {[=](){
        jniUtil->ReleasePrimitiveArrayCritical(env, fp32Array, dst_f32, 0);
        jniUtil->ReleasePrimitiveArrayCritical(env, fp16Array, src_bytes, JNI_ABORT);
    }};
```

It will release the arrays in its destructor, then code can be much simpler and safe.

jni/src/org_opensearch_knn_jni_SIMDDecoding.cpp
```diff
@@ -0,0 +1,38 @@
...
+    JNIEnv* env,
+    jclass,
+    jbyteArray fp16Array,
+    jfloatArray fp32Array,
+    jint count,
+    jint offset) {
+    try {
+        return knn_jni::decoding::convertFP16ToFP32(&jniUtil, env, fp16Array, fp32Array, count, offset);
+    } catch (...) {
+        return JNI_FALSE;
```
0ctopus13prime, line 32: Could you consider using `jniUtil.CatchCppExceptionAndThrowJava(env);` and not swallow the exception

jni/src/org_opensearch_knn_jni_SIMDEncoding.cpp
```diff
@@ -0,0 +1,53 @@
...
+JNIEXPORT jboolean JNICALL Java_org_opensearch_knn_jni_SIMDEncoding_convertFP32ToFP16(
+    JNIEnv* env,
+    jclass,
+    jfloatArray fp32Array,
+    jbyteArray fp16Array,
+    jint count) {
+    try {
+        return knn_jni::encoding::convertFP32ToFP16(&jniUtil, env, fp32Array, fp16Array, count);
+    } catch (...) {
+        return JNI_FALSE;
```
0ctopus13prime, line 31: Could you consider using jniUtil.CatchCppExceptionAndThrowJava(env); and not swallow the exception

src/main/java/org/opensearch/knn/index/codec/util/KNNVectorAsCollectionOfHalfFloatsSerializer.java
```diff
@@ -0,0 +1,117 @@
...
+     * Handles both little-endian and big-endian platforms.
+     *
+     * @param input  float array to convert
+     * @param output output byte array to hold FP16-encoded values
+     * @param dimension number of elements to convert
+     */
+    private void floatToByteArrayFallback(float[] input, byte[] output, int dimension) {
+        for (int i = 0; i < dimension; ++i) {
+            short fp16 = Float.floatToFloat16(input[i]);
+            if (IS_LITTLE_ENDIAN) {
```
0ctopus13prime: Could you consider taking the branch checking endianess out of the loop to minimize branching in hot loop.
```
if (IS_LITTLE_ENDIAL) {
    for (...) {

    }
} else {
    for (...) {

    }
}
```

src/main/java/org/opensearch/knn/jni/SIMDDecoding.java
```diff
@@ -0,0 +1,39 @@
...
+import java.security.PrivilegedAction;
+
+public class SIMDDecoding {
+
+    /**
+     * Checks if the platform supports SIMD-based FP16 decoding.
+     *
+     * @return true if SIMD is supported and enabled, false otherwise
+     */
+    public static native boolean isSIMDSupported();
```
0ctopus13prime: NIT : Can we cache value and avoid calling the method everytime?
```
Ex:

private static final boolean SIMD_SUPPORTED = isSIMDSupportedNative();

public static boolean isSIMDSupported() {
    return SIMD_SUPPORTED;
}

private static native boolean isSIMDSupportedNative();   -> this will call C++ to get the flag.
```

0ctopus13prime (commented):
Overall, looks good, but just one thing.

Generally, it is a typical pattern that we define SIMD function signatures in one header then in branching in CPP side to decide which CPP file to pick up.

For example, we could do something like below instead of have the branch in CMake side.

```
// encoding.h. Just declare signatures.

jboolean isSIMDSupported();

jboolean convertFP32ToFP16(knn_jni::JNIUtilInterface *, JNIEnv* env, jfloatArray fp32Array, jbyteArray fp16Array, jint count);

// encoding.cpp. Pick up the actual CPP file.
#if defined(__aarch64__) && defined(KNN_HAVE_ARM_FP16)
  #include "arm_encoding.cpp"
#elif defined(__x86_64__) && (defined(KNN_HAVE_AVX512) || defined(KNN_HAVE_F16C))
  #include "x86_encoding.cpp"
#else
  #include "default_encoding.cpp"
#endif

```


Then in CMake, we add two files encoding.h + encoding.cpp.
This is the general pattern including chip specific functions.

src/main/java/org/opensearch/knn/index/codec/util/KNNVectorAsCollectionOfHalfFloatsSerializer.java
```diff
@@ -0,0 +1,130 @@
...
+            throw new IllegalArgumentException("Input/output buffers cannot be null.");
+        }
+        if (dimension > input.length) {
+            throw new IllegalArgumentException("Count exceeds input float array length.");
+        }
+        if (output.length != input.length * BYTES_IN_HALF_FLOAT) {
+            throw new IllegalArgumentException("Output buffer size mismatch. Must be 2x input length.");
+        }
+
+        if (SIMDEncoding.isSIMDSupported() && SIMDEncoding.convertFP32ToFP16(input, output, dimension)) {
```
kotwanikunal: I think we'd also benefit from implementing a fallback approach here.
Currently we throw out the exception in case of CPP failures and could lead to cascading failures.

We can work it similar to what we have for remote index builds - 

```
if (SIMDEncoding.isSIMDSupported()) {
    try {
        SIMDEncoding.convertFP32ToFP16(input, output, dimension);
        return; 
    } catch (Exception e) {
        logger.debug("SIMD conversion failed, falling back to Java: {}", e.getMessage());
        // Continue to Java fallback
    }
}
```

build.gradle
```diff
@@ -416,7 +416,7 @@ tasks.register('buildJniLib', Exec) {
     args.add("--build")
     args.add("jni/build")
     args.add("--target")
-    def knn_libs = ['opensearchknn_faiss', 'opensearchknn_common', 'opensearchknn_nmslib']
+    def knn_libs = ['opensearchknn_simd', 'opensearchknn_faiss', 'opensearchknn_common', 'opensearchknn_nmslib']
```
kotwanikunal, line 419: Not directly related to this - but can you also add in support for a flag to disable SIMD.
It can be an index setting or an arg

```
-Dknn.simd.enabled=false
```

For scenarios where we want to avoid SIMD completely - an example can be compatibility issues with a version upgrade/missing libs. It'll be good to have a lever.

src/main/java/org/opensearch/knn/jni/SIMDEncoding.java
```diff
@@ -0,0 +1,72 @@
...
+                    System.loadLibrary(KNNConstants.SIMD_AVX512_SPR_JNI_LIBRARY_NAME);
+                } else if (PlatformUtils.isAVX512SupportedBySystem()) {
+                    System.loadLibrary(KNNConstants.SIMD_AVX512_JNI_LIBRARY_NAME);
+                } else if (PlatformUtils.isAVX2SupportedBySystem()) {
+                    System.loadLibrary(KNNConstants.SIMD_AVX2_JNI_LIBRARY_NAME);
+                } else {
+                    System.loadLibrary(KNNConstants.SIMD_JNI_LIBRARY_NAME);
+                }
+            } catch (UnsatisfiedLinkError e) {
+                throw new RuntimeException("[KNN] Failed to load native SIMD library", e);
```
kotwanikunal: This would also need handling for the fallback logic. 
I think it's better to design for correctness first and then performance. You can add in a log statement here calling out a failure to load these libraries - we do not want to fail all other cases.

jni/src/encoding/arm_encoding.cpp
```diff
@@ -0,0 +1,71 @@
...
+    // When 'release_arrays' goes out of scope, its lambda will run to ensure that the
+    // critical arrays are always released properly even if there's an error or early return
+    knn_jni::JNIReleaseElements release_arrays{[=]() {
+        // Release the destination FP16 array, mode 0 means changes are written back
+        jniUtil->ReleasePrimitiveArrayCritical(env, fp16Array, dst_bytes, 0);
+        // Release the source FP32 array, JNI_ABORT means we don't write changes back (read-only)
+        jniUtil->ReleasePrimitiveArrayCritical(env, fp32Array, src_f32, JNI_ABORT);
+    }};
+
+    if ((reinterpret_cast<uintptr_t>(dst_bytes) % alignof(uint16_t)) != 0) {
```
0ctopus13prime, line 42: We don't need this alignment check for vld1q_f32.
We can even pass an address that is multiple of 3.

jni/src/encoding/x86_encoding.cpp
```diff
@@ -0,0 +1,83 @@
...
+        jniUtil->ReleasePrimitiveArrayCritical(env, fp16Array, dst_bytes, 0);
+        // Release the source FP32 array, JNI_ABORT means we don't write changes back (read-only)
+        jniUtil->ReleasePrimitiveArrayCritical(env, fp32Array, src_f32, JNI_ABORT);
+    }};
+
+    // Ensure the FP16 destination buffer is 2-byte aligned.
+    // The JVM will almost always give you an 8 or at least 4-byte-aligned buffer for any primitive array,
+    // so the base pointer (dst_bytes or src_bytes) is naturally aligned to alignof(uint16_t)==2.
+    // A misalignment here is therefore extremely rare, but this guard prevents undefined behavior or crashes.
+    if ((reinterpret_cast<uintptr_t>(dst_bytes) % alignof(uint16_t)) != 0) {
```
0ctopus13prime, line 53: I don't think this alignment check is required for _mm512_loadu_ps
'loadu' represents load unaligned pointer

ManasviGoyal:
I will address @kotwanikunal 's comments in the next PRs based on the consensus on whether we should prioritize strict failure or favor a silent fallback with logging.

jni/cmake/init-simd.cmake
```diff
@@ -0,0 +1,107 @@
...
+set(KNN_HAVE_F16C OFF)
+set(KNN_HAVE_AVX512 OFF)
+set(KNN_HAVE_ARM_FP16 OFF)
+
+set(SIMD_OPT_LEVEL "")
+set(SIMD_FLAGS "")
+
+# AVX512 detection (compiler only)
+if(AVX512_ENABLED)
+    set(CMAKE_REQUIRED_FLAGS "-mavx512f -mf16c")
```
navneet1v: The whole detection logic is very interesting. But that is not how we would like to put this. So the way we should be thinking about these things is:
1. operator pass the flag which says enable SIMD_OPT for the knn build.
2. We either build the so file or we fail. This is similar to how faiss does. If some one pass avx_512 flag and system doesn't support it then build fails
3. When the whole k-NN plugin will be packaged we should build all the different version of opensearch_smid just like we do for faiss. 
4. At runtime we should check if SIMD with avx_512 is supported we pick the right so files.

src/main/java/org/opensearch/knn/index/codec/util/KNNVectorAsCollectionOfHalfFloatsSerializer.java
```diff
@@ -0,0 +1,130 @@
...
+import org.opensearch.knn.jni.SIMDEncoding;
+
+import java.nio.ByteOrder;
+
+/**
+ * Class implements KNNVectorSerializer based on serialization/deserialization of float array
+ * as a collection of individual half-precision values
+ */
+public class KNNVectorAsCollectionOfHalfFloatsSerializer {
+    private static final Logger log = LogManager.getLogger(KNNVectorAsCollectionOfHalfFloatsSerializer.class);
```
navneet1v: lets use @Log4j2 here

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNN990HalfFloatFlatVectorsFormat.java
```diff
@@ -0,0 +1,63 @@
...
+ */
+
+/**
+ * Custom FlatVectorsFormat implementation to support half-float vectors. This class is mostly identical to
+ * {@link org.apache.lucene.codecs.lucene99.Lucene99FlatVectorsFormat}, however we use the custom {@link KNN990HalfFloatFlatVectorsWriter}
+ * and {@link KNN990HalfFloatFlatVectorsReader} for storage and retrieval of half-float vectors.
+ */
+public final class KNN990HalfFloatFlatVectorsFormat extends FlatVectorsFormat {
+
+    static final String NAME = KNN990HalfFloatFlatVectorsFormat.class.getSimpleName();
```
navneet1v: lets do something like this: https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsFormat.java#L37

getSimpleName can create issue if the class name changes in future

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNN990HalfFloatFlatVectorsFormat.java
```diff
@@ -0,0 +1,63 @@
...
+    static final String META_CODEC_NAME = "Lucene99FlatVectorsFormatMeta";
+    static final String VECTOR_DATA_CODEC_NAME = "Lucene99FlatVectorsFormatData";
+    static final String META_EXTENSION = "vemf";
+    static final String VECTOR_DATA_EXTENSION = "vec";
+
+    public static final int VERSION_START = 0;
+    public static final int VERSION_CURRENT = VERSION_START;
```
navneet1v: can these be also package private

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNN990HalfFloatFlatVectorsWriter.java
```diff
@@ -0,0 +1,443 @@
...
+import org.apache.lucene.util.hnsw.UpdateableRandomVectorScorer;
+import org.apache.lucene.util.Bits;
+import org.apache.lucene.search.VectorScorer;
+import org.opensearch.knn.index.codec.util.KNNIOUtils;
+import org.opensearch.knn.index.codec.util.KNNVectorAsCollectionOfHalfFloatsSerializer;
+
+/**
+ * Writes half float vector values to index segments.
+ */
+public final class KNN990HalfFloatFlatVectorsWriter extends FlatVectorsWriter {
```
navneet1v: I think most of the reference has been taken from Lucene implemetation. Lets mention it on top of this file as doc

src/main/java/org/opensearch/knn/index/codec/util/KNNVectorAsCollectionOfHalfFloatsSerializer.java
navneet1v: I think I understand what we are trying to do here, we are asking c++ code to tell if SIMD is supported or not. THis is catched in the Encoding decode code. 

* if SIMD is enabled then lets not fallback to Java, If the code fails it fails. That is a bug from our side and we should fix it. This silent failing is hard to catch in automated tests.

jni/src/decoding/x86_decoding.cpp
```diff
@@ -0,0 +1,73 @@
...
+    float* dst = reinterpret_cast<float*>(dst_f32);
+    const uint16_t* src = reinterpret_cast<const uint16_t*>(src_bytes + offset);
+
+    int i = 0;
+#if defined(KNN_HAVE_AVX512)
+    for (; i + 16 <= count; i += 16) {
+        if (i + 64 < count) {
+            _mm_prefetch(reinterpret_cast<const char*>(&src[i + 64]), _MM_HINT_T0);
+        }
+        __m256i h = _mm256_loadu_si256(reinterpret_cast<const __m256i*>(&src[i]));
```
0ctopus13prime, line 51: You can't directly cast here, it can segfault because of the alignemnt requirement.
please use below:
```
__m256 vec = _mm256_loadu_ps(&src[i]);
```

jni/src/encoding/arm_encoding.cpp
```diff
@@ -0,0 +1,66 @@
...
+    knn_jni::JNIReleaseElements release_arrays{[=]() {
+        // Release the destination FP16 array, mode 0 means changes are written back
+        jniUtil->ReleasePrimitiveArrayCritical(env, fp16Array, dst_bytes, 0);
+        // Release the source FP32 array, JNI_ABORT means we don't write changes back (read-only)
+        jniUtil->ReleasePrimitiveArrayCritical(env, fp32Array, src_f32, JNI_ABORT);
+    }};
+
+    const float* src = reinterpret_cast<const float*>(src_f32);
+    uint16_t* dst = reinterpret_cast<uint16_t*>(dst_bytes);
+
```
0ctopus13prime, line 49: we have to check whether dst_bytes can be casted to uint16_t.
e.g. 2 byte aligned

jni/src/encoding/x86_encoding.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+    const float* src = reinterpret_cast<const float*>(src_f32);
+    uint16_t* dst = reinterpret_cast<uint16_t*>(dst_bytes);
+
+    int i = 0;
+#if defined(KNN_HAVE_AVX512)
+    for (; i + 16 <= count; i += 16) {
+        __m512 v = _mm512_loadu_ps(&src[i]);
+        __m256i h = _mm512_cvtps_ph(v, _MM_FROUND_TO_NEAREST_INT | _MM_FROUND_NO_EXC);
+        _mm256_storeu_si256(reinterpret_cast<__m256i*>(&dst[i]), h);
```
0ctopus13prime, line 57: Please avoid direct casting and use load instruction.

jni/src/encoding/x86_encoding.cpp
```diff
@@ -0,0 +1,74 @@
...
+    }
+#elif defined(KNN_HAVE_F16C)
+    for (; i + 8 <= count; i += 8) {
+        __m256 v = _mm256_loadu_ps(&src[i]);
+        __m128i h = _mm256_cvtps_ph(v, _MM_FROUND_TO_NEAREST_INT | _MM_FROUND_NO_EXC);
+        _mm_storeu_si128(reinterpret_cast<__m128i*>(&dst[i]), h);
```
0ctopus13prime, line 63: Please avoid direct casting and use load instruction.

jni/src/decoding/x86_decoding.cpp
```diff
@@ -0,0 +1,73 @@
...
+    // critical arrays are always released properly even if there's an error or early return
+    knn_jni::JNIReleaseElements release_arrays{[=]() {
+        // Release the destination FP32 array, mode 0 means changes are written back
+        jniUtil->ReleasePrimitiveArrayCritical(env, fp32Array, dst_f32, 0);
+        // Release the source FP16 array, JNI_ABORT means we don't write changes back (read-only)
+        jniUtil->ReleasePrimitiveArrayCritical(env, fp16Array, src_bytes, JNI_ABORT);
+    }};
+
+    float* dst = reinterpret_cast<float*>(dst_f32);
+    const uint16_t* src = reinterpret_cast<const uint16_t*>(src_bytes + offset);
```
0ctopus13prime, line 47: Please do an alignment check and return false if it can't be casted.

jni/src/decoding/arm_decoding.cpp
```diff
@@ -0,0 +1,62 @@
...
+    // critical arrays are always released properly even if there's an error or early return
+    knn_jni::JNIReleaseElements release_arrays{[=]() {
+        // Release the destination FP32 array, mode 0 means changes are written back
+        jniUtil->ReleasePrimitiveArrayCritical(env, fp32Array, dst_f32, 0);
+        // Release the source FP16 array, JNI_ABORT means we don't write changes back (read-only)
+        jniUtil->ReleasePrimitiveArrayCritical(env, fp16Array, src_bytes, JNI_ABORT);
+    }};
+
+    float* dst = reinterpret_cast<float*>(dst_f32);
+    const uint16_t* src = reinterpret_cast<const uint16_t*>(src_bytes + offset);
```
0ctopus13prime, line 47: Please do an alignment check and return false if it can't be casted.

src/main/java/org/opensearch/knn/jni/SIMDEncoding.java
```diff
@@ -0,0 +1,72 @@
...
+                // Load SIMD library based on support
+                if (PlatformUtils.isAVX512SPRSupportedBySystem()) {
+                    System.loadLibrary(KNNConstants.SIMD_AVX512_SPR_JNI_LIBRARY_NAME);
+                } else if (PlatformUtils.isAVX512SupportedBySystem()) {
+                    System.loadLibrary(KNNConstants.SIMD_AVX512_JNI_LIBRARY_NAME);
+                } else if (PlatformUtils.isAVX2SupportedBySystem()) {
+                    System.loadLibrary(KNNConstants.SIMD_AVX2_JNI_LIBRARY_NAME);
+                } else {
+                    System.loadLibrary(KNNConstants.SIMD_JNI_LIBRARY_NAME);
+                }
```
navneet1v: we need to move this into a common file. Since the native lib should be loaded once for encoding or decoding. The problem of keeping this only in encoding is there can be a case when encoding never happens and only decoding is happening. This happens if there is no indexing happening.

src/main/java/org/opensearch/knn/index/codec/util/KNNVectorAsCollectionOfHalfFloatsSerializer.java
```diff
@@ -0,0 +1,130 @@
...
+    /**
+     * Fallback Java implementation of float[] to byte[] conversion.
+     * Handles both little-endian and big-endian platforms.
+     *
+     * @param input  float array to convert
+     * @param output output byte array to hold FP16-encoded values
+     * @param dimension number of elements to convert
+     */
+    private void floatToByteArrayFallback(float[] input, byte[] output, int dimension) {
+        if (IS_LITTLE_ENDIAN) {
```
navneet1v: why we need little endian check here? since we are writing these files as new so we know we will always write either as little endian or big endian.

navneet1v (commented):
Overall code looks good to me. I think there some sharp corners that we need to trim off. Lets see how we can iterate faster on this PR

jni/cmake/init-simd.cmake
ManasviGoyal: @navneet1v I added these changes in this PR itself since the fix for the MACOS test was overlapping with the above. I will add the detailed documentation for SIMD code in the next PR since this PR is already quite big. Thanks!

src/main/java/org/opensearch/knn/index/codec/util/KNNIOUtils.java
```diff
@@ -0,0 +1,71 @@
...
+ * Minimal KNNIOUtils for resource and file cleanup.
+ */
+public class KNNIOUtils {
+    private KNNIOUtils() {}
+
+    /**
+     * Closes all given Closeables, suppressing all thrown Throwables in ex.
+     * Some of the Closeables may be null, they are ignored.
+     */
+    public static void closeWhileSuppressingExceptions(Throwable ex, Closeable... objects) {
```
kotwanikunal, line 23: Can we also add tests for this later with concurrency scenarios?
I just want to make sure any future work around these methods does not break any assumptions.

kotwanikunal (approved):

jni/include/decoding/decoding.h
```diff
@@ -0,0 +1,37 @@
...
+ * GitHub history for details.
+ */
+
+#ifndef OPENSEARCH_KNN_DECODING_H
+#define OPENSEARCH_KNN_DECODING_H
+
+#include "jni_util.h"
+#include <jni.h>
+namespace knn_jni {
+    namespace decoding {
```
shatejas, line 18: So encoding and decoding are inverse operations. I wouldn't want to separate them. How about we combine this into one namespace 

```
namspace knn_jni {

      namespace simd {
           jboolean isSIMDSupported();
      }

      namepace codec {
               namespace fp16 {
                        encodeFp32ToFp16()

                        decodeFp16ToFp32()
                }
     }
}
```

jni/src/decoding/arm_decoding.cpp
```diff
@@ -0,0 +1,64 @@
...
+
+    int i = 0;
+    for (; i + 8 <= count; i += 8) {
+        float16x4_t h0 = vld1_f16(reinterpret_cast<const __fp16*>(&src[i + 0]));
+        float16x4_t h1 = vld1_f16(reinterpret_cast<const __fp16*>(&src[i + 4]));
+        float32x4_t v0 = vcvt_f32_f16(h0);
+        float32x4_t v1 = vcvt_f32_f16(h1);
+        vst1q_f32(&dst[i + 0], v0);
+        vst1q_f32(&dst[i + 4], v1);
+    }
```
shatejas, line 57: Please add these comments. 
```suggestion
    // The loop processes 8 fp16 values at a time in a batch of 4. 
    for (; i + 8 <= count; i += 8) {
         // loads vector with 4 fp 16 values in a NEON register
        float16x4_t h0 = vld1_f16(reinterpret_cast<const __fp16*>(&src[i + 0]));
        float16x4_t h1 = vld1_f16(reinterpret_cast<const __fp16*>(&src[i + 4]));
        
        // converts fp16 and fp32 in parallel 
        float32x4_t v0 = vcvt_f32_f16(h0);
        float32x4_t v1 = vcvt_f32_f16(h1);
        
        //Copies the value of registers to memory location (dst)
        vst1q_f32(&dst[i + 0], v0);
        vst1q_f32(&dst[i + 4], v1);
    }

```




----
On a side note: I am wondering how did we decide on 8 elements, 4 elements would use less registers. On the contrary 16 elements would use more registers but may be faster Can you share your insights behind choosing 8?

jni/src/decoding/arm_decoding.cpp
```diff
@@ -0,0 +1,64 @@
...
+jboolean knn_jni::decoding::convertFP16ToFP32(knn_jni::JNIUtilInterface *jniUtil, JNIEnv* env, jbyteArray fp16Array, jfloatArray fp32Array, jint count, jint offset) {
+    // Return early if there's nothing to convert
+    if (count <= 0) return JNI_TRUE;
+
+    jfloat* dst_f32 = reinterpret_cast<jfloat*>(jniUtil->GetPrimitiveArrayCritical(env, fp32Array, nullptr));
+    jbyte* src_bytes = reinterpret_cast<jbyte*>(jniUtil->GetPrimitiveArrayCritical(env, fp16Array, nullptr));
+
+    // When 'release_arrays' goes out of scope, its lambda will run to ensure that the
+    // critical arrays are always released properly even if there's an error or early return
+    knn_jni::JNIReleaseElements release_arrays{[=]() {
```
shatejas, line 35: nit: I am assuming this is an equivalent of try with resources in java. Nice!

jni/src/encoding/arm_encoding.cpp
```diff
@@ -0,0 +1,68 @@
...
+        // release_arrays will cleanup the arrays automatically
+        return JNI_FALSE;
+    }
+
+    const float* src = reinterpret_cast<const float*>(src_f32);
+    uint16_t* dst = reinterpret_cast<uint16_t*>(dst_bytes);
+
+    int i = 0;
+    // ARM NEON bulk 8-wide
+    for (; i + 8 <= count; i += 8) {
```
shatejas, line 52: please add similar comments  as in decoding

jni/src/decoding/x86_decoding.cpp
```diff
@@ -0,0 +1,86 @@
...
+
+    size_t i = 0;
+#if defined(KNN_HAVE_AVX512_SPR)
+    for (; i + 32 <= count; i += 32) {
+        if (i + 128 < count) {
+            _mm_prefetch(reinterpret_cast<const char*>(&src[i + 128]), _MM_HINT_T0);
+        }
+        __m512h h = _mm512_loadu_ph(&src[i]);
```
shatejas: Since there is an alignment check above, is it safe to use `_mm512_load_ph` instead? It will be faster right? I understand this might be defensive but was wondering if the trade off is worth it

jni/src/decoding/arm_decoding.cpp
ManasviGoyal: @shatejas This is based on some experimentation I did while tuning the loop. During benchmarking, I found that processing 8 elements gave better performance than 4 likely because it allowed better pipelining of the operations without overwhelming the NEON registers. I also tested with 16 elements but didn’t see any improvement (possibly due to register pressure). 

also I plan on adding line by line comments and references to the instruction set for each SIMD instruction (already have it ready) in the next PR since this one is quite big already. Is it okay if we wait for that?

jni/src/decoding/x86_decoding.cpp
```diff
@@ -0,0 +1,86 @@
...
+        __m512 v = _mm512_cvtph_ps(h);
```
shatejas: I am a little confused here. so `_m512h` will hold 32 16 bit values (32*16=512), `_m512` should hold 16 32bit values right?  `_mm512_cvtph_ps` is probably converting the first 16 to single precision. what happens to other 16 values. there might be some gap in my understanding, can you clarify?

jni/src/decoding/x86_decoding.cpp
```diff
@@ -0,0 +1,86 @@
...
+
+    if ((reinterpret_cast<uintptr_t>(src_bytes + offset) % alignof(uint16_t)) != 0) {
+        return JNI_FALSE;  // release_arrays will still run its cleanup here
+    }
+
+    float* dst = reinterpret_cast<float*>(dst_f32);
+    const uint16_t* src = reinterpret_cast<const uint16_t*>(src_bytes + offset);
+
+    size_t i = 0;
+#if defined(KNN_HAVE_AVX512_SPR)
```
shatejas, line 50: please add comments similar to arm

jni/src/decoding/arm_decoding.cpp
shatejas: > This is based on some experimentation I did while tuning the loop. During benchmarking, I found that processing 8 elements gave better performance than 4 likely because it allowed better pipelining of the operations without overwhelming the NEON registers. I also tested with 16 elements but didn’t see any improvement (possibly due to register pressure).

Do add this in comments as well for context

> Is it okay if we wait for that?

Should be fine

jni/src/decoding/x86_decoding.cpp
ManasviGoyal: The alignment check only checks for 2-byte alignment to safely cast to `uint16_t*`.The aligned loads like `_mm512_load_ph` require strict 64-byte alignment (32-byte for `_mm256_load_ps`), which we can’t guarantee here since `GetPrimitiveArrayCritical` provides only 4-8 byte alignment. Using aligned loads here would be unsafe and can lead to undefined behavior. I think it's safer to use the unaligned versions for the minimal performance tradeoff.

jni/src/decoding/x86_decoding.cpp
ManasviGoyal: Thanks for pointing out. Yes, you are right. The upper half of the input is wrongly skipped here. I will update the code to address this by splitting the load into two `__m256h` chunks. 

I checked locally and I did have the correct version there. Might have missed it during the last push to this branch. Sorry for the confusion.

jni/include/decoding/decoding.h
ManasviGoyal: The current split was based on the suggestions before but it ultimately depends on how we want to maintain this in the future. The above suggestion makes sense to me since we can go for a similar convention with bfp16 later. Will address in the next PR since it is mainly code restructure,

src/main/java/org/opensearch/knn/jni/SIMDDecoding.class
shatejas, line 1: Is there are reason we are pushing .class files?

src/main/java/org/opensearch/knn/jni/SIMDEncoding.java
```diff
@@ -0,0 +1,38 @@
...
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ * Modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+package org.opensearch.knn.jni;
+
+public class SIMDEncoding {
```
shatejas, line 14: Similar to cpp, lets merge encoding and decoding into the same class

src/main/java/org/opensearch/knn/index/KNNSettings.java
```diff
@@ -93,10 +93,13 @@ public class KNNSettings {
     public static final String MODEL_CACHE_SIZE_LIMIT = "knn.model.cache.size.limit";
     public static final String ADVANCED_FILTERED_EXACT_SEARCH_THRESHOLD = "index.knn.advanced.filtered_exact_search_threshold";
     public static final String KNN_FAISS_AVX2_DISABLED = "knn.faiss.avx2.disabled";
+    public static final String KNN_SIMD_AVX2_DISABLED = "knn.simd.avx2.disabled";
     public static final String QUANTIZATION_STATE_CACHE_SIZE_LIMIT = "knn.quantization.cache.size.limit";
     public static final String QUANTIZATION_STATE_CACHE_EXPIRY_TIME_MINUTES = "knn.quantization.cache.expiry.minutes";
     public static final String KNN_FAISS_AVX512_DISABLED = "knn.faiss.avx512.disabled";
+    public static final String KNN_SIMD_AVX512_DISABLED = "knn.simd.avx512.disabled";
```
shatejas: Shouldn't there be one setting for avx. Its either enabled or disabled. I don't think we should separate out faiss and code paths which are non faiss. Why would users want to leverage it in one part and not do it in other parts of the system

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNN990HalfFloatFlatVectorsReader.java
```diff
@@ -0,0 +1,397 @@
...
+            @Override
+            public byte[] vectorValue(int ord) throws IOException {
+                slice.seek((long) ord * byteSize);
+                slice.readBytes(bytesBuffer, 0, bytesBuffer.length);
+                return bytesBuffer;
+            }
+
+            @Override
+            public ByteVectorValues copy() {
+                return this;
```
shatejas: Is this safe? The client might expect a copy, instead it will return the same ref

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNN990HalfFloatFlatVectorsWriter.java
```diff
@@ -0,0 +1,445 @@
...
+            OffHeapFloatVectorValues base = new OffHeapFloatVectorValues.DenseOffHeapVectorValues(
+                dim,
+                count,
+                finalVectorDataInput.slice("vector-data", 0, count * byteSize),
+                byteSize,
+                vectorsScorer,
+                fieldInfo.getVectorSimilarityFunction()
+            );
+
+            FloatVectorValues floatVectorValues = new FloatVectorValues() {
```
shatejas: This seems like the same code as in reader, I think we should pull it out and have it package-private access. That will require to create a package in KNN990Codec so no class apart from HalfFloat codecs can use it

jni/src/decoding/x86_decoding.cpp
```diff
@@ -52,9 +52,18 @@ jboolean knn_jni::decoding::convertFP16ToFP32(knn_jni::JNIUtilInterface *jniUtil
         if (i + 128 < count) {
             _mm_prefetch(reinterpret_cast<const char*>(&src[i + 128]), _MM_HINT_T0);
         }
-        __m512h h = _mm512_loadu_ph(&src[i]);
-        __m512 v = _mm512_cvtph_ps(h);
-        _mm512_storeu_ps(&dst[i], v);
+
+        // Load and convert first 16 FP16 values
```
shatejas, line 56: Is the processing of 2 chunks at a time based on experiments, I see Avx512 below processed in 1 chunks of 16, so was curious

ManasviGoyal: Yes. the avx512 spr case supports native FP16 instructions, so we load and convert two chunks per iteration to fully utilize the 512-bit FP32 registers. But for the avx512 case, we operate on packed 16-bit integers and don’t have native FP16 hardware. I did experiment with processing two chunks per iteration for this but saw no performance gain. Based on my understanding, this is likely because the conversion isn’t pipelined in hardware for avx512 case so batching just increases register pressure without improving performance.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/KNN990HalfFloatFlatVectorsReader.java
ManasviGoyal: Most of the code is adapted from Lucene. This is mirroring what Lucene [ByteVectorValues](https://github.com/apache/lucene/blob/fa6c87c4dfaa277d0f3da43d7118bf8b6775f2ac/lucene/core/src/java/org/apache/lucene/index/ByteVectorValues.java#L107-L109) does. It also returns the same ref not a deep copy. I think in Lucene the assumption is that vector access occurs in single-threaded contexts where this is safe.

shatejas: Thats the default implementation probably to have backward compatibility and not break users of lucene, here you are using base of OffheapVectorValues which returns a new object creating a new reference https://github.com/apache/lucene/blob/fa6c87c4dfaa277d0f3da43d7118bf8b6775f2ac/lucene/core/src/java/org/apache/lucene/codecs/lucene95/OffHeapByteVectorValues.java#L145. Ideally you want to call the copy function of the base and then create a new instance in this

generally its not a good idea to deviate from a behavior, It breaks https://en.wikipedia.org/wiki/Liskov_substitution_principle. It may have unknown side effects if the user does not check the implementation of an interface

Issue #2816, opened 2025-08-03
[FEATURE] Add GRPC support for k-NN queries


karenyrx:
## Is your feature request related to a problem? Please describe.
Currently, the k-NN plugin only supports REST API queries. With OpenSearch moving transport-grpc from a plugin to a core module ([PR #18897](https://github.com/opensearch-project/OpenSearch/pull/18897)), and real world benchmarking results (which will be published in a separate  future issue) showing a sizeable performance gains for GRPC KNN over HTTP queries, there's an opportunity to provide native GRPC support for k-NN queries.

## Describe the solution you'd like
Add native GRPC support for k-NN queries, enabling clients to perform vector searches using Protocol Buffers over GRPC transport instead of REST APIs.

Expected performance benefits include: 
- **Reduced Serialization Overhead**: Binary protocol buffers vs JSON
- **Lower Network Latency**: HTTP/2 multiplexing vs HTTP/1.1
- **Better Throughput**: Connection reuse and streaming capabilities
- **Type Safety**: Compile-time validation vs runtime JSON parsing
- **Streaming support**: Potential for streaming large result sets 

## Describe alternatives you've considered
N/A

###  Milestones  
- [x] Core PR to allow the transport-GRPC to be extensible 
- [ ] k-NN PR to implement the KNN GRPC Support by extending the transport-grpc plugin 
- [ ] Add integration and BWC tests for GRPC KNN  
- [ ] Add user documentation for GRPC k-NN queries, under https://docs.opensearch.org/latest/api-reference/grpc-apis/search/
- [ ] Publish issue benchmarking results of GRPC KNN compared to REST API

karenyrx:
Please feel free to assign this to me!

PR #2817, opened 2025-08-03
[GRPC] Extend transport-grpc module to support GRPC KNN queries

karenyrx:
### Description

1. Extend the `transport-grpc` module to support GRPC-based KNN queries. 
2. Upgrade the guava version to avoid dependency conflicts with the opensearch-protobufs library. 

Note: This PR is dependent on [OpenSearch PR #18897](https://github.com/opensearch-project/OpenSearch/pull/18897) moving the `transport-grpc` plugin to a module.

### Future enhancements
In a future PR, an integration test for GRPC KNN will be added. This requires an `OpenSearchGrpcTestCase` test fixture is added to OpenSearch core, similar to the existing [OpenSearchRestTestCase](https://github.com/opensearch-project/OpenSearch/blob/main/test/framework/src/main/java/org/opensearch/test/rest/OpenSearchRestTestCase.java). 

### Related Issues
Part of #2816 

### Check List


src/main/java/org/opensearch/knn/grpc/proto/request/search/query/KNNQueryBuilderProtoUtils.java
```diff
@@ -0,0 +1,180 @@
...
+ * into their corresponding OpenSearch KNNQueryBuilder implementations for search operations.
+ */
+public class KNNQueryBuilderProtoUtils {
+
+    // Registry for query conversion
+    private static QueryBuilderProtoConverterRegistry REGISTRY = new QueryBuilderProtoConverterRegistry();
+
+    private KNNQueryBuilderProtoUtils() {
+        // Utility class, no instances
+    }
```
navneet1v: we can use ~https://github.com/Utility~ @Utility annotation of Lombok here

src/main/java/org/opensearch/knn/grpc/proto/request/search/query/KNNQueryBuilderProtoUtils.java
```diff
@@ -0,0 +1,180 @@
...
+                        javaValue = value.getString();
+                        break;
+                    case BOOL:
+                        javaValue = value.getBool();
+                        break;
+                    // Handle other types as needed
+                    default:
+                        // Skip unsupported types
+                        continue;
+                }
```
navneet1v: this seems to be a good candidate for commons. Can we move this in https://github.com/opensearch-project/opensearch-protobufs ?

navneet1v:
> In a future PR, an integration test for GRPC KNN will be added. This requires an `OpenSearchGrpcTestCase` test fixture is added to OpenSearch core, similar to the existing [OpenSearchRestTestCase](https://github.com/opensearch-project/OpenSearch/blob/main/test/framework/src/main/java/org/opensearch/test/rest/OpenSearchRestTestCase.java).

Can we create a GH issue for this task. This will ensure that effort to add ITs is not lost.

We should also add the BWC test for the same.

src/main/java/org/opensearch/knn/grpc/proto/request/search/query/KNNQueryBuilderProtoUtils.java
```diff
@@ -0,0 +1,180 @@
...
+            } else if (rescoreProto.getKnnQueryRescoreCase() == KnnQueryRescore.KnnQueryRescoreCase.CONTEXT) {
+                // If context is provided, create a RescoreContext with the specified oversample factor
+                org.opensearch.protobufs.RescoreContext contextProto = rescoreProto.getContext();
+                if (contextProto.hasOversampleFactor()) {
+                    builder.rescoreContext(RescoreContext.builder().oversampleFactor(contextProto.getOversampleFactor()).build());
+                } else {
+                    builder.rescoreContext(RescoreContext.getDefault());
+                }
+            }
+        }
```
navneet1v: [minor] Let's just move this to a separate private function to ensure methods are not getting too long. Please see if we can also move some other parts of this `fromProto` function in smaller private functions

navneet1v (commented):
Mostly code looks good to me. Some minor comments. 

Also adding the discussion I had with @finnegancarroll here:
1. The protobuff endpoint will be present for newer OpenSearch clusters.
2. If there is a mix cluster case then responsibility of sending the protobuff based search request to right nodes is on operator.
3. Node to Node communication is still happing via REST.

src/main/java/org/opensearch/knn/grpc/proto/request/search/query/KNNQueryBuilderProtoConverter.java
```diff
@@ -0,0 +1,32 @@
...
+import org.opensearch.index.query.QueryBuilder;
+import org.opensearch.transport.grpc.proto.request.search.query.QueryBuilderProtoConverter;
+import org.opensearch.protobufs.QueryContainer;
+
+/**
+ * Converter for KNN queries.
+ * This class implements the QueryBuilderProtoConverter interface to provide KNN query support
+ * for the gRPC transport plugin.
+ */
+public class KNNQueryBuilderProtoConverter implements QueryBuilderProtoConverter {
```
shatejas, line 17: Instead can KNNQueryBuilder directly implement the converter. The reason is that there are similar convertors `toXContent` and `fromXcontent`. Its easy to skip making changes when there is an interface change if there is a separate class. The core logic can still be in utils

Let me know your thoughts

navneet1v: +1 on this.

src/main/java/org/opensearch/knn/grpc/proto/request/search/query/KNNQueryBuilderProtoUtils.java
karenyrx: could you elaborate how this private constructor would benefit from Utility?

src/main/java/org/opensearch/knn/grpc/proto/request/search/query/KNNQueryBuilderProtoUtils.java
karenyrx: the opensearch-protobufs repo does not handle protobuf parsing; it only stores raw protobuf files

src/main/java/org/opensearch/knn/grpc/proto/request/search/query/KNNQueryBuilderProtoConverter.java
karenyrx: it's currently following the pattern in core, for all other query converters: https://github.com/opensearch-project/OpenSearch/tree/main/modules/transport-grpc/src/main/java/org/opensearch/plugin/transport/grpc/proto/request/search/query 

the reason for not touching the original KNNQueryBuilder is because in core, these QueryBuilders are located in  the `/server` folder rather than the `/modules` folder where the transport-grpc plugin is located

ideally, to catch differences in toXContent vs. toProto and fromXContent vs fromProto, we can have some unit tests which will fail when one is changed without corresponding changes to the other 

the javadoc in the KNNPQueryBuilderProtoConverter file will also reference the KNNQueryBuilder#fromXContent so that is a soft link which maybe developers will check via IDE 

if it's a strong concern, i can have KNNQueryBuilder implement QueryBuilderProtoConverter direcrly, or we can just keep it as is to match the pattern in core currently. what would your preferences be?

karenyrx:
Latest changes require https://github.com/opensearch-project/OpenSearch/pull/18923/file to be merged

src/main/java/org/opensearch/knn/grpc/proto/request/search/query/KNNQueryBuilderProtoUtils.java
navneet1v: @karenyrx i think github add a wierd link here. So what I mean to say was @Utility annotation of Lombok will ensure that you don't have to write
```
private KNNQueryBuilderProtoUtils() {
        // Utility class, no instances
    }
```

src/main/java/org/opensearch/knn/grpc/proto/request/search/query/KNNQueryBuilderProtoUtils.java
navneet1v: In that case it will be good to move this in a private function.

navneet1v:
@karenyrx some minor comments, I think PR is ready for merge once we have successful CI runs.

src/main/java/org/opensearch/knn/grpc/proto/request/search/query/KNNQueryBuilderProtoUtils.java
karenyrx: updated

src/main/java/org/opensearch/knn/grpc/proto/request/search/query/KNNQueryBuilderProtoUtils.java
karenyrx: updated

src/main/java/org/opensearch/knn/grpc/proto/request/search/query/KNNQueryBuilderProtoUtils.java
karenyrx: done

finnroblin:
CI failures are likely related to PR, since [other recent CI does not fail](https://github.com/opensearch-project/k-NN/pull/2830) (has been running for >30 mins, whereas ITs fail in ~10 mins on this PR). Fails on [SPI load](https://github.com/opensearch-project/k-NN/actions/runs/16766002665/job/47473170575?pr=2817#step:7:3040):
```
»  java.util.ServiceConfigurationError: org.apache.lucene.codecs.KnnVectorsFormat: Provider org.opensearch.knn.index.codec.KNN990Codec.NativeEngines990KnnVectorsFormat could not be instantiated
»  	at java.base/java.util.ServiceLoader.fail(ServiceLoader.java:586) ~[?:?]
»  	at java.base/java.util.ServiceLoader$ProviderImpl.newInstance(ServiceLoader.java:813) ~[?:?]
»  	at java.base/java.util.ServiceLoader$ProviderImpl.get(ServiceLoader.java:729) ~[?:?]
»  	at java.base/java.util.ServiceLoader$3.next(ServiceLoader.java:1403) ~[?:?]
»  	at org.apache.lucene.util.NamedSPILoader.reload(NamedSPILoader.java:68) ~[lucene-core-10.2.2.jar:10.2.2 279eb7aaafe985e5d0552b7f2a10b63185a3f893 - 2025-06-17 09:30:59]
»  	at org.apache.lucene.codecs.KnnVectorsFormat.reloadKnnVectorsFormat(KnnVectorsFormat.java:82) ~[lucene-core-10.2.2.jar:10.2.2 279eb7aaafe985e5d0552b7f2a10b63185a3f893 - 2025-06-17 09:30:59]
»  	at org.opensearch.plugins.PluginsService.reloadLuceneSPI(PluginsService.java:842) ~[opensearch-3.2.0-SNAPSHOT.jar:3.2.0-SNAPSHOT]
»  	at org.opensearch.plugins.PluginsService.loadBundle(PluginsService.java:795) ~[opensearch-3.2.0-SNAPSHOT.jar:3.2.0-SNAPSHOT]
»  	at org.opensearch.plugins.PluginsService.loadBundles(PluginsService.java:615) ~[opensearch-3.2.0-SNAPSHOT.jar:3.2.0-SNAPSHOT]
```

finnroblin:
I'm also thinking that the BWC tests may break on this change. I recall @finnegancarroll mentioning failing BWC tests for his gRPC changes, but I'm unsure of the specifics.

navneet1v:
@karenyrx 

The error is this actually which is present here https://github.com/opensearch-project/k-NN/actions/runs/16766002665/job/47473170575?pr=2817

Caused by: java.lang.NoClassDefFoundError: com/google/common/collect/ImmutableList
»  	at org.opensearch.knn.common.featureflags.KNNFeatureFlags.getFeatureFlags(KNNFeatureFlags.java:39)
»  	at org.opensearch.knn.index.KNNSettings.<clinit>(KNNSettings.java:549)
»  	at org.opensearch.knn.index.codec.KNN990Codec.NativeEngines990KnnVectorsFormat.<init>(NativeEngines990KnnVectorsFormat.java:50)
»  	at org.opensearch.knn.index.codec.KNN990Codec.NativeEngines990KnnVectorsFormat.<init>(NativeEngines990KnnVectorsFormat.java:42)
»  	at java.base/jdk.internal.reflect.DirectConstructorHandleAccessor.newInstance(DirectConstructorHandleAccessor.java:62)
»  	at java.base/java.lang.reflect.Constructor.newInstanceWithCaller(Constructor.java:502)
»  	at java.base/java.lang.reflect.Constructor.newInstance(Constructor.java:486)
»  	at java.base/java.util.ServiceLoader$ProviderImpl.newInstance(ServiceLoader.java:789)
»  	... 20 more
»  Caused by: java.lang.ClassNotFoundException: com.google.common.collect.ImmutableList
»  	at java.base/java.net.URLClassLoader.findClass(URLClassLoader.java:445)
»  	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:593)
»  	at java.base/java.net.FactoryURLClassLoader.loadClass(URLClassLoader.java:872)
»  	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:526)
»  	... 28 more

which is just below the the KNNVectorFormat SPI issue.

Looks like the new google guava dependency you added doesn't have the ImmutableList class in it. Hence the constructor is failing. which fails the SPI. Please check in the version you added for google guava has the right classes. This should be easily reproducible in local via `./gradlew run` after clearing the gradle cache.

build.gradle
```diff
@@ -320,6 +320,14 @@ opensearchplugin {
     noticeFile = rootProject.file('NOTICE.txt')
 }
 
+// Force bundle Guava dependencies
+tasks.named("bundlePlugin").configure {
+    from(configurations.runtimeClasspath) {
+        include "guava-*.jar"
+        include "failureaccess-*.jar"
+    }
+}
```
navneet1v: this can potentially break other plugins. can we just try upgrading guava dependency to see if that is not present in plugin bundle.

navneet1v:
> It looks the addition of these compileOnly GRPC dependencies are causing the guava api dependencies to not be added
> 
> ```
> compileOnly "org.opensearch.plugin:transport-grpc:${opensearch_version}"
> compileOnly "org.opensearch:protobufs:0.6.0"  // Changed from implementation
> ```
> 
> The issue might be that when we have conflicting dependency scopes, Gradle's dependency resolution changes the bundling behavior. Continuing to investigate.

Thanks. that is my guess too ref: https://github.com/opensearch-project/k-NN/pull/2817#discussion_r2257761598

karenyrx:
Narrowed it down- looks like the addition of this compileOnly GRPC depedency is causing the guava api dependencies to not be added:

```
compileOnly "org.opensearch.plugin:transport-grpc:${opensearch_version}"
```
This seems to be because: 
* transport-grpc module itself has Guava as a dependency
* When Gradle sees transport-grpc as compileOnly, it assumes Guava will be provided by the runtime environment
* This changes Gradle's dependency resolution strategy for conflicting dependencies
Gradle excludes api Guava from bundling because it thinks it's already provided

We have 2 options to fix this: 
1. Force guava dep 
```
// Force bundle Guava dependencies
tasks.named("bundlePlugin").configure {
    from(configurations.runtimeClasspath) {
        include "guava-*.jar"
        include "failureaccess-*.jar"
    }
}
```
2. Change `transport-grpc` from a compileOnly to an implementation dependency 
```
implementation "org.opensearch.plugin:transport-grpc:${opensearch_version}"
```

build.gradle
```diff
@@ -336,10 +336,20 @@ task release(type: Copy, group: 'build') {
 //****************************************************************************/
 // Dependencies
 //****************************************************************************/
+// Force resolution of conflicting dependencies for tests
+configurations.matching { it.name.contains('test') || it.name.contains('Test') }.all {
+    resolutionStrategy {
+        force 'com.google.guava:guava:32.1.3-jre'
+        force 'com.google.guava:failureaccess:1.0.1'
+    }
+}
```
navneet1v: It will be better to add why we forcing this dependency for test only? what if it fails in actual distribution build? What mechanisms we have used to ensure that doesn't happen. any special handling for `test` here just make it hard in future to rely on the credibility of the tests running in CI

karenyrx: In production/runtime, it should use the transport-grpc dependency from core (rather than the compileOnly dependency added here). Since core already uses a different guava version from KNN already, and that does not seem to have issues, I think this should fall into that case as well

CHANGELOG.md
```diff
@@ -4,4 +4,27 @@ All notable changes to this project are documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html). See the [CONTRIBUTING guide](./CONTRIBUTING.md#Changelog) for instructions on how to add changelog entries.
 
+## [Unreleased 3.2](https://github.com/opensearch-project/k-NN/compare/main...HEAD)
```
andrross: Assuming this is planned to go in 3.2, then this entry should go in the [release notes](https://github.com/opensearch-project/k-NN/blob/main/release-notes/opensearch-k-NN.release-notes-3.2.0.0.md) as opposed to the changelog

finnroblin: Yes, I will move to release notes in a follow-up PR. I had a maintainer resolve the changelog merge conflicts to unblock CI.

karenyrx: Oops just moved it. Should I revert it?

karenyrx: Reverted.

finnroblin: That's fine to add to release notes, will ask maintainer to rerun CI. Every commit triggers rerun.

finnroblin: Could you please undo the revert @karenyrx ? will save one release notes PR

karenyrx: Done

build.gradle
karenyrx: I can change it to resolve for all too: 
```
// Force resolution of conflicting dependencies
configurations.all {
    resolutionStrategy {
        force 'com.google.guava:guava:32.1.3-jre'
        force 'com.google.guava:failureaccess:1.0.1'
    }
}
```
or just compile + tests: 
```
// Force resolution of conflicting dependencies for compile and test configurations
configurations.matching {
    it.canBeResolved && (
        it.name.toLowerCase().contains('compile') ||
        it.name.toLowerCase().contains('test')
    ) && !it.name.toLowerCase().contains('runtime')
}.all {
    resolutionStrategy {
        force 'com.google.guava:guava:32.1.3-jre'        // Override transport-grpc's 33.2.1-jre
        force 'com.google.guava:failureaccess:1.0.1'     // Override transport-grpc's 1.0.2
    }
}
```
@navneet1v @finnegancarroll wdyt?

navneet1v:
@karenyrx is the PR to fix the dependency of grpc plugin is merged? the latest failure tells me that the change of grpc is not merged as of now

navneet1v:
@karenyrx pasting the followup plan as suggested on the slack.


Followup I plan to take on after this PR is merged:
1. Short term: Make sure all the plugins that are dependent on KNN ( e.g. ML commons, Security Analytics, and Neural Search) can still be built + run.
   1. If they are using guava as a compileOnly dependency, they will have to be changed to be an implementation instead.
    2. Please correct if I missed any plugins that are dependent on KNN
2. Medium term: Figure out how to write integration + BWC tests for this
3. Long term: Migrate to use the transport-grpc SPI approach.
Benefit: No more unecessary deps introduced to KNN plugin by the transport-grpc plugin (e.g. guava)
4. Optional: Up to KNN maintainers whether the guava dep should be removed from KNN plugin or not - which would require a migration of all the Immutable collections to the JDK ones


cc: @vamshin , @shatejas , @finnegancarroll , @andrross

navneet1v:
@shatejas , @jmazanec15 can one of you take a look at this PR?

owaiskazi19:
Neural search build is failing
```

> Task :run FAILED
Exec output and error:
| Output for ./bin/opensearch-plugin:-> Installing file:/Users/kazabdu/.gradle/caches/modules-2/files-2.1/org.opensearch.plugin/opensearch-job-scheduler/3.2.0.0-SNAPSHOT/5d7eb58e5c0fbdf4796b61fb200f17aac7b76825/opensearch-job-scheduler-3.2.0.0-SNAPSHOT.zip
| -> Downloading file:/Users/kazabdu/.gradle/caches/modules-2/files-2.1/org.opensearch.plugin/opensearch-job-scheduler/3.2.0.0-SNAPSHOT/5d7eb58e5c0fbdf4796b61fb200f17aac7b76825/opensearch-job-scheduler-3.2.0.0-SNAPSHOT.zip
| -> Installed opensearch-job-scheduler with folder name opensearch-job-scheduler
| -> Installing file:/Users/kazabdu/.gradle/caches/modules-2/files-2.1/org.opensearch.plugin/opensearch-knn/3.2.0.0-SNAPSHOT/58e1b6a3e8e22b6ecf6a3f82cef5829a73a21ba2/opensearch-knn-3.2.0.0-SNAPSHOT.zip
| -> Downloading file:/Users/kazabdu/.gradle/caches/modules-2/files-2.1/org.opensearch.plugin/opensearch-knn/3.2.0.0-SNAPSHOT/58e1b6a3e8e22b6ecf6a3f82cef5829a73a21ba2/opensearch-knn-3.2.0.0-SNAPSHOT.zip
| -> Failed installing file:/Users/kazabdu/.gradle/caches/modules-2/files-2.1/org.opensearch.plugin/opensearch-knn/3.2.0.0-SNAPSHOT/58e1b6a3e8e22b6ecf6a3f82cef5829a73a21ba2/opensearch-knn-3.2.0.0-SNAPSHOT.zip
| -> Rolling back opensearch-job-scheduler
| -> Rolled back opensearch-job-scheduler
| -> Rolling back file:/Users/kazabdu/.gradle/caches/modules-2/files-2.1/org.opensearch.plugin/opensearch-knn/3.2.0.0-SNAPSHOT/58e1b6a3e8e22b6ecf6a3f82cef5829a73a21ba2/opensearch-knn-3.2.0.0-SNAPSHOT.zip
| -> Rolled back file:/Users/kazabdu/.gradle/caches/modules-2/files-2.1/org.opensearch.plugin/opensearch-knn/3.2.0.0-SNAPSHOT/58e1b6a3e8e22b6ecf6a3f82cef5829a73a21ba2/opensearch-knn-3.2.0.0-SNAPSHOT.zip
| Exception in thread "main" java.lang.IllegalArgumentException: Missing plugin [transport-grpc], dependency of [opensearch-knn]
|       at org.opensearch.plugins.PluginsService.addSortedBundle(PluginsService.java:596)
|       at org.opensearch.plugins.PluginsService.sortBundles(PluginsService.java:559)
|       at org.opensearch.plugins.PluginsService.checkJarHellForPlugin(PluginsService.java:403)
|       at org.opensearch.tools.cli.plugin.InstallPluginCommand.jarHellCheck(InstallPluginCommand.java:834)
|       at org.opensearch.tools.cli.plugin.InstallPluginCommand.loadPluginInfo(InstallPluginCommand.java:811)
|       at org.opensearch.tools.cli.plugin.InstallPluginCommand.installPlugin(InstallPluginCommand.java:846)
|       at org.opensearch.tools.cli.plugin.InstallPluginCommand.execute(InstallPluginCommand.java:277)
|       at org.opensearch.tools.cli.plugin.InstallPluginCommand.execute(InstallPluginCommand.java:251)
|       at org.opensearch.common.cli.EnvironmentAwareCommand.execute(EnvironmentAwareCommand.java:110)
|       at org.opensearch.cli.Command.mainWithoutErrorHandling(Command.java:138)
|       at org.opensearch.cli.MultiCommand.execute(MultiCommand.java:104)
|       at org.opensearch.cli.Command.mainWithoutErrorHandling(Command.java:138)
|       at org.opensearch.cli.Command.main(Command.java:101)
|       at org.opensearch.tools.cli.plugin.PluginCli.main(PluginCli.java:66)


```

karenyrx:
Confirmed that the neural-search `./gradlew check` completed successfully:
```
BUILD SUCCESSFUL in 21m 24s
91 actionable tasks: 91 executed
``` 
The ongoing opensearch-build RCs are also succeeding so far: https://build.ci.opensearch.org/blue/organizations/jenkins/distribution-build-opensearch/detail/distribution-build-opensearch/11325/pipeline
[https://build.ci.opensearch.org/blue/organizations/jenkins/distribution-build-opensearch[…]il/distribution-build-opensearch-dashboards/8587/pipeline](https://build.ci.opensearch.org/blue/organizations/jenkins/distribution-build-opensearch-dashboards/detail/distribution-build-opensearch-dashboards/8587/pipeline)

The error above is potentially a caching issue where local Gradle caches were still using an older version of opensearch core where [transport-grpc had not been moved to a module](https://github.com/opensearch-project/OpenSearch/pull/18897/files) yet. After the core PR to move the transport-grpc plugin to a module , transport-grpc should be auto-installed/bundled with all opensearch core distributions, so it should not be unavailable. `./gradlew clean && rm -rf ~/.gradle && rm -rf ~/.m2` may be needed.

Issue #2796, opened 2025-07-17
[RFC] Support GPU indexing FP16, byte and binary index.


0ctopus13prime:
# [RFC] Support FP16, Binary in GPU

# Abstract

The current GPU indexing pipeline only supports FP32, and lacks support for both FP16 and binary vectors. To address this limitation, we need to add support for these two formats. Additionally, the implementation should be compatible with memory-optimized search to ensure efficiency.

# Goal, Scope
* Only HNSW is supported, supporting IVF will be discussed later.
* Support FP16
    * Build FP16 index via GPU indexing pipeline
    * LuceneOnFaiss should be able to perform vector search on the built index.
* Binary Index
    * Build 4x, 8x, 16x and 32x quantized index via GPU indexing pipeline
    * LuceneOnFaiss should be able to perform vector search on the built index.

0ctopus13prime:
Status update:

Verified that it’s working for both FP16 and byte vector types (PR for byte is merged, PR for FP16 is not yet raised). Binary is the only remaining index type to test, which I'm currently working on with this [repo](https://github.com/tarang-jain/faiss/tree/binary-cagra-nnd). Fortunately, it looks like supporting all three types in remote build won’t require significant changes on the OpenSearch KNN side. I plan to wrap up all code changes—including tests—in a single PR. That PR will likely focus more on testings, with relatively minimal code changes.

Roughly, my PR plans are below:
1. Finish PR for binary in remote-vector-index-builder.
2. Finish PR for FP16 in remote-vector-index-builder.
3. Merge PR in knn to feature branch `dedup-vectors-optimization`
4. Bump up Faiss commit point in KNN to include all three types GPU indexing support.
5. Merge feature branch `dedup-vectors-optimization` into main.

navneet1v:
@0ctopus13prime great progress on this. Excited to see the benchmark runs

0ctopus13prime:
Similar patch like [this](https://github.com/opensearch-project/k-NN/blob/main/jni/patches/faiss/0005-Custom-patch-to-support-multi-vector-IndexHNSW-search_level_0.patch) should be made for [IndexBinaryHNSWCagra](https://github.com/facebookresearch/faiss/blob/580f2f9246686341f9933f7e4d479dc85bc0768b/faiss/IndexBinaryHNSW.h#L80) for nested search in binary.

0ctopus13prime:
[...]

PR #2815, opened 2025-08-02
Bumping up Faiss from 0cbc2a885cde923d80c4bf9c9d6f4d81665f3f64 to 2929bf471bcfbc161e6e97356d33582af29a6c05

0ctopus13prime:
### Description
Bumping up Faiss commit rom 0cbc2a885cde923d80c4bf9c9d6f4d81665f3f64 to 2929bf471bcfbc161e6e97356d33582af29a6c05.

This is required to have Faiss GPU support for FP16, Byte and Binary. 

### Related Issues
Resolves #[Issue number to be closed when this PR is merged]
N/A

### Check List
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
- [ ] Commits are signed per the DCO using `--signoff`.
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/k-NN/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #2819, opened 2025-08-04
Support GPU indexing for FP16, Byte and Binary.

0ctopus13prime:
### Description
RFC:  https://github.com/opensearch-project/k-NN/issues/2796

This PR introduces changes to enable GPU support for FP16, Byte, and Binary indexing. 

The key updates are as follows:

1. Support in FaissHNSWMethod
This change removes the restriction that previously prevented GPU indexing for FP16, Byte, and Binary vector types. With this update, Faiss HNSW indexing is consistently allowed on GPU across all supported types.

2. Quantized KNN Vector Values
A new abstraction is introduced to represent quantized vector values, intended for serialization as byte[] and upload to remote storage (e.g., S3). These values are later consumed by the remote index builder to construct Cagra2 binary HNSW indices. The implementation wraps FloatVectorValues alongside a QuantizationState, which performs quantization and returns the resulting byte representation.

3. LuceneOnFaiss Enhancements for Cagra2
To support FP16 and Cagra2 indexing, a new numeric_type field is added. Additionally, support for a new index type targeting Faiss Cagra Binary is introduced. LuceneOnFaiss is extended to handle these changes, enabling memory-efficient search over Cagra2-based indices.

### Related Issues

https://github.com/opensearch-project/k-NN/issues/2796

### Check List


0ctopus13prime:
Bumping up Faiss [PR](https://github.com/opensearch-project/k-NN/pull/2815) should be merged before this.

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/QuantizedKNNVectorValuesSupplier.java
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+package org.opensearch.knn.index.codec.nativeindex.remote;
+
+import org.opensearch.knn.index.codec.nativeindex.model.BuildIndexParams;
+import org.opensearch.knn.index.vectorvalues.KNNVectorValues;
+import org.opensearch.knn.index.vectorvalues.QuantizedKNNBinaryVectorValues;
+
+import java.util.function.Supplier;
+
```
0ctopus13prime: Will add Java doc soon.

src/main/java/org/opensearch/knn/index/vectorvalues/QuantizedKNNBinaryVectorValues.java
```diff
@@ -0,0 +1,48 @@
...
+ */
+
+package org.opensearch.knn.index.vectorvalues;
+
+import org.opensearch.knn.index.codec.nativeindex.IndexBuildSetup;
+import org.opensearch.knn.index.codec.nativeindex.QuantizationIndexUtils;
+import org.opensearch.knn.index.codec.nativeindex.model.BuildIndexParams;
+
+import java.io.IOException;
+
```
0ctopus13prime, line 13: Will add Java doc soon.

src/main/java/org/opensearch/knn/index/KNNIndexShard.java
```diff
@@ -74,15 +74,6 @@ public KNNIndexShard(IndexShard indexShard) {
...
 
-    /**
-     * Return the underlying IndexShard
-     *
-     * @return IndexShard
-     */
-    public IndexShard getIndexShard() {
-        return indexShard;
-    }
-
```
jed326, line 85: Are we just removing this because it's unused?

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/QuantizedKNNVectorValuesSupplier.java
```diff
@@ -0,0 +1,27 @@
...
+    private final BuildIndexParams indexInfo;
+
+    public QuantizedKNNVectorValuesSupplier(final BuildIndexParams indexInfo) {
+        this.orgSupplier = indexInfo.getKnnVectorValuesSupplier();
+        this.indexInfo = indexInfo;
+    }
+
+    @Override
+    public KNNVectorValues<?> get() {
+        return new QuantizedKNNBinaryVectorValues(orgSupplier.get(), indexInfo);
```
jed326: Any reason why we don't pass only `indexInfo` here?

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
```diff
@@ -331,17 +352,18 @@ static RemoteBuildRequest buildRemoteBuildRequest(
...
             default -> throw new IllegalArgumentException(
                 "Repository type " + repositoryType + " is not supported by the remote build service"
             );
         }
-        String vectorDataType = indexInfo.getVectorDataType().getValue();
 
-        KNNVectorValues<?> vectorValues = indexInfo.getKnnVectorValuesSupplier().get();
+        final String vectorDataType = determineVectorDataType(indexInfo.getVectorDataType(), parameters);
+
+        KNNVectorValues<?> vectorValues = decorateVectorValuesSupplier(indexInfo).get();
```
jed326, line 367: So we only use this to get the dimension here, do we still need to decorate it for that?

src/main/java/org/opensearch/knn/index/engine/faiss/FaissHNSWMethod.java
```diff
@@ -200,14 +195,131 @@ public RemoteIndexParameters createRemoteIndexingParameters(Map<String, Object>
      */
     @SuppressWarnings("unchecked")
     static boolean supportsRemoteIndexBuild(Map<String, Object> parameters) {
+        if (isFloat32Index(parameters)) {
```
jed326: A few thoughts on this:
1. The `isFloat32Index` name is a little misleading, as it's really evaluating the encoding *if* it's a Float32 index.
2. In each function we are re-checking the `VectorDataType` -- this feels a little awkward, because if it's a byte index then this logic would have been executed 5 times:
```
            final String dataType = getStringFromMap(parameters, VECTOR_DATA_TYPE_FIELD);
            final VectorDataType vectorDataType = VectorDataType.get(dataType);
            if (vectorDataType != VectorDataType.FLOAT) {
                return false;
            }
```
3. In addition to [2], should we just have a switch statement on `VectorDataType` instead?
4. What is actually not supported at this point? It seems like it might just be easier to check for that instead.

src/main/java/org/opensearch/knn/index/engine/faiss/FaissHNSWMethod.java
```diff
@@ -200,14 +195,131 @@ public RemoteIndexParameters createRemoteIndexingParameters(Map<String, Object>
...
+                return false;
+            }
+
+            // Check encoding is 'flat'
+            final Map<String, Object> innerMap = (Map<String, Object>) parameters.get(PARAMETERS);
+            final Map<String, Object> encoderMap = (Map<String, Object>) innerMap.get(METHOD_ENCODER_PARAMETER);
+            final String encoder = getStringFromMap(encoderMap, NAME);
+            return encoder.equals(ENCODER_FLAT);
+        } catch (final Exception e) {
+            // Ignore
```
jed326, line 240: Should we log the exception?

jed326 (commented):
I see there are no ITs on this PR, is that coming later? I would have thought we could at least add the remote build annotation to a few more tests after this PR.

src/main/java/org/opensearch/knn/index/KNNIndexShard.java
0ctopus13prime: I've removed it as we have @Getter already after I saw warning in Intellij.

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/QuantizedKNNVectorValuesSupplier.java
0ctopus13prime: QuantizedKNNBinaryVectorValues needs both original KNNVectorValues which is expected to return float[] along with `BuildIndexParams` which is needed for quantization.

In short, QuantizedKNNBinaryVectorValues will pull float[] from the given original vector, then use quantization state to apply compression upon it and return byte[] as a result.

src/main/java/org/opensearch/knn/index/KNNIndexShard.java
```diff
@@ -155,8 +146,14 @@ public void warmup() throws IOException {
...
                 final List<EngineFileContext> engineFileContexts = getAllEngineFileContexts(loadedFieldNames, leafReaderContext);
                 warmUpOffHeapIndex(engineFileContexts, directory);
-                log.info("[KNN] Loaded off-heap indices for fields {}", engineFileContexts.stream().map(ctx -> ctx.fieldName));
+                log.info(
+                    "[KNN] Loaded off-heap indices for fields {}",
+                    engineFileContexts.stream().map(ctx -> ctx.fieldName).collect(Collectors.toSet())
+                );
             }
+        } catch (Exception e) {
+            log.error("Failed warm-up", e);
```
navneet1v: why we are logging and throwing the exception here? The same exception will be logged twice and will create more confusion. If want to improve the logging I would say we catch the exception then wrap the exception with SomeKNNWarmupException and then throw it.

Also the log message added is not useful. Please add more details like which index failed to load and what shard failed.

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/DefaultVectorRepositoryAccessor.java
```diff
@@ -69,7 +69,8 @@ public void writeToRepository(
         initializeVectorValues(knnVectorValues);
         long vectorBlobLength = (long) knnVectorValues.bytesPerVector() * totalLiveDocs;
 
-        if (blobContainer instanceof AsyncMultiStreamBlobContainer asyncBlobContainer) {
+        // TODO : Once Lucene patch https://github.com/apache/lucene/issues/14992 is merged, remove vector data type check in condition.
```
navneet1v, line 72: Lets create a GH issue on k-NN side for this.

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
0ctopus13prime: `decorateVectorValuesSupplier` is needed for quantizing float vectors. 
It will decorate depend on the current context, and selectively decorate `indexInfo.getKnnVectorValuesSupplier()` to do quantization upon underlying float vectors.
So yeah, not only dimension, but it is about wrapping existing vector stream to have quantization layer on it.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissMemoryOptimizedSearcherFactory.java
```diff
@@ -33,6 +35,8 @@ public VectorSearcher createVectorSearcher(final Directory directory, final Stri
             // Try load it. Not all FAISS index types are currently supported at the moment.
             return new FaissMemoryOptimizedSearcher(indexInput);
         } catch (UnsupportedFaissIndexException e) {
+            log.error(e);
```
navneet1v: this is interesting, if index type is not supported with MemoryOptimized search why the call is coming here?

If we think in case of bad code the call can come here, in that case we should just throw the exception with proper messages on what cause the `UnsupportedFaissIndexException`

src/main/java/org/opensearch/knn/index/engine/faiss/FaissHNSWMethod.java
owenhalpert: +1 for `switch` statement and/or flip the condition.

src/main/java/org/opensearch/knn/index/engine/faiss/FaissHNSWMethod.java
```diff
@@ -200,14 +195,131 @@ public RemoteIndexParameters createRemoteIndexingParameters(Map<String, Object>
...
+        }
+
+        if (isByteIndex(parameters)) {
+            return true;
+        }
+
+        return false;
+    }
+
+    private static boolean isFloat32Index(final Map<String, Object> parameters) {
```
owenhalpert: Does a method like this need to exist in the `FaissHNSWMethod` class? Or is it a more generic method on the parameters, in which case should we move it to Faiss or some parameters class?

src/main/java/org/opensearch/knn/index/engine/faiss/FaissHNSWMethod.java
```diff
@@ -200,14 +195,131 @@ public RemoteIndexParameters createRemoteIndexingParameters(Map<String, Object>
...
+        }
+    }
+
+    /**
+     * From indexing library parameter, it determines whether configured index is FP16, scalar quantized.
+     *
+     * @param parameters KNN library indexing parameters.
+     * @return Trye if FP16, otherwise False.
+     */
+    public static boolean isFloat16Index(final Map<String, Object> parameters) {
```
owenhalpert: There is a lot of common logic in these methods. It might be more readable like this, but maybe consider getting dataType and encoderType in a helper/caller method then passing those into `is_X_index(dataType, encoderType)`

src/main/java/org/opensearch/knn/index/engine/faiss/FaissHNSWMethod.java
0ctopus13prime: Better branding for a variable name is welcome! This is my weak area.
And agree on the point 2. I think it would be better just to pass VectorDataType to each checking function instead.
But I think it would better to clearly describe the allowed cases respectively as we are keeping pushing things into the system. (Ex: BFP16, 4x compression, future features) 

For example `isQuantizedIndex` checking, it is counter intuitive, but VectorDataType is given as a `Float` for quantized index, not 'Binary' at the time the method is called. Also the encoder parameters map should be empty. 
This contract can be break in the future, in one day, we can add a new encoder parameter which will evaluate `isQuantizedIndex` as true, which is not what we want.

I thought listing out 'supported cases' would be easy and solid for us to check what are supported.

src/main/java/org/opensearch/knn/index/engine/faiss/FaissHNSWMethod.java
0ctopus13prime: I dont think we need log things there, but we could print error message in debug mode.
The reason I think we can ignore the exception in here is that, even an exception was thrown in here, it is not error. 
For Float32 index, what we expect are data type and 'flat' as an encoder parameter. Even inner map did not have encoder parameter map, technically it's not error. We are trying to pull 'flat' from a non-float32 index.

src/main/java/org/opensearch/knn/index/engine/faiss/FaissHNSWMethod.java
0ctopus13prime: At the moment, it is only specific to FaissHNSWMethod for now. 
If we have a new places could use it, then I think we can factor them out into a separate class.
Do you think it should be separated out? If it is the case, will update in the next rev.

src/main/java/org/opensearch/knn/index/engine/faiss/FaissHNSWMethod.java
0ctopus13prime: sure, will update in the next rev.

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
jed326: I meant in this `buildRemoteBuildRequest` method, we only need the dimension of the vectors to submit the remote build request so I don't think we need to decorate here?

src/main/java/org/opensearch/knn/index/engine/faiss/FaissHNSWMethod.java
jed326: I think even if you want to list out the suppported cases you can still just get the vector data type once at the top instead of in each method

src/main/java/org/opensearch/knn/index/engine/faiss/FaissHNSWMethod.java
jed326: I think a debug log would be good. After all we had the same issue with silent try/catch statements when debugging the repository-s3 byte vector issue.

src/main/java/org/opensearch/knn/index/engine/faiss/FaissHNSWMethod.java
owenhalpert: > Do you think it should be separated out?

No, only if this were a generic method. Thanks. Agree that we can move it out once it is used in other places.

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/QuantizedKNNVectorValuesSupplier.java
jed326: I meant `QuantizedKNNBinaryVectorValues` can just call `indexInfo.getKnnVectorValuesSupplier()` on it's own, but wasn't sure if there was another use case for `QuantizedKNNBinaryVectorValues` that required both parameters.

0ctopus13prime: not sure why have `QuantizedKNNBinaryVectorValues` call `getKnnVectorValuesSupplier()` twice while we have `orgSupplier` already

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
0ctopus13prime: For the quantization case, we should pass the number of bits as dimension because Faiss will interpret dimension as 'code-size' for binary case.
For example, with 32x quantization for 768 dimension float vector, the resulting bits (e.g. code size) would be 96. (e.g. (sizeof(float) * 768 * 8) / 32) / 8). And we need to pass 96 as a code size.

src/main/java/org/opensearch/knn/index/engine/faiss/FaissHNSWMethod.java
0ctopus13prime: We should perform the check whether GPU is supported on both VectorDataType + Encoding.
It should be working for now by merely checking vector data type, but the issue is encoding type as there are limited encoding types that now GPU supports. My concern is what if we introduced a new encoding type with FLOAT, then it can be misinterpreted and return 'true' and pass it to GPU while it does not support the new encoding. 

Therefore, I think we should check both VectorDataType + Encoding per each case to prevent this subtle error from happening.

src/main/java/org/opensearch/knn/index/engine/faiss/FaissHNSWMethod.java
0ctopus13prime: sounds good, will update in the next rev.

src/main/java/org/opensearch/knn/index/engine/faiss/FaissHNSWMethod.java
jed326: Yeah that makes sense. All I'm saying is you can do this part just once:
```
            final String dataType = getStringFromMap(parameters, VECTOR_DATA_TYPE_FIELD);
            final VectorDataType vectorDataType = VectorDataType.get(dataType);
```

0ctopus13prime: oh sure, will update in the next rev.
It would be something like this

```
final VectorDataType vectorDataType = extractVectorDataType(parameters);
final Map<String, Object> encoderMap = extractEncoderMap(parameters);

if (isFloat32Index(vectorDataType, encoderMap)) {
    return true;
}
...
```

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/QuantizedKNNVectorValuesSupplier.java
0ctopus13prime: Oh I got your point.
So we don't have to have `orgSupplier` at the first place in `QuantizedKNNVectorValuesSupplier`, and just passing indexInfo. 
Will update in the next rev.

0ctopus13prime:
> I see there are no ITs on this PR, is that coming later? I would have thought we could at least add the remote build annotation to a few more tests after this PR.

Yes, I will add ITs or add remote-build annotation in the next upcoming PR after 3.2 branch is cut. 
Focusing on adding feature into 3.2 first :melting_face:

src/main/java/org/opensearch/knn/index/engine/faiss/FaissHNSWMethod.java
```diff
@@ -244,86 +246,111 @@ private static boolean isFloat32Index(final Map<String, Object> parameters) {
...
             return encoder.equals(ENCODER_SQ);
         } catch (final Exception e) {
+            log.debug(e.getMessage());
             // Ignore
             return false;
         }
     }
 
-    private static boolean isBinaryIndex(final Map<String, Object> parameters) {
+    private static boolean isBinaryIndex(final VectorDataType vectorDataType, final Map<String, Object> encoderMap) {
```
owenhalpert, line 269: What do you think about simplifying these methods even further?
```
return vectorDataType == VectorDataType.BINARY && getStringFromMap(encoderMap, NAME).equals(ENCODER_FLAT)
```

0ctopus13prime: Yeah, i'm fine with it!
Just I'm prefer enumerating all variables for debugging purposes. 
With the simplified version, we have to execute separately `getStringFromMap(encoderMap, NAME)`. 

Will update in the next rev.

owenhalpert: I meant to simplify to this logic for each method, not just the binary index method. What do you think?

src/main/java/org/opensearch/knn/index/KNNIndexShard.java
0ctopus13prime: Will update in the next rev.

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/DefaultVectorRepositoryAccessor.java
0ctopus13prime: Will update in the next rev.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissMemoryOptimizedSearcherFactory.java
0ctopus13prime: Will update in the next rev.

src/main/java/org/opensearch/knn/index/KNNIndexShard.java
0ctopus13prime: This is a quick temporary fix, without this error log, OpenSearch will keep calling warmup forever.

src/main/java/org/opensearch/knn/index/engine/faiss/FaissHNSWMethod.java
```diff
@@ -199,19 +194,157 @@ public RemoteIndexParameters createRemoteIndexingParameters(Map<String, Object>
...
-            Map<String, Object> encoderMap = (Map<String, Object>) innerMap.get(METHOD_ENCODER_PARAMETER);
-            String dataType = getStringFromMap(parameters, VECTOR_DATA_TYPE_FIELD);
-            String encoder = getStringFromMap(encoderMap, NAME);
-            return SUPPORTED_REMOTE_INDEX_DATA_TYPES.contains(VectorDataType.get(dataType)) && ENCODER_FLAT.equals(encoder);
-        } catch (IllegalArgumentException e) {
-            log.error("Unrecognized indexing parameters in KNNLibraryIndexingContext", e);
+            // This index type is a binary case where user ingested binary vectors (e.g. bit stream)
+            // Therefore, we didn't do any quantization from our end, it is already done from user side.
+            // Check whether if vector type is binary
+            return vectorDataType == VectorDataType.BINARY && getStringFromMap(encoderMap, NAME).equals(ENCODER_FLAT);
```
owenhalpert, line 274: Not going to block the PR but I think the suggestion I had for the change you made here can go into the other methods too. In my opinion it is more readable.

owenhalpert (approved):
LGTM

src/main/java/org/opensearch/knn/index/engine/faiss/FaissHNSWMethod.java
0ctopus13prime: Sure, will update it in the next upcoming PRs

src/main/java/org/opensearch/knn/index/vectorvalues/QuantizedKNNBinaryVectorValues.java
```diff
@@ -0,0 +1,57 @@
...
+        final byte[] quantizedVector = (byte[]) QuantizationIndexUtils.processAndReturnVector(knnFloatVectorValues, indexBuildSetup);
+        this.dimension = quantizedVector.length * Byte.SIZE;
+        this.bytesPerVector = quantizedVector.length;
+        return quantizedVector;
+    }
+
+    @Override
+    public byte[] conditionalCloneVector() throws IOException {
+        throw new UnsupportedOperationException(
+            "Cloning vector is not supported in " + QuantizedKNNBinaryVectorValues.class.getSimpleName()
```
shatejas:
```suggestion
        return getVector();
```

src/main/java/org/opensearch/knn/index/engine/faiss/FaissHNSWMethod.java
```diff
@@ -199,19 +194,157 @@ public RemoteIndexParameters createRemoteIndexingParameters(Map<String, Object>
...
+                return true;
+            }
+
+            if (isQuantizedIndex(vectorDataType, encoderMap)) {
+                return true;
+            }
+
+            return isByteIndex(vectorDataType, encoderMap);
+        } catch (final Exception e) {
+            log.debug(e.getMessage());
```
shatejas:
```suggestion
            log.warn(e.getMessage());
```

src/main/java/org/opensearch/knn/index/engine/faiss/FaissHNSWMethod.java
```diff
@@ -199,19 +194,157 @@ public RemoteIndexParameters createRemoteIndexingParameters(Map<String, Object>
...
-            String encoder = getStringFromMap(encoderMap, NAME);
-            return SUPPORTED_REMOTE_INDEX_DATA_TYPES.contains(VectorDataType.get(dataType)) && ENCODER_FLAT.equals(encoder);
-        } catch (IllegalArgumentException e) {
-            log.error("Unrecognized indexing parameters in KNNLibraryIndexingContext", e);
+            // This index type is a binary case where user ingested binary vectors (e.g. bit stream)
+            // Therefore, we didn't do any quantization from our end, it is already done from user side.
+            // Check whether if vector type is binary
+            return vectorDataType == VectorDataType.BINARY && getStringFromMap(encoderMap, NAME).equals(ENCODER_FLAT);
+        } catch (final Exception e) {
+            log.debug(e.getMessage());
```
shatejas:
```suggestion
            log.warn(e.getMessage());
```

src/main/java/org/opensearch/knn/index/vectorvalues/QuantizedKNNBinaryVectorValues.java
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * This class is designed for uploading quantized binary vectors as part of the GPU indexing pipeline. Internally, we store the
+ * full-precision vectors. When needed, it retrieves these full-precision vectors, applies quantization based on its current state, and
+ * returns the quantized data as a byte[]. As a result, the `bytesPerVector` and `dimension` exposed externally refer to the quantized
+ * (binary) vectors.
+ * For instance, if the original vector has 768 dimensions and a 32× compression ratio is applied, the quantized vector size (code_size)
+ * becomes 96 bytes. In that case, `bytesPerVector` would be set to `96`, and the `dimension` would be set to 768 bits. (e.g. one float
+ * value became one bit)
+ */
+public class QuantizedKNNBinaryVectorValues extends KNNVectorValues<byte[]> {
```
shatejas, line 23: Since this is a public class, I am not sure if having a member variables like IndexInfo and a constructor which is dependent on BuildIndexParams makes sense here. Those POJOs were intended to aid index build. 

Shouldn't this directly take a dependency on QuantizationState and QuantizationService? and the client or vectorvalues factory be responsible for creating the right KNNVectorValues

shatejas (commented):
Can you check the build please

src/main/java/org/opensearch/knn/index/engine/faiss/FaissHNSWMethod.java
0ctopus13prime: Will update in the next rev.

src/main/java/org/opensearch/knn/index/vectorvalues/QuantizedKNNBinaryVectorValues.java
0ctopus13prime: Will update in the next rev.

src/main/java/org/opensearch/knn/index/vectorvalues/QuantizedKNNBinaryVectorValues.java
0ctopus13prime: Sounds good, for this one will cover in a separate PR.
I have to raise a PR for GPU support anyway.
Hoping to merge it as 3.2 branch cut is near.

shatejas (approved):

navneet1v (approved):

src/main/java/org/opensearch/knn/index/vectorvalues/QuantizedKNNBinaryVectorValues.java
navneet1v: Thanks @shatejas for pointing this out. I didn't realize it was present in vector values package. Yes we should either reduce the scope of this QuantizedKNNBinaryVectorValues or change the constructor parameters. 

Will wait for next PR for this.

PR #2822, opened 2025-08-04
Index Warmup Setting

### LINKED ISSUE #2642: [RFC] Preload Graph For Prevent slow knn queries when nodes restart

leoyu808:
### Description
Whenever a node restarts, a shard is rebalanced, or a large segment is merged, there is a large latency spike on the first query to the node due to graph files needing to be reloaded to native memory. This change creates a new index level setting: index.knn.warmup.enabled, which controls whether graph files are eagerly loaded into native memory as soon as a Lucene segment is opened.

### Related Issues
Resolves #2642 

### Check List


src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
```diff
@@ -397,4 +407,57 @@ public boolean isSet() {
...
+                if (memoryOptimizedFieldNames.contains(fieldInfo.getName())) {
+                    String dataTypeStr = fieldInfo.getAttribute(KNNConstants.VECTOR_DATA_TYPE_FIELD);
+                    if (dataTypeStr == null) {
+                        continue;
+                    }
+                    try {
+                        boolean isFloat = VectorDataType.get(dataTypeStr) == VectorDataType.FLOAT;
+                        trySearchWithMemoryOptimizedSearch(fieldInfo.getName(), null, null, null, isFloat);
+                    } catch (Exception e) {
+                        // Ignore
```
kotwanikunal, line 427: We should still log here for traceability.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
```diff
@@ -397,4 +407,57 @@ public boolean isSet() {
...
+                        final SpaceType spaceType = SpaceType.getSpace(spaceTypeName);
+                        final KNNEngine knnEngine = KNNEngine.getEngineNameFromPath(vectorIndexFileName);
+                        final QuantizationParams quantizationParams = QuantizationService.getInstance()
+                            .getQuantizationParams(fieldInfo, segmentInfo.getVersion());
+                        final VectorDataType vectorDataType = FieldInfoExtractor.determineVectorDataType(
+                            fieldInfo,
+                            quantizationParams,
+                            segmentInfo.getVersion()
+                        );
+                        cacheManager.get(
```
kotwanikunal: Is there a reason to not wire up existing warm up logic here?

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngineSegmentAttributeParser.java
```diff
@@ -0,0 +1,96 @@
...
+     * From segmentInfo, parse whether warmup is enabled
+     *
+     * @param segmentInfo {@link SegmentInfo}
+     * @return boolean of whether the warmup setting is enabled
+     */
+    public static boolean parseWarmup(SegmentInfo segmentInfo) {
+        if (segmentInfo == null) {
+            throw new IllegalArgumentException("SegmentInfo cannot be null");
+        }
+        return TRUE.equals(segmentInfo.getAttribute(WARMUP_ENABLED)) && segmentInfo.getAttribute(INDEX_NAME) != null;
```
kotwanikunal, line 41: Adding a field seems interesting. One of the solutions I was thinking for this was to listen for shard recovery events and enable warm up using index level settings.

Have you considered using https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/shard/IndexEventListener.java#L69-L77?

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
```diff
@@ -397,4 +407,57 @@ public boolean isSet() {
...
+                                segmentInfo.dir,
+                                cacheKey,
+                                NativeMemoryLoadStrategy.IndexLoadStrategy.getInstance(),
+                                getParametersAtLoading(spaceType, knnEngine, indexName, vectorDataType, quantizationParams),
+                                indexName
+                            ),
+                            true
+                        );
+                    } catch (Exception e) {
+                        log.warn("Failed to put file in memory: {}, exception: {}", vectorIndexFileName, e.toString());
```
naveentatikonda: 1. Can you log the exception object to print the full stack trace
2. catch a specific exception type and throw it
3. Probably log it as an error instead of warning

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngineSegmentAttributeParser.java
shatejas: @leoyu808 can we explore this approach, current approach stills tries to warmup when the reader is opened which doesn't isolate it rebalance cases. It will still do it for merges which isn't necessary if there are no search requests related to that index

leoyu808: Seems like a good idea, I'm checking out it out now

leoyu808: @shatejas @kotwanikunal The problem with using the hook afterIndexShardStarted is that by the time it runs, the shard is already visible, which doesn't solve the slow queries/timeouts mentioned in #711. I also tried using beforeIndexShardRecovery, but indexShard.acquireSearcher can't be called in the recovering state.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
leoyu808: The existing warm up logic uses LeafReaderContexts to build SegmentLevelQuantizationInfo. However, we don't have access to it in KnnVectorsReaders, which is why I circumvented it by only building the QuantizationParams.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngineSegmentAttributeParser.java
kotwanikunal: That would only be the case if a query hits before the warmup is completed right? 
I think OpenSearch core also has a warm up API which may not be as popularly used.

Can you look into the implementation here - https://github.com/opensearch-project/OpenSearch/blob/f967a72279a54032ba9ef3df4a33f8a7f850d66d/server/src/main/java/org/opensearch/index/IndexWarmer.java#L132-L136?

I think it gives you the appropriate hooks for warmup

PR #2833, opened 2025-08-06
[GRPC] Extend transport-grpc-spi to support GRPC KNN queries

karenyrx:
### Description
Used in conjunction with https://github.com/opensearch-project/OpenSearch/pull/18949

### Related Issues
Part of https://github.com/opensearch-project/k-NN/issues/2816 
Related to https://github.com/opensearch-project/OpenSearch/issues/18513

### Check List


src/main/java/org/opensearch/knn/grpc/proto/request/search/query/KNNQueryBuilderProtoConverter.java
```diff
@@ -16,6 +17,15 @@
  */
 public class KNNQueryBuilderProtoConverter implements QueryBuilderProtoConverter {
 
+    private QueryBuilderProtoConverterRegistry registry;
+
+    @Override
+    public void setRegistry(QueryBuilderProtoConverterRegistry registry) {
```
Vikasht34, line 23: Where this registry is getting used , What is relevance of having here?

src/main/java/org/opensearch/knn/grpc/proto/request/search/query/KNNQueryBuilderProtoUtils.java
```diff
@@ -46,7 +47,7 @@ void setRegistry(QueryBuilderProtoConverterRegistry registry) {
      *
      * @return The current registry
      */
-    QueryBuilderProtoConverterRegistry getRegistry() {
+    static QueryBuilderProtoConverterRegistry getRegistry() {
```
Vikasht34: Why GetRegistry is Private?

Vikasht34 (commented):
Can we have complete unit test and if possible unit test for this PR?

navneet1v:
@karenyrx can you please resolve the comments and get this PR in good shape. The KNN build is failing without this change

src/main/java/org/opensearch/knn/grpc/proto/request/search/query/KNNQueryBuilderProtoUtils.java
```diff
@@ -29,15 +29,16 @@
...
 
     /**
-     * Sets the registry for testing purposes.
+     * Sets the registry injected by the gRPC plugin.
+     * This method is called when the k-NN converter receives the populated registry.
      *
      * @param registry The registry to use
      */
-    void setRegistry(QueryBuilderProtoConverterRegistry registry) {
+    public static void setRegistry(QueryBuilderProtoConverterRegistry registry) {
```
navneet1v: this whole setting and getting of registry is not useful now. Since the registry is being injected by OpenSearch in `KNNQueryBuilderProtoConverter`. You can pass the injected registry directly to `KNNQueryBuilderProtoUtils.fromProto(queryContainer.getKnn());` from  `KNNQueryBuilderProtoConverter`. 

@karenyrx

src/main/java/org/opensearch/knn/grpc/proto/request/search/query/KNNQueryBuilderProtoConverter.java
karenyrx: The registry is used on line 96 of KNNQueryBuilderProtoUtils.java to convert filter clauses. k-NN queries support filters that can be any query type (Term, Bool, Match, etc.), and the registry provides access to the built-in converters for these standard query types. Without the registry, k-NN queries with filters would fail to convert.

src/main/java/org/opensearch/knn/grpc/proto/request/search/query/KNNQueryBuilderProtoUtils.java
karenyrx: The `setRegistry` is required because the k-NN plugin cannot directly access transport-grpc's internal registry - it only gets access through the setRegistry() injection method that transport-grpc plugin calls during plugin loading here: https://github.com/opensearch-project/OpenSearch/blob/73a431607ceeb639614b1e924290842387f137eb/modules/transport-grpc/src/main/java/org/opensearch/transport/grpc/GrpcPlugin.java#L296

So the static registry is necessary because:
1. the converter receives the registry via injection
2. k-NN cannot directly access transport-grpc's internal registry, and 
3. The injection happens after plugin creation but before usage

navneet1v: @karenyrx may be I am missing something here, let me give it another and let me know if I am missing something here.

So the setRegistry() happens here: `KNNQueryBuilderProtoConverter.setRegistry` and then when `KNNQueryBuilderProtoConverter.fromProto()` is calling you can do something like `KNNQueryBuilderProtoUtils.fromProto(..., registry)`. This `registry` is already a class member in `KNNQueryBuilderProtoConverter` so we should be able to do this. Let me know if I am missing anything here.

karenyrx: > you can do something like KNNQueryBuilderProtoUtils.fromProto(..., registry).

Is there a specific functional benefit of refactoring this? 

This would require changing the interface in core: https://github.com/opensearch-project/OpenSearch/blob/73a431607ceeb639614b1e924290842387f137eb/modules/transport-grpc/spi/src/main/java/org/opensearch/transport/grpc/spi/QueryBuilderProtoConverter.java#L33 , which is extended by all query types. However, not all query types required a registry, so this registry was not added as part of the interface in core. For example MatchAll, Term, Terms, do not require this registry as it does not have any "embedded queries" of other types inside of them.

src/main/java/org/opensearch/knn/grpc/proto/request/search/query/KNNQueryBuilderProtoUtils.java
karenyrx: Yes good point. It was only used in one test in `KNNQueryBuilderProtoUtilsTests.java`. I've removed this method and rewritten the tests. Thanks for the catch!

karenyrx:
> Can we have complete unit test and if possible unit test for this PR?

A good test is that that previous unit tests continue to pass (as this is mostly a refactor), but added some new ones as well to cover more edge cases

src/main/java/org/opensearch/knn/grpc/proto/request/search/query/KNNQueryBuilderProtoUtils.java
navneet1v: Not sure why you are saying that it needs interface change since `KNNQueryBuilderProtoUtils` is internal to k-NN. So when you say we need to change the interface I am not sure what exactly you mean here

karenyrx: It is not internal to KNN - it is defined in this SPI file: https://github.com/opensearch-project/k-NN/blob/8129f5ed3df445feecbc0d8b683f1febadba1338/src/main/resources/META-INF/services/org.opensearch.transport.grpc.spi.QueryBuilderProtoConverter 
which extends https://github.com/opensearch-project/OpenSearch/blob/73a431607ceeb639614b1e924290842387f137eb/modules/transport-grpc/spi/src/main/java/org/opensearch/transport/grpc/spi/QueryBuilderProtoConverter.java#L33 from core

navneet1v: @karenyrx I understand that the class `KNNQueryBuilderProtoConverter` is defined in the SPI, but your `KNNQueryBuilderProtoUtils` is not in the SPI. I am asking to change the `KNNQueryBuilderProtoUtils` class. Please check my comment again. and let me know if I am still missing anything

> So the setRegistry() happens here: KNNQueryBuilderProtoConverter.setRegistry and then when KNNQueryBuilderProtoConverter.fromProto() is calling you can do something like KNNQueryBuilderProtoUtils.fromProto(..., registry). This registry is already a class member in KNNQueryBuilderProtoConverter so we should be able to do this. Let me know if I am missing anything here.

karenyrx: Ah got it. Thanks for clarifying, good point! Revisited in latest commit.

navneet1v:
@karenyrx the CIs are failing

karenyrx:
Hi @navneet1v the reason the CIs are failing is because https://github.com/opensearch-project/OpenSearch/pull/19007 was just merged. The fix is in https://github.com/opensearch-project/k-NN/pull/2842 is missing. Would you mind helping take a look at that PR? Then I will rebase this PR and retry the CI.

navneet1v (approved):
Approving the code, since CI is only left. Overall code looks good to me.

navneet1v:
@Vikasht34 can you please review this PR now, this PR needs to be merged since this is blocking the builds.

bzhangam:
@karenyrx @finnroblin @Vikasht34 @navneet1v After this PR is merged we can run into new issues([test](https://github.com/opensearch-project/k-NN/actions/runs/17813490938/job/50642192582?pr=2868)):

```
> Could not resolve all files for configuration ':testFixturesCompileClasspath'.
   > Could not resolve com.google.errorprone:error_prone_annotations:2.26.1.
     Required by:
         root project : > com.google.guava:guava:33.2.1-jre
      > Conflict found for module 'com.google.errorprone:error_prone_annotations': between versions 2.41.0 and 2.26.1
> There is 1 more failure with an identical cause.
```

I have tried to do below to fix it in this [PR](https://github.com/opensearch-project/k-NN/pull/2874) and seems like it can pass the github checks.
```
    // com.google.guava:guava:33.2.1-jre is using com.google.errorprone:error_prone_annotations:2.26.1 but OpenSearch
    // core is using 2.41.0 so manually force to use 2.41.0 to avoid the conflict.
    testFixturesImplementation('com.google.guava:guava:33.2.1-jre'){
        exclude group: 'com.google.errorprone', module: 'error_prone_annotations'
    }
    testFixturesImplementation group: 'com.google.errorprone', name: 'error_prone_annotations', version: '2.41.0'
```

But when I start the cluster locally and try to test the knn query it will fail with error:
```
» java.lang.NoClassDefFoundError: com/google/common/util/concurrent/internal/InternalFutureFailureAccess » at java.base/java.lang.ClassLoader.defineClass1(Native Method) » at java.base/java.lang.ClassLoader.defineClass(ClassLoader.java:1027)
```
Which seems related to the change about 
```
   compileOnly("org.opensearch.plugin:transport-grpc:${opensearch_version}") {
        exclude group: 'com.google.guava', module: 'guava'
        exclude group: 'com.google.guava', module: 'failureaccess'
        exclude group: 'com.google.errorprone', module: 'error_prone_annotations'
    }
```

Could you help take a look?

karenyrx:
Opened up https://github.com/opensearch-project/k-NN/pull/2883/files for the fix

PR #2838, opened 2025-08-07
Integrating Lucene's Better Binary Quantization

### LINKED ISSUE #2805: [FEATURE] [RFC] Integrating Lucene's Better Binary Quantization into OpenSearch
## ***1. Problem Statement***

This project aims to integrate BBQ into the OpenSearch k-NN plugin to offer users a memory-efficient alternative, ideal for large-scale vector workloads in constrained compute environments.

The OpenSearch k-NN plugin currently supports two engines, each with distinct quantization capabilities:

- **FAISS Engine**: Supports Product Quantization, Scalar Quantization, Binary Quantization, and raw float32 storage
- **Lucene Engine**: Supports scalar quantization (7-bit) byte vectors, and raw float32 storage

While both engines offer quantization options, OpenSearch lacks access to Lucene’s Better Binary Quantization, which provides a better compress to quality ratio. BBQ is available in the underlying Lucene library, but not exposed through OpenSearch’s k-NN plugin.

This project aims to address this gap and integrate Lucene’s BBQ into OpenSearch’s k-NN plugin, which provides users with a new memory efficient quantization option. 

With this optimization binary quantization achieves a recall enhancement without oversampling from 0.18 using Faiss Binary Quantization to 0.32 using BBQ with the sift-128 dataset, and an improvement of 0.3 to 0.63 using Cohere's-1M dataset.

---

### **2.1 Introduction**

Better Binary Quantization is a vector compression technique developed by Lucene which focuses on reducing memory usage for approximate nearest neighbor search. It encodes high dimensional float vectors into compact binary representations using advanced quantization methods optimized for search efficiency. BBQ was developed with RaBitQ as a reference, drawing on its approach of combining product quantization with binary encoding.

---

### **2.2 Understanding Better Binary Quantization**

**Technical Process:**

This diagram illustrates the process of computing the global centroid of a set of vectors and how the vectors are then centered around this centroid. After centering, we calculate basic statistics and apply an optimization process to quantize the vectors efficiently. This quantization is ultimately used to produce the company binary representations of the vectors.

Centroid Calculation

Quantization Process

Compared to naive binary quantization, the centroid calculation is identical. Naive binary quantization does not, however, calculate stats, optimize intervals, and hold on to corrective factors. Once a centered vector has been calculated, naive binary quantization quantizes without making any optimization.

---

### **2.3 Comparison of Quantization Methods**

| Method                         | Compression Ratio | Recall   | Memory Usage |
| ----------------------------- | ----------------: | -------: | -----------: |
| Scalar Quantization           | 2x, 4x            | Great    | Moderate     |
| Faiss Binary Quantization     | 8x, 16x, 32x      | Moderate | Low          |
| Lucene Better Binary Quantization | 32x           | Good     | Low          |

---

## ***3. Proposed Solution***

- **Lucene BBQ Integration**: Incorporate Lucene’s `Lucene102BinaryQuantizedVectorsFormat` into OpenSearch.
- **Query Pipeline**: Ensure BBQ-based search scoring is invoked properly.
- **Parameterization**: Allow customizable BBQ configurations via mappings.

---

## ***4. User Interface***

Users will configure BBQ through the field mapping:

```json
{
  "settings": {
    "index": {
      "knn": true
    }
  },
  "mappings": {
    "properties": {
      "my_vector1": {
        "type": "knn_vector",
        "dimension": 2,
        "space_type": "l2",
        "method": {
          "name": "hnsw",
          "engine": "lucene",
          "parameters": {
            "encoder": {
              // CHANGED PARAMETER
              "name": "binary"
            },
            "ef_construction": 256,
            "m": 8
          }
        }
      }
    }
  }
}
```
---

### **5.1 Architecture Flow**

Architecture Flow

- User specifies `"encoder": {"name": "binary"}` in index mapping
- `LuceneHNSWMethod` resolves BBQ encoder from `SUPPORTED_ENCODERS` map
- `BasePerFieldKnnVectorsFormat` validates BBQ parameters via `KNNBBQVectorsFormatParams`
- Codec instantiates `Lucene102HnswBinaryQuantizedVectorsFormat` for BBQ-enabled fields
- Lucene's native BBQ implementation handles vector quantization and storage

---

#### LuceneBBQEncoder

Implements `Encoder` interface, not much configuration as it is set similar to SQ, but without 3 variables.

```java
public class LuceneBBQEncoder implements Encoder {
    private static final Set<VectorDataType> SUPPORTED_DATA_TYPES = ImmutableSet.of(VectorDataType.FLOAT);

    private final static MethodComponent METHOD_COMPONENT = MethodComponent.Builder.builder(ENCODER_BBQ)
        .addSupportedDataTypes(SUPPORTED_DATA_TYPES)
        .build();

    @Override
    public MethodComponent getMethodComponent() {
        return METHOD_COMPONENT;
    }

    @Override
    public CompressionLevel calculateCompressionLevel(
        MethodComponentContext methodComponentContext,
        KNNMethodConfigContext knnMethodConfigContext
    ) {
        return CompressionLevel.NOT_CONFIGURED;
    }
}
```
#### LuceneHNSWMethod

BBQ encoder is registered the same way as SQ encoder

```java
final static Encoder SQ_ENCODER = new LuceneSQEncoder();
final static Encoder BBQ_ENCODER = new LuceneBBQEncoder();
final static Map<String, Encoder> SUPPORTED_ENCODERS = Map.of(
    SQ_ENCODER.getName(), SQ_ENCODER,
    BBQ_ENCODER.getName(), BBQ_ENCODER
);
```

#### KNNBBQVectorsFormatParams

Used for validation, ensures proper bbq encoder identification

```java
public class KNNBBQVectorsFormatParams extends KNNVectorsFormatParams {
    @Override
    public boolean validate(Map<String, Object> params) {
            return false;
        }

            return false;
        }

        MethodComponentContext encoderMethodComponentContext = 
            (MethodComponentContext) params.get(METHOD_ENCODER_PARAMETER);
        return ENCODER_BBQ.equals(encoderMethodComponentContext.getName());
    }
}
```

#### BasePerFieldKnnVectorsFormat Changes

Check for logic that the user select BBQ

```java
    if (params != null && params.containsKey(METHOD_ENCODER_PARAMETER)) {
        KNNBBQVectorsFormatParams bbqParams = new KNNBBQVectorsFormatParams(
            params, defaultMaxConnections, defaultBeamWidth
        );
        if (bbqParams.validate(params)) {
            log.debug("Initialize KNN vector format for field [{}] with BBQ encoder", field);
            return bbqVectorsFormatSupplier.apply(bbqParams);
        }
       SQ encoder validation...
    }
}
```

#### KNN9120PerFieldVectorsFormat Implementation

 The concrete format supplier instantiates Lucene’s BBQ format

```java
knnBBQVectorsFormatParams -> {
    final Tuple<Integer, ExecutorService> mergeThreadCountAndExecutorService = 
        getMergeThreadCountAndExecutorService();
    return new Lucene102HnswBinaryQuantizedVectorsFormat(
        knnBBQVectorsFormatParams.getMaxConnections(),
        knnBBQVectorsFormatParams.getBeamWidth(),
        mergeThreadCountAndExecutorService.v1(),
        mergeThreadCountAndExecutorService.v2()
    );
}
```

#### 5.4.1 Leveraging Existing Infrastructure

BBQ integrates with OpenSearch k-NN's existing rescoring framework through Lucene's built-in search mechanism. No additional rescoring classes are required as Lucene's Lucene102HnswBinaryQuantizedVectorsFormat handles rescoring automatically.

#### 5.4.2 Automatic Search Implementation

Lucene's BBQ implementation automatically performs:

* HNSW graph traversal using binary quantized vectors for fast candidate retrieval
* Final scoring using original FP32 vectors from rawVectorDelegate for precision

#### 5.4.3 Query Parameter Usage

Users control BBQ rescoring through existing query structure:  

```json
{
  "query": {                                                                                                                                                                         
    "knn": {                                                                                                                                                                         
      "my_vector1": {                                                                                                                                                                
        "vector": [0.1, 0.2, 0.3],                                                                                                                                                   
        "k": 10,                                                                                                                                                                     
        "rescore": {                                                                                                                                                                 
          "oversample_factor": 3.0                                                                                                                                                   
        }                                                                                                                                                                            
      }                                                                                                                                                                              
    }                                                                                                                                                                                
  }                                                                                                                                                                                  
}                                                                                                                                                                                    
```
* Rescoring needs to happen through oversampling_factor parameter. 
* Scalar quantization by default sets a compression level of 4x. This compression level is not configured with BBQ, so instead a user must pass an oversample factor by query to handle these differences
* Unlike Faiss, Lucene mmaps the index, which allows search without fully loading the entire index into memory
* The segment size is larger because we store the .veb file in addition to the .vemb file, which contains extra information

#### 5.4.4 Built-in Performance Optimizations

Lucene's BBQ format automatically provides:

* Selective Vector Access: Only loads FP32 vectors when needed for final scoring
* Memory Efficiency: Leverages dual storage architecture to minimize memory usage
* Optimized Scoring: Uses corrective terms and binary representations for efficient similarity computation

## ***6. Benchmarking***
Currently, these benchmarking results do not contains tests with rescoring and oversampling enabled. Those tests are in progress.

## Config

| Config Item       | Sift-128 | Cohere-768-1M |
|------------------|:--------:|:-------------:|
| Nodes            |    1     |       1       |
| Primary Shards   |    1     |       1       |
| Replica Shards   |    1     |       1       |
| ef_construction  |   100    |      100      |
| ef_search        |   100    |      100      |
| m                |   16     |      16       |
| space type       |   l2     |      l2       |

## Metrics

| Metric                              | Task                     | Sift Dataset, Lucene BBQ | Sift Dataset, FaissBQ | Cohere-768-1M, Lucene BBQ | Cohere-768-1M, Faiss BQ | Unit      |
|------------------------------------|-------------------------|:-----------------------:|:--------------------:|:------------------------:|:----------------------:|:---------:|
| Mean recall@100                    | prod-queries            |          0.32           |         0.18         |           0.63           |          0.3           |           |
| Mean Throughput                    | custom-vector-bulk      |        6310.84          |       9569.38        |         3011.83          |        3112.96         | docs/s    |
| Median Throughput                  | custom-vector-bulk      |        6226.56          |       9582.13        |         3065.34          |        3152.94         | docs/s    |
| Mean Throughput                    | prod-queries            |          1.82           |         1.61         |           1.66           |          1.69          | ops/s     |
| Median Throughput                  | prod-queries            |          1.82           |         1.61         |           1.66           |          1.69          | ops/s     |
| 50th percentile service time       | prod-queries            |         5.69515         |       5.71014        |         5.88795          |        5.58729         | ms        |
| 90th percentile service time       | prod-queries            |         8.88647         |       7.59109        |         7.74107          |        6.48138         | ms        |
| 99th percentile service time       | prod-queries            |        18.26953         |      17.57099        |        19.85139          |       17.39427         | ms        |

Without any oversampling, there is a definite recall improvement when using BBQ as opposed to the existing Faiss binary quantization, with minor affects on latency and throughput.

adityamachiroutu:
## Description
Integrates Lucene’s Better Binary Quantization (BBQ) into the OpenSearch k-NN plugin, enabling a memory-efficient encoding option for high-dimensional vector search. Compared to Faiss binary quantization, BBQ has increased recall, while maintaining low memory usage and integrating well into existing query and rescoring pipelines.

This integration exposes BBQ through the Lucene engine with a new "encoder": "binary" parameter in index mappings, leveraging Lucene’s native vector quantization and storage formats. BBQ improves recall on large datasets with a modest trade-off in latency and throughput, as demonstrated in the benchmarks below. Users can adjust the oversample factor in queries to balance recall and performance. By default the oversample factor set is 3x for vectors with 1000+ dimensions, and 5x otherwise. This matches the FAISS 32x compression defaults set.

Users can additionally configure the BBQ encoder by setting their engine as Lucene and the compression level to 32x, from which BBQ is automatically enabled.

#### No Rescoring Tests Comparing FAISS BQ to Lucene BBQ
| Metric                        | Sift-128, Lucene BBQ | Sift-128, FaissBQ | Cohere-1m, Lucene BBQ | Cohere-1m, FaissBQ | Unit     |
|-------------------------------|-----------------------|-------------------|-----------------------|-------------------|----------|
| Mean recall@k                  | 0.32                  | 0.18              | 0.63                  | 0.3                |          |
| 50th percentile service time    | 5.69515               | 5.71014           | 5.88795               | 5.58729            | ms       |
| 90th percentile service time    | 8.88647               | 7.59109           | 7.74107               | 6.48138            | ms       |
| 99th percentile service time    | 18.26953              | 17.57099          | 19.85139              | 17.39427           | ms       |
| Mean Throughput                 | 6310.84               | 9569.38           | 3011.83               | 3112.96            | docs/s   |
| Median Throughput               | 6226.56               | 9582.13           | 3065.34               | 3152.94            | docs/s   |

Config:
1 shard, 1 segments, 10 indexing clients, k=100, ef_search: 100, ef_construction: 100

#### Single Node Tests Comparing FAISS BQ (on disk) to Lucene BBQ

| Dataset                  | 32x Compression Technique | Recall@100 | p90 ms latency (1 shard, 1 segment) |
|---------------------------|----------------------------|------------|-------------------------------------|
| clip-flickr-image-text-queries | Lucene BBQ                 | 0.91       | 96.1                                |
|                           | Faiss BQ (on disk)         | 0.81       | 58.24                               |
| cohere-v2-dbpedia          | Lucene BBQ                 | 0.75       | 52.4                                |
|                           | Faiss BQ (on disk)         | 0.76       | 12.29                               |
| cohere-v2-wiki             | Lucene BBQ                 | 0.9        | 58.9                                |
|                           | Faiss BQ (on disk)         | 0.9        | 15.74                               |
| cohere-v3-bioasq           | Lucene BBQ                 | 0.73       | 226.63                              |
|                           | Faiss BQ (on disk)         | 0.73       | 17.97                               |
| e5small-msmarco            | Lucene BBQ                 | 0.9        | 40.06                               |
|                           | Faiss BQ (on disk)         | 0.8        | 16.04                               |
| gist                       | Lucene BBQ                 | 0.64       | 49.67                               |
|                           | Faiss BQ (on disk)         | 0.12       | 46.84                               |
| glove                      | Lucene BBQ                 | 0.63       | 4.37                                |
|                           | Faiss BQ (on disk)         | 0.42       | 4.37                                |
| minilm-msmarco              | Lucene BBQ                 | 0.97       | 2.81                                |
|                           | Faiss BQ (on disk)         | 0.9        | 3.6                                 |
| mpnet-msmarco               | Lucene BBQ                 | 0.98       | 4.98                                |
|                           | Faiss BQ (on disk)         | 0.95       | 3.68                                |
| mbread_marco                | Lucene BBQ                 | 0.96       | 5.93                                |
|                           | Faiss BQ (on disk)         | 0.92       | 4.96                                |
| sift                       | Lucene BBQ                 | 0.69       | 5.23                                |
|                           | Faiss BQ (on disk)         | 0.39       | 3.72                                |
| snowflake-msmarco           | Lucene BBQ                 | 0.94       | 70.41                               |
|                           | Faiss BQ (on disk)         | 0.93       | 47.79                               |
| tasb-msmarco                | Lucene BBQ                 | 0.95       | 9.73                                |
|                           | Faiss BQ (on disk)         | 0.87       | 3.93                                |

Config:  ef_search: 256, ef_construction: 256, m: 16, 1 shard, 1 client, 1 segment

graph
size
spacetype
dimension
embeddingtype

* These tests compare Lucene’s better binary quantization with Faiss BQ (on disk), which is the current default quantization method used by OpenSearch when undergoing 32x compression.
* Overall, we see that Lucene BBQ achieves higher recall compared to Faiss BQ on disk. This occurs for all nearly datasets, regardless of space type, dimensionality, dataset size, or modality of the data. 
* FAISS binary quantization significantly outperforms BBQ in p90 latency across all datasets.
* Both models seems to struggle with image datasets compared to text datasets, although Lucene’s BBQ still outperformed FAISS
* As dataset sizes increases, both quantization techniques scaled effectively, and the trade off recall to latency persisted
* When accuracy is the focus, Lucene BBQ seems to be the use case fit, while Faiss BQ is more suitable for latency sensitive times. 

### Related Issues
Resolves #2805 

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/mapper/CompressionLevel.java
+++ b/src/main/java/org/opensearch/knn/index/mapper/CompressionLevel.java
@@ -138,7 +138,7 @@ public enum CompressionLevel {
         }
 
         // Special handling for Lucene BBQ (x32 compression)
-        if (this == x32 && engine == KNNEngine.LUCENE && version.onOrAfter(Version.V_3_2_0)) {
+        if (this == x32 && engine == KNNEngine.LUCENE && version.onOrAfter(Version.V_3_3_0)) {
             if (dimension <= RescoreContext.DIMENSION_THRESHOLD) {
                 return RescoreContext.builder()
                     .oversampleFactor(RescoreContext.OVERSAMPLE_FACTOR_BELOW_DIMENSION_THRESHOLD)


naveentatikonda:
@adityamachiroutu as mentioned offline, can you double check the oversampling factor of Faiss BQ with a debugger(check the value of this [firstPassK](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java#L88)). I believe it's 5x not 3x if dimension of vector is less than 1000 - https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/mapper/CompressionLevel.java#L119

src/testFixtures/java/org/opensearch/knn/KNNRestTestCase.java
```diff
@@ -2516,6 +2516,19 @@ protected boolean isApproximateThresholdSupported(final Optional<String> bwcVers
...
+    protected boolean isBBQEncoderSupported(final Optional<String> bwcVersion) {
+        if (bwcVersion.isEmpty()) {
+            return false;
+        }
+        String versionString = bwcVersion.get();
+        if (versionString.endsWith("-SNAPSHOT")) {
+            versionString = versionString.substring(0, versionString.length() - 9);
+        }
+        final Version version = Version.fromString(versionString);
+        return version.onOrAfter(Version.V_3_2_0);
```
finnroblin: V_3_3_0

CHANGELOG.md
```diff
@@ -4,4 +4,4 @@ All notable changes to this project are documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html). See the [CONTRIBUTING guide](./CONTRIBUTING.md#Changelog) for instructions on how to add changelog entries.
 
-## [Unreleased 3.3](https://github.com/opensearch-project/k-NN/compare/main...HEAD)
\ No newline at end of file
+## [Unreleased 3.3](https://github.com/opensearch-project/k-NN/compare/main...HEAD)
```
kotwanikunal, line 7: I'd also add a changelog entry while you are at it :)

qa/restart-upgrade/build.gradle
```diff
@@ -247,6 +248,7 @@ testClusters {
                 knn_bwc_version.startsWith("2.15.")) {
             filter {
                 excludeTestsMatching "org.opensearch.knn.bwc.IndexingIT.testKNNIndexLuceneQuantization"
+                excludeTestsMatching "org.opensearch.knn.bwc.IndexingIT.testKNNIndexLuceneBBQ"
```
kotwanikunal: nit: You can also your the `@AwaitsFix` to disable tests. 
Sample: https://github.com/opensearch-project/k-NN/blob/49ff9a8e63588679949ae6f2a6fd5e62fd81f123/qa/restart-upgrade/src/test/java/org/opensearch/knn/bwc/WarmupIT.java#L49

src/main/java/org/opensearch/knn/index/codec/BasePerFieldKnnVectorsFormat.java
```diff
@@ -44,7 +45,8 @@ public abstract class BasePerFieldKnnVectorsFormat extends PerFieldKnnVectorsFor
     private final int defaultBeamWidth;
     private final Supplier<KnnVectorsFormat> defaultFormatSupplier;
     private final Function<KNNVectorsFormatParams, KnnVectorsFormat> vectorsFormatSupplier;
-    private Function<KNNScalarQuantizedVectorsFormatParams, KnnVectorsFormat> scalarQuantizedVectorsFormatSupplier;
+    private final Function<KNNBBQVectorsFormatParams, KnnVectorsFormat> bbqVectorsFormatSupplier;
```
kotwanikunal: super nit: Add new parameters after the existing attributes. Makes it easier to follow with the method signatures and attribute sequences.

CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ All notable changes to this project are documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html). See the [CONTRIBUTING guide](./CONTRIBUTING.md#Changelog) for instructions on how to add changelog entries.
 
 ## [Unreleased 3.3](https://github.com/opensearch-project/k-NN/compare/main...HEAD)
+
+### Features
+* Integrated Lucene's better binary quantization [#2838](https://github.com/opensearch-project/k-NN/pull/2838)
```
finnroblin: Nit: present tense

qa/restart-upgrade/src/test/java/org/opensearch/knn/bwc/IndexingIT.java
```diff
@@ -659,4 +661,75 @@ public void testRandomRotationBWC() throws Exception {
             deleteKNNIndex(newIndex);
         }
     }
+
+    @AwaitsFix(bugUrl = "https://github.com/opensearch-project/k-NN/issues/2805")
```
finnroblin: Is this meant to be here?

src/main/java/org/opensearch/knn/index/codec/BasePerFieldKnnVectorsFormat.java
```diff
@@ -110,6 +114,11 @@ public KnnVectorsFormat getKnnVectorsFormatForField(final String field) {
 
         if (engine == KNNEngine.LUCENE) {
             if (params != null && params.containsKey(METHOD_ENCODER_PARAMETER)) {
+                KNNBBQVectorsFormatParams bbqParams = new KNNBBQVectorsFormatParams(params, defaultMaxConnections, defaultBeamWidth);
+                if (bbqParams.validate(params)) {
```
finnroblin, line 118: Can we please add debug log like the below?

src/main/java/org/opensearch/knn/index/codec/backward_codecs/KNN990Codec/KNN990PerFieldKnnVectorsFormat.java
```diff
@@ -37,6 +38,12 @@ public KNN990PerFieldKnnVectorsFormat(final Optional<MapperService> mapperServic
                 knnScalarQuantizedVectorsFormatParams.isCompressFlag(),
                 knnScalarQuantizedVectorsFormatParams.getConfidenceInterval(),
                 null
+            ),
+            knnBBQVectorsFormatParams -> new Lucene102HnswBinaryQuantizedVectorsFormat(
+                knnBBQVectorsFormatParams.getMaxConnections(),
+                knnBBQVectorsFormatParams.getBeamWidth(),
```
finnroblin, line 44: Is there no constructor parameter for bits like above?

src/main/java/org/opensearch/knn/index/codec/params/KNNBBQVectorsFormatParams.java
```diff
@@ -0,0 +1,46 @@
...
+
+        MethodComponentContext encoderMethodComponentContext = (MethodComponentContext) params.get(METHOD_ENCODER_PARAMETER);
+        return ENCODER_BBQ.equals(encoderMethodComponentContext.getName());
+    }
+
+    /**
+     * Check if BBQ is enabled
+     * @return true if BBQ is enabled, false otherwise
+     */
+    public boolean isBBQEnabled() {
```
finnroblin: What's the purpose of this method?

src/main/java/org/opensearch/knn/index/engine/lucene/LuceneMethodResolver.java
```diff
@@ -69,10 +72,18 @@ protected void resolveEncoder(KNNMethodContext resolvedKNNMethodContext, KNNMeth
         }
 
         MethodComponentContext methodComponentContext = resolvedKNNMethodContext.getMethodComponentContext();
-        MethodComponentContext encoderComponentContext = new MethodComponentContext(SQ_ENCODER.getName(), new HashMap<>());
+
+        String encoderName = (resolvedCompressionLevel == CompressionLevel.x32)
```
finnroblin: Nit: since there are multiple clauses let's make this one if check of x32 compression

src/main/java/org/opensearch/knn/index/mapper/CompressionLevel.java
```diff
@@ -108,12 +109,22 @@ public static boolean isConfigured(CompressionLevel compressionLevel) {
      *                  is invalid.
      */
     public RescoreContext getDefaultRescoreContext(Mode mode, int dimension, Version version) {
-        // TODO move this to separate class called resolver to resolve rescore context
```
finnroblin, line 111: Does your PR address this TODO? If not can we please leave it in?

src/test/java/org/opensearch/knn/index/engine/lucene/LuceneMethodResolverTests.java
```diff
@@ -178,21 +178,6 @@ public void testResolveMethod_whenInvalid_thenThrow() {
             )
         );
 
-        // Invalid compression
-        expectThrows(
```
finnroblin, line 182: Let's keep this in and change to x16 to still throw with a comment explaining that x32 added in this PR

finnroblin (changes requested):
Few small comments and questions, thanks Aditya!

src/main/java/org/opensearch/knn/index/codec/backward_codecs/KNN990Codec/KNN990PerFieldKnnVectorsFormat.java
adityamachiroutu: No, Lucene does not have that constructor.

src/main/java/org/opensearch/knn/index/codec/BasePerFieldKnnVectorsFormat.java
```diff
@@ -110,6 +114,19 @@ public KnnVectorsFormat getKnnVectorsFormatForField(final String field) {
...
+                    log.debug(
+                        "Initialize KNN vector format for field [{}] with params [{}] = \"{}\", [{}] = \"{}\"",
```
finnroblin: Will the log statement have BBQ in it? Idea is for the log to distinguish whether BBQ or ScalarQuantized format was instantiated.

src/main/java/org/opensearch/knn/index/codec/params/KNNBBQVectorsFormatParams.java
```diff
@@ -0,0 +1,37 @@
...
+        Map<String, Object> bbqEncoderParams = encoderMethodComponentContext.getParameters();
+    }
+
+    @Override
+    public boolean validate(Map<String, Object> params) {
+        if (params.get(METHOD_ENCODER_PARAMETER) == null) {
+            return false;
+        }
+
+        if (!(params.get(METHOD_ENCODER_PARAMETER) instanceof MethodComponentContext)) {
```
kotwanikunal: nit: use `== false` check. It's standard across OpenSearch

```
if ((params.get(METHOD_ENCODER_PARAMETER) instanceof MethodComponentContext) == false)
```

src/main/java/org/opensearch/knn/index/mapper/CompressionLevel.java
```diff
@@ -108,12 +109,23 @@ public static boolean isConfigured(CompressionLevel compressionLevel) {
...
+
+    @VisibleForTesting
+    RescoreContext getDefaultRescoreContext(Mode mode, int dimension) {
+        return getDefaultRescoreContext(mode, dimension, Version.CURRENT);
+    }
+
+    // Add new method signature with KNNEngine parameter
+    public RescoreContext getDefaultRescoreContext(Mode mode, int dimension, Version version, KNNEngine engine) {
         // TODO move this to separate class called resolver to resolve rescore context
         if (modesForRescore.contains(mode)) {
```
kotwanikunal, line 123: We would also need a version check similar to here - it could be a restore scenario across versions and the encoder might not be supported, unless we backport it.

adityamachiroutu: thanks for reviewing, addressed your comments.

qa/restart-upgrade/build.gradle
```diff
@@ -252,6 +252,7 @@ testClusters {
...
```
naveentatikonda: Why are you trying to exclude this test only until version 2.15 ? it won't work until version 3.2 ?

qa/restart-upgrade/src/test/java/org/opensearch/knn/bwc/IndexingIT.java
```diff
@@ -64,6 +64,8 @@ public class IndexingIT extends AbstractRestartUpgradeTestCase {
     private static final int NUM_DOCS = 10;
     private static int QUERY_COUNT = 0;
 
+    private static final String ALGO = "hnsw";
```
naveentatikonda: remove this constant

qa/restart-upgrade/src/test/java/org/opensearch/knn/bwc/IndexingIT.java
```diff
@@ -659,4 +661,74 @@ public void testRandomRotationBWC() throws Exception {
             deleteKNNIndex(newIndex);
         }
     }
+
+    public void testKNNIndexLuceneBBQ() throws Exception {
+        waitForClusterHealthGreen(NODES_BWC_CLUSTER);
```
naveentatikonda, line 664: We don't need to add the same condition to validate BWC version twice in if and else blocks, probably add it here after the cluster is green
```
if (!isBBQEncoderSupported(getBWCVersion())) {
                logger.info("Skipping testKNNIndexLuceneBBQ as BBQ encoder is not supported in version: {}", getBWCVersion());
                return;
            }
```

src/main/java/org/opensearch/knn/index/engine/lucene/LuceneBBQEncoder.java
```diff
@@ -0,0 +1,42 @@
...
+import static org.opensearch.knn.common.KNNConstants.ENCODER_BBQ;
+
+/**
+ * Lucene BBQ (Better Binary Quantization) encoder
+ */
+public class LuceneBBQEncoder implements Encoder {
+    private static final Set<VectorDataType> SUPPORTED_DATA_TYPES = ImmutableSet.of(VectorDataType.FLOAT);
+
+    private final static MethodComponent METHOD_COMPONENT = MethodComponent.Builder.builder(ENCODER_BBQ)
+        .addSupportedDataTypes(SUPPORTED_DATA_TYPES)
```
naveentatikonda, line 27: As discussed offline in the past, can you add support for **_bits_** parameter and set default to 1 bit (32x compression) such that in the future if Lucene supports 2 and 4 bits we can use this parameter.

naveentatikonda: For the naming convention, pls keep it consistent with Faiss BQ

naveentatikonda:
> 
> ### Benchmarking Results:
> #### Single Node Tests Comparing FAISS BQ (on disk) to Lucene BBQ
> Dataset	32x Compression Technique	Recall@100	p90 ms latency (1 shard, 1 segment)

@adityamachiroutu can you add the configuration used for running these tests specifically m, ef_construction

src/main/java/org/opensearch/knn/index/mapper/CompressionLevel.java
```diff
@@ -124,12 +136,22 @@ public RescoreContext getDefaultRescoreContext(Mode mode, int dimension, Version
...
             return defaultRescoreContext;
         }
-        return null;
-    }
 
-    @VisibleForTesting
-    RescoreContext getDefaultRescoreContext(Mode mode, int dimension) {
-        return getDefaultRescoreContext(mode, dimension, Version.CURRENT);
+        // Special handling for Lucene BBQ (x32 compression)
+        if (this == x32 && engine == KNNEngine.LUCENE && version.onOrAfter(Version.V_3_2_0)) {
```
finnroblin: Should be after `V_3_3_0`, since this was not added as a part of 3.2.

naveentatikonda: no, version.onOrAfter(Version.V_3_3_0)

PR #2826, opened 2025-08-05
Add "exact_knn" query clause

kaushikau:
### Description
Currently, if a user wants to perform exact search on a field with index.knn set to true, a score script query must be used. As mentioned in #2368 and #1079, script scoring has the extra hop of script compilation and requires users to have script score knowledge which may not be intuitive. Adding a separate KNN exact query clause `exact_knn` could simplify this use-case and make carrying out exact search less ambiguous for users. In addition, this query will have the option for users to specify a space type that will be used for exact search (which can be different from the one in the index), which is a benefit of script scoring that will be maintained.

Example of query with `exact_knn`:
```
{
  "size": 2,
  "query": {
    "exact_knn": {
      "my_vector": {
        "vector": [2, 3, 5, 6],
        "space_type": "l2"
      }
    }
  }
}
```

Note: This query type will also support nested fields/querying, will add that soon.

### Related Issues
Resolves Problem 2 of #2368

### Check List
- [X] New functionality includes testing.
- [ ] New functionality has been documented.
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
- [X] Commits are signed per the DCO using `--signoff`.
- [X] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/k-NN/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #2848, opened 2025-08-13
Add expandNested support to "exact_knn" query clause

kaushikau:
### Description
This is a follow-up to https://github.com/opensearch-project/k-NN/pull/2826, and adds support for the expandNested query parameter in the exact_knn query. This allows for all child document scores to be retrieved instead of only the highest-scoring one.

Example of query with expandNested:

```
{
  "size": 2,
  "query": {
    "exact_knn": {
      "my_vector": {
        "vector": [2, 3, 5, 6],
        "space_type": "l2",
        "expand_nested_docs": "true"
      }
    }
  }
}
```

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/query/iterators/NestedBinaryVectorIdsKNNIterator.java
+++ b/src/main/java/org/opensearch/knn/index/query/iterators/NestedBinaryVectorIdsKNNIterator.java
@@ -17,19 +17,23 @@ import java.io.IOException;
  * This iterator iterates filterIdsArray to scoreif filter is provided else it iterates over all docs.
  * However, it dedupe docs per each parent doc
  * of which ID is set in parentBitSet and only return best child doc with the highest score.
+ * When expandNested is true, it returns ALL child docs instead of just the best one.
  */
 public class NestedBinaryVectorIdsKNNIterator extends BinaryVectorIdsKNNIterator {
     private final BitSet parentBitSet;
+    private final boolean expandNested;
 
     public NestedBinaryVectorIdsKNNIterator(
         @Nullable final DocIdSetIterator filterIdsIterator,
         final byte[] queryVector,
         final KNNBinaryVectorValues binaryVectorValues,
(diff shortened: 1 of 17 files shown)


src/main/java/org/opensearch/knn/index/query/iterators/NestedBinaryVectorIdsKNNIterator.java
```diff
@@ -40,20 +44,34 @@ public NestedBinaryVectorIdsKNNIterator(
...
+        if (expandNested) {
+            int currentParent = parentBitSet.nextSetBit(docId);
+            if (currentParent != DocIdSetIterator.NO_MORE_DOCS && docId < currentParent) {
+                currentScore = computeScore();
+                int currentDocId = docId;
+                docId = getNextDocId();
+                return currentDocId;
+            }
+            docId = getNextDocId();
+            return nextDoc();
```
shatejas, line 72: Just for clarification, what case is this handling? I think if it hits this case then docId will already be -1 right? is it safe to return no more docs here?

src/main/java/org/opensearch/knn/index/query/iterators/NestedBinaryVectorIdsKNNIterator.java
```diff
@@ -40,20 +44,34 @@ public NestedBinaryVectorIdsKNNIterator(
...
      */
     @Override
     public int nextDoc() throws IOException {
         if (docId == DocIdSetIterator.NO_MORE_DOCS) {
             return DocIdSetIterator.NO_MORE_DOCS;
         }
 
+        if (expandNested) {
+            int currentParent = parentBitSet.nextSetBit(docId);
+            if (currentParent != DocIdSetIterator.NO_MORE_DOCS && docId < currentParent) {
```
shatejas, line 65: will there be a case where docId won't be less than parent? it will either be NO_MORE_DOCS or less than parent right?

PR #2850, opened 2025-08-14
Enable FP16 support for Exact Search and Lucene ANN

ManasviGoyal:
### Description

This is the final PR for adding FP16 support for Exact Search and Lucene ANN.

In this PR, I am using the SIMD C++ code and the Custom Format added in the previous PR to enable the `half_float` data type support for exact search and Lucene ANN

When  `index.knn.advanced.approximate_threshold` is set to -1 , which means no graphs are built, and the `data_type` is `half_float` , the custom `KNN990HalfFloatFlatVectorsFormat` is returned to store vectors as FP16 in the `.vec` file irrespective of the engine configuration. In other cases, the format is selected based on engine configuration and data type to perform ANN search.

| Graphs Built? | Engine         | Vector Type         | Format Used                          |
|------------|----------------|---------------------|--------------------------------------|
| No            | –              | half_float          | KNN990HalfFloatFlatVectorsFormat     |
| No            | –              | float, byte, binary         | Lucene99FlatVectorsFormat            |
| Yes           | Lucene         | half_float          | KNN9120HnswHalfFloatVectorsFormat    |
| Yes           | Lucene         | binary              | KNN9120HnswBinaryVectorsFormat       |
| Yes           | Lucene         | float, byte         | Lucene99HnswVectorsFormat            |
| Yes           | Native Engines | float, half_float**, byte, binary | NativeEngines990KnnVectorsFormat     |

**Currently, half_float is not supported for native engines ANN except when the encoder type is fp16 encoder with Faiss engine. The custom format I added will be used by the deduplication project to support it across Faiss ANN.

A new custom format `KNN9120HnswHalfFloatVectorsFormat` is added in this PR to support FP16 for Lucene ANN. This wrapper calls the `KNN990HalfFloatFlatVectorsFormat` for writing the flat vectors to the `.vec` file while performing ANN search. The graphs are built the same way as they are for other data types.

**Notes:**
- No logical changes have been made in the SIMD related code. It is only restructured and some files have been renamed based on the suggestion to merge the encoding and decoding code into a single namespace.
- Documentation is added for the SIMD C++ code
- The majority of the changes are the above 2 points (restructure and docs) and tests

### Related Issues
#2655 Feature Request
#2811 RFC

### Check List


src/main/java/org/opensearch/knn/index/codec/KNN9120Codec/KNN9120HnswHalfFloatVectorsFormat.java
```diff
@@ -0,0 +1,119 @@
...
+import static org.apache.lucene.codecs.lucene99.Lucene99HnswVectorsFormat.DEFAULT_BEAM_WIDTH;
+import static org.apache.lucene.codecs.lucene99.Lucene99HnswVectorsFormat.DEFAULT_MAX_CONN;
+import static org.apache.lucene.codecs.lucene99.Lucene99HnswVectorsFormat.DEFAULT_NUM_MERGE_WORKER;
+import static org.apache.lucene.codecs.lucene99.Lucene99HnswVectorsFormat.MAXIMUM_BEAM_WIDTH;
+import static org.apache.lucene.codecs.lucene99.Lucene99HnswVectorsFormat.MAXIMUM_MAX_CONN;
+import static org.opensearch.knn.index.engine.KNNEngine.getMaxDimensionByEngine;
+
+/**
+ * Custom KnnVectorsFormat implementation to support half-float vectors.  This class is mostly identical to
+ * {@link org.apache.lucene.codecs.lucene99.Lucene99HnswVectorsFormat}, however we use the custom however we
```
owenhalpert: Minor typo here with repeated phrase

src/main/java/org/opensearch/knn/index/codec/BasePerFieldKnnVectorsFormat.java
```diff
@@ -97,6 +99,12 @@ public KnnVectorsFormat getKnnVectorsFormatForField(final String field) {
             )
         ).fieldType(field);
 
+        // For now, we directly return the HalfFloatFlatVectorsFormat to perform Exact Search for FP16 based on the approximate threshold.
+        // This would change once the mapping field-level parameter `index` is added to disable graph creation.
```
owenhalpert: Let's mark this as a TODO with a link to Kaushika's PR or RFC

src/main/java/org/opensearch/knn/index/codec/BasePerFieldKnnVectorsFormat.java
```diff
@@ -168,6 +183,9 @@ private int getApproximateThresholdValue() {
         // This is private method and mapperService is already checked for null or valid instance type before this call
         // at caller, hence we don't need additional isPresent check here.
         final IndexSettings indexSettings = mapperService.get().getIndexSettings();
+        if (indexSettings == null) {
+            return KNNSettings.INDEX_KNN_ADVANCED_APPROXIMATE_THRESHOLD_DEFAULT_VALUE;
```
owenhalpert, line 193: Might have missed in a previous PR: why is this necessary now with your changes?

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -598,9 +598,10 @@ private boolean isFilteredExactSearchPreferred(final int filterIdsCount) {
     }
 
     private boolean isMDCGreaterThanFilterIdCnt(int filterIdsCount) {
-        return KNNConstants.MAX_DISTANCE_COMPUTATIONS >= filterIdsCount * (knnQuery.getVectorDataType() == VectorDataType.FLOAT
-            ? knnQuery.getQueryVector().length
-            : knnQuery.getByteQueryVector().length);
+        return KNNConstants.MAX_DISTANCE_COMPUTATIONS >= filterIdsCount * ((knnQuery.getVectorDataType() == VectorDataType.FLOAT
+            || knnQuery.getVectorDataType() == VectorDataType.HALF_FLOAT)
```
owenhalpert: Thanks for being diligent about these changes. In the future, if we add any other float types with similar functionality, it might be worth having a `knnQuery.getVectorDataType().isFloatType()` helper method. probably not necessary in this PR, but commenting to see what others think

src/main/java/org/opensearch/knn/jni/SimdFp16.java
```diff
@@ -21,20 +21,17 @@
  * Service to interact with SIMD jni layer. Class dependencies should be minimal
  * <p>
  * In order to compile C++ header file, run:
- * javac -h jni/include src/main/java/org/opensearch/knn/jni/SIMDNativeLibraryLoader.java
+ * javac -h jni/include src/main/java/org/opensearch/knn/jni/SimdFp16.java
  *      src/main/java/org/opensearch/knn/common/KNNConstants.java
  */
-public class SIMDNativeLibraryLoader {
+public class SimdFp16 {
```
owenhalpert, line 27: nit: should we capitalize SIMD?

src/main/java/org/opensearch/knn/index/KNNVectorScriptDocValues.java
```diff
@@ -173,7 +173,7 @@ protected T doGetValue() throws IOException {
      * @return An empty KNNVectorScriptDocValues object.
      */
     public static KNNVectorScriptDocValues<?> emptyValues(String fieldName, VectorDataType type) {
-        if (type == VectorDataType.FLOAT) {
+        if (type == VectorDataType.FLOAT || type == VectorDataType.HALF_FLOAT) {
```
0ctopus13prime: Can we consider adding a method `isFloatingDataType` in VectorDataType where evaluating this condition?
I think once we added BFP16, we need to change all the places likewise you did in this PR. but having it as a method, then I believe only we need less changes for floating family.

src/test/java/org/opensearch/knn/index/engine/faiss/FaissTests.java
```diff
@@ -45,11 +47,22 @@
 
 public class FaissTests extends KNNTestCase {
 
+    private final VectorDataType vectorDataType;
+
+    public FaissTests(VectorDataType vectorDataType) {
+        this.vectorDataType = vectorDataType;
+    }
+
+    @ParametersFactory
```
owenhalpert, line 56: Thanks for introducing this pattern. I like this.

owenhalpert (commented):
Minor comments. Thank you for your hard work! And thanks for breaking these big PRs down effectively.

src/main/java/org/opensearch/knn/jni/SimdFp16.java
ManasviGoyal: I did consider that but I was not sure what would be the best convention here. Possible alternatives can be SIMDFP16 or SIMDFp16

src/main/java/org/opensearch/knn/index/codec/BasePerFieldKnnVectorsFormat.java
ManasviGoyal: This is not related to any previous PR. 

The function was not checking for null cases which was throwing errors when no threshold setting was set. so I just added the case that when it is null use the default threshold value.

src/main/java/org/opensearch/knn/index/codec/BasePerFieldKnnVectorsFormat.java
```diff
@@ -149,6 +160,12 @@ public KnnVectorsFormat getKnnVectorsFormatForField(final String field) {
             return vectorsFormatSupplier.apply(knnVectorsFormatParams);
         }
 
+        if (mappedFieldType.getVectorDataType() == VectorDataType.HALF_FLOAT) {
+            throw new UnsupportedOperationException(
```
0ctopus13prime: Not sure I'm understanding this, but will `VectorDataType.FLOAT` be assigned when user configured FP16 encoding with Faiss engine?

jni/src/org_opensearch_knn_jni_SimdFp16.cpp
```diff
@@ -0,0 +1,68 @@
...
+
+#include "jni_util.h"
+#include "simd_fp16.h"
+
+static knn_jni::JNIUtil jniUtil;
+static const jint KNN_SIMDFP16_JNI_VERSION = JNI_VERSION_1_1;
+
+jint JNI_OnLoad(JavaVM* vm, void* reserved) {
+    JNIEnv* env;
+    if (vm->GetEnv((void**)&env, KNN_SIMDFP16_JNI_VERSION) != JNI_OK) {
```
0ctopus13prime, line 25: Why we need this check?

src/main/java/org/opensearch/knn/index/codec/util/KNNVectorAsCollectionOfHalfFloatsSerializer.java
```diff
@@ -71,8 +70,8 @@ public void byteToFloatArray(byte[] input, float[] output, int dimension, int of
             throw new IllegalArgumentException("Offset and dimension exceed input length.");
         }
 
-        if (SIMDDecoding.isSIMDSupported()) {
-            if (!SIMDDecoding.convertFP16ToFP32(input, output, dimension, offset)) {
+        if (SimdFp16.isSIMDSupported()) {
+            if (!SimdFp16.decodeFp16ToFp32(input, output, dimension, offset)) {
```
0ctopus13prime, line 74: I think if we're unlucky and had 16 bytes unaligned byte then `false` will be returned. 
If that's the case, we catch the exception and fallback to default? or search or indexing will halt?

src/main/java/org/opensearch/knn/index/engine/faiss/AbstractFaissMethod.java
```diff
@@ -50,7 +50,7 @@ protected PerDimensionValidator doGetPerDimensionValidator(
             return PerDimensionValidator.DEFAULT_BYTE_VALIDATOR;
         }
 
-        if (VectorDataType.FLOAT == vectorDataType) {
+        if (VectorDataType.FLOAT == vectorDataType || VectorDataType.HALF_FLOAT == vectorDataType) {
```
0ctopus13prime: Could you define a method `isFloatFamily` (you can give a better brand) within `VectorDataType`?

src/main/java/org/opensearch/knn/index/mapper/ModelFieldMapper.java
```diff
@@ -308,9 +308,8 @@ protected void parseCreateField(ParseContext context) throws IOException {
             int adjustedDimension = modelMetadata.getVectorDataType() == VectorDataType.BINARY
                 ? modelMetadata.getDimension() / Byte.SIZE
                 : modelMetadata.getDimension();
-            final VectorEncoding encoding = modelMetadata.getVectorDataType() == VectorDataType.FLOAT
-                ? VectorEncoding.FLOAT32
-                : VectorEncoding.BYTE;
+            final VectorEncoding encoding = (modelMetadata.getVectorDataType() == VectorDataType.FLOAT
```
0ctopus13prime: yeah... we definitely need `isFloatFamily` function to be defined.

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
0ctopus13prime: +1. Since BFP16 is coming, it would be better to have that helper function to minimize future changes.

src/main/java/org/opensearch/knn/index/codec/KNN9120Codec/KNN9120HnswHalfFloatVectorsFormat.java
```diff
@@ -0,0 +1,119 @@
...
+/**
+ * Custom KnnVectorsFormat implementation to support half-float vectors.  This class is mostly identical to
+ * {@link org.apache.lucene.codecs.lucene99.Lucene99HnswVectorsFormat}, however we use the custom
+ * {@link KNN990HalfFloatFlatVectorsFormat} for storing flat vectors as half-float.
+ */
+public final class KNN9120HnswHalfFloatVectorsFormat extends KnnVectorsFormat {
+
+    private final int maxConn;
+    private final int beamWidth;
+    private static final FlatVectorsFormat flatVectorsFormat = new KNN990HalfFloatFlatVectorsFormat();
```
shatejas, line 43: Instead of maintaining an new class which is more or less a copy of https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsFormat.java, Did we consider making flatVectorsFormat a member variable in `Lucene99HnswVectorsFormat`?

src/main/java/org/opensearch/knn/index/codec/BasePerFieldKnnVectorsFormat.java
ManasviGoyal: Thanks for pointing out. Added a condition so that HALF_FLOAT is supported for FAISS ANN when FP16 encoding is configured. For all other cases of Native engine ANN, an exception will be raised for HALF_FLOAT.

jni/src/org_opensearch_knn_jni_SimdFp16.cpp
ManasviGoyal: We need this check to prevent a crash from a null JNIEnv* because `GetEnv` can fail if the JVM doesn’t support the requested JNI version. It is same as used in other libraries like Faiss, JNICommons.

src/main/java/org/opensearch/knn/index/codec/BasePerFieldKnnVectorsFormat.java
```diff
@@ -97,6 +100,14 @@ public KnnVectorsFormat getKnnVectorsFormatForField(final String field) {
             )
         ).fieldType(field);
 
+        // TODO: Use the new `index` parameter to disable graph creation for exact search
+        // https://github.com/opensearch-project/k-NN/pull/2768
+        // For now, we directly return the KNN990HalfFloatFlatVectorsFormat to perform Exact Search for FP16 based on the approximate
+        // threshold.
+        if (getApproximateThresholdValue() < 0 && mappedFieldType.getVectorDataType() == VectorDataType.HALF_FLOAT) {
+            return new KNN990HalfFloatFlatVectorsFormat(FlatVectorScorerUtil.getLucene99FlatVectorsScorer());
+        }
```
shatejas, line 109: Whats the behavior when the approximate setting is changed for say faiss engine?

src/main/java/org/opensearch/knn/index/codec/KNN9120Codec/KNN9120HnswHalfFloatVectorsFormat.java
ManasviGoyal: Yes. Since, based on my knowledge the timeline to get a change merged in Lucene can be long, for now I have added it as a new class. I can make a separate PR for this change to Lucene and once that is merged we can remove this class. Similarly we can also remove `KNN9120HnswBinaryVectorsFormat` class. Thanks!

src/main/java/org/opensearch/knn/index/codec/BasePerFieldKnnVectorsFormat.java
ManasviGoyal: The behavior here is that if the threshold value is less than 0, no graphs built irrespective of engine. This forces exact search. For other cases graphs are built. The table in the PR description has more details.

This is sort of a temporary workaround for now (mentioned in TODO) as @kaushikau's project adds a index parameter to enable and disable graph creation which can be directly used here. We can just replace the `getApproximateThresholdValue() < 0` check with `index == false` once her change is merged.

src/main/java/org/opensearch/knn/index/codec/KNN9120Codec/KNN9120HnswHalfFloatVectorsFormat.java
ManasviGoyal: Made a PR in Lucene https://github.com/apache/lucene/pull/15090

src/main/java/org/opensearch/knn/index/codec/BasePerFieldKnnVectorsFormat.java
shatejas: This is a dynamic setting and users can set it -1 initially and then can have an actual threshold. For FAISS, they can index with -1 and later change it to say 1000, I think once they change it to 1000 it will start throwing validation errors. Coupling half_float with approx threshold setting doesn't seem right, we need a non-dynamic setting to do it. 

I think its best to block half float completely for faiss engine to start with

src/main/java/org/opensearch/knn/index/codec/KNN9120Codec/KNN9120HnswHalfFloatVectorsFormat.java
shatejas: Please add a TODO to remove this class and link the lucene PR

ManasviGoyal: Added already

PR #2868, opened 2025-09-08
Support native Maximal Marginal Relevance

### LINKED ISSUE #2804: [FEATURE]Native Support for Maximal Marginal Relevance (MMR) in Vector Search
**Is your feature request related to a problem? Please describe**
I'd like to request native support for Maximal Marginal Relevance (MMR) in OpenSearch vector search.

MMR is a widely used technique that balances query relevance and result diversity when selecting top-k results. Instead of returning the top-k nearest neighbors (which may be semantically redundant), MMR helps return a diverse yet relevant set of results — improving downstream performance in tasks like:

Retrieval-Augmented Generation (RAG)
Recommender systems
Search UX (less repetitive results)
This feature is especially useful in vector databases, where top-k similarity often results in clustered/redundant results. While this can be implemented client-side, it requires:

Fetching more than k documents (e.g., 3×k)
Computing similarities between each result
Re-ranking locally using custom code
Having MMR support built into OpenSearch would:

Reduce latency for large-scale systems
Simplify development (especially for those using OpenSearch via APIs or clients like LangChain)
For reference, Qdrant recently added native support for MMR ([LinkedIn post](https://www.linkedin.com/posts/zayarni_qdrant-maximal-marginal-relevance-mmr-activity-7354066357283434496-si1H)) and it has proven valuable in real-world scenarios. This could make OpenSearch competitive with other vector stores which offers native MMR.

**Describe the solution you'd like**
Query OpenSearch for a large number of candidate vectors (fetch_k)
Apply MMR algorithm to re-rank and select diverse results
Return top-k MMR-ranked documents
MMR Equation:
For each candidate d_i not yet selected:
MMR(d_i) = λ * Sim(d_i, query) - (1 - λ) * max_{d_j ∈ selected} Sim(d_i, d_j)

Where:
λ is the relevance/diversity trade-off (e.g., 0.7 = more relevance)
Sim = similarity (e.g., cosine similarity)

**Describe alternatives you've considered**
Client-side MMR using cosine similarity and re-ranking (works, but inefficient at scale)
Hybrid search with rerankers (does not solve semantic redundancy)

**Additional context**
Joinal Ahmed and Nikhil Rana presented on this topic at Europe Conference, the implementation of MMR was done externally OpenSearch. https://www.youtube.com/watch?v=rBcQLh77J_Y

LangChain’s MMR-based retriever: https://docs.langchain.com/docs/modules/data_connection/retrievers/vectorstores#maximal-marginal-relevance

bzhangam:
### Description
Support native Maximal Marginal Relevance. It relies on the system generated search pipeline to auto generate:
 - MMR oversample search request processor. It will modify the search request to oversample for MMR and also resolve the space type for response processor to use.
- MMR rerank search response processor. It will rerank the hits based on Maximal Marginal Relevance.

We also introduce the `MMRQueryTransformer` to support different transform logic for different query builder. Here we just implement the transformer for the knn query for now. And we will implement the transformer in neural plugin to handle the neural query properly.

Checks will fail until this [PR](https://github.com/opensearch-project/OpenSearch/pull/19128) is merged in core.

example query with mmr:
```
{
	"_source": {
		"excludes": ["passage_embedding"]
	},
	"query": {
		"knn": {
				"passage_embedding": {
					"vector": [1.0, 2.0, 3.0],
					"k": 10
				}
		}
  },
	"ext":{
		"mmr":{
			"diversity": 0.5,
			"candidates": 20
		}
	}
}
```

Check [benchmark](https://github.com/opensearch-project/k-NN/issues/2869#issuecomment-3324856131)
### Related Issues
Resolves #2804

### Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ### Bug Fixes
 * Use queryVector length if present in MDC check [#2867](https://github.com/opensearch-project/k-NN/pull/2867)
+* Fix derived source deserialization bug on invalid documents [#2882](https://github.com/opensearch-project/k-NN/pull/2882)
+
 ### Refactoring
 * Refactored the KNN Stat files for better readability.
 
--- a/src/main/java/org/opensearch/knn/index/codec/KNN10010Codec/KNN10010DerivedSourceStoredFieldsWriter.java
+++ b/src/main/java/org/opensearch/knn/index/codec/KNN10010Codec/KNN10010DerivedSourceStoredFieldsWriter.java
@@ -5,6 +5,7 @@
 
 package org.opensearch.knn.index.codec.KNN10010Codec;
 
+import lombok.extern.log4j.Log4j2;
 import org.apache.lucene.codecs.StoredFieldsWriter;
 import org.apache.lucene.index.FieldInfo;
 import org.apache.lucene.index.MergeState;
@@ -15,6 +16,7 @@ import org.opensearch.common.io.stream.BytesStreamOutput;
 import org.opensearch.common.xcontent.XContentHelper;
 import org.opensearch.common.xcontent.support.XContentMapValues;
 import org.opensearch.core.common.bytes.BytesReference;
+import org.opensearch.core.compress.NotXContentException;
 import org.opensearch.core.xcontent.MediaType;
 import org.opensearch.core.xcontent.MediaTypeRegistry;
 import org.opensearch.core.xcontent.XContentBuilder;
@@ -29,6 +31,7 @@ import java.util.Objects;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 
+@Log4j2
 public class KNN10010DerivedSourceStoredFieldsWriter extends StoredFieldsWriter {
 
     private final StoredFieldsWriter delegate;
@@ -113,11 +116,25 @@ public class KNN10010DerivedSourceStoredFieldsWriter extends StoredFieldsWriter 
             // here because writeField may not always follow the preindex step, so it might have the vector (think
             // merge). This may be overly cautious and we can remove/optimize it in the future. For now, its a safety
             // net.
-            Tuple<? extends MediaType, Map<String, Object>> mapTuple = XContentHelper.convertToMap(
-                BytesReference.fromByteBuffer(ByteBuffer.wrap(bytesRef.bytes, bytesRef.offset, bytesRef.length)),
-                true,
-                MediaTypeRegistry.JSON
-            );
+            Tuple<? extends MediaType, Map<String, Object>> mapTuple;
+
+            // If the content is not XContent, skip writing altogether. See:
+            // https://github.com/opensearch-project/k-NN/issues/2880
+            try {
+                mapTuple = XContentHelper.convertToMap(
+                    BytesReference.fromByteBuffer(ByteBuffer.wrap(bytesRef.bytes, bytesRef.offset, bytesRef.length)),
+                    true,
+                    MediaTypeRegistry.JSON
+                );
+            } catch (NotXContentException e) {
+                log.warn(
+                    "Encountered NotXContent while deserializing _source field. Instead found String: [{}]",
+                    // Limit max string length in case of long bytes object
+                    new String(bytesRef.bytes, 0, Math.min(bytesRef.bytes.length, 512)),
+                    e
+                );
+                return;
+            }
             Map<String, Object> filteredSource = vectorMask.apply(mapTuple.v2());
             BytesStreamOutput bStream = new BytesStreamOutput();
             MediaType actualContentType = mapTuple.v1();
--- a/src/test/java/org/opensearch/knn/integ/DerivedSourceIT.java
+++ b/src/test/java/org/opensearch/knn/integ/DerivedSourceIT.java
@@ -17,6 +17,7 @@ import org.opensearch.knn.Pair;
 import org.opensearch.knn.index.VectorDataType;
 import org.opensearch.knn.common.annotation.ExpectRemoteBuildValidation;
 
+import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Locale;
@@ -138,6 +139,48 @@ public class DerivedSourceIT extends DerivedSourceTestCase {
         }
 
         prepareOriginalIndices(indexConfigContexts);
+    }
+
+    /**
+     * Tests that kNN handles bad documents the same when derived source is enabled and disabled.
+     * @throws java.io.IOException
+     */
+    public void testDerivedSource_HandlesInvalidDocuments() throws IOException {
+        List<DerivedSourceUtils.IndexConfigContext> indexConfigContexts = getCustomAnalyzerIndexContexts("derivedit", true, true);
+
+        assertTrue(1 < indexConfigContexts.size());
+        DerivedSourceUtils.IndexConfigContext derivedSourceEnabledContext = indexConfigContexts.get(0);
+        DerivedSourceUtils.IndexConfigContext derivedSourceDisabledContext = indexConfigContexts.get(1);
+        createKnnIndex(
+            derivedSourceEnabledContext.indexName,
+            derivedSourceEnabledContext.getSettings(),
+            derivedSourceEnabledContext.getMapping()
+        );
+        createKnnIndex(
+            derivedSourceDisabledContext.indexName,
+            derivedSourceDisabledContext.getSettings(),
+            derivedSourceDisabledContext.getMapping()
+        );
+        for (int i = 0; i < derivedSourceDisabledContext.docCount; i++) {
+            String doc1 = derivedSourceEnabledContext.buildDoc();
+            String doc2 = derivedSourceDisabledContext.buildDoc();
+            assertEquals(doc1, doc2);
+            boolean dsEnabledException = false;
+            boolean dsDisabledException = false;
+            try {
+                addKnnDoc(derivedSourceEnabledContext.getIndexName(), String.valueOf(i + 1), doc1);
+            } catch (ResponseException e) {
+                assertTrue(e.getMessage().contains("number_format_exception"));
+                dsEnabledException = true;
+            }
+            try {
+                addKnnDoc(derivedSourceDisabledContext.getIndexName(), String.valueOf(i + 1), doc2);
+            } catch (ResponseException e) {
+                assertTrue(e.getMessage().contains("number_format_exception"));
+                dsDisabledException = true;
+            }
+            assertEquals(dsEnabledException, dsDisabledException);
+        }
     }
 
     /**
--- a/src/testFixtures/java/org/opensearch/knn/DerivedSourceTestCase.java
+++ b/src/testFixtures/java/org/opensearch/knn/DerivedSourceTestCase.java
@@ -9,11 +9,13 @@ import lombok.SneakyThrows;
 import org.apache.hc.core5.http.io.entity.EntityUtils;
 import org.opensearch.client.Request;
 import org.opensearch.client.Response;
+import org.opensearch.common.settings.Settings;
 import org.opensearch.common.xcontent.XContentFactory;
 import org.opensearch.core.rest.RestStatus;
 import org.opensearch.core.xcontent.XContentBuilder;
 import org.opensearch.index.query.MatchAllQueryBuilder;
 import org.opensearch.index.query.QueryBuilder;
+import org.opensearch.knn.index.KNNSettings;
 import org.opensearch.knn.index.VectorDataType;
 
 import java.io.IOException;
@@ -23,6 +25,8 @@ import java.util.Locale;
 import java.util.Map;
 import java.util.Random;
 import java.util.function.Supplier;
+
+import static org.opensearch.knn.TestUtils.BWC_VERSION;
 
 public class DerivedSourceTestCase extends KNNRestTestCase {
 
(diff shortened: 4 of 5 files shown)


src/main/java/org/opensearch/knn/index/engine/SpaceTypeResolver.java
```diff
@@ -38,15 +37,11 @@ private SpaceTypeResolver() {}
     public SpaceType resolveSpaceType(
         final KNNMethodContext knnMethodContext,
         final String topLevelSpaceTypeString,
-        final Settings indexSettings,
         final VectorDataType vectorDataType
     ) {
         SpaceType methodSpaceType = getSpaceTypeFromMethodContext(knnMethodContext);
         SpaceType topLevelSpaceType = getSpaceTypeFromString(topLevelSpaceTypeString);
 
-        // If we failed to find space type from both method context and top level
```
bzhangam, line 47: We cleaned up this logic in this [PR](https://github.com/opensearch-project/k-NN/commit/e7d5304b85d97d0b24f33552481423c6cefcf009#diff-a294ae365e1e18e6fea5d7f4964ecd4726b029bac6006fc441423f05b51fca5d) but didn't clean up the comments which can cause confusion. So cleaned it up here since this PR also needs to resolve the space type.

src/main/java/org/opensearch/knn/index/util/IndexMappingUtil.java
```diff
@@ -0,0 +1,50 @@
...
+
+            Map<String, Object> properties = (Map<String, Object>) propertiesObj;
+            Object fieldConfig = properties.get(part);
+            if (!(fieldConfig instanceof Map)) {
+                return null; // field not found
+            }
+
+            current = (Map<String, Object>) fieldConfig;
+
+            if (i == parts.length - 1) {
```
heemin32: We don’t need this code block. This condition will always be true before exiting the for loop.
Otherwise, you better call `break;` instead of `return null;`

src/main/java/org/opensearch/knn/index/util/IndexMappingUtil.java
```diff
@@ -0,0 +1,50 @@
...
+                return current; // reached target field
+            }
+        }
+        return null;
```
heemin32:
```suggestion
        return current;
```

src/main/java/org/opensearch/knn/search/processor/mmr/MMRConstants.java
```diff
@@ -0,0 +1,13 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.knn.search.processor.mmr;
+
+/**
+ * Constants for MMR
+ */
+public class MMRConstants {
+    public static final String MMR_RERANK_CONTEXT = "mmr.rerank_context";
```
heemin32: Why only this constant is here but all the others in `KNNConstants`?

src/main/java/org/opensearch/knn/search/processor/mmr/MMROverSampleProcessor.java
```diff
@@ -0,0 +1,333 @@
...
+import static org.opensearch.knn.search.processor.mmr.MMRUtil.*;
+
+/**
+ * A system generated search request processor for MMR. It will transform the search request to oversample and also
+ * collect and store the info in the PipelineProcessingContext for MMR later in a search response processor.
+ */
+public class MMROverSampleProcessor implements SearchRequestProcessor, SystemGeneratedProcessor {
+    public static final String TYPE = "mmr_over_sample";
+    public static final String DESCRIPTION = "This is a system generated processor that will modify the query size and"
+        + "k of the knn/neural query to oversample for Maximal Marginal Relevance rerank.";
```
heemin32:
```suggestion
        + "k of the knn query to oversample for Maximal Marginal Relevance rerank.";
```

src/main/java/org/opensearch/knn/plugin/KNNPlugin.java
```diff
@@ -250,6 +261,9 @@ public Collection<Object> createComponents(
 
         clusterService.addListener(TrainingJobClusterStateListener.getInstance());
 
+        // register mmr query transformer for knn query
+        MMRQueryTransformerRegistry.getInstance().register(KNNQueryBuilder.NAME, new MMRKnnQueryTransformer(client));
```
heemin32: Shouldn’t we follow this pattern? https://github.com/opensearch-project/job-scheduler/blob/main/README.md#getting-started

I see a few issues with the current registration approach:

Race condition – We could make the register method synchronous so it is small thing.
Overriding – One plugin can easily override another plugin’s MMR query transformer.
Registration timing – We don’t control when the MMR query transformer is registered and instead rely on each plugin’s timing, which might not be ideal. For example, if we block overriding, it wouldn’t be detected at boot time but only when the actual registration happens.

src/main/java/org/opensearch/knn/search/processor/mmr/MMRKnnQueryTransformer.java
```diff
@@ -0,0 +1,62 @@
...
+ * A transformer to transform the knn query for MMR
+ */
+@AllArgsConstructor
+public class MMRKnnQueryTransformer implements MMRQueryTransformer<KNNQueryBuilder> {
+    private final Client client;
+
+    @Override
+    public void transform(KNNQueryBuilder queryBuilder, ActionListener<Void> listener, MMRTransformContext mmrTransformContext) {
+        try {
+            if (queryBuilder.getMaxDistance() == null && queryBuilder.getMinScore() == null) {
```
heemin32, line 23: Does mmr still work with range query? Do we still increase the size value and rescore for range query?

src/main/java/org/opensearch/knn/search/processor/mmr/MMRUtil.java
```diff
@@ -0,0 +1,497 @@
...
+                        baseError,
+                        fieldPath,
+                        current.getClass().getName()
+                    )
+                );
+            }
+        }
+
+        // Should never reach here
+        throw new IllegalArgumentException(String.format("%s: unexpected error resolving field path [%s].", baseError, fieldPath));
```
heemin32: IllegalArgumentException should be used when user input is invalid I think. Here, we should use InternalStateException or something like that.

src/main/java/org/opensearch/knn/index/util/IndexMappingUtil.java
bzhangam: ok

src/main/java/org/opensearch/knn/search/processor/mmr/MMRConstants.java
bzhangam: Didn't notice we have KNNConstants when I created this. I can move it to the KNNConstants.

src/main/java/org/opensearch/knn/search/processor/mmr/MMROverSampleProcessor.java
bzhangam: We will use this processor to handle the neural query.

src/main/java/org/opensearch/knn/plugin/KNNPlugin.java
bzhangam: Yeah. We can use ExtensionLoader to load the transformers.

src/main/java/org/opensearch/knn/search/processor/mmr/MMRKnnQueryTransformer.java
bzhangam: Yeah I think we just need to increase the size value to oversample. I'm not sure what do you mean about rescore. For range query even we need to oversample I don't think we should relax the range.

src/main/java/org/opensearch/knn/search/processor/mmr/MMRUtil.java
bzhangam: yeah you are right

src/main/java/org/opensearch/knn/search/processor/mmr/MMROverSampleProcessor.java
heemin32: knn plugin shouldn't know about neural query. It does not have any dependency on neural plugin.

src/main/java/org/opensearch/knn/search/processor/mmr/MMRKnnQueryTransformer.java
heemin32: Got it. So we do support mmr for range query as well. Thanks.

src/main/java/org/opensearch/knn/search/processor/mmr/MMROverSampleProcessor.java
bzhangam: ok

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java
```diff
@@ -424,7 +424,6 @@ public Mapper.Builder<?> parse(String name, Map<String, Object> node, ParserCont
                 final SpaceType resolvedSpaceType = SpaceTypeResolver.INSTANCE.resolveSpaceType(
                     builder.originalParameters.getKnnMethodContext(),
                     builder.topLevelSpaceType.get(),
-                    parserContext.getSettings(),
```
Vikasht34, line 427: Why removing?

src/main/java/org/opensearch/knn/plugin/KNNPlugin.java
```diff
@@ -351,6 +367,27 @@ public void onIndexModule(IndexModule indexModule) {
         }
     }
 
+    @Override
+    public void loadExtensions(ExtensionLoader loader) {
+        // knn plugin cannot extend itself so we have to manually load the transformer implemented in knn plugin
+        mmrQueryTransformers.put(KNNQueryBuilder.NAME, new MMRKnnQueryTransformer());
+        for (MMRQueryTransformer<?> transformer : loader.loadExtensions(MMRQueryTransformer.class)) {
```
Vikasht34, line 379: Why it has to be in loop?

src/main/java/org/opensearch/knn/plugin/KNNPlugin.java
```diff
@@ -351,6 +367,27 @@ public void onIndexModule(IndexModule indexModule) {
...
+            String queryName = transformer.getQueryName();
+            if (mmrQueryTransformers.containsKey(queryName)) {
+                throw new IllegalStateException(
```
Vikasht34, line 382: What's the point to throw this exception to user , Can we check the condition in reverse way , if it's not present then only put in Map?

src/main/java/org/opensearch/knn/plugin/rest/RestTrainModelHandler.java
```diff
@@ -177,7 +177,6 @@ && ensureSpaceTypeNotSet(topLevelSpaceType)) {
         SpaceType resolvedSpaceType = SpaceTypeResolver.INSTANCE.resolveSpaceType(
             knnMethodContext,
             topLevelSpaceType.getValue(),
-            null,
```
Vikasht34, line 180: ?

src/main/java/org/opensearch/knn/search/extension/MMRSearchExtBuilder.java
```diff
@@ -0,0 +1,261 @@
...
+    // should auto resolve it from the index mapping.
+    public static final ParseField VECTOR_FIELD_SPACE_TYPE_FIELD = new ParseField(VECTOR_FIELD_SPACE_TYPE);
+
+    private Float diversity;
+    private Integer candidates;
+    private String vectorFieldPath;
+    private VectorDataType vectorFieldDataType;
+    private SpaceType spaceType;
+
+    public static class Builder {
```
Vikasht34, line 70: There is no real benefit of having builder here , Since Parse is the entry point and there is no Partial Construction.

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java
bzhangam: As I mentioned above in this [PR](https://github.com/opensearch-project/k-NN/commit/e7d5304b85d97d0b24f33552481423c6cefcf009#diff-a294ae365e1e18e6fea5d7f4964ecd4726b029bac6006fc441423f05b51fca5d) we already remove the logic checking the setting to resolve the space type. So clean up this code.

src/main/java/org/opensearch/knn/plugin/KNNPlugin.java
bzhangam: Since there can be multiple transformers from other plugins.

src/main/java/org/opensearch/knn/plugin/KNNPlugin.java
bzhangam: If plugin A defines transformer A for query Q, and plugin B also defines transformer B for the same query Q, we cannot determine which transformer will ultimately be mapped to query Q. To prevent this conflict, we should throw an exception when multiple transformers are registered for the same query type. Only one transformer should be associated with a given query type, and plugins A and B must coordinate to resolve the duplication.

src/main/java/org/opensearch/knn/plugin/rest/RestTrainModelHandler.java
bzhangam: Same as above. I cleaned up the resolveSpaceType function to remove the unused parameter.

src/main/java/org/opensearch/knn/search/extension/MMRSearchExtBuilder.java
bzhangam: True. But I think it's ok to keep it since we already have it. And I also think it's a little bit more cleaner. And in future we may want to support partial construct.

heemin32:
> > Will Continue reviewing the other files, Mean While can you attach the benchmarking/Testing Details here.
> 
> I think I can do a benchmark. But not sure what's the testing details you are looking for?

You could use this workload. https://github.com/opensearch-project/opensearch-benchmark-workloads/blob/main/vectorsearch/params/corpus/10million/faiss-cohere-768-dp.json

and paste the result with and without mmr.

src/main/java/org/opensearch/knn/plugin/KNNPlugin.java
Vikasht34: Can you explain more on this , Why other plugings extensions would be loaded in KNN Plugin?

src/main/java/org/opensearch/knn/plugin/KNNPlugin.java
Vikasht34: Since we have already the registry, how other plugins would come in picture here .

Vikasht34:
@bzhangam We need to make sure all the Integs are passing ..In case if it's not becuase of your change , Please ensure you work with respetive folks to sign off , we can't merge PR with Integs failed.

src/main/java/org/opensearch/knn/plugin/KNNPlugin.java
bzhangam: You can check this [PR](https://github.com/opensearch-project/neural-search/pull/1567) where we implement the transformer for the neural query. We want to use this pattern so that we can keep the core function in knn plugin but also allow neural plugin to inject its specific logic to process the neural query.

src/main/java/org/opensearch/knn/plugin/KNNPlugin.java
bzhangam: This pattern is designed to allow other plugins to extend an interface defined in the KNN plugin.

src/main/java/org/opensearch/knn/search/processor/mmr/MMROverSampleProcessor.java
```diff
@@ -0,0 +1,403 @@
...
+        }
+        boolean fetchingAll = sourceContext.fetchSource();
+        boolean noIncludes = sourceContext.includes().length == 0;
+        boolean noExcludes = sourceContext.excludes().length == 0;
+        return fetchingAll && noIncludes && noExcludes;
+    }
+
+    private void handleDisabledStoredFields(SearchRequest request, MMRRerankContext mmrContext, FetchSourceContext currentSourceContext) {
+        if (currentSourceContext != null) {
+            // stored_fields = _none_ + explicit _source → let core handle this case, this is an invalid combo and core
```
heemin32: Still let's throw exception here as well to be safe?

src/main/java/org/opensearch/knn/search/processor/mmr/MMRRerankProcessor.java
```diff
@@ -0,0 +1,307 @@
...
+        float diversity,
+        int targetSize,
+        boolean isFloatVector
+    ) {
+        List<SearchHit> selected = new ArrayList<>();
+        Map<Long, Float> simCache = new ConcurrentHashMap<>();
+
+        while (selected.size() < targetSize && !candidates.isEmpty()) {
+
+            Optional<SearchHit> bestCandidateOpt = candidates.parallelStream().max(Comparator.comparingDouble(candidate -> {
```
heemin32, line 201: Cannot tell for sure but there could be a way to improve this code part to utilize SIMD. Doesn't have to be handled in this PR though.

src/main/java/org/opensearch/knn/search/extension/MMRSearchExtBuilder.java
```diff
@@ -0,0 +1,285 @@
...
+                    throw unsupportedField(parser, currentFieldName);
+                }
+            } else {
+                throw unsupportedField(parser, currentFieldName);
+            }
+        }
+        return builder.build();
+    }
+
+    private static void ensureMMRProcessorsEnabled() {
```
heemin32, line 260: This implementation feels odd. I believe this check should be handled in the OpenSearch core rather than within each plugin. Otherwise, there’s no reliable way to ensure that all plugins consistently follow it.

bzhangam: This the plugin specific logic that we want to ensure the required factories are enabled when we find the mmr extension is set in the search request.

src/main/java/org/opensearch/knn/search/processor/mmr/MMROverSampleProcessor.java
bzhangam: ok

src/main/java/org/opensearch/knn/search/extension/MMRSearchExtBuilder.java
heemin32: Might be better for core to provide a method to check if certain factory is enabled or not?

Vikasht34: I agree with @heemin32 , I had same question but I got convinced through registry , since Knn was owning .. But it makes more sense to core to vend out interface which can be checked at plugin side.

heemin32: We can address this in a follow-up PR. Please open an issue in core and KNN so we don’t lose track.

bzhangam: ok We can have `SearchPipelineService` to expose this method since anyway it's tracking the enabled factories.

bzhangam: crated an issue to track it - https://github.com/opensearch-project/OpenSearch/issues/19427

Vikasht34 (approved):

heemin32 (approved):

Vikasht34 (approved):

Issue #2836, opened 2025-08-07
[BUG] Date range filters with knn query fails: [knn] requires exactly one of k, distance or score to be set


jzhao-b:
**What is the bug?**

Using a date range in the knn query filter produces the following error:
```
{
  "error": {
    "root_cause": [
      {
        "type": "illegal_argument_exception",
        "reason": "[knn] requires exactly one of k, distance or score to be set"
      }
    ],
    "type": "search_phase_execution_exception",
    "reason": "all shards failed",
    "phase": "query",
    "grouped": true,
    "failed_shards": [
      {
        "shard": 0,
        "index": "jz_test.v1",
        "node": "tCApmPvNTVSCVFAe2tRj9w",
        "reason": {
          "type": "illegal_argument_exception",
          "reason": "[knn] requires exactly one of k, distance or score to be set"
        }
      }
    ],
    "caused_by": {
      "type": "illegal_argument_exception",
      "reason": "[knn] requires exactly one of k, distance or score to be set",
      "caused_by": {
        "type": "illegal_argument_exception",
        "reason": "[knn] requires exactly one of k, distance or score to be set"
      }
    }
  },
  "status": 400
}
```

**How can one reproduce the bug?**
The minimal reproduction I've found, using an AWS opensearch 2.19 deployment:

Create an index `jz_test.v1` with settings:
```
{
  "index.knn": "true",
  "index.knn.algo_param.ef_search": "100",
  "index.number_of_replicas": "0",
  "index.number_of_shards": "1"
}
```

and mappings:
```
{
  "properties": {
    "date_field": {
      "type": "date"
    },
    "my_embedding": {
      "dimension": 2,
      "method": {
        "engine": "lucene",
        "space_type": "cosinesimil",
        "name": "hnsw",
        "parameters": {
          "ef_construction": 100,
          "m": 16
        }
      },
      "type": "knn_vector"
    },
    "numeric_field": {
      "type": "integer"
    }
  }
}
```

issue the request to run a knn search over `my_embedding` subject to `date_field` as a range filter and observe that it fails with the above error.
```
GET jz_test.v1/_search
{
  "query": {
    "knn": {
      "my_embedding": {
        "filter": {
          "range": {
            "date_field": {
              "gte": "2020-01-01"
            }
          }
        },
        "max_distance": 0.1,
        "vector": [1, 0]
      }
    }
  }
}
```

good to note that a similar query using "k" instead of "max_distance" behaves as expected...

[...]

PR #2871, opened 2025-09-10
Adds a process to make initially null k value null again on rewrite.

MrFlap:
### Description
The rewritten query references the QueryRewriteContext after a null `k` is rewritten as zero [here](https://github.com/opensearch-project/k-NN/blob/53e2c3a303e29162eb516d7db42a432bc1ab5966/src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java#L221). This change reverts `k` to null so that the query can be validated. This change also makes sure that `k = 0` is invalid for an initial query.

### Related Issues
Resolves [#2836](https://github.com/opensearch-project/k-NN/issues/2836)

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
+++ b/src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
@@ -739,11 +739,12 @@ public class KNNQueryBuilder extends AbstractQueryBuilder<KNNQueryBuilder> imple
         QueryBuilder rewrittenFilter;
         if (Objects.nonNull(filter)) {
             rewrittenFilter = filter.rewrite(queryShardContext);
+            Integer nulledK = this.k == 0 ? null : this.k;
             if (rewrittenFilter != filter) {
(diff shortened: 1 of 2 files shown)


src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
```diff
@@ -739,11 +739,12 @@ protected QueryBuilder doRewrite(QueryRewriteContext queryShardContext) throws I
         QueryBuilder rewrittenFilter;
         if (Objects.nonNull(filter)) {
             rewrittenFilter = filter.rewrite(queryShardContext);
+            Integer nulledK = this.k == 0 ? null : this.k;
```
Vikasht34, line 742: Please have better naming for Integer K

shatejas, line 742:
```suggestion
            Integer rewrittenK = this.k == 0 ? null : this.k;
```

Issue #2876, opened 2025-09-16
FIX github CI for mac os


navneet1v:
## Description
This issue tracks the fix for the GH actions on mac os. 

This is not a burning issue for now since OpenSearch doesn't do a release for mac os.

navneet1v:
We can try doing this:

```
    jobs:
      build:
        runs-on: macos-latest
        steps:
          - uses: actions/checkout@v4
          - name: Build
            env:
              CC: clang
              CXX: clang++
            run: |
              # Your build commands here
              # e.g., cmake . && make
```

or 
```
    jobs:
      build:
        runs-on: macos-latest
        steps:
          - uses: actions/checkout@v4
          - name: Install GCC
            run: brew install gcc
          - name: Build with specific GCC
            env:
              CC: /usr/local/bin/gcc-11 # Adjust version as needed
              CXX: /usr/local/bin/g++-11 # Adjust version as needed
            run: |
              # Your build commands here
```

to see if that resolves the issue

PR #2878, opened 2025-09-17
Fix KNN build due to Lucene 10.3 upgrade

navneet1v:
### Description
This change allows us to fix the failing builds which was happening to Lucene 10.3 and other changes in core.

### Changes done:
1. Created a new codec KNN1030Codec class for Lucene 10.3 version.
2. Fixed the interface for search for kNN queries to use AcceptedDocs rather than BitSets
3. Fix and cleaned up the guava dependencies for grpc.

### Known Issue
MacOS CI will fail and it is unrelated to this PR. Will raise another PR for the same. GH issue: https://github.com/opensearch-project/k-NN/issues/2876

### Related Issues
https://github.com/opensearch-project/k-NN/issues/2888

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
+++ b/src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
@@ -20,10 +20,15 @@ import org.apache.lucene.index.ByteVectorValues;
 import org.apache.lucene.index.FieldInfo;
 import org.apache.lucene.index.FloatVectorValues;
 import org.apache.lucene.index.SegmentReadState;
+import org.apache.lucene.search.AcceptDocs;
 import org.apache.lucene.search.KnnCollector;
 import org.apache.lucene.search.ScoreDoc;
 import org.apache.lucene.search.TopDocs;
 import org.apache.lucene.search.TotalHits;
+import org.apache.lucene.store.DataAccessHint;
+import org.apache.lucene.store.FileDataHint;
+import org.apache.lucene.store.FileTypeHint;
+import org.apache.lucene.store.IOContext;
 import org.apache.lucene.util.Bits;
 import org.apache.lucene.util.IOSupplier;
 import org.apache.lucene.util.IOUtils;
@@ -63,12 +68,13 @@ public class NativeEngines990KnnVectorsReader extends KnnVectorsReader {
     private final SegmentReadState segmentReadState;
     private final List<String> cacheKeys;
     private volatile Map<String, VectorSearcherHolder> vectorSearchers;
+    private final IOContext ioContext;
 
     public NativeEngines990KnnVectorsReader(final SegmentReadState state, final FlatVectorsReader flatVectorsReader) {
         this.flatVectorsReader = flatVectorsReader;
         this.segmentReadState = state;
         this.cacheKeys = getVectorCacheKeysFromSegmentReaderState(state);
-
+        ioContext = state.context.withHints(FileTypeHint.DATA, FileDataHint.KNN_VECTORS, DataAccessHint.RANDOM);
         loadCacheKeyMap();
         fillVectorSearcherTable();
     }
@@ -135,7 +141,7 @@ public class NativeEngines990KnnVectorsReader extends KnnVectorsReader {
      *                     if they are all allowed to match.
      */
     @Override
-    public void search(String field, float[] target, KnnCollector knnCollector, Bits acceptDocs) throws IOException {
+    public void search(String field, float[] target, KnnCollector knnCollector, AcceptDocs acceptDocs) throws IOException {
         // TODO: This is a temporary hack where we are using KNNCollector to initialize the quantization state.
         if (knnCollector instanceof QuantizationConfigKNNCollector) {
             String cacheKey = quantizationStateCacheKeyPerField.get(field);
@@ -185,7 +191,7 @@ public class NativeEngines990KnnVectorsReader extends KnnVectorsReader {
      *                     if they are all allowed to match.
      */
     @Override
-    public void search(String field, byte[] target, KnnCollector knnCollector, Bits acceptDocs) throws IOException {
+    public void search(String field, byte[] target, KnnCollector knnCollector, AcceptDocs acceptDocs) throws IOException {
         // searching with byte vector is not supported by ADC.
         if (trySearchWithMemoryOptimizedSearch(field, target, knnCollector, acceptDocs, false)) {
             return;
@@ -239,7 +245,7 @@ public class NativeEngines990KnnVectorsReader extends KnnVectorsReader {
         final String field,
         final Object target,
         final KnnCollector knnCollector,
-        final Bits acceptDocs,
+        final AcceptDocs acceptDocs,
         final boolean isFloatVector
     ) throws IOException {
         // Try with memory optimized searcher
@@ -363,7 +369,7 @@ public class NativeEngines990KnnVectorsReader extends KnnVectorsReader {
         // Start creating searcher
         final String fileName = KNNCodecUtil.getNativeEngineFileFromFieldInfo(fieldInfo, segmentReadState.segmentInfo);
         if (fileName != null) {
-            return () -> searcherFactory.createVectorSearcher(segmentReadState.directory, fileName, fieldInfo);
+            return () -> searcherFactory.createVectorSearcher(segmentReadState.directory, fileName, fieldInfo, ioContext);
         }
 
         // Not supported
--- a/src/main/java/org/opensearch/knn/index/query/common/QueryUtils.java
+++ b/src/main/java/org/opensearch/knn/index/query/common/QueryUtils.java
@@ -48,7 +48,7 @@ public class QueryUtils {
 
     /**
      * Returns a query that represents the specified TopDocs
-     * This is copied from {@link org.apache.lucene.search.AbstractKnnVectorQuery#createRewrittenQuery}
+     * This is copied from org.apache.lucene.search.AbstractKnnVectorQuery#createRewrittenQuery
      *
      * @param reader the index reader
      * @param topDocs the documents to be retured by the query
--- a/src/main/java/org/opensearch/knn/index/query/memoryoptsearch/MemoryOptimizedKNNWeight.java
+++ b/src/main/java/org/opensearch/knn/index/query/memoryoptsearch/MemoryOptimizedKNNWeight.java
@@ -9,6 +9,8 @@ import lombok.extern.log4j.Log4j2;
 import org.apache.lucene.index.FieldInfo;
 import org.apache.lucene.index.LeafReaderContext;
 import org.apache.lucene.index.SegmentReader;
+import org.apache.lucene.search.AcceptDocs;
+import org.apache.lucene.search.DocIdSetIterator;
 import org.apache.lucene.search.IndexSearcher;
 import org.apache.lucene.search.KnnCollector;
 import org.apache.lucene.search.TopDocs;
@@ -18,6 +20,8 @@ import org.apache.lucene.search.knn.KnnCollectorManager;
 import org.apache.lucene.search.knn.KnnSearchStrategy;
 import org.apache.lucene.search.knn.TopKnnCollectorManager;
 import org.apache.lucene.util.BitSet;
+import org.apache.lucene.util.BitSetIterator;
+import org.apache.lucene.util.Bits;
 import org.apache.lucene.util.Version;
 import org.opensearch.knn.index.SpaceType;
 import org.opensearch.knn.index.VectorDataType;
@@ -189,13 +193,13 @@ public class MemoryOptimizedKNNWeight extends KNNWeight {
 
         // Create a collector + bitset
         final KnnCollector knnCollector = knnCollectorManager.newCollector(visitedLimit, DEFAULT_HNSW_SEARCH_STRATEGY, context);
-        final BitSet bitSet = cardinality == 0 ? null : filterIdsBitSet;
+        final AcceptDocs acceptDocs = getAcceptedDocs(reader, cardinality, filterIdsBitSet);
 
         // Start searching index
         if (targetVector instanceof float[] floatTargetVector) {
-            reader.getVectorReader().search(knnQuery.getField(), floatTargetVector, knnCollector, bitSet);
+            reader.getVectorReader().search(knnQuery.getField(), floatTargetVector, knnCollector, acceptDocs);
         } else {
-            reader.getVectorReader().search(knnQuery.getField(), (byte[]) targetVector, knnCollector, bitSet);
+            reader.getVectorReader().search(knnQuery.getField(), (byte[]) targetVector, knnCollector, acceptDocs);
         }
 
         // Make results to return
@@ -207,4 +211,33 @@ public class MemoryOptimizedKNNWeight extends KNNWeight {
         addExplainIfRequired(topDocs, knnEngine, spaceType);
         return topDocs;
     }
+
+    private AcceptDocs getAcceptedDocs(SegmentReader reader, int cardinality, BitSet filterIdsBitSet) {
+        final AcceptDocs acceptDocs;
+        if (cardinality == 0) {
+            // We may want to use liveDocs here rather than null, to ensure that deleted docs are not considered in k-NN search
+            // But we are not doing this, because in that case it will break the current behavior of LOF which is similar to
+            // normal faiss based search.
+            acceptDocs = AcceptDocs.fromLiveDocs(null, reader.maxDoc());
+        } else {
+            acceptDocs = new AcceptDocs() {
+                @Override
+                public Bits bits() throws IOException {
+                    return filterIdsBitSet;
+                }
+
+                @Override
+                public DocIdSetIterator iterator() throws IOException {
+                    return new BitSetIterator(filterIdsBitSet, cardinality);
+                }
+
+                @Override
+                public int cost() throws IOException {
+                    return cardinality;
+                }
+            };
+        }
+        return acceptDocs;
+    }
+
 }
--- a/src/test/java/org/opensearch/knn/memoryoptsearch/AbstractFaissCagraHnswIndexTests.java
+++ b/src/test/java/org/opensearch/knn/memoryoptsearch/AbstractFaissCagraHnswIndexTests.java
@@ -9,10 +9,12 @@ import lombok.SneakyThrows;
 import org.apache.lucene.index.ByteVectorValues;
 import org.apache.lucene.index.FieldInfo;
 import org.apache.lucene.index.FloatVectorValues;
+import org.apache.lucene.search.AcceptDocs;
 import org.apache.lucene.search.KnnCollector;
 import org.apache.lucene.search.ScoreDoc;
 import org.apache.lucene.search.TopDocs;
 import org.apache.lucene.search.TopKnnCollector;
+import org.apache.lucene.search.knn.KnnSearchStrategy;
 import org.apache.lucene.store.IndexInput;
 import org.apache.lucene.util.IOConsumer;
 import org.opensearch.knn.KNNTestCase;
@@ -49,7 +51,7 @@ public abstract class AbstractFaissCagraHnswIndexTests extends KNNTestCase {
 
             // Make collector
             final int k = isApproximateSearch ? EF_SEARCH : TOTAL_NUMBER_OF_VECTORS;
-            final KnnCollector knnCollector = new TopKnnCollector(k, Integer.MAX_VALUE);
+            final KnnCollector knnCollector = new TopKnnCollector(k, Integer.MAX_VALUE, KnnSearchStrategy.Hnsw.DEFAULT);
 
             // Build a query
             final Object query;
@@ -72,12 +74,13 @@ public abstract class AbstractFaissCagraHnswIndexTests extends KNNTestCase {
                 }
                 query = binaryQuery;
             }
+            AcceptDocs acceptDocs = AcceptDocs.fromLiveDocs(null, TOTAL_NUMBER_OF_VECTORS);
 
             // Start searching
             if (vectorDataType == VectorDataType.FLOAT) {
-                searcher.search((float[]) query, knnCollector, null);
+                searcher.search((float[]) query, knnCollector, acceptDocs);
             } else {
-                searcher.search((byte[]) query, knnCollector, null);
+                searcher.search((byte[]) query, knnCollector, acceptDocs);
             }
             final TopDocs topDocs = knnCollector.topDocs();
             final ScoreDoc[] scoreDocs = topDocs.scoreDocs;
--- a/src/test/java/org/opensearch/knn/memoryoptsearch/FaissCagraHnswIndexTests.java
+++ b/src/test/java/org/opensearch/knn/memoryoptsearch/FaissCagraHnswIndexTests.java
@@ -8,11 +8,14 @@ package org.opensearch.knn.memoryoptsearch;
 import lombok.SneakyThrows;
 import org.apache.lucene.index.FloatVectorValues;
 import org.apache.lucene.index.VectorSimilarityFunction;
+import org.apache.lucene.search.AcceptDocs;
 import org.apache.lucene.search.KnnCollector;
 import org.apache.lucene.search.ScoreDoc;
 import org.apache.lucene.search.TopDocs;
 import org.apache.lucene.search.TopKnnCollector;
+import org.apache.lucene.search.knn.KnnSearchStrategy;
 import org.apache.lucene.store.IndexInput;
+import org.apache.lucene.util.Bits;
 import org.apache.lucene.util.IOConsumer;
 import org.opensearch.knn.KNNTestCase;
 import org.opensearch.knn.memoryoptsearch.faiss.FaissIndex;
@@ -47,7 +50,8 @@ public class FaissCagraHnswIndexTests extends KNNTestCase {
 
             // Make collector
             final int k = isApproximateSearch ? EF_SEARCH : TOTAL_NUMBER_OF_VECTORS;
-            final KnnCollector knnCollector = new TopKnnCollector(k, Integer.MAX_VALUE);
+            final KnnCollector knnCollector = new TopKnnCollector(k, Integer.MAX_VALUE, KnnSearchStrategy.Hnsw.DEFAULT);
+            AcceptDocs acceptDocs = AcceptDocs.fromLiveDocs(new Bits.MatchAllBits(TOTAL_NUMBER_OF_VECTORS), TOTAL_NUMBER_OF_VECTORS);
 
             // Build a query
             final float[] query = new float[DIMENSION];
@@ -56,7 +60,7 @@ public class FaissCagraHnswIndexTests extends KNNTestCase {
             }
 
             // Start searching
-            searcher.search(query, knnCollector, null);
+            searcher.search(query, knnCollector, acceptDocs);
             final TopDocs topDocs = knnCollector.topDocs();
             final ScoreDoc[] scoreDocs = topDocs.scoreDocs;
 
--- a/src/test/java/org/opensearch/knn/memoryoptsearch/FaissMemoryOptimizedSearcherTests.java
+++ b/src/test/java/org/opensearch/knn/memoryoptsearch/FaissMemoryOptimizedSearcherTests.java
@@ -12,6 +12,7 @@ import org.apache.lucene.index.FieldInfo;
 import org.apache.lucene.index.FieldInfos;
 import org.apache.lucene.index.SegmentInfo;
 import org.apache.lucene.index.SegmentReadState;
+import org.apache.lucene.search.AcceptDocs;
 import org.apache.lucene.search.DocIdSetIterator;
 import org.apache.lucene.search.KnnCollector;
 import org.apache.lucene.search.ScoreDoc;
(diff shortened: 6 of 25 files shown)


navneet1v:
Need to fix Lucene codec too since OpenSearch core is moved to 10.3 version of Lucene

finnegancarroll:
I see dependency conflict fixed on this branch but ITs still seem to fail to load failure access at runtime when i run locally.

```
./gradlew ':integTest' --tests 'org.opensearch.knn.recall.RecallTestsIT'
...

=== Standard error of node `node{::integTest-0}` ===
»   ↓ last 40 non error or warning messages from /Users/carrofin/fdev/repos/k-NN/build/testclusters/integTest-0/logs/opensearch.stderr.log ↓
» WARNING: Using incubator modules: jdk.incubator.vector
»  WARNING: Unknown module: org.apache.arrow.memory.core specified to --add-opens
»  fatal error in thread [opensearch[integTest-0][search][T#1]], exiting
»  java.lang.NoClassDefFoundError: com/google/common/util/concurrent/internal/InternalFutureFailureAccess
»       at java.base/java.lang.ClassLoader.defineClass1(Native Method)
»       at java.base/java.lang.ClassLoader.defineClass(ClassLoader.java:1027)
»       at java.base/java.security.SecureClassLoader.defineClass(SecureClassLoader.java:150)
»       at java.base/java.net.URLClassLoader.defineClass(URLClassLoader.java:524)
»       at java.base/java.net.URLClassLoader$1.run(URLClassLoader.java:427)
»       at java.base/java.net.URLClassLoader$1.run(URLClassLoader.java:421)
```

It's confusing to me this class is not loaded as I see it in the bundle:
https://github.com/opensearch-project/k-NN/pull/2885#issuecomment-3310030450

I think @karenyrx has an upstream change to resolve this:
https://github.com/opensearch-project/OpenSearch/pull/19339

My only thought would be it's kind of awkward that grpc transport SPI is supplying failure access when KNN uses it natively for `com.google.common.cache.LocalCache` in `org.opensearch.knn.index.memory.NativeMemoryCacheManager`.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
```diff
@@ -63,12 +68,13 @@ public class NativeEngines990KnnVectorsReader extends KnnVectorsReader {
...
     private final List<String> cacheKeys;
     private volatile Map<String, VectorSearcherHolder> vectorSearchers;
+    private final IOContext ioContext;
 
     public NativeEngines990KnnVectorsReader(final SegmentReadState state, final FlatVectorsReader flatVectorsReader) {
         this.flatVectorsReader = flatVectorsReader;
         this.segmentReadState = state;
         this.cacheKeys = getVectorCacheKeysFromSegmentReaderState(state);
-
+        ioContext = state.context.withHints(FileTypeHint.DATA, FileDataHint.KNN_VECTORS, DataAccessHint.RANDOM);
```
Vikasht34, line 77: Why this change?

src/main/java/org/opensearch/knn/index/query/TopApproxKnnCollector.java
```diff
@@ -20,7 +21,7 @@ public class TopApproxKnnCollector extends TopKnnCollector {
     private final SpaceType spaceType;
 
     public TopApproxKnnCollector(int k, KNNEngine engine, SpaceType spaceType) {
-        super(k, Integer.MAX_VALUE);
+        super(k, Integer.MAX_VALUE, KnnSearchStrategy.Hnsw.DEFAULT);
```
Vikasht34: Do we really need Integer.MAX_VALUE now? I would suggest to create the HNSW Strategy as generic interface in KNN to extend and pass rather than passing Default, this is not going to scale if we need custom implementation on Strategy

src/main/java/org/opensearch/knn/index/query/memoryoptsearch/MemoryOptimizedKNNWeight.java
```diff
@@ -189,13 +193,34 @@ private TopDocs queryIndex(
 
         // Create a collector + bitset
         final KnnCollector knnCollector = knnCollectorManager.newCollector(visitedLimit, DEFAULT_HNSW_SEARCH_STRATEGY, context);
-        final BitSet bitSet = cardinality == 0 ? null : filterIdsBitSet;
+        final AcceptDocs acceptDocs;
+        if (cardinality == 0) {
+            final Bits liveDocs = reader.getLiveDocs();
+            acceptDocs = AcceptDocs.fromLiveDocs(liveDocs, reader.maxDoc());
+        } else {
+            acceptDocs = new AcceptDocs() {
```
Vikasht34, line 223: This is too verbose to put in else statement, Please find a better way to write this.

navneet1v: Can you please elaborate more your comment. What is the meaning of this is too verbose to put in else comment?

src/main/java/org/opensearch/knn/index/query/TopApproxKnnCollector.java
navneet1v: > Do we really need Integer.MAX_VALUE now?

I am not sure what you mean when you say `now`. It is needed as part of the constructor args. So I kept the status go.
https://github.com/apache/lucene/blob/6f5451ca1aa1817a3b7a4d24576a23b71e9a1f95/lucene/core/src/java/org/apache/lucene/search/TopKnnCollector.java#L38-L46


> I would suggest to create the HNSW Strategy as generic interface in KNN to extend and pass rather than passing Default, this is not going to scale if we need custom implementation on Strategy

In terms of scaling I agree this won't scale. But at the same time in context of this PR, building an interface can lead to a premature interface, when we don't know what should be criteria for choosing a KNNSearchStrategy. I believe we should take it as part of a different GH issue discussion.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
navneet1v: We need IOContext object to open the faiss file during MemoryOptimized Search. [Ref](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissMemoryOptimizedSearcherFactory.java#L31-L34)

Now in the old code we were able to instantiate the IOContext with `new` keyword. It is not possible now. [Ref](https://github.com/apache/lucene/blob/branch_10_3/lucene/core/src/java/org/apache/lucene/store/IOContext.java#L31)

Now we need to use `withHints` interface to create a new instance of the IOContext, [ref](https://github.com/apache/lucene/blob/branch_10_3/lucene/core/src/java/org/apache/lucene/store/IOContext.java#L151). This is similar to what Lucene is also doing.[Ref](https://github.com/apache/lucene/blob/9135700a59f5d1eeff5e1fdf0393764ed3929c5f/lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99FlatVectorsReader.java#L72-L75)

So I am creating a similar to [previous one](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissMemoryOptimizedSearcherFactory.java#L31-L34) keeping it as class member and then pass to FaissMemoryOptimizedSearcherFactory when needed. 

Please let me know if you have more questions. I can see more how we can further improve this once build is fixed.

src/main/java/org/opensearch/knn/index/query/TopApproxKnnCollector.java
0ctopus13prime: I don't think we need to pass it, by default it will be null - [Code](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/TopKnnCollector.java#L35)
This one will only be used to transform Faiss raw score value to the contract, and the strategy will not be used.

src/main/java/org/opensearch/knn/index/query/memoryoptsearch/MemoryOptimizedKNNWeight.java
0ctopus13prime: I think we can factor this out as a small function in this class with a small comment. e.g. when `cardinality` is 0, it's indicating that there's no filtering which is combining user provided filter + liveDocs required..

src/test/java/org/opensearch/knn/memoryoptsearch/AbstractFaissCagraHnswIndexTests.java
```diff
@@ -49,7 +52,7 @@ protected void doTestKNNSearch(
 
             // Make collector
             final int k = isApproximateSearch ? EF_SEARCH : TOTAL_NUMBER_OF_VECTORS;
-            final KnnCollector knnCollector = new TopKnnCollector(k, Integer.MAX_VALUE);
+            final KnnCollector knnCollector = new TopKnnCollector(k, Integer.MAX_VALUE, KnnSearchStrategy.Hnsw.DEFAULT);
```
0ctopus13prime, line 54: I think we can just use `new TopKnnCollector(K, MAX_VALUE)` as it is, wondering is there particular reason to pass Hnsw.DEFAULT?

src/test/java/org/opensearch/knn/memoryoptsearch/AbstractFaissCagraHnswIndexTests.java
```diff
@@ -73,11 +76,14 @@ protected void doTestKNNSearch(
                 query = binaryQuery;
             }
 
+            Bits bits = new Bits.MatchAllBits(TOTAL_NUMBER_OF_VECTORS);
+            AcceptDocs acceptDocs = AcceptDocs.fromLiveDocs(bits, TOTAL_NUMBER_OF_VECTORS);
```
0ctopus13prime: NIT : `AcceptDocs acceptDocs = AcceptDocs.fromLiveDocs(null, TOTAL_NUMBER_OF_VECTORS);`
Just that the original test assumes there's no liveDocs or user provided filtering, I think passing null would match up better.

src/test/java/org/opensearch/knn/memoryoptsearch/FaissCagraHnswIndexTests.java
```diff
@@ -47,7 +50,9 @@ public void doTestKNNSearch(boolean isApproximateSearch) {
 
             // Make collector
             final int k = isApproximateSearch ? EF_SEARCH : TOTAL_NUMBER_OF_VECTORS;
-            final KnnCollector knnCollector = new TopKnnCollector(k, Integer.MAX_VALUE);
+            // TODO use KNNPlugin's collector here
```
0ctopus13prime: NIT : Hmm.. I don't think we can use `TopApproxKnnCollector` here as internal distance calculation is slightly different.

src/test/java/org/opensearch/knn/memoryoptsearch/FaissCagraHnswIndexTests.java
```diff
@@ -47,7 +50,9 @@ public void doTestKNNSearch(boolean isApproximateSearch) {
...
+            final KnnCollector knnCollector = new TopKnnCollector(k, Integer.MAX_VALUE, KnnSearchStrategy.Hnsw.DEFAULT);
```
0ctopus13prime, line 53: NIT : Can't we use `new TopKnnCollector(k, Integer.MAX_VALUE)`??

src/test/java/org/opensearch/knn/memoryoptsearch/FaissCagraHnswIndexTests.java
```diff
@@ -47,7 +50,9 @@ public void doTestKNNSearch(boolean isApproximateSearch) {
...
+            AcceptDocs acceptDocs = AcceptDocs.fromLiveDocs(new Bits.MatchAllBits(TOTAL_NUMBER_OF_VECTORS), TOTAL_NUMBER_OF_VECTORS);
```
0ctopus13prime, line 54: NIT : Can't we pass `null`? e.g. `AcceptDocs.fromLiveDocs(null, TOTAL_NUMBER_OF_VECTORS);`

src/test/java/org/opensearch/knn/memoryoptsearch/FaissMemoryOptimizedSearcherTests.java
```diff
@@ -500,15 +502,17 @@ private static KNNQueryResult[] doSearchViaVectorReader(
         // buildInfo.documentIds.size() + 1 -> Will force it to do exhaustive search.
         // buildInfo.documentIds.size() - 1 -> Will maximize search space, this is equivalent to set efSearch = len(N) - 1
         final int efSearch = exhaustiveSearch ? buildInfo.documentIds.size() + 1 : buildInfo.documentIds.size() - 1;
-        final KnnCollector knnCollector = new TopKnnCollector(efSearch, Integer.MAX_VALUE);
-        FixedBitSet acceptDocs = null;
+        final KnnCollector knnCollector = new TopKnnCollector(efSearch, Integer.MAX_VALUE, KnnSearchStrategy.Hnsw.DEFAULT);
```
0ctopus13prime: NIT : Can't we use `TopKnnCollector(efSearch, MAX_VALUE)` as it is?

0ctopus13prime (commented):
LGTM! Added few nit-pickings, thank you for the hard work!

src/test/java/org/opensearch/knn/memoryoptsearch/AbstractFaissCagraHnswIndexTests.java
navneet1v: if don't do this, then I was getting NPE in the test runs

src/test/java/org/opensearch/knn/memoryoptsearch/FaissMemoryOptimizedSearcherTests.java
navneet1v: This is resulting in NPE in tests.

src/main/java/org/opensearch/knn/index/query/memoryoptsearch/MemoryOptimizedKNNWeight.java
navneet1v: It was not clear from the comment. I can move this to a private function.

0ctopus13prime: NIT : If `cardinality == 0`, then can we just pass null?
```
acceptDocs = AcceptDocs.fromLiveDocs(null, reader.maxDoc());
```

Vikasht34: Here’s a working definition of verbose code style: using more words (or lines of code) than necessary, which reduces readability and maintainability.

Since verbose is a fairly common term in discussions about code style, I assumed you were already familiar with it — my mistake! If you’d like a deeper explanation, let me know.

For future reference, here are a couple of resources you may find useful:

Verbose Code Is Bad(https://www.divotion.com/blog/verbose-code-is-bad)
Verbose Mode: A Guide for the Curious, the Confused, and the Non-Techies(https://arc-sosangyo.medium.com/verbose-mode-a-guide-for-the-curious-the-confused-and-the-non-techies-259428dd30ca)
These should help you better understand the term in a coding context.

src/test/java/org/opensearch/knn/memoryoptsearch/AbstractFaissCagraHnswIndexTests.java
navneet1v: @0ctopus13prime I think original test was there to say that there is no filterDocs. 

Now with Accepted docs its like this, you always have to provide the docs on which you want to do the search. Ref: https://github.com/apache/lucene/blob/branch_10_3/lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java#L192-L208 So in case there are no filters we have to send all the docsIds. @0ctopus13prime let me know if you still think that is not the case.

src/test/java/org/opensearch/knn/memoryoptsearch/FaissCagraHnswIndexTests.java
navneet1v: I tried it, but the test failed with this exception.

```
FaissCagraHnswIndexTests > testKNNSearch FAILED
    java.lang.NullPointerException: Cannot invoke "org.apache.lucene.search.knn.KnnSearchStrategy.nextVectorsBlock()" because "this.originalStrategy" is null
        at __randomizedtesting.SeedInfo.seed([20196997EABB0E4A:867776256367AB8F]:0)
        at org.apache.lucene.search.knn.KnnSearchStrategy$Seeded.nextVectorsBlock(KnnSearchStrategy.java:169)
        at org.apache.lucene.util.hnsw.HnswGraphSearcher.searchLevel(HnswGraphSearcher.java:362)
        at org.apache.lucene.util.hnsw.SeededHnswGraphSearcher.searchLevel(SeededHnswGraphSearcher.java:71)
        at org.apache.lucene.util.hnsw.AbstractHnswGraphSearcher.search(AbstractHnswGraphSearcher.java:82)
        at org.apache.lucene.util.hnsw.HnswGraphSearcher.search(HnswGraphSearcher.java:162)
        at org.apache.lucene.util.hnsw.HnswGraphSearcher.search(HnswGraphSearcher.java:104)
        at org.opensearch.knn.memoryoptsearch.faiss.FaissMemoryOptimizedSearcher.search(FaissMemoryOptimizedSearcher.java:141)
        at org.opensearch.knn.memoryoptsearch.faiss.FaissMemoryOptimizedSearcher.search(FaissMemoryOptimizedSearcher.java:82)
        at org.opensearch.knn.memoryoptsearch.FaissCagraHnswIndexTests.lambda$doTestKNNSearch$0(FaissCagraHnswIndexTests.java:64)
        at org.opensearch.knn.memoryoptsearch.FaissCagraHnswIndexTests.doTestWithIndexInput(FaissCagraHnswIndexTests.java:114)
        at org.opensearch.knn.memoryoptsearch.FaissCagraHnswIndexTests.doTestKNNSearch(FaissCagraHnswIndexTests.java:46)
        at org.opensearch.knn.memoryoptsearch.FaissCagraHnswIndexTests.testKNNSearch(FaissCagraHnswIndexTests.java:39)



```

src/test/java/org/opensearch/knn/memoryoptsearch/FaissCagraHnswIndexTests.java
navneet1v: @0ctopus13prime I think this should ans the question: https://github.com/opensearch-project/k-NN/pull/2878/files/20f21f169c19ed7ceb1d0b464d5a6585e1bdc208#r2369595583

src/test/java/org/opensearch/knn/memoryoptsearch/AbstractFaissCagraHnswIndexTests.java
Vikasht34: @navneet1v , we have overloaded constructor with default strategy, we don't need to pass other parm , until unless we have any other strategy

https://github.com/apache/lucene/blob/a5a7e3417fb9180da3ebc14621a0d9ae6588c3bc/lucene/core/src/java/org/apache/lucene/search/TopKnnCollector.java#L29 

Let's fix the tests if fails rather that passing things which is not required.

src/main/java/org/opensearch/knn/index/query/TopApproxKnnCollector.java
Vikasht34: I hope this clears the confusion , https://github.com/apache/lucene/blob/a5a7e3417fb9180da3ebc14621a0d9ae6588c3bc/lucene/core/src/java/org/apache/lucene/search/TopKnnCollector.java#L29 there is overloaded constructor as well , which makes the method backward compatible.

src/test/java/org/opensearch/knn/memoryoptsearch/AbstractFaissCagraHnswIndexTests.java
0ctopus13prime: oh.. that's interesting... wondering why the previous test passes.

src/main/java/org/opensearch/knn/index/query/memoryoptsearch/MemoryOptimizedKNNWeight.java
navneet1v: > NIT : If `cardinality == 0`, then can we just pass null?
> 
> ```
> acceptDocs = AcceptDocs.fromLiveDocs(null, reader.maxDoc());
> ```

@0ctopus13prime I don't think we can do that. There has been some changes in how Lucene is now using AcceptedDocs. Earlier acceptedDocs was bitset, which when null means search on all docs. https://github.com/apache/lucene/blob/branch_10_3/lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java#L192-L208

src/test/java/org/opensearch/knn/memoryoptsearch/AbstractFaissCagraHnswIndexTests.java
navneet1v: https://github.com/opensearch-project/k-NN/pull/2878/files/20f21f169c19ed7ceb1d0b464d5a6585e1bdc208#r2369721220

src/test/java/org/opensearch/knn/memoryoptsearch/FaissCagraHnswIndexTests.java
navneet1v: Sure I can remove the TODO.

src/test/java/org/opensearch/knn/memoryoptsearch/FaissMemoryOptimizedSearcherTests.java
navneet1v: NVM, removed this. For this test its not needed

src/main/java/org/opensearch/knn/index/query/memoryoptsearch/MemoryOptimizedKNNWeight.java
0ctopus13prime: Yeah, agree for Lucene query, but we don't usee AbstractKnnVectorQuery in memory optimized search 😛 
Since we always pass non-null bitset (created in [here](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/query/KNNWeight.java#L356)), it will always fallback to [here](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/util/hnsw/HnswGraphSearcher.java#L295) I believe.

All I want to avoid is the bitset lookup operation. When we pass `null`, the lookup operation can be completely avoided, but with this change, since we're passing non-null bitset, we must do look up operation which probably affect performance. 

Of course, since I did not run debugging, so as long as we're passing null bitset when no filtering + no deleted docs, I'm fine with this!

src/main/java/org/opensearch/knn/index/query/common/QueryUtils.java
```diff
@@ -48,7 +48,7 @@ public static QueryUtils getInstance() {
 
     /**
      * Returns a query that represents the specified TopDocs
-     * This is copied from {@link org.apache.lucene.search.AbstractKnnVectorQuery#createRewrittenQuery}
+     * This is copied from org.apache.lucene.search.AbstractKnnVectorQuery#createRewrittenQuery
```
shatejas, line 51: nit: why remove?

src/main/java/org/opensearch/knn/index/query/memoryoptsearch/MemoryOptimizedKNNWeight.java
shatejas: @navneet1v just want to confirm we can't use [fromIteratorSupplier](https://github.com/apache/lucene/blob/a5a7e3417fb9180da3ebc14621a0d9ae6588c3bc/lucene/core/src/java/org/apache/lucene/search/AcceptDocs.java#L88) right?

Another thought would be, should we create acceptdocs when we create filterbit set and pass it in?

src/test/java/org/opensearch/knn/memoryoptsearch/AbstractFaissCagraHnswIndexTests.java
navneet1v: So the previous tests were passing because earlier Lucene was not getting the searchStratergy. https://github.com/apache/lucene/blob/branch_10_3/lucene/core/src/java/org/apache/lucene/util/hnsw/HnswGraphSearcher.java#L361-L363

Also in the src files for doing MemoryOptimizedSearch we are already setting the DefaultSearchStratrgy https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/query/memoryoptsearch/MemoryOptimizedKNNWeight.java#L191, which uses this: https://github.com/apache/lucene/blob/branch_10_3/lucene/core/src/java/org/apache/lucene/search/knn/TopKnnCollectorManager.java#L43-L47 to create the collector and SearchStrategy is set.

src/main/java/org/opensearch/knn/index/query/common/QueryUtils.java
navneet1v: it was failing the gradle build. I can revert it and see if it works this time.

src/main/java/org/opensearch/knn/index/query/memoryoptsearch/MemoryOptimizedKNNWeight.java
0ctopus13prime: Synced with Navneet, we can just pass liveDocs. If no filtering + no deleted docs, liveDocs will always be null.e

src/test/java/org/opensearch/knn/memoryoptsearch/AbstractFaissCagraHnswIndexTests.java
0ctopus13prime: Thank you for the details, this is helpful.

src/test/java/org/opensearch/knn/memoryoptsearch/FaissCagraHnswIndexTests.java
0ctopus13prime: oh sounds good. thank you for sharing.

src/main/java/org/opensearch/knn/index/query/common/QueryUtils.java
0ctopus13prime: `AbstractKnnVectorQuery` is a package scoped class, so we cannot import it for `@link` I believe

navneet1v:
> @navneet1v just want to confirm we can't use [fromIteratorSupplier](https://github.com/apache/lucene/blob/a5a7e3417fb9180da3ebc14621a0d9ae6588c3bc/lucene/core/src/java/org/apache/lucene/search/AcceptDocs.java#L88) right?

Another thought would be, should we create acceptdocs when we create filterbit set and pass it in?

@shatejas to use the [fromIteratorSupplier](https://github.com/apache/lucene/blob/a5a7e3417fb9180da3ebc14621a0d9ae6588c3bc/lucene/core/src/java/org/apache/lucene/search/AcceptDocs.java#L88), we have to pass the iterator right from the base KNNWeight.class . I thought about doing it first and then realized it will break the whole code and will lead to changes across many files, we are already on 25 files in this PR. I think with new interface we should move our code to completely using AcceptedDocs. Will take that change in upcoming PRs.

So, my strategy was to first make this change get the build working, and as a next steps see if we can use fromIteratorSupplier. Based on my deep-dive I can see AcceptedDocs was just a fancy wrapper over bits and DISI. For HNSW the usage of bits remain same, its just when Lucene starts to do exact search it uses iterator() to get the doc Ids. https://github.com/apache/lucene/blob/586dc103361c99fefa2f67946855691a073e805c/lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java#L234

src/main/java/org/opensearch/knn/index/query/memoryoptsearch/MemoryOptimizedKNNWeight.java
navneet1v: > @navneet1v just want to confirm we can't use [fromIteratorSupplier](https://github.com/apache/lucene/blob/a5a7e3417fb9180da3ebc14621a0d9ae6588c3bc/lucene/core/src/java/org/apache/lucene/search/AcceptDocs.java#L88) right?
> 
> Another thought would be, should we create acceptdocs when we create filterbit set and pass it in?


@shatejas to use the [fromIteratorSupplier](https://github.com/apache/lucene/blob/a5a7e3417fb9180da3ebc14621a0d9ae6588c3bc/lucene/core/src/java/org/apache/lucene/search/AcceptDocs.java#L88), we have to pass the iterator right from the base KNNWeight.class . I thought about doing it first and then realized it will break the whole code and will lead to changes across many files, we are already on 25 files in this PR. I think with new interface we should move our code to completely using AcceptedDocs. Will take that change in upcoming PRs.

So, my strategy was to first make this change get the build working, and as a next steps see if we can use fromIteratorSupplier. Based on my deep-dive I can see AcceptedDocs was just a fancy wrapper over bits and DISI. For HNSW the usage of bits remain same, its just when Lucene starts to do exact search it uses iterator() to get the doc Ids. https://github.com/apache/lucene/blob/586dc103361c99fefa2f67946855691a073e805c/lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java#L234

src/main/java/org/opensearch/knn/index/query/common/QueryUtils.java
navneet1v: I think it was always package scoped, not sure how it was working earlier. :P

0ctopus13prime: Earlier gradle does not check it, probably the checking rule has been added?
But I've been seeing warning signs in Intellij all the times with read highlight.

src/main/java/org/opensearch/knn/index/query/memoryoptsearch/MemoryOptimizedKNNWeight.java
shatejas: @navneet1v I think I am aligned, as long as we do it as a fast follow up

shatejas (approved):

src/main/java/org/opensearch/knn/index/query/memoryoptsearch/MemoryOptimizedKNNWeight.java
navneet1v: Sure, I will create a GH issue on this. and take it next.

PR #2879, opened 2025-09-17
FIX Potential Memory Leak

luyuncheng:
### Description
When we using [jemalloc](https://jemalloc.net/) to trace native memory allocation, and i found that `Convert2dJavaObjectArrayAndStoreToFloatVector` has a potential memory leak even there is nothing to index and after Full GC run. as following shows:
image

i think it is because that we do not delete local ref. 
like code: `auto vectorArray = (jfloatArray)env->GetObjectArrayElement(array2dJ, i); `
https://github.com/opensearch-project/k-NN/blob/ba979714f43fd3ace77b095ab7e9d3d6a132f718/jni/src/jni_util.cpp#L276-L294
After this patch, i traced again with jemalloc, it shows all memory cleaned.

### Check List


Vikasht34:
Hey @luyuncheng thanks for caching this !! , Looks good to me !!

navneet1v (dismissed):

jni/src/jni_util.cpp
```diff
@@ -291,6 +291,7 @@ void knn_jni::JNIUtil::Convert2dJavaObjectArrayAndStoreToFloatVector(JNIEnv *env
             vect->push_back(vector[j]);
         }
         env->ReleaseFloatArrayElements(vectorArray, vector, JNI_ABORT);
+        env->DeleteLocalRef(vectorArray);
```
0ctopus13prime, line 294: Could you use RAII to release resource? 
When an exception thrown at line 281, then we wont delete local ref, which is another leak I belive

JNIReleaseElements - https://github.com/opensearch-project/k-NN/blob/ba979714f43fd3ace77b095ab7e9d3d6a132f718/jni/include/jni_util.h#L230
```
// Acquire vector array
auto vectorArray = (jfloatArray)env->GetObjectArrayElement(array2dJ, i);

// Auto release when exiting scope
JNIReleaseElements localRefDeleter {[=]() {
    env->DeleteLocalRef(vectorArray);
}};
```

0ctopus13prime (commented):
Thank you for bring this up!
Could you use RAII to release resource to ensure that we always delete local ref, no matter what?

JNIReleaseElements - https://github.com/opensearch-project/k-NN/blob/ba979714f43fd3ace77b095ab7e9d3d6a132f718/jni/include/jni_util.h#L230

jni/src/jni_util.cpp
luyuncheng: i found there is many place need to refactor as RAII.  

@0ctopus13prime @navneet1v 
how about let us merge this PR
and let me create a new Refactor PR for all these places and rewrite it to RAII

navneet1v: I am good with that. @0ctopus13prime do you have any concerns?

0ctopus13prime: I think we can have RAII at least in the function in this PR, it just few lines 😛 
@luyuncheng Could you at least use RAII in the function in this PR? Remaining functions in KNN can be done in later as you said. 
If an exception thrown at line 281, it's clearly leak as we didn't call decrease local ref.

navneet1v: @luyuncheng can we do this and get this PR merged? It will be better if we can merge the memory leak fixes.

0ctopus13prime:
Raised a PR to use RAII to decrease local reference count.

https://github.com/opensearch-project/k-NN/pull/2916

Issue #2887, opened 2025-09-19
[BUG] Cosine similarity score range issue when memory optimized search is enabled.


0ctopus13prime:
**What is the bug?**
When cosine similarity is configured and memory optimized search is enabled, score value can be greater than 1 which suppose to be in range [0, 1). When it's turned off, score value now be bounded by 1.

Under the hood, Faiss only supports L2 or Inner product, therefore to support cosine similarity, we have a transform logic placed before Faiss.
And this was missed in memory optimized search, returning max inner product result back to client.
In the unit tests, I only checked the recall (technically, this won't affect recall), and missed score value range validation, that's why we did not find this earlier.

**How can one reproduce the bug?**
```
#!/bin/bash

# Create a mapping
curl -s -X PUT 'http://localhost:9200/my-knn-index-1' -H 'Content-Type: application/json' -d '
{
  "settings": {
    "index": {
      "knn": true,
      "use_compound_file": false,
      "knn.memory_optimized_search": true
    }
  },
  "mappings": {
    "properties": {
      "my_vector1": {
        "type": "knn_vector",
        "dimension": 2,
        "method": {
          "name": "hnsw",
          "space_type": "cosinesimil",
          "engine": "faiss"
        }
      }
    }
  }
}
'

sleep 2

curl -s -X POST 'http://localhost:9200/_bulk' -H 'Content-Type: application/json' -d '{ "index": { "_index": "my-knn-index-1", "_id": "1" } }
{ "my_vector1": [1.5, 2.5], "price": 12.2 }
{ "index": { "_index": "my-knn-index-1", "_id": "2" } }
{ "my_vector1": [2.5, 3.5], "price": 7.1 }
{ "index": { "_index": "my-knn-index-1", "_id": "3" } }
{ "my_vector1": [3.5, 4.5], "price": 12.9 }
{ "index": { "_index": "my-knn-index-1", "_id": "4" } }
{ "my_vector1": [5.5, 6.5], "price": 1.2 }
{ "index": { "_index": "my-knn-index-1", "_id": "5" } }
{ "my_vector1": [4.5, 5.5], "price": 3.7 }
'

sleep 2

curl -s -X GET 'http://localhost:9200/my-knn-index-1/_flush'

curl -X POST "http://localhost:9200/my-knn-index-1/_forcemerge?max_num_segments=1" -H "Content-Type: application/json"

[...]

PR #2892, opened 2025-09-23
Fix LuceneOnFaiss to have the same cosine score values with Faiss

0ctopus13prime:
### Description
This PR fixes score translation bug in LuceneOnFaiss when cosine space type is being used.
For more details, please refer to this [issue](https://github.com/opensearch-project/k-NN/issues/2887).

### Related Issues
https://github.com/opensearch-project/k-NN/issues/2887

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/query/MemoryOptimizedSearchScoreConverter.java
+++ b/src/main/java/org/opensearch/knn/index/query/MemoryOptimizedSearchScoreConverter.java
@@ -1,0 +1,102 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.knn.index.query;
+
+import org.apache.lucene.search.ScoreDoc;
+import org.opensearch.knn.index.SpaceType;
+import org.opensearch.knn.index.engine.KNNEngine;
+
+/**
+ * Utility class for converting between Faiss and Lucene score representations
+ * in memory-optimized search.
+ *
+ * <p>Memory-optimized search runs Lucene on top of a Faiss index. It leverages
+ * Lucene’s efficient algorithms and Lucene’s {@code Directory} architecture for efficient loading to
+ * produce the same results as when memory optimization is disabled.</p>
+ * With the same query, results are expected to be identical regardless of
+ * whether memory optimization is enabled.
+ *
+ * <p>However, unlike {@link KNNEngine},
+ * the input here is a Faiss score, which must be converted to Lucene’s
+ * scoring range.</p>
+ *
+ * <p>For example, Faiss uses inner product while Lucene uses
+ * maximum inner product. When converting distances, this class maps
+ * the Faiss score into the maximum inner product range so Lucene can
+ * interpret it correctly during search.</p>
+ *
+ * <p>Conversely, it also converts Lucene scores back into Faiss scores so that
+ * the same query produces consistent results across both implementations.
+ *
+ * <p>Note that this should be used only when memory_optimized_search is enabled.
+ *
+ */
+public final class MemoryOptimizedSearchScoreConverter {
+    /**
+     * Convert Faiss distance to Lucene score.
+     *
+     * @param distance Faiss distance
+     * @param spaceType Space type being used.
+     * @return Converted value to be used during Lucene search algorithm.
+     */
+    public static float distanceToRadialThreshold(final float distance, final SpaceType spaceType) {
+        if (spaceType != SpaceType.COSINESIMIL) {
+            return KNNEngine.LUCENE.distanceToRadialThreshold(distance, spaceType);
+        }
+
+        // For cosine similarity, `distance = 1 - inner_product_value`.
+        // therefore, we should extract it then convert it to max_inner_product_value
+        final float innerProductValue = KNNEngine.FAISS.distanceToRadialThreshold(distance, SpaceType.COSINESIMIL);
+
+        // Convert inner product value to max inner product value.
+        return SpaceType.INNER_PRODUCT.scoreTranslation(-innerProductValue);
+    }
+
+    /**
+     * Convert Faiss score to Lucene radial threshold.
+     *
+     * @param score Faiss score
+     * @param spaceType Space type that's being used
+     * @return Converted radial threshold for Lucene
+     */
+    public static float scoreToRadialThreshold(final float score, final SpaceType spaceType) {
+        if (spaceType != SpaceType.COSINESIMIL) {
+            return KNNEngine.LUCENE.scoreToRadialThreshold(score, spaceType);
+        }
+
+        // Since `score = (2 - (1 - inner_product_value)) / 2 = (1 + inner_product_value) / 2`,
+        // we should extract it then convert it to max inner product value.
+        final float innerProductValue = KNNEngine.FAISS.scoreToRadialThreshold(score, SpaceType.COSINESIMIL);
+
+        // Convert inner product value to max inner product value.
+        return SpaceType.INNER_PRODUCT.scoreTranslation(-innerProductValue);
+    }
+
+    /**
+     * This method converts Lucene's max inner product score to Faiss cosine score to ensure user
+     * to get the same results with the same query.
+     *
+     * @param scoreDocs Results from internal search before returning.
+     */
+    public static void convertToCosineScore(final ScoreDoc[] scoreDocs) {
+        for (final ScoreDoc scoreDoc : scoreDocs) {
+            // For cosine similarity, MAXIMUM_INNER_PRODUCT being used internally.
+            // Which maps negative values (which is plain inner product result value) to (0, 1], and maps positive values to (1, +inf).
+            // Below logic is to reverse back and extract the result of plain inner product.
+            final float innerProductValue;
+            if (scoreDoc.score >= 1) {
+                // Inner product value is positive.
+                innerProductValue = scoreDoc.score - 1;
+            } else {
+                // Inner product value is negative.
+                innerProductValue = 1 - 1 / scoreDoc.score;
+            }
+
+            // Then we need to transform the value to be bounded the desired range in cosine similarity space type.
+            scoreDoc.score = KNNEngine.FAISS.score(innerProductValue, SpaceType.COSINESIMIL);
+        }
+    }
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Bug Fixes
 * Use queryVector length if present in MDC check [#2867](https://github.com/opensearch-project/k-NN/pull/2867)
 * Fix derived source deserialization bug on invalid documents [#2882](https://github.com/opensearch-project/k-NN/pull/2882)
+* Fix invalid cosine score range in LuceneOnFaiss [#2892](https://github.com/opensearch-project/k-NN/pull/2892)
 
 ### Refactoring
 * Refactored the KNN Stat files for better readability.
--- a/src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
+++ b/src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
@@ -498,7 +498,7 @@ public class KNNQueryBuilder extends AbstractQueryBuilder<KNNQueryBuilder> imple
                 );
             }
             if (memoryOptimizedSearchEnabled) {
-                radius = KNNEngine.LUCENE.distanceToRadialThreshold(this.maxDistance, spaceType);
+                radius = MemoryOptimizedSearchScoreConverter.distanceToRadialThreshold(this.maxDistance, spaceType);
             } else {
                 radius = knnEngine.distanceToRadialThreshold(this.maxDistance, spaceType);
             }
@@ -511,7 +511,7 @@ public class KNNQueryBuilder extends AbstractQueryBuilder<KNNQueryBuilder> imple
                 );
             }
             if (memoryOptimizedSearchEnabled) {
-                radius = KNNEngine.LUCENE.scoreToRadialThreshold(this.minScore, spaceType);
+                radius = MemoryOptimizedSearchScoreConverter.scoreToRadialThreshold(this.minScore, spaceType);
             } else {
                 radius = knnEngine.scoreToRadialThreshold(this.minScore, spaceType);
             }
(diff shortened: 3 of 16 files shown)


src/main/java/org/opensearch/knn/index/query/MemoryOptimizedSearchScoreConverter.java
```diff
@@ -0,0 +1,102 @@
...
+ * the Faiss score into the maximum inner product range so Lucene can
+ * interpret it correctly during search.</p>
+ *
+ * <p>Conversely, it also converts Lucene scores back into Faiss scores so that
+ * the same query produces consistent results across both implementations.
+ *
+ * <p>Note that this should be used only when memory_optimized_search is enabled.
+ *
+ */
+public final class MemoryOptimizedSearchScoreConverter {
```
Vikasht34, line 37: This is good approach to fill the gap, So as per current logic , No Cosine Spaces falls back to KNNEngine.LUCENE methods while cosine gets special handling, but when  we have another space type like say Jaccard and if that needs special handling, we would be end up creating similar. Can we have Scorer translator inteface with per space inplementation , we can start with Cosine but it will to extend.

Right now there is already mapping logic leaves in SpaceType and KnnEngine Adapters. Can we not move this special case functions into spaceType having both MOS, NON MOS code at same path?

src/main/java/org/opensearch/knn/index/query/MemoryOptimizedSearchScoreConverter.java
```diff
@@ -0,0 +1,102 @@
...
+        return SpaceType.INNER_PRODUCT.scoreTranslation(-innerProductValue);
+    }
+
+    /**
+     * This method converts Lucene's max inner product score to Faiss cosine score to ensure user
+     * to get the same results with the same query.
+     *
+     * @param scoreDocs Results from internal search before returning.
+     */
+    public static void convertToCosineScore(final ScoreDoc[] scoreDocs) {
```
Vikasht34, line 84: As part of this , Can we add bound check (eg. [0,1]) like for cosine to catch errors in scores? Another point Since FAISS allows zero vectors [IP=0] but norm =0 causes division by zero in cosine, Can to safe guard , if IP extraction yeilds NAN , can we default to zero. Can we add test cases with Zero Vector.

if(Float.IsNan(scoreDoc.score) || scoreDoc.score <=0 then scoreDoc.score = 0.0f;

src/main/java/org/opensearch/knn/index/query/MemoryOptimizedSearchScoreConverter.java
```diff
@@ -0,0 +1,102 @@
...
+        for (final ScoreDoc scoreDoc : scoreDocs) {
+            // For cosine similarity, MAXIMUM_INNER_PRODUCT being used internally.
+            // Which maps negative values (which is plain inner product result value) to (0, 1], and maps positive values to (1, +inf).
+            // Below logic is to reverse back and extract the result of plain inner product.
+            final float innerProductValue;
+            if (scoreDoc.score > 1) {
```
Vikasht34: I think assumption here is Lucene MAX_IP emitted values strictly > for positive IP and (0,1] for negative IP, But at FP16/Quantized ADC boundry drift may happen, like a vlaue extermly close to 1 , if our boundry condition is strict > 1.0f and <= 1.0f with no tolerance then near boundry samples flip sides. 

FP Math 

fp32 ulp at 1.0 ~~1.1920929e-7 wheere as FP 16 ulp = 0.0009765625, 

if underliying engine accumilates in FP16 then effective uncernatiity around 1.0 is less. 

Can we add some clamping like if (s-1F) <1e-6f consider this as 1F ?

src/main/java/org/opensearch/knn/index/query/MemoryOptimizedSearchScoreConverter.java
```diff
@@ -0,0 +1,102 @@
...
+                // Inner product value is positive.
+                innerProductValue = scoreDoc.score - 1;
+            } else {
+                // Inner product value is negative.
+                innerProductValue = 1 - 1 / scoreDoc.score;
```
navneet1v, line 95: if scoreDoc.score = 0 then this code will lead divide by zero exception. can we have a check and a UT for this case.

src/main/java/org/opensearch/knn/index/query/MemoryOptimizedSearchScoreConverter.java
0ctopus13prime: This issue is a bit tricky and comes from a subtle difference in how Lucene and Faiss handle inner product scoring.

Lucene uses MaximumInnerProduct (MIP), which maps negative values to the range (0, 1) and non-negative values (including 0) to [1, +∞). Faiss uses a plain InnerProduct without that mapping.

Since Faiss only supports L2 and InnerProduct, we need special handling in LuceneOnFaiss to reconcile this difference. For the other metrics, we’ve already aligned the behavior, so scores from LuceneOnFaiss can be used interchangeably. **_Cosine is the only exception._**

Also, the SpaceType and KnnEngine adapters aren’t useful here, because they translate engine distances into engine-specific scores. What we need instead is to translate Lucene’s score into Faiss’s score, and only for the cosine case.

That’s why I don’t think this logic belongs in a general interface — it’s a LuceneOnFaiss-specific special case.

src/main/java/org/opensearch/knn/index/query/MemoryOptimizedSearchScoreConverter.java
0ctopus13prime: Hmm… thanks for the suggestion, but I think the score values are guaranteed to be non-NaN.
Also, the translation doesn’t involve division by zero: zero and all positive values are mapped to (value + 1), while only negative numbers go into the denominator.

src/main/java/org/opensearch/knn/index/query/MemoryOptimizedSearchScoreConverter.java
0ctopus13prime: Small correction is, negative IP will be mapped to (0, 1). If a value that is extremely close to 1, since the value is positive, it will be mapped to 1.99999 something similar. 
Also both Faiss and LuceneOnFaiss first converts FP16 to FP32 then do the distance calculation, I'm skeptical that drift may happen since all score calculation will be conducted in a FP32 variable.

src/main/java/org/opensearch/knn/index/query/MemoryOptimizedSearchScoreConverter.java
Vikasht34: Agree that It's LuceneOnFaiss use case specific , but the root issue is Lucene MIP vs Faiss plain IP and this pattern could re-occure. Lucene's VectorSimiliarityFunction enum already handles similar abstraction and per metrics translations, my worry is this approach would not scale much in future , if we have another score mismatching with lucene and faiss and when we support the Lucene on faiss , we end up creating similar. My Suggestion is see how we can generalize this to have less friction is future and make it state so that in future we don't end up creating this fragmentation in code.

src/main/java/org/opensearch/knn/index/query/MemoryOptimizedSearchScoreConverter.java
Vikasht34: This what I simply tried 

```
import numpy as np

def mip_encode(dot):
    if dot >= 0:
        return 1 + dot
    else:
        return 1.0 / (1 - dot)

def mip_decode(mip):  # Like PR's convertToCosineScore extraction
    if mip > 1:
        return mip - 1  # Positive ip
    else:
        return 1 - 1 / mip  # Negative ip (risky division)

# Case 1: Tiny negative dot (precision loss)
dot = np.float32(-1e-8)  # Tiny dissimilar
mip = mip_encode(dot)
print(f"Dot={dot}, MIP={mip}")  # MIP=1.0 (rounded, lost negative)
ip_decoded = mip_decode(mip)
print(f"Decoded IP={ip_decoded}, Cosine={(1 + ip_decoded)/2}")  # IP=0, Cosine=0.5 (neutral, but should be ~0.499999995)

# Case 2: Overflow to -inf (unnormalized huge negative)
dot = np.float32(-1e38)  # Beyond float32 max (~-3.4e38), becomes -inf
mip = mip_encode(dot)  # 1 / inf = 0.0
print(f"Dot={dot}, MIP={mip}")
ip_decoded = mip_decode(mip)  # 1 - 1/0.0 = -inf
print(f"Decoded IP={ip_decoded}, Cosine={(1 + ip_decoded)/2}")  # Cosine=-inf (bad)

# Case 3: NaN dot (invalid vector)
dot = np.float32(np.nan)
mip = mip_encode(dot)
print(f"Dot={dot}, MIP={mip}")
ip_decoded = mip_decode(mip)
print(f"Decoded IP={ip_decoded}, Cosine={(1 + ip_decoded)/2}")  # All NaN


Output 

1. Case 1: Dot=-1e-08, MIP=1.0
Decoded IP=0.0, Cosine=0.5  (Lost the tiny dissimilarity; should be <0.5)
2. Case 2: Dot=-inf, MIP=0.0
Decoded IP=-inf, Cosine=-inf  (Inf propagation from 1/mip)
3. Case 3: Dot=nan, MIP=nan
Decoded IP=nan, Cosine=nan  (Full propagation)
```

For huge negative dots  float32 can overflow dot to -inf, making (1 - dot)=inf, mip=1/inf=0.0. Decoding then hits 1 / mip = inf, yielding ip=-inf and cosine=NaN/inf

My sugestion is to keep the check.

src/main/java/org/opensearch/knn/index/query/MemoryOptimizedSearchScoreConverter.java
0ctopus13prime: I think it should be fine.
Regarding 

>  if we have another score mismatching with lucene and faiss and when we support the Lucene on faiss , we end up creating similar.

Except for Cosine, all metrics are identical between two, only Cosine is the issue.

src/main/java/org/opensearch/knn/index/query/MemoryOptimizedSearchScoreConverter.java
0ctopus13prime: Please note that Case 2 and Case 3 will not occur. Since we ensure all valid float numbers to be indexed.
And for the Case-1, treating -1e-08 as 0, I think this is reasonable, -1e-08 is almost 0.

0ctopus13prime: Also please note that, this PR is to ensure the same score values between Faiss vs LuceneOnFaiss!
When memory optimized search is disabled, it will encode the plain inner product values to max-inner-product and return. Therefore, we need to ensure the same value to be returned. And for the all cases, I got the same results!

```
public class DistanceTest {
    public static void main(String ... args) {
        printResults(-1e-08f);
        printResults(Float.NEGATIVE_INFINITY);
        printResults(Float.NaN);
    }

    public static float scaleMaxInnerProductScore(float vectorDotProductSimilarity) {
        return vectorDotProductSimilarity < 0.0F ? 1.0F / (1.0F + -1.0F * vectorDotProductSimilarity) : vectorDotProductSimilarity + 1.0F;
    }

    public static float cosine(float ip) {
        ip = 1 - ip;
        return Math.max((2.0F - ip) / 2.0F, 0.0F);
    }

    public static void printResults(final float ip) {
        // No MOS
        System.out.println("No MOS (e.g. Faiss C++): final score=" + cosine(ip));

        final float score = scaleMaxInnerProductScore(ip);
        final float innerProductValue;
        if (ip >= 1) {
            // Inner product value is positive.
            innerProductValue = score - 1;
        } else {
            // Inner product value is negative.
            innerProductValue = 1 - 1 / score;
        }

        System.out.println("Yes MOS (e.g. LuceneOnFaiss): final score=" + cosine(innerProductValue));
    }
}

Result:
No MOS (e.g. Faiss C++): final score=0.5
Yes MOS (e.g. LuceneOnFaiss): final score=0.5
No MOS (e.g. Faiss C++): final score=0.0
Yes MOS (e.g. LuceneOnFaiss): final score=0.0
No MOS (e.g. Faiss C++): final score=NaN
Yes MOS (e.g. LuceneOnFaiss): final score=NaN
```

0ctopus13prime: Since both are getting the same final score values with all the extreme values with the converting logic, I think it would be safe to keep it as it is?
Do you have thoughts on it?

src/main/java/org/opensearch/knn/index/query/MemoryOptimizedSearchScoreConverter.java
0ctopus13prime: `scoreDoc.score` is the result of MAX_IP conversion as shown below:

```
public static float scaleMaxInnerProductScore(float vectorDotProductSimilarity) {
    return vectorDotProductSimilarity < 0.0F ? 1.0F / (1.0F + -1.0F * vectorDotProductSimilarity) : vectorDotProductSimilarity + 1.0F;
}
```

If the score is 0, that means it had very large negative number (one example can be -Inf).
And for -Inf, the result of this cosine score encoding would be 0, which is exactly the one that we would have acquired with when memory optimized search disabled.

```
printResults(0);

    public static void printResults(final float ip) {
        // No MOS
        System.out.println("No MOS (e.g. Faiss C++): final score=" + cosine(ip));

        final float score = scaleMaxInnerProductScore(ip);
        final float innerProductValue;
        if (ip >= 1) {
            // Inner product value is positive.
            innerProductValue = score - 1;
        } else {
            // Inner product value is negative.
            innerProductValue = 1 - 1 / score;
        }

        System.out.println("Yes MOS (e.g. LuceneOnFaiss): final score=" + cosine(innerProductValue));
    }

// Result
No MOS (e.g. Faiss C++): final score=0.0
Yes MOS (e.g. LuceneOnFaiss): final score=0.0
```

The goal is to ensure both (LuceneOnFaiss, Faiss C++) yield the same score values.

src/main/java/org/opensearch/knn/index/query/MemoryOptimizedSearchScoreConverter.java
Vikasht34: Sure , Just want to Make Sure that , if anything changes on Lucene side on 10.4 , Please avoid making parallel changes in KNN , with this we will always creating parallel version of code in KNN and Lucene.

src/main/java/org/opensearch/knn/index/query/MemoryOptimizedSearchScoreConverter.java
Vikasht34: It should be fine , my concern was if we have differnce score from MOS vs Non Mos , as long as we have same it should be fine.

src/main/java/org/opensearch/knn/index/query/MemoryOptimizedSearchScoreConverter.java
Vikasht34: sure if you are 100% sure on above cases will be handle and no drift will ever happen , I am fine.

Vikasht34 (approved):
I am good with Changes

navneet1v (approved):

PR #2896, opened 2025-09-23
Changed query k functionality so that empty k field is null instead of 0

MrFlap:
### Description
KNNQuery k == 0 and k == null are used interchangeably, this change is to remove this logic (outside of Lucene library calls) by changing k from an int field to an Integer field.

### Related Issues
Resolves [#2836](https://github.com/opensearch-project/k-NN/issues/2836)
Intended to replace PR [#2871](https://github.com/opensearch-project/k-NN/pull/2871)

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
+++ b/src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
@@ -96,7 +96,7 @@ public class KNNQueryBuilder extends AbstractQueryBuilder<KNNQueryBuilder> imple
     private final String fieldName;
     private final float[] vector;
     @Getter
-    private int k;
+    private Integer k;
     @Getter
     private Float maxDistance;
     @Getter
@@ -218,7 +218,6 @@ public class KNNQueryBuilder extends AbstractQueryBuilder<KNNQueryBuilder> imple
 
         public KNNQueryBuilder build() {
             validate();
-            int k = this.k == null ? 0 : this.k;
             return new KNNQueryBuilder(
                 fieldName,
                 vector,
@@ -563,7 +562,7 @@ public class KNNQueryBuilder extends AbstractQueryBuilder<KNNQueryBuilder> imple
             throw new IllegalArgumentException(String.format(Locale.ROOT, "Engine [%s] does not support filters", knnEngine));
         }
 
-        if (k != 0) {
+        if (k != null && k != 0) {
             KNNQueryFactory.CreateQueryRequest createQueryRequest = KNNQueryFactory.CreateQueryRequest.builder()
                 .knnEngine(knnEngine)
                 .indexName(indexName)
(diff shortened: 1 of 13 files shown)


shatejas:
Changing int to Integer, increases the chances of NPE. Can we audit the unit test coverage with Jacoco and make sure it hits the code path

Also please make sure CIs pass

Vikasht34:
> HI @MrFlap Can we keep nullable `k` (e.g. Integer) in KnnQueryBuilder and keep it as 0 in KnnQuery side? We can pass `k == null ? 0 : k` to KnnQuery when building. Just that KnnQuery#k is being used in many places, not sure with this single PR we can ensure no NPE to be thrown in the future.

This is good Idea.

MrFlap:
> @MrFlap Hoping to understand, but what is the reason that should be changed at some point? Semantically, both `k == 0` and `k == null` are the same thing AFAIK, I thought things will work if we just keep `k == 0` in KnnQuery side. Just curious, and understand it more.

@0ctopus13prime 

The main cause of the bug in 2836 is that the KNNQueryBuilder.build() is called again on rewrite after a null k is set to 0. Since these have different behavior in the current code, this causes an error on a valid KNNQuery.

Ideally, we implement a change such that these kinds of errors are impossible to happen instead of a band-aid fix. One way to do this is to make a queryType abstract class that acts as a union between k, maxDistance, and minScore. We can make this into a separate PR for a later release, though, if this is too much for the 3.3 release date.

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
```diff
@@ -96,7 +96,7 @@ public class KNNQueryBuilder extends AbstractQueryBuilder<KNNQueryBuilder> imple
     private final String fieldName;
     private final float[] vector;
     @Getter
-    private int k;
+    private Integer k;
```
shatejas, line 99: I think this will throw an NPE https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/query/parser/KNNQueryBuilderParser.java#L237

shatejas: Please change this file https://github.com/opensearch-project/k-NN/blob/main/src/test/java/org/opensearch/knn/index/query/parser/KNNQueryBuilderParserTests.java to k = null wherever there is k = 0 in the tests

MrFlap: Acknowledged the first issue. Will investigate the second issue

MrFlap: Acknowledged and fixed

Vikasht34 (approved):
I am good with changes , Let's make sure all BWC passes and Integ test coverage for Radial Search.

MrFlap:
> @MrFlap some of the commits are not signed off. Can you squash all the commits in 1 commit and then signoff that commit please

Yes. Will do once I get confirmation that BWC is working.

MrFlap:
@navneet1v @Vikasht34 @0ctopus13prime Commits are squashed into one and signed off. This should pass k-NN build and test.

navneet1v (approved):

Vikasht34 (approved):

navneet1v:
> > @MrFlap does this change need backport to 2.19 version?
> 
> @navneet1v
> 
> I haven't learned what backporting is. Assuming this is asking if we need to make the same change to fix the bug in the older version. I would say to do the earlier bandaid fix so that BWC isn't broken (as this change modifies parsing). Let me know what you think.

I am adding a label on this. Please take care if the backport fails.

MrFlap:
@navneet1v @Vikasht34 

The plan for the 2.19 backporting is to reimplement PR [2871](https://github.com/opensearch-project/k-NN/pull/2871) in order to keep BWC intact. Otherwise there will be a tricky issue where KNNQueryBuilder is parsed in two different ways on two clusters with the same OS version.

PR #2903, opened 2025-09-26
Fixes integer overflow for efficient filtering

### LINKED ISSUE #2901: [BUG] Integer overflow for MDC calculations
**What is the bug?**

When there is a filter query present in KNN, the efficient filtering algorithm tries to estimate distance computation by multiplying query dimensions and filtered doc count. This multiplication causes integer overflow if filtered doc counts are high. 

This ultimately forces an exact KNN search behavior where ANN search is required causing higher latencies 

ANN should be executed when filterid count is large

 - OS:linux
 - Version 2.19
 - Plugins

shatejas:
Estimating distance calculation can run into int overflow if the filter docs cardinality is high. This happens because distance calculation involves multiplying dimension and filter docs cardinality

### Related Issues
Resolves [#[2901]](https://github.com/opensearch-project/k-NN/issues/2901)

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/query/KNNQuery.java
+++ b/src/main/java/org/opensearch/knn/index/query/KNNQuery.java
@@ -68,6 +68,10 @@ public class KNNQuery extends Query {
     // TODO: ThreadContext does not work with logger, remove this from here once its figured out
     private int shardId;
 
+    /**
+     * @deprecated Use builder instead
+     */
+    @Deprecated
     public KNNQuery(
         final String field,
         final float[] queryVector,
@@ -78,6 +82,10 @@ public class KNNQuery extends Query {
         this(field, queryVector, null, k, indexName, null, parentsFilter, VectorDataType.FLOAT, null);
     }
 
+    /**
+     * @deprecated Use builder instead
+     */
+    @Deprecated
     public KNNQuery(
         final String field,
         final float[] queryVector,
@@ -90,6 +98,10 @@ public class KNNQuery extends Query {
         this(field, queryVector, null, k, indexName, filterQuery, parentsFilter, VectorDataType.FLOAT, rescoreContext);
     }
 
+    /**
+     * @deprecated Use builder instead
+     */
+    @Deprecated
     public KNNQuery(
         final String field,
         final byte[] byteQueryVector,
@@ -103,6 +115,10 @@ public class KNNQuery extends Query {
         this(field, null, byteQueryVector, k, indexName, filterQuery, parentsFilter, vectorDataType, rescoreContext);
     }
 
+    /**
+     * @deprecated Use builder instead
+     */
+    @Deprecated
     private KNNQuery(
         final String field,
         final float[] queryVector,
@@ -126,13 +142,9 @@ public class KNNQuery extends Query {
     }
 
     /**
-     * Constructor for KNNQuery with query vector, index name and parent filter
-     *
-     * @param field field name
-     * @param queryVector query vector
-     * @param indexName index name
-     * @param parentsFilter parent filter
+     * @deprecated Use builder instead
      */
+    @Deprecated
     public KNNQuery(String field, float[] queryVector, String indexName, BitSetProducer parentsFilter) {
         this(field, queryVector, null, 0, indexName, null, parentsFilter, VectorDataType.FLOAT, null);
     }
@@ -168,6 +180,10 @@ public class KNNQuery extends Query {
     public KNNQuery filterQuery(Query filterQuery) {
         this.filterQuery = filterQuery;
         return this;
+    }
+
+    public int getQueryDimension() {
+        return this.queryVector != null ? this.queryVector.length : this.byteQueryVector.length;
     }
 
     /**
--- a/src/main/java/org/opensearch/knn/index/query/KNNQuery.java
+++ b/src/main/java/org/opensearch/knn/index/query/KNNQuery.java
@@ -183,7 +183,17 @@ public class KNNQuery extends Query {
     }
 
     public int getQueryDimension() {
-        return this.queryVector != null ? this.queryVector.length : this.byteQueryVector.length;
+        return switch (vectorDataType) {
+            case FLOAT, BYTE -> {
+                assert queryVector != null;
+                yield queryVector.length;
+            }
+            case BINARY -> {
+                assert byteQueryVector != null;
+                yield byteQueryVector.length * Byte.SIZE;
+            }
+            default -> queryVector.length;
+        };
     }
 
     /**
--- a/src/main/java/org/opensearch/knn/index/query/KNNQuery.java
+++ b/src/main/java/org/opensearch/knn/index/query/KNNQuery.java
@@ -184,15 +184,14 @@ public class KNNQuery extends Query {
 
     public int getQueryDimension() {
         return switch (vectorDataType) {
-            case FLOAT, BYTE -> {
-                assert queryVector != null;
-                yield queryVector.length;
-            }
             case BINARY -> {
                 assert byteQueryVector != null;
                 yield byteQueryVector.length * Byte.SIZE;
             }
-            default -> queryVector.length;
+            default -> {
+                assert queryVector != null;
+                yield queryVector.length;
+            }
         };
     }
 
--- a/src/main/java/org/opensearch/knn/index/query/KNNWeight.java
+++ b/src/main/java/org/opensearch/knn/index/query/KNNWeight.java
@@ -185,12 +185,12 @@ public abstract class KNNWeight extends Weight {
                     .append(filterThresholdValue)
                     .append(" is greater than or equal to cardinality = ")
                     .append(cardinality);
-            } else if (!isExactSearchThresholdSettingSet(filterThresholdValue) && isMDCGreaterThanFilterIdCnt(cardinality)) {
+            } else if (!isExactSearchThresholdSettingSet(filterThresholdValue) && isMaxDistCompGreaterThanEstimatedDistComp(cardinality)) {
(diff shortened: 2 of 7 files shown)


shatejas:
Will fix the CIs if they fail

src/main/java/org/opensearch/knn/index/query/KNNQuery.java
```diff
@@ -170,6 +182,10 @@ public KNNQuery filterQuery(Query filterQuery) {
         return this;
     }
 
+    public int getQueryDimension() {
+        return this.queryVector != null ? this.queryVector.length : this.byteQueryVector.length;
+    }
```
navneet1v, line 196: for binary vectors the dimension will be dimension*8.

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -182,14 +182,14 @@ private StringBuilder getLeafLevelExplanation(LeafReaderContext context) {
...
                     .append(cardinality);
-            } else if (!isExactSearchThresholdSettingSet(filterThresholdValue) && isMDCGreaterThanFilterIdCnt(cardinality)) {
+            } else if (!isExactSearchThresholdSettingSet(filterThresholdValue) && isMaxDistCompGreaterThanEstimatedDistComp(cardinality)) {
                 sb.append(KNNConstants.EXACT_SEARCH)
                     .append(" since max distance computation = ")
                     .append(KNNConstants.MAX_DISTANCE_COMPUTATIONS)
-                    .append(" is greater than or equal to cardinality = ")
-                    .append(cardinality);
+                    .append(" is greater than or equal to estimated distance computations = ")
+                    .append(((long) cardinality) * knnQuery.getQueryDimension());
```
Vikasht34, line 194: Will String will be trucated with explain API , Can we add test like filtercadinatility = 2^31 and Dimesion 10^3 to verify we don't truncate the string.

src/main/java/org/opensearch/knn/index/query/KNNQuery.java
```diff
@@ -68,6 +68,10 @@ public class KNNQuery extends Query {
     // TODO: ThreadContext does not work with logger, remove this from here once its figured out
     private int shardId;
 
+    /**
+     * @deprecated Use builder instead
+     */
+    @Deprecated
```
Vikasht34, line 74: Is this all part of Int Overflow issue?

src/main/java/org/opensearch/knn/index/query/KNNQuery.java
```diff
@@ -170,6 +182,10 @@ public KNNQuery filterQuery(Query filterQuery) {
         return this;
     }
 
+    public int getQueryDimension() {
```
Vikasht34, line 185: Since we are deprecating couple of Query Constructor, to safe guard method, Since we are doing query.length , Can we please add NPE check on Query?

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -593,13 +593,12 @@ private boolean isFilteredExactSearchPreferred(final int filterIdsCount) {
...
          */
-        return isMDCGreaterThanFilterIdCnt(filterIdsCount);
+        return isMaxDistCompGreaterThanEstimatedDistComp(filterIdsCount);
     }
 
-    private boolean isMDCGreaterThanFilterIdCnt(int filterIdsCount) {
-        return KNNConstants.MAX_DISTANCE_COMPUTATIONS >= filterIdsCount * (knnQuery.getQueryVector() != null
-            ? knnQuery.getQueryVector().length
-            : knnQuery.getByteQueryVector().length);
+    private boolean isMaxDistCompGreaterThanEstimatedDistComp(int filterDocsCardinality) {
```
Vikasht34, line 601: Should we add one test with Filtercadinality with Integer.max with Dim < 10 that can return correct boolean value. 

another test case would be filtercardinality* dim close to Long.Max_Value and check if estimates handle without warping negative.

Another test case what I was thinking , if we set Filtercardinality to max Value with annFilter null and if full estimation is used without overlflow.

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -593,13 +593,12 @@ private boolean isFilteredExactSearchPreferred(final int filterIdsCount) {
...
+        long estimatedDistanceComputations = ((long) filterDocsCardinality) * knnQuery.getQueryDimension();
```
Vikasht34, line 602: Will it be any issue if mutiple threads paraller running searchleaf with this estimator?

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
shatejas: > Should we add one test with Filtercadinality with Integer.max with Dim < 10 that can return correct boolean value.

There is already a unit test added for int overflow, are you asking for something else here? its a private function and I would want to keep it that way since its tested

> another test case would be filtercardinality* dim close to Long.Max_Value and check if estimates handle without warping negative

For OS to hit `Long.MAX_VALUE` it needs to have 92 trillion vectors with 10K dimensions (and more if the dimensions are less) in a segment which is not realistic, so this should be good for now

> Another test case what I was thinking , if we set Filtercardinality to max Value with annFilter null and if full estimation is used without overlflow.

If filter cardinality is `Integer.MAX_VALUE` that means the segment has more than 2 billion docs which does not seem to be realistic.

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
shatejas: yes its thread-safe. filterDocsCardinality is local to the thread

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
shatejas: filtercadinatility = 2^31 is unrealistic since this is more than 2 billion docs in a segment. Is there any evidence that explain strings are being truncated, can you clarify?

src/main/java/org/opensearch/knn/index/query/KNNQuery.java
shatejas: Nope, just a general cleanup

src/main/java/org/opensearch/knn/index/query/KNNQuery.java
shatejas: Not the right place for NPE check, I added an assert

src/main/java/org/opensearch/knn/index/query/KNNQuery.java
```diff
@@ -170,6 +182,20 @@ public KNNQuery filterQuery(Query filterQuery) {
...
+        return switch (vectorDataType) {
+            case FLOAT, BYTE -> {
+                assert queryVector != null;
+                yield queryVector.length;
+            }
+            case BINARY -> {
+                assert byteQueryVector != null;
+                yield byteQueryVector.length * Byte.SIZE;
+            }
+            default -> queryVector.length;
```
shatejas: I did not find any `vectorDataType != null` check so keeping a default to be safe

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
Vikasht34: Though String builder supports ~2B Character, 
Since we are appending (long) cardinality) * knnQuery.getQueryDimension() this , Can we verify in test with Max Values which is realistic with Max Dimension Supported  to verify string are not concatenated. ?
I guess in Knn weight test , we already testing with integer.MaxValue , Can we have add similar test here as well to verify the string being returned ?

src/main/java/org/opensearch/knn/index/query/KNNQuery.java
Vikasht34: assert queryVector != null is NPE Check , it bubbles up exception upwards.  I am good with assert too.

src/main/java/org/opensearch/knn/index/query/KNNQuery.java
Vikasht34: Thanks , I guess in Default also should be put Assert?

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
Vikasht34: I see test contains "when(reader.maxDoc()).thenReturn(Integer.MAX_VALUE) ~2.1 billion , should be fine for above mentioned cases.

Vikasht34 (dismissed):
Overall looks good , some nit picks with test cases.

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
shatejas: There is an [assert](https://github.com/opensearch-project/k-NN/blob/main/src/test/java/org/opensearch/knn/index/query/ExplainTests.java#L565) on explain. I agree its a `contains` check and not `equals` and it should be tightened but its a bit orthogonal and should be tackled as a improvement on tests

If the concern is `(long) cardinality) * knnQuery.getQueryDimension()` will increase the characters in String; a quintillion would only add about 19 characters, which won't breach the String limits

src/main/java/org/opensearch/knn/index/query/KNNQuery.java
shatejas: added

Vikasht34 (approved):
Looks Good

navneet1v (approved):

Issue #2875, opened 2025-09-16
[RFC] Use SIMD for FP16 in LuceneOnFaiss


0ctopus13prime:
# 1. Background

When LuceneOnFaiss is enabled, OpenSearch relies on Lucene’s vector search algorithms running over a Faiss index. Because Java does not natively support FP16, OpenSearch first copies the raw bytes into a byte[], then converts the FP16 vectors into FP32 before applying the similarity function. This additional copy and conversion introduces latency, making FP16 search with LuceneOnFaiss slower compared to Faiss C++, which leverages optimized SIMD intrinsics. 
In a force merge scenario with a single segment, the performance gap is significant, showing LuceneOnFaiss to be roughly 150%–200% slower than Faiss C++ even with early termination is enabled. Therefore, it is important for us to find an optimal approach to ensure that LuceneOnFaiss can achieve performance comparable to Faiss C++.

# 2. Goal

When MMapDirectory is used, acquire its internal pointer to pass directly to the C++ layer to avoid copying vectors. The C++ layer then leverages Faiss’s SIMD-optimized distance calculation logic to compute distances more efficiently, resulting in better performance.

# 3. Testing Cases

## 3.1. Implementation Notes

I implemented two versions:

1. A version that uses Faiss’s built-in SIMD distance calculator.
2. Since FP16 lacks the kind of bulk optimization available for FP32, I added a custom bulk SIMD inner-product distance routine with prefetching.

Both versions first obtain a raw pointer to the mapped memory region, then pass it to C++ to offload distance calculations. (Note: this approach is for POC purposes only. Technically, a file can have multiple mapped regions. By default, one arena size is configured as 16 GB in Lucene, so a file larger than 16 GB may span multiple arenas.) I also attempted to read vectors into a `byte[]` and pass them to JNI, as shown below, but performance was extremely slow compared to Faiss C++. This led me to decide not to explore that approach further.

```
// Use Faiss SIMD to calculate distance on the given byte[] from Java
// This shows twice slower performance than current LuceneOnFaiss implementation.
JNIEXPORT void Java_org_opensearch_knn_jni_FaissService_bulkScoring1
  (JNIEnv * env, jclass cls, jlong queryAddr, jobjectArray vectors, jint numVectors, jlong scoresAddr, jint scoreIdx, jint dimension) {
[...]

PR #2904, opened 2025-09-26
[BulkSimd PR1] Added MMapByteVectorValues for FP16 in LuceneOnFaiss.

0ctopus13prime:
### Description
This is the first PR to support native scoring in LuceneOnFaiss. 

RFC : https://github.com/opensearch-project/k-NN/issues/2875

In this first PR, FP16 flat index will return `MMapByteVectorValues` having `MemorySegment[]` extracted from MemorySegment$IndexInput. Each `MemorySegment` will have a raw pointer pointing to mapped region in memory.
Passing the raw pointer for distance calculation will be dealt with in the second PR. This PR is mostly set-up for the second PR.

Since the actual distance calculation is not there, still the distance calculation will be conducted in Java side.
This PR is mostly focusing on extracting `MemorySegment` from IndexInput.

### Related Issues
N/A

### Check List
- [O] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).

### DIFF
--- a/src/main/java/org/opensearch/knn/memoryoptsearch/MemorySegmentAddressExtractorJDK21.java
+++ b/src/main/java/org/opensearch/knn/memoryoptsearch/MemorySegmentAddressExtractorJDK21.java
@@ -1,0 +1,97 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.knn.memoryoptsearch;
+
+import lombok.extern.log4j.Log4j2;
+import org.apache.lucene.store.IndexInput;
+
+import java.lang.reflect.Array;
+import java.lang.reflect.Field;
+import java.lang.reflect.InaccessibleObjectException;
+import java.lang.reflect.InvocationTargetException;
+
+/**
+ * Since {@code MemorySegment} is not made officially in JDK21, we first extract {@code MemorySegment[]} from {@link IndexInput},
+ * and have it as Object then use reflection to collect address and size info.
+ */
+@Log4j2
+public final class MemorySegmentAddressExtractorJDK21 implements MemorySegmentAddressExtractor {
+    private static final java.lang.reflect.Method ADDRESS_METHOD;
+    private static final java.lang.reflect.Method BYTE_SIZE_METHOD;
+
+    // Get the handle when loading once.
+    static {
+        java.lang.reflect.Method addressMethod = null;
+        java.lang.reflect.Method byteSizeMethod = null;
+        try {
+            Class<?> clazz = Class.forName("java.lang.foreign.MemorySegment");
+            addressMethod = clazz.getMethod("address");
+            byteSizeMethod = clazz.getMethod("byteSize");
+        } catch (ClassNotFoundException | NoSuchMethodException e) {
+            // Running on JDK where MemorySegment/address() doesn't exist
+        }
+        ADDRESS_METHOD = addressMethod;
+        BYTE_SIZE_METHOD = byteSizeMethod;
+    }
+
+    /**
+     * Extracts address and size info of {@code MemorySegment[]} from the given {@code indexInput}.
+     * <p>
+     * When using {@code MMapDirectory}, the {@code indexInput} may be an instance of
+     * {@code MemorySegmentIndexInput$SingleSegmentImpl} or {@code MemorySegmentIndexInput$MultiSegmentImpl}.
+     * These classes wrap mapped pointers in {@code MemorySegment} objects stored in a field named {@code segments}.
+     * This method detects the {@code segments} field, extracts its value, and returns it.
+     * <p>
+     * If the corresponding `segments` cannot be found, this method simply returns {@code null}.
+     * In that case, the search logic falls back to the default scorer, which loads vectors
+     * into the JVM heap and performs distance calculations there.
+     * <p>
+     *
+     * @param indexInput the input from which to extract memory segments
+     * @return an array of address and size info extracted from the input, or null if it's not found.
+     *         ex: address_i = array[i], size_i = array[i + 1]
+     */
+    @Override
+    public long[] extractAddressAndSize(final IndexInput indexInput) {
+        if (ADDRESS_METHOD == null || BYTE_SIZE_METHOD == null) {
+            // Runtime JDK does not support MemorySegment.
+            return null;
+        }
+
+        try {
+            // MMapDirectory in Lucene will return MemorySegmentIndexInput$SingleSegmentImpl or .$MultiSegmentImpl.
+            // Thus, get the super class (e.g. MemorySegmentIndexInput) to acquire `MemorySegment[] segments`.
+            final Field f = indexInput.getClass().getSuperclass().getDeclaredField("segments");
+            f.setAccessible(true);
+            // We're expecting this to be MemorySegment[]
+            final Object objSegments = f.get(indexInput);
+            if (objSegments == null || !objSegments.getClass().isArray()) {
+                // It's not MemorySegment[]
+                return null;
+            }
+            final int numSegments = Array.getLength(objSegments);
+            final long[] addressAndSize = new long[2 * numSegments];
+            for (int segmentIndex = 0; segmentIndex < numSegments; segmentIndex++) {
+                final Object memorySegment = Array.get(objSegments, segmentIndex);
+                if (memorySegment == null) {
+                    // Memory segments does not have complete mapped regions.
+                    return null;
+                }
+                final int addressIndex = 2 * segmentIndex;
+                final int sizeIndex = 2 * segmentIndex + 1;
+                addressAndSize[addressIndex] = (long) ADDRESS_METHOD.invoke(memorySegment);
+                addressAndSize[sizeIndex] = (long) BYTE_SIZE_METHOD.invoke(memorySegment);
+            }
+
+            return addressAndSize;
+        } catch (NoSuchFieldException | IllegalAccessException | IllegalArgumentException | InaccessibleObjectException
+            | InvocationTargetException e) {
+            // Ignore
+            log.warn("Failed to extract MemorySegment[] from IndexInput, error message={}", e.getMessage());
+        }
+        return null;
+    }
+}
--- a/src/main/java/org/opensearch/knn/memoryoptsearch/MemorySegmentAddressExtractorJDK22.java
+++ b/src/main/java/org/opensearch/knn/memoryoptsearch/MemorySegmentAddressExtractorJDK22.java
@@ -1,0 +1,69 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.knn.memoryoptsearch;
+
+import lombok.extern.log4j.Log4j2;
+import org.apache.lucene.store.IndexInput;
+
+import java.lang.foreign.MemorySegment;
+import java.lang.reflect.Field;
+import java.lang.reflect.InaccessibleObjectException;
+
+/**
+ * For JDK 22 and later, the Foreign Function  &amp; Memory (FFM) API is part of the standard, so we no longer need to rely on reflection.
+ * This class relies on reflection to extract {@code MemorySegment[]} from {@code IndexInput}.
+ * And after acquired {@code MemorySegment[]}, it will use it right away to collect address and mapped size.
+ */
+@Log4j2
+public final class MemorySegmentAddressExtractorJDK22 implements MemorySegmentAddressExtractor {
+    /**
+     * Extracts address and size info of {@code MemorySegment[]} from the given {@code indexInput}.
+     * <p>
+     * When using {@code MMapDirectory}, the {@code indexInput} may be an instance of
+     * {@code MemorySegmentIndexInput$SingleSegmentImpl} or {@code MemorySegmentIndexInput$MultiSegmentImpl}.
+     * These classes wrap mapped pointers in {@code MemorySegment} objects stored in a field named {@code segments}.
+     * This method detects the {@code segments} field, extracts its value, and returns it.
+     * <p>
+     * If the corresponding `segments` cannot be found, this method simply returns {@code null}.
+     * In that case, the search logic falls back to the default scorer, which loads vectors
+     * into the JVM heap and performs distance calculations there.
+     * <p>
+     *
+     * @param indexInput the input from which to extract memory segments
+     * @return an array of address and size info extracted from the input, or null if it's not found.
+     *         ex: address_i = array[i], size_i = array[i + 1]
+     */
+    @Override
+    public long[] extractAddressAndSize(final IndexInput indexInput) {
+        try {
+            // MMapDirectory in Lucene will return MemorySegmentIndexInput$SingleSegmentImpl or .$MultiSegmentImpl.
+            // Thus, get the super class (e.g. MemorySegmentIndexInput) to acquire `MemorySegment[] segments`.
+            final Field f = indexInput.getClass().getSuperclass().getDeclaredField("segments");
+            f.setAccessible(true);
+            final MemorySegment[] segments = (MemorySegment[]) f.get(indexInput);
+            if (segments == null || segments.length == 0) {
+                return null;
+            }
+            final long[] addressAndSize = new long[2 * segments.length];
+            for (int segmentIndex = 0; segmentIndex < segments.length; ++segmentIndex) {
+                final int addressIndex = 2 * segmentIndex;
+                final int sizeIndex = 2 * segmentIndex + 1;
+                final MemorySegment segment = segments[segmentIndex];
+                if (segment == null) {
+                    // Memory segments does not have complete mapped regions.
+                    return null;
+                }
+                addressAndSize[addressIndex] = segments[segmentIndex].address();
+                addressAndSize[sizeIndex] = segments[segmentIndex].byteSize();
+            }
+            return addressAndSize;
+        } catch (NoSuchFieldException | IllegalAccessException | InaccessibleObjectException e) {
+            // Ignore
+            log.warn("Failed to extract MemorySegment[] from IndexInput, error message={}", e.getMessage());
+        }
+        return null;
+    }
+}
--- a/src/main/java/org/opensearch/knn/memoryoptsearch/MemorySegmentAddressExtractorUtil.java
+++ b/src/main/java/org/opensearch/knn/memoryoptsearch/MemorySegmentAddressExtractorUtil.java
@@ -1,0 +1,78 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.knn.memoryoptsearch;
+
+import lombok.extern.log4j.Log4j2;
+import org.apache.lucene.store.IndexInput;
+
+/**
+ * This class instantiates a {@code MemorySegment} address and size extractor when it is loaded.
+ * It first attempts to load MemorySegmentAddressExtractorJDK22+; if that fails,
+ * it falls back to {@link MemorySegmentAddressExtractorJDK21}.
+ *
+ * <p>
+ * Note that MemorySegmentAddressExtractorJDK22+ is only included when the project
+ * is compiled with JDK&nbsp;22+ compatibility.
+ * </p>
+ *
+ * <p>
+ * Both extractors use reflection to obtain the private {@code MemorySegment[]} field from
+ * {@link IndexInput}. In particular, {@link org.apache.lucene.store.MMapDirectory} returns
+ * an internal {@link IndexInput} implementation that stores {@code MemorySegment[]} in
+ * a private field named {@code segment}. Reflection is used to access this field and
+ * extract address and size information.
+ * </p>
+ *
+ * <p>
+ * For JDK&nbsp;21, {@link MemorySegmentAddressExtractorJDK21} also uses reflection to indirectly
+ * invoke the {@code address()} and {@code byteSize()} methods of {@code MemorySegment}.
+ * Since {@code MemorySegment} only became a standard API in JDK&nbsp;22+, reflection is the
+ * only option unless the {@code --enable-preview} flag is explicitly provided.
+ * </p>
+ *
+ * <pre>
+ * +-----------------------------+        (reflection)        +--------------------------------------+
+ * |          IndexInput         |  ----------------------->  |  MemorySegmentAddressExtractorJDK22  |
+ * |  segments: MemorySegment[]  |                            |  Address[], Size[]                   |
+ * +-----------------------------+                            +--------------------------------------+
+ *
+ * +-----------------------------+        (reflection)        +--------------------------------------+
+ * |          IndexInput         |  ----------------------->  |  MemorySegmentAddressExtractorJDK21  |
+ * |  segments: MemorySegment[]  |                     ^      |  Address[], Size[]                   |
+ * +-----------------------------+                     |      +--------------------------------------+
+ *                                                     |
+ *             (reflection on MemorySegment methods)   +--> MemorySegment#address()
+ *                                                         MemorySegment#byteSize()
+ * </pre>
+ */
+@Log4j2
+public final class MemorySegmentAddressExtractorUtil {
+    private static final MemorySegmentAddressExtractor INSTANCE;
+
+    static {
+        MemorySegmentAddressExtractor instance;
+        try {
+            try {
+                // Try to load the JDK22-optimized class which will be packaged only for compile_target = 22+.
+                Class<?> clazz = Class.forName("org.opensearch.knn.memoryoptsearch.MemorySegmentAddressExtractorJDK22");
+                instance = (MemorySegmentAddressExtractor) clazz.getDeclaredConstructor().newInstance();
+            } catch (ClassNotFoundException e) {
+                log.warn("Failed to load MemorySegmentAddressExtractorJDK22, falling back to MemorySegmentAddressExtractorJDK21", e);
+                // Class not found: fall back to JDK21 version
+                instance = new MemorySegmentAddressExtractorJDK21();
+            }
+        } catch (Throwable t) {
+            // Any other errors (constructor, reflection issues)
+            log.error("Failed to instantiate MemorySegmentAddressExtractor", t);
+            instance = (indexInput) -> null;
+        }
+        INSTANCE = instance;
+    }
+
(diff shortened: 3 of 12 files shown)


src/main/java/org/opensearch/knn/memoryoptsearch/MemorySegmentAddressExtractorJDK22.java
```diff
@@ -0,0 +1,57 @@
...
+                return null;
+            }
+            final long[] addressAndSize = new long[2 * segments.length];
+            for (int i = 0; i < segments.length; i += 2) {
+                addressAndSize[i] = segments[i].address();
+                addressAndSize[i + 1] = segments[i].byteSize();
+            }
+            return addressAndSize;
+        } catch (NoSuchFieldException | IllegalAccessException ea) {
+            // Ignore
```
shatejas, line 64: nit: log a warning

src/main/java/org/opensearch/knn/memoryoptsearch/MemorySegmentAddressExtractorJDK21.java
```diff
@@ -0,0 +1,87 @@
...
+        }
+
+        try {
+            // MMapDirectory in Lucene will return MemorySegmentIndexInput$SingleSegmentImpl or .$MultiSegmentImpl.
+            // Thus, get the super class (e.g. MemorySegmentIndexInput) to acquire `MemorySegment[] segments`.
+            final Field f = indexInput.getClass().getSuperclass().getDeclaredField("segments");
+            f.setAccessible(true);
+            // We're expecting this to be MemorySegment[]
+            final Object objSegments = f.get(indexInput);
+            if (objSegments == null || !objSegments.getClass().isArray()) {
```
shatejas, line 71: Just curious, `objSegments instanceof MemorySegment[]` does not work?

src/main/java/org/opensearch/knn/memoryoptsearch/MemorySegmentAddressExtractorJDK21.java
```diff
@@ -0,0 +1,87 @@
...
+     * If the corresponding `segments` cannot be found, this method simply returns {@code null}.
+     * In that case, the search logic falls back to the default scorer, which loads vectors
+     * into the JVM heap and performs distance calculations there.
+     * <p>
+     *
+     * @param indexInput the input from which to extract memory segments
+     * @return an array of address and size info extracted from the input, or null if it's not found.
+     *         ex: address_i = array[i], size_i = array[i + 1]
+     */
+    public long[] extractAddressAndSize(final IndexInput indexInput) {
```
shatejas, line 58: nit: `@Override` annotation

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissMemoryOptimizedSearcher.java
```diff
@@ -67,6 +69,16 @@ public FaissMemoryOptimizedSearcher(final IndexInput indexInput, final FieldInfo
         this.flatVectorsScorer = FlatVectorsScorerProvider.getFlatVectorsScorer(knnVectorSimilarityFunction, isAdc, spaceType);
 
         this.hnsw = extractFaissHnsw(faissIndex);
+
+        // Use native scoring logic to calculate distance outside JVM.
+        this.useNativeScorer = false;
+        try {
+            if (faissIndex.getByteValues(indexInput) instanceof MMapByteVectorValues) {
```
shatejas: Its not clear why this is needed, PR2 might give a complete picture though. Overall if we can avoid  `faissIndex.getByteValues(indexInput)` here we should

src/main/java/org/opensearch/knn/memoryoptsearch/MemorySegmentAddressExtractorUtil.java
```diff
@@ -0,0 +1,32 @@
...
+            // Try to load the JDK22-optimized class which will be packaged only for compile_target = 22+.
+            Class<?> clazz = Class.forName("org.opensearch.knn.memoryoptsearch.MemorySegmentAddressExtractorJDK22");
+            instance = (MemorySegmentAddressExtractor) clazz.getDeclaredConstructor().newInstance();
+        } catch (ClassNotFoundException e) {
+            // Class not found: fall back to JDK21 version
+            instance = new MemorySegmentAddressExtractorJDK21();
+        } catch (Throwable t) {
+            // Any other errors (constructor, reflection issues)
+            throw new RuntimeException("Failed to initialize MemorySegmentAddressExtractor", t);
+        }
```
shatejas, line 71: this should keep the instance null and then fallback to java dist computations? Throwing here might kill the node at startup I think, can you verify

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissIndexScalarQuantizedFlat.java
```diff
@@ -158,6 +159,23 @@ public ByteVectorValues copy() {
...
+                // Return MMapByteVectorValues having pointers pointing to mmap regions.
+                return new MMapByteVectorValues(
+                    indexInput,
+                    oneVectorByteSize,
+                    flatVectors.getBaseOffset(),
+                    dimension,
+                    totalNumberOfVectors,
+                    addressAndSize
+                );
+            }
```
shatejas, line 179: nit: consider maybe?
```
else {
  log.debug()
}
```

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/MMapByteVectorValues.java
```diff
@@ -0,0 +1,73 @@
...
+
+package org.opensearch.knn.memoryoptsearch.faiss;
+
+import lombok.Getter;
+import org.apache.lucene.index.ByteVectorValues;
+import org.apache.lucene.store.IndexInput;
+
+import java.io.IOException;
+
+public class MMapByteVectorValues extends ByteVectorValues {
```
navneet1v, line 23: please add java docs on all the public functions and classes

src/main/java/org/opensearch/knn/memoryoptsearch/MemorySegmentAddressExtractorJDK21.java
```diff
@@ -0,0 +1,87 @@
...
+            f.setAccessible(true);
+            // We're expecting this to be MemorySegment[]
+            final Object objSegments = f.get(indexInput);
+            if (objSegments == null || !objSegments.getClass().isArray()) {
+                // It's not MemorySegment[]
+                return null;
+            }
+            final int numSegments = Array.getLength(objSegments);
+            final long[] addressAndSize = new long[2 * numSegments];
+            for (int i = 0; i < numSegments; i += 2) {
```
Vikasht34: Hoping to understand !! , Is this i+=2 is intentional? Are we trying to skip odd number of segments? 

or it should be 
```
for (int i = 0; i < numSegments; i ++) {
addressAndSize[2i] = (long) ADDRESS_METHOD.invoke(memorySegment);
                addressAndSize[2i + 1] = (long) BYTE_SIZE_METHOD.invoke(memorySegment);
}
```

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/MMapByteVectorValues.java
```diff
@@ -0,0 +1,73 @@
...
+    public byte[] vectorValue(int internalVectorId) throws IOException {
+        indexInput.seek(baseOffset + internalVectorId * oneVectorByteSize);
+        // Lazy initialization, in general this method is not expected to be called during search.
+        // During search, distance calculation will be done in a separated C++ code.
+        if (buffer == null) {
+            buffer = new byte[(int) oneVectorByteSize];
+        }
+        indexInput.readBytes(buffer, 0, buffer.length);
+        return buffer;
+    }
```
navneet1v, line 79: the byte[] returned from here will have length 4*dimension right?

src/main/java/org/opensearch/knn/memoryoptsearch/MemorySegmentAddressExtractorJDK21.java
```diff
@@ -0,0 +1,87 @@
...
+                final Object memorySegment = Array.get(objSegments, i);
+                addressAndSize[i] = (long) ADDRESS_METHOD.invoke(memorySegment);
+                addressAndSize[i + 1] = (long) BYTE_SIZE_METHOD.invoke(memorySegment);
+            }
+
+            return addressAndSize;
+        } catch (NoSuchFieldException | IllegalAccessException | IllegalArgumentException e) {
+            // Ignore
```
navneet1v: we should add a log message here, warn atleast.

src/main/java/org/opensearch/knn/memoryoptsearch/MemorySegmentAddressExtractorJDK22.java
0ctopus13prime: Will add it in next rev.

src/main/java/org/opensearch/knn/memoryoptsearch/MemorySegmentAddressExtractorJDK21.java
0ctopus13prime: no, MemorySegment is under preview, so it wont work for JDK21. 
We can use `objSegments instanceof MemorySegment[]` after JDK22+. 
So unless we set 22+ as the min target class version, this is the only way 🤑

src/main/java/org/opensearch/knn/memoryoptsearch/MemorySegmentAddressExtractorJDK21.java
0ctopus13prime: Will add it in next rev.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissMemoryOptimizedSearcher.java
0ctopus13prime: This is needed to avoid calling `getBytesValues` for every FP16 query. 
For FP16 query, `getFloatValues` needs to be called and in order to check whether if bulk SIMD is supported, calling `getByteValues` is the only way. And it will be guaranteed that `getByteValues` returns the same class, we call it to check whether native scoring is supported upfront in constructor, and branch on the value.

Alternatively, we can create MMapFloatValues similar to what I added MMapByteVectorValues.
Do you think it would be better to create MMapFloatValues?

src/main/java/org/opensearch/knn/memoryoptsearch/MemorySegmentAddressExtractorUtil.java
0ctopus13prime: Yes, it will kill the node and this is by design.
If it fails to initialize the extractor, that means old JDK is being used.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissIndexScalarQuantizedFlat.java
0ctopus13prime: will update in the next rev.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/MMapByteVectorValues.java
0ctopus13prime: oh, I missed adding java doc to this class 😅 
Thank you, will add it in next rev.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/MMapByteVectorValues.java
0ctopus13prime: yes, it is! for example, 128 dimension, then 4 * 128 will be passed as `oneVectorByteSize`

src/main/java/org/opensearch/knn/memoryoptsearch/MemorySegmentAddressExtractorJDK21.java
0ctopus13prime: Sure, will add it in next rev.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/MMapByteVectorValues.java
navneet1v: So here what I would suggest, please add java doc and be little verbose in explaining what this function is doing and explanation of parameters. This will help in future.

src/main/java/org/opensearch/knn/memoryoptsearch/MemorySegmentAddressExtractorUtil.java
```diff
@@ -0,0 +1,32 @@
...
+
+    static {
+        MemorySegmentAddressExtractor instance = null;
+        try {
+            // Try to load the JDK22-optimized class which will be packaged only for compile_target = 22+.
+            Class<?> clazz = Class.forName("org.opensearch.knn.memoryoptsearch.MemorySegmentAddressExtractorJDK22");
+            instance = (MemorySegmentAddressExtractor) clazz.getDeclaredConstructor().newInstance();
+        } catch (ClassNotFoundException e) {
+            // Class not found: fall back to JDK21 version
+            instance = new MemorySegmentAddressExtractorJDK21();
```
navneet1v: please add warn log here. So that we know which extractor is picked

src/main/java/org/opensearch/knn/memoryoptsearch/MemorySegmentAddressExtractorUtil.java
navneet1v: @0ctopus13prime we might need to think little bit more here, if you can provide some details why you went with killing of node by design approach that can help.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissMemoryOptimizedSearcher.java
navneet1v: I think it will be better to use `MMapFloatValues` rather than `MMapByteVectorValues` .

src/main/java/org/opensearch/knn/memoryoptsearch/MemorySegmentAddressExtractorJDK21.java
```diff
@@ -0,0 +1,87 @@
...
+        if (ADDRESS_METHOD == null || BYTE_SIZE_METHOD == null) {
+            // Runtime JDK does not support MemorySegment.
+            return null;
+        }
+
+        try {
+            // MMapDirectory in Lucene will return MemorySegmentIndexInput$SingleSegmentImpl or .$MultiSegmentImpl.
+            // Thus, get the super class (e.g. MemorySegmentIndexInput) to acquire `MemorySegment[] segments`.
+            final Field f = indexInput.getClass().getSuperclass().getDeclaredField("segments");
+            f.setAccessible(true);
```
Vikasht34, line 68: JPMS Illegal Access(InAccessibleObjectException) can throw at runtime and it's not caught by below catch expection list, either we wrap the whole code on try catch or handle this expection as well.

src/main/java/org/opensearch/knn/memoryoptsearch/MemorySegmentAddressExtractorJDK21.java
```diff
@@ -0,0 +1,87 @@
...
+    public long[] extractAddressAndSize(final IndexInput indexInput) {
+        if (ADDRESS_METHOD == null || BYTE_SIZE_METHOD == null) {
+            // Runtime JDK does not support MemorySegment.
+            return null;
+        }
+
+        try {
+            // MMapDirectory in Lucene will return MemorySegmentIndexInput$SingleSegmentImpl or .$MultiSegmentImpl.
+            // Thus, get the super class (e.g. MemorySegmentIndexInput) to acquire `MemorySegment[] segments`.
+            final Field f = indexInput.getClass().getSuperclass().getDeclaredField("segments");
```
Vikasht34, line 67: this code assumes that ,immediate superclass has a private field names exactly "segments". That's an internal  to lucene and can be changed, if lucene changes it m, this will quietly fail. Do u think we should handle upfront when we are takling such assumption?

src/main/java/org/opensearch/knn/memoryoptsearch/MemorySegmentAddressExtractorJDK21.java
```diff
@@ -0,0 +1,87 @@
...
+            // We're expecting this to be MemorySegment[]
+            final Object objSegments = f.get(indexInput);
+            if (objSegments == null || !objSegments.getClass().isArray()) {
+                // It's not MemorySegment[]
+                return null;
+            }
+            final int numSegments = Array.getLength(objSegments);
+            final long[] addressAndSize = new long[2 * numSegments];
+            for (int i = 0; i < numSegments; i += 2) {
+                final Object memorySegment = Array.get(objSegments, i);
```
Vikasht34: Array.get can be null, if slot is unused , Can we wrap the whole code on genral expection so that if any other issue happens apart from listed catch list we don't crash.

src/main/java/org/opensearch/knn/memoryoptsearch/MemorySegmentAddressExtractorJDK21.java
```diff
@@ -0,0 +1,87 @@
...
+                addressAndSize[i] = (long) ADDRESS_METHOD.invoke(memorySegment);
+                addressAndSize[i + 1] = (long) BYTE_SIZE_METHOD.invoke(memorySegment);
+            }
+
+            return addressAndSize;
+        } catch (NoSuchFieldException | IllegalAccessException | IllegalArgumentException e) {
```
Vikasht34: Here we are swallowing these exception, how we will be identifying behaviour depending on which reflection step failed?

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissMemoryOptimizedSearcher.java
0ctopus13prime: Will update in the next rev.

src/main/java/org/opensearch/knn/memoryoptsearch/MemorySegmentAddressExtractorUtil.java
0ctopus13prime: Sure, the goal of this static initialization is to instantiate global extractor which should be working regardless JDK versions. If it fails then I think there are two possibilities.

1. Bugs in MemorySegmentAddressExtractorJDK22's constructor. But the constructor is empty, can't fail.
If it fails, then there's something wrong with the code. 
2. Bugs in MemorySegmentAddressExtractorJDK21's constructor. And equally, the constructor is empty.

Therefore, if we failed to instantiate MemorySegmentAddressExtractor, then there's something wrong with the current code since the whole try-catch statement is trying to instantiate an extractor whose constructor is completely empty. 
Rather than falling back to Java scoring silently, I think it would be better to abort the node.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/MMapByteVectorValues.java
0ctopus13prime: sure, will add in the next rev.

src/main/java/org/opensearch/knn/memoryoptsearch/MemorySegmentAddressExtractorJDK21.java
0ctopus13prime: yeah, the loop is semantically the same as you suggested. 
resulting addressAndSize[i] will have address_i and addressAndSize[i + 1] will have size_i.

src/main/java/org/opensearch/knn/memoryoptsearch/MemorySegmentAddressExtractorJDK21.java
0ctopus13prime: Will update in the next rev.

Vikasht34 (commented):
Hoping to understand more on this , Correct me if I am wrong !!. We are adding reflection on our side,  What it takes to enhance IndexInput on Lucene Side apis to give this instead of doing in KNN. Can we propose supported SPI from Lucene instead of creating in KNN., Instead of ceating mmap vectorValues , What it takes to use FloatVectorValues or Byte from Lcuene.

src/main/java/org/opensearch/knn/memoryoptsearch/MemorySegmentAddressExtractorJDK21.java
0ctopus13prime: But if there's such changes in Lucene, then it probably will be shipped into a newer version, in turn,  those related unit tests will fail anyway. So we have to fix the fixture if there's change in Lucene regardless

src/main/java/org/opensearch/knn/memoryoptsearch/MemorySegmentAddressExtractorJDK21.java
0ctopus13prime: Sounds good, will update in the next rev.

src/main/java/org/opensearch/knn/memoryoptsearch/MemorySegmentAddressExtractorJDK21.java
0ctopus13prime: Reflection failing indicates either non-MMapDirectory is being used or user intentionally blocked using reflection. 
For those case, nothing we can do about it to enforce using this bulk SIMD optimization. 
I think we can silently fallback to the default scoring which will use Java code, but I think we should log warnning at least.
Will update in the next rev.

src/main/java/org/opensearch/knn/memoryoptsearch/MemorySegmentAddressExtractorUtil.java
0ctopus13prime: sure, will do in next rev

src/main/java/org/opensearch/knn/memoryoptsearch/MemorySegmentAddressExtractorJDK21.java
Vikasht34: Suppose We have 3 segements , with the loop we are just doing 0, 2 and skipped 1. but the suggetion I provided , it does not skip odd number sgements.

shatejas: Would prefer 2 separate variables here, one for index of segments and one for index of addressAndSize for clarity

src/main/java/org/opensearch/knn/memoryoptsearch/MemorySegmentAddressExtractorUtil.java
navneet1v: > Rather than falling back to Java scoring silently, I think it would be better to abort the node.

This can cause a widespread outage even when non k-NN indices are present if there are bugs in this code. I think it will be better if you do lazy init.

src/main/java/org/opensearch/knn/memoryoptsearch/MemorySegmentAddressExtractorJDK21.java
0ctopus13prime: Will update in the next rev.

src/main/java/org/opensearch/knn/memoryoptsearch/MemorySegmentAddressExtractorUtil.java
0ctopus13prime: Hmm.. but if there's bug then lazy init will still fail anyway 🤔 
If this is concerning, then I can set global extractor as null and return null long[] instead. Then it will fallback to Java distance calculation logic. Will log error for that case. 

Will update in the next rev

0ctopus13prime:
> overall looks good. One thing I would suggest is for JDK21 and even JDK22 and above, we should have a jvm args that can switch off the `MemorySegmentAddressExtractor` feature. See in which PR you want to add this capability.

I can add this in the last PR.

src/main/java/org/opensearch/knn/memoryoptsearch/MemorySegmentAddressExtractorUtil.java
navneet1v: > Hmm.. but if there's bug then lazy init will still fail anyway 🤔


Yes it will fail but in that case the specific shard will fail. It will not fail the whole node. So that is less disastrous that failing a node during bootup.

0ctopus13prime: Why the whole node will fail? 
The util is being used in FaissIndexScalarQuantizedFlat for now, that means it is trying to do vector search operation on a shard level, if it fails, then vector search will fail and I believe the whole node will remain safe... 🤔 

Any failures within static initialization will throw an exception if it's being captured, then JVM process will keep alive.

But anyway, I've updated to assign empty extractor will make the logic fallback to Java based distance calculation with error logging.

src/main/java/org/opensearch/knn/memoryoptsearch/MemorySegmentAddressExtractorUtil.java
```diff
@@ -0,0 +1,78 @@
...
+
+    static {
+        MemorySegmentAddressExtractor instance;
+        try {
+            try {
+                // Try to load the JDK22-optimized class which will be packaged only for compile_target = 22+.
+                Class<?> clazz = Class.forName("org.opensearch.knn.memoryoptsearch.MemorySegmentAddressExtractorJDK22");
+                instance = (MemorySegmentAddressExtractor) clazz.getDeclaredConstructor().newInstance();
+            } catch (ClassNotFoundException e) {
+                log.warn("Failed to load MemorySegmentAddressExtractorJDK22, falling back to MemorySegmentAddressExtractorJDK21");
```
navneet1v: please add the exception you caught here with warn log.

0ctopus13prime: sure will do.

0ctopus13prime:
Hi @Vikasht34 @navneet1v @shatejas 
I think I've addressed all comments, could you take a look and and approve it if all look good on you?
Please note that this is to merge against the feature branch.
Thank you.

Vikasht34 (approved):
Looks Good , Thanks for addressing the commnets

PR #2905, opened 2025-09-26
Fix radial search for byte[] faiss

### LINKED ISSUE #2864: [BUG] Radial Search with byte[] Vector throws exception

Index with byte data type:

```
curl -XPUT "http://localhost:9200/my-knn-index-1" \
-H 'Content-Type: application/json' \
-d'{
  "settings": {
    "index": {
      "knn": true,
      "knn.algo_param.ef_search": 100
    }
  },
  "mappings": {
    "properties": {
      "embedding_3_large": {
        "type": "knn_vector",
        "dimension": 3,
        "data_type": "byte",
        "method": {
          "engine": "faiss",
          "name": "hnsw",
          "space_type": "l2",
          "parameters": {
            "m": 16,
            "ef_construction": 256
          }
        }
      }
    }
  }
}'

```

Add data:

```
curl -XPOST "http://localhost:9200/my-knn-index-1/_doc" \
-H 'Content-Type: application/json' \
-d'{
  "embedding_3_large": [10, 20, 30]
}'
```

Query with radius:

```
curl -XGET "http://localhost:9200/my-knn-index-1/_search" \
-H 'Content-Type: application/json' \
-d'{
  "size": 10,
  "query": {
    "knn": {
      "embedding_3_large": {
        "vector": [12, 22, 32],
        "max_distance": 0.8
      }
    }
  }
}'
```

We get exception:

```
Caused by: java.lang.Exception: Query Vector cannot be null
        at org.opensearch.knn.jni.FaissService.rangeSearchIndex(Native Method) ~[?:?]
        at org.opensearch.knn.jni.JNIService.radiusQueryIndex(JNIService.java:464) ~[?:?]
        at org.opensearch.knn.index.query.DefaultKNNWeight.doANNSearch(DefaultKNNWeight.java:146) ~[?:?]
        at org.opensearch.knn.index.query.KNNWeight.approximateSearch(KNNWeight.java:468) ~[?:?]
        at org.opensearch.knn.index.query.KNNWeight.searchLeaf(KNNWeight.java:332) ~[?:?]
        at org.opensearch.knn.index.query.KNNWeight$1.get(KNNWeight.java:271) ~[?:?]
```

The query should be able to succeed.

markwu-sde:
### Description

Fix bug to enable radial search for faiss engine for byte[] vectors. This includes faiss and lucene on faiss implementations.

### Related Issues
Resolves #2864 

### Check List


src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
```diff
@@ -585,8 +585,8 @@ protected Query doToQuery(QueryShardContext context) {
                 .knnEngine(knnEngine)
                 .indexName(indexName)
                 .fieldName(this.fieldName)
-                .vector(VectorDataType.FLOAT == vectorDataType ? this.vector : null)
-                .byteVector(VectorDataType.BYTE == vectorDataType ? byteVector : null)
+                .vector(getVectorForCreatingQueryRequest(vectorDataType, knnEngine))
```
neetikasinghal, line 590: it would be good to add unit tests in KNNQueryBuilderTests.java for these changes along with the ITs that you have already added.

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
```diff
@@ -531,29 +531,14 @@ protected Query doToQuery(QueryShardContext context) {
...
-            case BINARY:
+            case BINARY, BYTE:
                 byteVector = new byte[vector.length];
                 for (int i = 0; i < vector.length; i++) {
                     validateByteVectorValue(vector[i], knnVectorFieldType.getVectorDataType());
                     byteVector[i] = (byte) vector[i];
                 }
                 spaceType.validateVector(byteVector);
                 break;
-            case BYTE:
```
neetikasinghal, line 542: why are we removing this switch case?

markwu-sde: `isUsingLuceneQuery` returns true if memOptimized is enabled or if the engine is lucene and it's only used for supporting these byte vector values. Now we are supporting lucene and faiss there's no need for this check anymore since we won't go into the else block for either engine. The logic then for BYTE and BINARY is the same now.

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
markwu-sde: Yeah good idea, added.

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
shatejas: > Now we are supporting lucene and faiss there's no need for this check anymore since we won't go into the else block for either engine. The logic then for BYTE and BINARY is the same now.

I am not sure if this is true, we can still turn off `memoryOptimizedSearch` and FAISS will use float and internally use int8. This might break the case where we are loading graphs and not using memoryOpt search for byte

@naveentatikonda @0ctopus13prime can you verify?

Vikasht34 (commented):
I am not sure if this is right fix , hoping to understand , correct me if I am wrong , the bug was during radial search if Data type is byte and when we are passing float vector.  
in this code here , it goes into else condition and in else we just validate but byte vector array was never filled , 

```
case BYTE:
                if (isUsingLuceneQuery(knnEngine, memoryOptimizedSearchSupported)) {
                    byteVector = new byte[vector.length];
                    for (int i = 0; i < vector.length; i++) {
                        validateByteVectorValue(vector[i], knnVectorFieldType.getVectorDataType());
                        byteVector[i] = (byte) vector[i];
                    }
                    spaceType.validateVector(byteVector);
                } else {
                    for (float v : vector) {
                        validateByteVectorValue(v, knnVectorFieldType.getVectorDataType());
                    }
                    spaceType.validateVector(vector);
                }
```

Simple fix would be according to me without removing the switch case , is that 
if vectortype is byte and if we have min/max or radius then this is radial search query and in this case in else condition byte vector array has to be filled. 

I feel if we are removing switch case without impact analysis will lead to failure. 
Hoping to understand more on this. 





then vector will be null  here 
```
.vector(VectorDataType.FLOAT == vectorDataType ? this.vector : null) 
```

```
RNNQueryFactory.CreateQueryRequest createQueryRequest = RNNQueryFactory.CreateQueryRequest.builder()
                .knnEngine(knnEngine)
                .indexName(indexName)
                .fieldName(this.fieldName)
                .vector(VectorDataType.FLOAT == vectorDataType ? this.vector : null)
                .byteVector(VectorDataType.BYTE == vectorDataType ? byteVector : null)
                .vectorDataType(vectorDataType)
                .radius(radius)
                .methodParameters(this.methodParameters)
                .filter(this.filter)
                .context(context)
                .memoryOptimizedSearchSupported(memoryOptimizedSearchSupported)
                .build();
```

markwu-sde:
> I am not sure if this is right fix , hoping to understand , correct me if I am wrong , the bug was during radial search if Data type is byte and when we are passing float vector. in this code here , it goes into else condition and in else we just validate but byte vector array was never filled ,
> 
> ```
> case BYTE:
>                 if (isUsingLuceneQuery(knnEngine, memoryOptimizedSearchSupported)) {
>                     byteVector = new byte[vector.length];
>                     for (int i = 0; i < vector.length; i++) {
>                         validateByteVectorValue(vector[i], knnVectorFieldType.getVectorDataType());
>                         byteVector[i] = (byte) vector[i];
>                     }
>                     spaceType.validateVector(byteVector);
>                 } else {
>                     for (float v : vector) {
>                         validateByteVectorValue(v, knnVectorFieldType.getVectorDataType());
>                     }
>                     spaceType.validateVector(vector);
>                 }
> ```
> 
> Simple fix would be according to me without removing the switch case , is that if vectortype is byte and if we have min/max or radius then this is radial search query and in this case in else condition byte vector array has to be filled.
> 
> I feel if we are removing switch case without impact analysis will lead to failure. Hoping to understand more on this.
> 
> then vector will be null here
> 
> ```
> .vector(VectorDataType.FLOAT == vectorDataType ? this.vector : null) 
> ```
> 
> ```
> RNNQueryFactory.CreateQueryRequest createQueryRequest = RNNQueryFactory.CreateQueryRequest.builder()
>                 .knnEngine(knnEngine)
>                 .indexName(indexName)
>                 .fieldName(this.fieldName)
>                 .vector(VectorDataType.FLOAT == vectorDataType ? this.vector : null)
>                 .byteVector(VectorDataType.BYTE == vectorDataType ? byteVector : null)
>                 .vectorDataType(vectorDataType)
>                 .radius(radius)
>                 .methodParameters(this.methodParameters)
>                 .filter(this.filter)
>                 .context(context)
>                 .memoryOptimizedSearchSupported(memoryOptimizedSearchSupported)
>                 .build();
> ```

Yeah I agree with this I think modifying the switch logic might impact too many other flows. We can just check the existence of the `radius` field and populate the byte vector field appropriately without modifying the existing logic.

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
markwu-sde: Yeah I think I'll defer to keeping the changes to a minimum and just populate the necessary fields for radial search.

src/main/java/org/opensearch/knn/index/query/memoryoptsearch/MemoryOptimizedKNNWeight.java
```diff
@@ -156,7 +156,7 @@ protected TopDocs doANNSearch(
             } else {
                 // Radius search
                 return queryIndex(
-                    knnQuery.getQueryVector(),
+                    knnQuery.getVectorDataType() == VectorDataType.BYTE ? knnQuery.getByteQueryVector() : knnQuery.getQueryVector(),
```
shatejas: Maybe create a `Object getVector()` method in knnQuery instead of doing it here?

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
```diff
@@ -686,15 +686,9 @@ private float[] getVectorForCreatingQueryRequest(VectorDataType vectorDataType,
...
-        VectorDataType vectorDataType,
-        KNNEngine knnEngine,
-        byte[] byteVector,
-        boolean memoryOptimizedSearchEnabled
-    ) {
+    private byte[] getVectorForCreatingQueryRequest(VectorDataType vectorDataType, byte[] byteVector) {
 
-        if (VectorDataType.BINARY == vectorDataType
-            || (VectorDataType.BYTE == vectorDataType && isUsingLuceneQuery(knnEngine, memoryOptimizedSearchEnabled))) {
+        if (VectorDataType.BINARY == vectorDataType || (VectorDataType.BYTE == vectorDataType)) {
```
shatejas: There is still a behavior change here, Why is memoryOpt not being considered here? this impacts KNNQuery as well. 

My understanding is
```
if faiss engine and memory opt disabled => populates byte data type as float vector
if faiss and memory opt enabled => populates byte data type as byte vector
if lucene => populates byte data type as byte vector
```
Am I missing something?

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
```diff
@@ -540,7 +540,7 @@ protected Query doToQuery(QueryShardContext context) {
                 spaceType.validateVector(byteVector);
                 break;
             case BYTE:
-                if (isUsingLuceneQuery(knnEngine, memoryOptimizedSearchEnabled)) {
+                if (isUsingLuceneQuery(knnEngine, memoryOptimizedSearchEnabled) || radius != null) {
```
shatejas: Does that mean all byte radial queries is using byteVector? 

[JNI layer](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/jni/JNIService.java#L440) does not seem to accept bytevector

markwu-sde: We actually don't need this check here because for FAISS engine with non-memory optimized we use the query vector anyways.

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
markwu-sde: Yeah - Right now the radial search path goes into 2 different flows (1 for memOpt and another for JNI)

The memory optimized flow exposes a different pathing for byte[] vector but it requires the object to be the of type byte[]. So we'll need to use the byteVector[] field here instead of queryVector. 

The non-memory opt flow for faiss will use the queryVector[] anyways as our JNILib only exposes float[] radial interface.

src/main/java/org/opensearch/knn/index/query/memoryoptsearch/MemoryOptimizedKNNWeight.java
markwu-sde: Sure yeah.

src/main/java/org/opensearch/knn/index/query/KNNQuery.java
```diff
@@ -251,6 +251,10 @@ private Weight getFilterWeight(IndexSearcher searcher) throws IOException {
         return null;
     }
 
+    public Object getVector() {
+        return vectorDataType == VectorDataType.BYTE ? byteQueryVector : queryVector;
+    }
```
shatejas, line 265:
```suggestion
    public Object getVector() {
        return switch(vectorDataType) {
                  BYTE -> {
                         if (isMemoryOptimizedSearch) {
                              yield byteQueryVector;
                         }
                         yield queryVector;
                  }
                  BINARY ->  byteQueryVector;
                  FLOAT -> queryVector
                  
        }
    }
```

Also please add a unit test here https://github.com/opensearch-project/k-NN/blob/main/src/test/java/org/opensearch/knn/index/query/KNNQueryTests.java#L11

src/test/java/org/opensearch/knn/index/query/KNNQueryBuilderTests.java
```diff
@@ -1125,6 +1126,62 @@ public void testRadialSearch_whenEfSearchIsSet_whenFaissEngine_thenSuccess() {
...
+        when(mockQueryShardContext.index()).thenReturn(dummyIndex);
+        when(mockKNNVectorField.getVectorDataType()).thenReturn(VectorDataType.BYTE);
+        when(mockQueryShardContext.fieldMapper(anyString())).thenReturn(mockKNNVectorField);
+        IndexSettings indexSettings = mock(IndexSettings.class);
+        when(mockQueryShardContext.getIndexSettings()).thenReturn(indexSettings);
+        when(indexSettings.getMaxResultWindow()).thenReturn(1000);
+
+        KNNQuery query = (KNNQuery) knnQueryBuilder.doToQuery(mockQueryShardContext);
+        assertEquals(1 / MIN_SCORE - 1, query.getRadius(), 0);
+        assertEquals(queryVector, query.getQueryVector());
```
shatejas, line 1181:
```suggestion
        assertEquals(queryVector, query.getQueryVector());
        assertNull(query.getByteQueryVector())
```

src/main/java/org/opensearch/knn/index/query/KNNQuery.java
markwu-sde: Added.

src/test/java/org/opensearch/knn/index/query/KNNQueryBuilderTests.java
markwu-sde: revised.

shatejas (dismissed):

Vikasht34 (approved):
Looks good to me

PR #2909, opened 2025-09-29
Add lateInteractionFunction

mingshl:
### Description
Moved the lateInteraction Function PR to K-nn, previous details in core https://github.com/opensearch-project/OpenSearch/pull/18727 

This is an interim solution for reranking using lateInteraction models, we still plan to migrate to native Lucene implementation when available, keeping user API consistent

Implemented SIMD-optimized dot product using Java Vector API (jdk.incubator.vector)

### Related Issues
https://github.com/opensearch-project/OpenSearch/issues/18091

### Check List


mingshl:
build test success for linux 21 and 24. will add in changlog after CI tests complete. 

Successful in 7m
[Build and Test k-NN / Build and Test k-NN Plugin on Linux (21) (pull_request)](https://github.com/opensearch-project/k-NN/actions/runs/18109377607/job/51531833331?pr=2909)
Build and Test k-NN / Build and Test k-NN Plugin on Linux (21) (pull_request)Successful in 54m
[Build and Test k-NN / Build and Test k-NN Plugin on Linux (24) (pull_request)](https://github.com/opensearch-project/k-NN/actions/runs/18109377607/job/51531833363?pr=2909)
Build and Test k-NN / Build and Test k-NN Plugin on Linux (24) (pull_request)Successful in 54m

src/main/java/org/opensearch/knn/plugin/script/KNNPainlessScriptUtils.java
```diff
@@ -0,0 +1,85 @@
...
+                }
+            }
+            totalMaxSim += maxDocTokenSim;
+        }
+        return totalMaxSim;
+    }
+
+    private static double dotProduct(List<Double> vec1, List<Double> vec2) {
+        if (vec1 == null || vec2 == null || vec1.size() != vec2.size()) {
+            return 0.0;
```
martin-gaievski: I'm not sure if returning 0.0 in case of dimensionality mismatch is optimal behavior, my vote is for failing fast with exception. this way user will see the problem is the query or data earlier. Am I correct assuming that such mismatch in vector sizes can be in case we're sending query vector that has been generated with the model of different dimensionality than the one used for data ingestion?
If you still prefer returning 0.0 then at least add log information to ease the potential troubleshooting

src/main/java/org/opensearch/knn/plugin/script/KNNPainlessScriptUtils.java
```diff
@@ -0,0 +1,85 @@
...
+
+import jdk.incubator.vector.DoubleVector;
+import jdk.incubator.vector.VectorOperators;
+import jdk.incubator.vector.VectorSpecies;
+
+public final class KNNPainlessScriptUtils {
+
+    private static final VectorSpecies<Double> SPECIES = DoubleVector.SPECIES_PREFERRED;
+
+    private KNNPainlessScriptUtils() {}
```
martin-gaievski: nit: such boilerplate code can be simplified with lombok, knn has support for it

src/main/java/org/opensearch/knn/plugin/script/KNNPainlessScriptUtils.java
```diff
@@ -0,0 +1,85 @@
...
+                    continue;
+                }
+
+                double currentSim = dotProduct(q_vec, doc_token_vec);
+
+                if (currentSim > maxDocTokenSim) {
+                    maxDocTokenSim = currentSim;
+                }
+            }
+            totalMaxSim += maxDocTokenSim;
```
martin-gaievski: are we going with assumption that vectors are normalized, any restrictions we do have to final dot product?

src/main/java/org/opensearch/knn/plugin/script/KNNPainlessScriptUtils.java
```diff
@@ -0,0 +1,85 @@
...
+        }
+        return totalMaxSim;
+    }
+
+    private static double dotProduct(List<Double> vec1, List<Double> vec2) {
+        if (vec1 == null || vec2 == null || vec1.size() != vec2.size()) {
```
martin-gaievski: in addition to equality of vector sizes we need to check if any of those arrays exceeds the max dimensions value, I think it can be different for different engines.

src/main/java/org/opensearch/knn/plugin/script/KNNPainlessScriptUtils.java
```diff
@@ -0,0 +1,85 @@
...
+            return 0.0;
+        }
+        double[] a = vec1.stream().mapToDouble(d -> d).toArray();
+        double[] b = vec2.stream().mapToDouble(d -> d).toArray();
+        double res = 0;
+        int i = 0;
+        for (; i < SPECIES.loopBound(a.length); i += SPECIES.length()) {
+            DoubleVector va = DoubleVector.fromArray(SPECIES, a, i);
+            DoubleVector vb = DoubleVector.fromArray(SPECIES, b, i);
+            res += va.mul(vb).reduceLanes(VectorOperators.ADD);
```
martin-gaievski: with direct summation like this with high number of dimensions it's possible to get high enough accumulated error. check with knn folks, if they are using any library function for such operations with floating point numbers.

src/main/java/org/opensearch/knn/plugin/script/KNNPainlessScriptUtils.java
```diff
@@ -0,0 +1,85 @@
...
+        }
+        for (; i < a.length; i++) {
+            res += a[i] * b[i];
```
martin-gaievski: same problem a above, accumulated error can be significant with the large number of dimensions

src/main/java/org/opensearch/knn/plugin/script/KNNPainlessScriptUtils.java
```diff
@@ -0,0 +1,85 @@
...
+            totalMaxSim += maxDocTokenSim;
+        }
+        return totalMaxSim;
+    }
+
+    private static double dotProduct(List<Double> vec1, List<Double> vec2) {
+        if (vec1 == null || vec2 == null || vec1.size() != vec2.size()) {
+            return 0.0;
+        }
+        double[] a = vec1.stream().mapToDouble(d -> d).toArray();
```
martin-gaievski: any chance you can avoid conversion of list to array of doubles? Number of such temporary array can go very high which is not good for already memory intense application like knn. example - for query with 32 tokens and document with 512 tokens this creates 32K of temporary arrays.

src/main/java/org/opensearch/knn/plugin/script/KNNPainlessScriptUtils.java
```diff
@@ -0,0 +1,85 @@
...
+package org.opensearch.knn.plugin.script;
+
+import java.util.List;
+import java.util.Map;
+
+import jdk.incubator.vector.DoubleVector;
+import jdk.incubator.vector.VectorOperators;
+import jdk.incubator.vector.VectorSpecies;
+
+public final class KNNPainlessScriptUtils {
```
navneet1v: Lets add java doc on all the public functions and public classes for better redability.

src/main/java/org/opensearch/knn/plugin/script/KNNPainlessScriptUtils.java
```diff
@@ -0,0 +1,85 @@
...
+            }
+
+            double maxDocTokenSim = 0.0;
+
+            for (List<Double> doc_token_vec : docVectors) {
+                if (doc_token_vec == null || doc_token_vec.isEmpty()) {
+                    continue;
+                }
+
+                double currentSim = dotProduct(q_vec, doc_token_vec);
```
navneet1v: why we just have dotProduct here? what about cosine and L2?

src/main/java/org/opensearch/knn/plugin/script/KNNPainlessScriptUtils.java
```diff
@@ -0,0 +1,85 @@
...
+import java.util.List;
+import java.util.Map;
+
+import jdk.incubator.vector.DoubleVector;
+import jdk.incubator.vector.VectorOperators;
+import jdk.incubator.vector.VectorSpecies;
+
+public final class KNNPainlessScriptUtils {
+
+    private static final VectorSpecies<Double> SPECIES = DoubleVector.SPECIES_PREFERRED;
```
navneet1v: This is not needed. Please use KNNSpaceType for distance calculation as suggested

src/main/java/org/opensearch/knn/plugin/script/KNNPainlessScriptUtils.java
```diff
@@ -0,0 +1,85 @@
...
+     * For each query vector, finds the maximum dot product with any document vector and sums these maxima.
+     * This implements a ColBERT-style late interaction pattern for token-level matching.
+     *
+     * @param queryVectors List of query vectors, each a list of doubles
+     * @param docFieldName Name of the field in the document containing vectors
+     * @param doc Document source as a map
+     * @return Sum of maximum similarity scores
+     */
+    @SuppressWarnings("unchecked")
+    public static double lateInteractionScore(List<List<Double>> queryVectors, String docFieldName, Map<String, Object> doc) {
```
navneet1v: please make all the args final and @NonNull.

Also why we are using `List<List<Double>> queryVectors` why not `List<float[]>` .

src/main/java/org/opensearch/knn/plugin/script/KNNPainlessScriptUtils.java
```diff
@@ -0,0 +1,85 @@
...
+        int i = 0;
+        for (; i < SPECIES.loopBound(a.length); i += SPECIES.length()) {
+            DoubleVector va = DoubleVector.fromArray(SPECIES, a, i);
+            DoubleVector vb = DoubleVector.fromArray(SPECIES, b, i);
+            res += va.mul(vb).reduceLanes(VectorOperators.ADD);
+        }
+        for (; i < a.length; i++) {
+            res += a[i] * b[i];
+        }
+        return res;
```
navneet1v: Lets use the scoring function from KNN rather than rewriting it. Please use https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/SpaceType.java and score translation functionality. This will ensure consistency across all distance computation and also performance

src/test/java/org/opensearch/knn/integ/LateInteractionScoreIT.java
```diff
@@ -0,0 +1,90 @@
...
+import org.opensearch.core.rest.RestStatus;
+import org.opensearch.common.xcontent.XContentFactory;
+import org.opensearch.core.xcontent.XContentBuilder;
+
+import java.util.List;
+import java.util.ArrayList;
+
+public class LateInteractionScoreIT extends KNNRestTestCase {
+
+    public void testLateInteractionScore() throws Exception {
```
navneet1v: in kNN the tests names should be `testLateInteractionScore_whenABC_thenXYZ()`. Please rename all the test functions.

build.gradle
```diff
@@ -246,6 +248,11 @@ allprojects {
         project.licenseFile = rootProject.file('LICENSE.txt')
         project.noticeFile = rootProject.file('NOTICE.txt')
         project.forbiddenApis.ignoreFailures = true
+        if (project.tasks.findByName("forbiddenApisMain")) {
+            tasks.named("forbiddenApisMain").configure {
+                enabled = false
+            }
+        }
```
navneet1v: why we need this?

build.gradle
```diff
@@ -233,6 +233,8 @@ allprojects {
         sourceCompatibility = targetCompatibility = "21"
     }
 
+    
+
```
navneet1v: please remove this

build.gradle
```diff
@@ -289,6 +296,7 @@ publishing {
 
 compileJava {
     options.compilerArgs.addAll(["-processor", 'lombok.launch.AnnotationProcessorHider$AnnotationProcessor'])
+    options.compilerArgs += ['--add-modules', 'jdk.incubator.vector']
```
navneet1v: You don't need this use SpaceType abstraction for vector apis

build.gradle
```diff
@@ -352,6 +370,7 @@ dependencies {
     testImplementation group: 'org.objenesis', name: 'objenesis', version: "${versions.objenesis}"
     testImplementation group: 'net.bytebuddy', name: 'byte-buddy-agent', version: "${versions.bytebuddy}"
     testImplementation("com.google.guava:guava:${versions.guava}")
+    testImplementation "org.opensearch:opensearch-core:${opensearch_version}"
```
navneet1v: why we need this?

src/main/java/org/opensearch/knn/plugin/script/KNNPainlessScriptUtils.java
navneet1v: @martin-gaievski I think we should not even create our own dot product function. Lets use SpaceType abstraction for doing the distance computations.

build.gradle
```diff
@@ -441,6 +461,7 @@ test {
     systemProperty "java.library.path", "$rootDir/jni/build/release"
     //this change enables mockito-inline that supports mocking of static classes/calls
     systemProperty "jdk.attach.allowAttachSelf", true
+    jvmArgs '--add-modules', 'jdk.incubator.vector'
```
Vikasht34: Hoping to understand more on this !! What happens when Opensearch server/jvm where the plugin module is not loaded so even a single refrence to double vector will make the jvm try to link incubator at load time. If the server is not started with --add-modules', 'jdk.incubator.vector'' , we will get NoClassDefFoundError/ClassNotFoundException during loading.

Or Do we even need this ?

src/main/java/org/opensearch/knn/plugin/script/KNNPainlessScriptUtils.java
mingshl: nice call, 
added these following: 
- • innerproduct 
- • cosinesimil 
- • l2
- • l1 
- • linf

build.gradle
mingshl: oh I no longer need this because I use the KNNScoringUtils.

build.gradle
mingshl: removed

src/main/java/org/opensearch/knn/plugin/script/KNNPainlessScriptUtils.java
mingshl: good callout, do we have limitation for dimension value in KNN repo? I will try look for reference, I cannot make up the magic number

src/main/java/org/opensearch/knn/plugin/script/KNNPainlessScriptUtils.java
mingshl: used lombok.

src/main/java/org/opensearch/knn/plugin/script/KNNPainlessScriptUtils.java
mingshl: agreed. even though this risk the waste of resources of searching, but we won't return 0 in scoring which is not useful at all. I used non-null in the parameters definition, so they cannot be null now.

applied fast error check.

src/main/java/org/opensearch/knn/plugin/script/KNNPainlessScriptUtils.java
mingshl: changed to use KNNScoringUtils.innerProduct, this part is removed

src/main/java/org/opensearch/knn/plugin/script/KNNPainlessScriptUtils.java
mingshl: changed to used list of float array to avoid the conversions. in fact the KNNScoringUtil has the spaceType calculation also use  list of float array

build.gradle
mingshl: removed

build.gradle
mingshl: removed

src/main/java/org/opensearch/knn/plugin/script/KNNPainlessScriptUtils.java
mingshl: applied!!!

src/main/java/org/opensearch/knn/plugin/script/KNNPainlessScriptUtils.java
mingshl: changed to list of float array

src/main/java/org/opensearch/knn/plugin/script/KNNPainlessScriptUtils.java
mingshl: changed to spaceType, this is making my life easier!!!

src/test/java/org/opensearch/knn/integ/LateInteractionScoreIT.java
mingshl: changed

src/main/java/org/opensearch/knn/plugin/script/KNNPainlessScriptUtils.java
mingshl: added

src/main/java/org/opensearch/knn/plugin/script/KNNPainlessScriptUtils.java
```diff
@@ -0,0 +1,129 @@
...
+     * @param docFieldName Name of the field in the document containing vectors
+     * @param doc Document source as a map
+     * @return Sum of maximum similarity scores
+     */
+    public static float lateInteractionScore(
+        @NonNull final List<List<Number>> queryVectors,
+        @NonNull final String docFieldName,
+        @NonNull final Map<String, Object> doc
+    ) {
+        return lateInteractionScore(queryVectors, docFieldName, doc, "innerproduct");
```
martin-gaievski: can we retrieve space name from knn enum, something like


```suggestion
        return lateInteractionScore(queryVectors, docFieldName, doc, SpaceType.INNER_PRODUCT.getValue());
```

martin-gaievski (commented):
minor comment, overall looks good

src/main/java/org/opensearch/knn/plugin/script/KNNPainlessScriptUtils.java
```diff
@@ -0,0 +1,129 @@
...
+     * @param docFieldName Name of the field in the document containing vectors
+     * @param doc Document source as a map
+     * @return Sum of maximum similarity scores
+     */
+    public static float lateInteractionScore(
+        @NonNull final List<List<Number>> queryVectors,
+        @NonNull final String docFieldName,
+        @NonNull final Map<String, Object> doc
+    ) {
+        return lateInteractionScore(queryVectors, docFieldName, doc, SpaceType.INNER_PRODUCT.getValue());
```
navneet1v: The default for k-NN is L2, can we use L2 as a default here

src/main/java/org/opensearch/knn/plugin/script/KNNPainlessScriptUtils.java
```diff
@@ -0,0 +1,129 @@
...
+     * @return Sum of maximum similarity scores
+     */
+    @SuppressWarnings("unchecked")
+    public static float lateInteractionScore(
+        @NonNull final List<List<Number>> queryVectors,
+        @NonNull final String docFieldName,
+        @NonNull final Map<String, Object> doc,
+        @NonNull final String spaceType
+    ) {
+        List<List<Number>> docVectors = (List<List<Number>>) doc.get(docFieldName);
```
navneet1v: can we add try catch on this, as this typecasting can lead to NPE, cast exceptions

src/main/java/org/opensearch/knn/plugin/script/KNNPainlessScriptUtils.java
```diff
@@ -0,0 +1,129 @@
...
+            case COSINESIMIL:
+                return KNNScoringUtil.cosinesimil(vec1, vec2);
+            case L2:
+                return KNNScoringUtil.l2Squared(vec1, vec2);
+            case L1:
+                return KNNScoringUtil.l1Norm(vec1, vec2);
+            case LINF:
+                return KNNScoringUtil.lInfNorm(vec1, vec2);
+            default:
+                return KNNScoringUtil.innerProduct(vec1, vec2);
```
navneet1v: we should throw exception from here rather than defaulting to IP. We have already set the default

src/main/java/org/opensearch/knn/plugin/script/KNNPainlessScriptUtils.java
```diff
@@ -0,0 +1,129 @@
...
+     * @param doc Document source as a map
+     * @param spaceType Space type for similarity calculation: "innerproduct", "cosinesimil", "l2", "l1", "linf"
+     * @return Sum of maximum similarity scores
+     */
+    @SuppressWarnings("unchecked")
+    public static float lateInteractionScore(
+        @NonNull final List<List<Number>> queryVectors,
+        @NonNull final String docFieldName,
+        @NonNull final Map<String, Object> doc,
+        @NonNull final String spaceType
```
navneet1v: does these NonNull checks valid here? In my understanding the painless script will directly call these functions, and it can happen that docFieldName is null and can lead to NPE due to @NonNull annotation. It will be better if you can add a validation utility here which can validate all the fields and then throw a user friendly exception and message to user

src/main/java/org/opensearch/knn/plugin/script/KNNPainlessScriptUtils.java
```diff
@@ -0,0 +1,129 @@
...
+                    maxDocTokenSim = Math.max(maxDocTokenSim, similarity);
+                }
+            }
+
+            totalMaxSim += maxDocTokenSim;
+        }
+        return totalMaxSim;
+    }
+
+    private static float calculateSimilarity(@NonNull final float[] vec1, @NonNull final float[] vec2, @NonNull final SpaceType spaceType) {
```
navneet1v: Please avoid adding @NonNull on private functions.

src/main/java/org/opensearch/knn/plugin/script/KNNPainlessScriptUtils.java
```diff
@@ -0,0 +1,129 @@
...
+            if (queryVector == null || queryVector.isEmpty()) {
+                throw new IllegalArgumentException("Every single vector within query vectors cannot be empty or null");
+            }
+
+            float[] qVec = new float[queryVector.size()];
+            for (int i = 0; i < queryVector.size(); i++) {
+                qVec[i] = queryVector.get(i).floatValue();
+            }
+
+            float maxDocTokenSim = isDistanceMetric ? Float.MAX_VALUE : Float.MIN_VALUE;
```
navneet1v: I don't think you should use this. KNNScoringUtil will already be taking care of this, where large score means better result.

src/main/java/org/opensearch/knn/plugin/script/KNNPainlessScriptUtils.java
```diff
@@ -0,0 +1,129 @@
...
+                }
+
+                float similarity = calculateSimilarity(qVec, dVec, space);
+
+                if (isDistanceMetric) {
+                    maxDocTokenSim = Math.min(maxDocTokenSim, similarity);
+                } else {
+                    maxDocTokenSim = Math.max(maxDocTokenSim, similarity);
+                }
+            }
```
navneet1v: We should parallelize this calculation, for better throughput.

src/main/java/org/opensearch/knn/plugin/script/KNNPainlessScriptUtils.java
mingshl: followed knn to use L2 as default

src/main/java/org/opensearch/knn/plugin/script/KNNPainlessScriptUtils.java
mingshl: added catch

src/main/java/org/opensearch/knn/plugin/script/KNNPainlessScriptUtils.java
mingshl: will throw exception when it's not supported spacetype, also added IT and UT to verify the exception

src/main/java/org/opensearch/knn/plugin/script/KNNPainlessScriptUtils.java
mingshl: removed NonNull and also add validation method

src/main/java/org/opensearch/knn/plugin/script/KNNPainlessScriptUtils.java
mingshl: this method is removed as I use KNNVectorSimilarityFunction.compare() now

src/main/java/org/opensearch/knn/plugin/script/KNNPainlessScriptUtils.java
mingshl: adapted KNNVectorSimilarityFunction.compare()

src/main/java/org/opensearch/knn/plugin/script/KNNPainlessScriptUtils.java
mingshl: use streaming method to parallelize each docVector compute

src/main/java/org/opensearch/knn/plugin/script/KNNPainlessScriptUtils.java
```diff
@@ -0,0 +1,129 @@
...
+                    for (int i = 0; i < docVector.size(); i++) {
+                        dVec[i] = docVector.get(i).floatValue();
+                    }
+                    return similarityFunction.compare(qVec, dVec);
+                })
+                .max()
+                .orElse(Double.NEGATIVE_INFINITY);
+
+            if (bestRawScore != Double.NEGATIVE_INFINITY) {
+                totalScore += space.scoreTranslation(bestRawScore);
```
navneet1v: I think we don't need score translation here since similarityFunction is already doing it for us.

src/main/java/org/opensearch/knn/plugin/script/KNNPainlessScriptUtils.java
```diff
@@ -0,0 +1,129 @@
...
+        final Map<String, Object> doc,
+        final String spaceType
+    ) {
+        if (queryVectors == null) {
+            throw new IllegalArgumentException("Query vectors cannot be null");
+        }
+        if (queryVectors.isEmpty()) {
+            throw new IllegalArgumentException("Query vectors cannot be empty");
+        }
+        if (docFieldName == null || docFieldName.trim().isEmpty()) {
```
navneet1v: See if you want to use StringUtils.isEmpty function here.

src/main/java/org/opensearch/knn/plugin/script/KNNPainlessScriptUtils.java
```diff
@@ -0,0 +1,129 @@
...
+            throw new IllegalArgumentException("Document field name cannot be null or empty");
+        }
+        if (doc == null) {
+            throw new IllegalArgumentException("Document cannot be null");
+        }
+        if (spaceType == null || spaceType.trim().isEmpty()) {
```
navneet1v: Same as above

src/main/java/org/opensearch/knn/plugin/script/KNNPainlessScriptUtils.java
mingshl: https://github.com/opensearch-project/k-NN/blob/b93ba10a403b2b409ad988f4de1c059d4ea36fe3/src/main/java/org/opensearch/knn/index/SpaceType.java#L58 it doesn't call scoreTranslation here, right?  it just call lucene's org.apache.lucene.index.VectorSimilarityFunction;

mingshl: oh, I see, here in KNN, it doesn't call scoreTranslate, but we don't need translate, as lucene already did https://github.com/apache/lucene/blob/836c8e76ad15fcd2894df215a21c8cbb293062a9/lucene/core/src/java/org/apache/lucene/index/VectorSimilarityFunction.java#L35-L37

src/main/java/org/opensearch/knn/plugin/script/KNNPainlessScriptUtils.java
mingshl: used Strings.isNullOrEmpty, which is better

src/main/java/org/opensearch/knn/plugin/script/KNNPainlessScriptUtils.java
mingshl: used Strings.isNullOrEmpty, which is better

src/main/java/org/opensearch/knn/plugin/script/KNNPainlessScriptUtils.java
mingshl: removed scoreTranslate

martin-gaievski (approved):
Looks good to me, thank you!

mingshl:
my local tests are passed on unit test and IT too 

```
./gradlew :test --tests "*KNNPainlessScriptUtilsTests*" -x cmakeJniLib -x buildJniLib


BUILD SUCCESSFUL in 42s

 ./gradlew ':integTest' --tests 'org.opensearch.knn.integ.LateInteractionScoreIT.testLateInteractionScore_whenValidVectors_thenReturnsCorrectScore'   -x cmakeJniLib -x buildJniLib

> Configure project :
Java home directory used by gradle: /Users/mingshl/Library/Java/JavaVirtualMachines/corretto-21.0.4/Contents/Home
CMake command: cmake -S jni -B jni/build -DKNN_PLUGIN_VERSION=3.3.0-SNAPSHOT -DCOMMIT_LIB_PATCHES=true -DAPPLY_LIB_PATCHES=true -DCMAKE_POLICY_VERSION_MINIMUM=3.5
Build command: cmake --build jni/build --target opensearchknn_faiss opensearchknn_common opensearchknn_nmslib --parallel 1


BUILD SUCCESSFUL in 1m 18s

```

src/main/java/org/opensearch/knn/plugin/script/KNNPainlessScriptUtils.java
```diff
@@ -0,0 +1,129 @@
...
+            if (queryVector == null || queryVector.isEmpty()) {
+                throw new IllegalArgumentException("Every single vector within query vectors cannot be empty or null");
+            }
+
+            float[] qVec = new float[queryVector.size()];
+            for (int i = 0; i < queryVector.size(); i++) {
+                qVec[i] = queryVector.get(i).floatValue();
+            }
+
+            float bestRawScore = (float) docVectors.parallelStream()
```
shatejas: I wouldn't prefer using `parallelStream()`. It uses a common forkjoin pool that can cause resource starvation/ increase in CPU. Its best to use common OS pool for better resource management

mingshl: This is a static utility class and doesn't have access to OpenSearch thread pool, if you are concerning about ForkJoinPool, we can use regular stream, but this won't be parallel @navneet1v

navneet1v: @mingshl lets change this to normal stream and then take it as an optimization in next PR to see how we can use OpenSearch Common Pool

navneet1v: Thanks @shatejas for pointing out. :)

mingshl: sounds good, let's do normal stream. It's better in stability than potential performance 
updated in new commit: 
[use stream rather than parallel stream](https://github.com/opensearch-project/k-NN/pull/2909/commits/0469df3c152b70968197fe152e882af189522d17)

src/test/java/org/opensearch/knn/integ/LateInteractionScoreIT.java
```diff
@@ -0,0 +1,201 @@
...
+    private void createIndexWithMapping() throws Exception {
+        String mapping = "{\n"
+            + "  \"properties\": {\n"
+            + "    \"my_vector\": {\n"
+            + "      \"type\": \"object\",\n"
+            + "      \"enabled\": false\n"
+            + "    }\n"
+            + "  }\n"
+            + "}";
+        createKnnIndex(INDEX_NAME, mapping);
```
navneet1v, line 175: this is not a k-NN field. I thought we will be using nested K-NN fields only for late interaction. 

@vamshin can you please correct me what was decided on this.

mingshl:
> Putting this as request for changes since I am not sure if we want this painless script to work on fields other than k-NN field.

lateInteractionScore is for reranking phase, we can support different field types later. Also reranking phase doesn't have to depend on the search phase. If user would love to have nested vector field for search and have a float/object field for reranking, that's doable. 

We decided to support float/object first as the lateInteraction models like Colbert, produce nested float array, and we can use mean pooling to a single knn field for fast retrieval and simple float/object field for second level reranking. 

If there are use cases more favorable in searching using nested vector field, rather than knn query type, we can support vested vector field type as well. It's not one way decision.


Now I have the function interface as following

```
   public static float lateInteractionScore(
        final List<List<Number>> queryVectors,
        final String docFieldName,
        final Map<String, Object> doc,
        final String spaceType
    ) 

```
I can easily achieve support nested vector field in two ways,

1. I can also add another interface that accept two docFieldNames that can support nested vector which needs two field names. 

```
   public static float lateInteractionScore(
        final List<List<Number>> queryVectors,
        final String nestedVectordocFieldName,
        final String innerVectordocFieldName,
        final Map<String, Object> doc,
        final String spaceType
    ) 

```
for example `lateInteractionScore(queryVectors, "vectors", "embeddings", doc, spaceType)`

2. keep the interface the same but use FieldPath to specific the nested path
for example, `lateInteractionScore(queryVectors, FieldPath.of("vectors", "embeddings"), doc, spaceType)`

navneet1v (approved):
Approving the PR. In the next later version we will add the lateInteraction field as a first class citizen

src/test/java/org/opensearch/knn/integ/LateInteractionScoreIT.java
navneet1v: Had a big discussion on this with @vamshin , @mingshl and @vigyasharma. The idea of using nested fields for these kind of multi-vector is not ideal since, multi-vectors generated from models like colbert and copali are dependent on one another. What a nested fields represent is multi-vector which are independent of one another. So in a way representing dependent multi-vector using nested fields is not a good idea.

Also, doing ANN search on these dependent multi-vector is computationally high, to go around this, either you fuse multi dependent vectors in 1 vector or use another model to generate a single vector. Along with this use dependent multi-vector(represented as a separate field) for doing re-ranking.

As next step, we will start exposing a lateInteraction field of Lucene via mappings and then integrate it with painless or expose it via a separate query clause.

PR #2911, opened 2025-09-29
Use right doc id for MMR.

bzhangam:
### Description
Use right doc id for MMR. We should use the unique doc id rather than the internal lucene doc id. That doc id is not always unique especially when we have multiple nodes. That id of the search hit can be -1 if the hit is passed from one node to another one.

### Related Issues
N/A

### Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 * Fix invalid cosine score range in LuceneOnFaiss [#2892](https://github.com/opensearch-project/k-NN/pull/2892)
 * Allows k to be nullable to fix filter bug [#2836](https://github.com/opensearch-project/k-NN/issues/2836)
 * Fix integer overflow for while estimating distance computations for efficient filtering [#2903](https://github.com/opensearch-project/k-NN/pull/2903)
+* Use the unique doc id for MMR rerank rather than internal lucenue doc id which is not unique for multiple shards case. [#2911](https://github.com/opensearch-project/k-NN/pull/2911)
 
 ### Refactoring
 * Refactored the KNN Stat files for better readability.
--- a/src/main/java/org/opensearch/knn/search/processor/mmr/MMRRerankProcessor.java
+++ b/src/main/java/org/opensearch/knn/search/processor/mmr/MMRRerankProcessor.java
@@ -6,6 +6,7 @@
 package org.opensearch.knn.search.processor.mmr;
 
 import lombok.AllArgsConstructor;
+import lombok.extern.log4j.Log4j2;
 import org.opensearch.action.search.SearchRequest;
 import org.opensearch.action.search.SearchResponse;
 import org.opensearch.action.search.SearchResponseSections;
@@ -41,6 +42,7 @@ import static org.opensearch.knn.search.processor.mmr.MMRUtil.shouldGenerateMMRP
 /**
  * A system generated search response processor that rerank the response based on the Maximal Marginal Relevance
  */
+@Log4j2
 @AllArgsConstructor
 public class MMRRerankProcessor implements SearchResponseProcessor, SystemGeneratedProcessor {
     public static final String TYPE = "mmr_rerank";
@@ -72,7 +74,7 @@ public class MMRRerankProcessor implements SearchResponseProcessor, SystemGenera
         final boolean isFloatVector = VectorDataType.FLOAT.equals(mmrContext.getVectorDataType());
 
         final List<SearchHit> candidates = new ArrayList<>(List.of(searchResponse.getHits().getHits()));
-        final Map<Integer, Object> docVectors = extractVectors(
+        final Map<String, Object> docVectors = extractVectors(
             candidates,
             mmrContext.getVectorFieldPath(),
             mmrContext.getIndexToVectorFieldPathMap(),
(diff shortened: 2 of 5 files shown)


heemin32 (dismissed):

src/main/java/org/opensearch/knn/search/processor/mmr/MMRRerankProcessor.java
navneet1v: why we are modifying the took time? when tookTime generally means the query phase latency

navneet1v, line 93: if we are apply this, why we need to get newSelectedScore, newHits etc. We could just return from here.

navneet1v: Also do we have IT for the case when _source is false, or user has removed a vector field from source

src/test/java/org/opensearch/knn/search/extension/MMRSearchExtBuilderIT.java
```diff
@@ -184,8 +186,16 @@ private void verifyResults(List<KNNResult> results, boolean excludeVector) {
             results.forEach(r -> assertNotNull("Vector should be included", r.getVector()));
         }
         assertEquals(QUERY_SIZE, results.size());
-        assertEquals("0", results.get(0).getDocId());
```
bzhangam, line 187: Since we have multiple shards now the first doc is no longer 0. So change to compare the vectors.

heemin32 (dismissed):

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 * Fix invalid cosine score range in LuceneOnFaiss [#2892](https://github.com/opensearch-project/k-NN/pull/2892)
 * Allows k to be nullable to fix filter bug [#2836](https://github.com/opensearch-project/k-NN/issues/2836)
 * Fix integer overflow for while estimating distance computations for efficient filtering [#2903](https://github.com/opensearch-project/k-NN/pull/2903)
+* Use right doc id for MMR.
```
navneet1v: Please add more details here. This changelog entry provides no information

src/main/java/org/opensearch/knn/search/processor/mmr/MMRRerankProcessor.java
bzhangam: I thought it means the overall time used for the search request. If it's just for query phase I think we can remove it.

bzhangam: This is just the hits and we cannot directly set it in the existing response instance. That's why we need to construct a new one.

We have unit tests and integ tests to cover the case when _source is disabled.

src/test/java/org/opensearch/knn/search/extension/MMRSearchExtBuilderIT.java
```diff
@@ -184,8 +186,16 @@ private void verifyResults(List<KNNResult> results, boolean excludeVector) {
...
+            assertArrayEquals(SIMILAR_VECTOR, results.get(0).getVector(), DELTA);
+        }
         assertEquals("Should pick up the hit with diversity.", "8", results.get(1).getDocId());
-        assertEquals("1", results.get(2).getDocId());
+        if (results.get(1).getVector() != null) {
+            assertArrayEquals("Should pick up the hit with diversity.", DIVERSE_VECTOR_1, results.get(1).getVector(), DELTA);
+        }
+        if (results.get(2).getVector() != null) {
+            assertArrayEquals(SIMILAR_VECTOR, results.get(2).getVector(), DELTA);
+        }
```
navneet1v, line 201: we should also validate the scores here. Just checking docId is not enough

src/main/java/org/opensearch/knn/search/processor/mmr/MMRRerankProcessor.java
navneet1v: Yes please check this and fix it. This can screw up a lot of metrics.

navneet1v: BTW I would suggest to log the time as a debug time so that it can be used later.

src/test/java/org/opensearch/knn/search/extension/MMRSearchExtBuilderIT.java
bzhangam: ok we can validate

CHANGELOG.md
bzhangam: ok

src/main/java/org/opensearch/knn/search/processor/mmr/MMRRerankProcessor.java
bzhangam: ok

src/main/java/org/opensearch/knn/search/processor/mmr/MMRRerankProcessor.java
```diff
@@ -41,6 +42,7 @@
 /**
  * A system generated search response processor that rerank the response based on the Maximal Marginal Relevance
  */
+@Slf4j
```
navneet1v: in k-NN we use @Log4j2, so please use that

bzhangam: ok

Issue #2917, opened 2025-10-01
[BUG] Block using memory optimized search for indices created before 2.19


0ctopus13prime:
**What is the bug?**
When an index was created before 2.17, and user upgrades the cluster to 3.1 and turn on memory optimized search, then NPE will be raised from `SegmentReader.getVectorReader` as old codec does not have the implementation. We added `getVectorReader` to return non-null `NativeEngines990KnnVectorsReader` at version 3.1. 

We should fix [this check](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/engine/MemoryOptimizedSearchSupportSpec.java#L45) to check the index created version and return `false` accordingly. 

**How can one reproduce the bug?**
1. Spin up 2.17 cluster
2. Configure Faiss knn field.
3. Upgrade the cluster to 3.1
4. Turn on memory optimized search.

**What is the expected behavior?**

**What is your host/environment?**
 - OS: Linux
 - Version : 3.1
 - Plugins : kNN

**Do you have any screenshots?**

**Do you have any additional context?**
Add any other context about the problem.

0ctopus13prime:
Official doc update :  https://github.com/opensearch-project/documentation-website/pull/11173

PR #2918, opened 2025-10-01
Do not apply memory optimized search for old indices.

0ctopus13prime:
### Description
For old indices created before 2.17, will use its old codec where returning null as VectorReader leading NPE issue when memory optimized search is enabled.
This fix PR is adding index created check to return `false` for isMemoryOptimizedSearchEnabled so that the flow can fallback to default logic which will then delegate C++ to load vectors into off-heap.

For indices created after 2.18+, will use the codec has LuceneOnFaiss implementation, will delegate it to search on Faiss index.

### Related Issues
https://github.com/opensearch-project/k-NN/issues/2917

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/engine/MemoryOptimizedSearchSupportSpec.java
+++ b/src/main/java/org/opensearch/knn/index/engine/MemoryOptimizedSearchSupportSpec.java
@@ -32,6 +32,7 @@ import static org.opensearch.knn.common.KNNConstants.METHOD_HNSW;
  * The overall logic will be made based on the given method context and quantization configuration.
  */
 public class MemoryOptimizedSearchSupportSpec {
+    private static final Version MIN_VERSION_SUPPORTS_MEM_OPT_SEARCH = Version.V_2_17_0;
     private static final Set<String> SUPPORTED_HNSW_ENCODING = Set.of(ENCODER_FLAT, ENCODER_SQ, ENCODER_BINARY);
 
     /**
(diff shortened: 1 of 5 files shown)


src/main/java/org/opensearch/knn/index/engine/MemoryOptimizedSearchSupportSpec.java
```diff
@@ -69,8 +70,14 @@ public static boolean isSupportedFieldType(final KNNVectorFieldType fieldType, f
     public static boolean isSupportedFieldType(
         final Optional<KNNMethodContext> methodContextOpt,
         final QuantizationConfig quantizationConfig,
-        final Optional<String> modelId
+        final Optional<String> modelId,
+        final Version indexCreatedVersion
     ) {
+        // Since LuceneOnFaiss logic sits in newer codec, we don't support LuceneOnFaiss for older codec whose version < 2.19
+        if (indexCreatedVersion.before(Version.V_2_19_0)) {
+            return false;
```
shatejas: should this throw an exception? The idea is that if user tries to turn on the setting for indices before 2.19 it shouldn't be allowed rather than silently falling back to default behavior

0ctopus13prime: Currently, there's no way to block user from enabling memory optimized search as in the event listener, we can't look up mapping from IndicesService. And the setting sits at index level, requires user to close, update and reopen it again. If we throw an exception, then all traffics will be hitting 400, then to recover, user should close the index again, and turn it off, and reopen again where loading all bytes will happen. At the worst case, if there's no sufficient memory space, then whole cluster can be ruined. 😓 

The best way is to block user from enabling it if the target index is not supported, but technically we can't do it right now.

shatejas: @navneet1v Thoughts on throwing vs silently falling back to graph loading?

I am okay with silently falling back for now, please add a debug log and update the documentation for memory optimized search

Vikasht34: There were some setting we depreacated in 3.0 and when I was trying in 3.1 , it was throwinng as undefined setting !! So if we have launched memoryoptimzed from 2.19 , would user able to set the setting since there is no updated code which has memoryoptimzed?

Vikasht34 (approved):
Looks Good to me !!

src/main/java/org/opensearch/knn/index/engine/MemoryOptimizedSearchSupportSpec.java
navneet1v: @0ctopus13prime I think the better experience here will be to throw the exception when the setting is getting set at the index. But since we cannot do this, as an application if somehow user has set this setting on old indices we should throw exception during query to ensure that user know they are doing something incorrect, rather than silently falling back.

@0ctopus13prime another thing, what is the reason for enabling this setting for indices >=2.19 ? it should be 2.17 because from 2.17 we started using KNNVectorField of Lucene.

0ctopus13prime: @navneet1v 
But isn't it better to fallback to default rather than throwing an error?
Once it's configured, then all traffic would end up getting 400 if we throw an exception there. 🤔 
Would it be better if we could give a chance to user to pick window for closing index and reopen it?

navneet1v: > But isn't it better to fallback to default rather than throwing an error?

No its not better to fallback since it gives a wrong impression to users and operators that the feature is working. If we can throw exception during turning on of this setting that will be great. But looks like it is not possible. So atleast on search path we should throw exceptions.

0ctopus13prime: sure, sounds good! Will change the code

navneet1v (approved):

shatejas (approved):

PR #2921, opened 2025-10-04
Fix rescoring logic for nested exact search

### LINKED ISSUE #2895: [BUG] Nested exact search does not return the best child score.
**What is the bug?**
Nested exact search does not return the highest child score. While the result set itself is correct and contains all the expected document IDs, the score values are not reflecting the best child score.

Apply 8x quantization with inner product, and ingest documents and get the results.
Recall is not impacted, just that the score value is not correct.

Score values should be the highest child score.

Opensearch : 3.1+

N/A

Root cause analysis 
1. In exact search, NestedVectorIdsKNNIterator sets the last score to `currentScore` - [Code](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/query/iterators/NestedVectorIdsKNNIterator.java#L82)
2. `currentScore` then returned and put into ScoreDoc - [Code](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/query/iterators/VectorIdsKNNIterator.java#L89)
3. But `currentScore` is not the best child score, it just have the last child score. - [Code](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/query/iterators/NestedVectorIdsKNNIterator.java#L82)

Therefore, to fix this, we need to track of the best score in NestedVectorIdsKNNIterator then return it in `score()`.

neetikasinghal:
### Description
Details of the issue are described in https://github.com/opensearch-project/k-NN/issues/2895

### Related Issues
Resolves #2895
[](https://github.com/opensearch-project/k-NN/issues/2895)

### Check List
~~- [ ] New functionality includes testing.~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 * Fix byte[] radial search for faiss [#2905](https://github.com/opensearch-project/k-NN/pull/2905)
 * Use the unique doc id for MMR rerank rather than internal lucenue doc id which is not unique for multiple shards case. [#2911](https://github.com/opensearch-project/k-NN/pull/2911)
 * Fix local ref leak in JNI [#2916](https://github.com/opensearch-project/k-NN/pull/2916)
+* Fix rescoring logic for nested exact search [#2921](https://github.com/opensearch-project/k-NN/pull/2921)
 
 ### Refactoring
 * Refactored the KNN Stat files for better readability.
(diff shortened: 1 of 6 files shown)


neetikasinghal:
> Looks Good to me !! , Would u able to put the testing details.

All the tests extending AbstractMemoryOptimizedKnnSearchIT and having disk based vector search will check and assert if the best child score is returned here: https://github.com/opensearch-project/k-NN/blob/main/src/testFixtures/java/org/opensearch/knn/generate/Documents.java#L138-L158

Vikasht34 (approved):

PR #2922, opened 2025-10-06
Native scoring for FP16 V1 implementation. 

0ctopus13prime:
### Description
RFC : https://github.com/opensearch-project/k-NN/issues/2875

This PR introduces a new build structure in k-NN for SIMD-based computation, implementing the V1 design described in the RFC linked above.
V1 relies on Faiss distance calculation functions for similarity scoring. The structure introduced here is designed to make it easy to add bulk SIMD operations in subsequent versions. Starting with V1 helps reduce the initial review complexity for maintainers.

The core concept is to extract mapped memory pointers from MemorySegmentIndexInput and leverage native SIMD acceleration to enhance search performance.

### Related Issues
[](https://github.com/opensearch-project/k-NN/issues/2875)

### Check List


0ctopus13prime:
## Performance improvement summarization
With early termination + Native scoring for FP16,
I'm seeing 70% QPS improvement compared to when using Faiss C++ for multi segments scenario with Cohere-10M.
For the single segment case, this is showing slightly less QPS than Faiss C++ which I suspect it's due to the JNI call + Reflection logic extracting MemorySegment[] overhead.

jni/src/simd/similarity_function/x86_avx2_similarity_function.cpp
```diff
@@ -0,0 +1,562 @@
+#include <algorithm>
```
0ctopus13prime: oops, mistakenly included V2. 
This will be removed in the next revision.

jni/src/simd/similarity_function/x86_avx512_similarity_function.cpp
```diff
@@ -0,0 +1,538 @@
+#include <algorithm>
```
0ctopus13prime: oops, mistakenly included V2. 
This will be removed in the next revision.

jni/src/simd/similarity_function/arm_simd_similarity_function.cpp
```diff
@@ -0,0 +1,690 @@
+#include <algorithm>
```
0ctopus13prime: oops, mistakenly included V2. 
This will be removed in the next revision.

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldType.java
```diff
@@ -65,6 +67,7 @@ public KNNVectorFieldType(String name, Map<String, String> metadata, VectorDataT
             annConfig.getQuantizationConfig(),
             annConfig.getModelId()
         );
+        log.info("Field [{}] memory_optimized_search_enabled={}", name, memoryOptimizedSearchAvailable);
```
navneet1v, line 70: should this be a debug log? Also, we should see how we can add this information in our explain API?

src/main/java/org/opensearch/knn/indices/ModelDao.java
```diff
@@ -187,7 +187,7 @@ final class OpenSearchKNNModelDao implements ModelDao {
         private int numberOfShards;
         private int numberOfReplicas;
 
-        private static OpenSearchKNNModelDao INSTANCE;
+        private static volatile OpenSearchKNNModelDao INSTANCE;
```
navneet1v, line 190: any reason for making this variable volatile?

src/main/java/org/opensearch/knn/index/query/memoryoptsearch/MemoryOptimizedKNNWeight.java
```diff
@@ -48,14 +50,43 @@ public class MemoryOptimizedKNNWeight extends KNNWeight {
 
     private final KnnCollectorManager knnCollectorManager;
 
+    // Ported from Lucene as since 10.3.0, TopKnnCollectorManager does not use MultiLeafKnnCollector no more.
+    private static class MultiLeafTopKnnCollectorManager implements KnnCollectorManager {
```
0ctopus13prime, line 54: This will not be put into main, it's for benchmarking.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/WrappedFloatVectorValues.java
```diff
@@ -0,0 +1,49 @@
...
+     * any nested {@link WrappedFloatVectorValues} layers until it reaches the base
+     * {@link FloatVectorValues}.
+     *
+     * @param knnVectorValues the {@link KnnVectorValues} to unwrap
+     * @return the innermost {@link FloatVectorValues}, or {@code null} if not applicable
+     */
+    public static FloatVectorValues getBottomFloatVectorValues(KnnVectorValues knnVectorValues) {
+        if (knnVectorValues instanceof FloatVectorValues floatVectorValues) {
+            while (floatVectorValues instanceof WrappedFloatVectorValues wrappedFloatVectorValues) {
+                floatVectorValues = wrappedFloatVectorValues.nestedVectorValues;
```
shatejas: Just curious, are we expecting multiple layers of this? From what I understand these will always be flat values from faiss graph. Do we really need a while loop here? maybe just an `assert !(wrappedFloatVectorValues.nestedVectorValues instanceof  WrappedFloatVectorValues` is sufficient before returning

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/WrappedFloatVectorValues.java
```diff
@@ -0,0 +1,49 @@
...
+ * <p>
+ * This class is typically used when multiple wrappers are stacked around a base
+ * {@link FloatVectorValues}, and utility methods are needed to access the innermost
+ * (bottom-level) implementation.
+ */
+@RequiredArgsConstructor
+public abstract class WrappedFloatVectorValues extends FloatVectorValues {
+
+    // The wrapped (nested) {@link FloatVectorValues} instance.
+    protected final FloatVectorValues nestedVectorValues;
```
shatejas: The prefix `nested` can be confusing considering knn supported nested index mapping. I understand this is in terms of how faiss stores index in nested fashion, but maybe flatVectorValues or something else to avoid the confusion?

src/main/java/org/opensearch/knn/memoryoptsearch/MemorySegmentAddressExtractorJDK21.java
```diff
@@ -74,23 +74,58 @@ public long[] extractAddressAndSize(final IndexInput indexInput) {
...
+                    // If this chunk contains `baseOffset`, then force its address value to be `baseOffset`
+                    if (startOffset < baseOffset) {
+                        chunkSize = endOffsetExclusive - baseOffset;
+                        address += baseOffset - startOffset;
+                    }
+
+                    // Collect only chunk that overlap with baseOffset or placed after baseOffset
+                    addressAndSize[addressIndex] = address;
+                    addressIndex += 2;
+                    addressAndSize[sizeIndex] = chunkSize;
```
shatejas: Just checking if we need to manipulate the final chunkSize based on size of faissSection? You only need access to flat vectors right? this will give access till the end of the faiss file I think. Let me know if I am missing something here

src/main/java/org/opensearch/knn/memoryoptsearch/MemorySegmentAddressExtractorJDK21.java
```diff
@@ -55,7 +55,7 @@ public final class MemorySegmentAddressExtractorJDK21 implements MemorySegmentAd
      *         ex: address_i = array[i], size_i = array[i + 1]
      */
     @Override
-    public long[] extractAddressAndSize(final IndexInput indexInput) {
+    public long[] extractAddressAndSize(final IndexInput indexInput, long baseOffset) {
```
shatejas: So the responsibility of this method was to extract addressAndSize, Shouldn't the manipulation of base and sectionSize happen in MMap*VectorValues?

I think this still works considering addressAndSize is not a member variable, just want to make sure if like baseOffset, we need the total size as parameter in this method

src/main/java/org/opensearch/knn/memoryoptsearch/MemorySegmentAddressExtractorJDK22.java
```diff
@@ -48,21 +48,49 @@ public long[] extractAddressAndSize(final IndexInput indexInput) {
...
+                }
+                startOffset += originalChunkSize;
             }
+
+            if (addressIndex != addressAndSize.length) {
+                // There was a chunk that excluded, shrink it
+                long[] newAddressAndSize = new long[addressIndex];
+                System.arraycopy(addressAndSize, 0, newAddressAndSize, 0, addressIndex);
+                return newAddressAndSize;
+            }
```
shatejas: nit: can we extract this into a common method and avoid duplicates

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissIndexScalarQuantizedFlat.java
```diff
@@ -127,6 +127,28 @@ public FloatVectorValuesImpl copy() {
...
+                    indexInput,
+                    oneVectorByteSize,
+                    flatVectors.getBaseOffset(),
+                    dimension,
+                    totalNumberOfVectors,
+                    addressAndSize,
+                    reconstructor
+                );
+            } else {
+                log.debug("Failed to extract mapped pointers from IndexInput, falling back to ByteVectorValuesImpl.");
```
shatejas: nit: `FloatVectorValuesImpl`

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/WrappedFloatVectorValues.java
0ctopus13prime: No, we are not expecting it to be multi-layers. I just added a general way to handle wrapper.
Will get rid of loop in the next rev.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/WrappedFloatVectorValues.java
0ctopus13prime: Sure, I will just use `flatVectorValues`.
Will update in the next rev.

src/main/java/org/opensearch/knn/memoryoptsearch/MemorySegmentAddressExtractorJDK21.java
0ctopus13prime: yes, you're right, you're not missing anything. What we need is exactly the flat vector section. I think the starting offset is already addressed, but the end offset of the section, as you said, the size of file is being used. 
Do you think would it be better to make it explicitly requires startOffset and size?

src/main/java/org/opensearch/knn/memoryoptsearch/MemorySegmentAddressExtractorJDK21.java
0ctopus13prime: MMap*VectorValues is a general VectorValues that returns addressAndSize, then it's FaissIndex's (ex: FaissIndexScalarQuantizedFlat) responsibility to slice necessary parts from it and give it to scorer. Since each FaissIndex implementation knows the flat vector section size in Faiss index, I divided responsibility into 2 parts: 
1. Let MMap*VectorValues return whole addressAndSize for underlying file.
2. Each FaissIndex implementation then slice it based on flat vector FaissSection.

Will add `size` parameter to make it return exactly the region it needs for scoring.

src/main/java/org/opensearch/knn/memoryoptsearch/MemorySegmentAddressExtractorJDK22.java
0ctopus13prime: Sure, will do in next rev.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissIndexScalarQuantizedFlat.java
0ctopus13prime: will update in the next rev.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/NativeRandomVectorScorer.java
```diff
@@ -0,0 +1,117 @@
...
+        final float[] query,
+        final KnnVectorValues knnVectorValues,
+        final MMapVectorValues mmapVectorValues,
+        final SimdVectorComputeService.SimilarityFunctionType similarityFunctionType
+    ) {
+        this.knnVectorValues = knnVectorValues;
+        this.addressAndSize = mmapVectorValues.getAddressAndSize();
+        this.maxOrd = knnVectorValues.size();
+        this.nativeFunctionTypeOrd = similarityFunctionType.ordinal();
+        SimdVectorComputeService.saveSearchContext(query, addressAndSize, nativeFunctionTypeOrd);
```
shatejas, line 59: I am wondering if the use of thread local should be minimized to `queryVectorSimdAligned`. Everything other parameter in search context can be local to the thread. This increases maintaibility of the code while keeping the optimization. let me know what you think

src/main/java/org/opensearch/knn/memoryoptsearch/MemorySegmentAddressExtractorJDK21.java
shatejas: > Do you think would it be better to make it explicitly requires startOffset and size?

Yes, I know its being managed correctly but best to give only access to the required section

jni/src/simd/similarity_function/faiss_score_to_lucene_transform.cpp
```diff
@@ -0,0 +1,76 @@
...
+    static float ipToMaxIpTransform(const float innerProductValue) noexcept {
+        return innerProductValue < 0 ? 1 / (1 - innerProductValue) : (1 + innerProductValue);
+    }
+
+    // Convert Faiss inner product values to Max Inner Product scheme whose range is in [0, +Inf)
+    // Initially, `scores` have Faiss inner product values, after this transform, it will have Max IP value.
+    static void ipToMaxIpTransformBulk(float* scores, const int32_t numScores) noexcept {
+        int32_t i = 0;
+        for (; (i + 8) <= numScores ; i += 8, scores += 8) {
+            scores[0] = scores[0] < 0 ? 1 / (1 - scores[0]) : (1 + scores[0]);
```
shatejas, line 16: just trying to figure out the intent here, why not loop through and call ipToMaxIpTransform instead doing 8 together? will this be replaced by SIMD intrinsics in the same file?

src/main/java/org/opensearch/knn/memoryoptsearch/MemorySegmentAddressExtractorJDK21.java
0ctopus13prime: sure, will update in the next rev.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/NativeRandomVectorScorer.java
0ctopus13prime: Sure, sounds good.
But I think it's better to save a list of address so the pointer of vector calculation can be done quickly.
Other than that, I think mostly are native types, we can pass them as parameters. 
Will update in the next rev.

jni/src/simd/similarity_function/faiss_score_to_lucene_transform.cpp
0ctopus13prime: Just typical loop unrolling optimization likewise Lucene is doing in [here](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/util/packed/BulkOperationPacked20.java#L29) 😛- https://en.wikipedia.org/wiki/Loop_unrolling
Not only the SIMD, but it can take advantage of CPU instruction pipeline. 
Did some simple tests, and the loop unrolling version showed 30% - 50% faster performance than simple looping.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/NativeRandomVectorScorer.java
0ctopus13prime: Hi @shatejas,
I tried to reflect your comment, but I think we need to save the similarity function in context to create and keep Faiss distance calculator. Otherwise, we need to create the distance calculator each time whenever nodes visited.

jni/include/platform_defs.h
```diff
@@ -0,0 +1,13 @@
...
+    #define LIKELY(x)   (__builtin_expect(!!(x), 1))
+    #define UNLIKELY(x) (__builtin_expect(!!(x), 0))
+#elif defined(_MSC_VER)
+    // MSVC doesn't have __builtin_expect; just pass through
+    #define LIKELY(x)   (x)
+    #define UNLIKELY(x) (x)
+#else
+    #define LIKELY(x)   (x)
+    #define UNLIKELY(x) (x)
+#endif
```
navneet1v, line 13: can you please add proper docs on what are these macros doing.

jni/src/simd/similarity_function/default_simd_similarity_function.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+#include "simd_similarity_function_common.cpp"
+#include "faiss_score_to_lucene_transform.cpp"
+#include "parameter_utils.h"
+
+//
+// FP16
+//
+
+using BulkScoreTransform = void (*)(float*/*scores*/, int32_t/*num scores to transform*/);
```
navneet1v, line 15:
```suggestion
// scores and num scores to transform*
using BulkScoreTransform = void (*)(float*, int32_t);
```

jni/src/simd/similarity_function/default_simd_similarity_function.cpp
```diff
@@ -0,0 +1,74 @@
...
+using ScoreTransform = float (*)(float/*score*/);
```
navneet1v, line 16: same as above

jni/src/simd/similarity_function/faiss_score_to_lucene_transform.cpp
```diff
@@ -0,0 +1,76 @@
+#include <cmath>
+#include <cstdint>
+
+// This class is responsible to convert Faiss distance value to Lucene similarity score.
+struct FaissScoreToLuceneScoreTransform final {
```
navneet1v, line 5: why cannot we move this code to java? is there a reason we need to keep these translations in C++ side? Reason being we already have multiple translations of scores on java side. Also, once we are in java we can actually use Lucene VectorUtil class to covert IP to MaxIP.

https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/util/VectorUtil.java#L367-L376

Also, do we really need this conversion from IP to MaxIP? do we not need to convert back to IP and then do faiss based score translation to keep scores consistent with non-memory optimized search.

jni/src/simd/similarity_function/faiss_score_to_lucene_transform.cpp
```diff
@@ -0,0 +1,76 @@
...
+            ++i;
+            ++scores;
+        }
+    }
+
+    // Transform Faiss L2 distance to be bounded (0, 1]
+    static float l2Transform(float l2Distance) noexcept {
+        return 1.0f / (1.0f + l2Distance);
+    }
+
```
navneet1v, line 44: same as above

jni/src/simd/similarity_function/simd_similarity_function_common.cpp
```diff
@@ -0,0 +1,269 @@
...
+}
+
+SimdVectorSearchContext::~SimdVectorSearchContext() {
+    if (queryVectorSimdAligned) {
+        free_aligned_memory(queryVectorSimdAligned);
+    }
+}
+
+// Thread static local SimdVectorSearchContext
+thread_local SimdVectorSearchContext THREAD_LOCAL_SIMD_VEC_SRCH_CTX {};
```
navneet1v, line 160: what is the benefit of storing the SimdVectorSearchContext in thread local here?

and if I am not wrong, this assumes that every java search thread is getting mapped to a new C++ thread right?

jni/src/simd/similarity_function/simd_similarity_function_common.cpp
```diff
@@ -0,0 +1,269 @@
...
+           int32_t queryByteSize,
+           int32_t dimension,
+           int64_t* mmapAddressAndSize,
+           int32_t numAddressAndSize,
+           int32_t nativeFunctionTypeOrd) {
+    // Free tmp buffer
+    THREAD_LOCAL_SIMD_VEC_SRCH_CTX.tmpBuffer = {};
+
+    // Allocate query vector space
+    if (THREAD_LOCAL_SIMD_VEC_SRCH_CTX.queryVectorByteSize < queryByteSize) {
```
navneet1v, line 178: what would be the case in which this condition will not be true?

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldType.java
0ctopus13prime: I think this should be at info level, the message format should be revised though.
I think adding information in explain API is another scope, can be dealt with in a separate PR

src/main/java/org/opensearch/knn/indices/ModelDao.java
0ctopus13prime: It's accessed by multiple threads, and in that case, we should add volatile to make sure each thread can fetch the latest reference.

jni/include/platform_defs.h
0ctopus13prime: sure, will do

jni/src/simd/similarity_function/default_simd_similarity_function.cpp
0ctopus13prime: Hm.. I think removing variable name can give user confusions on what each parameter is for.
Will keep variable name, and add the comment.

jni/src/simd/similarity_function/default_simd_similarity_function.cpp
0ctopus13prime: I think it's better to keep the variable names. Will add the comment.

jni/src/simd/similarity_function/faiss_score_to_lucene_transform.cpp
0ctopus13prime: This is bulk conversion optimization in C++ side which is showing better performance, it's not about using which API. 

And we do need conversion from IP to MaxIP. One of the reason is for radial search. We are converting radius to MaxIP score for radial search when memory optimized search is enabled.

jni/src/simd/similarity_function/faiss_score_to_lucene_transform.cpp
0ctopus13prime: Addressed the comment.

jni/src/simd/similarity_function/simd_similarity_function_common.cpp
0ctopus13prime: Java thread is itself a native thread, there's no distinction between C++ thread and Java thread. They are just plain POSIX thread.

Having the thread local is prevent frequent memory allocation and reuse memory space as possible. From the benchmark, I found memory allocation was the bottleneck. (We can call JNI function to get raw pointer every time from `float[]`, but it drags down performance) So, having the thread local to keep the memory space to minimize the allocations possible.

jni/src/simd/similarity_function/simd_similarity_function_common.cpp
0ctopus13prime: Imagine a scenario where one field has 768 dimensions, and another field can have 1024 dimensions. 
In that case, a query for the first field will allocate 768 * 4 bytes initially, and if it's the next query is targeting for the second query, then the internal space needs to be grow. In that case, the condition will be evaluated as false.

jni/cmake/init-simd.cmake
```diff
@@ -0,0 +1,138 @@
...
+#
+
+include(CheckCXXSourceCompiles)
+
+# Allow user overrides
+if(NOT DEFINED AVX2_ENABLED)
+    set(AVX2_ENABLED true)   # set default value as true if the argument is not set
+endif()
+
+if(NOT DEFINED AVX512_ENABLED)
```
Vikasht34: Why same statement twice? 

```
if(NOT DEFINED AVX2_ENABLED)
    set(AVX2_ENABLED true)   # set default value as true if the argument is not set
endif()
```

jni/cmake/init-simd.cmake
```diff
@@ -0,0 +1,138 @@
...
+    set(AVX512_ENABLED true)   # set default value as true if the argument is not set
+endif()
+
+if(NOT DEFINED AVX512_SPR_ENABLED)
+    # Check if the system is Intel(R) Sapphire Rapids or a newer-generation processor
+    execute_process(COMMAND bash -c "lscpu | grep -q 'GenuineIntel' && lscpu | grep -i 'avx512_fp16' | grep -i 'avx512_bf16' | grep -i 'avx512_vpopcntdq'" OUTPUT_VARIABLE SPR_FLAGS OUTPUT_STRIP_TRAILING_WHITESPACE)
```
Vikasht34: is this what we wanted to do here?

```
if(NOT DEFINED AVX512_SPR_ENABLED)
    set(AVX512_SPR_ENABLED false)
    
    if(AVX512_ENABLED AND ${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
        find_program(LSCPU_PROGRAM lscpu)
        if(LSCPU_PROGRAM)
            execute_process(
                COMMAND ${LSCPU_PROGRAM}
                OUTPUT_VARIABLE CPU_INFO
                ERROR_QUIET
                OUTPUT_STRIP_TRAILING_WHITESPACE
            )
            if(CPU_INFO MATCHES "GenuineIntel" AND 
               CPU_INFO MATCHES "avx512_fp16" AND 
               CPU_INFO MATCHES "avx512_bf16" AND 
               CPU_INFO MATCHES "avx512_vpopcntdq")
                set(AVX512_SPR_ENABLED true)
            endif()
        endif()
    endif()
endif()
```

Vikasht34: Did you mean RESULT_VARIBALE?

jni/cmake/init-simd.cmake
```diff
@@ -0,0 +1,138 @@
+#
+# Copyright OpenSearch Contributors
+# SPDX-License-Identifier: Apache-2.0
+#
+
+include(CheckCXXSourceCompiles)
+
+# Allow user overrides
+if(NOT DEFINED AVX2_ENABLED)
```
Vikasht34, line 9: So we are setting default AVX2, AVX512 and SPR to true, Should it to default false?

jni/cmake/init-simd.cmake
Vikasht34: Should we gauard "COMMAND bash -c "lscpu " WITH below if ?

if(CMAKE_HOST_SYSTEM_NAME STREQUAL "Linux" and CMAKE_CROSS_COMPILINGH)

jni/cmake/init-simd.cmake
```diff
@@ -0,0 +1,138 @@
...
+
+# Default SIMD state
+set(KNN_HAVE_AVX2_F16C OFF)
+set(KNN_HAVE_AVX512 OFF)
+set(KNN_HAVE_AVX512_SPR OFF)
+set(KNN_HAVE_ARM_FP16 OFF)
+set(SIMD_OPT_LEVEL "")
+set(SIMD_FLAGS "")
+
+if(${CMAKE_SYSTEM_NAME} STREQUAL "Windows" OR (NOT AVX2_ENABLED AND NOT AVX512_ENABLED AND NOT AVX512_SPR_ENABLED))
```
Vikasht34, line 43: MSVC now supports AVX2 AND 512 , do we plan to enable for that?

jni/include/simd/similarity_function/similarity_function.h
```diff
@@ -0,0 +1,86 @@
...
+    };
+
+    // This class's responsibility is to calculate similarity between query and vectors.
+    // It first saves search context first in static thread local object, then use it during search.
+    struct SimilarityFunction {
+        virtual ~SimilarityFunction() = default;
+
+        // Save required information during search in static thread local storage.
+        // The maximum thread local storage is bounded by O(SizeOf(Query Vector Size)).
+        static SimdVectorSearchContext* saveSearchContext(
```
Vikasht34, line 60: That's great that we are promising static thread local object but if savesearchcontext stores state in thread_local singleton , can we be explict in the .cc and define lifetime?

jni/include/simd/similarity_function/similarity_function.h
```diff
@@ -0,0 +1,86 @@
...
+                                               int32_t* internalVectorIds,
+                                               float* scores,
+                                               int32_t numVectors) = 0;
+
+        // Similar to `calculateSimilarityInBulk`, but this targets a single vector and returns a score.
+        virtual float calculateSimilarity(SimdVectorSearchContext* srchContext, int32_t internalVectorId) = 0;
+
+      private:
+        // Select similarity function based on the function type.
+        static SimilarityFunction* selectSimilarityFunction(NativeSimilarityFunctionType nativeFunctionType);
```
Vikasht34, line 82: Seems this is non-owning pointer with no explict lifetime? if it's per type can we return to static and document it , and if it's per query can we use std:unique_ptr?

jni/include/simd/similarity_function/similarity_function.h
```diff
@@ -0,0 +1,86 @@
...
+    struct SimilarityFunction {
+        virtual ~SimilarityFunction() = default;
+
+        // Save required information during search in static thread local storage.
+        // The maximum thread local storage is bounded by O(SizeOf(Query Vector Size)).
+        static SimdVectorSearchContext* saveSearchContext(
+                   uint8_t* queryPtr,
+                   int32_t queryByteSize,
+                   int32_t dimension,
+                   int64_t* mmapAddressAndSize,
```
Vikasht34, line 64: can we use POD Struct for mmapAddressAndSize?

jni/include/simd/similarity_function/similarity_function.h
```diff
@@ -0,0 +1,86 @@
...
+                   int32_t numAddressAndSize,
```
Vikasht34, line 65: Right now we are passing address through an integral type - will it hit UB on 32bit?

jni/src/simd/similarity_function/default_simd_similarity_function.cpp
```diff
@@ -0,0 +1,74 @@
...
+            func, "Unexpected distance function acquired. Expected SQDistanceComputer, but it was something else");
+
+        for (int32_t i = 0 ; i < numVectors ; ++i) {
+            // Calculate distance
+            auto vector = reinterpret_cast<uint8_t*>(srchContext->getVectorPointer(internalVectorIds[i]));
+            scores[i] = func->query_to_code(vector);
+        }
+
+        // Transform score values if it needs to
+        return BulkScoreTransformFunc(scores, numVectors);
```
Vikasht34: Why we are returning with void?

jni/src/simd/similarity_function/default_simd_similarity_function.cpp
```diff
@@ -0,0 +1,74 @@
+#include <algorithm>
+#include <array>
+#include <cstddef>
+#include <stdint.h>
+#include <cmath>
+
+#include "simd_similarity_function_common.cpp"
```
Vikasht34, line 7: Since we are pulling cpp file into another CPP files , Will it create build chrun? Can we split them into header?

jni/src/simd/similarity_function/default_simd_similarity_function.cpp
```diff
@@ -0,0 +1,74 @@
...
+        // Transform score value if it needs to
+        return ScoreTransformFunc(score);
+    }
+};
+
+//
+// FP16
+//
+// 1. Max IP
+DefaultFP16SimilarityFunction<FaissScoreToLuceneScoreTransform::ipToMaxIpTransformBulk, FaissScoreToLuceneScoreTransform::ipToMaxIpTransform> DEFAULT_FP16_MAX_INNER_PRODUCT_SIMIL_FUNC;
```
Vikasht34, line 60: Should this be static and other one also?

jni/src/simd/similarity_function/default_simd_similarity_function.cpp
```diff
@@ -0,0 +1,74 @@
...
+            scores[i] = func->query_to_code(vector);
+        }
+
+        // Transform score values if it needs to
+        return BulkScoreTransformFunc(scores, numVectors);
+    }
+
+    float calculateSimilarity(SimdVectorSearchContext* srchContext, const int32_t internalVectorId) final {
+        // Prepare distance calculation
+        auto vector = reinterpret_cast<uint8_t*>(srchContext->getVectorPointer(internalVectorId));
```
Vikasht34, line 42: Can we add null check if vector is null for InternalVectorId?

jni/src/simd/similarity_function/default_simd_similarity_function.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+template <BulkScoreTransform BulkScoreTransformFunc, ScoreTransform ScoreTransformFunc>
+struct DefaultFP16SimilarityFunction final : SimilarityFunction {
+    void calculateSimilarityInBulk(SimdVectorSearchContext* srchContext,
+                                   int32_t* internalVectorIds,
+                                   float* scores,
+                                   const int32_t numVectors) final {
+
+        // Prepare similarity calculation
+        auto func = dynamic_cast<faiss::ScalarQuantizer::SQDistanceComputer*>(srchContext->faissFunction.get());
```
Vikasht34, line 26: Since we are doinng this dynamic_cast here as well as in below function as well , this would be slow and unnecessary if context always carries SQDistanceComputer, either can we store a typed pointer in the context once or check one at entry and then static cast inside?

jni/cmake/init-simd.cmake
0ctopus13prime: This is following the current structure in CMakeLists.txt!
Idea is that setting each flag as True, unless user explicitly disable it.

jni/cmake/init-simd.cmake
0ctopus13prime: yeah, I think this is better! thank you :)

0ctopus13prime: No, we want stdoutput in here, not return code.

0ctopus13prime: I think FIND_PROGRAM will work as you suggested above!
Will update it to use FIND_PROGRAM

jni/cmake/init-simd.cmake
0ctopus13prime: Em.. not sure I want to  spend time on Windows.. 
I think we can use the default settings in Windows 😅

jni/include/simd/similarity_function/similarity_function.h
0ctopus13prime: will update in the next rev.

jni/include/simd/similarity_function/similarity_function.h
0ctopus13prime: I think I'm already doing that in [here](https://github.com/opensearch-project/k-NN/blob/8bead91579cbd2c319643deb2a2d8ae0903d6bb8/jni/src/simd/similarity_function/simd_similarity_function_common.cpp)!

In there, I've explicitly defined it in cpp file

jni/include/simd/similarity_function/similarity_function.h
0ctopus13prime: This is the size of `int64_t* mmapAddressAndSize`, which at most 10.

jni/include/simd/similarity_function/similarity_function.h
0ctopus13prime: `SimilarityFunction*` is a stateless static function pointer whose lifetime is static.

jni/src/simd/similarity_function/default_simd_similarity_function.cpp
0ctopus13prime: But the issue is that, we won't know whether context always carries it or not. To check actual type it's pointing to, we need at least one time dynamic cast anyway.

jni/src/simd/similarity_function/default_simd_similarity_function.cpp
0ctopus13prime: Good catch! will fix it in next rev.

jni/src/simd/similarity_function/default_simd_similarity_function.cpp
0ctopus13prime: sure, will update in the next rev.

jni/src/simd/similarity_function/default_simd_similarity_function.cpp
0ctopus13prime: By default, it's static scope for all variable defined in `.cpp` file.

jni/src/simd/similarity_function/default_simd_similarity_function.cpp
0ctopus13prime: No it won't! Since #include is just merely copy things in, it will be fine.

jni/include/simd/similarity_function/similarity_function.h
0ctopus13prime: Em.. but on the second thought, void* pointer and int64_t size will be kept in separate vectors anyway, can we keep the structure?

Issue #2852, opened 2025-08-14
[BUG] _routing missing when knn true on index settings


dan-acsc:
**What is the bug?**
`_routing` isn't returned on hits when an index has knn: true set.

This is a big problem if an index has _routing required set to true.

This results in a number of issues including delete_by_query failing because it lacks the _routing field to perform deletions with.

**How can one reproduce the bug?**
Example script to reproduces the issue.

Simple add credentials and toggle SUCCEED between True and False.

The script creates a simple index with a knn property.
In then inserts a document with the knn property (if succeed = false)
Or it inserts a document with no knn property (if succeed = false)

It then attempts to query for the document and get the _routing field no the hit.

It then attempts t query by delete (this fails due to the lack of the _routing field)

```python
# USERNAME = "cluster-username"
# PASSWORD = "cluster-password"

import random
from opensearchpy import OpenSearch

# Toggle to switch between indexing knn data and not indexing any knn data to see the difference.
SUCCEED = True

# Connection settings
host = "https://localhost:9204"
port = 9200
auth = (USERNAME, PASSWORD)  # Basic Auth

# Create OpenSearch client
client = OpenSearch(hosts=host, http_auth=auth, verify_certs=False, ssl_show_warn=False)

# Pick index name depending if you index any documents with knn or not
if SUCCEED:
    index_name = "non-knn-index"
else:
    index_name = "knn-index"

# Create index with mapping (knn in both is fine)
index_body = {
    "settings": {"index": {"knn": True}},
    # Using knn.derived_source enabled false bypasses the issue
    # "settings": {"index": {"knn": True, "knn.derived_source": {"enabled": False}}},
    "mappings": {
        # Routing required
        "_routing": {"required": True},
        "properties": {
            # Random keyword used as an id and for routing
            "random_id": {"type": "keyword"},
            # Knn data.
            "my_vector": {
                "type": "knn_vector",
                "dimension": 4,
                "data_type": "byte",
                "space_type": "cosinesimil",
                "method": {"name": "hnsw", "engine": "lucene"},
            },
            "description": {"type": "text"},
        },
    },
}


response = client.indices.create(index=index_name, body=index_body, ignore=400)
print("Index creation:", response)

# Step 2: Insert one example document
random_extra = ""
for i in range(10):
    random_extra += random.choice("abcdefghjiklmnopqrstuvwxyz")
id = random_extra + "custom_id"

if SUCCEED:  # Succeeds because vector data is not indexed.
[...]

PR #2925, opened 2025-10-07
Update Visitor to delegate for other fields

VijayanB:
### Description
When creating a custom Visitor for Derived Source, the implementation needed to invoke the delegate's methods (such as 'stringField', 'intField', 'longField') rather than calling the superclass methods, which were empty implementations.

Metadata fields like routing rely on the stringField method to transfer information from documents to SearchHit objects. Due to the previous implementation, the 'routing' field was missing from search responses because the methods weren't properly delegated to FieldVisitor's implementations.

### Related Issues
#2852 

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/codec/derivedsource/DerivedSourceStoredFieldVisitor.java
+++ b/src/main/java/org/opensearch/knn/index/codec/derivedsource/DerivedSourceStoredFieldVisitor.java
@@ -36,4 +36,29 @@ public class DerivedSourceStoredFieldVisitor extends StoredFieldVisitor {
     public Status needsField(FieldInfo fieldInfo) throws IOException {
         return delegate.needsField(fieldInfo);
     }
+
+    @Override
+    public void stringField(FieldInfo fieldInfo, String value) throws IOException {
+        delegate.stringField(fieldInfo, value);
+    }
+
+    @Override
+    public void intField(FieldInfo fieldInfo, int value) throws IOException {
+        delegate.intField(fieldInfo, value);
+    }
+
+    @Override
+    public void longField(FieldInfo fieldInfo, long value) throws IOException {
+        delegate.longField(fieldInfo, value);
+    }
+
+    @Override
+    public void floatField(FieldInfo fieldInfo, float value) throws IOException {
+        delegate.floatField(fieldInfo, value);
+    }
+
+    @Override
+    public void doubleField(FieldInfo fieldInfo, double value) throws IOException {
+        delegate.doubleField(fieldInfo, value);
+    }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 * Use the unique doc id for MMR rerank rather than internal lucenue doc id which is not unique for multiple shards case. [#2911](https://github.com/opensearch-project/k-NN/pull/2911)
 * Fix local ref leak in JNI [#2916](https://github.com/opensearch-project/k-NN/pull/2916)
 * Fix rescoring logic for nested exact search [#2921](https://github.com/opensearch-project/k-NN/pull/2921)
+* Update Visitor to delegate for other fields [#2925](https://github.com/opensearch-project/k-NN/pull/2925)
 
 ### Refactoring
 * Refactored the KNN Stat files for better readability.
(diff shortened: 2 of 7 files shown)


src/main/java/org/opensearch/knn/index/codec/derivedsource/DerivedSourceStoredFieldVisitor.java
```diff
@@ -15,14 +17,21 @@
...
 
+    @Delegate
     private final StoredFieldVisitor delegate;
     private final Integer documentId;
     private final DerivedSourceVectorTransformer derivedSourceVectorTransformer;
 
+    @Override
+    public void binaryField(final FieldInfo fieldInfo, final StoredFieldDataInput value) throws IOException {
+        super.binaryField(fieldInfo, value);
+    }
```
navneet1v: if we are using Delegate on `StoredFieldVisitor` then why we need this function?

VijayanB: At runtime delegate is pointing to FieldsVisitor which was not getting me call back to public void binaryField(FieldInfo fieldInfo, byte[] value) without this change when i tested . i will run the test and post the stack trace here

src/main/java/org/opensearch/knn/index/codec/derivedsource/DerivedSourceStoredFieldVisitor.java
```diff
@@ -15,14 +17,21 @@
...
 public class DerivedSourceStoredFieldVisitor extends StoredFieldVisitor {
 
+    @Delegate
     private final StoredFieldVisitor delegate;
     private final Integer documentId;
     private final DerivedSourceVectorTransformer derivedSourceVectorTransformer;
 
+    @Override
+    public void binaryField(final FieldInfo fieldInfo, final StoredFieldDataInput value) throws IOException {
+        super.binaryField(fieldInfo, value);
```
navneet1v: @VijayanB If I understand this change correctly whats happening here is if there is not any overridden method present in the class `DerivedSourceStoredFieldVisitor` of parent class `StoredFieldVisitor`, the delegate annotation of Lombok will call those functions from variable `private final StoredFieldVisitor delegate`. Is this correct? 

Also looking at the docs of Lombok, this annotation is pretty unfriendly and may lead to some unexpected behaviors. Ref: https://projectlombok.org/features/experimental/Delegate Can we avoid usage of this annotation?

Vikasht34: +1 I think changes is very confusing , delegate is still expremental and there is no point to have delegate as well overridden method.

shatejas: @Delegate has been experimental for a while now, I don't see issues in using it. The biggest advantage of the annotation is that any changes in the StoredFieldVisitor interface will be picked up. Its easy to miss it without and not update the method to delegate just like this one

either way works for me but I agree there are some gotchas around annotation

src/main/java/org/opensearch/knn/index/codec/derivedsource/DerivedSourceStoredFieldVisitor.java
shatejas: Add a comment please in the code

src/main/java/org/opensearch/knn/index/codec/derivedsource/DerivedSourceStoredFieldVisitor.java
VijayanB: Delegate will delegate to the method instead of parent class. I agree it gets messy when you have inheritance and delegation as part of same class. I will update this class to explicitly call the delegate's method. However, like @shatejas mentioned, i think delegate annotation will be helpful in other scenarios where there is no inheritance. The comments from this PR, shouldn't discourage us from trying this annotation when we could use this to our advantage.

Vikasht34: @VijayanB I agree we should never discourage using the annotation when it's helpful , and impact is not negative. 

@shatejas It does not matter how long we have experimental annotation is  , if health is negative we should not use.

These all the exprimental https://projectlombok.org/features/experimental/ and some of them is neutral health , some of them is good to use and some them is negative. 

So bottom line is it really does not matter if any feature is experimental for how long and I can use that .
Unless until we know we know  how Lombok intervenes by acting as an annotation processor that modifies the AST using javac internals and reflection, generating code transparently during compilation

let me know it makes sense.

https://projectlombok.org/features/experimental/Delegate

```
Experimental because:
Not used that much.
Difficult to support for edge cases, such as recursive delegation.
API is rather unfriendly; it would be a lot nicer if you can simply implement some methods and let @Delegate generate delegates for whatever you didn't manually implement, but due to issues with generics erasure this also can't be made to work without caveats.
Current status: negative - Currently we feel this feature will not move out of experimental status anytime soon, and support for this feature may be dropped if future versions of javac or ecj make it difficult to continue to maintain the feature.
```

Issue #2924, opened 2025-10-06
[FP16 Default] Enable optimistic search in memory optimized search


0ctopus13prime:
### Description
Recently Lucene has added optimistic search in place of early termination enabled parallel search - https://github.com/apache/lucene/pull/14226
This is not yet available in memory optimized search, which needs to be pulled in.
From the coarse performance benchmark with Cohere-10M with FP32, optimistic search is showing 10% better QPS than early termination based parallel search.

0ctopus13prime:
SIMD enabled memory optimized search with optimistic search seems showing slightly better performance and recall than early termination based method in multi segments scenario. Recall went up from 93% to 94% and mean QPS went up from 347 to 367 (5% improvement)



## SIMD + Memory optimized search + Early termination
```
|                                                 Min Throughput |   prod-queries |      306.12 |  ops/s |
|                                                Mean Throughput |   prod-queries |      341.07 |  ops/s |
|                                              Median Throughput |   prod-queries |      343.28 |  ops/s |
|                                                 Max Throughput |   prod-queries |      345.76 |  ops/s |
|                                        50th percentile latency |   prod-queries |     54.2734 |     ms |
|                                        90th percentile latency |   prod-queries |     77.5883 |     ms |
|                                        99th percentile latency |   prod-queries |     101.928 |     ms |
|                                      99.9th percentile latency |   prod-queries |     126.095 |     ms |
|                                     99.99th percentile latency |   prod-queries |     147.692 |     ms |
|                                       100th percentile latency |   prod-queries |     153.879 |     ms |
|                                   50th percentile service time |   prod-queries |     54.2734 |     ms |
|                                   90th percentile service time |   prod-queries |     77.5883 |     ms |
|                                   99th percentile service time |   prod-queries |     101.928 |     ms |
|                                 99.9th percentile service time |   prod-queries |     126.095 |     ms |
|                                99.99th percentile service time |   prod-queries |     147.692 |     ms |
|                                  100th percentile service time |   prod-queries |     153.879 |     ms |
|                                                     error rate |   prod-queries |           0 |      % |
|                                                  Mean recall@k |   prod-queries |        0.93 |        |
|                                                  Mean recall@1 |   prod-queries |        0.97 |        |
```

### SIMD + Memory optimized search + Optimistic search
```
|                                                 Min Throughput | warmup-indices |       26.06 |  ops/s |
|                                                Mean Throughput | warmup-indices |       26.06 |  ops/s |
[...]

PR #2933, opened 2025-10-08
Enable optimistic search to memory optimized search.

0ctopus13prime:
### Description
RFC : https://github.com/opensearch-project/k-NN/issues/2924

Enable optimistic search for memory optimized search and deprecate MultiLeafKnnCollector which has an early termination logic.

This PR has three big changes:
1. Now, when memory-optimized search is enabled, all queries use `NativeEngineKnnVectorQuery`.
`KnnQuery`, which only provides a `ScorerSupplier` and performs search within a single leaf segment (with the resulting `Scorer` being consumed by an external `BulkScorer` under the standard Lucene search flow). But optimistic search requires coordination across segments. It needs to run an initial (first-phase) search, then identify and revisit only the segments likely to contain promising results.
To support this coordinated two-phase process, `NativeEngineKnnVectorQuery` is a more suitable entry point than `KnnQuery`.

2. Backported Lucene components required for optimistic search, specifically:
  2.1. ReentrantKnnCollectorManager
  2.2. SeededMappedDISI
  2.3. SeededTopDocsDISI

### Related Issues
#2924 

### Check List


src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java
```diff
@@ -58,6 +64,11 @@
 @Getter
 @RequiredArgsConstructor
 public class NativeEngineKnnVectorQuery extends Query {
+    private static final boolean FORCE_REENTER_TESTING;
```
0ctopus13prime, line 76: Oops, will add comment on this.
This is for testing purpose only. Tuning it on to force IT to always kicks of 2nd revisit to maximize coverage.

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -320,18 +321,17 @@ public PerLeafResult searchLeaf(LeafReaderContext context, int k) throws IOExcep
...
+            final TopDocs result = doExactSearch(context, new BitSetIterator(filterBitSet, filterCardinality), filterCardinality, k);
+            return new PerLeafResult(
+                filterWeight == null ? null : filterBitSet,
+                filterCardinality,
+                result,
+                PerLeafResult.SearchMode.EXACT_SEARCH
+            );
         }
 
-        /*
```
0ctopus13prime, line 327: This logic has been moved to approximateSearch

src/main/java/org/opensearch/knn/index/query/PerLeafResult.java
```diff
@@ -7,21 +7,189 @@
...
+ * the global {@code TopDocs} result set.
+ */
 @Getter
 public class PerLeafResult {
-    public static final PerLeafResult EMPTY_RESULT = new PerLeafResult(new Bits.MatchNoBits(0), TopDocsCollector.EMPTY_TOPDOCS);
+    /**
+     * An immutable, empty {@link BitSet} implementation used to represent
+     * the absence of filter bits without incurring null checks or allocations.
+     */
+    public static final BitSet MATCH_ALL_BIT_SET = new BitSet() {
```
Vikasht34, line 33: Curious why we are not uysing Lucene's MathAllBits or MatcNoBits?

src/main/java/org/opensearch/knn/index/query/PerLeafResult.java
```diff
@@ -7,21 +7,189 @@
...
+            throw new UnsupportedOperationException();
+        }
+
+        @Override
+        public boolean get(int i) {
+            return true;
+        }
+
+        @Override
+        public int length() {
```
Vikasht34, line 85: Are we doing any  iteration on Bitset , this gooona break if we are doing?

src/main/java/org/opensearch/knn/index/query/memoryoptsearch/optimistic/Optimistic2ndSearchUtils.java
```diff
@@ -0,0 +1,83 @@
...
+
+import lombok.experimental.UtilityClass;
+import org.apache.lucene.search.ScoreDoc;
+import org.apache.lucene.util.hnsw.FloatHeap;
+import org.opensearch.knn.index.query.PerLeafResult;
+
+import java.util.List;
+
+@UtilityClass
+public class Optimistic2ndSearchUtils {
```
navneet1v: Please use better names for this class. `2ndSearchUtils` what is first search utils? and why this class is not part of the same class.

src/main/java/org/opensearch/lucene/ReentrantKnnCollectorManager.java
```diff
@@ -0,0 +1,132 @@
...
+ * <p>
+ * This implementation reuses top-ranked results (from the 1st-phase search presumably)
+ * as entry points for a 2nd-phase vector search. It converts previously collected
+ * {@link TopDocs} into corresponding vector entry points using
+ * {@link SeededTopDocsDISI} and {@link SeededMappedDISI}, enabling the internal searcher to start from these known points
+ * instead of beginning from random or default graph entry nodes.
+ */
+@Log4j2
+@RequiredArgsConstructor
+public class ReentrantKnnCollectorManager implements KnnCollectorManager {
```
navneet1v, line 39: This class is a copy/has taken references from https://github.com/apache/lucene/blob/71e822e6240878018a6ff3c28381a0d88bebdc72/lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java#L368

It will be better to mention this somewhere in this class.

src/main/java/org/opensearch/knn/index/query/memoryoptsearch/MemoryOptimizedKNNWeight.java
```diff
@@ -47,6 +48,47 @@ public class MemoryOptimizedKNNWeight extends KNNWeight {
     private static final KnnSearchStrategy.Hnsw DEFAULT_HNSW_SEARCH_STRATEGY = new KnnSearchStrategy.Hnsw(60);
 
     private final KnnCollectorManager knnCollectorManager;
+    @Setter
+    private KnnCollectorManager optimistic2ndKnnCollectorManager;
+
+    public static class OptimisticKnnCollectorManager implements KnnCollectorManager {
```
navneet1v: can we move this to a separate file?

src/main/java/org/opensearch/knn/index/query/PerLeafResult.java
```diff
@@ -7,21 +7,189 @@
...
+    private final int filterBitsCardinality;
+
+    // Top document results for this leaf segment.
     @Setter
     private TopDocs result;
 
-    public PerLeafResult(final Bits filterBits, final TopDocs result) {
-        this.filterBits = filterBits == null ? new Bits.MatchAllBits(0) : filterBits;
+    // Indicates whether this result was produced via exact or approximate search.
+    private final SearchMode searchMode;
```
navneet1v, line 179: what is the use of this parameter?

src/main/java/org/opensearch/knn/index/query/PerLeafResult.java
navneet1v: +1

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -413,9 +424,33 @@ private TopDocs doExactSearch(
...
+     * @param context the {@link LeafReaderContext} representing the current index segment
+     * @param filterIdsBitSet an optional {@link BitSet} indicating document IDs to include in the search;
+     *                        may be {@code null} if no filtering is required
+     * @param filterCardinality the number of documents included in {@code filterIdsBitSet};
+     *                          used to optimize search filtering
+     * @param k the number of nearest neighbors to retrieve
+     * @return a {@link TopDocs} object containing the top {@code k} approximate search results
+     * @throws IOException if an error occurs while reading index data or accessing vector fields
+     */
+    public TopDocs approximateSearch(
```
navneet1v, line 441: why we are making this function public?

src/main/java/org/opensearch/lucene/ReentrantKnnCollectorManager.java
0ctopus13prime: sounds good, will update in the next rev.

src/main/java/org/opensearch/knn/index/query/memoryoptsearch/optimistic/Optimistic2ndSearchUtils.java
0ctopus13prime: I think `2nd` is bit misleading.. will find a better name for that.

src/main/java/org/opensearch/knn/index/query/memoryoptsearch/MemoryOptimizedKNNWeight.java
0ctopus13prime: sure, will update in the next rev.

src/main/java/org/opensearch/knn/index/query/PerLeafResult.java
0ctopus13prime: Optimistic search would do deep dive HNSW search with acquired top k results as seeds. And this will not be needed if the results acquired via exact search. Hence having search mode here, and let it bypass the second search if possible.

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
0ctopus13prime: We need this particular function in optimistic second search. Otherwise, if using searchLeaf, then we will end up building filter bitset twice.

src/main/java/org/opensearch/knn/index/query/PerLeafResult.java
0ctopus13prime: Only place using this one is when we're getting siblings in nested case.
In there, we don't do iteration.

src/main/java/org/opensearch/knn/index/query/PerLeafResult.java
0ctopus13prime: I could not use both, as it's sub class of `Bits` while we need BitSet in here. 😵‍💫 
Since optimistic search will call `approximateSearch` twice, we need to keep `BitSet` for reusing.

src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java
```diff
@@ -58,6 +64,17 @@
 @Getter
 @RequiredArgsConstructor
 public class NativeEngineKnnVectorQuery extends Query {
+    /**
+     * A special flag used for testing purposes that forces execution of the second (exact) search
+     * in optimistic search mode, regardless of the results returned by the first approximate search.
+     * <p>
+     * This flag should never be enabled in production; it is intended for testing and debugging only.
+     */
+    private static final boolean FORCE_REENTER_TESTING;
```
navneet1v, line 76: lets remove this in next iteration of PR.

src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java
```diff
@@ -236,11 +263,111 @@ private List<PerLeafResult> doSearch(
...
             tasks.add(() -> searchLeaf(leafReaderContext, knnWeight, k));
         }
-        return indexSearcher.getTaskExecutor().invokeAll(tasks);
+
+        // Execute search tasks
+        final List<PerLeafResult> perLeafResults = indexSearcher.getTaskExecutor().invokeAll(tasks);
+
+        // For memory optimized search, it should kick off 2nd search if optimistic
+        if (knnQuery.isMemoryOptimizedSearch() && perLeafResults.size() > 1) {
+            run2ndOptimisticSearch(perLeafResults, knnWeight, leafReaderContexts, k, indexSearcher);
```
navneet1v: lets add a debug log here which states that we are running 2nd phase of search.

navneet1v: Also lets add the debug logs for profiling of first phase and second phase of optimistic searches

src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java
```diff
@@ -236,11 +263,111 @@ private List<PerLeafResult> doSearch(
...
+            totalResults += perLeafResult.getResult().scoreDocs.length;
+        }
+
+        // If we got empty results, then return immediately
+        if (totalResults == 0) {
+            return;
+        }
+
+        // Build segment to results table
+        final Map<Integer, TopDocs> segmentOrdToResults = new HashMap<>(leafReaderContexts.size());
```
navneet1v: can we not use an array of TopDocs here where index of the array is the segmentIndex.

src/main/java/org/opensearch/lucene/OptimisticKnnCollectorManager.java
```diff
@@ -0,0 +1,85 @@
...
+        if (delegate.isOptimistic()) {
+            @SuppressWarnings("resource")
+            float leafProportion = context.reader().maxDoc() / (float) context.parent.reader().maxDoc();
+            int perLeafTopK = perLeafTopKCalculation(k, leafProportion);
+            // if we divided by zero above, leafProportion can be NaN and then this would be 0
+            assert perLeafTopK > 0;
+            return delegate.newOptimisticCollector(visitedLimit, searchStrategy, context, perLeafTopK);
+        }
+        // We don't support optimistic collection, so just do regular execution path
+        return delegate.newCollector(visitedLimit, searchStrategy, context);
```
navneet1v, line 73: Not sure where else we will pass a non delegateble collector. But since this class is coming from Lucene, its fine to keep it like this.

src/main/java/org/opensearch/knn/index/query/PerLeafResult.java
navneet1v: The cases when results will be acquired from exact search is the filters case right?

navneet1v (commented):
Overall looks good to me.

I am not seeing how this new memory optimized search is getting integrated with the explain api. 


Also let put the destination branch as main branch.

build.gradle
```diff
@@ -563,6 +565,8 @@ def commonIntegTestClusters(OpenSearchCluster cluster, _numNodes){
 
 testClusters.integTest {
     commonIntegTestClusters(it, _numNodes)
+    // Forcing optimistic search for testing
+    systemProperty 'mem_opt_srch.force_reenter', 'true'
```
navneet1v, line 556: do we need this ?

build.gradle
```diff
@@ -575,6 +579,8 @@ testClusters.integTestRemoteIndexBuild {
     keystore 's3.client.default.access_key', "${System.getProperty("access_key")}"
     keystore 's3.client.default.secret_key', "${System.getProperty("secret_key")}"
     keystore 's3.client.default.session_token', "${System.getProperty("session_token")}"
+    // Forcing optimistic search for testing
+    systemProperty 'mem_opt_srch.force_reenter', 'true'
```
navneet1v, line 570: same as above.

build.gradle
navneet1v: If you want to run search for memory optimized cases then lets create another gradle task and also a new CI that runs that task.

src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java
0ctopus13prime: This is needed for testing the second reentrance in optimistic search.
It's bit tricky to make the data to ensure there's at least one segment whose min topk score is greater than the minimum score in the merged top-k results

src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java
0ctopus13prime: will update in the next rev.

0ctopus13prime: will update in the next rev.

src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java
0ctopus13prime: I think it should be fine, I copied this code from lucene.
Do you have any concerns?

src/main/java/org/opensearch/lucene/OptimisticKnnCollectorManager.java
0ctopus13prime: In this case, we always produce optimistic collector for plain vector search. (e.g. HNSW knn approximate search)
For nested, radial search, we use different collectors which do not use optimistic search.

build.gradle
0ctopus13prime: We need this to force it to run 2nd search in optimistic.
Since the 2nd search will kick off only if there's segment whose min score > the min score in merged results, it was tricky for me to make the data.

build.gradle
0ctopus13prime: We need this to force it to run 2nd search in optimistic.
Since the 2nd search will kick off only if there's segment whose min score > the min score in merged results, it was tricky for me to make the data.

src/main/java/org/opensearch/knn/index/query/PerLeafResult.java
0ctopus13prime: Yes! Whenever running exact search for whatever reason, it will set the mode as EXACT_SEARCH.
If that's the case, then we should not run 2nd search as it's pointless.

src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java
0ctopus13prime: Hmm... I think shard id grows linearly. at the same time, first generation would be gone after merging, so later on there can be segment ids from 1000 to 1100. 
Let me create a hash map only for segments running 2nd search in optimistic search.
In general, there will be far fewer segments we would run 2nd search against, it will reduce memory space further significantly.

build.gradle
navneet1v: Thanks for the clarification. We also run remote cluster tests will this system property works for those cases too?

src/main/java/org/opensearch/lucene/OptimisticKnnCollectorManager.java
navneet1v: make sense. thanks

build.gradle
0ctopus13prime: Yes, this will enforce all cluster during integration to have that flag, will force it to kick off 2nd search.

navneet1v: > Yes, this will enforce all cluster during integration to have that flag, will force it to kick off 2nd search.

Wait is this for cluster setup? if yes then it will fail for release ITs since the cluster setup is done separately. :(

src/main/java/org/opensearch/knn/index/query/memoryoptsearch/MemoryOptimizedKNNWeight.java
```diff
@@ -47,6 +49,8 @@ public class MemoryOptimizedKNNWeight extends KNNWeight {
     private static final KnnSearchStrategy.Hnsw DEFAULT_HNSW_SEARCH_STRATEGY = new KnnSearchStrategy.Hnsw(60);
 
     private final KnnCollectorManager knnCollectorManager;
+    @Setter
+    private KnnCollectorManager optimistic2ndKnnCollectorManager;
```
shatejas: nit: Do you really need KNNCollectorManager interface here? can we just have ReentrantKNNCollectorManager?

src/main/java/org/opensearch/knn/index/query/memoryoptsearch/MemoryOptimizedKNNWeight.java
```diff
@@ -184,7 +188,10 @@ private TopDocs queryIndex(
         }
 
         // Create a collector + bitset
-        final KnnCollector knnCollector = knnCollectorManager.newCollector(visitedLimit, DEFAULT_HNSW_SEARCH_STRATEGY, context);
+        final KnnCollectorManager collectorManager = optimistic2ndKnnCollectorManager != null
+            ? optimistic2ndKnnCollectorManager
+            : knnCollectorManager;
```
shatejas: I would have rather preferred client of this method passing in the collectorManager, I understand that its a larger change. Can we have github issue attached to this and have a fast follow?

src/main/java/org/opensearch/knn/index/query/memoryoptsearch/MemoryOptimizedKNNWeight.java
0ctopus13prime: Sure, will update in the next rev.

src/main/java/org/opensearch/knn/index/query/memoryoptsearch/MemoryOptimizedKNNWeight.java
0ctopus13prime: Sure, will do.

build.gradle
0ctopus13prime: It's passing JVM args -Dmem_opt_srch.force_reenter=true to IT cluster, it should work, and I verified it

src/main/java/org/opensearch/knn/index/query/memoryoptsearch/MemoryOptimizedKNNWeight.java
0ctopus13prime: Created a link : https://github.com/opensearch-project/k-NN/issues/2944

build.gradle
0ctopus13prime: It's an environment variable passing to force the algorithm to start 2nd search in Github CI. 
So as long as CI is enforcing it, it will cover optimistic search testing functionality and won't be affecting release IT.

src/main/java/org/opensearch/knn/index/query/memoryoptsearch/MemoryOptimizedKNNWeight.java
```diff
@@ -47,6 +50,8 @@ public class MemoryOptimizedKNNWeight extends KNNWeight {
     private static final KnnSearchStrategy.Hnsw DEFAULT_HNSW_SEARCH_STRATEGY = new KnnSearchStrategy.Hnsw(60);
 
     private final KnnCollectorManager knnCollectorManager;
+    @Setter
+    private ReentrantKnnCollectorManager optimistic2ndKnnCollectorManager;
```
shatejas: nit: rename to `reentrantKNNCollectorManager`

src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java
```diff
@@ -236,11 +266,116 @@ private List<PerLeafResult> doSearch(
...
+                        )
+                    );
+                    contextIndices.add(i);
+                }
+            }
+        }
+
+        // Kick off 2nd search tasks
+        if (secondDeepDiveTasks.isEmpty() == false) {
+            final ReentrantKnnCollectorManager knnCollectorManagerPhase2 = new ReentrantKnnCollectorManager(
```
shatejas: `reentrantCollectorManager`

src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java
```diff
@@ -236,11 +266,116 @@ private List<PerLeafResult> doSearch(
...
+                    + (knnWeight == null ? "null" : knnWeight.getClass().getSimpleName())
+                    + "], expected="
+                    + MemoryOptimizedKNNWeight.class.getSimpleName()
+            );
+            return;
+        }
+
+        assert (perLeafResults.size() == leafReaderContexts.size());
+
+        // Get collector manager first
```
shatejas: nit: is this the right comment?

src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java
```diff
@@ -236,11 +266,116 @@ private List<PerLeafResult> doSearch(
...
+        final List<Callable<TopDocs>> secondDeepDiveTasks = new ArrayList<>();
+        final List<Integer> contextIndices = new ArrayList<>();
+        final Map<Integer, TopDocs> segmentOrdToResults = new HashMap<>();
+
+        for (int i = 0; i < leafReaderContexts.size(); ++i) {
+            final LeafReaderContext leafReaderContext = leafReaderContexts.get(i);
+            final PerLeafResult perLeafResult = perLeafResults.get(i);
+            final TopDocs perLeaf = perLeafResults.get(i).getResult();
+            if (perLeaf.scoreDocs.length > 0 && perLeafResult.getSearchMode() == PerLeafResult.SearchMode.APPROXIMATE_SEARCH) {
+                if (FORCE_REENTER_TESTING || perLeaf.scoreDocs[perLeaf.scoreDocs.length - 1].score >= minTopKScore) {
```
shatejas, line 339: nit: Its not ideal to have a testing flag in final code, can't we create a test case where we mock scores?

src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java
```diff
@@ -236,11 +266,116 @@ private List<PerLeafResult> doSearch(
...
+            );
+            final StopWatch stopWatch = startStopWatch(log);
+            run2ndOptimisticSearch(perLeafResults, knnWeight, leafReaderContexts, k, indexSearcher);
+            stopStopWatchAndLog(log, stopWatch, "2ndOptimisticSearch", knnQuery.getShardId(), "All Shards", knnQuery.getField());
+        }
+
+        return perLeafResults;
+    }
+
+    private void run2ndOptimisticSearch(
```
shatejas: nit: name is misleading. the first pass is optimistic, the second pass is not. maybe `reentrantSearch`

src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java
```diff
@@ -236,11 +266,116 @@ private List<PerLeafResult> doSearch(
...
+
+            // Make weight use reentrant collector manager
+            memoryOptKNNWeight.setOptimistic2ndKnnCollectorManager(knnCollectorManagerPhase2);
+
+            final List<TopDocs> deepDiveTopDocs = indexSearcher.getTaskExecutor().invokeAll(secondDeepDiveTasks);
+
+            // Override results for target context
+            for (int i = 0; i < deepDiveTopDocs.size(); ++i) {
+                // Override with the new results
+                final TopDocs resultsFrom2ncDeepDive = deepDiveTopDocs.get(i);
```
shatejas:
```suggestion
                final TopDocs resultsFromDeepDive = deepDiveTopDocs.get(i);
```

shatejas (dismissed):
Looks good overall, some minor comments

src/main/java/org/opensearch/knn/index/query/memoryoptsearch/MemoryOptimizedKNNWeight.java
0ctopus13prime: will update in the next rev

src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java
0ctopus13prime: will update in the next rev

src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java
0ctopus13prime: will update in the next rev

src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java
0ctopus13prime: Since whether if it kicks off the second search will be entirely depending on data. Which is hard to be made, and this was the only solution I found so far to always enforce it to enter 2nd search in optimistic.

src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java
0ctopus13prime: will update in the next rev

src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java
0ctopus13prime: will update in the next rev

0ctopus13prime:
@navneet1v 
It's timed out

```
2025-10-22T19:33:55.9624590Z REPRODUCE WITH: ./gradlew ':integTest' --tests 'org.opensearch.knn.index.FaissIT.testEndToEnd_whenDoRadiusSearch_whenDistanceThreshold_whenMethodIsHNSWFlat_thenSucceed' -Dtests.seed=FD2D022924B4F513 -Dtests.security.manager=false -Dtests.locale=jmc-Latn-TZ -Dtests.timezone=Etc/Zulu -Druntime.java=21
2025-10-22T19:33:55.9640570Z FaissIT > classMethod FAILED
2025-10-22T19:33:55.9643660Z 
2025-10-22T19:33:55.9644380Z     java.lang.Exception: Suite timeout exceeded (>= 1200000 msec).
```

src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java
```diff
@@ -236,11 +266,116 @@ private List<PerLeafResult> doSearch(
...
+        final List<Callable<TopDocs>> secondDeepDiveTasks = new ArrayList<>();
+        final List<Integer> contextIndices = new ArrayList<>();
+        final Map<Integer, TopDocs> segmentOrdToResults = new HashMap<>();
+
+        for (int i = 0; i < leafReaderContexts.size(); ++i) {
+            final LeafReaderContext leafReaderContext = leafReaderContexts.get(i);
+            final PerLeafResult perLeafResult = perLeafResults.get(i);
+            final TopDocs perLeaf = perLeafResults.get(i).getResult();
+            if (perLeaf.scoreDocs.length > 0 && perLeafResult.getSearchMode() == PerLeafResult.SearchMode.APPROXIMATE_SEARCH) {
+                if (FORCE_REENTER_TESTING || perLeaf.scoreDocs[perLeaf.scoreDocs.length - 1].score >= minTopKScore) {
```
shatejas, line 339: Add a debug log indicating its reentering with the value of `FORCE_REENTER_TESTING`

0ctopus13prime: Will update in the next rev.

PR #2941, opened 2025-10-14
Make 2x compression default for faiss.

MrFlap:
### Description
Make 2x compression default for faiss. Change the testing dataset so that the vectors won't be clipped.

### Check List


0ctopus13prime:
Hi @MrFlap, thank you for raising PR.
I think we should get align on what would be the 'default' settings for Faiss engine. 
With this change, we will give 2x compression for `ivf`, and do we want to give 2x for all methods including hnsw + ivf in Faiss?

MrFlap:
Hey @0ctopus13prime, I think that should be fine, right? Unless we see somewhere that IVF has a significant recall or throughput drop when using fp16.

0ctopus13prime:
@MrFlap 
It's not my call 😅 
I think we should make it clear in here whether if the direction is the one that we want to go.
It's clear for us to make FP16 default for HNSW as LuceneOnFaiss is coming up, but for IVF, not sure we have aligned on the benefits yet. Just saying that we need to make it clear that we want to make FP16 default for all.

cc @navneet1v @shatejas @vamshin @Vikasht34

Issue #2938, opened 2025-10-14
Unnecessary usage of Heap during Opening of NativeEngines990KnnVectorsReader when memory optimized search is enabled


navneet1v:
## Description
After going through the code, I can see that we are creating a reference to `VectorSearcherHolder` for all the vector fields per `NativeEngines990KnnVectorsReader`. The code assumes that for every vector field per segment there will be 1 instance of `NativeEngines990KnnVectorsReader`. This is an incorrect assumption since KNNVectorFormat is at a field level so for every field when SegmentReaders get opened Lucene initialize a `KnnVectorsReader` [here](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/codecs/perfield/PerFieldKnnVectorsFormat.java#L206-L240). 

So if we have 2 vector field in the index then the hash map having `VectorSearcherHolder` for every vector field will be initialized twice. Which can start to become an overhead and unnecessary usage of heap when we have a lot of segments and shards per node.

Another problem here is the inherent assumption the code is making which is `NativeEngines990KnnVectorsReader` is one per segment. This is incorrect and may lead to severe issues in future. Ideally we should fix this to ensure that in future if we move from lazy loading of graphs to eager loading of graphs we don't hit other memory bottlenecks.

The same: `NativeEngines990KnnVectorsReader` also gets opened from `DerviedSourceReader` [here](https://github.com/opensearch-project/k-NN/blob/b285075876b347d5f0977259181604e1b4629bb4/src/main/java/org/opensearch/knn/index/codec/KNN1030Codec/KNN1030Codec.java#L84-L88). This I am not sure if we can avoid but opening of this reader also adds overhead to heap. 

**_Note:_ Current code is not creating any major issues but it can lead to major issues in future if lets say we move from lazy loading of graphs to eager loading of graphs** so better we fix this now.

### Steps to Repro:
1. Create a KNN index with more than 1 vector field.
2. Index data for both the fields and ensure that 1 segment is created.
3. Add break points/logs in `NativeEngines990KnnVectorsReader` constructor to see `NativeEngines990KnnVectorsReader` is called from Derived Source and also from `KnnVectorsReader` [here](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/codecs/perfield/PerFieldKnnVectorsFormat.java#L206-L240) for every field.

PR #2948, opened 2025-10-18
Removed VectorSearchHolders map from NativeEngines990KnnVectorsReader

navneet1v:
### Description
Removed VectorSearchHolders map from NativeEngines990KnnVectorsReader since NativeEngines990KnnVectorsReader is per field. Hence it doesn't need to understand other fields and have to lazy init the VectorSearchHolder object.

The current merged code is functionally correct, but it assumes that we have a single NativeEngines990KnnVectorsReader for a segment for all the fields, which is not the case. Hence correcting the code to ensure that in future we don't build any extra logic on top of this assumption.

### Related Issues
Resolves https://github.com/opensearch-project/k-NN/issues/2938

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
+++ b/src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
@@ -60,14 +60,16 @@ import static org.opensearch.knn.index.mapper.KNNVectorFieldMapper.KNN_FIELD;
  */
 @Log4j2
 public class NativeEngines990KnnVectorsReader extends KnnVectorsReader {
-    private static final int RESERVE_TWICE_SPACE = 2;
-    private static final float SUFFICIENT_LOAD_FACTOR = 0.6f;
 
     private final FlatVectorsReader flatVectorsReader;
     private Map<String, String> quantizationStateCacheKeyPerField;
     private final SegmentReadState segmentReadState;
     private final List<String> cacheKeys;
-    private volatile Map<String, VectorSearcherHolder> vectorSearchers;
+    private volatile VectorSearcherHolder vectorSearcherHolder;
+    // This lock object ensure that only one thread can initialize vectorSearcherHolder object.
+    // This is needed since we are mappings graphs to memory for memory optimized search lazily. But once we make it eager
+    // the lock object will not be needed
+    private final Object vectorSearcherHolderLockObject;
     private final IOContext ioContext;
 
     public NativeEngines990KnnVectorsReader(final SegmentReadState state, final FlatVectorsReader flatVectorsReader) {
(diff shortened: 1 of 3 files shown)


Vikasht34 (approved):
Looks Good !!

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
```diff
@@ -300,48 +288,28 @@ private static List<String> getVectorCacheKeysFromSegmentReaderState(SegmentRead
...
         }
 
-        if (searcherHolder.isSet()) {
-            return searcherHolder.getVectorSearcher();
-        }
-
-        synchronized (searcherHolder) {
-            if (searcherHolder.isSet()) {
-                return searcherHolder.getVectorSearcher();
+        synchronized (vectorSearcherHolderLockObject) {
```
0ctopus13prime, line 295: Nit : we can just do `synchronized(this)`

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
```diff
@@ -223,11 +226,10 @@ public void close() throws IOException {
         final List<Closeable> closeables = new ArrayList<>();
         closeables.add(flatVectorsReader);
 
-        // Close vector searchers if loaded.
-        if (vectorSearchers != null) {
-            closeables.addAll(
-                vectorSearchers.values().stream().filter(VectorSearcherHolder::isSet).map(VectorSearcherHolder::getVectorSearcher).toList()
-            );
+        // Close Vector Search
+        if (vectorSearcherHolder != null) {
```
0ctopus13prime, line 230: Nit : Do we need this check? 🤔 It never be null I think, are we being conservative??

navneet1v: I think we can remove it too. But just being conservative here. Let keep it I would say

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
navneet1v: > Nit : we can just do synchronized(this)

Synchronizing on the whole class looks very enticing but its actually not a good idea. The main reason is, if lets say outside of this code someone tries to synchronize on the object of NativeEnfines990KNNVectorsReader then one thread will try to block the other one. 

With this lock object, we ensure that if other parts of the code try to synchronize on instance of NativeEnfines990KNNVectorsReader there is no locking both threads can work in parallel.

0ctopus13prime: Sounds good, but then we can acquire lock from `searcherHolder`?

navneet1v:
merging the change. BWC is failing due to some cgroup issues. We will fix it in another PR.

Issue #2962, opened 2025-10-27
[BUG] Fix different hit counts when memory optimized search is enabled.


0ctopus13prime:
## Description

Memory Optimized Search allows users to partially load only the parts of the index required for vector search, whereas when this mode is turned off, all bytes are loaded fully into memory.

Under the hood, Memory Optimized Search uses Lucene’s search algorithm on top of a Faiss index. In other words, it still relies on the graph and vectors from the Faiss index but performs search and result collection using Lucene’s HNSW algorithm and related logic.

Because Lucene’s components are used for result collection, there is a difference in how hit counts are reported between the two modes: when Memory Optimized Search is turned off and when it is turned on.

When using the native Faiss C++ path (that is, when Memory Optimized Search is turned off), the hit count corresponds to the number of results obtained per Lucene segment [Code](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/query/DefaultKNNWeight.java#L171-L177). In contrast, Lucene’s internal definition of hit count represents the number of vectors visited during the search to produce the final top-k results. [Code](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/TopKnnCollector.java#L70) As a result, enabling Memory Optimized Search typically leads to higher hit count values than when it is disabled.

Consequently, for Byte, FP32, and FP16 vector data types, users may observe different hit counts depending on whether Memory Optimized Search is enabled.

## Analysis
Fixing doANNSearch to return hits matching the result size does not appear to be a simple task. The NativeEngineKnnVectorQuery first retrieves the results and then prunes each result set per Lucene segment to ensure that only ScoreDoc entries with scores greater than or equal to the minimum score of the global result are retained. [Code](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java#L119)

For example, suppose there are three Faiss indexes, each returning the following results with k=5:
- Faiss index1: [(doc=7, score=7), (doc=6, score=6), (doc=5, score=5)]
- Faiss index2: [(doc=17, score=17), (doc=16, score=16), (doc=15, score=15)]
- Faiss index3: [(doc=27, score=27), (doc=26, score=26), (doc=25, score=25)]

After merging and sorting all results in descending order of score, we get:
[...]

PR #2965, opened 2025-10-28
[Bug Fix] Fix NativeEngineKNNQuery to return correct results as 'totalHits' when memory optimized search is enabled.

0ctopus13prime:
### Description
NativeEngineKnnVectorQuery currently always returns `k` results, which causes `hits.total` in response to always be k, regardless of how many results are actually returned from each Lucene shard.
With optimistic search in memory optimized search, it's using NativeEngineKnnVectorQuery to return the results. 
However, the total hit count does not match up where memory optimized search is disabled, which returns the sum of the result size acquired from each segment. 

For example, we got 20, 15, 20 from underlying three segments, then the total hit count would be 55. 
But since memory optimized search is using Lucene's component where treating the number of vectors visited as totalHits, usually it's giving bigger value for total hit. 

This CR fixes that behavior so that all matched results are passed to DocsMatchQuery$Weight, ensuring that `hits.total` correctly reflects the sum of the partial results from all Lucene shards.

For more details, please refer to GH Issue : https://github.com/opensearch-project/k-NN/issues/2962

### Related Issues
N/A

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java
+++ b/src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java
@@ -117,8 +117,15 @@ public class NativeEngineKnnVectorQuery extends Query {
             log.debug("Rescoring results took {} ms. oversampled k:{}, segments:{}", rescoreTime, firstPassK, leafReaderContexts.size());
         }
 
-        if (expandNestedDocs) {
+        // Since memory optimized search is using this class, we need to return the same totalHits value when it's disabled.
+        // e.g. Let's say we got 100 result per each segment where #segments=3, then 300 should be returned as total hit.
+        // Therefore, when memory optimized search is enabled, we should skip taking top-k and return whatever we got from approximate
+        // search.
+        if (knnQuery.isMemoryOptimizedSearch() == false) {
             ResultUtil.reduceToTopK(perLeafResults, finalK);
+        }
+
+        if (expandNestedDocs) {
             StopWatch stopWatch = new StopWatch().start();
             perLeafResults = retrieveAll(indexSearcher, leafReaderContexts, knnWeight, perLeafResults, rescoreContext == null);
             long time_in_millis = stopWatch.stop().totalTime().millis();
@@ -128,32 +135,21 @@ public class NativeEngineKnnVectorQuery extends Query {
             }
         }
 
-        // Get the total result size
-        int totalResultSize = 0;
-        for (PerLeafResult perLeafResult : perLeafResults) {
-            totalResultSize += perLeafResult.getResult().scoreDocs.length;
+        TopDocs[] topDocs = new TopDocs[perLeafResults.size()];
+        for (int i = 0; i < perLeafResults.size(); i++) {
+            TopDocs leafTopDocs = perLeafResults.get(i).getResult();
+            for (ScoreDoc scoreDoc : leafTopDocs.scoreDocs) {
+                scoreDoc.doc += leafReaderContexts.get(i).docBase;
+            }
+            topDocs[i] = leafTopDocs;
         }
 
-        // If empty? then return MatchNoDocs$Weight
-        if (totalResultSize == 0) {
+        TopDocs topK = TopDocs.merge(getTotalTopDoc(topDocs), topDocs);
+
+        if (topK.scoreDocs.length == 0) {
             return new MatchNoDocsQuery().createWeight(indexSearcher, scoreMode, boost);
         }
 
-        // Combine all part results we obtained
-        final ScoreDoc[] scoreDocs = new ScoreDoc[totalResultSize];
-        for (int i = 0, j = 0; i < perLeafResults.size(); i++) {
-            final TopDocs leafTopDocs = perLeafResults.get(i).getResult();
-            final int docBase = leafReaderContexts.get(i).docBase;
-            for (ScoreDoc scoreDoc : leafTopDocs.scoreDocs) {
-                scoreDoc.doc += docBase;
-            }
-            System.arraycopy(leafTopDocs.scoreDocs, 0, scoreDocs, j, leafTopDocs.scoreDocs.length);
-            j += leafTopDocs.scoreDocs.length;
-        }
-        // We don't need to sort docs by score as `DocAndScoreQuery` will sort them by doc id anyway, sorting is redundant.
-        final TopDocs topK = new TopDocs(new TotalHits(totalResultSize, TotalHits.Relation.EQUAL_TO), scoreDocs);
-
-        // Return DocAndScoreQuery$Weight
         return queryUtils.createDocAndScoreQuery(reader, topK, knnWeight).createWeight(indexSearcher, scoreMode, boost);
     }
 
@@ -169,15 +165,15 @@ public class NativeEngineKnnVectorQuery extends Query {
      * @return the total number of documents in the topDocs
      */
     private int getTotalTopDoc(TopDocs[] topDocs) {
-        if (expandNestedDocs == false) {
-            return knnQuery.getK();
+        if (knnQuery.isMemoryOptimizedSearch() || expandNestedDocs) {
+            int sum = 0;
+            for (TopDocs topDoc : topDocs) {
+                sum += topDoc.scoreDocs.length;
+            }
+            return sum;
         }
 
-        int sum = 0;
-        for (TopDocs topDoc : topDocs) {
-            sum += topDoc.scoreDocs.length;
-        }
-        return sum;
+        return knnQuery.getK();
     }
 
     /**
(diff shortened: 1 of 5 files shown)


navneet1v:
> correctly reflects the sum of the partial results from all Lucene shards.

is it shards or segments?

src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java
```diff
@@ -128,20 +128,32 @@ public Weight createWeight(IndexSearcher indexSearcher, ScoreMode scoreMode, flo
...
-        for (int i = 0; i < perLeafResults.size(); i++) {
-            TopDocs leafTopDocs = perLeafResults.get(i).getResult();
-            for (ScoreDoc scoreDoc : leafTopDocs.scoreDocs) {
-                scoreDoc.doc += leafReaderContexts.get(i).docBase;
-            }
-            topDocs[i] = leafTopDocs;
+        // Get the total result size
+        int totalResultSize = 0;
+        for (PerLeafResult perLeafResult : perLeafResults) {
+            totalResultSize += perLeafResult.getResult().scoreDocs.length;
```
shatejas: Can this cause int overflow?

If it does then we have to incorporate GREATER_THAN relation as well

src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java
```diff
@@ -116,9 +116,9 @@ public Weight createWeight(IndexSearcher indexSearcher, ScoreMode scoreMode, flo
             long rescoreTime = stopWatch.stop().totalTime().millis();
             log.debug("Rescoring results took {} ms. oversampled k:{}, segments:{}", rescoreTime, firstPassK, leafReaderContexts.size());
         }
-        ResultUtil.reduceToTopK(perLeafResults, finalK);
```
shatejas, line 119: Can you verify that non-rescore case is covered in unit tests? I don't see why it shouldn't work since doc and score query takes care of it but please validate once

src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java
```diff
@@ -128,20 +128,32 @@ public Weight createWeight(IndexSearcher indexSearcher, ScoreMode scoreMode, flo
...
+        for (int i = 0, j = 0; i < perLeafResults.size(); i++) {
+            final TopDocs leafTopDocs = perLeafResults.get(i).getResult();
+            final int docBase = leafReaderContexts.get(i).docBase;
+            for (ScoreDoc scoreDoc : leafTopDocs.scoreDocs) {
+                scoreDoc.doc += docBase;
+            }
+            System.arraycopy(leafTopDocs.scoreDocs, 0, scoreDocs, j, leafTopDocs.scoreDocs.length);
+            j += leafTopDocs.scoreDocs.length;
+        }
+        // We don't need to sort docs by score as `DocAndScoreQuery` will sort them by doc id anyway, sorting is redundant.
```
shatejas: Nice!

shatejas (commented):
Looks good overall!

Please resolve the conflicts and this is good to go!

src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java
0ctopus13prime: Sure, will add it in next rev.

0ctopus13prime: Hi Tejas, [this test case](https://github.com/opensearch-project/k-NN/blob/main/src/test/java/org/opensearch/knn/index/query/nativelib/NativeEngineKNNVectorQueryTests.java#L469) is covering non-rescore context (which is expanding nested siblings)

src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java
```diff
@@ -128,20 +128,32 @@ public Weight createWeight(IndexSearcher indexSearcher, ScoreMode scoreMode, flo
...
+        final TopDocs topK = new TopDocs(new TotalHits(totalResultSize, TotalHits.Relation.EQUAL_TO), scoreDocs);
+
+        // Return DocAndScoreQuery$Weight
         return queryUtils.createDocAndScoreQuery(reader, topK, knnWeight).createWeight(indexSearcher, scoreMode, boost);
     }
```
navneet1v: If I understand this correctly, since we are not reducing docs to topK count, and just sending K results per segment to doc and score query we are saying that hits count will be correct is that right?

If above is true, then this logic will only work for non disk based vector search, because in disk based vector search we get oversampledK from segments and then reduce them to K. Please let me know if I am missing something here.

src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java
```diff
@@ -116,9 +116,9 @@ public Weight createWeight(IndexSearcher indexSearcher, ScoreMode scoreMode, flo
             long rescoreTime = stopWatch.stop().totalTime().millis();
             log.debug("Rescoring results took {} ms. oversampled k:{}, segments:{}", rescoreTime, firstPassK, leafReaderContexts.size());
         }
-        ResultUtil.reduceToTopK(perLeafResults, finalK);
 
         if (expandNestedDocs) {
+            ResultUtil.reduceToTopK(perLeafResults, finalK);
             StopWatch stopWatch = new StopWatch().start();
             perLeafResults = retrieveAll(indexSearcher, leafReaderContexts, knnWeight, perLeafResults, rescoreContext == null);
             long time_in_millis = stopWatch.stop().totalTime().millis();
```
navneet1v: As a behavior what I am not able to understand is when nested docs are there what would be the right behavior for totalHits? what exactly should be the totalHits in the current implementation and what should be the right behavior?

src/main/java/org/opensearch/knn/index/query/memoryoptsearch/MemoryOptimizedKNNWeight.java
```diff
@@ -203,7 +204,11 @@ private TopDocs queryIndex(
         }
 
         // Make results to return
-        final TopDocs topDocs = knnCollector.topDocs();
+        TopDocs topDocs = knnCollector.topDocs();
+        // Align `hitCount` logic with the non-memory-optimized path by setting it to the size of the result set.
+        // Note: DefaultKNNWeight defines `hitCount` as the number of results returned per Lucene segment,
+        // while Lucene’s implementation interprets it as the total number of vectors visited during search.
+        topDocs = new TopDocs(new TotalHits(topDocs.scoreDocs.length, TotalHits.Relation.EQUAL_TO), topDocs.scoreDocs);
```
navneet1v, line 211: what is the topDocs behavior from earlier in MemoryOptimized cases?

src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java
0ctopus13prime: Yes, that's correct. And whatever we're oversampling or not, the definition of `hitCount` is the size of result we got from each Lucene segment with `k` given.

src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java
0ctopus13prime: Since we are expanding the nested docs of top results, reducing results to top-k is needed.
Otherwise it will expand all parent documents that are not in top-k which is not the desirable behavior. 
In this case, the right behavior is debatable, we can meet half way to define `hitCount` as the total number of expanded child documents whose parent docs in top-k results.

src/main/java/org/opensearch/knn/index/query/memoryoptsearch/MemoryOptimizedKNNWeight.java
0ctopus13prime: Since we're using Lucene, the definition of hitCount also follows it's definition where the number of vectors visited during search (e.g. the number of collector.collect calls)

src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java
navneet1v: >  the definition of hitCount is the size of result we got from each Lucene segment with k given.

I think then I am missing something here, how for disk based vector search this is mantained?

0ctopus13prime: For disk-based queries that require rescoring, the baseline hitCount is always k, because, as you mentioned, the second rescoring phase reduces the aggregated results down to k. (If there are not enough results, hitCount may be less than k, but that’s a rare case.)

We want to change this so that hitCount correctly reflects the actual number of results returned from the Lucene segment using the given k. This behavior is equally being applied consistently to FP32 and FP16, both of which use KNNQuery::scorerSupplier.

PR #2974, opened 2025-11-01
[Bug Fix] Fix race condition on transforming vector in KNNQueryBuilder.

0ctopus13prime:
### Description
We are seeing random and inconsistent score values when a query runs across multiple shards that are all located on a single data node. Specifically, when using cosine similarity, the score for the same document varies between queries even though the query vector does not change. For example, a document with id 321004 may return a cosine score of 0.998 in one query, but sometimes the score drops to around 0.6 with the same query being used. This issue occurs only with cosine similarity. Other similarity types behave consistently. Additionally, when we force the request to target a single shard using the preference parameter ([doc](https://docs.opensearch.org/latest/search-plugins/searching-data/search-shard-routing/)), the score remains stable. The inconsistency appears only when multiple shards on one data node in the search.

### Root cause analysis
When the first query arrives, the engine parses the JSON query and builds a SearchSourceBuilder. During this step, it also calls KNNQueryBuilderParser to create a KNNQueryBuilder instance. [[Code](https://github.com/opensearch-project/OpenSearch/blob/3.3/server/src/main/java/org/opensearch/search/builder/SearchSourceBuilder.java#L1353)]

```
class SearchSourceBuilder {

    public void parseXContent(XContentParser parser, boolean checkTrailingTokens) throws IOException {
        ...
        
        } else if (token == XContentParser.Token.START_OBJECT) {
            if (QUERY_FIELD.match(currentFieldName, parser.getDeprecationHandler())) {
                queryBuilder = parseInnerQueryBuilder(parser);
            } else if (POST_FILTER_FIELD.match(currentFieldName, parser.getDeprecationHandler())) {
                postQueryBuilder = parseInnerQueryBuilder(parser);
        ...
    }
}
```

Next, the engine goes through all shards on the node and creates a SearchShardRequest for each shard, attaching the same SearchSourceBuilder to every request [[Code](https://github.com/opensearch-project/OpenSearch/blob/3.3/server/src/main/java/org/opensearch/action/search/AbstractSearchAsyncAction.java#L266)]. This means multiple SearchShardRequest instances share the same SearchSourceBuilder, and therefore also share the same KNNQueryBuilder instance. [[Code](https://github.com/opensearch-project/OpenSearch/blob/3.3/server/src/main/java/org/opensearch/search/internal/ShardSearchRequest.java#L220)]

```
    private ShardSearchRequest(
        OriginalIndices originalIndices,
        ShardId shardId,
        int numberOfShards,
        SearchType searchType,
        SearchSourceBuilder source,
        Boolean requestCache,
        AliasFilter aliasFilter,
        float indexBoost,
        boolean allowPartialSearchResults,
        String[] indexRoutings,
        String preference,
        Scroll scroll,
        long nowInMillis,
        @Nullable String clusterAlias,
        ShardSearchContextId readerId,
        TimeValue keepAlive
    ) {
        this.shardId = shardId;
        this.numberOfShards = numberOfShards;
        this.searchType = searchType;
        this.source = source;
        ...
```

When each shard executes its search [[Code](https://github.com/opensearch-project/OpenSearch/blob/3.3/server/src/main/java/org/opensearch/search/SearchService.java#L893)], its thread eventually calls doToQuery() while creating the search context.

```
public class SearchService {
    public void executeQueryPhase(
        InternalScrollSearchRequest request,
        SearchShardTask task,
        ActionListener<ScrollQuerySearchResult> listener
    ) {
        final LegacyReaderContext readerContext = (LegacyReaderContext) findReaderContext(request.contextId(), request);
        final Releasable markAsUsed;
        try {
            markAsUsed = readerContext.markAsUsed(getScrollKeepAlive(request.scroll()));
        } catch (Exception e) {
            // We need to release the reader context of the scroll when we hit any exception (here the keep_alive can be too large)
            freeReaderContext(readerContext.id());
            throw e;
        }
        runAsync(getExecutor(null, readerContext.indexShard()), () -> {
            final ShardSearchRequest shardSearchRequest = readerContext.getShardSearchRequest(null);
            try (
                SearchContext searchContext = createContext(readerContext, shardSearchRequest, task, false);
                SearchOperationListenerExecutor executor = new SearchOperationListenerExecutor(searchContext)
            ) {
    ...
```

The problem occurs here : Inside doToQuery(), the query vector in KNNQueryBuilder is normalized in place. But there is no locking or synchronization. [[Code](https://github.com/opensearch-project/k-NN/blob/3.3/src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java#L446)] [[VectorUtil Code](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/internal/vectorization/DefaultVectorUtilSupport.java#L414)]

```
final class DefaultVectorUtilSupport implements VectorUtilSupport {
  @Override
  public float[] l2normalize(float[] v, boolean throwOnZero) {
    ...
    int dim = v.length;
    double l2norm = Math.sqrt(l1norm);
    for (int i = 0; i < dim; i++) {
      v[i] /= (float) l2norm;  // In-place update!
    }
    return v;
  }
  ...
}
```

Since all shard search threads share the same query builder, multiple threads try to normalize the same vector at the same time. This causes the given query vector values to be overwritten by different threads concurrently, resulting in random float values.

In short:

1. Multiple threads share the same query vector instance (e.g. float[]).
2. Each thread modifies it without synchronization. (e.g. l2normalize)

### Related Issues
N/A

### Check List


src/main/java/org/opensearch/knn/plugin/KNNPlugin.java
```diff
@@ -199,11 +198,9 @@ public class KNNPlugin extends Plugin
     private final Map<String, MMRQueryTransformer<? extends QueryBuilder>> mmrQueryTransformers = new HashMap<>();
 
     static {
-        ForkJoinPool.commonPool().execute(() -> {
```
0ctopus13prime, line 202: KNNSettingTests is failing since the commonPool has forked a thread but never kills it at the end of test.

Error:


1 thread leaked from SUITE scope at org.opensearch.knn.index.KNNSettingsTests: 1) Thread[id=39, name=ForkJoinPool.commonPool-worker-1, state=TIMED_WAITING, group=TGRP-KNNSettingsTests] at java.base/jdk.internal.misc.Unsafe.park(Native Method) at java.base/java.util.concurrent.locks.LockSupport.parkUntil(LockSupport.java:449) at java.base/java.util.concurrent.ForkJoinPool.awaitWork(ForkJoinPool.java:1891) at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1809) at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:188) com.carrotsearch.randomizedtesting.ThreadLeakError: 1 thread leaked from SUITE scope at org.opensearch.knn.index.KNNSettingsTests: 1) Thread[id=39, name=ForkJoinPool.commonPool-worker-1, state=TIMED_WAITING, group=TGRP-KNNSettingsTests] at java.base/jdk.internal.misc.Unsafe.park(Native Method) at java.base/java.util.concurrent.locks.LockSupport.parkUntil(LockSupport.java:449) at java.base/java.util.concurrent.ForkJoinPool.awaitWork(ForkJoinPool.java:1891) at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1809) at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:188) at __randomizedtesting.SeedInfo.seed([5956DDA4BD65A868]:0)

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java
```diff
@@ -787,7 +787,7 @@ protected void parseCreateField(ParseContext context, int dimension, VectorDataT
             }
             final float[] array = floatsArrayOptional.get();
             getVectorValidator().validateVector(array);
-            getVectorTransformer().transform(array);
+            getVectorTransformer().transform(array, true);
```
Vikasht34, line 790: Why we need this Inplace boolean?

Vikasht34 (commented):
Changes looks good !!

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java
0ctopus13prime: Since it's also being used in ingestion workflow which is fine to do L2 normalization in-place.
So I've added a flag there to have it conditionally apply the transformation to reduce memory usage as possible.
If we ingest 100K docs, then without the flag, we will end up allocating float[] for 100K times when cosine space type being used

src/main/java/org/opensearch/knn/index/query/BaseQueryFactory.java
```diff
@@ -52,6 +52,7 @@ public static class CreateQueryRequest {
         private String indexName;
         private String fieldName;
         private float[] vector;
+        private float[] originalVector;
```
shatejas, line 55: nit: Please add a comment that its a normalized vector

0ctopus13prime: Sure, will update in next rev

Vikasht34 (approved):

PR #2977, opened 2025-11-03
Cosine Similarity Multi Shard Testing

MrFlap:
### Description
Adds testing for multi shard cosine similarity.

### Related Issues
Adds more general integration tests to PR #[2974](https://github.com/opensearch-project/k-NN/pull/2974)

### Check List


src/test/java/org/opensearch/knn/KNNMultiNodeTestCase.java
```diff
@@ -0,0 +1,323 @@
...
+import org.opensearch.common.settings.Settings;
+import org.opensearch.core.xcontent.XContentBuilder;
+import org.opensearch.common.xcontent.XContentFactory;
+import org.opensearch.knn.quantization.models.quantizationState.QuantizationStateCacheManager;
+import org.opensearch.core.rest.RestStatus;
+import org.opensearch.test.OpenSearchIntegTestCase;
+import org.opensearch.test.hamcrest.OpenSearchAssertions;
+
+import java.io.IOException;
+import java.util.*;
```
shatejas: nit: don't use * imports

src/test/java/org/opensearch/knn/KNNMultiNodeTestCase.java
```diff
@@ -0,0 +1,323 @@
...
+        request.source(xContentBuilder);
+        OpenSearchAssertions.assertAcked(client().admin().indices().putMapping(request).actionGet());
+    }
+
+    protected void updateIndexSetting(String indexName, Settings setting) {
+        UpdateSettingsRequest request = new UpdateSettingsRequest(setting, indexName);
+        OpenSearchAssertions.assertAcked(client().admin().indices().updateSettings(request).actionGet());
+    }
+
+    protected void createKnnNestedIndexMapping(String indexName, String fieldPath, Integer dimensions) throws IOException {
```
shatejas: Is this being used? Lets not add methods which are not being used.

src/test/java/org/opensearch/knn/KNNMultiNodeTestCase.java
```diff
@@ -0,0 +1,323 @@
...
+    protected Settings getKNNDefaultIndexSettingsBuildsGraphAlways() {
+        return Settings.builder()
+            .put("number_of_shards", 1)
+            .put("number_of_replicas", 0)
+            .put("index.knn", true)
+            .put(KNNSettings.INDEX_KNN_ADVANCED_APPROXIMATE_THRESHOLD, 0)
+            .build();
+    }
+
+    protected Settings getKNNDefaultMultiShardIndexSettingsBuildsGraphAlways() {
```
shatejas: I think we should start adding a testFixture instead of putting it in a base class. This limits the user to 8 shards and 0 replicas. Take this as an example for test fixture: https://github.com/opensearch-project/k-NN/blob/main/src/testFixtures/java/org/opensearch/knn/KNNJsonIndexMappingsBuilder.java we can have KNNIndexSettingsBuilder similar to it.

src/test/java/org/opensearch/knn/KNNSingleNodeTestCase.java
```diff
@@ -45,11 +45,7 @@
 import com.carrotsearch.randomizedtesting.annotations.ThreadLeakFilters;
 
 import java.io.IOException;
-import java.util.Base64;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.EnumSet;
-import java.util.Map;
+import java.util.*;
```
shatejas: Nit: avoid * imports

src/test/java/org/opensearch/knn/KNNSingleNodeTestCase.java
```diff
@@ -116,13 +112,27 @@ protected IndexService createKNNIndex(String indexName) {
...
         builder.put(MEMORY_OPTIMIZED_KNN_SEARCH_MODE, true);
         return createIndex(indexName, builder.build());
     }
 
+    protected IndexService createMemoryOptimizedSearchEnabledKNNIndexMultiShard(final String indexName) {
+        Settings settings = getKNNDefaultMultiShardIndexSettingsBuildsGraphAlways();
+        final Settings.Builder builder = Settings.builder().put(settings);
+        builder.put(MEMORY_OPTIMIZED_KNN_SEARCH_MODE, true);
+        return createIndex(indexName, builder.build());
+    }
```
shatejas: Lets not add utility methods for each setting, Let the test actually control creating mapping and settings using test fixtures. Adding method for each setting can easily create an explosion of methods

src/test/java/org/opensearch/knn/KNNSingleNodeTestCase.java
```diff
@@ -309,6 +331,17 @@ protected void searchKNNIndex(String index, String fieldName, float[] vector, in
         assertEquals(response.status(), RestStatus.OK);
     }
 
+    /**
+     * Validates that a k=1 search with a vector from the dataset returns that same vector
+     */
+    protected void validateExactMatch(String index, String fieldName, float[] queryVector, String expectedDocId) {
+        SearchResponse response = client().prepareSearch(index).setQuery(new KNNQueryBuilder(fieldName, queryVector, 1)).get();
```
shatejas: Lets use builder instead of constructor for KNNQueryBuilder

src/test/java/org/opensearch/knn/KNNMultiNodeTestCase.java
```diff
@@ -0,0 +1,323 @@
...
+import static org.opensearch.knn.common.KNNConstants.MODEL_DESCRIPTION;
+import static org.opensearch.knn.common.KNNConstants.MODEL_ERROR;
+import static org.opensearch.knn.common.KNNConstants.MODEL_ID;
+import static org.opensearch.knn.common.KNNConstants.MODEL_INDEX_NAME;
+import static org.opensearch.knn.common.KNNConstants.MODEL_STATE;
+import static org.opensearch.knn.common.KNNConstants.MODEL_TIMESTAMP;
+import static org.opensearch.knn.common.KNNConstants.VECTOR_DATA_TYPE_FIELD;
+import static org.opensearch.knn.index.KNNSettings.MEMORY_OPTIMIZED_KNN_SEARCH_MODE;
+
+@OpenSearchIntegTestCase.ClusterScope(scope = OpenSearchIntegTestCase.Scope.SUITE, numDataNodes = 2)
```
shatejas: I am not sure if want one size fits all for numOfDataNodes. Lets keep this on individual test level?

src/test/java/org/opensearch/knn/KNNMultiNodeTestCase.java
```diff
@@ -0,0 +1,323 @@
...
+    }
+
+    @Override
+    public void tearDown() throws Exception {
+        NativeMemoryCacheManager.getInstance().invalidateAll();
+        NativeMemoryCacheManager.getInstance().close();
+        QuantizationStateCacheManager.getInstance().close();
+        NativeMemoryLoadStrategy.IndexLoadStrategy.getInstance().close();
+        NativeMemoryLoadStrategy.TrainingLoadStrategy.getInstance().close();
+        NativeMemoryLoadStrategy.AnonymousLoadStrategy.getInstance().close();
```
shatejas: Do we need this for all multi-node tests? QuantizationState has caused issues, Lets not have this at base class level?

src/test/java/org/opensearch/knn/index/CosineSimilarityMultiNodeTest.java
```diff
@@ -0,0 +1,83 @@
...
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ExecutionException;
+
+public class CosineSimilarityMultiNodeTest extends KNNMultiNodeTestCase {
+
+    public void testCosineSimilaritySearchMultiNode() throws IOException, InterruptedException, ExecutionException {
```
shatejas: Lets make this a parametrized test. We want this to be extensible to a combination of different space type and engines. A good example would be https://github.com/opensearch-project/OpenSearch/blob/main/server/src/internalClusterTest/java/org/opensearch/index/IndexSortIT.java#L73. 

the parameters can be spaceType, engine, maybe some index settings, a ground truth file. This will make it easier to test other spacetypes as well. To start with we can just have cosinesim-faiss combination

src/test/java/org/opensearch/knn/index/SimilarityMultiNodeTestIT.java
```diff
@@ -0,0 +1,83 @@
...
+            Float[] vector = new Float[dimensions];
+            for (int i = 0; i < dimensions; i++) {
+                vector[i] = ((Number) vectorList.get(i)).floatValue();
+            }
+            if (queryVector == null) {
+                queryVector = new float[dimensions];
+                for (int i = 0; i < dimensions; i++) {
+                    queryVector[i] = vector[i];
+                }
+            }
```
shatejas: nit: maye create a function to improve readability

src/test/java/org/opensearch/knn/KNNMultiNodeTestCase.java
MrFlap: @shatejas This is tricky because it's a decorator level fix. Meaning that if we want to make it test by test, we need to call this decorator every time. Whoever tries to extend this will be confused why the multi node test isn't spinning up multiple nodes. I'm going to investigate if there is a way that we can make this default and add a way to override the number of nodes.

src/test/java/org/opensearch/knn/KNNMultiNodeTestCase.java
MrFlap: @shatejas I'm just going off of what was already in the single node test case: https://github.com/opensearch-project/k-NN/blob/e534928c142ffaba06cd9d69ea37701c76a24524/src/test/java/org/opensearch/knn/KNNSingleNodeTestCase.java#L98

src/test/java/org/opensearch/knn/KNNSingleNodeTestCase.java
MrFlap: Okay, I can make a fixture for this

src/test/java/org/opensearch/knn/KNNMultiNodeTestCase.java
MrFlap: @shatejas I don't think that there is a good way to do this unless there is something in Java that I don't know about. Are we fine with expecting every multi node test case to be annotated with OpenSearchIntegTestCase.ClusterScope(scope = OpenSearchIntegTestCase.Scope.SUITE, numDataNodes = {num_nodes})?

src/test/java/org/opensearch/knn/index/SimilarityMultiNodeTest.java
```diff
@@ -0,0 +1,121 @@
...
+        String fieldName = "vector";
+        int dimensions = 16;
+
+        createIndex(indexName, KNNCommonSettingsBuilder.defaultSettings().memOptSearch().multiShard().build());
+        createMapping(indexName, fieldName, dimensions);
+
+        List<Object> groundTruth = loadGroundTruth();
+        float[] queryVector = indexDocuments(indexName, fieldName, dimensions, groundTruth);
+
+        SearchResponse response = client().prepareSearch(indexName).setQuery(new KNNQueryBuilder(fieldName, queryVector, 10)).get();
```
0ctopus13prime: Nit: new KNNQueryBuilder(...) is deprecated.

src/test/java/org/opensearch/knn/index/SimilarityMultiNodeTest.java
```diff
@@ -0,0 +1,121 @@
...
+
+        validateResults(response, groundTruth);
+    }
+
+    private void createMapping(String indexName, String fieldName, int dimensions) throws IOException, ExecutionException,
```
0ctopus13prime: Nit : ExecutionException, InterruptedException are never being thrown.

src/test/java/org/opensearch/knn/index/SimilarityMultiNodeTest.java
```diff
@@ -0,0 +1,121 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ExecutionException;
+
+@OpenSearchIntegTestCase.ClusterScope(scope = OpenSearchIntegTestCase.Scope.SUITE, numDataNodes = 2)
+public class SimilarityMultiNodeTest extends KNNMultiNodeTestCase {
```
0ctopus13prime: Add 'IT' at the end to make sure it's included in integration test suites.
In [here](https://github.com/opensearch-project/k-NN/blob/main/build.gradle#L603), we are filtering class files end with 'IT' for integration test.

src/test/java/org/opensearch/knn/KNNMultiNodeTestCase.java
```diff
@@ -0,0 +1,130 @@
...
+import org.opensearch.test.hamcrest.OpenSearchAssertions;
+
+import java.io.IOException;
+import java.util.Map;
+import java.util.concurrent.ExecutionException;
+
+import static org.opensearch.knn.common.KNNConstants.KNN_ENGINE;
+import static org.opensearch.knn.common.KNNConstants.METHOD_HNSW;
+
+public class KNNMultiNodeTestCase extends OpenSearchIntegTestCase {
```
0ctopus13prime: Add 'IT' at the end to make sure it's included in integration test suites.
In [here](https://github.com/opensearch-project/k-NN/blob/main/build.gradle#L603), we are filtering class files end with 'IT' for integration test.

src/test/java/org/opensearch/knn/index/SimilarityMultiShardTestIT.java
```diff
@@ -0,0 +1,118 @@
...
+        float[] queryVector = null;
+        for (Object docObj : groundTruth) {
+            Map<String, Object> doc = (Map<String, Object>) docObj;
+            String id = (String) doc.get("id");
+            List<Object> vectorList = (List<Object>) doc.get("vector");
+            Float[] vector = new Float[dimensions];
+            for (int i = 0; i < dimensions; i++) {
+                vector[i] = ((Number) vectorList.get(i)).floatValue();
+            }
+            if (queryVector == null) {
```
0ctopus13prime: Nit : Seems like we're picking the first vector as a query vector. Could you add a comment like 'Returning the first one as a query vector' etc?

src/test/java/org/opensearch/knn/KNNMultiNodeTestCase.java
```diff
@@ -0,0 +1,130 @@
...
+        KNNQueryBuilder knnQueryBuilder = KNNQueryBuilder.builder().fieldName(fieldName).vector(vector).k(k).build();
+        SearchResponse response = client().prepareSearch(index).setQuery(knnQueryBuilder).get();
+        assertEquals(response.status(), RestStatus.OK);
+    }
+
+    public Map<String, Object> xContentBuilderToMap(XContentBuilder xContentBuilder) {
+        return XContentHelper.convertToMap(BytesReference.bytes(xContentBuilder), true, xContentBuilder.contentType()).v2();
+    }
+
+    public void assertTrainingSucceeds(ModelDao modelDao, String modelId, int attempts, int delayInMillis) throws InterruptedException,
```
0ctopus13prime: NIT : Can you remove this?

0ctopus13prime (commented):
Overall looks good. 
But could you add the suffix "IT" for each case? I think your added test cases are not being running. 
Also please ensure add 'test' as a prefix for the testing method to make sure have it executed.

src/test/java/org/opensearch/knn/KNNMultiNodeTestCase.java
MrFlap: This file doesn't need IT as it is the base class for other tests

src/test/java/org/opensearch/knn/index/SimilarityMultiShardTestIT.java
```diff
@@ -0,0 +1,94 @@
...
+                { SpaceType.COSINESIMIL.getValue(), KNNEngine.FAISS.getName(), "/data/test_sanity_ground_truth_cosine_1000.json" } }
+        );
+    }
+
+    public void testSimilaritySearchMultiShard() throws IOException, InterruptedException, ExecutionException {
+        String indexName = "similarity-test-index";
+        String fieldName = "vector";
+        int dimensions = 16;
+
+        createIndex(indexName, KNNCommonSettingsBuilder.defaultSettings().memOptSearch().multiShard().build());
```
shatejas: Make settings as as a parameter and add both memOpt and faiss cpp settings as separate runs. We need to test it for both

src/test/java/org/opensearch/knn/index/SimilarityMultiShardTestIT.java
```diff
@@ -0,0 +1,94 @@
...
+        createMapping(indexName, fieldName, dimensions);
+
+        List<Object> groundTruth = loadGroundTruth(this.groundTruthFile);
+        float[] queryVector = indexDocuments(indexName, fieldName, dimensions, groundTruth);
```
shatejas: nit: its a bit weird that indexingDocs returns a queryVector :)

src/test/java/org/opensearch/knn/index/SimilarityMultiShardTestIT.java
```diff
@@ -0,0 +1,94 @@
...
+
+        SearchResponse response = client().prepareSearch(indexName).setQuery(new KNNQueryBuilder(fieldName, queryVector, 10)).get();
+
+        validateResults(response, groundTruth);
```
shatejas, line 79: Please delete the index either here or as a part of tearDown

MrFlap: super().tearDown() already deletes every index in org.opensearch.test.OpenSearchSingleNodeTestCase#client 👍

src/test/java/org/opensearch/knn/index/SimilarityMultiShardTestIT.java
MrFlap: Agreed. Renamed to make more sense

navneet1v:
@MrFlap can we fix the merge conflicts

navneet1v:
@MrFlap can we fix the merge conflicts

PR #2983, opened 2025-11-23
Fix patch to have a valid score conversion for BinaryCagra.

0ctopus13prime:
#### Summary

This PR fixes a bug in the HNSW Cagra binary path where float-typed Hamming distances were being incorrectly interpreted as integers. The issue originates from the patch below, where the non-Cagra binary conversion code was not applied to the Cagra binary code path.

Relevant patch location:
https://github.com/opensearch-project/k-NN/blob/3.2/jni/patches/faiss/0006-Add-nested-search-support-for-IndexBinaryHNSWCagra.patch#L141

#### Details

HNSW internally expects distances as float*.
For binary vectors, the raw distances are actually int* (Hamming distances).
To pass them through HNSW, they are cast from int* to float*.
In the non-Cagra binary implementation, these floats are converted back to integers after search:

```
#pragma omp parallel for
for (int i = 0; i < n * k; ++i) {
    distances[i] = std::round(distances_f[i]);  // float -> int
}
```

In the Cagra binary implementation, this conversion step is missing. As a result, the code interprets the float bit pattern as an integer rather than converting the numeric value. For example:

```
float dis = 333;
auto* ptr = (int*)&dis;
int dist_int = *ptr;   // becomes 1134985216
```

This is not a cast; it's a reinterpretation of the underlying 32-bit pattern.
Hamming distance is converted to a similarity score using: 1 / (1 + hamming_dist).
With a raw Hamming distance of 333, the expected score is approximately 0.00299, but the buggy interpretation yields about 0.000000000881.
After adding the missing conversion, the scores become correct.

#### Patch Issue

The existing Cagra binary patch incorrectly applies only negation logic for similarity metrics and does not restore integer distances:
```
if (is_similarity_metric(this->metric_type)) {
    // revert negated distances
    #pragma omp parallel for
    for (int64_t i = 0; i < k * n; i++) {
        distances[i] = -distances[i];
    }
}
```

The float-to-int restoration logic is missing entirely.

#### Impact

Luckily, this bug affects score values but does not change ranking or ordering.
Because Hamming distances are always non-negative integers, interpreting their float bit patterns as integers happens to preserve ordering. For any float A and B representing valid Hamming distances:
```
float A = ...;
int A_int = *((int*)&A);

float B = ...;
int B_int = *((int*)&B);

// A == 0 => A_int == 0
// Ordering is preserved:
A < B  =>  A_int < B_int
A == B =>  A_int == B_int
A > B  =>  A_int > B_int
```

Therefore, search results remain in the correct order, but the scores are invalid and may affect logging or explain-feature output. This is still a correctness bug and should be fixed.

### Related Issues
N/A

### Check List


navneet1v:
@0ctopus13prime lets add a test to ensure that if anything changes in future we are able to catch this bug. We already have remote index build test and also `.faiss` file which can help in building the tests.

0ctopus13prime:
@navneet1v 
Sure, will add tests.

PR #2986, opened 2025-11-25
QuantizationConfig merge fix

finnroblin:
### Description
[Describe what this change achieves]

### Related Issues

### Check List

### DIFF
--- a/qa/rolling-upgrade/src/test/java/org/opensearch/knn/bwc/IndexingIT.java
+++ b/qa/rolling-upgrade/src/test/java/org/opensearch/knn/bwc/IndexingIT.java
@@ -189,7 +189,6 @@ public class IndexingIT extends AbstractRollingUpgradeTestCase {
         validateKNNSearch(testIndex, TEST_FIELD, DIMENSIONS, totalDocsCount, K);
     }
 
-
     /**
      * Test to verify that NMSLIB index creation is blocked in OpenSearch 3.0.0 and later,
      * while ensuring backward compatibility (BWC) for existing indexes created in OpenSearch 2.19,


src/main/java/org/opensearch/knn/index/engine/qframe/QuantizationConfigParser.java
```diff
@@ -66,12 +66,15 @@ public static String toCsv(QuantizationConfig quantizationConfig) {
      * @param csv Csv format of quantization config
      * @return Quantization config
      */
-    public static QuantizationConfig fromCsv(String csv, org.apache.lucene.util.Version luceneVersion) {
+    public static QuantizationConfig fromCsv(String csv, org.apache.lucene.util.Version _luceneVersion) {
         if (csv == null || csv.isEmpty()) {
             return QuantizationConfig.EMPTY;
         }
+        String[] csvArray = CSVUtil.parse(csv);
+        int csvArrayLength = csvArray.length;
```
Vikasht34, line 74: We don't need two conditions for 4 and 2 , Just fail if less than 2 , which is not a valid Binary Quantization.

finnroblin: Sure, will refactor in next rev.

navneet1v: I think we should fail if csv length is not 2 or 4. Any other length of csv is invalid data. I would suggest having the check for 2 and 4.

Vikasht34: May I know the use case why we need to have check for both 2 or 4 !! It's redundant ..  The logic right now , if it is not 2 fail ..which was original thing when we had only Binary ..now if we have >=2 then it's difinetly Vanila Binary Quantization with extra flavour for ADC and RR .. I would love to know the reason and data point why we think we should have both the checks.

Below is correctly written , it handles both 2 or 4

if (csvArrayLength < 2) {
            throw new IllegalArgumentException(String.format(Locale.ROOT, "Invalid csv (length < 2) for quantization config: \"%s\"", csv));
        }

navneet1v: what if in some case the length is 3, or 5? Since we only 2 and 4 are valid lengths we should make sure that we are doing those validations and in all other case we fail. This will ensure that no data corruption is happening.

The check `< 2` is only handling the case when length is less than 2 and nothing else. So we say that `< 2` check handles both `length == 2` and `length == 4` then its not what this code is doing.

Vikasht34: It can never be 3 or 5  according to me ..Quntization writes into csv , it's not dynamic .,. bringing hypothesis of 3 or 5 ,   Needs Concrete data points to support this theory . Could you list out cases for supporting this, would be curious to know what would be cases which can bring  this.

Yes check < 2 handles the what's valid for Binary Quantization.  For BQ either we can have 2 or 4 . if Min is 2 < 2 handles both.Again I need some code pointers how you think it's not handling

navneet1v: > It can never be 3 or 5 according to me ..Quntization writes into csv , it's not dynamic .,. bringing hypothesis of 3 or 5 , Needs Concrete data points to support this theory . Could you list out cases for supporting this, would be curious to know what would be cases which can bring this.

The simple thing could be corruption of file or wrong logical parsing of the file. To avoid these cases we should have a strong validation here. 

Rest I don't know what code pointer you need.

Vikasht34: May I know what is wrong logical parsing of the file means? Do u have any technical example for this? What it means by logical parsing on file. Who does it ..and how file is logically parsed?? Is this happening at Operating System or Who is the Actor ..That's where we need Some Code examples to Understand that?

Writing theory does not help , Please explain in technical means .. 

As far I understand file system if file is corrput , cases will broken before this only ... it will not even work for 2 or 4 or anything else.  Again Please provide how you can read the corrupted file with some code examples so that I can learn from this.

finnroblin: @navneet1v @Vikasht34 Any concerns with the current implementation? Currently we throw if `!= 2` and `!= 4`. I did remove some duplicated parsing code which I think was the spirit of Vikash's original comment. Let me know if I should handle anything differently.

Issue #2771, opened 2025-06-23
[BUG] Efficient filtering: cannot mix parent and nested filters


uhinze:
**What is the bug?**
For ANN efficient filtering, mixing parent and child filters returns 0 results.

**How can one reproduce the bug?**
A minimal example:

```
PUT /vector-test
{
  "settings": {
    "index": {
      "number_of_shards": "1",
      "knn": "true",
      "number_of_replicas": "2"
    }
  },
  "mappings": {
    "properties": {
      "nested_chunks": {
        "type": "nested",
        "properties": {
          "chunk_embedding": {
            "type": "knn_vector",
            "dimension": 2,
            "method": {
              "engine": "faiss",
              "space_type": "l2",
              "name": "hnsw",
              "parameters": {}
            }
          },
          "chunk_type": {
            "type": "keyword"
          }
        }
      },
      "parent_type": {
        "type": "keyword"
      }
    }
  }
}
```

```
POST /_bulk
{ "index": { "_index": "vector-test", "_id": "1" } }
{ "parent_type": "foo", "nested_chunks":[{ "chunk_type": "bar", "chunk_embedding": [5.2, 4.4]}] }
{ "index": { "_index": "vector-test", "_id": "2" } }
{ "parent_type": "baz", "nested_chunks":[{ "chunk_type": "bar", "chunk_embedding": [0.9, 0.6]}, { "chunk_type": "qux", "chunk_embedding": [0.3, 0.1]}] }
```

```
GET /vector-test/_search
{
  "query": {
    "nested": {
      "inner_hits": {
        "_source": "nested_chunks.chunk_embedding"
      },
      "path": "nested_chunks",
      "query": {
        "knn": {
          "nested_chunks.chunk_embedding": {
            "k": 5,
            "vector": [
              5.1,
              5.4
            ],
            "filter": {
              "bool": {
                "must": [
                  {
                    "term": {
                      "parent_type": "foo"
                    }
                  },
                  {
                    "term": {
                      "nested_chunks.chunk_type": "bar"
                    }
                  }
                ]
              }
            }
          }
        }
      }
    }
  }
}
```

Should return 1 result, returns 0 results.

Removing either filter gives the expected result set.

[...]

PR #2990, opened 2025-11-26
Join filter clauses of nested k-NN queries to root-parent scope

### LINKED ISSUE #2222: [BUG] Filter on Parent Doc fields inside Nested knn query fails for many Query types
**What is the bug?**

When a document contains vectors in nested documents, and we perform a nested knn query with filters set on the parent documents fields, the filters can only be specific Query types (like TermQuery). If for example a Phrase Query is specified as a filter, the knn query fails to retrieve any results at all. There are several other Query types (like exists, range etc) which also fail.

1. Create a simple index with nested vector objects & text fields on the parent document to apply filters over
2. Index a couple of example documents
3. Perform a nested neural search or nested knn search with filter in the neural search query /knn query set on the parent document field - (let the term be an exact match with one of the doucments in index)
```
{
    "query_string": {
        "query": "field_standard: \"Hello World\""
    }
}
```
5. You can see that no results are returned

Evene when filters specify PhraseQuery or range query etc. the filters should be applied & results should be returned if any.

 - OS: Windows
 - Version OpenSearch 2.17
 - Plugins - Neural Serach & knn plugin

On analysis, we found that @navneet1v added the functionality to support applying filters on parent documents here: https://github.com/opensearch-project/k-NN/issues/1356

The code uses the NestedHelper.mightMatchNestedDocs  method determine whether to filter is applied on Parent doucment or nested document. Unfortunately, mightMatchNestedDocs method checks for speicifc Query types individually to see if they contain "field" & check if it is present in the parent or the nested doc. This list of Query types in not complete. Many commonly uses Query types which have "field" are missing like Phrase query, Range query etc. 

https://github.com/opensearch-project/OpenSearch/blob/f1c98a4da0cf6583212eecc9ed8ebc3cd426a918/server/src/main/java/org/opensearch/index/search/NestedHelper.java#L65

liamwhite:
### Description
Fixes parent document efficient filtering for nested k-NN queries by making the efficient filter query always operate in parent context, and joining down to children.

This is a minimal fix for the issue. It explicitly removes special casing for the efficient filter query to "select" whether it would execute in parent or child context, because that did not work correctly. Since the existing documentation suggests that the efficient filter query only operates in parent context, the efficient filter query is now always required to be in parent context, and test cases are updated accordingly. (To query in child context instead, use a nested query node within the efficient filter--this is tested by the integration cases and supported.)

### Related Issues
Resolves #2222
Unknown #2771

### Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 * Update validation for cases when k is greater than total results [#3038](https://github.com/opensearch-project/k-NN/pull/3038)
 * Add regex support to derived source transformer include / exclude check [#3031](https://github.com/opensearch-project/k-NN/pull/3031)
 * Correct ef_search parameter for Lucene engine and override mergeLeafResults to return top K results [#3037](https://github.com/opensearch-project/k-NN/pull/3037)
+* Fix efficient filtering in nested k-NN queries [#2990](https://github.com/opensearch-project/k-NN/pull/2990)
 
 ### Refactoring
 * Change ordering of build task and added tests to catch uninitialized libraries [#3024](https://github.com/opensearch-project/k-NN/pull/3024)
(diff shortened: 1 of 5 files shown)


navneet1v:
@liamwhite can you rebase the code and we can try merging the PR in 3.5 release

navneet1v (approved):

PR #3002, opened 2025-12-05
Block index creation for byte-cosinesimilarity-faiss indices

### LINKED ISSUE #2955: [BUG] Block index creation for byte-cosinesimilarity-faiss combination
**What is the bug?**
Currently when data_type is byte, spaceType is consinesimilarity and engine is faiss, an error is thrown during ingestion of document
```
"caused_by": { "type": "unsupported_operation_exception", "reason": "Byte array normalization is not supported" }
```

Ask is to block index creation instead of throwing at ingestion

kasundra07:
### Description
Block index creation for Faiss engine with cosine similarity and byte vectors due to normalization incompatibility.

### Testing
```
% curl -X PUT "http://localhost:9200/test-byte-cosine" -H "Content-Type: application/json" -d'        
{
  "settings": {
    "index.knn": true
  },
  "mappings": {
    "properties": {
      "my_vector": {
        "type": "knn_vector",
        "dimension": 3,
        "data_type": "byte",
        "space_type": "cosinesimil",
        "method": {
          "name": "hnsw",
          "engine": "faiss"
        }
      }
    }
  }
}'
{"error":{"root_cause":[{"type":"mapper_parsing_exception","reason":"Failed to parse mapping [_doc]: Faiss engine does not support cosine similarity with byte vectors. Cosine similarity requires vector normalization which is only supported for float vectors. Please use float data type or choose a different space type like l2 or inner_product."}],"type":"mapper_parsing_exception","reason":"Failed to parse mapping [_doc]: Faiss engine does not support cosine similarity with byte vectors. Cosine similarity requires vector normalization which is only supported for float vectors. Please use float data type or choose a different space type like l2 or inner_product.","caused_by":{"type":"illegal_argument_exception","reason":"Faiss engine does not support cosine similarity with byte vectors. Cosine similarity requires vector normalization which is only supported for float vectors. Please use float data type or choose a different space type like l2 or inner_product."}},"status":400}
```

### Related Issues
Resolves #2955 

### Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 * Fix NativeEngineKnnQuery to return all part results for valid totalHits in response [#2965](https://github.com/opensearch-project/k-NN/pull/2965)
 * Fix unsafe concurrent update query vector in KNNQueryBuilder [#2974](https://github.com/opensearch-project/k-NN/pull/2974)
 * Fix Backwards Compatability on Segment Merge for Disk-Based vector search [#2994](https://github.com/opensearch-project/k-NN/pull/2994) 
-* Fix score to distance calculation for inner product in faiss [#2992](https://github.com/opensearch-project/k-NN/pull/2992) 
+* Fix score to distance calculation for inner product in faiss [#2992](https://github.com/opensearch-project/k-NN/pull/2992)
+* Block index creation for Faiss engine with cosine similarity and byte vectors due to normalization incompatibility [#3002](https://github.com/opensearch-project/k-NN/pull/3002)
 
 ### Refactoring
 * Refactor to not use parallel for MMR rerank. [#2968](https://github.com/opensearch-project/k-NN/pull/2968)
--- a/src/main/java/org/opensearch/knn/index/engine/faiss/Faiss.java
+++ b/src/main/java/org/opensearch/knn/index/engine/faiss/Faiss.java
@@ -6,8 +6,10 @@
 package org.opensearch.knn.index.engine.faiss;
 
 import com.google.common.collect.ImmutableMap;
+import org.opensearch.common.ValidationException;
 import org.opensearch.knn.common.KNNConstants;
 import org.opensearch.knn.index.SpaceType;
+import org.opensearch.knn.index.VectorDataType;
 import org.opensearch.knn.index.engine.KNNLibraryIndexingContext;
 import org.opensearch.knn.index.engine.KNNMethod;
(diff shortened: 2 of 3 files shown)


src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java
```diff
@@ -526,6 +526,9 @@ private void validateFromKNNMethod(KNNVectorFieldMapper.Builder builder) {
                 throw validationException;
             }
 
+            // Validate Faiss engine specific constraints
+            validateFaissEngineConstraints(builder);
```
Vikasht34: Though This solution will work but it will be harder to scale if we keep dumping validations on validateFromKNNMethod. I think there would be two Places we can do this validation better 

1. Engine Level Validation :- we already have Faiss.java class and KnnEngine Enum , if we can restrict the validation with help of these two classes , we can all the validation regarding specific engine at places where it's functionality is there. 
2. Data Type Level Validation :- In this Class we are reading data type like this 

```
protected final Parameter<VectorDataType> vectorDataType = new Parameter<>(
            VECTOR_DATA_TYPE_FIELD,
            false,
            () -> DEFAULT_VECTOR_DATA_TYPE_FIELD,
            (n, c, o) -> VectorDataType.get((String) o),
            m -> toType(m).originalMappingParameters.getVectorDataType()
        )
```

and Parent class for this is ParametrizedFieldMapper , where we already have functinality to do setValidation , If we can write a validator function difrectly attached to vectorDataType that will also be scalable solution. 

```
public Parameter<T> setValidator(Consumer<T> validator) {
            this.validator = validator;
            return this;
        }
```

But I will prefer approach 1.

navneet1v: +1 on the comment. Dumping validation in the mapper is not a scalable idea. One way to think where to put this validation is from where this validation is coming from? Like is this a data type validation or an Engine Validation. If we see space type constraint on valid datatype(byte being a valid datatype) comes from Engine(since a engine might not support a certain datatype like Faiss may not support L1, or in this case cosine with Byte vector) we should move this validation inside FaissEngine class via KNNEngine being base class.

kasundra07: Thanks for the feedback. I agree that having engine-level validations makes it more scalable. I've moved the validation to `Faiss.java` class.

MrFlap:
CIs are failing due to the issue here: https://github.com/opensearch-project/k-NN/issues/3005

VijayanB (approved):
LGTM

Issue #2898, opened 2025-09-24
[BUG] Indexing and search fails for 16x and 8x compression for dimension <8


shatejas:
**What is the bug?**
While indexing vectors with <8 dimension Indexing has an error trace in the logs and query fails
```
Caused by: java.lang.Exception: Number of IDs does not match number of vectors
        at org.opensearch.knn.jni.FaissService.insertToBinaryIndex(Native Method) ~[?:?]
        at org.opensearch.knn.jni.JNIService.insertToIndex(JNIService.java:80) ~[?:?]
        at org.opensearch.knn.index.codec.nativeindex.MemOptimizedNativeIndexBuildStrategy.lambda$buildAndWriteIndex$1(MemOptimizedNativeIndexBuildStrategy.java:90) ~[?:?]
        at java.base/java.security.AccessController.doPrivileged(AccessController.java:319) ~[?:?]
        at org.opensearch.knn.index.codec.nativeindex.MemOptimizedNativeIndexBuildStrategy.buildAndWriteIndex(MemOptimizedNativeIndexBuildStrategy.java:89) ~[?:?]
        ... 31 more
```

**How can one reproduce the bug?**
1. Create a knn mapping with `compression_level` 16 or 8x and dimensions = 2
2. Index documents 
3. refresh
4. query

**What is the expected behavior?**
Query should not fail

**What is your host/environment?**
 - OS: MacOS
 - Version 2.19

PR #3019, opened 2025-12-10
Fix indexing for 16x and 8x compression

anntians:
### Description
Currently indexing fails for 16x and 8x compression with error `Caused by: java.lang.Exception: Number of IDs does not match number of vectors`, which is thrown at https://github.com/opensearch-project/k-NN/blob/main/jni/src/faiss_index_service.cpp#L224. 

The issue is that `(dim / 8)` at https://github.com/opensearch-project/k-NN/blob/main/jni/src/faiss_index_service.cpp#L218 is overestimating the number of bytes per vector, resulting in a mismatch between `numVectors` and `numIds`. The `dim` comes from `getDimensions` of `MultiBitScalarQuantizationState` at https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/quantization/models/quantizationState/MultiBitScalarQuantizationState.java#L182. Currently, the `getDimensions` method first aligns original vector dimension to nearest multiple of 8, then multiplies by bit quantization. This is incorrect, the method is supposed to first multiply original vector dimension by bit quantization, then align to nearest multiple of 8. See below example:

#### 2D vector with 16x compression:
Current `getDimensions` (Wrong):
- align 2D vector to 8 -> multiply 8 by 2 bit quantization = 16 -> at JNI layer `16 dim / 8` = 2 bytes per vector (wrong)

PR Fix:
- multiply 2D by 2 bit quantization = 4 -> align 4 to 8 -> at JNI layer `8 dim / 8` = 1 byte per vector (correct)

During deep dive, we also found that `getBytesPerVector` is returning bits instead of bytes, even though the code uses it as bytes, ref https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/codec/transfer/OffHeapVectorTransfer.java#L43.

So this PR updates both `getDimensions` and `getBytesPerVector` methods to fix 16x and 8x compression.

### Related Issues
Resolves https://github.com/opensearch-project/k-NN/issues/2898

### Check List

### DIFF
--- a/release-notes/opensearch-k-NN.release-notes-3.4.0.0.md
+++ b/release-notes/opensearch-k-NN.release-notes-3.4.0.0.md
@@ -22,3 +22,6 @@ Compatible with OpenSearch and OpenSearch Dashboards version 3.4.0
 ### Enhancements
 * Removed VectorSearchHolders map from NativeEngines990KnnVectorsReader ([#2948](https://github.com/opensearch-project/k-NN/pull/2948))
 * Native scoring for FP16 ([#2922](https://github.com/opensearch-project/k-NN/pull/2922))
+
+### Infrastructure
+* Include opensearchknn_simd in build configurations ([#3025](https://github.com/opensearch-project/k-NN/pull/3025))
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -130,7 +130,7 @@ fi
 # Build k-NN lib and plugin through gradle tasks
 cd $work_dir
 ./gradlew build --no-daemon --refresh-dependencies -x integTest -x test -Dopensearch.version=$VERSION -Dbuild.snapshot=$SNAPSHOT -Dbuild.version_qualifier=$QUALIFIER -Dbuild.lib.commit_patches=false -Dnmslib_simd_flags=$NMSLIB_SIMD_FLAGS
-./gradlew :buildJniLib -Pknn_libs=opensearchknn_faiss -Davx512.enabled=false -Davx512_spr.enabled=false -Davx2.enabled=false -Dbuild.lib.commit_patches=false -Dnproc.count=${NPROC_COUNT:-1} -Dbuild.snapshot=$SNAPSHOT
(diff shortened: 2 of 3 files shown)


navneet1v:
@anntians can you fix the CIs

src/main/java/org/opensearch/knn/quantization/models/quantizationState/MultiBitScalarQuantizationState.java
```diff
@@ -175,7 +175,8 @@ public int getBytesPerVector() {
         }
 
         // Calculate the number of bytes required for multi-bit quantization
-        return thresholds.length * thresholds[0].length;
+        int totalBits = thresholds.length * thresholds[0].length;
+        return (totalBits + 7) / 8;
```
jed326: Can we use constants for `7` and `8` to make it clear why we are using these "magic" values?

naveentatikonda (approved):
LGTM! Thanks for explaining the bug and the fix clearly in the description

src/main/java/org/opensearch/knn/quantization/models/quantizationState/MultiBitScalarQuantizationState.java
navneet1v: agree

src/main/java/org/opensearch/knn/common/KNNConstants.java
```diff
@@ -206,4 +206,8 @@ public class KNNConstants {
     public static final String VECTOR_FIELD_DATA_TYPE = "vector_field_data_type";
     public static final String VECTOR_FIELD_SPACE_TYPE = "vector_field_space_type";
     public static final String MMR_RERANK_CONTEXT = "mmr.rerank_context";
+
+    // Bit manipulation constants for quantization
+    public static final int BITS_PER_BYTE = 8;
```
jed326: Nit: This can just be `Byte.SIZE`, we don't need to re-define a new constant

jed326 (approved):

Issue #2937, opened 2025-10-11
[FEATURE] Efficient Filtering Optimizations


navneet1v:
## Description
As mentioned in this GH [issue comment](https://github.com/opensearch-project/k-NN/issues/2936#issuecomment-3393666442), I am creating this another issue with ideas which can help optimizing the Efficient filtering. 

Reference on how efficient filters work: https://opensearch.org/blog/efficient-filters-in-knn/

Working of Efficient filters: https://docs.opensearch.org/latest/vector-search/filter-search-knn/efficient-knn-filtering/#faiss-k-nn-filter-implementation

### Optimizations

1. When we fallback to exact search after filtered ANN search, we should not compute results for all the filtered docIds because we have already have some docIds with the scores, we should just compute the scores of remaining ones. This will at-least reduce the latency at some extent.

2. In cases of fp32, fp16, byte data type rather than iterating over VectorValues, we should use faiss Index which is already loaded in memory to do exact search(using FlatIndex present in IndexHNSW). The main benefit of this will we won't need to load the vectors in memory via .vec file since they are already in memory. We can do this for fallback exact search and threshold based exact search.

These ideas can be extended to memory optimized search too. But will not work quantized index since full precision vectors are just present in `.vec` file and not in faiss Index.

PR #3022, opened 2025-12-11
Index setting to disable exact search after ANN Search with Faiss efficient filters

### LINKED ISSUE #2936: ExactSearch brute force pathway for Faiss is causing queries to timeout
When running queries against indices hosted on OpenSearch, I have multiple queries where attempting to query n=200+ results causes the queries to timeout due to the ExactSearch pathway being triggered during knn filtering as described here https://docs.opensearch.org/latest/vector-search/filter-search-knn/efficient-knn-filtering/ — I want to request a patch to add config that can disable the brute force path and just return however many results were retrieved post-ANN.

This would require a change here https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/query/KNNWeight.java#L632 
I consistently see this message in the logs for the queries that are timing out, which notably, have very exclusive filters.

naveentatikonda:
### Description
Adds a dynamic index setting `index.knn.faiss.efficient_filter.disable_exact_search` which when set to true will disable the fallback mechanism to exact search after ANN search.

### Related Issues
Resolves #2936 
https://github.com/opensearch-project/k-NN/issues/2937

### Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,28 +4,17 @@ All notable changes to this project are documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html). See the [CONTRIBUTING guide](./CONTRIBUTING.md#Changelog) for instructions on how to add changelog entries.
 
-## [Unreleased 3.4](https://github.com/opensearch-project/k-NN/compare/main...HEAD)
+## [Unreleased 3.5](https://github.com/opensearch-project/k-NN/compare/main...HEAD)
 ### Features
-* Memory optimized search warmup ([#2954](https://github.com/opensearch-project/k-NN/pull/2954))
+
 ### Maintenance
-* Onboard to s3 snapshots ([#2943](https://github.com/opensearch-project/k-NN/pull/2943))
-* Gradle 9.2.0 and GitHub Actions JDK 25 Upgrade ([#2984](https://github.com/opensearch-project/k-NN/pull/2984))
-* Swapped macos-13 image with macos-15 ([#3014](https://github.com/opensearch-project/k-NN/pull/3014))
 
 ### Bug Fixes
-* Fix blocking old indices created before 2.18 to use memory optimized search. [#2918](https://github.com/opensearch-project/k-NN/pull/2918)
-* Fix NativeEngineKnnQuery to return all part results for valid totalHits in response [#2965](https://github.com/opensearch-project/k-NN/pull/2965)
-* Fix unsafe concurrent update query vector in KNNQueryBuilder [#2974](https://github.com/opensearch-project/k-NN/pull/2974)
-* Fix Backwards Compatability on Segment Merge for Disk-Based vector search [#2994](https://github.com/opensearch-project/k-NN/pull/2994) 
-* Fix score to distance calculation for inner product in faiss [#2992](https://github.com/opensearch-project/k-NN/pull/2992) 
 * Fix indexing for 16x and 8x compression [#3019](https://github.com/opensearch-project/k-NN/pull/3019)
 * Block index creation for Faiss engine with cosine similarity and byte vectors due to normalization incompatibility [#3002](https://github.com/opensearch-project/k-NN/pull/3002)
 
 ### Refactoring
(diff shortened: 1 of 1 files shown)


naveentatikonda:
Here are the benchmarking results using the cohere-1m dataset which shows the improvement in query latencies when the fallback mechanism to exact search has been disabled compared to the query latencies of exact search after ANN Search. But, as expected there is a huge drop in the recall when the exact search is disabled especially when the filter is too restrictive. 

| filter percent | Exact Search disabled | p50 (ms) | p90 (ms) | p99 (ms) | recall |
| --- | --- | --- | --- | --- | --- |
| 0.1 | yes | 2.85 | 3.16 | 3.49 | 0.02 |
| 0.1 | no | 3.62 | 3.94 | 4.54 | 1.0 |
| 1.0 | yes | 2.87 | 3.19 | 3.51 | 0.11 |
| 1.0 | no | 11.08 | 11.29 | 12.8 | 1.0 |
| 5.0 | yes | 3.29 | 3.59 | 3.9 | 0.31 |
| 5.0 | no | 41.35 | 42.57 | 43.23 | 0.74 |
| 10.0 | yes | 3.53 | 3.82 | 4.1 | 0.42 |
| 10.0 | no | 3.54 | 3.85 | 75.17 | 0.43 |
| 25.0 | N/A | 5.13 | 5.42 | 5.72 | 0.58 |
| 50.0 | N/A | 7.25 | 7.54 | 7.86 | 0.68 |
| 75.0 | N/A | 10.09 | 10.37 | 10.67 | 0.73 |
| 90.0 | N/A | 11.48 | 11.77 | 12.08 | 0.75 |
| 99.0 | N/A | 12.28 | 12.57 | 12.87 | 0.76 |

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -663,6 +663,16 @@ private boolean isExactSearchRequire(final LeafReaderContext context, final int
...
         }
+
+        // Disable the fallback mechanism to exact search after ANN Search when the index setting is set to true
+        if (KNNSettings.isKnnIndexFaissEfficientFilterExactSearchDisabled(knnQuery.getIndexName())) {
+            log.debug(
+                "ExactSearch is disabled after ANN Search because the index setting: {} is set to true",
+                KNNSettings.INDEX_KNN_FAISS_EFFICIENT_FILTER_DISABLE_EXACT_SEARCH
+            );
+            return false;
+        }
```
navneet1v: we should add the information about not falling on this in the explain api.

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -663,6 +663,16 @@ private boolean isExactSearchRequire(final LeafReaderContext context, final int
             log.debug("Perform exact search after approximate search since no native engine files are available");
             return true;
         }
+
+        // Disable the fallback mechanism to exact search after ANN Search when the index setting is set to true
```
navneet1v: Please add UTs for this change

CHANGELOG.md
```diff
@@ -28,3 +28,4 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Enhancements
 * Removed VectorSearchHolders map from NativeEngines990KnnVectorsReader [#2948](https://github.com/opensearch-project/k-NN/pull/2948)
 * Native scoring for FP16 [#2922](https://github.com/opensearch-project/k-NN/pull/2922)
```
navneet1v: These changes are already released. Can we remove them? I know its not part of this PR. But something to cleanup.

navneet1v (approved):

Vikasht34 (approved):

OrestisNer:
Thanks for looking into this @naveentatikonda, I think that's a very helpful improvement to avoid falling back to exact search when the cardinality is high. I was thinking that it might be useful if OpenSearch provided a cardinality threshold setting for falling back to exact search, instead of completely disabling it. For example, we have observed from our vector database that the exact search performs good enough when the cardinality is <=2000, but we would like to avoid falling back to exact search if the cardinality is >=2000. 

I would like to know your thoughts on this suggestion

OrestisNer:
Thanks @naveentatikonda for coming back to me.
I created [this draft PR](https://github.com/opensearch-project/k-NN/pull/3109) which is heavily inspired from your changes. Happy to discuss further there:)

PR #3030, opened 2025-12-15
Added check-files job to ensure that required workflows runs when there are changes right files

navneet1v:
### Description
Added check-files job to ensure that required workflows runs when there are changes right files. This change adds 2 more steps for CI.yml, test-security.yml and remote-index.yml

1. Check Files: This ensures that only a specific set of files are changed and returns if those files are changed or not.
2. Check results: This ensures that if check files is true we go and validate if the ITs and other gradle runs are successful or not. If they are not successful, error is thrown and step fails.

With this change we will not make the `Check results` mandatory to be successful before we can merge any PR.

## What was problem with older approach?
In the older approach we made few CIs required but those CIs doesn't run if a PR has release notes or changes only in `.md` files. Due to this the PRs were not able to get merged since required steps were not completed. More details can be found here: https://opensearch.slack.com/archives/C04UTNM338A/p1765568895862329

### Related Issues
NA

### Check List

### DIFF
--- a/.github/actions/check-files/action.yml
+++ b/.github/actions/check-files/action.yml
@@ -1,0 +1,30 @@
+name: 'KNN Check Files Changed'
+description: 'Check if specific files have changed for KNN repo'
+inputs:
+  files:
+    description: 'Files to check for changes'
+outputs:
+  files_changed:
+    description: 'Whether any files changed'
+    value: ${{ steps.changed-files.outputs.any_changed }}
+runs:
+  using: 'composite'
+  steps:
+    - name: Combine files
+      id: combine-files
+      shell: bash
+      run: |
+        DEFAULT_FILES="build.gradle,settings.gradle,src/**,build-tools/**,buildSrc/**,gradle/**,jni/**"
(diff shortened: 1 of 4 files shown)


Vikasht34 (commented):
This is a good CI hygiene improvement. It unblocks doc-only / release-note-only PRs that were previously stuck due to required workflows never starting. Thanks for raising this !! 

One thing I can see in this PR is CheckFiles is getting duplicated in RUN_BUILD_AND_TEST appears both in the workflow trigger (on: pull_request: paths) and again in the changed-files step. 

So if we need to make or add any new file in list we need to add this in atleast 2 files. 
This is right opportunity to use Composite Action .
https://docs.github.com/en/actions/tutorials/create-actions/create-a-composite-action
 Create a composite action that contains the path list once and Use it in any workflow. What do u think?

navneet1v:
> Create a composite action that contains the path list once and Use it in any workflow. What do u think?

All the files are not same across the actions. There is difference in the last file in each of the action. I need to see if there is a way to do that since it is not a list that can just be extended.

naveentatikonda (approved):

PR #3031, opened 2025-12-16
Regex for derived source support

### LINKED ISSUE #3029: [BUG] Derived Source Regex Bug
**What is the bug?**
Currently, when passing includes and excludes in queries, derived source only recognizes full names, and doesn't resolve regex. Additionally, if a vector fields are partially excluded, and some are included, then it will go through the derived path for all vector fields, and will only get excluded from the source at the core level, which is inefficient. 

It will make no difference for the end output in either case, but this is not correct behavior, and can be optimized here.

This bug has no effect on output, but it leads to unnecessary work being done. 
To reproduce:
1. Create an index with 2 knn fields. 
2. Index one document.
3. Fire a query with excluded wildcard that should exclude both, set breakpoint inside ShouldInject function in DerivedSourceVectorTransformaer.java. It will not properly exclude. 

The regex should be evaluated and exclude properly, and excluded fields, even when partially excluded, should not trigger the derived flow for that field.

buddharajusahil:
### Description
Add regex support to derived source transformer exclusion. Also avoid applying derived source path on excluded or non included fields:

https://github.com/opensearch-project/k-NN/issues/3029

### Related Issues
Resolves #3029 

### Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Bug Fixes
 * Fix indexing for 16x and 8x compression [#3019](https://github.com/opensearch-project/k-NN/pull/3019)
 * Block index creation for Faiss engine with cosine similarity and byte vectors due to normalization incompatibility [#3002](https://github.com/opensearch-project/k-NN/pull/3002)
+* Update validation for cases when k is greater than total results [#3038](https://github.com/opensearch-project/k-NN/pull/3038)
 * Add regex support to derived source transformer include / exclude check [#3031](https://github.com/opensearch-project/k-NN/pull/3031)
 
 ### Refactoring
--- a/src/main/java/org/opensearch/knn/index/query/memoryoptsearch/optimistic/OptimisticSearchStrategyUtils.java
+++ b/src/main/java/org/opensearch/knn/index/query/memoryoptsearch/optimistic/OptimisticSearchStrategyUtils.java
@@ -51,7 +51,7 @@ public class OptimisticSearchStrategyUtils {
      *                      used for boundary checks or optimizations
      * @return the score value that would appear at position {@code k} if all scores
      *         were globally sorted in descending order
-     * @throws IllegalArgumentException if {@code k} is less than 1 or greater than {@code totalResults}
+     * @throws IllegalArgumentException if either {@code k} or {@code totalResults} is less than 1
      */
     public static float findKthLargestScore(final List<PerLeafResult> results, final int k, final int totalResults) {
         if (totalResults <= 0) {
@@ -60,11 +60,8 @@ public class OptimisticSearchStrategyUtils {
         if (k <= 0) {
             throw new IllegalArgumentException("K must be greater than zero, got=" + k);
         }
-        if (k > totalResults) {
-            throw new IllegalArgumentException("K must be less than total results, got=" + k + ", totalResults=" + totalResults);
-        }
 
-        // If fewer than k scores, return the minimum score
+        // If fewer than or equal to k scores, return the minimum score
         if (totalResults <= k) {
             float min = Float.MAX_VALUE;
             for (final PerLeafResult result : results) {
--- a/src/test/java/org/opensearch/knn/memoryoptsearch/OptimisticSearchStrategyUtilsTests.java
+++ b/src/test/java/org/opensearch/knn/memoryoptsearch/OptimisticSearchStrategyUtilsTests.java
@@ -68,7 +68,6 @@ public class OptimisticSearchStrategyUtilsTests {
     public void testInvalidK() {
         List<PerLeafResult> results = List.of(perLeaf(1.0f, 2.0f));
         assertThrows(IllegalArgumentException.class, () -> findKthLargestScore(results, 0, 2));
-        assertThrows(IllegalArgumentException.class, () -> findKthLargestScore(results, 3, 2));
     }
 
     @Test
@@ -89,5 +88,13 @@ public class OptimisticSearchStrategyUtilsTests {
         List<PerLeafResult> results = List.of(perLeaf(10f, 9f, 8f), perLeaf(7f, 6f), perLeaf(5f, 4f));
         // Combined sorted: [10, 9, 8, 7, 6, 5, 4]
         assertEquals(4f, findKthLargestScore(results, 7, 7), 1e-6);
+    }
+
+    @Test
+    public void testTotalResultsLessThanK() {
+        List<PerLeafResult> results = List.of(perLeaf(9.0f, 7.0f), perLeaf(5.0f));
+        // When totalResults < k, should return minimum score
+        float score = findKthLargestScore(results, 5, 3);
+        assertEquals(5.0f, score, 1e-6);
     }
 }


src/main/java/org/opensearch/knn/index/codec/derivedsource/DerivedSourceVectorTransformer.java
```diff
@@ -55,12 +57,75 @@ public DerivedSourceVectorTransformer(
...
     }
 
+    /**
+     * Initialize the transformer with the fields that should be injected based on includes/excludes.
+     * This filters perFieldDerivedVectorTransformers and builds the derivedSourceVectorTransformer.
+     * Should be called once before the first call to injectVectors().
+     *
+     * @param excludes List of field patterns that should not be injected
+     */
+    public void initialize(String[] includes, String[] excludes) {
```
Vikasht34, line 70: Earlier change was not using regex pattern and was using excat , Since we started using Regex , and If customer is let's say happeing patern like _source: { excludes: ["test_vector_*"] } , regex can be costly  since this would be running per doc,  we should improve this to run pattern matching only once per filepath.

buddharajusahil: This won't happen on a per document basis, the includes excludes should be resolved in the DerivedSourceVectorTransformer at the first time document is called. Any document call after this will not go down this initialize path.

Vikasht34: You are right , I saw the code in Reader. Make Sense.

src/main/java/org/opensearch/knn/index/codec/derivedsource/DerivedSourceVectorTransformer.java
```diff
@@ -55,12 +57,75 @@ public DerivedSourceVectorTransformer(
...
+        derivedSourceVectorTransformer = XContentMapValues.transform(transformerFunctions, true);
+
+    }
+
+    private Set<String> getFieldsToExclude(String[] includes, String[] excludes) {
+        Set<String> fieldsToRemove = new HashSet<>();
+        Set<String> allFields = perFieldDerivedVectorTransformers.keySet();
+
+        // If includes are specified, start by excluding everything that doesn't match
+        if (includes != null && includes != Strings.EMPTY_ARRAY) {
```
Vikasht34: This check is not correct , **includes != Strings.EMPTY_ARRAY** checks object identity, not whether the array is empty. 

try this 
```
String[] a = new String[0];
String[] b = new String[0];

a == b                 
a == Strings.EMPTY_ARRAY 

it comes as false, 
```

src/main/java/org/opensearch/knn/index/codec/derivedsource/DerivedSourceVectorTransformer.java
```diff
@@ -55,12 +57,75 @@ public DerivedSourceVectorTransformer(
...
+            for (String fieldName : allFields) {
```
Vikasht34, line 89: Below two loops are duplicating the work , We don't need two pass for all the fields , 

Right now for everyfeild we do this 

First loop:
- Run simpleMatch() against every include pattern

Second loop:
- Run simpleMatch() against every exclude pattern.

if Field is already included why we need to run the Second Loop against all exclue pattern again for the same feild.

src/main/java/org/opensearch/knn/index/codec/derivedsource/DerivedSourceVectorTransformer.java
buddharajusahil: Makes sense, will update in next revision

src/main/java/org/opensearch/knn/index/codec/derivedsource/DerivedSourceVectorTransformer.java
buddharajusahil: So the behavior for source includes and excludes from my understanding is this:

First pass through the includes to figure out what fields to include in source. If any fields are explicitly included, then fields that are not explicitly included are practically excluded. This is what the first loop does.

Then it passes through the excluded fields to exclude from the remaining fields that were initially included from the included field.

For example a query like this:
```
{
  "query": { "match_all": {} },
  "_source": {
    "includes": ["t*"],
    "excludes": ["test"]
  }
}'
```
will include all fields that match the t* regex, but then exclude the test field, so the output source won't have the "test" field.

Vikasht34: My understanding was if field was not in Include that means it's already excluded then why we need to run second loop again on that field.  Includes act as a whitelist. Excludes only apply to fields that survive the include gate

so I have test, temp, and user as feilds.
Since we are lopping for all fields , why I need to do Regex Matching for user field on second loop , Since I know it's not in include and already excluded. 

Example 

```
"_source": {
  "includes": ["t*"],
  "excludes": ["test"]
}
``` 

@navneet1v What do u think?

src/main/java/org/opensearch/knn/index/codec/derivedsource/DerivedSourceVectorTransformer.java
buddharajusahil: This makes sense, will change in next revision.

src/main/java/org/opensearch/knn/index/codec/derivedsource/DerivedSourceVectorTransformer.java
buddharajusahil: Oh, I think I misunderstood your first comment. Yea, I think this makes sense, if a field is already not in includes and includes is explicitly defined, then we do not need to go through the excludes logic for that field.

Vikasht34 (approved):
Thanks for Addressing the Comments , Looks Good to Me !!

src/main/java/org/opensearch/knn/index/codec/derivedsource/DerivedSourceVectorTransformer.java
navneet1v: @buddharajusahil or @Vikasht34 did any one of you checked the code of the SourceFieldMapper in the OpenSeach core to confirm the behavior?

Vikasht34: I looked at this https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/mapper/SourceFieldMapper.java , 

```
final boolean filtered = CollectionUtils.isEmpty(includes) == false || CollectionUtils.isEmpty(excludes) == false;
this.filter = enabled && filtered ? XContentMapValues.filter(includes, excludes, true) : null;

```

So:

- if includes or excludes is set and _source is enabled → filter becomes a function that will filter a JSON map.

- otherwise filter == null


They behave like:

- includes is a whitelist gate: if includes is present, fields not matching it are effectively dropped.

- excludes is applied after that: it removes fields from whatever survived the include step.

Let me know if this is what I am geting

buddharajusahil: Here is the filtering logic used in XContentMapValues.java:
```
    private static Function<Map<String, ?>, Map<String, Object>> createSetBasedFilter(
        String[] includes,
        String[] excludes,
        boolean caseSensitive
    ) {
        Set<String> includeSet = (includes == null || includes.length == 0) ? null : toSet(includes, caseSensitive);
        Set<String> excludeSet = (excludes == null || excludes.length == 0) ? Collections.emptySet() : toSet(excludes, caseSensitive);

        return (map) -> {
            Map<String, Object> filtered = new HashMap<>();
            for (Map.Entry<String, ?> entry : map.entrySet()) {
                String key = entry.getKey();
                int dotPos = key.indexOf('.');
                if (dotPos > 0) {
                    key = key.substring(0, dotPos);
                }
                String k = caseSensitive ? key : key.toLowerCase(Locale.ROOT);
                if ((includeSet == null || includeSet.contains(k)) && !excludeSet.contains(k)) {
                    filtered.put(entry.getKey(), entry.getValue());
                }
            }
            return filtered;
        };
    }
```
    
In particular  :  if ((includeSet == null || includeSet.contains(k)) && !excludeSet.contains(k)) {

Based on this logic it first checks if a field is included, whether excplicitly or because includes is null, then it checks if excludes contains this field. So the behavior should match here. 

Have also manually tested these cases with text fields to confirm this behavior.

navneet1v: > Have also manually tested these cases with text fields to confirm this behavior.

Would it be good to have UTs and ITs for this change? WDYT?

navneet1v: > They behave like:
> 
> * includes is a whitelist gate: if includes is present, fields not matching it are effectively dropped.
> * excludes is applied after that: it removes fields from whatever survived the include step.
> 
> Let me know if this is what I am geting

Based on the code read and attached links I also feel this is what is happening.

buddharajusahil: I have added IT's covering all combinations of includes excludes scenarios, but will also add UT's for the specific includes / excludes function.

src/main/java/org/opensearch/knn/index/codec/derivedsource/DerivedSourceVectorTransformer.java
```diff
@@ -55,12 +56,78 @@ public DerivedSourceVectorTransformer(
...
     }
 
+    /**
+     * Initialize the transformer with the fields that should be injected based on includes/excludes.
+     * This filters perFieldDerivedVectorTransformers and builds the derivedSourceVectorTransformer.
+     * Should be called once before the first call to injectVectors().
+     *
+     * @param excludes List of field patterns that should not be injected
+     */
+    public void initialize(String[] includes, String[] excludes) {
```
navneet1v, line 70: @buddharajusahil this can be basic question but want to make sure that we have covered it. Will this initialize function will look at the fields? or we are just going to include/exclude the vector fields? Because in this PR I am not able to find how we are making sure that we are apply this transformer for vector fields only. Is that logic present somewhere else?

buddharajusahil: So this initialize function's includes and excludes parameter will contain all included and excluded fields, regardless of field type. However, initialize then calls getFieldsToExclude:

```
private Set<String> getFieldsToExclude(String[] includes, String[] excludes) {
        Set<String> fieldsToRemove = new HashSet<>();
        Set<String> allFields = perFieldDerivedVectorTransformers.keySet();
```

which sets allFields to perFieldDerivedVectorTransformers.keySet();. This keyset will only contain the vector fields that we need to apply derived source transformation for, this vector field filtering logic was in KNN9120DerivedSourceStoredFieldsFormat.fieldsReader function. So it will go through the fields from this keyset and apply the includes and excludes filter. 

Short answer: No this filtering will only be applied to vector fields, which we filtered out in KNN9120DerivedSourceStoredFieldsFormat.fieldsReader.

navneet1v (approved):

PR #3035, opened 2025-12-18
Derived source dynamic template fix

### LINKED ISSUE #3012: [BUG] KNN 512-dimensional vector has broken value with derived_source: true and dynamic template
**What is the bug?**

KNN vector field has value `1`

**Edit**: I have minimally recreated the bug here: https://github.com/devk-scayle/os-knn-derived-source-bug

We upgraded to OS3.3 (from 2.19) and did reindexing. At some point we ended up with large number of KNN vectors with value `1`. These documents had expected 512 dimensional vector in the old index.

The mapping is done using dynamic template:
```
{
    "knn_vector_template": {
        "path_match": "similar_products_vector.*.clip_vit_base_patch32",
        "mapping": {
            "dimension": 512,
            "method": {
                "engine": "faiss",
                "space_type": "cosinesimil",
                "name": "hnsw",
                "parameters": {
                    "ef_search": 128,
                    "ef_construction": 128,
                    "m": 16
                }
            },
            "type": "knn_vector"
        }
    }
}
```

Further reindexing also failed with expected reason:
```
{
    "type": "illegal_argument_exception",
    "reason": "Vector dimension mismatch. Expected: 512, Given: 1"
}
```

Reindexing should not break existing vectors. Fields with knn_vector type should not accept broken values

_note_: manually updating the document kkn value to `1` fails, as expected.

AWS managed OpenSearch 3.3 with KNN plugin.

Proof of correct mappings, but broken value. See `similar_products_vector.2088.clip_vit_base_patch32`:

Image

I have also tried manually setting the new `derived_source` to `false` and could no longer reproduce the issue. We had this set to default (`false` pre-GA on 2.19, `true` since upgrade to 3.3).
I cannot 100% confirm this is the cause, but so far I could not reproduce it with it set to `false`, while it happened regularly with it set to `true`

**Edit:**

Upon further research, it looks like it this happens when _target_ index has `derived_source: true`. The source index is not the issue.

Image

It also looks like the issue might be the dynamic template. We have multiple knn graps (`similar_products_vector.1856`, `similar_products_vector.2167`, etc) and to the best of my findings: either all values under a graph are broken (1856 in the above screenshot) or all are correct (2167 in the above screenshot). Though, mapping for both shows knn_field as I would expect

buddharajusahil:
### Description
Fixes a bug where bulk indexing with dynamic templates and knn.derived_source.enabled=true caused vectors to be incorrectly returned as 1 instead of the actual vector array.

The Problem:
When bulk indexing documents that create new dynamic field mappings (e.g., similar_products_vector.*.clip_vit_base_patch32), only the first document's vector was correctly reconstructed. All subsequent documents returned the mask value 1 instead of the actual vector.

Root Cause:
Segment attributes (which track which fields need derived source reconstruction) were written in fieldsWriter() at segment creation time. At that point, only the first document's dynamic mapping existed in mapperService.fieldTypes(). Subsequent documents' dynamic mappings were created as they were parsed, after the segment attributes had already been written with an incomplete field list.

The Fix:
Move segment attribute writing from fieldsWriter() to finish(), which is called after all documents have been written to the segment and all dynamic mappings exist.
Add explicit segment attribute handling in merge() for merge operations, since finish() is not called during merge (the delegate's finish is called instead).

Changes:
KNN10010DerivedSourceStoredFieldsFormat: Removed early attribute writing from fieldsWriter(), now passes segmentInfo and mapperService to writer.
KNN10010DerivedSourceStoredFieldsWriter: Added attribute writing in finish() for normal indexing and in merge() for merge operations.
KNN10010DerivedSourceStoredFieldsReader: Added getDerivedVectorFields() getter to support field collection during merge.
Added regression test testDerivedSource_withDynamicTemplates_andBulkIndexing to prevent future regressions.

### Related Issues
Resolves #3012 

### Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased 3.6](https://github.com/opensearch-project/k-NN/compare/main...HEAD)
 ### Features
+* Support Lucene BBQ Flat for 1 bit [#3154](https://github.com/opensearch-project/k-NN/pull/3154)
 
 ### Maintenance
 * Improve unit tests by tightening asserts [#3112](https://github.com/opensearch-project/k-NN/pull/3112)
--- a/src/main/java/org/opensearch/knn/common/KNNConstants.java
+++ b/src/main/java/org/opensearch/knn/common/KNNConstants.java
@@ -17,6 +17,7 @@ public class KNNConstants {
     public static final String NAME = "name";
     public static final String PARAMETERS = "parameters";
     public static final String METHOD_HNSW = "hnsw";
+    public static final String METHOD_FLAT = "flat";
     public static final String TYPE = "type";
     public static final String TYPE_NESTED = "nested";
     public static final String PATH = "path";
@@ -104,6 +105,8 @@ public class KNNConstants {
     public static final double MAXIMUM_CONFIDENCE_INTERVAL = 1.0;
     public static final String LUCENE_SQ_BITS = "bits";
     public static final int LUCENE_SQ_DEFAULT_BITS = 7;
+    public static final String MAX_CONNECTIONS = "max_connections";
+    public static final String BEAM_WIDTH = "beam_width";
 
     // nmslib specific constants
     @Deprecated(since = "2.19.0", forRemoval = true)
--- a/src/main/java/org/opensearch/knn/index/codec/KNN1040BasePerFieldKnnVectorsFormat.java
+++ b/src/main/java/org/opensearch/knn/index/codec/KNN1040BasePerFieldKnnVectorsFormat.java
@@ -1,0 +1,110 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.knn.index.codec;
+
+import lombok.extern.log4j.Log4j2;
+import org.apache.lucene.codecs.KnnVectorsFormat;
+import org.apache.lucene.codecs.perfield.PerFieldKnnVectorsFormat;
+import org.opensearch.index.mapper.MapperService;
+import org.opensearch.knn.index.codec.nativeindex.NativeIndexBuildStrategyFactory;
+import org.opensearch.knn.index.engine.CodecFormatResolver;
+import org.opensearch.knn.index.engine.KNNEngine;
+import org.opensearch.knn.index.engine.KNNMethodContext;
+import org.opensearch.knn.index.mapper.KNNMappingConfig;
+import org.opensearch.knn.index.mapper.KNNVectorFieldType;
+
+import java.util.Map;
+import java.util.Optional;
+import java.util.function.Supplier;
+
+import static org.opensearch.knn.common.KNNConstants.BEAM_WIDTH;
+import static org.opensearch.knn.common.KNNConstants.MAX_CONNECTIONS;
+
+/**
+ * Standalone registry-based base class for per-field KNN vectors format resolution,
+ * used by current and future codecs (KNN1040+).
+ *
+ * <p>
+ * This class extends {@link PerFieldKnnVectorsFormat} directly and delegates format
+ * construction to {@link CodecFormatResolver} instances for each engine type.
+ * </p>
+ */
+@Log4j2
+public abstract class KNN1040BasePerFieldKnnVectorsFormat extends PerFieldKnnVectorsFormat {
+
+    private final Optional<MapperService> mapperService;
+    private final int defaultMaxConnections;
+    private final int defaultBeamWidth;
+    private final Supplier<KnnVectorsFormat> defaultFormatSupplier;
+    private final CodecFormatResolver luceneFormatResolver;
+    private final CodecFormatResolver nativeFormatResolver;
+    private final NativeIndexBuildStrategyFactory nativeIndexBuildStrategyFactory;
+
+    protected KNN1040BasePerFieldKnnVectorsFormat(
+        Optional<MapperService> mapperService,
+        int defaultMaxConnections,
+        int defaultBeamWidth,
+        Supplier<KnnVectorsFormat> defaultFormatSupplier,
+        CodecFormatResolver luceneFormatResolver,
+        CodecFormatResolver nativeFormatResolver,
+        NativeIndexBuildStrategyFactory nativeIndexBuildStrategyFactory
+    ) {
+        this.mapperService = mapperService;
+        this.defaultMaxConnections = defaultMaxConnections;
+        this.defaultBeamWidth = defaultBeamWidth;
+        this.defaultFormatSupplier = defaultFormatSupplier;
+        this.luceneFormatResolver = luceneFormatResolver;
+        this.nativeFormatResolver = nativeFormatResolver;
+        this.nativeIndexBuildStrategyFactory = nativeIndexBuildStrategyFactory;
+    }
+
+    @Override
+    public KnnVectorsFormat getKnnVectorsFormatForField(final String field) {
+        if (isKnnVectorFieldType(field) == false) {
+            log.debug(
+                "Initialize KNN vector format for field [{}] with default params [{}] = \"{}\" and [{}] = \"{}\"",
+                field,
+                MAX_CONNECTIONS,
+                defaultMaxConnections,
+                BEAM_WIDTH,
+                defaultBeamWidth
+            );
+            return defaultFormatSupplier.get();
+        }
+        KNNVectorFieldType mappedFieldType = (KNNVectorFieldType) mapperService.orElseThrow(
+            () -> new IllegalStateException(
+                String.format("Cannot read field type for field [%s] because mapper service is not available", field)
+            )
+        ).fieldType(field);
+
+        final KNNMappingConfig knnMappingConfig = mappedFieldType.getKnnMappingConfig();
+        if (knnMappingConfig.getModelId().isPresent()) {
+            return nativeFormatResolver.resolve();
+        }
+
+        final KNNMethodContext knnMethodContext = knnMappingConfig.getKnnMethodContext()
+            .orElseThrow(() -> new IllegalArgumentException("KNN method context cannot be empty"));
+        nativeIndexBuildStrategyFactory.setKnnLibraryIndexingContext(knnMappingConfig.getKnnLibraryIndexingContext());
+        final KNNEngine engine = knnMethodContext.getKnnEngine();
+        final Map<String, Object> params = knnMethodContext.getMethodComponentContext().getParameters();
+
+        if (engine == KNNEngine.LUCENE) {
+            return luceneFormatResolver.resolve(field, knnMethodContext, params, defaultMaxConnections, defaultBeamWidth);
+        }
+
+        // All native engines to use nativeFormatResolver
+        return nativeFormatResolver.resolve();
+    }
+
+    @Override
+    public int getMaxDimensions(String fieldName) {
+        return getKnnVectorsFormatForField(fieldName).getMaxDimensions(fieldName);
+    }
+
+    private boolean isKnnVectorFieldType(final String field) {
+        return mapperService.isPresent() && mapperService.get().fieldType(field) instanceof KNNVectorFieldType;
+    }
+}
--- a/src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/KNN1040Codec.java
+++ b/src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/KNN1040Codec.java
@@ -18,7 +18,6 @@ import org.opensearch.index.mapper.MapperService;
 import org.opensearch.knn.index.codec.KNN10010Codec.KNN10010DerivedSourceStoredFieldsFormat;
 import org.opensearch.knn.index.codec.KNN80Codec.KNN80CompoundFormat;
 import org.opensearch.knn.index.codec.KNN80Codec.KNN80DocValuesFormat;
-import org.opensearch.knn.index.codec.KNN9120Codec.KNN9120PerFieldKnnVectorsFormat;
 import org.opensearch.knn.index.codec.derivedsource.DerivedSourceReadersSupplier;
 
 import java.util.Optional;
@@ -26,7 +25,7 @@ import java.util.Optional;
 public class KNN1040Codec extends FilterCodec {
     private static final String NAME = "KNN1040Codec";
     public static final Codec DEFAULT_DELEGATE = new Lucene104Codec();
-    private static final PerFieldKnnVectorsFormat DEFAULT_KNN_VECTOR_FORMAT = new KNN9120PerFieldKnnVectorsFormat(Optional.empty());
+    private static final PerFieldKnnVectorsFormat DEFAULT_KNN_VECTOR_FORMAT = new KNN1040PerFieldKnnVectorsFormat(Optional.empty());
 
     private final PerFieldKnnVectorsFormat perFieldKnnVectorsFormat;
     private final StoredFieldsFormat storedFieldsFormat;
--- a/src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/KNN1040PerFieldKnnVectorsFormat.java
+++ b/src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/KNN1040PerFieldKnnVectorsFormat.java
@@ -1,0 +1,110 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.knn.index.codec.KNN1040Codec;
+
+import org.apache.lucene.backward_codecs.lucene99.Lucene99RWHnswScalarQuantizedVectorsFormat;
+import org.apache.lucene.codecs.KnnVectorsFormat;
+import org.apache.lucene.codecs.lucene99.Lucene99HnswVectorsFormat;
+import org.apache.lucene.codecs.lucene104.Lucene104ScalarQuantizedVectorsFormat;
+import org.opensearch.common.collect.Tuple;
+import org.opensearch.index.mapper.MapperService;
+import org.opensearch.knn.index.KNNSettings;
+import org.opensearch.knn.index.SpaceType;
+import org.opensearch.knn.index.codec.KNN1040BasePerFieldKnnVectorsFormat;
+import org.opensearch.knn.index.codec.KnnVectorsFormatContext;
+import org.opensearch.knn.index.codec.LuceneVectorsFormatType;
+import org.opensearch.knn.index.codec.KNN9120Codec.KNN9120HnswBinaryVectorsFormat;
+import org.opensearch.knn.index.codec.nativeindex.NativeIndexBuildStrategyFactory;
+import org.opensearch.knn.index.codec.params.KNNScalarQuantizedVectorsFormatParams;
+import org.opensearch.knn.index.codec.params.KNNVectorsFormatParams;
+import org.opensearch.knn.index.engine.KNNEngine;
+import org.opensearch.knn.index.engine.faiss.FaissCodecFormatResolver;
+import org.opensearch.knn.index.engine.lucene.LuceneCodecFormatResolver;
+
+import java.util.Map;
+import java.util.Optional;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.function.Function;
+
+/**
+ * Per-field KNN vectors format for the KNN1040 codec. Uses {@link Lucene99HnswVectorsFormat}
+ * for HNSW, {@link Lucene99RWHnswScalarQuantizedVectorsFormat} for scalar quantization (to
+ * preserve the {@code confidenceInterval} parameter), and
+ * {@link Lucene104ScalarQuantizedVectorsFormat} with {@code SINGLE_BIT_QUERY_NIBBLE} encoding
+ * for the flat BBQ method.
+ */
+public class KNN1040PerFieldKnnVectorsFormat extends KNN1040BasePerFieldKnnVectorsFormat {
+
+    private static final Tuple<Integer, ExecutorService> DEFAULT_MERGE_THREAD_COUNT_AND_EXECUTOR_SERVICE = Tuple.tuple(1, null);
+
+    public KNN1040PerFieldKnnVectorsFormat(final Optional<MapperService> mapperService) {
+        this(mapperService, new NativeIndexBuildStrategyFactory());
+    }
+
+    public KNN1040PerFieldKnnVectorsFormat(
+        final Optional<MapperService> mapperService,
+        NativeIndexBuildStrategyFactory nativeIndexBuildStrategyFactory
+    ) {
+        super(
+            mapperService,
+            Lucene99HnswVectorsFormat.DEFAULT_MAX_CONN,
+            Lucene99HnswVectorsFormat.DEFAULT_BEAM_WIDTH,
+            Lucene99HnswVectorsFormat::new,
+            new LuceneCodecFormatResolver(buildLuceneFormatResolvers()),
+            new FaissCodecFormatResolver(mapperService, nativeIndexBuildStrategyFactory),
+            nativeIndexBuildStrategyFactory
+        );
+    }
+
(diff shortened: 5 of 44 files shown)


src/main/java/org/opensearch/knn/index/codec/KNN10010Codec/KNN10010DerivedSourceStoredFieldsWriter.java
```diff
@@ -88,21 +111,63 @@ public void writeField(FieldInfo info, StoredFieldDataInput value) throws IOExce
...
-        // In case of backwards compatibility, with old segments, we need to perform a non-optimal merge. Basically, it
-        // will repopulate each source and then inject the vector and then remove it. This allows us to migrate
-        // segments from filter approach to mask approach
         if (KNN9120DerivedSourceStoredFieldsReader.doesMergeContainLegacySegments(mergeState)) {
             return super.merge(mergeState);
         }
 
-        // We wrap the segments to avoid injecting back vectors and then removing. If this is not done, then we will
-        // inject and then just write to disk potentially.
+        Set<String> vectorFields = new HashSet<>();
```
Vikasht34, line 176: Nit:- We are using **HashSet**  in Merge and **mapperService.fieldTypes()** in finish , both can make non-deterministic order which might be difficult in debugging , Can we sort the field lists before writing segment attributes.

buddharajusahil: This makes sense, will add.

navneet1v (approved):

src/main/java/org/opensearch/knn/index/codec/KNN10010Codec/KNN10010DerivedSourceStoredFieldsWriter.java
```diff
@@ -43,12 +57,22 @@ public class KNN10010DerivedSourceStoredFieldsWriter extends StoredFieldsWriter
...
+        this.mapperService = mapperService;
+
+        List<String> vectorFieldTypes = new ArrayList<>();
+        for (MappedFieldType fieldType : mapperService.fieldTypes()) {
+            if (fieldType instanceof KNNVectorFieldType knnVectorFieldType) {
+                if (IndexUtil.isDerivedEnabledForField(knnVectorFieldType, mapperService)) {
+                    vectorFieldTypes.add(fieldType.name().toLowerCase());
+                }
+            }
+        }
```
sohami: this will not fix the issue, as writer is created only on first document so mapper service has knowledge about first document field only. For rest of the document adding new fields won't be present in mapper service yet and it will not create transformers for those fields. We will need to do it at per document level unfortunately and add transformers there for dynamic template case. 

Basically in our test, can we somehow validate that indeed mask is written to the `_source` for the vector fields because from search query it will always return the vector field irrespective of it is masked in source or not. Or another work around if we cannot validate this easily is use integer values for each dimension in vector and I think if it is masked it will return floats back. So we can assert if returned vectors have integer values that means it is not masked and stored as is ?

buddharajusahil: So actually original masking is done in derived source pre-index operation listener. So in all our IT's we have, the behavior will be correct, in that all fields will get masked. The masking done in the fields writer, as far as I know, is redundant, but just used a safety net as jack pointed out in his original PR: https://github.com/opensearch-project/k-NN/pull/2704

Which makes sense because otherwise the original customer queries would not have contained the mask as stated in the bug, it would include the full vector (since the vector would never have gotten masked in the first place). The case here that would fail would be that if writeField is called before preIndex, on documents that are not already masked, where all fields that are "to be" added don't yet exist in the mapperService, which I am not sure is a possible case.

sohami: Right, so we are changing that behavior here by relying only on the masking in the Listener. Original linked PR had a caveat where author was not sure if listener based approach is handling all the cases correctly or not. So either we completely remove the masking logic from codec and rely only on the listener side of the things or we keep both in correct way which is not true in current state of the PR.

>The case here that would fail would be that if writeField is called before preIndex, on documents that are not already masked, where all fields that are "to be" added don't yet exist in the mapperService, which I am not sure is a possible case.

I think meta question is if there are cases when `writeField` is called without `preIndex` getting invoked. Both of these get exercised as part of indexing operation which can happen either as part of bulk request or as part of recovery (translog replay or sequence number based recovery in peer recovery). In all these ingestion path, `writeField` and `preIndex` both will be invoked. `preIndex` gets invoked as part of the [IndexShard private method index](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/shard/IndexShard.java#L1281) which takes care of indexing documents into the Lucene layer. With this we can remove the transformation from codec layer but there is another big caveat. If we keep transformation only in listener layer, and lets say there is any failure in the listener and it [throws exception, that will be caught and swallowed](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/shard/IndexingOperationListener.java#L113) and request path will flow to the codec layer. So if we remove the transformation logic from codec layer we will not mask vector fields for those operations. With this I don't see that we can escape the code layer transformation here.

Vikasht34: Nice , In Depth Explanation @sohami , learned from your comment !! But Still I have question , Are u suggesting document level or What is your suggestion?

buddharajusahil: Got it. Yea I think the codec layer mask's original intention was for merge cases, as this will go through the writeField flow without masking from preIndex, but even then it might be redundant. My thought was if any merge case existed where we needed codec masking, then we would anyways have all the fields in mapperservice anyways.

However the preindex failure case probably makes the codec dynamic masking unavoidable.

sohami: >But Still I have question , Are u suggesting document level or What is your suggestion?

Immediate fix is document level, don't see any other way out of it. If there is a way to differentiate dynamic template vs static mapping probably we can limit the overhead (if any) to dynamic template case itself.

As a follow-up we should work on https://github.com/opensearch-project/k-NN/issues/3058 which **may** remove the need of doing in codec or listener layer completely.

Vikasht34: Thanks Make Sense

sohami: >Got it. Yea I think the codec layer mask's original intention was for merge cases, as this will go through the writeField flow without masking from preIndex

@buddharajusahil  But for merge cases, `_source` is already having mask for vector values. So we should not require to mask it again or apply transformers again right ? But yes segmentInfo needs to be updated with the list of fields for the newly created segment

buddharajusahil: @sohami I think for standard merge, documents will already have the mask from indexing so codec masking is not necessary. But I think the original PR was cautious about some edge cases. For example if a legacy KNN9120 segment is being merged with a 10010 segment, as far as I can tell, codec masking might be necessary, as KNN9120 codec filtered vectors completely out of its source in derived source path. On merge, the old segment source is read, which will include the entire vector.

sohami: @buddharajusahil [Setting is static](https://docs.opensearch.org/latest/vector-search/settings/#index-settings) so ideally it is not intended to change on existing index. So the case of old segments without mask should not exist. But I guess transformation during `merge` time can be separate discussion, please create a issue if transformation is happening today during merges. We can discuss and see if we need to optimize that as well.

buddharajusahil: @sohami Right the derived source setting isn't changing, I am talking about legacy KNN9120 format using derived source. In 9120 format, in derived source path, we just remove the vector entirely, and on read we put it back, there is no concept of masking, only vector field filtering. If we merge this with a 10010 format segment, we will need to apply the mask to the legacy segment in codec level. This is more about bwc.

sohami: So is 10010 format handling the case where a mask is not present for a document to still fetch the vector value from reader which may or may not be present (e.g. a doc without vector field) ? I think it is best to discuss the merge part separately as we need to consider all the different possibilities here. For the scope of this PR we can keep the merge behavior same as today.

buddharajusahil: @sohami Agreed, for this PR I will add the dynamic masking. 

Not sure I fully understand the question in the comment. The 10010 format reader finds the fields to inject from the segment info. If a documents source does not have any of the vector fields to inject, then it is treated as a no-op.

sohami: Ignore my previous case, that was assuming a scenario when an older segment in an index on latest codec is not merged so it will not have masks for vector fields. How is that handled with new codec derived readers, however in that case for older segments, reader will still be from older codec so should be fine. 

Performing transformation during merge for bwc is fine. Thanks for sharing that use case. Wondering if we should have logic to ignore transformations when segments from 10010 codecs are merged so all the new indices atleast does not have to again do transformation in merge phase. It will be worth it if there is any performance gain with it.

src/main/java/org/opensearch/knn/index/codec/KNN10010Codec/KNN10010DerivedSourceStoredFieldsWriter.java
```diff
@@ -88,62 +112,219 @@ public void writeField(FieldInfo info, StoredFieldDataInput value) throws IOExce
...
+     */
+    private boolean containsUnmaskedVectors(Map<String, Object> source) {
+        for (Map.Entry<String, Object> entry : source.entrySet()) {
+            Object value = entry.getValue();
+
+            // Check if this is an unmasked vector (List of Numbers with more than 1 element)
+            if (value instanceof List<?> list && !list.isEmpty()) {
+                Object first = list.get(0);
+                if (first instanceof Number && list.size() > 1) {
+                    return true;
```
sohami: I would recommend not to have utility method like this where we are assuming based on the field value type if it is a vector field or not. For reference, users can pass list of integers for an integer field type as well which will be treated as a multi-valued field.

src/main/java/org/opensearch/knn/index/codec/KNN10010Codec/KNN10010DerivedSourceStoredFieldsWriter.java
```diff
@@ -88,62 +112,219 @@ public void writeField(FieldInfo info, StoredFieldDataInput value) throws IOExce
...
+                "Encountered NotXContent while deserializing _source field. Instead found String: [{}]",
+                new String(bytesRef.bytes, 0, Math.min(bytesRef.bytes.length, 512)),
+                e
+            );
+            return;
+        }
+
+        // Check if source is already masked from preindex masking path.
+        // If it is, we do not need to remask the source here
+        if (!containsUnmaskedVectors(mapTuple.v2())) {
```
sohami: We can remove this to preserve the older behavior of always masking in codec for now

src/main/java/org/opensearch/knn/index/codec/KNN10010Codec/KNN10010DerivedSourceStoredFieldsWriter.java
```diff
@@ -20,41 +22,63 @@
...
+            if (fieldType instanceof KNNVectorFieldType knnVectorFieldType) {
+                if (IndexUtil.isDerivedEnabledForField(knnVectorFieldType, mapperService)) {
+                    knownVectorFields.add(fieldType.name().toLowerCase());
+                }
+            }
+        }
+
+        if (!knownVectorFields.isEmpty()) {
+            vectorMask = buildMask(knownVectorFields);
         }
```
sohami: we can move the check for empty collections inside `buildMask` which can return `null` or  valid transformation function.

src/main/java/org/opensearch/knn/index/codec/KNN10010Codec/KNN10010DerivedSourceStoredFieldsWriter.java
```diff
@@ -88,62 +112,219 @@ public void writeField(FieldInfo info, StoredFieldDataInput value) throws IOExce
...
+        }
+
+        // This path is only if source is not masked from preindex path
+        // Eg: preindex failed, merging legacy segment with modern segment - need to mask
+        // Find KNN fields in this document
+        if (vectorMask != null) {
+            Map<String, Object> filteredSource = vectorMask.apply(mapTuple.v2());
+
+            // Check if existing masking worked (no unmasked vectors remain)
+            if (!containsUnmaskedVectors(filteredSource)) {
```
sohami: 1. I think instead of handling like this where we iterate on `_source` to check if mask is applied on some field or not, can we use `mapperService` here to iterate and find all the fields with derived source enabled. Update `vectorMask` to initialize the masking lambda function for all the discovered fields and then just apply the mask on the current document ? Basically move the logic from constructor [here](https://github.com/opensearch-project/k-NN/pull/3035/files#diff-28fdc09fa0ef5c86ccc4f8a2ef9712c745485a41c0ff423ecafee3481ab32416R72) to this function. That will be much stronger and safer check based on fieldType and will be contained to this `writeField` method as well.
2. We should definitely do some perf benchmark to see if there is any performance impact on the ingestion and merges with this change.

src/main/java/org/opensearch/knn/index/codec/KNN10010Codec/KNN10010DerivedSourceStoredFieldsWriter.java
```diff
@@ -157,8 +338,55 @@ public void finishDocument() throws IOException {
...
+                }
+
+                boolean isNested = mapperService.documentMapper().mappers().getNestedScope(fieldType.name()) != null;
+                if (isNested) {
+                    nestedVectorFieldTypes.add(fieldType.name());
+                } else {
+                    vectorFieldTypes.add(fieldType.name());
+                }
+            }
+        }
```
sohami: I think we can maintain both nested and non-nested derived field types as part of evaluating in `writeField` itself which can be later used here to update the segment info. Can we double confirm if writers will be recreated for each segment or reused such we will need to reset these states related to derivedFields or vectorMask

src/main/java/org/opensearch/knn/index/codec/KNN10010Codec/KNN10010DerivedSourceStoredFieldsWriter.java
```diff
@@ -88,62 +112,219 @@ public void writeField(FieldInfo info, StoredFieldDataInput value) throws IOExce
...
         for (int i = 0; i < mergeState.storedFieldsReaders.length; i++) {
             mergeState.storedFieldsReaders[i] = KNN10010DerivedSourceStoredFieldsReader.wrapForMerge(mergeState.storedFieldsReaders[i]);
         }
-        return delegate.merge(mergeState);
+
+        int result = delegate.merge(mergeState);
+
+        if (!vectorFields.isEmpty()) {
+            List<String> sortedVectorFields = new ArrayList<>(vectorFields);
+            Collections.sort(sortedVectorFields);
```
sohami: any reason to sort the fields here and in other places ?

src/main/java/org/opensearch/knn/index/codec/KNN10010Codec/KNN10010DerivedSourceStoredFieldsWriter.java
```diff
@@ -88,62 +112,219 @@ public void writeField(FieldInfo info, StoredFieldDataInput value) throws IOExce
...
+            DerivedSourceSegmentAttributeParser.addDerivedVectorFieldsSegmentInfoAttribute(
+                segmentInfo,
+                new ArrayList<>(vectorFields),
```
sohami: can we just pass set into `addDerivedVectorFieldsSegmentInfoAttribute` utility instead of converting it to list ? `Set` seems to provide better control to avoid duplicate fields from being written to segment info

src/main/java/org/opensearch/knn/index/codec/KNN10010Codec/KNN10010DerivedSourceStoredFieldsWriter.java
buddharajusahil: Sure, makes sense.

src/main/java/org/opensearch/knn/index/codec/KNN10010Codec/KNN10010DerivedSourceStoredFieldsWriter.java
buddharajusahil: I think this was to address @Vikasht34 comment that it would make debugging easier, but on second thought, I don't think it is applicable in this case.

src/main/java/org/opensearch/knn/index/codec/KNN10010Codec/KNN10010DerivedSourceStoredFieldsWriter.java
buddharajusahil: Makes sense.

src/main/java/org/opensearch/knn/index/codec/KNN10010Codec/KNN10010DerivedSourceStoredFieldsWriter.java
buddharajusahil: Writers are created when segment is created and are not reused across multiple segments. Sounds good.

src/main/java/org/opensearch/knn/index/codec/KNN10010Codec/KNN10010DerivedSourceStoredFieldsWriter.java
buddharajusahil: Makes sense, sure.

src/main/java/org/opensearch/knn/index/codec/KNN10010Codec/KNN10010DerivedSourceStoredFieldsWriter.java
buddharajusahil: Got it.

navneet1v:
@buddharajusahil anything else left on this PR any reason why this PR is not getting merged?

src/main/java/org/opensearch/knn/index/codec/KNN10010Codec/KNN10010DerivedSourceStoredFieldsWriter.java
```diff
@@ -88,62 +99,150 @@ public void writeField(FieldInfo info, StoredFieldDataInput value) throws IOExce
...
+                    } else {
+                        knownVectorFields.add(fieldName.toLowerCase());
+                    }
+                }
+            }
+        }
+        // Build mask from both sets combined
+        Set<String> allFields = new HashSet<>();
+        allFields.addAll(knownVectorFields);
+        allFields.addAll(knownNestedVectorFields);
```
sohami: nit: this can be populated in above loop itself.

src/main/java/org/opensearch/knn/index/codec/KNN10010Codec/KNN10010DerivedSourceStoredFieldsWriter.java
```diff
@@ -88,62 +99,150 @@ public void writeField(FieldInfo info, StoredFieldDataInput value) throws IOExce
...
             MediaType actualContentType = mapTuple.v1();
             XContentBuilder builder = MediaTypeRegistry.contentBuilder(actualContentType, bStream).map(filteredSource);
             builder.close();
-            BytesReference bytesReference = bStream.bytes();
-            delegate.writeField(fieldInfo, bytesReference.toBytesRef());
+            delegate.writeField(fieldInfo, bStream.bytes().toBytesRef());
             return;
         }
+
         delegate.writeField(fieldInfo, bytesRef);
```
sohami:
```suggestion
        if (!Objects.equals(fieldInfo.name, SourceFieldMapper.NAME)) {
            delegate.writeField(fieldInfo, bytesRef);
            return;
        }
        
        // Check if mapperService has new fields (cheap integer comparison)
        int currentFieldCount = getFieldCount();
        if (currentFieldCount != lastKnownFieldCount) {
            refreshKnownVectorFields();
            lastKnownFieldCount = currentFieldCount;
        }
    
         // if we don't have any vector fields just pass it to delegate writer
         if (vectorMask == null) {
             delegate.writeField(fieldInfo, bytesRef);
             return;
         }

         // Reference: https://github.com/opensearch-project/OpenSearch/blob/2.18.0/server/src/main/java/org/opensearch/index/mapper/SourceFieldMapper.java#L322
            
         // In the Index operation listener, preindex, we apply the mask operation. We apply it again for safety
         // here because writeField may not always follow the preindex step, so it might have the vector (think
         // merge). This may be overly cautious and we can remove/optimize it in the future. For now, its a safety net.
        Tuple<? extends MediaType, Map<String, Object>> mapTuple;
        try {
            mapTuple = XContentHelper.convertToMap(
                BytesReference.fromByteBuffer(ByteBuffer.wrap(bytesRef.bytes, bytesRef.offset, bytesRef.length)),
                true,
                MediaTypeRegistry.JSON
            );
        } catch (NotXContentException e) {
            // If the content is not XContent, skip writing altogether. See:
            // https://github.com/opensearch-project/k-NN/issues/2880
            log.warn(
                "Encountered NotXContent while deserializing _source field. Instead found String: [{}]",
                new String(bytesRef.bytes, 0, Math.min(bytesRef.bytes.length, 512)),
                e
            );
            return;
        }

        // Apply mask on vector fields in parsed source (idempotent - safe to apply even if already masked)
        Map<String, Object> filteredSource = vectorMask.apply(mapTuple.v2());
        BytesStreamOutput bStream = new BytesStreamOutput();
        MediaType actualContentType = mapTuple.v1();
        XContentBuilder builder = MediaTypeRegistry.contentBuilder(actualContentType, bStream).map(filteredSource);
        builder.close();
        delegate.writeField(fieldInfo, bStream.bytes().toBytesRef());
```

sohami (commented):
@buddharajusahil Did you get chance to run perf benchmark with this change ? Curious to see if we are observing any latency impact here.

src/main/java/org/opensearch/knn/index/codec/KNN10010Codec/KNN10010DerivedSourceStoredFieldsWriter.java
buddharajusahil: Sure

buddharajusahil:
@navneet1v @sohami Need review / approval again after merge conflict resolution.

Vikasht34 (dismissed):

src/main/java/org/opensearch/knn/index/codec/KNN10010Codec/KNN10010DerivedSourceStoredFieldsWriter.java
```diff
@@ -90,66 +106,158 @@ public void writeField(FieldInfo info, StoredFieldDataInput value) throws IOExce
...
 
+    /**
+     * Builds a mask transform function for the given fields.
+     */
+    private Function<Map<String, Object>, Map<String, Object>> buildMask(Set<String> fields) {
+        if (fields == null || fields.isEmpty()) {
+            return null;
+        }
+        return XContentMapValues.transform(
+            fields.stream().collect(Collectors.toMap(k -> k, k -> (Object o) -> o == null ? o : MASK)),
```
shatejas, line 117: nit: I know its just copied but everytime I look at it its confusing so just a suggestion

```suggestion
            fields.stream().collect(Collectors.toMap(field -> , field -> (Object vector) -> vector == null ? vector : MASK)),
```

Vikasht34 (approved):

PR #3037, opened 2025-12-18
Correct ef_search parameter for Lucene engine and reduce to top K

### LINKED ISSUE #2940: [BUG] Correct EF_Search parameter is not passed for Lucene Engine when ef_search not set in method_parameters
## Description
For Lucene engine we use Lucene library to do the K-NN search. Lucene library doesn't take `ef_search` as a parameter and only takes `k` as a parameter. Due to this in k-NN query when `ef_search` is not passed as method parameters we just pass `k` which acts as `ef_search` and may lead to drop in recall and also less number of results. Ref: https://github.com/opensearch-project/k-NN/issues/2881

## Solution
Ideally we should update the logic for Lucene engine and start doing this:

1. If ef_search is not passed in method parameters then
   1. Check if ef_search is set in index_mappings if not then use default ef_search based in index version else return the ef_search set on the index mappings

2. ef_search = max(k, ef_search)
3. Once Lucene query is completed reduce the number of results to final topK results.

This will ensure that we are always passing right value of ef_search parameter for Lucene engine.

naveentatikonda:
### Description
This PR fixes a recall degradation issue in k-NN search (especially with small k values) by implementing dynamic ef_search resolution for the Lucene engine.

1. Determine the ef_search value using the following priority order and use it as luceneK
    a. Use ef_search from method parameters if specified in the query
    b. Otherwise, use ef_search from index setting (`knn.algo_param.ef_search`)
    c. If neither exists, fall back to default ef_search value based on index version
2. Calculate the effective Lucene k value as max(k, luceneK)
3. Override mergeLeafResults to return top K results    

### Related Issues
Resolves #2940 

### Check List

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -97,9 +97,10 @@ jobs:
 
 
       - name: Upload Coverage Report
-        uses: codecov/codecov-action@v4
+        uses: codecov/codecov-action@v5
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
+          use_pypi: true
 
   Build-k-NN-MacOS:
     strategy:


src/main/java/org/opensearch/knn/index/query/common/DocAndScoreQuery.java
```diff
@@ -55,6 +55,10 @@ public Explanation explain(LeafReaderContext context, int doc) {
                 if (found < 0) {
                     return Explanation.noMatch("not in top " + k);
                 }
+                if (knnWeight == null) {
```
VijayanB: either update the method doc or add comment here on why you are doing this

src/main/java/org/opensearch/knn/index/query/lucene/LuceneEngineKnnVectorQuery.java
```diff
@@ -47,33 +54,30 @@ public Weight createWeight(IndexSearcher searcher, ScoreMode scoreMode, float bo
...
         }
         return weight;
     }
 
-    @Override
-    public String toString(String s) {
-        return luceneQuery.toString();
-    }
+    private Query reduceToTopK(Query query, IndexSearcher searcher) throws IOException {
+        if (luceneK == k || query instanceof ExpandNestedDocsQuery) {
```
VijayanB: add method doc or explain the logic here

src/main/java/org/opensearch/knn/index/query/lucene/LuceneEngineKnnVectorQuery.java
```diff
@@ -24,10 +28,13 @@
  * of Lucene based k-NN queries.
  */
 @AllArgsConstructor
+@EqualsAndHashCode(callSuper = false)
 @Log4j2
 public class LuceneEngineKnnVectorQuery extends Query {
     @Getter
     private final Query luceneQuery;
+    private final int luceneK;
```
0ctopus13prime: NIT : Could you add one line comment please? I think our memory will vague, not knowing the difference between k and luceneK

src/main/java/org/opensearch/knn/index/query/lucene/LuceneEngineKnnVectorQuery.java
```diff
@@ -47,33 +54,34 @@ public Weight createWeight(IndexSearcher searcher, ScoreMode scoreMode, float bo
...
-    public void visit(QueryVisitor queryVisitor) {
-        queryVisitor.visitLeaf(this);
+        // Skip reducing to top-k in two cases:
+        // 1. When luceneK equals k (no reduction needed)
+        // 2. When query is ExpandNestedDocsQuery (reducing would exclude required child documents)
+        if (luceneK == k || query instanceof ExpandNestedDocsQuery) {
+            return query;
+        }
+
+        TopDocs topK = searcher.search(query, k);
```
0ctopus13prime: I think `search` method will eventually create a new task and delegate it to another thread. And seems like it will need too much effort compared to what we're trying to achieve here -- reduce it to top-k docs.
Can we just get iterator from it, and reduce the results directly on it?
```
DocIdSetIterator iterator = ...;  // Get it from Query
iterator.nextDoc();
TopKnnCollector collector = new TopKnnCollector(k, MAX_INT);
while (iterator.doc() != NO_MORE_DOCS) {
  collector.collect(iterator.doc(), scorer.score());
  iterator.nextDoc();
}
return QueryUtils.getInstance().createDocAndScoreQuery(indexReader, collector.topDocs());
```

naveentatikonda: Ack and as discussed updated the code. Thanks for calling it out

0ctopus13prime (approved):
LGTM!
Please add one IT for this. It can be in the next PR.

src/main/java/org/opensearch/knn/index/query/common/DocAndScoreQuery.java
```diff
@@ -55,6 +55,13 @@ public Explanation explain(LeafReaderContext context, int doc) {
                 if (found < 0) {
                     return Explanation.noMatch("not in top " + k);
                 }
+
+                // This is copied from org.apache.lucene.search.DocAndScoreQuery#createWeight
+                // for LuceneEngineKnnVectorQuery where knnWeight is null
+                if (knnWeight == null) {
```
navneet1v: what is the case when KNNWeight is null?

src/main/java/org/opensearch/knn/index/query/lucene/LuceneEngineKnnVectorQuery.java
```diff
@@ -24,10 +31,13 @@
  * of Lucene based k-NN queries.
  */
 @AllArgsConstructor
+@EqualsAndHashCode(callSuper = false)
```
navneet1v: any reason for making callSuper as false?

src/main/java/org/opensearch/knn/index/query/lucene/LuceneEngineKnnVectorQuery.java
```diff
@@ -47,33 +57,46 @@ public Weight createWeight(IndexSearcher searcher, ScoreMode scoreMode, float bo
...
+        for (LeafReaderContext context : searcher.getIndexReader().leaves()) {
+            Scorer scorer = weight.scorer(context);
+            if (scorer != null) {
+                DocIdSetIterator iterator = scorer.iterator();
+                int doc;
+                while ((doc = iterator.nextDoc()) != DocIdSetIterator.NO_MORE_DOCS) {
+                    collector.collect(doc + context.docBase, scorer.score());
+                }
+            }
+        }
```
navneet1v: should we not parallelize this using the executor present in the IndexSearcher?

src/main/java/org/opensearch/knn/index/query/lucene/LuceneEngineKnnVectorQuery.java
```diff
@@ -75,18 +75,21 @@ private Query reduceToTopK(Query query, IndexSearcher searcher, ScoreMode scoreM
...
-            Scorer scorer = weight.scorer(context);
-            if (scorer != null) {
-                DocIdSetIterator iterator = scorer.iterator();
-                int doc;
-                while ((doc = iterator.nextDoc()) != DocIdSetIterator.NO_MORE_DOCS) {
-                    collector.collect(doc + context.docBase, scorer.score());
-                }
+        // Collect results from all leaves in parallel
+        List<LeafReaderContext> leaves = searcher.getIndexReader().leaves();
+        List<Map<Integer, Float>> leafResults = QueryUtils.getInstance().doSearch(searcher, leaves, weight);
```
0ctopus13prime: Hmm, do we really need parallelizing results reduction?? I'm bit skeptical about it. 
Taking top-k within a single thread will always give the better results than parallel map-reduce.
With below sample example, I'm always getting better latency in single thread.

```
package main.java.kdy;

import java.util.ArrayList;
import java.util.Comparator;
import java.util.List;
import java.util.PriorityQueue;
import java.util.Random;
import java.util.concurrent.ExecutorService;
import java.util.concurrent.Executors;
import java.util.concurrent.Future;

public class ParallelReduction {
    static class DocScore {
        final int doc;
        final float score;

        DocScore(int doc, float score) {
            this.doc = doc;
            this.score = score;
        }
    }

    public static List<DocScore> topK(int N, int numThreads, int k, ExecutorService pool, boolean print) throws Exception {
        // 1. Generate random doc IDs + scores
        int[] docs = new int[N];
        float[] scores = new float[N];
        Random rand = new Random();

        for (int i = 0; i < N; i++) {
            docs[i] = i;
            scores[i] = rand.nextFloat();
        }

        long s = System.nanoTime();

        List<Future<List<DocScore>>> futures = new ArrayList<>();

        int chunkSize = (N + numThreads - 1) / numThreads;

        // 2. Fork p threads, each computing local top-k
        for (int t = 0; t < numThreads; t++) {
            final int start = t * chunkSize;
            final int end = Math.min(start + chunkSize, N);

            if (start >= end) {
                break;
            }

            futures.add(pool.submit(() -> {
                PriorityQueue<DocScore> pq =
                    new PriorityQueue<>(Comparator.comparingDouble(ds -> ds.score));

                for (int i = start; i < end; i++) {
                    if (pq.size() < k) {
                        pq.add(new DocScore(docs[i], scores[i]));
                    } else if (scores[i] > pq.peek().score) {
                        pq.poll();
                        pq.add(new DocScore(docs[i], scores[i]));
                    }
                }

                return new ArrayList<>(pq);
            }));
        }

        // 3. Merge local top-k results into final top-k
        PriorityQueue<DocScore> finalPQ =
            new PriorityQueue<>(Comparator.comparingDouble(ds -> ds.score));

        for (Future<List<DocScore>> f : futures) {
            for (DocScore ds : f.get()) {
                if (finalPQ.size() < k) {
                    finalPQ.add(ds);
                } else if (ds.score > finalPQ.peek().score) {
                    finalPQ.poll();
                    finalPQ.add(ds);
                }
            }
        }

        // Return sorted descending
        List<DocScore> result = new ArrayList<>(finalPQ);
        result.sort((a, b) -> Float.compare(b.score, a.score));

        long e = System.nanoTime();
        long took = e - s;
        if (print) {
            System.out.println("Took " + took + " nano, " + (took / 1e6) + "ms");
        }

        return result;
    }


    public static List<DocScore> justTopK(int N, int k, boolean print) {
        Random rand = new Random();
        int[] docs = new int[N];
        float[] scores = new float[N];

        for (int i = 0; i < N; i++) {
            docs[i] = i;
            scores[i] = rand.nextFloat();
        }

        long s = System.nanoTime();

        PriorityQueue<DocScore> pq =
            new PriorityQueue<>(Comparator.comparingDouble(ds -> ds.score));

        // Generate docs + scores and keep top-k
        for (int i = 0; i < N; i++) {
            int doc = docs[i];
            float score = scores[i];

            if (pq.size() < k) {
                pq.add(new DocScore(doc, score));
            } else if (score > pq.peek().score) {
                pq.poll();
                pq.add(new DocScore(doc, score));
            }
        }

        // Sort result descending
        List<DocScore> result = new ArrayList<>(pq);
        result.sort((a, b) -> Float.compare(b.score, a.score));

        long e = System.nanoTime();
        long took = e - s;
        if (print) {
            System.out.println("Took " + took + " nano, " + (took / 1e6) + "ms");
        }

        return result;
    }

    // Example usage
    public static void main(String[] args) throws Exception {
        int N = 500;
        int numThreads = 10;
        int k = 30;

        ExecutorService pool = Executors.newFixedThreadPool(numThreads);
        for (int i = 0 ; i < 100 ; ++i) {
            List<DocScore> topK = topK(N, numThreads, k, pool, false);
            justTopK(N, k, false);
        }

        List<DocScore> topK = topK(N, numThreads, k, pool, true);
        justTopK(N, k, true);
    }
}

```

src/main/java/org/opensearch/knn/index/query/common/DocAndScoreQuery.java
naveentatikonda: It will be null for all the lucene engine queries

src/main/java/org/opensearch/knn/index/query/lucene/LuceneEngineKnnVectorQuery.java
naveentatikonda: Here Query is an abstract class and has no instance fields and meaningful state, so I set callsuper as false.

src/main/java/org/opensearch/knn/index/query/lucene/LuceneEngineKnnVectorQuery.java
naveentatikonda: Got it, we shouldn't parallelize because here we will be reducing top efSearch results among all the segments into topK results, which is usually very small so better to process sequentially than to parallelize. Also, validated with a small benchmarking test. Thanks!

0ctopus13prime (approved):

0ctopus13prime:
Hi @naveentatikonda, LGTM and approved the PR, but I think I found the better solution and I think we can revisit later.
The idea is to override `mergeLeafResults`  to take top-k directly without introducing additional phase.
https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java#L358-L372

```
  /**
   * Merges all segment-level kNN results to get the index-level kNN results.
   *
   * <p>The default implementation delegates to {@link TopDocs#merge(int, TopDocs[])} to find the
   * overall top {@link #k}, which requires input results to be sorted.
   *
   * <p>This method is useful for reading and / or modifying the final results as needed.
   *
   * @param perLeafResults array of segment-level kNN results.
   * @return index-level kNN results (no constraint on their ordering).
   * @lucene.experimental
   */
  protected TopDocs mergeLeafResults(TopDocs[] perLeafResults) {
    return TopDocs.merge(k, perLeafResults);
  }
``` 

So we have four classes need to be expanded for this result reduction and we won't need reiterating for reduction no more.
- KnnFloatVectorQuery
- KnnByteVectorQuery
- DiversifyingChildrenFloatKnnVectorQuery
- DiversifyingChildrenByteKnnVectorQuery


Pseudocode:
```
class XXX extends YYY {  // YYY can be one of four listed above ^^
  int k;  // `k` that user want results to be included.

  public XXX(int luceneK, int k, ...) {
    super(luceneK, ...);  // Pass augmented `k` to super to user bigger search budget
    this.k = k;
  }

  ....

    protected TopDocs mergeLeafResults(TopDocs[] perLeafResults) {
        // Just take top-k in here
        return TopDocs.merge(k, perLeafResults);
    }
}
```

```
// Then, in KNNQueryFactory#getKnnVectorQuery we just need to replace those four classes. That's it.
    private static Query getKnnVectorQuery(
        final String fieldName,
        final float[] floatQueryVector,
        final byte[] byteQueryVector,
        final int luceneK, <-------------------- augmented k
        final int k,
        final Query filterQuery,
        final BitSetProducer parentFilter,
        final boolean expandNested,
        @NonNull final VectorDataType vectorDataType
    ) {
        if (parentFilter == null) {
                           vvvvvvvvvvvv OSKnnFloatVectorQuery inherits KnnFloatVectorQuery
                ? new OSKnnFloatVectorQuery(fieldName, floatQueryVector, luceneK, k, filterQuery)
                                                                                                                        ^^^^^^ pass augmented k
                          vvvvvvvvvvvv OSKnnByteVectorQuery inherits KnnByteVectorQuery
                : new OSKnnByteVectorQuery(fieldName, byteQueryVector, luceneK, k, filterQuery);
                                                                                                                       ^^^^^^ pass augmented k
        }
        // If parentFilter is not null, it is a nested query. Therefore, we delegate creation of query to {@link
        // NestedKnnVectorQueryFactory}
        // which will create query to dedupe search result per parent so that we can get k parent results at the end.
            ? NestedKnnVectorQueryFactory.createNestedKnnVectorQuery(
                fieldName,
                floatQueryVector,
                luceneK, <------------ augmented k
                k,
                filterQuery,
                parentFilter,
                expandNested
            )
            : NestedKnnVectorQueryFactory.createNestedKnnVectorQuery(
                fieldName,
                byteQueryVector,
                luceneK, <------------ augmented k
                k,
                filterQuery,
                parentFilter,
                expandNested
            );
    }

    // In nested case
    public static Query createNestedKnnVectorQuery(
        final String fieldName,
        final float[] vector,
        final int luceneK, <-------------------- augmented k
        final int k,
        final Query filterQuery,
        final BitSetProducer parentFilter,
        final boolean expandNestedDocs
    ) {
        if (expandNestedDocs) {
            return new ExpandNestedDocsQuery.ExpandNestedDocsQueryBuilder().internalNestedKnnVectorQuery(
                new InternalNestedKnnFloatVectoryQuery(fieldName, vector, filterQuery, k, parentFilter)
            ).queryUtils(QueryUtils.getInstance()).build();
        }
        vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv OSDiversifyingChildrenFloatKnnVectorQuery inherits DiversifyingChildrenFloatKnnVectorQuery
        return new OSDiversifyingChildrenFloatKnnVectorQuery(fieldName, vector, filterQuery, luceneK, k, parentFilter);
                                                                                                                                                            ^^^^^^ pass augmented k
    }

    public static Query createNestedKnnVectorQuery(
        final String fieldName,
        final byte[] vector,
        final int luceneK, <-------------------- augmented k
        final int k,
        final Query filterQuery,
        final BitSetProducer parentFilter,
        final boolean expandNestedDocs
    ) {
        if (expandNestedDocs) {
            return new ExpandNestedDocsQuery.ExpandNestedDocsQueryBuilder().internalNestedKnnVectorQuery(
                new InternalNestedKnnByteVectoryQuery(fieldName, vector, filterQuery, k, parentFilter)
            ).queryUtils(QueryUtils.getInstance()).build();
        }
        vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv OSDiversifyingChildrenByteKnnVectorQuery inherits DiversifyingChildrenByteKnnVectorQuery
        return new OSDiversifyingChildrenByteKnnVectorQuery(fieldName, vector, filterQuery, luceneK, k, parentFilter);
                                                                                                                                                            ^^^^^^ pass augmented k
    }
```

naveentatikonda:
> Hi @naveentatikonda, LGTM and approved the PR, but I think I found the better solution and I think we can revisit later. The idea is to override `mergeLeafResults` to take top-k directly without introducing additional phase. e.g. giving augmented k (luceneK) to Lucene query and let it use it for search budget, and use `k` when finalizing results. With this, we don't need to reiterate results and put it back to DocsAndQuery no more. I think this will be faster than this raised version. Only downside is bit cumbersome, we need to expand 4 classes. https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java#L358-L372

Makes sense, will address it in the next PR. Thanks!

src/main/java/org/opensearch/knn/index/query/lucene/LuceneEngineKnnVectorQuery.java
naveentatikonda: @navneet1v as Dooyong called out and I have mentioned in this [thread](https://github.com/opensearch-project/k-NN/pull/3037#discussion_r2652166162), we will be only reducing top luceneK(efSearch) results among all the segments into topK results, which is usually very small so better to process sequentially than to parallelize. Pls let us know if there is any other edge case that we are missing

navneet1v: @naveentatikonda I understand that it will be small, but when CSS is enabled then we are breaking the contract to process the segments in parallel. Hence I am little skeptical on not processing them in parallel.

navneet1v: Think of cases when K becomes as high as 10K across different shards and large datasets.

src/main/java/org/opensearch/knn/index/query/lucene/LuceneEngineKnnVectorQuery.java
navneet1v: but if you call super will that create any problem?

src/main/java/org/opensearch/knn/index/query/KNNQueryFactory.java
```diff
@@ -128,25 +129,38 @@ public static Query create(CreateQueryRequest createQueryRequest) {
...
+    // 3. If neither exists, fall back to default ef_search value based on index version
+    private static int getEfSearch(final Map<String, ?> methodParameters, final String indexName) {
+        if (methodParameters != null && methodParameters.containsKey(METHOD_PARAMETER_EF_SEARCH)) {
+            return (Integer) methodParameters.get(METHOD_PARAMETER_EF_SEARCH);
+        }
+
+        // Returns ef_search from index setting (knn.algo_param.ef_search) or
+        // falls back to default ef_search value based on index version
+        return KNNSettings.getEfSearchParam(indexName);
+    }
```
navneet1v: shouldn't this be part of any utility, rather than a private function in this class? Also there should be a similar function already present

src/main/java/org/opensearch/knn/index/query/lucene/LuceneEngineKnnVectorQuery.java
```diff
@@ -47,33 +57,46 @@ public Weight createWeight(IndexSearcher searcher, ScoreMode scoreMode, float bo
...
             profiler.pollLastElement();
         }
         return weight;
     }
 
-    @Override
-    public String toString(String s) {
-        return luceneQuery.toString();
-    }
+    private Query reduceToTopK(Query query, IndexSearcher searcher, ScoreMode scoreMode, float boost) throws IOException {
```
navneet1v: Should we move this function in ResultUtil class as there is already a method to reduce the results to topk. can we not reuse the same method?

src/main/java/org/opensearch/knn/index/query/lucene/LuceneEngineKnnVectorQuery.java
naveentatikonda: It doesn't break anything right now but if Lucene adds state to Query class, it may cause issues in the future. But, I think we always explicitly define equality at the concrete class level. This was the code before refactoring with annotation which was validating just based on field luceneQuery

```
@Override
    public boolean equals(Object o) {
        if (this == o) return true;
        if (o == null || getClass() != o.getClass()) return false;
        LuceneEngineKnnVectorQuery otherQuery = (LuceneEngineKnnVectorQuery) o;
        return luceneQuery.equals(otherQuery.luceneQuery);
    }

    @Override
    public int hashCode() {
        return luceneQuery.hashCode();
    }

File: src/main/java/org/opensearch/knn/index/query/lucene/LuceneEngineKnnVectorQuery.java (hunk 7 above)
Comment (naveentatikonda) [reply src/main/java/org/opensearch/knn/index/query/lucene/LuceneEngineKnnVectorQuery.java]: Got it, but practically we often might not use such a big number for K. Also, @0ctopus13prime was saying that with his test that he ran below sequential search performs better upto 100K. @0ctopus13prime can you pls confirm ?

Also, @navneet1v I need to refactor this logic and change my implementation in the next PR based on this [suggestion](https://github.com/opensearch-project/k-NN/pull/3037#issuecomment-3701456911) which will cleanup all this logic

File: src/main/java/org/opensearch/knn/index/query/KNNQueryFactory.java (hunk 9 above)
Comment (naveentatikonda) [reply src/main/java/org/opensearch/knn/index/query/KNNQueryFactory.java]: Ack, refactored into HyperparametersUtil class. We already have a method but it just returns efSearch based on index version

File: src/main/java/org/opensearch/knn/index/query/lucene/LuceneEngineKnnVectorQuery.java (hunk 7 above)
Comment (0ctopus13prime) [reply src/main/java/org/opensearch/knn/index/query/lucene/LuceneEngineKnnVectorQuery.java]: Hi @navneet1v 
I don't think we should make it parallel. Underlying the search is finished already and what we're doing is results reduction and not search. Hence, we are not breaking the contract (concurrent segment 'search'). Everything we need is already in memory. Even K becomes as high as 10K across different shards and large dataset, parallel results reduction is always slower than single thread loop. 

Also in my finding (https://github.com/opensearch-project/k-NN/pull/3037#issuecomment-3701456911), we don't even need this extra results reduction (which might be dealt with in the next PR). Turns out there's the extension point for customizing results reduction in Lucene and we can use that.

File: src/main/java/org/opensearch/knn/index/query/lucene/LuceneEngineKnnVectorQuery.java (hunk 10 above)
Comment (naveentatikonda) [reply src/main/java/org/opensearch/knn/index/query/lucene/LuceneEngineKnnVectorQuery.java]: We can't reuse it as we also need to iterate over the scorer and fetch the results here for every leaf. Anyways all this logic will be cleaned in the next PR with this [approach](https://github.com/opensearch-project/k-NN/pull/3037#issuecomment-3701456911) so I'm not refactoring it for now

Comment (navneet1v) [reply src/main/java/org/opensearch/knn/index/query/lucene/LuceneEngineKnnVectorQuery.java]: @naveentatikonda if we are going to refactor this whole logic in the next PR what I am not able to understand is why we need to review this PR? Why cannot we just implement once and review everything in 1 PR? 

Please let me know if I am missing something here.

File: src/main/java/org/opensearch/knn/index/query/lucene/LuceneEngineKnnVectorQuery.java (hunk 7 above)
Comment (navneet1v) [reply src/main/java/org/opensearch/knn/index/query/lucene/LuceneEngineKnnVectorQuery.java]: > I don't think we should make it parallel. Underlying the search is finished already and what we're doing is results reduction and not search. Hence, we are not breaking the contract (concurrent segment 'search')

This is a valid point. But a point to consider here is what we are calling as Search. Overfetching the results and reducing them is part of search as per my point of view. But if there is a way to avoid the whole reduction that is pretty good I would say and we should aim for that. 

Nevertheless, I would say if we have a better way to make the change we should directly aim for that. Having intermediate things in between I am not very much comfortable as we will be sitting on a better version of code for sometime. Please let me know if I am going in a different direction here.

File: src/main/java/org/opensearch/knn/index/query/lucene/LuceneEngineKnnVectorQuery.java (hunk 6 above)
Comment (navneet1v) [reply src/main/java/org/opensearch/knn/index/query/lucene/LuceneEngineKnnVectorQuery.java]: [Recommendation]
We should improve on this and call super here, as that will be the right thing to do.

**Review by 0ctopus13prime — DISMISSED** [review]

File: src/main/java/org/opensearch/knn/index/query/KNNQueryFactory.java [hunk 11]
```diff
@@ -128,20 +128,17 @@ public static Query create(CreateQueryRequest createQueryRequest) {
...
         int overSampledK = k;
         boolean needsRescore = shouldRescore(rescoreContext);
         if (needsRescore) {
             // Will always do shard level rescoring whenever rescore is required.
             overSampledK = rescoreContext.getFirstPassK(k, false, getDimension(vector, byteVector));
         }
-        int luceneK = requestEfSearch == null ? overSampledK : Math.max(overSampledK, requestEfSearch);
+
+        int luceneK = Math.max(overSampledK, IndexHyperParametersUtil.getHNSWEFSearchValue(methodParameters, indexName));
         log.debug("Creating Lucene k-NN query for index: {}, field:{}, k: {}", indexName, fieldName, luceneK);
```
navneet1v: can you log the luceneK also in this debug logs.

src/main/java/org/opensearch/knn/index/query/lucenelib/OSKnnFloatVectorQuery.java
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * OpenSearch wrapper around Lucene's KnnByteVectorQuery that customizes
+ * result merging to honor the original k parameter for queries with float vectors.
+ *
+ * <p>This wrapper ensures that when merging results from multiple segments, only the top k
+ * documents are returned, maintaining consistency with OpenSearch's k-NN query behavior.
+ */
+public final class OSKnnFloatVectorQuery extends KnnFloatVectorQuery {
+    int k;
```
navneet1v: any reason for making this as package private?

navneet1v (dismissed):
Minor comments. Overall looks good to me

src/main/java/org/opensearch/knn/index/query/KNNQueryFactory.java
naveentatikonda: We are actually logging luceneK as the value for K here. Should we log both actual k and luceneK ?

navneet1v: Yes lets log both of them.

src/main/java/org/opensearch/knn/index/query/lucenelib/OSKnnFloatVectorQuery.java
naveentatikonda: Oh missed it. Will make it private and add final

0ctopus13prime (approved):

PR #3048, opened 2026-01-05
Efficient filters Optimization: Skip Bitset Creation in Exact Search

### LINKED ISSUE #2981: [Enhancement] Remove conversion of FilterIdIterater to Bitset and DISI in case of Efficient filter falling back to exact search
### Description
Currently when doing Efficient filters based search either we do exact search if thresholds are correct or we do ANN search with filter bitset. In the first case when we fallback to exact search since we need DISI to iterate over filterIds we don't need to convert the FiteredDocIdSetIterator to Bitset and then again to DISI. Since exact search works on DISI. We can avoid this conversion of FiteredDocIdSetIterator to Bitset. Since converting a FiteredDocIdSetIterator to Bitset literally mean accessing all docIds and then set the bits.

Ref: 
Conversion to Bitset: 

https://github.com/opensearch-project/k-NN/blob/5388ea3171664929c78536775d5dda337efef9ab/src/main/java/org/opensearch/knn/index/query/KNNWeight.java#L392-L393

Then in Lucene:

https://github.com/apache/lucene/blob/branch_10_3/lucene/core/src/java/org/apache/lucene/util/BitSet.java#L33-L43

https://github.com/apache/lucene/blob/branch_10_3/lucene/core/src/java/org/apache/lucene/util/BitSet.java#L118-L123

Currently I am not how much of an improvement we will get, it can be a micro optimization, but this is something to consider for large filterIds

finnroblin:
### Description
Efficient Filters optimization that avoids creating a bitset in the case where it is not needed. In the KNNWeight code path there is a call to exact search if there are fewer documents that pass through the filter than the `k` value. In this case we bypass approximate search entirely. Approximate search requires a `BitSet` but exact search does not; the filter iterator can be passed directly to the `doExactSearch` method. Making a `BitSet` is expensive since it requires a linear iteration through the documents that pass the filters.

This PR adds the optimization to avoid creating a BitSet in the case where approximate search is unnecessary.

It works by using the `filterIterator.cost()` method to estimate the cardinality of the filter. If we want to get the cardinality exactly I believe we will need to materialize the BitSet and calculate the number of set bits (if a bit is set than the document at that location passes through the filter). I'd like some feedback in the PR about how to validate that the `.cost()` estimation function is a good enough approximation for different types of filter iterators. Below is a table summarizing the behavior of `cost()` for some types of Lucene iterators.

### cost() Return Values by Iterator Type

| Iterator Type              | cost() Returns                                             | Source                  |
| -------------------------- | ---------------------------------------------------------- | ----------------------- |
| TermQuery                  | te.docFreq() - exact count from posting list metadata      | O(1) lookup             |
| BooleanQuery (MUST/FILTER) | min(cost of all MUST/FILTER clauses)                       | Minimum of sub-costs    |
| BooleanQuery (SHOULD)      | Sum of SHOULD costs, capped by min required cost           | Computed from sub-costs |
| ConjunctionDISI (AND)      | lead1.cost() - cost of the lead (cheapest) iterator        | Overestimate            |
| DisjunctionDISI (OR)       | Sum of all sub-iterator costs                              | Sum of sub-costs        |
| PointRangeQuery            | values.estimateDocCount(visitor) - estimated from BKD tree | Computed lazily         |
| BitSetIterator             | Stored cost value (typically cardinality)                  | Passed at construction  |
| FilteredDocIdSetIterator   | innerIter.cost() - delegates to wrapped iterator

### Related Issues
Resolves #2981

### Check List
- [ ] New functionality has been documented. NA — this PR is an optimization. 
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md). NA — this PR is an optimization. 
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose). NA — this PR is an optimization. 

### DIFF
--- a/IteratorCostAccuracyResults.md
+++ b/IteratorCostAccuracyResults.md
@@ -22,6 +22,19 @@ re are the results:
 - **PointRangeQuery**: Significantly overestimates, especially at low cardinality (BKD tree estimation)
 - **BitSetIterator**: Always exact (cardinality passed at construction)
 
+Other mixed (boolean AND query): 
+> Results for mixed queries:
+> 
+> | Test | cost() | actual | accuracy | exact? |
+> |------|--------|--------|----------|--------|
+> | **Mixed-PointAndTerm-Low** (PointRange leads) | 1000 | 5 | 200x | ❌ |
+> | **Mixed-TermLeads-Low** (Term leads) | 5 | 5 | 1.00 | ✅ |
+> | **Mixed-TermAndPoint** (Term leads) | 10 | 10 | 1.00 | ✅ |
+> 
+> **Key insight**: BooleanMUST uses `min(cost)` of all clauses. When TermQuery has lower cost, it becomes the lead iterator and `cost()` is exact. When PointRangeQuery has lower estimated cost (even if inaccurate), it leads and the overestimate propagates.
+> [Tool uses: none]
+
+
 Full test results: 
 ./gradlew :test --tests "org.opensearch.knn.index.query.IteratorCostAccuracyTests" --info | pbcopy
 
--- a/src/main/java/org/opensearch/knn/index/query/KNNWeight.java
+++ b/src/main/java/org/opensearch/knn/index/query/KNNWeight.java
@@ -41,6 +41,11 @@ import org.opensearch.knn.indices.ModelMetadata;
 import org.opensearch.knn.indices.ModelUtil;
 import org.opensearch.knn.plugin.stats.KNNCounter;
 
+import org.apache.lucene.search.BooleanClause;
+import org.apache.lucene.search.BooleanQuery;
+import org.apache.lucene.search.Query;
+import org.apache.lucene.search.TermQuery;
+
 import java.io.IOException;
 import java.util.Arrays;
 import java.util.List;
@@ -340,9 +345,24 @@ public abstract class KNNWeight extends Weight {
         }
 
         final DocIdSetIterator filterIterator = scorer.iterator();
-        // filterCardinality uses the cost() estimate instead of the exact cardinality to save work.
-        // Without the cost() estimate we would have to materialize a bitset and compute the accepted bits in linear time.
-        final int filterCardinality = Math.toIntExact(Math.min(filterIterator.cost(), maxDoc));
+
+        // Determine if cost() provides accurate cardinality based on filter query type
+        final boolean canUseCostForCardinality = canUseCostForCardinality(filterWeight.getQuery());
+
+        final int filterCardinality;
+        BitSet filterBitSet = null;
+
+        if (canUseCostForCardinality) {
+            // TermQuery and BooleanQuery with MUST clauses provide accurate cost()
+            filterCardinality = Math.toIntExact(Math.min(filterIterator.cost(), maxDoc));
+        } else {
+            // For other query types (e.g., PointRangeQuery), materialize BitSet for accurate cardinality
+            final StopWatch stopWatch = startStopWatch(log);
+            filterBitSet = createBitSet(context, filterIterator, liveDocs, maxDoc);
+            filterCardinality = filterBitSet.cardinality();
+            stopStopWatchAndLog(log, stopWatch, "FilterBitSet creation", knnQuery.getShardId(), segmentName, knnQuery.getField());
+        }
+
         if (filterCardinality == 0) {
             return PerLeafResult.EMPTY_RESULT;
         }
@@ -360,16 +380,24 @@ public abstract class KNNWeight extends Weight {
          * because nested doc handling requires the BitSet to be materialized.
          */
         if (isFilteredExactSearchPreferred(filterCardinality) && knnQuery.getParentsFilter() == null) {
+            if (filterBitSet != null) {
+                // BitSet already created, use it
+                final BitSetIterator docs = new BitSetIterator(filterBitSet, filterCardinality);
+                final TopDocs result = doExactSearch(context, docs, filterCardinality, k);
+                return new PerLeafResult(filterBitSet, filterCardinality, result, PerLeafResult.SearchMode.EXACT_SEARCH);
+            }
             final DocIdSetIterator liveFilterIterator = createLiveDocsFilteredIterator(filterIterator, liveDocs);
             final TopDocs result = doExactSearch(context, liveFilterIterator, filterCardinality, k);
             return new PerLeafResult(null, filterCardinality, result, PerLeafResult.SearchMode.EXACT_SEARCH);
         }
 
         // BitSet is required for: nested docs, approximate search, or exact search fallback after ANN
-        final StopWatch stopWatch = startStopWatch(log);
-        final BitSet filterBitSet = createBitSet(context, filterIterator, liveDocs, maxDoc);
+        if (filterBitSet == null) {
+            final StopWatch stopWatch = startStopWatch(log);
+            filterBitSet = createBitSet(context, filterIterator, liveDocs, maxDoc);
+            stopStopWatchAndLog(log, stopWatch, "FilterBitSet creation", knnQuery.getShardId(), segmentName, knnQuery.getField());
+        }
         final int actualCardinality = filterBitSet.cardinality();
-        stopStopWatchAndLog(log, stopWatch, "FilterBitSet creation", knnQuery.getShardId(), segmentName, knnQuery.getField());
 
         // For nested docs: check if exact search is preferred before doing ANN
         if (knnQuery.getParentsFilter() != null && isFilteredExactSearchPreferred(actualCardinality)) {
@@ -705,6 +733,34 @@ public abstract class KNNWeight extends Weight {
      */
     private boolean isExactSearchThresholdSettingSet(int filterThresholdValue) {
         return filterThresholdValue != KNNSettings.ADVANCED_FILTERED_EXACT_SEARCH_THRESHOLD_DEFAULT_VALUE;
+    }
+
+    /**
+     * Determines if the filter query type provides accurate cost() estimates.
+     * TermQuery and BooleanQuery with only MUST/FILTER clauses provide exact cost().
+     * Other query types (e.g., PointRangeQuery) may significantly overestimate.
+     *
+     * @param query the filter query to check
+     * @return true if cost() can be trusted for cardinality estimation
+     */
+    @VisibleForTesting
+    static boolean canUseCostForCardinality(Query query) {
+        if (query instanceof TermQuery) {
+            return true;
+        }
+        if (query instanceof BooleanQuery) {
+            BooleanQuery bq = (BooleanQuery) query;
+            for (BooleanClause clause : bq.clauses()) {
+                if (clause.occur() == BooleanClause.Occur.SHOULD) {
+                    return false;
+                }
+                if (!canUseCostForCardinality(clause.query())) {
+                    return false;
+                }
+            }
+            return true;
+        }
+        return false;
     }
 
     /**
--- a/src/test/java/org/opensearch/knn/index/query/IteratorCostAccuracyTests.java
+++ b/src/test/java/org/opensearch/knn/index/query/IteratorCostAccuracyTests.java
@@ -97,8 +97,7 @@ public class IteratorCostAccuracyTests extends KNNTestCase {
     // ==================== BooleanQuery (MUST/AND) Tests ====================
 
     public void testBooleanMustQueryCost_LowCardinality() throws IOException {
-        Query query = new BooleanQuery.Builder()
-            .add(new TermQuery(new Term("category_low", "match")), BooleanClause.Occur.MUST)
+        Query query = new BooleanQuery.Builder().add(new TermQuery(new Term("category_low", "match")), BooleanClause.Occur.MUST)
             .add(new TermQuery(new Term("category_high", "match")), BooleanClause.Occur.MUST)
             .build();
         // AND of low and high should return low (intersection)
@@ -106,8 +105,7 @@ public class IteratorCostAccuracyTests extends KNNTestCase {
     }
 
     public void testBooleanMustQueryCost_MediumCardinality() throws IOException {
-        Query query = new BooleanQuery.Builder()
-            .add(new TermQuery(new Term("category_medium", "match")), BooleanClause.Occur.MUST)
+        Query query = new BooleanQuery.Builder().add(new TermQuery(new Term("category_medium", "match")), BooleanClause.Occur.MUST)
             .add(new TermQuery(new Term("category_high", "match")), BooleanClause.Occur.MUST)
             .build();
         verifyCostAccuracy(query, MEDIUM_CARDINALITY, "BooleanMUST-Medium");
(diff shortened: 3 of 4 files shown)


Vikasht34:
> It works by using the `filterIterator.cost()` method to estimate the cardinality of the filter. If we want to get the cardinality exactly I believe we will need to materialize the BitSet and calculate the number of set bits (if a bit is set than the document at that location passes through the filter). I'd like some feedback in the PR about how to validate that the `.cost()` estimation function is a good enough approximation for different types of filter iterators. Below is a table summarizing the behavior of `cost()` for some types of Lucene iterators.


cost() is only "generally an upper bound"  - can be wildly inaccurate(https://lucene.apache.org/core/8_0_0/core/org/apache/lucene/search/DocIdSetIterator.html), so let's take an example if cost() returns 1000, but actual matches = 5. We do ANN instead of exact search, losing recall and as we know cost() varies lots on querytype. 


so we should use cost very wisely to handle near the boundary use cases. we should do two phase,
1. if cost() suggests very small or very large, use it 
2. For boundary cases, probe exact cardinality up to threshold

 1: k=10, actual matches=5
Correctly chooses exact search

2: k=10, actual matches=1000  
 Correctly chooses ANN search

Scenario 3: k=10, actual matches=10
Correctly chooses exact search


So bottom line is Just relying on Cost IMO is not enough.

Vikasht34:
> I ran some experiments with different types of filter iterators. The experiment creates different types of queries (like `TermQuery`, boolean composites, and `PointRangeQuery`), calls the iterator `cost()` method, and then exhausts the iterator to count the actual number of docs.
> 
> The takeaway I have from these tests is that `TermQuery` should use the cost() estimate, and boolean AND/MUST queries should use cost() estimate as long as the underlying query are termQuery. What do you think? @navneet1v @Vikasht34
> 
> Here is the summary, full tests are in `IteratorCostAccuracyResults.md` ([link](https://github.com/opensearch-project/k-NN/pull/3048/changes#diff-7656750584896da64a8ede23662d90dcd62931bc2c17660dac1b09c714d7adb9R3-R23)) and test/experiment in `src/test/java/org/opensearch/knn/index/query/IteratorCostAccuracyTests.java` ([link](https://github.com/opensearch-project/k-NN/pull/3048/changes#diff-d093073cd26bb3f7a7452f93dc8ceb5b2dcbacc98a09f1b13d3562dbf79b2432R59-R67)).
> 
> Iterator Type	Cardinality	cost()	actual	accuracy	exact?
> **TermQuery**	Low (5)	5	5	1.00	✅
> **TermQuery**	Medium (10)	10	10	1.00	✅
> **TermQuery**	High (1000)	1000	1000	1.00	✅
> **BooleanMUST**	Low (5)	5	5	1.00	✅
> **BooleanMUST**	Medium (10)	10	10	1.00	✅
> **BooleanSHOULD**	Low (5)	5	5	1.00	✅
> **BooleanSHOULD**	High	15	10	1.50	❌ (overestimate)
> **PointRangeQuery**	Low (5)	385	5	77.00	❌ (huge overestimate)
> **PointRangeQuery**	Medium (10)	385	10	38.50	❌ (huge overestimate)
> **PointRangeQuery**	High (1000)	1155	1000	1.16	❌ (slight overestimate)
> **BitSetIterator**	All	exact	exact	1.00	✅
> **Mixed (Term+Point)**	10	385	10	38.50	❌ (PointRange dominates)
> **Key findings:**
> 
> * **TermQuery**: Always exact - safe to use `cost()`
> * **BooleanMUST (AND)**: Exact when sub-queries are exact
> * **BooleanSHOULD (OR)**: Sums costs, can overestimate with overlapping docs
> * **PointRangeQuery**: Significantly overestimates, especially at low cardinality (BKD tree estimation)
> * **BitSetIterator**: Always exact (cardinality passed at construction)


This is really good experiment, IMO we should use this, 

Fast path: get the cost , if Cost > ExcatSearchThresold*10(RandomNumber) ..we should defi etly use ANN
now to handle boundary cases we should probe upto threshold plus 1 maybe and see if we need filteredKnnExcatSearch or not.

finnroblin:
Hi @navneet1v @Vikasht34 , I ran some benchmarks on an isolated instance to measure the performance improvement in the exact search path. The results actually show some slowdown.

Experiment setup:
- r6g.8xl , run opensearch with 64GB JVM heap. Run OSB on same machine.
- Ingest Cohere 1M 768d with filter attributes. Each doc has fields `filter01pct`, `filter1pct`, `filter5pct`, `filter10pct` which are boolean T/F. 0.1% of docs have `filter01pct = True` and so on.
- Force merge to 5 segments (only 1 pri shard, no replicas)
- Search: perform 5 warmup runs, discard them, perform 5 measured runs.
- Stop the cluster, rebuild with optimized code, restore via the data directory (stored segments).
- Search: for optimized branch perform 5 warmup runs, discard them, perform 5 measured runs.

So the path with exact search is `filter01pct` will have 1000 docs, 200 per segment on average (it won't be exactly even). This is because exact search is preferred when: filterCardinality * dimension <= 2,048,000
  filter01pct: 200 * 768 = 153600 <= 2,048,000

Experiment results:
```
=== BASELINE MEASURED ===
run1: 1366.5 ops/s | p50: 3.67ms | p99: 5.57ms | p99.9: 6.86ms
run2: 1567.6 ops/s | p50: 2.97ms | p99: 5.11ms | p99.9: 5.86ms
run3: 1700.2 ops/s | p50: 2.73ms | p99: 5.52ms | p99.9: 6.86ms
run4: 1496.9 ops/s | p50: 3.12ms | p99: 5.01ms | p99.9: 33.99ms
run5: 1628.5 ops/s | p50: 2.84ms | p99: 5.35ms | p99.9: 5.71ms

=== OPTIMIZED MEASURED ===
run1: 1065.6 ops/s | p50: 5.38ms | p99: 5.69ms | p99.9: 5.94ms
run2: 1144.7 ops/s | p50: 5.19ms | p99: 5.39ms | p99.9: 6.56ms
run3: 1156.2 ops/s | p50: 5.23ms | p99: 5.42ms | p99.9: 5.47ms
run4: 1133.6 ops/s | p50: 5.09ms | p99: 5.64ms | p99.9: 28.18ms
run5: 1264.8 ops/s | p50: 4.47ms | p99: 5.37ms | p99.9: 7.57ms
```
| Run | Baseline | | | | Optimized | | | |
|-----|----------|------|------|-------|-----------|------|------|-------|
| | ops/s | p50 | p99 | p99.9 | ops/s | p50 | p99 | p99.9 |
| 1 | 1366 | 3.67 | 5.57 | 6.86 | 1066 | 5.38 | 5.69 | 5.94 |
| 2 | 1568 | 2.97 | 5.11 | 5.86 | 1145 | 5.19 | 5.39 | 6.56 |
| 3 | 1700 | 2.73 | 5.52 | 6.86 | 1156 | 5.23 | 5.42 | 5.47 |
| 4* | 1497 | 3.12 | 5.01 | 33.99 | 1134 | 5.09 | 5.64 | 28.18 |
| 5 | 1629 | 2.84 | 5.35 | 5.71 | 1265 | 4.47 | 5.37 | 7.57 |

run4 has outliers in both.

Baseline is consistently faster (~1500-1700 ops/s vs ~1100-1265 ops/s) with lower p50 latency (~2.7-3.7ms vs ~4.5-5.4ms).


Profiles:
I pulled two profiles with async profiler 3.0. One had the current k-NN main branch and the other this optimized branch. With the latency degradation I was expecting to see high number of CPU cycles spent in `cost()` method. However this was not the case.

Optimized ( 
[flamegraph_optimized.html](https://github.com/user-attachments/files/24576682/flamegraph_optimized.html)
) :
Screenshot 2026-01-12 at 4 41 16 PM

Baseline ( 
[flamegraph_baseline_filter01pct.html](https://github.com/user-attachments/files/24576681/flamegraph_baseline_filter01pct.html)
)
Screenshot 2026-01-12 at 4 43 42 PM


Some observations:

- We see in the optimized branch that "termQuery.get()" is called (<1% of flame graph). So this sanity check shows that the optimized code is picked up and the direct branch is called.
- In the baseline branch for the selective filter, only <1% of profiler time is spent in `createBitSet`.
Screenshot 2026-01-12 at 4 42 53 PM

- Looking at the cost of `nextDoc`, for baseline the magenta are highlighted. The larger two bars are VectorIdsKNNIterator::NextDoc , the smaller bars are FilteredDISI::nextDoc or BitSetIterator::nextDoc
Screenshot 2026-01-12 at 4 45 52 PM

Optimized shows similar bars, with one difference: More time spent in `org/apache/lucene/codecs/lucene103/Lucene103PostingsReader$BlockPostingsEnum.nextDoc`
Screenshot 2026-01-12 at 4 48 46 PM

Around 0.8% of baseline is spent in `BlockPostingsEnum.nextDoc`, 1.5% in optimized is spent in the method.

My thoughts:
- It may be that the exactsearcher is hanging more when it is iterating over the FilterDISI. In the baseline we do some initial work of bringing up the BitSet , but the code path is solely responsible for doing the iteration (to set the bits). Then we pass through the generated bitset iterator to exact searcher. Vs in the optimized method, we defer iteration over the FilteredDISI instead of a BitSetIterator. Calling `FilterDISI::nextDoc` in the Exact Searcher may be more costly since we are context switching to distance computation and back to iteration — it may be cheaper in terms of memory locality to handle `filterDISI::nextDoc` all at once when we are making the BitSetIterator.
- The Lucnee103PostingsReader seems very expensive! I don't have full context on the `nextDoc` method ([link](https://github.com/apache/lucene/blob/9e7381cc5cc76234eb232b3adab43bbabd167add/lucene/backward-codecs/src/java/org/apache/lucene/backward_codecs/lucene103/Lucene103PostingsReader.java#L891)), but it seems to open a segment and iterate over blocks of memory. It could be the case that if we are generating the BitSetIterator all at once, we open fewer segment files and have better memory location when iterating over the filterDISI.

What do you folks think? Any ideas about where I can look next in the flamegraphs to figure out where the slowdown exactly is coming from? Does my hypothesis that the slowdown is from not batching the filterDISI.nextDoc iteration seem plausible (I am unsure about it myself).

Vikasht34:
> Here are some results from a single segment search. The profiles look cleaner. We are seeing more speedup of the tail latencies.
> 
> ```
> Baseline Results:
>   Run 1: 807.6 ops/s, P50: 3.24ms, P90: 3.31ms, P99.9: 7.69ms
>   Run 2: 802.9 ops/s, P50: 3.25ms, P90: 3.37ms, P99.9: 7.66ms
>   Run 3: 815.2 ops/s, P50: 3.21ms, P90: 3.27ms, P99.9: 8.06ms
>   Run 4: 805.5 ops/s, P50: 3.23ms, P90: 3.34ms, P99.9: 8.18ms
>   Run 5: 807.1 ops/s, P50: 3.22ms, P90: 3.28ms, P99.9: 18.29ms
> 
> Optimized Results:
>   Run 1: 637.1 ops/s, P50: 3.84ms, P90: 3.97ms, P99.9: 8.53ms
>   Run 2: 619.6 ops/s, P50: 3.74ms, P90: 3.86ms, P99.9: 8.90ms
>   Run 3: 819.3 ops/s, P50: 3.17ms, P90: 3.24ms, P99.9: 7.39ms
>   Run 4: 800.2 ops/s, P50: 3.19ms, P90: 3.34ms, P99.9: 7.54ms
>   Run 5: 693.6 ops/s, P50: 3.63ms, P90: 3.73ms, P99.9: 7.80ms
> 
> ============================================================
> PERFORMANCE ANALYSIS: Baseline vs Optimized
> ============================================================
> Throughput:
>   Baseline: 807.67 ops/s
>   Optimized: 713.97 ops/s
>   Change: -11.6%
> 
> P50 Latency:
>   Baseline: 3.23 ms
>   Optimized: 3.52 ms
>   Change: -8.9% (worse)
> 
> P90 Latency:
>   Baseline: 3.32 ms
>   Optimized: 3.63 ms
>   Change: -9.4% (worse)
> 
> P99.9 Latency:
>   Baseline: 9.98 ms
>   Optimized: 8.03 ms
>   Change: +19.5% (better)
> ```
> 
> Profiles: [flamegraph_1segment_baseline_filter01pct.html](https://github.com/user-attachments/files/24599261/flamegraph_1segment_baseline_filter01pct.html) [flamegraph_1segment_optimized_filter01pct.html](https://github.com/user-attachments/files/24599262/flamegraph_1segment_optimized_filter01pct.html)
> 
> One other thing that Navneet pointed out offline is that we can test the "BitSet.of() saves work when iterating over the FilterDISI" hypothesis by directly making a bitset in KNNIterator before it is passed to exact search in this [code path](https://github.com/opensearch-project/k-NN/blob/fd71813fa984293516245631d291c23298f54716/src/main/java/org/opensearch/knn/index/query/ExactSearcher.java#L167). Then we don't need to iterate over the FilterDISI in ExactSearch and are instead iterating over a BitSetIterator. I will run that test next.

Actually it's not deterministic, P100 is way better but P90 is not .. how come?

finnroblin:
I did some runs with the following change to identify if the lack of a `BitSetIterator` is the cause of slowdown:
In ExactSearcher.java (changes made in optimized branch, baseline branch remains the same):
```
    private KNNIterator getKNNIterator(LeafReaderContext leafReaderContext, ExactSearcherContext exactSearcherContext) throws IOException {
// replace -        final DocIdSetIterator matchedDocs = exactSearcherContext.getMatchedDocsIterator(); with below:
        final DocIdSetIterator mdi = exactSearcherContext.getMatchedDocsIterator();
        final BitSetIterator matchedDocs = new BitSetIterator(BitSet.of(mdi, leafReaderContext.reader().maxDoc()), exactSearcherContext.getNumberOfMatchedDocs());
```


```
Baseline Results:
  Run 1: 818.1 ops/s, P50: 3.16ms, P90: 3.23ms, P99.9: 7.68ms
  Run 2: 830.0 ops/s, P50: 3.13ms, P90: 3.19ms, P99.9: 8.09ms
  Run 3: 824.7 ops/s, P50: 3.13ms, P90: 3.19ms, P99.9: 7.61ms
  Run 4: 827.8 ops/s, P50: 3.13ms, P90: 3.19ms, P99.9: 8.05ms
  Run 5: 831.0 ops/s, P50: 3.14ms, P90: 3.20ms, P99.9: 7.52ms

Optimized Results:
  Run 1: 803.8 ops/s, P50: 3.22ms, P90: 3.28ms, P99.9: 7.23ms
  Run 2: 811.7 ops/s, P50: 3.22ms, P90: 3.30ms, P99.9: 7.52ms
  Run 3: 811.3 ops/s, P50: 3.23ms, P90: 3.34ms, P99.9: 7.52ms
  Run 4: 807.5 ops/s, P50: 3.22ms, P90: 3.34ms, P99.9: 7.70ms
  Run 5: 796.0 ops/s, P50: 3.27ms, P90: 3.38ms, P99.9: 7.04ms

============================================================
PERFORMANCE ANALYSIS: Baseline vs Optimized
============================================================
Throughput:
  Baseline: 826.32 ops/s
  Optimized: 806.04 ops/s
  Change: -2.5%

P50 Latency:
  Baseline: 3.14 ms
  Optimized: 3.23 ms
  Change: -2.9% (worse)

P90 Latency:
  Baseline: 3.20 ms
  Optimized: 3.33 ms
  Change: -4.0% (worse)

P99.9 Latency:
  Baseline: 7.79 ms
  Optimized: 7.40 ms
  Change: +5.0% (better)

============================================================
SUMMARY:
✗ Throughput decreased by 2.5%
✗ P50 latency worsened by 2.9%
✓ P99.9 latency improved by 5.0%
```

@navneet1v / @Vikasht34: We see that the latencies are "less bad" when materializing the bitset in the optimized branch, suggesting that there is a benefit to passing a `BitSetIterator` vs a direct `FilterDISI` to the exact searcher, possibly due to a more efficient implementation of `nextDoc` within the `BitSetIterator`.

PR #3049, opened 2026-01-05
Field exclusion in source indexing handling

### LINKED ISSUE #3034: [BUG] Field excluded from source bug
**What is the bug?**
More of an optimization than a bug.
Field excluded from source. We are still going through derived logic for the field, but just excluding at the end. To optimize, we avoid adding the derived field when field is excluded from source, so that the field is never processed in the derived path.

We do this by adding a check for field excluded in source in this index util function:

```
public static boolean isDerivedEnabledForField(KNNVectorFieldType knnVectorFieldType, MapperService mapperService) {
    // Skip copy to fields
    if (mapperService.documentMapper().mappers().getMapper(knnVectorFieldType.name()) instanceof FieldMapper mapper) {
        return mapper.copyTo() == null || mapper.copyTo().copyToFields() == null || mapper.copyTo().copyToFields().isEmpty() != false;
    }
    return true;
}
```

buddharajusahil:
### Description
Add a check for is derived enabled for field to return false if the field is excluded from the source, as if its excluded we do not need to go through derived source logic.

### Related Issues
Resolves #3034

### Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,3 +30,4 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ### Enhancements
 * Index setting to disable exact search after ANN Search with Faiss efficient filters [#3022](https://github.com/opensearch-project/k-NN/pull/3022)
+* Bulk SIMD for FP16 distance computation [#3071](https://github.com/opensearch-project/k-NN/pull/3071)
--- a/jni/cmake/init-simd.cmake
+++ b/jni/cmake/init-simd.cmake
@@ -5,6 +5,7 @@
 
 include(CheckCXXSourceCompiles)
 
+
 # Allow user overrides
 if(NOT DEFINED AVX2_ENABLED)
     set(AVX2_ENABLED true)   # set default value as true if the argument is not set
@@ -68,7 +69,11 @@ elseif(${CMAKE_SYSTEM_PROCESSOR} MATCHES "aarch64" OR ${CMAKE_SYSTEM_PROCESSOR} 
     endif()
 
 elseif(${CMAKE_SYSTEM_NAME} STREQUAL "Linux" AND AVX512_SPR_ENABLED)
-    set(CMAKE_REQUIRED_FLAGS "-mavx512f -mavx512fp16 -mf16c")
+    # -mavx512f	: 512-bit Registers & FMA. Ex: __m512, _mm512_fmadd_ps
+    # -mavx512bw : 16-bit Masking. Ex: __mmask16, mask generation
+    # -mavx512vl : 256-bit "Sub-vector" support. Ex: _mm256_maskz_loadu_epi16
+    # -mavx512fp16 : Native FP16 / EVEX Conversion.
+    set(CMAKE_REQUIRED_FLAGS "-mavx512f -mavx512bw -mavx512vl -mavx512fp16")
     check_cxx_source_compiles("
         #include <immintrin.h>
         int main() {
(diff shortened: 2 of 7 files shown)


src/main/java/org/opensearch/knn/index/util/IndexUtil.java
```diff
@@ -480,10 +482,53 @@ public static boolean isDerivedEnabledForIndex(MapperService mapperService) {
...
+                    break;
+                }
+            }
+            if (!matchedInclude) {
+                return true;  // Not in includes = excluded
+            }
+        }
+
+        // Check if field matches any exclude pattern
+        if (excludes != null && excludes.length > 0) {
```
Vikasht34: Why this is again for if matchedInclude is true Since we are matching feild Name?

buddharajusahil: I might be misunderstanding the question. Even if a field matches the include pattern, it can still be excluded.

Vikasht34: so the Problem with SimpleMatch regex is if I do simpleMatch("user.*", "user.name.first") it return false, but it should be included.

buddharajusahil: My understanding is simpleMatch("user.*", "user.name.first") will return true, as simplematch handles wildcard *.

Vikasht34: You are right , That's correct.

src/main/java/org/opensearch/knn/index/util/IndexUtil.java
```diff
@@ -480,10 +482,53 @@ public static boolean isDerivedEnabledForIndex(MapperService mapperService) {
...
     }
+
+    private static boolean isFieldExcludedFromSource(String fieldName, MapperService mapperService) {
+        SourceFieldMapper sourceMapper = mapperService.documentMapper().metadataMapper(SourceFieldMapper.class);
+        if (sourceMapper == null) {
+            return false;
+        }
+
+        String[] includes = sourceMapper.getIncludes();
+        String[] excludes = sourceMapper.getExcludes();
```
sohami: these will change based on the method definition change [here](https://github.com/opensearch-project/OpenSearch/pull/20290/files)

buddharajusahil: Good catch, will revise.

sohami (approved):

Vikasht34 (dismissed):

src/main/java/org/opensearch/knn/index/util/IndexUtil.java
```diff
@@ -480,10 +483,53 @@ public static boolean isDerivedEnabledForIndex(MapperService mapperService) {
...
+        }
+
         // Skip copy to fields
         if (mapperService.documentMapper().mappers().getMapper(knnVectorFieldType.name()) instanceof FieldMapper mapper) {
             return mapper.copyTo() == null || mapper.copyTo().copyToFields() == null || mapper.copyTo().copyToFields().isEmpty() != false;
         }
         return true;
     }
+
+    private static boolean isFieldExcludedFromSource(String fieldName, MapperService mapperService) {
```
shatejas, line 501: Can we have unit tests for isDerivedEnabledForField covering include exclude cases extensively?

Also see if there is an existing integration test that can be used to add a simple include and exclude works and check if it works e2e

buddharajusahil: Sure will add this testing.

Vikasht34 (approved):

Issue #3026, opened 2025-12-13
[BUG] Getting EOF exception when running disk based vector search with Nested Fields and when expand_nested_docs with in_memory and disk mode


navneet1v:
### Description
Getting EOF exception when running disk based vector search with Nested Fields and when expand_nested_docs with in_memory and disk mode

### Root Cause:

1. Create an index and ingest docs in OpenSearch where vector is part of a nested field.

2. If the nested field child docs has vectors in it, then there is no issue. But if in a nested field if one of the child doesn't have vector(which can be possible), then in that case due to [this](https://github.com/opensearch-project/k-NN/blob/4f627ad1de6143af51ed12422550940853fb4a13/src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java#L267) which calls [this](https://github.com/opensearch-project/k-NN/blob/4f627ad1de6143af51ed12422550940853fb4a13/src/main/java/org/opensearch/knn/index/query/common/QueryUtils.java#L138-L150) we are getting all the child docs too where a child doc might not have vectors in it. Due to this the iterator is running pass the vector file  since DISI has the docId but there is no vector present in the .vec file.

### Replication Script
```
#!/bin/bash

if [ $# -lt 1 ] || [ $# -gt 3 ]; then
    echo "Usage: $0 <number_of_docs> [mode] [expand_nested_docs]"
    echo "Mode: in_memory | on_disk (default: on_disk)"
    exit 1
fi

NUM_DOCS=$1
MODE=${2:-"on_disk"}
OS_URL="http://localhost:9200"
INDEX_NAME="knn_test_index"
EXPAND_NESTED_DOCS=${3:-"false"}

# Delete index if it exists
echo "Deleting index if exists..."
curl -s -X DELETE "$OS_URL/$INDEX_NAME" | jq .

echo "Updating settings"
curl -s -X PUT "$OS_URL/_cluster/settings" -H "Content-Type: application/json" -d '{
    "persistent": {
        "logger.org.opensearch.knn.index.query": "DEBUG"
    }
}' | jq .


# Create index with nested knn vector field
echo "Creating index with mode: $MODE"
jq -n --arg mode "$MODE" '{
  "settings": {
    "index": {
      "knn": true,
      "number_of_shards": 2
    }
  },
  "mappings": {
    "properties": {
      "title": {"type": "text"},
      "nested_field": {
        "type": "nested",
        "properties": {
          "title": {"type": "text"},
[...]

PR #3051, opened 2026-01-06
Fix nested docs query when some child docs has no vector field present

navneet1v:
### Description
Fix nested docs query when some child docs has no vector field present.

This change fixes issues that can happen during various cases which includes:
1. When expand nested docs = true 
2. When on_disk mode is true and rescoring is required
3. Filtering cases when search fall back to exact search.

#### Root Cause
The common case in all of these scenarios is how we are iterating over docIds(child docIds) of matched parentIds while doing exact search. Earlier the code was assuming that every child of a parent doc will have vector field in it. But this is not. Due to lack of vector field, vectorValues iterator was getting exhausted while there are child docIds still present. This leads to an EOF exception on `.vec` file. Ref GH issue for actual exception trace and script to repro the issue.

Below is an example explaining what is happening:
Consider this below case at a segment level:

```
Parent DocIds: [3, 6]
ChildDocsIds: [0, 1, 2, 4, 5]
VectorValuesDocIds : [0, 1 , 2, 4]
```

when exact search is running for this segment what will happen is we will start iterating over the parentDocIds using [NestedVectorIdsKNNIterator](https://github.com/opensearch-project/k-NN/blob/88792e42f121b050f2fc9cf32b039052aab62128/src/main/java/org/opensearch/knn/index/query/iterators/NestedVectorIdsKNNIterator.java#L22) and then for every parent docId we will [expand](https://github.com/opensearch-project/k-NN/blob/88792e42f121b050f2fc9cf32b039052aab62128/src/main/java/org/opensearch/knn/index/query/iterators/NestedVectorIdsKNNIterator.java#L64-L88) it to child docs score the vectors for those child docs. This is where VectorValues moves ahead if some child docs has no vector in it, leading to failure. This happens even when when we ask the VectorValues to [advance to a specific docId](https://github.com/opensearch-project/k-NN/blob/db822193495e367e087039e1168c5c60287f76fa/src/main/java/org/opensearch/knn/index/query/iterators/VectorIdsKNNIterator.java#L106-L116) because calling advance with a docId move the iterator to first docId which is >= provided docId ([Lucene ref](https://github.com/apache/lucene/blob/ae6484f43e61a65504f9b270e16c27e92d31875c/lucene/core/src/java/org/apache/lucene/search/DocIdSetIterator.java#L196-L223)) 

### Solution
In the PR we create an Intersection Iterator using the VectorValuesIterator and MachedDocsDISI which will be passed to KNNIterator to score the docs. This will ensure that only docs with Vectors are getting scored and EOF exception doesn't happen. The way this new Iterator works is using the Leap Frog fashion, hence if one of the iterator gets exhausted first we will exist out. To make sure that all the usecases are tested, added new tests in ExpandNestedDocsIT and NestedDocsIT class.

Few more changes in PR include:
1. Removal of few redundant variables.
2. To reduce the blast radius of test data creation, created some overloaded functions in tests.
3. Moved KNNIterator and its implementation to an exact search package which ensure that the KNNIterators are only used for Exact Search. 
4. Renamed KNNIterators to ExactKNNIterators to clearly define what the iterator is supposed to do.

### Related Issues
Resolves https://github.com/opensearch-project/k-NN/issues/3026

### Check List


src/main/java/org/opensearch/knn/index/query/ExactSearcher.java
```diff
@@ -179,83 +181,131 @@ private KNNIterator getKNNIterator(LeafReaderContext leafReaderContext, ExactSea
...
-            spaceType,
-            quantizedQueryVector,
-            segmentLevelQuantizationInfo
-        );
+        // align the docIds of the iterators to same docId.
+        while (originalMatchedDocsDISI.docID() != DocIdSetIterator.NO_MORE_DOCS
+            && vectorValuesDISI.docID() != DocIdSetIterator.NO_MORE_DOCS) {
+            int originalMatchedDocsDocId = originalMatchedDocsDISI.docID();
+            int vectorValuesDocId = vectorValuesDISI.docID();
+            if (originalMatchedDocsDocId == vectorValuesDocId) {
```
kotwanikunal: The default value for ids would be -1 for both the iterators here right?

Based on https://github.com/apache/lucene/blob/branch_10_1/lucene/core/src/java/org/apache/lucene/search/DocIdSetIterator.java#L43-L45

I think we will run into the same scenario. We need to advance both the iterators before performing this check.

src/main/java/org/opensearch/knn/index/mapper/Mode.java
```diff
@@ -25,14 +25,12 @@ public enum Mode {
...
 
     // Internally, an empty string is easier to deal with them null. However, from the mapping,
-    // we do not want users to pass in the empty string and instead want null. So we make the conversion herex
+    // we do not want users to pass in the empty string and instead want null. So we make the conversion here
     public static final String[] NAMES_ARRAY = Arrays.stream(Mode.values())
         .map(mode -> mode == NOT_CONFIGURED ? null : mode.getName())
         .collect(Collectors.toList())
         .toArray(new String[0]);
 
-    private static final Mode DEFAULT = IN_MEMORY;
```
Vikasht34, line 34: Is this not getting used anywhere?

src/main/java/org/opensearch/knn/index/query/ExactSearcher.java
Vikasht34: When initial DocId is -1 and -1 for both and it breaks the while loop and then intersectIterators will start throwing the error that ids are not aligned.

src/main/java/org/opensearch/knn/index/query/exactsearch/ExactSearcher.java
```diff
@@ -179,83 +181,131 @@ private KNNIterator getKNNIterator(LeafReaderContext leafReaderContext, ExactSea
...
+
+        if (originalMatchedDocsDISI.docID() == DocIdSetIterator.NO_MORE_DOCS || vectorValuesDISI.docID() == DocIdSetIterator.NO_MORE_DOCS) {
+            // return an empty iterator since one of the iterator is exhausted
+            return DocIdSetIterator.empty();
+        }
+
+        final List<DocIdSetIterator> disiList = new ArrayList<>();
+        disiList.add(originalMatchedDocsDISI);
+        disiList.add(vectorValuesDISI);
+        return ConjunctionUtils.intersectIterators(disiList);
```
Vikasht34, line 282: As far as I can read ConjunctionUtils , It brings some of the overhead as use case will be greatly benefit it for Two phase search (https://github.com/apache/lucene/blob/26b118872f617edb6688b6439bf87a669487cfe2/lucene/core/src/java/org/apache/lucene/search/ConjunctionDISI.java), So overhead includes allocating lists and wrapper objects. And the reason I believe we are aligning first becuase createConjunction,  the iterators must be on the same docID at construction time. 

Since we already have aligned , then why not create simple IntersectDocIdSetIterator which will avoids all the TwoPhaseIterator?

src/main/java/org/opensearch/knn/index/mapper/Mode.java
navneet1v: Yeah its not getting used.

src/main/java/org/opensearch/knn/index/query/ExactSearcher.java
navneet1v: This is not the final code. I don't think this is needed. I was testing some fixes and left it here.

src/main/java/org/opensearch/knn/index/query/exactsearch/ExactSearcher.java
navneet1v: This is not the final code. We don't need any alignment. Some UTs are failing because they are not mocking the iterators properly. So I need to fix that first. But at the end we just need to an iterator from `ConjunctionUtils.intersectIterators`. Once I clean up the code then it will make more sense.

src/main/java/org/opensearch/knn/index/query/ExactSearcher.java
navneet1v: Yes you are right, docIds will be -1. I was testing some UT fix and then I added this alignment code. Actually we don't need any alignment. It was just some bad mocking on UTs which caused the tests to crash(which let me believe I need an alignment code, but actually I don't need it). I am fixing UTs mocking one by one. I will ping on the PR once all CIs are passing.

Vikasht34:
> > So IMO we should fix the place which is causing the issue . Excatsearcher is caller so ideally this is not the place which will have the fix. We need to fix iterator (Next Doc ) for Filtered Document and Skip Child Where we Don't have Vector.
> 
> Using an intersection iterator is exactly what will skip the docIds which doesn't have a vector. We can add the logic to skip the docIds where vectors are not present but that would mean implementing the intersection logic of our. Rather having a wrapper iterator class provided by Lucene is much more safer and cleaner. The internal KNNIterators doesn't now need to handle anything rather than just pure iteration over FilterIdsIterator and VectorValues.



This is what I have in Mind , Just below lines of Code without bringing the Intersector .. and this will fix paramenetly for any use case which goes through getNextDocId  for filktered uses case.

```
protected int getNextDocId() throws IOException {
    if (filterIdsIterator == null) {
        return knnFloatVectorValues.nextDoc();
    }
    int nextDocID = this.filterIdsIterator.nextDoc();
    if (nextDocID != DocIdSetIterator.NO_MORE_DOCS) {
        int vectorDocId = knnFloatVectorValues.advance(nextDocID); 
        while (vectorDocId != nextDocID && nextDocID != DocIdSetIterator.NO_MORE_DOCS) {
            nextDocID = this.filterIdsIterator.nextDoc();
            if (nextDocID != DocIdSetIterator.NO_MORE_DOCS) {
                vectorDocId = knnFloatVectorValues.advance(nextDocID);
            }
        }
    }
    return nextDocID;
}
```

Vikasht34:
> > > > So IMO we should fix the place which is causing the issue . Excatsearcher is caller so ideally this is not the place which will have the fix. We need to fix iterator (Next Doc ) for Filtered Document and Skip Child Where we Don't have Vector.
> > > 
> > > 
> > > Using an intersection iterator is exactly what will skip the docIds which doesn't have a vector. We can add the logic to skip the docIds where vectors are not present but that would mean implementing the intersection logic of our. Rather having a wrapper iterator class provided by Lucene is much more safer and cleaner. The internal KNNIterators doesn't now need to handle anything rather than just pure iteration over FilterIdsIterator and VectorValues.
> > 
> > 
> > This is what I have in Mind , Just below lines of Code without bringing the Intersector .. and this will fix paramenetly for any use case which goes through getNextDocId for filktered uses case.
> > [Block (17 lines)](#)
> > ```
> > protected int getNextDocId() throws IOException {
> >     if (filterIdsIterator == null) {
> >         return knnFloatVectorValues.nextDoc();
> >     }
> >     int nextDocID = this.filterIdsIterator.nextDoc();
> >     if (nextDocID != DocIdSetIterator.NO_MORE_DOCS) {
> >         int vectorDocId = knnFloatVectorValues.advance(nextDocID); 
> >         while (vectorDocId != nextDocID && nextDocID != DocIdSetIterator.NO_MORE_DOCS) {
> >             nextDocID = this.filterIdsIterator.nextDoc();
> >             if (nextDocID != DocIdSetIterator.NO_MORE_DOCS) {
> >                 vectorDocId = knnFloatVectorValues.advance(nextDocID);
> >             }
> >         }
> >     }
> >     return nextDocID;
> > }
> > ```
> 
> Yes exactly I implemented this code in my local and didn't want to use this idea since we should not be maintaining these things if we already have utilities from Lucene. Also the code which looks simple has missing edge cases, which will be harder to maintain going forward.

Yeah , let's solve the missing use case rather that bringing dependency which is not intended to solve the all issue.  I tried Small Bencmark suggested by Q and Saw notificable Perfomace degradation using Intreactor

This was generated By Q .. You can try as well,   

Benchmark: 10,000 docs, 100 missing vectors

ConjunctionDISI: 2.5ms
Simple Fix:         1.1ms (2.3x faster!)

Memory:
ConjunctionDISI: 120 bytes allocated
Simple Fix:         0 bytes allocated

```
/*
 * Simple Benchmark: ConjunctionDISI vs Skip-on-Advance
 * No dependencies required - just compile and run
 * 
 * Compile: javac SimpleBenchmark.java
 * Run: java SimpleBenchmark
 */

import java.util.*;

public class SimpleBenchmark {

    public static void main(String[] args) {
        System.out.println("Iterator Approach Benchmark");
        System.out.println("============================\n");

        int[] docCounts = {1000, 10000, 100000};
        double[] missingRatios = {0.01, 0.05, 0.10};

        for (int totalDocs : docCounts) {
            for (double missingRatio : missingRatios) {
                runBenchmark(totalDocs, missingRatio);
            }
        }
    }

    static void runBenchmark(int totalDocs, double missingRatio) {
        Random random = new Random(42);
        float[] queryVector = {0.5f, 0.5f, 0.5f};

        // Setup data
        int[] filterDocs = new int[totalDocs];
        for (int i = 0; i < totalDocs; i++) {
            filterDocs[i] = i;
        }

        List<Integer> vectorList = new ArrayList<>();
        for (int i = 0; i < totalDocs; i++) {
            if (random.nextDouble() > missingRatio) {
                vectorList.add(i);
            }
        }
        int[] vectorDocs = vectorList.stream().mapToInt(Integer::intValue).toArray();

        int missingCount = totalDocs - vectorDocs.length;

        // Warmup
        for (int i = 0; i < 100; i++) {
            benchmarkConjunction(filterDocs, vectorDocs, queryVector);
            benchmarkSkipOnAdvance(filterDocs, vectorDocs, queryVector);
        }

        // Measure
        int iterations = 1000;
        
        long conjunctionTime = 0;
        for (int i = 0; i < iterations; i++) {
            long start = System.nanoTime();
            benchmarkConjunction(filterDocs, vectorDocs, queryVector);
            conjunctionTime += System.nanoTime() - start;
        }

        long skipTime = 0;
        for (int i = 0; i < iterations; i++) {
            long start = System.nanoTime();
            benchmarkSkipOnAdvance(filterDocs, vectorDocs, queryVector);
            skipTime += System.nanoTime() - start;
        }

        double conjunctionAvg = conjunctionTime / (double) iterations / 1000.0;
        double skipAvg = skipTime / (double) iterations / 1000.0;
        double speedup = conjunctionAvg / skipAvg;

        System.out.printf("Docs: %,6d | Missing: %4.1f%% (%,5d) | Conjunction: %7.2f μs | Skip: %7.2f μs | Speedup: %.2fx\n",
            totalDocs, missingRatio * 100, missingCount, conjunctionAvg, skipAvg, speedup);
    }

    static int benchmarkConjunction(int[] filterDocs, int[] vectorDocs, float[] queryVector) {
        MockFilterIterator filter = new MockFilterIterator(filterDocs);
        MockVectorIterator vector = new MockVectorIterator(vectorDocs);

        List<Iterator> iters = Arrays.asList(filter, vector);
        ConjunctionIterator conj = new ConjunctionIterator(iters);

        int count = 0;
        int doc;
        while ((doc = conj.nextDoc()) != -1) {
            float[] v = vector.getVector();
            float score = computeScore(queryVector, v);
            count++;
        }
        return count;
    }

    static int benchmarkSkipOnAdvance(int[] filterDocs, int[] vectorDocs, float[] queryVector) {
        MockFilterIterator filter = new MockFilterIterator(filterDocs);
        MockVectorIterator vector = new MockVectorIterator(vectorDocs);

        int count = 0;
        int nextDocID;
        while ((nextDocID = filter.nextDoc()) != -1) {
            int vectorDocId = vector.advance(nextDocID);
            
            while (vectorDocId != nextDocID && nextDocID != -1) {
                nextDocID = filter.nextDoc();
                if (nextDocID != -1) {
                    vectorDocId = vector.advance(nextDocID);
                }
            }
            
            if (nextDocID != -1) {
                float[] v = vector.getVector();
                float score = computeScore(queryVector, v);
                count++;
            }
        }
        return count;
    }

    static float computeScore(float[] q, float[] v) {
        float sum = 0;
        for (int i = 0; i < q.length; i++) {
            float diff = q[i] - v[i];
            sum += diff * diff;
        }
        return 1.0f / (1.0f + sum);
    }

    interface Iterator {
        int docID();
        int nextDoc();
        int advance(int target);
    }

    static class ConjunctionIterator implements Iterator {
        private final Iterator lead;
        private final Iterator[] others;

        ConjunctionIterator(List<Iterator> iters) {
            this.lead = iters.get(0);
            this.others = iters.subList(1, iters.size()).toArray(new Iterator[0]);
        }

        public int docID() {
            return lead.docID();
        }

        public int nextDoc() {
            return doNext(lead.nextDoc());
        }

        public int advance(int target) {
            return doNext(lead.advance(target));
        }

        private int doNext(int doc) {
            for (;;) {
                if (doc == -1) return -1;
                
                int next = doc;
                for (Iterator other : others) {
                    int otherDoc = other.advance(doc);
                    if (otherDoc > doc) {
                        doc = lead.advance(otherDoc);
                        next = otherDoc;
                        break;
                    }
                }
                
                if (next == doc) return doc;
            }
        }
    }

    static class MockFilterIterator implements Iterator {
        private final int[] docs;
        private int pos = -1;

        MockFilterIterator(int[] docs) {
            this.docs = docs;
        }

        public int docID() {
            return pos < 0 ? -1 : (pos >= docs.length ? -1 : docs[pos]);
        }

        public int nextDoc() {
            pos++;
            return docID();
        }

        public int advance(int target) {
            while (pos < docs.length - 1 && docs[pos + 1] <= target) {
                pos++;
            }
            if (pos >= 0 && pos < docs.length && docs[pos] >= target) {
                return docs[pos];
            }
            pos = docs.length;
            return -1;
        }
    }

    static class MockVectorIterator implements Iterator {
        private final int[] docs;
        private int pos = -1;
        private final Random random = new Random(42);

        MockVectorIterator(int[] docs) {
            this.docs = docs;
        }

        public int docID() {
            return pos < 0 ? -1 : (pos >= docs.length ? -1 : docs[pos]);
        }

        public int nextDoc() {
            pos++;
            return docID();
        }

        public int advance(int target) {
            while (pos < docs.length - 1 && docs[pos + 1] <= target) {
                pos++;
            }
            if (pos >= 0 && pos < docs.length && docs[pos] >= target) {
                return docs[pos];
            }
            pos = docs.length;
            return -1;
        }

        float[] getVector() {
            return new float[]{random.nextFloat(), random.nextFloat(), random.nextFloat()};
        }
    }
}

```

navneet1v:
Another thing if we really want to benchmark the performance we should do micro benchmarks with Lucene since that has the exact code for intersect. Plus another reason of favoring Lucene over our own implementation is any performance improvement Lucene adds we will get for free. Improvements related to specific iterator types etc.

Vikasht34:
> Another thing if we really want to benchmark the performance we should do micro benchmarks with Lucene since that has the exact code for intersect. Plus another reason of favoring Lucene over our own implementation is any performance improvement Lucene adds we will get for free. Improvements related to specific iterator types etc.

Agree on Lucene Advancements , yes Definitely but this use case does not fit. !!

navneet1v:
> yes Definitely but this use case does not fit. !!

Please explain more how that doesn't fit here?

Vikasht34:
> The reason for making vector as the lead is because that has the minimum cost. Refer this: https://github.com/apache/lucene/blob/fd75807350604fbff5d571ec9b31e3669d17059e/lucene/core/src/java/org/apache/lucene/search/ConjunctionDISI.java#L97-L107 This is what happen in background when we call to get intersect iterator.
> 
> The idea is which should be true for our skipping logic or conjuction, which ever iterator cost is low we should use that to iterate. Since that will lead to best latency.
> 
> >



> > yes Definitely but this use case does not fit. !!
> 
> Please explain more how that doesn't fit here?

Sure , Here it is ,Lucene optimizes for two phase where we have iteraror that can give broder results which has to be cheap and then verfication which is excauhustive so that's what util is doing ..It's heavy on Verfication and due to that we are bringing lots of wrpper intiallized which is not needed since we alkready have custom iterator, and we control nextDoc , it's overiding method. If I buy your logic then Next doc method is not useful and use Super.NextDoc ..

src/main/java/org/opensearch/knn/index/query/exactsearch/ExactSearcher.java
```diff
@@ -179,83 +181,112 @@ private KNNIterator getKNNIterator(LeafReaderContext leafReaderContext, ExactSea
...
+        // We need to create a new VectorValues instances as the new one will be used to iterate over the docIds in
+        // conjunction with Matched Docs.
+        final DocIdSetIterator matchedDocs = getMatchedDocsIterator(
+            exactSearcherContext.getMatchedDocsIterator(),
+            KNNVectorValuesFactory.getVectorValues(fieldInfo, reader).getVectorValuesIterator().getDocIdSetIterator()
+        );
+
+        final KNNIterator knnIterator;
+        final KNNVectorValues<T> vectorValues = KNNVectorValuesFactory.getVectorValues(fieldInfo, reader);
         if (VectorDataType.BINARY == vectorDataType) {
```
kotwanikunal, line 185: nit: While you are at cleaning up the branching logic - I think it would make it easier to read if we break it out into descriptive methods. The current flow can get convoluted to read through.

```
if (VectorDataType.BINARY == vectorDataType) {
    knnIterator = getBinaryIterator(isNestedRequired, ...)
} else if (VectorDataType.BYTE == vectorDataType) {
     knnIterator = getByteIterator(isNestedRequired, ...)
}
....
```

kotwanikunal (approved):

src/main/java/org/opensearch/knn/index/query/exactsearch/ExactSearcher.java
navneet1v: Thank you. I also don't like the code structure. Let me take a stab it in this PR if there are more comments. otherwise I will take it as a separate PR.

Vikasht34 (dismissed):
I think I am convinced on usage of ConjuctionUtil after reading some more. Looks Good to me !!

src/main/java/org/opensearch/knn/index/query/ExactSearcher.java
```diff
@@ -179,83 +181,112 @@ private KNNIterator getKNNIterator(LeafReaderContext leafReaderContext, ExactSea
...
-                spaceType
-            );
-        }
-
-        if (VectorDataType.BYTE == vectorDataType) {
-            final KNNVectorValues<byte[]> vectorValues = KNNVectorValuesFactory.getVectorValues(fieldInfo, reader);
+        } else if (VectorDataType.BYTE == vectorDataType) {
             if (isNestedRequired) {
-                return new NestedByteVectorIdsKNNIterator(
+                knnIterator = new NestedByteVectorIdsKNNIterator(
```
shatejas: nit: was wondering the reason to change short-circuit return to this?  

Looking around there isn't a unit test covering all cases for exact search class, it mostly covers radial cases so just wanted to see if we can kee the changes minimal

src/main/java/org/opensearch/knn/index/query/exactsearch/ExactSearcher.java
```diff
@@ -179,83 +181,112 @@ private KNNIterator getKNNIterator(LeafReaderContext leafReaderContext, ExactSea
         final SpaceType spaceType = FieldInfoExtractor.getSpaceType(modelDao, fieldInfo);
         boolean isNestedRequired = exactSearcherContext.getParentsFilter() != null;
 
+        // We need to create a new VectorValues instances as the new one will be used to iterate over the docIds in
+        // conjunction with Matched Docs.
+        final DocIdSetIterator matchedDocs = getMatchedDocsIterator(
+            exactSearcherContext.getMatchedDocsIterator(),
+            KNNVectorValuesFactory.getVectorValues(fieldInfo, reader).getVectorValuesIterator().getDocIdSetIterator()
+        );
```
shatejas, line 184: Overall I think `ConjunctionDISI` is great reuse of existing class, One thing worth thinking about is the overall correctness of all implementations of `KNNIterator`. These are public classes and if used standalone (outside of exactsearcher) they won't be functionally correct (can hit EOF)

Can we use the `ConjunctionUtils` inside the constructor of `KNNIterator` and make sure the case is covered in the unit test for those iterators individually? as far as `KNNVectorValues` are concerned we can pass in a supplier in the constructor

example
```
    public VectorIdsKNNIterator(
        @Nullable final DocIdSetIterator filterIdsIterator,
        final float[] queryVector,
        final Supplier<KNNFloatVectorValues> knnFloatVectorValuesSupplier,
        final SpaceType spaceType,
        ....
    ) throws IOException {   
       
       if (filterIdsIterator != null) {
           this.finalDocIdIterator = ConjunctionUtils.intersect(filterIdsIterator, 
                      knnFloatVectorValues.get().getIterator());
       }
        this.queryVector = queryVector;
        this.knnFloatVectorValues = knnFloatVectorValues.get();
        this.spaceType = spaceType;
       //everything else
    }
```

Let me know your thoughts on this

src/test/java/org/opensearch/knn/index/query/exactsearch/ExactSearcherTests.java
```diff
@@ -195,6 +197,10 @@ private void doTestRadialSearch_whenNoEngineFiles_thenSuccess(final boolean memo
 
             // Mocking float vector values
             KNNFloatVectorValues floatVectorValues = mock(KNNFloatVectorValues.class);
```
shatejas, line 275: It might be worth adding a unit test to cover the case, It might turn into a component test but since exact search is extensively used in various part want to make sure this component is extensively tested

src/test/java/org/opensearch/knn/integ/ExpandNestedDocsIT.java
```diff
@@ -194,6 +199,53 @@ public void testExpandNestedDocs_whenMultiShards_thenReturnCorrectResult() {
         }
     }
 
+    @SneakyThrows
+    public void testExpandNestedDocs_whenFewChildHasNoVectors_thenReturnCorrectResult() {
```
shatejas, line 204: was wondering if we want to add in a case which has filtering involved with missing vector values in doc

src/main/java/org/opensearch/knn/index/query/ExactSearcher.java
navneet1v: There is no specific reason of making the change. It got changed as I was working through the code. I can revert these changes. My thought was first to remove this short circuit and next time fix it once and for all. 

Let me know if you consider the revert of these lines.

src/main/java/org/opensearch/knn/index/query/exactsearch/ExactSearcher.java
navneet1v: > One thing worth thinking about is the overall correctness of all implementations of KNNIterator. These are public classes and if used standalone (outside of exactsearcher) they won't be functionally correct (can hit EOF)

In terms of correctness my thought process is right now also it is correct. As the caller its the responsibility to make sure that iterators provided are correct and can be iterated correctly. 

Another point here is KNNIterators are not used outside of ExactSearcher. I can make them all to package private because the only case where these iterators are used and should be used is for doing exact search.

src/test/java/org/opensearch/knn/index/query/exactsearch/ExactSearcherTests.java
navneet1v: > It might be worth adding a unit test to cover the case

@shatejas can you please elaborate here, what case you are talking about. I am able to get it from your comment here. :D

src/test/java/org/opensearch/knn/integ/ExpandNestedDocsIT.java
navneet1v: The interesting part is both filters and non filter cases work exactly same. Because expand nested docs is basically expanding the child doc for a parent docId. Now even if we add filter case here, since the parent docId will be matched due to any of child, expand nested docs will work exactly the same it will still get all the child docs and then hit same code path, since expanding of child docs is after filters are done.

navneet1v: added

src/test/java/org/opensearch/knn/index/query/exactsearch/ExactSearcherTests.java
navneet1v: added another test.

src/main/java/org/opensearch/knn/index/query/exactsearch/ExactSearcher.java
navneet1v: Moved all iterators and ExactSearch code to one java package and made classes package private to ensure that KNNIterator is not getting used outside of exact search.

src/main/java/org/opensearch/knn/index/query/ExactSearcher.java
navneet1v: Reverted all the code back.

Vikasht34 (approved):

Issue #3060, opened 2026-01-09
[BUG] MemoryOptimizedKNNWeight doesn't use updated k value for oversampling


MrFlap:
# Bug: MemoryOptimizedKNNWeight doesn't use updated k value for oversampling

## Summary
The `MemoryOptimizedKNNWeight.queryIndex()` method ignores the `k` parameter passed from `doANNSearch()`, causing incorrect behavior during oversampling scenarios where a different k value should be used for collecting results.

## Problem Description
In `MemoryOptimizedKNNWeight`, the `doANNSearch()` method receives a `k` parameter that may differ from the original query's k value (e.g., during oversampling). However, the `queryIndex()` method doesn't accept this parameter and always uses the collector manager created during construction with the original k value.

## Root Cause
1. The `queryIndex()` method signature doesn't include the `k` parameter
2. The method uses `knnCollectorManager` (created at construction time) instead of creating a new collector with the updated k value
3. This affects both non-nested (`TopKnnCollectorManager`) and nested (`DiversifyingNearestChildrenKnnCollectorManager`) cases

## Impact
- Oversampling doesn't work correctly as the collector still uses the original k value
- May return fewer results than expected during oversampling scenarios
- Affects search quality and recall in filtered search scenarios

## Code Location
**File:** `src/main/java/org/opensearch/knn/index/query/memoryoptsearch/MemoryOptimizedKNNWeight.java`

**Method:** `queryIndex()` (line ~150)

## Current Problematic Code
```java
private TopDocs queryIndex(
    final Object targetVector,
    final int cardinality,
    final int visitLimitWhenFilterExists,
    final LeafReaderContext context,
    final BitSet filterIdsBitSet,
    final SegmentReader reader,
    final KNNEngine knnEngine,
    final SpaceType spaceType
    // Missing k parameter here
) throws IOException {
    // ...
    // Always uses knnCollectorManager created at construction time
    final KnnCollector knnCollector = knnCollectorManager.newCollector(visitedLimit, DEFAULT_HNSW_SEARCH_STRATEGY, context);
    // ...
}
```

## Expected Behavior
The `queryIndex()` method should:
1. Accept the `k` parameter from `doANNSearch()`
2. Create a new collector manager with the updated k value when needed
3. Handle both non-nested and nested cases appropriately

## Proposed Solution
1. Add `k` parameter to `queryIndex()` method signature
2. Update all `queryIndex()` calls to pass the k parameter
[...]

PR #3061, opened 2026-01-10
Fix MemoryOptimizedKNNWeight to use expanded 'k' for disk based index.

0ctopus13prime:
### Description
In disk based index, we run 2 phased search for better recall. At the first phase, we first expand the given `k` [Code](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/query/rescore/RescoreContext.java#L81). 
Then use it to run approximate search, and rescore the collected vectors from the 1st phase with the full precision vectors. 

But when memory optimized search is enabled, the expanded `k` is not being used in search.
This is because, [MemoryOptimizedKNNWeight](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/query/memoryoptsearch/MemoryOptimizedKNNWeight.java) is created first with the given `k` and we do expand it later in [here](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java#L108).

When [MemoryOptimizedKNNWeight](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/query/memoryoptsearch/MemoryOptimizedKNNWeight.java) is created, internally it instantiates `KnnCollectorManager` with the original `k` given from user and being used for entire search for disk based index. 

Therefore, when `k` is expanded, `KnnCollectorManager` in [MemoryOptimizedKNNWeight](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/query/memoryoptsearch/MemoryOptimizedKNNWeight.java) should be re-instantiated with the expanded `k`.

For disk-based indexes, we run a two-phase search to improve recall. In the first phase, we expand the given k (see [Code](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/query/rescore/RescoreContext.java#L81)) and use the expanded value to perform the approximate search. Then, in the second phase, we rescore the collected vectors using full-precision vectors.

However, when memory-optimized search is enabled, the expanded k is not used. This happens because [MemoryOptimizedKNNWeight](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/query/memoryoptsearch/MemoryOptimizedKNNWeight.java) is created using the original k provided by the user, while the k expansion happens later. When [MemoryOptimizedKNNWeight](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/query/memoryoptsearch/MemoryOptimizedKNNWeight.java) is constructed, it internally creates a `KnnCollectorManager` using this original k, and that instance is used for the entire disk-based search.

Therefore, when k is expanded, the `KnnCollectorManager` inside [MemoryOptimizedKNNWeight](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/query/memoryoptsearch/MemoryOptimizedKNNWeight.java) must be reinstantiated using the expanded k, so that the memory-optimized path behaves consistently with the disk-based two-phase search logic.

### Related Issues
https://github.com/opensearch-project/k-NN/issues/3060

### Check List

### DIFF
--- a/qa/restart-upgrade/build.gradle
+++ b/qa/restart-upgrade/build.gradle
@@ -31,7 +31,6 @@ testClusters {
         } else {
             systemProperty "java.library.path", "${buildDir}/testclusters/${baseName}-0/distro/${knn_bwc_version_no_qualifier}-ARCHIVE/plugins/opensearch-knn/knnlib:${buildDir}/testclusters/${baseName}-0/distro/${knn_bwc_version_no_qualifier}-ARCHIVE/plugins/opensearch-knn/lib"
         }
-
     }
 }
 
@@ -174,6 +173,9 @@ task testAgainstOldCluster(type: StandaloneRestIntegTestTask) {
     nonInputProperties.systemProperty('tests.rest.cluster', "${-> testClusters."${baseName}".allHttpSocketURI.join(",")}")
     nonInputProperties.systemProperty('tests.clustername', "${-> testClusters."${baseName}".getName()}")
     systemProperty 'tests.security.manager', 'false'
+    
+    testClassesDirs = sourceSets.test.output.classesDirs
+    classpath = sourceSets.test.runtimeClasspath
 }
 
 // All nodes are upgraded to latest version and run the tests
@@ -324,4 +326,7 @@ task testRestartUpgrade(type: StandaloneRestIntegTestTask) {
     nonInputProperties.systemProperty('tests.rest.cluster', "${-> testClusters."${baseName}".allHttpSocketURI.join(",")}")
     nonInputProperties.systemProperty('tests.clustername', "${-> testClusters."${baseName}".getName()}")
     systemProperty 'tests.security.manager', 'false'
+
+    testClassesDirs = sourceSets.test.output.classesDirs
+    classpath = sourceSets.test.runtimeClasspath
 }
--- a/qa/rolling-upgrade/build.gradle
+++ b/qa/rolling-upgrade/build.gradle
@@ -45,6 +45,9 @@ task testAgainstOldCluster(type: StandaloneRestIntegTestTask) {
     nonInputProperties.systemProperty('tests.rest.cluster', "${-> testClusters."${baseName}".allHttpSocketURI.join(",")}")
     nonInputProperties.systemProperty('tests.clustername', "${-> testClusters."${baseName}".getName()}")
     systemProperty 'tests.security.manager', 'false'
+
+    testClassesDirs = sourceSets.test.output.classesDirs
+    classpath = sourceSets.test.runtimeClasspath
 }
 
 
@@ -67,6 +70,9 @@ task testAgainstOneThirdUpgradedCluster(type: StandaloneRestIntegTestTask) {
     nonInputProperties.systemProperty('tests.rest.cluster', "${-> testClusters."${baseName}".allHttpSocketURI.join(",")}")
     nonInputProperties.systemProperty('tests.clustername', "${-> testClusters."${baseName}".getName()}")
     systemProperty 'tests.security.manager', 'false'
+
+    testClassesDirs = sourceSets.test.output.classesDirs
+    classpath = sourceSets.test.runtimeClasspath
 }
 
 // Part of rolling upgrade. Upgrades the second node to new OpenSearch version with upgraded plugin version after the
@@ -86,6 +92,9 @@ task testAgainstTwoThirdsUpgradedCluster(type: StandaloneRestIntegTestTask) {
     nonInputProperties.systemProperty('tests.rest.cluster', "${-> testClusters."${baseName}".allHttpSocketURI.join(",")}")
     nonInputProperties.systemProperty('tests.clustername', "${-> testClusters."${baseName}".getName()}")
     systemProperty 'tests.security.manager', 'false'
+
+    testClassesDirs = sourceSets.test.output.classesDirs
+    classpath = sourceSets.test.runtimeClasspath
 }
 
 // Part of rolling upgrade. Upgrades the third node to new OpenSearch version with upgraded plugin version after the
@@ -105,4 +114,7 @@ task testRollingUpgrade(type: StandaloneRestIntegTestTask) {
     nonInputProperties.systemProperty('tests.rest.cluster', "${-> testClusters."${baseName}".allHttpSocketURI.join(",")}")
     nonInputProperties.systemProperty('tests.clustername', "${-> testClusters."${baseName}".getName()}")
     systemProperty 'tests.security.manager', 'false'
+
+    testClassesDirs = sourceSets.test.output.classesDirs
+    classpath = sourceSets.test.runtimeClasspath
 }


src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -171,6 +171,10 @@ public Explanation explain(LeafReaderContext context, int doc, float score) {
             : Explanation.match(score, highLevelExplanation, Explanation.match(score, leafLevelExplanation.toString()));
     }
 
+    public void updateStatus(int newTopK) {
```
navneet1v: 1. Lets rename this to `setFirstPassK`, and add java docs to explain when to use it.
2. Throw Exception from here so that we don't accidentally call this.

src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java
```diff
@@ -106,6 +106,7 @@ public Weight createWeight(IndexSearcher indexSearcher, ScoreMode scoreMode, flo
             boolean isShardLevelRescoringDisabled = KNNSettings.isShardLevelRescoringDisabledForDiskBasedVector(knnQuery.getIndexName());
             int dimension = knnQuery.getQueryVector().length;
             int firstPassK = rescoreContext.getFirstPassK(finalK, isShardLevelRescoringDisabled, dimension);
+            knnWeight.updateStatus(firstPassK);
```
navneet1v: should we add a check here to updateFirstPassK only if we are using MOS?

src/main/java/org/opensearch/knn/index/query/memoryoptsearch/MemoryOptimizedKNNWeight.java
```diff
@@ -50,29 +51,43 @@ public class MemoryOptimizedKNNWeight extends KNNWeight {
...
+        if (isRadialSearch == false) {
+            log.debug("Re-instantiating KNN collector manager with new top-k {}", newTopK);
+
+            if (parentsFilter == null) {
+                // Non-nested case
+                this.knnCollectorManager = new OptimisticKnnCollectorManager(newTopK, new TopKnnCollectorManager(newTopK, searcher));
+            } else {
+                // Nested case
+                this.knnCollectorManager = new DiversifyingNearestChildrenKnnCollectorManager(newTopK, parentsFilter, searcher);
+            }
```
navneet1v: call this function multiple times will lead to creation of objects. We should see how we can limit creation of these instances.

Vikasht34 (dismissed):
Thanks for taking suggestion for late declaration on Weight, Looks Good to me !!

Vikasht34 (approved):

src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java
```diff
@@ -85,36 +86,51 @@ public class NativeEngineKnnVectorQuery extends Query {
...
+        // Create Weight depending on whether 2-phase search is needed
+        final boolean isShardLevelRescoringDisabled = KNNSettings.isShardLevelRescoringDisabledForDiskBasedVector(knnQuery.getIndexName());
+        final Integer firstPassKFor2PhaseSearch = getFirstPassK(isShardLevelRescoringDisabled);
+        final IOSupplier<KNNWeight> weightSupplier;
+        if (firstPassKFor2PhaseSearch == null) {
+            weightSupplier = () -> (KNNWeight) knnQuery.createWeight(indexSearcher, scoreMode, 1);
+        } else {
+            // Create weight with expanded `k`.
+            weightSupplier = () -> (KNNWeight) knnQuery.createWeight(indexSearcher, scoreMode, 1, firstPassKFor2PhaseSearch);
+        }
```
navneet1v: can we move this code in a separate function named `getKNNWeightSupplier()` ?

src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java
```diff
@@ -153,6 +176,18 @@ public Weight createWeight(IndexSearcher indexSearcher, ScoreMode scoreMode, flo
...
+        final RescoreContext rescoreContext = knnQuery.getRescoreContext();
+        if (rescoreContext != null && rescoreContext.isRescoreEnabled()) {
+            // We need 2-phase search where using expanded `k` for the first stage search.
+            final int dimension = knnQuery.getQueryVector().length;
+            return rescoreContext.getFirstPassK(knnQuery.getK(), isShardLevelRescoringDisabled, dimension);
+        }
+
+        // We don't need 2-phase, hence there's no first pass k.
+        return null;
+    }
```
navneet1v, line 202: rather than using `null` why not use `-1`.

src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java
```diff
@@ -85,36 +86,51 @@ public class NativeEngineKnnVectorQuery extends Query {
...
         }
+
+        // Run search
+        final IndexReader reader = indexSearcher.getIndexReader();
         List<LeafReaderContext> leafReaderContexts = reader.leaves();
         List<PerLeafResult> perLeafResults;
-        RescoreContext rescoreContext = knnQuery.getRescoreContext();
         final int finalK = knnQuery.getK();
-        if (rescoreContext == null || !rescoreContext.isRescoreEnabled()) {
+        if (firstPassKFor2PhaseSearch == null) {
```
navneet1v: why we need this change? can we abstract this in a small function like `isRescoreRequired()`

0ctopus13prime:
@navneet1v 
Thank you for the suggestion.
Can I cover it in the next PR and merge this?

src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java
0ctopus13prime: Updated in the next rev

src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java
0ctopus13prime: Updated in the next rev

src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java
0ctopus13prime: I keep the return type as null as I think it's natural to return null to represent 'No value' instead of the magic number -1.

navneet1v: Yeah I understand that. You can put any value here. My thought was use primitives as much as possible rather than using boxed primitives. :)

0ctopus13prime: I have a slightly different idea. Returning null and fail fast if possible would be better than `-1`. 
Since the returned value should not used when it's null, instead `knnQuery.getK()` should be used.
If that's the case, then I think it's better to thrown NPE so that we can find the bug much earlier as it might swallow -1 value somewhere.

Vikasht34: @navneet1v  do u still see this as blocker for this PR, we need to merge for AVR.

PR #3063, opened 2026-01-13
Fix MemoryOptimizedSearchWarmup vector search function selection for compressed indices

MrFlap:
## Description

This PR fixes a bug in `MemoryOptimizedSearchWarmup` where the logic for selecting between `byte[]` and `float[]` vector search was flawed. For compressed indices using quantization (e.g., 1-bit, 2-bit, 4-bit with BinaryHNSWIndex), the warmup would incorrectly use `float[]` search when the VectorDataType was FLOAT, even though the index required `byte[]` search.

## Problem

The original implementation in `MemoryOptimizedSearchWarmup` used a simple check based on `VectorDataType`:
- If `VectorDataType == BINARY` or `BYTE` → use `byte[]` search
- Otherwise → use `float[]` search

This logic failed for quantized indices where:
- The original vector type is FLOAT
- But quantization transforms vectors to BINARY for storage/search
- The warmup would use `float[]` search, but actual queries would use `byte[]` search
- This mismatch meant warmup didn't properly prepare the JVM for the actual search operations

## Solution

Created a unified `SearchVectorTypeResolver` utility class that determines the correct search function based on field configuration:

1. **Check quantization configuration first**: Uses `QuantizationService.getVectorDataTypeForTransfer()` to detect if quantization is configured
   - If quantization returns `BINARY` → use `byte[]` search
2. **Check native vector type**: If `VectorDataType` is `BYTE` or `BINARY` → use `byte[]` search  
3. **Default to float search**: For regular FLOAT vectors and ADC cases

This ensures both `MemoryOptimizedKNNWeight` (actual search) and `MemoryOptimizedSearchWarmup` use identical logic.

### New Files
- `VectorSearchFunction.java` - Functional interface for vector search operations
- `SearchVectorTypeResolver.java` - Utility class to determine correct search function
- `SearchVectorTypeResolverTests.java` - Comprehensive parameterized unit tests (9 tests)
- `MemoryOptimizedSearchIndexWarmupTests.java` - Comprehensive parameterized integration tests (27 tests)

### Modified Files
- `MemoryOptimizedKNNWeight.java` - Refactored to use `SearchVectorTypeResolver`
- `MemoryOptimizedSearchWarmup.java` - Refactored to use `SearchVectorTypeResolver`

## Testing

Added comprehensive parameterized tests in two test classes:

### SearchVectorTypeResolverTests (9 tests)
Unit tests for the `SearchVectorTypeResolver` utility class covering:
- 1-bit, 2-bit, and 4-bit quantized indices
- Quantized indices with ADC enabled
- Quantized indices with random rotation enabled
- Native BYTE vector type
- Native BINARY vector type
- Regular FLOAT vectors (no quantization)
- ADC transformed FLOAT vectors

All 9 test configurations verify that the resolver returns the correct search function.

### MemoryOptimizedSearchIndexWarmupTests (27 tests)
Integration tests for the warmup functionality covering all 9 index configurations with 3 test scenarios each:
- **testWarmupWithDifferentIndexConfigurations**: Verifies warmup completes without crashing
- **testWarmupWithVectorData**: Tests warmup with actual vector documents
- **testWarmupWithEmptyIndex**: Tests warmup with empty index (no documents)

All 27 tests (9 configurations × 3 scenarios) pass successfully, ensuring the warmup works correctly for all index types.

## Benefits

1. **Correctness**: Warmup now uses the same search function as actual queries
2. **Maintainability**: Single source of truth for search function selection logic
3. **Type Safety**: Returns function directly instead of boolean flag
4. **Testability**: Comprehensive test coverage for all index configurations

## Backward Compatibility

This change is backward compatible - it fixes incorrect behavior without changing any public APIs or breaking existing functionality.

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/query/memoryoptsearch/SearchVectorTypeResolver.java
+++ b/src/main/java/org/opensearch/knn/index/query/memoryoptsearch/SearchVectorTypeResolver.java
@@ -1,0 +1,67 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.knn.index.query.memoryoptsearch;
+
+import org.apache.lucene.index.FieldInfo;
+import org.apache.lucene.index.SegmentReader;
+import org.opensearch.knn.index.VectorDataType;
+import org.opensearch.knn.index.quantizationservice.QuantizationService;
+
+/**
+ * Utility class to determine the correct vector search function for memory-optimized search operations.
+ * This ensures consistency between actual search (MemoryOptimizedKNNWeight) and warmup (MemoryOptimizedSearchWarmup).
+ */
+public class SearchVectorTypeResolver {
+
+    /**
+     * Returns the appropriate search function based on the field configuration.
+     *
+     * @param reader the segment reader
+     * @param fieldInfo the field information
+     * @param vectorDataType the vector data type from the field
+     * @return a VectorSearchFunction that performs the appropriate search
+     */
+    public static VectorSearchFunction getSearchFunction(
+        final SegmentReader reader,
+        final FieldInfo fieldInfo,
+        final VectorDataType vectorDataType
+    ) {
+        final boolean useByteSearch = shouldUseByteVectorSearch(fieldInfo, vectorDataType);
+
+        if (useByteSearch) {
+            return (fieldName, vector, knnCollector, acceptDocs) -> reader.getVectorReader()
+                .search(fieldName, (byte[]) vector, knnCollector, acceptDocs);
+        } else {
+            return (fieldName, vector, knnCollector, acceptDocs) -> reader.getVectorReader()
+                .search(fieldName, (float[]) vector, knnCollector, acceptDocs);
+        }
+    }
+
+    /**
+     * Determines whether to use byte[] or float[] for the search operation based on the field configuration.
+     *
+     * @param fieldInfo the field information
+     * @param vectorDataType the vector data type from the field
+     * @return true if byte[] search should be used, false if float[] search should be used
+     */
+    private static boolean shouldUseByteVectorSearch(final FieldInfo fieldInfo, final VectorDataType vectorDataType) {
+        // Check if quantization is configured - this determines the search vector type
+        final QuantizationService<?, ?> quantizationService = QuantizationService.getInstance();
+        final VectorDataType transferDataType = quantizationService.getVectorDataTypeForTransfer(fieldInfo);
+        if (transferDataType != null) {
+            // Quantization is configured - check if it uses BINARY transfer
+            return transferDataType == VectorDataType.BINARY;
+        }
+
+        // When data_type is set to byte or binary
+        if (vectorDataType == VectorDataType.BINARY || vectorDataType == VectorDataType.BYTE) {
+            return true;
+        }
+
+        // Default float case (includes ADC which uses float[] search)
+        return false;
+    }
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 * Correct ef_search parameter for Lucene engine and override mergeLeafResults to return top K results [#3037](https://github.com/opensearch-project/k-NN/pull/3037)
 * Fix efficient filtering in nested k-NN queries [#2990](https://github.com/opensearch-project/k-NN/pull/2990)
 * Changed warmup seek to use long instead of int to avoid overflow [#3067](https://github.com/opensearch-project/k-NN/pull/3067)
+* Fix MemoryOptimizedSearchWarmup vector search function selection for compressed indices [#3063](https://github.com/opensearch-project/k-NN/pull/3063)
 
 ### Refactoring
 * Change ordering of build task and added tests to catch uninitialized libraries [#3024](https://github.com/opensearch-project/k-NN/pull/3024)
--- a/src/main/java/org/opensearch/knn/index/query/memoryoptsearch/MemoryOptimizedKNNWeight.java
+++ b/src/main/java/org/opensearch/knn/index/query/memoryoptsearch/MemoryOptimizedKNNWeight.java
@@ -24,7 +24,6 @@ import org.apache.lucene.search.knn.TopKnnCollectorManager;
 import org.apache.lucene.util.BitSet;
 import org.apache.lucene.util.BitSetIterator;
 import org.apache.lucene.util.Bits;
-import org.apache.lucene.util.Version;
 import org.opensearch.knn.index.SpaceType;
 import org.opensearch.knn.index.VectorDataType;
 import org.opensearch.knn.index.engine.KNNEngine;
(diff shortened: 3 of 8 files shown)


src/main/java/org/opensearch/knn/index/query/memoryoptsearch/SearchVectorTypeResolver.java
```diff
@@ -0,0 +1,67 @@
...
+    public static VectorSearchFunction getSearchFunction(
+        final SegmentReader reader,
+        final FieldInfo fieldInfo,
+        final VectorDataType vectorDataType
+    ) {
+        final boolean useByteSearch = shouldUseByteVectorSearch(fieldInfo, vectorDataType);
+
+        if (useByteSearch) {
+            return (fieldName, vector, knnCollector, acceptDocs) -> reader.getVectorReader()
+                .search(fieldName, (byte[]) vector, knnCollector, acceptDocs);
```
Vikasht34, line 36: How this is going to work with Binary Quantization, For Binary Quantization useByteSearch is going to be true , but we don't caste Query to Byte Vector.

Vikasht34 (approved):

Vikasht34 (commented):
We need to have integs around Binary Quantization , Byte and Float to cover the New Functional Interface. I feel casting is wrong.

src/main/java/org/opensearch/knn/index/query/memoryoptsearch/SearchVectorTypeResolver.java
Vikasht34: NVM I found the place where we are quantizing , looks good

Vikasht34:
> Hey @MrFlap, thank you for the PR. The purpose of calling `search` method with a dummy query is to trigger [loadMemoryOptimizedSearcherIfRequired](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java#L290) method in NativeEngines990KnnVectorsReader. In there, it will partial load `.faiss` file into memory so we can serve future subsequent queries right away without this lazy loading.
> 
> I think we can just catch UnsupportedOperationException and that's it.
> 
> ```
>         try {
>             final String dataTypeStr = field.getAttribute(VECTOR_DATA_TYPE_FIELD);
>             if (dataTypeStr == null) {
>                 return true;
>             }
>             final VectorDataType vectorDataType = VectorDataType.get(dataTypeStr);
>             // We expect a NPE here:
>             if (vectorDataType == VectorDataType.FLOAT) {
>                 segmentReader.getVectorReader().search(field.getName(), (float[]) null, null, null);
>             } else {
>                 segmentReader.getVectorReader().search(field.getName(), (byte[]) null, null, null);
>             }
>         } catch (UnsupportedOperationException | NullPointerException ignored) {} <-------- Catch UnsupportedOperationException
> ```
> 
> I know this is somewhat hacky-ish way to do it, but we don't at least maintain two code pieces for getting the actual type of underlying index.
> 
> ## Why this works?
> [loadMemoryOptimizedSearcherIfRequired](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java#L290) method called at the beginning of `search` method as you can see below:
> 
> ```
>     @Override
>     public void search(String field, float[] target, KnnCollector knnCollector, AcceptDocs acceptDocs) throws IOException {
>         // TODO: This is a temporary hack where we are using KNNCollector to initialize the quantization state.
>         if (knnCollector instanceof QuantizationConfigKNNCollector) {
>             ...
>             return;
>         }
>         // vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv
>         if (trySearchWithMemoryOptimizedSearch(field, target, knnCollector, acceptDocs, true)) {
>             return;
>         }
> 
>         throw new UnsupportedOperationException("Search functionality using codec is not supported with Native Engine Reader");
>     }
> 
> 
>     private boolean trySearchWithMemoryOptimizedSearch(
>         final String field,
>         final Object target,
>         final KnnCollector knnCollector,
>         final AcceptDocs acceptDocs,
>         final boolean isFloatVector
>     ) throws IOException {
>         // Try with memory optimized searcher vvvvvvvvvvvvvvvvvvvvvvvvvv
>         final VectorSearcher memoryOptimizedSearcher = loadMemoryOptimizedSearcherIfRequired(field);
> 
>         if (memoryOptimizedSearcher != null) {
>             if (isFloatVector) {
>                 memoryOptimizedSearcher.search((float[]) target, knnCollector, acceptDocs);
>             } else {
>                 memoryOptimizedSearcher.search((byte[]) target, knnCollector, acceptDocs);
>             }
>             return true;
>         }
>         return false;
>     }
> 


@0ctopus13prime I am super curious to know , what you are suggesting is, how going to fix the issue.

Vikasht34:
> @Vikasht34 sure,
> 
> ```
> // We expect an NPE here
> if (vectorDataType == VectorDataType.FLOAT) {
>     segmentReader.getVectorReader().search(field.getName(), (float[]) null, null, null);
> } else {
>     segmentReader.getVectorReader().search(field.getName(), (byte[]) null, null, null);
> }
> ```
> 
> The code above internally creates a FaissMemoryOptimizedSearcher and then calls FaissMemoryOptimizedSearcher.search(...).
> 
> During warm-up, our goal is simply to instantiate FaissMemoryOptimizedSearcher so that it opens the IndexInput and performs all required bookkeeping. We do not really care about the actual search results at this stage, which is why we currently pass a null query.
> 
> However, for quantized indices, the VectorDataType is VectorDataType.FLOAT. Because we pass a float[], FaissMemoryOptimizedSearcher attempts to create FloatVectorValues (see [code](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissMemoryOptimizedSearcher.java#L88)). At that point, it throws an UnsupportedOperationException, because binary indices do not support FloatVectorValues.
> 
> This search failure is not a problem for warm-up. By the time the exception is thrown, FaissMemoryOptimizedSearcher has already been created, which means our goal, partial loading of the .faiss index has been achieved.
> 
> Therefore, a simple solution is to just catch UnsupportedOperationException for binary indices and ignore it:
> 
> ```java
> try {
>     final String dataTypeStr = field.getAttribute(VECTOR_DATA_TYPE_FIELD);
>     if (dataTypeStr == null) {
>         return true;
>     }
> 
>     final VectorDataType vectorDataType = VectorDataType.get(dataTypeStr);
> 
>     // We expect an NPE here
>     if (vectorDataType == VectorDataType.FLOAT) {
>         segmentReader.getVectorReader().search(field.getName(), (float[]) null, null, null);
>     } else {
>         segmentReader.getVectorReader().search(field.getName(), (byte[]) null, null, null);
>     }
> } catch (***** UnsupportedOperationException ***** | NullPointerException ignored) {
>     // intentionally ignored
> }
> ```
> 
> Ideally, we would verify that the UnsupportedOperationException originates from FaissMemoryOptimizedSearcher, but I'm just giving my idea.
> 
> I agree this approach is not ideal and is somewhat hacky. However, I haven’t found a simpler way to trigger partial loading of the .faiss index. Similar to what @MrFlap did in their PR, we could precisely target each case, but I’m hesitant to introduce changes and maintenance duties in the Weight class solely for warm-up purposes 😅

Makes Sense .. yeah with this we can localized the changes to Memory Optimized only , It works . The only caviat I am seeing is we should not ignore actual error by thinking this is expoected , so we should add some mechanism to localize the error also so that we can identify them,

MrFlap:
@Vikasht34 @0ctopus13prime I can make the other version in another PR to see if we like the shorter one more. I would try to isolate the exception to the function that throws it by using the stack trace. I think this will be equally as hard to maintain as the original method if not harder.

What happens when we change the naming of the function/class? We will have to probably do a string comparison, so the class name will need to be hardcoded.

What happens if somebody changes where the exception is caught? This may change it so that we will no longer actually do the warmup.

Vikasht34:
> @Vikasht34 @0ctopus13prime I can make the other version in another PR to see if we like the shorter one more. I would try to isolate the exception to the function that throws it by using the stack trace. I think this will be equally as hard to maintain as the original method if not harder.
> 
> What happens when we change the naming of the function/class? We will have to probably do a string comparison, so the class name will need to be hardcoded.
> 
> What happens if somebody changes where the exception is caught? This may change it so that we will no longer actually do the warmup.

As we disscussed .

1. We will Localise the error spoecifically for this case by creating new error case so that we don't have to haerdcode. 

2. We need testing for 1. 1X, 2X, 4X, 8X, 16X, 32X with Adc RR and Separate Binary VECTOR as well.

Issue #2284, opened 2024-11-21
Add IT and BWC tests with Indices containing both Vector and Non Vector documents


navneet1v:
## Description
Currently in k-NN plugin all the ITs and BWC created has indices with vector fields and all the documents contain vector field. But in production indices it is not necessary that a k-NN index documents will always have the vector field in it or to say all the vector fields in it. Due to these kind of tests being missing we are not able to catch issues which are fixed in these PRs:
1. NPE exception during Disk based vector search due to segment not containing a vector field. Ref: https://github.com/opensearch-project/k-NN/issues/2277
2. The [feature](https://github.com/opensearch-project/k-NN/pull/1946) of releasing the memory during closing of the index introduced a bug where if a segment has a knn_vector field but no docs with this field present, then an index OOB exception will be thrown. This was fixed in  https://github.com/opensearch-project/k-NN/pull/2182.
```
Caused by: NotSerializableExceptionWrapper[index_out_of_bounds_exception: Index 0 out of bounds for length 0]
    at jdk.internal.util.Preconditions.outOfBounds(Preconditions.java:100)
    at jdk.internal.util.Preconditions.outOfBoundsCheckIndex(Preconditions.java:106)
    at jdk.internal.util.Preconditions.checkIndex(Preconditions.java:302)
    at java.util.Objects.checkIndex(Objects.java:385)
    at java.util.ArrayList.get(ArrayList.java:427)
    at org.opensearch.knn.index.codec.KNN80Codec.KNN80DocValuesProducer.<init>(KNN80DocValuesProducer.java:78)
    at org.opensearch.knn.index.codec.KNN80Codec.KNN80DocValuesFormat.fieldsProducer(KNN80DocValuesFormat.java:44)
    at org.apache.lucene.index.SegmentDocValues.newDocValuesProducer(SegmentDocValues.java:52)
```

## Proposal
To catch the above issues during PRs we should add tests(BWC and ITs) for all 3 engines and disk based vector search. For 1, I added the integration tests with the fix https://github.com/opensearch-project/k-NN/blob/2d1a4080d5b1601bf3362fecd85384348af1f326/src/test/java/org/opensearch/knn/integ/ModeAndCompressionIT.java#L225 . We need to similar thing for BWC and other engines.

### Tests to be added
1. BWC test for all versions where an index has 10 docs where 9 contain vector fields and 1 is no vector field. The ingestion should happen such that document with no vector field gets its own segment. ref proposal section.
[...]

Issue #2277, opened 2024-11-17
[BUG] NPE while calling ANN search when deleted docs or a segment with no vector field present in the index


navneet1v:
## Description
When a k-NN index(with on_disk mode) has deleted documents/ in it, then while doing the search the search is failing with NPE. 
### Impacted cases, refer below sections for workarounds:
1. This issue will happen only in case of on_disk mode is set to true with rescoring.
2. If on_disk mode index has deleted docs and segment not commited.
3. If on_disk mode index has some segments which has no vector field where query is happening.
4. If on_disk mode index has shards which has no vector field query is happening on the field. 

Please refer the below steps for reproduction.

### Create Index
```
PUT my-knn-index-1
{
  "settings": {
    "index": {
      "number_of_shards": 1,
      "number_of_replicas": 0,
      "knn": true
    }
  },
  "mappings": {
    "properties": {
      "my_vector": {
        "type": "knn_vector",
        "dimension": 8,
        "mode": "on_disk"
        }
      }
    }
  }
}
```

### Ingest 2 documents
```
PUT _bulk?refresh=true
{ "index": { "_index": "my-knn-index-1", "_id": "1" } }
{"my_vector":[1,1,1,1,1,1,1,1]}
{ "index": { "_index": "my-knn-index-1", "_id": "2" } }
{"my_vector":[1,1,1,1,1,1,1,1]}
```

### Search working as expected
```
GET my-knn-index-1/_search
{
  "query": {
    "knn": {
      "my_vector": {
        "vector": [
          1,
          1,
          1,
          1,
          1,
          1,
          1,
          1
        ],
        "k": 3
      }
    }
  }
}

```

### Delete a document
```
DELETE my-knn-index-1/_doc/1
```
### Search Again with error
```
GET my-knn-index-1/_search
{
  "query": {
    "knn": {
      "my_vector": {
        "vector": [
          1,
          1,
          1,
          1,
          1,
          1,
          1,
          1
        ],
        "k": 3
      }
    }
  }
}
```

### Error Response
```
{
  "error": {
    "root_cause": [
      {
        "type": "null_pointer_exception",
        "reason": "Cannot invoke \"org.apache.lucene.index.FieldInfo.getAttribute(String)\" because \"fieldInfo\" is null"
      }
    ],
    "type": "search_phase_execution_exception",
    "reason": "all shards failed",
    "phase": "query",
    "grouped": true,
    "failed_shards": [
      {
        "shard": 0,
[...]

PR #3064, opened 2026-01-13
Add IT and bwc test with indices containing both vector and non-vector docs

anntians:
### Description
Referencing the proposal section of the github issue (https://github.com/opensearch-project/k-NN/issues/2284), we want to test existing functionality in ITs and BWC tests with indices containing both vector and non-vector docs. The core problem we want to test for is segment has a knn_vector field but no docs with this field present. More specifically, these are the test cases we want to add:

1. Index vector and non-vector docs, then delete a vector doc, then check search. As mentioned in issue (https://github.com/opensearch-project/k-NN/issues/2277), when docs are deleted, a new segment is created with deleted docs in it, but there are no docs present in this segment. So this tests deleted doc segment in an index with mixed-fields. This test is already added as `testCompressionIndexWithNonVectorFieldsSegment_whenValid_ThenSucceed` in `ModeAndCompressionIT.java` (ref github issue), in this PR we add this test for BWC and other engines, etc. This validates handling of segments that lost their vector data through deletion.
2. Index vector, flush, index non-vector, delete a vector doc, flush, then check search. This is similar to above but ensures that the non-vector doc has its own segment. This way we can also test functionality when segment doesn't contain the vector field. This test case is referenced in section `Tests to be added` in the github issue. This validates handling of segments that never had vector data from creation.
3. Index doc with both vector and non-vector field, update doc to remove vector field, check search. This is similar to test case 2 where we have an isolated segment with non-vector docs, just we get to that state through an update operation. This validates handling of segments that lost their vector data through update.

These test cases are added to ITs and BWC to test vector search with Faiss, filters tests with Faiss, disk-based vector search with default compression, and Lucene engine tests.

### Related Issues
Resolves https://github.com/opensearch-project/k-NN/issues/2284

### Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,9 +23,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 * Fix nested docs and exact search query when some docs has no vector field present. [#3051](https://github.com/opensearch-project/k-NN/pull/3051)
 * Changed warmup seek to use long instead of int to avoid overflow [#3067](https://github.com/opensearch-project/k-NN/pull/3067)
 * Fix memory-optimized-search reentrant search bug in byte index. [#3071](https://github.com/opensearch-project/k-NN/pull/3071)
+* Fix MemoryOptimizedKNNWeight to use expanded 'k' for disk based index. [#3061](https://github.com/opensearch-project/k-NN/pull/3061)
+* Added new exception type to signify expected warmup behavior [#3070](https://github.com/opensearch-project/k-NN/pull/3070)
+* Add derived source handling for when field is excluded in source on indexing [#3049](https://github.com/opensearch-project/k-NN/pull/3049)
 
 ### Refactoring
 * Change ordering of build task and added tests to catch uninitialized libraries [#3024](https://github.com/opensearch-project/k-NN/pull/3024)
 
 ### Enhancements
 * Index setting to disable exact search after ANN Search with Faiss efficient filters [#3022](https://github.com/opensearch-project/k-NN/pull/3022)
+* Bulk SIMD for FP16 distance computation [#3071](https://github.com/opensearch-project/k-NN/pull/3071)
--- a/jni/cmake/init-simd.cmake
+++ b/jni/cmake/init-simd.cmake
@@ -5,6 +5,7 @@
 
 include(CheckCXXSourceCompiles)
 
+
 # Allow user overrides
 if(NOT DEFINED AVX2_ENABLED)
     set(AVX2_ENABLED true)   # set default value as true if the argument is not set
@@ -68,7 +69,11 @@ elseif(${CMAKE_SYSTEM_PROCESSOR} MATCHES "aarch64" OR ${CMAKE_SYSTEM_PROCESSOR} 
     endif()
 
 elseif(${CMAKE_SYSTEM_NAME} STREQUAL "Linux" AND AVX512_SPR_ENABLED)
-    set(CMAKE_REQUIRED_FLAGS "-mavx512f -mavx512fp16 -mf16c")
+    # -mavx512f	: 512-bit Registers & FMA. Ex: __m512, _mm512_fmadd_ps
+    # -mavx512bw : 16-bit Masking. Ex: __mmask16, mask generation
+    # -mavx512vl : 256-bit "Sub-vector" support. Ex: _mm256_maskz_loadu_epi16
+    # -mavx512fp16 : Native FP16 / EVEX Conversion.
+    set(CMAKE_REQUIRED_FLAGS "-mavx512f -mavx512bw -mavx512vl -mavx512fp16")
     check_cxx_source_compiles("
         #include <immintrin.h>
         int main() {
@@ -83,7 +88,7 @@ elseif(${CMAKE_SYSTEM_NAME} STREQUAL "Linux" AND AVX512_SPR_ENABLED)
     if(HAVE_AVX512_SPR_COMPILER)
         set(KNN_HAVE_AVX512_SPR ON)
         set(SIMD_OPT_LEVEL "avx512_spr")
-        set(SIMD_FLAGS -mavx512f -mavx512fp16 -mf16c)
+        set(SIMD_FLAGS -mavx512f -mavx512bw -mavx512vl -mavx512fp16)
         add_definitions(-DKNN_HAVE_AVX512_SPR)
         message(STATUS "[SIMD] AVX512_SPR supported by compiler.")
     else()
@@ -91,7 +96,10 @@ elseif(${CMAKE_SYSTEM_NAME} STREQUAL "Linux" AND AVX512_SPR_ENABLED)
     endif()
 
 elseif(${CMAKE_SYSTEM_NAME} STREQUAL "Linux" AND AVX512_ENABLED)
-    set(CMAKE_REQUIRED_FLAGS "-mavx512f -mf16c")
+    # -mavx512f	: 512-bit Registers & FMA. Ex: __m512, _mm512_fmadd_ps
+    # -mavx512bw : 16-bit Masking. Ex: __mmask16, mask generation
+    # -mavx512vl : 256-bit "Sub-vector" support. Ex: _mm256_maskz_loadu_epi16
+    set(CMAKE_REQUIRED_FLAGS "-mavx512f -mavx512vl -mavx512bw")
     check_cxx_source_compiles("
         #include <immintrin.h>
         int main() {
@@ -105,7 +113,7 @@ elseif(${CMAKE_SYSTEM_NAME} STREQUAL "Linux" AND AVX512_ENABLED)
     if(HAVE_AVX512_COMPILER)
         set(KNN_HAVE_AVX512 ON)
         set(SIMD_OPT_LEVEL "avx512") # Keep optimization level as avx512 to improve performance on Linux. This is not present on mac systems, and presently not supported on Windows OS.
-        set(SIMD_FLAGS -mavx512f -mf16c)
+        set(SIMD_FLAGS -mavx512f -mavx512vl -mavx512bw)
         add_definitions(-DKNN_HAVE_AVX512)
         message(STATUS "[SIMD] AVX512 + F16C supported by compiler.")
     else()
--- a/jni/src/simd/similarity_function/arm_neon_simd_similarity_function.cpp
+++ b/jni/src/simd/similarity_function/arm_neon_simd_similarity_function.cpp
@@ -1,0 +1,181 @@
+#ifndef __ARM_FEATURE_FP16_VECTOR_ARITHMETIC
+#define __ARM_FEATURE_FP16_VECTOR_ARITHMETIC 1
+#endif
+#include <arm_neon.h>
+#include <algorithm>
+#include <array>
+#include <cstddef>
+#include <stdint.h>
+#include <cmath>
+#include <iostream>
+
+#include "simd_similarity_function_common.cpp"
+#include "faiss_score_to_lucene_transform.cpp"
+
+
+
+//
+// FP16
+//
+
+template <BulkScoreTransform BulkScoreTransformFunc, ScoreTransform ScoreTransformFunc>
+struct ArmNeonFP16MaxIP final : BaseSimilarityFunction<BulkScoreTransformFunc, ScoreTransformFunc> {
+    void calculateSimilarityInBulk(SimdVectorSearchContext* srchContext,
+                                   int32_t* internalVectorIds,
+                                   float* scores,
+                                   const int32_t numVectors) {
+        // Bulk inner product with 4 batch
+        int32_t processedCount = 0;
+        constexpr int32_t vecBlock = 4;
+        const uint8_t* vectors[vecBlock];
+        const auto* queryPtr = (const float*) srchContext->queryVectorSimdAligned;
+        const int32_t dim = srchContext->dimension;
+        constexpr int32_t dimensionBatch = 8;
+
+        for ( ; (processedCount + vecBlock) <= numVectors ; processedCount += vecBlock) {
+            srchContext->getVectorPointersInBulk((uint8_t**)vectors, &internalVectorIds[processedCount], vecBlock);
+
+            // Score accumulator per each vector
+            float32x4_t acc0 = vdupq_n_f32(0.0f);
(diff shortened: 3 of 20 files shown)


qa/restart-upgrade/src/test/java/org/opensearch/knn/bwc/IndexingIT.java
```diff
@@ -731,4 +731,129 @@ public void testDiskBasedMergeBWC() throws Exception {
             deleteKNNIndex(indexName);
         }
     }
+
+    public void testMixedFieldsWithFaissRestartUpgrade() throws Exception {
```
Vikasht34, line 740: Thanks for Adding all these tests , Which covers most of it, But Do u think we need to add Multi Shard with 2-3 fields with K large and small to cover the correctness

src/test/java/org/opensearch/knn/integ/IndexIT.java
```diff
@@ -70,6 +71,188 @@ public void testFaissHnsw_when1000Data_thenRecallIsAboveNinePointZero() {
         }
     }
 
+    @SneakyThrows
+    public void testIndexWithNonVectorFields_whenValid_thenSucceed() {
```
Vikasht34, line 80: Do u think we can have one more test where we intially have vector Fierld then in update remove that vector feild to capture to case.

src/test/java/org/opensearch/knn/integ/FilteredSearchANNSearchIT.java
```diff
@@ -56,4 +57,121 @@ public void testFilteredSearchWithFaissHnsw_whenFiltersMatchAllDocs_thenReturnCo
...
+
+        Response response = searchKNNIndex(INDEX_NAME, query, 3);
+        assertOK(response);
+        String entity = EntityUtils.toString(response.getEntity());
+        List<String> docIds = parseIds(entity);
+        assertTrue(docIds.size() > 0);
+    }
+
+    @SneakyThrows
+    public void testMixedSegmentsFilteredSearch_whenValid_thenSucceed() {
```
Vikasht34, line 161: Since we are adding mixed feild, we should add case where filter matches 0 docs , filter matches only non-vector docs.

qa/restart-upgrade/src/test/java/org/opensearch/knn/bwc/IndexingIT.java
anntians: The goal of the github issue is to test segment-level handling when segments have no docs with vector fields. The current tests ensure segment isolation and validates this. Not sure if testing multi-shard/multi-field scenario would bring additional coverage. Thoughts @navneet1v ?

src/test/java/org/opensearch/knn/integ/IndexIT.java
anntians: Do you mean we add a vector doc with doc id 1, then update doc id 1 with a non-vector doc? Is it possible to update a doc from vector field to a non-vector field?

Vikasht34: What I meant was to delete that filed now there is no vector vector feild in Doc.

qa/restart-upgrade/src/test/java/org/opensearch/knn/bwc/IndexingIT.java
```diff
@@ -731,4 +731,129 @@ public void testDiskBasedMergeBWC() throws Exception {
...
+            addNonKNNDoc(testIndex, String.valueOf(NUM_DOCS + 1), "description", "Test document");
+            deleteKnnDoc(testIndex, "0");
+            flush(testIndex, true);
+        } else {
+            validateKNNSearch(testIndex, TEST_FIELD, DIMENSIONS, NUM_DOCS - 1, K);
+            deleteKNNIndex(testIndex);
+        }
+    }
+
+    public void testMixedSegmentsWithFaissRestartUpgrade() throws Exception {
```
kotwanikunal, line 822: nit: Can we add in comments describing the scenario we are testing?
It will be easier for folks to understand the reasoning behind tests with text only segments without coming back to the PR.

CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Maintenance
 * Added gradle task to generate task dependency graph ([#3032](https://github.com/opensearch-project/k-NN/pull/3032))
 * Added new gradle task validateLibraryUsage so that System.loadLibrary cannot be run outside KNNLibraryLoader ([#3033](https://github.com/opensearch-project/k-NN/pull/3033))
+* Add IT and bwc test with indices containing both vector and non-vector docs ([#]())
```
kotwanikunal: Please add in the issue link here.

qa/restart-upgrade/src/test/java/org/opensearch/knn/bwc/IndexingIT.java
```diff
@@ -731,4 +731,129 @@ public void testDiskBasedMergeBWC() throws Exception {
...
         }
     }
+
+    public void testMixedFieldsWithFaissRestartUpgrade() throws Exception {
+        waitForClusterHealthGreen(NODES_BWC_CLUSTER);
+        if (isRunningAgainstOldCluster()) {
+            createKnnIndex(testIndex, getKNNDefaultIndexSettings(), createKnnIndexMapping(TEST_FIELD, DIMENSIONS, METHOD_HNSW, FAISS_NAME));
+            addKNNDocs(testIndex, TEST_FIELD, DIMENSIONS, DOC_ID, NUM_DOCS);
+            addNonKNNDoc(testIndex, String.valueOf(NUM_DOCS + 1), "description", "Test document");
+            deleteKnnDoc(testIndex, "0");
```
kotwanikunal, line 747: Can we ensure the doc count post delete returns no result?

anntians: Are you referring to the doc count reflects the deleted doc? Meaning we add 10 vector docs, delete 1 and check that doc count is 9? Or are you referring to the deleted doc doesn't show up in the search results?

kotwanikunal: I was pointing to the latter - the goal though is to ensure the doc is deleted.

qa/restart-upgrade/src/test/java/org/opensearch/knn/bwc/IndexingIT.java
```diff
@@ -731,4 +731,292 @@ public void testDiskBasedMergeBWC() throws Exception {
...
+     * Validates k-NN search functionality works without errors after upgrade with Faiss engine.
+     */
+    public void testMixedFieldsWithFaissRestartUpgrade() throws Exception {
+        waitForClusterHealthGreen(NODES_BWC_CLUSTER);
+        if (isRunningAgainstOldCluster()) {
+            createKnnIndex(testIndex, getKNNDefaultIndexSettings(), createKnnIndexMapping(TEST_FIELD, DIMENSIONS, METHOD_HNSW, FAISS_NAME));
+            addKNNDocs(testIndex, TEST_FIELD, DIMENSIONS, DOC_ID, NUM_DOCS);
+            addNonKNNDoc(testIndex, String.valueOf(NUM_DOCS + 1), "description", "Test document");
+            assertEquals(NUM_DOCS + 1, getDocCount(testIndex));
+            deleteKnnDoc(testIndex, "0");
```
naveentatikonda, line 747: Should we validate the segment count before deletion of doc and after upgrade?

qa/rolling-upgrade/src/test/java/org/opensearch/knn/bwc/IndexingIT.java
```diff
@@ -252,4 +253,350 @@ public void testBlockNMSLIBIndexCreationPost3_0_0_RollingUpgrade() throws Except
...
+        waitForClusterHealthGreen(NODES_BWC_CLUSTER);
+        switch (getClusterType()) {
+            case OLD:
+                createKnnIndex(testIndex, getKNNDefaultIndexSettings(), createKnnIndexMapping(TEST_FIELD, DIMENSIONS, ALGO, FAISS_NAME));
+                addKNNDocs(testIndex, TEST_FIELD, DIMENSIONS, 0, NUM_DOCS);
+                addNonKNNDoc(testIndex, String.valueOf(NUM_DOCS + 1), "description", "Test document");
+                assertEquals(NUM_DOCS + 1, getDocCount(testIndex));
+                deleteKnnDoc(testIndex, "0");
+                assertEquals(NUM_DOCS, getDocCount(testIndex));
+                break;
```
naveentatikonda, line 272: Don't we need to do flush here like in restart upgrade tests ?

qa/rolling-upgrade/src/test/java/org/opensearch/knn/bwc/IndexingIT.java
```diff
@@ -252,4 +253,350 @@ public void testBlockNMSLIBIndexCreationPost3_0_0_RollingUpgrade() throws Except
...
+            case MIXED:
+                validateKNNSearch(testIndex, TEST_FIELD, DIMENSIONS, NUM_DOCS - 1, K);
+                break;
+            case UPGRADED:
+                validateKNNSearch(testIndex, TEST_FIELD, DIMENSIONS, NUM_DOCS - 1, K);
+                deleteKNNIndex(testIndex);
+                break;
```
naveentatikonda: nit: We don't need break in the UPGRADED case

src/test/java/org/opensearch/knn/integ/ModeAndCompressionIT.java
```diff
@@ -513,6 +513,120 @@ public void testCompressionIndexWithNonVectorFieldsSegment_whenValid_ThenSucceed
...
     }
 
+    /**
+     * Test segment with knn_vector field mapping but no docs containing the vector field.
+     * Creates separate segments: one with vector docs, one with only non-vector doc.
+     * Validates k-NN search functionality works without errors for ON_DISK mode with compression.
+     */
+    @SneakyThrows
+    public void testMixedSegmentsWithNonVectorDoc_whenValid_ThenSucceed() {
+        CompressionLevel compressionLevel = CompressionLevel.x32;
```
naveentatikonda: Can we loop through all the compression levels, to validate this use case with all of them

qa/restart-upgrade/src/test/java/org/opensearch/knn/bwc/IndexingIT.java
```diff
@@ -731,4 +731,292 @@ public void testDiskBasedMergeBWC() throws Exception {
...
+        }
+    }
+
+    /**
+     * Test segment with knn_vector field mapping but no docs containing the vector field.
+     * Deletes a vector doc, creating a new segment with deleted docs but no docs present.
+     * Validates k-NN search functionality works without errors after upgrade with ON_DISK mode and compression.
+     */
+    public void testMixedFieldsWithCompressionRestartUpgrade() throws Exception {
+        waitForClusterHealthGreen(NODES_BWC_CLUSTER);
```
naveentatikonda, line 787: @anntians wondering how the BWC CI is succeeding on older versions <2.17 as you didn't exclude this test on those versions because ON_DISK isn't supported on older versions ?

Ideally, you should have excluded it here https://github.com/opensearch-project/k-NN/blob/main/qa/restart-upgrade/build.gradle#L109-L118

Am I missing something ?

anntians: I think the BWC tests are being skipped in the CI right now. In the output of the BWC github action, I see 
```
> Task :qa:restart-upgrade:testAgainstOldCluster NO-SOURCE
> Task :qa:restart-upgrade:testRestartUpgrade NO-SOURCE
```
at https://github.com/opensearch-project/k-NN/actions/runs/21084193124/job/60644236182?pr=3064. This means that the task is skipped because no source files to process. I do not see `NO-SOURCE` for other github actions such as for `Remote-Index-Build-IT` where the output shows 
```
> Task :integTestRemoteIndexBuild
```

I also see this `NO-SOURCE` output for the BWC action of other recent PRs. This might need some additional deep dive, but it may be related to some stale build artifacts or old action dependency.

naveentatikonda: Yes, looks like the BWC CI is also succeeding in appx. 4 min which also justifies that it is not running the tests but just building JNI libs in those 4 min. We need to prioritize and fix this asap to catch BWC issues.
cc: @navneet1v @shatejas 

 @anntians meanwhile can you validate your changes on dev desktop or any ec2 ?

anntians: I created an issue to track BWC bug https://github.com/opensearch-project/k-NN/issues/3073

qa/restart-upgrade/src/test/java/org/opensearch/knn/bwc/IndexingIT.java
```diff
@@ -731,4 +731,313 @@ public void testDiskBasedMergeBWC() throws Exception {
...
+            createKnnIndex(testIndex, getKNNDefaultIndexSettings(), createKnnIndexMapping(TEST_FIELD, DIMENSIONS, METHOD_HNSW, FAISS_NAME));
+            addKNNDocs(testIndex, TEST_FIELD, DIMENSIONS, DOC_ID, NUM_DOCS);
+            flush(testIndex, true);
+            addNonKNNDoc(testIndex, String.valueOf(NUM_DOCS + 1), "description", "Test document");
+            flush(testIndex, true);
+            int segmentCount = getTotalSegmentCount(testIndex);
+            assertTrue(segmentCount >= 2);
+        } else {
+            int segmentCountAfterUpgrade = getTotalSegmentCount(testIndex);
+            assertTrue(segmentCountAfterUpgrade > 0);
```
naveentatikonda, line 834: Why not `assertTrue(segmentCountAfterUpgrade >= 2);` ?

anntians: There is flakiness in whether there are segment merges during upgrade. I see sometimes after upgrade there is `>= 2` segments but other times there isn't. So keeping condition to `> 0` to reduce flakiness, validate index is functional, and ensures search functionality after upgrade.

PR #3067, opened 2026-01-14
Changed warmup seek to use long instead of int to avoid overflow

### LINKED ISSUE #3066: [BUG] Integer Overflow in Memory-Optimized Search Warmup for Large Index Files
**What is the bug?**
The `MemoryOptimizedSearchWarmup.warmUpField()` method uses an `int` loop variable when iterating through index files during the warmup process. This causes an integer overflow when warming up index files larger than ~2GB (2,147,483,647 bytes), preventing proper warmup of large vector indices.

1. Create a k-NN index with a large number of high-dimensional vectors
2. Ensure the resulting index file exceeds 2GB in size
3. Trigger a warmup operation on the index
4. Observe that the warmup process fails or doesn't complete properly

The warmup process should successfully iterate through the entire index file regardless of size, reading bytes at 4KB intervals to load the file into memory.

MrFlap:
### Description
This PR fixes an integer overflow bug in the memory-optimized search warmup functionality that prevented proper warmup of large index files (>2GB).

### Related Issues
Resolves #3066

### Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 * Add regex support to derived source transformer include / exclude check [#3031](https://github.com/opensearch-project/k-NN/pull/3031)
 * Correct ef_search parameter for Lucene engine and override mergeLeafResults to return top K results [#3037](https://github.com/opensearch-project/k-NN/pull/3037)
 * Fix efficient filtering in nested k-NN queries [#2990](https://github.com/opensearch-project/k-NN/pull/2990)
+* Changed warmup seek to use long instead of int to avoid overflow [#3067](https://github.com/opensearch-project/k-NN/pull/3067)
 
 ### Refactoring
 * Change ordering of build task and added tests to catch uninitialized libraries [#3024](https://github.com/opensearch-project/k-NN/pull/3024)
--- a/src/main/java/org/opensearch/knn/index/warmup/MemoryOptimizedSearchWarmup.java
+++ b/src/main/java/org/opensearch/knn/index/warmup/MemoryOptimizedSearchWarmup.java
@@ -86,7 +86,7 @@ public class MemoryOptimizedSearchWarmup {
 
         try (IndexInput input = directory.openInput(indexPath.toString(), IOContext.READONCE)) {
             if (input.length() != 0) {
-                for (int i = 0; i < input.length(); i += 4096) {
+                for (long i = 0; i < input.length(); i += 4096) {
                     input.seek(i);
                     input.readByte();
                 }


CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 * Add regex support to derived source transformer include / exclude check [#3031](https://github.com/opensearch-project/k-NN/pull/3031)
 * Correct ef_search parameter for Lucene engine and override mergeLeafResults to return top K results [#3037](https://github.com/opensearch-project/k-NN/pull/3037)
 * Fix efficient filtering in nested k-NN queries [#2990](https://github.com/opensearch-project/k-NN/pull/2990)
+* Changed warmup seek to use long instead of int to avoid overflow
```
shatejas: This needs a PR link

src/main/java/org/opensearch/knn/index/warmup/MemoryOptimizedSearchWarmup.java
```diff
@@ -86,7 +86,7 @@ private boolean warmUpField(FieldInfo field, SegmentReader segmentReader, Direct
 
         try (IndexInput input = directory.openInput(indexPath.toString(), IOContext.READONCE)) {
             if (input.length() != 0) {
-                for (int i = 0; i < input.length(); i += 4096) {
+                for (long i = 0; i < input.length(); i += 4096) {
```
navneet1v, line 89: I would be interested if we can add the test similar to how we are able to catch this issue. That will be awesome

MrFlap: @navneet1v Yes... the issue is that it only pops up for 2 GB indices. I've been waiting on the test index to force merge for the past two hours to properly reproduce the issue. That would not be a good procedure to add to integ tests.

I was thinking about how to catch issues like these, @finnroblin mentioned that we plan to do scheduled benchmarking by pulling from main branch. That would be nice. What progress do we have on that?

finnroblin: @MrFlap I don't know if we've planned it out/scoped the work. The pt I'll make here is that this probably could've been caught by a performance benchmark if we had one with memory optimized search enabled, since we would call the warmup API before searching like we do with the faiss workloads to skip cold start . Having these tests run pre-code merge would give us confidence not only into performance but also into E2E correctness with some representative customer workloads. Basically get some E2E tests for free, since we have to run with big dataset to get sensible signals into performance.

Vikasht34 (approved):

src/main/java/org/opensearch/knn/index/warmup/MemoryOptimizedSearchWarmup.java
navneet1v: @MrFlap I think you could have catched this easily using a Unit test rather than an IT.

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 * Add regex support to derived source transformer include / exclude check [#3031](https://github.com/opensearch-project/k-NN/pull/3031)
 * Correct ef_search parameter for Lucene engine and override mergeLeafResults to return top K results [#3037](https://github.com/opensearch-project/k-NN/pull/3037)
 * Fix efficient filtering in nested k-NN queries [#2990](https://github.com/opensearch-project/k-NN/pull/2990)
+* Changed warmup seek to use long instead of int to avoid overflow [#3067](https://github.com/opensearch-project/k-NN/pull/3067)
```
navneet1v, line 21: this should have come under bug.

Issue #3069, opened 2026-01-16
[BUG] Memory optimized search reentrant search bug in Byte index.


0ctopus13prime:
**What is the bug?**
A clear and concise description of the bug.

The `.vec` file stores the raw input vectors.
For FP32, FP16, and quantized indices, it stores the float vectors exactly as provided.
For byte indices, it stores byte vectors.

Optimistic search is a two-phase search process:
An approximate search is executed first, using a search budget proportional to the target segment size.
A reentrant search may then be triggered, using the document IDs collected in the first phase as seed documents to discover additional relevant results.
The second phase only runs when there are at least two segments and the search logic determines that additional relevant documents are likely to be found. [[Code](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java#L287)]

During the second phase, a [ReentrantKnnCollectorManager](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/lucene/ReentrantKnnCollectorManager.java) is created. It mixes:
DocIndexIterator instances backed by vectors read from the .vec file, and the document IDs collected during the first phase.

However, the current implementation assumes that the `.vec` file always contains float vectors. As a result, it always attempts to retrieve FloatVectorValues, even when the underlying index is a byte index [[Code](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/lucene/ReentrantKnnCollectorManager.java#L86)]

```
// Obtain the per-segment vector values
final LeafReader reader = ctx.reader();

// Always tries to get FloatVectorValues even when the `.vec` file contains byte vectors
// vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv
final FloatVectorValues vectorValues = reader.getFloatVectorValues(field);

if (vectorValues == null) {
    log.error("Acquired null {} for field [{}]", FloatVectorValues.class.getSimpleName(), field);
    // Validates the field exists, otherwise throws an informative exception
    FloatVectorValues.checkField(reader, field);
    return null;
}
```

For byte indices, this results in an exception because FloatVectorValues are not supported for byte-backed `.vec` files.

To fix this, the logic should branch based on the underlying vector data type and call `reader.getByteVectorValues(field)` when operating on a byte index.

**Why it's not caught in IT**
[...]

PR #3071, opened 2026-01-17
Fix MOS reentrant search bug in byte index.

0ctopus13prime:
### Description
Fix reentrant search when memory optimized search is enabled for byte index.
For byte index, now it passes byte[] query, and create ByteVectorValues.

### Related Issues
https://github.com/opensearch-project/k-NN/issues/3069

### Check List

### DIFF
--- a/src/main/java/org/opensearch/lucene/ReentrantKnnCollectorManager.java
+++ b/src/main/java/org/opensearch/lucene/ReentrantKnnCollectorManager.java
@@ -5,9 +5,11 @@
 
 package org.opensearch.lucene;
 
+import lombok.NonNull;
 import lombok.RequiredArgsConstructor;
 import lombok.extern.log4j.Log4j2;
 import org.apache.lucene.codecs.lucene90.IndexedDISI;
+import org.apache.lucene.index.ByteVectorValues;
 import org.apache.lucene.index.FloatVectorValues;
 import org.apache.lucene.index.KnnVectorValues;
 import org.apache.lucene.index.LeafReader;
@@ -18,6 +20,7 @@ import org.apache.lucene.search.TopDocs;
 import org.apache.lucene.search.VectorScorer;
 import org.apache.lucene.search.knn.KnnCollectorManager;
 import org.apache.lucene.search.knn.KnnSearchStrategy;
+import org.apache.lucene.util.IOSupplier;
 
 import java.io.IOException;
 import java.util.Map;
@@ -32,22 +35,28 @@ import java.util.Map;
  * {@link SeededTopDocsDISI} and {@link SeededMappedDISI}, enabling the internal searcher to start from these known points
  * instead of beginning from random or default graph entry nodes.
  * <p>
- * See <a href="https://github.com/apache/lucene/blob/71e822e6240878018a6ff3c28381a0d88bebdc72/lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java#L368">...</a>
+ * See
+ * <a href="https://github.com/apache/lucene/blob/71e822e6240878018a6ff3c28381a0d88bebdc72/lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java#L368">...</a>
  */
 @Log4j2
 @RequiredArgsConstructor
 public class ReentrantKnnCollectorManager implements KnnCollectorManager {
 
     // The underlying (delegate) KNN collector manager used to create collectors.
+    @NonNull
     private final KnnCollectorManager knnCollectorManager;
 
     // Mapping from segment ordinal to previously collected {@link TopDocs}.
+    @NonNull
     private final Map<Integer, TopDocs> segmentOrdToResults;
 
     // Query vector used for scoring during vector similarity search.
-    private final float[] query;
+    // It must be float[] or byte[]
+    @NonNull
+    private final Object query;
 
     // Name of the vector field being searched.
+    @NonNull
     private final String field;
 
     /**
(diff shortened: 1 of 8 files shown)


src/main/java/org/opensearch/lucene/ReentrantKnnCollectorManager.java
```diff
@@ -83,16 +92,32 @@ public KnnCollector newCollector(int visitLimit, KnnSearchStrategy searchStrateg
...
-        final FloatVectorValues vectorValues = reader.getFloatVectorValues(field);
-        if (vectorValues == null) {
-            log.error("Acquired null {} for field [{}]", FloatVectorValues.class.getSimpleName(), field);
-            // Validates the field exists, otherwise throws informative exception
-            FloatVectorValues.checkField(reader, field);
-            return null;
+
+        final IOSupplier<VectorScorer> vectorScorerSupplier;
+        if (query instanceof float[] floatQuery) {
+            final FloatVectorValues vectorValues = reader.getFloatVectorValues(field);
```
navneet1v, line 98: Use KNNVectorValuesFactory class to get the VectorValuesIterator rather than directly calling `reader.getFloatVectorValues(field)`

0ctopus13prime: We can't use KNNVectorValuesFactory in here. 
The goal of this class is to get [KnnVectorValues.DocIndexIterator](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/index/KnnVectorValues.java#L95) and pass vector ids as seed down to its collector so that it can start graph searching at the bottom layer.

navneet1v: I see. Looks like the whole file is inspired from Lucene.

0ctopus13prime: Yes, I blindly copied it from Lucene and missed for byte index type checking.

navneet1v: If you are just looking for DISI and in any order, KNNVectorValues of k-NN plugin provides the DISI. I can see in the code we are creating a scorer and then getting DISI out of it. We can avoid that since we already have DISI which is same as what is present here KnnVectorValues.DocIndexIterator

0ctopus13prime: We still cannot use DISI because the logic is not simply looking for a DISI. It specifically requires a DocIndexIterator, which returns internal vector IDs, whereas DISI returns document IDs. For example, if 10 documents are indexed but only 4 of them contain kNN vector fields, using document IDs for reentrant search will not work. In this case, the search logic expects internal vector IDs in the range [0, 3], not document IDs.

0ctopus13prime:
Not sure why ITs are failing, both Mac + Linux in my local development environment, all are passing.
Will do the double check.

src/main/java/org/opensearch/lucene/ReentrantKnnCollectorManager.java
Vikasht34: After taking look on ReentrantKnnCollectorManager , SeededMappedDISI , SeededTopDocsDISI . Here is my understanding .

```
Phase 1 Results (TopDocs)
    doc IDs: [5, 23, 42]
SeededTopDocsDISI
    iterates: 5 → 23 → 42
SeededMappedDISI
    Doc ID 5  → advance indexedDISI to 5  → get vector index 2
    Doc ID 23 → advance indexedDISI to 23 → get vector index 8
    Doc ID 42 → advance indexedDISI to 42 → get vector index 15
    Returns: [2, 8, 15]  ← Vector indices!
KnnSearchStrategy.Seeded
    Uses vector indices [2, 8, 15] as HNSW entry points
```

Now the reason we can't use VectorValues Factory becuase , it ultimatey gives DOCIDsetIerator which we already have from SeededTopDocsDISI and now we need DOCID - VectorID Mapping which is DocIndexIterator that we can have intersection on with TopDocs. If this Understanding is correct then yes we can't use KNNVectorValuesFactory class.

src/main/java/org/opensearch/lucene/ReentrantKnnCollectorManager.java
```diff
@@ -83,16 +92,32 @@ public KnnCollector newCollector(int visitLimit, KnnSearchStrategy searchStrateg
...
+                return null;
+            }
+            vectorScorerSupplier = () -> vectorValues.scorer(byteQuery);
+        } else {
+            throw new IllegalStateException("Unknown query type: " + query.getClass());
         }
 
         // Create a vector scorer for the query vector
-        final VectorScorer scorer = vectorValues.scorer(query);
+        final VectorScorer scorer = vectorScorerSupplier.get();
```
Vikasht34, line 120: I think we should keep the same behaviour from above , if vectorValues is null we are returning null not delegateCollector but if scorer is null we are returning delegateCollector, Any reason for this?

src/main/java/org/opensearch/lucene/ReentrantKnnCollectorManager.java
Vikasht34: @0ctopus13prime  can we add one integ test around this "For example, if 10 documents are indexed but only 4 of them contain kNN vector fields, using document IDs for reentrant search will not work. In this case, the search logic expects internal vector IDs in the range [0, 3], not document IDs." 

for both Postive and negative cases. I did not find it integs for this. Let me know if we already have.

0ctopus13prime: Hi Vikash, yes, thank you for the analysis. What we need is a list of internal vector ids so we need to convert doc id to internal vector ids. Therefore, I don't think we can use the factory.

src/main/java/org/opensearch/lucene/ReentrantKnnCollectorManager.java
0ctopus13prime: It's the same behavior, I didn't change anything from behavioral perspective, what did you see?

src/main/java/org/opensearch/lucene/ReentrantKnnCollectorManager.java
0ctopus13prime: Hi @Vikasht34 
For the memory optimized search, it's running all possible cases exhaustively. Below links are covering sparse vector fields case in your analysis.

For the non-nested case - https://github.com/opensearch-project/k-NN/blob/5602c8f3a0921cc7ad310aaeb2f4b2245a143f9c/src/test/java/org/opensearch/knn/memoryoptsearch/AbstractMemoryOptimizedKnnSearchIT.java#L111

For the nested case - https://github.com/opensearch-project/k-NN/blob/5602c8f3a0921cc7ad310aaeb2f4b2245a143f9c/src/test/java/org/opensearch/knn/memoryoptsearch/AbstractMemoryOptimizedKnnSearchIT.java#L149

Vikasht34: Thanks ..

src/main/java/org/opensearch/lucene/ReentrantKnnCollectorManager.java
Vikasht34: I was thinking if VectorValue is null , we are returning null but if scorer is null then defaultCollector, should we also return defaultCollector if VectorValues is null?

0ctopus13prime: Ah, got it.
We're not returning null actually, as `XXXVectorValues.checkField(reader, field)` will throw an error.

Vikasht34: so like 

``FloatVectorValues.checkField(reader, field);
                return null;

so return null; will neveer hit

0ctopus13prime: Yes! that's true.

Issue #3092, opened 2026-02-03
[BUG] The KNN1030Codec does not properly support delegation for non-default codec(s)


reta:
**What is the bug?**
In the latest upcoming 3.5 release of the OpenSearch core we introduced an alternative way to register additional codecs without the need to provide custom `CodecService` instance (see please [1] for details). It opens up the possibility to use `k-nn` with custom codecs (like `zstd` from https://github.com/opensearch-project/custom-codecs).

```
"settings": {
    "index": {
      "codec": "zstd",
      "knn": true
    }
 }
```

It works on the write path but fails on the read path with:

```
    java.lang.IllegalStateException: missing value for Lucene90StoredFieldsFormat.mode for segment: _0
        at __randomizedtesting.SeedInfo.seed([58DBE7C5F887978:96A46AB764488226]:0)
        at org.apache.lucene.codecs.lucene90.Lucene90StoredFieldsFormat.fieldsReader(Lucene90StoredFieldsFormat.java:133)
        at org.opensearch.knn.index.codec.KNN10010Codec.KNN10010DerivedSourceStoredFieldsFormat.fieldsReader(KNN10010DerivedSourceStoredFieldsFormat.java:64)
        at org.apache.lucene.index.SegmentCoreReaders.<init>(SegmentCoreReaders.java:118)
        at org.apache.lucene.index.SegmentReader.<init>(SegmentReader.java:95)
        at org.apache.lucene.index.CheckIndex.testSegment(CheckIndex.java:1003)
        at org.apache.lucene.index.CheckIndex.checkIndex(CheckIndex.java:788)
        at org.apache.lucene.index.CheckIndex.checkIndex(CheckIndex.java:558)
        at org.apache.lucene.tests.util.TestUtil.checkIndex(TestUtil.java:345)
        ...
```

The issue here is that when the KNN1030Codec is returned by the `CodecService`, it wraps (and delegates) whatever the default codec for the index is (see please the snippet below):

```java
    @Override
    public Codec codec(String name) {
        return KNN1030Codec.builder()
            .delegate(super.codec(name))
            .mapperService(mapperService)
            .knnVectorsFormat(new KNN9120PerFieldKnnVectorsFormat(Optional.ofNullable(mapperService), nativeIndexBuildStrategyFactory))
            .build();
    }
```

Which for `zstd` would translate roughy to:

```java
    @Override
    public Codec codec(String name) {
        return KNN1030Codec.builder()
[...]

PR #3093, opened 2026-02-03
The KNN1030Codec does not properly support delegation for non-default codec(s)

reta:
### Description
In the latest upcoming 3.5 release of the OpenSearch core we introduced an alternative way to register additional codecs without the need to provide custom `CodecService` instance (see please [1] for details). It opens up the possibility to use `k-nn` with custom codecs (like `zstd` from https://github.com/opensearch-project/custom-codecs).

```
"settings": {
    "index": {
      "codec": "zstd",
      "knn": true
    }
 }
```

It works on the write path but fails on the read path with:

```
    java.lang.IllegalStateException: missing value for Lucene90StoredFieldsFormat.mode for segment: _0
        at __randomizedtesting.SeedInfo.seed([58DBE7C5F887978:96A46AB764488226]:0)
        at org.apache.lucene.codecs.lucene90.Lucene90StoredFieldsFormat.fieldsReader(Lucene90StoredFieldsFormat.java:133)
        at org.opensearch.knn.index.codec.KNN10010Codec.KNN10010DerivedSourceStoredFieldsFormat.fieldsReader(KNN10010DerivedSourceStoredFieldsFormat.java:64)
        at org.apache.lucene.index.SegmentCoreReaders.<init>(SegmentCoreReaders.java:118)
        at org.apache.lucene.index.SegmentReader.<init>(SegmentReader.java:95)
        at org.apache.lucene.index.CheckIndex.testSegment(CheckIndex.java:1003)
        at org.apache.lucene.index.CheckIndex.checkIndex(CheckIndex.java:788)
        at org.apache.lucene.index.CheckIndex.checkIndex(CheckIndex.java:558)
        at org.apache.lucene.tests.util.TestUtil.checkIndex(TestUtil.java:345)
        ...
```

The suggested fix is to store the delegating codec name in the Apache Lucene segment attributes so it could be properly restored on read, the scope is limited to `storedFieldsFormat()` only since it does not have per-field equivalents.

### Related Issues
Resolves https://github.com/opensearch-project/k-NN/issues/3092

### Check List

### DIFF
--- a/src/test/java/org/opensearch/knn/integ/codecs/CustomCodecsIT.java
+++ b/src/test/java/org/opensearch/knn/integ/codecs/CustomCodecsIT.java
@@ -9,6 +9,7 @@ import java.net.URL;
 import java.util.List;
 
 import org.apache.hc.core5.http.io.entity.EntityUtils;
+import org.junit.After;
 import org.junit.Before;
 import org.opensearch.client.Response;
 import org.opensearch.common.settings.Settings;
@@ -41,6 +42,12 @@ public class CustomCodecsIT extends KNNRestTestCase {
         assert testIndexVectors != null;
         assert testQueries != null;
         testData = new TestUtils.TestData(testIndexVectors.getPath(), testQueries.getPath());
+    }
+
+    @After
+    @SneakyThrows
+    public final void cleanUp() {
+        deleteKNNIndex(INDEX_NAME);
     }
 
     // Test KNN index with ZSTD codec enabled


src/test/java/org/opensearch/knn/index/codec/KNNCodecTestCase.java
```diff
@@ -155,7 +157,8 @@ protected ResourceWatcherService createDisabledResourceWatcherService() {
 
     public void testMultiFieldsKnnIndex(Codec codec) throws Exception {
         setUpMockClusterService();
-        try (Directory dir = newFSDirectory(createTempDir())) {
+        final Path path = createTempDir();
```
reta, line 160: Slightly modifying test case to have writer and reader separated, functionally-wise no changes but reproduces the issue right away

src/main/java/org/opensearch/knn/index/codec/KNN10010Codec/KNN10010DerivedSourceStoredFieldsFormat.java
```diff
@@ -54,20 +61,46 @@ public StoredFieldsReader fieldsReader(Directory directory, SegmentInfo segmentI
...
+            return delegate;
+        }
+    }
+
     @Override
     public StoredFieldsWriter fieldsWriter(Directory directory, SegmentInfo segmentInfo, IOContext ioContext) throws IOException {
+        // Store delegate codec name to be used by reader side
+        String previous = segmentInfo.putAttribute(KNN_DELEGATE_CODEC_NAME, name);
+        if (previous != null && previous.equals(name) == false) {
+            throw new IllegalStateException(
```
Vikasht34, line 92: Do we really need to throw this exception, Just wanted to understand if we already had previous match , Should we even put?

reta: This is more like a precaution to be fair, if we for some reasons the attribute have different value(s), the segments really messed up, should never happen.

navneet1v:
@reta since we are already building a composable codec in OpenSearch core do you really think we need this change? I understand why this change is needed but I am little bit skeptical of storing the delegating codec in segmentInfo file since it can become a one way door.

src/main/java/org/opensearch/knn/index/codec/KNN10010Codec/KNN10010DerivedSourceStoredFieldsFormat.java
```diff
@@ -31,7 +33,10 @@
 
 @AllArgsConstructor
 public class KNN10010DerivedSourceStoredFieldsFormat extends StoredFieldsFormat {
+    // Stores the delegate codec name (in case it is different from the default one)
+    static final String KNN_DELEGATE_CODEC_NAME = "knn_delegate_codec_key";
```
shatejas: I understand this is not public but would want to be specific just in case we need it for compound or anything else
```suggestion
    // Stores the delegate codec name (in case it is different from the default one)
    static final String KNN_DELEGATE_CODEC_NAME = "knn_delegate_stored_fields_codec_key";
```

shatejas (commented):
@reta Do you think we should make it future proof while doing the same for compound format

@navneet1v thoughts?

navneet1v:
@shatejas I keep on forgetting that we have overridden the compound format thanks for bringing this up. I think I am good right now in terms of this PR. We might want to take the compound format codec as a separate item and we can take a similar approach what we are doing for stored fieldFormat here.

navneet1v:
@reta can we add an IT with and without derived source enabled with this index setting

```
"settings": {
    "index": {
      "codec": "zstd",
      "knn": true
    }
 }
```

This will help ensure that if anything gets broken in future we can catch it. I know you added a test but that is at the bottom layer. We should have a test from OpenSearch side too.

Rest code looks good to me. Once we have the IT I can approve the code.

navneet1v:
> > @reta can we add an IT with and without derived source enabled with this index setting
> 
> Thanks @navneet1v, I could add test case without derived source enabled, sure.
> 
> Regarding the index settings, it is good to understand that it means bringing test cluster with the custom-codec plugin installed (it is not installed / bundled by default). Technically, I could do that but I really think this is not a good idea since most of the maintenance of build scripts is on your shoulders. However, we do have a place where we exercise a full distribution through the series of smoke tests (see please [opensearch-project/opensearch-build#5223](https://github.com/opensearch-project/opensearch-build/issues/5223)), I think this would be a good tradeoff, we have all plugins installed already and would add one test case for knn + zstd, sounds like a plan?
> 
> **Update**: [opensearch-project/opensearch-build#5999](https://github.com/opensearch-project/opensearch-build/pull/5999)

I am little hesitant on not having IT for 
```
"settings": {
    "index": {
      "codec": "zstd",
      "knn": true
    }
 }
```
can somehow create a test setup with custom plugin? would that be easier? I think we do something similar for neural where we install KNN and ML commons plugin just for tests.

shatejas (approved):
Thanks for the integ test, LGTM!

src/test/java/org/opensearch/knn/integ/codecs/CustomCodecsIT.java
```diff
@@ -0,0 +1,121 @@
...
+        URL testIndexVectors = CustomCodecsIT.class.getClassLoader().getResource("data/test_vectors_1000x128.json");
+        URL testQueries = CustomCodecsIT.class.getClassLoader().getResource("data/test_queries_100x128.csv");
+
+        assert testIndexVectors != null;
+        assert testQueries != null;
+        testData = new TestUtils.TestData(testIndexVectors.getPath(), testQueries.getPath());
+    }
+
+    // Test KNN index with ZSTD codec enabled
+    public void testCreateIndexWithZstdCodec() throws Exception {
```
shatejas, line 54: nit: delete the index after test run

reta: Thanks, added cleanUp()

build.gradle
```diff
@@ -687,6 +688,21 @@ def commonIntegTestClusters(OpenSearchCluster cluster, _numNodes){
     // Optionally install security
     if (System.getProperty("security.enabled") != null) {
         configureSecurityPlugin(cluster)
+    } else {
```
navneet1v, line 691: any reason for putting this in else? Shouldn't we be installing custom-codec plugin for both security and non security cases?

navneet1v: @reta am I missing something here?

reta: > @reta am I missing something here?

@navneet1v Yes, the `configureSecurityPlugin()` installs all plugins with `zipArchive` scope, including custom-codecs

navneet1v: Make sense. Good to go from my side

PR #3095, opened 2026-02-03
Add support for pre-quantized vector exact search

### LINKED ISSUE #2215: [Enhancement] Reducing double vector reading and distance calculation for Disk based Filtered Vector Search

VijayanB:
### Description
Implement exact KNN search using pre-quantized vectors from native engines to avoid redundant quantization during query execution. This optimization retrieves quantized vectors directly from FAISS memory-optimized indexes and uses them for Hamming distance calculations.

Performance metrics for 10M https://github.com/opensearch-project/k-NN/issues/2215#issuecomment-3882739298

### Related Issues
Resolves #2215 

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
+++ b/src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
@@ -125,6 +125,7 @@ public class NativeEngines990KnnVectorsReader extends KnnVectorsReader {
             if (quantizedVectorValues != null) {
                 return quantizedVectorValues;
             }
+            log.warn("No quantized vectors found for field [{}]", field);
         }
         return flatVectorsReader.getByteVectorValues(field);
     }
--- a/src/test/java/org/opensearch/knn/index/query/exactsearch/ExactSearcherTests.java
+++ b/src/test/java/org/opensearch/knn/index/query/exactsearch/ExactSearcherTests.java
@@ -189,6 +189,86 @@ public class ExactSearcherTests extends KNNTestCase {
     }
 
     @SneakyThrows
+    public void testExactSearch_whenQuantizationWithoutADC_thenSuccess() {
+        try (
+            MockedStatic<KNNVectorValuesFactory> valuesFactoryMockedStatic = Mockito.mockStatic(KNNVectorValuesFactory.class);
+            MockedStatic<org.opensearch.knn.index.query.SegmentLevelQuantizationInfo> quantizationInfoMockedStatic = Mockito.mockStatic(
+                org.opensearch.knn.index.query.SegmentLevelQuantizationInfo.class
+            );
+            MockedStatic<org.opensearch.knn.index.query.SegmentLevelQuantizationUtil> quantizationUtilMockedStatic = Mockito.mockStatic(
+                org.opensearch.knn.index.query.SegmentLevelQuantizationUtil.class
+            )
+        ) {
+            final float[] queryVector = new float[] { 0.1f, 2.0f, 3.0f };
+            final byte[] quantizedQueryVector = new byte[] { 1, 2, 3 };
+            final List<float[]> floatVectors = List.of(
+                new float[] { 1, 2, 3 },
+                new float[] { 4, 5, 6 },
+                new float[] { 2, 3, 4 },
+                new float[] { 7, 8, 9 },
+                new float[] { 0, 1, 2 }
+            );
+            final List<byte[]> quantizedDataVectors = List.of(
+                new byte[] { 1, 2, 3 },
+                new byte[] { 4, 5, 6 },
+                new byte[] { 2, 3, 4 },
+                new byte[] { 7, 8, 9 },
+                new byte[] { 0, 1, 2 }
+            );
+            final SpaceType spaceType = SpaceType.L2;
+            final int k = 10;
+
+            final ExactSearcher.ExactSearcherContext exactSearcherContext = ExactSearcher.ExactSearcherContext.builder()
+                .field(FIELD_NAME)
+                .floatQueryVector(queryVector)
+                .useQuantizedVectorsForSearch(true)
+                .k(k)
+                .build();
+
+            ExactSearcher exactSearcher = new ExactSearcher(null);
+            final LeafReaderContext leafReaderContext = mock(LeafReaderContext.class);
+            final SegmentReader reader = mock(SegmentReader.class);
+            when(leafReaderContext.reader()).thenReturn(reader);
+
+            final FieldInfos fieldInfos = mock(FieldInfos.class);
+            final FieldInfo fieldInfo = mock(FieldInfo.class);
+            when(fieldInfo.getAttribute(SPACE_TYPE)).thenReturn(spaceType.getValue());
+            when(reader.getFieldInfos()).thenReturn(fieldInfos);
+            when(fieldInfos.fieldInfo(FIELD_NAME)).thenReturn(fieldInfo);
+
+            final org.opensearch.knn.index.query.SegmentLevelQuantizationInfo quantizationInfo = mock(
+                org.opensearch.knn.index.query.SegmentLevelQuantizationInfo.class
+            );
+            quantizationInfoMockedStatic.when(
+                () -> org.opensearch.knn.index.query.SegmentLevelQuantizationInfo.build(reader, fieldInfo, FIELD_NAME)
+            ).thenReturn(quantizationInfo);
+            quantizationUtilMockedStatic.when(
+                () -> org.opensearch.knn.index.query.SegmentLevelQuantizationUtil.isAdcEnabled(quantizationInfo)
+            ).thenReturn(false);
+            quantizationUtilMockedStatic.when(
+                () -> org.opensearch.knn.index.query.SegmentLevelQuantizationUtil.quantizeVector(queryVector, quantizationInfo)
+            ).thenReturn(quantizedQueryVector);
+
+            final KNNVectorValues knnFloatVectorValues = TestVectorValues.createKNNFloatVectorValues(floatVectors);
+            final KNNVectorValues knnByteVectorValues = TestVectorValues.createKNNBinaryVectorValues(quantizedDataVectors);
+            valuesFactoryMockedStatic.when(() -> KNNVectorValuesFactory.getVectorValues(fieldInfo, reader))
+                .thenReturn(knnFloatVectorValues);
+            valuesFactoryMockedStatic.when(() -> KNNVectorValuesFactory.getVectorValues(fieldInfo, reader, true))
+                .thenReturn(knnByteVectorValues);
+
+            TopDocs docs = exactSearcher.searchLeaf(leafReaderContext, exactSearcherContext);
+            assertEquals(quantizedDataVectors.size(), docs.scoreDocs.length);
+
+            final List<Float> expectedScores = quantizedDataVectors.stream()
+                .map(quantizedVector -> SpaceType.HAMMING.getKnnVectorSimilarityFunction().compare(quantizedQueryVector, quantizedVector))
+                .sorted(Comparator.reverseOrder())
+                .toList();
+            final List<Float> actualScores = Arrays.stream(docs.scoreDocs).map(scoreDoc -> scoreDoc.score).toList();
+            assertEquals(expectedScores, actualScores);
+        }
+    }
+
+    @SneakyThrows
     private void doTestRadialSearch_whenNoEngineFiles_thenSuccess(final boolean memoryOptimizedSearchEnabled) {
         try (MockedStatic<KNNVectorValuesFactory> valuesFactoryMockedStatic = Mockito.mockStatic(KNNVectorValuesFactory.class)) {
             // Prepare data
--- a/src/test/java/org/opensearch/knn/index/vectorvalues/KNNVectorValuesFactoryTests.java
+++ b/src/test/java/org/opensearch/knn/index/vectorvalues/KNNVectorValuesFactoryTests.java
@@ -20,6 +20,9 @@ import org.opensearch.knn.index.VectorDataType;
 import java.util.List;
 import java.util.Map;
 
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.verify;
+
 public class KNNVectorValuesFactoryTests extends KNNTestCase {
     private static final int COUNT = 10;
     private static final int DIMENSION = 10;
@@ -63,6 +66,52 @@ public class KNNVectorValuesFactoryTests extends KNNTestCase {
             byteVectorMap
         );
         Assert.assertNotNull(binaryVectorValues);
+    }
+
+    @SneakyThrows
+    public void testGetVectorValuesFromFieldInfo_whenQueryVectorQuantized_thenSuccess() {
+        final List<float[]> floatArrayList = List.of(new float[] { 1.3f, 2.2f, 3.2f });
+        final List<byte[]> quantizedByteArrayList = List.of(new byte[] { 1, 2, 3 });
+        final FieldInfo fieldInfo = Mockito.mock(FieldInfo.class);
+        final SegmentReader reader = Mockito.mock(SegmentReader.class);
+        Mockito.when(fieldInfo.hasVectorValues()).thenReturn(true);
+        Mockito.when(fieldInfo.getName()).thenReturn("test_field");
+        Mockito.when(fieldInfo.getVectorEncoding()).thenReturn(VectorEncoding.FLOAT32);
+        Mockito.when(fieldInfo.getAttribute(KNNConstants.VECTOR_DATA_TYPE_FIELD)).thenReturn(VectorDataType.FLOAT.getValue());
+        Mockito.when(reader.getFloatVectorValues("test_field"))
+            .thenReturn(new TestVectorValues.PreDefinedFloatVectorValues(floatArrayList));
+
+        final org.apache.lucene.codecs.KnnVectorsReader vectorsReader = Mockito.mock(org.apache.lucene.codecs.KnnVectorsReader.class);
+        Mockito.when(reader.getVectorReader()).thenReturn(vectorsReader);
+        Mockito.when(vectorsReader.getByteVectorValues("test_field"))
+            .thenReturn(new TestVectorValues.PreDefinedByteVectorValues(quantizedByteArrayList));
+
+        final KNNVectorValues<byte[]> quantizedVectorValues = KNNVectorValuesFactory.getVectorValues(fieldInfo, reader, true);
+        quantizedVectorValues.nextDoc();
+        Assert.assertArrayEquals(quantizedByteArrayList.getFirst(), quantizedVectorValues.getVector());
+        Assert.assertNotNull(quantizedVectorValues);
+        verify(reader).getFloatVectorValues("test_field");
+    }
+
+    @SneakyThrows
+    public void testGetVectorValuesFromFieldInfo_whenQueryVectorNotQuantized_thenSuccess() {
+        final List<float[]> floatArrayList = List.of(new float[] { 1.3f, 2.2f, 3.2f });
+        final FieldInfo fieldInfo = Mockito.mock(FieldInfo.class);
+        final SegmentReader reader = Mockito.mock(SegmentReader.class);
+        Mockito.when(fieldInfo.hasVectorValues()).thenReturn(true);
+        Mockito.when(fieldInfo.getName()).thenReturn("test_field");
+        Mockito.when(fieldInfo.getVectorEncoding()).thenReturn(VectorEncoding.FLOAT32);
+        Mockito.when(fieldInfo.getAttribute(KNNConstants.VECTOR_DATA_TYPE_FIELD)).thenReturn(VectorDataType.FLOAT.getValue());
+        Mockito.when(reader.getFloatVectorValues("test_field"))
+            .thenReturn(new TestVectorValues.PreDefinedFloatVectorValues(floatArrayList));
+
+        final KNNVectorValues<float[]> vectorValues = KNNVectorValuesFactory.getVectorValues(fieldInfo, reader, false);
+        Assert.assertNotNull(vectorValues);
+        vectorValues.nextDoc();
+        Assert.assertNotNull(vectorValues.getVector());
+        verify(reader).getFloatVectorValues("test_field");
+        verify(reader, never()).getVectorReader();
+
     }
 
     @SneakyThrows
@@ -146,7 +195,7 @@ public class KNNVectorValuesFactoryTests extends KNNTestCase {
         Assert.assertArrayEquals(binaryArrayList.get(0), binaryVectorValues.getVector());
         Assert.assertNotNull(binaryVectorValues);
 
-        Mockito.verify(fieldInfo, Mockito.times(0)).getVectorEncoding();
+        verify(fieldInfo, Mockito.times(0)).getVectorEncoding();
     }
 
 }
--- a/src/test/java/org/opensearch/knn/index/vectorvalues/TestVectorValues.java
+++ b/src/test/java/org/opensearch/knn/index/vectorvalues/TestVectorValues.java
@@ -421,6 +421,10 @@ public class TestVectorValues {
         return new KNNBinaryVectorValues(new KNNVectorValuesIterator.DocIdsIteratorValues(binaryDocValues));
     }
 
+    public static KNNVectorValues createKNNBinaryVectorValues(final List<byte[]> vectors) {
+        return new KNNBinaryVectorValues(new KNNVectorValuesIterator.DocIdsIteratorValues(new PreDefinedBinaryVectorValues(vectors)));
+    }
+
     public static KNNVectorValues createKNNFloatVectorValues(final List<float[]> vectors) {
         return new KNNFloatVectorValues(
             new KNNVectorValuesIterator.DocIdsIteratorValues(new PredefinedFloatVectorBinaryDocValues(vectors))


src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
```diff
@@ -349,6 +350,26 @@ private IOSupplier<VectorSearcher> getVectorSearcherSupplier(final FieldInfo fie
...
+    public ByteVectorValues getQuantizedVectorValues(final String fieldName) {
+        try {
+            VectorSearcher vectorSearcher = loadMemoryOptimizedSearcherIfRequired(fieldName);
+            if (vectorSearcher instanceof FaissMemoryOptimizedSearcher searcher) {
+                return searcher.getByteVectorValues();
+            }
+        } catch (Exception e) {
+            log.error("Failed to get quantized vector values for field [{}]", fieldName, e);
+        }
+        return null;
```
0ctopus13prime: I think we should throw an exception as this is not expected.

src/main/java/org/opensearch/knn/index/query/exactsearch/ExactSearcher.java
```diff
@@ -261,6 +280,20 @@ private ExactKNNIterator getKNNIterator(LeafReaderContext leafReaderContext, Exa
...
+    private ByteVectorValues getQuantizedVectorValues(SegmentReader reader, String fieldName) {
+        try {
+            if (reader.getVectorReader() instanceof PerFieldKnnVectorsFormat.FieldsReader fieldsReader) {
+                final KnnVectorsReader fieldReader = fieldsReader.getFieldReader(fieldName);
+                if (fieldReader instanceof QuantizedVectorsReader quantizedReader) {
+                    return quantizedReader.getQuantizedVectorValues(fieldName);
+                }
+            }
+        } catch (Exception e) {
+            log.debug("Failed to retrieve quantized vectors for field {}", fieldName, e);
```
0ctopus13prime: Curious, what is the case will we get this exception?

src/main/java/org/opensearch/knn/index/query/exactsearch/ExactSearcher.java
```diff
@@ -261,6 +280,20 @@ private ExactKNNIterator getKNNIterator(LeafReaderContext leafReaderContext, Exa
         );
     }
 
+    private ByteVectorValues getQuantizedVectorValues(SegmentReader reader, String fieldName) {
```
0ctopus13prime: Also do we have to check whether MOS is enabled in here?
ExactSearcher context should have the info

src/main/java/org/opensearch/knn/index/query/exactsearch/QuantizedVectorIdsExactKNNIterator.java
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    protected int getNextDocId() throws IOException {
+        if (filterIdsIterator == null) {
+            return knnFloatVectorValuesIndexIterator.nextDoc();
+        }
+        int nextDocID = this.filterIdsIterator.nextDoc();
+        // For filter case, advance vector values to corresponding doc id from filter bit set
+        if (nextDocID != DocIdSetIterator.NO_MORE_DOCS) {
+            knnFloatVectorValuesIndexIterator.advance(nextDocID);
```
0ctopus13prime: I think this will not work for sparse case where not all docs having knn field.
for example, imagine a scenario where the match filter gives 5, but doc 5 does not knn vector field.

Could you use 

```
List<DocIdIterator> disiList = Arrays.asList(filterIdsIterator, knnFloatVectorValuesIndexIterator);
this.conjunctionIterator = ConjunctionUtils.intersectIterators(disiList)

then, we don't need this method. we can just call conjunctionIterator.nextDoc().
```

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissMemoryOptimizedSearcher.java
```diff
@@ -143,6 +144,16 @@ public void search(byte[] target, KnnCollector knnCollector, AcceptDocs acceptDo
         );
     }
 
+    /**
+     * Returns the byte vector values from the underlying FAISS index.
+     *
+     * @return {@link ByteVectorValues} containing the quantized vectors
+     * @throws IOException if an I/O error occurs while reading the index
+     */
```
0ctopus13prime, line 152: Nit : Missed `@Override`

src/main/java/org/opensearch/knn/index/query/exactsearch/QuantizedVectorIdsExactKNNIterator.java
```diff
@@ -0,0 +1,104 @@
...
+        final KnnVectorValues.DocIndexIterator knnFloatVectorValuesIndexIterator,
+        final ByteVectorValues quantizedByteVectorValues,
+        final byte[] quantizedQueryVector
+    ) throws IOException {
+        this.filterIdsIterator = filterIdsIterator;
+        this.knnFloatVectorValuesIndexIterator = knnFloatVectorValuesIndexIterator;
+        this.quantizedByteVectorValues = quantizedByteVectorValues;
+        // This cannot be moved inside nextDoc() method since it will break when we have nested field, where
+        // nextDoc should already be referring to next knnVectorValues
+        this.docId = getNextDocId();
```
0ctopus13prime: I think we can make `NestedQuantizedVectorIdsExactKNNIterator` call `getNextDocId()` in its constructor and cut this dependency. 
What do you think?

src/main/java/org/opensearch/knn/index/query/exactsearch/NestedQuantizedVectorIdsExactKNNIterator.java
```diff
@@ -0,0 +1,55 @@
...
+    public int nextDoc() throws IOException {
+        if (docId == DocIdSetIterator.NO_MORE_DOCS) {
+            return DocIdSetIterator.NO_MORE_DOCS;
+        }
+
+        currentScore = Float.NEGATIVE_INFINITY;
+        int currentParent = parentBitSet.nextSetBit(docId);
+        int bestChild = -1;
+
+        while (docId != DocIdSetIterator.NO_MORE_DOCS && docId < currentParent) {
```
0ctopus13prime: NIT : we don't need `docId != NO_MORE_DOCS`. 
NO_MORE_DOCS has Integer.MAX_VALUE.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
VijayanB: Ack

src/main/java/org/opensearch/knn/index/query/exactsearch/ExactSearcher.java
VijayanB: Not necessary. We are not actually doing MOS, we are just using the searcher class only to get access to the FaissIndex. This is path for both MOS enabled and disabled for now

src/main/java/org/opensearch/knn/index/query/exactsearch/QuantizedVectorIdsExactKNNIterator.java
VijayanB: I believe this was already added here https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/query/exactsearch/ExactSearcher.java#L179.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissMemoryOptimizedSearcher.java
VijayanB: I thought we discussed that this will not be added to vectorsearcher interface, will use instance check to get ByteVectorValues

src/main/java/org/opensearch/knn/index/query/exactsearch/ExactSearcher.java
VijayanB: Good catch. will remove it. i implemented this first and then added try-catch to getQuantizedVectorValues which made this case impossible..

src/main/java/org/opensearch/knn/index/query/exactsearch/NestedQuantizedVectorIdsExactKNNIterator.java
VijayanB: Agree, however i see this pattern (docId != NO_MORE_DOCS ) in Lucene and other iterators where it  protects against edge cases and makes the intent obvious.

src/main/java/org/opensearch/knn/index/query/exactsearch/QuantizedVectorIdsExactKNNIterator.java
VijayanB: i am not following this comment. can you elaborate on this?

VijayanB:
> @VijayanB Luckily, thanks to MOS exhaustive IT test, it will naturally cover all possible nested cases listed below:
> 
> 1. Dense or Sparse
> 2. Compression ratio : 8x, 16x 32x
> 3. Single or multi segments
> 4. Faiss build skip or build Faiss index
> 
> Other than this, please add ITs for areas that above is not covering.

Not sure whether MOS test will cover or not since the new iterator will hit only when we have restrictive filters. The current PR already contains Unit test and IT test which will make sure that it always trigger the Exact search and both iterators will be used to compute the score.

VijayanB:
> Hi @VijayanB Could you run cohere-10m benchmark to verify the performance improvement?

It might not be helpful, however will run filter tests with 1M dataset and post the results

src/main/java/org/opensearch/knn/index/query/exactsearch/ExactSearcher.java
```diff
@@ -238,7 +241,22 @@ private ExactKNNIterator getKNNIterator(LeafReaderContext leafReaderContext, Exa
                 );
             }
         }
-
+        if (quantizedQueryVector != null) {
+            // Use pre-quantized vectors from native engine to avoid quantization multiple times
+            final ByteVectorValues quantizedVectorValues = getQuantizedVectorValues(reader, exactSearcherContext.getField());
```
Vikasht34: So Logic here is if quantizedQueryVector is not null , get quantizedVectorValues but when I see Downmethod , it can have null also , and we are not checking null anywhere in code and this quantizedVectorValues is fed to iterator and then we are calling quantizedVectorValues.vectorValue which will throw null pointer exception.

src/main/java/org/opensearch/knn/index/query/exactsearch/ExactSearcher.java
```diff
@@ -261,6 +279,16 @@ private ExactKNNIterator getKNNIterator(LeafReaderContext leafReaderContext, Exa
         );
     }
 
+    private ByteVectorValues getQuantizedVectorValues(SegmentReader reader, String fieldName) throws IOException {
+        if (reader.getVectorReader() instanceof PerFieldKnnVectorsFormat.FieldsReader fieldsReader) {
```
Vikasht34: I am not quite confortable with this change , so according to this change ,  now ExcatSearcher has to know  about PerFieldKnnVectorsFormat which is codec details, now what will happen in future , if we need to have more quantization methods , we need to again modify the excat searcher to accomdate that. 

When I see this ,  I see we need answer for this question "do you have quantized vectors" today it's binary and tomorrow it could be somthing else. 

Somehow I was thinking , what we need for Excat ssearcher , Vector Source, Which is nothing but Our VectorValue Factory , so with this change at some place we are using KNNVectorValues or at some place we are using DirectLuceneByteVectorValue, if Somehow we can have that vector Source logic abstracted from here and Excat Seracher will have just depende on source like be it float vector value or Byte or Binary or simething else we can avoid this coupling.


@shatejas  @navneet1v what you guys think? I am open for suggestion here.

src/main/java/org/opensearch/knn/index/codec/quantization/QuantizedVectorsReader.java
```diff
@@ -0,0 +1,26 @@
...
+package org.opensearch.knn.index.codec.quantization;
+
+import org.apache.lucene.index.ByteVectorValues;
+
+import java.io.IOException;
+
+/**
+ * Interface for reading quantized vector values from an index.
+ */
+public interface QuantizedVectorsReader {
```
Vikasht34: On this interface , Second thought , do we really need it??  When I see this why we need it , I feel we are trying to utlize loadMemorySearcherIfNeeded, we already have all the Infromation needed to call this function upstream, what was your tought to force a NativeEnginerKNNReader with this inteface?

VijayanB: My thought process on having interface for QunatizationVectorsReader is due to
1. It follows Lucene's pattern - Lucene99ScalarQuantizedVectorsReader implements QuantizedVectorsReader
2.  The interface provides a clean contract for quantized vector access

Vikasht34: Thanks @VijayanB , Make Sense ,  But The Name of Reader in Lucene in Lucene99ScalarQuantizedVectorsReader not Lucene99ScalarVectorsReader so it's ovibious to implement QuantizedVectorsReader , this Interface is not implemented with Lucene99ScalarVectorsReader , but in our case we have NativeKNNVectorReader not NativeQuantizedKNNVectorReader so I am not sure it follows as Black and White but yeah I see your point.

src/main/java/org/opensearch/knn/index/query/exactsearch/ExactSearcher.java
VijayanB: If we move this to VectorValueFactory, i think we have to maintain map to hold Vector Source so that it can be returned when either MOS or ExactSearch asks for given field. Is that your recommendation or do you see any other way to do this without maintaining it in map?

src/main/java/org/opensearch/knn/index/codec/quantization/QuantizedVectorsReader.java
VijayanB: another example: Lucene99HnswVectorsReader

src/main/java/org/opensearch/knn/index/query/exactsearch/ExactSearcher.java
Vikasht34: My Suggestion is very Simple .

1. Remove the interface .
2. Keep the Same Method in NativeRaeder Since we are trying to utlize MemorySearcher(Or Push this to FaissMemeorySearcher to give always Singleton rather than forcing to have singleton in reader)
3. In Excat Seracher if Quantized is not null 
4. Call VectorValueFactory to give getNativeQuantizedVectorValues (The If Else logic of Checking format will be again Localised to VectorValueFactory) so in future if other places if we need to read QuantizedVectorValue we can always go through factory which we have , rathee that writing same logic which we have in excat Searcher.
5. Which will talk to NativeRaeder and get your ByteVectorValue


With this there is no interface and there is no format level check , and it alings with what we have in Excat searcher which is totally dependent of KNNVectorValue Factory

src/main/java/org/opensearch/knn/index/codec/quantization/QuantizedVectorsReader.java
Vikasht34: Thanks @VijayanB  for Pointing that out , This is my Understanding , Pardon me if I am wrong. As far as I undersatnd 

Lucene stores the graph structure separately from vector storage:

Graph: Node connections, levels

Vectors: Actual vector data (could be quantized or not)

But in Lucene99HnswVectorsReader we need vector data as well for Distance calculation , and since graph strtcure is common , and we need vector data , it makes sense to implement QuantizedVectorsReader.

In our case we need actual vector data. 

My Suggestion over here is , Right now we are writing below logic in Excat Searcher. 

Excat search has already a way to get VectorValues which is nothing but VectorValueFactory. If we could abstract this logic into VectorValueFactory , it will give us more flexibility , I will tell you why. so in future we again need to read vector from faiss file , we can directly read from factory else we need to write below logic again.

```
 if (reader.getVectorReader() instanceof PerFieldKnnVectorsFormat.FieldsReader fieldsReader) {
```

src/main/java/org/opensearch/knn/index/query/exactsearch/QuantizedVectorIdsExactKNNIterator.java
```diff
@@ -0,0 +1,104 @@
...
+import org.opensearch.common.Nullable;
+import org.opensearch.knn.index.SpaceType;
+
+import java.io.IOException;
+
+/**
+ * Iterator for exact KNN search using pre-quantized vectors from native engines.
+ * Compares quantized query vector against pre-quantized document vectors using Hamming distance.
+ */
+class QuantizedVectorIdsExactKNNIterator implements ExactKNNIterator {
```
Vikasht34: This class seems to be also not required, we already have BinaryVectorIdsExactKNNIterator, ByteVectorIdsExactKNNIterator, The reason why we ended up impolemting the class becuase of Constuctor mismatch Since we started using DirectLuceneVectorValues , if we map to KNNVectorValues, those classes can be untilized with minimal change , just to add extensibility for IndexSetIterator.

src/main/java/org/opensearch/knn/index/query/exactsearch/NestedQuantizedVectorIdsExactKNNIterator.java
```diff
@@ -0,0 +1,55 @@
...
+import org.apache.lucene.util.BitSet;
+import org.opensearch.common.Nullable;
+
+import java.io.IOException;
+
+/**
+ * Iterator for exact KNN search using quantized vectors with nested field support.
+ * Deduplicates docs per parent and returns the best child doc with the highest score.
+ */
+class NestedQuantizedVectorIdsExactKNNIterator extends QuantizedVectorIdsExactKNNIterator {
```
Vikasht34: I feel we already have another class NestedBinaryVectorIdsExactKNNIterator , I feel the above reason is again valid for this case.

src/main/java/org/opensearch/knn/index/query/exactsearch/ExactSearcher.java
Vikasht34: > If we move this to VectorValueFactory, i think we have to maintain map to hold Vector Source so that it can be returned when either MOS or ExactSearch asks for given field. Is that your recommendation or do you see any other way to do this without maintaining it in map?

We need DocIndexSetIerator Support in VectorValueFactory, if we can do that we can get rid of following 

1. New Inteface, 
2. New NestedQuantizedIteartor Class 
3. New QuantizedIteratorClass.

shatejas: @VijayanB as discussed offline, can we override getByteVectorValues to get the quantized vectors from the reader? you can also considert a flatformatwrapper over it if needed

shatejas: > We need DocIndexSetIerator Support in VectorValueFactory,

This I agree with, we shouldn't be bypassing the interface, @VijayanB Can we look into blockers of supporting this in KNNVectorValues?

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
```diff
@@ -108,14 +111,22 @@ public FloatVectorValues getFloatVectorValues(final String field) throws IOExcep
...
-     * @param field {@link String}
+     * @param field the vector field name
+     * @return {@link ByteVectorValues} for the field, never {@code null}
+     * @throws IOException if an I/O error occurs or no byte vectors are available for the field
      */
     @Override
     public ByteVectorValues getByteVectorValues(final String field) throws IOException {
+        final FieldInfo fieldInfo = segmentReadState.fieldInfos.fieldInfo(field);
+        if (fieldInfo.getVectorEncoding() == VectorEncoding.FLOAT32) {
+            final Optional<ByteVectorValues> quantizedVectorValues = getQuantizedVectorValues(fieldInfo);
```
shatejas: nit: I have seen usage of Optional adding an overhead in profiling, since this can be in search code path reconsider and use null with proper handling

src/main/java/org/opensearch/knn/index/vectorvalues/KNNVectorValuesFactory.java
```diff
@@ -115,10 +133,23 @@ public static <T> KNNVectorValues<T> getVectorValues(final FieldInfo fieldInfo,
                 new KNNVectorValuesIterator.DocIdsIteratorValues(leafReader.getByteVectorValues(fieldInfo.getName()))
             );
         } else if (fieldInfo.getVectorEncoding() == VectorEncoding.FLOAT32) {
+            final FloatVectorValues floatVectorValues = leafReader.getFloatVectorValues(fieldInfo.getName());
+            // for quantized query vectors, we need to pass the original byte vectors to the engine if supported, otherwise fallback
+            if (isQueryVectorQuantized && leafReader.getVectorReader() instanceof PerFieldKnnVectorsFormat.FieldsReader fieldsReader) {
```
shatejas: `leafReader.getVectorReader() instanceof PerFieldKnnVectorsFormat.FieldsReader fieldsReader` why is this condition needed?

src/main/java/org/opensearch/knn/index/query/exactsearch/NestedQuantizedVectorIdsExactKNNIterator.java
VijayanB: Using NestedBinaryVectorIdsExactKNNIterator as suggested

src/main/java/org/opensearch/knn/index/query/exactsearch/QuantizedVectorIdsExactKNNIterator.java
VijayanB: used BinaryVectorIdsExactKNNIterator as suggested

src/main/java/org/opensearch/knn/index/codec/quantization/QuantizedVectorsReader.java
VijayanB: Removed interface and incorporated the changes in getByteVectorValues as recommended. Thanks @Vikasht34

src/main/java/org/opensearch/knn/index/query/exactsearch/ExactSearcher.java
VijayanB: Updated factory to support iterator and vector values to support this use case.

src/test/java/org/opensearch/knn/integ/DiskBasedFilteredExactSearchIT.java
```diff
@@ -0,0 +1,290 @@
...
+    private void testDiskBasedFilteredExactSearchWhenFiltersMatchAllDocs(CompressionLevel compressionLevel) throws Exception {
+        String indexName = INDEX_NAME + "_" + compressionLevel.getName();
+        String filterFieldName = "color";
+        final int expectResultSize = randomIntBetween(1, 3);
+        final String filterValue = "red";
+
+        createDiskBasedIndex(indexName, FIELD_NAME, 8, compressionLevel);
+
+        for (int i = 0; i < 5; i++) {
+            addKnnDocWithAttributes(
```
Vikasht34, line 56: I think , we should improve this integ tests , as you can see for all the docs we are adding red color, we need diversification here , somedocs red , some without color and some differnt .

src/test/java/org/opensearch/knn/integ/DiskBasedFilteredExactSearchIT.java
```diff
@@ -0,0 +1,290 @@
...
+            .filterValue(filterValue)
+            .rescoreEnabled(false)
+            .build()
+            .getQueryString();
+
+        Response response = searchKNNIndex(indexName, query, expectResultSize);
+        String entity = EntityUtils.toString(response.getEntity());
+        List<String> docIds = parseIds(entity);
+        assertEquals(expectResultSize, docIds.size());
+        assertEquals(expectResultSize, parseTotalSearchHits(entity));
```
Vikasht34, line 84: We need to check if the same color docments with excat count we are getting.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
```diff
@@ -349,6 +361,23 @@ private IOSupplier<VectorSearcher> getVectorSearcherSupplier(final FieldInfo fie
...
     }
 
+    /**
+     * Retrieves quantized byte vectors from Faiss memory-optimized searcher if available.
+     *
+     * @param fieldInfo the field to retrieve vectors for
+     * @return {@link Optional} containing quantized byte vectors, or empty if not available
+     * @throws IOException if an I/O error occurs
+     */
+    private Optional<ByteVectorValues> getQuantizedVectorValues(@NonNull final FieldInfo fieldInfo) throws IOException {
```
Vikasht34: I was wondering how this is going to work , when we have 32x compression without filter , in Second phase search , now with this change , would it start using ByteValues?

VijayanB: When you say second phase search, do you mean rescore phase? in that case only getFloatVectorValues are called.

src/test/java/org/opensearch/knn/integ/DiskBasedFilteredExactSearchIT.java
VijayanB: i believe this test testDiskBasedFilteredExactSearch_whenNonVectorFields_thenSucceed covers non vector fields/multi filter value. Do you see any gap even with this?

VijayanB:
> After refactoring, Changes looks really good !! Thank you for all hard work !! Could you also complete Cohere 10m/1m bechmarking with filter like 0%, 5%, 10%, 15% etc.. and post result.

I am current working on running Benchmark results. I will publish the result in the github issue once i complete

src/main/java/org/opensearch/knn/index/query/exactsearch/ExactSearcher.java
```diff
@@ -238,7 +238,26 @@ private ExactKNNIterator getKNNIterator(LeafReaderContext leafReaderContext, Exa
                 );
             }
         }
-
+        // Quantized search path: retrieve quantized byte vectors from reader as KNNBinaryVectorValues and perform exact search
+        // using Hamming distance.
+        if (quantizedQueryVector != null) {
```
shatejas, line 243: quantizedVector will be null for ADC cases, so it seems with this logic for ADC it won't use byte vectors from faiss, can we handle ADC cases in the binary extractors?

Once we handle ADC here, quantization logic  from [VectorIdsKNNIterator](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/query/exactsearch/VectorIdsExactKNNIterator.java#L98) should be removed

shatejas (commented):
Yet to look at tests but they might change after incorporating ADC so will hold on to it

src/main/java/org/opensearch/knn/index/vectorvalues/KNNVectorValuesFactory.java
VijayanB: I need to type cast to PerFieldKnnVectorsFormat.FieldsReader in order to call getFieldReader to get underlying knnvector reader

src/main/java/org/opensearch/knn/index/query/exactsearch/ExactSearcher.java
VijayanB: For ADC, we need query vector as float vector instead of byte, this needs more refactoring,  i am thinking about doing it as part of follow up PR which includes handles ADC case and make [VectorIdsKNNIterator](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/query/exactsearch/VectorIdsExactKNNIterator.java#L98) independent of quantization. What do you think?

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissMemoryOptimizedSearcher.java
shatejas: > I thought we discussed that this will not be added to vectorsearcher interface, will use instance check to get ByteVectorValues

Can you elaborate about this decision why is interface not updated?

src/main/java/org/opensearch/knn/index/vectorvalues/KNNVectorValuesFactory.java
shatejas: > I need to type cast to PerFieldKnnVectorsFormat.FieldsReader in order to call getFieldReader to get underlying knnvector reader

whats blocking you to do `leafReader.getByteVectorVaues(fieldName)` ? what am I missing?

src/main/java/org/opensearch/knn/index/query/exactsearch/ExactSearcher.java
shatejas: This is being merged directly into main can we do it as a part of this please so its feature complete

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
VijayanB: updated to use null check

src/main/java/org/opensearch/knn/index/vectorvalues/KNNVectorValuesFactory.java
VijayanB: Yeah leafReader is not NativeEngines990KnnVectorReader, it is stored inside FieldsReader, so we have to first get FieldsReader and then get the reader for our field,

shatejas: > Yeah leafReader is not NativeEngines990KnnVectorReader, it is stored inside FieldsReader, so we have to first get FieldsReader and then get the reader for our field,

Doesn't it ultimately call NativeEngines990KnnVectorReader?

src/main/java/org/opensearch/knn/index/query/exactsearch/ExactSearcher.java
VijayanB: As discussed, this will be taken as next PR.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
VijayanB: As discussed this will not be called since rescore is always sets useQuantizedVectorSearch as false.

src/test/java/org/opensearch/knn/integ/DiskBasedFilteredExactSearchIT.java
VijayanB: This is addressed in next test case

src/test/java/org/opensearch/knn/integ/DiskBasedFilteredExactSearchIT.java
VijayanB: As discussed, this is addressed in the testDiskBasedFilteredExactSearch_whenNonVectorFields_thenSucceed

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
```diff
@@ -108,14 +110,22 @@ public FloatVectorValues getFloatVectorValues(final String field) throws IOExcep
...
+     * @throws IOException if an I/O error occurs or no byte vectors are available for the field
      */
     @Override
     public ByteVectorValues getByteVectorValues(final String field) throws IOException {
+        final FieldInfo fieldInfo = segmentReadState.fieldInfos.fieldInfo(field);
+        if (fieldInfo.getVectorEncoding() == VectorEncoding.FLOAT32) {
+            final ByteVectorValues quantizedVectorValues = getQuantizedVectorValues(fieldInfo);
+            if (quantizedVectorValues != null) {
+                return quantizedVectorValues;
+            }
```
shatejas, line 127: nit: Can you add a warn message after if. Ideally it should never be hit since its guarded by VectorEncoding.FLOAT32

src/test/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReaderTests.java
```diff
@@ -42,12 +50,76 @@ public void testWhenMemoryOptimizedSearchIsEnabled_emptyCase() {
...
+
+        final FlatVectorsReader flatVectorsReader = mock(FlatVectorsReader.class);
+        try (MockedStatic<KNNEngine> mockedStatic = mockStatic(KNNEngine.class)) {
+            mockedStatic.when(() -> KNNEngine.getEngine(any())).thenReturn(mockFaiss);
+            final Set<String> filesInSegment = Set.of("_0_165_field1.faiss");
+            mockedStatic.when(KNNEngine::getEnginesThatCreateCustomSegmentFiles).thenReturn(ImmutableSet.of(mockFaiss));
+            NativeEngines990KnnVectorsReader reader = createReader(fieldInfos, filesInSegment, flatVectorsReader);
+            reader.getByteVectorValues("field1");
+            verify(mockSearcher).getByteVectorValues();
+            verify(flatVectorsReader, Mockito.never()).getByteVectorValues("field1");
```
shatejas, line 102: nit: 
- Seems like a weak test. Can you use [Predefined values](https://github.com/opensearch-project/k-NN/blob/main/src/test/java/org/opensearch/knn/index/vectorvalues/TestVectorValues.java#L269) and iterate through to make sure they are not manipulated
- try not to use any()

src/test/java/org/opensearch/knn/integ/DiskBasedFilteredExactSearchIT.java
```diff
@@ -0,0 +1,290 @@
...
+            .build()
+            .getQueryString();
+
+        Response response = searchKNNIndex(indexName, query, 10);
+        assertOK(response);
+        String entity = EntityUtils.toString(response.getEntity());
+        List<String> docIds = parseIds(entity);
+        assertEquals(2, docIds.size());
+        assertEquals("3", docIds.get(0));
+        assertEquals("1", docIds.get(1));
```
shatejas, line 241: nit: Do we want to check the scores as well?

VijayanB: Since quantization is happening inside indexing it is not possible to assert unless i pre-calculate the quantization, given docIds are returning in correct order, this should give signal that it is valid.

src/test/java/org/opensearch/knn/index/query/exactsearch/ExactSearcherTests.java
```diff
@@ -188,6 +188,86 @@ public void testRadialSearch_whenNoEngineFiles_thenSuccess() {
...
+                new float[] { 2, 3, 4 },
+                new float[] { 7, 8, 9 },
+                new float[] { 0, 1, 2 }
+            );
+            final List<byte[]> quantizedDataVectors = List.of(
+                new byte[] { 1, 2, 3 },
+                new byte[] { 4, 5, 6 },
+                new byte[] { 2, 3, 4 },
+                new byte[] { 7, 8, 9 },
+                new byte[] { 0, 1, 2 }
```
shatejas, line 216: nit: quantized vectors are 0 or 1 right, ideally we should have close to real data

src/test/java/org/opensearch/knn/index/vectorvalues/KNNVectorValuesFactoryTests.java
```diff
@@ -65,6 +68,52 @@ public void testGetVectorValuesUsingDocWithFieldSet_whenValidInput_thenSuccess()
...
+        final FieldInfo fieldInfo = Mockito.mock(FieldInfo.class);
+        final SegmentReader reader = Mockito.mock(SegmentReader.class);
+        Mockito.when(fieldInfo.hasVectorValues()).thenReturn(true);
+        Mockito.when(fieldInfo.getName()).thenReturn("test_field");
+        Mockito.when(fieldInfo.getVectorEncoding()).thenReturn(VectorEncoding.FLOAT32);
+        Mockito.when(fieldInfo.getAttribute(KNNConstants.VECTOR_DATA_TYPE_FIELD)).thenReturn(VectorDataType.FLOAT.getValue());
+        Mockito.when(reader.getFloatVectorValues("test_field"))
+            .thenReturn(new TestVectorValues.PreDefinedFloatVectorValues(floatArrayList));
+
+        final org.apache.lucene.codecs.KnnVectorsReader vectorsReader = Mockito.mock(org.apache.lucene.codecs.KnnVectorsReader.class);
```
shatejas, line 84: nit: use imports instead?

src/test/java/org/opensearch/knn/index/vectorvalues/KNNVectorValuesFactoryTests.java
```diff
@@ -65,6 +68,52 @@ public void testGetVectorValuesUsingDocWithFieldSet_whenValidInput_thenSuccess()
...
+        Mockito.when(reader.getVectorReader()).thenReturn(vectorsReader);
+        Mockito.when(vectorsReader.getByteVectorValues("test_field"))
+            .thenReturn(new TestVectorValues.PreDefinedByteVectorValues(quantizedByteArrayList));
+
+        final KNNVectorValues<byte[]> quantizedVectorValues = KNNVectorValuesFactory.getVectorValues(fieldInfo, reader, true);
+        quantizedVectorValues.nextDoc();
+        Assert.assertArrayEquals(quantizedByteArrayList.getFirst(), quantizedVectorValues.getVector());
+        Assert.assertNotNull(quantizedVectorValues);
```
shatejas, line 92: nit: add multiple vectors in input and loop through for assertions to make it tighter, we don't want out-of-order or less or more values

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
VijayanB: Sure

src/test/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReaderTests.java
VijayanB: I added new test which uses Predefined values to test scores. Here, kept verify scope to only whether method is called or not.  w.r.t to any() sure will clean it up in next PR

src/test/java/org/opensearch/knn/integ/DiskBasedFilteredExactSearchIT.java
VijayanB: since it is not possible to compare the score ,i added test in ExactSearcher which will check for scores

src/test/java/org/opensearch/knn/index/vectorvalues/KNNVectorValuesFactoryTests.java
VijayanB: Ack

src/test/java/org/opensearch/knn/index/query/exactsearch/ExactSearcherTests.java
VijayanB: Ack

src/test/java/org/opensearch/knn/index/vectorvalues/KNNVectorValuesFactoryTests.java
VijayanB: Ack. This needs update to all test in the same class. Will include it in next PR

PR #3110, opened 2026-02-10
Fix score conversion logic for filtered radial exact search

### LINKED ISSUE #3099: [BUG] min_score Threshold Ignored with FAISS Efficient Filtering
**What is the bug?**
When using FAISS engine with radial search (`min_score`) and efficient filtering (filter inside knn query), OpenSearch returns results with scores **below** the specified `min_score` threshold. The threshold is correctly enforced when no filter is applied or when using post-filtering.

Step 1: Create index with FAISS engine

```bash
PUT /test-faiss-bug
{
  "settings": {
    "index": {
      "knn": true
    }
  },
  "mappings": {
    "properties": {
      "embedding": {
        "type": "knn_vector",
        "dimension": 768,
        "method": {
          "name": "hnsw",
          "engine": "faiss",
          "space_type": "cosinesimil",
          "parameters": {
            "ef_construction": 100,
            "m": 16
          }
        }
      },
      "category": {
        "type": "keyword"
      }
    }
  }
}
```

Step 2: Index sample documents

```bash
# Index 20 documents with similar vectors in category "electronics"
POST /test-faiss-bug/_doc/1
{
  "embedding": [0.1, 0.2, ..., 0.5],  # 768-dimensional vector
  "category": "electronics"
}

# Repeat for documents 2-20 with slightly varied vectors
```

Step 3: Query with min_score and efficient filtering

```bash
POST /test-faiss-bug/_search
{
  "size": 100,
  "query": {
    "knn": {
      "embedding": {
        "vector": [0.1, 0.2, ..., 0.5],
        "min_score": 0.935,
        "filter": {
          "term": {
            "category": "electronics"
          }
        }
      }
    }
  }
}
```

Step 4: Observe incorrect results

Response contains documents with `_score` values **below 0.935** (e.g., 0.934, 0.918, 0.900).

**Expected**: All returned documents should have `_score >= 0.935`

**Actual**: Documents with scores below 0.935 are returned when efficient filtering is used

- **OS**: AL2023
- **OpenSearch Version**: 3.1.0
- **Engine**: FAISS

Current workaround: Use post-filtering
```json
{
  "query": {
    "bool": {
      "must": [
        {
          "knn": {
            "embedding": {
              "vector": [...],
              "min_score": 0.935
            }
          }
        }
      ],
      "filter": [
        { "term": { "category": "electronics" } }
      ]
    }
  }
}
```

kotwanikunal:
### Description
- Fixes score translation logic for exact radial search with cosine space type

### Related Issues
Resolves #3099 

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/query/exactsearch/ExactSearcher.java
+++ b/src/main/java/org/opensearch/knn/index/query/exactsearch/ExactSearcher.java
@@ -238,7 +238,26 @@ public class ExactSearcher {
                 );
             }
         }
-
+        // Quantized search path: retrieve quantized byte vectors from reader as KNNBinaryVectorValues and perform exact search
+        // using Hamming distance.
+        if (quantizedQueryVector != null) {
+            final KNNVectorValues<byte[]> vectorValues = KNNVectorValuesFactory.getVectorValues(fieldInfo, reader, true);
+            if (isNestedRequired) {
+                return new NestedBinaryVectorIdsExactKNNIterator(
+                    matchedDocs,
+                    quantizedQueryVector,
+                    (KNNBinaryVectorValues) vectorValues,
+                    SpaceType.HAMMING,
+                    exactSearcherContext.getParentsFilter().getBitSet(leafReaderContext)
+                );
+            }
+            return new BinaryVectorIdsExactKNNIterator(
+                matchedDocs,
+                quantizedQueryVector,
+                (KNNBinaryVectorValues) vectorValues,
+                SpaceType.HAMMING
+            );
+        }
         final KNNVectorValues<float[]> vectorValues = KNNVectorValuesFactory.getVectorValues(fieldInfo, reader);
         if (isNestedRequired) {
             return new NestedVectorIdsExactKNNIterator(
--- a/.github/workflows/backwards_compatibility_tests_workflow.yml
+++ b/.github/workflows/backwards_compatibility_tests_workflow.yml
@@ -36,7 +36,7 @@ jobs:
         java: [ 21 ]
         os: [ubuntu-latest]
         bwc_version : [ "2.0.1", "2.1.0", "2.2.1", "2.3.0", "2.4.1", "2.5.0", "2.6.0", "2.7.0", "2.8.0", "2.9.0", "2.10.0", "2.11.0", "2.12.0", "2.13.0", "2.14.0", "2.15.0", "2.16.0", "2.17.0", "2.18.0","2.19.1", "2.20.0-SNAPSHOT"]
-        opensearch_version : [ "3.5.0-SNAPSHOT" ]
+        opensearch_version : [ "3.6.0-SNAPSHOT" ]
         exclude:
           - os: windows-latest
             bwc_version: "2.0.1"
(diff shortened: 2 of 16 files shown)


src/main/java/org/opensearch/knn/index/query/exactsearch/ExactSearcher.java
```diff
@@ -104,9 +104,9 @@ private TopDocs doRadialSearch(LeafReaderContext leafReaderContext, ExactSearche
         // We need to use the given radius when memory optimized search is enabled. Since it relies on Lucene's scoring framework, the given
         // max distance is already converted min score then saved in `radius`. Thus, we don't need a score translation which does not make
         // sense as it is treating min score as a max distance otherwise.
-        final float minScore = context.isMemoryOptimizedSearchEnabled
-            ? context.getRadius()
-            : spaceType.scoreTranslation(context.getRadius());
+        // For FAISS exact search, radius is in FAISS distance space (e.g., inner product for cosine).
+        // We need to convert it to OpenSearch score space using the reverse translation.
+        final float minScore = context.isMemoryOptimizedSearchEnabled ? context.getRadius() : engine.score(context.getRadius(), spaceType);
```
Vikasht34, line 109: Curious : Same Problem could also happen with MemoryOptimized Search as well , or it's like score calaulation is differnt?

Vikasht34 (commented):
Overall Looks Good , Some. Clarification Question

src/main/java/org/opensearch/knn/index/query/exactsearch/ExactSearcher.java
VijayanB: Can we also make sure we have IT for MOS if it is not already there?

src/test/java/org/opensearch/knn/index/FaissIT.java
```diff
@@ -236,6 +236,83 @@ public void testEndToEnd_whenDoRadiusSearch_whenScoreThreshold_whenMethodIsHNSWF
...
+            new float[][] { queryVector },
+            null,
+            minScore,
+            spaceType,
+            filterQuery,
+            null
+        );
+
+        for (KNNResult result : results.get(0)) {
+            assertTrue(String.format("Score %.4f below threshold %.3f", result.getScore(), minScore), result.getScore() >= minScore);
```
shatejas, line 310: Ideally we want to be at a place where we match the exact scores with an error factor

src/main/java/org/opensearch/knn/index/query/exactsearch/ExactSearcher.java
kotwanikunal: MOS does not need this fix as it has a separate score calculation flow. 

References:
- https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java#L516-L521
- MOS then uses - https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/engine/KNNEngine.java#L199-L201
but Native uses https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/engine/faiss/Faiss.java#L111-L117
- MOS stays in the score space vs Native needs to be converted from radial distance to score - leading to the bug

I'll add in a test for MOS as a followup

src/test/java/org/opensearch/knn/index/FaissIT.java
kotwanikunal: I'll update the test with the MOS follow up.

PR #3113, opened 2026-02-11
Use pre-quantized vectors for ADC

VijayanB:
### Description
This PR implements support for using pre-quantized vectors with Asymmetric Distance Computation (ADC) in the k-NN plugin. This optimization allows the system to work directly with quantized vectors, improving performance by avoiding redundant quantization operations during search.

### Related Issues

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/query/exactsearch/BinaryVectorIdsExactKNNIterator.java
+++ b/src/main/java/org/opensearch/knn/index/query/exactsearch/BinaryVectorIdsExactKNNIterator.java
@@ -28,24 +28,6 @@ class BinaryVectorIdsExactKNNIterator implements ExactKNNIterator {
     protected float currentScore = Float.NEGATIVE_INFINITY;
     protected int docId;
 
-    private BinaryVectorIdsExactKNNIterator(
-        @Nullable final DocIdSetIterator docIdSetIterator,
-        final byte[] byteQueryVector,
-        final float[] floatQueryVector,
-        final KNNBinaryVectorValues binaryVectorValues,
-        final SpaceType spaceType
-    ) throws IOException {
-        assert (floatQueryVector == null) != (byteQueryVector == null)
-            : "Exactly one of byteQueryVector or floatQueryVector must be non-null";
-        this.docIdSetIterator = docIdSetIterator;
-        this.byteQueryVector = byteQueryVector;
-        this.floatQueryVector = floatQueryVector;
-        this.binaryVectorValues = binaryVectorValues;
-        this.spaceType = spaceType;
-        this.docId = getNextDocId();
-
-    }
-
     public BinaryVectorIdsExactKNNIterator(
         @Nullable final DocIdSetIterator docIdSetIterator,
         final byte[] byteQueryVector,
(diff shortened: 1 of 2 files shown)


VijayanB:
@shatejas @Vikasht34 Can you take a look at this PR? Thank you.

Vikasht34 (commented):
Overall Looks Good , Will Take a final look tomorrow

src/main/java/org/opensearch/knn/index/query/exactsearch/BinaryVectorIdsExactKNNIterator.java
```diff
@@ -20,33 +21,55 @@
...
+    public BinaryVectorIdsExactKNNIterator(
+        @Nullable final DocIdSetIterator docIdSetIterator,
+        final float[] floatQueryVector,
+        final KNNBinaryVectorValues binaryVectorValues,
+        final SpaceType spaceType
+    ) throws IOException {
+        this(docIdSetIterator, null, floatQueryVector, binaryVectorValues, spaceType);
+    }
+
+    public BinaryVectorIdsExactKNNIterator(
```
navneet1v: since this is a new iterator, I would suggest reducing the number of constructors. There are too many here.

src/main/java/org/opensearch/knn/plugin/script/KNNScoringUtil.java
```diff
@@ -174,6 +174,27 @@ public static float innerProductADC(float[] queryVector, byte[] inputVector) {
...
+     * @throws UnsupportedOperationException if the space type is not supported for ADC
+     */
+    public static float scoreWithADC(float[] queryVector, byte[] documentVector, SpaceType spaceType) {
+        if (spaceType.equals(SpaceType.L2)) {
+            return SpaceType.L2.scoreTranslation(l2SquaredADC(queryVector, documentVector));
+        } else if (spaceType.equals(SpaceType.INNER_PRODUCT)) {
+            return SpaceType.INNER_PRODUCT.scoreTranslation((-1 * innerProductADC(queryVector, documentVector)));
+        } else if (spaceType.equals(SpaceType.COSINESIMIL)) {
+            return SpaceType.COSINESIMIL.scoreTranslation(1 - innerProductADC(queryVector, documentVector));
+        }
```
navneet1v, line 194: can this distance computations be moved under SpaceType enum?

src/main/java/org/opensearch/knn/index/query/exactsearch/BinaryVectorIdsExactKNNIterator.java
VijayanB: will remove one that was only used by test, to reuse other iterator.

src/main/java/org/opensearch/knn/plugin/script/KNNScoringUtil.java
VijayanB: Sure. I just moved this method from existing iterator to here, i can take this up in follow up PR as refactoring. It needs some investigation if we need to merge this to space type and reuse in both scorer and iterator cc @finnroblin

PR #3117, opened 2026-02-14
[Perf-Enhancement] Use right Vector Scorer when segments are initialized using SPI and also corrected the maxConn for MOS

navneet1v:
#### Enhancements

This change contains 2 improvements for Memory Optimized Search:
1. When a segment is initialized using SPI in case of shard movements or a cluster restart etc, then use the correct VectorScorer coming from `FlatVectorScorerUtil.getLucene99FlatVectorsScorer()`. This same scorer is used when segment is created for the [first time](https://github.com/opensearch-project/k-NN/blob/0495458f3832242956f72a171cd2b4e752b409c9/src/main/java/org/opensearch/knn/index/codec/BasePerFieldKnnVectorsFormat.java#L160-L164). FlatVectorScorerUtil either returns a DefaultVectorScorer or returns a PanamaVectorScorer. For our usecases mostly we should get `PanamaVectorScorer` which is superior than `DefaultVectorScorer`
2. For Memory Optimized search, return 16 as max conn when M=16 rather than 32. Since last layer of HNSW has 2*M neighbors we were returning neighbors list size as max conn, but ideally it should be M/2 . Lucene itself takes cares of doing 2*M in [graph searcher](https://github.com/apache/lucene/blob/65d91569c6abc08bead550534c5a5d35444946a2/lucene/core/src/java/org/apache/lucene/util/hnsw/HnswGraphSearcher.java#L282-L285). This is also consistent with Lucene where they return M as [maxConn](https://github.com/apache/lucene/blob/65d91569c6abc08bead550534c5a5d35444946a2/lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsReader.java#L545).

#### Bug fix
1. Fixed a bug with approximate graph threshold where if we have 2 indices with 2 different thresholds then the index which will get created at last its thresholds will be used in the older indices new segment. I added a unit test to ensure that to repro the test.
2. NativeEngine format was using Lucene engine for its max dimensions. Fixed it to use Faiss Engine. Even though both Lucene and Faiss uses 16K as max dimension so no customer impact but this is still a bug. Added UT for the same.

### Expectation
1. With change 1, we should see some improvements in latency, although it will not be drastic but it will be consistent to how Lucene works. 
2. For change 2, there should be little bit less pressure on heap per query per segment, but its a good to have optimization which may come handy in high QPS scenarios.
3. Bug fixes should now ensure that approx graph threshold is used correctly.

### Related Issues
There is no GH for this. I found these inconsistencies in the code during my code reading and deep-dive.

### Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,4 +18,5 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Enhancements
 * Make Merge in nativeEngine can Abort [#2529](https://github.com/opensearch-project/k-NN/pull/2529)
 * Use pre-quantized vectors from native engines for exact search [#3095](https://github.com/opensearch-project/k-NN/pull/3095)
+* Use right Vector Scorer when segments are initialized using SPI and also corrected the maxConn for MOS [#3117](https://github.com/opensearch-project/k-NN/pull/3117)
 
--- a/src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsFormat.java
+++ b/src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsFormat.java
@@ -15,7 +15,7 @@ import lombok.extern.log4j.Log4j2;
 import org.apache.lucene.codecs.KnnVectorsFormat;
 import org.apache.lucene.codecs.KnnVectorsReader;
 import org.apache.lucene.codecs.KnnVectorsWriter;
-import org.apache.lucene.codecs.hnsw.DefaultFlatVectorScorer;
+import org.apache.lucene.codecs.hnsw.FlatVectorScorerUtil;
 import org.apache.lucene.codecs.hnsw.FlatVectorsFormat;
 import org.apache.lucene.codecs.lucene99.Lucene99FlatVectorsFormat;
 import org.apache.lucene.index.SegmentReadState;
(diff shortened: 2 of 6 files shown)


navneet1v:
Added 2 more bug fix in the PR. Details in PR description

CHANGELOG.md
```diff
@@ -16,6 +16,5 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Refactoring
 
 ### Enhancements
-* Make Merge in nativeEngine can Abort [#2529](https://github.com/opensearch-project/k-NN/pull/2529)
```
kotwanikunal, line 19: I don't think this caught the 3.5 release train. https://github.com/opensearch-project/k-NN/blob/3.5/release-notes/opensearch-k-NN.release-notes-3.5.0.0.md

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissHnswGraph.java
```diff
@@ -195,6 +195,6 @@ public int neighborCount() {
 
     @Override
     public int maxConn() {
-        return faissHnsw.getMaxNumNeighbors();
+        return faissHnsw.getMaxNumNeighbors() / 2;
```
owenhalpert, line 201: Should we add a comment here to explain?

CHANGELOG.md
```diff
@@ -16,6 +16,5 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
...
-* Use pre-quantized vectors from native engines for exact search [#3095](https://github.com/opensearch-project/k-NN/pull/3095)
+* Use right Vector Scorer when segments are initialized using SPI and also corrected the maxConn for MOS [#3117](https://github.com/opensearch-project/k-NN/pull/)
```
owenhalpert: Don't forget the pull request number in the link :)

CHANGELOG.md
navneet1v: @kotwanikunal I see it got backported to 3.5 Hence I removed it. Ref: https://github.com/opensearch-project/k-NN/pull/3111

Let me check were are the tags for 3.5

navneet1v: You are right, we merged in 3.5 but its not released. Not sure why we did that. Let me fix this change log

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissHnswGraph.java
navneet1v: Good catch. I added it in the description but code level is also needed.

CHANGELOG.md
navneet1v: Good catch.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissHnswGraph.java
```diff
@@ -195,6 +195,9 @@ public int neighborCount() {
 
     @Override
     public int maxConn() {
-        return faissHnsw.getMaxNumNeighbors();
+        // The maximum number of neighbors in bottom layer of HNSW in faiss and also in lucene is 2*M. But the
+        // expectation from this function is to return M. The factor of 2 is already taken care by Lucene in
+        // graph searcher class. Hence, we are dividing here by 2 to ensure that we return correct value of M.
+        return faissHnsw.getMaxNumNeighbors() / 2;
```
0ctopus13prime, line 201: One edge case is that `getMaxNumNeighbors` can return 0, I know which is extremely rare case. Not sure what will happens in LuceneHnswSearcher when maxConn returned 0. 

In such case, I think it would be better to return non-zero value, what do you think?

neetikasinghal:
would be good if we can add some integration tests for the bug fixes also along with unit tests

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissHnswGraph.java
navneet1v: I think 0/2 will give 0 only. So we should be good here.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsFormat.java
```diff
@@ -33,35 +33,27 @@
 @Log4j2
 public class NativeEngines990KnnVectorsFormat extends KnnVectorsFormat {
     /** The format for storing, reading, merging vectors on disk */
-    private static FlatVectorsFormat flatVectorsFormat;
+    private static final FlatVectorsFormat flatVectorsFormat = new Lucene99FlatVectorsFormat(
+        FlatVectorScorerUtil.getLucene99FlatVectorsScorer()
+    );
     private static final String FORMAT_NAME = "NativeEngines990KnnVectorsFormat";
-    private static int approximateThreshold;
+    private final int approximateThreshold;
```
shatejas, line 40: nice catch!

navneet1v: Thanks

Issue #2964, opened 2025-10-28
[FEATURE] Enforce min value of index.knn.remote_index_build.size.min


jed326:
**Is your feature request related to a problem?**
`index.knn.remote_index_build.size.min` should not be allowed to be set below a certain point, as below that GPU acceleration will not be useful and in fact potentially harmful.

**What solution would you like?**
The `index.knn.remote_index_build.size.min` setting should have a min value.

**What alternatives have you considered?**

**Do you have any additional context?**
This change is somewhat involved because all the current remote build tests rely on setting the setting to `0`, so any min value will break this and require a lot of test fixes.

navneet1v:
@Gankris96 please go ahead

PR #3118, opened 2026-02-15
Add minimum value validation for remote_index_build.size.min

drtootsie:
Addresses #2964

## Problem
`index.knn.remote_index_build.size.min` can currently be set to any value including 0, but GPU acceleration for remote index builds is not useful below a certain threshold and can actually be harmful for very small segments.

## Solution
Add a minimum value of 1KB to the setting validation.

## Changes
```diff
public static final Setting<ByteSizeValue> KNN_INDEX_REMOTE_VECTOR_BUILD_SIZE_MIN_SETTING = Setting.byteSizeSetting(
    KNN_INDEX_REMOTE_VECTOR_BUILD_SIZE_MIN,
    KNN_INDEX_REMOTE_VECTOR_BUILD_THRESHOLD_DEFAULT_VALUE,
+   new ByteSizeValue(1, ByteSizeUnit.KB),
    Dynamic,
    IndexScope
);
```

## Impact
- Prevents setting the threshold too low where GPU acceleration is ineffective
- **Breaking change for tests:** Many existing tests set this to 0 and will need updates
- Users attempting to set values below 1KB will receive validation errors

## Testing Notes
As mentioned in the issue, this change will require updating all remote build tests that currently rely on setting this value to 0. The tests will need to use a value >= 1KB instead.

## Alternative Considered
Could use a higher minimum (e.g., 10KB or 100KB) based on actual GPU acceleration performance characteristics, but starting with 1KB as a conservative minimum.


src/main/java/org/opensearch/knn/index/KNNSettings.java
```diff
@@ -444,6 +444,7 @@ public void validate(String key, Object value, Object dependency) {
     public static final Setting<ByteSizeValue> KNN_INDEX_REMOTE_VECTOR_BUILD_SIZE_MIN_SETTING = Setting.byteSizeSetting(
         KNN_INDEX_REMOTE_VECTOR_BUILD_SIZE_MIN,
         KNN_INDEX_REMOTE_VECTOR_BUILD_THRESHOLD_DEFAULT_VALUE,
+        new ByteSizeValue(1, ByteSizeUnit.KB),
```
kotwanikunal, line 447: Is the intention here to prevent user error?
The default value `KNN_INDEX_REMOTE_VECTOR_BUILD_THRESHOLD_DEFAULT_VALUE` should help unless the user explicitly overrides the value.

Gankris96: Yes the intention is to prevent users from setting the value to be something too low, such that the remote index build doesn't really show much performance benefit.

Issue #3102, opened 2026-02-05
Native thread leak during HNSW indexing when index_thread_qty > 1


Jacob411:
### What is the bug?
There is a significant native thread leak in the k-NN plugin during HNSW indexing. The operating system reports a continuous, linear increase in threads (LWPs) that does not plateau. In our environment, this reached over **77,000 threads**, eventually causing the OS to kill the OpenSearch process or hit `max_user_processes` limits.

### How can one reproduce the bug?
1. Use OpenSearch **3.3.2** or **3.4.0** with the k-NN plugin installed.
2. Configure an HNSW index.
3. Ensure `knn.algo_param.index_thread_qty` is set to a value > 1 (default on most multi-core systems).
4. Perform continuous bulk indexing with a standard `refresh_interval` (e.g., 1s or 60s).
5. Monitor the OS thread count for the OpenSearch PID using `ps -o nlwp <PID>` or `btop`.
6. Observe that the thread count rises steadily and is never reclaimed, even after indexing stops.

### What is the expected behavior?
The native thread pool used for HNSW graph construction should properly `join` and terminate worker threads or reuse a fixed-size pool. The OS-level thread count should remain stable and proportional to the configured thread pool settings.

### What is your host/environment?
- **OS:**  Ubuntu 24.04
- **Version:** 3.3.2 (also confirmed on 3.4.0)
- **Plugins:** k-NN (HNSW)

vamshin:
Thanks @Jacob411. Do you see this in older versions may be 3.1.0 or 3.2.0. This will help narrow down issue faster.

Jacob411:
Yes I believe so. However, the only versions where I specifically reproduced were 3.3.2 and 3.4.0

vamshin:
could you please help validate if issue is with 3.1 and 3.2 as well? Trying to narrow down the release introduce the blog.

Jacob411:
@navneet1v No the threads would not decrease even when no indexing/merge/searches were happening

dbackeus:
> So I am bit confused here that is your issue is thread leak or something else.

@navneet1v  just look at the graph above, all our cluster nodes started growing up to ~125k threads followed by a crash on all our cluster nodes just after we started indexing using `knn`. So yes, this bug exists on lucene as well.

I can also confirm that setting `knn.algo_param.index_thread_qty: 1` on the cluster stopped the thread leakage:
[...]

PR #3119, opened 2026-02-16
Made threading persistent for KNN9120PerFieldKnnVectorsFormat

MrFlap:
### Description
Fix thread leak in HNSW merge executor

KNN9120PerFieldKnnVectorsFormat.getMergeThreadCountAndExecutorService() was creating a new Executors.newFixedThreadPool() on every invocation. Since this is called per field per segment during flush/merge, thread pools accumulated unboundedly under continuous indexing (77k+ leaked threads observed).

This PR replaces the per-call allocation with a cached singleton executor that is reused across calls and only recreated when knn.algo_param.index_thread_qty changes. The previous executor is gracefully shut down on recreation or when the setting drops to ≤ 1.

thread_leak_comparison

### Related Issues
Resolves (https://github.com/opensearch-project/k-NN/issues/3102)

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/codec/KNN9120Codec/KNN9120PerFieldKnnVectorsFormat.java
+++ b/src/main/java/org/opensearch/knn/index/codec/KNN9120Codec/KNN9120PerFieldKnnVectorsFormat.java
@@ -24,6 +24,8 @@ import java.util.concurrent.Executors;
  */
 public class KNN9120PerFieldKnnVectorsFormat extends BasePerFieldKnnVectorsFormat {
     private static final Tuple<Integer, ExecutorService> DEFAULT_MERGE_THREAD_COUNT_AND_EXECUTOR_SERVICE = Tuple.tuple(1, null);
+    private static volatile ExecutorService cachedMergeExecutorService;
+    private static volatile int cachedMergeThreadCount;
 
     public KNN9120PerFieldKnnVectorsFormat(final Optional<MapperService> mapperService) {
         this(mapperService, new NativeIndexBuildStrategyFactory());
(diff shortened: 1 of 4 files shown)


src/main/java/org/opensearch/knn/index/codec/KNN9120Codec/KNN9120PerFieldKnnVectorsFormat.java
```diff
@@ -91,16 +93,26 @@ public int getMaxDimensions(String fieldName) {
...
+            shutdownCachedExecutor();
             return DEFAULT_MERGE_THREAD_COUNT_AND_EXECUTOR_SERVICE;
-        } else {
-            return Tuple.tuple(mergeThreadCount, Executors.newFixedThreadPool(mergeThreadCount));
+        }
+        if (cachedMergeExecutorService != null && cachedMergeThreadCount == mergeThreadCount) {
+            return Tuple.tuple(cachedMergeThreadCount, cachedMergeExecutorService);
+        }
+        shutdownCachedExecutor();
+        cachedMergeExecutorService = Executors.newFixedThreadPool(mergeThreadCount);
```
navneet1v, line 106: Here rather than using a new Executor can we create a static executor and use it all the time.

MrFlap: @navneet1v Is there a case where we would change the thread count? This is why I have it dynamic.

navneet1v: Yes the thread counts value is dynamic. but it we should pick the thread counts once when we are creating the format.

PR #3120, opened 2026-02-17
Keep a static, dynamically threaded executor to seamlessly switch between thread counts

MrFlap:
### Description
Fix thread leak in HNSW merge executor

KNN9120PerFieldKnnVectorsFormat.getMergeThreadCountAndExecutorService() was creating a new Executors.newFixedThreadPool() on every invocation. Since this is called per field per segment during flush/merge, thread pools accumulated unboundedly under continuous indexing (77k+ leaked threads observed).

This PR replaces the per-call allocation with a static, dynamically sized executor that is reused across calls and changes thread count when knn.algo_param.index_thread_qty changes.

### Related Issues
Resolves https://github.com/opensearch-project/k-NN/issues/3102

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/KNN1040PerFieldKnnVectorsFormat.java
+++ b/src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/KNN1040PerFieldKnnVectorsFormat.java
@@ -11,7 +11,6 @@ import org.apache.lucene.codecs.lucene99.Lucene99HnswVectorsFormat;
 import org.apache.lucene.codecs.lucene104.Lucene104ScalarQuantizedVectorsFormat;
 import org.opensearch.common.collect.Tuple;
 import org.opensearch.index.mapper.MapperService;
-import org.opensearch.knn.index.KNNSettings;
 import org.opensearch.knn.index.SpaceType;
 import org.opensearch.knn.index.codec.KNN1040BasePerFieldKnnVectorsFormat;
 import org.opensearch.knn.index.codec.KnnVectorsFormatContext;
@@ -27,7 +26,9 @@ import org.opensearch.knn.index.engine.lucene.LuceneCodecFormatResolver;
 import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
+import java.util.concurrent.SynchronousQueue;
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
 import java.util.function.Function;
 
 /**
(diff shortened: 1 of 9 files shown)


src/main/java/org/opensearch/knn/index/codec/KNN9120Codec/KNN9120PerFieldKnnVectorsFormat.java
```diff
@@ -9,21 +9,47 @@
...
 /**
  * Class provides per field format implementation for Lucene Knn vector type
  */
 public class KNN9120PerFieldKnnVectorsFormat extends BasePerFieldKnnVectorsFormat {
     private static final Tuple<Integer, ExecutorService> DEFAULT_MERGE_THREAD_COUNT_AND_EXECUTOR_SERVICE = Tuple.tuple(1, null);
+    private static final ThreadPoolExecutor mergeExecutor;
+    private static volatile int mergeThreadCount;
+
+    static {
+        mergeExecutor = new ThreadPoolExecutor(0, 1, 60L, TimeUnit.SECONDS, new LinkedBlockingQueue<>());
```
shatejas: This fundamentally changes the semantics of index thread quantity setting for lucene engine. Till now setting this was giving `index_thrd_quantity` threads per merge. This moves the setting to `index_thrd_quantity` per shard. This can create a backlog of merges creating a bottleneck or slowing them down depending on how lucene uses. Is my understanding not correct here? Am I missing something  where merges won't be a bottleneck?

src/main/java/org/opensearch/knn/index/codec/backward_codecs/KNN9120Codec/KNN9120PerFieldKnnVectorsFormat.java
```diff
@@ -9,21 +9,47 @@
...
+
+    /**
+     * Called by KNNSettings when knn.algo_param.index_thread_qty changes at runtime.
+     * Resizes the shared merge executor pool without creating a new instance.
+     */
+    public static void updateMergeThreadCount(int newCount) {
+        mergeThreadCount = newCount;
+        int targetCore = Math.max(newCount, 0);
+        int targetMax = Math.max(newCount, 1);
+        // Always adjust in the safe order to avoid core > max invariant violation
```
shatejas, line 55: > Unbounded queues. Using an unbounded queue (for example a [LinkedBlockingQueue](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/LinkedBlockingQueue.html) without a predefined capacity) will cause new tasks to wait in the queue when all corePoolSize threads are busy. Thus, no more than corePoolSize threads will ever be created. (And the value of the maximumPoolSize therefore doesn't have any effect.) This may be appropriate when each task is completely independent of others, so tasks cannot affect each others execution; for example, in a web page server. While this style of queuing can be useful in smoothing out transient bursts of requests, it admits the possibility of unbounded work queue growth when commands continue to arrive on average faster than they can be processed.

With LinkedBlockingQueue without a bound I don't think maximumPoolSize takes into affect.

src/main/java/org/opensearch/knn/index/codec/KNN9120Codec/KNN9120PerFieldKnnVectorsFormat.java
Vikasht34: Good Point !!

src/main/java/org/opensearch/knn/index/codec/backward_codecs/KNN9120Codec/KNN9120PerFieldKnnVectorsFormat.java
```diff
@@ -9,21 +9,50 @@
...
+        if (targetMax >= mergeExecutor.getMaximumPoolSize()) {
+            mergeExecutor.setMaximumPoolSize(targetMax);
+            mergeExecutor.setCorePoolSize(targetCore);
+        } else {
+            mergeExecutor.setCorePoolSize(targetCore);
+            mergeExecutor.setMaximumPoolSize(targetMax);
+        }
```
vamshin, line 62: Does the ordering of these statements matter?Otherwise I see same logic inside **if** and **else**.

MrFlap: Yes. ThreadPoolExecutor throws IllegalArgumentException if corePoolSize > maximumPoolSize.

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/KNN1040PerFieldKnnVectorsFormat.java
```diff
@@ -40,6 +41,41 @@
...
+    private static volatile int mergeThreadCount;
+
+    static {
+        // Use SynchronousQueue so maximumPoolSize governs actual thread creation (matching Lucene's
+        // ConcurrentMergeScheduler.CachedExecutor pattern). CallerRunsPolicy provides graceful
+        // degradation if a resize races with an in-flight merge submission.
+        mergeExecutor = new ThreadPoolExecutor(
+            0,
+            1,
+            60L,
```
navneet1v, line 54: can we create constants for these things in the class so that we know what these parameters mean.

src/main/java/org/opensearch/knn/index/codec/KNN9120Codec/KNN9120PerFieldKnnVectorsFormat.java
navneet1v: @shatejas I don't think this is now at shard level. Since the code is in static block, the threadpool is now at the node level. So if we have more shards on one node all of them will share the same pool.

PR #3124, opened 2026-02-20
[BugFix] Fix lucene reduce to topK when rescoring is enabled

naveentatikonda:
### Description
This PR addresses two bugs related to k-NN rescoring functionality:

#### Bug 1: Premature Result Reduction with Rescoring Enabled
A regression was introduced in [PR #3037](https://github.com/opensearch-project/k-NN/pull/3037) where oversampled k results are incorrectly reduced to the actual k value before the rescoring phase, even when rescoring is enabled. This defeats the purpose of oversampling, as rescoring should operate on the full oversampled result set to improve accuracy.

**Fix:** Ensure that result reduction occurs after rescoring completes.

#### Bug 2: ExpandNested Incompatibility with Rescoring
An existing bug before [PR #3037](https://github.com/opensearch-project/k-NN/pull/3037) affects the interaction between expandNested and rescoring. When both features are enabled, the RescoreKNNVectorQuery wraps the NestedKnnQuery (similar to other Lucene queries). However, ExpandNestedDocsQuery fetches all child documents before rescoring, and these are then reduced to k results after rescoring—preventing all relevant child documents from being returned.

Created this issue https://github.com/opensearch-project/k-NN/issues/3125 for tracking

**Temporary Solution:** Rescoring is now disabled when expandNested is enabled to prevent incorrect results.

**Future Work:** A TODO has been added to properly support rescoring with expandNested, which will require additional refactoring to do rescoring and reducing them to top K before fetching child documents for all those k results.

### Related Issues
https://github.com/opensearch-project/k-NN/issues/2940

### Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 * Fix score conversion logic for radial exact search [#3110](https://github.com/opensearch-project/k-NN/pull/3110)
 * Fix lucene reduce to topK when rescoring is enabled [#3124](https://github.com/opensearch-project/k-NN/pull/3124)
 * Fix bugs in optimistic search for nested Cagra index [#3155](https://github.com/opensearch-project/k-NN/pull/3155)
+* Fix integer overflow for memory optimized search [#3130](https://github.com/opensearch-project/k-NN/pull/3130)
 
 ### Refactoring
 
--- a/build.gradle
+++ b/build.gradle
@@ -501,7 +501,7 @@ dependencies {
     implementation "org.slf4j:slf4j-api:${versions.slf4j}"
     zipArchive group: 'org.opensearch.plugin', name:'opensearch-security', version: "${opensearch_build}"
     // Commenting this out till CustomCodec plugin is ready after Lucene 10.4.0 upgrade in OS Core.
-    //zipArchive group: 'org.opensearch.plugin', name:'opensearch-custom-codecs', version: "${opensearch_build}"
+    zipArchive group: 'org.opensearch.plugin', name:'opensearch-custom-codecs', version: "${opensearch_build}"
 }
 
 task windowsPatches(type:Exec) {
@@ -654,8 +654,6 @@ integTest {
     commonIntegTest(it, project, integTestDependOnJniLib, opensearch_tmp_dir, _numNodes)
     filter {
         excludeTestsMatching "org.opensearch.knn.index.RelocationIT"
-        // Excluding till custom codec plugin upgrades
-        excludeTestsMatching "org.opensearch.knn.integ.codecs.CustomCodecsIT"
     }
     mustRunAfter test
 }
@@ -666,8 +664,6 @@ task integTestRemoteIndexBuild(type: RestIntegTestTask) {
         // Temporarily skipping this test class, see: https://github.com/opensearch-project/k-NN/issues/2726
         excludeTestsMatching "org.opensearch.knn.index.SegmentReplicationIT"
         excludeTestsMatching "org.opensearch.knn.index.RelocationIT"
-        // Excluding till custom codec plugin upgrades
-        excludeTestsMatching "org.opensearch.knn.integ.codecs.CustomCodecsIT"
     }
     systemProperty("test.remoteBuild", System.getProperty("test.remoteBuild"))
     systemProperty("test.bucket", System.getProperty("test.bucket"))
@@ -694,7 +690,7 @@ def commonIntegTestClusters(OpenSearchCluster cluster, _numNodes){
     if (System.getProperty("security.enabled") != null) {
         configureSecurityPlugin(cluster)
     }
-    /*else {
+    else {
         // Only install custom codecs plugin
         configurations.zipArchive.asFileTree.filter(f -> f.name.contains("custom-codecs")).each {
             cluster.plugin(provider(new Callable<RegularFile>() {
@@ -709,7 +705,7 @@ def commonIntegTestClusters(OpenSearchCluster cluster, _numNodes){
             }
           }))
        }
-    }*/
+    }
 
     cluster.plugin(project.tasks.bundlePlugin.archiveFile)
     if (Os.isFamily(Os.FAMILY_WINDOWS)) {
--- a/src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissHNSW.java
+++ b/src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissHNSW.java
@@ -74,7 +74,7 @@ public class FaissHNSW {
 
         // Load `offsets` into memory.
         size = input.readLong();
-        offsetsReader = MonotonicIntegerSequenceEncoder.encode(Math.toIntExact(size), input);
+        offsetsReader = MonotonicIntegerSequenceEncoder.encode(Math.toIntExact(size), input, false);
         Objects.requireNonNull(offsetsReader);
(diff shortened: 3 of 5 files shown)


src/main/java/org/opensearch/knn/index/query/lucenelib/NestedKnnVectorQueryFactory.java
```diff
@@ -67,13 +68,23 @@ public static Query createNestedKnnVectorQuery(
...
         final BitSetProducer parentFilter,
         final boolean expandNestedDocs,
-        final int k
+        final int k,
+        final boolean needsRescore
     ) {
         if (expandNestedDocs) {
             return new ExpandNestedDocsQuery.ExpandNestedDocsQueryBuilder().internalNestedKnnVectorQuery(
-                new InternalNestedKnnFloatVectorQuery(fieldName, vector, filterQuery, luceneK, parentFilter, k)
+                new InternalNestedKnnFloatVectorQuery(
```
kotwanikunal: Do we need to apply the same fix to byte based indexes?

src/main/java/org/opensearch/knn/index/query/KNNQueryFactory.java
```diff
@@ -138,9 +138,25 @@ public static Query create(CreateQueryRequest createQueryRequest) {
...
+                needsRescore
+            )
         );
-        return needsRescore ? new RescoreKNNVectorQuery(luceneKnnQuery, fieldName, k, vector, shardId) : luceneKnnQuery;
+
+        // Skip wrapping with RescoreKNNVectorQuery for expandNested as it will reduce the results to k after rescoring
+        // which will not return all the child documents
+        // TODO: Skip retrieving child docs in ExpandNestedDocsQuery if rescoring is enabled and instead retrieve them after rescoring and
+        // reducing to top K.
+        return needsRescore && !expandNested ? new RescoreKNNVectorQuery(luceneKnnQuery, fieldName, k, vector, shardId) : luceneKnnQuery;
```
kotwanikunal: We should warn or document this behavior with explain to ensure the users understand these compatibility issues.

kotwanikunal: Can you also add in ITs for this scenario?

src/main/java/org/opensearch/knn/index/query/lucenelib/InternalNestedKnnFloatVectorQuery.java
```diff
@@ -50,7 +52,9 @@ public InternalNestedKnnFloatVectorQuery(
             filter,
             luceneK,
             parentFilter,
-            k
+            k,
+            needsRescore,
+            expandNestedDocs
         );
```
Vikasht34, line 57: Right now this class already takes aprox 10 params , adding two another seems bit overloaded, Can we reduce this , class Name is Nested so expandNested Docs should not be another cosntructor param, rather than functional params and same as rescore , these should be determined at run time I feel rather than while creating the class .Let me know what do u think?

src/main/java/org/opensearch/knn/index/query/lucenelib/NestedKnnVectorQueryFactory.java
```diff
@@ -67,13 +68,23 @@ public static Query createNestedKnnVectorQuery(
...
+                    luceneK,
+                    parentFilter,
+                    k,
+                    needsRescore,
+                    expandNestedDocs
+                )
             ).queryUtils(QueryUtils.getInstance()).build();
         }
-        return new OSDiversifyingChildrenFloatKnnVectorQuery(fieldName, vector, filterQuery, luceneK, parentFilter, k);
+        return new OSDiversifyingChildrenFloatKnnVectorQuery(fieldName, vector, filterQuery, luceneK, parentFilter, k, needsRescore);
```
Vikasht34, line 79: As I see this calss takes expandNestedDocs also as param , but I am not seeing it's getting passed when we are crreating the class,?

src/main/java/org/opensearch/knn/index/query/lucenelib/NestedKnnVectorQueryFactory.java
naveentatikonda: No, we need rescoring only for quantization. Quantization only works when the vector datatype is float. So, we don't need it for byte index.

src/main/java/org/opensearch/knn/index/query/lucenelib/NestedKnnVectorQueryFactory.java
naveentatikonda: Yes, I created another constructor in OSDiversifyingChildrenFloatKnnVectorQuery without expandNestedDocs param

src/main/java/org/opensearch/knn/index/query/lucenelib/InternalNestedKnnFloatVectorQuery.java
naveentatikonda: It makes sense to skip passing the expandNestedDocs param to InternalNestedKnnFloatVectorQuery class because this class will only be called in case of expandNested but we still need to pass needsRescore param.

src/test/java/org/opensearch/knn/integ/NestedSearchIT.java
```diff
@@ -148,6 +148,31 @@ public void testNestedSearchWithFaiss_whenRescoreEnabled_thenSucceed() {
         assertEquals("13", parseIds(entity).get(1));
     }
 
+    @SneakyThrows
+    public void testNestedSearchWithLuceneOnDisk_whenExpandNestedAndRescoreEnabled_thenSucceed() {
```
kotwanikunal: Shouldnt this test exist in `ExpandNestedDocsIT`?
I dont see the expand nested initialization in this class similar to https://github.com/opensearch-project/k-NN/blob/main/src/test/java/org/opensearch/knn/integ/ExpandNestedDocsIT.java#L506

Vikasht34 (approved):
I am good with change , thanks ...

src/test/java/org/opensearch/knn/integ/NestedSearchIT.java
naveentatikonda: Good catch. I removed the test here and added config to run it in ExpandNestedDocsIT

src/main/java/org/opensearch/knn/index/query/KNNQueryFactory.java
```diff
@@ -175,12 +195,13 @@ private static Query getKnnVectorQuery(
...
-        final int k
+        final int k,
+        final boolean needsRescore
     ) {
         if (parentFilter == null) {
             assert expandNested == false : "expandNested is allowed to be true only for nested fields.";
             return vectorDataType == VectorDataType.FLOAT
-                ? new OSKnnFloatVectorQuery(fieldName, floatQueryVector, luceneK, filterQuery, k)
+                ? new OSKnnFloatVectorQuery(fieldName, floatQueryVector, luceneK, filterQuery, k, needsRescore)
                 : new OSKnnByteVectorQuery(fieldName, byteQueryVector, luceneK, filterQuery, k);
```
0ctopus13prime, line 205: Curious, is there a reason not passing `needsRescore` to OSKnnByteVectorQuery?

naveentatikonda: As of now, Lucene does rescoring only for 4x compression and for that the vector data type should be float and we doesn't support any compression levels on byte vector data type. So, we don't need it for byte

PR #3139, opened 2026-03-01
[Backport 3.5] Simplify DerivedSourceReaders lifecycle by removing manual ref-counting

shatejas:
https://github.com/opensearch-project/k-NN/pull/3138

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/codec/KNN10010Codec/KNN10010DerivedSourceStoredFieldsReader.java
+++ b/src/main/java/org/opensearch/knn/index/codec/KNN10010Codec/KNN10010DerivedSourceStoredFieldsReader.java
@@ -99,7 +99,7 @@ public class KNN10010DerivedSourceStoredFieldsReader extends StoredFieldsReader 
             return new KNN10010DerivedSourceStoredFieldsReader(
                 delegate.clone(),
                 derivedVectorFields,
-                derivedSourceReaders.cloneWithMerge(),
+                derivedSourceReaders.clone(),
                 segmentReadState,
                 shouldInject
             );
@@ -130,7 +130,7 @@ public class KNN10010DerivedSourceStoredFieldsReader extends StoredFieldsReader 
             return new KNN10010DerivedSourceStoredFieldsReader(
                 delegate.getMergeInstance(),
                 derivedVectorFields,
-                derivedSourceReaders.cloneWithMerge(),
+                derivedSourceReaders.getMergeInstance(),
                 segmentReadState,
                 false
             );
--- a/src/main/java/org/opensearch/knn/index/codec/derivedsource/DerivedSourceReaders.java
+++ b/src/main/java/org/opensearch/knn/index/codec/derivedsource/DerivedSourceReaders.java
@@ -6,70 +6,97 @@
 package org.opensearch.knn.index.codec.derivedsource;
 
 import lombok.Getter;
-import lombok.RequiredArgsConstructor;
 import org.apache.lucene.codecs.DocValuesProducer;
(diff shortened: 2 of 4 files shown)


navneet1v (dismissed):

src/main/java/org/opensearch/knn/index/codec/derivedsource/DerivedSourceReaders.java
```diff
@@ -6,70 +6,97 @@
...
-    private void decRef() throws IOException {
-        if (ref.decrementAndGet() == 0) {
-            IOUtils.close(knnVectorsReader, docValuesProducer);
-        }
+    /**
+     * Closes the underlying readers if this is an owning instance. No-op for cloned or merge instances.
+     */
+    @Override
+    public void close() throws IOException {
+        onClose.close();
```
navneet1v: should we check if close is already called then subsequent calls of close shouldn't do anything

shatejas: done

navneet1v (dismissed):

src/main/java/org/opensearch/knn/index/codec/derivedsource/DerivedSourceReaders.java
```diff
@@ -15,61 +14,107 @@
...
+     * Creates an owning instance. Closing this instance will close both underlying readers.
+     *
+     * @param knnVectorsReader  reader for raw vector values; may be {@code null}.
+     * @param docValuesProducer reader for doc-values; may be {@code null}.
+     */
+    public DerivedSourceReaders(KnnVectorsReader knnVectorsReader, DocValuesProducer docValuesProducer) {
+        assert knnVectorsReader != null || docValuesProducer != null : "At least one reader must be non-null";
+        this.knnVectorsReader = knnVectorsReader;
+        this.docValuesProducer = docValuesProducer;
+        this.onClose = () -> IOUtils.close(knnVectorsReader, docValuesProducer);
```
sohami: should just call `this(knnVectorsReader, docValuesProducer, () -> IOUtils.close(knnVectorsReader, docValuesProducer))`

src/main/java/org/opensearch/knn/index/codec/KNN10010Codec/KNN10010DerivedSourceStoredFieldsReader.java
sohami, line 97: I am thinking to be consistent with delegate reader (LuceneXXCompressionStoredFieldsReader) implementation, probably this class should also implement the `ensureOpen` which can be checked for both `clone` and `getMergeInstance | cloneForMerge` on this reader. I understand that underlying readers will throw if original instance of this reader is closed and later `clone` is called on it, but same is true for delegate implementation, so not sure why this pattern is followed in delegate.

shatejas: We can take this as a part of separate PR, Would want to keep this one only related to the bug

I think there are some minor improvements in `KNN10010DerivedSourceStoredFieldsReader.java` that are needed. For instance, initalizing transformer is currently not thread safe either. Will take both in separate PR

src/main/java/org/opensearch/knn/index/codec/derivedsource/DerivedSourceReaders.java
shatejas: Done

sohami (approved):

vamshin (approved):

VijayanB (approved):

PR #3140, opened 2026-03-03
[BUGFIX]Catch Merge IOException

luyuncheng:
### Description
#2529 we made faiss engine can abort.  so we catch some exceptions. 

## BUGFIX1: Exception
but when there is some runtime exception in native io, like disk io exception, other runtime exceptions. it should be catch by `Lucene#ConcurrentMergeScheduler`

https://github.com/opensearch-project/k-NN/blob/7f19fa7dc38e6315957fa99a1b95fbc543206381/src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexWriter.java#L115-L136

we need made runtime exception throw to lucene level which can be handled by `ConcurrentMergeScheduler#handleMergeException` in NativeIndexWriter.java#L134

## BUGFIX2: permission Access
In MergeAbortChecker we changed MergeThread accessibility in code: https://github.com/opensearch-project/k-NN/blob/7f19fa7dc38e6315957fa99a1b95fbc543206381/src/main/java/org/apache/lucene/index/MergeAbortChecker.java#L33-L42

so we need add `suppressAccessChecks` and `accessDeclaredMemberspermission` into plugin-security.policy

### Related Issues

### Check List

### DIFF
--- a/jni/src/org_opensearch_knn_jni_FaissService.cpp
+++ b/jni/src/org_opensearch_knn_jni_FaissService.cpp
@@ -417,9 +417,6 @@ JNIEXPORT jobjectArray JNICALL Java_org_opensearch_knn_jni_FaissService_queryInd
         return knn_jni::faiss_wrapper::QueryIndex(&jniUtil, env, indexPointerJ, queryVectorJ, kJ, methodParamsJ, parentIdsJ);
 
     }
-    catch (const faiss::FaissException& e) {
-        std::cout << "====> query get faiss exception:" << e.what() << "\n";
-    }
     catch (...) {
         jniUtil.CatchCppExceptionAndThrowJava(env);
     }
--- a/src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexWriter.java
+++ b/src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexWriter.java
@@ -166,7 +166,6 @@ public class NativeIndexWriter {
                 knnVectorValuesSupplier.get(),
                 nativeIndexParams
             );
-            System.out.println(indexBuilder.getClass());
             indexBuilder.buildAndWriteIndex(nativeIndexParams);
             CodecUtil.writeFooter(output);
         }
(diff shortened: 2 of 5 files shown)


src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexWriter.java
```diff
@@ -132,6 +132,7 @@ public void mergeIndex(final Supplier<KNNVectorValues<?>> knnVectorValuesSupplie
             throw new MergePolicy.MergeAbortedException("KNN Merge aborted.");
         } catch (Exception ex) {
             log.error("Merge exception happened for field {}", fieldInfo.name, ex);
+            throw new IOException("Merge exception happened for field: " + fieldInfo.name, ex);
```
shatejas, line 135: Its not a good idea to catch a generic exception and throw IOException. Can we catch a IOException and rethrow?

jni/src/org_opensearch_knn_jni_FaissService.cpp
```diff
@@ -417,9 +417,6 @@ JNIEXPORT jobjectArray JNICALL Java_org_opensearch_knn_jni_FaissService_queryInd
         return knn_jni::faiss_wrapper::QueryIndex(&jniUtil, env, indexPointerJ, queryVectorJ, kJ, methodParamsJ, parentIdsJ);
 
     }
-    catch (const faiss::FaissException& e) {
-        std::cout << "====> query get faiss exception:" << e.what() << "\n";
-    }
```
shatejas, line 422: The change seems right but out out curiosity was the query returning empty response when faiss exception was thrown?

src/main/plugin-metadata/plugin-security.policy
```diff
@@ -11,5 +11,6 @@ grant {
     permission java.lang.RuntimePermission "loadLibrary.opensearchknn_faiss_avx512_spr";
     permission java.net.SocketPermission "*", "connect,resolve";
     permission java.lang.RuntimePermission "accessDeclaredMembers";
+    permission java.lang.reflect.ReflectPermission "suppressAccessChecks";
```
shatejas, line 14: Can you explain how does it affect/what errors are seen if this is not present, was just curious?

PR #3144, opened 2026-03-05
1 bit compression support for the Lucene Scalar Quantizer

### LINKED ISSUE #2805: [FEATURE] [RFC] Integrating Lucene's Better Binary Quantization into OpenSearch

jack-hung-lgtm:
### Description
Integrates Lucene's BBQ and makes it the default when possible. Building off of this PR from @adityamachiroutu: https://github.com/opensearch-project/k-NN/pull/2838

Users can use the new quantizer for 32x compression by specifying bits = 1 in the lucene sq encoder, or by choosing the lucene engine with on_disk mode without specifying an encoder (the code will infer 32x compression by default).

```diff
 {
   "settings": {
     "index": {
       "knn": true
     }
   },
   "mappings": {
     "properties": {
       "my_vector1": {
         "type": "knn_vector",
         "dimension": 2,
         "space_type": "l2",
         "method": {
           "name": "hnsw",
           "engine": "lucene",
           "parameters": {
             "encoder": {
               "name": "sq"
 .             "parameters": {
+                   "bits": 1
                 }   
             },
             "ef_construction": 256,
             "m": 8
           }
         }
       }
     }
   }
 }
```

### Related Issues
Resolves #2805

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/engine/faiss/FaissSQEncoder.java
+++ b/src/main/java/org/opensearch/knn/index/engine/faiss/FaissSQEncoder.java
@@ -141,6 +141,11 @@ public class FaissSQEncoder implements Encoder {
         return CompressionLevel.x2;
     }
 
+    // TODO: The Encoder interface's validateEncoderConfig uses TrainingConfigValidation* types that were
+    // designed for model training. We should add a general-purpose validation method to the Encoder interface
+    // (e.g. Encoder.validate(KNNMethodContext, KNNMethodConfigContext)) that both Faiss and Lucene resolvers
+    // can delegate to, decoupled from training concerns. See LuceneHNSWMethodResolver.validateEncoderParams()
+    // for the Lucene equivalent that avoids the training interface.
     @Override
     public TrainingConfigValidationOutput validateEncoderConfig(TrainingConfigValidationInput validationInput) {
         TrainingConfigValidationOutput.TrainingConfigValidationOutputBuilder builder = TrainingConfigValidationOutput.builder();
@@ -163,6 +168,7 @@ public class FaissSQEncoder implements Encoder {
         boolean isV360OrLater = version != null && version.onOrAfter(Version.V_3_6_0);
         Object bitsObj = encoderParams.get(SQ_BITS);
         boolean hasType = encoderParams.containsKey(FAISS_SQ_TYPE);
+        boolean hasClip = encoderParams.containsKey(FAISS_SQ_CLIP);
 
         // On 3.6.0+, bits is required when the user explicitly specifies the sq encoder for FLOAT data
         if (isV360OrLater && bitsObj == null && configContext.getVectorDataType() == VectorDataType.FLOAT) {
@@ -183,21 +189,40 @@ public class FaissSQEncoder implements Encoder {
         if (bitsObj instanceof Integer) {
             int bits = (Integer) bitsObj;
 
-            // bits=1 does not support the type parameter
-            if (bits == Bits.ONE.getValue() && hasType) {
-                return builder.valid(false)
-                    .errorMessage(
-                        String.format(
-                            Locale.ROOT,
-                            "Parameter [%s] is not supported when [%s=%d] for encoder [%s]. "
-                                + "The 1-bit scalar quantization path does not use the type parameter.",
-                            FAISS_SQ_TYPE,
-                            SQ_BITS,
-                            bits,
-                            ENCODER_SQ
+            // type and clip is only applicable for fp16 (bits=16)
+            if (Bits.SIXTEEN.getValue() != bits) {
+                if (hasType) {
+                    return builder.valid(false)
+                        .errorMessage(
+                            String.format(
+                                Locale.ROOT,
+                                "Parameter [%s] is not supported when [%s=%d] for encoder [%s]. "
+                                    + "The type parameter is only applicable for fp16 quantization (bits=16).",
+                                FAISS_SQ_TYPE,
+                                SQ_BITS,
+                                bits,
+                                ENCODER_SQ
+                            )
                         )
-                    )
-                    .build();
+                        .build();
+                }
+
+                if (hasClip) {
+                    return builder.valid(false)
+                        .errorMessage(
+                            String.format(
+                                Locale.ROOT,
+                                "Parameter [%s] is not supported when [%s=%d] for encoder [%s]. "
+                                    + "Clipping is only applicable for fp16 quantization (bits=16).",
+                                FAISS_SQ_CLIP,
+                                SQ_BITS,
+                                bits,
+                                ENCODER_SQ
+                            )
+                        )
+                        .build();
+                }
+
             }
 
             // Validate compression level compatibility if explicitly set
--- a/src/test/java/org/opensearch/knn/common/FieldInfoExtractorTests.java
+++ b/src/test/java/org/opensearch/knn/common/FieldInfoExtractorTests.java
@@ -13,6 +13,7 @@ import org.mockito.MockedStatic;
 import org.mockito.Mockito;
 import org.opensearch.knn.KNNTestCase;
 import org.opensearch.knn.index.VectorDataType;
+import org.opensearch.knn.index.engine.faiss.SQConfig;
 import org.opensearch.knn.indices.ModelMetadata;
 import org.opensearch.knn.indices.ModelUtil;
 
@@ -107,5 +108,30 @@ public class FieldInfoExtractorTests extends KNNTestCase {
         Mockito.when(fieldInfos.fieldInfo("valid")).thenReturn(fieldInfo);
         Assert.assertNull(FieldInfoExtractor.getFieldInfo(leafReader, "invalid"));
         Assert.assertEquals(fieldInfo, FieldInfoExtractor.getFieldInfo(leafReader, "valid"));
+    }
+
+    public void testIsSQField_whenAttributePresent_thenTrue() {
+        FieldInfo fieldInfo = Mockito.mock(FieldInfo.class);
+        Mockito.when(fieldInfo.getAttribute(KNNConstants.SQ_CONFIG)).thenReturn("bits=1");
+        Assert.assertTrue(FieldInfoExtractor.isSQField(fieldInfo));
+    }
+
+    public void testIsSQField_whenAttributeAbsent_thenFalse() {
+        FieldInfo fieldInfo = Mockito.mock(FieldInfo.class);
+        Mockito.when(fieldInfo.getAttribute(KNNConstants.SQ_CONFIG)).thenReturn(null);
+        Assert.assertFalse(FieldInfoExtractor.isSQField(fieldInfo));
+    }
+
+    public void testExtractSQConfig_whenPresent_thenReturnConfig() {
+        FieldInfo fieldInfo = Mockito.mock(FieldInfo.class);
+        Mockito.when(fieldInfo.getAttribute(KNNConstants.SQ_CONFIG)).thenReturn("bits=1");
+        SQConfig config = FieldInfoExtractor.extractSQConfig(fieldInfo);
+        Assert.assertEquals(1, config.getBits());
+    }
+
+    public void testExtractSQConfig_whenAbsent_thenReturnEmpty() {
+        FieldInfo fieldInfo = Mockito.mock(FieldInfo.class);
+        Mockito.when(fieldInfo.getAttribute(KNNConstants.SQ_CONFIG)).thenReturn(null);
+        Assert.assertSame(SQConfig.EMPTY, FieldInfoExtractor.extractSQConfig(fieldInfo));
     }
 }
--- a/src/test/java/org/opensearch/knn/index/engine/faiss/FaissSQEncoderTests.java
+++ b/src/test/java/org/opensearch/knn/index/engine/faiss/FaissSQEncoderTests.java
@@ -24,6 +24,7 @@ import java.util.Map;
 import static org.opensearch.knn.common.KNNConstants.ENCODER_SQ;
 import static org.opensearch.knn.common.KNNConstants.FAISS_FLAT_DESCRIPTION;
 import static org.opensearch.knn.common.KNNConstants.SQ_BITS;
+import static org.opensearch.knn.common.KNNConstants.FAISS_SQ_CLIP;
 import static org.opensearch.knn.common.KNNConstants.FAISS_SQ_TYPE;
 import static org.opensearch.knn.common.KNNConstants.INDEX_DESCRIPTION_PARAMETER;
 import static org.opensearch.knn.common.KNNConstants.METHOD_ENCODER_PARAMETER;
@@ -110,6 +111,30 @@ public class FaissSQEncoderTests extends KNNTestCase {
         assertTrue(output.getErrorMessage().contains("not supported"));
     }
 
+    // --- Validation: bits=1 + clip not allowed ---
+
+    public void testValidate_whenBits1WithClip_thenError() {
+        FaissSQEncoder encoder = new FaissSQEncoder();
+        TrainingConfigValidationOutput output = encoder.validateEncoderConfig(
+            buildValidationInput(Version.CURRENT, CompressionLevel.NOT_CONFIGURED, Map.of(SQ_BITS, 1, FAISS_SQ_CLIP, true))
+        );
+        assertNotNull(output.getValid());
+        assertFalse(output.getValid());
+        assertTrue(output.getErrorMessage().contains("clip"));
+    }
+
+    public void testValidate_whenBits16WithClip_thenOk() {
+        FaissSQEncoder encoder = new FaissSQEncoder();
+        TrainingConfigValidationOutput output = encoder.validateEncoderConfig(
+            buildValidationInput(
+                Version.CURRENT,
+                CompressionLevel.NOT_CONFIGURED,
+                Map.of(SQ_BITS, 16, FAISS_SQ_TYPE, "fp16", FAISS_SQ_CLIP, true)
+            )
+        );
+        assertNull(output.getValid());
+    }
+
     // --- Validation: compression level compatibility ---
 
     public void testValidate_whenBits1WithX32Compression_thenOk() {
@@ -176,5 +201,37 @@ public class FaissSQEncoderTests extends KNNTestCase {
         );
 
         return TrainingConfigValidationInput.builder().knnMethodContext(methodContext).knnMethodConfigContext(configContext).build();
+    }
+
+    // --- isSQOneBit utility ---
+
+    public void testIsSQOneBit_whenSQWithBits1_thenTrue() {
+        MethodComponentContext encoderCtx = new MethodComponentContext(ENCODER_SQ, Map.of(SQ_BITS, 1));
+        Map<String, Object> params = Map.of(METHOD_ENCODER_PARAMETER, encoderCtx);
+        assertTrue(FaissSQEncoder.isSQOneBit(params));
+    }
+
+    public void testIsSQOneBit_whenSQWithBits16_thenFalse() {
+        MethodComponentContext encoderCtx = new MethodComponentContext(ENCODER_SQ, Map.of(SQ_BITS, 16));
+        Map<String, Object> params = Map.of(METHOD_ENCODER_PARAMETER, encoderCtx);
+        assertFalse(FaissSQEncoder.isSQOneBit(params));
+    }
+
+    public void testIsSQOneBit_whenNonSQEncoder_thenFalse() {
+        MethodComponentContext encoderCtx = new MethodComponentContext("flat", Map.of());
+        Map<String, Object> params = Map.of(METHOD_ENCODER_PARAMETER, encoderCtx);
+        assertFalse(FaissSQEncoder.isSQOneBit(params));
+    }
+
+    public void testIsSQOneBit_whenNullParams_thenFalse() {
+        assertFalse(FaissSQEncoder.isSQOneBit(null));
+    }
+
+    public void testIsSQOneBit_whenNoEncoder_thenFalse() {
+        assertFalse(FaissSQEncoder.isSQOneBit(Map.of()));
+    }
+
+    public void testIsSQOneBit_whenEncoderNotMethodComponentContext_thenFalse() {
+        assertFalse(FaissSQEncoder.isSQOneBit(Map.of(METHOD_ENCODER_PARAMETER, "not_a_context")));
     }
 }
--- a/src/test/java/org/opensearch/knn/index/engine/faiss/SQConfigParserTests.java
+++ b/src/test/java/org/opensearch/knn/index/engine/faiss/SQConfigParserTests.java
@@ -1,0 +1,61 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.knn.index.engine.faiss;
+
+import org.opensearch.knn.KNNTestCase;
+
+public class SQConfigParserTests extends KNNTestCase {
+
+    public void testToCsv_whenValidConfig_thenReturnCsv() {
+        SQConfig config = SQConfig.builder().bits(1).build();
+        assertEquals("bits=1", SQConfigParser.toCsv(config));
+    }
+
+    public void testToCsv_whenBits16_thenReturnCsv() {
+        SQConfig config = SQConfig.builder().bits(16).build();
+        assertEquals("bits=16", SQConfigParser.toCsv(config));
+    }
+
+    public void testToCsv_whenNull_thenReturnEmpty() {
+        assertEquals("", SQConfigParser.toCsv(null));
+    }
+
+    public void testToCsv_whenEmpty_thenReturnEmpty() {
+        assertEquals("", SQConfigParser.toCsv(SQConfig.EMPTY));
+    }
+
+    public void testFromCsv_whenValidCsv_thenReturnConfig() {
+        SQConfig config = SQConfigParser.fromCsv("bits=1");
+        assertEquals(1, config.getBits());
+    }
+
+    public void testFromCsv_whenBits16_thenReturnConfig() {
+        SQConfig config = SQConfigParser.fromCsv("bits=16");
+        assertEquals(16, config.getBits());
+    }
+
+    public void testFromCsv_whenNull_thenReturnEmpty() {
+        assertSame(SQConfig.EMPTY, SQConfigParser.fromCsv(null));
+    }
+
+    public void testFromCsv_whenEmptyString_thenReturnEmpty() {
+        assertSame(SQConfig.EMPTY, SQConfigParser.fromCsv(""));
+    }
+
+    public void testFromCsv_whenInvalidFormat_thenThrow() {
+        expectThrows(IllegalArgumentException.class, () -> SQConfigParser.fromCsv("invalid"));
+    }
+
+    public void testFromCsv_whenWrongKey_thenThrow() {
+        expectThrows(IllegalArgumentException.class, () -> SQConfigParser.fromCsv("wrong=1"));
+    }
+
+    public void testRoundTrip() {
+        SQConfig original = SQConfig.builder().bits(1).build();
+        SQConfig parsed = SQConfigParser.fromCsv(SQConfigParser.toCsv(original));
+        assertEquals(original, parsed);
+    }
+}
--- a/src/test/java/org/opensearch/knn/integ/FaissSQMappingIT.java
+++ b/src/test/java/org/opensearch/knn/integ/FaissSQMappingIT.java
@@ -1,0 +1,442 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.knn.integ;
+
+import org.opensearch.client.ResponseException;
+import org.opensearch.common.xcontent.XContentFactory;
+import org.opensearch.core.xcontent.XContentBuilder;
+import org.opensearch.knn.KNNRestTestCase;
+import org.opensearch.knn.index.SpaceType;
+
+import java.util.Map;
+import org.opensearch.knn.index.VectorDataType;
+import org.opensearch.knn.index.mapper.CompressionLevel;
+import org.opensearch.knn.index.mapper.Mode;
+
+import static org.opensearch.knn.common.KNNConstants.COMPRESSION_LEVEL_PARAMETER;
+import static org.opensearch.knn.common.KNNConstants.DIMENSION;
+import static org.opensearch.knn.common.KNNConstants.ENCODER_SQ;
+import static org.opensearch.knn.common.KNNConstants.FAISS_NAME;
+import static org.opensearch.knn.common.KNNConstants.FAISS_SQ_CLIP;
+import static org.opensearch.knn.common.KNNConstants.FAISS_SQ_ENCODER_FP16;
+import static org.opensearch.knn.common.KNNConstants.FAISS_SQ_TYPE;
+import static org.opensearch.knn.common.KNNConstants.KNN_ENGINE;
+import static org.opensearch.knn.common.KNNConstants.KNN_METHOD;
+import static org.opensearch.knn.common.KNNConstants.METHOD_ENCODER_PARAMETER;
+import static org.opensearch.knn.common.KNNConstants.METHOD_HNSW;
+import static org.opensearch.knn.common.KNNConstants.METHOD_IVF;
+import static org.opensearch.knn.common.KNNConstants.METHOD_PARAMETER_NLIST;
+import static org.opensearch.knn.common.KNNConstants.METHOD_PARAMETER_NPROBES;
+import static org.opensearch.knn.common.KNNConstants.METHOD_PARAMETER_SPACE_TYPE;
+import static org.opensearch.knn.common.KNNConstants.MODE_PARAMETER;
+import static org.opensearch.knn.common.KNNConstants.NAME;
+import static org.opensearch.knn.common.KNNConstants.PARAMETERS;
+import static org.opensearch.knn.common.KNNConstants.PROPERTIES;
+import static org.opensearch.knn.common.KNNConstants.SQ_BITS;
+import static org.opensearch.knn.common.KNNConstants.TYPE;
+import static org.opensearch.knn.common.KNNConstants.TYPE_KNN_VECTOR;
+import static org.opensearch.knn.common.KNNConstants.VECTOR_DATA_TYPE_FIELD;
+
+/**
+ * Integration tests for Faiss SQ encoder mapping validation.
+ * Covers bits parameter requirements, type/bits conflicts, compression compatibility,
+ * and various mapping combinations.
+ */
+public class FaissSQMappingIT extends KNNRestTestCase {
+
+    private static final int TEST_DIMENSION = 128;
+    private static final int NUM_DOCS = 10;
+    private static final int K = 5;
+
+    /**
+     * Validates a bits=16 (fp16) index
+     */
+    private void validateFP16Index(String indexName, String mapping, Integer expectedBits, String expectedType, Boolean expectedClip)
+        throws Exception {
+        createKnnIndex(indexName, mapping);
+        validateMapping(indexName, expectedBits, expectedType, expectedClip);
+        addKNNDocs(indexName, FIELD_NAME, TEST_DIMENSION, 0, NUM_DOCS);
+        validateKNNSearch(indexName, FIELD_NAME, TEST_DIMENSION, NUM_DOCS, K);
+        deleteKNNIndex(indexName);
+    }
+
+    /**
+     * Validates a bits=1 index at mapping level
+     */
+    private void validateMappingOnly(String indexName, String mapping, Integer expectedBits) throws Exception {
+        createKnnIndex(indexName, mapping);
+        validateMapping(indexName, expectedBits, null, null);
+        deleteKNNIndex(indexName);
+    }
+
+    @SuppressWarnings("unchecked")
+    private void validateMapping(String indexName, Integer expectedBits, String expectedType, Boolean expectedClip) throws Exception {
+        Map<String, Object> actualMapping = getIndexMappingAsMap(indexName);
+        assertNotNull(actualMapping);
+        Map<String, Object> actualProps = (Map<String, Object>) actualMapping.get(PROPERTIES);
+        assertNotNull("Mapping properties should not be null", actualProps);
+        Map<String, Object> actualField = (Map<String, Object>) actualProps.get(FIELD_NAME);
+        assertNotNull("Field [" + FIELD_NAME + "] should exist in mapping", actualField);
+        assertEquals(TYPE_KNN_VECTOR, actualField.get(TYPE));
+        assertEquals(TEST_DIMENSION, actualField.get(DIMENSION));
+
+        Map<String, Object> method = (Map<String, Object>) actualField.get(KNN_METHOD);
+        assertNotNull("Method should not be null", method);
+        assertEquals(FAISS_NAME, method.get(KNN_ENGINE));
+        Map<String, Object> params = (Map<String, Object>) method.get(PARAMETERS);
+        assertNotNull("Method parameters should not be null", params);
+        Map<String, Object> encoder = (Map<String, Object>) params.get(METHOD_ENCODER_PARAMETER);
+        assertNotNull("Encoder should not be null", encoder);
+        assertEquals(ENCODER_SQ, encoder.get(NAME));
+
+        Map<String, Object> encoderParams = (Map<String, Object>) encoder.get(PARAMETERS);
+        assertNotNull("Encoder parameters should not be null", encoderParams);
+        if (expectedBits != null) {
+            assertEquals(expectedBits, encoderParams.get(SQ_BITS));
+        }
+        if (expectedType != null) {
+            assertEquals(expectedType, encoderParams.get(FAISS_SQ_TYPE));
+        }
+        if (expectedClip != null) {
+            assertEquals(expectedClip, encoderParams.get(FAISS_SQ_CLIP));
+        }
+    }
+
+    /**
+     * Validates that index creation fails with a ResponseException containing the expected message.
+     */
+    private void validateMappingFails(String mapping, String expectedError) {
+        ResponseException ex = expectThrows(ResponseException.class, () -> createKnnIndex(INDEX_NAME, mapping));
+        assertTrue(
+            "Expected error containing [" + expectedError + "] but got: " + ex.getMessage(),
+            ex.getMessage().contains(expectedError)
+        );
+    }
+
+    // --- bits required on 3.6.0+ ---
+
+    public void testSQEncoder_whenNoBitsOnCurrentVersion_thenFail() throws Exception {
+        String mapping = buildSQMapping(null, FAISS_SQ_ENCODER_FP16, null, null, null);
+        validateMappingFails(mapping, "bits");
+    }
+
+    // --- bits=16 with type=fp16 (valid) ---
+
+    public void testSQEncoder_whenBits16WithTypeFP16_thenSucceed() throws Exception {
+        String mapping = buildSQMapping(16, FAISS_SQ_ENCODER_FP16, null, null, null);
+        validateFP16Index(INDEX_NAME, mapping, 16, FAISS_SQ_ENCODER_FP16, null);
+    }
+
+    public void testSQEncoder_whenBits16WithTypeFP16AndCompressionX2_thenSucceed() throws Exception {
+        String mapping = buildSQMapping(16, FAISS_SQ_ENCODER_FP16, CompressionLevel.x2.getName(), null, null);
+        validateFP16Index(INDEX_NAME, mapping, 16, FAISS_SQ_ENCODER_FP16, null);
+    }
+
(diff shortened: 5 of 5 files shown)


qa/rolling-upgrade/build.gradle
```diff
@@ -8,6 +8,11 @@ import org.apache.tools.ant.taskdefs.condition.Os
 
 apply from : "$rootDir/qa/build.gradle"
 
+dependencies {
+    // Ensure Guava is available for BWC test compilation
+    testImplementation group: 'com.google.guava', name: 'guava', version:'33.2.1-jre'
```
navneet1v: use something similar to https://github.com/opensearch-project/k-NN/blob/main/build.gradle#L478

src/main/java/org/opensearch/knn/common/KNNConstants.java
```diff
@@ -104,6 +104,9 @@ public class KNNConstants {
     public static final double MAXIMUM_CONFIDENCE_INTERVAL = 1.0;
     public static final String LUCENE_SQ_BITS = "bits";
     public static final int LUCENE_SQ_DEFAULT_BITS = 7;
+    public static final String ENCODER_BBQ = "binary";
```
navneet1v: we should use encode as BBQ or LVQ. Binary can conflict with other encoder.

src/main/java/org/opensearch/knn/index/codec/params/KNNBBQVectorsFormatParams.java
```diff
@@ -0,0 +1,37 @@
...
+
+import org.opensearch.knn.index.engine.MethodComponentContext;
+import java.util.Map;
+import static org.opensearch.knn.common.KNNConstants.ENCODER_BBQ;
+import static org.opensearch.knn.common.KNNConstants.METHOD_ENCODER_PARAMETER;
+
+/**
+ * Class provides params for Lucene102HnswBinaryQuantizedVectorsFormat
+ */
+public class KNNBBQVectorsFormatParams extends KNNVectorsFormatParams {
```
navneet1v: Probably rename this to `KNNScalerQuantizerVectorsFormatParams`

src/main/java/org/opensearch/knn/index/codec/BasePerFieldKnnVectorsFormat.java
```diff
@@ -129,6 +169,21 @@ public KnnVectorsFormat getKnnVectorsFormatForField(final String field) {
                     return scalarQuantizedVectorsFormatSupplier.apply(knnScalarQuantizedVectorsFormatParams);
                 }
             }
+            // BBQ encoder should be implied for on_disk mode or compression level 32x
+            if (knnMappingConfig.getIndexCreatedVersion().onOrAfter(Version.V_3_6_0)
+                && ((knnMappingConfig.getCompressionLevel() == CompressionLevel.NOT_CONFIGURED
+                    && knnMappingConfig.getMode() == Mode.ON_DISK) || knnMappingConfig.getCompressionLevel() == CompressionLevel.x32)) {
+                KNNBBQVectorsFormatParams bbqParams = new KNNBBQVectorsFormatParams(params, defaultMaxConnections, defaultBeamWidth);
```
navneet1v: This is weird. Why all these checks are not happening at the Mapper Layer?

src/main/resources/META-INF/services/org.opensearch.transport.grpc.proto.request.search.query.QueryBuilderProtoConverter
```diff
@@ -0,0 +1 @@
+org.opensearch.knn.grpc.proto.request.search.query.KNNQueryBuilderProtoConverter
```
navneet1v: is this new? Why we are seeing this in this PR?

src/main/java/org/opensearch/knn/index/codec/KNN9120Codec/KNN9120PerFieldKnnVectorsFormat.java
```diff
@@ -76,6 +78,16 @@ public KNN9120PerFieldKnnVectorsFormat(
                     mergeThreadCountAndExecutorService.v2()
                 );
             },
+            knnBBQVectorsFormatParams -> {
```
naveentatikonda: We need to add this in new codec KNN1040Codec

src/main/java/org/opensearch/knn/index/codec/params/KNNBBQVectorsFormatParams.java
```diff
@@ -0,0 +1,37 @@
...
+
+package org.opensearch.knn.index.codec.params;
+
+import org.opensearch.knn.index.engine.MethodComponentContext;
+import java.util.Map;
+import static org.opensearch.knn.common.KNNConstants.ENCODER_BBQ;
+import static org.opensearch.knn.common.KNNConstants.METHOD_ENCODER_PARAMETER;
+
+/**
+ * Class provides params for Lucene102HnswBinaryQuantizedVectorsFormat
```
naveentatikonda: It is Lucene104HnswScalarQuantizedVectorsFormat

src/main/java/org/opensearch/knn/index/mapper/CompressionLevel.java
```diff
@@ -124,12 +136,22 @@ public RescoreContext getDefaultRescoreContext(Mode mode, int dimension, Version
...
-    }
 
-    @VisibleForTesting
-    RescoreContext getDefaultRescoreContext(Mode mode, int dimension) {
-        return getDefaultRescoreContext(mode, dimension, Version.CURRENT);
+        // Special handling for Lucene BBQ (x32 compression)
+        if (this == x32 && engine == KNNEngine.LUCENE && version.onOrAfter(Version.V_3_6_0)) {
+            if (dimension <= RescoreContext.DIMENSION_THRESHOLD) {
+                return RescoreContext.builder()
+                    .oversampleFactor(RescoreContext.OVERSAMPLE_FACTOR_BELOW_DIMENSION_THRESHOLD)
```
naveentatikonda: I think based on our discussion after reviewing the benchmarking results we want to use a default oversampling factor of 2.0 irrespective of the dimension of the vector.

Also, why are we adding this outside of the above if condition ?
`if (modesForRescore.contains(mode)) {`

src/test/java/org/opensearch/knn/index/codec/backward_codecs/KNN990Codec/KNN990PerFieldKnnVectorsFormatTests.java
```diff
@@ -0,0 +1,60 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+
+import static org.opensearch.knn.common.KNNConstants.ENCODER_BBQ;
+import static org.opensearch.knn.common.KNNConstants.METHOD_ENCODER_PARAMETER;
+import static org.opensearch.knn.common.KNNConstants.METHOD_PARAMETER_EF_CONSTRUCTION;
+import static org.opensearch.knn.common.KNNConstants.METHOD_PARAMETER_M;
+
+public class KNN990PerFieldKnnVectorsFormatTests extends TestCase {
```
naveentatikonda: Why do we need these tests with KNN990 format ?

src/test/java/org/opensearch/knn/index/codec/KNN9120Codec/KNN9120PerFieldKnnVectorsFormatTests.java
```diff
@@ -0,0 +1,86 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+
+import static org.opensearch.knn.common.KNNConstants.ENCODER_BBQ;
+import static org.opensearch.knn.common.KNNConstants.METHOD_ENCODER_PARAMETER;
+import static org.opensearch.knn.common.KNNConstants.METHOD_PARAMETER_EF_CONSTRUCTION;
+import static org.opensearch.knn.common.KNNConstants.METHOD_PARAMETER_M;
+
+public class KNN9120PerFieldKnnVectorsFormatTests extends TestCase {
```
naveentatikonda: Pls update these tests with KNN1040 format

src/test/java/org/opensearch/knn/integ/IndexIT.java
```diff
@@ -124,6 +126,39 @@ public void testIndexWithNonVectorFields_whenValid_thenSucceed() {
...
+
+        String mapping = XContentFactory.jsonBuilder()
+            .startObject()
+            .startObject("properties")
+            .startObject(FIELD_NAME)
+            .field("type", "knn_vector")
+            .field("dimension", 128)
+            .field("mode", Mode.ON_DISK.getName())
+            .field("engine", KNNEngine.LUCENE.getName())
+            .field("compression_level", CompressionLevel.x4.getName())
```
naveentatikonda, line 146: Is this intentionally 4x ?

src/test/java/org/opensearch/knn/index/LuceneEngineIT.java
```diff
@@ -751,6 +755,173 @@ private void createKnnIndexMappingWithLuceneEngineAndSQEncoder(
...
+            .endObject()
+            .endObject()
+            .endObject()
+            .endObject()
+            .endObject();
+
+        String mapping = builder.toString();
+        createKnnIndex(INDEX_NAME, mapping);
+    }
+
```
naveentatikonda, line 930: Also, pls add an IT for nested case

src/test/java/org/opensearch/knn/index/codec/backward_codecs/KNN990Codec/KNN990PerFieldKnnVectorsFormatTests.java
jack-hung-lgtm: Yeah actually we don't need these anymore

src/test/java/org/opensearch/knn/integ/IndexIT.java
jack-hung-lgtm: Yup, just wanted to verify top level lucene works

src/main/java/org/opensearch/knn/index/codec/BasePerFieldKnnVectorsFormat.java
jack-hung-lgtm: Oh yeah thats better, I'll update that

src/test/java/org/opensearch/knn/recall/RecallTestsIT.java
```diff
@@ -668,6 +669,59 @@ public void testRecall_whenFaissHNSWPQFP32_thenRecallAbove50percent() {
...
+                .field(DIMENSION, TEST_DIMENSION)
+                .startObject(KNN_METHOD)
+                .field(METHOD_PARAMETER_SPACE_TYPE, spaceType.getValue())
+                .field(KNN_ENGINE, LUCENE_NAME)
+                .field(NAME, METHOD_HNSW)
+                .startObject(PARAMETERS)
+                .field(METHOD_PARAMETER_EF_CONSTRUCTION, HNSW_EF_CONSTRUCTION)
+                .field(METHOD_PARAMETER_M, HNSW_M)
+                .startObject(METHOD_ENCODER_PARAMETER)
+                .field(NAME, "binary")
```
kotwanikunal: This might be causing issues with your tests. Use the same name as the code and make it a constant ideally.

src/main/java/org/opensearch/knn/index/codec/params/KNN1040ScalarQuantizedVectorsFormatParams.java
```diff
@@ -0,0 +1,65 @@
...
+ */
+@Getter
+public class KNN1040ScalarQuantizedVectorsFormatParams extends KNNVectorsFormatParams {
+    private static final Set<String> SUPPORTED_ENCODERS = Set.of(ENCODER_BBQ);
+    private String encoderName;
+    private ScalarEncoding bitEncoding;
+
+    public KNN1040ScalarQuantizedVectorsFormatParams(Map<String, Object> params, int defaultMaxConnections, int defaultBeamWidth) {
+        super(params, defaultMaxConnections, defaultBeamWidth);
+        MethodComponentContext encoderMethodComponentContext = (MethodComponentContext) params.get(METHOD_ENCODER_PARAMETER);
```
kotwanikunal: Do not use a raw cast here without null checks - we've had issues in the past with BwC and defaults.
Can you instead follow the same pattern as `resolveEncoderName`?

src/main/java/org/opensearch/knn/index/codec/params/KNN1040ScalarQuantizedVectorsFormatParams.java
```diff
@@ -0,0 +1,65 @@
...
+            return null;
+        }
+
+        MethodComponentContext encoderMethodComponentContext = (MethodComponentContext) params.get(METHOD_ENCODER_PARAMETER);
+        return encoderMethodComponentContext.getName();
+    }
+
+    @Override
+    public boolean validate(final Map<String, Object> params) {
+        return encoderName != null && SUPPORTED_ENCODERS.contains(encoderName);
```
kotwanikunal: We would also need to validate the bits right? We only support 1-bit currently

qa/restart-upgrade/src/test/java/org/opensearch/knn/bwc/IndexingIT.java
```diff
@@ -674,6 +674,153 @@ public void testRandomRotationBWC() throws Exception {
...
+        waitForClusterHealthGreen(NODES_BWC_CLUSTER);
+
+        // Skip test if BBQ encoder is not supported in the old cluster version
+        if (isBBQEncoderSupported(getBWCVersion()) == false) {
+            logger.info("Skipping testKNNIndexLuceneBBQ as BBQ encoder is not supported in version: {}", getBWCVersion());
+            return;
+        }
+
+        int k = 4;
+        int dimension = 2;
```
kotwanikunal: We should use higher dimension values for tests.

src/test/java/org/opensearch/knn/recall/RecallTestsIT.java
jack-hung-lgtm: Oh yeah I found this out and pushed a fix for it, thanks for noting it.

src/main/java/org/opensearch/knn/index/codec/params/KNN1040ScalarQuantizedVectorsFormatParams.java
jack-hung-lgtm: There is a validation for the LuceneBBQEncoder class that ensures the user only provides bits = 1 with BBQ, is that enough? I have a test case that also makes sure it throws exception if the user provides an invalid number of bits for BBQ

qa/restart-upgrade/src/test/java/org/opensearch/knn/bwc/IndexingIT.java
```diff
@@ -231,24 +231,24 @@ public void testKNNIndexLuceneQuantization() throws Exception {
...
             createKnnIndex(testIndex, getKNNDefaultIndexSettings(), mapping);
 
-            Float[] vector1 = { -10.6f, 25.48f };
-            Float[] vector2 = { -10.8f, 25.48f };
-            Float[] vector3 = { -11.0f, 25.48f };
-            Float[] vector4 = { -11.2f, 25.48f };
+            Float[] vector1 = { -10.6f, 25.48f, 1.2f, 3.4f, -5.6f, 7.8f, -9.0f, 11.2f };
+            Float[] vector2 = { -10.8f, 25.48f, 1.4f, 3.6f, -5.8f, 8.0f, -9.2f, 11.4f };
+            Float[] vector3 = { -11.0f, 25.48f, 1.6f, 3.8f, -6.0f, 8.2f, -9.4f, 11.6f };
+            Float[] vector4 = { -11.2f, 25.48f, 1.8f, 4.0f, -6.2f, 8.4f, -9.6f, 11.8f };
```
navneet1v, line 237: why this change?

src/main/java/org/opensearch/knn/index/codec/LuceneVectorsFormatType.java
```diff
@@ -31,6 +31,11 @@ public enum LuceneVectorsFormatType {
      */
     SCALAR_QUANTIZED,
 
+    /**
+     * HNSW format with scalar quantization (SQ) and binary quantization encoding.
+     */
+    LUCENE_SCALAR_QUANTIZED,
```
navneet1v: what is the diff between both of them? better name here will be OptimizedScalerQuantizer this is what lucene is also using.

src/main/java/org/opensearch/knn/index/mapper/CompressionLevel.java
```diff
@@ -109,19 +110,38 @@ public static boolean isConfigured(CompressionLevel compressionLevel) {
...
             return RescoreContext.builder().oversampleFactor(FLAT_OVERSAMPLE_FACTOR).userProvided(false).build();
         }
         if (modesForRescore.contains(mode)) {
+            // Special handling for Lucene Scalar Quantized (x32 compression)
+            if (this == x32 && engine == KNNEngine.LUCENE && version.onOrAfter(Version.V_3_6_0)) {
+                return RescoreContext.builder()
+                    .oversampleFactor(RescoreContext.OVERSAMPLE_FACTOR_DEFAULT_FOR_LUCENE_SCALAR_QUANTIZED)
+                    .userProvided(false)
+                    .build();
+            }
```
navneet1v: I would suggest adding a comment here saying that engine check is temporary. once faiss also has BBQ we should just use same oversampling factor for both of them.

qa/restart-upgrade/src/test/java/org/opensearch/knn/bwc/IndexingIT.java
jack-hung-lgtm: @kotwanikunal suggested increasing dimensions for the IT

src/main/java/org/opensearch/knn/index/codec/params/KNN1040ScalarQuantizedVectorsFormatParams.java
kotwanikunal: Sounds good. Thats what I ended up doing as well. Thanks

kotwanikunal (approved):

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/KNN1040PerFieldKnnVectorsFormat.java
```diff
@@ -89,6 +91,20 @@ private static Map<LuceneVectorsFormatType, Function<KnnVectorsFormatContext, Kn
                 p.getConfidenceInterval(),
                 merge.v2()
             );
+        }, LuceneVectorsFormatType.OPTIMIZED_SCALAR_QUANTIZER, ctx -> {
```
kotwanikunal: @naveentatikonda - based on your dive and changes - would there be a way out where we can reuse the existing `LuceneVectorsFormatType` and override the format with `Lucene104HnswScalarQuantizedVectorsFormat` based on number of bits?

src/main/java/org/opensearch/knn/index/codec/LuceneVectorsFormatType.java
```diff
@@ -31,6 +31,11 @@ public enum LuceneVectorsFormatType {
      */
     SCALAR_QUANTIZED,
 
+    /**
+     * HNSW format for the optimized scalar quantizer.
+     */
+    OPTIMIZED_SCALAR_QUANTIZER,
```
navneet1v: I thought we are going to use SQ as name only. Why we are using OPTIMIZED_SCALAR_QUANTIZER in the name now?

src/main/java/org/opensearch/knn/index/engine/lucene/LuceneCodecFormatResolver.java
```diff
@@ -105,6 +106,24 @@ private LuceneVectorsFormatType determineFormatType(
                     sqParams.getBits()
                 );
                 return LuceneVectorsFormatType.SCALAR_QUANTIZED;
+            } else {
+                // Temporary route for OptimizedScalarQuantizer - eventually this should be used for SQ as well
```
navneet1v: since this change is going in main. Will it be next PR where we remove OSQ completely and just use SQ

src/main/java/org/opensearch/knn/index/engine/lucene/LuceneHNSWMethodResolver.java
```diff
@@ -105,6 +134,10 @@ private CompressionLevel getDefaultCompressionLevel(KNNMethodConfigContext knnMe
             return knnMethodConfigContext.getCompressionLevel();
         }
         if (knnMethodConfigContext.getMode() == Mode.ON_DISK) {
+            // Starting with version 3.6, supporting Optimized Scalar Quantizer by default
+            if (knnMethodConfigContext.getVersionCreated().onOrAfter(Version.V_3_6_0)) {
```
navneet1v: we should flip the check here.

```
Version.V_3_6_0.onOrBefore(knnMethodConfigContext.getVersionCreated())
```

This will avoid NPE if `knnMethodConfigContext.getVersionCreated()` is null.

src/main/java/org/opensearch/knn/index/engine/lucene/LuceneSQEncoder.java
```diff
@@ -6,32 +6,69 @@
...
 
 /**
  * Lucene scalar quantization encoder
  */
 public class LuceneSQEncoder implements Encoder {
     private static final Set<VectorDataType> SUPPORTED_DATA_TYPES = ImmutableSet.of(VectorDataType.FLOAT);
+    private static final Set<Integer> LUCENE_SQ_BITS_SUPPORTED = Arrays.stream(Bits.values())
+        .map(Bits::getValue)
+        .collect(Collectors.toUnmodifiableSet());
+    private static final Bits LUCENE_PRE_360_SUPPORTED_SQ_BITS = Bits.SEVEN;
```
navneet1v: [Suggestion]
You can move this in Bits Enum class.
```
ONE(1, CompressionLevel.x32, Version.3_6_0),
SEVEN(7, CompressionLevel.x4, <something same>);
```

where version parameter defines the min version supported for the bits.

src/main/java/org/opensearch/knn/index/engine/lucene/LuceneSQEncoder.java
```diff
@@ -53,12 +90,122 @@ public MethodComponent getMethodComponent() {
...
+        if (knnMethodConfigContext == null) {
+            return CompressionLevel.x4;
+        }
+
+        if (knnMethodConfigContext.getCompressionLevel() == CompressionLevel.NOT_CONFIGURED) {
+            if (knnMethodConfigContext.getVersionCreated().onOrAfter(Version.V_3_6_0)) {
+                return CompressionLevel.x32;
+            }
+            return CompressionLevel.x4;
+        }
```
navneet1v, line 107: lets add some comments explaining the logic.

src/test/java/org/opensearch/knn/recall/RecallTestsIT.java
```diff
@@ -668,6 +671,62 @@ public void testRecall_whenFaissHNSWPQFP32_thenRecallAbove50percent() {
...
+     *      "dimension": {TEST_DIMENSION},
+     *      "method": {
+     *          "name":"hnsw",
+     *          "engine":"lucene",
+     *          "space_type": "{SPACE_TYPE}",
+     *          "parameters":{
+     *              "m":{HNSW_M},
+     *              "ef_construction": {HNSW_EF_CONSTRUCTION},
+     *              "encoder": {
+     *                  "name": "binary",
```
navneet1v: I don't think we suppport binary in Lucene

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/KNN1040PerFieldKnnVectorsFormat.java
naveentatikonda: We can probably do something like this and reuse the existing `LuceneVectorsFormatType.SCALAR_QUANTIZED` for all the SQ bits with HNSW. We can add 2 and 4 bits later into the set when we support it. `ScalarEncoding.fromNumBits(p.getBits())` will fetch the encoding from LuceneCodecFormat based on bits. We will get rid of `Lucene99RWHnswScalarQuantizedVectorsFormat` after we remove confidence_interval

```
private static final Set<Integer> LUCENE104_SUPPORTED_BITS = Set.of(1);
LuceneVectorsFormatType.SCALAR_QUANTIZED, ctx -> {
            final KNNScalarQuantizedVectorsFormatParams p = new KNNScalarQuantizedVectorsFormatParams(
                ctx.getParams(),
                ctx.getDefaultMaxConnections(),
                ctx.getDefaultBeamWidth()
            );
            final Tuple<Integer, ExecutorService> merge = getMergeThreadCountAndExecutorService();
            if (LUCENE104_SUPPORTED_BITS.contains(p.getBits())) {
                return new Lucene104HnswScalarQuantizedVectorsFormat(
                    ScalarEncoding.fromNumBits(p.getBits()),
                    p.getMaxConnections(),
                    p.getBeamWidth(),
                    merge.v1(),
                    merge.v2()
                );
            }
            return new Lucene99RWHnswScalarQuantizedVectorsFormat(
                p.getMaxConnections(),
                p.getBeamWidth(),
                merge.v1(),
                p.getBits(),
                p.isCompressFlag(),
                p.getConfidenceInterval(),
                merge.v2()
            );
        }
```

src/main/java/org/opensearch/knn/index/codec/params/KNN1040ScalarQuantizedVectorsFormatParams.java
```diff
@@ -0,0 +1,64 @@
...
+        if (this.encoderName != null) this.initBits(encoderParams);
+    }
+
+    @Override
+    public boolean validate(final Map<String, Object> params) {
+        return this.encoderName != null && SUPPORTED_ENCODERS.contains(this.encoderName);
+    }
+
+    private ScalarEncoding getOrDefaultBitsForEncoder(final Map<String, Object> params, String bitsKey, int defaultBits) {
+        if (params != null && params.containsKey(bitsKey)) {
```
naveentatikonda: Ideally we should not run into this case, because we will throw exception if the bits is not set. Am i missing something ?

src/main/java/org/opensearch/knn/index/engine/lucene/LuceneCodecFormatResolver.java
```diff
@@ -105,6 +106,24 @@ private LuceneVectorsFormatType determineFormatType(
...
+                KNN1040ScalarQuantizedVectorsFormatParams optimizedScalarQuantizerParams = new KNN1040ScalarQuantizedVectorsFormatParams(
+                    params,
+                    defaultMaxConnections,
+                    defaultBeamWidth
+                );
+                if (optimizedScalarQuantizerParams.validate(params)) {
+                    log.debug(
+                        "Initialize KNN vector format for field [{}] with scalar/binary quantization, params [{}] = \"{}\", [{}] = \"{}\"",
```
naveentatikonda: Lucene doesn't support binary quantization, I mean in terms of UX. Pls fix the log message

src/main/java/org/opensearch/knn/index/engine/lucene/LuceneHNSWMethodResolver.java
```diff
@@ -75,14 +83,35 @@ protected void resolveEncoder(KNNMethodContext resolvedKNNMethodContext, KNNMeth
...
             knnMethodConfigContext
         );
         encoderComponentContext.getParameters().putAll(resolvedParams);
         methodComponentContext.getParameters().put(METHOD_ENCODER_PARAMETER, encoderComponentContext);
+
+        TrainingConfigValidationInput.TrainingConfigValidationInputBuilder inputBuilder = TrainingConfigValidationInput.builder();
+
+        Encoder encoder = SUPPORTED_ENCODERS.get(encoderName);
+
+        TrainingConfigValidationOutput validationOutput = encoder.validateEncoderConfig(
```
naveentatikonda: Why are we using TrainingConfigValidationInput and TrainingConfigValidationOutput. I believe we use these only for models

jack-hung-lgtm: Talked to Kunal, I think we should be good - it is a bit of a misnomer but seems to be used for all validation in faiss already

src/main/java/org/opensearch/knn/index/engine/lucene/LuceneSQEncoder.java
jack-hung-lgtm: I think I will keep it untouched for now because its nice for logging too

src/main/java/org/opensearch/knn/index/engine/lucene/LuceneSQEncoder.java
```diff
@@ -53,12 +90,131 @@ public MethodComponent getMethodComponent() {
...
         KNNMethodConfigContext knnMethodConfigContext
     ) {
-        // Hard coding to 4x for now, given thats all that is supported.
-        return CompressionLevel.x4;
+        if (knnMethodConfigContext == null) {
+            return CompressionLevel.x4;
+        }
+
+        // resolve compression level based on bits if its specified - the two must be equivalent
+        if (methodComponentContext != null && methodComponentContext.getParameters() != null) {
```
jack-hung-lgtm, line 102: Just realized this overrides user provided compression level. I will be pushing a revision that gives precedence to user given compression

src/main/java/org/opensearch/knn/index/engine/lucene/LuceneHNSWMethodResolver.java
kotwanikunal: Hey Jack!
Sorry I did not realize we only called this flow for faiss based flow by default. Looking this flow - I think it would be better to add in a new method to the Encoder interface that can help with general encoder specific validation. It makes sense even for the faiss interfaces to have a similar method. 

Can you adapt this implementation in this method instead and add a TODO for the interfacing? We can take care of it as a follow up 

```
// TODO: The Encoder interface currently only has validateEncoderConfig() which uses
//  TrainingConfigValidation* types designed for model training. We should add a general-purpose
//  validation method to the Encoder interface (e.g. Encoder.validate(KNNMethodContext, KNNMethodConfigContext))
//  that both Faiss and Lucene resolvers can delegate to, decoupled from training concerns.
//  See: FaissSQEncoder.validateEncoderConfig() for the Faiss equivalent of this validation.
private void validateEncoderParams(
    KNNMethodContext resolvedMethodContext,
    KNNMethodConfigContext configContext
) {
    if (resolvedMethodContext == null || configContext == null) {
        return;
    }

    Object encoderObj = resolvedMethodContext.getMethodComponentContext()
        .getParameters()
        .get(METHOD_ENCODER_PARAMETER);
    if (encoderObj instanceof MethodComponentContext == false) {
        return;
    }

    MethodComponentContext encoderCtx = (MethodComponentContext) encoderObj;
    if (ENCODER_SQ.equals(encoderCtx.getName()) == false) {
        return;
    }

    Map<String, Object> encoderParams = encoderCtx.getParameters();
    Object bitsObj = encoderParams.get(LUCENE_SQ_BITS);

    // On 3.6.0+, bits is required when user explicitly specifies the sq encoder
    Version version = configContext.getVersionCreated();
    if (version != null && version.onOrAfter(Version.V_3_6_0) && bitsObj == null) {
        ValidationException ex = new ValidationException();
        ex.addValidationError(
            String.format(Locale.ROOT,
                "Parameter [%s] is required for encoder [%s] on indices created with version 3.6.0 or later.",
                LUCENE_SQ_BITS, ENCODER_SQ)
        );
        throw ex;
    }

  .. other validations
}

```

src/main/java/org/opensearch/knn/index/engine/lucene/LuceneSQEncoder.java
kotwanikunal: Resolving as this is done. Thanks for updating this interface - looks much cleaner.

src/main/java/org/opensearch/knn/index/query/rescore/RescoreContext.java
```diff
@@ -22,6 +22,8 @@ public final class RescoreContext {
 
     public static final int MAX_FIRST_PASS_RESULTS = 10000;
     public static final int DIMENSION_THRESHOLD = 1000;
+
+    public static final float OVERSAMPLE_FACTOR_DEFAULT_FOR_LUCENE_SCALAR_QUANTIZER_AFTER_V360 = 2.0f;
```
kotwanikunal, line 25: nit: This will need a cleanup post faiss merges - I think we are duplicating the value across FLAT, here and faiss sq 1 bit

src/main/java/org/opensearch/knn/index/engine/lucene/LuceneHNSWMethodResolver.java
```diff
@@ -105,8 +171,117 @@ private CompressionLevel getDefaultCompressionLevel(KNNMethodConfigContext knnMe
...
         }
         return CompressionLevel.x1;
     }
+
+    // TODO: The Encoder interface currently only has validateEncoderConfig() which uses
+    // TrainingConfigValidation* types designed for model training. We should add a general-purpose
+    // validation method to the Encoder interface (e.g. Encoder.validate(KNNMethodContext, KNNMethodConfigContext))
+    // that both Faiss and Lucene resolvers can delegate to, decoupled from training concerns.
+    // See: FaissSQEncoder.validateEncoderConfig() for the Faiss equivalent of this validation.
+    protected static TrainingConfigValidationOutput validateEncoderParams(TrainingConfigValidationInput validationInput) {
```
kotwanikunal: Can we just throw an exception from here instead of using `TrainingConfigValidationOutput/TrainingConfigValidationInput`? You have all the parameters in the caller above
Lets not use the training interfaces please.

kotwanikunal (approved):

src/main/java/org/opensearch/knn/index/engine/lucene/LuceneHNSWMethodResolver.java
```diff
@@ -105,8 +159,107 @@ private CompressionLevel getDefaultCompressionLevel(KNNMethodConfigContext knnMe
             return knnMethodConfigContext.getCompressionLevel();
         }
         if (knnMethodConfigContext.getMode() == Mode.ON_DISK) {
+            // Starting with version 3.6, supporting 32x compression by default
+            if (Version.V_3_6_0.onOrBefore(knnMethodConfigContext.getVersionCreated())) {
+                return CompressionLevel.x32;
+            }
```
navneet1v, line 165: Why we are adding this check here? I thought the default compression to move from 32x will be in separate PR. @jack-hung-lgtm and @kotwanikunal

kotwanikunal: @navneet1v - For Lucene - we are moving to 32x as default instead of 4x starting 3.6.0 -  this should not be an issue.

navneet1v (approved):

PR #3152, opened 2026-03-09
Add KNNSourceExcludesProcessor to exclude vector fields from source

shatejas:
Adds a search request processor that intercepts queries containing KNNQueryBuilder and automatically adds the knn vector field names to _source excludes. User-provided includes are preserved.

### Check List


src/main/java/org/opensearch/knn/search/processor/KNNSourceExcludesProcessor.java
```diff
@@ -0,0 +1,277 @@
...
+        String description,
+        boolean ignoreFailure,
+        ClusterService clusterService,
+        IndexNameExpressionResolver indexNameExpressionResolver,
+        List<InnerHitBuilder> innerHitBuilders
+    ) {
+        super(tag, description, ignoreFailure);
+        this.clusterService = clusterService;
+        this.indexNameExpressionResolver = indexNameExpressionResolver;
+        this.innerHitBuilders = innerHitBuilders;
```
buddharajusahil: I believe the processor is a singleton class. Should we declare innerHitBuilders in the processRequest function, otherwise there could be race condition with multiple requests fighting over the same innerHitBuilders object?

build.gradle
```diff
@@ -837,6 +834,12 @@ transport.port: ${coordTransportPort}
 path.data: ${coordDir}/data
 path.logs: ${coordDir}/logs
 """
+        if (System.getProperty("coordinator.debug") != null) {
```
buddharajusahil, line 837: Is this change to build.gradle intended?

src/main/java/org/opensearch/knn/search/processor/KNNSourceExcludesProcessor.java
Vikasht34: That's true , I was about to put same commnet.

build.gradle
shatejas: yes, it was helpful for me to test this making sure its not executed on data node with a debugger

src/main/java/org/opensearch/knn/search/processor/KNNSourceExcludesProcessor.java
shatejas: I believe its per search request
- The create method in [this PR](https://github.com/opensearch-project/k-NN/pull/3152/changes#diff-110b41cd6690dd3b8df86ed74dc8bef9fa5ac0c206dbe894f2631ce015f035baR246) makes sure it creates a new object
- It creates a processor [decides whether to add processor](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/search/pipeline/SystemGeneratedPipelineWithMetrics.java#L157) and creates one[ here](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/search/pipeline/SystemGeneratedPipelineWithMetrics.java#L169)
- generateProcessors is called when the [pipeline is created](https://github.com/opensearch-project/OpenSearch/blob/e3831ffc74b9e198fa59f09a6571376b46dcb905/server/src/main/java/org/opensearch/search/pipeline/SearchPipelineService.java#L524)
- And resolvePipeline is done on [execute](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/action/search/TransportSearchAction.java#L488) of TransportSearchAction per request

Let me know if I missed something here

shatejas: figured factory is singleton and processor is not so cannot cache it

Vikasht34: Yes that's correct

src/main/java/org/opensearch/knn/search/processor/KNNSourceExcludesProcessor.java
```diff
@@ -0,0 +1,253 @@
...
+        }
+
+        final List<InnerHitBuilder> innerHitBuilders = innerHitsExtractor.getInnerHitBuilders();
+        for (InnerHitBuilder innerHitBuilder : innerHitBuilders) {
+            FetchSourceContext fetchSourceContext = innerHitBuilder.getFetchSourceContext();
+            List<String> fetchFields = innerHitBuilder.getFetchFields() != null
+                ? innerHitBuilder.getFetchFields().stream().map(fieldAndFormat -> fieldAndFormat.field).toList()
+                : List.of();
+
+            if (sourceExplicitTrue(fetchSourceContext)
```
Vikasht34: What happens in this case , Just Source True without fetch feild , as I can see sourceExplicitTrue will be true but fetchFields is empty so processor applies excludes. 

```
{
  "nested": {
    "path": "docs",
    "inner_hits": { "_source": true }
  }
}
```

src/main/java/org/opensearch/knn/search/processor/KNNSourceExcludesProcessor.java
```diff
@@ -0,0 +1,253 @@
...
+                && Stream.concat(Arrays.stream(fetchSourceContext.includes()), fetchFields.stream()).anyMatch(vectorFields::contains)) {
+                // If source is explicitly true for inner hits we should not apply excludes at top level as it will return
+                // mask 1 in inner hits. https://github.com/opensearch-project/k-NN/issues/3303
+                return request;
+            }
+
+            innerHitBuilder.setFetchSourceContext(applyExcludes(innerHitBuilder.getFetchSourceContext(), vectorFields));
```
Vikasht34: so this loop is already mutating innerHitBuilder, The loop mutates inner hit builders BEFORE checking if the next one triggers . If you have 2+ nested queries and the second one passes, the first one is already corrupted?

src/main/java/org/opensearch/knn/search/processor/KNNSourceExcludesProcessor.java
```diff
@@ -0,0 +1,253 @@
...
+     * Collect all knn_vector field names from the index mappings for the target indices.
+     */
+    private Set<String> getVectorFieldsFromMappings(SearchRequest request) {
+        String[] indices = request.indices();
+        if (indices == null || indices.length == 0) {
+            return Set.of();
+        }
+
+        ClusterState state = clusterService.state();
+        String[] concreteIndices = indexNameExpressionResolver.concreteIndexNames(state, request);
```
Vikasht34, line 145: Does this throw expection on concreteIndexNames, if Index not found or index closed, do we want to wrap this?

src/main/java/org/opensearch/knn/search/processor/KNNSourceExcludesProcessor.java
```diff
@@ -0,0 +1,253 @@
...
+import java.util.Set;
+import java.util.stream.Stream;
+
+/**
+ * A system-generated search request processor that automatically adds _source excludes for knn vector fields
+ * found in the index mappings. This avoids returning large vector data in search responses by default.
+ * User-provided includes are always preserved.
+ */
+@Log4j2
+public final class KNNSourceExcludesProcessor extends AbstractProcessor implements SearchRequestProcessor, SystemGeneratedProcessor {
```
Vikasht34, line 73: Question :- What if Source is already disabled at Index Level , but this guy still fires and add one again excludes right?

src/main/java/org/opensearch/knn/search/processor/KNNSourceExcludesProcessor.java
```diff
@@ -0,0 +1,253 @@
...
+
+        final List<InnerHitBuilder> innerHitBuilders = innerHitsExtractor.getInnerHitBuilders();
+        for (InnerHitBuilder innerHitBuilder : innerHitBuilders) {
+            FetchSourceContext fetchSourceContext = innerHitBuilder.getFetchSourceContext();
+            List<String> fetchFields = innerHitBuilder.getFetchFields() != null
+                ? innerHitBuilder.getFetchFields().stream().map(fieldAndFormat -> fieldAndFormat.field).toList()
+                : List.of();
+
+            if (sourceExplicitTrue(fetchSourceContext)
+                && Stream.concat(Arrays.stream(fetchSourceContext.includes()), fetchFields.stream()).anyMatch(vectorFields::contains)) {
```
Vikasht34: why we need second condistion ? As I can see If sourceExplicitTrue is true, then includes().length == 0 so  So Arrays.stream(fetchSourceContext.includes()) is always an empty stream?

src/main/java/org/opensearch/knn/search/processor/KNNSourceExcludesProcessor.java
```diff
@@ -0,0 +1,253 @@
...
+        ClusterService clusterService,
+        IndexNameExpressionResolver indexNameExpressionResolver
+    ) {
+        super(tag, description, ignoreFailure);
+        this.clusterService = clusterService;
+        this.indexNameExpressionResolver = indexNameExpressionResolver;
+    }
+
+    @Override
+    public SearchRequest processRequest(SearchRequest request) {
```
Vikasht34, line 98: what happens when I have this , ansd I have feild like x_vector,. x_text

```
{ "_source": { "includes": ["x_*"] } }
```

this will create a bug since user is including but we re excluding?

src/main/java/org/opensearch/knn/search/processor/KNNSourceExcludesProcessor.java
```diff
@@ -0,0 +1,253 @@
...
+        final Set<String> vectorFields = getVectorFieldsFromMappings(request);
+        if (vectorFields.isEmpty()) {
+            return request;
+        }
+
+        final InnerHitsExtractor innerHitsExtractor = new InnerHitsExtractor();
+        if (request.source().query() != null) {
```
Vikasht34, line 105: Should have NPE check in source or it's always guranted that it can never become null?

src/main/java/org/opensearch/knn/search/processor/KNNSourceExcludesProcessor.java
```diff
@@ -0,0 +1,253 @@
...
+    }
+
+    @Getter
+    private static class InnerHitsExtractor implements QueryBuilderVisitor {
+
+        private final List<InnerHitBuilder> innerHitBuilders = new ArrayList<>();
+
+        @Override
+        public void accept(QueryBuilder qb) {
+            if (qb instanceof NestedQueryBuilder nestedQueryBuilder && nestedQueryBuilder.innerHit() != null) {
```
Vikasht34, line 360: so we are handling nested for innter hit , will it be case for any other pattern which alsio has inner hits but when thewy are mixed with KNN , they won't exclude https://docs.opensearch.org/latest/query-dsl/joining/has-child/

src/main/java/org/opensearch/knn/plugin/KNNPlugin.java
```diff
@@ -515,6 +516,8 @@ public Map<String, SystemGeneratedProcessor.SystemGeneratedFactory<SearchRequest
     ) {
         KNNClusterUtil.instance().setSearchPipelineService(parameters.searchPipelineService);
         return Map.of(
+            KNNSourceExcludesProcessor.Factory.TYPE,
```
Vikasht34, line 519: Since we have one more proecessor registered here , how can we be determistic about behaviour, becuase one can override each other?

src/main/java/org/opensearch/knn/search/processor/KNNSourceExcludesProcessor.java
```diff
@@ -0,0 +1,253 @@
...
+    @Override
+    public SearchRequest processRequest(SearchRequest request) {
+        final Set<String> vectorFields = getVectorFieldsFromMappings(request);
+        if (vectorFields.isEmpty()) {
+            return request;
+        }
+
+        final InnerHitsExtractor innerHitsExtractor = new InnerHitsExtractor();
+        if (request.source().query() != null) {
+            request.source().query().visit(innerHitsExtractor);
```
Vikasht34, line 106: Can we have NPE check on source?

src/main/java/org/opensearch/knn/search/processor/KNNSourceExcludesProcessor.java
```diff
@@ -0,0 +1,253 @@
...
+        }
+        return vectorFields;
+    }
+
+    /**
+     * Recursively walk the mapping tree and collect field names with type "knn_vector".
+     */
+    @SuppressWarnings("unchecked")
+    @VisibleForTesting
+    static void collectVectorFields(final Map<String, Object> mappingMap, final String prefix, final Set<String> vectorFields) {
```
Vikasht34, line 186: In case of Derived Source , vectors are not in Source then what is point of walking the mapping?

src/main/java/org/opensearch/knn/search/processor/KNNSourceExcludesProcessor.java
```diff
@@ -0,0 +1,253 @@
...
+        final Map<String, Object> properties = (Map<String, Object>) mappingMap.get("properties");
```
Vikasht34, line 187:
```{
  "content": {
    "type": "text",
    "fields": {
      "vector": { "type": "knn_vector", "dimension": 768 }
    }
  }
}
```

in this case this colection will not work since we are looking at property not in the fileds and exclude will not be appiled.

src/main/java/org/opensearch/knn/search/processor/KNNSourceExcludesProcessor.java
shatejas: it won't reach the code in processRequest if source is null,[ it will return false](https://github.com/opensearch-project/k-NN/pull/3152/changes#diff-110b41cd6690dd3b8df86ed74dc8bef9fa5ac0c206dbe894f2631ce015f035baR193) in should generate so its moot

src/main/java/org/opensearch/knn/search/processor/KNNSourceExcludesProcessor.java
shatejas: As long as the query correctly executes visitor pattern it would identify all innerhits, the interface is desined to walk each node of query tree

src/main/java/org/opensearch/knn/plugin/KNNPlugin.java
shatejas: One is s [pre](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/search/processor/mmr/MMRRerankProcessor.java#L319) and this one is [post](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/search/pipeline/SystemGeneratedProcessor.java#L19) so they are not overriding each other. I can add an integ test to guard us in future

src/main/java/org/opensearch/knn/search/processor/KNNSourceExcludesProcessor.java
shatejas: https://github.com/opensearch-project/k-NN/pull/3152#discussion_r3476346051

src/main/java/org/opensearch/knn/search/processor/KNNSourceExcludesProcessor.java
shatejas: true will update to use regex match

src/main/java/org/opensearch/knn/search/processor/KNNSourceExcludesProcessor.java
shatejas: true it should be an or condition instead of && there

src/main/java/org/opensearch/knn/search/processor/KNNSourceExcludesProcessor.java
shatejas: Derived source doesn't change the behavior in terms of input and output so you still need all the vector fields to exclude for response

src/main/java/org/opensearch/knn/search/processor/KNNSourceExcludesProcessor.java
Vikasht34: they have differnet builder but we are only checking nested ( qb instanceof NestedQueryBuilder nestedQueryBuilder) not NestedQueryBuilder ,

src/main/java/org/opensearch/knn/search/processor/KNNSourceExcludesProcessor.java
Vikasht34: this feture is for source excluds , when we don't have vector in feild , what we are exluding?

src/main/java/org/opensearch/knn/search/processor/KNNSourceExcludesProcessor.java
shatejas: I don't think multi-field support is there for knn, I tried it and failed at ingestion - I used this as mapping
```
  curl -X PUT "localhost:9200/test-multifield" -H 'Content-Type: application/json' -d'
  {
    "settings": { "index.knn": true },
    "mappings": {
      "properties": {
        "content": {
          "type": "text",
          "fields": {
            "vector": { "type": "knn_vector", "dimension": 3 }
          }
        },
        "title": { "type": "text" }
      }
    }
  }'

  # Index a document
  curl -X PUT "localhost:9200/test-multifield/_doc/1" -H 'Content-Type: application/json' -d'
  {
    "content": [1.0, 2.0, 3.0],
    "title": "hello"
  }'

```

src/main/java/org/opensearch/knn/plugin/KNNPlugin.java
Vikasht34: You are looking at wrong precessor , we don't register re-rank we register oversample MMR and that's POST

src/main/java/org/opensearch/knn/search/processor/KNNSourceExcludesProcessor.java
Vikasht34: Sounds Good

src/main/java/org/opensearch/knn/search/processor/KNNSourceExcludesProcessor.java
Vikasht34: Sounds good

src/main/java/org/opensearch/knn/plugin/KNNPlugin.java
shatejas: Changed to pre

src/main/java/org/opensearch/knn/search/processor/KNNSourceExcludesProcessor.java
shatejas: if this throws an exception the search request will fail in later stages so I think it should be fine to throw

src/main/java/org/opensearch/knn/search/processor/KNNSourceExcludesProcessor.java
shatejas: changed to || instead of &&

src/main/java/org/opensearch/knn/search/processor/KNNSourceExcludesProcessor.java
shatejas: moved it outside

src/main/java/org/opensearch/knn/search/processor/KNNSourceExcludesProcessor.java
shatejas: added a check for source.enabled: true

src/main/java/org/opensearch/knn/search/processor/KNNSourceExcludesProcessor.java
shatejas: Core only has NestedQueryBuilder with innerhits. I wouldn't want to use reflection, one way is to have an interface but will need changes in core

Vikasht34: Make Sense ..Thanks

Vikasht34 (dismissed):
LGTM!!

src/test/java/org/opensearch/knn/integ/KNNSourceExcludesProcessorIT.java
```diff
@@ -0,0 +1,1271 @@
...
+            .endObject()
+            .endObject()
+            .endObject()
+            .toString();
+
+        Response response = performSearch(indexName, query);
+        List<Map<String, Object>> hits = parseHits(response);
+
+        assertEquals("MMR should return 3 results", 3, hits.size());
+        for (Map<String, Object> hit : hits) {
```
bzhangam, line 983: It's better to verify the results contain the two diverse docs.

shatejas: added more asserts

Vikasht34 (approved):

src/main/java/org/opensearch/knn/search/processor/KNNSourceExcludesProcessor.java
```diff
@@ -0,0 +1,370 @@
...
+        ClusterState state = clusterService.state();
+        String[] concreteIndices = indexNameExpressionResolver.concreteIndexNames(state, request);
+        Set<String> vectorFields = new HashSet<>();
+
+        for (String indexName : concreteIndices) {
+            IndexMetadata indexMetadata = state.metadata().index(indexName);
+            if (indexMetadata != null) {
+                collectExposedVectorFields(indexMetadata, vectorFields);
+            }
+        }
```
navneet1v, line 153: can we parallelize this?

src/main/java/org/opensearch/knn/search/processor/KNNSourceExcludesProcessor.java
```diff
@@ -0,0 +1,370 @@
...
+        return vectorFields;
+    }
+
+    /**
+     * Adds all {@code knn_vector} field paths from the given index to {@code vectorFields},
+     * skipping the index if {@code _source} is disabled, and filtering out fields already
+     * covered by the index-level {@code _source.excludes}.
+     */
+    private static void collectExposedVectorFields(final IndexMetadata indexMetadata, final Set<String> vectorFields) {
```
navneet1v, line 162: why this function is static?

src/main/java/org/opensearch/knn/search/processor/KNNSourceExcludesProcessor.java
```diff
@@ -0,0 +1,370 @@
...
+     * {@code type} is {@code knn_vector} to {@code vectorFields}.
+     * Nested and object fields are traversed by recursing into their {@code properties} block.
+     *
+     * @param mappingMap   the mapping node to inspect (top-level or a nested object block)
+     * @param prefix       dot-separated path prefix for this node (empty string at the root)
+     * @param vectorFields accumulator for discovered field paths
+     */
+    @SuppressWarnings("unchecked")
+    @VisibleForTesting
+    static void collectVectorFields(final Map<String, Object> mappingMap, final String prefix, final Set<String> vectorFields) {
```
navneet1v, line 186: same as above.

src/main/java/org/opensearch/knn/plugin/KNNPlugin.java
```diff
@@ -515,6 +516,8 @@ public Map<String, SystemGeneratedProcessor.SystemGeneratedFactory<SearchRequest
...
+            new KNNSourceExcludesProcessor.Factory(clusterService, indexNameExpressionResolver),
```
navneet1v, line 520: what kind of stats is being emitted for this new processor?

VijayanB (approved):

navneet1v (commented):
I see the PR is merged. Some comments were left in my review. just adding it.

PR #3153, opened 2026-03-10
Faiss bbq index build logic - MemOptimizedBBQIndexBuildStrategy

0ctopus13prime:
### Description
Adding MemOptimizedBBQIndexBuildStrategy to build HNSW based on Lucene Lucene102BinaryQuantizedVectorsWriter.

### Related Issues
N/A

### Check List


src/main/java/org/opensearch/lucene/lucene102/Lucene102BinaryQuantizedVectorsWriter.java
```diff
@@ -0,0 +1,952 @@
...
+
+import static org.apache.lucene.index.VectorSimilarityFunction.COSINE;
+import static org.apache.lucene.search.DocIdSetIterator.NO_MORE_DOCS;
+import static org.apache.lucene.util.RamUsageEstimator.shallowSizeOfInstance;
+import static org.apache.lucene.util.quantization.OptimizedScalarQuantizer.discretize;
+import static org.apache.lucene.util.quantization.OptimizedScalarQuantizer.packAsBinary;
+import static org.apache.lucene.util.quantization.OptimizedScalarQuantizer.transposeHalfByte;
+
+/** Copied from Lucene, replace with Lucene's implementation sometime after Lucene 10 */
+class Lucene102BinaryQuantizedVectorsWriter extends FlatVectorsWriter {
```
navneet1v: @0ctopus13prime can we remove these classes, I think it was added because Lucene upgrade was failing.

0ctopus13prime: Working on it, will remove in next rev.

jni/patches/faiss/0008-Added-skipping-flat-storage-availability-in-write_in.patch
```diff
@@ -0,0 +1,87 @@
...
+@@ -1021,7 +1021,12 @@ void write_index_binary(const IndexBinary* idx, IOWriter* f) {
+         }
+ 
+         write_HNSW(&idxhnsw->hnsw, f);
+-        write_index_binary(idxhnsw->storage, f);
++        if (io_flags & IO_FLAG_SKIP_STORAGE) {
++            uint32_t n4 = fourcc("null");
++            WRITE1(n4);
++        } else {
++            write_index_binary(idxhnsw->storage, f, io_flags);
```
navneet1v, line 35: are we planning to contribute this IO_FLAG skip to faiss?

0ctopus13prime: yes! absolutely! 
I'm hoping to raise a single PR to cover both write_index and write_binary_index

src/main/java/org/opensearch/lucene/lucene102/Lucene102BinaryQuantizedVectorsReader.java
```diff
@@ -0,0 +1,401 @@
...
+        }
+
+        @Override
+        public DocIndexIterator iterator() {
+            return rawVectorValues.iterator();
+        }
+
+        @Override
+        public VectorScorer scorer(float[] query) throws IOException {
+            return quantizedVectorValues.scorer(query);
```
VijayanB: can we add this?
```
    @Override
    public VectorScorer rescorer(float[] query) throws IOException {
      return rawVectorValues.rescorer(query);
    }
```

This could be useful if someone wants to do scoring based full precision.

src/main/java/org/opensearch/knn/jni/JNIService.java
```diff
@@ -465,4 +473,69 @@ public static KNNQueryResult[] radiusQueryIndex(
...
+            FaissService.passBBQVectorsWithCorrectionFactors(indexMemoryAddress, buffer, loopSize);
+            return;
+        }
+
+        throw new IllegalArgumentException(
+            String.format(Locale.ROOT, "passBBQVectorsWithCorrectionFactors not supported for provided engine : %s", knnEngine.getName())
+        );
+    }
+
+    public static void releaseFaissBBQIndex(final long indexMemoryAddress, final KNNEngine knnEngine) {
```
VijayanB: nit: since you are validating the engine, i don't think you have to include faiss in the method name.

0ctopus13prime: yeah, good point, will update in the next rev.

src/main/java/org/opensearch/lucene/lucene102/Lucene102BinaryQuantizedVectorsReader.java
0ctopus13prime: I removed all Lucene 102 stuff and replaced them with Lucene104. 
So the method will come naturally.

src/main/java/org/opensearch/knn/index/codec/nativeindex/MemOptimizedBBQIndexBuildStrategy.java
```diff
@@ -0,0 +1,414 @@
...
+ * To preserve distance accuracy, each quantized vector is accompanied by correction factors:
+ * <ul>
+ *   <li>{@code lowerInterval} (float) — lower bound of the quantization interval</li>
+ *   <li>{@code upperInterval} (float) — upper bound of the quantization interval</li>
+ *   <li>{@code additionalCorrection} (float) — residual correction term for score adjustment</li>
+ *   <li>{@code quantizedComponentSum} (int) — sum of quantized component values, used in
+ *       asymmetric distance computation (ADC)</li>
+ * </ul>
+ * These correction factors enable accurate approximate distance calculations between a 4-bit
+ * quantized query and 1-bit quantized data vectors during search (see docs/5_bulk_simd_bbq_adc.md).
```
navneet1v, line 63: i dont' see these .md files in the PR.

src/main/java/org/opensearch/knn/index/codec/nativeindex/MemOptimizedBBQIndexBuildStrategy.java
```diff
@@ -0,0 +1,414 @@
...
+            writeState.context,
+            indexInfo.getFieldInfo().getName()
+        );
+
+        // Open the full-precision flat vectors reader (.vec file) — this is required as a
+        // dependency for the binary quantized reader, which needs access to the original
+        // vectors for centroid computation and corrective term derivation.
+        final FlatVectorsReader fullPrecisionVectorsReader = new Lucene99FlatVectorsFormat(
+            FlatVectorScorerUtil.getLucene99FlatVectorsScorer()
+        ).fieldsReader(readState);
```
navneet1v, line 135: This all can be avoided once you start to hook the right HNSWVectorsFormat for Faiss BBQ.

src/main/java/org/opensearch/knn/index/codec/nativeindex/MemOptimizedBBQIndexBuildStrategy.java
```diff
@@ -0,0 +1,414 @@
...
+     *
+     * <h3>Byte ordering</h3>
+     * All float and int values are serialized in little-endian byte order to match the
+     * native (x86/ARM) memory layout, avoiding byte-swapping overhead in the C++ layer.
+     *
+     * @param indexMemoryAddress    pointer to the native Faiss BBQ index
+     * @param binarizedVectorValues provides quantized binary codes and correction factors
+     * @param quantizedVecBytes     byte length of a single quantized binary code
+     */
+    private void passQuantizedVectorsAndCorrectionFactors(
```
navneet1v, line 334: any reason for not moving this under VectorTransfer abstraction?

navneet1v (commented):
Will do another round of review tomorrow.

src/main/java/org/opensearch/knn/index/codec/nativeindex/MemOptimizedBBQIndexBuildStrategy.java
```diff
@@ -0,0 +1,414 @@
...
+            fullPrecisionVectorsReader.close();
+            throw e;
+        }
+
+        try (faissBBQVectorsReader) {
+            // Retrieve the binarized vector values for this field. The returned FloatVectorValues
+            // is actually a BinarizedVectorValues instance that provides both the quantized binary
+            // codes and their associated correction factors.
+            final FloatVectorValues floatVectorValues = faissBBQVectorsReader.getFloatVectorValues(indexInfo.getFieldInfo().getName());
+            final Field f = floatVectorValues.getClass().getDeclaredField("quantizedVectorValues");
```
kotwanikunal, line 163: Can we work this out without using reflection?

kotwanikunal:
@0ctopus13prime - do you mind rebasing your branch off the latest main commits?

src/main/java/org/opensearch/knn/index/codec/nativeindex/MemOptimizedBBQIndexBuildStrategy.java
0ctopus13prime: Yes, correct.
Will change once Kunal's done.

src/main/java/org/opensearch/knn/index/codec/nativeindex/MemOptimizedBBQIndexBuildStrategy.java
0ctopus13prime: thank you for the hawk eye, will remove it.

src/main/java/org/opensearch/knn/index/codec/nativeindex/MemOptimizedBBQIndexBuildStrategy.java
```diff
@@ -0,0 +1,414 @@
...
+ * <h2>SIMD-Accelerated Search</h2>
+ * During search, the ADC distance computation between 4-bit query vectors and 1-bit data vectors
+ * is offloaded to native SIMD code (AVX512 / ARM NEON) via {@code FaissBBQNativeRandomVectorScorer}
+ * for maximum throughput (see docs/5_bulk_simd_bbq_adc.md).
+ *
+ * @see NativeIndexBuildStrategy
+ * @see NativeIndexBuildStrategyFactory — returns this strategy when field info contains faiss_bbq_config
+ */
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+public class MemOptimizedBBQIndexBuildStrategy implements NativeIndexBuildStrategy {
```
navneet1v, line 81: This class will need some refactoring. For now since this is a feature branch, I am okay on this. But let revisit this, once we have other pieces tied together

jni/src/faiss_index_service.cpp
```diff
@@ -203,6 +204,49 @@ jlong BinaryIndexService::initIndex(
     return reinterpret_cast<jlong>(idMap.release());
 }
 
+jlong BinaryIndexService::initFaissBBQIndex(knn_jni::JNIUtilInterface *jniUtil, JNIEnv *env, faiss::MetricType metric,
+                                            std::string indexDescription, int dim, int numVectors, int threadCount,
+                                            std::unordered_map<std::string, jobject> parameters, float centroidDp, int quantizedVectorBytes) {
+    // Extract `m` from the binary index
+    int32_t m;
+    {
```
navneet1v, line 212: what about ef_construction? do we not need to set that? also what about ef_search?

src/main/java/org/opensearch/knn/index/codec/nativeindex/MemOptimizedBBQIndexBuildStrategy.java
0ctopus13prime: Unfortunately, I could not find a way to extract it.
`QuantizedByteVectorValues` is hidden in `ScalarQuantizedVectorValues` which is already in protected scope and it's not reachable.

navneet1v (approved):

src/main/java/org/opensearch/knn/index/codec/nativeindex/MemOptimizedBBQIndexBuildStrategy.java
0ctopus13prime: Yes, VectorTransfer is copying from Java heap to C++ std::vector, then calling `insertToIndex` to push vectors into Faiss.
But for BBQ, we can do better than this, directly passing Java heap to FaissBBQFlat's buffer avoiding 2 copies. (Java heap -> std::vector -> FaissBBQFlat)
Also insertion could be faster since we're sending only doc ids.

src/main/java/org/opensearch/knn/index/codec/nativeindex/MemOptimizedBBQIndexBuildStrategy.java
0ctopus13prime: sure, sounds good!

jni/src/faiss_index_service.cpp
0ctopus13prime: ef_construction and ef_search are both being handled in SetExtraParameters.
Only `m` is missing there since it's given as string descriptor like "BHNSW32...".

src/main/java/org/opensearch/knn/index/codec/nativeindex/MemOptimizedBBQIndexBuildStrategy.java
navneet1v: Lets see if we can hide all these under right private functions to ensure code is maintainable and once we fix Lucene to expose these QuantizedVectorValues this becomes a easy swap

src/main/java/org/opensearch/knn/index/codec/nativeindex/MemOptimizedBBQIndexBuildStrategy.java
navneet1v: @0ctopus13prime the reason for not passing the java heap pointer is to ensure that we don't put extra pressure on heap + if vectors are larger than heap size we can build the indices. The same will be true for BBQ also.

0ctopus13prime: Yeah, it's also true for this current implementation as we're batching vectors + correction factors.
It's not loading entire vectors in heap then pass it, it's batching with 64kb unit.
This is an optimization to pass all quantized vectors to C++ layer without double copying. 
Then we can then only pass document ids during insertion.

navneet1v: @0ctopus13prime okay I think I got what you are saying. Now this tells me that, we should our other transfer vectors code then. :P 

The reason for doing the copy was, before we made our index build iterative we used to transfer everything to off heap and then build the index. then we added the iterative index build logic. At that time we could have used the vectors present in heap. Thats a good optimization we missed. Lets create a GH issue for that. I think we are missing a good optimization there

Vikasht34 (approved):
Ready for Feature branch ..

PR #3154, opened 2026-03-10
Support lucene bbq flat for 1 bit

naveentatikonda:
### Description
This PR adds support for Lucene BBQ Flat using `Lucene104ScalarQuantizedVectorsFormat` by introducing a new method called `flat` and only supports 1 bit or 32x compression for now. Also, adds validation to reject unsupported compression levels, engines, mode parameter and other validations.

The sample mapping is shared below:

```
    "mappings": {
        "properties": {
            "target-field": {
                "type": "knn_vector",
                "dimension": 4,
                "compression_level": "32x", // Optional because 32x is default
                "method": {
                    "name": "flat"
                }
            }
        }
    }
```

### Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 * Fix score conversion logic for radial exact search [#3110](https://github.com/opensearch-project/k-NN/pull/3110)
 * Fix lucene reduce to topK when rescoring is enabled [#3124](https://github.com/opensearch-project/k-NN/pull/3124)
 * Fix bugs in optimistic search for nested Cagra index [#3155](https://github.com/opensearch-project/k-NN/pull/3155)
+* Fixed generating random entry points for CagraIndex in MOS when numVectors < entryPoints [#3161](https://github.com/opensearch-project/k-NN/pull/3161)
 * Fix integer overflow for memory optimized search [#3130](https://github.com/opensearch-project/k-NN/pull/3130)
 
 ### Refactoring
--- a/src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissMemoryOptimizedSearcher.java
+++ b/src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissMemoryOptimizedSearcher.java
@@ -251,6 +251,9 @@ public class FaissMemoryOptimizedSearcher implements VectorSearcher {
         }
 
         private static DocIdSetIterator generateRandomEntryPoints(final int numberOfEntryPoints, int totalNumberOfVectors) {
+            if (numberOfEntryPoints >= totalNumberOfVectors) {
+                return DocIdSetIterator.all(totalNumberOfVectors);
+            }
             return new DocIdSetIterator() {
                 final RobustUniqueRandomIterator robustUniqueRandomIterator = new RobustUniqueRandomIterator(
                     totalNumberOfVectors,
--- a/src/test/java/org/opensearch/knn/memoryoptsearch/faiss/RandomEntryPointsKnnSearchStrategyTests.java
+++ b/src/test/java/org/opensearch/knn/memoryoptsearch/faiss/RandomEntryPointsKnnSearchStrategyTests.java
@@ -36,4 +36,26 @@ public class RandomEntryPointsKnnSearchStrategyTests {
             assertTrue(internalVectorId < numVectors);
         }
     }
+
+    @Test
+    @SneakyThrows
+    public void testRandomEntryPointsGeneration_whenNumVectorsIsLessThanNumEntries_thenSuccess() {
+        final int numEntries = 100;
+        final long numVectors = 10;
+
+        // Create strategy
+        final FaissMemoryOptimizedSearcher.RandomEntryPointsKnnSearchStrategy strategy =
+            new FaissMemoryOptimizedSearcher.RandomEntryPointsKnnSearchStrategy(numEntries, numVectors, mock(KnnSearchStrategy.class));
+
+        // Validate #entry points
+        assertEquals(numEntries, strategy.numberOfEntryPoints());
+
+        // We should get exactly `numEntries` vector ids.
+        final DocIdSetIterator iterator = strategy.entryPoints();
+        for (int i = 0; i < numVectors; ++i) {
+            final int internalVectorId = iterator.nextDoc();
+            assertTrue(internalVectorId >= 0);
+            assertTrue(internalVectorId < numVectors);
+        }
+    }
 }


src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldType.java
```diff
@@ -141,11 +142,19 @@ public Object valueForDisplay(Object value) {
...
         if (userProvidedContext != null) {
             return userProvidedContext;
         }
         KNNMappingConfig knnMappingConfig = getKnnMappingConfig();
+        Optional<KNNMethodContext> methodContext = knnMappingConfig.getKnnMethodContext();
+
+        // Flat method uses BBQ (1-bit quantization) without an HNSW graph, set a default oversample factor of 2.0
+        if (methodContext.isPresent() && METHOD_FLAT.equals(methodContext.get().getMethodComponentContext().getName())) {
+            return RescoreContext.builder().oversampleFactor(FLAT_OVERSAMPLE_FACTOR).userProvided(false).build();
+        }
```
navneet1v: why we are adding a if/else here, shouldn't this be responsibility of `compressionLevel.getDefaultRescoreContext` method? since its a default resolution of rescoreContext

src/main/java/org/opensearch/knn/index/engine/lucene/LuceneMethodResolver.java
```diff
@@ -37,6 +43,12 @@ public ResolvedMethodContext resolveMethod(
         boolean shouldRequireTraining,
         final SpaceType spaceType
     ) {
+        // Check if the user specified flat method
+        if (isFlatMethod(knnMethodContext)) {
+            validateNotTrainingContext(shouldRequireTraining, knnMethodConfigContext);
+            return resolveFlatMethod(knnMethodContext, knnMethodConfigContext, spaceType);
+        }
```
navneet1v: why we are not creating something as FlatMethodResolver?

src/main/java/org/opensearch/knn/index/codec/backward_codecs/KNN920Codec/KNN920PerFieldKnnVectorsFormat.java
```diff
@@ -22,10 +25,15 @@ public KNN920PerFieldKnnVectorsFormat(final Optional<MapperService> mapperServic
             Lucene92HnswVectorsFormat.DEFAULT_MAX_CONN,
             Lucene92HnswVectorsFormat.DEFAULT_BEAM_WIDTH,
             () -> new Lucene92HnswVectorsFormat(),
-            knnVectorsFormatParams -> new Lucene92HnswVectorsFormat(
-                knnVectorsFormatParams.getMaxConnections(),
-                knnVectorsFormatParams.getBeamWidth()
-            )
+            Map.of(LuceneVectorsFormatType.HNSW, ctx -> {
```
navneet1v: Its pretty counterintuitive to see that we have to change these backward codecs, do we really need it?

src/main/java/org/opensearch/knn/index/codec/KNN9120Codec/KNN9120PerFieldKnnVectorsFormat.java
```diff
@@ -38,50 +46,51 @@ public KNN9120PerFieldKnnVectorsFormat(
...
+                merge.v1(),
+                p.getBits(),
+                p.isCompressFlag(),
+                p.getConfidenceInterval(),
+                merge.v2()
+            );
+        },
+            LuceneVectorsFormatType.FLAT,
+            ctx -> new Lucene104ScalarQuantizedVectorsFormat(Lucene104ScalarQuantizedVectorsFormat.ScalarEncoding.SINGLE_BIT_QUERY_NIBBLE)
+        );
```
navneet1v, line 95: since we are adding this. Why we are still using KNN9120PerFieldKNNVectorsFormat? we should create KNN1040PerFieldKNNVectorsFormat now, and we don't need to touch this format so that all old code keep on working, and new segments goes through new path.

Vikasht34 (commented):
@naveentatikonda Kudos for structuring code neat way .. Thank you !! Overall Functionality  looks correct, Will take a deeper look tomorrow.

src/main/java/org/opensearch/knn/index/codec/backward_codecs/KNN920Codec/KNN920PerFieldKnnVectorsFormat.java
naveentatikonda: Makes sense, I have updated the implementation to not change old codec formats

src/main/java/org/opensearch/knn/index/codec/KNN9120Codec/KNN9120PerFieldKnnVectorsFormat.java
naveentatikonda: Makes sense, I updated it. I thought we intentionally didn't change it with Lucene 10.4 version bump change

src/main/java/org/opensearch/knn/index/codec/KnnVectorsFormatContext.java
```diff
@@ -0,0 +1,46 @@
...
+
+    /**
+     * The method component parameters (may be null).
+     */
+    Map<String, Object> params;
+
+    /**
+     * Default max connections if not specified in params.
+     */
+    int defaultMaxConnections;
```
kotwanikunal, line 40: nit: Should these be Optional given we don't need them for `flat`?

src/main/java/org/opensearch/knn/index/codec/BasePerFieldKnnVectorsFormat.java
```diff
@@ -107,48 +160,126 @@ public KnnVectorsFormat getKnnVectorsFormatForField(final String field) {
...
+
+    /**
+     * Routes the method context to the appropriate {@link LuceneVectorsFormatType}.
+     */
+    private LuceneVectorsFormatType determineLuceneFormatType(
+        final String field,
+        final KNNMethodContext methodContext,
+        final Map<String, Object> params
+    ) {
+        if (METHOD_FLAT.equals(methodContext.getMethodComponentContext().getName())) {
```
kotwanikunal: I think we should have a registry check here as well to check for registered method types. It can just be along the lines of check registry is not null and contains flat - that way we do not have to worry about version compat going forward.

```
if (METHOD_FLAT.equals(methodContext.getMethodComponentContext().getName())) {
       if (luceneFormatResolvers.containsKey(LuceneVectorsFormatType.FLAT)) {
           log.debug("Initialize KNN vector format for field [{}] with Lucene BBQ flat format", field);
           return LuceneVectorsFormatType.FLAT;
       }
      log.error("BBQ flat is not supported"); //OR
      throw new IllegalStateException("BBQ flat is not supported"))
   }
```

naveentatikonda: makes sense, done

src/main/java/org/opensearch/knn/index/codec/KnnVectorsFormatContext.java
naveentatikonda: IMO making those fields optional adds complexity at every consumer site — each factory would need to check isPresent() or call orElse(). Right now the FLAT factory in KNN1040PerFieldKnnVectorsFormat simply ignores those fields

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/KNN1040PerFieldKnnVectorsFormat.java
```diff
@@ -0,0 +1,107 @@
...
+import java.util.function.Function;
+
+/**
+ * Per-field KNN vectors format for the KNN1040 codec. Uses {@link Lucene99HnswVectorsFormat}
+ * for HNSW, {@link Lucene99RWHnswScalarQuantizedVectorsFormat} for scalar quantization (to
+ * preserve the {@code confidenceInterval} parameter), and
+ * {@link Lucene104ScalarQuantizedVectorsFormat} with {@code SINGLE_BIT_QUERY_NIBBLE} encoding
+ * for the flat BBQ method.
+ */
+public class KNN1040PerFieldKnnVectorsFormat extends BasePerFieldKnnVectorsFormat {
```
navneet1v: since we added this, please move KNN9120 in backwards_codec package.

src/main/java/org/opensearch/knn/index/codec/BasePerFieldKnnVectorsFormat.java
```diff
@@ -29,25 +28,40 @@
...
 
     private final Optional<MapperService> mapperService;
     private final int defaultMaxConnections;
     private final int defaultBeamWidth;
     private final Supplier<KnnVectorsFormat> defaultFormatSupplier;
+    // Legacy path: used by backward codecs (KNN920 – KNN9120)
     private final Function<KNNVectorsFormatParams, KnnVectorsFormat> vectorsFormatSupplier;
-    private Function<KNNScalarQuantizedVectorsFormatParams, KnnVectorsFormat> scalarQuantizedVectorsFormatSupplier;
+    private final Function<KNNScalarQuantizedVectorsFormatParams, KnnVectorsFormat> scalarQuantizedVectorsFormatSupplier;
+    // Registry path: used by current codecs (KNN1040+)
```
navneet1v: Should we create PerEngineFormatResolvers? rather than having these different different parameters for different engines and quantizations.

src/main/java/org/opensearch/knn/index/codec/BasePerFieldKnnVectorsFormat.java
```diff
@@ -29,25 +28,40 @@
...
+    private final Map<LuceneVectorsFormatType, Function<KnnVectorsFormatContext, KnnVectorsFormat>> luceneFormatResolvers;
```
navneet1v: this seems right to me, but I don't see how we can fix the other code paths which are added as vectorsFormatSupplier and scalarQuantizedVectorsFormatSupplier

src/main/java/org/opensearch/knn/index/codec/BasePerFieldKnnVectorsFormat.java
```diff
@@ -76,6 +90,45 @@ public BasePerFieldKnnVectorsFormat(
         );
     }
 
+    protected BasePerFieldKnnVectorsFormat(
```
navneet1v: Should we create another BasePerFieldKNNVectorsFormat for codec 10.4? We are trying to shove everything in 1 BasePerFieldKNNVectorsFormat which is adding more and more conditional logic.

src/main/java/org/opensearch/knn/index/engine/lucene/LuceneFlatMethodResolver.java
```diff
@@ -0,0 +1,104 @@
...
+            knnMethodContext,
+            KNNEngine.LUCENE,
+            spaceType,
+            METHOD_FLAT
+        );
+        resolveMethodParams(resolvedKNNMethodContext.getMethodComponentContext(), knnMethodConfigContext, FLAT_METHOD_COMPONENT);
+
+        CompressionLevel resolvedCompressionLevel = CompressionLevel.isConfigured(knnMethodConfigContext.getCompressionLevel())
+            ? knnMethodConfigContext.getCompressionLevel()
+            : DEFAULT_COMPRESSION;
```
navneet1v: This guy assumes that validation on the compression level has already been set. Should we first resolve everything and then do the validation? WDYT?

src/main/java/org/opensearch/knn/index/engine/lucene/LuceneFlatSearchContext.java
```diff
@@ -0,0 +1,22 @@
...
+package org.opensearch.knn.index.engine.lucene;
+
+import org.opensearch.knn.index.engine.KNNLibrarySearchContext;
+import org.opensearch.knn.index.engine.Parameter;
+import org.opensearch.knn.index.engine.model.QueryContext;
+
+import java.util.Collections;
+import java.util.Map;
+
+public class LuceneFlatSearchContext implements KNNLibrarySearchContext {
```
navneet1v, line 20: Please add java doc on all classes and public functions

src/main/java/org/opensearch/knn/index/engine/EngineResolver.java
```diff
@@ -95,6 +96,11 @@ private KNNEngine resolveKNNEngine(
             return KNNEngine.FAISS;
         }
 
+        // Flat method is only supported by Lucene
+        if (knnMethodContext != null && METHOD_FLAT.equals(knnMethodContext.getMethodComponentContext().getName())) {
```
navneet1v: can we use enum comparision with `==` rather than equals method.

kotwanikunal: The current `.equals()` approach will take care of null checks within the method context object as well. I like the existing approach :)

src/main/java/org/opensearch/knn/index/codec/BasePerFieldKnnVectorsFormat.java
kotwanikunal: We will have additional formats for #3153 follow ups as well.

src/main/java/org/opensearch/knn/index/codec/BasePerFieldKnnVectorsFormat.java
Vikasht34: Seems like we already have in this class ? Curious to know what exactly intent here?

src/main/java/org/opensearch/knn/index/engine/lucene/LuceneFlatMethodResolver.java
Vikasht34: This is correct sequnece  ..Validation , if validations fails , there is no point resolving it .

src/main/java/org/opensearch/knn/index/codec/BasePerFieldKnnVectorsFormat.java
navneet1v: yeah its all in one class. the idea is can be break down into separate class and make the code easy to maintain, understand and contribute.

src/main/java/org/opensearch/knn/index/engine/EngineResolver.java
navneet1v: I think when we use `==` it also takes care of those things. My worry is on the string comparisons with upper case and lower cases.

src/main/java/org/opensearch/knn/index/codec/backward_codecs/KNN920Codec/KNN920PerFieldKnnVectorsFormat.java
Vikasht34: I feel @naveentatikonda the first version was correct , now we have concept of Supplier and resolver just to not change the old Codec, I  will be more aligned to have one , which can be resolver for old and new formats so that we have one thing with Lucene and it will avoid branching and we really don't want to create another format just for Flat since this all belong to Lucene , else we could have created one for HNSW , One for SQ . This will make code more complex.

src/main/java/org/opensearch/knn/index/codec/BasePerFieldKnnVectorsFormat.java
Vikasht34: Make Sense , This we should do it definitely

src/main/java/org/opensearch/knn/index/codec/backward_codecs/BasePerFieldKnnVectorsFormat.java
navneet1v, line 46: not a blocker for the PR, but I thought the name of this format will be LegacyBasePerFieldKnnVectorsFormat

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/KNN1040PerFieldKnnVectorsFormat.java
```diff
@@ -0,0 +1,110 @@
...
+        final Optional<MapperService> mapperService,
+        NativeIndexBuildStrategyFactory nativeIndexBuildStrategyFactory
+    ) {
+        super(
+            mapperService,
+            Lucene99HnswVectorsFormat.DEFAULT_MAX_CONN,
+            Lucene99HnswVectorsFormat.DEFAULT_BEAM_WIDTH,
+            Lucene99HnswVectorsFormat::new,
+            new LuceneCodecFormatResolver(buildLuceneFormatResolvers()),
+            new FaissCodecFormatResolver(mapperService, nativeIndexBuildStrategyFactory),
```
navneet1v, line 58: [Not a blocker]
why cannot we make the FaissCodecFormatResolvers similar to `LuceneCodecFormatResolver`

src/main/java/org/opensearch/knn/index/engine/CodecFormatResolver.java
```diff
@@ -0,0 +1,37 @@
...
+     * @param defaultBeamWidth      default beam width for HNSW
+     * @return the resolved {@link KnnVectorsFormat}
+     */
+    KnnVectorsFormat resolve(
+        String field,
+        KNNMethodContext methodContext,
+        Map<String, Object> params,
+        int defaultMaxConnections,
+        int defaultBeamWidth
+    );
```
navneet1v, line 34: [not a blocker for PR]
is this is only for Faiss, we should see how we can remove it.

Vikasht34 (approved):

src/main/java/org/opensearch/knn/index/codec/backward_codecs/BasePerFieldKnnVectorsFormat.java
naveentatikonda: I didn't change it because it will change in all the classes in old codecs and instead I changed the name of new class

src/main/java/org/opensearch/knn/index/engine/CodecFormatResolver.java
naveentatikonda: This is for Lucene, we either need to create separate interfaces or support both method signatures in one interface

PR #3155, opened 2026-03-11
[Bug] MOS optimistic search bug fix on nested Cagra index.

0ctopus13prime:
### Description
# Problems

We have three problems.

1. RandomEntryPointsKnnSearchStrategy generates duplicated elements.
2. Second deep-dive search for nested index
  2.1. The seed doc IDs from the second deep dive are not honored. Instead, random entry points are still used.
  2.2. Nested index search uses TopKnnCollectorManager instead of DiversifyingNearestChildrenKnnCollectorManager.
  
# Unique Random Entry Points Generation
The fix should be straightforward.

A naive implementation would use a Set to remove duplicates, but we can instead use a random sampling algorithm with O(1) memory.

# Second Deep-Dive Search for Nested Index
The memory-optimized search uses an optimistic search strategy:
1. Run ANN search as usual.
2. Identify "promising segments" that are expected to contain vectors closer to the query vector.
3. Re-run ANN search on those segments using the doc IDs collected from the first phase as entry points.

This second phase is called the second deep-dive search.

The doc IDs collected during the first phase must be used as entry points. Otherwise, the purpose of the second deep dive is defeated.

However, for CAGRA, we currently use random entry points regardless.

```
if (hnsw instanceof FaissCagraHNSW cagraHNSW) {
    return new KnnCollector.Decorator(ordinalTranslatedKnnCollector) {
        @Override
        public KnnSearchStrategy getSearchStrategy() {
            return new RandomEntryPointsKnnSearchStrategy(
                cagraHNSW.getNumBaseLevelSearchEntryPoints(),
                cagraHNSW.getTotalNumberOfVectors(),
                knnCollector.getSearchStrategy()
            );
        }
    };
} else {
```

## Fix

If the KnnCollector's search strategy is KnnSearchStrategy.Seeded, we must honor it and ensure that the search starts from the provided entry points.

# Use DiversifyingNearestChildrenKnnCollectorManager

For a nested index, we track the best parent score, not individual child vectors.

For example:
- Child docs 1, 2, 3 belong to parent doc 4.
- The best child among 1, 2, 3 should represent the parent.
- Collecting both 1 and 2 is not allowed.

This behavior is implemented by DiversifyingNearestChildrenKnnCollectorManager.

However, we currently use TopKnnCollectorManager for nested indices.
```
final ReentrantKnnCollectorManager reentrantCollectorManager =
    new ReentrantKnnCollectorManager(
        new TopKnnCollectorManager(k, indexSearcher),
        segmentOrdToResults,
        knnQuery.getVectorDataType() == VectorDataType.FLOAT
            ? knnQuery.getQueryVector()
            : knnQuery.getByteQueryVector(),
        knnQuery.getField()
    );
```

# Fix: Select Collector Manager Based on Index Type
We should create the collector manager depending on whether the index is nested.

```
final KnnCollectorManager knnCollectorManager =
    knnQuery.getParentsFilter() == null
        ? new TopKnnCollectorManager(k, indexSearcher)
        : new DiversifyingNearestChildrenKnnCollectorManager(
              k,
              knnQuery.getParentsFilter(),
              indexSearcher
          );

final ReentrantKnnCollectorManager reentrantCollectorManager =
    new ReentrantKnnCollectorManager(
        knnCollectorManager,
        segmentOrdToResults,
        knnQuery.getVectorDataType() == VectorDataType.FLOAT
            ? knnQuery.getQueryVector()
            : knnQuery.getByteQueryVector(),
        knnQuery.getField()
    );
```

### Related Issues

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/common/RobustUniqueRandomIterator.java
+++ b/src/main/java/org/opensearch/knn/common/RobustUniqueRandomIterator.java
@@ -13,12 +13,12 @@ import static org.apache.lucene.search.DocIdSetIterator.NO_MORE_DOCS;
  * An iterator that provides unique random integers in the range [0, maxValExclusive)
  * using O(1) memory and O(1) complexity (amortized).
  *
- * <h3>Theory of Operation</h3>
+ * <h2>Theory of Operation</h2>
  * This class utilizes a <b>Linear Congruential Generator (LCG)</b> configured to produce
- * a "Full Period" (or Full Cycle). A standard LCG follows the recurrence relation:
- * <center>X_{n+1} = (a * X_n + c) mod M</center>
+ * a "Full Period" (or Full Cycle). A standard LCG follows the recurrence relation.
+ * Formula: X_{n+1} = (a * X_n + c) mod M
  *
- * <h4>1. The Hull-Dobell Theorem</h4>
+ * <h3>1. The Hull-Dobell Theorem</h3>
  * To ensure the generator visits every single number in the range [0, M) exactly once
  * before repeating (uniqueness), the parameters must satisfy:
  * <ul>
@@ -27,7 +27,7 @@ import static org.apache.lucene.search.DocIdSetIterator.NO_MORE_DOCS;
  * <li>If M is a multiple of 4, a - 1 must be a multiple of 4 (satisfied by a = 4k + 1).</li>
  * </ul>
  *
- * <h4>2. The "Sandbox" & Rejection Sampling</h4>
+ * <h3>2. The "Sandbox" &amp; Rejection Sampling</h3>
  * LCGs are most efficient when M is a power of two, as the modulo operation becomes a
  * bitwise {@code & (M - 1)}. However, the user's requested {@code maxValExclusive} might not
  * be a power of two.
@@ -35,10 +35,9 @@ import static org.apache.lucene.search.DocIdSetIterator.NO_MORE_DOCS;
  * This class solves this by finding the smallest power of two (M) that is greater than or
  * equal to the requested range. When the LCG generates a value X such that
  * {@code maxValExclusive <= X < M}, the value is "rejected," and the next value in the
- * sequence is generated. Because M < 2 * maxValExclusive, the expected number of
+ * sequence is generated. Because M &lt; 2 * maxValExclusive, the expected number of
  * iterations is less than 2, preserving O(1) amortized time.
  * </p>
- *
  */
 public final class RobustUniqueRandomIterator {
     /** The user's requested max value (exclusive). */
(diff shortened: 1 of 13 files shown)


src/main/java/org/opensearch/knn/common/RobustUniqueRandomIterator.java
```diff
@@ -0,0 +1,125 @@
...
+ * be a power of two.
+ * <p>
+ * This class solves this by finding the smallest power of two (M) that is greater than or
+ * equal to the requested range. When the LCG generates a value X such that
+ * {@code maxValExclusive <= X < M}, the value is "rejected," and the next value in the
+ * sequence is generated. Because M &lt; 2 * maxValExclusive, the expected number of
+ * iterations is less than 2, preserving O(1) amortized time.
+ * </p>
+ */
+public final class RobustUniqueRandomIterator {
```
0ctopus13prime, line 42: LCG based random number generation is almost four times faster than HashSet based deduplication.
Also it does not require additional memory.

HashSet based micro performance
```
        Min: 541 nano
        50%: 1959 nano
        90%: 3458 nano
        95%: 3750 nano
        99%: 5291 nano
        Max: 43125 nano
```

LCG based micro performance
```
        Min: 125 nano
        50%: 333 nano
        90%: 458 nano
        95%: 1417 nano
        99%: 1625 nano
        Max: 12458 nano
```

src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java
```diff
@@ -413,8 +415,12 @@ private void reentrantSearch(
 
         // Kick off 2nd search tasks
         if (secondDeepDiveTasks.isEmpty() == false) {
+            final KnnCollectorManager collectorManager = knnQuery.getParentsFilter() == null
+                ? new TopKnnCollectorManager(k, indexSearcher)
+                : new DiversifyingNearestChildrenKnnCollectorManager(k, knnQuery.getParentsFilter(), indexSearcher);
```
navneet1v, line 420: [Clarification needed]
I am little surprised that this is not just related to remote index build it can be for other search path too right?

0ctopus13prime: Unfortunately, yes. Not only for the Cagra case this was affecting all nested case.

navneet1v: Thank you. Now it makes more sense to me. :)

navneet1v (approved):

Vikasht34 (approved):

PR #3162, opened 2026-03-13
Fix copy_to functionality with vector fields.

### LINKED ISSUE #2636: CopyTo does not work with knn_vector fields
## Description

We expose a field parameter called "copy_to". This parameter is supposed to copy one field to another. The main purpose is for searching text where you typically want to search over a couple fields. See https://opensearch.org/docs/latest/field-types/mapping-parameters/copy-to/.

For vector fields, this does not work. For one thing, we do not support multi-fields, so its unclear how this would work, but I suspect that it would result in some kind of multi-field effect. For another thing, in parsing, we process tokens and go over them. However, it seems like copyTo will just parse the context and then copy it to the copy to context, and then call the parser again. See: 
1. https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/mapper/DocumentParser.java#L527-L528
2. https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/mapper/DocumentParser.java#L900-L951

This works for text fields which do not skip the text value (See: https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/mapper/TextFieldMapper.java#L1038). However, it does not work for vectors fields which we call next token for.

The error message ends up looking something like:
```
    {"error":{"root_cause":[{"type":"mapper_parsing_exception","reason":"failed to parse field [copyto_vector] of type [knn_vector] in document with id '1'. Preview of field's value: 'null'"}],"type":"mapper_parsing_exception","reason":"failed to parse field [copyto_vector] of type [knn_vector] in document with id '1'. Preview of field's value: 'null'","caused_by":{"type":"x_content_parse_exception","reason":"[1:76] Failed to parse list:  expecting START_ARRAY but got END_ARRAY"}},"status":400}

# With a mapping like:
    "mappings" : {
      "properties" : {
        "copyto_vector" : {
          "type" : "knn_vector",
          "dimension" : 16
        },
        "test_vector" : {
          "type" : "knn_vector",
          "dimension" : 16,
          "copy_to" : [
            "copyto_vector"
          ]
        }
      }
    },
```

krocky-cooky:
### Description
Previously, adding the copy_to attribute to a vector field would cause an error during indexing. The root cause of this issue is the same as the cause of geo_point issue(opensearch-project/OpenSearch#20540) and has been resolved by opensearch-project/OpenSearch#20542. This PR includes integration tests that verify the copy_to attribute works correctly on vector fields.

### Related Issues
Resolves #2636

### Check List

### DIFF
--- a/src/test/java/org/opensearch/knn/index/OpenSearchIT.java
+++ b/src/test/java/org/opensearch/knn/index/OpenSearchIT.java
@@ -1776,40 +1776,11 @@ public class OpenSearchIT extends KNNRestTestCase {
         String targetField1 = "target_vector_1";
         String targetField2 = "target_vector_2";
 
-        String mapping = XContentFactory.jsonBuilder()
-            .startObject()
-            .startObject("properties")
-            .startObject(sourceField)
-            .field("type", "knn_vector")
-            .field("dimension", 2)
-            .startObject("method")
-            .field(KNNConstants.NAME, KNNConstants.METHOD_HNSW)
-            .field(KNNConstants.METHOD_PARAMETER_SPACE_TYPE, SpaceType.L2.getValue())
-            .field(KNNConstants.KNN_ENGINE, KNNEngine.FAISS.getName())
-            .endObject()
-            .field("copy_to", new String[] { targetField1, targetField2 })
-            .endObject()
-            .startObject(targetField1)
-            .field("type", "knn_vector")
-            .field("dimension", 2)
-            .startObject("method")
-            .field(KNNConstants.NAME, KNNConstants.METHOD_HNSW)
-            .field(KNNConstants.METHOD_PARAMETER_SPACE_TYPE, SpaceType.L2.getValue())
-            .field(KNNConstants.KNN_ENGINE, KNNEngine.FAISS.getName())
-            .endObject()
-            .endObject()
-            .startObject(targetField2)
-            .field("type", "knn_vector")
-            .field("dimension", 2)
-            .startObject("method")
-            .field(KNNConstants.NAME, KNNConstants.METHOD_HNSW)
-            .field(KNNConstants.METHOD_PARAMETER_SPACE_TYPE, SpaceType.L2.getValue())
-            .field(KNNConstants.KNN_ENGINE, KNNEngine.LUCENE.getName())
-            .endObject()
-            .endObject()
-            .endObject()
-            .endObject()
-            .toString();
+        XContentBuilder builder = XContentFactory.jsonBuilder().startObject().startObject("properties");
+        buildKnnVectorFieldMapping(builder, sourceField, 2, SpaceType.L2, KNNEngine.FAISS, new String[] { targetField1, targetField2 });
+        buildKnnVectorFieldMapping(builder, targetField1, 2, SpaceType.L2, KNNEngine.FAISS, null);
+        buildKnnVectorFieldMapping(builder, targetField2, 2, SpaceType.L2, KNNEngine.LUCENE, null);
+        String mapping = builder.endObject().endObject().toString();
 
         createKnnIndex(indexName, mapping);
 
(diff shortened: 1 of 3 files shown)


src/test/java/org/opensearch/knn/index/OpenSearchIT.java
```diff
@@ -1769,6 +1769,275 @@ public void testKNNSearchWithProfilerEnabled_Radial() throws Exception {
...
     }
 
+    @SneakyThrows
+    public void testCopyTo_whenSearchOnTargetField_thenSuccess() {
+        String indexName = "test_copy_to_search";
+        String sourceField = "source_vector";
+        String targetField1 = "target_vector_1";
+        String targetField2 = "target_vector_2";
+
+        String mapping = XContentFactory.jsonBuilder()
```
kotwanikunal: There is a lot of duplication with the mapping creation logic. Can you please pull it out into a helper method or extend an existing one?

src/test/java/org/opensearch/knn/index/OpenSearchIT.java
```diff
@@ -1769,6 +1769,275 @@ public void testKNNSearchWithProfilerEnabled_Radial() throws Exception {
...
+            .startObject("method")
+            .field(KNNConstants.NAME, KNNConstants.METHOD_HNSW)
+            .field(KNNConstants.METHOD_PARAMETER_SPACE_TYPE, SpaceType.L2.getValue())
+            .field(KNNConstants.KNN_ENGINE, KNNEngine.FAISS.getName())
+            .endObject()
+            .field("copy_to", new String[] { targetField1, targetField2 })
+            .endObject()
+            .startObject(targetField1)
+            .field("type", "knn_vector")
+            .field("dimension", 2)
```
kotwanikunal: Can you add in a case where the dimensions across the mappings mismatch?

src/test/java/org/opensearch/knn/index/OpenSearchIT.java
```diff
@@ -1769,6 +1769,275 @@ public void testKNNSearchWithProfilerEnabled_Radial() throws Exception {
...
+        float[] queryVector = { 1.0f, 1.0f };
+        List<KNNResult> results1 = getResults(indexName, targetField1, queryVector, 1);
+        assertEquals(1, results1.size());
+        assertEquals("1", results1.get(0).getDocId());
+
+        List<KNNResult> results2 = getResults(indexName, targetField2, queryVector, 1);
+        assertEquals(1, results2.size());
+        assertEquals("1", results2.get(0).getDocId());
+
+        deleteKNNIndex(indexName);
```
kotwanikunal, line 1822: Please wrap the delete behind a finally block - that should avoid any flakiness.

src/test/java/org/opensearch/knn/index/OpenSearchIT.java
```diff
@@ -1769,6 +1769,275 @@ public void testKNNSearchWithProfilerEnabled_Radial() throws Exception {
         deleteKNNIndex(INDEX_NAME);
     }
 
+    @SneakyThrows
+    public void testCopyTo_whenSearchOnTargetField_thenSuccess() {
```
kotwanikunal, line 1791: Can you also please add in a new test within `DerivedSourceIT` with the copy to functionality? That path is crucial and we should validate copyTo works with derived source enabled before closing out the issue

kotwanikunal (commented):
@krocky-cooky - Thanks for raising the PR for testing the `copy_to` functionality. Before we can close out the linked issue - we need some additional tests with DerivedSource enabled as well as non happy path test cases.

src/test/java/org/opensearch/knn/index/OpenSearchIT.java
krocky-cooky: Added!

src/test/java/org/opensearch/knn/index/OpenSearchIT.java
krocky-cooky: Since ODFERestTestCase already has index cleanup logic in its @After method ([wipeAllODFEIndices](https://github.com/opensearch-project/k-NN/blob/afa7c5c749ce96a978aaaa06ea08e631ef6e93e5/src/testFixtures/java/org/opensearch/knn/ODFERestTestCase.java#L153))
, I believe flakiness should be mitigated. That said, would you still prefer adding try-finally 
blocks for clarity? My concern is that it would increase nesting depth in the test methods.

src/test/java/org/opensearch/knn/index/OpenSearchIT.java
krocky-cooky: Added non happy path test cases!

kotwanikunal (approved):

navneet1v (dismissed):

navneet1v (approved):

navneet1v (approved):

PR #3164, opened 2026-03-13
Add MUVERA processors for multi-vector search

### LINKED ISSUE #3163: [FEATURE] Add MUVERA ingest and search processors for multi-vector ANN prefetch

Multi-vector retrieval models like ColBERT and ColPali produce per-token embeddings that require MaxSim scoring across all token pairs. This is expensive at scale because there's no way to do ANN prefetch on variable-length multi-vector representations — you're forced to either brute-force score every document or rely on external tooling to pre-encode vectors client-side.

Currently, the k-NN plugin supports `[lateInteractionScore](https://docs.opensearch.org/latest/query-dsl/specialized/script-score/#late-interaction-score)` for MaxSim reranking, but the inner query is typically `match_all` or a text filter, meaning every matching document gets scored. There's no native way to narrow candidates using the multi-vector embeddings themselves.

Add two new processors implementing the MUVERA algorithm (Multi-Vector Retrieval via Fixed Dimensional Encodings, [[paper](https://arxiv.org/abs/2405.19504)](https://arxiv.org/abs/2405.19504)):

1. **`muvera` ingest processor** — Converts variable-length multi-vector embeddings into a single fixed-dimensional encoding (FDE) vector using SimHash clustering and random projections. The FDE is stored in a `knn_vector` field for ANN indexing. The original multi-vectors remain in `_source` for reranking.

2. **`muvera_query` search request processor** — Intercepts `script_score` queries containing `query_vectors` in script params, MUVERA-encodes them into an FDE, and replaces the inner `match_all` with a `knn` query on the FDE field. The `lateInteractionScore` script wrapper stays intact for MaxSim reranking on the prefetched candidates.

### User flow

**Step 1: Create ingest pipeline**

```json
PUT _ingest/pipeline/muvera-ingest
{
  "description": "MUVERA FDE encoding for ColBERT vectors",
  "processors": [
    {
      "muvera": {
        "source_field": "colbert_vectors",
        "target_field": "muvera_fde",
        "dim": 128,
        "fde_dimension": 2560
      }
    }
  ]
}
```

Defaults: `k_sim=4`, `dim_proj=8`, `r_reps=20`, `seed=42`. FDE dimension = `r_reps * 2^k_sim * dim_proj` = 2560. The `fde_dimension` parameter validates the computed value so the user explicitly acknowledges the output size.

**Step 2: Create index**

```json
PUT muvera-index
{
  "settings": {
    "index.knn": true,
    "default_pipeline": "muvera-ingest"
  },
  "mappings": {
    "dynamic": false,
    "properties": {
      "muvera_fde": {
        "type": "knn_vector",
        "dimension": 2560,
        "method": {
          "name": "hnsw",
          "space_type": "innerproduct",
          "engine": "faiss"
        }
      },
      "title": { "type": "text" }
    }
  }
}
```

Note: `colbert_vectors` is intentionally left unmapped, it stays in `_source` for reranking but doesn't need its own field mapping.

**Step 3: Index documents**

```json
POST muvera-index/_doc/1
{
  "title": "example document",
  "colbert_vectors": [
    [0.1, 0.2, ...],
    [0.3, 0.4, ...],
    [0.5, 0.6, ...]
  ]
}
```

The ingest processor reads `colbert_vectors`, produces the FDE, and writes it to `muvera_fde`. Both fields end up in the stored document.

**Step 4: Create search pipeline**

```json
PUT _search/pipeline/muvera-search
{
  "request_processors": [
    {
      "muvera_query": {
        "target_field": "muvera_fde",
        "dim": 128,
        "fde_dimension": 2560,
        "oversample_factor": 4
      }
    }
  ]
}
```

Same MUVERA hyperparams as ingest (must match). `oversample_factor` controls how many candidates the knn prefetch retrieves relative to the requested result size.

**Step 5: Search**

```json
POST muvera-index/_search?search_pipeline=muvera-search
{
  "size": 10,
  "query": {
    "script_score": {
      "query": { "match_all": {} },
      "script": {
        "source": "lateInteractionScore(params.query_vectors, 'colbert_vectors', params._source, params.space_type)",
        "params": {
          "query_vectors": [[0.1, 0.2, ...], [0.3, 0.4, ...]],
          "space_type": "innerproduct"
        }
      }
    }
  }
}
```

What happens:
1. Search processor extracts `query_vectors` from script params
2. MUVERA-encodes them into a query FDE
3. Replaces `match_all` with `knn` on `muvera_fde` (k = size × oversample_factor = 40)
4. `lateInteractionScore` reranks the 40 candidates using exact MaxSim on original multi-vectors
5. Top 10 returned to user

- Client-side MUVERA encoding (works but requires users to maintain encoding logic outside OpenSearch)
- Binary quantization of multi-vectors (lossy, doesn't preserve MaxSim structure)
- Text-based prefetch with BM25 (misses semantic signal from embeddings)

- MUVERA is already implemented in [[fastembed](https://github.com/qdrant/fastembed)](https://github.com/qdrant/fastembed) (Python) and used in production with [Qdrant](https://qdrant.tech/articles/muvera-embeddings/)
- We have a working implementation with unit tests, stable across multiple iterations with random seeds
- The implementation uses only public APIs — no reflection or core OpenSearch modifications required
- Tested end-to-end on a live cluster: ingest pipeline creates FDE vectors, search pipeline rewrites queries, `lateInteractionScore` reranking produces correct MaxSim scores

praveenMprasad:
### Description

Add MUVERA (Multi-Vector Retrieval via Fixed Dimensional Encodings) ingest and search request processors to enable ANN prefetch for multi-vector models like ColBERT and ColPali.

The MUVERA algorithm converts variable-length multi-vector embeddings into a single fixed-dimensional encoding (FDE) vector using SimHash clustering and random projections. This enables approximate nearest neighbor search on multi-vector representations, which previously required brute-force scoring via `lateInteractionScore`.

**Two new processors:**

- `muvera` ingest processor — reads multi-vectors from a source field, produces an FDE vector, writes it to a `knn_vector` target field. Original multi-vectors remain in `_source` for reranking.
- `muvera_query` search request processor — intercepts `script_score` queries containing `query_vectors` in script params, MUVERA-encodes them, and replaces the inner `match_all` with a `knn` query on the FDE field. The `lateInteractionScore` script wrapper stays intact for MaxSim reranking.

**Key design decisions:**
- All hyperparameters have sensible defaults (k_sim=4, dim_proj=8, r_reps=20, seed=42)
- FDE dimension is computed and logged at pipeline creation time
- Optional `fde_dimension` parameter validates against computed value
- Tested end-to-end on a live cluster with correct MaxSim reranking scores

Reference: [MUVERA paper](https://arxiv.org/abs/2405.19504), [fastembed implementation](https://github.com/qdrant/fastembed)

### Related Issues
Resolves #3163

### Check List
  - 34 unit tests across 3 test classes, stable across 10 iterations with random seeds
  - Documentation PR to follow after initial review
  - No REST API changes — processors are configured via existing ingest/search pipeline APIs
  - Will create after initial review


navneet1v:
@praveenMprasad before we can kickoff the review lets ensure that all CIs are passing.

cc: @Gankris96 and @kotwanikunal

src/main/java/org/opensearch/knn/plugin/KNNPlugin.java
```diff
@@ -504,4 +508,16 @@ public Map<String, SystemGeneratedProcessor.SystemGeneratedFactory<SearchRespons
     ) {
         return Map.of(MMRRerankProcessor.MMRRerankProcessorFactory.TYPE, new MMRRerankProcessor.MMRRerankProcessorFactory());
     }
+
+    @Override
+    public Map<String, org.opensearch.ingest.Processor.Factory> getProcessors(org.opensearch.ingest.Processor.Parameters parameters) {
```
0ctopus13prime: NIT : Could you not use full package path in here?
org.opensearch.ingest.Processor -> Processor.

src/main/java/org/opensearch/knn/processor/muvera/MuveraIngestProcessor.java
```diff
@@ -0,0 +1,281 @@
...
+                "MUVERA encoder produced FDE of dimension ["
+                    + fde.length
+                    + "] but expected ["
+                    + fdeDimension
+                    + "]. This should not happen — please report this as a bug."
+            );
+        }
+
+        // Convert float[] to List<Float> for IngestDocument compatibility
+        List<Float> fdeList = new java.util.ArrayList<>(fde.length);
```
0ctopus13prime: Nit : java.util.ArrayList.

praveenMprasad: Done

src/main/java/org/opensearch/knn/plugin/KNNPlugin.java
praveenMprasad: Done, Thanks.

src/main/java/org/opensearch/knn/processor/muvera/MuveraEncoder.java
```diff
@@ -0,0 +1,196 @@
...
+        }
+        if (rReps <= 0) {
+            throw new IllegalArgumentException("r_reps must be positive, got: " + rReps);
+        }
+
+        this.dim = dim;
+        this.kSim = kSim;
+        this.dimProj = dimProj;
+        this.rReps = rReps;
+        this.numPartitions = 1 << kSim;
```
Gankris96, line 107: we should probably consider adding an upper bound check for kSim too. this ksim controlls the FDE dimensionality and if kSim is too large, i think it might actually be blocked by k-NN max dimensionality check. 
https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/engine/KNNEngine.java#L47-L52 

This will fail for some kSim values.

src/main/java/org/opensearch/knn/processor/muvera/MuveraSearchRequestProcessor.java
```diff
@@ -0,0 +1,363 @@
...
+                    "required property is missing. Set this to your embedding model's vector dimension (e.g. 128)."
+                );
+            }
+            int dim = dimValue;
+
+            int kSim = ConfigurationUtils.readIntProperty(TYPE, tag, config, "k_sim", 4);
+            int dimProj = ConfigurationUtils.readIntProperty(TYPE, tag, config, "dim_proj", 8);
+            int rReps = ConfigurationUtils.readIntProperty(TYPE, tag, config, "r_reps", 20);
+            long seed = readLongProperty(TYPE, tag, config, "seed", 42L);
+            int oversampleFactor = ConfigurationUtils.readIntProperty(TYPE, tag, config, "oversample_factor", 4);
```
Gankris96: Can we have validation for this? This should have validation, `>0`

src/main/java/org/opensearch/knn/processor/muvera/MuveraEncoder.java
praveenMprasad: @Gankris96 Added FDE dimension upper bound check (16,000, matching k-NN engine max). The encoder now validates rReps * 2^kSim * dimProj <= 16,000 at construction time with a descriptive error message. Please check commit ba68cbdc.

src/main/java/org/opensearch/knn/processor/muvera/MuveraSearchRequestProcessor.java
praveenMprasad: Added oversample_factor > 0 validation in the search processor factory. Also confirmed all other hyperparams are validated — dim, k_sim, dim_proj, r_reps are checked in the MuveraEncoder constructor. Please check commit ba68cbdc.

src/main/java/org/opensearch/knn/processor/muvera/MuveraEncoder.java
```diff
@@ -0,0 +1,221 @@
...
+    private final int rReps;
+    private final int numPartitions;
+    private final double[][] simhashVectors;
+    private final double[][][] dimReductionProjections;
+
+    /**
+     * Maximum allowed FDE dimension. Matches the k-NN engine max dimension limit (16,000)
+     * to ensure the FDE vector can be indexed in any supported engine.
+     */
+    static final int MAX_FDE_DIMENSION = 16_000;
```
mingshl: is there chances people want to configure the max dimension? this can be a configuration setting in the processor ?

praveenMprasad: @mingshl The 16,000 cap mirrors the k-NN engine's max vector dimension, going above it would fail at indexing regardless. Users already control the actual FDE size through the MUVERA parameters (dim, k_sim, dim_proj, r_reps). Adding a configurable max would let users set a tighter bound, but it feels like an edge case since the engine enforces the real ceiling. Happy to add it as an optional max_fde_dimension processor config if you think it's worth it though.

praveenMprasad:
## MUVERA Benchmark Results — nfcorpus + ColBERTv2

Benchmarked on BeIR nfcorpus (3,633 docs, 323 queries) with ColBERTv2 multi-vector embeddings.

**MUVERA params:** `k_sim=5`, `dim_proj=16`, `r_reps=20`, FDE dim=10,240

**Index:** faiss HNSW, `m=16`, `ef_construction=512`, `ef_search=512`, 1 shard, 1 segment

| Approach | NDCG@1 | NDCG@5 | NDCG@10 | % of Exact MaxSim |
| Exact MaxSim (brute-force) | 0.483 | 0.386 | 0.344 | 100% |
| MUVERA + MaxSim rerank (4x) | 0.474 | 0.379 | 0.339 | 98.5% |
| MUVERA-only | 0.464 | 0.369 | 0.329 | 95.6% |
| Mean pool + MaxSim rerank | 0.249 | 0.183 | 0.145 | 42.2% |

MUVERA+rerank recovers **98.5%** of exact MaxSim quality. MUVERA-only gets **95.6%** without any reranking. Mean pooling (the naive alternative) retains only **42%** — demonstrating why MUVERA is needed.

Full benchmark code and results: https://github.com/praveenMprasad/OpenSearch-k-NN/blob/feature/muvera-benchmark/benchmarks/muvera/BENCHMARK_RESULTS.md

src/main/java/org/opensearch/knn/processor/muvera/MuveraEncoder.java
```diff
@@ -0,0 +1,221 @@
...
+ *
+ * The random seed must be identical between index time and query time to produce compatible encodings.
+ */
+public class MuveraEncoder {
+    private final int dim;
+    private final int kSim;
+    private final int dimProj;
+    private final int rReps;
+    private final int numPartitions;
+    private final double[][] simhashVectors;
```
kotwanikunal: Any particular reason to use double instead of float? The return type signatures are all floats.
The operations for similarity match can benefit with SIMD/Panama calculations and the memory consumption can also reduce by half.

src/main/java/org/opensearch/knn/processor/muvera/MuveraSearchRequestProcessor.java
```diff
@@ -0,0 +1,371 @@
...
+                    + queryFde.length
+                    + "] but expected ["
+                    + fdeDimension
+                    + "]. This should not happen — please report this as a bug."
+            );
+        }
+
+        // Determine result size from the request
+        int resultSize = request.source().size() > 0 ? request.source().size() : 10;
+        int prefetchK = resultSize * oversampleFactor;
```
kotwanikunal: Can you limit this to the some considerable size?
```
int prefetchK = Math.min(resultSize * oversampleFactor, 10_000); 
```

src/main/java/org/opensearch/knn/processor/muvera/MuveraSearchRequestProcessor.java
```diff
@@ -0,0 +1,371 @@
...
+                    parser.skipChildren();
+                }
+            }
+        }
+        throw new IllegalStateException("Failed to extract script from ScriptScoreQueryBuilder via XContent serialization");
+    }
+
+    @Override
+    @SuppressWarnings("unchecked")
+    public SearchRequest processRequest(SearchRequest request) throws Exception {
```
kotwanikunal: Please break this method down into smaller methods 

```
validateUserRequest()
extractRequestParams()
createKnnRequest()
```

src/main/java/org/opensearch/knn/processor/muvera/MuveraSearchRequestProcessor.java
```diff
@@ -0,0 +1,371 @@
...
+            return request;
+        }
+
+        // Extract multi-vectors from script params
+        Object queryVectorsObj = params.get(QUERY_VECTORS_PARAM);
+        if (queryVectorsObj instanceof List == false) {
+            throw new IllegalArgumentException("[" + QUERY_VECTORS_PARAM + "] in script params must be a list of vectors");
+        }
+
+        List<List<Number>> queryVectorsList = (List<List<Number>>) queryVectorsObj;
```
kotwanikunal: This is unsafe. This will lead to a cast exception - can you add validations before raw conversions?

src/main/java/org/opensearch/knn/processor/muvera/MuveraIngestProcessor.java
```diff
@@ -0,0 +1,282 @@
...
+                    + dimProj
+                    + ", seed="
+                    + seed
+                    + ")";
+            }
+
+            return new MuveraIngestProcessor(tag, description, sourceField, targetField, encoder, dim, computedDimension);
+        }
+
+        private static long readLongProperty(
```
kotwanikunal: This method seems to be duplicated across classes. Move this to a utility class.

src/test/java/org/opensearch/knn/processor/muvera/MuveraIngestProcessorTests.java
```diff
@@ -0,0 +1,183 @@
...
+        assertTrue(e.getMessage().contains("dim"));
+    }
+
+    public void testFactoryThrowsOnFdeDimensionMismatch() {
+        MuveraIngestProcessor.Factory factory = new MuveraIngestProcessor.Factory();
+
+        Map<String, Object> config = new HashMap<>();
+        config.put("source_field", SOURCE_FIELD);
+        config.put("target_field", TARGET_FIELD);
+        config.put("dim", 4);
```
mingshl, line 162: every processor has ignoreMissing and ignoreFailure flag, would you add the UT and verify the edge cases? 

for example,
in two documents, one has the source field, and the the second document doesn't have field.

when setting muvera pipeline when ignoreMissing is true, the pipeline should success even when one document has missing source field. For two documents, one document has muvera embedding target field, the second document doesn't have the embedding target field

when setting muvera pipeline when ignoreMissing is false, the second document should fail. 

you might want to write down these four combinations down for ingest and search pipeline in the PR description. so that it make sense for us to review in these different configurations. And I would love to see the logic verified in the Unit tests.

src/test/java/org/opensearch/knn/integ/MuveraProcessorIT.java
```diff
@@ -0,0 +1,436 @@
...
+    private static final int R_REPS = 2;
+    private static final int FDE_DIM = R_REPS * (1 << K_SIM) * DIM_PROJ; // 8
+    private static final long SEED = 42L;
+
+    /**
+     * Tests the full MUVERA ingest + search pipeline flow end-to-end.
+     * Creates pipelines, indexes documents with multi-vectors, and searches
+     * using script_score with lateInteractionScore reranking over ANN prefetch.
+     */
+    public void testMuveraEndToEnd_whenIngestAndSearch_thenReturnsResults() throws Exception {
```
kotwanikunal, line 48: Test cases are lacking. Can you create an at scale test and ensure the complete functionality (ordering etc) is appropriate?

src/test/java/org/opensearch/knn/processor/muvera/MuveraIngestProcessorTests.java
praveenMprasad: @mingshl Thanks for the review! Added `ignore_missing` support to the ingest processor with unit tests covering the combinations.

**Ingest Pipeline (`muvera` processor):**

| `ignore_missing` | Source field present | Behavior |
| --- | --- | --- |
| `false` (default) | Yes | Produces FDE in target field |
| `false` | No / null | Throws `IllegalArgumentException` |
| `true` | Yes | Produces FDE in target field |
| `true` | No / null | Skips silently, no target field added |

Two-document scenario is also tested: with `ignore_missing=true`, doc1 (has source) gets FDE, doc2 (missing source) is skipped. With `ignore_missing=false`, doc2 throws.

**Search Pipeline (`muvera_query` processor):**

| `ignore_failure` | Query has `script_score` with `query_vectors` | Behavior |
| --- | --- | --- |
| `false` (default) | Yes | Rewrites query with FDE KNN |
| `false` | No | Returns request unchanged (no error) |
| `true` | Yes | Rewrites query with FDE KNN |
| `true` | No / malformed | Exception caught by framework, original request passed through |

The search processor already handles `ignore_failure` via OpenSearch's `AbstractProcessor` base class — the pipeline framework catches exceptions when `ignore_failure=true`. The processor itself returns the request unchanged if the query isn't a `script_score` type, so missing/non-matching queries don't error.

**Two-document scenario (unit tested + integration tested):**

- `ignore_missing=true`: Doc1 (has `colbert_vectors`) → gets `muvera_fde` target field. Doc2 (missing `colbert_vectors`) → indexed successfully, no `muvera_fde` field added.
- `ignore_missing=false`: Doc1 → gets `muvera_fde`. Doc2 → throws `IllegalArgumentException`, indexing fails.

Unit tests: `testTwoDocuments_IgnoreMissingTrue_OneWithSourceOneWithout`, `testTwoDocuments_IgnoreMissingFalse_SecondDocFails`

Integration tests: `testMuveraIngest_whenIgnoreMissingTrue_thenMixedDocsSucceed`, `testMuveraIngest_whenIgnoreMissingFalse_thenMissingSourceFails`


See commit: e1e40468

src/main/java/org/opensearch/knn/processor/muvera/MuveraEncoder.java
praveenMprasad: Good point on the memory savings. The internal arrays use double for numerical precision during accumulation — with ~1000 vectors per document, the cluster center sums can get large and float accumulation could introduce meaningful rounding errors. The output is cast to float at the end since that's what the knn_vector field stores.

That said, for the simhash and projection matrices (which are generated once at init), float would be fine. I'll convert those to float in a follow-up to reduce the encoder's memory footprint. The cluster center accumulation should stay double to avoid precision loss during summation.

Open to converting everything to float if you think the precision tradeoff is acceptable — happy to benchmark the quality impact.

src/main/java/org/opensearch/knn/processor/muvera/MuveraSearchRequestProcessor.java
praveenMprasad: Good catch — added the cap. `int prefetchK = Math.min(resultSize * oversampleFactor, 10_000);`

src/main/java/org/opensearch/knn/processor/muvera/MuveraSearchRequestProcessor.java
praveenMprasad: Refactored into three methods: `validateUserRequest()`, `extractRequestParams()`, `createKnnRequest()`.

src/main/java/org/opensearch/knn/processor/muvera/MuveraSearchRequestProcessor.java
praveenMprasad: Fixed — now validates each element type before casting: outer list → inner list → Number. No more unchecked cast.

src/main/java/org/opensearch/knn/processor/muvera/MuveraIngestProcessor.java
praveenMprasad: Moved to `MuveraProcessorUtils` utility class, both processors now reference it.

src/test/java/org/opensearch/knn/integ/MuveraProcessorIT.java
praveenMprasad: Added two new integration tests:
- `testMuveraEndToEnd_whenSearchWithKnownVectors_thenOrderingIsCorrect`: indexes 3 docs with known vectors (close/medium/far), queries, verifies scores are in descending order and the closest doc ranks first.
- `testMuveraIngest_whenIgnoreMissingTrue_thenMixedDocsSucceed`: indexes 2 docs through pipeline with `ignore_missing=true` — one with vectors (gets FDE), one without (skipped silently). Verifies both are indexed.

src/main/java/org/opensearch/knn/processor/muvera/MuveraEncoder.java
kotwanikunal: > for the simhash and projection matrices (which are generated once at init), float would be fine. I'll convert those to float in a follow-up to reduce the encoder's memory footprint. The cluster center accumulation should stay double to avoid precision loss during summation.

I think this makes sense. Let's reduce the footprint as much as we can

src/main/java/org/opensearch/knn/processor/muvera/MuveraEncoder.java
```diff
@@ -0,0 +1,221 @@
...
+     * @return FDE vector of length getEmbeddingSize()
+     */
+    public float[] process(double[][] vectors, boolean fillEmpty, boolean normalize) {
+        int nVectors = vectors.length;
+        float[] output = new float[getEmbeddingSize()];
+        int outOffset = 0;
+        double scale = 1.0 / Math.sqrt(dimProj);
+
+        for (int r = 0; r < rReps; r++) {
+            double[][] centers = new double[numPartitions][dim];
```
kotwanikunal: This seems too heavy. Can we optimize this by reusing the arrays somehow? Allocation for every loop can get expensive.

src/main/java/org/opensearch/knn/processor/muvera/MuveraSearchRequestProcessor.java
```diff
@@ -0,0 +1,381 @@
...
+                "MUVERA encoder produced query FDE of dimension ["
+                    + queryFde.length
+                    + "] but expected ["
+                    + fdeDimension
+                    + "]. This should not happen — please report this as a bug."
+            );
+        }
+
+        int resultSize = request.source().size() > 0 ? request.source().size() : 10;
+        int prefetchK = Math.min(resultSize * oversampleFactor, 10_000);
```
kotwanikunal: Thanks for adding this check in. Please add comments explaining why it's restricted.

src/main/java/org/opensearch/knn/processor/muvera/MuveraEncoder.java
praveenMprasad: Converted simhash and projection matrices to float. Cluster center accumulation stays double to avoid precision loss during summation. See latest commit.

src/main/java/org/opensearch/knn/processor/muvera/MuveraEncoder.java
praveenMprasad: Pre-allocated centers, counts, and clusterVecIndices arrays outside the loop and reset them per repetition. No more per-loop allocation.

src/main/java/org/opensearch/knn/processor/muvera/MuveraSearchRequestProcessor.java
praveenMprasad: Added comment explaining the cap

praveenMprasad:
Thanks for the suggestion @mingshl  / @vamshin. I've switched the search processor to use a template query instead of the script_score + match_all pattern. The user now sends:

```
{
  "query": {
    "template": {
      "script_score": {
        "query": { "knn": { "muvera_fde": { "vector": "${muvera_fde}", "k": 40 } } },
        "script": {
          "source": "lateInteractionScore(params.query_vectors, 'colbert_vectors', params._source, params.space_type)",
          "params": { "query_vectors": [[...]], "space_type": "innerproduct" }
        }
      }
    }
  }
}
```
The processor extracts query_vectors from the template content, encodes the FDE, and sets it as a PipelineProcessingContext attribute keyed on target_field. The template query resolves ${muvera_fde} during doRewrite using those context attributes, so the KNN query runs with the encoded FDE and script_score reranks the prefetched candidates.

Unit tests (MuveraSearchRequestProcessorTests) and integration tests (MuveraProcessorIT) have been updated for the new format and are passing.

src/main/java/org/opensearch/knn/processor/muvera/MuveraIngestProcessor.java
```diff
@@ -0,0 +1,269 @@
...
+ * With defaults (r_reps=20, k_sim=4, dim_proj=8): 20 * 16 * 8 = 2560.
+ * Use this value as the "dimension" in your knn_vector field mapping.
+ *
+ * The computed FDE dimension is logged at pipeline creation time and included
+ * in the processor description visible via GET _ingest/pipeline.
+ */
+public class MuveraIngestProcessor extends AbstractProcessor {
+
+    public static final String TYPE = "muvera";
+    private static final Logger logger = LogManager.getLogger(MuveraIngestProcessor.class);
```
navneet1v: use @Log4j here

src/main/java/org/opensearch/knn/processor/muvera/MuveraEncoder.java
```diff
@@ -0,0 +1,229 @@
...
+    }
+
+    /**
+     * Returns the output FDE dimension: rReps * 2^kSim * dimProj.
+     */
+    public int getEmbeddingSize() {
+        return rReps * numPartitions * dimProj;
+    }
+
+    private int getClusterId(double[] vector, int repIndex) {
```
navneet1v: lets use float[] wherever possible rather than double[]

src/main/java/org/opensearch/knn/processor/muvera/MuveraEncoder.java
```diff
@@ -0,0 +1,229 @@
...
+        int clusterId = 0;
+        for (int k = 0; k < kSim; k++) {
+            float dot = 0;
+            int offset = k * dim;
+            for (int d = 0; d < this.dim; d++) {
+                dot += (float) vector[d] * simhashVectors[repIndex][offset + d];
+            }
```
navneet1v: use SIMD here to ensure that we are able to do this FMA faster

src/main/java/org/opensearch/knn/processor/muvera/MuveraEncoder.java
```diff
@@ -0,0 +1,229 @@
...
+     * @return FDE vector of length getEmbeddingSize()
+     */
+    public float[] process(double[][] vectors, boolean fillEmpty, boolean normalize) {
+        int nVectors = vectors.length;
+        float[] output = new float[getEmbeddingSize()];
+        int outOffset = 0;
+        float scale = (float) (1.0 / Math.sqrt(dimProj));
+
+        // Pre-allocate reusable arrays to avoid per-repetition allocation
+        double[][] centers = new double[numPartitions][dim];
```
navneet1v: lets use float[][] here

src/main/java/org/opensearch/knn/processor/muvera/MuveraEncoder.java
```diff
@@ -0,0 +1,229 @@
...
+        int[] counts = new int[numPartitions];
+        List<List<Integer>> clusterVecIndices = new ArrayList<>(numPartitions);
```
navneet1v: why we are not using a 2D array here instead of List<List<Integer>>

src/main/java/org/opensearch/knn/processor/muvera/MuveraEncoder.java
```diff
@@ -0,0 +1,229 @@
...
+        for (int i = 0; i < numPartitions; i++) {
+            clusterVecIndices.add(new ArrayList<>());
+        }
+
+        for (int r = 0; r < rReps; r++) {
+            // Reset reusable arrays
+            for (int c = 0; c < numPartitions; c++) {
+                java.util.Arrays.fill(centers[c], 0.0);
```
navneet1v: import Arrays rather than using `java.util.Arrays`

src/main/java/org/opensearch/knn/processor/muvera/MuveraEncoder.java
```diff
@@ -0,0 +1,229 @@
...
+            }
+
+            // Assign vectors to clusters and accumulate
+            for (int v = 0; v < nVectors; v++) {
+                int cid = getClusterId(vectors[v], r);
+                counts[cid]++;
+                clusterVecIndices.get(cid).add(v);
+                for (int d = 0; d < dim; d++) {
+                    centers[cid][d] += vectors[v][d];
+                }
```
navneet1v: see if we can use SIMD here

src/main/java/org/opensearch/knn/processor/muvera/MuveraEncoder.java
```diff
@@ -0,0 +1,229 @@
...
+            // Normalize by count (document mode)
+            if (normalize) {
+                for (int c = 0; c < numPartitions; c++) {
+                    if (counts[c] > 0) {
+                        for (int d = 0; d < dim; d++) {
+                            centers[c][d] /= counts[c];
+                        }
+                    }
+                }
+            }
```
navneet1v, line 244: same as above

src/main/java/org/opensearch/knn/processor/muvera/MuveraIngestProcessor.java
```diff
@@ -0,0 +1,269 @@
...
+
+        @Override
+        public MuveraIngestProcessor create(
+            Map<String, Processor.Factory> registry,
+            String tag,
+            String description,
+            Map<String, Object> config
+        ) throws Exception {
+            String sourceField = ConfigurationUtils.readStringProperty(TYPE, tag, config, "source_field");
+            String targetField = ConfigurationUtils.readStringProperty(TYPE, tag, config, "target_field");
```
navneet1v: move all these strings to private static final field

src/main/java/org/opensearch/knn/processor/muvera/MuveraIngestProcessor.java
```diff
@@ -0,0 +1,269 @@
...
+                "[{}] processor [{}]: computed FDE dimension = {} (r_reps={} * 2^k_sim={} * dim_proj={}). "
+                    + "Use this as the 'dimension' in your knn_vector field mapping for [{}].",
+                TYPE,
+                tag,
+                computedDimension,
+                rReps,
+                (1 << kSim),
+                dimProj,
+                targetField
+            );
```
navneet1v: doing this log will add the log at the cluster level. They will not be able to see it.

src/main/java/org/opensearch/knn/processor/muvera/MuveraProcessorUtils.java
```diff
@@ -0,0 +1,48 @@
...
+package org.opensearch.knn.processor.muvera;
+
+import java.util.Map;
+
+/**
+ * Shared utility methods for MUVERA ingest and search processors.
+ */
+final class MuveraProcessorUtils {
+
+    private MuveraProcessorUtils() {}
```
navneet1v: You can use @NoArgsConstructor(access = PACKAGE.PRIVATE)

src/main/java/org/opensearch/knn/processor/muvera/MuveraSearchRequestProcessor.java
```diff
@@ -0,0 +1,349 @@
...
+                    + fdeDimension
+                    + "]. This should not happen — please report this as a bug."
+            );
+        }
+
+        // Convert float[] to List<Float> for JSON serialization in template resolution
+        List<Float> fdeList = new ArrayList<>(queryFde.length);
+        for (float v : queryFde) {
+            fdeList.add(v);
+        }
```
navneet1v: why we need this conversion? we should try avoiding this.

src/main/java/org/opensearch/knn/processor/muvera/MuveraSearchRequestProcessor.java
```diff
@@ -0,0 +1,349 @@
...
+        return multiVectors;
+    }
+
+    @Override
+    public String getType() {
+        return TYPE;
+    }
+
+    public static class Factory implements Processor.Factory<SearchRequestProcessor> {
```
navneet1v, line 259: for all public classes and functions please add the java docs

navneet1v (commented):
Please remove jni folder from the PR and also fix the conflicts.

praveenMprasad:
> Please remove jni folder from the PR and also fix the conflicts.

@navneet1v  thanks for the detailed review. Below is a per-comment response, all addressed in commits 7bd886d4 and 93ee708f. Verified end-to-end on a live cluster — identical scores to the prior path, no correctness regression.

1. Remove jni folder + fix conflicts

Issue: jni/external/faiss and jni/external/nmslib submodule pointers were dirty from local cmake builds.
Solution: Reset both submodule pointers. Diff is now isolated to src/, KNNPlugin.java, and CHANGELOG.md. The CHANGELOG.md conflict against main will be resolved in a rebase before merge.

2. Use @Log4j2 in MuveraIngestProcessor

Issue: Manual LogManager.getLogger(...) instead of the Lombok annotation used elsewhere in k-NN.
Solution: Both processors now use @Log4j2.

3. Use float[] wherever possible instead of double[] in MuveraEncoder

Issue: Internal arrays were double[]/double[][], doubling memory and blocking SIMD.
Solution: Hot path is now fully float[][] — input vectors, cluster centers, projection rows, SimHash hyperplanes. Both processors parse incoming Number via floatValue(). Live test gave identical scores (3.5 / 3.0) to the previous path, so float precision is sufficient for our parameter ranges.

4. Use SIMD on the SimHash dot product

Issue: Scalar dot loop in getClusterId.
Solution: Switched to Lucene's SIMD-backed VectorUtil.dotProduct(float[], float[]). Each SimHash plane is stored as its own contiguous float[dim] so no slicing/copying happens at the call site.

5. Use float[][] for the centers array

Solution: centers is now float[numPartitions][dim], reset per repetition with Arrays.fill(centers[c], 0f).

6. Use a 2D array instead of List<List<Integer>> for cluster vec indices

Issue: List-of-List boxing in the hot path.
Solution: Replaced with int[numPartitions][nVectors] plus an int[] counts. Pre-allocated to the worst case so the per-repetition loop is allocation-free.

7. SIMD on the cluster-center accumulation centers[cid][d] += tokenVec[d]

Solution: Replaced the scalar loop with VectorUtil.add(centers[cid], tokenVec) — Lucene's in-place SIMD add.

8. SIMD on the normalize step centerVec[d] *= inv

Issue: Lucene's VectorUtil does not expose a scale-by-scalar primitive (its float-array API is add, dotProduct, cosine, squareDistance, l2normalize).
Solution: Kept the scalar loop with a comment explaining the rationale. The pattern (single load, multiply by constant, single store) is reliably auto-vectorized by HotSpot on x86/ARM. Happy to add a direct Panama jdk.incubator.vector path for this one loop if you'd prefer — but it would be the only direct incubator vector use in the plugin (everything else goes through VectorUtil).

9. Move config-key strings to private static final constants

Solution: All processor config keys are now static final String CONFIG_* (e.g. CONFIG_SOURCE_FIELD, CONFIG_TARGET_FIELD, CONFIG_DIM, CONFIG_K_SIM, CONFIG_DIM_PROJ, CONFIG_R_REPS, CONFIG_SEED, CONFIG_FDE_DIMENSION, CONFIG_IGNORE_MISSING). Default values are constants too (DEFAULT_K_SIM, DEFAULT_DIM_PROJ, DEFAULT_R_REPS, DEFAULT_SEED, DEFAULT_IGNORE_MISSING).

10. Cluster-level Factory log line — "users will not see it"

Issue: The FDE dimension was logged at pipeline-creation time via cluster logs, which aren't user-visible.
Solution: Removed both Factory logger.info(...) calls (ingest + search). The computed FDE dimension is now surfaced through the auto-generated processor description, visible via GET _ingest/pipeline and GET _search/pipeline. Format: "MUVERA FDE encoder: dim=128, fde_dimension=2560 (r_reps=20, k_sim=4, dim_proj=8, seed=42)". Only auto-generated when the user didn't supply their own description.

11. Use @NoArgsConstructor(access = AccessLevel.PRIVATE) in MuveraProcessorUtils

Solution: Done — matches the pattern in KNNVectorUtil, QuantizerFactory, KNNVectorFieldMapperUtil, etc.

12. Avoid float[] → List<Float> conversion

Issue: Boxing every output element on the search path.
Solution: For the search processor, requestContext.setAttribute(targetField, queryFde) now passes the raw float[]. The pipeline framework serializes context attributes via XContentBuilder.value(Object) at template-resolution time, and XContentBuilder has a registered writer for float[].class (writers.put(float[].class, (b, v) -> b.values((float[]) v))), so it serializes natively as a JSON array.
The ingest processor still converts to List<Float> because IngestDocument.deepCopy(Object) only supports Map, List, primitives, and byte[] — passing float[] to setFieldValue would throw IllegalArgumentException("unexpected value type [class [F]") later in the ingest path. Code comment explains this.

13. Add javadocs on all public classes and functions

Solution: Class-level Javadoc on MuveraEncoder, MuveraIngestProcessor, MuveraSearchRequestProcessor, MuveraProcessorUtils, and inner Factory classes. Method-level Javadoc on the constructor, getEmbeddingSize, processDocument, processQuery, process, getClusterId, the processRequest overrides, extraction helpers, parseMultiVectors, and MuveraProcessorUtils.readLongProperty.
Latest commits: 93ee708f (VectorUtil.add for accumulation), 7bd886d4 (everything else above). Verification: End-to-end test on a live 3.6.0-SNAPSHOT cluster — same scores (3.5 / 3.0) as the pre-refactor code, all unit + integration tests pass locally.

Vikasht34:
# MUVERA End-to-End Design(Current)

## Problem Statement

ColBERT/ColPali models produce **N token vectors per document** (variable N).
Standard ANN indices require **1 fixed-size vector per document**.
MaxSim scoring (sum of per-query-token max similarities) is O(N×M) — too expensive for first-pass retrieval.

## Solution: Two-Phase Retrieval

```
Phase 1: ANN search on FDE vectors (fast, approximate)
Phase 2: Rerank top-K with exact MaxSim (accurate, expensive)
```

---

## Architecture Overview

```
╔══════════════════════════════════════════════════════════════════════╗
║                        INDEX TIME                                    ║
╠══════════════════════════════════════════════════════════════════════╣
║                                                                      ║
║  User Document                                                       ║
║  {                                                                   ║
║    "text": "hello world",                                            ║
║    "colbert_vectors": [[0.1, 0.2, ...], [0.3, 0.4, ...], ...]      ║
║  }                                                                   ║
║       │                                                              ║
║       ▼                                                              ║
║  ┌─────────────────────────────────────┐                            ║
║  │      MuveraIngestProcessor          │                            ║
║  │                                     │                            ║
║  │  1. Read colbert_vectors            │                            ║
║  │  2. Validate dimensions             │                            ║
║  │  3. MuveraEncoder.processDocument() │                            ║
║  │  4. Write FDE to target field       │                            ║
║  └─────────────────────────────────────┘                            ║
║       │                                                              ║
║       ▼                                                              ║
║  Indexed Document                                                    ║
║  {                                                                   ║
║    "text": "hello world",                                            ║
║    "colbert_vectors": [[0.1, 0.2, ...], ...],  ← kept for reranking ║
║    "muvera_fde": [0.42, -0.13, 0.87, ...]      ← ANN searchable    ║
║  }                                                                   ║
║                                                                      ║
╚══════════════════════════════════════════════════════════════════════╝


╔══════════════════════════════════════════════════════════════════════╗
║                        QUERY TIME                                    ║
╠══════════════════════════════════════════════════════════════════════╣
║                                                                      ║
║  User Search Request                                                 ║
║  {                                                                   ║
║    "query": {                                                        ║
║      "template": {                                                   ║
║        "script_score": {                                             ║
║          "query": { "knn": { "muvera_fde": {                        ║
║              "vector": "${muvera_fde}",  ← placeholder               ║
║              "k": 40                     ← prefetch K                ║
║          }}},                                                        ║
║          "script": {                                                 ║
║            "source": "lateInteractionScore(...)",                    ║
║            "params": {                                               ║
║              "query_vectors": [[q1], [q2], ...],  ← raw query tokens ║
║              "space_type": "innerproduct"                            ║
║            }                                                         ║
║          }                                                           ║
║        }                                                             ║
║      }                                                               ║
║    }                                                                 ║
║  }                                                                   ║
║       │                                                              ║
║       ▼                                                              ║
║  ┌─────────────────────────────────────────────┐                    ║
║  │   MuveraSearchRequestProcessor              │                    ║
║  │                                             │                    ║
║  │  1. Detect TemplateQueryBuilder             │                    ║
║  │  2. Navigate: script_score→script→params    │                    ║
║  │  3. Extract query_vectors                   │                    ║
║  │  4. MuveraEncoder.processQuery()            │                    ║
║  │  5. Store FDE in PipelineProcessingContext  │                    ║
║  └─────────────────────────────────────────────┘                    ║
║       │                                                              ║
║       ▼                                                              ║
║  ┌─────────────────────────────────────────────┐                    ║
║  │   Template Query Resolution                  │                    ║
║  │                                             │                    ║
║  │  ${muvera_fde} → [0.55, -0.22, 0.91, ...]  │                    ║
║  │  (reads from pipeline context)              │                    ║
║  └─────────────────────────────────────────────┘                    ║
║       │                                                              ║
║       ▼                                                              ║
║  ┌─────────────────────────────────────────────┐                    ║
║  │   Phase 1: KNN ANN Search                   │                    ║
║  │                                             │                    ║
║  │  dot_product(query_FDE, doc_FDE)            │                    ║
║  │  Returns top-40 candidates                  │                    ║
║  └─────────────────────────────────────────────┘                    ║
║       │                                                              ║
║       ▼                                                              ║
║  ┌─────────────────────────────────────────────┐                    ║
║  │   Phase 2: script_score Reranking           │                    ║
║  │                                             │                    ║
║  │  lateInteractionScore(                      │                    ║
║  │    query_tokens,                            │                    ║
║  │    doc.colbert_vectors,                     │                    ║
║  │    space_type                               │                    ║
║  │  )                                          │                    ║
║  │  = exact MaxSim on top-40 candidates        │                    ║
║  │  Returns top-10 final results               │                    ║
║  └─────────────────────────────────────────────┘                    ║
║                                                                      ║
╚══════════════════════════════════════════════════════════════════════╝
```

---

## MuveraEncoder Internals

```
Input: N token vectors, each of dimension D
       e.g., tokens = [[t1], [t2], ..., [tN]]  where each ti ∈ ℝ^D

For each repetition r = 1..rReps:
┌────────────────────────────────────────────────────────────┐
│                                                            │
│  Step 1: SimHash Partitioning                              │
│  ────────────────────────────                              │
│  kSim random hyperplanes → 2^kSim = P partitions          │
│                                                            │
│  For each token ti:                                        │
│    cluster_id = 0                                          │
│    for k = 0..kSim-1:                                      │
│      if dot(ti, hyperplane[r][k]) > 0:                     │
│        cluster_id |= (1 << k)                              │
│    assign ti → cluster[cluster_id]                         │
│                                                            │
│  ┌──────┐ ┌──────┐ ┌──────┐     ┌──────┐                 │
│  │ C_0  │ │ C_1  │ │ C_2  │ ... │C_P-1 │                 │
│  │t1,t5 │ │t2,t7 │ │  ∅   │     │ t3   │                 │
│  └──────┘ └──────┘ └──────┘     └──────┘                 │
│                                                            │
│  Step 2: Compute Cluster Centers                           │
│  ───────────────────────────────                           │
│  Document mode: center = mean(tokens in cluster)           │
│  Query mode:    center = sum(tokens in cluster)            │
│                                                            │
│  Step 3: Fill Empty Clusters (document mode only)          │
│  ────────────────────────────────────────────────          │
│  For empty C_2: find Hamming-nearest non-empty cluster     │
│    hamming(2=0b10, 0=0b00) = 1  ← nearest!                │
│    hamming(2=0b10, 1=0b01) = 2                             │
│    Copy center of C_0 into C_2                             │
│                                                            │
│  Step 4: Random Projection (dimensionality reduction)      │
│  ─────────────────────────────────────────────────         │
│  For each cluster c = 0..P-1:                              │
│    For each j = 0..dimProj-1:                              │
│      output[r*P*dimProj + c*dimProj + j] =                 │
│        (1/√dimProj) * dot(center[c], projection[r][j])     │
│                                                            │
│  projection[r][j] ∈ {-1, +1}^D  (Rademacher random)       │
│                                                            │
└────────────────────────────────────────────────────────────┘

Output: Concatenate all repetitions
        FDE ∈ ℝ^(rReps × 2^kSim × dimProj)

With defaults (rReps=20, kSim=4, dimProj=8):
  FDE dimension = 20 × 16 × 8 = 2560
```

---

## Why This Approximates MaxSim

```
MaxSim(Q, D) = Σ_i max_j dot(q_i, d_j)

For each query token q_i:
  - SimHash assigns it to a partition
  - The document center in that partition is the average of 
    document tokens that hash similarly
  - dot(q_i, center) ≈ max_j dot(q_i, d_j) for tokens in same partition

Multiple repetitions (rReps) reduce variance from random hash collisions.
Random projection preserves dot products (Johnson-Lindenstrauss lemma).

Therefore:
  dot(FDE_query, FDE_doc) ≈ MaxSim(Q, D)
```

---

## Configuration Tuning Guide

```
┌──────────┬─────────┬───────────────────────────────────────────┐
│ Param    │ Default │ Effect                                     │
├──────────┼─────────┼───────────────────────────────────────────┤
│ dim      │ (req'd) │ Must match embedding model dimension       │
│ k_sim    │ 4       │ More bits → more partitions → better       │
│          │         │ separation, but FDE grows exponentially    │
│ dim_proj │ 8       │ Higher → preserves more info per cluster   │
│          │         │ but FDE grows linearly                     │
│ r_reps   │ 20      │ More reps → less variance, but FDE grows  │
│          │         │ linearly and compute grows linearly        │
│ seed     │ 42      │ Must be SAME for ingest and search         │
└──────────┴─────────┴───────────────────────────────────────────┘

FDE dimension = r_reps × 2^k_sim × dim_proj

Trade-offs:
  Higher FDE dim → better recall → more storage + slower ANN
  Lower FDE dim  → worse recall  → less storage + faster ANN

Recommended starting points:
  ColBERT-128d:  dim=128, k_sim=4, dim_proj=8, r_reps=20 → FDE=2560
  ColPali-128d:  dim=128, k_sim=3, dim_proj=8, r_reps=20 → FDE=1280
  Large models:  dim=768, k_sim=3, dim_proj=4, r_reps=10 → FDE=320
```

---

## Pipeline Setup (User-Facing)

```json
// 1. Create ingest pipeline
PUT _ingest/pipeline/muvera-ingest
{
  "processors": [{
    "muvera": {
      "source_field": "colbert_vectors",
      "target_field": "muvera_fde",
      "dim": 128,
      "k_sim": 4,
      "dim_proj": 8,
      "r_reps": 20,
      "seed": 42
    }
  }]
}

// 2. Create search pipeline  
PUT _search/pipeline/muvera-search
{
  "request_processors": [{
    "muvera_query": {
      "target_field": "muvera_fde",
      "dim": 128,
      "k_sim": 4,
      "dim_proj": 8,
      "r_reps": 20,
      "seed": 42
    }
  }]
}

// 3. Create index
PUT my-colbert-index
{
  "settings": {
    "index.knn": true,
    "default_pipeline": "muvera-ingest"
  },
  "mappings": {
    "properties": {
      "colbert_vectors": { "type": "flat_object" },
      "muvera_fde": {
        "type": "knn_vector",
        "dimension": 2560,
        "method": { "name": "hnsw", "engine": "faiss", "space_type": "innerproduct" }
      }
    }
  }
}

// 4. Search
POST my-colbert-index/_search?search_pipeline=muvera-search
{
  "query": {
    "template": {
      "script_score": {
        "query": { "knn": { "muvera_fde": { "vector": "${muvera_fde}", "k": 40 }}},
        "script": {
          "source": "lateInteractionScore(params.query_vectors, 'colbert_vectors', params._source, params.space_type)",
          "params": {
            "query_vectors": [[0.1, 0.2, ...], [0.3, 0.4, ...]],
            "space_type": "innerproduct"
          }
        }
      }
    }
  }
}
```

---

## Component Responsibilities

```
┌───────────────────────┬────────────────────────────────────────────┐
│ Component             │ Responsibility                              │
├───────────────────────┼────────────────────────────────────────────┤
│ MuveraEncoder         │ Pure math: multi-vectors → FDE vector      │
│                       │ Thread-safe, immutable after construction   │
│                       │ Two modes: processDocument / processQuery   │
├───────────────────────┼────────────────────────────────────────────┤
│ MuveraIngestProcessor │ Pipeline integration for index-time         │
│                       │ Parses List<List<Number>> from _source      │
│                       │ Validates dimensions                        │
│                       │ Writes FDE as List<Float> to target field   │
├───────────────────────┼────────────────────────────────────────────┤
│ MuveraSearchRequest-  │ Pipeline integration for query-time         │
│ Processor             │ Extracts query_vectors from TemplateQuery   │
│                       │ Encodes query FDE                           │
│                       │ Stores in PipelineProcessingContext         │
├───────────────────────┼────────────────────────────────────────────┤
│ MuveraProcessorUtils  │ Shared config parsing (readLongProperty)    │
├───────────────────────┼────────────────────────────────────────────┤
│ lateInteractionScore  │ Script for exact MaxSim reranking           │
│ (existing)            │ Runs on top-K candidates from Phase 1       │
└───────────────────────┴────────────────────────────────────────────┘
```

---

## Data Flow Diagram

```
                    INDEX TIME
                    ══════════

  Client                  OpenSearch                     Lucene
    │                        │                             │
    │  PUT doc with          │                             │
    │  colbert_vectors       │                             │
    │───────────────────────►│                             │
    │                        │                             │
    │                   IngestPipeline                     │
    │                        │                             │
    │                  MuveraIngest                        │
    │                  Processor                           │
    │                        │                             │
    │                   ┌────┴────┐                        │
    │                   │Muvera   │                        │
    │                   │Encoder  │                        │
    │                   │.process │                        │
    │                   │Document │                        │
    │                   └────┬────┘                        │
    │                        │                             │
    │                  doc + FDE field                     │
    │                        │─────────────────────────────►
    │                        │          Index doc           │
    │                        │          (HNSW graph for FDE)│
    │◄───────────────────────│                             │
    │   201 Created          │                             │


                    QUERY TIME
                    ══════════

  Client                  OpenSearch                     Lucene
    │                        │                             │
    │  POST _search          │                             │
    │  template + vectors    │                             │
    │───────────────────────►│                             │
    │                        │                             │
    │                  SearchPipeline                      │
    │                        │                             │
    │                  MuveraSearch                        │
    │                  RequestProcessor                    │
    │                        │                             │
    │                   ┌────┴────┐                        │
    │                   │Muvera   │                        │
    │                   │Encoder  │                        │
    │                   │.process │                        │
    │                   │Query    │                        │
    │                   └────┬────┘                        │
    │                        │                             │
    │                  context.set(FDE)                    │
    │                        │                             │
    │                  TemplateQuery                       │
    │                  resolves ${fde}                     │
    │                        │                             │
    │                  Phase 1: KNN                        │
    │                        │─────────────────────────────►
    │                        │   ANN search on HNSW graph  │
    │                        │◄────────────────────────────│
    │                        │   top-40 candidates         │
    │                        │                             │
    │                  Phase 2: script_score               │
    │                  lateInteractionScore                │
    │                  (exact MaxSim on 40 docs)           │
    │                        │                             │
    │◄───────────────────────│                             │
    │  top-10 results        │                             │
    │  scored by MaxSim      │                             │
```

---

## Key Invariants

1. **Seed must match** — ingest and search pipelines MUST use the same seed, or FDE vectors are incompatible
2. **Dimension must match** — `dim` in processor config must equal the embedding model's output dimension
3. **FDE dimension must match mapping** — `knn_vector` field dimension must equal `rReps × 2^kSim × dimProj`
4. **Space type must be innerproduct** — FDE approximation relies on dot product
5. **Original vectors must be preserved** — stored in _source for Phase 2 reranking

PR #3167, opened 2026-03-14
Enhance ADC scoring with SIMD Vector API and add comprehensive tests

### LINKED ISSUE #3150: [Enhancement] Integrate SIMD with ADC distance computation
## Description
While reading through the code of KNN what I found was ADC feature is not using SIMD instruction set for doing distance computation. Creating this gh issue to track the SIMD implementation for ADC. I just see bunch of TODOs in the code

Code Ref which needs a change:

https://github.com/opensearch-project/k-NN/blob/0d05a101a25317ab21503e9a6f9a9a39b608b74a/src/main/java/org/opensearch/knn/plugin/script/KNNScoringUtil.java#L114-L175

ajw711:
### Description
Resolves #3150

Implements SIMD optimization for `l2SquaredADC` and `innerProductADC` using Java Vector API (`FloatVector.SPECIES_PREFERRED`) as suggested in the linked issue.

**Key Improvements & Stability Considerations:**
- **Robust Build Configuration:** Configured `--add-modules jdk.incubator.vector` and `--enable-preview` globally in `build.gradle` using `tasks.withType`. This ensures the Vector API is consistently injected not only during `JavaCompile` but also during `Test` and `JavaExec` phases, preventing unexpected `NoClassDefFoundError` at runtime.
- **Comprehensive Testing:** Added detailed unit tests covering bit-level operations, multi-byte arrays, and specifically a large **1024-dimension edge case**. This guarantees both the SIMD loop bounds and the tail-loop logic function perfectly under heavy operations.
- **Documentation:** Updated `CHANGELOG.md` to reflect the enhancement.

### Related Issues
Resolves #3150

### Check List


0ctopus13prime:
Thank you @ajw711
Taking it a look and get back to you!

finnroblin (approved):
Looks good to me, thanks!

ajw711:
Hi @0ctopus13prime and team,

Just following up on this PR 🙂
It has one approval already, and all review comments have been addressed.

Could someone help:
1) Approve the workflows so CI can run
2) Provide an additional review if needed

Thanks a lot!

navneet1v:
@ajw711 can you please fix the CI errors.

CHANGELOG.md
```diff
@@ -18,3 +18,20 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 
 ### Enhancements
+* Make Merge in nativeEngine can Abort [#2529](https://github.com/opensearch-project/k-NN/pull/2529)
+* Use pre-quantized vectors from native engines for exact search [#3095](https://github.com/opensearch-project/k-NN/pull/3095)
+* Use right Vector Scorer when segments are initialized using SPI and also corrected the maxConn for MOS [#3117](https://github.com/opensearch-project/k-NN/pull/3117)
+* Use pre-quantized vectors for ADC [#3113](https://github.com/opensearch-project/k-NN/pull/3113)
```
navneet1v: I think whole change log completely cleared. can we please fix it

ajw711:
Hi @navneet1v,

You're right — the changelog was unintentionally overwritten during rebase.

I've now restored it based on main and only added my entry under Unreleased 3.7.

Could you please take another look?

Thanks!

src/main/java/org/opensearch/knn/plugin/script/KNNScoringUtil.java
```diff
@@ -5,21 +5,58 @@
...
+                    MethodType.methodType(float.class, float[].class, byte[].class)
+            );
+            innerProduct = lookup.findStatic(
+                    simdClass,
+                    "innerProductADC",
+                    MethodType.methodType(float.class, float[].class, byte[].class)
+            );
+        } catch (Throwable t) {
+            // Fallback to scalar implementation if Vector API module or SIMD class is unavailable.
+            logger.debug("SIMD Vector API not available, falling back to scalar ADC computation");
```
navneet1v: can we convert this to warn log.

src/main/java/org/opensearch/knn/plugin/script/KNNScoringUtil.java
```diff
@@ -143,23 +190,35 @@ public static float l2SquaredADC(float[] queryVector, byte[] inputVector) {
...
+
+        if (SIMD_ENABLED && INNER_PRODUCT_ADC_SIMD != null) {
+            try{
+                return (float) INNER_PRODUCT_ADC_SIMD.invokeExact(queryVector, inputVector);
+            } catch (Throwable t) {
+                SIMD_ENABLED = false;
+                logger.warn("SIMD ADC failed unexpectedly. Permanent fallback to scalar enabled.", t);
+            }
+        }
+        return innerProductADCScalar(queryVector, inputVector);
```
navneet1v: Here is a thought, can we not determine this in a static block which function we have to use SIMD vs non SIMD? This will ensure that we don't do this if else check in the hot search path. Same for L2 distance too.

src/main/java/org/opensearch/knn/plugin/script/KNNScoringUtil.java
ajw711: Good suggestion. Updated this to WARN since falling back from SIMD to scalar can have performance impact and should be visible in production logs.

src/main/java/org/opensearch/knn/plugin/script/KNNScoringUtil.java
ajw711: Great suggestion, thanks for pointing this out.

You're absolutely right that having a conditional check in the hot path can introduce unnecessary overhead, especially with repeated branch evaluation.

By moving the SIMD vs scalar decision to the static initialization block, we can eliminate this branching and allow the JIT to treat the call site as stable, improving inlining opportunities and overall performance.

I've refactored the code accordingly for both ADC and L2 paths.

ajw711:
Hi @navneet1v,

Thanks again for the detailed feedback earlier.

I’ve applied all the suggested changes:
- Updated the logging to WARN level for SIMD fallback
- Refactored the SIMD vs scalar selection into static initialization to avoid branching in the hot path
- Cleaned up the implementation accordingly

Everything from the previous review should now be addressed.

I’d really appreciate another look when you have a chance 🙂

ajw711:
Thanks for pointing this out.

In the current implementation, `step` is derived from `SPECIES.length()`, and the Vector API species widths used here are expected to be multiples of 8 (e.g. 8, 16, 32 lanes). Therefore, `(step / 8)` correctly represents the number of bytes required for unpacking in the current SIMD path.

ajw711:
Hi @navneet1v,

While investigating the previous CI issues locally, I had run into a :forbiddenApisMain failure related to the SIMD Vector API classes:

ClassNotFoundException: jdk.incubator.vector.VectorSpecies

At the time, this happened with an older Gradle setup (~9.2.x). It seemed that compileJava and test tasks had:

--add-modules=jdk.incubator.vector

configured, but the forbiddenApisMain task JVM did not.

After upgrading Gradle (currently 9.4.1), I can no longer reproduce the issue in CI, so this may have been environment/version specific. However, since the SIMD implementation still references the incubator Vector API directly, I wanted to mention it in case it resurfaces in other environments/toolchains.

If it does become an issue again, I was considering either:

Adding the incubator module to the forbidden-apis task JVM
Excluding KNNScoringUtilSIMD from forbidden-apis scanning

Just wanted to share the context in case it helps with future CI/debugging.

ajw711:
The failing Code-Diff-Analyzer is caused by upstream commit #3323
(Add Maven cache mirror before mavenCentral to reduce 429 errors).

This repository change was rebased from upstream and not introduced by this PR.
Could you please review and add the skip-diff-analyzer label?

Issue #3169, opened 2026-03-16
[Enhancement/META] Make 32x compression default for Vector Engine


navneet1v:
## Description
Creating this Meta issue to track the effort for making the 32x compression default in KNN. Below are some of the high level task.

- [ ] Initial Benchmarks comparing BQ and BBQ implementation. @naveentatikonda 
- [ ] Integrate BBQ in Lucene Engine and Faiss engine : https://github.com/opensearch-project/k-NN/pull/3144
- [ ] Integrate BBQ based indexing with Faiss Engine : https://github.com/opensearch-project/k-NN/pull/3153
- [ ] Mapping changes to make 32x compression default with BBQ.
- [X] Integrate Flat Index with 32x compression : https://github.com/opensearch-project/k-NN/pull/3154
- [ ] Backward compatible tests
- [ ] Testing
- [ ] Documentation Changes

PR #3170, opened 2026-03-16
Faiss BBQ Bulk SIMD

0ctopus13prime:
### Description
Adding Bulk SIMD support for Faiss BBQ.

### Related Issues
https://github.com/opensearch-project/k-NN/issues/3169

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/memoryoptsearch/faiss/Faiss104ScalerQuantizedVectorScorer.java
+++ b/src/main/java/org/opensearch/knn/memoryoptsearch/faiss/Faiss104ScalerQuantizedVectorScorer.java
@@ -1,0 +1,278 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.knn.memoryoptsearch.faiss;
+
+import org.apache.lucene.codecs.hnsw.FlatVectorsScorer;
+import org.apache.lucene.codecs.lucene104.Lucene104ScalarQuantizedVectorScorer;
+import org.apache.lucene.codecs.lucene104.Lucene104ScalarQuantizedVectorsFormat;
+import org.apache.lucene.codecs.lucene104.QuantizedByteVectorValues;
+import org.apache.lucene.index.KnnVectorValues;
+import org.apache.lucene.index.VectorSimilarityFunction;
+import org.apache.lucene.store.IndexInput;
+import org.apache.lucene.util.ArrayUtil;
+import org.apache.lucene.util.Bits;
+import org.apache.lucene.util.hnsw.RandomVectorScorer;
+import org.apache.lucene.util.quantization.OptimizedScalarQuantizer;
+import org.opensearch.knn.jni.SimdVectorComputeService;
+import org.opensearch.knn.memoryoptsearch.MemorySegmentAddressExtractorUtil;
+
+import java.io.IOException;
+
+import static org.apache.lucene.codecs.lucene104.Lucene104ScalarQuantizedVectorsFormat.ScalarEncoding.SINGLE_BIT_QUERY_NIBBLE;
+
+/**
+ * A specialized {@link Lucene104ScalarQuantizedVectorScorer} that leverages
+ * FAISS-style SIMD-accelerated scoring for scalar-quantized vectors with fallback.
+ * Will hit fallback only if we cannot use native bulk simd scorer.
+ *
+ * <p>This scorer attempts to use a native SIMD-backed bulk scoring path when:
+ * <ul>
+ *   <li>The underlying vector values are {@link QuantizedByteVectorValues}</li>
+ *   <li>The backing storage can expose a raw memory address</li>
+ *   <li>The scalar encoding matches the expected FAISS-compatible format</li>
+ * </ul>
+ *
+ * <p>If these conditions are not met, it falls back to the default Lucene scoring implementation.
+ *
+ * <p>The SIMD path uses a precomputed search context and performs scoring in native code
+ * (e.g., AVX-512), significantly improving throughput for large-scale vector search.
+ */
+public class Faiss104ScalerQuantizedVectorScorer extends Lucene104ScalarQuantizedVectorScorer {
+    /**
+     * Creates a new scorer that wraps a non-quantized delegate scorer.
+     *
+     * @param delegate fallback scorer used when SIMD acceleration is not applicable
+     */
+    public Faiss104ScalerQuantizedVectorScorer(final FlatVectorsScorer delegate) {
+        super(delegate);
+    }
+
+    /**
+     * Returns a {@link RandomVectorScorer} for the given query vector.
+     *
+     * <p>This method attempts to construct a SIMD-accelerated scorer when the input vectors
+     * are quantized and backed by memory that can be accessed directly (e.g., via a memory segment).
+     * Otherwise, it falls back to the default implementation.
+     *
+     * @param similarityFunction the similarity function (e.g., inner product or L2)
+     * @param vectorValues       the vector storage
+     * @param target             the query vector (float32)
+     * @return a scorer capable of computing similarity scores
+     * @throws IOException if an error occurs while accessing vector data
+     */
+    @Override
+    public RandomVectorScorer getRandomVectorScorer(
+        VectorSimilarityFunction similarityFunction,
+        KnnVectorValues vectorValues,
+        float[] target
+    ) throws IOException {
+        if (vectorValues instanceof QuantizedByteVectorValues quantizedByteVectorValues) {
+            final IndexInput indexInput = quantizedByteVectorValues.getSlice();
+            final long[] addressAndSize = MemorySegmentAddressExtractorUtil.tryExtractAddressAndSize(indexInput, 0, indexInput.length());
+            if (addressAndSize != null) {
+                return bulkSimdRandomVectorScorer(quantizedByteVectorValues, target, addressAndSize, similarityFunction);
+            }
+        }
+
+        // Fallback to Lucene scorer
+        return super.getRandomVectorScorer(similarityFunction, vectorValues, target);
+    }
+
+    /**
+     * Builds a SIMD-accelerated scorer using quantized vectors and a precomputed query.
+     *
+     * <p>This method:
+     * <ol>
+     *   <li>Validates the scalar encoding format</li>
+     *   <li>Quantizes the query vector into the same representation as stored vectors</li>
+     *   <li>Applies transformations (e.g., nibble transposition) required for SIMD efficiency</li>
+     *   <li>Initializes a native SIMD search context</li>
+     * </ol>
+     *
+     * <p>The resulting scorer uses bulk SIMD instructions to compute similarity scores.
+     *
+     * @param quantizedByteVectorValues the quantized vector storage
+     * @param target                    the query vector (float32)
+     * @param addressAndSize            raw memory address and size of vector data
+     * @param similarityFunction        similarity function to use
+     * @return a SIMD-accelerated scorer
+     * @throws IOException if quantization or initialization fails
+     */
+    private RandomVectorScorer bulkSimdRandomVectorScorer(
+        final QuantizedByteVectorValues quantizedByteVectorValues,
+        final float[] target,
+        final long[] addressAndSize,
+        final VectorSimilarityFunction similarityFunction
+    ) throws IOException {
+        // Check encoding type
+        final Lucene104ScalarQuantizedVectorsFormat.ScalarEncoding scalarEncoding = quantizedByteVectorValues.getScalarEncoding();
+
+        // We only support 32x quantization with 4 bit query quantization for search.
+        if (scalarEncoding != SINGLE_BIT_QUERY_NIBBLE) {
+            throw new IllegalStateException("Faiss BBQ only supports SINGLE_BIT_QUERY_NIBBLE encoding.");
+        }
+
+        // Validate dimensionality
+        FlatVectorsScorer.checkDimensions(target.length, quantizedByteVectorValues.dimension());
+
+        // Transpose query vector if it needs to
+        final OptimizedScalarQuantizer quantizer = quantizedByteVectorValues.getQuantizer();
+        final byte[] scratch = new byte[scalarEncoding.getDiscreteDimensions(quantizedByteVectorValues.dimension())];
+        final byte[] targetQuantized;
+        if (scalarEncoding.isAsymmetric() == false) {
+            targetQuantized = scratch;
+        } else {
+            // Asymmetric encoding requires packed representation
+            targetQuantized = new byte[scalarEncoding.getQueryPackedLength(scratch.length)];
+        }
+
+        // We make a copy as the quantization process mutates the input
+        final float[] targetCopy = ArrayUtil.copyOfSubArray(target, 0, target.length);
+
+        // Perform scalar quantization
+        final OptimizedScalarQuantizer.QuantizationResult targetCorrectiveTerms = quantizer.scalarQuantize(
+            targetCopy,
+            scratch,
+            scalarEncoding.getQueryBits(),
+            quantizedByteVectorValues.getCentroid()
+        );
+
+        // Transpose half-bytes (nibbles) for SIMD-friendly layout
+        OptimizedScalarQuantizer.transposeHalfByte(scratch, targetQuantized);
+
+        // Return Bulk SIMD scorer
+        return new BulkSimdRandomVectorScorer(
+            targetQuantized,
+            quantizedByteVectorValues.size(),
+            targetCorrectiveTerms,
+            addressAndSize,
+            quantizedByteVectorValues,
+            similarityFunction,
+            targetCopy.length,
+            quantizedByteVectorValues.getCentroidDP()
+        );
+    }
+
+    /**
+     * A {@link RandomVectorScorer} implementation backed by native SIMD computation.
+     *
+     * <p>This scorer delegates all similarity computations to a native service
+     * ({@link SimdVectorComputeService}), which uses preloaded query state and raw
+     * vector memory to compute scores efficiently using SIMD instructions.
+     *
+     * <p>The query is preprocessed (quantized + transformed) once during construction,
+     * and reused across all scoring calls.
+     */
+    private static class BulkSimdRandomVectorScorer implements RandomVectorScorer {
+        private final int maxOrd;
+        private final QuantizedByteVectorValues knnVectorValues;
+
+        /**
+         * Constructs a SIMD-backed scorer and initializes the native search context.
+         *
+         * <p>This constructor pushes all necessary query state into native memory,
+         * including quantized query values and correction terms required for accurate scoring.
+         *
+         * @param targetQuantized        quantized query vector
+         * @param maxOrd                 total number of vectors
+         * @param targetCorrectiveTerms  correction terms from quantization
+         * @param addressAndSize         raw memory location of vector data
+         * @param knnVectorValues        vector storage abstraction
+         * @param similarityFunction     similarity function (IP or L2)
+         * @param dimension              vector dimensionality
+         * @param centroidDp             centroid dot-product correction
+         */
+        public BulkSimdRandomVectorScorer(
+            final byte[] targetQuantized,
+            final int maxOrd,
+            final OptimizedScalarQuantizer.QuantizationResult targetCorrectiveTerms,
+            final long[] addressAndSize,
+            final QuantizedByteVectorValues knnVectorValues,
+            final VectorSimilarityFunction similarityFunction,
+            final int dimension,
+            final float centroidDp
+        ) {
+
+            this.maxOrd = maxOrd;
+            this.knnVectorValues = knnVectorValues;
+
+            // Initialize native SIMD search context
+            SimdVectorComputeService.saveBBQSearchContext(
+                targetQuantized,
+                targetCorrectiveTerms.lowerInterval(),
+                targetCorrectiveTerms.upperInterval(),
+                targetCorrectiveTerms.additionalCorrection(),
+                targetCorrectiveTerms.quantizedComponentSum(),
+                addressAndSize,
+                similarityFunction == VectorSimilarityFunction.MAXIMUM_INNER_PRODUCT
+                    ? SimdVectorComputeService.SimilarityFunctionType.BBQ_IP.ordinal()
+                    : SimdVectorComputeService.SimilarityFunctionType.BBQ_L2.ordinal(),
+                dimension,
+                centroidDp
+            );
+        }
+
+        /**
+         * Computes similarity scores for multiple vectors in bulk using native SIMD code.
+         *
+         * <p>This method is optimized for throughput and should be preferred when scoring
+         * large batches of vectors.
+         *
+         * @param internalVectorIds vector ordinals to score
+         * @param scores            output buffer for similarity scores
+         * @param numVectors        number of vectors to process
+         * @return implementation-defined value (typically unused aggregate)
+         */
+        @Override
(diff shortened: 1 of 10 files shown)


0ctopus13prime:
Yeah, the test is not picking up new built shared library, not sure why though.
I've tested in Linux test environment, i've verified that it passes

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/Faiss104ScalerQuantizedVectorScorer.java
```diff
@@ -0,0 +1,186 @@
...
+        super(nonQuantizedDelegate);
+    }
+
+    @Override
+    public RandomVectorScorer getRandomVectorScorer(
+        VectorSimilarityFunction similarityFunction,
+        KnnVectorValues vectorValues,
+        float[] target
+    ) throws IOException {
+        if (vectorValues instanceof QuantizedByteVectorValues quantizedByteVectorValues) {
```
navneet1v, line 72: should we also not check the HasSlice interface also implemented or not?

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/Faiss104ScalerQuantizedVectorScorer.java
```diff
@@ -0,0 +1,186 @@
...
+            final IndexInput indexInput = quantizedByteVectorValues.getSlice();
+            final long[] addressAndSize = MemorySegmentAddressExtractorUtil.tryExtractAddressAndSize(indexInput, 0, indexInput.length());
+            if (addressAndSize != null) {
+                return bulkSimdRandomVectorScorer(quantizedByteVectorValues, target, addressAndSize, similarityFunction);
+            }
+        }
+
```
navneet1v, line 79: can we have a warn log here?

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissLucene104BBQBulkSimdScorer.java
```diff
@@ -0,0 +1,186 @@
...
+import org.opensearch.knn.jni.SimdVectorComputeService;
+import org.opensearch.knn.memoryoptsearch.MemorySegmentAddressExtractorUtil;
+
+import java.io.IOException;
+
+import static org.apache.lucene.codecs.lucene104.Lucene104ScalarQuantizedVectorsFormat.ScalarEncoding.SINGLE_BIT_QUERY_NIBBLE;
+
+public class FaissLucene104BBQBulkSimdScorer extends Lucene104ScalarQuantizedVectorScorer {
+    public FaissLucene104BBQBulkSimdScorer(final FlatVectorsScorer nonQuantizedDelegate) {
+        super(nonQuantizedDelegate);
```
navneet1v: shouldn't this be the delegate which doesn't use BulkSIMD?

0ctopus13prime:
This is frustrating, it's passing from my end
```
BUILD SUCCESSFUL in 2m 15s
21 actionable tasks: 10 executed, 11 up-to-date
```

Will force it to refresh in workflow.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissLucene104BBQBulkSimdScorer.java
```diff
@@ -0,0 +1,186 @@
...
+        public BulkSimdRandomVectorScorer(
+            final byte[] quantized,
+            final OptimizedScalarQuantizer.QuantizationResult correctionFactors,
+            final long[] addressAndSize,
+            final int maxOrd,
+            final int dimension,
+            final float centroidDp,
+            final QuantizedByteVectorValues knnVectorValues,
+            final VectorSimilarityFunction similarityFunction
+        ) {
```
navneet1v, line 197: I won't suggest passing everything to the constructor here and just initialize the SimdVectorComputeService upfront.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissLucene104BBQBulkSimdScorer.java
```diff
@@ -0,0 +1,186 @@
...
+            this.maxOrd = maxOrd;
+            this.knnVectorValues = knnVectorValues;
+            SimdVectorComputeService.saveBBQSearchContext(
```
navneet1v, line 203: how you are able to access what you have set here, in bulk score function since SimdVectorComputeService is calling all static functions

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/Faiss104ScalerQuantizedVectorScorer.java
0ctopus13prime: sure, but technically, I think we can thrown an exception in here as this will be used exclusively in search path.
What do you think?

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/Faiss104ScalerQuantizedVectorScorer.java
0ctopus13prime: QuantizedByteVectorValues already has HasIndexSlice, so I think it should be fine.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissLucene104BBQBulkSimdScorer.java
0ctopus13prime: `nonQuantizedDelegate` will be used when caller is requesting to do distance calculations on FP32 vectors. (i.e. KnnVectorValues vectorValues is not instance of QuantizedByteVectorValues)

0ctopus13prime: I think we should pass NULL, since this will be used for quantized index only.
What do you think?

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissLucene104BBQBulkSimdScorer.java
0ctopus13prime: So internally, it will set all these info in internal thread local right before starting HNSW searcher in FaissMemoryOptimizedSearcher.

```
// Set up required components for vector search
final RandomVectorScorer scorer = scorerSupplier.get(); <-------------- Set infos in thread local object
final KnnCollector collector = createKnnCollector(knnCollector, scorer);
final Bits acceptedOrds = scorer.getAcceptOrds(acceptDocs.bits());

if (knnCollector.k() < scorer.maxOrd()) {
    // Do ANN search with Lucene's HNSW graph searcher.
    HnswGraphSearcher.search(scorer, collector, new FaissHnswGraph(hnsw, getSlicedIndexInput()), acceptedOrds); <---------- Start searching.
```

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissLucene104BBQBulkSimdScorer.java
0ctopus13prime: hmm.. but we need to call the method in its constructor so that it can save those info in thread local right before searching in HNSW.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissLucene104BBQBulkSimdScorer.java
navneet1v: can you not do this in Line number 85 of this file? since that is where you are creating the scorer

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissLucene104BBQBulkSimdScorer.java
navneet1v: I won't suggest using NULL. The way I will see the fallback to happen is

1. Use bulkSIMD Scorer which uses C++.
2. If somehow that scorer doesn't work, use the Delegate scorer of Lucene which doesn't do BulkSIMD.

and in the 2, Lucene should then see if it wants to fallback to Fp32 based scoring

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/Faiss104ScalerQuantizedVectorScorer.java
navneet1v: replied above

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissLucene104BBQBulkSimdScorer.java
0ctopus13prime: yeah, we can do this, but i feel like it's bit weird to put it there.. 😅 

I've put it in BulkSimdRandomVectorScorer's constructor for capsulation purpose. 
Likewise how we're doing it in NativeRandomScorer - https://github.com/0ctopus13prime/k-NN/blob/6327a79d1aee3b43a76fff24ae787e4eeef3443f/src/main/java/org/opensearch/knn/memoryoptsearch/faiss/NativeRandomVectorScorer.java#L59 

Do you think putting it in line 85 would be better?

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissLucene104BBQBulkSimdScorer.java
0ctopus13prime: But I think that 'somehow' should never happen since it will be created in quantized index context. 
I'm thinking instead of falling back to the delegate silently, it should throw an exception explicitly.

0ctopus13prime: oh wait. 
I think we should pass `Lucene104ScalarQuantizedVectorScorer` to its delegate scorer.
nvm, please ignore mine.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/Faiss104ScalerQuantizedVectorScorer.java
0ctopus13prime: will update in the next rev.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissLucene104BBQBulkSimdScorer.java
0ctopus13prime: will update in the next rev.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissLucene104BBQBulkSimdScorer.java
0ctopus13prime: will update in the next rev.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/NativeRandomVectorScorer.java
```diff
@@ -70,8 +70,10 @@ public NativeRandomVectorScorer(
     public float bulkScore(final int[] internalVectorIds, final float[] scores, final int numVectors) {
         SimdVectorComputeService.scoreSimilarityInBulk(internalVectorIds, scores, numVectors);
         float max = scores[0];
-        for (float v : scores) {
-            if (v > max) max = v;
+        for (int i = 1; i < numVectors; ++i) {
+            if (scores[i] > max) {
+                max = scores[i];
+            }
```
navneet1v: Ideally we should do this in C++. Lets have a TODO for this.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissLucene104BBQBulkSimdScorer.java
```diff
@@ -0,0 +1,188 @@
...
+import org.apache.lucene.util.hnsw.RandomVectorScorer;
+import org.apache.lucene.util.quantization.OptimizedScalarQuantizer;
+import org.opensearch.knn.jni.SimdVectorComputeService;
+import org.opensearch.knn.memoryoptsearch.MemorySegmentAddressExtractorUtil;
+
+import java.io.IOException;
+
+import static org.apache.lucene.codecs.lucene104.Lucene104ScalarQuantizedVectorsFormat.ScalarEncoding.SINGLE_BIT_QUERY_NIBBLE;
+
+public class FaissLucene104BBQBulkSimdScorer extends Lucene104ScalarQuantizedVectorScorer {
```
navneet1v: Lets name this to Faiss104ScalerQuantizedVectorScorer

src/test/java/org/opensearch/knn/memoryoptsearch/faiss/FaissBBQBulkSimdScorerTests.java
```diff
@@ -0,0 +1,252 @@
...
+    private static final String FIELD_NAME = "vector";
+    private static final int NUM_VECTORS = 500;
+
+    @Test
+    public void testBBQEuclideanScoring() {
+        // TODO : Remove this once generic version of bulk simd is added
+        // for (int dim : Arrays.asList(1, 7, 77, 128, 512, 777, 1024, 10240, 30000, 65535)) {
+        // System.out.println("Dimension=" + dim);
+        // doTest(VectorSimilarityFunction.EUCLIDEAN, dim);
+        // }
```
navneet1v, line 67: any reason for keeping these tests?

src/test/java/org/opensearch/knn/memoryoptsearch/faiss/FaissBBQBulkSimdScorerTests.java
```diff
@@ -0,0 +1,252 @@
...
+ * Tests that the native SIMD BBQ scorer ({@link FaissLucene104BBQBulkSimdScorer}) produces scores
+ * matching Lucene's {@link Lucene104ScalarQuantizedVectorScorer} (the source of truth).
+ * <p>
+ * Uses the Lucene codec pipeline directly:
+ * 1. {@link Lucene104ScalarQuantizedVectorsFormat#fieldsWriter} to quantize and write vectors.
+ * 2. {@link Lucene104ScalarQuantizedVectorsReader} with {@link Lucene104ScalarQuantizedVectorScorer} → truth.
+ * 3. {@link Lucene104ScalarQuantizedVectorsReader} with {@link FaissLucene104BBQBulkSimdScorer} → test subject.
+ * 4. Compare scores.
+ */
+@TimeoutSuite(millis = 100 * TimeUnits.HOUR)
```
navneet1v: why we need this? and also such a large value

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissLucene104BBQBulkSimdScorer.java
navneet1v: add java docs here

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/NativeRandomVectorScorer.java
0ctopus13prime: oh shoot, C++ already doing this, will remove this

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissLucene104BBQBulkSimdScorer.java
0ctopus13prime: Sounds good

src/test/java/org/opensearch/knn/memoryoptsearch/faiss/FaissBBQBulkSimdScorerTests.java
0ctopus13prime: I've kept it for running in my ARM Neon environment

src/test/java/org/opensearch/knn/memoryoptsearch/faiss/FaissBBQBulkSimdScorerTests.java
0ctopus13prime: shoot, this is temp code that I used for running all the dimensions which took few couple of hours.
Will remove it

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissLucene104BBQBulkSimdScorer.java
0ctopus13prime: Sounds good

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/Faiss104ScalerQuantizedVectorScorer.java
```diff
@@ -0,0 +1,278 @@
...
+     */
+    @Override
+    public RandomVectorScorer getRandomVectorScorer(
+        VectorSimilarityFunction similarityFunction,
+        KnnVectorValues vectorValues,
+        float[] target
+    ) throws IOException {
+        if (vectorValues instanceof QuantizedByteVectorValues quantizedByteVectorValues) {
+            final IndexInput indexInput = quantizedByteVectorValues.getSlice();
+            final long[] addressAndSize = MemorySegmentAddressExtractorUtil.tryExtractAddressAndSize(indexInput, 0, indexInput.length());
```
kotwanikunal, line 74: Any chances this could fail given that this is a branched/conditional code path?
In that case it will be better to have a try catch based fallback.

kotwanikunal: Ahh nevermind. Followed the code flow to realize we handle exceptions in the called methods rather than the caller

https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/memoryoptsearch/AbstractMemorySegmentAddressExtractor.java#L36-L47

kotwanikunal (approved):

Vikasht34 (approved):
LGTM ..Thanks @0ctopus13prime

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/Faiss104ScalerQuantizedVectorScorer.java
```diff
@@ -0,0 +1,278 @@
...
+     * A {@link RandomVectorScorer} implementation backed by native SIMD computation.
+     *
+     * <p>This scorer delegates all similarity computations to a native service
+     * ({@link SimdVectorComputeService}), which uses preloaded query state and raw
+     * vector memory to compute scores efficiently using SIMD instructions.
+     *
+     * <p>The query is preprocessed (quantized + transformed) once during construction,
+     * and reused across all scoring calls.
+     */
+    private static class BulkSimdRandomVectorScorer implements RandomVectorScorer {
```
navneet1v, line 169: You can use RandomVectorScorer.AbstractRadomVectorScorer class here. It provides better override methods as compared to RandomVectorScorer and will reduce the code in this class

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/Faiss104ScalerQuantizedVectorScorer.java
```diff
@@ -0,0 +1,278 @@
...
+        if (vectorValues instanceof QuantizedByteVectorValues quantizedByteVectorValues) {
+            final IndexInput indexInput = quantizedByteVectorValues.getSlice();
+            final long[] addressAndSize = MemorySegmentAddressExtractorUtil.tryExtractAddressAndSize(indexInput, 0, indexInput.length());
+            if (addressAndSize != null) {
+                return bulkSimdRandomVectorScorer(quantizedByteVectorValues, target, addressAndSize, similarityFunction);
+            }
+        }
+
+        // Fallback to Lucene scorer
+        return super.getRandomVectorScorer(similarityFunction, vectorValues, target);
```
navneet1v, line 81: why this is not delegate.getRandomVectorScorer(similarityFunction, vectorValues, target); and its `super.getRandomVectorScorer(similarityFunction, vectorValues, target);`

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/Faiss104ScalerQuantizedVectorScorer.java
```diff
@@ -0,0 +1,278 @@
...
+ *   <li>The backing storage can expose a raw memory address</li>
+ *   <li>The scalar encoding matches the expected FAISS-compatible format</li>
+ * </ul>
+ *
+ * <p>If these conditions are not met, it falls back to the default Lucene scoring implementation.
+ *
+ * <p>The SIMD path uses a precomputed search context and performs scoring in native code
+ * (e.g., AVX-512), significantly improving throughput for large-scale vector search.
+ */
+public class Faiss104ScalerQuantizedVectorScorer extends Lucene104ScalarQuantizedVectorScorer {
```
navneet1v, line 43: why we are extending the `Lucene104ScalerQuantizedVectorScorer` ? We should extend FlatVectorScorer and then in constructor pass the instance of `Lucene104ScalarQuantizedVectorScorer` as the delegate ? Because right now if delegate and Faiss104ScalerQuantizedVectorScorer extends Lucene104ScalarQuantizedVectorScorer are one and the same thing.

0ctopus13prime: This is Faiss104 scorer which will be passed to Lucene104ScalarQuantizedVectorsReader which only takes Lucene104ScalarQuantizedVectorScorer. 
The purpose of this class is to hijack the method and return bulk simd scorer if it's available.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/Faiss104ScalerQuantizedVectorScorer.java
0ctopus13prime: Because super.getRandomVectorScorer will take care of all the fallback logic.
`delegate` in this class is for .vec flat vector reader, we should not call it.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/Faiss104ScalerQuantizedVectorScorer.java
0ctopus13prime: oh this is good. will update it

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/Faiss104ScalerQuantizedVectorScorer.java
navneet1v: Make sense. Thank you providing the details. I checked it. Then yes we cannot do anything here.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/Faiss104ScalerQuantizedVectorScorer.java
navneet1v: > Because super.getRandomVectorScorer will take care of all the fallback logic. `delegate` in this class is for .vec flat vector reader, we should not call it.

Make sense. I was not aware that `delegate` is for the .vec file. Sounds good.

Issue #2577, opened 2025-03-04
[META] Use Prefetch api of IndexInput for Vectors


navneet1v:
## Description
Creating this Meta issue to track all the ideas and places where we can use `prefetch` functionality of Lucene for Vectors.

Lucene Issue on Prefetch: https://github.com/apache/lucene/issues/13179

Based on the deep-dive what I can see is currently the prefetch is not added for KNNVectorValues interface, but the IndexInput has the prefetch. We might need to make some changes in the KNNVectorValues so that we can expose the prefetch logic on a docId(like this: https://github.com/apache/lucene/blob/e8e179cd4553b503dc529179beb7c6290202acf0/lucene/core/src/java/org/apache/lucene/index/TermVectors.java#L44) rather than use the low level apis of IndexInput.

### Tasks

- [ ] implement docId based prefetch in Lucene for KNNVectorValues : https://github.com/apache/lucene/pull/15722
- [ ] https://github.com/opensearch-project/k-NN/issues/2423
- [ ] Use prefetch to reduce the merge time for vector segments
- [ ] Use prefetch in HNSW based search
- [ ] integrate Prefetch in filtered vector search
- [ ] Integrate prefetch in rescoring

PR #3173, opened 2026-03-16
Add Prefetch functionality to prefetch vectors during ANN Search for MemoryOptimizedSearch

navneet1v:
### Description
Add Prefetch functionality to prefetch vectors during ANN Search for MemoryOptimizedSearch.

What changes are present:
1. A new PrefetchableVectorScorer is created, which wraps FlatVectorScorer, which is then passed to FlatVectorFormat.
2. PrefetchableVectorScorer mainly calls the delegate scorer, just in case of bulkScore it calls prefetch first.
3. Moved the prefetch behind a FF.
4. Passed the FlatVectorScorer from FlatVectorFormat to FaissMemoryOptimizedSearcher to ensure that searcher can start using PrefetchableVectorScorer.
5. As of now, Prefetch is working for Byte, Fp32 and in ANN flow.

### Next Steps
1. Integrate the Prefetchable Scorer with NativeScorer(which uses BulkSIMD).
2. Integrate changes in the ADCScorer.
3. @VijayanB is working on implementing Scorer with ExactSearch which will bring Prefetch functionality to ExactSearch too.

### Related Issues
https://github.com/opensearch-project/k-NN/issues/2577

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/common/featureflags/KNNFeatureFlags.java
+++ b/src/main/java/org/opensearch/knn/common/featureflags/KNNFeatureFlags.java
@@ -26,6 +26,8 @@ public class KNNFeatureFlags {
 
     // Feature flags
     private static final String KNN_FORCE_EVICT_CACHE_ENABLED = "knn.feature.cache.force_evict.enabled";
+    private static final String KNN_PREFETCH_ENABLED = "knn.feature.prefetch.enabled";
+    private static final boolean KNN_PREFETCH_ENABLED_DEFAULT_VALUE = true;
 
     @VisibleForTesting
     public static final Setting<Boolean> KNN_FORCE_EVICT_CACHE_ENABLED_SETTING = Setting.boolSetting(
@@ -35,8 +37,15 @@ public class KNNFeatureFlags {
         Dynamic
     );
 
+    public static final Setting<Boolean> KNN_PREFETCH_ENABLED_SETTING = Setting.boolSetting(
+        KNN_PREFETCH_ENABLED,
+        KNN_PREFETCH_ENABLED_DEFAULT_VALUE,
+        NodeScope,
+        Dynamic
+    );
+
     public static List<Setting<?>> getFeatureFlags() {
-        return ImmutableList.of(KNN_FORCE_EVICT_CACHE_ENABLED_SETTING);
+        return ImmutableList.of(KNN_FORCE_EVICT_CACHE_ENABLED_SETTING, KNN_PREFETCH_ENABLED_SETTING);
     }
 
     /**
@@ -45,5 +54,12 @@ public class KNNFeatureFlags {
      */
     public static boolean isForceEvictCacheEnabled() {
         return Booleans.parseBoolean(KNNSettings.state().getSettingValue(KNN_FORCE_EVICT_CACHE_ENABLED).toString(), false);
+    }
+
+    public static boolean isPrefetchEnabled() {
+        return Booleans.parseBoolean(
+            KNNSettings.state().getSettingValue(KNN_PREFETCH_ENABLED).toString(),
+            KNN_PREFETCH_ENABLED_DEFAULT_VALUE
+        );
     }
 }
--- a/src/main/java/org/opensearch/knn/index/codec/scorer/PrefetchHelper.java
+++ b/src/main/java/org/opensearch/knn/index/codec/scorer/PrefetchHelper.java
@@ -1,0 +1,112 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.knn.index.codec.scorer;
+
+import lombok.AccessLevel;
+import lombok.NoArgsConstructor;
+import lombok.extern.log4j.Log4j2;
+import org.apache.lucene.store.IndexInput;
+import org.opensearch.knn.common.featureflags.KNNFeatureFlags;
+
+import java.io.IOException;
+import java.util.Arrays;
+
+/**
+ * Helper class to prefetch vector data from disk to improve query performance.
+ * <p>
+ * Prefetches vector data using exact byte range strategy. Vectors are grouped together when
+ * the end of a vector (start offset + vector size) falls within 128KB from the group's start offset.
+ * This minimizes I/O operations while fetching only the exact bytes needed.
+ * <p>
+ * Prefetching is controlled by {@link KNNFeatureFlags#isPrefetchEnabled()}. When disabled,
+ * no prefetch operations are performed.
+ * <p>
+ * <b>Example:</b><br>
+ * Given vectors at offsets [100KB, 120KB, 300KB] with 10KB vector size:
+ * <ul>
+ *   <li>Group 1: Prefetch 40KB (100KB to 130KB, covers vectors at 100KB and 120KB)</li>
+ *   <li>Group 2: Prefetch 10KB (300KB to 310KB, covers vector at 300KB)</li>
+ * </ul>
+ * Result: 2 prefetch operations, 50KB total prefetched (vs 6 random seeks without prefetch)
+ */
+@Log4j2
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+public class PrefetchHelper {
+
+    // TODO: If needed we can get this value via Cluster Settings
+    private static final long BYTES_128 = 128 * 1024;
+
+    /**
+     * Prefetches vector data from disk using exact byte range strategy.
+     * <p>
+     * Vectors are sorted by offset and grouped when the end of a vector falls within 128KB
+     * from the group start. Each group is prefetched with exact byte range needed.
+     * Returns early if prefetch is disabled, ordinals are null, or fewer than 2 vectors.
+     *
+     * @param indexInput the index input to prefetch from
+     * @param baseOffset the base offset in the file where vectors start
+     * @param oneVectorByteSize the size of one vector in bytes
+     * @param ordsToPrefetch array of vector ordinals to prefetch
+     * @param numOrds number of valid ordinals in the array
+     * @throws IOException if an I/O error occurs during prefetch
+     */
+    public static void prefetch(
+        final IndexInput indexInput,
+        final long baseOffset,
+        final long oneVectorByteSize,
+        final int[] ordsToPrefetch,
+        final int numOrds
+    ) throws IOException {
+        if (ordsToPrefetch == null || numOrds <= 1) {
+            return;
+        }
+        if (KNNFeatureFlags.isPrefetchEnabled()) {
+            prefetchExactVectorSize(indexInput, baseOffset, oneVectorByteSize, ordsToPrefetch, numOrds);
+        } else {
+            log.debug("KNNVectors Prefetch is disabled");
+        }
+    }
+
+    /**
+     * Prefetches vectors using exact byte ranges.
+     * <p>
+     * Groups vectors within 128KB ranges and prefetches only the exact bytes needed for each group.
+     *
+     * @param indexInput the index input to prefetch from
+     * @param baseOffset the base offset in the file where vectors start
+     * @param oneVectorByteSize the size of one vector in bytes
+     * @param ordsToPrefetch array of vector ordinals to prefetch
+     * @param numOrds number of valid ordinals in the array
+     * @throws IOException if an I/O error occurs during prefetch
+     */
+    private static void prefetchExactVectorSize(
+        final IndexInput indexInput,
+        final long baseOffset,
+        final long oneVectorByteSize,
+        final int[] ordsToPrefetch,
+        final int numOrds
+    ) throws IOException {
+        Arrays.sort(ordsToPrefetch, 0, numOrds);
+
+        int groupCount = 1;
+        long groupStartOffset = baseOffset + (long) ordsToPrefetch[0] * oneVectorByteSize;
+
+        for (int i = 1; i < numOrds; i++) {
+            long currentOffset = baseOffset + (long) ordsToPrefetch[i] * oneVectorByteSize;
+            if ((currentOffset + oneVectorByteSize) - groupStartOffset > BYTES_128) {
+                long prevOffset = baseOffset + (long) ordsToPrefetch[i - 1] * oneVectorByteSize;
+                indexInput.prefetch(groupStartOffset, (prevOffset + oneVectorByteSize) - groupStartOffset);
+                groupCount++;
+                groupStartOffset = currentOffset;
+            }
+        }
+        // Prefetch final group
+        long lastOffset = baseOffset + (long) ordsToPrefetch[numOrds - 1] * oneVectorByteSize;
+        indexInput.prefetch(groupStartOffset, (lastOffset + oneVectorByteSize) - groupStartOffset);
+
+        log.trace("Prefetching grouped [{}] vectors where num of ords was [{}] using exact prefetch size", groupCount, numOrds);
+    }
+}
--- a/src/main/java/org/opensearch/knn/index/codec/scorer/PrefetchableFlatVectorScorer.java
+++ b/src/main/java/org/opensearch/knn/index/codec/scorer/PrefetchableFlatVectorScorer.java
@@ -1,0 +1,135 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.knn.index.codec.scorer;
+
+import lombok.extern.log4j.Log4j2;
+import org.apache.lucene.codecs.hnsw.FlatVectorsScorer;
+import org.apache.lucene.index.KnnVectorValues;
+import org.apache.lucene.index.VectorSimilarityFunction;
+import org.apache.lucene.util.Bits;
+import org.apache.lucene.util.hnsw.RandomVectorScorer;
+import org.apache.lucene.util.hnsw.RandomVectorScorerSupplier;
+
+import java.io.IOException;
+
+/**
+ * A Prefetchable {@link FlatVectorsScorer} that wraps another scorer and adds prefetching capabilities for doing
+ * search. For all other cases it just delegates to the underlying scorer.
+ */
+@Log4j2
+public class PrefetchableFlatVectorScorer implements FlatVectorsScorer {
+
+    private final FlatVectorsScorer delegateScorer;
+
+    /**
+     * Constructs a new prefetchable scorer wrapper.
+     *
+     * @param delegateScorer the underlying scorer to delegate to
+     */
+    public PrefetchableFlatVectorScorer(final FlatVectorsScorer delegateScorer) {
+        this.delegateScorer = delegateScorer;
+    }
+
+    @Override
+    public RandomVectorScorerSupplier getRandomVectorScorerSupplier(
+        VectorSimilarityFunction similarityFunction,
+        KnnVectorValues vectorValues
+    ) throws IOException {
+        return delegateScorer.getRandomVectorScorerSupplier(similarityFunction, vectorValues);
+    }
+
+    @Override
+    public RandomVectorScorer getRandomVectorScorer(
+        VectorSimilarityFunction similarityFunction,
+        KnnVectorValues vectorValues,
+        float[] target
+    ) throws IOException {
+        return new PrefetchableRandomVectorScorer(
+            (RandomVectorScorer.AbstractRandomVectorScorer) delegateScorer.getRandomVectorScorer(similarityFunction, vectorValues, target)
+        );
+    }
+
+    @Override
+    public RandomVectorScorer getRandomVectorScorer(
+        VectorSimilarityFunction similarityFunction,
+        KnnVectorValues vectorValues,
+        byte[] target
+    ) throws IOException {
+        return new PrefetchableRandomVectorScorer(
+            (RandomVectorScorer.AbstractRandomVectorScorer) delegateScorer.getRandomVectorScorer(similarityFunction, vectorValues, target)
+        );
+    }
+
+    @Override
+    public String toString() {
+        return "PrefetchableFlatVectorScorer()";
+    }
+
+    /**
+     * A {@link RandomVectorScorer} that prefetches vector data before bulk scoring operations during search.
+     *
+     * <p>This scorer delegates all operations to an underlying scorer, but intercepts {@link
+     * #bulkScore} to prefetch the required vectors before scoring.
+     */
+    @Log4j2
+    static class PrefetchableRandomVectorScorer extends RandomVectorScorer.AbstractRandomVectorScorer {
+
+        private final RandomVectorScorer.AbstractRandomVectorScorer delegate;
+
+        /**
+         * Constructs a new prefetchable random vector scorer.
+         *
+         * @param delegate the underlying scorer to delegate to
+         */
+        public PrefetchableRandomVectorScorer(final RandomVectorScorer.AbstractRandomVectorScorer delegate) {
+            super(delegate.values());
+            this.delegate = delegate;
+        }
+
+        @Override
+        public float score(int node) throws IOException {
+            return delegate.score(node);
+        }
+
+        /**
+         * Scores multiple nodes with prefetching optimization.
+         *
+         * <p>Prefetches vector data before delegating to the underlying scorer for improved
+         * performance.
+         *
+         * @param nodes    array of node ordinals to score
+         * @param scores   output array for computed scores
+         * @param numNodes number of nodes to score
+         * @return the maximum score computed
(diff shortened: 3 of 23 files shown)


Vikasht34:
Good PR !! Wrapping FlatVectorScorer and passing to format is really good !!

src/main/java/org/opensearch/knn/common/featureflags/KNNFeatureFlags.java
```diff
@@ -26,6 +26,8 @@ public class KNNFeatureFlags {
 
     // Feature flags
     private static final String KNN_FORCE_EVICT_CACHE_ENABLED = "knn.feature.cache.force_evict.enabled";
+    private static final String KNN_PREFETCH_ENABLED = "knn.feature.prefetch.enabled";
+    private static final boolean KNN_PREFETCH_ENABLED_DEFAULT_VALUE = true;
```
Vikasht34, line 30: Are we enabling by default True ? Why?

src/main/java/org/opensearch/knn/index/codec/scorer/PrefetchableFlatVectorScorer.java
```diff
@@ -0,0 +1,135 @@
...
+         *
+         * @param nodes    array of node ordinals to score
+         * @param scores   output array for computed scores
+         * @param numNodes number of nodes to score
+         * @return the maximum score computed
+         * @throws IOException if an I/O error occurs
+         */
+        @Override
+        public float bulkScore(int[] nodes, float[] scores, int numNodes) throws IOException {
+            PrefetchableVectorValuesHelper.mayBeDoPrefetch(values(), nodes, numNodes);
```
Vikasht34, line 111: so with this everytime bulkScore gets called , we are making System call? Curious how this will be helpful when we have high throughput

src/main/java/org/opensearch/knn/index/codec/scorer/PrefetchHelper.java
```diff
@@ -0,0 +1,128 @@
...
+        final long baseOffset,
+        final long oneVectorByteSize,
+        final int[] ordsToPrefetch,
+        final int numOrds
+    ) throws IOException {
+        long[] offsets = calculateAndSortOffsets(baseOffset, oneVectorByteSize, ordsToPrefetch, numOrds);
+        int groupCount = 1;
+        long groupStartOffset = offsets[0];
+
+        for (int i = 1; i < numOrds; i++) {
```
Vikasht34, line 97: When offsets[i] is too far, we close the current group (prefetch up to i-1) ,then We set groupStartOffset = offsets[i] — starting a new group with the current vectorThe loop continues, and this vector becomes part of the new group.  This is somehow bit difficult to follow , I feel we should track groupEnd explicitly instead of recalculating.

```
if (offsets[i] - groupStart <= BYTES_128) {
            groupEnd = currentEnd;
        } else {
            indexInput.prefetch(groupStart, groupEnd - groupStart);
            groupCount++;
            groupStart = offsets[i];
            groupEnd = currentEnd;
        }
```

src/main/java/org/opensearch/knn/common/featureflags/KNNFeatureFlags.java
navneet1v: We want to provide the benefits hence enabling it by default. With this change and our nightlies we can guage the performance before release too

src/main/java/org/opensearch/knn/index/codec/scorer/PrefetchableFlatVectorScorer.java
navneet1v: Not sure what is the question here. In high throughput if everything is mapped the sys call will return. Also, we know that Prefetch is done in exponential backoff scenario, so at high throughput there will be no prefetches.

src/main/java/org/opensearch/knn/index/codec/scorer/PrefetchHelper.java
navneet1v: No this is not what is happening. We are checking if vector end is in the 128KB range and not the vector start. 
```
(offsets[i] + oneVectorByteSize) - groupStartOffset > BYTES_128
```

Simple idea here is, take the group start offset, add the next vector bytes to it and see if its in 128KB range or not. If its out side do prefetch other wise move to next vector.

src/main/java/org/opensearch/knn/index/codec/scorer/PrefetchableVectorValuesHelper.java
```diff
@@ -0,0 +1,66 @@
...
+            case FaissIndexBinaryFlat.ByteVectorValuesImpl binaryImpl:
+                binaryImpl.prefetch(nodes, numNodes);
+                break;
+            case HasIndexSlice luceneKNNVectorValues:
+                // Since Lucene uses HasIndexSlice, we can use the slice to prefetch and for lucene base offset for
+                // sliced index input is always 0.
+                PrefetchHelper.prefetch(luceneKNNVectorValues.getSlice(), 0, vectorValues.getVectorByteLength(), nodes, numNodes);
+                break;
+            default:
+                log.info("Not able to do prefetch on instance {}", vectorValues.getClass().getSimpleName());
```
0ctopus13prime: NIT : Should we have `log.debug` instead?

src/main/java/org/opensearch/knn/index/codec/scorer/PrefetchableVectorValuesHelper.java
```diff
@@ -0,0 +1,66 @@
...
+            case FaissIndexFloatFlat.FloatVectorValuesImpl floatImpl:
+                floatImpl.prefetch(nodes, numNodes);
+                break;
+            case FaissIndexBinaryFlat.ByteVectorValuesImpl binaryImpl:
+                binaryImpl.prefetch(nodes, numNodes);
+                break;
+            case HasIndexSlice luceneKNNVectorValues:
+                // Since Lucene uses HasIndexSlice, we can use the slice to prefetch and for lucene base offset for
+                // sliced index input is always 0.
+                PrefetchHelper.prefetch(luceneKNNVectorValues.getSlice(), 0, vectorValues.getVectorByteLength(), nodes, numNodes);
```
0ctopus13prime, line 59: I think this always allocates memory for sorting ords.
From my understanding, the flow is:
- PrefetchHelper.prefetch
- PrefetchHelper.prefetchExactVectorSize
- calculateAndSortOffsets → this allocates a new long[]

In the cohere-10m case, the node visit count is typically around 10K, with an average of 16 neighbors. That results in about 1.22 MB of allocation per query (10000 x 16 x 8 bytes).

I think this can be improved to reduce per-query memory allocation.

src/main/java/org/opensearch/knn/index/codec/scorer/PrefetchableVectorValuesHelper.java
```diff
@@ -0,0 +1,66 @@
...
+     * If the provided {@link KnnVectorValues} implementation supports prefetching, this method will issue
+     * prefetch hints for the specified nodes. Otherwise, it logs an informational message and returns without
+     * performing any prefetch.
+     *
+     * @param vectorValues the vector values instance to prefetch from
+     * @param nodes        array of node ordinals whose vector data should be prefetched
+     * @param numNodes     number of valid entries in the {@code nodes} array to prefetch
+     * @throws IOException if an I/O error occurs during prefetching
+     */
+    public static void mayBeDoPrefetch(final KnnVectorValues vectorValues, final int[] nodes, final int numNodes) throws IOException {
```
0ctopus13prime, line 48: NIT : I'm  not English expert, but this should be maybeDoPrefetch?

src/main/java/org/opensearch/knn/index/codec/scorer/PrefetchableVectorValuesHelper.java
```diff
@@ -0,0 +1,66 @@
...
+        switch (vectorValues) {
+            case FaissIndexFloatFlat.FloatVectorValuesImpl floatImpl:
+                floatImpl.prefetch(nodes, numNodes);
+                break;
+            case FaissIndexBinaryFlat.ByteVectorValuesImpl binaryImpl:
+                binaryImpl.prefetch(nodes, numNodes);
+                break;
+            case HasIndexSlice luceneKNNVectorValues:
```
0ctopus13prime, line 56: Hoping to understand, in what case this will hit this line? 🤔

src/main/java/org/opensearch/knn/index/codec/scorer/PrefetchHelper.java
```diff
@@ -0,0 +1,128 @@
...
+ * the end of a vector (start offset + vector size) falls within 128KB from the group's start offset.
+ * This minimizes I/O operations while fetching only the exact bytes needed.
+ * <p>
+ * Prefetching is controlled by {@link KNNFeatureFlags#isPrefetchEnabled()}. When disabled,
+ * no prefetch operations are performed.
+ * <p>
+ * <b>Example:</b><br>
+ * Given vectors at offsets [100KB, 120KB, 300KB] with 10KB vector size:
+ * <ul>
+ *   <li>Group 1: Prefetch 40KB (100KB to 130KB, covers vectors at 100KB and 120KB)</li>
```
0ctopus13prime, line 30: Ah.. not sure I'm following, why prefetch 40KB while we can do prefetch 10KB for the first vector and prefetch 10KB for the second vector and third vector individually.

src/main/java/org/opensearch/knn/index/codec/scorer/PrefetchHelper.java
```diff
@@ -0,0 +1,128 @@
...
+     * @throws IOException if an I/O error occurs during prefetch
+     */
+    private static void prefetchExactVectorSize(
+        final IndexInput indexInput,
+        final long baseOffset,
+        final long oneVectorByteSize,
+        final int[] ordsToPrefetch,
+        final int numOrds
+    ) throws IOException {
+        long[] offsets = calculateAndSortOffsets(baseOffset, oneVectorByteSize, ordsToPrefetch, numOrds);
```
0ctopus13prime: Can we do in-place sort in `ordsToPrefetch` then do prefetch?
I'm trying to avoid allocating long[] every single node visit.
I think then prefetch requests will be in sequentially increasing pattern which disk driver usually prefers.

src/main/java/org/opensearch/knn/index/codec/scorer/PrefetchHelper.java
```diff
@@ -0,0 +1,128 @@
...
+            prefetchExactVectorSize(indexInput, baseOffset, oneVectorByteSize, ordsToPrefetch, numOrds);
+        } else {
+            log.debug("KNNVectors Prefetch is disabled");
+        }
+    }
+
+    /**
+     * Prefetches vectors using exact byte ranges.
+     * <p>
+     * Groups vectors within 128KB ranges and prefetches only the exact bytes needed for each group.
```
0ctopus13prime, line 76: Curious, why we waste potential IO bandwidth?
With 768 dimension, one vector occupies 3KB, then at the worst case, disk can scan 128kb blocks having only two vectors.
Hoping to understand why we are not prefetching each individual vector. Is this design conceiving upcoming reordering?

src/main/java/org/opensearch/knn/index/codec/scorer/PrefetchHelper.java
```diff
@@ -0,0 +1,128 @@
...
+     * Calculates file offsets for vector ordinals and sorts them.
+     *
+     * @param baseOffset the base offset in the file where vectors start
+     * @param oneVectorByteSize the size of one vector in bytes
+     * @param ordsToPrefetch array of vector ordinals
+     * @param numOrds number of valid ordinals in the array
+     * @return sorted array of file offsets
+     */
+    private static long[] calculateAndSortOffsets(long baseOffset, long oneVectorByteSize, int[] ordsToPrefetch, int numOrds) {
+        long[] offsets = new long[numOrds];
```
0ctopus13prime: 😩

src/main/java/org/opensearch/knn/index/codec/scorer/PrefetchableVectorValuesHelper.java
```diff
@@ -0,0 +1,66 @@
...
+     *
+     * @param vectorValues the vector values instance to prefetch from
+     * @param nodes        array of node ordinals whose vector data should be prefetched
+     * @param numNodes     number of valid entries in the {@code nodes} array to prefetch
+     * @throws IOException if an I/O error occurs during prefetching
+     */
+    public static void mayBeDoPrefetch(final KnnVectorValues vectorValues, final int[] nodes, final int numNodes) throws IOException {
+        switch (vectorValues) {
+            case FaissIndexFloatFlat.FloatVectorValuesImpl floatImpl:
+                floatImpl.prefetch(nodes, numNodes);
```
0ctopus13prime, line 51: I'm struggling to understand why this is needed, can't we make FloatVectorValuesImpl implement HasIndexSlice 🤔?
the impl will use PrefetchHelper anyway.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/binary/FaissIndexBinaryFlat.java
```diff
@@ -57,35 +58,40 @@ public FloatVectorValues getFloatValues(IndexInput indexInput) throws IOExceptio
...
-            public ByteVectorValues copy() {
-                return new ByteVectorValuesImpl(indexInput.clone());
-            }
-        }
 
         return new ByteVectorValuesImpl(indexInput);
     }
+
+    @RequiredArgsConstructor
+    public class ByteVectorValuesImpl extends ByteVectorValues {
```
0ctopus13prime, line 66: Can we make this implement HasIndexSlice?

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissIndexFloatFlat.java
```diff
@@ -81,39 +82,43 @@ public VectorEncoding getVectorEncoding() {
...
         return new FloatVectorValuesImpl(indexInput);
     }
 
     @Override
     public ByteVectorValues getByteValues(IndexInput indexInput) {
         throw new UnsupportedOperationException(getClass().getSimpleName() + " does not support " + ByteVectorValues.class.getSimpleName());
     }
+
+    @RequiredArgsConstructor
+    public class FloatVectorValuesImpl extends FloatVectorValues {
```
0ctopus13prime, line 94: Added the same comment above, but can we have this to implement HasIndexSlice?

src/main/java/org/opensearch/knn/index/codec/scorer/PrefetchableVectorValuesHelper.java
navneet1v: let me make it warn log.

src/main/java/org/opensearch/knn/index/codec/scorer/PrefetchableVectorValuesHelper.java
navneet1v: You have to create it, since each query is using a different neighbors. One way to fix this would be sort the ords and then calculate the offsets during prefetch. We can add this may be in next PR in terms of optimizations.

src/main/java/org/opensearch/knn/index/codec/scorer/PrefetchableVectorValuesHelper.java
navneet1v: sure, I can make the change.

src/main/java/org/opensearch/knn/index/codec/scorer/PrefetchableVectorValuesHelper.java
navneet1v: You are asking about this line:
> case HasIndexSlice luceneKNNVectorValues:

?

src/main/java/org/opensearch/knn/index/codec/scorer/PrefetchableVectorValuesHelper.java
0ctopus13prime: But allocating 1.22MB per query seems bit huge, will the next PR gets merged for 3.6?

src/main/java/org/opensearch/knn/index/codec/scorer/PrefetchableVectorValuesHelper.java
0ctopus13prime: yes! in which case it will hit that line?

src/main/java/org/opensearch/knn/index/codec/scorer/PrefetchHelper.java
navneet1v: since we are grouping the vectors which lies in 128KB blocks. Hence we will first fetch from 100 to 130KB that is 30KB prefetch and then another one 10KB. So in this case we might be prefetching 10KB extra but number of calls to prefetch has been reduced from 3 calls to 2 calls.

src/main/java/org/opensearch/knn/index/codec/scorer/PrefetchHelper.java
navneet1v: Replied here: https://github.com/opensearch-project/k-NN/pull/3173#discussion_r2950673974

src/main/java/org/opensearch/knn/index/codec/scorer/PrefetchHelper.java
navneet1v: So idea is, to avoid the amount of prefetch calls too. So we need some kind of grouping. Here grouping 128KB. We can further fine tuning as this prefetch evolves. I have added a comment on this class to see if we want to bring this 128KB group size via cluster settings.

src/main/java/org/opensearch/knn/index/codec/scorer/PrefetchableVectorValuesHelper.java
navneet1v: I have fixed it in the next PR which is only my local. Another thing to keep in mind here is even if we get indexInput slice we would need baseOffset, since FaissIndexFloatFlat.FloatVectorValuesImpl base offset is not 0 just like Lucene Vectorvalues.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/binary/FaissIndexBinaryFlat.java
navneet1v: https://github.com/opensearch-project/k-NN/pull/3173#discussion_r2950691543

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissIndexFloatFlat.java
navneet1v: https://github.com/opensearch-project/k-NN/pull/3173#discussion_r2950691543

src/main/java/org/opensearch/knn/index/codec/scorer/PrefetchHelper.java
0ctopus13prime: I get that logic, but not sure I'm understanding any benefits of reducing the number of prefetch calls with large scan. i.e. the benefit of speculative IO amplification

0ctopus13prime: I mean, in memory abundant environment, prefetch essentially has no effect since kernel will just ignore the request.
so the only scenario where prefetch shines is memory starving environment.
I'm trying to understand what benefits we will have with this speculative IO amplification versus just submit N times of prefetch requests individually.

src/main/java/org/opensearch/knn/index/codec/scorer/PrefetchableVectorValuesHelper.java
navneet1v: Does this work:

```
private static void prefetchExactVectorSize(
        final IndexInput indexInput,
        final long baseOffset,
        final long oneVectorByteSize,
        final int[] ordsToPrefetch,
        final int numOrds
    ) throws IOException {
        Arrays.sort(ordsToPrefetch, 0, numOrds);

        int groupCount = 1;
        long groupStartOffset = baseOffset + (long) ordsToPrefetch[0] * oneVectorByteSize;

        for (int i = 1; i < numOrds; i++) {
            long currentOffset = baseOffset + (long) ordsToPrefetch[i] * oneVectorByteSize;
            if ((currentOffset + oneVectorByteSize) - groupStartOffset > BYTES_128) {
                long prevOffset = baseOffset + (long) ordsToPrefetch[i - 1] * oneVectorByteSize;
                indexInput.prefetch(groupStartOffset, (prevOffset + oneVectorByteSize) - groupStartOffset);
                groupCount++;
                groupStartOffset = currentOffset;
            }
        }
        // Prefetch final group
        long lastOffset = baseOffset + (long) ordsToPrefetch[numOrds - 1] * oneVectorByteSize;
        indexInput.prefetch(groupStartOffset, (lastOffset + oneVectorByteSize) - groupStartOffset);

        log.trace("Prefetching grouped [{}] vectors where num of ords was [{}] using exact prefetch size", groupCount, numOrds);
    }
```

BTW I really hate when functions change the input which was provided. But this is the best I can think of without adding extra overhead.

src/main/java/org/opensearch/knn/index/codec/scorer/PrefetchHelper.java
0ctopus13prime: What I'm trying to understand is not on the number, rather than on the motivation behind it.
Have we saw performance degradation between submitting prefetch requests per each vector versus 128kb prefetch approach??

src/main/java/org/opensearch/knn/index/codec/scorer/PrefetchableVectorValuesHelper.java
0ctopus13prime: Thank you, if you're ok with this, hoping to revisit sorting ords without allocating spaces.

Got some helps from AI, we can make simpler further:
```
for (int i = 1; i < numOrds; i++) {
    long currentOffset = baseOffset + (long) ordsToPrefetch[i] * oneVectorByteSize;
    long endOfLastVector = baseOffset + (long) ordsToPrefetch[i-1] * oneVectorByteSize + oneVectorByteSize;
    
    // Check if adding this vector exceeds 128KB OR if the gap is too large
    boolean tooLong = (currentOffset + oneVectorByteSize) - groupStartOffset > BYTES_128;
    
    if (tooLong) {
        indexInput.prefetch(groupStartOffset, endOfLastVector - groupStartOffset);
        groupCount++;
        groupStartOffset = currentOffset;
    }
}
```

src/main/java/org/opensearch/knn/index/codec/scorer/PrefetchableVectorValuesHelper.java
0ctopus13prime: I don't think it would need baseOffset,
we can make it return indexInput.slice(baseOffset, ...) right?

0ctopus13prime (approved):

src/main/java/org/opensearch/knn/index/codec/scorer/PrefetchableVectorValuesHelper.java
navneet1v: updated.

src/main/java/org/opensearch/knn/index/codec/scorer/PrefetchableVectorValuesHelper.java
navneet1v: Yes I was trying to avoid it. But lets me raise it in the next PR, since I was going to use HasSlice interface there too.

src/main/java/org/opensearch/knn/index/codec/scorer/PrefetchHelper.java
navneet1v: this is removed now.

src/main/java/org/opensearch/knn/index/codec/scorer/PrefetchHelper.java
navneet1v: Yes we saw performance degradation with EFS. I will do more test on this and update the benchmarks results.

src/main/java/org/opensearch/knn/index/codec/scorer/PrefetchableVectorValuesHelper.java
navneet1v: The case when we are trying to do the prefetch over VectorValues that comes from Lucene. So think of cases like where we do exact search using the vector values which are on `.vec` file

PR #3174, opened 2026-03-16
Fix unit test: testCreate_whenRescoreWithoutExpandNested_thenWrapWithRescoreQuery

naveentatikonda:
### Description
Fix unit test `testCreate_whenRescoreWithoutExpandNested_thenWrapWithRescoreQuery` which was failing in some scenarios with the below error

```
REPRODUCE WITH: ./gradlew ':test' --tests 'org.opensearch.knn.index.query.KNNQueryFactoryTests.testCreate_whenRescoreWithoutExpandNested_thenWrapWithRescoreQuery' -Dtests.seed=7BFAFB7F2D14E1BA -Dtests.security.manager=false -Dtests.locale=en-DM -Dtests.timezone=Africa/Djibouti -Druntime.java=21
  2> java.lang.NullPointerException: Cannot invoke "org.opensearch.common.settings.Settings.getSecureSettings()" because "settings" is null
        at __randomizedtesting.SeedInfo.seed([7BFAFB7F2D14E1BA:1FF71CC5F9C9830C]:0)
        at org.opensearch.common.settings.Setting.innerGetRaw(Setting.java:572)
        at org.opensearch.common.settings.Setting.getRaw(Setting.java:561)
        at org.opensearch.common.settings.Setting.get(Setting.java:494)
        at org.opensearch.common.settings.Setting.get(Setting.java:490)
        at org.opensearch.knn.indices.ModelDao$OpenSearchKNNModelDao.<init>(ModelDao.java:214)
        at org.opensearch.knn.indices.ModelDao$OpenSearchKNNModelDao.getInstance(ModelDao.java:202)
        at org.opensearch.knn.index.query.RescoreKNNVectorQuery.<init>(RescoreKNNVectorQuery.java:64)
        at org.opensearch.knn.index.query.KNNQueryFactory.create(KNNQueryFactory.java:163)
        at org.opensearch.knn.index.query.KNNQueryFactoryTests.testCreate_whenRescoreWithoutExpandNested_thenWrapWithRescoreQuery(KNNQueryFactoryTests.java:610)
```
* The test `testCreate_whenRescoreWithoutExpandNested_thenWrapWithRescoreQuery` was flaky due to a static singleton ordering hazard in `ModelDao.OpenSearchKNNModelDao`. Its `INSTANCE` field is only initialized on first call to `getInstance()`, which invokes `new OpenSearchKNNModelDao()` — reading the static `settings` field. In production this is safe because `initialize(client, clusterService, settings)` is always called during plugin startup. In tests, if this test ran before any other test that triggered `ModelDao` initialization, `settings` was `null`, causing the NPE. Fixed by intercepting the `RescoreKNNVectorQuery` construction with `Mockito.mockConstruction`, avoiding the `ModelDao.getInstance()` call entirely.

### Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,4 +28,5 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 * Use right Vector Scorer when segments are initialized using SPI and also corrected the maxConn for MOS [#3117](https://github.com/opensearch-project/k-NN/pull/3117)
 * Use pre-quantized vectors for ADC [#3113](https://github.com/opensearch-project/k-NN/pull/3113)
 * Upgrade Lucene to 10.4.0 [#3135](https://github.com/opensearch-project/k-NN/pull/3135)
+* Speedup FP16 bulk similarity by precomputing the tail mask [#3172](https://github.com/opensearch-project/k-NN/pull/3172)
 * Optimize ByteVectorIdsExactKNNIterator by moving array conversion to constructor [#3171](https://github.com/opensearch-project/k-NN/pull/3171)
--- a/jni/src/simd/similarity_function/avx512_simd_similarity_function.cpp
+++ b/jni/src/simd/similarity_function/avx512_simd_similarity_function.cpp
@@ -30,6 +30,14 @@ struct AVX512SPRFP16MaxIP final : BaseSimilarityFunction<BulkScoreTransformFunc,
         constexpr int32_t vecBlock = 8;
         // Maximum number of elements to load at the same time
         constexpr int32_t elemPerLoad = 16;
+
+        // SIMD-aligned dim and tail dim
+        const int32_t simdDim = (dim / elemPerLoad) * elemPerLoad;
+        const int32_t tailDim = dim - simdDim;
+
+        // Precompute tail mask
+        const __mmask16 tailMask = tailDim > 0 ? (__mmask16)((1U << tailDim) - 1) : 0;
+
         // Tracking accumulated summation per each vector
         // FYI : IP = Sum(v1[i] * v2[i])
         __m512 sum[vecBlock];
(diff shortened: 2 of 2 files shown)


src/test/java/org/opensearch/knn/index/query/KNNQueryFactoryTests.java
```diff
@@ -607,10 +607,13 @@ public void testCreate_whenRescoreWithoutExpandNested_thenWrapWithRescoreQuery()
...
-        Query query = KNNQueryFactory.create(createQueryRequest);
 
-        // Should return RescoreKNNVectorQuery wrapper
-        assertEquals(RescoreKNNVectorQuery.class, query.getClass());
+        try (MockedConstruction<RescoreKNNVectorQuery> mockedConstruction = Mockito.mockConstruction(RescoreKNNVectorQuery.class)) {
+            Query query = KNNQueryFactory.create(createQueryRequest);
+            // Should return RescoreKNNVectorQuery wrapper
+            assertEquals(1, mockedConstruction.constructed().size());
+            assertSame(mockedConstruction.constructed().get(0), query);
+        }
```
navneet1v, line 618: rather than doing this, won't it be better to mock the call of `ModelDao.OpenSearchKNNModelDao.getInstance()` which is happening in the `RescoreKNNVectorQuery` ? Also RescoreKNNVectorQuery is a Lucene query, it shouldn't have instance of ExactSearcher and if its getting that then I would probably fix that pattern.

src/main/java/org/opensearch/knn/indices/ModelDao.java
```diff
@@ -210,6 +211,11 @@ public static void initialize(Client client, ClusterService clusterService, Sett
             OpenSearchKNNModelDao.settings = settings;
         }
 
+        @VisibleForTesting
+        public static synchronized void resetInstance() {
+            INSTANCE = null;
+        }
```
navneet1v: Lets avoid such changes

src/test/java/org/opensearch/knn/index/query/KNNQueryFactoryTests.java
navneet1v: Example:
```
try (MockedStatic<ModelDao.OpenSearchKNNModelDao> mockedModelDao = mockStatic(ModelDao.OpenSearchKNNModelDao.class)) {
            ModelDao.OpenSearchKNNModelDao mockInstance = mock(ModelDao.OpenSearchKNNModelDao.class);
            mockedModelDao.when(ModelDao.OpenSearchKNNModelDao::getInstance).thenReturn(mockInstance);
            Query query = KNNQueryFactory.create(createQueryRequest);
            // Should return RescoreKNNVectorQuery wrapper
            assertEquals(RescoreKNNVectorQuery.class, query.getClass());
        }
```

Also a similar fix is needed in all the tests where you are using RescoreKNNVectorQuery

naveentatikonda: mocking makes sense but we need `ExactSearcher` in `RescoreKNNVectorQuery ` as we need to do exact search(rescore) with fp32 vectors. Should I avoid initializing it in the constructor  ?

navneet1v: Sure. Lets move to mocking based thing here.

PR #3179, opened 2026-03-17
Add NestedBestChildVectorScorer and KnnBinaryDocValuesScorer for exact search

VijayanB:
### Description
Introduce VectorScorer implementations for cases where Lucene's built-in vector values do not provide a scorer:

- NestedBestChildVectorScorer: Groups child documents by parent and returns the best-scoring child per parent. Adapted from Lucene's DiversifyingChildrenVectorScorer to implement VectorScorer. Supports both filtered and unfiltered iteration.

- KnnBinaryDocValuesScorer: Scores documents backed by BinaryDocValues by deserializing stored vectors and comparing against the query vector using SpaceType. Supports both float[] and byte[] query vectors via overloaded factory methods.

### Related Issues

### Check List


src/main/java/org/opensearch/knn/index/query/exactsearch/scorers/KNNBinaryDocValuesScorer.java
```diff
@@ -0,0 +1,90 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.knn.index.query.exactsearch.scorers;
```
navneet1v:
```suggestion
package org.opensearch.knn.index.query.scorers;
```

src/main/java/org/opensearch/knn/index/query/exactsearch/scorers/NestedBestChildVectorScorer.java
```diff
@@ -0,0 +1,194 @@
...
+     * @param acceptedChildrenIterator iterator over the accepted child doc ids (i.e. children that
+     *                                 pass the filter). Pass {@code null} for the unfiltered case
+     *                                 where all vector documents are considered.
+     * @param parentBitSet             a {@link BitSet} with bits set at every parent doc id.
+     *                                 Used to determine parent boundaries for grouping children.
+     * @param vectorScorer             the underlying scorer that computes similarity scores for
+     *                                 individual child documents against the query vector.
+     */
+    public NestedBestChildVectorScorer(
+        @Nullable DocIdSetIterator acceptedChildrenIterator,
```
navneet1v: I would suggest having DocIdSetIterator.all() rather than null iterator for matching all the docs.

src/main/java/org/opensearch/knn/index/query/scorers/NestedBestChildVectorScorer.java
```diff
@@ -0,0 +1,194 @@
...
+ *
+ * @see org.apache.lucene.search.VectorScorer
+ * @see org.apache.lucene.search.join.DiversifyingChildrenFloatKnnVectorQuery
+ */
+public class NestedBestChildVectorScorer implements VectorScorer {
+    private final VectorScorer vectorScorer;
+    private final DocIdSetIterator vectorIterator;
+    private final DocIdSetIterator childIterator;
+    private final BitSet parentBitSet;
+    private final DocIdSetIterator iterator;
```
navneet1v, line 72: what is the use of this third iterator? from the name of the variable it is not clear.

src/main/java/org/opensearch/knn/index/query/scorers/NestedBestChildVectorScorer.java
```diff
@@ -0,0 +1,194 @@
...
+     * best-scoring child for each successive parent. The same instance is returned on every call.
+     *
+     * @return the iterator over best-child doc ids.
+     */
+    @Override
+    public DocIdSetIterator iterator() {
+        return iterator;
+    }
+
+    private DocIdSetIterator createIterator() {
```
navneet1v, line 141: need docs on this function

src/main/java/org/opensearch/knn/index/query/scorers/NestedBestChildVectorScorer.java
```diff
@@ -0,0 +1,194 @@
...
+            @Override
+            public int docID() {
+                return bestChild;
+            }
+
+            @Override
+            public int nextDoc() throws IOException {
+                int nextChild = nextChildDoc();
+                if (nextChild == NO_MORE_DOCS) {
+                    bestChild = NO_MORE_DOCS;
```
navneet1v, line 155: why if nextChild is NO_MORE_DOCS we are setting bestChild as NO_MORE_DOCS

src/main/java/org/opensearch/knn/index/query/exactsearch/scorers/NestedBestChildVectorScorer.java
```diff
@@ -0,0 +1,194 @@
...
+                return docId == -1 ? childIterator.nextDoc() : docId;
+            }
+
+            private void advanceVectorIterator(int target) throws IOException {
+                if (childIterator == null) {
+                    // Unfiltered: vectorIterator is already positioned via nextChildDoc
+                    vectorIterator.nextDoc();
+                } else {
+                    vectorIterator.advance(target);
+                    childIterator.nextDoc();
```
navneet1v: I am hoping that we don't face EOF exception here if 1 iterator has less DocIds than the other

src/main/java/org/opensearch/knn/index/query/scorers/NestedBestChildVectorScorer.java
```diff
@@ -0,0 +1,194 @@
...
+                } else {
+                    vectorIterator.advance(target);
+                    childIterator.nextDoc();
+                }
+            }
+
+            @Override
+            public int advance(int target) {
+                throw new UnsupportedOperationException();
+            }
```
navneet1v, line 183: why unsupported exception here?

VijayanB: Thank you for raising the question about the UnsupportedOperationException in the advance(int target) method. I've provided my reasoning below. If you have any suggestions for improvement, alternative approaches, or feedback, please share—I'd definitely like to consider them.

### Reasoning

The `advance(int target)` method throws `UnsupportedOperationException` because implementing it correctly would violate the semantic contract of `advance()` for nested document scoring.

### The Fundamental Problem

According to the `DocIdSetIterator.advance()` contract, the method should advance to "the first document whose number is greater than or equal to target." However, for `NestedBestChildVectorScorer`, this creates an impossible situation:

**What if the best-scoring child within a parent group appears *before* the target document?**

Consider this scenario:
- Parent document at position 100
- Child documents at positions 95, 96, 97, 98, 99 (all belong to parent 100)
- `advance(98)` is called
- The best-scoring child is at position 96

If we honor the `advance()` contract and skip to position 98, we would:
1. **Miss the best child** (at position 96) for this parent group
2. **Return incorrect results** since we're supposed to return the best-scoring child per parent
3. **Violate the scorer's core semantic** of finding the optimal child within each parent group

### Why Sequential Iteration is Required

The scorer must process **all children within a parent group sequentially** to:
- Evaluate each child's score against the query vector
- Track the best-scoring child for that parent
- Return the parent's position with the best child's score

Jumping to an arbitrary target with `advance()` would require either:
- Backtracking to process earlier children (violating the forward-only iterator contract)
- Accepting incomplete/incorrect results (violating the scoring semantics)

src/main/java/org/opensearch/knn/index/query/exactsearch/scorers/NestedBestChildVectorScorer.java
VijayanB: i believe Using DocIdSetIterator.all() would include both parent and child documents in the iteration. Let me see whether i can improve the code to make it more simpler and straight forward.

src/main/java/org/opensearch/knn/index/query/exactsearch/scorers/KNNBinaryDocValuesScorer.java
VijayanB: Will update

src/main/java/org/opensearch/knn/index/query/scorers/NestedBestChildVectorScorer.java
VijayanB: This is basically the iterator of nested scorer. This will be returned if caller asks iterator from the scorer.

src/main/java/org/opensearch/knn/index/query/scorers/NestedBestChildVectorScorer.java
VijayanB: Sure. Will add.

src/main/java/org/opensearch/knn/index/query/scorers/NestedBestChildVectorScorer.java
VijayanB: so that docId can return NO_MORE_DOCS

src/main/java/org/opensearch/knn/index/query/exactsearch/scorers/NestedBestChildVectorScorer.java
VijayanB: Good point. Will take a look again and see whether this is possible or not

src/main/java/org/opensearch/knn/index/query/scorers/KNNBinaryDocValuesScorer.java
```diff
@@ -0,0 +1,90 @@
...
+import org.apache.lucene.search.VectorScorer;
+import org.apache.lucene.util.ArrayUtil;
+import org.apache.lucene.util.BytesRef;
+import org.opensearch.knn.index.KNNVectorSimilarityFunction;
+import org.opensearch.knn.index.SpaceType;
+import org.opensearch.knn.index.codec.util.KNNVectorAsCollectionOfFloatsSerializer;
+
+import java.io.IOException;
+
+/**
```
Vikasht34, line 18: After reading this Class summary , I am not getting what this class is doing? What is intent of this class and how this will be used?

```
Document vectors stored as serialized bytes in Lucene's {@link BinaryDocValues} format are
 * deserialized on the fly and compared against the query vector using the provided
```

src/main/java/org/opensearch/knn/index/query/scorers/KNNBinaryDocValuesScorer.java
```diff
@@ -0,0 +1,90 @@
...
+ *       {@link BytesRef}.</li>
+ * </ul>
+ */
+public class KNNBinaryDocValuesScorer implements VectorScorer {
+
+    private final BinaryDocValues binaryDocValues;
+    private final ScoreFunction scoreFunction;
+
+    @FunctionalInterface
+    private interface ScoreFunction {
```
Vikasht34, line 51: Who uses this Inteface? and there is not test for to invoke this, how are we planning to use this functional inteface and why do u think it's needed here?

src/main/java/org/opensearch/knn/index/query/scorers/KNNBinaryDocValuesScorer.java
```diff
@@ -0,0 +1,90 @@
...
+    public static KNNBinaryDocValuesScorer create(byte[] queryVector, BinaryDocValues binaryDocValues, SpaceType spaceType) {
+        return new KNNBinaryDocValuesScorer(binaryDocValues, bytesRef -> {
+            byte[] docVector = ArrayUtil.copyOfSubArray(bytesRef.bytes, bytesRef.offset, bytesRef.offset + bytesRef.length);
+            return spaceType.getKnnVectorSimilarityFunction().compare(queryVector, docVector);
+        });
+    }
+
+    @Override
+    public float score() throws IOException {
+        return scoreFunction.score(binaryDocValues.binaryValue());
```
Vikasht34, line 92: Why I need functional inteface here?

src/main/java/org/opensearch/knn/index/query/exactsearch/scorers/NestedBestChildVectorScorer.java
VijayanB: As we discussed, we can't assume that acceptedIterator will always contain documents with vectors. I'll use Conjunction.intersection to handle this properly. Do you have any other suggestions for addressing this issue?

src/main/java/org/opensearch/knn/index/query/exactsearch/scorers/ConjunctionUtils.java
```diff
@@ -0,0 +1,35 @@
...
+import java.util.List;
+
+/**
+ * Utility for creating conjunction (intersection) iterators over multiple {@link DocIdSetIterator}s.
+ * <p>
+ * This is used to ensure that an accepted-children iterator only yields document IDs that also
+ * exist in the vector values, preventing score misattribution when a child document passes a
+ * filter but has no vector.
+ */
+public final class ConjunctionUtils {
```
Vikasht34: we already have same in Lucene ?

Vikasht34 (dismissed):
Looks Good , Please put all the comments and explanation

src/main/java/org/opensearch/knn/index/query/scorers/KNNBinaryDocValuesScorer.java
VijayanB: updated java doc

src/main/java/org/opensearch/knn/index/query/scorers/KNNBinaryDocValuesScorer.java
VijayanB: updated in java doc

src/main/java/org/opensearch/knn/index/query/scorers/KNNBinaryDocValuesScorer.java
VijayanB: added explanation at top

src/main/java/org/opensearch/knn/index/query/exactsearch/scorers/ConjunctionUtils.java
VijayanB: Good call out. I removed and added as inline method to Scorer

Vikasht34 (approved):
@VijayanB  Really Liked the Scorer , especially right usage of Functional Interface

src/main/java/org/opensearch/knn/index/query/exactsearch/scorers/NestedBestChildVectorScorer.java
navneet1v: Lets use the Conjunction.intersection to handle this

src/main/java/org/opensearch/knn/index/query/scorers/NestedBestChildVectorScorer.java
```diff
@@ -0,0 +1,191 @@
...
+     *
+     * @param filterIdsIterator iterator over the accepted child doc ids (i.e. children that
+     *                                 pass the filter). Pass {@code null} for the unfiltered case
+     *                                 where all vector documents are considered.
+     * @param parentBitSet             a {@link BitSet} with bits set at every parent doc id.
+     *                                 Used to determine parent boundaries for grouping children.
+     * @param childrenVectorScorer             the underlying scorer that computes similarity scores for
+     *                                 individual child documents against the query vector.
+     */
+    public NestedBestChildVectorScorer(
```
navneet1v, line 92: any reason for keeping this constructor public but for the BinaryDocValuesScorer we are using a `create` method? can we keep the creation interfaces same?

VijayanB: Good point. I actually made it package private in next PR  #3183

Issue #3176, opened 2026-03-17
[Faiss BBQ] Search on Faiss BBQ index.


0ctopus13prime:
## Description
Implement memory-optimized search for the Faiss BBQ index.

When io_flags=IO_FLAG_SKIP_STORAGE is used, the .faiss file will not contain flat vectors. Because of this, the codec must provide a supplier of FaissIndex so that Faiss BBQ storage can be plugged in at runtime.

The reason for plugging FaissIndex in as the bottom storage layer is to support vector-ordinal to document-ID conversion. [code](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissIdMapIndex.java#L135) The KnnVectorValues returned from the storage layer will be wrapped by FaissIdMappingIndex, which handles this conversion.

Therefore, the codec should supply a FaissIndex instance and place it under the FaissIdMappingIndex hierarchy so that the mapping logic can function correctly.

## Implementations
For this we should expand parameters in below interface to take Supplier<FaissIndex>
- [VectorSearcherFactory.createVectorSearcher](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/memoryoptsearch/VectorSearcherFactory.java#L29)
- [FaissIndex.load](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissIndex.java#L52)

Then in [FaissMemoryOptimizedSearcher](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissMemoryOptimizedSearcher.java#L84), we should make get `FlatVectorsScorer` from `FaissIndex` which will have vector scorer for BBQ.

navneet1v:
closing this GH as the feature is released.

PR #3182, opened 2026-03-17
Faiss SQ 1 bit MOS changes

naveentatikonda:
### Description

Enables memory-optimized search (MOS) for Faiss SQ (1-bit) indices. When SQ with 1 bit is used, Faiss writes the HNSW graph without flat vector storage (`IO_FLAG_SKIP_STORAGE`), and quantized vectors are stored separately via Lucene's `Lucene104ScalarQuantizedVectorsFormat`. This PR bridges the two by wiring Lucene's quantized reader into the Faiss index hierarchy at search time, and routes scoring through a SIMD-accelerated path.

#### Key Changes

**`AbstractNativeEnginesKnnVectorsReader` (new class)**
- Extracted shared infrastructure from `NativeEngines990KnnVectorsReader` into an abstract base class
- Owns `flatVectorsReader`, `segmentReadState`, `ioContext`, `vectorSearcherHolder`, and the double-checked locking pattern for lazy searcher initialization
- `getVectorSearcherSupplier` passes the full `flatVectorsReader` to `VectorSearcherFactory.createVectorSearcher` so factories can access both scorer and vector data
- `NativeEngines990KnnVectorsReader` and `Faiss1040ScalarQuantizedKnnVectorsReader` both extend this class

**`Faiss1040ScalarQuantizedKnnVectorsReader`**
- Extends `AbstractNativeEnginesKnnVectorsReader`; always forces memory-optimized search, not gated by the index-level setting
- No quantization state cache — quantization is handled by Lucene, not k-NN's framework
- No `NativeMemoryCacheManager` invalidation on close
- Loads the searcher before the null-target check so warmup triggers memory-mapping; throws `WarmupInitializationException` on null target
- Byte vector search is not supported; throws `UnsupportedOperationException`

**`FaissEmptyIndex` (new class)**
- Sentinel `FaissIndex` returned by `FaissIndex.load` when a `"null"` section name is read (Faiss SQ with 1 bit skips flat storage via `IO_FLAG_SKIP_STORAGE`)
- All operations throw `UnsupportedOperationException`; the caller (`FaissFlatIndexFactory`) is expected to replace it with a concrete flat index

**`FaissScalarQuantizedFlatIndex` (new class)**
- Virtual `FaissIndex` backed by a `FlatVectorsReader`, no-op `doLoad`
- Delegates `getFloatValues` to the reader; `getByteValues` throws `UnsupportedOperationException`
- Installed as flat storage under `AbstractFaissHNSWIndex` for SQ (1-bit) indices

**`FaissFlatIndexFactory` (new class)**
- `create(fieldInfo, flatVectorsReader)`: returns a `FaissScalarQuantizedFlatIndex` when the field is SQ (1-bit), otherwise `null`
- `maybeSetFlatIndex(faissIndex, fieldInfo, flatVectorsReader)`: walks the `FaissIdMapIndex → AbstractFaissHNSWIndex` chain; if flat storage is a `FaissEmptyIndex`, wires in the result of `create`; throws `IllegalStateException` if `create` returns null unexpectedly
- Both methods are package-private

**`FlatVectorsScorerProvider.getFlatVectorsScorer`**
- Added SQ (1-bit) branch: Verifies FieldInfo and if it is an SQ index with 1 bit, returns `new Faiss104ScalarQuantizedVectorScorer(delegateScorer)`

**`FaissIndex.load`**
- Returns `FaissEmptyIndex` when it reads a `"null"` section name (SQ with 1 bit skips flat storage via `IO_FLAG_SKIP_STORAGE`)

**`FaissMemoryOptimizedSearcherFactory.createVectorSearcher`**
- Calls `FaissIndex.load(indexInput)` then `FaissFlatIndexFactory.maybeSetFlatIndex(faissIndex, fieldInfo, flatVectorsReader)` separately before constructing the searcher

**`FaissMemoryOptimizedSearcher` constructor**
- Changed from 3-arg `(IndexInput, FieldInfo, FlatVectorsScorer)` to 4-arg `(IndexInput, FaissIndex, FieldInfo, FlatVectorsScorer)`
- The factory now owns loading and transforming the `FaissIndex` before passing it in

**`AbstractFaissHNSWIndex`**
- Added `@Setter` on `flatVectors` to allow `FaissFlatIndexFactory` to wire in `FaissScalarQuantizedFlatIndex`

### Related Issues
https://github.com/opensearch-project/k-NN/issues/3176

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss1040ScalarQuantizedKnnVectorsReader.java
+++ b/src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss1040ScalarQuantizedKnnVectorsReader.java
@@ -8,7 +8,6 @@ package org.opensearch.knn.index.codec.KNN1040Codec;
 import lombok.extern.log4j.Log4j2;
 import org.apache.lucene.codecs.hnsw.FlatVectorsReader;
 import org.apache.lucene.index.ByteVectorValues;
-import org.apache.lucene.index.FieldInfo;
 import org.apache.lucene.index.SegmentReadState;
 import org.apache.lucene.search.AcceptDocs;
 import org.apache.lucene.index.FloatVectorValues;
@@ -38,8 +37,6 @@ import java.io.IOException;
  */
 @Log4j2
 public class Faiss1040ScalarQuantizedKnnVectorsReader extends AbstractNativeEnginesKnnVectorsReader {
-    private volatile FieldInfo fieldInfo;
-
     Faiss1040ScalarQuantizedKnnVectorsReader(SegmentReadState state, FlatVectorsReader flatVectorsReader) {
         super(state, flatVectorsReader);
     }
@@ -56,14 +53,7 @@ public class Faiss1040ScalarQuantizedKnnVectorsReader extends AbstractNativeEngi
      */
     @Override
     public void search(String field, float[] target, KnnCollector knnCollector, AcceptDocs acceptDocs) throws IOException {
-        if (this.fieldInfo == null) {
-            synchronized (vectorSearcherHolderLockObject) {
-                if (this.fieldInfo == null) {
-                    this.fieldInfo = segmentReadState.fieldInfos.fieldInfo(field);
-                }
-            }
-        }
-        final VectorSearcher memoryOptimizedSearcher = loadMemoryOptimizedSearcherIfRequired(this.fieldInfo);
+        final VectorSearcher memoryOptimizedSearcher = loadMemoryOptimizedSearcherIfRequired(fieldInfos.fieldInfo(field));
 
         // On warmup, target is null. We load the searcher first to trigger memory-mapping of vectors(partial load),
         // then throw instead of returning silently so the warmup call is detectable and can be validated in tests.
--- a/src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
+++ b/src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
@@ -70,7 +70,7 @@ public class NativeEngines990KnnVectorsReader extends AbstractNativeEnginesKnnVe
      */
     @Override
     public ByteVectorValues getByteVectorValues(final String field) throws IOException {
-        final FieldInfo fieldInfo = segmentReadState.fieldInfos.fieldInfo(field);
+        final FieldInfo fieldInfo = fieldInfos.fieldInfo(field);
         if (fieldInfo.getVectorEncoding() == VectorEncoding.FLOAT32) {
             final ByteVectorValues quantizedVectorValues = getQuantizedVectorValues(fieldInfo);
             if (quantizedVectorValues != null) {
@@ -111,7 +111,7 @@ public class NativeEngines990KnnVectorsReader extends AbstractNativeEnginesKnnVe
         // TODO: This is a temporary hack where we are using KNNCollector to initialize the quantization state.
         if (knnCollector instanceof QuantizationConfigKNNCollector) {
             String cacheKey = quantizationStateCacheKeyPerField.get(field);
-            FieldInfo fieldInfo = segmentReadState.fieldInfos.fieldInfo(field);
+            FieldInfo fieldInfo = fieldInfos.fieldInfo(field);
             QuantizationState quantizationState = QuantizationStateCacheManager.getInstance()
                 .getQuantizationState(
                     new QuantizationStateReadConfig(
@@ -124,7 +124,7 @@ public class NativeEngines990KnnVectorsReader extends AbstractNativeEnginesKnnVe
             ((QuantizationConfigKNNCollector) knnCollector).setQuantizationState(quantizationState);
             return;
         }
-        final FieldInfo fieldInfo = segmentReadState.fieldInfos.fieldInfo(field);
+        final FieldInfo fieldInfo = fieldInfos.fieldInfo(field);
         if (trySearchWithMemoryOptimizedSearch(fieldInfo, target, knnCollector, acceptDocs, true)) {
             return;
         }
@@ -159,7 +159,7 @@ public class NativeEngines990KnnVectorsReader extends AbstractNativeEnginesKnnVe
      */
     @Override
     public void search(String field, byte[] target, KnnCollector knnCollector, AcceptDocs acceptDocs) throws IOException {
-        final FieldInfo fieldInfo = segmentReadState.fieldInfos.fieldInfo(field);
+        final FieldInfo fieldInfo = fieldInfos.fieldInfo(field);
         // searching with byte vector is not supported by ADC.
         if (trySearchWithMemoryOptimizedSearch(fieldInfo, target, knnCollector, acceptDocs, false)) {
             return;
@@ -226,7 +226,7 @@ public class NativeEngines990KnnVectorsReader extends AbstractNativeEnginesKnnVe
 
     private void loadCacheKeyMap() {
         quantizationStateCacheKeyPerField = new HashMap<>();
-        for (FieldInfo fieldInfo : segmentReadState.fieldInfos) {
+        for (FieldInfo fieldInfo : fieldInfos) {
             String cacheKey = UUIDs.base64UUID();
             quantizationStateCacheKeyPerField.put(fieldInfo.getName(), cacheKey);
         }
--- a/src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissFlatIndexFactory.java
+++ b/src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissFlatIndexFactory.java
@@ -9,10 +9,11 @@ import lombok.experimental.UtilityClass;
 import org.apache.lucene.codecs.hnsw.FlatVectorsReader;
 import org.apache.lucene.index.FieldInfo;
 import org.opensearch.knn.common.FieldInfoExtractor;
+import org.opensearch.knn.index.engine.faiss.FaissSQEncoder;
 
 /**
  * Factory that creates the appropriate {@link FaissIndex} flat storage implementation
- * based on the field's configuration.
+ * based on the field's configuration, and wires it into the index tree when needed.
  */
 @UtilityClass
 public class FaissFlatIndexFactory {
@@ -21,10 +22,34 @@ public class FaissFlatIndexFactory {
      * Returns a {@link FaissIndex} to use as flat storage for the given field, or {@code null}
      * if the FAISS file's own flat storage should be used.
      */
-    public static FaissIndex create(final FieldInfo fieldInfo, final FlatVectorsReader flatVectorsReader) {
-        if (FieldInfoExtractor.isFaissBBQ(fieldInfo)) {
+    static FaissIndex create(final FieldInfo fieldInfo, final FlatVectorsReader flatVectorsReader) {
+        if (FieldInfoExtractor.isSQField(fieldInfo)
+            && FieldInfoExtractor.extractSQConfig(fieldInfo).getBits() == FaissSQEncoder.Bits.ONE.getValue()) {
             return new FaissScalarQuantizedFlatIndex(flatVectorsReader, fieldInfo.getName());
         }
         return null;
+    }
+
+    /**
+     * If the loaded FAISS index has no flat storage (e.g. SQ with 1-bit skips it via IO_FLAG_SKIP_STORAGE),
+     * wires in the appropriate flat index via {@link #create}.
+     */
+    static void maybeSetFlatIndex(final FaissIndex faissIndex, final FieldInfo fieldInfo, final FlatVectorsReader flatVectorsReader) {
+        if (!(faissIndex instanceof FaissIdMapIndex idMapIndex)) return;
+        final FaissIndex nested = idMapIndex.getNestedIndex();
+        if (!(nested instanceof AbstractFaissHNSWIndex hnswIndex) || !(hnswIndex.getFlatVectors() instanceof FaissEmptyIndex)) return;
+
+        final FaissIndex flatIndex = create(fieldInfo, flatVectorsReader);
+        if (flatIndex == null) {
+            throw new IllegalStateException(
+                String.format(
+                    "%s found for field [%s] but %s returned null — cannot wire flat storage.",
+                    FaissEmptyIndex.class.getName(),
+                    fieldInfo.getName(),
+                    FaissFlatIndexFactory.class.getName()
+                )
+            );
+        }
+        hnswIndex.flatVectors = flatIndex;
     }
 }
--- a/src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissIndex.java
+++ b/src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissIndex.java
@@ -6,9 +6,7 @@
 package org.opensearch.knn.memoryoptsearch.faiss;
 
 import lombok.Getter;
-import org.apache.lucene.codecs.hnsw.FlatVectorsReader;
 import org.apache.lucene.index.ByteVectorValues;
-import org.apache.lucene.index.FieldInfo;
 import org.apache.lucene.index.FloatVectorValues;
 import org.apache.lucene.index.VectorEncoding;
 import org.apache.lucene.store.IndexInput;
@@ -65,43 +63,6 @@ public abstract class FaissIndex {
         final FaissIndex faissIndex = IndexTypeToFaissIndexMapping.getFaissIndex(indexType);
         faissIndex.doLoad(input);
         return faissIndex;
-    }
-
-    /**
-     * Loads a FAISS index and, if required based on the field's configuration, wires in the appropriate
-     * flat vector storage via {@link FaissFlatIndexFactory}.
-     *
-     * @param input             Input stream to a FAISS index
-     * @param fieldInfo         Field metadata used to determine the flat index type
-     * @param flatVectorsReader Reader providing both the scorer and, for certain index types (e.g. SQ (with 1 bit)),
-     *                          the backing flat vector storage
-     * @return Top level {@link FaissIndex}
-     * @throws IOException
-     */
-    public static FaissIndex load(IndexInput input, FieldInfo fieldInfo, FlatVectorsReader flatVectorsReader) throws IOException {
-        final FaissIndex faissIndex = load(input);
-        maybeSetFlatIndex(faissIndex, fieldInfo, flatVectorsReader);
-        return faissIndex;
-    }
-
-    // If the HNSW index has no flat storage (e.g. SQ (with 1 bit) skips it via IO_FLAG_SKIP_STORAGE), wire in the appropriate flat index.
-    static void maybeSetFlatIndex(final FaissIndex faissIndex, final FieldInfo fieldInfo, final FlatVectorsReader flatVectorsReader) {
-        if (!(faissIndex instanceof FaissIdMapIndex idMapIndex)) return;
-        final FaissIndex nested = idMapIndex.getNestedIndex();
-        if (!(nested instanceof AbstractFaissHNSWIndex hnswIndex) || !(hnswIndex.getFlatVectors() instanceof FaissEmptyIndex)) return;
-
-        final FaissIndex flatIndex = FaissFlatIndexFactory.create(fieldInfo, flatVectorsReader);
-        if (flatIndex == null) {
-            throw new IllegalStateException(
-                String.format(
-                    "%s found for field [%s] but %s returned null — cannot wire flat storage.",
-                    FaissEmptyIndex.class.getName(),
-                    fieldInfo.getName(),
-                    FaissFlatIndexFactory.class.getName()
-                )
-            );
-        }
-        hnswIndex.flatVectors = flatIndex;
     }
 
     protected abstract void doLoad(IndexInput input) throws IOException;
--- a/src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissMemoryOptimizedSearcherFactory.java
+++ b/src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissMemoryOptimizedSearcherFactory.java
@@ -38,7 +38,8 @@ public class FaissMemoryOptimizedSearcherFactory implements VectorSearcherFactor
         final IndexInput indexInput = directory.openInput(fileName, ioContext);
         try {
             // Try load it. Not all FAISS index types are currently supported at the moment.
-            final FaissIndex faissIndex = FaissIndex.load(indexInput, fieldInfo, flatVectorsReader);
+            final FaissIndex faissIndex = FaissIndex.load(indexInput);
+            FaissFlatIndexFactory.maybeSetFlatIndex(faissIndex, fieldInfo, flatVectorsReader);
             final FlatVectorsScorer vectorScorer = FlatVectorsScorerProvider.getFlatVectorsScorer(
                 fieldInfo,
                 faissIndex.getVectorSimilarityFunction(),
--- a/src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FlatVectorsScorerProvider.java
+++ b/src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FlatVectorsScorerProvider.java
@@ -16,6 +16,7 @@ import org.apache.lucene.util.hnsw.RandomVectorScorerSupplier;
 import org.opensearch.knn.common.FieldInfoExtractor;
 import org.opensearch.knn.index.KNNVectorSimilarityFunction;
 import org.opensearch.knn.index.SpaceType;
+import org.opensearch.knn.index.engine.faiss.FaissSQEncoder;
 import org.opensearch.knn.plugin.script.KNNScoringUtil;
 
 import java.io.IOException;
@@ -51,16 +52,20 @@ public class FlatVectorsScorerProvider {
         final KNNVectorSimilarityFunction similarityFunction,
         final FlatVectorsScorer delegateScorer
     ) {
+        // TODO: Refactor with a Resolver
         // Handle Special case of ADC first.
         if (FieldInfoExtractor.isAdc(fieldInfo)) {
             return ADC_FLAT_SCORERS.get(FieldInfoExtractor.getSpaceType(null, fieldInfo));
-        } else if (KNNVectorSimilarityFunction.HAMMING == similarityFunction) {
-            // Since Lucene doesn't provide hamming distance scorer, we return our own hamming distance scorer
-            return HAMMING_VECTOR_SCORER;
-        } else {
-            // For all other cases just return the delegate scorer.
-            return delegateScorer;
-        }
+        } else if (FieldInfoExtractor.isSQField(fieldInfo)
+            && FieldInfoExtractor.extractSQConfig(fieldInfo).getBits() == FaissSQEncoder.Bits.ONE.getValue()) {
+                return new Faiss104ScalarQuantizedVectorScorer(delegateScorer);
+            } else if (KNNVectorSimilarityFunction.HAMMING == similarityFunction) {
+                // Since Lucene doesn't provide hamming distance scorer, we return our own hamming distance scorer
+                return HAMMING_VECTOR_SCORER;
+            } else {
+                // For all other cases just return the delegate scorer.
+                return delegateScorer;
+            }
     }
 
     private static class ADCFlatVectorsScorer implements FlatVectorsScorer {
--- a/src/main/java/org/opensearch/knn/index/codec/nativeindex/AbstractNativeEnginesKnnVectorsReader.java
+++ b/src/main/java/org/opensearch/knn/index/codec/nativeindex/AbstractNativeEnginesKnnVectorsReader.java
@@ -11,6 +11,7 @@ import lombok.extern.log4j.Log4j2;
 import org.apache.lucene.codecs.KnnVectorsReader;
 import org.apache.lucene.codecs.hnsw.FlatVectorsReader;
 import org.apache.lucene.index.FieldInfo;
+import org.apache.lucene.index.FieldInfos;
 import org.apache.lucene.index.FloatVectorValues;
 import org.apache.lucene.index.SegmentReadState;
 import org.apache.lucene.store.DataAccessHint;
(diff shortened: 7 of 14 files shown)


src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissBBQFlatIndex.java
```diff
@@ -0,0 +1,57 @@
...
+import java.util.function.Supplier;
+
+/**
+ * A virtual FaissIndex that serves as a proxy for Lucene's BinaryQuantized vectors reader.
+ * When Faiss BBQ is used, the HNSW graph is stored in the .faiss file without storage (IO_FLAG_SKIP_STORAGE),
+ * and quantized vectors are stored separately via Lucene's format. This class bridges the two by installing
+ * itself as the flat storage under FaissHNSWIndex, providing access to the quantized vectors reader for scoring.
+ */
+public class FaissBBQFlatIndex extends FaissIndex {
+    public static final String FAISS_BBQ_FLAT_INDEX = "FaissBBQFlatIndex";
```
navneet1v: any reason for making this public?

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/binary/FaissBinaryHnswIndex.java
```diff
@@ -43,15 +44,15 @@ public FaissBinaryHnswIndex(final String indexType, final FaissHNSW faissHnsw) {
...
+    protected void doLoad(IndexInput input, Supplier<FaissIndex> nullFlatIndexSupplier) throws IOException {
         // Read common binary index header
         readBinaryCommonHeader(input);
 
         // Partial load HNSW graph
         faissHnsw.load(input, getTotalNumberOfVectors());
 
         // Partial load storage
-        storage = FaissIndexLoadUtils.toBinaryIndex(FaissIndex.load(input));
+        storage = FaissIndexLoadUtils.toBinaryIndex(FaissIndex.load(input, null));
```
0ctopus13prime: We should pass `nullFlatIndexSupplier` instead of null here.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissBBQFlatIndex.java
```diff
@@ -0,0 +1,57 @@
...
+
+    @Override
+    protected void doLoad(IndexInput input, Supplier<FaissIndex> nullFlatIndexSupplier) throws IOException {
+        // No-op: quantized vectors are managed by bbqFlatReader, not loaded from the faiss file.
+    }
+
+    @Override
+    public VectorEncoding getVectorEncoding() {
+        return VectorEncoding.FLOAT32;
+    }
```
navneet1v, line 45: is this encoding be Float32?

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissBBQFlatIndex.java
```diff
@@ -0,0 +1,57 @@
...
+
+    @Override
+    public FloatVectorValues getFloatValues(IndexInput indexInput) throws IOException {
+        return bbqFlatReader.getFloatVectorValues(fieldName);
+    }
+
+    @Override
+    public ByteVectorValues getByteValues(IndexInput indexInput) throws IOException {
+        throw new UnsupportedOperationException("FaissBBQFlatIndex does not support byte vector values.");
+    }
```
navneet1v, line 62: we don't want to return quantized vectors from here?

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
```diff
@@ -354,7 +354,7 @@ private IOSupplier<VectorSearcher> getVectorSearcherSupplier(final FieldInfo fie
         // Start creating searcher
         final String fileName = KNNCodecUtil.getNativeEngineFileFromFieldInfo(fieldInfo, segmentReadState.segmentInfo);
         if (fileName != null) {
-            return () -> searcherFactory.createVectorSearcher(segmentReadState.directory, fileName, fieldInfo, ioContext);
+            return () -> searcherFactory.createVectorSearcher(segmentReadState.directory, fileName, fieldInfo, ioContext, null);
```
navneet1v: instead of null can we use identity functions?

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissBBQFlatIndex.java
```diff
@@ -0,0 +1,57 @@
...
+    public FaissBBQFlatIndex(final FlatVectorsReader bbqFlatReader, final String fieldName) {
+        super(FAISS_BBQ_FLAT_INDEX);
+        this.bbqFlatReader = bbqFlatReader;
+        this.fieldName = fieldName;
+    }
+
+    @Override
+    protected void doLoad(IndexInput input, Supplier<FaissIndex> nullFlatIndexSupplier) throws IOException {
+        // No-op: quantized vectors are managed by bbqFlatReader, not loaded from the faiss file.
+    }
```
navneet1v: any reason of not passing the quantized vectors here which are stored in Lucene?

src/main/java/org/opensearch/knn/memoryoptsearch/VectorSearcherFactory.java
```diff
@@ -8,23 +8,39 @@
...
      * @return Null instance if it is not supported, otherwise return {@link VectorSearcher}
      * @throws IOException
      */
-    VectorSearcher createVectorSearcher(Directory directory, String fileName, FieldInfo fieldInfo, IOContext ioContext) throws IOException;
+    VectorSearcher createVectorSearcher(
+        Directory directory,
+        String fileName,
+        FieldInfo fieldInfo,
+        IOContext ioContext,
+        Supplier<FaissIndex> nullFlatIndexSupplier
```
navneet1v: is there a way we can avoid this parameter in the interface?

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/cagra/FaissHNSWCagraBinaryIndex.java
```diff
@@ -39,6 +40,6 @@ protected void doLoad(final IndexInput input) throws IOException {
         ((FaissCagraHNSW) faissHnsw).setNumBaseLevelSearchEntryPoints(numBaseLevelSearchEntryPoint);
 
         faissHnsw.load(input, getTotalNumberOfVectors());
-        storage = FaissIndexLoadUtils.toBinaryIndex(FaissIndex.load(input));
+        storage = FaissIndexLoadUtils.toBinaryIndex(FaissIndex.load(input, null));
```
0ctopus13prime: I think we should pass nullFlatIndexSupplier instead of null

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsReader.java
0ctopus13prime: The supplier will only be used when "null" is found with IO_SKIP flag in Faiss.
I don't think we can use identify function in here.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/cagra/FaissHNSWCagraIndex.java
```diff
@@ -76,7 +77,7 @@ protected void doLoad(final IndexInput input) throws IOException {
         ((FaissCagraHNSW) faissHnsw).setNumBaseLevelSearchEntryPoints(numBaseLevelSearchEntryPoint);
 
         // Partial load flat vector storage
-        flatVectors = FaissIndex.load(input);
+        flatVectors = FaissIndex.load(input, null);
```
0ctopus13prime: Pass `nullFlatIndexSupplier`

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissBBQFlatIndex.java
0ctopus13prime: +1, I think we can make it package-private

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissBBQFlatIndex.java
0ctopus13prime: hmm... maybe I'm missing something but we have `bbqFlatReader` already, what do you mean by passing 'quantized vectors` in here?
`bbqFlatReader` will be used for scorer creation in FaissMemoryOptimizedSearcher

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissBBQFlatIndex.java
0ctopus13prime: Yes, this is more about the type of query provided. 
If this returns FLOAT32, then it will use `search(float[])`

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissBBQFlatIndex.java
0ctopus13prime: No, I think it's slightly meaningless just returning quantized vectors without correction factors.

src/main/java/org/opensearch/knn/memoryoptsearch/VectorSearcherFactory.java
0ctopus13prime: Would be appreciated if you have better ideas, but I could not find a way to inject pluggable FaissIndex into FaissMemoryOptimizedSearcher.
We need to provide FaissIndex for "null" section to hook into IndexHNSW.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/binary/FaissBinaryHnswIndex.java
```diff
@@ -43,15 +44,15 @@ public FaissBinaryHnswIndex(final String indexType, final FaissHNSW faissHnsw) {
      * @throws IOException
      */
     @Override
-    protected void doLoad(IndexInput input) throws IOException {
+    protected void doLoad(IndexInput input, Supplier<FaissIndex> nullFlatIndexSupplier) throws IOException {
```
kotwanikunal: Can we overload the method instead?

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/binary/FaissIndexBinaryFlat.java
```diff
@@ -37,10 +39,11 @@ public FaissIndexBinaryFlat() {
      * FYI FAISS - <a href="https://github.com/facebookresearch/faiss/blob/main/faiss/impl/index_read.cpp#L1363">...</a>
      *
      * @param input Index input reading bytes from Faiss index.
+     * @param nullFlatIndexSupplier Supplier for null flat index handling.
      * @throws IOException
      */
     @Override
-    protected void doLoad(IndexInput input) throws IOException {
+    protected void doLoad(IndexInput input, Supplier<FaissIndex> nullFlatIndexSupplier) throws IOException {
```
kotwanikunal: Same here - I think overloading makes more sense

naveentatikonda:
> @naveentatikonda can you explain bit in more details what are doing in this PR and how the interfaces are getting changed?

Updated PR description with all the details

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissBBQFlatIndex.java
naveentatikonda: Ack

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/AbstractFaissHNSWIndex.java
```diff
@@ -6,10 +6,13 @@
...
 
 import lombok.Getter;
+import lombok.Setter;
 
 public abstract class AbstractFaissHNSWIndex extends FaissIndex implements FaissHNSWProvider {
     @Getter
     protected FaissHNSW faissHnsw;
+    @Getter
+    @Setter
     protected FaissIndex flatVectors;
```
navneet1v, line 14: can we avoid this and set the flatVectors once?

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissMemoryOptimizedSearcherFactory.java
```diff
@@ -49,4 +51,20 @@ public VectorSearcher createVectorSearcher(
...
+    /**
+     * When the FAISS HNSW index has no flat vector storage (e.g., BBQ), wire in the provided
+     * {@link FlatVectorsReader} as the flat vector source.
+     */
+    private void maybeSetFlatVectorsFromReader(
+        final FaissIndex faissIndex,
+        final FlatVectorsReader flatVectorsReader,
+        final FieldInfo fieldInfo
+    ) {
+        if (faissIndex instanceof FaissIdMapIndex idMapIndex) {
```
0ctopus13prime: I think calling `FiedInfoExtractor.isBBQ()` is missing in if statement.
We cannot blindly set FaissBBQFlatIndex for null flat vectors as this is being used for FP32 index as well.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissBBQFlatIndex.java
```diff
@@ -0,0 +1,56 @@
...
+
+    @Override
+    public VectorEncoding getVectorEncoding() {
+        return VectorEncoding.FLOAT32;
+    }
+
+    @Override
+    public FloatVectorValues getFloatValues(IndexInput indexInput) throws IOException {
+        return bbqFlatReader.getFloatVectorValues(fieldName);
+    }
```
navneet1v, line 55: will this give full precision vectors? if Yes please update the java doc.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissIndex.java
```diff
@@ -44,13 +47,19 @@ public FaissIndex(final String indexType) {
...
      *
      * @param input Input stream to a FAISS index
-     * @return Top level {@link FaissIndex}.
+     * @return Top level {@link FaissIndex}, or {@code null} if a "null" section is encountered.
      * @throws IOException
      */
     public static FaissIndex load(IndexInput input) throws IOException {
         final String indexType = FaissIndexLoadUtils.readIndexType(input);
+        if (NULL_INDEX_TYPE.equals(indexType)) {
+            return null;
```
navneet1v: rather than returning null can we create something like an EmptyIndex?

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss104ScalarQuantizedKnnVectorsReader.java
```diff
@@ -59,14 +59,14 @@ public ByteVectorValues getByteVectorValues(String field) throws IOException {
      */
     @Override
     public void search(String field, float[] target, KnnCollector knnCollector, AcceptDocs acceptDocs) throws IOException {
+        final FieldInfo fieldInfo = segmentReadState.fieldInfos.fieldInfo(field);
```
0ctopus13prime: Since this reader will only handle one field, can we cache FieldInfo and `VectorSearcher`?

```
private FieldInfo fieldInfo;
private VectorSearcher memoryOptimizedSearcher;

if (this.fieldInfo == null) {
    // First search
    synchronized(vectorSearcherHolderLockObject) {
        if (this.fieldInfo == null) {
            // initialize fieldInfo + memoryOptimizedSearcher
        }
    }
}

// search with memoryOptimizedSearcher
```

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissMemoryOptimizedSearcherFactory.java
```diff
@@ -32,13 +32,15 @@ public VectorSearcher createVectorSearcher(
...
-        final FlatVectorsScorer vectorScorer
+        final FlatVectorsReader flatVectorsReader
     ) throws IOException {
         final IndexInput indexInput = directory.openInput(fileName, ioContext);
 
         try {
             // Try load it. Not all FAISS index types are currently supported at the moment.
-            return new FaissMemoryOptimizedSearcher(indexInput, fieldInfo, vectorScorer);
+            FaissIndex faissIndex = FaissIndex.load(indexInput);
+            maybeSetFlatVectorsFromReader(faissIndex, flatVectorsReader, fieldInfo);
```
navneet1v: can we combine this and create a function on FaissIndex with name `loadWithFlatIndex()` or just `load(indexInput, FlatBBQIndex)`.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissMemoryOptimizedSearcherFactory.java
```diff
@@ -49,4 +51,20 @@ public VectorSearcher createVectorSearcher(
...
+     */
+    private void maybeSetFlatVectorsFromReader(
+        final FaissIndex faissIndex,
+        final FlatVectorsReader flatVectorsReader,
+        final FieldInfo fieldInfo
+    ) {
+        if (faissIndex instanceof FaissIdMapIndex idMapIndex) {
+            final FaissIndex nested = idMapIndex.getNestedIndex();
+            if (nested instanceof AbstractFaissHNSWIndex hnswIndex && hnswIndex.getFlatVectors() == null) {
+                hnswIndex.setFlatVectors(new FaissBBQFlatIndex(flatVectorsReader, fieldInfo.getName()));
```
navneet1v: this logic is very very specific to BBQ. I wonder if you can read the fieldInfo or any other parameter that can help you deicide what type of `FaissFlatIndex` you need to use. One way I can think to do this would be to create a factory that takes fieldInfo, flatVectorsReader and gives you a FaissFlatIndex.

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss104ScalarQuantizedKnnVectorsReader.java
```diff
@@ -59,14 +59,14 @@ public ByteVectorValues getByteVectorValues(String field) throws IOException {
...
+        final VectorSearcher memoryOptimizedSearcher = loadMemoryOptimizedSearcherIfRequired(fieldInfo);
+
         // Null target is the warmup signal — handle before attempting searcher load
```
0ctopus13prime: NIT : I think this is stale comment

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissMemoryOptimizedSearcherFactory.java
naveentatikonda: I moved it to FaissIndex

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissMemoryOptimizedSearcherFactory.java
naveentatikonda: Ack, created a new FaissFlatIndexFactory

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss1040ScalarQuantizedKnnVectorsReader.java
```diff
@@ -59,14 +61,19 @@ public ByteVectorValues getByteVectorValues(String field) throws IOException {
...
     public void search(String field, float[] target, KnnCollector knnCollector, AcceptDocs acceptDocs) throws IOException {
-        // Null target is the warmup signal — handle before attempting searcher load
+        if (this.fieldInfo == null) {
+            synchronized (vectorSearcherHolderLockObject) {
+                if (this.fieldInfo == null) {
+                    this.fieldInfo = segmentReadState.fieldInfos.fieldInfo(field);
+                }
+            }
+        }
+        final VectorSearcher memoryOptimizedSearcher = loadMemoryOptimizedSearcherIfRequired(this.fieldInfo);
```
0ctopus13prime: You can have it in next PR, 
but this should be put under line 67.

And have `VectorSearcher memoryOptimizedSearcher` as a member variable.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissIndex.java
```diff
@@ -44,18 +49,61 @@ public FaissIndex(final String indexType) {
...
+     * @throws IOException
+     */
+    public static FaissIndex load(IndexInput input, FieldInfo fieldInfo, FlatVectorsReader flatVectorsReader) throws IOException {
+        final FaissIndex faissIndex = load(input);
+        maybeSetFlatIndex(faissIndex, fieldInfo, flatVectorsReader);
+        return faissIndex;
+    }
+
+    // If the HNSW index has no flat storage (e.g. BBQ skips it via IO_FLAG_SKIP_STORAGE), wire in the appropriate flat index.
+    static void maybeSetFlatIndex(final FaissIndex faissIndex, final FieldInfo fieldInfo, final FlatVectorsReader flatVectorsReader) {
```
0ctopus13prime: Alright, so we are not moving this logic outside of FaissIndex.

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss1040ScalarQuantizedKnnVectorsReader.java
naveentatikonda: it is already cached in NativeEngines990KnnVectorsReader. Then why do we need to move it inside ?

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissIndex.java
naveentatikonda: Yes

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss1040ScalarQuantizedKnnVectorsReader.java
```diff
@@ -59,14 +60,19 @@ public ByteVectorValues getByteVectorValues(String field) throws IOException {
...
+
         if (target == null) {
             throw new FaissMemoryOptimizedSearcher.WarmupInitializationException("Null vector supplied for warmup");
         }
 
-        final FieldInfo fieldInfo = segmentReadState.fieldInfos.fieldInfo(field);
-        final VectorSearcher memoryOptimizedSearcher = loadMemoryOptimizedSearcherIfRequired(fieldInfo);
```
navneet1v, line 68: why we are removing this?

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss1040ScalarQuantizedKnnVectorsReader.java
```diff
@@ -59,14 +60,19 @@ public ByteVectorValues getByteVectorValues(String field) throws IOException {
...
     public void search(String field, float[] target, KnnCollector knnCollector, AcceptDocs acceptDocs) throws IOException {
-        // Null target is the warmup signal — handle before attempting searcher load
+        if (this.fieldInfo == null) {
+            synchronized (vectorSearcherHolderLockObject) {
+                if (this.fieldInfo == null) {
+                    this.fieldInfo = segmentReadState.fieldInfos.fieldInfo(field);
+                }
+            }
+        }
+        final VectorSearcher memoryOptimizedSearcher = loadMemoryOptimizedSearcherIfRequired(this.fieldInfo);
```
navneet1v: why we want to initialize this early?

naveentatikonda: When target is null, that means it is warmup request and we want to load the searcher to do partial loading before returning it. Added a comment below

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss1040ScalarQuantizedKnnVectorsReader.java
naveentatikonda: moved it to the top

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissFlatIndexFactory.java
```diff
@@ -0,0 +1,53 @@
...
+ */
+@UtilityClass
+public class FaissFlatIndexFactory {
+
+    /**
+     * Returns a {@link FaissIndex} to use as flat storage for the given field, or {@code null}
+     * if the FAISS file's own flat storage should be used.
+     */
+    static FaissIndex create(final FieldInfo fieldInfo, final FlatVectorsReader flatVectorsReader) {
+        if (FieldInfoExtractor.isFaissBBQ(fieldInfo)) {
```
kotwanikunal: Please replace this check with 
```
if (FieldInfoExtractor.isSQField(fieldInfo)
            && FieldInfoExtractor.extractSQConfig(fieldInfo).getBits() == FaissSQEncoder.Bits.ONE.getValue())
{
   ...
}
```

src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexBuildStrategyFactory.java#63 for reference

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissFlatIndexFactory.java
```diff
@@ -0,0 +1,55 @@
...
+            throw new IllegalStateException(
+                String.format(
+                    "%s found for field [%s] but %s returned null — cannot wire flat storage.",
+                    FaissEmptyIndex.class.getName(),
+                    fieldInfo.getName(),
+                    FaissFlatIndexFactory.class.getName()
+                )
+            );
+        }
+        hnswIndex.flatVectors = flatIndex;
```
navneet1v, line 53: [not a blocker]
This is okay for now. But need to improve here. I feel seeing of the flat index is something we should avoid.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissScalarQuantizedFlatIndex.java
```diff
@@ -0,0 +1,63 @@
...
+    public FaissScalarQuantizedFlatIndex(final FlatVectorsReader flatVectorsReader, final String fieldName) {
+        super(FAISS_SCALAR_QUANTIZED_FLAT_INDEX);
+        this.flatVectorsReader = flatVectorsReader;
+        this.fieldName = fieldName;
+    }
+
+    @Override
+    protected void doLoad(IndexInput input) throws IOException {
+        // No-op: quantized vectors are managed by LuceneSQFlatReader, not loaded from the faiss file.
+    }
```
navneet1v, line 40: can we throw exception from here?

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissScalarQuantizedFlatIndex.java
```diff
@@ -0,0 +1,63 @@
...
+
+    @Override
+    public VectorEncoding getVectorEncoding() {
+        return VectorEncoding.FLOAT32;
+    }
```
navneet1v, line 45: why this is FP32? it should not be. We should pick it from the `FlatVectorsReader`. This can break assumptions we have on the VectorValues.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissScalarQuantizedFlatIndex.java
```diff
@@ -0,0 +1,63 @@
...
+    @Override
+    public FloatVectorValues getFloatValues(IndexInput indexInput) throws IOException {
+        return flatVectorsReader.getFloatVectorValues(fieldName);
+    }
+
+    @Override
+    public ByteVectorValues getByteValues(IndexInput indexInput) throws IOException {
+        throw new UnsupportedOperationException(
+            String.format("%s does not support byte vector values.", FAISS_SCALAR_QUANTIZED_FLAT_INDEX)
+        );
```
navneet1v, line 61: you should call `flatVectorsReader.getByteVectorValues(fieldName);` here

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FlatVectorsScorerProvider.java
```diff
@@ -51,16 +52,20 @@ public static FlatVectorsScorer getFlatVectorsScorer(
...
+        } else if (FieldInfoExtractor.isSQField(fieldInfo)
+            && FieldInfoExtractor.extractSQConfig(fieldInfo).getBits() == FaissSQEncoder.Bits.ONE.getValue()) {
+                return new Faiss104ScalarQuantizedVectorScorer(delegateScorer);
+            } else if (KNNVectorSimilarityFunction.HAMMING == similarityFunction) {
+                // Since Lucene doesn't provide hamming distance scorer, we return our own hamming distance scorer
+                return HAMMING_VECTOR_SCORER;
+            } else {
+                // For all other cases just return the delegate scorer.
+                return delegateScorer;
+            }
```
navneet1v, line 68: [not a blocker]
not sure if formatting is correct here.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissFlatIndexFactory.java
navneet1v: We should if we can use BuilderPatter here to build the index. Main problem is `hnswIndex.flatVectors` should not be open. We should just set the flatVectors index once and then no one should change it

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissScalarQuantizedFlatIndex.java
0ctopus13prime: I think this should return FP32 as query type will be float[] for faiss sq always.
The encoding type determines 'acceptable' query type.

But, I think your intention is to make this 'underlying format' agnostic. For now, we're using 104 format, but this can be changed in the future.

The thing is that the reader might return both FloatKnnVectorValues and ByteKnnVectorValues. 
Otherwise we can simply check vector values type to determine the encoding.

navneet1v: Atleast in Lucene I have seen it on VectorValues and not on the Index. We are adding the encoding at the FaissIndex level, what was the purpose of that? Are we using it somewhere?

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissFlatIndexFactory.java
naveentatikonda: Ack, will address these in my next PR

PR #3183, opened 2026-03-18
Introduce VectorScorers to create VectorScorer instances based on the underlying vector storage format

VijayanB:
### Description
Introduce VectorScorerFactory to create VectorScorer instances based on
the underlying vector storage format (BinaryDocValues, FloatVectorValues,
ByteVectorValues). Add ScoreMode interface to abstract scorer vs rescorer
creation.

- Add ScoreMode with SCORE and RESCORE strategies
- Add VectorScorerFactory with float and byte target overloads
- Add ADC scorer with TODO to remove once ByteVectorValues.scorer() is implemented

### Related Issues

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/query/scorers/VectorScorers.java
+++ b/src/main/java/org/opensearch/knn/index/query/scorers/VectorScorers.java
@@ -18,6 +18,7 @@ import org.apache.lucene.search.VectorScorer;
 import org.apache.lucene.util.BitSet;
 import org.apache.lucene.util.hnsw.RandomVectorScorer;
 import org.opensearch.common.Nullable;
+import org.opensearch.knn.common.FieldInfoExtractor;
 import org.opensearch.knn.index.SpaceType;
 import org.opensearch.knn.index.codec.scorer.PrefetchableFlatVectorScorer;
 import org.opensearch.knn.index.vectorvalues.KNNVectorValuesIterator;
@@ -48,7 +49,7 @@ public final class VectorScorers {
      * @param target    the float query vector
      * @param vectorScorerMode determines whether to use scoring or rescoring
      * @param spaceType the space type defining the similarity function
-     * @param fieldInfo the field info for the vector field, or null if not available
+     * @param fieldInfo the field info for the vector field
      * @return a {@link VectorScorer} appropriate for the underlying vector storage format
      * @throws IOException if an I/O error occurs
      */
@@ -57,7 +58,7 @@ public final class VectorScorers {
         final float[] target,
         final VectorScorerMode vectorScorerMode,
         final SpaceType spaceType,
-        @Nullable final FieldInfo fieldInfo
+        final FieldInfo fieldInfo
     ) throws IOException {
         return createScorer(docIdsIteratorValues, target, vectorScorerMode, spaceType, fieldInfo, null, null);
     }
@@ -71,7 +72,7 @@ public final class VectorScorers {
      * @param target    the float query vector
      * @param vectorScorerMode determines whether to use scoring or rescoring
      * @param spaceType the space type defining the similarity function
-     * @param fieldInfo the field info for the vector field, or null if not available
+     * @param fieldInfo the field info for the vector field
      * @param filteredIdsIterator iterator over accepted child documents, or null if not nested
      * @param parentBitSet bit set identifying parent documents, or null if not nested
      * @return a {@link VectorScorer} appropriate for the underlying vector storage format
@@ -82,7 +83,7 @@ public final class VectorScorers {
         final float[] target,
         final VectorScorerMode vectorScorerMode,
         final SpaceType spaceType,
-        @Nullable final FieldInfo fieldInfo,
+        final FieldInfo fieldInfo,
         @Nullable final DocIdSetIterator filteredIdsIterator,
         @Nullable final BitSet parentBitSet
     ) throws IOException {
@@ -98,7 +99,6 @@ public final class VectorScorers {
      * @param target    the byte query vector
      * @param vectorScorerMode determines whether to use scoring or rescoring
      * @param spaceType the space type defining the similarity function
-     * @param fieldInfo the field info for the vector field, or null if not available
      * @return a {@link VectorScorer} appropriate for the underlying vector storage format
      * @throws IOException if an I/O error occurs
      */
@@ -106,10 +106,9 @@ public final class VectorScorers {
         final KNNVectorValuesIterator.DocIdsIteratorValues docIdsIteratorValues,
         final byte[] target,
         final VectorScorerMode vectorScorerMode,
-        final SpaceType spaceType,
-        @Nullable final FieldInfo fieldInfo
+        final SpaceType spaceType
     ) throws IOException {
-        return createScorer(docIdsIteratorValues, target, vectorScorerMode, spaceType, fieldInfo, null, null);
+        return createScorer(docIdsIteratorValues, target, vectorScorerMode, spaceType, null, null);
     }
 
     /**
@@ -121,7 +120,6 @@ public final class VectorScorers {
      * @param target    the byte query vector
      * @param vectorScorerMode determines whether to use scoring or rescoring
      * @param spaceType the space type defining the similarity function
-     * @param fieldInfo the field info for the vector field, or null if not available
      * @param acceptedChildrenIterator iterator over accepted child documents, or null if not nested
      * @param parentBitSet bit set identifying parent documents, or null if not nested
      * @return a {@link VectorScorer} appropriate for the underlying vector storage format
@@ -132,11 +130,10 @@ public final class VectorScorers {
         final byte[] target,
         final VectorScorerMode vectorScorerMode,
         final SpaceType spaceType,
-        @Nullable final FieldInfo fieldInfo,
         @Nullable final DocIdSetIterator acceptedChildrenIterator,
         @Nullable final BitSet parentBitSet
     ) throws IOException {
-        final VectorScorer scorer = getBaseScorer(docIdsIteratorValues, target, vectorScorerMode, spaceType, fieldInfo);
+        final VectorScorer scorer = getBaseScorer(docIdsIteratorValues, target, vectorScorerMode, spaceType);
         return maybeWrapWithNestedScorer(scorer, acceptedChildrenIterator, parentBitSet);
     }
 
@@ -145,7 +142,7 @@ public final class VectorScorers {
         final float[] target,
         final VectorScorerMode vectorScorerMode,
         final SpaceType spaceType,
-        @Nullable final FieldInfo fieldInfo
+        final FieldInfo fieldInfo
     ) throws IOException {
         final DocIdSetIterator docIdSetIterator = docIdsIteratorValues.getDocIdSetIterator();
 
@@ -158,18 +155,17 @@ public final class VectorScorers {
         if (knnVectorValues instanceof FloatVectorValues floatVectorValues) {
             return vectorScorerMode.createScorer(floatVectorValues, target);
         }
-        if (knnVectorValues instanceof ByteVectorValues byteVectorValues) { // ADC case
+        if (knnVectorValues instanceof ByteVectorValues byteVectorValues && FieldInfoExtractor.isAdc(fieldInfo)) {
             return createADCScorer(fieldInfo, byteVectorValues, target, spaceType);
         }
-        throw new IllegalArgumentException("Unsupported KnnVectorValues type: " + knnVectorValues.getClass().getSimpleName());
+        throw new IllegalArgumentException("Unsupported KnnVectorValues type: " + knnVectorValues.getClass());
     }
 
     private static VectorScorer getBaseScorer(
         final KNNVectorValuesIterator.DocIdsIteratorValues docIdsIteratorValues,
         final byte[] target,
         final VectorScorerMode vectorScorerMode,
-        final SpaceType spaceType,
-        @Nullable final FieldInfo fieldInfo
+        final SpaceType spaceType
     ) throws IOException {
         final DocIdSetIterator docIdSetIterator = docIdsIteratorValues.getDocIdSetIterator();
 
--- a/src/test/java/org/opensearch/knn/index/query/scorers/VectorScorerModeTests.java
+++ b/src/test/java/org/opensearch/knn/index/query/scorers/VectorScorerModeTests.java
@@ -5,17 +5,17 @@
 
 package org.opensearch.knn.index.query.scorers;
 
-import junit.framework.TestCase;
 import lombok.SneakyThrows;
 import org.apache.lucene.index.ByteVectorValues;
 import org.apache.lucene.index.FloatVectorValues;
 import org.apache.lucene.search.VectorScorer;
+import org.opensearch.knn.KNNTestCase;
 
 import static org.mockito.Mockito.mock;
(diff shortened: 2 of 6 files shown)


src/main/java/org/opensearch/knn/index/query/scorers/VectorScorerMode.java
```diff
@@ -0,0 +1,79 @@
...
+        }
+
+        @Override
+        public VectorScorer createScorer(ByteVectorValues vectorValues, byte[] target) throws IOException {
+            return vectorValues.scorer(target);
+        }
+    };
+
+    /** Creates a scorer that recomputes a higher-fidelity score from unquantized vectors. */
+    VectorScorerMode RESCORE = new VectorScorerMode() {
```
Vikasht34, line 48: why score and rescore is differnt ...things , for quantized it had to score .

src/main/java/org/opensearch/knn/index/query/scorers/VectorScorerMode.java
```diff
@@ -0,0 +1,79 @@
...
+ *   <li>{@link #SCORE} — creates a scorer via {@code vectorValues.scorer(target)},
+ *       which computes the primary similarity score (e.g. dot product, L2).</li>
+ *   <li>{@link #RESCORE} — creates a scorer via {@code vectorValues.rescorer(target)},
+ *       which recomputes a higher-fidelity score from the original (unquantized) vectors.
+ *       This is typically used after an initial approximate search with quantized vectors.</li>
+ * </ul>
+ *
+ * @see VectorScorers
+ */
+public interface VectorScorerMode {
```
Vikasht34, line 32: who is using this Interface , it's hard to review the PR when we have just interface without uses?

VijayanB: This will be used by ExactSearcher. That will be my next PR.

src/main/java/org/opensearch/knn/index/query/scorers/VectorScorerMode.java
VijayanB: score will call scorer and rescore will call rescorer. For phase 1 we will call scorer/ and for phase 2 we call rescorer from FloatVectorValues.

Vikasht34: both does same right? What is the difference? Just for my Understanding

VijayanB: Good question! While both ultimately compute similarity scores, they serve different purposes in the two-phase search process:

**SCORE (Phase 1):**
- Calls `scorer()` method on the vector values
- Works with quantized/compressed vectors during the initial search phase
- Used to quickly identify candidate results from the full dataset
- Optimized for speed over precision

**RESCORE (Phase 2):**
- Calls `rescorer()` method on FloatVectorValues
- Recomputes scores using the original unquantized vectors
- Applied only to the top candidates from Phase 1
- Provides higher-fidelity scores for final ranking

The separation allows us to benefit from fast quantized search in Phase 1, then improve accuracy by rescoring with full-precision vectors in Phase 2. The `VectorScorerMode` interface abstracts this distinction so the `ExactSearcher` (in the next PR) can handle both phases uniformly.

Vikasht34: Thanks

src/main/java/org/opensearch/knn/index/query/scorers/VectorScorerMode.java
```diff
@@ -0,0 +1,79 @@
...
+
+    /**
+     * Creates a {@link VectorScorer} for a byte query vector against {@link ByteVectorValues}.
+     *
+     * @param vectorValues the byte vector values for the segment
+     * @param target       the byte query vector
+     * @return a scorer positioned over the given vector values
+     * @throws IOException if an I/O error occurs
+     */
+    VectorScorer createScorer(ByteVectorValues vectorValues, byte[] target) throws IOException;
```
navneet1v, line 78: Looking at the abstractions it looks fine, but one thing where we are going to get confused is when we will see something like this in the code

```
VectorScorer scorer = vectorScorerMode.createScorer();
```

The only way we will know if the `scorer` is doing rescoring or not will be via instance of `vectorScorerMode`. just like createScorer, if create two more functions like `createRescorer` and remove the `vectorScorerMode` will that be better?

src/main/java/org/opensearch/knn/index/query/scorers/VectorScorers.java
```diff
@@ -0,0 +1,244 @@
...
+     * @param fieldInfo the field info for the vector field, or null if not available
+     * @return a {@link VectorScorer} appropriate for the underlying vector storage format
+     * @throws IOException if an I/O error occurs
+     */
+    public static VectorScorer createScorer(
+        final KNNVectorValuesIterator.DocIdsIteratorValues docIdsIteratorValues,
+        final float[] target,
+        final VectorScorerMode vectorScorerMode,
+        final SpaceType spaceType,
+        @Nullable final FieldInfo fieldInfo
```
navneet1v: why we think FieldInfo should be null? I would suggest lets make it NonNull.

src/main/java/org/opensearch/knn/index/query/scorers/VectorScorers.java
```diff
@@ -0,0 +1,244 @@
...
+            return KNNBinaryDocValuesScorer.create(target, binaryDocValues, spaceType);
+        }
+
+        final KnnVectorValues knnVectorValues = docIdsIteratorValues.getKnnVectorValues();
+        if (knnVectorValues instanceof FloatVectorValues floatVectorValues) {
+            return vectorScorerMode.createScorer(floatVectorValues, target);
+        }
+        if (knnVectorValues instanceof ByteVectorValues byteVectorValues) { // ADC case
+            return createADCScorer(fieldInfo, byteVectorValues, target, spaceType);
+        }
```
navneet1v: check fieldInfo here. Without fieldInfo check that its ADC, lets not make any assumption here.

src/main/java/org/opensearch/knn/index/query/scorers/VectorScorers.java
```diff
@@ -0,0 +1,244 @@
...
+        throw new IllegalArgumentException("Unsupported KnnVectorValues type: " + knnVectorValues.getClass().getSimpleName());
```
navneet1v:
```suggestion
        throw new IllegalArgumentException("Unsupported KnnVectorValues type: " + knnVectorValues.getClass());
```

Lets use full class name. rather than simple names

src/main/java/org/opensearch/knn/index/query/scorers/VectorScorers.java
```diff
@@ -0,0 +1,244 @@
...
+        final ByteVectorValues byteVectorValues,
+        final float[] target,
+        final SpaceType spaceType
+    ) throws IOException {
+        // We don't need to delegate since we know it is already ADC.
+        // This will be removed once ADC Scorer is integrated into the reader.
+        final FlatVectorsScorer adcFlatVectorsScorer = FlatVectorsScorerProvider.getFlatVectorsScorer(
+            fieldInfo,
+            spaceType.getKnnVectorSimilarityFunction(),
+            null
```
navneet1v, line 213: Lets still pass the delegate. Lets not make any assumptions. If necessasry lets add nonNull for all parameters of `getFlatVectorsScorer` function

src/test/java/org/opensearch/knn/index/query/scorers/VectorScorersTests.java
```diff
@@ -0,0 +1,267 @@
...
+
+import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.mockStatic;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+import static org.opensearch.knn.common.KNNConstants.QFRAMEWORK_CONFIG;
+import static org.opensearch.knn.common.KNNConstants.SPACE_TYPE;
+
+public class VectorScorersTests extends TestCase {
```
navneet1v: why we not extending KNNTestCase here?

src/test/java/org/opensearch/knn/index/query/scorers/VectorScorerModeTests.java
```diff
@@ -0,0 +1,79 @@
...
+import lombok.SneakyThrows;
+import org.apache.lucene.index.ByteVectorValues;
+import org.apache.lucene.index.FloatVectorValues;
+import org.apache.lucene.search.VectorScorer;
+
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+public class VectorScorerModeTests extends TestCase {
```
navneet1v: why we are not extending KNNTestCase here?

src/test/java/org/opensearch/knn/index/query/scorers/VectorScorersTests.java
```diff
@@ -0,0 +1,267 @@
...
+        DocIdSetIterator iterator = scorer.iterator();
+        assertEquals(0, iterator.nextDoc());
+        float scoreIdentical = scorer.score();
+
+        assertEquals(1, iterator.nextDoc());
+        float scoreFar = scorer.score();
+
+        assertTrue("Identical vector should score higher", scoreIdentical > scoreFar);
+        assertEquals(DocIdSetIterator.NO_MORE_DOCS, iterator.nextDoc());
+    }
```
navneet1v, line 64: assert Score values here and for all the tests. Since these are scorers and rescores I am expecting them to work perfectly with values and ids correctly being returned

src/main/java/org/opensearch/knn/index/query/scorers/VectorScorerMode.java
VijayanB: Good point! I had thought about this as well.

Your concern about readability is completely valid — vectorScorerMode.createScorer() at the call site is genuinely ambiguous, and the only way to know what’s happening is to trace back to how the mode was constructed. 


The reason I went with ScoreMode + createScorer() rather than separate createScorer() / createRescorer() methods on the factory is mainly around flexibility — if ScoreMode needs to be passed around, injected, or varied at runtime (e.g., configured per-request or per-segment), having it as an abstraction keeps that door open. Flattening it into two explicit factory methods works well when rescoring is a pure binary choice that never changes after construction.

That being said if this is something anti pattern or concerning , I am open to replacing with explicit methods.

src/main/java/org/opensearch/knn/index/query/scorers/VectorScorers.java
VijayanB: Ack

src/test/java/org/opensearch/knn/index/query/scorers/VectorScorersTests.java
VijayanB: isn't KNNTestCase integ test base class?

src/test/java/org/opensearch/knn/index/query/scorers/VectorScorerModeTests.java
VijayanB: wanted unit test, not integ test

src/test/java/org/opensearch/knn/index/query/scorers/VectorScorersTests.java
VijayanB: Ack

src/main/java/org/opensearch/knn/index/query/scorers/VectorScorers.java
VijayanB: You are right. Good catch

src/test/java/org/opensearch/knn/index/query/scorers/VectorScorersTests.java
navneet1v: no its not.

src/main/java/org/opensearch/knn/index/query/scorers/VectorScorers.java
VijayanB: In this specific code path, we're handling the ADC (Asymmetric Distance Computation) case where `knnVectorValues` is already a `ByteVectorValues` instance. The `getFlatVectorsScorer` function expects a delegate parameter for fallback scoring when the primary scorer can't handle certain cases. However, in the ADC scenario:

1. **No delegation needed**: We already know the vector format is ADC-compatible (`ByteVectorValues`), so there's no fallback case that would require delegation
2. **Unnecessary coupling**: Passing a non-null delegate would create a dependency on another scorer implementation that will never be invoked in this code path
3. **Temporary workaround**: This entire code block exists only because `ByteVectorValues.scorer()` isn't yet implemented in Lucene. Once that's available, this workaround will be removed entirely

**Proposed alternative**: Instead of passing `null` or adding `@NonNull` constraints that would be violated here, we should expose the ADC scorer directly through a dedicated method (e.g., `createADCScorer()`) that doesn't require a delegate parameter. This makes the intent explicit and avoids the ambiguity of passing `null` to a parameter that might later be marked `@NonNull`.

What do you think?

src/test/java/org/opensearch/knn/index/query/scorers/VectorScorersTests.java
VijayanB: I see. The class java doc was misleading then.

https://github.com/opensearch-project/k-NN/blob/ae119622e682c49936ba217e57d3d8f1b300c26c/src/test/java/org/opensearch/knn/KNNTestCase.java#L50-L54

src/main/java/org/opensearch/knn/index/query/scorers/VectorScorerMode.java
navneet1v: > if ScoreMode needs to be passed around, injected, or varied at runtime (e.g., configured per-request or per-segment), having it as an abstraction keeps that door open.

is this the case? Thats what exact question here. If the case is we need to pass around the ScoreMode then make sense. So @VijayanB what I would like to know here is how we are thinking to implement the exact search here.

VijayanB: Currently, the isQuantization flag is used by the query path to signal to the exact searcher whether scoring or rescoring is intended. I want to eliminate this hint-based approach and instead allow the query and reader to directly control whether they perform scoring or rescoring.

navneet1v: > I want to eliminate this hint-based approach and instead allow the query and reader to directly control whether they perform scoring or rescoring.

How would you even do that? since a query doesn't know if it has to do rescoring or not. The flag tells what to do.

VijayanB: As discussed, this will be used in follow up PR where native reader knows how to do scoring/rescoring. We will keep this implementation as it is

src/main/java/org/opensearch/knn/index/query/scorers/VectorScorers.java
VijayanB: Implemented third approach, where we check for null only during delegation.

navneet1v (approved):

PR #3185, opened 2026-03-18
Stiching Faiss104ScalarQuantizedKnnVectorsWriter to MemOptimizedScalarQuantizedIndexBuildStrategy

0ctopus13prime:
### Description
This PR introduces the Faiss BBQ (Better Binary Quantization) indexing pipeline and refactors the native engine writer hierarchy to share common flush/merge logic.

### New BBQ codec classes:

- Faiss104ScalarQuantizedKnnVectorsFormat — dedicated format for BBQ fields, using Lucene's Lucene104ScalarQuantizedVectorsFormat with 1-bit quantization for flat vector storage and native Faiss for HNSW graph construction.
- Faiss104ScalarQuantizedKnnVectorsWriter — single-field writer that flushes flat vectors via Lucene's BBQ writer, then delegates HNSW graph building to MemOptimizedScalarQuantizedIndexBuildStrategy through the native index writer pipeline.
- Faiss104ScalarQuantizedKnnVectorsReader — reader that always forces memory-optimized search, bypassing the index-level setting gate.
- MemOptimizedScalarQuantizedIndexBuildStrategy — build strategy that reads quantized vectors and correction factors from Lucene's on-disk format, transfers them to off-heap memory via JNI in batches, builds the HNSW graph, and writes only the graph structure to disk (skipping flat storage duplication).

### Refactoring:

- Extracted AbstractNativeEnginesKnnVectorsWriter from NativeEngines990KnnVectorsWriter, pulling common doFlush/doMergeOneField logic into a shared base class. Both NativeEngines990KnnVectorsWriter and Faiss104ScalarQuantizedKnnVectorsWriter now extend this base class, with each subclass providing its own quantization strategy and build parameters.
- NativeIndexWriter.getWriter now accepts buildStrategyParams to pass format-specific parameters (e.g., the FlatVectorsReader supplier for BBQ) through to the build strategy.

### Bug fixes found during development:

- Fixed NativeIndexWriter.getWriter (5-arg) silently dropping buildStrategyParams by always passing Collections.emptyMap() to createWriter.
- Fixed Faiss104ScalarQuantizedKnnVectorsWriter.mergeOneField NPE — this.fieldInfo was null during merge because addField is only called during flush.
- Fixed MemOptimizedScalarQuantizedIndexBuildStrategy passing data_type=float to writeIndex, causing Faiss to fail with "don't know how to serialize this type of index" since the BBQ index is a binary index type.
- Fixed AbstractNativeEnginesKnnVectorsWriter.doFlush creating unnecessary KNNVectorValues suppliers for empty fields before the totalLiveDocs == 0 early return.
- Fixed Faiss104ScalarQuantizedKnnVectorsWriter not closing flatVectorsWriter before the native build, causing file handle conflicts when the build strategy tries to open a reader on the just-written flat files.

### Related Issues

### Check List


src/main/java/org/opensearch/knn/index/codec/nativeindex/model/BuildIndexParams.java
```diff
@@ -38,4 +37,5 @@ public class BuildIndexParams {
     int totalLiveDocs;
     SegmentWriteState segmentWriteState;
     boolean isFlush;
+    Map<String, Object> buildStrategyParams;
```
0ctopus13prime: This is parameters to be passed down to build strategy from VectorsWriter.

src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java
```diff
@@ -196,10 +196,7 @@ private void writeToRepository(RepositoryContext repositoryContext, BuildIndexPa
             );
```
0ctopus13prime, line 196: No major changes in this file.

src/main/java/org/opensearch/knn/index/codec/nativeindex/DefaultIndexBuildStrategy.java
```diff
@@ -73,7 +73,7 @@ public void buildAndWriteIndex(final BuildIndexParams indexInfo) throws IOExcept
             }
```
0ctopus13prime, line 73: No major changes

src/main/java/org/opensearch/knn/index/codec/nativeindex/MemOptimizedNativeIndexBuildStrategy.java
```diff
@@ -47,15 +47,15 @@ public static MemOptimizedNativeIndexBuildStrategy getInstance() {
      * enabled, the vectors are quantized before being transferred off-heap. Once all vectors are transferred, they are
```
0ctopus13prime, line 47: No major changes

src/main/java/org/opensearch/knn/index/codec/nativeindex/AbstractNativeEnginesKnnVectorsWriter.java
```diff
@@ -0,0 +1,172 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.function.Supplier;
+
+import static org.opensearch.knn.common.FieldInfoExtractor.extractVectorDataType;
+import static org.opensearch.knn.index.vectorvalues.KNNVectorValuesFactory.getKNNVectorValuesSupplierForMerge;
+import static org.opensearch.knn.index.vectorvalues.KNNVectorValuesFactory.getVectorValuesSupplier;
+
+@Log4j2
+public abstract class AbstractNativeEnginesKnnVectorsWriter extends KnnVectorsWriter {
```
0ctopus13prime, line 34: Will add detailed java doc.

Vikasht34 (approved):
LGTM ... I am not getting place to put comments ... All are perfect at first look , let me take a deeper look in night

src/main/java/org/opensearch/knn/index/codec/nativeindex/MemOptimizedScalarQuantizedIndexBuildStrategy.java
```diff
@@ -114,52 +115,29 @@ public void buildAndWriteIndex(final BuildIndexParams indexInfo) throws IOExcept
...
+        if ((flatVectorsReaderSupplierObj instanceof IOSupplier) == false) {
+            throw new IllegalArgumentException(
+                "Expected IOSupplier<FlatVectorsReader>, but got [" + flatVectorsReaderSupplierObj.getClass() + "]"
             );
-        } catch (final Exception e) {
-            fullPrecisionVectorsReader.close();
-            throw e;
         }
 
+        final Object flatVectorsReaderObj = ((IOSupplier<?>) flatVectorsReaderSupplierObj).get();
```
kotwanikunal: Any reason to not move all of this down to the format classes? We are breaking separation of concerns now with this flow. Passing down general objects and doing typecasting in a hot code path - I feel this is a concern of readers and writers.

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss104ScalarQuantizedKnnVectorsWriter.java
```diff
@@ -44,43 +43,45 @@
  * <p>No quantization training is needed — Lucene's flat format handles quantization internally.
  */
 @Log4j2
-class Faiss104ScalarQuantizedKnnVectorsWriter extends KnnVectorsWriter {
+class Faiss104ScalarQuantizedKnnVectorsWriter extends AbstractNativeEnginesKnnVectorsWriter {
```
Vikasht34: What was reason for New AbstractNativeEnginesKnnVectorsWriter class?

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriter.java
```diff
@@ -13,39 +13,31 @@
...
-import static org.opensearch.knn.common.FieldInfoExtractor.extractVectorDataType;
-import static org.opensearch.knn.index.vectorvalues.KNNVectorValuesFactory.getKNNVectorValuesSupplierForMerge;
-import static org.opensearch.knn.index.vectorvalues.KNNVectorValuesFactory.getVectorValuesSupplier;
-
 /**
  * A KNNVectorsWriter class for writing the vector data strcutures and flat vectors for Native Engines.
  */
 @Log4j2
-public class NativeEngines990KnnVectorsWriter extends KnnVectorsWriter {
+public class NativeEngines990KnnVectorsWriter extends AbstractNativeEnginesKnnVectorsWriter {
```
Vikasht34, line 39: Can we not touch this class as part of this , It will may increase blast Radius? we can have next attemept to refactor this.

src/main/java/org/opensearch/knn/index/codec/nativeindex/model/BuildIndexParams.java
Vikasht34: What Object Represent here ?

src/main/java/org/opensearch/knn/index/codec/nativeindex/MemOptimizedScalarQuantizedIndexBuildStrategy.java
0ctopus13prime: Sorry not following, I believe since this is the build strategy, so it has `faiss bbq index build` responsibility. 
In here, FlacVectorsReader (i.e. .veb reader) is required to build Faiss BBQ index.
So the logic is to load quantized vectors into C++ off-heap, so that C++ HNSW can perform distance calculation on them. 
I can't think of another way to put this in format layer or writer layer as this is a part of building process.

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss104ScalarQuantizedKnnVectorsWriter.java
0ctopus13prime: Native990Writer and Faiss104ScalarQuantizedKnnVectorsWriter shares the common indexing logic. 
So I factored them out in abstract class to avoid duplicated copy-and-paste.

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriter.java
0ctopus13prime: Will need votes for this, I got feedbacks that avoid copy-and-paste code as possible
@kotwanikunal @navneet1v 
Can you share your thoughts on this?

src/main/java/org/opensearch/knn/index/codec/nativeindex/model/BuildIndexParams.java
0ctopus13prime: It's a general reference (i.e. similar to void* in C++), it's build strategy's responsibility to cast the Object with key.

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss104ScalarQuantizedKnnVectorsWriter.java
Vikasht34: MakeSense !!

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriter.java
Vikasht34: Intent is right ... If we are okay to test all the code path we are touching as part of this , should be fine

src/main/java/org/opensearch/knn/index/codec/nativeindex/MemOptimizedScalarQuantizedIndexBuildStrategy.java
Vikasht34: @0ctopus13prime  earlier there we had SeGMENT STATE all part of This change , but now we wrapped into Supplier and passing the Supplier with All Segment State make no difference with earlier code to this.

src/main/java/org/opensearch/knn/index/codec/nativeindex/model/BuildIndexParams.java
Vikasht34: Yeah my question is What are differnt types we are expoecting here , Why it can not be concrete

src/main/java/org/opensearch/knn/index/codec/nativeindex/MemOptimizedScalarQuantizedIndexBuildStrategy.java
0ctopus13prime: But what it tries to load is for the new segment that's being baked right now. 
So SegmentWriteState does not have any references to it, because .veb is new one being built.

src/main/java/org/opensearch/knn/index/codec/nativeindex/model/BuildIndexParams.java
0ctopus13prime: so far the only type that's being used is IOSupplier<FlatVectorsReader>.

src/main/java/org/opensearch/knn/index/codec/nativeindex/MemOptimizedScalarQuantizedIndexBuildStrategy.java
kotwanikunal: I understand the strategy needs quantized vectors to build the Faiss BBQ index that's not in question. The concern is about where the Lucene reader opening and reflection logic lives, not whether it happens.

The issue is consistency with the existing strategy contract. Every other strategy in this codebase receives its input data through typed fields on BuildIndexParams:

DefaultIndexBuildStrategy → `indexInfo.getKnnVectorValuesSupplier().get()`
MemOptimizedNativeIndexBuildStrategy → `indexInfo.getKnnVectorValuesSupplier().get()`

None of them open Lucene readers, do reflection, or manage reader lifecycles. They receive prepared data and build an index.

Can we pass the result as a typed Supplier<QuantizedByteVectorValues> on BuildIndexParams — the same pattern as `knnVectorValuesSupplier`. The strategy then just calls `indexInfo.getQuantizedVectorValuesSupplier().get()` and does the native build. This keeps all Lucene format I/O in the writer layer and all native index construction in the strategy layer.

I do have a sample implementation here: https://github.com/kotwanikunal/k-NN/commit/aef426b1b263d4dcc8baa3a2cc110a5bb8987eb1#diff-f8c4eb218bf94ceee3be08f3489f3736f701652b0a4273d8b3cb103548eb2c13

0ctopus13prime: I understand your concern, but that's slow since it first copies vectors to std::vector first, then it copies what's in std::vector to Faiss index builder.
Even removed this double copy, from the benchmark this bbq indexing is showing -5% indexing throughput decrease compared to BQ. And if we move to the existing way, then the gap will be widen.
Furthermore, it's designed to pass quantized vectors, not designed for passing quantized vectors + correction factors I believe. 

I think it's worth to sacrificing the vectors loading abstraction for better performance.
Please share your thoughts on this.

kotwanikunal: The performance concern about the native transfer path is valid and completely separate from the ask is what I think. The suggestion is not that you use the existing VectorTransfer/KNNVectorValues for the data transfer — the custom `passBBQVectorsWithCorrectionFactors` JNI path with the packed layout  can be in the strategy.

The ask is smaller: move the lines that open the `FlatVectorsReader` from the strategy into the writer. The strategy still receives `QuantizedByteVectorValues` and does the exact same native transfer it does today. No change to the JNI path, no double copy, no performance impact.

Out of the strategy:
```
// These lines move to the writer
Object supplier = indexInfo.getBuildStrategyParams().get(KEY);
IOSupplier<?> ioSupplier = (IOSupplier<?>) supplier;
FlatVectorsReader reader = (FlatVectorsReader) ioSupplier.get();
FloatVectorValues fvv = reader.getFloatVectorValues(field);
Field f = fvv.getClass().getDeclaredField("quantizedVectorValues");
f.setAccessible(true);
QuantizedByteVectorValues qv = (QuantizedByteVectorValues) f.get(fvv);
```

Into the strategy (unchanged):

```
// Strategy still does all the native work exactly as before
QuantizedByteVectorValues qv = indexInfo.getQuantizedVectorValuesSupplier().get();
transferQuantizedVectors(indexMemoryAddress, qv, quantizedVecBytes, knnEngine);
buildGraph(indexMemoryAddress, knnVectorValues, knnEngine);
```

The `transferQuantizedVectors` method, the batch packing, the `passBBQVectorsWithCorrectionFactors` call, the `addDocsToBBQIndex` call — all of that stays exactly where it is.

0ctopus13prime: Ah! got it, somehow i misread your comments.
Good point, will update in the next rev.

0ctopus13prime: @kotwanikunal 
Updated

src/main/java/org/opensearch/knn/common/KNNConstants.java
```diff
@@ -76,6 +76,7 @@ public class KNNConstants {
     public static final String SEARCH_SIZE_PARAMETER = "search_size";
 
     public static final String QFRAMEWORK_CONFIG = "qframe_config";
+    public static final String FAISS_BBQ_CONFIG = "faiss_bbq_config";
```
navneet1v, line 79: Need to revisit this naming.

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss104ScalarQuantizedKnnVectorsFormat.java
```diff
@@ -29,39 +28,26 @@
...
- * <p>Uses Lucene's max dimension limit since the flat storage is Lucene-managed.
- *
  * @see Faiss104ScalarQuantizedKnnVectorsWriter
  * @see Faiss104ScalarQuantizedKnnVectorsReader
  */
 @Log4j2
 public class Faiss104ScalarQuantizedKnnVectorsFormat extends KnnVectorsFormat {
 
-    private static final String FORMAT_NAME = "Faiss104ScalarQuantizedKnnVectorsFormat";
+    private static final String FORMAT_NAME = Faiss104ScalarQuantizedKnnVectorsFormat.class.getSimpleName();
```
navneet1v: We should not do this, class names can change in future, but format name get serialized on the disk. So it should not get changed. Please keep it 

```
private static final String FORMAT_NAME = "Faiss104ScalarQuantizedKnnVectorsFormat";
```

src/main/java/org/opensearch/knn/common/KNNConstants.java
0ctopus13prime: sure, when will it be finalized? 🤔

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss104ScalarQuantizedKnnVectorsFormat.java
```diff
@@ -29,39 +28,26 @@
...
-     */
-    public Faiss104ScalarQuantizedKnnVectorsFormat(int approximateThreshold) {
         super(FORMAT_NAME);
-        this.approximateThreshold = approximateThreshold;
     }
 
     @Override
     public KnnVectorsWriter fieldsWriter(SegmentWriteState state) throws IOException {
-        return new Faiss104ScalarQuantizedKnnVectorsWriter(state, bbqFlatFormat.fieldsWriter(state), approximateThreshold);
+        return new Faiss104ScalarQuantizedKnnVectorsWriter(state, bbqFlatFormat.fieldsWriter(state), bbqFlatFormat::fieldsReader);
```
navneet1v: why is `approximateThreshold` is removed?

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss1040ScalarQuantizedKnnVectorsFormat.java
```diff
@@ -29,39 +28,26 @@
...
 @Log4j2
 public class Faiss104ScalarQuantizedKnnVectorsFormat extends KnnVectorsFormat {
 
-    private static final String FORMAT_NAME = "Faiss104ScalarQuantizedKnnVectorsFormat";
+    private static final String FORMAT_NAME = Faiss104ScalarQuantizedKnnVectorsFormat.class.getSimpleName();
 
     // Shared across all format instances; Lucene104ScalarQuantizedVectorsFormat is stateless.
     private static final Lucene104ScalarQuantizedVectorsFormat bbqFlatFormat = new Lucene104ScalarQuantizedVectorsFormat(
         ScalarEncoding.SINGLE_BIT_QUERY_NIBBLE
     );
```
navneet1v, line 43: This will work since we are only supporting 32x compression here. But this is not extensible for other compression level. If possible I would suggest to refactor this, or atleast add comments.

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss104ScalarQuantizedKnnVectorsFormat.java
```diff
@@ -29,39 +28,26 @@
  * {@code "encoder": {"name": "faiss_bbq"}}. See {@code FaissCodecFormatResolver} for the
  * routing logic in {@code BasePerFieldKnnVectorsFormat.getKnnVectorsFormatForField}.
  *
- * <p>Uses Lucene's max dimension limit since the flat storage is Lucene-managed.
- *
  * @see Faiss104ScalarQuantizedKnnVectorsWriter
  * @see Faiss104ScalarQuantizedKnnVectorsReader
  */
 @Log4j2
 public class Faiss104ScalarQuantizedKnnVectorsFormat extends KnnVectorsFormat {
```
navneet1v: this should be `Faiss1040ScalarQuantizedKnnVectorsFormat`, since 10.4 represent lucene version and 0 represent 0th version from KNN plugin. This is how we name all the classes. Ref: https://github.com/opensearch-project/k-NN/blob/main/DEVELOPER_GUIDE.md#codec-versioning

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss104ScalarQuantizedKnnVectorsFormat.java
navneet1v: if both readers and writers are needed here, why don't we just pass the `Lucene104ScalarQuantizedVectorsFormat` to the writer?

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss104ScalarQuantizedKnnVectorsWriter.java
```diff
@@ -175,17 +165,66 @@ public void finish() throws IOException {
...
+                    + "]. This may indicate an incompatible Lucene version.",
+                e
+            );
+        }
+    }
+
+    /**
+     * Builds the strategy params map containing the QuantizedByteVectorValues for the build strategy.
+     */
+    private Map<String, Object> buildStrategyParams(final QuantizedByteVectorValues quantizedValues) {
```
navneet1v: Lets avoid generic `Object` in map. Please make it a concrete. Its becomes harder to debug.

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss1040ScalarQuantizedKnnVectorsWriter.java
```diff
@@ -175,17 +165,66 @@ public void finish() throws IOException {
...
     @Override
     public long ramBytesUsed() {
-        return SHALLOW_SIZE + flatVectorsWriter.ramBytesUsed() + (field != null ? field.ramBytesUsed() : 0);
+        return SHALLOW_SIZE + flatVectorsWriter.ramBytesUsed() + (fieldWriter != null ? fieldWriter.ramBytesUsed() : 0);
+    }
+
+    /**
+     * Opens a FlatVectorsReader scoped to this single field from the already-written flat files.
+     */
+    private FlatVectorsReader openFlatVectorsReader() throws IOException {
```
navneet1v, line 176: closing the writer and then opening the reader like this, looks very fragile to me. Is there any precedence in lucene around this? also how Lucene is building the graph with quantized vectors? I feel we are missing something here.

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss1040ScalarQuantizedKnnVectorsWriter.java
```diff
@@ -44,129 +45,118 @@
...
-        );
-        int totalLiveDocs = NativeEngines990KnnVectorsWriter.getLiveDocs(knnVectorValuesSupplier.get());
-        if (totalLiveDocs == 0) {
-            log.debug("[Merge] No live docs for field {}", fieldInfo.getName());
-            return;
-        }
+        // Merge, finish, and close the flat writer so that files are readable.
+        flatVectorsWriter.mergeOneField(fieldInfo, mergeState);
+        flatVectorsWriter.finish();
+        IOUtils.close(flatVectorsWriter);
```
navneet1v, line 138: you don't need to close writer on merge. You can get access of VectorValues during merge. https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/vectorvalues/KNNVectorValuesFactory.java#L218-L238

There should be a similar way to get the QuantizedVectorValues too

src/main/java/org/opensearch/knn/index/codec/AbstractNativeEnginesKnnVectorsWriter.java
```diff
@@ -0,0 +1,172 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.knn.index.codec;
```
navneet1v: move this class under org.opensearch.knn.index.codec.native

src/main/java/org/opensearch/knn/index/codec/nativeindex/AbstractNativeEnginesKnnVectorsWriter.java
```diff
@@ -0,0 +1,172 @@
...
+        final SegmentWriteState segmentWriteState,
+        final NativeIndexBuildStrategyFactory nativeIndexBuildStrategyFactory,
+        final Map<String, Object> buildStrategyParameters
+    ) throws IOException {
+        // Check total live docs first to avoid unnecessary supplier creation for empty fields
+        final int totalLiveDocs;
+        if (vectors instanceof Map) {
+            totalLiveDocs = ((Map) vectors).size();
+        } else {
+            totalLiveDocs = ((List) vectors).size();
```
navneet1v, line 52: why there is a map or a list?

navneet1v (changes requested):
The changes are quite big and some of the patterns used in the code to close the writer during flush and merge and then open it back seems pretty counterintuitive to me. Lets see how lucene is handling  quantized vectors and hnsw index creation to see if we can take some inspiration from there.

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss104ScalarQuantizedKnnVectorsFormat.java
0ctopus13prime: will update in the next rev

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss104ScalarQuantizedKnnVectorsFormat.java
0ctopus13prime: It's not being used in quantized index

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss1040ScalarQuantizedKnnVectorsFormat.java
0ctopus13prime: Navneet, can we keep the scope smallest as possible? 
This is for 32x, and we can extend it later, it is not that hard to do it

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss104ScalarQuantizedKnnVectorsFormat.java
0ctopus13prime: will update in the next rev

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss104ScalarQuantizedKnnVectorsFormat.java
0ctopus13prime: I don’t think that’s the right approach. The writer shouldn’t be responsible for the format. Instead, the format should determine the specific reader and writer types, and then those should be passed into the writer to use. In other words, it’s better for the format to decide and provide the appropriate reader–writer pair.

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss104ScalarQuantizedKnnVectorsWriter.java
0ctopus13prime: It could be expressed with generic parameters. I know that’s not ideal, but if we fix the value type, it won’t scale well in the future. In that case, the map becomes meaningless—we might as well remove it and just use IOSupplier<QuantizedBytesVectorsValue> quantizedVectorValuesSupplier.

Do you want me to simplify it that way and remove the map? Would that be a better approach?

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss1040ScalarQuantizedKnnVectorsWriter.java
0ctopus13prime: We have to close the writer then open the reader otherwise, it will throw an exception saying the file is still being written. 
And I don't think it's fragile, why close first then open a new reader is fragile? if anything, i believe it's a robust.

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss1040ScalarQuantizedKnnVectorsWriter.java
0ctopus13prime: This is the new `.veb` file that needs to be loaded. 
We have to close it before reopening it.
MergeState has the readers that are already committed, but the new one is not yet committed, it's not included in MergeState.

src/main/java/org/opensearch/knn/index/codec/AbstractNativeEnginesKnnVectorsWriter.java
0ctopus13prime: sounds good

src/main/java/org/opensearch/knn/index/codec/nativeindex/AbstractNativeEnginesKnnVectorsWriter.java
0ctopus13prime: Will add a comment, but FlatFieldVectorsWriter returns List instead of Map.
```
public abstract List<T> getVectors();
```

Native990Writer internally maintains it as a map, but Lucene104Writer maintains it as a list.

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss104ScalarQuantizedKnnVectorsWriter.java
0ctopus13prime: I'll get rid of the map and make Faiss1040ScalarQuantizedKnnVectorsWriter pass QuantizedByteVectorValues instead.

src/main/java/org/opensearch/knn/index/codec/nativeindex/AbstractNativeEnginesKnnVectorsWriter.java
```diff
@@ -0,0 +1,172 @@
...
+        }
+
+        if (totalLiveDocs == 0) {
+            log.debug("[Flush] No live docs for field {}", fieldInfo.getName());
+            return;
+        }
```
navneet1v, line 58: [Suggestion]
this logic we can move to the proper writer classes. We don't need it here.

navneet1v (approved):

src/main/java/org/opensearch/knn/index/codec/nativeindex/AbstractNativeEnginesKnnVectorsWriter.java
0ctopus13prime: Sure, will add this in my queue.

PR #3186, opened 2026-03-19
AVX512, default Bulk SIMD for Faiss Scalar Quantized Index.

0ctopus13prime:
### Description
1. Fix: Undefined behavior in int4BitDotProduct scalar fallback
The int4BitDotProduct function in both avx512_simd_similarity_function.cpp and arm_neon_simd_similarity_function.cpp used reinterpret_cast<const uint64_t*> on pointers that may not be 8-byte aligned when binaryCodeBytes % 8 != 0. This is undefined behavior per the C++ standard.

Replaced the casts with std::memcpy into local uint64_t variables. Compilers optimize this into a single mov — zero runtime cost, no UB.

2. Feature: BBQ similarity support for the default (non-SIMD) fallback
The default similarity function (default_simd_similarity_function.cpp) only supported FP16. BBQ queries on platforms without AVX512 or NEON would fail with "Invalid native similarity function type".

Added the full BBQ scoring pipeline to the default implementation:

readDataCorrections — safe unaligned reads via std::memcpy
int4BitDotProduct — scalar weighted popcount dot product (with the alignment fix)
default4bitDotProductBatch<BATCH_SIZE> — batched version, pure C++, no SIMD intrinsics
DefaultBBQSimilarityFunction<IsMaxIP> — full scoring struct with batch-8/batch-4/scalar-tail pattern
Wired up BBQ_IP and BBQ_L2 in selectSimilarityFunction

### Related Issues

### Check List

### DIFF
--- a/jni/src/simd/similarity_function/avx512_simd_similarity_function.cpp
+++ b/jni/src/simd/similarity_function/avx512_simd_similarity_function.cpp
@@ -2,6 +2,7 @@
 #include <algorithm>
 #include <array>
 #include <cstddef>
+#include <cstring>
 #include <stdint.h>
 #include <cmath>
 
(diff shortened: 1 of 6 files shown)


src/test/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss104ScalarQuantizedKnnVectorsReaderTests.java
```diff
@@ -70,7 +70,7 @@ public void testSearchFloat_whenNullTarget_thenThrowsException() {
         KNNEngine mockFaiss = spy(KNNEngine.FAISS);
         VectorSearcherFactory mockFactory = mock(VectorSearcherFactory.class);
         when(mockFaiss.getVectorSearcherFactory()).thenReturn(mockFactory);
-        when(mockFactory.createVectorSearcher(any(), any(), any(), any())).thenReturn(mock(VectorSearcher.class));
+        when(mockFactory.createVectorSearcher(any(), any(), any(), any(), any())).thenReturn(mock(VectorSearcher.class));
```
naveentatikonda, line 73: nit: We should avoid using any() but that's fine for now, I'm addressing this in my PR as I will make changes in interface of createVectorSearcher

src/test/java/org/opensearch/knn/memoryoptsearch/faiss/FaissScalarQuantizedBulkSimdScorerTests.java
```diff
@@ -53,27 +54,25 @@
...
 
     @Test
     public void testBBQEuclideanScoring() {
-        // TODO : Remove this once generic version of bulk simd is added
-        // for (int dim : Arrays.asList(1, 7, 77, 128, 512, 777, 1024, 10240, 30000, 65535)) {
-        // System.out.println("Dimension=" + dim);
-        // doTest(VectorSimilarityFunction.EUCLIDEAN, dim);
-        // }
+        for (int dim : Arrays.asList(1, 7, 77, 128, 512, 777, 1024, 10240, 30000, 65535)) {
+            System.out.println("Dimension=" + dim);
```
naveentatikonda, line 65: nit: Do we need to log these dimensions ? Let us use logger if required

jni/src/simd/similarity_function/avx512_simd_similarity_function.cpp
```diff
@@ -262,6 +263,345 @@ struct AVX512SPRFP16L2 final : BaseSimilarityFunction<BulkScoreTransformFunc, Sc
 };
 
 
+//
+// BBQ (ADC: 4-bit query x 1-bit data) - AVX512 SIMD implementation
```
naveentatikonda, line 267: Does SPR also uses this logic or will we add separate intrinsics for AVX512_SPR ?

src/test/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss104ScalarQuantizedKnnVectorsReaderTests.java
0ctopus13prime: thank you Naveen!

jni/src/simd/similarity_function/avx512_simd_similarity_function.cpp
0ctopus13prime: Planning to add it later, I think we can improve the performance a bit wit spr instruction set, but avx512 bulk simd will give a descent performance boost

src/test/java/org/opensearch/knn/memoryoptsearch/faiss/FaissScalarQuantizedBulkSimdScorerTests.java
0ctopus13prime: yeah, this line is required for debugging
can I take it as a follow up pr to cover this?
will use `log` instead.

naveentatikonda (approved):
LGTM!

PR #3189, opened 2026-03-19
#3182 Faiss BBQ MOS changes Mac CI Crash Issue

0ctopus13prime:
### Description
#3182 keeps failing with crash in Mac.
From the seg dump, it's related to bbq indexing logic, specifically, it was related to scoring logic with unaligned pointers.

```
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGSEGV (0xb) at pc=0x0000000125762680, pid=11729, tid=45363
#
# JRE version: OpenJDK Runtime Environment Temurin-25.0.2+10 (25.0.2+10) (build 25.0.2+10-LTS)
# Java VM: OpenJDK 64-Bit Server VM Temurin-25.0.2+10 (25.0.2+10-LTS, mixed mode, sharing, tiered, compressed oops, compressed class ptrs, g1 gc, bsd-aarch64)
# Problematic frame:
# C  [libopensearchknn_faiss.jnilib+0x192680]  faiss::search_neighbors_to_add(faiss::HNSW&, faiss::DistanceComputer&, std::__1::priority_queue<faiss::HNSW::NodeDistCloser, std::__1::vector<faiss::HNSW::NodeDistCloser, std::__1::allocator<faiss::HNSW::NodeDistCloser>>, std::__1::less<faiss::HNSW::NodeDistCloser>>&, int, float, int, faiss::VisitedTable&, bool)+0x254
#
# No core dump will be written. Core dumps have been disabled. To enable core dumping, try "ulimit -c unlimited" before starting Java again
#
# An error report file with more information is saved as:
# /Users/runner/work/k-NN/k-NN/build/testrun/test/hs_err_pid11729.log
#
# If you would like to submit a bug report, please visit:
#   https://github.com/adoptium/adoptium-support/issues
```

New failure 2 - https://github.com/opensearch-project/k-NN/actions/runs/23278017625/job/67685102279?pr=3182
Failure 1 - https://github.com/opensearch-project/k-NN/actions/runs/23275754432/job/67678337407?pr=3182

Once verified that all CIs pass, will close this issue

Issue #2510, opened 2025-02-08
Add support for BFloat16 with Faiss Scalar Quantizer for extended range


naveentatikonda:
### Description
The fp16 support using Faiss scalar quantizer(SQfp16) provides 50% memory reduction besides providing at par performance and mostly similar recall compared to fp32 vectors. But, it has a range limitation where the input vectors needs to be within the range of [-65504, 65504] which is a bottleneck to use fp16 as a default datatype instead of fp32.

To overcome this problem, we can add support for BFloat16 with Faiss Scalar Quantizer(SQbf16). BFloat16 comes with an extended range(same as fp32) by trading off precision(supports upto 2 or 3 decimal values or 7 mantissa bits) and still uses 16 bits per dimension (provides 50% memory reduction).   

Intel AVX512 also has [BF16 instruction set](https://www.intel.com/content/www/us/en/docs/cpp-compiler/developer-guide-reference/2021-8/intrinsics-for-avx-512-bf16-instructions.html) which can be used to further improve the performance on newer-generation processors.

PR #3190, opened 2026-03-19
Add BF16 scalar quantization support for FAISS-backed k-NN indices.

mulugetam:
### Description
This PR adds BFloat16 (BF16) scalar quantization support for FAISS-backed k-NN indices. Key changes/additions:

  - Implement bulk BF16 vector similarity implementation for inner product and L2 distance.
  - Add AVX512-BF16 SIMD kernels for BF16 vector similarity (for `avx512_spr` with up to **45%** speed up on IP).
  - Register "bf16" as a FAISS SQ encoder type in KNN constants.
  - Add FaissBF16Util with validation and clipping logic for BF16 vectors.
  - Update FAISS index builders, memory-optimized searchers, and reconstructors for BF16.
  - Add FaissBF16Reconstructor for decoding BF16 quantized vectors.
  - Add integration tests for carga-to-hnsw BF16 index creation and search.
  - Add JNI-level unit tests for BF16 similarity functions.
  - Include a binary test fixture with 300 BF16 vectors of 768 dimensions.

The BF16 bulk distance implementations live in `avx512_simd_similarity_function.cpp` and `avx512_spr_simd_similarity_function.cpp`.

* In the first one, we convert BF16 to FP32 and just do everything in FP32 for both inner product (IP) and L2.
* In the second one, we use AVX512-BF16 instructions for IP. L2 falls back to BF16 → FP32 conversion because (1) BF16 doesn’t have a *sub* instruction yet, and (2) using `||a - b||^2 = ||a||^2 - 2·a·b + ||b||^2` ends up being slower.

Below are the results comparing the BF16 IP implementation in `avx512` vs `avx512_spr`.

Source: https://gist.github.com/mulugetam/f23317bbb9057e9798b86f4d02713fd7

```text
--------------------------------------------------------------------------------------------------------------------
Benchmark                   Time             CPU   Iterations    bytes/s        dim     elem/s   num_vecs      vec/s
--------------------------------------------------------------------------------------------------------------------
BM_Fma/384/64           0.854 us        0.852 us       825096 59.4851G/s        384 28.8413G/s         64 75.1074M/s
BM_Fma/384/256           3.37 us         3.36 us       208303 58.9559G/s        384 29.2494G/s        256 76.1704M/s
BM_Fma/768/64            1.65 us         1.64 us       426300 61.6926G/s        768 29.9115G/s         64 38.9473M/s
BM_Fma/768/256           6.58 us         6.57 us       106817  60.279G/s        768 29.9059G/s        256 38.9399M/s
BM_Fma/1536/64           3.23 us         3.23 us       218414 62.7522G/s     1.536k 30.4253G/s         64 19.8082M/s
BM_Fma/1536/256          12.8 us         12.8 us        54715   62.04G/s     1.536k 30.7795G/s        256 20.0388M/s
BM_DpBf16/384/64        0.610 us        0.609 us      1147908 81.9618G/s        384 40.3504G/s         64 105.079M/s
BM_DpBf16/384/256        2.41 us         2.41 us       291026 82.0689G/s        384 40.8748G/s        256 106.445M/s
BM_DpBf16/768/64         1.17 us         1.17 us       601518 85.5797G/s        768 42.1316G/s         64 54.8588M/s
BM_DpBf16/768/256        4.54 us         4.53 us       155123 87.0562G/s        768 43.3588G/s        256 56.4567M/s
BM_DpBf16/1536/64        2.31 us         2.31 us       304559 86.6281G/s     1.536k 42.6477G/s         64 27.7654M/s
BM_DpBf16/1536/256       9.00 us         8.98 us        77854 87.8933G/s     1.536k 43.7756G/s        256 28.4998M/s
```

### Related Issues
[#3189 ](https://github.com/opensearch-project/k-NN/issues/2510)

### Check List


mulugetam:
@vamshin The scores are just the min, max, and average distances from the query vector to the database vectors in my bulk similarity benchmark. The goal is just to see the loss in precision in the bulk similarity for FP16 and BF16 against FP32, nothing more.

Yeah, I’m already working on Cohere and will share the data once it’s ready.

jni/src/simd/similarity_function/simd_similarity_function_common.cpp
```diff
@@ -231,6 +231,38 @@ SimdVectorSearchContext* SimilarityFunction::saveSearchContext(
...
+
+        // Reset Faiss function for single vector similarity calculation
+        THREAD_LOCAL_SIMD_VEC_SRCH_CTX.faissFunction.reset(
+             faiss::ScalarQuantizer {static_cast<size_t>(dimension), faiss::ScalarQuantizer::QuantizerType::QT_bf16}
+                                    .get_distance_computer(faiss::MetricType::METRIC_INNER_PRODUCT));
+
+        // Assign query to Faiss function
+        THREAD_LOCAL_SIMD_VEC_SRCH_CTX.faissFunction->set_query(
+            reinterpret_cast<float*>(THREAD_LOCAL_SIMD_VEC_SRCH_CTX.queryVectorSimdAligned));
+    } else if (nativeFunctionTypeOrd == static_cast<int32_t>(NativeSimilarityFunctionType::BF16_L2)) {
```
ManasviGoyal: Can we put the common code in a helper function since most of the code is identical except some parameters?

src/test/java/org/opensearch/knn/index/FaissIT.java
```diff
@@ -1303,6 +1304,242 @@ public void testIVFSQFP16_whenClipToFp16isTrueAndIndexedWithOutOfFP16Range_thenS
         validateGraphEviction();
     }
 
+    @SneakyThrows
+    public void testHNSWSQBF16_whenIndexedAndQueried_thenSucceed() {
+        String indexName = "test-index-hnsw-sqbf16";
+        String fieldName = "test-field-hnsw-sqbf16";
+        SpaceType[] spaceTypes = { SpaceType.L2, SpaceType.INNER_PRODUCT };
+        Random random = new Random();
```
ManasviGoyal: You can directly use unseeded random() rather that creating this.

jni/cmake/init-simd.cmake
ManasviGoyal, line 91: `-mavx512bf16` flag is missing here

jni/src/simd/similarity_function/default_simd_similarity_function.cpp
```diff
@@ -45,6 +45,13 @@ DefaultFP16SimilarityFunction<FaissScoreToLuceneScoreTransform::ipToMaxIpTransfo
 // 2. L2
 DefaultFP16SimilarityFunction<FaissScoreToLuceneScoreTransform::l2TransformBulk, FaissScoreToLuceneScoreTransform::l2Transform> DEFAULT_FP16_L2_SIMIL_FUNC;
 
+//
+// BF16 - Uses the same default (Faiss SQDistanceComputer-based) approach as FP16
+//
+// 1. Max IP
+DefaultFP16SimilarityFunction<FaissScoreToLuceneScoreTransform::ipToMaxIpTransformBulk, FaissScoreToLuceneScoreTransform::ipToMaxIpTransform> DEFAULT_BF16_MAX_INNER_PRODUCT_SIMIL_FUNC;
```
ManasviGoyal: `DefaultFP16SimilarityFunction` should be renamed to a more generic name now since it covers for FP16 and BF16

src/main/java/org/opensearch/knn/common/KNNConstants.java
```diff
@@ -160,6 +161,10 @@ public class KNNConstants {
     public static final Float FP16_MAX_VALUE = 65504.0f;
     public static final Float FP16_MIN_VALUE = -65504.0f;
 
+    // BF16 has the same exponent range as float32.
+    public static final Float BF16_MAX_VALUE = Float.MAX_VALUE;
+    public static final Float BF16_MIN_VALUE = -Float.MAX_VALUE;
```
ManasviGoyal: Are these used anywhere in the code?

jni/src/simd/similarity_function/avx512_spr_simd_similarity_function.cpp
```diff
@@ -0,0 +1,806 @@
...
+// BF16 -> FP32: zero-extend 16 -> 32 bits then shift left by 16
+static inline __m512 cvtbf16_ps(__m256i bf16x16) {
+    return _mm512_castsi512_ps(_mm512_slli_epi32(_mm512_cvtepu16_epi32(bf16x16), 16));
+}
+
+//
+// FP16
+//
+
+template <BulkScoreTransform BulkScoreTransformFunc, ScoreTransform ScoreTransformFunc>
```
ManasviGoyal: A lot of the code in this file is same as `avx512_simd_similarity_function.cpp` can we restructure the common code in a separate file to avoid duplication?

ManasviGoyal (changes requested):

mulugetam:
@ManasviGoyal I believe I have addressed all your comments. Could you please take a look?

ManasviGoyal:
> @ManasviGoyal I believe I have addressed all your comments. Could you please take a look?

LGTM!

ManasviGoyal (approved):

mulugetam:
Rebased. @ManasviGoyal could you tag someone who can review and approve? Thanks!

0ctopus13prime:
Hi @mulugetam 
Taking it a look at it

0ctopus13prime:
Hi @mulugetam, will need more times for reviewing this PR. It was bigger than I anticipated at the beginning 😅

0ctopus13prime:
One caveat is remote index building with `clip` property. (I don't think we need this property at all for bfp16)
With `clip == false`, bfp16 still be able to serve a vector whose value is out of fp16 value range. Ex: [100_000, ...]

And since we're still using FP16 (half_float) as a data type for GPU index build, therefore we might get unexpected result for such case where a vector have a bigger value that does not fit into fp16 value range. Note that remote index build for bfp16 is not yet supported.

Ideally we should first make remote indexer support bfp16, then make engine pass bfp16 as a data type to it.
Let me get back to you, will need committer's alignment in here.

In the meantime, will start review C++ code.

0ctopus13prime:
Hi @mulugetam 
Let's block remote index build for bfp16 for now. 
I know in most cases, the value range should not be a problem, but in very rare case, a model can produce a value outside of fp16 range, then it will end up building a hnsw with truncated value which is really hard for us to debug and identify the root cause.

You can add the blocking condition in RemoteIndexBuildStrategy.shouldBuildIndexRemotely ([link](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/RemoteIndexBuildStrategy.java#L101))

mulugetam:
@0ctopus13prime I've pushed the changes. I believe the CI failures are a pre-existing, repo-wide flaky/latent bug on main rather than something introduced by this PR (I can reproduce them on main with the same test seed). Could you double-check?

src/main/java/org/opensearch/knn/index/engine/faiss/AbstractFaissMethod.java
```diff
@@ -79,6 +84,9 @@ protected PerDimensionProcessor doGetPerDimensionProcessor(
             if (isFaissSQClipToFP16RangeEnabled(knnMethodContext.getMethodComponentContext())) {
                 return FaissFP16Util.CLIP_TO_FP16_PROCESSOR;
             }
+            if (isFaissSQClipToBF16RangeEnabled(knnMethodContext.getMethodComponentContext())) {
```
0ctopus13prime: For BF16, I think we don't need `clip` property. I simply does not make sense for bfp16 to have `clip` property. 
Can we make it throw an exception when `clip` was configured for bfp16?

I think `FaissSQEncoder` is the good place to add that validation. [code_link](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/engine/faiss/FaissSQEncoder.java#L81)

src/main/java/org/opensearch/knn/index/engine/faiss/FaissBF16Util.java
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    /**
+     * Verify mapping and return the value of "clip" parameter(default false) for a "faiss" Index
+     * using "sq" encoder of type "bf16".
+     *
+     * @param methodComponentContext MethodComponentContext
+     * @return boolean value of "clip" parameter
+     */
+    static boolean isFaissSQClipToBF16RangeEnabled(MethodComponentContext methodComponentContext) {
```
0ctopus13prime: With rejecting `clip` for bfp16, we can remove this along with FaissBF16Util.CLIP_TO_BF16_PROCESSOR

jni/src/simd/similarity_function/avx512_spr_simd_similarity_function.cpp
```diff
@@ -0,0 +1,135 @@
...
+        const int32_t tailDim = dim - simdDim;
+
+        // Precompute tail mask
+        const __mmask32 tailMask = tailDim > 0 ? (__mmask32)((1ULL << tailDim) - 1) : 0;
+
+        // Preconvert fp32 query to bf16 once.
+        // Heap-allocated (not a stack VLA) so large dimensions can't overflow the stack; all
+        // accesses below use unaligned loads, so no explicit over-alignment is required.
+        const int32_t queryBF16Len = ((dim + 31) / 32) * 32;
+        std::vector<uint16_t> queryBF16Storage(queryBF16Len);
```
0ctopus13prime: Can we avoid allocating vector every time?
For 10M data set, usually the number of nodes visit count is around 10K. And if assuming qps is 100/sec, then there will be 1M memory allocation for every second. 

You can use https://github.com/opensearch-project/k-NN/blob/main/jni/include/simd/similarity_function/similarity_function.h#L57. But it needs to configured 4096 aligned though - https://github.com/opensearch-project/k-NN/blob/main/jni/include/memory_util.h#L42

0ctopus13prime (commented):
Thank you for rasing a new revision.
Added few comments.

jni/src/simd/similarity_function/avx512_spr_simd_similarity_function.cpp
mulugetam: Thanks @0ctopus13prime Instead of allocating `std::vector<uint16_t> queryBF16Storage` on every `calculateSimilarityInBulk` call, I added a dedicated `queryBF16Buffer` to `SimdVectorSearchContext` and grow-never-shrink it, so the query is converted into a reused buffer with no per-call allocation:

```cpp
auto& queryBF16Buffer = srchContext->queryBF16Buffer;
if ((int32_t) queryBF16Buffer.size() < queryBF16Len) {
    queryBF16Buffer.resize(queryBF16Len);
}
uint16_t* queryBF16 = queryBF16Buffer.data();
```

I used a dedicated buffer (mirroring the existing `queryFP16Buffer`) rather than the shared `tmpBuffer` you linked, since `getVectorPointer()` can resize/reallocate `tmpBuffer` mid-scan when reassembling a vector that straddles two mmap regions, which would dangle the query pointer. On the alignment note, all accesses in this kernel use unaligned loads (`_mm512_loadu_si512` / `_mm256_storeu_si256`), so no 4096-byte alignment is required.

mulugetam:
Thanks, @0ctopus13prime. I've updated the PR based on your comments. Please let me know if you have any further feedback, since you know the codebase better than I do.

src/main/java/org/opensearch/knn/index/engine/faiss/FaissBF16Util.java
mulugetam: I added the validation to `FaissSQEncoder.validate()` as you suggested, so a bf16 encoder with `clip` enabled is now rejected up front (d7b41dff).

One deliberate choice worth calling out: I reject only an explicit `clip=true`, not the mere presence of the `clip` key. `clip=false` is the default and a genuine no-op for bf16, and it can be present in the resolved context without the user having set it, so rejecting it could break otherwise-valid configs. Rejecting `clip=true` covers the meaningful case (a user asking for clipping behavior that bf16 can't provide) while leaving `clip=false`/unset alone.

If you'd rather reject any `clip` key for bf16 regardless of value, that's a one-line change — happy to switch if you prefer the stricter behavior.

0ctopus13prime: Ideally, I think we should reject `clip` property for bfp16, like bfp16 does not even know what `clip` is.
So I think it would be better to block `clip` property entirely for bfp16, including `clip=false`

jni/src/simd/similarity_function/avx512_spr_simd_similarity_function.cpp
0ctopus13prime: This is good approach. 
But can we make search context have one dedicated 4096 aligned query buffer `std::vector<uint8_t>` instead of adding a new one?
We already have `queryFP16Buffer`, and in this PR we're trying to add one more `queryBF16Buffer`. 

You can use https://github.com/opensearch-project/k-NN/blob/main/jni/include/simd/similarity_function/similarity_function.h#L57. But it needs to configured 4096 aligned though - https://github.com/opensearch-project/k-NN/blob/main/jni/include/memory_util.h#L42.
Ex: `std::vector<uint8_t, knn_jni::NBytesAlignedAllocator<uint8_t, 4096>>`

0ctopus13prime:
Thank you @mulugetam!
I think we're getting closer to be ready.

Added few comments, and I think overall looks good to be merged.
Also could you please rebase for the next revision?

Thank you.

jni/src/simd/similarity_function/avx512_spr_simd_similarity_function.cpp
0ctopus13prime: The reason I'm suggesting 4096 alignment is, with this we don't need to care alignment constraints. 
It will be fit virtually any data types.

mulugetam: Fixed. (I was looking into a different commit)

mulugetam:
@0ctopus13prime Done. Rebased with latest changes.

0ctopus13prime:
@mulugetam 
Can you rebase it again?
People are pushing things as 3.9 door is closing. 
Once all in one place, i think we are good to go

0ctopus13prime (approved):

naveentatikonda (approved):

PR #3192, opened 2026-03-19
Add scorer-aware ByteVectorValues wrapper for FAISS Index

VijayanB:
### Description
The FAISS-backed ByteVectorValues is a random-access store that cannot produce a DocIndexIterator on its own, which is required by VectorScorer to drive scoring. Introduce FaissScorableByteVectorValues to wrap raw byte vectors with scoring capability via FlatVectorsScorer, and update VectorSearcher.getByteVectorValues to accept a DocIndexIterator that the scorer uses for document traversal.

### Related Issues

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissMemoryOptimizedSearcher.java
+++ b/src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissMemoryOptimizedSearcher.java
@@ -10,6 +10,7 @@ import org.apache.lucene.codecs.hnsw.FlatVectorsScorer;
 import org.apache.lucene.index.ByteVectorValues;
 import org.apache.lucene.index.FieldInfo;
 import org.apache.lucene.index.KnnVectorValues;
+import org.apache.lucene.index.KnnVectorValues.DocIndexIterator;
 import org.apache.lucene.index.VectorEncoding;
 import org.apache.lucene.index.VectorSimilarityFunction;
 import org.apache.lucene.search.AcceptDocs;
@@ -70,7 +71,6 @@ public class FaissMemoryOptimizedSearcher implements VectorSearcher {
 
         this.isAdc = FieldInfoExtractor.isAdc(fieldInfo);
         this.flatVectorsScorer = flatVectorsScorer;
-
         this.hnsw = extractFaissHnsw(faissIndex);
     }
 
@@ -107,14 +107,18 @@ public class FaissMemoryOptimizedSearcher implements VectorSearcher {
     }
 
     /**
-     * Returns byte vector values from the FAISS index.
-     *
-     * @return byte vector values
-     * @throws IOException if an I/O error occurs
+     * Returns a {@link FaissScorableByteVectorValues} that wraps the raw byte vectors from the
+     * FAISS index with scoring support via {@link FlatVectorsScorer}.
+     * <p>Each call creates a new instance backed by a fresh index input slice.
      */
     @Override
-    public ByteVectorValues getByteVectorValues() throws IOException {
-        return faissIndex.getByteValues(indexInput.clone());
+    public ByteVectorValues getByteVectorValues(DocIndexIterator iterator) throws IOException {
+        return new FaissScorableByteVectorValues(
+            faissIndex.getByteValues(indexInput.clone()),
+            flatVectorsScorer,
+            vectorSimilarityFunction,
+            iterator
+        );
     }
 
     @Override
--- a/src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissScorableByteVectorValues.java
+++ b/src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissScorableByteVectorValues.java
@@ -1,0 +1,126 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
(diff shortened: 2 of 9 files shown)


src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissMemoryOptimizedSearcher.java
```diff
@@ -109,14 +110,34 @@ public void search(byte[] target, KnnCollector knnCollector, AcceptDocs acceptDo
...
+     * {@inheritDoc}
+     * <p>The returned instance is a lazily initialized, shared {@link FaissMemoryOptimizedByteVectorValues}
+     * that wraps the raw byte vectors from the FAISS index and adds scorer support.
+     * Initialization is guarded by a private lock to avoid exposing synchronization to external callers.
+     * Callers that need an independent read cursor should use {@link ByteVectorValues#copy()}.
      */
     @Override
     public ByteVectorValues getByteVectorValues() throws IOException {
-        return faissIndex.getByteValues(getSlicedIndexInput());
+        synchronized (byteVectorValuesLock) {
```
0ctopus13prime: this requires locking per every request.
Can we do this vvvvv
```
if (byteVectorValues == null) {
    synchronized (byteVectorValuesLock) {
        if (byteVectorValuesLock == null) {
            byteVectorValues = createByteVectorValues();
        }
    }
}

return byteVectorValues;
```

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissMemoryOptimizedSearcher.java
```diff
@@ -109,14 +110,34 @@ public void search(byte[] target, KnnCollector knnCollector, AcceptDocs acceptDo
...
+            if (byteVectorValues == null) {
+                byteVectorValues = createByteVectorValues();
+            }
+            return byteVectorValues;
```
navneet1v: why do we need a lock here? any specific reason for that?

navneet1v: in general you should be able to create multiple instances of ByteVectorValues.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissScorableByteVectorValues.java
```diff
@@ -0,0 +1,117 @@
...
+
+    /**
+     * Returns a {@link VectorScorer} for {@code target}, or {@code null} for an empty index.
+     *
+     * <p>A fresh copy is used so the scorer's read cursor is independent from any outer iteration.
+     * The scorer delegates to {@link RandomVectorScorer#score(int)} using the iterator's current
+     * ordinal ({@code iterator.index()}).
+     */
+    @Override
+    public VectorScorer scorer(byte[] target) throws IOException {
```
navneet1v, line 102: why we cannot add this scorer implementation in the FaissByteVectorValues?

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissMemoryOptimizedSearcher.java
VijayanB: After discussing this with @navneet1v , i changed logic to always return new instance. This is aligned with how Lucene has implemented

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissScorableByteVectorValues.java
VijayanB: **@navneet1v and I collaborated to develop a strategy for implementing this feature. I've updated the Javadoc to include:**
- **Rationale** for the chosen approach
- **Alternative strategies** we considered
- **Next steps** for future
This provides a clearer explanation of our decision-making process and the path ahead.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissScorableByteVectorValues.java
```diff
@@ -0,0 +1,134 @@
...
+        final DocIndexIterator overrideIterator
+    ) {
+        if (delegate == null) {
+            throw new IllegalArgumentException("delegate must not be null");
+        }
+        this.delegate = delegate;
+        this.flatVectorsScorer = flatVectorsScorer;
+        this.similarityFunction = similarityFunction;
+        this.overrideIterator = overrideIterator;
+    }
```
navneet1v: Please keep one constructor.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissScorableByteVectorValues.java
Vikasht34: Good collaboration Guys ..Keep it Coming @VijayanB @navneet1v

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissScorableByteVectorValues.java
VijayanB: Good catch. Removed the one that was used only by test.

PR #3193, opened 2026-03-19
Add Faiss SQ encoder support for 32x compression

kotwanikunal:
### Description
Adds the 1 bit support for `sq` encoder for the Faiss HNSW method. When users explicitly request 32x compression (via compression_level: 32x or mode: on_disk), indices created on 3.6.0+ will use sq with 1 bit instead of the older binary quantizer (QFrameBitEncoder). 

#### Mapping

```
{
  "mappings": {
    "properties": {
      "my_vector": {
        "type": "knn_vector",
        "dimension": 1024,
        "method": {
          "name": "hnsw",
          "engine": "faiss",
          "parameters": {
            "encoder": {
              "name": "sq",
              "parameters": {
                "bits": 1
              }
            }
          }
        }
      }
    }
  }
}
```

### Related Issues
Related - #3169 

### Check List


src/main/java/org/opensearch/knn/index/engine/faiss/FaissBBQEncoder.java
```diff
@@ -0,0 +1,103 @@
...
+        .collect(Collectors.toUnmodifiableSet());
+
+    /**
+     * Supported bit counts for BBQ quantization. Each value maps to a specific scalar encoding
+     * in Lucene's quantized vectors format. Only 1-bit is supported for now — additional bit
+     * widths can be added here as Lucene gains support for them.
+     */
+    @Getter
+    @RequiredArgsConstructor
+    public enum Bits {
```
navneet1v: do we need this as public?

src/main/java/org/opensearch/knn/index/engine/faiss/FaissBBQEncoder.java
```diff
@@ -0,0 +1,103 @@
...
+        return METHOD_COMPONENT;
+    }
+
+    @Override
+    public CompressionLevel calculateCompressionLevel(
+        MethodComponentContext encoderContext,
+        KNNMethodConfigContext knnMethodConfigContext
+    ) {
+        return CompressionLevel.x32;
+    }
```
navneet1v: this is okay for now. But can we use Bits to find the compression level somehow.

src/main/java/org/opensearch/knn/index/engine/faiss/FaissBBQEncoder.java
kotwanikunal: This should be the reference point across the code for checking valid bits - so yes. We do have `VALID_BIT_COUNTS`. But this gives more flexibility.

src/main/java/org/opensearch/knn/index/engine/faiss/FaissBBQEncoder.java
kotwanikunal: Great point. I'll take care of it with the next PR for defaults - I added in extensibility for bits - but missed this calc part.

src/main/java/org/opensearch/knn/common/KNNConstants.java
```diff
@@ -76,6 +76,7 @@ public class KNNConstants {
     public static final String SEARCH_SIZE_PARAMETER = "search_size";
 
     public static final String QFRAMEWORK_CONFIG = "qframe_config";
+    public static final String FAISS_BBQ_CONFIG = "faiss_bbq_config";
```
kotwanikunal: Naming is still TBD here. We should treat this as WIP naming.

src/main/java/org/opensearch/knn/index/engine/faiss/FaissBBQEncoder.java
```diff
@@ -0,0 +1,103 @@
...
+            return VALID_BIT_COUNTS.contains(v);
+        }))
+        .setKnnLibraryIndexingContextGenerator(((methodComponent, methodComponentContext, knnMethodConfigContext) -> {
+            // We include the encoder name and bits in the parameters map so they get serialized
+            // into the segment's PARAMETERS field attribute. This makes segments self-describing —
+            // any code inspecting the segment can identify the encoder without needing the mapping.
+            int bits = (int) methodComponentContext.getParameters().getOrDefault(BITCOUNT_PARAM, DEFAULT_BITS);
+            return KNNLibraryIndexingContextImpl.builder().parameters(new HashMap<>() {
+                {
+                    put(INDEX_DESCRIPTION_PARAMETER, FAISS_FLAT_DESCRIPTION);
```
0ctopus13prime: Could you add TODO to revisit this?
Actually, this should not be used for bbq as we manually create the hierarchy.

src/main/java/org/opensearch/knn/index/engine/faiss/FaissBBQEncoder.java
```diff
@@ -0,0 +1,103 @@
...
+            if (context != null && context.getVersionCreated() != null && context.getVersionCreated().before(Version.V_3_6_0)) {
+                return false;
+            }
+            return VALID_BIT_COUNTS.contains(v);
+        }))
+        .setKnnLibraryIndexingContextGenerator(((methodComponent, methodComponentContext, knnMethodConfigContext) -> {
+            // We include the encoder name and bits in the parameters map so they get serialized
+            // into the segment's PARAMETERS field attribute. This makes segments self-describing —
+            // any code inspecting the segment can identify the encoder without needing the mapping.
+            int bits = (int) methodComponentContext.getParameters().getOrDefault(BITCOUNT_PARAM, DEFAULT_BITS);
```
0ctopus13prime: Curious, we don't do validation in here? I remember validation component is separated, but asking purely out of curiosity.

kotwanikunal: Correct - the validation is separated. The bits parameter is validated earlier in the flow through `MethodComponent.validate()` → `IntegerParameter.validate()`, which calls our validator lambda that checks `VALID_BIT_COUNTS.contains(v)`.

0ctopus13prime: thanks Kunal

src/main/java/org/opensearch/knn/index/engine/faiss/FaissBBQEncoder.java
kotwanikunal: Makes sense. I'll add a TODO with the default mapping changes.

src/main/java/org/opensearch/knn/index/engine/faiss/FaissBBQEncoder.java
kotwanikunal: Updated.

src/main/java/org/opensearch/knn/index/engine/faiss/FaissMethodResolver.java
```diff
@@ -124,9 +131,22 @@ private void resolveEncoder(
...
 
         if (CompressionLevel.x32 == resolvedCompressionLevel) {
-            encoderComponentContext = new MethodComponentContext(QFrameBitEncoder.NAME, new HashMap<>());
-            encoder = encoderMap.get(QFrameBitEncoder.NAME);
-            encoderComponentContext.getParameters().put(QFrameBitEncoder.BITCOUNT_PARAM, CompressionLevel.x32.numBitsForFloat32());
+            // Starting 3.6.0, x32 compression uses sq(bits=1) instead of the older QFrameBitEncoder (binary).
+            // 1-bit quantization delegates to Lucene's flat format rather than the k-NN quantization
+            // framework, which gives better recall. We also guard on encoderMap.containsKey because
+            // IVF method doesn't support 1-bit quantization — only HNSW does.
+            if (knnMethodConfigContext.getVersionCreated() != null
```
Vikasht34: what can we do to this , Can we have decider else this If condidtion is too scary

src/main/java/org/opensearch/knn/index/engine/faiss/FaissMethodResolver.java
```diff
@@ -135,6 +155,14 @@ private void resolveEncoder(
             knnMethodConfigContext
         );
         encoderComponentContext.getParameters().putAll(resolvedParams);
+
+        // When auto-resolved to bits=1, remove the type default that was injected — the
+        // 1-bit quantization path doesn't use it, and validateEncoderConfig would reject it.
+        if (encoderComponentContext.getParameters().get(FAISS_SQ_BITS) instanceof Integer bitsVal
```
Vikasht34: what is Integer bitsVal? and why we are checking instance of .. ?

kotwanikunal: it is just the variable name - getParameters() returns `Map<String, Object>` and we need to check the type to be sure. It is equivalent to  
```
Object obj = encoderComponentContext.getParameters().get(FAISS_SQ_BITS);
if (obj instanceof Integer) {
 ...
}
```

src/main/java/org/opensearch/knn/index/engine/faiss/FaissMethodResolver.java
kotwanikunal: I can pull it out in a separate method with the next set of changes - but the other solution would be a strategy pattern for all the compression levels.

Vikasht34: sure

navneet1v: +1. The same is true for other part of the code like finding rescore context.

src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexBuildStrategyFactory.java
```diff
@@ -59,7 +59,7 @@ public NativeIndexBuildStrategy getBuildStrategy(
         final KNNEngine knnEngine = extractKNNEngine(fieldInfo);
         final boolean isTemplate = fieldInfo.attributes().containsKey(MODEL_ID);
         final boolean iterative = !isTemplate && KNNEngine.FAISS == knnEngine;
-        final boolean isFaissScalarQuantizedField = FieldInfoExtractor.isFaissBBQ(fieldInfo);
+        final boolean isFaissScalarQuantizedField = FieldInfoExtractor.isSQOneBitField(fieldInfo);
```
navneet1v: This is interesting. Should we name the function as `isFaissScalarQuantizedField(fieldInfo)` and then we can see how many bits we are using. Based on that we can figure out what to do.

src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexBuildStrategyFactory.java
navneet1v: Since Fp16 and 1 bit all are SQ. We would need some changes here.

src/main/java/org/opensearch/knn/index/mapper/EngineFieldMapper.java
```diff
@@ -302,4 +317,31 @@ void updateEngineStats() {
...
+
+    /**
+     * Builds the sq onebit config attribute value as a JSON string.
+     */
+    private static String getSQOneBitConfigValue(KNNMethodContext methodContext) {
+        MethodComponentContext encoderCtx = (MethodComponentContext) methodContext.getMethodComponentContext()
+            .getParameters()
+            .get(METHOD_ENCODER_PARAMETER);
+        Object bits = encoderCtx.getParameters().getOrDefault(FAISS_SQ_BITS, FaissSQEncoder.Bits.ONE.getValue());
+        return "{\"" + FAISS_SQ_BITS + "\":" + bits + "}";
```
navneet1v: do we want add FAISS here? can we not just have `{SQ : {bits: }}`  or may be something better. But I can think of this

kotwanikunal: `FAISS_SQ_BITS` points to `bits` - but I'll rename it to `SQ_BITS` to make it a bit more generic.

src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexBuildStrategyFactory.java
kotwanikunal: Good catch `isSQOneBitField` checks for just SQ and not one bit. Let me update the check

src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexBuildStrategyFactory.java
kotwanikunal: I had that in the original flow - but I somehow mixed up the flow with the latest rebase. Good call. Updating it.

src/main/java/org/opensearch/knn/index/engine/faiss/FaissMethodResolver.java
kotwanikunal: Added a TODO for a followup.

navneet1v (approved):

PR #3196, opened 2026-03-20
[Bug] Fix FaissIdMap honor the given acceptOrds for sparse case.

0ctopus13prime:
### Bug
In FaissIdMapIndex, the getAcceptOrds method in both SparseByteVectorValuesImpl and SparseFloatVectorValuesImpl applied the ordinal→docID mapping twice.

The old code:
```
final Bits internalBits = vectorValues.getAcceptOrds(acceptDocs);
return internalBits.get((int) idMappingReader.get(internalVectorId));
```

vectorValues.getAcceptOrds(acceptDocs) already returns a Bits indexed by vector ordinal — per the Lucene KnnVectorValues.getAcceptOrds contract, it internally translates ordinal→docID via ordToDoc() before checking acceptDocs. Then the wrapper applied idMappingReader.get() (another ordinal→docID mapping) on top, passing a docID into internalBits.get() which expects an ordinal.

With the current nested HNSW flat index (where ordToDoc is identity), the double mapping was masked. But it would produce incorrect filtering results with any nested index that has a non-identity ordToDoc, such as Lucene104ScalarQuantizedVectorsReader's sparse OffHeapScalarQuantizedFloatVectorValues.

### Fix
Skip the nested index's getAcceptOrds and map directly:
```
return acceptDocs.get((int) idMappingReader.get(internalVectorId));
```

This matches Lucene's own SparseOffHeapVectorValues.getAcceptOrds pattern: take the vector ordinal, convert to docID via the mapping, check acceptDocs at that docID. One translation, no delegation.

Test changes
- Moved the ordinal→docID assertions from the shared triggerLoadAndGetIndex setup (where they were on the now-unused vectorValues.getAcceptOrds mock) into each test method, asserting directly on the acceptDocs mock.
- Added length() assertions to verify the returned Bits reports the correct size.
- Added ordToDoc() assertions to verify consistency between ordToDoc and getAcceptOrds.
- Removed dead mock setup (mockBitsFromByteVectors, mockBitsFromFloatVectors and their getAcceptOrds stubs).
- Stubbed mockByteValues.size() and mockFloatValues.size() since length() now delegates to vectorValues.size().

### Related Issues
N/A

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissIdMapIndex.java
+++ b/src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissIdMapIndex.java
@@ -138,18 +138,16 @@ public class FaissIdMapIndex extends FaissBinaryIndex implements FaissHNSWProvid
             @Override
             public Bits getAcceptOrds(final Bits acceptDocs) {
                 if (acceptDocs != null) {
-                    final Bits internalBits = vectorValues.getAcceptOrds(acceptDocs);
-
                     return new Bits() {
                         @Override
                         public boolean get(int internalVectorId) {
-                            // Apply filtering with a converted Lucene document id.
-                            return internalBits.get((int) idMappingReader.get(internalVectorId));
+                            // Convert internal vector ordinal to Lucene document id, then check acceptDocs directly.
+                            return acceptDocs.get((int) idMappingReader.get(internalVectorId));
                         }
 
                         @Override
                         public int length() {
-                            return internalBits.length();
+                            return vectorValues.size();
                         }
                     };
                 }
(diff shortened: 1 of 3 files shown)


src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissIdMapIndex.java
```diff
@@ -205,18 +203,16 @@ public int ordToDoc(int internalVectorId) {
...
-                            // Apply filtering with a converted Lucene document id.
-                            return internalBits.get((int) idMappingReader.get(internalVectorId));
+                            // Convert internal vector ordinal to Lucene document id, then check acceptDocs directly.
+                            return acceptDocs.get((int) idMappingReader.get(internalVectorId));
                         }
 
                         @Override
                         public int length() {
-                            return internalBits.length();
+                            return floatVectorValues.size();
```
navneet1v, line 215: which code path hits this?

navneet1v: what IT we can add to simulate the error.

navneet1v (approved):

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissIdMapIndex.java
0ctopus13prime: https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissMemoryOptimizedSearcher.java#L150

So any filtering cases from [[here](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/query/memoryoptsearch/MemoryOptimizedKNNWeight.java#L223)] will hit the line 😓

Vikasht34 (approved):

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissIdMapIndex.java
0ctopus13prime: This is a subtle since things mostly worked internally. The bug was using doc IDs against acceptDocs instead of ords. This mainly affects recall, but recall is typically less critical in filtered queries as we expected it to degrade with filtering.

PR #3200, opened 2026-03-22
Fix bugs in Faiss Scalar Quantization and added basic ITs.

0ctopus13prime:
### Description
Fix bugs in Faiss SQ 1-bit indexing/search and get MOSFaissSQIndexIT passing

Fixes bugs in both the indexing and search sides for Faiss SQ 1-bit that were causing low recall, and gets MOSFaissSQIndexIT (integration tests for SQ 1-bit with inner product) to pass.

What MOSFaissSQIndexIT covers
- Non-nested multi segments
  - Dense + MAX_IP
  - Dense + MAX_IP + Filters
  - Dense + L2
  - Dense + L2 + Filters
  - Spase + MAX_IP
  - Spase + MAX_IP + Filters
  - Sparse + L2
  - Sparse + L2 + Filters
- Non-nested single segments
  - Dense + MAX_IP
  - Dense + MAX_IP + Filters
  - Dense + L2
  - Dense + L2 + Filters
  - Spase + MAX_IP
  - Spase + MAX_IP + Filters
  - Sparse + L2
  - Sparse + L2 + Filters
- Nested multi segments
  - Dense + MAX_IP
  - Dense + MAX_IP + Filters
  - Dense + L2
  - Dense + L2 + Filters
  - Spase + MAX_IP
  - Spase + MAX_IP + Filters
  - Sparse + L2
  - Sparse + L2 + Filters
- Nested single segments
  - Dense + MAX_IP
  - Dense + MAX_IP + Filters
  - Dense + L2
  - Dense + L2 + Filters
  - Spase + MAX_IP
  - Spase + MAX_IP + Filters
  - Sparse + L2
  - Sparse + L2 + Filters

Key fixes:

- FaissBBQHnsw: propagate code_size from storage so hnsw_add_vertices uses the correct 112-byte stride instead of 96.
- FaissBBQDistanceComputer: negate IP scores during graph construction since Faiss HNSW minimizes distance.
- FaissFlatIndexFactory: resolve metric type from the serialized binary index header and update space type accordingly, enabling L2 and IP support instead of defaulting to Hamming.
- Various fixes across the read/write path: FaissBinaryHnswIndex handles empty storage via FaissEmptyIndex.isEmptyIndex(), Faiss104ScalarQuantizedVectorScorer moved to codec package, shared Faiss1040ScalarQuantizedUtils for quantized value extraction, and updated tests to use FaissBinaryHnswIndex.

### Related Issues
N/A

### Check List


navneet1v:
@0ctopus13prime please fix the code coverage. also attach the GH issue.

jni/include/sq/faiss_sq_flat.h
```diff
@@ -106,13 +106,15 @@ namespace knn_jni {
 
             if constexpr (IsMaxIP) {
                 score += queryAdditional + additional - centroidDp;
+                // Negate: Faiss HNSW always minimizes distance (CMax comparator).
+                // For IP, higher score = more similar, so we negate so that
+                // minimizing -score = maximizing score.
+                return -score;
```
navneet1v, line 112: will it be better to do something similar to faiss. 

```
if (is_similarity_metric(this->metric_type)) {
        // we need to revert the negated distances
        for (size_t i = 0; i < result->lims[result->nq]; i++) {
            result->distances[i] = -result->distances[i];
        }
    }
```

or we want to keep this file free from any faiss namespace?

jni/src/faiss_index_service.cpp
```diff
@@ -207,25 +209,13 @@ jlong BinaryIndexService::initIndex(
...
-            std::string actualType = "nullptr";
-            if (originalIndexBinaryHNSW) {
-                // Dereferencing a simple pointer variable is technically a "side effect"
-                // in the eyes of the warning, but this is the standard way to use it.
-                actualType = typeid(*originalIndexBinaryHNSW).name();
-            }
-
-            throw std::runtime_error("Failed to cast " + actualType + " to faiss::IndexBinaryHNSW*");
-        }
+    if (auto it = parameters.find(M); it == parameters.end()) {
```
navneet1v, line 212: any reason we don't need all this?

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss1040ScalarQuantizedUtils.java
```diff
@@ -0,0 +1,73 @@
...
+     * @param throwExceptionIfNotFound if {@code true}, throws {@link IOException} when the field
+     *                                 cannot be found (e.g., incompatible Lucene version);
+     *                                 if {@code false}, returns {@code null} silently
+     * @return the extracted {@link QuantizedByteVectorValues}, or {@code null} if not found
+     *         and {@code throwExceptionIfNotFound} is {@code false}
+     * @throws IOException if extraction fails and {@code throwExceptionIfNotFound} is {@code true}
+     */
+    public static QuantizedByteVectorValues extractQuantizedByteVectorValues(
+        final KnnVectorValues floatVectorValues,
+        final boolean throwExceptionIfNotFound
```
navneet1v: I am trying to see what places set `throwExceptionIfNotFound` to false. Atleast I don't see it src files. can remove this extra parameter?

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/binary/FaissBinaryHnswIndex.java
```diff
@@ -27,6 +29,8 @@ public class FaissBinaryHnswIndex extends FaissBinaryIndex implements FaissHNSWP
 
     @Getter
     protected FaissHNSW faissHnsw;
+    @Getter
+    @Setter
     protected FaissBinaryIndex storage;
```
navneet1v, line 35: Lets move to builder pattern here. having both setters and getters here will mean the index can change at any point of time which is quite dangerous to be honest.

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss104ScalarQuantizedVectorScorer.java
```diff
@@ -68,16 +69,29 @@ public RandomVectorScorer getRandomVectorScorer(
...
+
+        // Try bulk SIMD
+        final IndexInput indexInput = quantizedByteVectorValues.getSlice();
+        final long[] addressAndSize = MemorySegmentAddressExtractorUtil.tryExtractAddressAndSize(indexInput, 0, indexInput.length());
+        if (addressAndSize != null) {
+            return bulkSimdRandomVectorScorer(quantizedByteVectorValues, target, addressAndSize, similarityFunction);
+        }
+
+        // Fallback
+        return super.getRandomVectorScorer(similarityFunction, quantizedByteVectorValues, target);
```
navneet1v, line 96: Please add a warn log here. This we discussed in the last PR too.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/binary/FaissBinaryIndex.java
```diff
@@ -41,11 +43,10 @@ protected void readBinaryCommonHeader(final IndexInput inputStream) throws IOExc
         // Consume `is_trained`, which is always true.
         inputStream.readByte();
 
-        // Consume `metric type`. We don't rely on this metric type as internally,
-        // as all distance calculation will be done with hamming distance calculator.
-        inputStream.readInt();
+        // Consume `metric type`.
+        metricType = inputStream.readInt();
```
navneet1v, line 47: we are reading and setting this, is there a use of this metric type?

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissFlatIndexFactory.java
```diff
@@ -9,20 +9,25 @@
...
 @UtilityClass
 public class FaissFlatIndexFactory {
 
     /**
-     * Returns a {@link FaissIndex} to use as flat storage for the given field, or {@code null}
+     * Returns a {@link FaissBinaryIndex} to use as flat storage for the given field, or {@code null}
      * if the FAISS file's own flat storage should be used.
      */
-    static FaissIndex create(final FieldInfo fieldInfo, final FlatVectorsReader flatVectorsReader) {
+    static FaissBinaryIndex create(final FieldInfo fieldInfo, final FlatVectorsReader flatVectorsReader) {
```
navneet1v: the return type of function was correct which is FaissIndex, why we are changing it to FaissBinaryIndex?

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissFlatIndexFactory.java
```diff
@@ -34,22 +39,40 @@ static FaissIndex create(final FieldInfo fieldInfo, final FlatVectorsReader flat
...
+        // Update the space type accordingly.
+        final int metricType = idMapIndex.getMetricType();
+        try {
+            final SpaceType spaceType = FaissMetricType.values()[metricType].spaceType;
+            idMapIndex.setSpaceType(spaceType);
+        } catch (ArrayIndexOutOfBoundsException e) {
+            throw new IllegalArgumentException(
+                "Unsupported metric type: " + metricType + ", only support " + Arrays.asList(FaissMetricType.values())
+            );
+        }
```
navneet1v, line 76: This logic will create problems in future. As of now it make sense that the only binary index with no spaceType set will be SQ index(faiss BBQ). But that will not be true. At that time it would be difficult to make things work. Is there a reason why we want to make SQ index of Faiss as BinaryIndex? If you don't treat HNSW index has binaryHNSW then this problem will never happen.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissIndex.java
```diff
@@ -35,6 +36,7 @@ public abstract class FaissIndex {
     // Total number of vectors saved within this index.
     protected int totalNumberOfVectors;
     // Space type used to index vectors in this index.
+    @Setter
     protected SpaceType spaceType;
```
navneet1v, line 40: Please avoid updating parameters after construction of objects specially like index aka FaissIndex.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissMemoryOptimizedSearcherFactory.java
```diff
@@ -39,7 +39,7 @@ public VectorSearcher createVectorSearcher(
         try {
             // Try load it. Not all FAISS index types are currently supported at the moment.
             final FaissIndex faissIndex = FaissIndex.load(indexInput);
-            FaissFlatIndexFactory.maybeSetFlatIndex(faissIndex, fieldInfo, flatVectorsReader);
+            FaissFlatIndexFactory.maybeSetFlatBinaryIndex(faissIndex, fieldInfo, flatVectorsReader);
```
navneet1v, line 42: older name was correct. why we want to say that this is a binary index?

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissMetricType.java
```diff
@@ -0,0 +1,32 @@
...
+ * {@link SpaceType} for search-time scoring.
+ * See MetricType.h for more details.
+ *
+ * @see FaissFlatIndexFactory#maybeSetFlatBinaryIndex
+ */
+public enum FaissMetricType {
+    /** Faiss {@code METRIC_INNER_PRODUCT} (ordinal 0). */
+    INNER_PRODUCT(SpaceType.INNER_PRODUCT),
+    /** Faiss {@code METRIC_L2} (ordinal 1). */
+    L2(SpaceType.L2);
```
navneet1v, line 25: don't rely on the ordinals here, I would suggest setting the ordinals directly in the enum rather than relying on the order in which INNER_PRODUCT and L2 is defined.

jni/include/bbq/faiss_bbq_hnsw.h
```diff
@@ -14,6 +14,11 @@ namespace knn_jni {
             faiss_bbq_flat(_faiss_bbq_flat) {
             // This has the ownership of FaissBBQFlat, setting this true to make it free'd when this class is freed
             own_fields = true;
+            // Set metric type that was given, not just blindly use default space type.
+            metric_type = _faiss_bbq_flat->metric_type;
```
naveentatikonda: We want to remove bbq terminology everywhere and use SQ. Pls do it in next PR

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss1040ScalarQuantizedUtils.java
```diff
@@ -0,0 +1,73 @@
...
+    ) throws IOException {
+        try {
+            final Field f = floatVectorValues.getClass().getDeclaredField("quantizedVectorValues");
```
naveentatikonda: nit: Let's define a constant for `quantizedVectorValues`

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss1040ScalarQuantizedUtils.java
```diff
@@ -0,0 +1,73 @@
...
+            f.setAccessible(true);
+            return (QuantizedByteVectorValues) f.get(floatVectorValues);
+        } catch (NoSuchFieldException | IllegalAccessException e) {
+            if (throwExceptionIfNotFound) {
+                throw new IOException(
+                    "Failed to extract QuantizedByteVectorValues from floatVectorValues ["
```
naveentatikonda: nit: Use String.format

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss104ScalarQuantizedVectorScorer.java
```diff
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-package org.opensearch.knn.memoryoptsearch.faiss;
+package org.opensearch.knn.index.codec.KNN1040Codec;
```
naveentatikonda, line 6: nit: I thought of moving this to the codec.KNN1040Codec package initially but later realized that all the scorers are in memoryoptsearch.faiss package. Shall we leave it as is ?

src/test/java/org/opensearch/knn/memoryoptsearch/MOSFaissSQIndexIT.java
```diff
@@ -0,0 +1,45 @@
...
+
+package org.opensearch.knn.memoryoptsearch;
+
+import org.opensearch.knn.index.SpaceType;
+import org.opensearch.knn.index.VectorDataType;
+import org.opensearch.knn.index.mapper.CompressionLevel;
+import org.opensearch.knn.index.mapper.Mode;
+
+/**
+ * This is testing disk modes (32x, 16x, 8x) with LuceneOnFaiss is on.
```
naveentatikonda: Pls fix this java doc

src/test/java/org/opensearch/knn/memoryoptsearch/MOSFaissSQIndexIT.java
```diff
@@ -0,0 +1,45 @@
...
+ * Note that Faiss does not support 4x, and Faiss uses FP16 as 2x which is already covered in {@link MOSFaissFP16IndexIT}.
+ */
+public class MOSFaissSQIndexIT extends AbstractMemoryOptimizedKnnSearchIT {
```
naveentatikonda, line 16: We want to add basic ITs for encoder mapping(without compression) and with compression and mode. Also, pls add it for other space types

jni/include/sq/faiss_sq_flat.h
0ctopus13prime: This file is separated from faiss namespace, this can be treated as a plugin.
Also the negation part is the last step, in order to have it, we should add codes to wrap the max heap to min heap etc, which requires another patch. 
Please note that this is not the how Faiss works for binary index, Faiss only supports max heap for binary index.

jni/src/faiss_index_service.cpp
0ctopus13prime: I didn't know at that time `m` is passed in parameters.
The if-else was there to get `m`, but since it can be found in parameters, it's no longer needed.

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss1040ScalarQuantizedUtils.java
0ctopus13prime: ah, good point. sure will do

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/binary/FaissBinaryHnswIndex.java
0ctopus13prime: Converting this to the builder pattern is quite tricky as `void doLoad(IndexInput)` is a member method that initializes member fields. Hard to picture how we can dynamically update `storage` for the one built with SKIP_STORAGE flags

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss104ScalarQuantizedVectorScorer.java
0ctopus13prime: will do in next rev.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/binary/FaissBinaryIndex.java
0ctopus13prime: Yes, we use this info in FaissFlatIndexFactory to determine `SpaceType` of this index.
Faiss ignores this metric and always assumes hamming distance for FaissBinaryIndex, and this assumption breaks with Faiss sq from our side. 
Therefore, we should read the metric type then get the SpaceType for similarity function.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissFlatIndexFactory.java
0ctopus13prime: No, it wasn't correct. 
Faiss sq is built upon FaissBinaryIndex not FaissIndex.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissFlatIndexFactory.java
0ctopus13prime: First of all, we can't use IndexHNSW in Faiss since it does not have `codesSize` concept. 
Every offset is calculated as `x + (pt_id - n0) * d` where d represents for dimension.
Also from semantic perspective, since we are handling binary quantized vectors + correction factors, it is natural to use IndexBinaryHNSW which also has `codeSize` with `dis->set_query(
                            (float*)(x + (pt_id - n0) * index_hnsw.code_size));`

Also `metric_type` is `space_type`. Just enum classes in Java and C++ that's all.
Future binary quantized index should have space_type. I don't think it would be difficult to make things work. 
I think it will either L2 or inner product for years.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissIndex.java
0ctopus13prime: We have 2 options.
1. Either to rebuild every hierarchical structure.
2. Allow mutable state.

Which one do you suggest?

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissMemoryOptimizedSearcherFactory.java
0ctopus13prime: because old name was not correct.
Who says Faiss sq is built on FaissIndex?

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissMetricType.java
0ctopus13prime: Not sure what's the difference.
Probably keep it as it is, if you're not strongly insisting.

jni/include/bbq/faiss_bbq_hnsw.h
0ctopus13prime: ah got it. sure will update in next rev.

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss1040ScalarQuantizedUtils.java
0ctopus13prime: sounds good

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss1040ScalarQuantizedUtils.java
0ctopus13prime: sounds good

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss104ScalarQuantizedVectorScorer.java
0ctopus13prime: Hi Naveen, i had to move this under KNN1040Codec as now it uses Faiss1040ScalarQuantizedUtils which is package scope.

src/test/java/org/opensearch/knn/memoryoptsearch/MOSFaissSQIndexIT.java
0ctopus13prime: sounds good

src/test/java/org/opensearch/knn/memoryoptsearch/MOSFaissSQIndexIT.java
0ctopus13prime: Hi Naveen, that's the plan for the next PR.
This PR is fixing Faiss SQ and adding IT for it, for other index types, I'll cover in next PR.
Also AbstractMemoryOptimizedKnnSearchIT will cover all supported space types, only we need to care is method parameters.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissFlatIndexFactory.java
kotwanikunal: +1. Are we locking this feature down to just binary indexes? I think there is a disconnect in extensibility and architectural patterns. What about when we want to reuse this code for deduplication? We will have to rewrite the `FlatIndexFactory` constructs because it can only do binary indexes.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissFlatIndexFactory.java
kotwanikunal: I missed this comment. I echoed the same thing here: https://github.com/opensearch-project/k-NN/pull/3200/changes#r2976117841

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissFlatIndexFactory.java
0ctopus13prime: @kotwanikunal 
What we have to do is separating out FaissIndex and FaissBinaryIndex. They are different. 
To support FaissIndex, we have to have a dedicated factory for FaissIndex. 
Maybe we should change the method name to `createFaissBinaryIndex`, but I believe that's what we need to do. 
Making the return type to FaissBinaryIndex is not feature locking Kunal, we still be able to extend factory logic for FaissBinaryIndex and we can also continuously add factory logic for FaissIndex in parallel

jni/include/sq/faiss_sq_flat.h
navneet1v: Make sense. Then lets keep it like this.

jni/src/faiss_index_service.cpp
navneet1v: thanks

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/binary/FaissBinaryHnswIndex.java
navneet1v: [Not a blocker]
We can add a TODO here. But there are many ways to do this. We can actually do all this during de-serialization, or while doing de-serialization rather than getting a FaissIndex in return you can get a FaissIndexBuilder in return that you can change based on your needs. 

My main aim here is to ensure that once FaissIndex is built, its built. Noone should be able to change it.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/binary/FaissBinaryIndex.java
navneet1v: So if I am not wrong, what we are doing here we are using `FaissBinaryIndex` as the base index. Would it be good if we create another MetricType like Hamming in Java side and not C++ side?

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissFlatIndexFactory.java
navneet1v: @0ctopus13prime lets got with name change for the function.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissIndex.java
navneet1v: check this: https://github.com/opensearch-project/k-NN/pull/3200#discussion_r2979506626 Its fine right now if you want to keep it mutable. but I would suggest to add a TODO here. Since this will need refactoring.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissMemoryOptimizedSearcherFactory.java
navneet1v: @0ctopus13prime I thought we were using FaissIndex. But looks like we are not. I think we can resolve this comment

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/binary/FaissBinaryHnswIndex.java
0ctopus13prime: We can't do it during the deserialization. Null storage can be known at the end of the whole deserialization.
Naveen had a great idea to introduce `SetOnce<FaissBinaryIndex>` to allow assign-once and I like the idea.
Will add TODO there.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/binary/FaissBinaryIndex.java
0ctopus13prime: I'm not leaning to ward to having HAMMING metric in Java side, as C++ does not even have HAMMING as metric type. Simply Faiss does not honor the metric type there, it just use HAMMING as a default distance calculator regardless of metric_type.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissFlatIndexFactory.java
0ctopus13prime: sure will do

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissIndex.java
0ctopus13prime: sounds good

Issue #3178, opened 2026-03-17
[FEATURE] Lucene BBQ Flat Integration for 1 bit 


naveentatikonda:
### Description

The goal is to integrate Lucene's Better Binary Quantization (BBQ) in flat mode(no graph file created) into the OpenSearch k-NN plugin, exposing brute-force BBQ vector search as a first-class option for workloads where ingest efficiency, filtered search with very high selectivity, exact recall, and memory compression are the primary constraints.

#### Indexing
When a document is ingested, OpenSearch k-NN determines the appropriate vector format for the field. The request flows through `KNN1040BasePerFieldKnnVectorsFormat`, where the format is resolved by `LuceneCodecFormatResolver`. Based on the methodContext, this process instantiates `Lucene104ScalarQuantizedVectorsFormat` with `ScalarEncoding.SINGLE_BIT_QUERY_NIBBLE`, which corresponds to Lucene’s native BBQ flat codec.

#### Querying
Query execution has two phases, driven by `LuceneEngineKnnVectorQuery` wrapped inside `RescoreKnnVectorQuery`. It does not require any changes in k-NN but we are setting a `default oversampling factor` of `2.0` based on our benchmarking results.

#### UX Changes
To support BBQ flat through OpenSearch, we need to make some changes to the index mapping and introduce a new method named `flat` like hnsw which can be later supported with all the compression levels.

```
"mappings": {
    "properties": {
      "my_vector": {
        "type": "knn_vector",
        "dimension": 768,
        "space_type": "l2", -> must
        "compression_level": "32x" -> optional but default
        "method": {
          "name": "flat" -> must
        }
      }
    }
  }
``` 

In terms of query mapping, no changes and the existing knn query is used.

PR #3203, opened 2026-03-24
[BugFix] Fix Scorer, and max dimension for Lucene SQ 1 bit 

naveentatikonda:
### Description
This PR includes the following bug fixes for Lucene SQ 1 bit (both HNSW and Flat):
* Set max dimension to `16000` instead of `1024`
* Fix the scorer to use Bulk SIMD Scorer instead of Lucene scorer

#### Changes

* Add `KNN1040ScalarQuantizedVectorsFormat` and `KNN1040HnswScalarQuantizedVectorsFormat` to use custom bulk SIMD scorer for the `FLAT` and `HNSW` index type

* Wire `KNN1040ScalarQuantizedVectorsFormat` and `KNN1040HnswScalarQuantizedVectorsFormat` into `KNN1040PerFieldKnnVectorsFormat` for the FLAT resolver

* Add unit tests for the new `KNN1040ScalarQuantizedVectorsFormat` and `KNN1040HnswScalarQuantizedVectorsFormat` and codec resolver

* Rename `Faiss104ScalarQuantizedVectorScorer.java => KNN1040ScalarQuantizedVectorScorer.java` and `Faiss1040ScalarQuantizedUtils.java => KNN1040ScalarQuantizedUtils.java`

### Related Issues
https://github.com/opensearch-project/k-NN/issues/3178

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/KNN1040ScalarQuantizedUtils.java
+++ b/src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/KNN1040ScalarQuantizedUtils.java
@@ -8,8 +8,6 @@ package org.opensearch.knn.index.codec.KNN1040Codec;
 import lombok.experimental.UtilityClass;
 import org.apache.lucene.codecs.lucene104.QuantizedByteVectorValues;
 import org.apache.lucene.index.KnnVectorValues;
-import org.apache.lucene.index.VectorSimilarityFunction;
-import org.apache.lucene.util.VectorUtil;
 
 import java.io.IOException;
 import java.lang.reflect.Field;
@@ -46,19 +44,6 @@ class KNN1040ScalarQuantizedUtils {
      * and {@code throwExceptionIfNotFound} is {@code false}
      * @throws IOException if extraction fails and {@code throwExceptionIfNotFound} is {@code true}
      */
-    /**
-     * Normalizes the vector in place if the similarity function is COSINE and the vector is not already a unit vector.
-     * FAISS already normalizes query vectors for cosine, so this avoids redundant work while ensuring
-     * Lucene's {@link org.apache.lucene.util.quantization.OptimizedScalarQuantizer} assertion is satisfied.
-     *
-     * @param vector the vector to conditionally normalize
-     * @param similarityFunction the similarity function in use
-     */
-    public static void normalizeIfNeeded(final float[] vector, final VectorSimilarityFunction similarityFunction) {
-        if (similarityFunction == VectorSimilarityFunction.COSINE && !VectorUtil.isUnitVector(vector)) {
-            VectorUtil.l2normalize(vector);
-        }
-    }
 
     public static QuantizedByteVectorValues extractQuantizedByteVectorValues(final KnnVectorValues floatVectorValues) throws IOException {
         try {
--- a/src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/KNN1040ScalarQuantizedVectorScorer.java
+++ b/src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/KNN1040ScalarQuantizedVectorScorer.java
@@ -155,10 +155,9 @@ public class KNN1040ScalarQuantizedVectorScorer extends Lucene104ScalarQuantized
         // We make a copy as the quantization process mutates the input
         final float[] targetCopy = ArrayUtil.copyOfSubArray(target, 0, target.length);
 
-        // For cosine similarity, Lucene's OptimizedScalarQuantizer asserts the query is a unit vector.
-        // FAISS already normalizes the query, so we only normalize if it isn't already. But, we need to
-        // normalize the query vector when using Lucene engine.
-        KNN1040ScalarQuantizedUtils.normalizeIfNeeded(targetCopy, similarityFunction);
+        // For cosine similarity, the query vector is expected to already be normalized.
+        // Normalization is performed upfront in KNNQueryBuilder via VectorTransformerFactory
+        // for Lucene cosine with SQ 1-bit and flat methods and for Faiss.
 
         // Perform scalar quantization
         final OptimizedScalarQuantizer.QuantizationResult targetCorrectiveTerms = quantizer.scalarQuantize(
--- a/src/main/java/org/opensearch/knn/index/engine/faiss/AbstractFaissMethod.java
+++ b/src/main/java/org/opensearch/knn/index/engine/faiss/AbstractFaissMethod.java
@@ -169,6 +169,6 @@ public abstract class AbstractFaissMethod extends AbstractKNNMethod {
 
     @Override
     protected VectorTransformer getVectorTransformer(SpaceType spaceType) {
-        return VectorTransformerFactory.getVectorTransformer(KNNEngine.FAISS, spaceType);
+        return VectorTransformerFactory.getVectorTransformer(KNNEngine.FAISS, spaceType, null);
     }
 }
--- a/src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldType.java
+++ b/src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldType.java
@@ -210,13 +210,18 @@ public class KNNVectorFieldType extends MappedFieldType {
         final Optional<KNNMethodContext> knnMethodContext = knnMappingConfig.getKnnMethodContext();
         if (knnMethodContext.isPresent()) {
             KNNMethodContext context = knnMethodContext.get();
-            return VectorTransformerFactory.getVectorTransformer(context.getKnnEngine(), context.getSpaceType()).transform(vector, false);
+            return VectorTransformerFactory.getVectorTransformer(
+                context.getKnnEngine(),
+                context.getSpaceType(),
+                context.getMethodComponentContext()
+            ).transform(vector, false);
         }
         final Optional<String> modelId = knnMappingConfig.getModelId();
         if (modelId.isPresent()) {
             ModelDao modelDao = ModelDao.OpenSearchKNNModelDao.getInstance();
             final ModelMetadata metadata = modelDao.getMetadata(modelId.get());
-            return VectorTransformerFactory.getVectorTransformer(metadata.getKnnEngine(), metadata.getSpaceType()).transform(vector, false);
+            return VectorTransformerFactory.getVectorTransformer(metadata.getKnnEngine(), metadata.getSpaceType(), null)
+                .transform(vector, false);
         }
         throw new IllegalStateException("Either KNN method context or Model Id should be configured");
     }
--- a/src/main/java/org/opensearch/knn/index/mapper/VectorTransformerFactory.java
+++ b/src/main/java/org/opensearch/knn/index/mapper/VectorTransformerFactory.java
@@ -9,6 +9,14 @@ import lombok.AccessLevel;
 import lombok.NoArgsConstructor;
 import org.opensearch.knn.index.SpaceType;
 import org.opensearch.knn.index.engine.KNNEngine;
+import org.opensearch.knn.index.engine.MethodComponentContext;
+
+import java.util.Map;
+
+import static org.opensearch.knn.common.KNNConstants.ENCODER_SQ;
+import static org.opensearch.knn.common.KNNConstants.LUCENE_SQ_BITS;
+import static org.opensearch.knn.common.KNNConstants.METHOD_ENCODER_PARAMETER;
+import static org.opensearch.knn.common.KNNConstants.METHOD_FLAT;
 
 /**
  * Factory class responsible for creating appropriate vector transformers.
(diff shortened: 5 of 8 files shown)


src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/PrefetchingLucene104ScalarQuantizedVectorsFormat.java
```diff
@@ -0,0 +1,63 @@
...
+import org.apache.lucene.index.SegmentWriteState;
+import org.opensearch.knn.index.codec.scorer.PrefetchableLucene104ScalarQuantizedVectorScorer;
+
+import java.io.IOException;
+
+/**
+ * A {@link Lucene104ScalarQuantizedVectorsFormat} that uses a {@link PrefetchableLucene104ScalarQuantizedVectorScorer}
+ * to take advantage of prefetch optimizations during search.
+ */
+public class PrefetchingLucene104ScalarQuantizedVectorsFormat extends Lucene104ScalarQuantizedVectorsFormat {
```
Vikasht34: Why we need a new format ,..What was problem with Old One ?

src/main/java/org/opensearch/knn/index/codec/scorer/PrefetchableLucene104ScalarQuantizedVectorScorer.java
```diff
@@ -0,0 +1,53 @@
...
+import org.apache.lucene.util.hnsw.RandomVectorScorer;
+
+import java.io.IOException;
+
+/**
+ * A {@link Lucene104ScalarQuantizedVectorScorer} that wraps returned {@link RandomVectorScorer}
+ * instances with {@link PrefetchableFlatVectorScorer.PrefetchableRandomVectorScorer} to enable
+ * prefetch optimizations during search on the quantized scoring path.
+ */
+public class PrefetchableLucene104ScalarQuantizedVectorScorer extends Lucene104ScalarQuantizedVectorScorer {
```
Vikasht34: Again same here , Why new class?

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/PrefetchingLucene104ScalarQuantizedVectorsFormat.java
navneet1v: This is needed because we want to pass the correct Prefetchable scorer to Lucene104ScalarQuantizedVectorsFormat. So we need to override the format

src/main/java/org/opensearch/knn/index/codec/scorer/PrefetchableLucene104ScalarQuantizedVectorScorer.java
navneet1v: @naveentatikonda we already have a scorer which we were using for BulkSIMD. We should use that rather than creating a new one

src/test/java/org/opensearch/knn/index/codec/scorer/PrefetchableLucene104ScalarQuantizedVectorScorerTests.java
```diff
@@ -0,0 +1,132 @@
...
+import java.io.IOException;
+
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+/**
+ * Tests for {@link PrefetchableLucene104ScalarQuantizedVectorScorer}.
+ */
+public class PrefetchableLucene104ScalarQuantizedVectorScorerTests extends TestCase {
```
navneet1v: use KNNTestCase here rather than TestCase

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/PrefetchingLucene104ScalarQuantizedVectorsFormat.java
VijayanB: why not KNN104ScalarQuantizedVectorsFormat? Do we have to use prefix prefetch as part of format name?

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/KNN1040ScalarQuantizedVectorsFormat.java
```diff
@@ -0,0 +1,62 @@
...
+ */
+public class KNN1040ScalarQuantizedVectorsFormat extends Lucene104ScalarQuantizedVectorsFormat {
+
+    private static final Lucene104ScalarQuantizedVectorScorer SIMD_SCORER = new KNN1040ScalarQuantizedVectorScorer(
+        FlatVectorScorerUtil.getLucene99FlatVectorsScorer()
+    );
+
+    private static final Lucene99FlatVectorsFormat RAW_VECTOR_FORMAT = new Lucene99FlatVectorsFormat(
+        FlatVectorScorerUtil.getLucene99FlatVectorsScorer()
+    );
```
navneet1v:
```suggestion
    private static final Lucene104ScalarQuantizedVectorScorer SIMD_SCORER = new KNN1040ScalarQuantizedVectorScorer(
        FlatVectorScorerUtil.getLucene99FlatVectorsScorer()
    );

    private static final Lucene99FlatVectorsFormat RAW_VECTOR_FORMAT = new Lucene99FlatVectorsFormat(
        SIMD_SCORER
    );
```

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/KNN1040ScalarQuantizedVectorsFormat.java
```diff
@@ -0,0 +1,62 @@
...
+
+/**
+ * A {@link Lucene104ScalarQuantizedVectorsFormat} that uses a {@link KNN1040ScalarQuantizedVectorScorer}
+ * to take advantage of SIMD-accelerated scoring during search.
+ */
+public class KNN1040ScalarQuantizedVectorsFormat extends Lucene104ScalarQuantizedVectorsFormat {
+
+    private static final Lucene104ScalarQuantizedVectorScorer SIMD_SCORER = new KNN1040ScalarQuantizedVectorScorer(
+        FlatVectorScorerUtil.getLucene99FlatVectorsScorer()
+    );
```
navneet1v: rather than creating scorer like this you can move the into FlatVectorScorerProvider similar to how we have `FlatVectorScorerUtil.getLucene99FlatVectorsScorer()`. 

You can create a function named" getKNN1040ScalarQuantizedVectorScorer(), and then internally it can call do 
```
return new KNN1040ScalarQuantizedVectorScorer(
        FlatVectorScorerUtil.getLucene99FlatVectorsScorer()
    );
```

Benefits:
1. in all the format where we are using `new KNN1040ScalarQuantizedVectorScorer`, we can move them  to call this new function.
2. When we want to wrap the scorers with prefetch scorer, we can just update 1 single function `getKNN1040ScalarQuantizedVectorScorer` and every piece of code will get the PrefetchScorer.

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss1040PrefetchSupportKnnVectorReader.java
navneet1v: There is nothing related to prefetch on this reader. We should update the name of the reader here.

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss1040PrefetchSupportKnnVectorReader.java
```diff
@@ -71,7 +71,7 @@ public FloatVectorValues getFloatVectorValues(String field) throws IOException {
         final FloatVectorValues floatVectorValues = delegateFlatVectorsReader.getFloatVectorValues(field);
         return new ScalarQuantizedFloatVectorValuesWithIndexInputSlice(
```
navneet1v: a better naming for this class could have been `ScalarQuantizedFloatVectorValues` not sure why we think having `WithIndexInputSlice` at the end of class name.

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/KNN1040ScalarQuantizedVectorsFormat.java
naveentatikonda: Why do we need to pass SIMD_SCORER to RAW_VECTOR_FORMAT ?

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/KNN1040ScalarQuantizedVectorsFormat.java
naveentatikonda: Makes sense

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss1040PrefetchSupportKnnVectorReader.java
naveentatikonda: @0ctopus13prime is there any reason to create with this name `ScalarQuantizedFloatVectorValuesWithIndexInputSlice ` ?

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/KNN1040ScalarQuantizedVectorsFormat.java
navneet1v: my question would be why not? a format initialize readers and writers. So we should use the same scorer. The only reason I would not use the same scorer here is if KNN1040ScalarQuantizedVectorScorer is not compatible with Lucene99FlatVectorsFormat.

navneet1v: This will also ensure that when we are doing rescoring using Full precision vectors which will come from `Lucene99FlatVectorsFormat` we are able to use the right scorer.

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/KNN1040ScalarQuantizedVectorsFormat.java
```diff
@@ -0,0 +1,63 @@
...
+import org.apache.lucene.index.SegmentWriteState;
+import org.opensearch.knn.memoryoptsearch.faiss.FlatVectorsScorerProvider;
+
+import java.io.IOException;
+
+/**
+ * A {@link Lucene104ScalarQuantizedVectorsFormat} that uses a {@link KNN1040ScalarQuantizedVectorScorer}
+ * to take advantage of SIMD-accelerated scoring during search.
+ */
+public class KNN1040ScalarQuantizedVectorsFormat extends Lucene104ScalarQuantizedVectorsFormat {
```
0ctopus13prime, line 26: I think this should have a separate format name for recovery, in recovery engine will use the format name persisted in disk I believe.

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/KNN1040ScalarQuantizedVectorsFormat.java
```diff
@@ -0,0 +1,63 @@
...
+
```
0ctopus13prime, line 27: Also curious, max dimensions will still 1024?
Technically, this flat can take more than 65535 dimension. 

Lucene104ScalarQuantizedVectorsFormat will give only 1024, which seems too tight.
```
  @Override
  public int getMaxDimensions(String fieldName) {
    return 1024;
  }
```

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/ScalarQuantizedFloatVectorValues.java
```diff
@@ -31,7 +31,7 @@
  * slice that contains the quantized vector data and correction factors used during scoring.
  */
 @RequiredArgsConstructor
-class ScalarQuantizedFloatVectorValuesWithIndexInputSlice extends FloatVectorValues implements HasIndexSlice {
+class ScalarQuantizedFloatVectorValues extends FloatVectorValues implements HasIndexSlice {
```
0ctopus13prime, line 34: lol I thought this name is too long.
good change

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss1040PrefetchSupportKnnVectorReader.java
0ctopus13prime: yeah, I thought it's a good name combining
1. ScalarQuantized
2. and it's a FloatVectorValues
3. also it supports IndexInputSlice.

but feel free to change the name, I'm open to it also I thought the name is bit long

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/KNN1040ScalarQuantizedVectorsFormat.java
naveentatikonda: Right now, the max dimensions we are supporting for lucene engine is 16000
https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/engine/KNNEngine.java#L46-L53

0ctopus13prime: yeah, if that's the case then we should make this format override `getmaxDimensions` method

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/KNN1040ScalarQuantizedVectorScorer.java
```diff
@@ -147,6 +155,11 @@ private RandomVectorScorer bulkSimdRandomVectorScorer(
         // We make a copy as the quantization process mutates the input
         final float[] targetCopy = ArrayUtil.copyOfSubArray(target, 0, target.length);
 
+        // For cosine similarity, Lucene's OptimizedScalarQuantizer asserts the query is a unit vector.
+        // FAISS already normalizes the query, so we only normalize if it isn't already. But, we need to
+        // normalize the query vector when using Lucene engine.
+        KNN1040ScalarQuantizedUtils.normalizeIfNeeded(targetCopy, similarityFunction);
```
0ctopus13prime: If we end up choosing normalizing query vector within flat vector scorer, then can we avoid checking whether it's an unit vector?
We can just normalize `targetCopy`, as it's idempotent. 
`Normalize(Normalize(Normalize(vector))) == Normalize(vector)`

0ctopus13prime: Also curious, why we are not normalizing query vector in QueryBuilder side likewise we're doing for other cosine index type?

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/KNN1040ScalarQuantizedUtils.java
```diff
@@ -44,6 +46,20 @@ class Faiss1040ScalarQuantizedUtils {
...
+    /**
+     * Normalizes the vector in place if the similarity function is COSINE and the vector is not already a unit vector.
+     * FAISS already normalizes query vectors for cosine, so this avoids redundant work while ensuring
+     * Lucene's {@link org.apache.lucene.util.quantization.OptimizedScalarQuantizer} assertion is satisfied.
+     *
+     * @param vector the vector to conditionally normalize
+     * @param similarityFunction the similarity function in use
+     */
+    public static void normalizeIfNeeded(final float[] vector, final VectorSimilarityFunction similarityFunction) {
+        if (similarityFunction == VectorSimilarityFunction.COSINE && !VectorUtil.isUnitVector(vector)) {
```
0ctopus13prime: We don't need to check `isUnitVector`, it's waste of CPU cycles, we can just do the normalization.
Even the given query is an unit vector, normalization path will not change the vector itself

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/KNN1040ReadAdviceOverridingDirectory.java
```diff
@@ -0,0 +1,51 @@
...
+ * OS read-ahead prefetching. However, quantized vector data is read sequentially during search
+ * and merge operations. This wrapper intercepts {@link #openInput} and forces
+ * {@link DataAccessHint#SEQUENTIAL} for {@code .veb} files to enable OS read-ahead behavior,
+ * improving sequential I/O throughput. Raw vector data ({@code .vec}) is left unchanged since
+ * it is accessed randomly during rescoring.
+ *
+ * <p>This is used by both the Lucene SQ flat path ({@link KNN1040ScalarQuantizedVectorsFormat})
+ * and the Lucene SQ HNSW path ({@link KNN1040HnswScalarQuantizedVectorsFormat}).
+ */
+final class KNN1040ReadAdviceOverridingDirectory extends FilterDirectory {
```
navneet1v: This is not needed because in OpenSearch version > 3.1 and less than equal to 3.6  readAdvise is not honored

navneet1v: Ref: https://github.com/opensearch-project/OpenSearch/issues/21012

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/KNN1040ReadAdviceOverridingDirectory.java
```diff
@@ -0,0 +1,51 @@
...
+import org.apache.lucene.store.DataAccessHint;
+import org.apache.lucene.store.Directory;
+import org.apache.lucene.store.FilterDirectory;
+import org.apache.lucene.store.IOContext;
+import org.apache.lucene.store.IndexInput;
+
+import java.io.IOException;
+
+/**
+ * Directory wrapper that enforces {@link DataAccessHint#SEQUENTIAL} for {@code .veb} files
```
0ctopus13prime: NIT : .veb -> .veq

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/KNN1040ScalarQuantizedVectorScorer.java
naveentatikonda: moved it to VectorTransformer

Issue #3105, opened 2026-02-07
[FEATURE] Use BulkScorer from VectorScorer Interface of Lucene to do Exact Search


navneet1v:
## Description
Currently in KNN when we do Exact Search, we typically build a KNNExactSearchIterator and feed it docIds and VectorValues to iterate and score documents 1 by 1. Even though in the scorer we use the Lucene VectorSimilarity functions but scoring is done doc by doc.

## Suggestion
We can implement a new RandomVectorScorer to score the documents which can then use [bulkScore](https://github.com/apache/lucene/blob/787f11a96e7f651a3768b541a183477bad1f93e5/lucene/core/src/java/org/apache/lucene/util/hnsw/RandomVectorScorer.java#L50-L54) method to score multiple documents at a time. 

PR of lucene: https://github.com/apache/lucene/pull/14978 where they recommend scoring 64 docs at a time in exhaustive search. 

### Places where it can be used:
1. Efficient Filters
2. Disk based vector search rescoring phase.
3. Script score based exact search

navneet1v:
@Vikasht34 what I added was the interface and the default implementation. Here is one of the implementation of bulk scoring function: https://github.com/apache/lucene/blob/787f11a96e7f651a3768b541a183477bad1f93e5/lucene/core/src/java21/org/apache/lucene/internal/vectorization/Lucene99MemorySegmentFloatVectorScorer.java#L84-L97

The idea here is if start using the RandomScorer interface of Lucene then by default we will get distance computations one by one, but using the right implementation of VectorScorer  we can easily reuse the bulk scoring. Some ref: https://github.com/apache/lucene/blob/2df0351931f529cb9b85719a5eb9502c114621d4/lucene/core/src/java21/org/apache/lucene/internal/vectorization/Lucene99MemorySegmentFlatVectorsScorer.java#L83-L99

vamshin:
[like]  Nakkirtha, Vamshi Vijay reacted to your message:
________________________________
From: Navneet Verma ***@***.***>
Sent: Saturday, February 7, 2026 6:47:11 PM
To: opensearch-project/k-NN ***@***.***>
Cc: Subscribed ***@***.***>
Subject: [opensearch-project/k-NN] [FEATURE] Use BulkScorer from VectorScorer Interface of Lucene to do Exact Search (Issue #3105)

[...]

PR #3207, opened 2026-03-24
Refactor ExactSearcher to use VectorScorer instead of ExactKNNIterator

VijayanB:
### Description
Replace the ExactKNNIterator-based exact search with Lucene's VectorScorer API, delegating scorer creation to VectorScorers factory. This eliminates the need for multiple iterator subclasses (VectorIdsExactKNNIterator, BinaryVectorIdsExactKNNIterator, and their nested variants) by using VectorScorers.createScorer which handles vector type dispatch and nested wrapping via NestedBestChildVectorScorer.

- Replace getKNNIterator with createVectorScorer using VectorScorers factory
- Use VectorScorerMode (SCORE/RESCORE) to control quantization path
- Remove getMatchedDocsIterator conjunction logic (handled by scorer)
- Remove direct dependency on KNNFloatVectorValues, KNNByteVectorValues, KNNBinaryVectorValues and all ExactKNNIterator subclasses

### Related Issues
https://github.com/opensearch-project/k-NN/issues/3105

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/query/scorers/VectorScorers.java
+++ b/src/main/java/org/opensearch/knn/index/query/scorers/VectorScorers.java
@@ -28,6 +28,8 @@ import org.opensearch.knn.memoryoptsearch.faiss.FlatVectorsScorerProvider;
 
 import java.io.IOException;
 
+import static org.opensearch.knn.index.query.MemoryOptimizedSearchScoreConverter.convertInnerProductScoreToCosineScore;
+
 /**
  * Static factory for creating {@link VectorScorer} instances from {@link KNNVectorValuesIterator.DocIdsIteratorValues}.
  *
@@ -254,11 +256,15 @@ public final class VectorScorers {
     }
 
     /**
-         * Wraps an ADC {@link FlatVectorsScorer} to convert INNER_PRODUCT-format scores to
-         * COSINESIMIL-format. The ADCFlatVectorsScorer uses INNER_PRODUCT.scoreTranslation for
-         * cosine, which the MemoryOptimized path post-converts. In the exact search path there
-         * is no post-conversion, so this wrapper applies it at the scorer level.
-         */
+     * Wraps an ADC {@link FlatVectorsScorer} to convert INNER_PRODUCT-format scores to
+     * COSINESIMIL-format. The ADCFlatVectorsScorer uses INNER_PRODUCT.scoreTranslation for
+     * cosine, which the MemoryOptimized path post-converts. In the exact search path there
+     * is no post-conversion, so this wrapper applies it at the scorer level.
+     */
+    // TODO: Move this cosine score conversion into ADCFlatVectorsScorer itself so that it directly
+    // produces COSINESIMIL-format scores. This would eliminate the need for both this wrapper and
+    // the post-conversion in MemoryOptimizedKNNWeight (convertToCosineScore), keeping the
+    // conversion logic in a single place.
     private record CosineADCFlatVectorsScorer(FlatVectorsScorer delegate) implements FlatVectorsScorer {
 
         @Override
@@ -279,11 +285,7 @@ public final class VectorScorers {
             return new RandomVectorScorer.AbstractRandomVectorScorer(vectorValues) {
                 @Override
                 public float score(int node) throws IOException {
-                    float ipScore = inner.score(node);
-                    // Reverse INNER_PRODUCT.scoreTranslation to recover the raw inner product value,
-                    // then apply the COSINESIMIL score translation.
-                    float ip = ipScore >= 1 ? ipScore - 1 : 1 - 1 / ipScore;
-                    return SpaceType.COSINESIMIL.scoreTranslation(1 - ip);
+                    return convertInnerProductScoreToCosineScore(inner.score(node));
                 }
             };
         }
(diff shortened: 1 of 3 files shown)


src/main/java/org/opensearch/knn/index/query/scorers/VectorScorers.java
```diff
@@ -215,11 +216,17 @@ private static VectorScorer createADCScorer(
...
             fieldInfo,
             spaceType.getKnnVectorSimilarityFunction(),
             null
         );
+        // For COSINESIMIL, the ADCFlatVectorsScorer produces scores in INNER_PRODUCT format
+        // (used by MemoryOptimizedKNNWeight which post-converts via convertToCosineScore).
+        // In the exact search path there is no post-conversion, so we wrap the scorer to convert here.
+        if (spaceType == SpaceType.COSINESIMIL) {
+            adcFlatVectorsScorer = new CosineADCFlatVectorsScorer(adcFlatVectorsScorer);
+        }
```
navneet1v, line 231: we should move this wrapping to `FlatVectorsScorerProvider.getFlatVectorsScorer` for ADC case and other cases too. This will ensure that we can remove the `used by MemoryOptimizedKNNWeight which post-converts via convertToCosineScore` from the code and everything is handled with Scorers

src/main/java/org/opensearch/knn/index/query/scorers/VectorScorers.java
```diff
@@ -246,6 +253,51 @@ public Bulk bulk(final DocIdSetIterator matchingDocs) {
...
+            float[] target
+        ) throws IOException {
+            final RandomVectorScorer inner = delegate.getRandomVectorScorer(similarityFunction, vectorValues, target);
+            return new RandomVectorScorer.AbstractRandomVectorScorer(vectorValues) {
+                @Override
+                public float score(int node) throws IOException {
+                    float ipScore = inner.score(node);
+                    // Reverse INNER_PRODUCT.scoreTranslation to recover the raw inner product value,
+                    // then apply the COSINESIMIL score translation.
+                    float ip = ipScore >= 1 ? ipScore - 1 : 1 - 1 / ipScore;
```
navneet1v: is this not logic not already abstracted in some common utility function?

src/main/java/org/opensearch/knn/index/query/scorers/VectorScorers.java
```diff
@@ -279,12 +331,13 @@ private static VectorScorer createHammingDistanceScorer(
...
             target
         );
+
         return new VectorScorer() {
             final KnnVectorValues.DocIndexIterator iterator = byteVectorValues.iterator();
 
             @Override
             public float score() throws IOException {
-                return randomVectorScorer.score(iterator.docID());
+                return randomVectorScorer.score(iterator.index());
```
navneet1v, line 342: why we need this change?

src/main/java/org/opensearch/knn/index/query/scorers/VectorScorers.java
```diff
@@ -246,6 +253,51 @@ public Bulk bulk(final DocIdSetIterator matchingDocs) {
         };
     }
 
+    /**
+         * Wraps an ADC {@link FlatVectorsScorer} to convert INNER_PRODUCT-format scores to
+         * COSINESIMIL-format. The ADCFlatVectorsScorer uses INNER_PRODUCT.scoreTranslation for
+         * cosine, which the MemoryOptimized path post-converts. In the exact search path there
+         * is no post-conversion, so this wrapper applies it at the scorer level.
+         */
+    private record CosineADCFlatVectorsScorer(FlatVectorsScorer delegate) implements FlatVectorsScorer {
```
navneet1v, line 268: we need a Unit test for this scorer, if present already please let me know where it is.

src/main/java/org/opensearch/knn/index/query/scorers/VectorScorers.java
VijayanB: Good question! This change is necessary because of the refactoring from `ExactKNNIterator` to Lucene's `VectorScorer` API.

When we moved to `VectorScorer.score(int ordinal)`, the scorer expects an **ordinal** (the index position in the vector values array), not a Lucene document ID. The `DocIdSetIterator.index()` method provides this ordinal, while `docID()` would return the actual document ID, which would cause incorrect scoring for nested or sparse values.

src/main/java/org/opensearch/knn/index/query/scorers/VectorScorers.java
VijayanB: Added unit test now, it was previously caught by ADCIT integration test.

src/main/java/org/opensearch/knn/index/query/scorers/VectorScorers.java
VijayanB: Good point. Extracted this into method and is being shared now.

src/main/java/org/opensearch/knn/index/query/scorers/VectorScorers.java
VijayanB: I agree, however this is more complicated than making this change as part of this PR. It needs more testing . I added TODO and will clean this up as part of it in follow up PR

PR #3209, opened 2026-03-24
Support derived source for knn with other fields

### LINKED ISSUE #3058: [FEATURE] Converge OpenSearch core and k-NN Derived Source feature
k-NN plugin has implemented its own version of Derived Source mechanism to exclude the vector field values from the `_source` field. Meanwhile now OpenSearch core now also offers the Derived source feature. We should see if k-NN implementation can be converged to use the core support as [discussed in the past](https://github.com/opensearch-project/OpenSearch/issues/9568#issuecomment-2606078308) to avoid maintaining the full feature in plugin side.

Converge the Derived Source implementation with the one present in OpenSearch core

Keep both the implementation separate which will result in the overhead of maintaining both

NA

neetikasinghal:
### Description
Implement derived source for k-NN fields with core implementation, with a fallback to k-NN plugin's implementation. 

Related PR: https://github.com/opensearch-project/OpenSearch/pull/20991

Before this change, if `index.derived_source.enabled` and there is a knn field type, there will be error thrown saying that knn vector field doesnt supprt derived source feature.
After this change, if `index.derived_source.enabled` and there is a knn field type, the derived source feature will be turned on so that users can benefit out of both knn fields and other fields.

### Related Issues
Resolves #3058

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java
+++ b/src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java
@@ -31,11 +31,14 @@ import org.opensearch.core.common.Strings;
 import org.opensearch.core.xcontent.ToXContent;
 import org.opensearch.core.xcontent.XContentBuilder;
 import org.opensearch.core.xcontent.XContentParser;
+import org.opensearch.index.mapper.DerivedFieldGenerator;
 import org.opensearch.index.mapper.FieldMapper;
+import org.opensearch.index.mapper.FieldValueType;
 import org.opensearch.index.mapper.Mapper;
 import org.opensearch.index.mapper.MapperParsingException;
 import org.opensearch.index.mapper.ParametrizedFieldMapper;
 import org.opensearch.index.mapper.ParseContext;
+import org.opensearch.index.mapper.StoredFieldFetcher;
 import org.opensearch.knn.common.KNNConstants;
 import org.opensearch.knn.index.DerivedKnnByteVectorField;
 import org.opensearch.knn.index.DerivedKnnFloatVectorField;
@@ -921,6 +924,20 @@ public abstract class KNNVectorFieldMapper extends ParametrizedFieldMapper {
         if (includeDefaults || ignoreMalformed.explicit()) {
             builder.field(Names.IGNORE_MALFORMED, ignoreMalformed.value());
         }
+    }
+
+    @Override
+    protected void canDeriveSourceInternal() {
+        // skipping any checks here
+    }
+
+    /**
+     * Derive source using stored field, which would always be present for derived source enabled index field
+     */
+    @Override
+    protected DerivedFieldGenerator derivedFieldGenerator() {
+        // using knn vector fetcher to fetch the vector and passing it as a doc values fetcher
+        return new DerivedFieldGenerator(mappedFieldType, new KnnVectorValuesFetcher((KNNVectorFieldType) mappedFieldType, simpleName()), null);
     }
 
     /**
--- a/src/main/java/org/opensearch/knn/index/mapper/KnnVectorValuesFetcher.java
+++ b/src/main/java/org/opensearch/knn/index/mapper/KnnVectorValuesFetcher.java
@@ -1,0 +1,77 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch.knn.index.mapper;
+
+import org.apache.lucene.index.FieldInfo;
+import org.apache.lucene.index.LeafReader;
+import org.opensearch.index.mapper.FieldValueFetcher;
+import org.opensearch.knn.index.vectorvalues.KNNVectorValues;
+import org.opensearch.knn.index.vectorvalues.KNNVectorValuesFactory;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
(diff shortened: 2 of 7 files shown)


src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java
```diff
@@ -923,6 +926,20 @@ protected void doXContentBody(XContentBuilder builder, boolean includeDefaults,
...
+        // skipping any checks here
+    }
+
+    /**
+     * Derive source using stored field, which would always be present for derived source enabled index field
+     */
+    @Override
+    protected DerivedFieldGenerator derivedFieldGenerator() {
+        // using knn vector fetcher to fetch the vector and passing it as a doc values fetcher
+        return new DerivedFieldGenerator(mappedFieldType, new KnnVectorValuesFetcher((KNNVectorFieldType) mappedFieldType, simpleName()), null);
```
Vikasht34, line 940: I have one quick question Here , So Dervied Source is enabled at Index Time  and can be mappedFieldType.hasDocValues() is false so DerivedFieldGenerator will choose stored path but storedFieldFetcher = null in this PR , will it not cause NullPointerException .

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java
```diff
@@ -923,6 +926,20 @@ protected void doXContentBody(XContentBuilder builder, boolean includeDefaults,
         }
     }
 
+    @Override
+    protected void canDeriveSourceInternal() {
+        // skipping any checks here
```
Vikasht34, line 931: So I understand why we are skipping here because KNN  is having Mapping Validation , correct me if I am wrong but how we will make sure if I excluded-from-source fields are never later considered derivable accidentally

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java
neetikasinghal: yes for that, i was planning to make a change like: 
```
return new DerivedFieldGenerator(mappedFieldType, new KnnVectorValuesFetcher((KNNVectorFieldType) mappedFieldType, simpleName()), null) {
            @Override
            public FieldValueType getDerivedFieldPreference() {
                return FieldValueType.DOC_VALUES;
            }
        };
```

I was planning to use FieldValueType.DOC_VALUES always for knn case, KnnVectorValuesFetcher can be used as the FieldValueFetcher. Ideally we should have a new FieldValueType something like FieldValueType.KNN_VALUES, but since i dont see knn using DOC_VALUES, we can leverage FieldValueType.DOC_VALUES for vector values. what do u think?

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java
neetikasinghal: in other mappers, `canDeriveSourceInternal` calls `checkStoredAndDocValuesForDerivedSource` but for knn field mapper it is not needed for either of the doc values and stored flag to return true, so dont see a need to put any checks here.
```
protected void checkStoredAndDocValuesForDerivedSource() {
        if (!mappedFieldType.isStored() && !mappedFieldType.hasDocValues()) {
            throw new UnsupportedOperationException("Unable to derive source for [" + name() + "] with stored and " + "docValues disabled");
        }
    }
```

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java
buddharajusahil: I don't think mappedFieldType.hasDocValues() can ever return false for KNNVectorFieldType right? In KNNVectorFieldType constructor, it hard passes true for hasDocValues in super() which cannot be changed. 

https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldType.java#L111

https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/mapper/MappedFieldType.java#L100

neetikasinghal: @buddharajusahil 
>I don't think mappedFieldType.hasDocValues() can ever return false for KNNVectorFieldType right? In KNNVectorFieldType constructor, it hard passes true for hasDocValues in super() which cannot be changed.
https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldType.java#L111
https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/mapper/MappedFieldType.java#L100

Its set to false if Index >= 3.0.0
https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java#L241

buddharajusahil: @neetikasinghal I believe there are two different hasDocValues, one in KNNVectorFieldMapper and one on the KNNVectorFieldType, I believe the line you linked sets the one in KNNVectorFieldMapper to false, but I don't think the one in KNNVectorFieldType can ever be false, or set to false, and thats the one we are checking here.

neetikasinghal: i see got it, that makes sense. But this looks to be a bug to me in the plugin. Ideally the docValues in the KNNVectorFieldType should be inline with hasDocValues in the KNNVectorFieldMapper.

neetikasinghal:
https://github.com/opensearch-project/k-NN/pull/3260 - added pr for main

PR #3210, opened 2026-03-24
[Bug Fix] Fix default encoder to SQ 1 bit for faiss 32x compression

kotwanikunal:
### Description
- Fixes the faiss 32x compression to use the SQ 1 bit encoder

### Related Issues
Related - #3169 

### Check List

### DIFF
--- a/src/test/java/org/opensearch/knn/integ/ModeAndCompressionIT.java
+++ b/src/test/java/org/opensearch/knn/integ/ModeAndCompressionIT.java
@@ -514,7 +514,9 @@ public class ModeAndCompressionIT extends KNNRestTestCase {
             // since we are going to delete a document, so its better to have 1 more extra doc so that we can re-use some tests
             addKNNDocs(indexName, FIELD_NAME, DIMENSION, 0, NUM_DOCS + 1);
             addNonKNNDoc(indexName, String.valueOf(NUM_DOCS + 2), FIELD_NAME_NON_KNN, "Hello world");
-            deleteKnnDoc(indexName, "0");
+            // Delete the last doc (furthest from query vector) to avoid SQ 1-bit's better recall
+            // pulling a deleted doc into top-k results
+            deleteKnnDoc(indexName, String.valueOf(NUM_DOCS));
             flushIndex(indexName);
             validateGreenIndex(indexName);
             validateSearch(


src/main/java/org/opensearch/knn/index/engine/faiss/FaissMethodResolver.java
```diff
@@ -153,11 +154,12 @@ private void resolveEncoder(
         );
         encoderComponentContext.getParameters().putAll(resolvedParams);
 
-        // When auto-resolved to bits=1, remove the type default that was injected — the
-        // 1-bit quantization path doesn't use it, and validateEncoderConfig would reject it.
+        // When auto-resolved to bits=1, remove the type and clip defaults that were injected —
+        // the 1-bit quantization path doesn't use them, and validateEncoderConfig would reject them.
```
navneet1v, line 158: why there is injection of Clip when 1-bit is there?

kotwanikunal: The default for SQ faiss injects clip as a parameter by default - https://github.com/opensearch-project/k-NN/blame/44df2eb082b2ca68c6b2d5d2519aa23f2008931b/src/main/java/org/opensearch/knn/index/engine/faiss/FaissSQEncoder.java#L88-L93

For 1-bit we want to prevent clip or type from being resolved within the mappings.

Vikasht34: Needed to understand more , what problem it was causing?

kotwanikunal: To clarify the full flow:

1. When x32 compression auto resolves to sq(bits=1), the resolver creates a new MethodComponentContext with only bits=1 set
2. getParameterMapWithDefaultsAdded() then injects ALL registered parameter defaults for the SQ encoder this includes type=fp16 (default) and clip=false (default), even though they're not applicable for the 1-bit path
3. validateEncoderConfig() then rejects type and clip when bits != 16, since these parameters are only meaningful for fp16 quantization
4. So we remove these injected defaults before validation to avoid a false rejection

This only affects the default/implicit resolution path. When a user explicitly provides sq(bits=1, type=fp16) or sq(bits=1, clip=true), the validation correctly rejects them with a clear error message.

Vikasht34: Make Sense ... Thanks

navneet1v (approved):

naveentatikonda (approved):

PR #3216, opened 2026-03-27
Store vector L2 norm to restore original vectors for Faiss cosine with derived source

### LINKED ISSUE #3083: [FEATURE] Enable cosine similarity to return original vectors rather than normalized vectors with Faiss Engine
## Description
Currently, when using cosine similarity with Faiss Engine, OpenSearch/k-NN does not return the original non-normalized vectors with Derived Source Enabled. Since we normalize the vectors before storing them to avoid doing normalization at every distance computation phase(segment flush, merges, search query).

Due to this limitation a user who needs original vectors for re-ranking or any other purpose has to do either of this:
1. Store vectors outside of OpenSearch too.
2. Store the norm with the document and then reconstruct vectors at the final step using script score(which is very slow) to recover original vectors, resulting in overheads.

## Request:
Add the ability to store the norm of the vector along side flat vectors and then return the original vector in derived source eliminating the need for the workarounds

krocky-cooky:
#### Problem
When Faiss engine with cosine similarity and derived source are both enabled, vectors are L2-normalized in-place 
during indexing. Since derived source reconstructs `_source` from doc values, it returns normalized vectors instead 
of the user's original input.

#### Solution
Store the pre-normalization L2 norm as a `FloatDocValuesField` (`_knn_norm_<field_name>`) alongside the vector. On `_source` reconstruction, read the norm and multiply it back to recover the original vector.

#### Key changes

Write path:
- `VectorTransformer` / `NormalizeVectorTransformer`: Added `computeL2Norm()` to calculate norm before normalization
- `KNNVectorFieldMapper.parseCreateField`: Computes norm, passes it to `getFieldsForFloatVector` which adds a 
`FloatDocValuesField` when derived source is enabled and norm ≠ 1.0
- `DerivedSourceIndexOperationListener`: Denormalizes vectors before injecting into translog source
- `KNN10010DerivedSourceStoredFieldsWriter`: Records norm field names in segment attributes 

Read path:
- `DerivedSourceNormSupplier`: Functional interface with `UNIT` (no-op) and `fromDocValues` (reads from `NumericDocValues`) implementations
- `PerFieldDerivedVectorTransformerFactory`: Looks up norm FieldInfo and creates the appropriate `DerivedSourceNormSupplier`
- `AbstractPerFieldDerivedVectorTransformer.formatVector`: Lazily reads norm and applies denormalization on both 
byte[] (legacy BinaryDocValues) and float[] (Lucene-based vector format) paths

Utilities:
- `KNNVectorUtil`: Added `denormalize(vector, norm, inplace)` and `getNormFieldName(fieldName)`

#### Backward compatibility
- Segments without norm attributes fall back to norm = 1.0 (no denormalization)
- Only Faiss + cosinesimil + derived source triggers norm storage; all other configurations are unaffected

### Related Issues
Resolves #3083 

### Check List
- [ ] API changes companion pull request [[created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md)](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
- [ ] Public documentation issue/PR [[created](https://github.com/opensearch-project/documentation-website/issues/new/choose)](https://github.com/opensearch-project/documentation-website/issues/new/choose).

For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/k-NN/blob/main/CONTRIBUTING.md#developer-certificate-of-origin)

### DIFF
--- a/src/main/java/org/opensearch/knn/common/KNNConstants.java
+++ b/src/main/java/org/opensearch/knn/common/KNNConstants.java
@@ -219,6 +219,7 @@ public class KNNConstants {
     // Define here: https://github.com/opensearch-project/remote-vector-index-builder/blob/main/API.md#index-parameters
     public static final int MIN_DOCS_FOR_REMOTE_INDEX_BUILD = 4;
 
-    // Norm field
+    // Prefix for the NumericDocValues field that stores the L2 norm of vectors before normalization.
+    // Used with derived source to restore original vectors when Faiss + cosinesimil is configured.
     public static final String NORM_FIELD_PREFIX = "_knn_norm_";
 }
--- a/src/main/java/org/opensearch/knn/common/KNNConstants.java
+++ b/src/main/java/org/opensearch/knn/common/KNNConstants.java
@@ -218,4 +218,7 @@ public class KNNConstants {
     public static final int BYTE_ALIGNMENT_MASK = 7; // Used for rounding up to nearest byte (Byte.SIZE - 1)
     // Define here: https://github.com/opensearch-project/remote-vector-index-builder/blob/main/API.md#index-parameters
     public static final int MIN_DOCS_FOR_REMOTE_INDEX_BUILD = 4;
+
+    // Norm field
+    public static final String NORM_FIELD_PREFIX = "_knn_norm_";
 }
--- a/src/main/java/org/opensearch/knn/common/KNNVectorUtil.java
+++ b/src/main/java/org/opensearch/knn/common/KNNVectorUtil.java
@@ -81,6 +81,10 @@ public class KNNVectorUtil {
      * @return the denormalized vector
      */
     public static float[] denormalize(float[] vector, float norm, boolean inplace) {
+        Objects.requireNonNull(vector, "vector must no be null");
+        if (norm <= 0 || !Float.isFinite(norm)) {
+            throw new IllegalArgumentException(("norm must be a positive finite value, got: " + norm));
+        }
         float[] result = inplace ? vector : Arrays.copyOf(vector, vector.length);
         for (int i = 0; i < result.length; i++) {
             result[i] *= norm;
--- a/src/main/java/org/opensearch/knn/common/KNNVectorUtil.java
+++ b/src/main/java/org/opensearch/knn/common/KNNVectorUtil.java
@@ -81,9 +81,9 @@ public class KNNVectorUtil {
      * @return the denormalized vector
      */
     public static float[] denormalize(float[] vector, float norm, boolean inplace) {
-        Objects.requireNonNull(vector, "vector must no be null");
+        Objects.requireNonNull(vector, "vector must not be null");
         if (norm <= 0 || !Float.isFinite(norm)) {
-            throw new IllegalArgumentException(("norm must be a positive finite value, got: " + norm));
+            throw new IllegalArgumentException("norm must be a positive finite value, got: " + norm);
         }
         float[] result = inplace ? vector : Arrays.copyOf(vector, vector.length);
         for (int i = 0; i < result.length; i++) {
--- a/src/main/java/org/opensearch/knn/common/KNNVectorUtil.java
+++ b/src/main/java/org/opensearch/knn/common/KNNVectorUtil.java
@@ -7,8 +7,8 @@ package org.opensearch.knn.common;
 
 import lombok.AccessLevel;
 import lombok.NoArgsConstructor;
-import org.apache.lucene.util.VectorUtil;
 
+import java.util.Arrays;
 import java.util.List;
 import java.util.Objects;
 
@@ -60,5 +60,31 @@ public class KNNVectorUtil {
             intArray[i] = integerList.get(i);
         }
         return intArray;
+    }
+
+    /**
+     * Get the doc values field name used to store the L2 norm for a given vector field.
+     *
+     * @param fieldName the vector field name
+     * @return the norm field name
+     */
+    public static String getNormFieldName(String fieldName) {
+        return KNNConstants.NORM_FIELD_PREFIX + fieldName;
+    }
+
+    /**
+     * Denormalize a normalized vector by multiplying each element by the given L2 norm.
+     *
+     * @param vector the normalized vector
+     * @param norm the L2 norm to restore
+     * @param inplace if true, modifies the input array; if false, returns a new array
+     * @return the denormalized vector
+     */
+    public static float[] denormalize(float[] vector, float norm, boolean inplace) {
+        float[] result = inplace ? vector : Arrays.copyOf(vector, vector.length);
+        for (int i = 0; i < result.length; i++) {
+            result[i] *= norm;
+        }
+        return result;
     }
 }
--- a/src/main/java/org/opensearch/knn/common/KNNVectorUtil.java
+++ b/src/main/java/org/opensearch/knn/common/KNNVectorUtil.java
@@ -7,6 +7,7 @@ package org.opensearch.knn.common;
 
 import lombok.AccessLevel;
 import lombok.NoArgsConstructor;
+import org.apache.lucene.util.VectorUtil;
 
 import java.util.List;
 import java.util.Objects;
--- a/src/main/java/org/opensearch/knn/index/DerivedKnnFloatVectorField.java
+++ b/src/main/java/org/opensearch/knn/index/DerivedKnnFloatVectorField.java
@@ -28,6 +28,10 @@ public class DerivedKnnFloatVectorField extends KnnFloatVectorField {
      * @param vector vector for the field
      * @param isDerivedEnabled boolean to indicate if derived source is enabled
      */
+    public DerivedKnnFloatVectorField(String name, float[] vector, boolean isDerivedEnabled) {
+        this(name, vector, isDerivedEnabled, 1.0f);
+    }
+
     public DerivedKnnFloatVectorField(String name, float[] vector, boolean isDerivedEnabled, float norm) {
         super(name, vector);
         this.isDerivedEnabled = isDerivedEnabled;
@@ -41,6 +45,10 @@ public class DerivedKnnFloatVectorField extends KnnFloatVectorField {
      * @param fieldType FieldType of the field
      * @param isDerivedEnabled boolean to indicate if derived source is enabled
      */
+    public DerivedKnnFloatVectorField(String name, float[] vector, FieldType fieldType, boolean isDerivedEnabled) {
+        this(name, vector, fieldType, isDerivedEnabled, 1.0f);
+    }
+
     public DerivedKnnFloatVectorField(String name, float[] vector, FieldType fieldType, boolean isDerivedEnabled, float norm) {
         super(name, vector, fieldType);
         this.isDerivedEnabled = isDerivedEnabled;
--- a/src/main/java/org/opensearch/knn/index/DerivedKnnFloatVectorField.java
+++ b/src/main/java/org/opensearch/knn/index/DerivedKnnFloatVectorField.java
@@ -19,15 +19,19 @@ public class DerivedKnnFloatVectorField extends KnnFloatVectorField {
     @Getter
     private final boolean isDerivedEnabled;
 
+    @Getter
+    private final float vectorNorm;
+
     /**
      *
      * @param name Name of the field
      * @param vector vector for the field
      * @param isDerivedEnabled boolean to indicate if derived source is enabled
      */
-    public DerivedKnnFloatVectorField(String name, float[] vector, boolean isDerivedEnabled) {
+    public DerivedKnnFloatVectorField(String name, float[] vector, boolean isDerivedEnabled, float norm) {
         super(name, vector);
         this.isDerivedEnabled = isDerivedEnabled;
+        this.vectorNorm = norm;
     }
 
     /**
@@ -37,8 +41,9 @@ public class DerivedKnnFloatVectorField extends KnnFloatVectorField {
      * @param fieldType FieldType of the field
      * @param isDerivedEnabled boolean to indicate if derived source is enabled
      */
-    public DerivedKnnFloatVectorField(String name, float[] vector, FieldType fieldType, boolean isDerivedEnabled) {
+    public DerivedKnnFloatVectorField(String name, float[] vector, FieldType fieldType, boolean isDerivedEnabled, float norm) {
         super(name, vector, fieldType);
         this.isDerivedEnabled = isDerivedEnabled;
+        this.vectorNorm = norm;
     }
 }
--- a/src/main/java/org/opensearch/knn/index/codec/KNN10010Codec/KNN10010DerivedSourceStoredFieldsWriter.java
+++ b/src/main/java/org/opensearch/knn/index/codec/KNN10010Codec/KNN10010DerivedSourceStoredFieldsWriter.java
@@ -25,9 +25,12 @@ import org.opensearch.core.xcontent.XContentBuilder;
 import org.opensearch.index.mapper.MappedFieldType;
 import org.opensearch.index.mapper.MapperService;
 import org.opensearch.index.mapper.SourceFieldMapper;
+import org.opensearch.knn.index.SpaceType;
 import org.opensearch.knn.index.codec.backward_codecs.KNN9120Codec.KNN9120DerivedSourceStoredFieldsReader;
 import org.opensearch.knn.index.codec.derivedsource.DerivedFieldInfo;
 import org.opensearch.knn.index.codec.derivedsource.DerivedSourceSegmentAttributeParser;
+import org.opensearch.knn.index.engine.KNNEngine;
+import org.opensearch.knn.index.engine.KNNMethodContext;
 import org.opensearch.knn.index.mapper.KNNVectorFieldType;
 import org.opensearch.knn.index.util.IndexUtil;
 
@@ -36,6 +39,7 @@ import java.nio.ByteBuffer;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.Set;
 import java.util.function.Function;
 import java.util.stream.Collectors;
@@ -50,6 +54,7 @@ public class KNN10010DerivedSourceStoredFieldsWriter extends StoredFieldsWriter 
 
     private final Set<String> knownVectorFields = new HashSet<>();
     private final Set<String> knownNestedVectorFields = new HashSet<>();
+    private final Set<String> knownNormFields = new HashSet<>();
 
     private Function<Map<String, Object>, Map<String, Object>> vectorMask;
 
@@ -149,10 +154,25 @@ public class KNN10010DerivedSourceStoredFieldsWriter extends StoredFieldsWriter 
                         knownVectorFields.add(fieldName);
                     }
                     allFields.add(fieldName);
+
+                    if (needsNormField(knnVectorFieldType)) {
+                        knownNormFields.add(fieldName);
+                    }
                 }
             }
         }
         vectorMask = buildMask(allFields);
+    }
+
+    /**
+     * Determines if a vector field needs a norm doc values field for denormalization.
+     * This is only needed for Faiss engine with cosine similarity, where vectors are normalized at index time.
+     */
+    private static boolean needsNormField(KNNVectorFieldType fieldType) {
+        Optional<KNNMethodContext> ctx = fieldType.getKnnMappingConfig().getKnnMethodContext();
+        return ctx.isPresent()
+            && ctx.get().getKnnEngine() == KNNEngine.FAISS
+            && ctx.get().getSpaceType() == SpaceType.COSINESIMIL;
     }
 
     /**
@@ -175,6 +195,7 @@ public class KNN10010DerivedSourceStoredFieldsWriter extends StoredFieldsWriter 
 
         Set<String> vectorFields = new HashSet<>();
         Set<String> nestedVectorFields = new HashSet<>();
+        Set<String> normFields = new HashSet<>();
 
         for (int i = 0; i < mergeState.storedFieldsReaders.length; i++) {
             StoredFieldsReader reader = mergeState.storedFieldsReaders[i];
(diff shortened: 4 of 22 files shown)


krocky-cooky:
Among several implementation approaches, I believe this implementation is the best option, though there are a few
considerations:

- **Due to floating-point rounding errors, there will be minor differences between the original vector and the vector returned in `_source.`** 
Since the difference is less than 1e-4, this should not be a significant issue in practice.
- **The norm is stored as doc values under `_knn_norm_<vector_field>`, which incurs additional storage usage.** 
Since it is a single floating-point value and is only stored under the cosinesimil + faiss condition, the impact is
minimal.
- **During the read phase, there is additional overhead from reading the norm from disk.** 
Since the norm is stored in 
the same file as the vectors, OS-level caching can be leveraged, keeping the impact small.

Happy to address any feedback!

src/main/java/org/opensearch/knn/common/KNNVectorUtil.java
```diff
@@ -60,4 +61,30 @@ public static int[] intListToArray(final List<Integer> integerList) {
...
+    /**
+     * Denormalize a normalized vector by multiplying each element by the given L2 norm.
+     *
+     * @param vector the normalized vector
+     * @param norm the L2 norm to restore
+     * @param inplace if true, modifies the input array; if false, returns a new array
+     * @return the denormalized vector
+     */
+    public static float[] denormalize(float[] vector, float norm, boolean inplace) {
+        float[] result = inplace ? vector : Arrays.copyOf(vector, vector.length);
```
kotwanikunal, line 88: nit: add in checks here to prevent NPEs

```
public static float[] denormalize(float[] vector, float norm, boolean inplace) {
   Objects.requireNonNull(vector, "vector must not be null");
   if (norm <= 0 || !Float.isFinite(norm)) {
       throw new IllegalArgumentException("norm must be a positive finite value, got: " + norm);
   }
  ...
}
```

src/main/java/org/opensearch/knn/index/codec/derivedsource/DerivedSourceNormSupplier.java
```diff
@@ -0,0 +1,46 @@
...
+    /**
+     * Create a DerivedSourceNormSupplier backed by NumericDocValues.
+     *
+     * @param supplier supplies a fresh NumericDocValues iterator on each call
+     * @return DerivedSourceNormSupplier that reads norm from doc values
+     */
+    static DerivedSourceNormSupplier fromDocValues(CheckedSupplier<NumericDocValues, IOException> supplier) {
+        return (docId) -> {
+            NumericDocValues dv = supplier.get();
+            dv.advance(docId);
```
kotwanikunal: This could go past to NO_MORE_DOCS if we are missing the norm in doc values (compatibility or otherwise).
Can you add in checks to ensure it returns the correct value? Or better use `advanceExact` - https://lucene.apache.org/core/10_4_0/core/org/apache/lucene/index/NumericDocValues.html#advanceExact(int)

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java
```diff
@@ -787,8 +792,10 @@ protected void parseCreateField(ParseContext context, int dimension, VectorDataT
             }
             final float[] array = floatsArrayOptional.get();
             getVectorValidator().validateVector(array);
-            getVectorTransformer().transform(array, true);
-            context.doc().addAll(getFieldsForFloatVector(array, isDerivedEnabled(context)));
+            VectorTransformer transformer = getVectorTransformer();
```
kotwanikunal, line 795: Shouldn't this be limited just to the Cosine space? Given that we have access to `VectorDataType`?

src/main/java/org/opensearch/knn/index/codec/derivedsource/AbstractPerFieldDerivedVectorTransformer.java
```diff
@@ -28,14 +29,48 @@ protected Object formatVector(
...
-        // If the vector value is a byte[], we must deserialize
+        // If the vector value is a byte[], we must deserialize.
+        // This is the Faiss/nmslib path where vectors are stored as BinaryDocValues.
         if (vectorValue instanceof byte[]) {
             BytesRef vectorBytesRef = new BytesRef((byte[]) vectorValue);
             VectorDataType vectorDataType = FieldInfoExtractor.extractVectorDataType(fieldInfo);
-            return KNNVectorFieldMapperUtil.deserializeStoredVector(vectorBytesRef, vectorDataType);
+            Object deserialized = KNNVectorFieldMapperUtil.deserializeStoredVector(vectorBytesRef, vectorDataType);
+            if (deserialized instanceof float[] floatVector) {
+                float norm = normSupplier.get();
```
kotwanikunal: nit: This code seems duplicated. Consider moving it into a method

```
float[] vector = ...
float norm = normSupplier.get();
if (norm != 1.0f) {
        KNNVectorUtil.denormalize(vector, norm, true);
}
```

src/test/java/org/opensearch/knn/integ/DerivedSourceIT.java
```diff
@@ -783,4 +783,109 @@ private void assertMappingLevelSourceFiltering(String indexName, String[] expect
...
+        );
+        // Create index with derived source disabled (baseline)
+        createKnnIndex(
+            indexDisabled,
+            Settings.builder().put("index.knn", true).put("index.knn.derived_source.enabled", false).build(),
+            mapping
+        );
+
+        // Index documents with non-unit vectors
+        Random random = new Random(42);
```
kotwanikunal, line 830: Can we also test with [0, 0,...] values? Just to ensure we cover the multiplicative factors correctly?

src/main/java/org/opensearch/knn/index/DerivedKnnFloatVectorField.java
```diff
@@ -19,15 +19,19 @@ public class DerivedKnnFloatVectorField extends KnnFloatVectorField {
...
+    private final float vectorNorm;
+
     /**
      *
      * @param name Name of the field
      * @param vector vector for the field
      * @param isDerivedEnabled boolean to indicate if derived source is enabled
      */
-    public DerivedKnnFloatVectorField(String name, float[] vector, boolean isDerivedEnabled) {
+    public DerivedKnnFloatVectorField(String name, float[] vector, boolean isDerivedEnabled, float norm) {
```
kotwanikunal, line 35: Can we maintain the original signature with a 1.0 value?

```
public DerivedKnnFloatVectorField(String name, float[] vector, boolean isDerivedEnabled) {
    ...
    this.vectorNorm = 1.0f;
}
```

kotwanikunal (commented):
Thanks for making the changes. Took an initial pass through the changes. 
I'll revisit it again.

src/main/java/org/opensearch/knn/common/KNNVectorUtil.java
krocky-cooky: added.

src/main/java/org/opensearch/knn/index/codec/derivedsource/DerivedSourceNormSupplier.java
krocky-cooky: Changed to return 1.0f if NO_MORE_DOCS emerged.

src/main/java/org/opensearch/knn/index/codec/derivedsource/AbstractPerFieldDerivedVectorTransformer.java
krocky-cooky: added!

src/test/java/org/opensearch/knn/integ/DerivedSourceIT.java
krocky-cooky: Added.

src/main/java/org/opensearch/knn/index/DerivedKnnFloatVectorField.java
krocky-cooky: Added original signature.

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java
krocky-cooky: Adding a cosinesimil check there might duplicate the branching logic that's already encapsulated in 
[VectorTransformerFactory](https://github.com/opensearch-project/k-NN/blob/658c0650d635afb66ab51ae7749ff08caccd4227/src/main/java/org/opensearch/knn/index/mapper/VectorTransformerFactory.java#L58). Instead, I renamed computeL2Norm to getTransformNormFactor — the previous name gave the impression that it always computes an L2 norm, which made the default 1.0f return value confusing. The new name makes it clear that the method returns the norm factor associated with the transformation (1.0f when no normalization is applied). What do you think?

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java
```diff
@@ -698,12 +700,15 @@ private Field createVectorField(byte[] vectorValue, boolean isDerivedEnabled) {
...
      */
-    protected List<Field> getFieldsForFloatVector(final float[] array, boolean isDerivedEnabled) {
+    protected List<Field> getFieldsForFloatVector(final float[] array, boolean isDerivedEnabled, float norm) {
         final List<Field> fields = new ArrayList<>();
-        fields.add(createVectorField(array, isDerivedEnabled));
+        fields.add(createVectorField(array, isDerivedEnabled, norm));
         if (this.stored) {
             fields.add(createStoredFieldForFloatVector(name(), array));
         }
+        if (isDerivedEnabled && norm != 1.0f) {
```
Vikasht34, line 709: So this is fine , here we are checking norm != 1.0 but in writer we have different condition though   so vector that happens to already be unit-length won't get a norm field   
  stored, which is fine (1.0f fallback), but it means the segment attribute will still list the field as having norms even if no documents actually wrote one right?

src/main/java/org/opensearch/knn/index/codec/derivedsource/DerivedSourceNormSupplier.java
```diff
@@ -0,0 +1,48 @@
...
+
+    /**
+     * Create a DerivedSourceNormSupplier backed by NumericDocValues.
+     *
+     * @param supplier supplies a fresh NumericDocValues iterator on each call
+     * @return DerivedSourceNormSupplier that reads norm from doc values
+     */
+    static DerivedSourceNormSupplier fromDocValues(CheckedSupplier<NumericDocValues, IOException> supplier) {
+        return (docId) -> {
+            NumericDocValues dv = supplier.get();
```
Vikasht34, line 41: Every call to getNorm(docId) creates a fresh NumericDocValues iterator. This is safe (no stale state) but potentially expensive. DocValuesProducer.getNumeric() typically returns a new iterator. In the read path, this is called once per document per field during _source reconstruction. For a GET or search with _source enabled, this is fine. But during a force merge that touches millions of docs, the overhead could add up.  @navneet1v  what do u think here?

src/main/java/org/opensearch/knn/common/KNNVectorUtil.java
```diff
@@ -60,4 +61,34 @@ public static int[] intListToArray(final List<Integer> integerList) {
...
+    /**
+     * Denormalize a normalized vector by multiplying each element by the given L2 norm.
+     *
+     * @param vector the normalized vector
+     * @param norm the L2 norm to restore
+     * @param inplace if true, modifies the input array; if false, returns a new array
+     * @return the denormalized vector
+     */
+    public static float[] denormalize(float[] vector, float norm, boolean inplace) {
+        Objects.requireNonNull(vector, "vector must no be null");
```
Vikasht34: nit :- vector must not be null

src/main/java/org/opensearch/knn/common/KNNConstants.java
```diff
@@ -218,4 +218,8 @@ public class KNNConstants {
     public static final int BYTE_ALIGNMENT_MASK = 7; // Used for rounding up to nearest byte (Byte.SIZE - 1)
     // Define here: https://github.com/opensearch-project/remote-vector-index-builder/blob/main/API.md#index-parameters
     public static final int MIN_DOCS_FOR_REMOTE_INDEX_BUILD = 4;
+
+    // Prefix for the NumericDocValues field that stores the L2 norm of vectors before normalization.
+    // Used with derived source to restore original vectors when Faiss + cosinesimil is configured.
+    public static final String NORM_FIELD_PREFIX = "_knn_norm_";
```
Vikasht34, line 224: do we already prevent user-defined mappings from colliding with this prefix? I think we should either reserve it explicitly

src/main/java/org/opensearch/knn/index/codec/KNN10010Codec/KNN10010DerivedSourceStoredFieldsWriter.java
```diff
@@ -202,6 +224,9 @@ public int merge(MergeState mergeState) throws IOException {
         if (!nestedVectorFields.isEmpty()) {
             DerivedSourceSegmentAttributeParser.addDerivedVectorFieldsSegmentInfoAttribute(segmentInfo, nestedVectorFields, true);
         }
+        if (!normFields.isEmpty()) {
```
Vikasht34: Can we have a test that has an old segment with no norm metadata + a new segment with norm metadata, followed by merge and _source reconstruction?

src/main/java/org/opensearch/knn/index/mapper/VectorTransformer.java
```diff
@@ -35,4 +35,8 @@ default void transform(final byte[] vector) {
             throw new IllegalArgumentException("Input vector cannot be null");
         }
     }
+
+    default float getTransformNormFactor(final float[] vector) {
```
Vikasht34: nit:- Add Java Doc here

src/main/java/org/opensearch/knn/index/mapper/NormalizeVectorTransformer.java
```diff
@@ -43,4 +43,16 @@ private void validateVector(float[] vector) {
...
     }
+
+    /**
+     * Calculate L2 norm of a given float vector as the norm factor for transformation.
+     *
+     * @param vector the vector
+     * @return L2 norm of the vector
+     */
+    @Override
+    public float getTransformNormFactor(final float[] vector) {
```
Vikasht34: so this is expensive becuase now we parseCreateField calls this sequence, 
float norm = transformer.getTransformNormFactor(array);                                                                                                                           
  transformer.transform(array, true); 

The L2 norm is computed twice.  For high-dimensional vectors (768, 1024 dims), dotProduct is the expensive part. This is a ~2x overhead on the normalization hot path during indexing and when search will be happening in parallel , this is more overhead specially for 32x compression with smaller machine

src/main/java/org/opensearch/knn/common/KNNVectorUtil.java
```diff
@@ -60,4 +61,34 @@ public static int[] intListToArray(final List<Integer> integerList) {
...
+        if (norm <= 0 || !Float.isFinite(norm)) {
+            throw new IllegalArgumentException(("norm must be a positive finite value, got: " + norm));
```
Vikasht34: nit double parathensis

src/main/java/org/opensearch/knn/index/codec/derivedsource/NestedPerFieldDerivedVectorTransformer.java
```diff
@@ -34,7 +42,12 @@ public Object apply(Object object) {
...
 
         try {
-            Object vector = formatVector(childFieldInfo, vectorValues::getVector, vectorValues::conditionalCloneVector);
+            Object vector = formatVector(
+                childFieldInfo,
+                vectorValues::getVector,
+                vectorValues::conditionalCloneVector,
+                () -> normSupplier.getNorm(currentOffset)
+            );
             vectorValues.nextDoc();
```
Vikasht34, line 51: currentOffset is set once in setCurrentDoc and never updated.  for multi-valued nested fields, all nested vectors would use the first child's norm?

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java
krocky-cooky: I implemented writing norm info to the segment attribute, but it turned out to be dead metadata — the norm read path doesn't actually use it (it relies on FieldInfos instead). So I went ahead and removed it. What do you think?

src/main/java/org/opensearch/knn/common/KNNVectorUtil.java
krocky-cooky: fixed.

src/main/java/org/opensearch/knn/index/codec/KNN10010Codec/KNN10010DerivedSourceStoredFieldsWriter.java
krocky-cooky: added.

src/main/java/org/opensearch/knn/common/KNNVectorUtil.java
krocky-cooky: fixed.

src/main/java/org/opensearch/knn/index/codec/derivedsource/NestedPerFieldDerivedVectorTransformer.java
krocky-cooky: added offset incrementation.

src/main/java/org/opensearch/knn/index/mapper/NormalizeVectorTransformer.java
krocky-cooky: I implemented transformAndGetNorm which performs l2 normalization and returns the norm in a single pass. The implementation — computing the norm from dotProduct then dividing each element — follows the same approach as VectorUtil.l2normalize, so there's no additional overhead.

src/main/java/org/opensearch/knn/common/KNNConstants.java
krocky-cooky: Thanks for the important callout. Unlike _source, this won't throw a duplicate mapping exception, so we need to fully prevent the collision. It seems like we'd need to add an explicit rejection on the OpenSearch core side — can you think of any other approaches?

src/main/java/org/opensearch/knn/index/codec/derivedsource/DerivedSourceNormSupplier.java
krocky-cooky: During merges, since [shouldInject is set to false](https://github.com/opensearch-project/k-NN/blob/4cf03358dfd9f6f85f29fbe133b71ece99ed5c5b/src/main/java/org/opensearch/knn/index/codec/KNN10010Codec/KNN10010DerivedSourceStoredFieldsReader.java#L135) and doc values are not read, the overhead from this processing may be somewhat mitigated.

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java
```diff
@@ -698,12 +700,15 @@ private Field createVectorField(byte[] vectorValue, boolean isDerivedEnabled) {
...
+            fields.add(new FloatDocValuesField(KNNVectorUtil.getNormFieldName(name()), norm));
+        }
```
navneet1v, line 711: > Store the pre-normalization L2 norm as a FloatDocValuesField (_knn_norm_<field_name>) alongside the vector. On _source reconstruction, read the norm and multiply it back to recover the original vector.

question on the choice of creating a new field here, why we are thinking to create a separate field, rather than just using the same field and at the codec layer store the norm in a separate file, which will be more cleaner in terms of keeping the information within 1 field. 

Can you share any reasoning/evaluation we have done to know which approach is better?

src/main/java/org/opensearch/knn/index/codec/derivedsource/DerivedSourceNormSupplier.java
navneet1v: > Every call to getNorm(docId) creates a fresh NumericDocValues iterator. 

Why we are doing this? Even for search this is just extra latency we are adding. Why we are not going with the behavior where per search request/merge we just have 1 Norms producer.

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java
navneet1v: @krocky-cooky

navneet1v (changes requested):

src/main/java/org/opensearch/knn/common/KNNVectorUtil.java
```diff
@@ -60,4 +61,34 @@ public static int[] intListToArray(final List<Integer> integerList) {
...
+        }
+        float[] result = inplace ? vector : Arrays.copyOf(vector, vector.length);
+        for (int i = 0; i < result.length; i++) {
+            result[i] *= norm;
+        }
+        return result;
```
navneet1v, line 92: Please use SIMD here rather than just plain for loop, otherwise the latency will be high.

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java
krocky-cooky: @navneet1v 
When persisting the norm into the same DocValues as the vector, the merge logic that the codec layer would have to implement looked significantly complex. I considered the following two options:

1. Writing norms to a separate file.
2. Appending the norm to the tail of the vector (`[f0 … f_{d-1} | norm]`).

Option (2) looked realistic because it doesn't require any extra file read overhead.

However, I landed on a different approach: deferring vector normalization to refresh time, so that the (unnormalized) original vectors end up in DocValues as-is. I believe this resolves the various problems the earlier direction was running into. (#3294 )

PR #3217, opened 2026-03-27
[Bug-fix] Ensure Prefetch setting doesn't trigger native cache rebuild

navneet1v:
### Description
Ensure Prefetch setting doesn't trigger native cache rebuild. 

The problem was all the FF was going into cache rebuild consumer. I have separated them out to ensure that we have different function to get settings which rebuilds cache.

### Related Issues
NA

### Check List

### DIFF
--- a/jni/cmake/init-faiss.cmake
+++ b/jni/cmake/init-faiss.cmake
@@ -25,6 +25,7 @@ if(NOT DEFINED APPLY_LIB_PATCHES OR "${APPLY_LIB_PATCHES}" STREQUAL true)
     list(APPEND PATCH_FILE_LIST "${CMAKE_CURRENT_SOURCE_DIR}/patches/faiss/0007-Modify-minor-Faiss-structures-to-support-Faiss-SQ.patch")
     list(APPEND PATCH_FILE_LIST "${CMAKE_CURRENT_SOURCE_DIR}/patches/faiss/0008-Added-skipping-flat-storage-availability-in-write_in.patch")
     list(APPEND PATCH_FILE_LIST "${CMAKE_CURRENT_SOURCE_DIR}/patches/faiss/0009-Fixing-radial-search-for-IndexHNSWCagra.patch")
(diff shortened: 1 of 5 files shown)


src/main/java/org/opensearch/knn/common/featureflags/KNNFeatureFlags.java
```diff
@@ -44,10 +44,22 @@ public class KNNFeatureFlags {
...
     public static List<Setting<?>> getFeatureFlags() {
         return ImmutableList.of(KNN_FORCE_EVICT_CACHE_ENABLED_SETTING, KNN_PREFETCH_ENABLED_SETTING);
     }
 
+    /**
+     * All feature flags which when changed should trigger a cache rebuild
+     * @return List of Feature flag settings
+     */
+    public static List<Setting<?>> getFeatureFlagsWhichRebuildsCache() {
+        return ImmutableList.of(KNN_FORCE_EVICT_CACHE_ENABLED_SETTING);
```
naveentatikonda, line 60: Now, we need to add the flag in 2 places if it is related to cache rebuild which is easy to miss. Instead shall we maintain a single source of truth something like this. WDYT ?

```
// Key: the setting, Value: true if changing this flag should trigger a native cache rebuild.
private static final Map<Setting<?>, Boolean> FEATURE_FLAG_MAP = Map.of(
    KNN_FORCE_EVICT_CACHE_ENABLED_SETTING, true,
    KNN_PREFETCH_ENABLED_SETTING, false
);

public static List<Setting<?>> getFeatureFlags() {
    return List.copyOf(FEATURE_FLAG_MAP.keySet());
}

public static List<Setting<?>> getFeatureFlagsWhichRebuildsCache() {
    return FEATURE_FLAG_MAP.entrySet()
        .stream()
        .filter(Map.Entry::getValue)
        .map(Map.Entry::getKey)
        .collect(Collectors.toUnmodifiableList());
}

navneet1v: I don't think its a good idea as of now. If we have more settings then we should just break the class into two parts and move all setting related to cache at one place. I think its okay for now.

PR #3240, opened 2026-03-31
[Bug Fix] Fix for prefetch failure due to out of bound exception 

VijayanB:
### Description
`FaissScorableByteVectorValues` was wrapping `ByteVectorValues` delegates but not properly exposing the `HasIndexSlice` interface when the underlying delegate supported it. This prevented Lucene's prefetch optimization from accessing the underlying `IndexInput` slice, potentially impacting search performance. It also causes out of range exception since the byteVectorLength is not delegated

### Related Issues
#2577 

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissScorableByteVectorValues.java
+++ b/src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissScorableByteVectorValues.java
@@ -7,6 +7,7 @@ package org.opensearch.knn.memoryoptsearch.faiss;
 import org.apache.lucene.codecs.hnsw.FlatVectorsScorer;
 import org.apache.lucene.codecs.lucene95.HasIndexSlice;
 import org.apache.lucene.index.ByteVectorValues;
+import org.apache.lucene.index.VectorEncoding;
 import org.apache.lucene.index.VectorSimilarityFunction;
 import org.apache.lucene.search.DocIdSetIterator;
 import org.apache.lucene.search.VectorScorer;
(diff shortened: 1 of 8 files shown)


src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissScorableByteVectorValues.java
```diff
@@ -123,4 +124,16 @@ public Bulk bulk(final DocIdSetIterator matchingDocs) {
...
+     * Returns the underlying {@link IndexInput} slice if the delegate supports
+     * {@link HasIndexSlice}, otherwise returns {@code null}.
+     */
+    @Override
+    public IndexInput getSlice() {
+        if (delegate instanceof HasIndexSlice supportsIndexSlice) {
+            return supportsIndexSlice.getSlice();
+        }
+        return null;
+    }
```
navneet1v, line 154: We should do something like this:

https://github.com/opensearch-project/k-NN/blob/ad5684af86921b6d2d03cb0367724d3b35588271/src/main/java/org/opensearch/knn/memoryoptsearch/faiss/MMapFloatVectorValues.java#L81-L87

instead of returning null.

VijayanB: As discussed changed the [MMapFloatVectorValues.java](https://github.com/opensearch-project/k-NN/blob/ad5684af86921b6d2d03cb0367724d3b35588271/src/main/java/org/opensearch/knn/memoryoptsearch/faiss/MMapFloatVectorValues.java#L81-L87) to return null to keep it consistent with method documentation

navneet1v (approved):

Vikasht34 (approved):

PR #3257, opened 2026-04-03
[Bug Fix] Block radial search on Faiss SQ

0ctopus13prime:
### Description
Radial search should not allowed on Faiss SQ.
This PR is explicitly blocking radial search on Faiss SQ.

### Related Issues
N/A

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
+++ b/src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
@@ -35,6 +35,7 @@ import org.opensearch.knn.index.engine.MemoryOptimizedSearchSupportSpec;
 import org.opensearch.knn.index.engine.MethodComponentContext;
 import org.opensearch.knn.index.engine.model.QueryContext;
 import org.opensearch.knn.index.engine.qframe.QuantizationConfig;
+import org.opensearch.knn.index.mapper.CompressionLevel;
 import org.opensearch.knn.index.mapper.KNNMappingConfig;
 import org.opensearch.knn.index.mapper.KNNVectorFieldType;
 import org.opensearch.knn.index.query.parser.KNNQueryBuilderParser;
@@ -59,6 +60,7 @@ import static org.opensearch.knn.common.KNNConstants.METHOD_PARAMETER_NPROBES;
 import static org.opensearch.knn.common.KNNConstants.MIN_SCORE;
 import static org.opensearch.knn.common.KNNValidationUtil.validateByteVectorValue;
 import static org.opensearch.knn.index.engine.KNNEngine.ENGINES_SUPPORTING_RADIAL_SEARCH;
+import static org.opensearch.knn.index.engine.KNNEngine.FAISS;
 import static org.opensearch.knn.index.engine.validation.ParameterValidator.validateParameters;
 import static org.opensearch.knn.index.query.parser.MethodParametersParser.validateMethodParameters;
 import static org.opensearch.knn.index.query.parser.RescoreParser.RESCORE_OVERSAMPLE_PARAMETER;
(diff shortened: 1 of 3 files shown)


src/test/java/org/opensearch/knn/memoryoptsearch/FaissSQRadialSearchBlockedIT.java
```diff
@@ -0,0 +1,142 @@
...
+
+import java.util.concurrent.ThreadLocalRandom;
+
+import static org.opensearch.knn.index.KNNSettings.KNN_INDEX;
+
+/**
+ * Integration test to verify that radial search (max_distance / min_score) is blocked
+ * for Faiss SQ (scalar quantized, 32x compression) indices.
+ */
+public class FaissSQRadialSearchBlockedIT extends KNNRestTestCase {
```
navneet1v, line 25: why we are not adding the ITs in radial search IT class?

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
```diff
@@ -486,7 +488,10 @@ protected Query doToQuery(QueryShardContext context) {
                 throw new UnsupportedOperationException(String.format(Locale.ROOT, "Binary data type does not support radial search"));
             }
 
-            if (knnMappingConfig.getQuantizationConfig() != QuantizationConfig.EMPTY) {
+            if ((knnMappingConfig.getQuantizationConfig() != QuantizationConfig.EMPTY)
+                // If engine is Faiss and compression level is 32x, then it's using Faiss SQ which has null QuantizationConfig.
+                // In that case, we should block radial search as well.
+                || (knnEngine == FAISS && knnMappingConfig.getCompressionLevel() == CompressionLevel.x32)) {
```
navneet1v, line 494: does this mean that Lucene Radial Search is supported with 1 bit?

src/test/java/org/opensearch/knn/memoryoptsearch/FaissSQRadialSearchBlockedIT.java
0ctopus13prime: Because, I've separated this out for us to easily remove the class when we need to support.
Then we can just remove the file

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
0ctopus13prime: Hm... not sure, we probably need to bring Jack in here?
This PR is to block radial search on Faiss SQ

src/test/java/org/opensearch/knn/memoryoptsearch/FaissSQRadialSearchBlockedIT.java
navneet1v: Too much just for a specific thing. I don't know why we are doing such things.

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
naveentatikonda: I think we need to block for Lucene too

navneet1v: Let me be specific why we are just blocking Faiss SQ for radial search. If we merge this PR this will lead to split in behavior where Faiss engine with 1 bit doesn't support Radial Search and Lucene does. Since we are close to the release. Lets target 1 PR.

Vikasht34: Yeah this makes sense

navneet1v (approved):

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
navneet1v: @jack-hung-lgtm can we have a followup PR for Lucene?

naveentatikonda (approved):

PR #3260, opened 2026-04-03
Support derived source for knn with other fields

### LINKED ISSUE #3058: [FEATURE] Converge OpenSearch core and k-NN Derived Source feature

neetikasinghal:
### Description
Implement derived source for k-NN fields with core implementation, with a fallback to k-NN plugin's implementation. 

Related PR: https://github.com/opensearch-project/OpenSearch/pull/20991

Before this change, if `index.derived_source.enabled` and there is a knn field type, there will be error thrown saying that knn vector field doesnt supprt derived source feature.
After this change, if `index.derived_source.enabled` and there is a knn field type, the derived source feature will be turned on so that users can benefit out of both knn fields and other fields.

### Related Issues
Resolves #3058

### Check List

### DIFF
--- a/qa/restart-upgrade/src/test/java/org/opensearch/knn/bwc/DerivedSourceBWCRestartIT.java
+++ b/qa/restart-upgrade/src/test/java/org/opensearch/knn/bwc/DerivedSourceBWCRestartIT.java
@@ -92,6 +92,30 @@ public class DerivedSourceBWCRestartIT extends DerivedSourceTestCase {
         }
     }
 
+    public void testDerivedEnabledSettingPreservedOnUpgrade() throws IOException {
+        waitForClusterHealthGreen(NODES_BWC_CLUSTER);
+        String indexName = getIndexName("knn-bwc", "derived-enabled-", false);
+        if (isRunningAgainstOldCluster()) {
+            String fieldName = "test";
+            int dimension = 16;
+            XContentBuilder builder = XContentFactory.jsonBuilder()
+                .startObject()
+                .startObject("properties")
+                .startObject(fieldName)
+                .field("type", "knn_vector")
+                .field("dimension", dimension)
+                .endObject()
+                .endObject()
+                .endObject();
+            String mapping = builder.toString();
+            Settings settings = Settings.builder().put("index.knn", true).put("index.knn.derived_source.enabled", true).build();
+            createKnnIndex(indexName, settings, mapping);
+            validateDerivedSetting(indexName, true);
+        } else {
+            validateDerivedSetting(indexName, true);
+        }
+    }
+
     @Override
     protected final boolean preserveIndicesUponCompletion() {
         return true;
--- a/src/main/java/org/opensearch/knn/index/mapper/KnnVectorValuesFetcher.java
+++ b/src/main/java/org/opensearch/knn/index/mapper/KnnVectorValuesFetcher.java
@@ -1,0 +1,68 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch.knn.index.mapper;
+
+import org.apache.lucene.index.FieldInfo;
+import org.apache.lucene.index.LeafReader;
+import org.apache.lucene.util.BytesRef;
+import org.opensearch.common.lucene.Lucene;
+import org.opensearch.index.mapper.FieldValueFetcher;
+import org.opensearch.knn.index.vectorvalues.KNNVectorValues;
+import org.opensearch.knn.index.vectorvalues.KNNVectorValuesFactory;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * FieldValueFetcher that retrieves KNN vector values from doc values or native KNN vector format.
+ *
+ */
+public class KnnVectorValuesFetcher extends FieldValueFetcher {
+    KNNVectorFieldType mappedFieldType;
+
+    public KnnVectorValuesFetcher(KNNVectorFieldType mappedFieldType, String simpleName) {
+        super(simpleName);
+        this.mappedFieldType = mappedFieldType;
+    }
+
+    @Override
+    public List<Object> fetch(LeafReader reader, int docId) throws IOException {
+        List<Object> values = new ArrayList<>(1);
+        try {
+            FieldInfo fieldInfo = reader.getFieldInfos().fieldInfo(mappedFieldType.name());
+            KNNVectorValues<?> vectorValues = fieldInfo != null
+                ? KNNVectorValuesFactory.getVectorValues(fieldInfo, Lucene.segmentReader(reader))
+                : null;
+            if (vectorValues == null) {
+                return values;
+            }
+            if (vectorValues.advance(docId) == docId) {
+                values.add(vectorValues.getVector());
+            }
+        } catch (Exception e) {
+            throw new IOException("Failed to read vector values for document " + docId + " in field " + mappedFieldType.name(), e);
+        }
+        return values;
+    }
+
+    /**
+     * Converts byte[] vectors into int[] so that XContentBuilder serializes them as numeric JSON
+     * arrays instead of Base64 binary data. float[] needs no conversion — XContentBuilder natively
+     * handles float[] via field(String, float[]).
+     * Uses same approach as {@link org.opensearch.knn.index.codec.derivedsource.DerivedSourceIndexOperationListener#formatVector}.
+     */
+    @Override
+    public Object convert(Object value) {
+        if (value instanceof byte[] vector) {
+            return KNNVectorFieldMapperUtil.deserializeStoredVector(new BytesRef(vector), mappedFieldType.getVectorDataType());
+        }
+        return value;
+    }
+}
--- a/src/main/java/org/opensearch/knn/index/util/IndexUtil.java
+++ b/src/main/java/org/opensearch/knn/index/util/IndexUtil.java
@@ -461,12 +461,27 @@ public class IndexUtil {
             .equals(VectorDataType.BYTE.getValue());
     }
 
+    /**
+    * Checks whether the k-NN plugin's derived source feature is enabled for the given index.
+    * Returns {@code false} when core's {@code index.derived_source.enabled} is on (core takes precedence),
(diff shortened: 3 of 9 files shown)


neetikasinghal:
Will add change log once all the reviews are completed

src/main/java/org/opensearch/knn/index/util/IndexUtil.java
```diff
@@ -470,6 +470,11 @@ public static boolean isDerivedEnabledForIndex(MapperService mapperService) {
             return false;
         }
 
+        // if core based setting is turned on, then it takes precedence
```
VijayanB, line 482: can we add java doc for this method?

src/main/java/org/opensearch/knn/index/mapper/KnnVectorValuesFetcher.java
```diff
@@ -0,0 +1,77 @@
...
+    /**
+     * Converts the raw vector (float[] or byte[]) into a List of Numbers so that
+     * XContentBuilder serializes all vector types as numeric JSON arrays.
+     * Without this, byte[] would be serialized as Base64 binary data.
+     */
+
+    @Override
+    public Object convert(Object value) {
+        if (value instanceof float[]) {
+            float[] vector = (float[]) value;
```
VijayanB: can be simplified using pattern matching

src/main/java/org/opensearch/knn/index/mapper/KnnVectorValuesFetcher.java
```diff
@@ -0,0 +1,77 @@
...
+            List<Number> result = new ArrayList<>(vector.length);
+            for (float v : vector) {
+                result.add(v);
+            }
+            return result;
+        } else if (value instanceof byte[]) {
```
VijayanB:
```suggestion
        } 
        if (value instanceof byte[]) {
```

src/main/java/org/opensearch/knn/index/mapper/KnnVectorValuesFetcher.java
```diff
@@ -0,0 +1,77 @@
...
+            FieldInfo fieldInfo = reader.getFieldInfos().fieldInfo(mappedFieldType.name());
+            KNNVectorValues<?> vectorValues = fieldInfo != null ? KNNVectorValuesFactory.getVectorValues(fieldInfo, reader) : null;
+            if (vectorValues == null) {
+                return values;
+            }
+            if (vectorValues.advance(docId) == docId) {
+                values.add(vectorValues.getVector());
+            }
+        } catch (Exception e) {
+            throw new IOException("Failed to read doc values for document " + docId + " in field " + mappedFieldType.name(), e);
```
navneet1v:
```suggestion
            throw new IOException("Failed to vector values for document " + docId + " in field " + mappedFieldType.name(), e);
```

src/main/java/org/opensearch/knn/index/mapper/KnnVectorValuesFetcher.java
```diff
@@ -0,0 +1,77 @@
...
+            }
+            return result;
+        } else if (value instanceof byte[]) {
+            byte[] vector = (byte[]) value;
+            List<Number> result = new ArrayList<>(vector.length);
+            for (byte v : vector) {
+                result.add((int) v);
+            }
+            return result;
+        }
```
navneet1v: this is going to be expensive. Can we see how we are doing this in our current derived source code.

navneet1v:
@neetikasinghal few questions:
1. Have we tested the code when include and exclude for a vector field is present(mappings or search) then KnnVectorValuesFetcher is not getting triggered?

Vikasht34:
@neetikasinghal I see old PR closed , how did we addressed this https://github.com/opensearch-project/k-NN/pull/3209#discussion_r3011835246

src/main/java/org/opensearch/knn/index/util/IndexUtil.java
neetikasinghal: sure

neetikasinghal:
> @neetikasinghal I see old PR closed , how did we addressed this [#3209 (comment)](https://github.com/opensearch-project/k-NN/pull/3209#discussion_r3011835246)

@Vikasht34 https://github.com/opensearch-project/k-NN/pull/3209#discussion_r3012017846 - please refer to my comment here and let me know what you think about it

src/main/java/org/opensearch/knn/index/mapper/KnnVectorValuesFetcher.java
neetikasinghal: updated

neetikasinghal:
> @neetikasinghal few questions:
> 
> 1. Have we tested the code when include and exclude for a vector field is present(mappings or search) then KnnVectorValuesFetcher is not getting triggered?

yes

navneet1v:
> > @neetikasinghal few questions:
> > 
> > 1. Have we tested the code when include and exclude for a vector field is present(mappings or search) then KnnVectorValuesFetcher is not getting triggered?
> 
> yes

@neetikasinghal are there ITs related or some way to confirm this or verify? I may be missing something here.

neetikasinghal:
> > > @neetikasinghal few questions:
> > > 
> > > 1. Have we tested the code when include and exclude for a vector field is present(mappings or search) then KnnVectorValuesFetcher is not getting triggered?
> > 
> > 
> > yes
> 
> @neetikasinghal are there ITs related or some way to confirm this or verify? I may be missing something here.

actually, I double checked this and this looks to be an issue in the core side, where the includes/excludes are not getting honored. I will create an issue there and see if can find a way to fix it. This should be handled at the core side and not something especially for knn plugin, what do u think?

navneet1v:
> > > > @neetikasinghal few questions:
> > > > 
> > > > 1. Have we tested the code when include and exclude for a vector field is present(mappings or search) then KnnVectorValuesFetcher is not getting triggered?
> > > 
> > > 
> > > yes
> > 
> > 
> > @neetikasinghal are there ITs related or some way to confirm this or verify? I may be missing something here.
> 
> actually, I double checked this and this looks to be an issue in the core side, where the includes/excludes are not getting honored. I will create an issue there and see if can find a way to fix it. This should be handled at the core side and not something especially for knn plugin, what do u think?

Yes it make sense that core handles it. But it will mean that we cannot merge this PR till that issue is fixed. WDYT?

neetikasinghal:
> > > > > @neetikasinghal few questions:
> > > > > 
> > > > > 1. Have we tested the code when include and exclude for a vector field is present(mappings or search) then KnnVectorValuesFetcher is not getting triggered?
> > > > 
> > > > 
> > > > yes
> > > 
> > > 
> > > @neetikasinghal are there ITs related or some way to confirm this or verify? I may be missing something here.
> > 
> > 
> > actually, I double checked this and this looks to be an issue in the core side, where the includes/excludes are not getting honored. I will create an issue there and see if can find a way to fix it. This should be handled at the core side and not something especially for knn plugin, what do u think?
> 
> Yes it make sense that core handles it. But it will mean that we cannot merge this PR till that issue is fixed. WDYT?

I think, we can merge this PR and a separate PR can take care of this issue. It will fix include/exclude issue with text fields as well. Can do a quick follow-up on that. This issue is not specific to the knn plugin, so we can push this PR, imo. 
Are you on-board with that?

https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/IndexSettings.java#L908 - By default, the derived source index setting is false.

neetikasinghal:
The CIs are failing as my latest change in OpenSearch: https://github.com/opensearch-project/OpenSearch/pull/21005 is not taken in the opensearch core snapshot jar, I see the tests are passing with the latest jar on my local.

@peterzhuamazon do u know how can i fix this issue?

peterzhuamazon:
Verified with Neetika and we confirmed that the latest jar having the change.
Given linux is passing it is possible this is either due to mac/windows cleanup issue or something else: https://github.com/opensearch-project/k-NN/actions/runs/24163591704/job/70714635405?pr=3260

Re-run to verify again to see if it passes. Thanks.

navneet1v:
Please fix the changelog CI by adding an entry in changelog

src/main/java/org/opensearch/knn/index/util/IndexUtil.java
```diff
@@ -470,6 +479,11 @@ public static boolean isDerivedEnabledForIndex(MapperService mapperService) {
...
+        if (mapperService.getIndexSettings().isDerivedSourceEnabled()) {
+            return false;
+        }
```
navneet1v, line 486: should we add a version check here such that if an old index is migrated to new cluster this check might say that KNN derived source is disabled.

neetikasinghal: makes sense, sure will do

navneet1v: This brings me to a point we need to BWC tests here. So that we can catch these issues.

neetikasinghal: added

qa/restart-upgrade/src/test/java/org/opensearch/knn/bwc/DerivedSourceBWCRestartIT.java
```diff
@@ -92,6 +92,30 @@ public void testOldSettingPreservedOnUpgrade() throws IOException {
         }
     }
 
+    public void testDerivedEnabledSettingPreservedOnUpgrade() throws IOException {
```
navneet1v, line 95: This is just checking the setting is preserved or not. how do we validate that merges and searches are working as expected?

neetikasinghal: there are already existing tests in the same test class that validate this behavior. If knn derived source setting is there, then we validate if the merging and searches work fine.

buddharajusahil (approved):

navneet1v (approved):

0ctopus13prime (approved):

PR #3269, opened 2026-04-10
Fix updating circuit breaker setting

### LINKED ISSUE #3205: [BUG] Exception while updating circuit breaker setting
**What is the bug?**
We are making KNN search requests with a user "test_user" which doesnt have admin permissions. When the KNN [cache eviction is triggered](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/memory/NativeMemoryCacheManager.java#L531), we get an error:
```
 Exception while updating circuit breaker setting {"knn.circuit_breaker.triggered":"true"} to no permissions for [cluster:admin/settings/update] and User [name=test_user, backend_roles=[index_management_full_access, CustomRole, readall], requestedTenant=null]
```

It seems that KNN Plugin tries to set knn.circuit_breaker.trigerred=true using the user of the current session - https://github.com/opensearch-project/k-NN/blob/3.5.0.0/src/main/java/org/opensearch/knn/index/KNNSettings.java#L1015
But it requires admin privileges to perform a search, which the user doesn't have.

1. Create a user without permissions for the cluster:admin/settings/update.
2. Create a knn index.
3. Overload the cluster with requests to trigger the circuit breaker.

The circuit breaker was triggered without any errors.

 - OS - linux
 - Version- 3.2.0
 - k-NN plugin, Faiss engine

HUSTERGS:
### Description
Stash thread context first before update settings to solve the permissions problem, just like what `ModelDao` does with `runWithStashedThreadContext`.

### Related Issues
Resolves #3205

### Check List


navneet1v:
@HUSTERGS can we add UTs and ITs for this change?

navneet1v:
Also add an entry in change log. Rest of the code looks good to me. Will wait for CIs to get completed

HUSTERGS:
> @HUSTERGS can we add UTs and ITs for this change?

@navneet1v Hi, thanks for the suggestion. I did look into adding UT/IT for this change. The challenge is that the core logic being modified here only takes effect when the security plugin is installed, and I haven't found a clean way to simulate that in UT/IT without pulling in the security plugin as a dependency.
A couple of options I considered:
- Mocking the ThreadContext.stashContext() call — but it doesn't fully replicate the security plugin's behavior.
- Adding an integration test with the security plugin enabled — this would work but may be out of scope for this PR.

Would either of these approaches work for you, or do you have a preferred way to test this kind of scenario?

navneet1v:
> > @HUSTERGS can we add UTs and ITs for this change?
> 
> @navneet1v Hi, thanks for the suggestion. I did look into adding UT/IT for this change. The challenge is that the core logic being modified here only takes effect when the security plugin is installed, and I haven't found a clean way to simulate that in UT/IT without pulling in the security plugin as a dependency. A couple of options I considered:
> 
> * Mocking the ThreadContext.stashContext() call — but it doesn't fully replicate the security plugin's behavior.
> * Adding an integration test with the security plugin enabled — this would work but may be out of scope for this PR.
> 
> Would either of these approaches work for you, or do you have a preferred way to test this kind of scenario?

Hi @HUSTERGS 

There is CI where we create a security cluster and run ITs against it. So what you can do is add an IT which checks if security is enabled then run the IT, else just exit the IT with success. 

On the mocking part, it would still good to have it. We can use that as a UT. So my recommendation will be do have both IT and UT

navneet1v:
@HUSTERGS please add the UTs and ITs as suggested above. Once that is done we can merge this PR.

HUSTERGS:
> @HUSTERGS please add the UTs and ITs as suggested above. Once that is done we can merge this PR.

Sorry, have been busy these days. I tried to add some UT/IT according to your suggestion. : )

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 * Update Gradle wrapper to 9.4.1 and Jacoco to 0.8.14 to match core OpenSearch [#3308](https://github.com/opensearch-project/k-NN/pull/3308)
 
 ### Bug Fixes
+* Fix filter fields with derived source enabled [#21209](https://github.com/opensearch-project/OpenSearch/pull/21209)
 * Use KNN1040ScalarQuantizedVectorsFormat for Faiss SQ flat format to enable prefetch [#3302](https://github.com/opensearch-project/k-NN/pull/3302)
```
navneet1v: @HUSTERGS can we fix the change log entry.

navneet1v (dismissed):
overall looks good. Lets just fix the changelog entry.

CHANGELOG.md
HUSTERGS: @navneet1v I've fixed the change log entry, thanks for the reminder : )

navneet1v:
> ## PR Code Suggestions ✨
> Explore these optional code suggestions:
> 
> **Category**	**Suggestion                                                                                                                                    **	**Impact**
> Possible issue	
> Restore context in async callbacks
> **The `StoredContext` is closed immediately after initiating the async `updateSettings` call, which may cause the stashed context to be restored before the async callbacks execute. This defeats the purpose of stashing the context. Consider restoring the context inside the callbacks instead, or use a different approach to ensure the callbacks execute with the stashed context.**
> 
> [src/main/java/org/opensearch/knn/index/KNNSettings.java [1023-1043]](https://github.com/opensearch-project/k-NN/pull/3269/files#diff-d5240888e4ed8c6fb277d8b9b9fc1680f632535629ededfe04311303777f04baR1023-R1043)
> 
> ```diff
> -try (ThreadContext.StoredContext context = client.threadPool().getThreadContext().stashContext()) {
> -    client.admin().cluster().updateSettings(clusterUpdateSettingsRequest, new ActionListener<ClusterUpdateSettingsResponse>() {
> -        @Override
> -        public void onResponse(ClusterUpdateSettingsResponse clusterUpdateSettingsResponse) {
> +ThreadContext.StoredContext context = client.threadPool().getThreadContext().stashContext();
> +client.admin().cluster().updateSettings(clusterUpdateSettingsRequest, new ActionListener<ClusterUpdateSettingsResponse>() {
> +    @Override
> +    public void onResponse(ClusterUpdateSettingsResponse clusterUpdateSettingsResponse) {
> +        try {
>              logger.debug(
>                  "Cluster setting {}, acknowledged: {} ",
>                  clusterUpdateSettingsRequest.persistentSettings(),
>                  clusterUpdateSettingsResponse.isAcknowledged()
>              );
> +        } finally {
> +            context.restore();
>          }
> +    }
>  
> -        @Override
> -        public void onFailure(Exception e) {
> +    @Override
> +    public void onFailure(Exception e) {
> +        try {
>              logger.info(
>                  "Exception while updating circuit breaker setting {} to {}",
>                  clusterUpdateSettingsRequest.persistentSettings(),
>                  e.getMessage()
>              );
> +        } finally {
> +            context.restore();
>          }
> -    });
> -}
> +    }
> +});
> ```
> 
> Suggestion importance[1-10]: 9
> __
> 
> Why: This is a critical issue. The `try-with-resources` block closes the `StoredContext` immediately after initiating the asynchronous `updateSettings` call, which restores the original context before the async callbacks execute. This defeats the purpose of stashing the context and could cause the callbacks to run with incorrect thread context state. The suggestion correctly identifies this timing issue and proposes restoring the context within the callbacks themselves.
> 
> High

@HUSTERGS can you check this. Looks to me a valid concern

HUSTERGS:
> @HUSTERGS can you check this. Looks to me a valid concern

@navneet1v I checked this concern. I do not think we should move the `StoredContext` restore into the async callbacks here.

The purpose of the stash in this PR is to clear the request/user security context while submitting the internal `cluster:admin/settings/update` request. The security-sensitive part is the execution/authorization of the update action itself, which happens when `client.admin().cluster().updateSettings(...)` is invoked. The callbacks in this method only log the response or failure and do not perform any privileged follow-up work.

Holding the `StoredContext` until an async callback would also be risky: the callback may run on a different thread, so restoring the caller's stored context there can overwrite that callback thread's current context, while leaving the original caller thread in the stashed context longer than intended. The scoped try-with-resources is the same pattern used by `ModelDao.runWithStashedThreadContext` around async client calls.

The added UT verifies the update request is submitted with the user/auth context removed and then restores the original context afterwards. The added security IT covers the original issue by validating that a limited user can trip the circuit breaker without requiring `cluster:admin/settings/update` permission. So I believe the current scope is intentional and should be kept.

navneet1v (dismissed):

HUSTERGS:
@navneet1v hi, I noticed this PR was blocked by Codecov requirement, so I've added some test, hope the Codecov can pass : )

navneet1v (approved):

PR #3270, opened 2026-04-10
Added support for 1 bit SQ with remote build

### LINKED ISSUE #3274: [FEATURE] Support 1 bit SQ with GPU remote build feature
k-NN plugin has switched to 1 bit SQ for 32x compression. We need to add support for the GPU remote build feature with this new compression scheme. Currently, GPU remote build only support BQ for quantization. 

Since there is currently no support in `cuvs` (NVIDIA library used by GPU remote builder) for 1 bit SQ, we can instead build the full precision FP32 graph on the remote index builder, and stitch together the graph with the 1 bit SQ flat vector storage on the k-NN data nodes at search time. 

- When 32x compression is enabled, send the full precision FP32 vectors to remote builder
- Remote Index Builder builds the CAGRA graph with FP32 vectors. 
- Remote Index Builder transfers only the graph structure back to k-NN
- At search time, k-NN attaches the 1 bit SQ flat storage to the FP32 CAGRA graph

Note that the remote index builder already supports transferring the FP32 graph only: https://github.com/opensearch-project/remote-vector-index-builder/pull/127 with the `skip_stored_vectors` parameter

Also note that this will increase latency on vector upload and index build time, since FP32 vectors take up more space than BQ vectors. Will add benchmarks for FP32 v.s. BQ for the remote index builder. Will also do a 1M OSB benchmark for sanity testing recall, with a single node OS cluster and single GPU worker. 

Instead of building an FP32 CAGRA graph on the remote index builder, we could build an FP16 graph instead to save on memory and indexing costs. However, this will be tricky since we would need to do a range check on the vectors somewhere - if the vectors are out of FP16 range, then we would need to fallback to FP32. This could be a future optimization. 

rchitale7:
### Description
This enables support for 1 bit SQ with remote build. For this change, we transfer the FP32 vectors to the GPU worker, the GPU worker sends back the FP32 graph, and then the k-NN data node stitches together the FP32 graph with the 1 bit SQ vectors. The GPU worker now supports only transferring the FP32 graph back without the flat vectors with the new `skip_stored_vectors` parameter. See this PR for the GPU worker change: https://github.com/opensearch-project/remote-vector-index-builder/pull/127 

### Testing

- Unit tests
- Manual test:
    - Ran a single node OS cluster with `./gradlew run` and the `opensearchstaging/remote-vector-index-builder:api-snapshot` GPU worker docker image. 
    - Created a 32x compression index, with remote index build enabled. Indexed some documents and then ran search, validated search worked as expected: 
```
curl -XPUT "localhost:9200/_snapshot/my-repo" -H 'Content-Type: application/json' -d '{
  "type": "s3",
  "settings": {
    "bucket": "testbucket-rchital",
    "base_path": "test-data/more-data",
    "region": "us-west-2"
  }
}'
```

```
curl -XPUT "localhost:9200/_cluster/settings" -H 'Content-Type: application/json' -d '{
  "persistent": {
    "knn.remote_index_build.enabled": true,
    "knn.remote_index_build.repository": "my-repo",
    "knn.remote_index_build.service.endpoint": "http://0.0.0.0:80"
  }
}'
```

```
curl -XPUT "localhost:9200/sq-test" -H 'Content-Type: application/json' -d '{
  "settings": {
    "index.knn": true,
    "index.knn.remote_index_build.enabled": true,
    "index.knn.remote_index_build.size.min": "0b",
    "number_of_shards": 1,
    "number_of_replicas": 0
  },
  "mappings": {
    "properties": {
      "vec": {
        "type": "knn_vector",
        "dimension": 8,
        "mode": "on_disk",
        "compression_level": "32x",
        "method": { "name": "hnsw", "engine": "faiss", "space_type": "l2" }
      }
    }
  }
}'
```

```
for i in $(seq 1 6); do
  curl -XPOST "localhost:9200/sq-test/_doc" -H 'Content-Type: application/json' \
    -d "{\"vec\": [$i.0, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0, 8.0]}"
done
```

```
curl -XPOST "localhost:9200/sq-test/_search" -H 'Content-Type: application/json' -d '{
  "query": { "knn": { "vec": { "vector": [1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0, 8.0], "k": 2 } } }
}'
```

- Integration tests: 
    - Added `ExpectRemoteBuildValidation` to MOSFaissSQIndexIT tests. 
- 1M OSB benchmark with faiss cohere
    - Setup is explained here: https://github.com/opensearch-project/k-NN/issues/3274#issuecomment-4238881163
    - Validated the recall was similar to 1 bit SQ non remote build. 

### Related Issues
Resolves #3274

### Check List
- [ ] New functionality has been documented - N/A
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md) - N/A

### DIFF
--- a/src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss1040ScalarQuantizedKnnVectorsFormat.java
+++ b/src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss1040ScalarQuantizedKnnVectorsFormat.java
@@ -13,6 +13,7 @@ import org.apache.lucene.codecs.lucene104.Lucene104ScalarQuantizedVectorsFormat;
 import org.apache.lucene.codecs.lucene104.Lucene104ScalarQuantizedVectorsFormat.ScalarEncoding;
 import org.apache.lucene.index.SegmentReadState;
 import org.apache.lucene.index.SegmentWriteState;
+import org.opensearch.knn.index.codec.nativeindex.NativeIndexBuildStrategyFactory;
 import org.opensearch.knn.index.engine.KNNEngine;
 
 import java.io.IOException;
@@ -42,13 +43,25 @@ public class Faiss1040ScalarQuantizedKnnVectorsFormat extends KnnVectorsFormat {
         ScalarEncoding.SINGLE_BIT_QUERY_NIBBLE
     );
 
+    private final NativeIndexBuildStrategyFactory nativeIndexBuildStrategyFactory;
+
     public Faiss1040ScalarQuantizedKnnVectorsFormat() {
+        this(new NativeIndexBuildStrategyFactory());
+    }
+
+    public Faiss1040ScalarQuantizedKnnVectorsFormat(final NativeIndexBuildStrategyFactory nativeIndexBuildStrategyFactory) {
         super(FORMAT_NAME);
+        this.nativeIndexBuildStrategyFactory = nativeIndexBuildStrategyFactory;
     }
 
     @Override
     public KnnVectorsWriter fieldsWriter(SegmentWriteState state) throws IOException {
-        return new Faiss1040ScalarQuantizedKnnVectorsWriter(state, faissSqFlatFormat.fieldsWriter(state), faissSqFlatFormat::fieldsReader);
+        return new Faiss1040ScalarQuantizedKnnVectorsWriter(
+            state,
+            faissSqFlatFormat.fieldsWriter(state),
+            faissSqFlatFormat::fieldsReader,
+            nativeIndexBuildStrategyFactory
+        );
     }
 
     /**
--- a/src/main/java/org/opensearch/knn/index/engine/faiss/FaissHNSWMethod.java
+++ b/src/main/java/org/opensearch/knn/index/engine/faiss/FaissHNSWMethod.java
@@ -41,6 +41,7 @@ import static org.opensearch.knn.common.KNNConstants.METHOD_PARAMETER_M;
 import static org.opensearch.knn.common.KNNConstants.NAME;
 import static org.opensearch.knn.common.KNNConstants.PARAMETERS;
 import static org.opensearch.knn.common.KNNConstants.SPACE_TYPE;
+import static org.opensearch.knn.common.KNNConstants.SQ_BITS;
 import static org.opensearch.knn.common.KNNConstants.VECTOR_DATA_TYPE_FIELD;
 
 /**
@@ -199,6 +200,10 @@ public class FaissHNSWMethod extends AbstractFaissMethod {
             final VectorDataType vectorDataType = extractVectorDataType(parameters);
             final Map<String, Object> encoderMap = extractEncoderMap(parameters);
 
+            if (isSQOneBitIndex(vectorDataType, parameters)) {
+                return true;
+            }
+
             if (isFloat32Index(vectorDataType, encoderMap)) {
                 return true;
             }
@@ -258,7 +263,12 @@ public class FaissHNSWMethod extends AbstractFaissMethod {
             // Check encoding is 'sq' meaning fp32 is being scalar quantized to fp16
             final Map<String, Object> encoderMap = extractEncoderMap(parameters);
             final String encoder = getStringFromMap(encoderMap, NAME);
-            return encoder.equals(ENCODER_SQ);
+            if (encoder.equals(ENCODER_SQ) == false) {
+                return false;
+            }
+            // bits is null for legacy pre-3.6.0 indexes which default to fp16
+            Object bits = encoderMap.get(SQ_BITS);
+            return bits == null || (bits instanceof Integer && (Integer) bits == FaissSQEncoder.Bits.SIXTEEN.getValue());
         } catch (final Exception e) {
             log.debug(e.getMessage());
             // Ignore
@@ -309,6 +319,36 @@ public class FaissHNSWMethod extends AbstractFaissMethod {
             return encoder.equals(ENCODER_FLAT);
         } catch (final Exception e) {
             log.debug(e.getMessage());
+            // Ignore
+            return false;
+        }
+    }
+
+    /**
+     * Checks whether the given parameters represent an SQ 1-bit index (encoder: sq, bits: 1).
+     *
+     * TODO: Consolidate the logic in this function with {@link FaissSQEncoder#isSQOneBit} into one function,
+     * so that there is a single source of truth. Currently, this is not possible because
+     * {@link FaissSQEncoder#isSQOneBit} assumes the encoder object is a {@link MethodComponentContext}.
+     *
+     * @param vectorDataType The data type for the vector field
+     * @param parameters KNN library indexing parameters
+     * @return true if SQ 1 bit, false otherwise
+     */
+    public static boolean isSQOneBitIndex(final VectorDataType vectorDataType, final Map<String, Object> parameters) {
+        try {
+            if (vectorDataType != VectorDataType.FLOAT) {
+                return false;
+            }
+            final Map<String, Object> encoderMap = extractEncoderMap(parameters);
+            final String encoder = getStringFromMap(encoderMap, NAME);
+            if (encoder.equals(ENCODER_SQ) == false) {
+                return false;
+            }
+            Object bits = encoderMap.get(SQ_BITS);
+            return bits instanceof Integer && (Integer) bits == FaissSQEncoder.Bits.ONE.getValue();
+        } catch (final Exception e) {
+            log.error("Failed to check if method parameters contain an sq encoder with bits=1", e);
             // Ignore
             return false;
         }
--- a/src/main/java/org/opensearch/knn/index/vectorvalues/KNNVectorValuesIterator.java
+++ b/src/main/java/org/opensearch/knn/index/vectorvalues/KNNVectorValuesIterator.java
@@ -157,6 +157,8 @@ public interface KNNVectorValuesIterator {
      */
     class FieldWriterIteratorValues<T> extends AbstractVectorValuesIterator {
         private final Function<Integer, T> vectorGetter;
+        private int index = -1;
+        private int lastDocId = Integer.MIN_VALUE;
 
         FieldWriterIteratorValues(@NonNull final DocsWithFieldSet docsWithFieldSet, @NonNull final Map<Integer, T> vectors) {
             super(docsWithFieldSet.iterator());
@@ -171,8 +173,19 @@ public interface KNNVectorValuesIterator {
             // Dense case -> Easy. doc_id == vector_ordinal and doc_id will be given as 0, 1, ..., N - 1
             // Sparse case -> doc_id will be given in increasing order 1, 4, 7, 8, 10, ...
             // but its corresponding vector ordinal is 0, 1, 2, ...
-            final int[] index = { 0 };
-            this.vectorGetter = (docId) -> vectors.get(index[0]++);
+            // The getter is idempotent — multiple getVector() calls for the same doc return the same vector
+            this.vectorGetter = (docId) -> {
+                if (docId != this.lastDocId) {
+                    if (docId < this.lastDocId) {
+                        throw new IllegalStateException(
+                            "Doc IDs must be in increasing order, but got " + docId + " after " + this.lastDocId
+                        );
+                    }
+                    this.index++;
+                    this.lastDocId = docId;
+                }
+                return vectors.get(this.index);
+            };
         }
 
         /**
--- a/src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissFlatIndexFactory.java
+++ b/src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissFlatIndexFactory.java
@@ -13,21 +13,26 @@ import org.opensearch.knn.index.SpaceType;
 import org.opensearch.knn.index.engine.faiss.FaissSQEncoder;
 import org.opensearch.knn.memoryoptsearch.faiss.binary.FaissBinaryHnswIndex;
 import org.opensearch.knn.memoryoptsearch.faiss.binary.FaissBinaryIndex;
+import org.opensearch.knn.memoryoptsearch.faiss.cagra.FaissHNSWCagraIndex;
 
 import java.util.Arrays;
 
 /**
- * Factory that creates the appropriate {@link FaissBinaryIndex} flat storage implementation
- * based on the field's configuration, and wires it into the index tree when needed.
+ * Factory that creates the appropriate flat storage implementation based on the field's configuration,
+ * and wires it into the index tree when needed.
  */
 @UtilityClass
 public class FaissFlatIndexFactory {
 
     /**
-     * Returns a {@link FaissBinaryIndex} to use as flat storage for the given field, or {@code null}
-     * if the FAISS file's own flat storage should be used.
+     * Creates a {@link FaissIndex} flat storage backed by Lucene's {@link FlatVectorsReader} for
+     * indices where flat storage was skipped (IO_FLAG_SKIP_STORAGE). Returns {@code null} if the field
+     * does not require externally-provided flat storage.
+     *
+     * <p>Currently supports SQ 1-bit. To add support for other flat storage types (e.g., fp32 flat),
+     * add new conditions here.
      */
-    static FaissBinaryIndex createBinaryIndex(final FieldInfo fieldInfo, final FlatVectorsReader flatVectorsReader) {
+    static FaissIndex createFlatIndex(final FieldInfo fieldInfo, final FlatVectorsReader flatVectorsReader) {
         if (FieldInfoExtractor.isSQField(fieldInfo)
             && FieldInfoExtractor.extractSQConfig(fieldInfo).getBits() == FaissSQEncoder.Bits.ONE.getValue()) {
             return new FaissScalarQuantizedFlatIndex(flatVectorsReader, fieldInfo.getName());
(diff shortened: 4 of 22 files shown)


src/main/java/org/opensearch/knn/index/vectorvalues/KNNVectorValuesIterator.java
```diff
@@ -171,8 +171,18 @@ class FieldWriterIteratorValues<T> extends AbstractVectorValuesIterator {
             // Dense case -> Easy. doc_id == vector_ordinal and doc_id will be given as 0, 1, ..., N - 1
             // Sparse case -> doc_id will be given in increasing order 1, 4, 7, 8, 10, ...
             // but its corresponding vector ordinal is 0, 1, 2, ...
-            final int[] index = { 0 };
-            this.vectorGetter = (docId) -> vectors.get(index[0]++);
+            // The getter is idempotent — multiple getVector() calls for the same doc return the same vector.
```
rchitale7: i had to make this change since `getVector()` gets called twice for the first vector read, during the remote build flow. This change ensures that if the same doc id is read more than once while iterating through the vectors, then it will return the same vector. Note however that this doesn't allow for random access, like the `Map` approach. 

`getVector` gets called here: https://github.com/opensearch-project/k-NN/blob/4cf03358dfd9f6f85f29fbe133b71ece99ed5c5b/src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/VectorValuesInputStream.java#L67. 

This is done in order to populate the dimension and bytesPerVector: https://github.com/opensearch-project/k-NN/blob/4cf03358dfd9f6f85f29fbe133b71ece99ed5c5b/src/main/java/org/opensearch/knn/index/vectorvalues/KNNFloatVectorValues.java#L28

Then it gets called again here when reloading the buffer: https://github.com/opensearch-project/k-NN/blob/4cf03358dfd9f6f85f29fbe133b71ece99ed5c5b/src/main/java/org/opensearch/knn/index/codec/nativeindex/remote/VectorValuesInputStream.java#L203

So the first time we upload data to the remote store, vector 1 is in the buffer instead of vector 0, causing an off by one error. With this change, we store the last read doc id in `lastDocId`.

src/main/java/org/opensearch/knn/index/vectorvalues/KNNVectorValuesIterator.java
```diff
@@ -171,8 +171,18 @@ class FieldWriterIteratorValues<T> extends AbstractVectorValuesIterator {
...
+            // (once in initializeVectorValues for dimension/bytesPerVector, once in reloadBuffer for data).
+            final int[] index = { -1 };
+            final int[] lastDocId = { Integer.MIN_VALUE };
+            this.vectorGetter = (docId) -> {
+                if (docId != lastDocId[0]) {
+                    index[0]++;
+                    lastDocId[0] = docId;
+                }
+                return vectors.get(index[0]);
+            };
```
navneet1v: can we add a unit test for this to ensure that we can validate the functionality

src/main/java/org/opensearch/knn/index/engine/faiss/FaissHNSWMethod.java
```diff
@@ -258,7 +263,12 @@ public static boolean isFloat16Index(final VectorDataType vectorDataType, final
             // Check encoding is 'sq' meaning fp32 is being scalar quantized to fp16
             final Map<String, Object> encoderMap = extractEncoderMap(parameters);
             final String encoder = getStringFromMap(encoderMap, NAME);
-            return encoder.equals(ENCODER_SQ);
+            if (!encoder.equals(ENCODER_SQ)) {
```
0ctopus13prime: NIT : we do `encoder.equals() == false` (Please see https://github.com/opensearch-project/k-NN/blob/main/DEVELOPER_GUIDE.md?plain=1#L131)

src/main/java/org/opensearch/knn/index/engine/faiss/FaissHNSWMethod.java
```diff
@@ -314,6 +324,32 @@ private static boolean isByteIndex(final VectorDataType vectorDataType, final Ma
...
+     * @return True if SQ 1 bit, False otherwise
+     */
+    public static boolean isSQOneBitIndex(final VectorDataType vectorDataType, final Map<String, Object> parameters) {
+        try {
+            if (vectorDataType != VectorDataType.FLOAT) {
+                return false;
+            }
+            final Map<String, Object> encoderMap = extractEncoderMap(parameters);
+            final String encoder = getStringFromMap(encoderMap, NAME);
+            if (!encoder.equals(ENCODER_SQ)) {
```
0ctopus13prime: NIT : we do encoder.equals() == false (Please see https://github.com/opensearch-project/k-NN/blob/main/DEVELOPER_GUIDE.md?plain=1#L131)

src/main/java/org/opensearch/knn/index/engine/faiss/FaissHNSWMethod.java
```diff
@@ -314,6 +324,32 @@ private static boolean isByteIndex(final VectorDataType vectorDataType, final Ma
...
+                return false;
+            }
+            Object bits = encoderMap.get(SQ_BITS);
+            return bits instanceof Integer && (Integer) bits == FaissSQEncoder.Bits.ONE.getValue();
+        } catch (final Exception e) {
+            log.debug(e.getMessage());
```
0ctopus13prime: Could you consider doing 
```
log.debug("<summarized message>", e);
```

to have stack trace in log? otherwise we will lose stacktrace, making us hard to debug which threw the exception

src/main/java/org/opensearch/knn/index/vectorvalues/KNNVectorValuesIterator.java
0ctopus13prime: Can we add a minimum validation check on increasing doc id order?
doc_id must > lastDocId[0]

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/cagra/FaissHNSWCagraIndex.java
```diff
@@ -75,8 +85,12 @@ protected void doLoad(final IndexInput input) throws IOException {
...
         ((FaissCagraHNSW) faissHnsw).setNumBaseLevelSearchEntryPoints(numBaseLevelSearchEntryPoint);
 
-        // Partial load flat vector storage
-        flatVectors = FaissIndex.load(input);
+        // Partial load flat vector storage.
+        // When IO_FLAG_SKIP_STORAGE was used (e.g., skip_stored_vectors remote build for SQ 1-bit),
+        // Faiss writes a "null" section. FaissFlatIndexFactory will wire in the real flat
+        // storage (e.g., FaissScalarQuantizedFlatIndex) after loading.
+        final FaissIndex loaded = FaissIndex.load(input);
+        flatVectors = FaissEmptyIndex.isEmptyIndex(loaded) ? null : loaded;
```
0ctopus13prime: Could you consider using `FaissEmptyIndex`?
I think you have reason for using `null`, could you elaborate? 🤔 
FaissIndex.load will give you `FaissEmptyIndex` for null section in faiss.

src/main/java/org/opensearch/knn/index/engine/faiss/FaissHNSWMethod.java
```diff
@@ -314,6 +324,32 @@ private static boolean isByteIndex(final VectorDataType vectorDataType, final Ma
...
     }
 
+    /**
+     * Checks whether the given parameters represent an SQ 1-bit index (encoder: sq, bits: 1).
+     *
+     * @param vectorDataType The data type for the vector field
+     * @param parameters KNN library indexing parameters
+     * @return True if SQ 1 bit, False otherwise
+     */
+    public static boolean isSQOneBitIndex(final VectorDataType vectorDataType, final Map<String, Object> parameters) {
```
navneet1v, line 338: are there not functions that already does this checking?

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/cagra/FaissHNSWCagraIndex.java
```diff
@@ -49,6 +50,15 @@ public FaissHNSWCagraIndex(final String indexType) {
...
     }
 
+    /**
+     * Sets the flat vector storage for this skip_stored_vectors index.
+     * Used by {@link org.opensearch.knn.memoryoptsearch.faiss.FaissFlatIndexFactory} to wire in flat storage
+     * for skip_stored_vectors indices (e.g., SQ 1-bit remote build with IO_FLAG_SKIP_STORAGE).
+     */
+    public void setFlatVectors(final FaissIndex flatVectors) {
+        this.flatVectors = flatVectors;
+    }
```
navneet1v: are setting the flat index like this at other place too?

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/cagra/FaissHNSWCagraIndex.java
rchitale7: yeah i'll change this, i wanted to match the behavior of `FaissBinaryHnswIndex` because `storage` is null when uninitialized there. But thinking about this more `FaissEmptyIndex` is cleaner

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss1040ScalarQuantizedKnnVectorsFormat.java
```diff
@@ -42,13 +43,25 @@ public class Faiss1040ScalarQuantizedKnnVectorsFormat extends KnnVectorsFormat {
         ScalarEncoding.SINGLE_BIT_QUERY_NIBBLE
     );
 
+    private final NativeIndexBuildStrategyFactory nativeIndexBuildStrategyFactory;
+
     public Faiss1040ScalarQuantizedKnnVectorsFormat() {
+        this(new NativeIndexBuildStrategyFactory());
```
navneet1v, line 49: do we need to create a new instance of `NativeIndexBuildStrategyFactory` per format creation which is basically gets created per segment?

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss1040ScalarQuantizedKnnVectorsFormat.java
```diff
@@ -42,13 +43,25 @@ public class Faiss1040ScalarQuantizedKnnVectorsFormat extends KnnVectorsFormat {
...
+    }
+
+    public Faiss1040ScalarQuantizedKnnVectorsFormat(NativeIndexBuildStrategyFactory nativeIndexBuildStrategyFactory) {
```
navneet1v:
```suggestion
    public Faiss1040ScalarQuantizedKnnVectorsFormat(final NativeIndexBuildStrategyFactory nativeIndexBuildStrategyFactory) {
```

src/main/java/org/opensearch/knn/index/engine/faiss/FaissHNSWMethod.java
```diff
@@ -314,6 +324,32 @@ private static boolean isByteIndex(final VectorDataType vectorDataType, final Ma
...
+            }
+            final Map<String, Object> encoderMap = extractEncoderMap(parameters);
+            final String encoder = getStringFromMap(encoderMap, NAME);
+            if (encoder.equals(ENCODER_SQ) == false) {
+                return false;
+            }
+            Object bits = encoderMap.get(SQ_BITS);
+            return bits instanceof Integer && (Integer) bits == FaissSQEncoder.Bits.ONE.getValue();
+        } catch (final Exception e) {
+            log.debug(e.getMessage(), e);
```
navneet1v: Please a put error message. This should be not be a debug log. Also the string in the exception message should be more verbose.

src/main/java/org/opensearch/knn/index/vectorvalues/KNNVectorValuesIterator.java
```diff
@@ -171,8 +171,21 @@ class FieldWriterIteratorValues<T> extends AbstractVectorValuesIterator {
             // Dense case -> Easy. doc_id == vector_ordinal and doc_id will be given as 0, 1, ..., N - 1
             // Sparse case -> doc_id will be given in increasing order 1, 4, 7, 8, 10, ...
             // but its corresponding vector ordinal is 0, 1, 2, ...
-            final int[] index = { 0 };
-            this.vectorGetter = (docId) -> vectors.get(index[0]++);
+            // The getter is idempotent — multiple getVector() calls for the same doc return the same vector.
+            // This is required by VectorValuesInputStream which calls getVector() twice for the first doc
+            // (once in initializeVectorValues for dimension/bytesPerVector, once in reloadBuffer for data).
```
navneet1v: I won't say this is required by `VectorValuesInputStream`. the getter should be idempotent since that is the contract of the VectorValues.

src/main/java/org/opensearch/knn/index/vectorvalues/KNNVectorValuesIterator.java
```diff
@@ -171,8 +171,21 @@ class FieldWriterIteratorValues<T> extends AbstractVectorValuesIterator {
...
+            final int[] index = { -1 };
+            final int[] lastDocId = { Integer.MIN_VALUE };
```
navneet1v: why these 2 variables are arrays? based on the code read I can see they can just be 2 variables of the class `FieldWriterIteratorValues`.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/cagra/FaissHNSWCagraIndex.java
```diff
@@ -49,6 +49,16 @@ public FaissHNSWCagraIndex(final String indexType) {
...
 
+    /**
+     * Sets the flat vector storage for this skip_stored_vectors index.
+     * Used by {@link org.opensearch.knn.memoryoptsearch.faiss.FaissFlatIndexFactory} to wire in flat storage
+     * for skip_stored_vectors indices (e.g., SQ 1-bit remote build with IO_FLAG_SKIP_STORAGE).
+     * @param flatVectors Flat vector storage
+     */
+    public void setFlatVectors(final FaissIndex flatVectors) {
+        this.flatVectors = flatVectors;
+    }
```
navneet1v: are we setting the flat index like this at other place too?

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissFlatIndexFactory.java
```diff
@@ -13,66 +13,103 @@
...
+        return null;
+    }
+
     /**
      * If the loaded FAISS index has no flat storage (e.g. SQ with 1-bit skips it via IO_FLAG_SKIP_STORAGE),
-     * wires in the appropriate flat index via {@link #createBinaryIndex}.
+     * wires in the appropriate flat index. Handles both binary HNSW (local JNI build) and float CAGRA HNSW
+     * (remote build) skip_stored_vectors indices.
      */
     static void maybeSetFlatBinaryIndex(final FaissIndex faissIndex, final FieldInfo fieldInfo, final FlatVectorsReader flatVectorsReader) {
```
navneet1v, line 61: this function is going bit out of hands here, and would need refactoring. Shouldn't be part of this PR, but somehow we have to have to take care of it.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissFlatIndexFactory.java
```diff
@@ -13,66 +13,103 @@
...
-            throw new IllegalStateException(
-                String.format(
-                    "%s found for field [%s] but %s returned null — cannot wire binary flat storage.",
-                    FaissEmptyIndex.class.getName(),
-                    fieldInfo.getName(),
-                    FaissFlatIndexFactory.class.getName()
-                )
-            );
+        // Binary HNSW path (local JNI build with IO_FLAG_SKIP_STORAGE)
+        if (nested instanceof FaissBinaryHnswIndex binaryHnswIndex && binaryHnswIndex.getStorage() == null) {
```
navneet1v, line 68: can we change `binaryHnswIndex.getStorage() == null` to use `FaissEmptyIndex.isEmptyIndex()`.

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss1040ScalarQuantizedKnnVectorsFormat.java
rchitale7: i was matching what `NativeEngines990KnnVectorsFormat` does: https://github.com/opensearch-project/k-NN/blob/9332fbc5911335814b53de846fd279886cd7ed17/src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsFormat.java#L49. I thought I would follow the convention, but technically it could be null? Since this constructor isn't used by the write path.

src/main/java/org/opensearch/knn/index/engine/faiss/FaissHNSWMethod.java
rchitale7: There is one here: https://github.com/opensearch-project/k-NN/blob/9332fbc5911335814b53de846fd279886cd7ed17/src/main/java/org/opensearch/knn/index/engine/faiss/FaissSQEncoder.java#L262. But it casts the encoder map as a `MethodComponentContext`, so it wont work with this file.

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/cagra/FaissHNSWCagraIndex.java
rchitale7: For `FaissBinaryHnswIndex` we use the `Setter` annotation: https://github.com/opensearch-project/k-NN/blame/9332fbc5911335814b53de846fd279886cd7ed17/src/main/java/org/opensearch/knn/memoryoptsearch/faiss/binary/FaissBinaryHnswIndex.java#L35. I can make this consistent with that

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissFlatIndexFactory.java
rchitale7: So the problem is that `storage` for `FaissBinaryHnswIndex` is of type `FaissBinaryIndex`. We can't store `FaissEmptyIndex` in a `FaissBinaryIndex`, because `FaissEmpyIndex` doesn't extend it. I thought it would be too intrusive to make that change, so i left that as is.

rchitale7:
CI was failing due to `DerivedSourceIT.testSourceFiltering_withVariousIncludeExcludeCombinations`. But that was unrelated to my changes. I fixed it here: https://github.com/opensearch-project/k-NN/pull/3280. Based on @0ctopus13prime feedback I made a small name change to the IT test method to `searchWithSourceIncludesExcludes` to make things more clear.

navneet1v (approved):

0ctopus13prime (approved):

Issue #3284, opened 2026-04-21
[Enhancement] Use BulkScore to Score the Vectors in MOS path when K > number of docs in Segment


navneet1v:
## Description
While reading through the code what I identified that in MOS case when K > number of docs in a segment we do exhaustive search, but that search is not using bulk score functionality of RandomVectorScorer. Due to this we are missing these things:
1. Since bulk score interface is not used we are missing on performance optimizations.
2. Since prefetch only works with bulkScore, we are missing the prefetching of vectors.

KNN code ref: https://github.com/opensearch-project/k-NN/blob/b9ca7bc16b1bfc87d937fb0c10dc78d12562a4d9/src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissMemoryOptimizedSearcher.java#L165-L179

We can build a similar logic like Lucene: https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsReader.java#L352-L385

This will ensure that we use both bulk score api and also the Prefetch capabilities.

The gains with this might be minimal, but this will ensure that for larger values of K we are missing an obvious optimization

navneet1v:
PR raised: https://github.com/opensearch-project/k-NN/pull/3285

navneet1v:
code is merged. Resolving the issue.

PR #3285, opened 2026-04-21
Add the bulkscore logic in MOS when K is greater than number of docs in segment

navneet1v:
## Description

When k >= number of vectors in a segment, Memory Optimized Search (MOS) falls back to an exhaustive search path instead of using HNSW graph traversal. Previously, this path scored vectors one at a time using `scorer.score(i)`, which incurs per-vector overhead from individual method calls and misses SIMD/vectorization opportunities.

This PR replaces the single-vector scoring with batch bulk scoring using Lucene's `RandomVectorScorer.bulkScore()` API. Vectors are accumulated into fixed-size batches of `EXHAUSTIVE_BULK_SCORE_ORDS` and scored together, enabling the JVM to apply SIMD optimizations across the batch. This aligns with how Lucene's own Lucene99HnswVectorsReader handles the same exhaustive search case. Using `RandomVectorScorer.bulkScore()` ensure that we are also doing prefetch before doing vector search

### Changes

1. Replaced per-vector scorer.score(i) + knnCollector.collect() loop with batched scorer.bulkScore(ords, scores, numOrds) approach

2. Vectors are accumulated into int[] ords batches of size EXHAUSTIVE_BULK_SCORE_ORDS

3. Each batch is bulk-scored and only collected if bulkScore > minCompetitiveSimilarity, allowing entire batches to be skipped

4. Remainder vectors after the last full batch are handled in a separate block

5. Visited count is now incremented per-batch instead of per-vector

### Tests

1. testExhaustiveBulkScoring_whenKExceedsMaxOrd_thenAllVectorsCollected — Verifies all vectors are collected when k > totalVectors, exercising both full-batch and remainder paths

2. testExhaustiveBulkScoring_whenFiltered_thenOnlyAcceptedVectorsCollected — Verifies filtering via acceptedOrds works correctly with bulk scoring (only even-numbered ordinals accepted)

3. testExhaustiveBulkScoring_resultsMatchAnnSearch — Validates correctness by comparing exhaustive bulk scoring top-k against ANN search results (recall > 0.8) and verifying descending score order

### Related Issues
https://github.com/opensearch-project/k-NN/issues/3284

### Check List


src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissMemoryOptimizedSearcher.java
```diff
@@ -163,16 +165,36 @@ private void search(
...
+            int[] ords = new int[EXHAUSTIVE_BULK_SCORE_ORDS];
+            float[] scores = new float[EXHAUSTIVE_BULK_SCORE_ORDS];
+            int numOrds = 0;
+            for (int i = 0; i < numVectors; i++) {
                 if (acceptedOrds == null || acceptedOrds.get(i)) {
-                    if (!knnCollector.earlyTerminated()) {
-                        knnCollector.incVisitedCount(1);
-                        knnCollector.collect(scorer.ordToDoc(i), scorer.score(i));
-                    } else {
+                    if (knnCollector.earlyTerminated()) {
```
Vikasht34, line 176: since this branch only runs for k >= scorer.maxOrd(), it seems unlikely that earlyTerminated() / minCompetitiveSimilarity() used much here. Would this path be simpler if it just batched all accepted ords, called bulkScore, and collected unconditionally?

src/test/java/org/opensearch/knn/memoryoptsearch/FaissMemoryOptimizedSearcherTests.java
```diff
@@ -1343,6 +1343,147 @@ public long cost() {
...
+        final TopDocs topDocs = knnCollector.topDocs();
+        assertEquals(expectedCount, topDocs.scoreDocs.length);
+    }
+
+    /**
+     * Exhaustive bulk scoring results should match ANN search results with high recall,
+     * confirming the bulk scoring path produces correct scores.
+     */
+    @SneakyThrows
+    public void testExhaustiveBulkScoring_resultsMatchAnnSearch() {
```
Vikasht34: I don’t think this test is validating the new bulk-scoring path strongly enough. It compares ANN results to the exhaustive path and checks recall > 0.8, but ANN is approximate, so this can pass/fail for reasons unrelated to the bulk-score implementation. how we will validating , just wanted to learn?

src/test/java/org/opensearch/knn/memoryoptsearch/FaissMemoryOptimizedSearcherTests.java
```diff
@@ -1343,6 +1343,147 @@ public long cost() {
...
+        final int dimension = 768;
+        final int totalVectors = 300;
+        final int k = 30;
+
+        final float[] query = new float[dimension];
+        for (int i = 0; i < dimension; i++) {
+            query[i] = (float) Math.random();
```
Vikasht34: Could we make this deterministic with a fixed seed or a canned query vector?

src/test/java/org/opensearch/knn/memoryoptsearch/FaissMemoryOptimizedSearcherTests.java
```diff
@@ -1343,6 +1343,147 @@ public long cost() {
...
+        Mockito.when(fieldInfo.getAttribute(KNNConstants.SPACE_TYPE)).thenReturn(SpaceType.L2.getValue());
+        final FaissMemoryOptimizedSearcher searcher = new FaissMemoryOptimizedSearcher(
+            input,
+            FaissIndex.load(input),
+            fieldInfo,
+            FlatVectorsScorerProvider.getFlatVectorsScorer(fieldInfo, KNNVectorSimilarityFunction.EUCLIDEAN, SCORER)
+        );
+
+        // k > totalVectors triggers the exhaustive bulk scoring path
+        final KnnCollector knnCollector = new TopKnnCollector(totalVectors + 1, Integer.MAX_VALUE);
```
Vikasht34: code path is actually for k >= scorer.maxOrd() so the exact boundary k == totalVectors still not tested?

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissMemoryOptimizedSearcher.java
navneet1v: > Would this path be simpler if it just batched all accepted ords, called bulkScore, and collected unconditionally?

The idea of checking minCompetitiveSimilarity, is to ensure that if the best document in this batch is worst than minCompetitiveSimilarity, we can avoid collecting it in our min heap. This is very minor optimization, but its good to have.

early terminated will be used if this segment which has docs < K is coming after some expensive segment search in case of Concurrent Segment Search.

src/test/java/org/opensearch/knn/memoryoptsearch/FaissMemoryOptimizedSearcherTests.java
navneet1v: will fix this

src/test/java/org/opensearch/knn/memoryoptsearch/FaissMemoryOptimizedSearcherTests.java
navneet1v: I can add 2 search both > and = .

src/test/java/org/opensearch/knn/memoryoptsearch/FaissMemoryOptimizedSearcherTests.java
navneet1v: This was more of a way to check. I will just add a proper ground truth on this to ensure that we can check with actual values

Vikasht34 (dismissed):

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissMemoryOptimizedSearcher.java
```diff
@@ -163,16 +165,36 @@ private void search(
             // Do ANN search with Lucene's HNSW graph searcher.
             HnswGraphSearcher.search(scorer, collector, new FaissHnswGraph(hnsw, indexInput.clone()), acceptedOrds);
         } else {
-            // If k is larger than the number of vectors, we can just iterate over all vectors
-            // and collect them.
-            for (int i = 0; i < scorer.maxOrd(); i++) {
+            // if k is larger than the number of vectors we expect to visit in an HNSW search,
+            // we can just iterate over all vectors and collect them.
+            int numVectors = scorer.maxOrd();
```
shatejas, line 170: Functionally this looks good to me

Was wondering if all exact search should be consolidated at one place, the previous code seemed simple enough to have its own implementation. From maintenance perspective it should try reusing existing exact search construct

navneet1v: Its a good point. I was thinking about the same. but then I thought exact search which we have right now should be triggered in special cases. But in long term we should try to merge the paths. We have to get out of the exactsearcher and move everything understand readers.

navneet1v:
@shatejas and @Vikasht34 can I get another round of review for this PR?

Vikasht34 (commented):
Looks good to me , So moving to Excat Scoprer to unify the logic will be seperate effort ?

Vikasht34 (approved):

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/FaissMemoryOptimizedSearcher.java
VijayanB: nit: can we move `int numVectors = scorer.maxOrd();` outside and share it at line 164

VijayanB (approved):

VijayanB:
@navneet1v can this be merged?

Issue #3283, opened 2026-04-20
[FEATURE] Sandbox Incubator: Experimental Algorithms & Optimizations


vamshin:
**Is your feature request related to a problem?**
As the codebase grows with increasingly complex capabilities like new approximate algorithms, quantization techniques, optimizations, tiered storage architectures etc, there is no structured home for experimental work that is not yet production-ready. Merging unproven algorithms directly into the main package risks instability, increases maintenance burden, and creates ambiguity for users about what is stable versus experimental. The absence of a clear incubation path also discourages contributors from proposing bold, high-risk/high-reward ideas, since there is no safe space to bake and validate them before full commitment.

**What solution would you like?**
Create a sandbox/ folder or package within the repository that serves as a first-class incubation environment for experimental algorithms, optimizations, and architectural components. The sandbox should have a clear lifecycle: new ideas land in sandbox/, are exposed to early adopters and community feedback under an explicit experimental label, and graduate to the main package once they meet defined stability, performance, and test coverage thresholds.

**What alternatives have you considered?**
**Separate repository / fork** — spinning up a standalone repo for experiments. This provides strong isolation but fragments contribution workflows, duplicates CI infrastructure, and makes it harder to graduate code back into the main package.

rcpeken:
Hi! I'd like to work on this issue. I plan to start by creating a new sandbox module and configuring the build scripts (Gradle) to isolate it from the main package. Before I submit a draft PR, are there any specific CI workflows or architectural guidelines I should be aware of for this new module?

rcpeken:
@navneet1v

navneet1v:
@rcpeken do you have a plan in mind what you are thinking to add in the new sandbox module? Atleast as of now I need to put some thoughts on what exactly we need in this sandbox module. But if you have some draft ready we can start with that.

rcpeken:
My plan is to start with the structural skeleton rather than specific algorithms.

I can open a Draft PR that includes:

A new isolated sandbox module with its own build.gradle.

A simple marker interface (e.g., ExperimentalAlgorithm) to set the architectural boundary.

CI pipeline integration to ensure the new module builds correctly.

This will give us a clean, isolated environment to discuss and build upon. I'll prepare this draft so we have something concrete to review. How does that sound? @navneet1v

navneet1v:
[...]

PR #3288, opened 2026-04-22
Replace KNNEngine enumeration usage with VectorSearchEngine API

reta:
### Description
This is a second change towards supporting extensibility (with respect to KNN/Vector libraries and engines) in the k-NN plugin. The change is small and contained, and does two things:
 - introduces `VectorSearchEngine ` interface (by pivoting off the enumeration) - https://github.com/opensearch-project/k-NN/pull/3443
 - replaces `KNNEngine` usage with `VectorSearchEngine`

#### Why? 
In scope of https://github.com/opensearch-project/OpenSearch/issues/20050, we have discussed the possible ways how new KNN/Vector engines could be introduced, and most importantly, co-exist with the ones provided by K-NN plugin. The `KNNEngine` is one of the key component here, unfortunately is it designed as enumeration and is not extensible (from the outside) but fixed. 

To break that barrier, this first change converts the `KNNEngine` into `VectorSearchEngine` interface but keeps the previous revision of the enumeration. Built on top of https://github.com/opensearch-project/k-NN/pull/3443, the pull request replaces `KNNEngine` -> `VectorSearchEngine`,  at the same time keeps the API / usage mostly unaffected. 

#### What is next? 

This changes unblock the possibility to extract KNN Engine/Library SPIs.

### Related Issues
Towards https://github.com/opensearch-project/OpenSearch/issues/20050

### Check List


reta:
@vamshin @navneet1v would really appreciate your thoughts folks, the pull request look large but 99% of changes is just name change  `KNNEngine` -> `BuiltinKNNEngine` whereas `KNNEngine` becomes an interface, that's about to be fair. Thank you.

vamshin:
> @vamshin @navneet1v would really appreciate your thoughts folks, the pull request look large but 99% of changes is just name change `KNNEngine` -> `BuiltinKNNEngine` whereas `KNNEngine` becomes an interface, that's about to be fair. Thank you.

Thanks @reta. We have a similar idea here to create a Sandbox module/package. https://github.com/opensearch-project/k-NN/issues/3283. Should we incorporate the code changes in this direction?

Also regarding the changes, wondering if we could reserve kNNEngine name for built in Engine(we can be clear about it in the javadoc) and choose another name for interface and other implementations can prefix their name to kNNEngine?

reta:
Thanks @vamshin 

> Thanks @reta. We have a similar idea here to create a Sandbox module/package. #3283. Should we incorporate the code changes in this direction?

Yes. totally possible as well, the end goal would be to have SPI module, which also aligns well with the sandbox idea. I didn't want to pull out new module for a sake of more contained change but that would come next.

> Also regarding the changes, wondering if we could reserve kNNEngine name for built in Engine(we can be clear about it in the javadoc) and choose another name for interface and other implementations can prefix their name to kNNEngine?

This is possible, however that would explode the amount of changes. `KNNEngine` is a very key type which is used literally everywhere (including https://github.com/Opensearch-project/neural-search), however we could introduce another interface `XxxKNNEngine` -> `KNNEngine` if necessary indeed.

reta:
@vamshin @navneet1v Wondering if we could move forward with this change folks? It should help a lot with sandboxing stuff as well, since it should be possible to sandbox not only part of the vector (K-NN) engine, but the complete engine itself. The next step would be to have an SPI module - this change detaches pretty much 99% of the abstractions from concrete implementations. If naming could be better - it could be always changed. Thanks again folks!

src/main/java/org/opensearch/knn/index/engine/AbstractKNNEngineBase.java
```diff
@@ -0,0 +1,154 @@
...
+import org.opensearch.knn.index.mapper.CompressionLevel;
+import org.opensearch.knn.index.mapper.Mode;
+import org.opensearch.knn.index.query.rescore.RescoreContext;
+import org.opensearch.knn.memoryoptsearch.VectorSearcherFactory;
+import org.opensearch.remoteindexbuild.model.RemoteIndexParameters;
+
+/**
+ * Abstract base implementation of the {@link KNNEngine}.
+ */
+public abstract class AbstractKNNEngineBase implements KNNEngine {
```
navneet1v: what is the use of this class?

navneet1v:
> @vamshin @navneet1v Wondering if we could move forward with this change folks? It should help a lot with sandboxing stuff as well, since it should be possible to sandbox not only part of the vector (K-NN) engine, but the complete engine itself. The next step would be to have an SPI module - this change detaches pretty much 99% of the abstractions from concrete implementations. If naming could be better - it could be always changed. Thanks again folks!

The change looks good to me in terms of what we want to do. But this change is just renaming and changing code all around. Just to avoid such blast can we do this:

1. KNNEngine enum remains same and we don't need to change it BuiltInKNNEngine.
2. A new interface can be `VectorSearchEngine` or `VectorEngine`. Which then can be implemented by `KNNEngine` enum.
3. The AbstractKNNEngine I am not sure what is the use of it since its not being used by any other class. So we can remove it. 

Overall change what you are doing in terms of new classes is same, but just avoid this huge 196 files PR.

```
PR.    ->.   Suggestion
KNNEngine. ->  VectorEngine or VectorSearchEngine
BuiltInKNNEngine -> KNNEngine (same as mainline)
AbstractKNNEngine -> not needed.
```

@vamshin WDYT?

reta:
> Overall change what you are doing in terms of new classes is same, but just avoid this huge 196 files PR.

> PR.    ->.   Suggestion
> KNNEngine. ->  VectorEngine or VectorSearchEngine
> BuiltInKNNEngine -> KNNEngine (same as mainline)
> AbstractKNNEngine -> not needed.

Thanks @navneet1v, the whole purpose of `BuiltInKNNEngine` -> `KNNEngine` is this: all usages of `KNNEngine` do not use enum anymore but interface, looking identical. Replacing usage of KNNEngine with VectorEngine / VectorSearchEngine or literally anything will require order of magnitude more changes (I am up for it, but 196 would look small comparing to what we will end up with).

src/main/java/org/opensearch/knn/index/engine/AbstractKNNEngineBase.java
reta: Yes, it used by `KNNEngine.UNDEFINED`

navneet1v:
> > Overall change what you are doing in terms of new classes is same, but just avoid this huge 196 files PR.
> 
> > PR.    ->.   Suggestion
> > KNNEngine. ->  VectorEngine or VectorSearchEngine
> > BuiltInKNNEngine -> KNNEngine (same as mainline)
> > AbstractKNNEngine -> not needed.
> 
> Thanks @navneet1v, the whole purpose of `BuiltInKNNEngine` -> `KNNEngine` is this: all usages of `KNNEngine` do not use enum anymore but interface, looking identical. Replacing usage of KNNEngine with VectorEngine / VectorSearchEngine or literally anything will require order of magnitude more changes (I am up for it, but 196 would look small comparing to what we will end up with).

I am not sure if I fully understand your point. If we move to `VectorSearchEngine` as the name, all new engines can just adopt that. and whatever is in the KNNPlugin will just keep on using KNNEngine enum. This is what I am thinking. 

Why you say that it will lead to order of magnitude more changes? can you share some light on this.

reta:
> I am not sure if I fully understand your point. If we move to `VectorSearchEngine` as the name, all new engines can just adopt that. and whatever is in the KNNPlugin will just keep on using KNNEngine enum. This is what I am thinking.

Ah I see, so this is the problem: if KNNPlugin will continue using KNNEngine, nothing changes. By changing KNNEngine to interface, KNNPlugin still thinks it deals with same enum-like KNNEngine, but under the hood - it is not, it will be possible to supply **any** sandboxed or experimental engine, and KNNPlugin  will still work the same way. 

> Why you say that it will lead to order of magnitude more changes? can you share some light on this.

That basically stems from the fact if we keep KNNEngine as enum that implements VectorSearchEngine , all references to KNNEngine  have to be replaced with VectorSearchEngine  to make it useful and let KNNPlugin to support  **any** sandboxed or experimental engine.

reta:
> KNNEngine. -> VectorEngine or VectorSearchEngine
> BuiltInKNNEngine -> KNNEngine (same as mainline)

@navneet1v finally found time to get back to this, I did the changes from above as well as replaced all usages of `KNNEngine` with `VectorSearchEngine`, the changeset turned out to be smaller (to my complete surprise), would appreciate the look, thanks!

navneet1v:
> > KNNEngine. -> VectorEngine or VectorSearchEngine
> > BuiltInKNNEngine -> KNNEngine (same as mainline)
> 
> @navneet1v finally found time to get back to this, I did the changes from above as well as replaced all usages of `KNNEngine` with `VectorSearchEngine`, the changeset turned out to be smaller (to my complete surprise), would appreciate the look, thanks!

sure will check

navneet1v:
@reta I was checking the code, and I was thinking what if we don't change the `KNNEngine` being used in KNN plugin, and just keep the `enum class extends VectorSearchEngine`. Would that leave the changes limited to just 2-4 files. I know we should move towards using `VectorSearchEngine` rather than `KNNEngine`, but could that change be done slowly as we change the part of code that needs `VectorSearchEngine` interfaces? Thoughts? 

other than that code looks good to me, please update the PR description with new details of the interfaces

reta:
> . I know we should move towards using `VectorSearchEngine` rather than `KNNEngine`, but could that change be done slowly as we change the part of code that needs `VectorSearchEngine` interfaces? Thoughts?

Thanks @navneet1v , sure, I could split pull request in two part, will do that shortly, thanks! Here we are https://github.com/opensearch-project/k-NN/pull/3443, I will rebase this pull request right after.

navneet1v:
@reta should we close this PR now since the other Pr is merged. https://github.com/opensearch-project/k-NN/pull/3443

reta:
> @reta should we close this PR now since the other Pr is merged. #3443

@navneet1v this is the next step to move off the KNNEngine enumeration, so we could progress towards SPI layers, I have rebased against `main` and changed the description (as we discussed  here https://github.com/opensearch-project/k-NN/pull/3288#issuecomment-5027972938), hope it makes sense, thanks!

reta:
@navneet1v wondering if we could move forward with adopting `VectorSearchEngine`? it is quite a low risk change (in my opinion, since we didn't really cut anything out, just replacing enum with interface), keeping change up to date for so long takes considerable amount of time, thank you.

reta:
No interests from maintainers, closing

PR #3299, opened 2026-04-29
Overrides visits to honor filter query

shatejas:
### Description
Corrects the visitor patter for knn querybuilder to visit the underlying filter. 

Visitor pattern is intended to visit each underlying query and take action against it. By not having an inner query, the high level call to `query.visit(QueryVisitor)` won't be able to visit the filter query and stops at k-NN query which can cause silent unintended behavior. The visitor pattern relies on right implementation of `visits` to be able to avoid recursion logic in core or elsewhere to evaluate a query

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
+++ b/src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
@@ -10,6 +10,7 @@ import lombok.AllArgsConstructor;
 import lombok.Getter;
 import lombok.Setter;
 import lombok.extern.log4j.Log4j2;
+import org.apache.lucene.search.BooleanClause;
 import org.apache.lucene.search.MatchNoDocsQuery;
 import org.apache.lucene.search.Query;
 import org.opensearch.common.ValidationException;
@@ -21,6 +22,7 @@ import org.opensearch.core.xcontent.XContentBuilder;
 import org.opensearch.index.mapper.MappedFieldType;
 import org.opensearch.index.query.AbstractQueryBuilder;
 import org.opensearch.index.query.QueryBuilder;
+import org.opensearch.index.query.QueryBuilderVisitor;
 import org.opensearch.index.query.QueryRewriteContext;
 import org.opensearch.index.query.QueryShardContext;
 import org.opensearch.index.query.WithFieldName;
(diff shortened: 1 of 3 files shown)


src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
```diff
@@ -713,6 +715,25 @@ private byte[] getByteVectorForCreatingQueryRequest(
...
+     * the filter's visitation to it. This enables traversal of the full query tree including any
+     * nested filter queries.
+     *
+     * @param visitor the {@link QueryBuilderVisitor} used to traverse the query tree
+     */
+    @Override
+    public void visit(QueryBuilderVisitor visitor) {
+        visitor.accept(this);
+        if (filter != null) {
+            final QueryBuilderVisitor subVisitor = visitor.getChildVisitor(BooleanClause.Occur.FILTER);
```
navneet1v, line 732: Why we are using boolean.Filter here?

shatejas: Because its a filter clause and we need an `AND` behavior here. 

The ChildVisitor behavior assumes that it can have any Occur clause and all query visitors are forced to behave differently (if they need to) based on these occur clauses.

Vikasht34 (commented):
Was this a bug? Can u write bit more on what's the problem statement if we don't have this change? So that I can have bit more context.

shatejas:
> Was this a bug? Can u write bit more on what's the problem statement if we don't have this change? So that I can have bit more context.

Added more context in description. If we don't have this change, any query evaluation might not evaluate the query correctly. For instance, if a vistor wants to check if there is a term query in the overall query it will return false negative with filtered k-NN query

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
navneet1v: @shatejas can you please add this in the comments of the code for reference. Rest code looks good to me.

shatejas:
```
 /**
     * Visits this query builder and its filter clause using the provided visitor.
     * <p>
     * This method first accepts the visitor for the current {@link KNNQueryBuilder}, then if a filter
     * is present, it obtains a child visitor for {@link BooleanClause.Occur#FILTER} for `AND` behavior and delegates
     * the filter's visitation to it. This enables traversal of the full query tree including any
     * nested filter queries.
     *
     * @param visitor the {@link QueryBuilderVisitor} used to traverse the query tree
     */
```
Made a minor change, any more suggestions?

navneet1v (approved):

Vikasht34 (approved):

Issue #3298, opened 2026-04-29
[RFC] Rescoring Post Radial Search on Quantized Index


0ctopus13prime:
# Low-Level Design: Rescoring After Radial Search on Quantized Index

## 1. Problem Statement

Radial search (`max_distance` / `min_score`) is blocked for quantized indices because scores computed on quantized
vectors contain error. The user-provided radius cannot be reliably applied to quantized scores. With scalar
quantization (32x), score conversion is now feasible, enabling us to introduce a rescoring layer.

### Correctness issue: false positives

Vectors whose quantized score places them inside the radius but whose true score is outside. The rescoring layer
eliminates these by recomputing scores against full-precision vectors.

Note: false negatives (vectors truly within the radius but excluded by quantized scoring) are not addressed. This is
acceptable — the quantized first pass is an approximation, and users opting into quantization accept this recall
tradeoff.

### Why this approach

- Engine-agnostic: one wrapper handles both Faiss (`KNNQuery`) and Lucene (`FloatVectorSimilarityQuery`) paths
- Follows existing patterns: `RescoreKNNVectorQuery` (Lucene top-K) and `NativeEngineKnnVectorQuery.doRescore` (Faiss
  top-K) both use the same "run inner query, rescore with ExactSearcher" pattern
- Clean separation of concerns: the inner query handles approximate search, the wrapper handles rescoring
- Natural path to P1 optimizations (prefetch, parallel rescoring)

### Alternatives considered

**Alternative A: Rescore inside KNNWeight.searchLeaf()**

- Fastest path to P0 but doesn't cover Lucene engine path (FloatVectorSimilarityQuery uses Lucene's own Weight)
- Mixes approximate search and rescoring concerns in an already complex class
- Rejected because: not engine-agnostic, harder to extend for P1

### 3.1 New Class: `RescoreRadialSearchQuery`

**Package:** `org.opensearch.knn.index.query`

A Lucene `Query` wrapper that:

1. Runs the inner query (radial search on quantized vectors) with the user's radius as-is
2. Collects first-pass results (doc IDs) per leaf
3. Loads full-precision vectors for those doc IDs and recomputes true scores
4. Filters out docs whose true score falls outside the user's radius

```
RescoreRadialSearchQuery
  - innerQuery: Query                  // KNNQuery or FloatVectorSimilarityQuery
  - field: String
  - queryVector: float[]
  - radius: Float                      // user's radius (used for both passes)
  - maxResultWindow: int
[...]

Issue #3290, opened 2026-04-23
RFC: Default 32x Compression for k-NN Vector Indices


kotwanikunal:
## Summary

Starting at `V_3_7_0`, `EngineResolver` will default new HNSW indices to `compression_level=32x` when compression is not explicitly set in the mapping. The resolved encoder is  SQ 1-bit quantization. This applies to both the Faiss and Lucene engines. The mechanism is a version gate on `indexVersionCreated`: the resolver checks `indexVersionCreated >= V_3_7_0` before injecting the default, so existing indices are unaffected. 

There is no cluster-level setting; the gate is immutable and tied to the index. Scope is limited to new index creation with `type=knn_vector`, `data_type=float`, and `method.algorithm=hnsw` (explicit or resolved). 

To opt out, a user sets `compression_level=1x` and derives `mode=in_memory` automatically. The `mode` parameter will be deprecated starting V_3_7_0; storage mode is derived from compression level. The change shifts the default from the zero-compression FP32 baseline to the cost-performance optimum validated by the benchmarking plan in this RFC.

## Motivation

The current default for a `knn_vector` field with `data_type=float` and `method.algorithm=hnsw` is FP32 in-memory storage with no compression. Every vector is stored as four bytes per dimension, loaded into the JVM heap or native memory at segment open time, and scored with exact FP32 arithmetic. This is the zero-compression baseline: maximum recall, maximum memory cost.

Most users do not set `compression_level` or `mode` explicitly. They accept the default. The current default forces every such user onto the highest-cost configuration regardless of whether their workload requires it. Users who need exact FP32 recall can opt out explicitly; the default should serve the common case, not the edge case.

The proposed change shifts the default to the cost-performance optimum: SQ 1-bit with rescore-assisted recall. The version gate on `indexVersionCreated` ensures the change is applied only to indices created at `V_3_7_0` or later, leaving all existing indices unaffected.

### Current Behavior

`EngineResolver.resolveKNNEngine()` is the entry point for all engine and method resolution. It receives an `OriginalMappingParameters` instance, which holds only the values the user explicitly provided in the mapping. The current logic short-circuits immediately when both `mode` and `compression` are `NOT_CONFIGURED`:

```
if (mode == Mode.NOT_CONFIGURED && compression == CompressionLevel.NOT_CONFIGURED) {
    // fall through to engine-specific defaults -- no compression applied
}
```

[...]

PR #3300, opened 2026-04-30
Rescoring after radial search on quantized index. [Task 1 - 4]

0ctopus13prime:
### Description
RFC : https://github.com/opensearch-project/k-NN/issues/3298

Enable radial search on 32x scalar quantized indices by introducing a rescoring wrapper query. Previously, radial search was blocked for all quantized indices.

- Selectively removed the radial search block: allows 32x SQ through, still blocks BQ and non-32x SQ
- Added compressionLevel and quantizationConfig fields to CreateQueryRequest to propagate quantization metadata to RNNQueryFactory
- Created RescoreRadialSearchQuery — a wrapper Query that delegates to the inner radial search query via ScorerSupplier, designed for Lucene's slice-based parallelism
- Wired the wrapper in RNNQueryFactory: detects CompressionLevel.x32 and wraps the inner KNNQuery (Faiss) or FloatVectorSimilarityQuery (Lucene)
- Refactored RNNQueryFactory.create() into createNativeEngineRadialQuery() and createLuceneRadialQuery()
- Currently pass-through (no actual rescoring yet) — full-precision rescoring via ExactSearcher will be added in a follow-up

### Related Issues
N/A

### Check List


navneet1v:
@0ctopus13prime can we add the response to the code suggestion by the bot and also ensure that coverage is upto the mark.

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
```diff
@@ -488,10 +488,27 @@ protected Query doToQuery(QueryShardContext context) {
...
+            // - NOT_CONFIGURED: no compression, fp32
+            // - x1: no compression, fp32
+            // - x2: fp16, not quantized
+            // - x32: 1-bit SQ, supported via rescoring (RescoreRadialSearchQuery)
+            // All other levels (x4, x8, x16) are SQ without rescoring support and are blocked.
+            final CompressionLevel compressionLevel = knnMappingConfig.getCompressionLevel();
+            if (compressionLevel != CompressionLevel.NOT_CONFIGURED
+                && compressionLevel != CompressionLevel.x1
+                && compressionLevel != CompressionLevel.x2
+                && compressionLevel != CompressionLevel.x32) {
```
navneet1v: Lets move this under compression level in a function which can encapsulate this logic.

src/main/java/org/opensearch/knn/index/query/RNNQueryFactory.java
```diff
@@ -63,63 +63,106 @@ public static Query create(
...
         }
     }
 
+    /**
+     * Determines whether the index uses 32x scalar quantization, which requires rescoring
+     * after radial search to eliminate false positives from quantization error.
+     */
+    private static boolean isQuantizedForRescore(CreateQueryRequest request) {
+        return request.getCompressionLevel() == CompressionLevel.x32;
+    }
```
navneet1v: why we are checking its just 32x compression? the name of the function is `isQuantizedForRescore` then all the cases when compression is not 1x and 2x are all quantizedForRescore.

src/test/java/org/opensearch/knn/memoryoptsearch/FaissSQRadialSearchBlockedIT.java
```diff
@@ -1,142 +0,0 @@
...
-
-import java.util.concurrent.ThreadLocalRandom;
-
-import static org.opensearch.knn.index.KNNSettings.KNN_INDEX;
-
-/**
- * Integration test to verify that radial search (max_distance / min_score) is blocked
- * for Faiss SQ (scalar quantized, 32x compression) indices.
- */
-public class FaissSQRadialSearchBlockedIT extends KNNRestTestCase {
```
navneet1v, line 25: since we are removing this IT, lets add an IT which tells that Radial Search is working. As we are progressing towards making radial search, we should ensure that we start with basic IT and then keeps on improving that IT as we make more and more changes

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
0ctopus13prime: sure, sounds good!

src/main/java/org/opensearch/knn/index/query/RNNQueryFactory.java
0ctopus13prime: I think we should give it a better name. like `isRescoringRequired`.
Since this is one to decide whether rescoring is required.
I'll change the method name.

src/test/java/org/opensearch/knn/memoryoptsearch/FaissSQRadialSearchBlockedIT.java
0ctopus13prime: Sounds good, will add it in next rev.

0ctopus13prime:
Dear bot, thank you for the suggestions and I reject them all.

1. Null Dereference Risk : It's an established pattern even before this commit. It was not introduced in this PR.

2. Incomplete Implementation : Yes, this is intended. The rescoring part will be dealt with in the next PR, note that we're merging this change into the feature branch.

3. Inconsistent Guard Logic : This is intended.  BQ has quantization config while SQ does not. So it's inconsistent, but it's fine.

4. Array Equality Issue : Yeah, it's confirmed. No issue

src/main/java/org/opensearch/knn/index/query/RNNQueryFactory.java
navneet1v: Whether rescoring is needed or not doesn't only depend on the compression level. It also depends on the rescore context provided by the user or not. So just compression level check is not enough. What you are actually trying to ask from this function is:
1. Should we wrap the Radial Search query with a rescore query or not by default?

Suggestion for a better name will be isDefaultRescoringRequired, and then the decision of rescoring needed for radial search or not should still come from the CompressionLevel enum.

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
```diff
@@ -488,10 +488,16 @@ protected Query doToQuery(QueryShardContext context) {
...
+            // BQ or other non-SQ quantized index do not support radial search
+            if ((knnMappingConfig.getQuantizationConfig() != QuantizationConfig.EMPTY)) {
+                throw new UnsupportedOperationException(
+                    "Radial search is not supported for non-32x-SQ quantized indices, quantization config was not empty."
+                );
+            }
+
+            final CompressionLevel compressionLevel = knnMappingConfig.getCompressionLevel();
+            if (CompressionLevel.supportsRadialSearch(compressionLevel) == false) {
+                throw new UnsupportedOperationException("Radial search is not supported for compression level=" + compressionLevel);
```
navneet1v: can we move this logic in the `KNNVectorFieldType`. generally in OpenSearch MappedFieldType suggest which query is supported or not on a field. This is similar to that, where we are saying for SQ 1 bit with 32x compression supports Radial Search. I would suggest moving the logic there.

src/test/java/org/opensearch/knn/index/query/RadialSearchOnQuantizedIndexIT.java
```diff
@@ -0,0 +1,230 @@
...
+import static org.opensearch.knn.index.KNNSettings.KNN_INDEX;
+
+/**
+ * Integration tests verifying that radial search (max_distance / min_score) works on
+ * 32x SQ quantized indices for both Faiss and Lucene engines, with both HNSW and FLAT methods.
+ *
+ * These tests validate that the query pipeline executes without error and returns results.
+ * Recall/precision validation will be added once ExactSearcher rescoring is implemented.
+ */
+public class RadialSearchOnQuantizedIndexIT extends KNNRestTestCase {
```
navneet1v: The way we have organize the tests is based on engines, can we move the ITs to their respective engine IT classes.

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
```diff
@@ -488,10 +488,16 @@ protected Query doToQuery(QueryShardContext context) {
...
-                throw new UnsupportedOperationException("Radial search is not supported for indices which have quantization enabled");
+            // BQ or other non-SQ quantized index do not support radial search
+            if ((knnMappingConfig.getQuantizationConfig() != QuantizationConfig.EMPTY)) {
+                throw new UnsupportedOperationException(
+                    "Radial search is not supported for non-32x-SQ quantized indices, quantization config was not empty."
+                );
+            }
+
+            final CompressionLevel compressionLevel = knnMappingConfig.getCompressionLevel();
+            if (CompressionLevel.supportsRadialSearch(compressionLevel) == false) {
```
navneet1v: any reason for keeping supportsRadialSearch a static function? why not make it nonstatic and do something like

`compressionLevel.supportsRadialSearch()` this way every compression level can tell if supports radial search or not. and on the CompressionLevel class you just need to remove static from there and rest of the code works as it is.

src/main/java/org/opensearch/knn/index/mapper/CompressionLevel.java
```diff
@@ -90,6 +90,24 @@ public static boolean isConfigured(CompressionLevel compressionLevel) {
...
+     *   <li>{@code NOT_CONFIGURED} — no compression (fp32)</li>
+     *   <li>{@code x1} — no compression (fp32)</li>
+     *   <li>{@code x2} — fp16, not quantized</li>
+     *   <li>{@code x32} — 1-bit SQ, supported via rescoring ({@code RescoreRadialSearchQuery})</li>
+     * </ul>
+     *
+     * @param compressionLevel the compression level to check
+     * @return {@code true} if radial search is supported for this compression level
+     */
+    public static boolean supportsRadialSearch(final CompressionLevel compressionLevel) {
```
VijayanB: Hey @0ctopus13prime, thanks for the PR! Just a small suggestion on the supportsRadialSearch method in CompressionLevel — since radial search support can depend on more than just the compression level (e.g., engine type, quantization config, mode), it might be worth considering extracting this into a separate utility/resolver that takes in all the relevant candidates and derives the decision holistically. This way, CompressionLevel stays focused on compression semantics, and the logic becomes easier to extend and test as new configurations are added down the line. Just a thought — happy to discuss if you see it differently!

src/main/java/org/opensearch/knn/index/query/RescoreRadialSearchQuery.java
```diff
@@ -0,0 +1,201 @@
...
+
+    /** The original query vector provided by the user. */
+    private final float[] queryVector;
+
+    /**
+     * The engine-specific radius threshold.
+     * For Faiss, this is a raw distance converted via {@code KNNEngine.distanceToRadialThreshold()}.
+     * For Lucene, this is a similarity value converted via {@code KNNEngine.scoreToRadialThreshold()}.
+     */
+    private final float radius;
```
VijayanB, line 62: I see that field, queryVector, and radius are stored in RescoreRadialSearchQuery but don't have any consumers in this PR yet — they're only referenced in toString(). 

What do you think about deferring these fields until the PR that actually wires up the rescoring logic? That way the class stays minimal for now, with just innerQuery, and we add field, queryVector, and radius alongside their real consumers. Keeps the diff focused and easier to review.

src/main/java/org/opensearch/knn/index/query/RescoreRadialSearchQuery.java
```diff
@@ -0,0 +1,201 @@
...
+
+    /**
+     * Constructs a new rescoring wrapper for radial search on a quantized index.
+     *
+     * @param innerQuery  the inner radial search query (must not be null)
+     * @param field       the knn_vector field name (must not be null)
+     * @param queryVector the query vector (must not be null)
+     * @param radius      the radius threshold for the search
+     */
+    public RescoreRadialSearchQuery(final Query innerQuery, final String field, final float[] queryVector, float radius) {
```
VijayanB, line 72: why not use Builder pattern ?

src/main/java/org/opensearch/knn/index/query/RescoreRadialSearchQuery.java
```diff
@@ -0,0 +1,201 @@
...
+
+    /**
+     * Weight implementation that wraps the inner weight and provides per-leaf scorer suppliers.
+     *
+     * <p>The {@link #scorerSupplier(LeafReaderContext)} method returns a {@link ScorerSupplier}
+     * whose {@code get()} method executes the full per-leaf pipeline:</p>
+     * <ol>
+     *   <li>Run the inner weight's scorer (quantized radial search on this leaf)</li>
+     *   <li>Collect first-pass candidate doc IDs</li>
+     *   <li>(TODO) Rescore candidates with {@code ExactSearcher} using full-precision vectors</li>
```
VijayanB: since you are merging in feature branch and your subsequent PR will add the wiring, i would prefer to include this TODO inside class and keep class doc to what you want to do

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
0ctopus13prime: sure

src/test/java/org/opensearch/knn/index/query/RadialSearchOnQuantizedIndexIT.java
0ctopus13prime: sure, will update in the next rev

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
0ctopus13prime: sure, will update in the next rev

src/main/java/org/opensearch/knn/index/mapper/CompressionLevel.java
0ctopus13prime: sounds good!
I will follow @navneet1v 's suggestion, putting the logic as a member function. then we can call `compressionLevel.supportRadialSearch`

src/main/java/org/opensearch/knn/index/query/RescoreRadialSearchQuery.java
0ctopus13prime: sure, will update in the next rev

src/main/java/org/opensearch/knn/index/query/RescoreRadialSearchQuery.java
0ctopus13prime: Usually, I prefer builder pattern to create something that's complicate to instantiate which needs another steps to get parameters.
In this case, just passing few parameters, I think it would be simple to just call constructor.
What do you think?

src/main/java/org/opensearch/knn/index/query/RescoreRadialSearchQuery.java
0ctopus13prime: good idea. will update in the next rev

src/main/java/org/opensearch/knn/index/query/RescoreRadialSearchQuery.java
VijayanB: When you have multiple field with different field type, it is better to use Builder pattern

0ctopus13prime: I don't think gain is much. Usually the benefit of builder pattern comes when it's complicated to construct parameters and when it has a long list of parameters to pass.
In this case, we only pass few of them nor, all of them need additional steps to be prepared.
We can't abuse builder pattern just because we have few parameters blindly.
I probably keep them as they are.

src/main/java/org/opensearch/knn/index/mapper/CompressionLevel.java
VijayanB: Right now isRescoringRequired only looks at compression level in isolation, so it can't tell if the combination actually makes sense. Pass all the relevant params so it can make a holistic decision, not one based on a fragment.

0ctopus13prime: I see, so you want a general predicate function for 'is rescoring needed for radial search'.
Got it, like the idea.

src/main/java/org/opensearch/knn/index/query/RescoreRadialSearchQuery.java
0ctopus13prime: Hey Vijay, thought twice, but will keep it as it is since the next PR will come right away after this.
I dont think keeping it will bring huge impact.

src/main/java/org/opensearch/knn/index/query/RescoreRadialSearchQuery.java
VijayanB: @0ctopus13prime I understand your perspective, but I'd like to push back here. The Builder pattern isn't just about the number of parameters — it's about type safety and readability at the call site. With two parameters of similar types (String and float[], plus a bare float), it's easy to silently swap arguments or misread what each positional arg means at the call site. A Builder makes the intent explicit. Additionally, this class might evolve in the future with additional configuration fields. A Builder gives us a stable construction API that's easy to extend without breaking existing call sites or ending up with telescoping constructors.

This isn't about over-engineering — it's about setting up the class correctly from the start. Let's get it right now while the class is new rather than refactoring later

0ctopus13prime: @VijayanB 
Thank you for the suggestion, I prefer to leave it as it is. 
I think for 4 parameters using builder pattern would make things much verbose than it's required.
I don't see a huge gain for using builder pattern. I understand if we are trying to instantiate Http client which needs complicated setting, but this is not the case. Simple one line of instantiating would be much easier to understand. 

I think using builder pattern in here is overkill. We don't need 'get it right' anything. It's already 'right' without builder pattern.

0ctopus13prime: @VijayanB 
Attaching AI's answer

Question:
```
do you think it's better to use builder pattern to replace this? if so, what's the gain? public RescoreRadialSearchQuery(final Query innerQuery, final String field, final float[] queryVector, float radius)
```

```
Short answer: probably not. A builder pattern only pays off if object construction is getting complex or error-prone. Your constructor is still pretty small and readable.

Let’s break it down.

Right now you have:

4 parameters
2 primitives + 1 array + 1 object
no obvious ambiguity in ordering (maybe field vs innerQuery, but still manageable)

That’s still within the “constructor is fine” zone.

Where a builder starts to make sense
You’d get real value if one or more of these become true:

Too many parameters (usually ~5–7+)
Especially if several share the same type (e.g., multiple float or String values).

Optional parameters
If radius or others become optional, or you start adding flags/configs:

new Query(..., ..., ..., ..., true, false, 0.7f, ...)

That’s where readability collapses.

Validation or invariants
If you need logic like:
radius > 0
queryVector.length == expectedDim
field must match mapping
A builder can centralize and enforce that before build().
Extensibility
If you expect this class to evolve (which seems likely for a search query type), builders prevent constructor explosion.

What you actually gain with a builder

Clarity at call site:

RescoreRadialSearchQuery query =
    RescoreRadialSearchQuery.builder()
        .innerQuery(q)
        .field("embedding")
        .queryVector(vec)
        .radius(0.8f)
        .build();

This avoids:

new RescoreRadialSearchQuery(q, "embedding", vec, 0.8f);

The difference is small now, but scales much better as arguments grow.

Safer construction:
You can enforce constraints in build() instead of relying on callers to “do the right thing.”

Flexibility:
You can add new parameters later without breaking existing code.

What you lose

More boilerplate (especially in Java without Lombok)
Slightly heavier usage for simple cases
Overengineering if the class stays small

A pragmatic take for your case

Given your current constructor:

(Query innerQuery, String field, float[] queryVector, float radius)

I would not switch yet.

But I would switch if you expect things like:

similarityMetric
boost
minScore
filterQuery
normalization
vectorType / dim

Search query objects tend to grow, so a builder is often justified eventually.

A middle-ground option

If you're on the fence, you can:

Keep the constructor
Add a builder later without breaking API

or even:

Add static factory methods for common cases:
RescoreRadialSearchQuery.of(innerQuery, field, vec, radius);

Bottom line

Builder is not better here yet — just more verbose.
It becomes better when your query object starts behaving like a configuration object rather than a simple data holder.
```

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
0ctopus13prime: I've moved the validation logic under `RadialSearchSupport` rather than putting it under KNNVectorFieldType.
I thought it would be better to isolate validation logics to `RadialSearchSupport`.

src/main/java/org/opensearch/knn/index/query/RNNQueryFactory.java
0ctopus13prime: @navneet1v 
Whether we need rescoring or not on radial search solely depends on compression level. 
If 32x, rescoring is mandatory. The goal of rescoring is to prevent false positive in radial search.

src/main/java/org/opensearch/knn/index/mapper/CompressionLevel.java
Vikasht34: What's the advantage of using static method and have in Compression level Class , why can't be bake in with  x4(4, "4x", RescoreContext.builder().oversampleFactor(1.0f).userProvided(false).build(), Set.of(Mode.ON_DISK)), compression level itself . With this any future added compression level will have to handled with compression level itself.

src/main/java/org/opensearch/knn/index/util/RadialSearchSupport.java
```diff
@@ -0,0 +1,86 @@
...
+import static org.opensearch.knn.index.engine.KNNEngine.ENGINES_SUPPORTING_RADIAL_SEARCH;
+
+/**
+ * Utility class for radial search validation and rescoring decisions.
+ *
+ * <p>Centralizes the logic for determining whether radial search is supported for a given
+ * index configuration, and whether rescoring with full-precision vectors is required.</p>
+ */
+@UtilityClass
+public class RadialSearchSupport {
```
Vikasht34: We don't need this class and any of the dependent class , any decision making happening with respect to engine has to be with Engine itself, we already have engine interface , bake it there , we want logic to be stays together , we don't want decision making logic scattered at every place.

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
```diff
@@ -42,6 +42,7 @@
 import org.opensearch.knn.index.query.parser.RescoreParser;
 import org.opensearch.knn.index.query.rescore.RescoreContext;
 import org.opensearch.knn.index.util.IndexUtil;
+import org.opensearch.knn.index.util.RadialSearchSupport;
```
Vikasht34: Bake it in engine level.

src/main/java/org/opensearch/knn/index/query/BaseQueryFactory.java
```diff
@@ -59,6 +61,10 @@ public static class CreateQueryRequest {
         private RescoreContext rescoreContext;
         private boolean expandNested;
         private boolean memoryOptimizedSearchEnabled;
+        @Builder.Default
```
Vikasht34: Why we need this in Base Query Factory.

src/main/java/org/opensearch/knn/index/query/RNNQueryFactory.java
```diff
@@ -8,9 +8,9 @@
 import static org.opensearch.knn.common.KNNConstants.DEFAULT_LUCENE_RADIAL_SEARCH_TRAVERSAL_SIMILARITY_RATIO;
 import static org.opensearch.knn.common.KNNConstants.VECTOR_DATA_TYPE_FIELD;
 import static org.opensearch.knn.index.VectorDataType.SUPPORTED_VECTOR_DATA_TYPES;
+import static org.opensearch.knn.index.util.RadialSearchSupport.isRescoringRequired;
```
Vikasht34: so we have KnnQuery Factory , why we are scattering decision here and there, the job of this factory to just create the final query.

src/main/java/org/opensearch/knn/index/query/BaseQueryFactory.java
0ctopus13prime: We need this for radial search determination logic

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
```diff
@@ -479,20 +480,7 @@ protected Query doToQuery(QueryShardContext context) {
...
-            if (vectorDataType == VectorDataType.BINARY) {
-                throw new UnsupportedOperationException(String.format(Locale.ROOT, "Binary data type does not support radial search"));
-            }
-
-            if ((knnMappingConfig.getQuantizationConfig() != QuantizationConfig.EMPTY)
-                // If compression level is 32x, then radial search should be blocked.
-                || (knnMappingConfig.getCompressionLevel() == CompressionLevel.x32)) {
-                throw new UnsupportedOperationException("Radial search is not supported for indices which have quantization enabled");
-            }
+            RadialSearchSupport.validate(knnEngine, vectorDataType, knnMappingConfig);
```
Vikasht34: It's not needed here , please bake it engine level.

0ctopus13prime: It's needed there Vikash for validating

src/main/java/org/opensearch/knn/index/query/RNNQueryFactory.java
0ctopus13prime: We need this `isRescoringRequired` to determine the final class

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
0ctopus13prime: We can't bake things in engine level. 
We need other information, refer to RadialSupport class that I added

src/main/java/org/opensearch/knn/index/util/RadialSearchSupport.java
0ctopus13prime: We need this class to factor out radial search support related logics

0ctopus13prime: Will keep it as it is

src/main/java/org/opensearch/knn/index/mapper/CompressionLevel.java
0ctopus13prime: We can't just bake things in compression level.
32x BQ does not support radial search.
32x SQ does support radial search.

src/main/java/org/opensearch/knn/index/util/RadialSearchSupport.java
Vikasht34: I don't think so ..

Vikasht34 (changes requested):

src/main/java/org/opensearch/knn/index/util/RadialSearchSupport.java
0ctopus13prime: We need a dedicated class for radial support as it needs sort of information. 
I think it would be better to put all validation, predicate in one place
Will keep it as it is

src/main/java/org/opensearch/knn/index/mapper/CompressionLevel.java
Vikasht34: CompressionLevel.supportsRadialSearch(compressionLevel)  also have some issue

Vikasht34:
Hey, thanks for putting this together — I appreciate the effort here. I do have some concerns about the current approach that I think are worth discussing before we move forward.

The core issue is that the PR uses CompressionLevel and QuantizationConfig as proxies to infer which encoder was used at index time, and then makes radial search decisions based on that inference. This indirect approach is fragile and doesn't scale well.

For example, imagine we add a new FaissIVFSQEncoder that also produces x32 compression but doesn't set QuantizationConfig. At query time we'd see CompressionLevel=x32, QuantizationConfig=EMPTY — which is identical to what FaissSQEncoder(bits=1) produces. The system would then incorrectly assume radial search is supported, even though the new encoder might not have a rescoring path yet. Working around cases like this would require a growing chain of if-else checks, which gets harder to maintain over time.

I'd love to get @navneet1v's and @shatejas  thoughts on this as well. Could we set up a quick discussion to align on a cleaner design before iterating further? I'm happy to brainstorm together — I'm confident we can find an approach that's both clean and future-proof.

src/main/java/org/opensearch/knn/index/mapper/CompressionLevel.java
kotwanikunal: This is conflicting directly with the refactoring around mappers and rescore that we are thinking of. Can you figure out a better way to achieve this check? `isSQ1bit()` checks work for now.

src/main/java/org/opensearch/knn/index/query/RescoreRadialSearchQuery.java
```diff
@@ -0,0 +1,198 @@
...
+            };
+        }
+
+        /**
+         * Returns {@code true} because the rescore result is deterministic for the same
+         * query parameters and segment state — safe to cache.
+         */
+        @Override
+        public boolean isCacheable(final LeafReaderContext ctx) {
+            return true;
```
kotwanikunal, line 203: Shouldn't this check the delegate rather than assuming underlying values are cacheable?
```
return innerWeight.isCacheable(ctx);
```

src/main/java/org/opensearch/knn/index/util/RadialSearchSupport.java
```diff
@@ -0,0 +1,80 @@
...
+     *
+     * <p>Currently, rescoring is needed for 32x SQ (1-bit scalar quantization) to eliminate
+     * false positives caused by quantization error. This will be extended to other compression
+     * levels as rescoring support is added.</p>
+     *
+     * @param compressionLevel the compression level of the index
+     * @return {@code true} if rescoring is required after radial search
+     */
+    public static boolean isRescoringRequired(final CompressionLevel compressionLevel) {
+        return compressionLevel == CompressionLevel.x32;
```
kotwanikunal: Let's please use direct isSQ1bit checks for now - this will add in a lot of unnecessary routing and delegation. We have the refactoring changes lined up which can help with these checks.

kotwanikunal (commented):
Thanks for the changes! The query wrapping pattern (`RescoreRadialSearchQuery`) looks solid and aligns with where we want to go for radial + SQ 1-bit.

My concern is with the branching validation. We have an active design for 32x defaults ([RFC #3290](https://github.com/opensearch-project/k-NN/issues/3290)) that replaces scattered CompressionLevel checks with capability based dispatch through  the `IndexCapabilities` value object. This PR adds new type based check call sites that go in the opposite direction:

1. CompressionLevel.supportsRadialSearch(level) - should be IndexCapabilities.supportsRadialSearch (boolean, set by encoder during resolution)
2. RadialSearchSupport.isRescoringRequired(level) - should be IndexCapabilities.needsRescore() (derived from RescoreStrategy)
3. CompressionLevel + QuantizationConfig threaded through CreateQueryRequest - should be a single IndexCapabilities object

The core issue (same as @Vikasht34's point): using CompressionLevel as a proxy for encoder behavior doesn't scale. When we add 4-bit SQ or other encoders, every == x32 check becomes a maintenance burden. The encoder should declare its own capabilities, and consumers should read those and not inspect compression levels.

src/main/java/org/opensearch/knn/index/util/RadialSearchSupport.java
0ctopus13prime: Yeah, sounds good.
Is it `isSQOneBit`? I think that's the one

src/main/java/org/opensearch/knn/index/query/RescoreRadialSearchQuery.java
0ctopus13prime: No, regardless of whether innerWeight is cacheable or not (technically this should not true for radial search, but assuming that it's not cacheable), this weight can be reused.
This is because the result per query is deterministic, given the same query, the result will always be the same.

0ctopus13prime:
@Vikasht34 
Your concern will be addressed with `isSQOneBit` from Kunal's.

src/main/java/org/opensearch/knn/index/query/RescoreRadialSearchQuery.java
```diff
@@ -0,0 +1,198 @@
...
+ * <p>This class is currently a pass-through skeleton that delegates entirely to the inner query.
+ * The rescoring logic (using {@code ExactSearcher} with full-precision vectors) will be added
+ * in a subsequent task.</p>
+ *
+ * @see RescoreKNNVectorQuery similar pattern for Lucene engine top-K rescoring
+ * @see org.opensearch.knn.index.query.nativelib.NativeEngineKnnVectorQuery similar pattern for Faiss engine top-K rescoring
+ */
+@Getter
+@EqualsAndHashCode(callSuper = false)
+public class RescoreRadialSearchQuery extends Query {
```
shatejas, line 46: Have we tried reusing `FunctionScoreQuery` here? Its on lucene side and it already has an implementation of `FloatVectorSimilarityValuesSource` which might be able to do rescoring for us. Was just wondering if we can reuse it and not reinvent the wheel

src/main/java/org/opensearch/knn/index/query/RescoreRadialSearchQuery.java
shatejas: IMO we should not take AI's answer as justification to sway one way or the other. It can very well hallucinate and give a biased perspective. 

I don't mind going either way for now but lets not put AI's justification on pedestal here at least for something subjective

Vikasht34:
> @Vikasht34 Your concern will be addressed with `isSQOneBit` from Kunal's.



> @Vikasht34 Your concern will be addressed with `isSQOneBit` from Kunal's.

I am not sure how this will be addressed, the fundamental problem with this approach is , we are not deciding correctly on what parameters would be applied for find radial search is enabled or not. 



two choices here

1. If we enable at felid level  then problem would be it global and if any of the encoding or quantization does not support then we need to write logic to block it , which is also not good in terms of descsion point.
2. I feel we should have cleaner interface that descide based on encoder which encoder supports raidal or not.

0ctopus13prime:
Hi @Vikasht34 
Thank you for the suggestions.
So the first alignment was to move validation logic + predicate under `KNNVectorFieldType`.
And reuse whatever we have for now (ex: isSQOneBit) and use it as a predicate to determine whether rescoring is needed.

I think the initial alignment will resolve your concerns.
Thanks

Vikasht34:
> Hi @Vikasht34 Thank you for the suggestions. So the first alignment was to move validation logic + predicate under `KNNVectorFieldType`. And reuse whatever we have for now (ex: isSQOneBit) and use it as a predicate to determine whether rescoring is needed.
> 
> I think the initial alignment will resolve your concerns. Thanks

I don't think moving to KNNVectorFieldType how it resolves.

Vikasht34:
> > Hi @Vikasht34 Thank you for the suggestions. So the first alignment was to move validation logic + predicate under `KNNVectorFieldType`. And reuse whatever we have for now (ex: isSQOneBit) and use it as a predicate to determine whether rescoring is needed.
> > I think the initial alignment will resolve your concerns. Thanks
> 
> I don't think moving to KNNVectorFieldType how it resolves.

I will wait foR PR changes to see what 's intention

src/main/java/org/opensearch/knn/index/query/RescoreRadialSearchQuery.java
```diff
@@ -0,0 +1,198 @@
...
+     *
+     * @param searcher  the index searcher
+     * @param scoreMode the score mode requested by the collector
+     * @param boost     the boost factor to apply to rescored document scores
+     * @return a weight that delegates scoring to the inner weight, with rescoring to be added
+     * @throws IOException if an I/O error occurs during weight creation
+     */
+    @Override
+    public Weight createWeight(final IndexSearcher searcher, final ScoreMode scoreMode, final float boost) throws IOException {
+        final Query rewritten = searcher.rewrite(innerQuery);
```
shatejas: Can you please confirm if its the right place to do a rewrite. I was looking at https://github.com/apache/lucene/blob/main/lucene/queries/src/java/org/apache/lucene/queries/function/FunctionScoreQuery.java#L126 and it overrides the rewrite for innerQuery. 

Please confirm this as it might have side effects

src/main/java/org/opensearch/knn/index/query/RescoreRadialSearchQuery.java
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    @Override
+    public String toString(String field) {
+        return "RescoreRadialSearchQuery[field=" + this.field + ", radius=" + radius + ", innerQuery=" + innerQuery.toString(field) + "]";
+    }
+
+    @Override
+    public void visit(QueryVisitor visitor) {
+        visitor.visitLeaf(this);
```
shatejas: Similarly please double check this implementation. This won't visit the innerQuery and will stop here. The visitor pattern is intended to visit the entire query tree and this stops it at RescoreRadialSearchQuery

Ref: https://github.com/apache/lucene/blob/main/lucene/queries/src/java/org/apache/lucene/queries/function/FunctionScoreQuery.java#L135

0ctopus13prime: Yeah, I found the similar pattern in wrapper Query like BoostQuery is using the same.
Will update it in next rev.

src/main/java/org/opensearch/knn/index/query/RescoreRadialSearchQuery.java
0ctopus13prime: Discussed in online, will update it in next rev.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,7 +1,7 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionSha256Sum=16f2b95838c1ddcf7242b1c39e7bbbb43c842f1f1a1a0dc4959b6d4d68abcac3
-distributionUrl=https\://services.gradle.org/distributions/gradle-9.2.0-all.zip
+distributionSha256Sum=708d2c6ecc97ca9a11838ef64a6c2301151b8dd10387e22dc1a12c30557cab5b
```
navneet1v, line 3: see if you just want to rebase your base branch to remove all this.

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldType.java
```diff
@@ -225,4 +227,85 @@ public float[] transformQueryVector(float[] vector) {
...
+            }
+        }
+    }
+
+    /**
+     * Determines whether rescoring with full-precision vectors is required after radial search.
+     *
+     * <p>This method should only be called for knn field types that have already been validated
+     * to support radial search via {@link #validateSupportRadialSearch(KNNEngine)}.
+     * Calling this on an unsupported configuration may return incorrect results.</p>
```
navneet1v, line 294: This give me little chills when I see this. We should think of a better way here so that as a caller I don't need to worry about the order of calling the functions.

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldType.java
```diff
@@ -225,4 +227,85 @@ public float[] transformQueryVector(float[] vector) {
...
+     * SQ encoder with {@code bits=1}. Other quantization types may be added in the future.</p>
+     *
+     * @return {@code true} if rescoring is required after radial search
+     */
+    public boolean isRescoringRequiredForRadial() {
+        final KNNMappingConfig mappingConfig = getKnnMappingConfig();
+        final Optional<KNNMethodContext> methodContext = mappingConfig.getKnnMethodContext();
+        if (methodContext.isPresent() == false) {
+            return false;
+        }
```
navneet1v, line 308: Lets a comment on why we think, if methodContext is null then rescoring is not required for radial search.

src/main/java/org/opensearch/knn/index/query/BaseQueryFactory.java
```diff
@@ -71,10 +75,6 @@ public Optional<QueryShardContext> getContext() {
         public Optional<RescoreContext> getRescoreContext() {
             return Optional.ofNullable(rescoreContext);
         }
-
-        public boolean isExpandNested() {
-            return expandNested;
-        }
```
navneet1v, line 77: where is this code moved now?

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldType.java
```diff
@@ -225,4 +227,85 @@ public float[] transformQueryVector(float[] vector) {
...
+        // support radial search (via rescoring). Non-quantized indices are always allowed.
+        final Optional<KNNMethodContext> methodContext = mappingConfig.getKnnMethodContext();
+        if (methodContext.isPresent()) {
+            // Check compression level first (cheap) before SQ encoder lookup (hash map access)
+            final boolean isQuantizedIndex = CompressionLevel.isConfigured(mappingConfig.getCompressionLevel())
+                && mappingConfig.getCompressionLevel() != CompressionLevel.x1
+                && mappingConfig.getCompressionLevel() != CompressionLevel.x2;
+            if (isQuantizedIndex) {
+                final boolean isFlatMethod = mappingConfig.getCompressionLevel() == CompressionLevel.x32
+                    && METHOD_FLAT.equals(methodContext.get().getMethodComponentContext().getName());
```
navneet1v: why isFlatMethod needs a compression level check?

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldType.java
```diff
@@ -225,4 +227,85 @@ public float[] transformQueryVector(float[] vector) {
...
+                final boolean isSQOneBit = isFlatMethod
+                    || FaissSQEncoder.isSQOneBit(methodContext.get().getMethodComponentContext().getParameters());
```
navneet1v: `isSQOneBit = FaissSQEncoder.isSQOneBit(methodContext.get().getMethodComponentContext().getParameters())`

not sure why we think `isFlatMethod` is needed in `or` condition

src/main/java/org/opensearch/knn/index/query/RNNQueryFactory.java
```diff
@@ -62,59 +61,94 @@ public static Query create(
...
-                .isMemoryOptimizedSearch(memoryOptimizedSearchEnabled)
-                .filterQuery(filterQuery)
-                .build();
+            innerQuery = createNativeEngineRadialQuery(createQueryRequest);
+        } else {
+            innerQuery = createLuceneRadialQuery(createQueryRequest);
         }
 
-        log.debug(String.format("Creating Lucene r-NN query for index: %s \"\", field: %s \"\", k: %f", indexName, fieldName, radius));
+        if (createQueryRequest.getVectorFieldType() != null && createQueryRequest.getVectorFieldType().isRescoringRequiredForRadial()) {
```
navneet1v, line 76: what will be the case when `createQueryRequest.getVectorFieldType()` will be null?

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldType.java
```diff
@@ -225,4 +227,85 @@ public float[] transformQueryVector(float[] vector) {
...
+     * </ul>
+     *
+     * <p>Uses the field's own {@code vectorDataType} and {@code knnMappingConfig} for validation.
+     * The engine must be passed explicitly because model-based fields resolve their engine via
+     * {@code ModelDao}, which is not available on this class.</p>
+     *
+     * @param knnEngine the engine resolved for this field (from method context or model metadata)
+     * @throws UnsupportedOperationException if radial search is not supported
+     */
+    public void validateSupportRadialSearch(final KNNEngine knnEngine) {
```
navneet1v, line 252: Please add UTs for the 2 functions we have in this class.

gradle/wrapper/gradle-wrapper.properties
0ctopus13prime: Since protection is enabled (not sure why they did it), I could not rebase it from my end. We need to open up a new PR and need to get 2 approvals on it. 
Hoping to skip that part

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldType.java
0ctopus13prime: To be conservative. 
Currently flat + 32x is the default, but things could change.

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldType.java
0ctopus13prime: It's needed in there, isSQOneBit is looking into method parameter, and `flat` does not have method parameter and Flat IT is failing.

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldType.java
0ctopus13prime: Sounds good, can I ship the comment in the next PR?
Hoping to close out this, it has been almost 2 weeks

gradle/wrapper/gradle-wrapper.properties
navneet1v: The problem is your changes in main will keep on getting accumulated. This is not a blocker for this PR. But something if want to talk to @peterzhuamazon and see if we can remove some restrictions from feature branches. As I said not a blocker.

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldType.java
navneet1v: Yes that is correct. Hence if you want to say isFlatMethod then you should just be checking method name is flat or not, that it.

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldType.java
0ctopus13prime: It's trade off I think.
Since the place calling each function is different, unless we pass `isRescoringAvailable` flag, the place calling `isRescoringRequiredForRadial` will need to call `validateSupportRadialSearch` (or we can put it inside of isRescoringRequiredForRadial)

Simple solution is to put `validateSupportRadialSearch` in isRescoringRequiredForRadial, but I wanted to avoid this and add the responsibility to caller.

src/main/java/org/opensearch/knn/index/query/RNNQueryFactory.java
0ctopus13prime: In normal case, I don't think field type is null. Just being conservative for the future bug that we might mistakenly bring in.

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldType.java
navneet1v: This is where we should take a step back and think like this:

`isSQOneBit is not == (flatMethod or SQ Encoder 1 bit). `

`isSQOneBit is == (SQ Encoder 1 bit). `

what you are looking here is `radialSearchSupported == (flatMethod or SQ Encoder 1 bit)`. I know it will look like just a variable name change, but what I want to put forward is this `isSQOneBit is not == (flatMethod or SQ Encoder 1 bit)`

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldType.java
0ctopus13prime: That's not the case. Code should be written to be strict especially for this predicate thing, now flat only supports 32x, we strictly impose such constrained exactly for flat 32x.

src/main/java/org/opensearch/knn/index/query/BaseQueryFactory.java
0ctopus13prime: `@Getter` is doing it.

gradle/wrapper/gradle-wrapper.properties
0ctopus13prime: Thank you for the understanding. Yeah, we need to talk this thing as it's a feature branch, and eventually we will have 2 approval when merging feature branch into main.

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldType.java
0ctopus13prime: sure

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldType.java
0ctopus13prime: ok, let me update it

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldType.java
```diff
@@ -225,4 +227,85 @@ public float[] transformQueryVector(float[] vector) {
...
+        if (ENGINES_SUPPORTING_RADIAL_SEARCH.contains(knnEngine) == false) {
+            throw new UnsupportedOperationException(String.format(Locale.ROOT, "Engine [%s] does not support radial search", knnEngine));
+        }
+        if (getVectorDataType() == VectorDataType.BINARY) {
+            throw new UnsupportedOperationException(String.format(Locale.ROOT, "Binary data type does not support radial search"));
```
VijayanB, line 257: Did we consider using Validation Exception? if it is not valid, update the method name so it use term "Supported" instead of "Validate"

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldType.java
Vikasht34, line 252: @0ctopus13prime  Thanks for putting this , I appreciate your effort and hard work !! By Looking at current code , I feel it will give you very hard time scaling to N BIT .... 

I would definitely gives thanks to @navneet1v for suggesting behind KnnFieldType and I understand his vision where he wanted to go , but looking at code , it seems we are closer but we still needs to do some work to make it right. 

Below code is not going to scale for combinations 

```
Today we have:

Engines:      2 (Faiss, Lucene)
Methods:      3 (HNSW, Flat, IVF)
Encoders:     5 (FaissSQ, FaissFlat, FaissPQ, LuceneSQ, QFrameBQ)
Bit widths:   5 (1, 2, 4, 7, 16)


That's 2 × 3 × 5 × 5 = 150 possible combinations. 
The current code handles ~6 of them with hardcoded if branches. Every new combination requires modifying this method.
```


```
INDEX CREATION TIME
═══════════════════

User creates index with mapping
        │
        ▼
┌─────────────────────┐
│   MethodResolver    │  (FaissMethodResolver / LuceneHNSWMethodResolver / LuceneFlatMethodResolver)
│                     │
│  resolves encoder   │
│  calls encoder:     │
│    .calculateCompressionLevel()
│    .supportsRadialSearch()    ← NEW
│    .requiresRescore()         ← NEW
└─────────┬───────────┘
          │
          ▼
┌─────────────────────────┐
│  ResolvedMethodContext  │
│                         │
│  - knnMethodContext     │
│  - compressionLevel     │
│  - supportsRadialSearch │  ← NEW boolean
│  - requiresRescore      │  ← NEW boolean
└─────────┬───────────────┘
          │
          ▼
┌─────────────────────┐
│  Engine  │  (builds the KNNMappingConfig anonymous class)
│                     │
│  sets all fields    │
│  from resolved ctx  │
└─────────┬───────────┘
          │
          ▼
┌─────────────────────┐
│   KNNMappingConfig  │  (interface, stored on field type)
│                     │
│  .getCompressionLevel()
│  .getQuantizationConfig()
│  .supportsRadialSearch()  ← reads stored boolean
│  .requiresRescore()       ← reads stored boolean
└─────────┬───────────┘
          │
          ▼
┌─────────────────────────┐
│   KNNVectorFieldType    │  (the field type stored in the index mapping)
│                         │
│  .getKnnMappingConfig() │
│  .validateSupportRadialSearch()  ← just reads boolean
│  .isRescoringRequiredForRadial() ← just reads boolean
└─────────────────────────┘


QUERY TIME
══════════

User sends radial search query
        │
        ▼
┌──────────────────────┐
│  KNNQueryBuilder     │
│  .doToQuery()        │
│                      │
│  1. Gets KNNVectorFieldType from shard context
│  2. Calls fieldType.validateSupportRadialSearch(engine)
│     → reads boolean → throws if false
│  3. Builds inner query via RNNQueryFactory.create()
│  4. Checks fieldType.isRescoringRequiredForRadial()
│     → reads boolean
│  5. If true: wraps with RescoreRadialSearchQuery
│     If false: returns inner query directly
└──────────────────────┘

```


```
## How it scales — every future scenario

Cases:
FaissSQ(bits=1)  → supportsRadial=true,  requiresRescore=true
FaissSQ(bits=16) → supportsRadial=true,  requiresRescore=false
LuceneSQ(bits=1) → supportsRadial=true,  requiresRescore=true
LuceneSQ(bits=7) → supportsRadial=false, requiresRescore=false
QFrameBQ         → supportsRadial=false, requiresRescore=false
FaissPQ          → supportsRadial=false, requiresRescore=false
FaissFlat        → supportsRadial=true,  requiresRescore=false
LuceneFlat(x32)  → supportsRadial=true,  requiresRescore=true




Case1 1: Lucene 4-bit (x4) gets rescoring support

Change ONE method in LuceneSQEncoder:
java
if (b == 7) return true;  // in supportsRadialSearch()
if (b == 7) return true;  // in requiresRescore()

Nothing else changes. Query path, validation, RNNQueryFactory — all untouched.

Case2 2: Lucene 4-bit works natively (no rescore needed)

java
if (b == 7) return true;   // supportsRadialSearch
if (b == 7) return false;  // requiresRescore — not needed!

Radial search runs without the RescoreRadialSearchQuery wrapper. Zero other code changes.



Case 3: New encoder FaissIVFSQEncoder added

Just implement the interface:
java
class FaissIVFSQEncoder implements Encoder {
    // ... existing methods ...

    @Override
    public boolean supportsRadialSearch(...) { return true; }

    @Override
    public boolean requiresRescore(...) { return true; }
}

Radial search works automatically. No changes to KNNVectorFieldType, KNNQueryBuilder, RNNQueryFactory, or any validation code.


Case4 4: BQ (QFrame) gets radial support someday

Change ONE method in QFrameBitEncoder:
java
@Override
public boolean supportsRadialSearch(...) { return bits == 1; } // only 1-bit BQ

Done. The QuantizationConfig != EMPTY check that currently blocks all BQ is gone — each encoder decides for itself.



Case5 : Top-k also wants to know if rescoring is needed

requiresRescore() is already there — it's not radial-specific. Top-k path can read the same boolean:
java
if (knnVectorFieldType.getKnnMappingConfig().requiresRescore()) {
    // use NativeEngineKnnVectorQuery with rescore
}




Case 6: Someone adds a completely new quantization method (e.g., "xyz")

They create xyzEncoder implements Encoder. The compiler forces them to implement calculateCompressionLevel(). The defaults for supportsRadialSearch() and requiresRescore() are both false — safe by default. Radial is blocked until they 
explicitly opt in. No silent bugs.


## Why the current code doesn't scale

Every scenario above would require modifying KNNVectorFieldType.validateSupportRadialSearch() — adding more if branches, more isSomethingSpecific() checks, more knowledge about encoder internals leaking into the field type. The method grows into
a god-method that knows about every encoder.


public void validateSupportRadialSearch(final KNNEngine knnEngine) {
    if (ENGINES_SUPPORTING_RADIAL_SEARCH.contains(knnEngine) == false) {    // 1
        throw ...
    }
    if (getVectorDataType() == VectorDataType.BINARY) {                     // 2
        throw ...
    }
    if (mappingConfig.getQuantizationConfig() != QuantizationConfig.EMPTY) { // 3
        throw ...
    }
    if (methodContext.isPresent()) {                                         // 4
        if (isQuantizedIndex) {                                             // 5 (compound: isConfigured AND != x1 AND != x2)
            // isConfigured                                                 // 6
            // != x1                                                        // 7
            // != x2                                                        // 8
            if (isQuantizedIndex) {
                final boolean isFlatMethod = ... == x32 && METHOD_FLAT...   // 9, 10
                final boolean isSQOneBit = isFlatMethod || FaissSQ...       // 11
                if (isSQOneBit == false) {                                  // 12
                    throw ...
                }
            }
        }
    }
}


Current cyclomatic complexity: ~12

Plus isRescoringRequiredForRadial():
java
public boolean isRescoringRequiredForRadial() {
    if (methodContext.isPresent() == false) {  // 1
        return false;
    }
    return FaissSQEncoder.isSQOneBit(...)      // 2 (internal: null check + name check + bits check = +3)
}


That method: ~5

Total for radial logic: ~17

and this is going to increase if we have to put combinations 
```

0ctopus13prime: @Vikasht34 
This is the closed loop introduced on purpose to ensure only available cases to support radial search.
So far, only SQ 1bit is supporting radial search, and the condition is making that super clear. 
I don't get why it needs to be scaled. In the future, if we introduce a new quantized index type to support radial, then of course we need the change.

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldType.java
0ctopus13prime: This is the existing code, I just moved to here.
I think `UnsupportedOperationException` will be easier for user to understand. Also I don't want to change user experience on it.

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldType.java
Vikasht34: I agree , but for future changes I would not like to add any if condition in this class, that's what I meant. It feels like right now we removed code from one place to here.

0ctopus13prime: Yeah, I think Navneet has a plan for that to make it scale. 
For now, it should be good start with SQ 1bit within the method.

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldType.java
Vikasht34: how we will be scalled to if we have flkat +8x , does not requires but flat + 16x does . If you see I can go ahread and add another if condition , then that's not scalling , it's increasing complexcity for code.

0ctopus13prime: Navneet has the plan for that, he is planning to adding a new enum. 
But for now, checking SQ 1 bit in the method is fine I believe. Without clear timeline when we introduce 16x and whatnot, change the code for that seems overkill

Vikasht34: we should design for open for extension not for closed for extension , https://en.wikipedia.org/wiki/Open%E2%80%93closed_principle. Right now it's closed.

0ctopus13prime: That's the trade off. I respect OOP principal, but we can consider starting small and gradually make it scalable. 
Theory is good, but some time we could find a practical approach. 
For now, only SQ 1 bit supports radial, we can explicitly limit those cases at the beginning.

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldType.java
VijayanB: Are you changing the method name then?

0ctopus13prime: No, it will be kept as it is. 
Throwing an exception is giving useful info for user. I think it would be better to keep it as it as it makes very clear that `it's not supported` and user will get the sense of it immediately.

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
```diff
@@ -479,20 +476,7 @@ protected Query doToQuery(QueryShardContext context) {
...
-            if (vectorDataType == VectorDataType.BINARY) {
-                throw new UnsupportedOperationException(String.format(Locale.ROOT, "Binary data type does not support radial search"));
-            }
-
-            if ((knnMappingConfig.getQuantizationConfig() != QuantizationConfig.EMPTY)
-                // If compression level is 32x, then radial search should be blocked.
-                || (knnMappingConfig.getCompressionLevel() == CompressionLevel.x32)) {
-                throw new UnsupportedOperationException("Radial search is not supported for indices which have quantization enabled");
-            }
+            knnVectorFieldType.validateSupportRadialSearch(knnEngine);
```
VijayanB, line 479: @navneet1v is this aligned with your suggestion that we move validation there ? IMO it looks disconnected from rest of the validation that we added in this method.

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldType.java
Vikasht34: I would let @navneet1v @shatejas @vamshin   to give opinion.

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
0ctopus13prime: Yes, this is aligned where letting the field type decide whether supporting radial search. 
What's disconnected?

src/main/java/org/opensearch/knn/index/query/BaseQueryFactory.java
```diff
@@ -50,8 +51,11 @@ public static class CreateQueryRequest {
         private float[] vector;
         private float[] originalVector;
         private byte[] byteVector;
+        // TODO : This needs clean-up, we have vectorFieldType.
```
VijayanB: Could you expand this TODO with more context so that someone else can also pick this up in the future?

0ctopus13prime: Yeah, will do in the next PR if you're fine.

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldType.java
Vikasht34: I would let @navneet1v  @shatejas  @vamshin to give opinion right way to put forward , but I feel current code right now is going to have major change if we will be supporting others, right now I am bit lost in nested if else.

src/main/java/org/opensearch/knn/index/query/BaseQueryFactory.java
VijayanB: Why spill this to next PR, lets do it in this. so that we can close out now and your next PR will be very clean

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldType.java
kotwanikunal: @Vikasht34 @0ctopus13prime Good discussion. I agree with Vikash's scaling concern. The current `validateSupportRadialSearch()` has 10+ branches of encoder-specific logic sitting inside `KNNVectorFieldType`. Every new encoder/bit-width combo means modifying this method. It doesn't scale.

The problem is `KNNVectorFieldType` knows too much about encoder internals. It shouldn't be reasoning about `isSQOneBit`,  `isFlatMethod`, `CompressionLevel != xx`, or `QuantizationConfig.EMPTY`. Those are encoder implementation details leaking into the field type.

kotwanikunal: What I'd propose instead: capability delegation

Encoder declares its own capabilities. Each encoder answers "do I support radial?" and "do I need rescore?" based on its own config. The encoder owns that knowledge.

Engine declares engine-level capabilities (already exists as ENGINES_SUPPORTING_RADIAL_SEARCH).

`KNNVectorFieldType` just delegates. It asks the encoder and engine, combines the answers. No encoder-specific branches, no compression-level proxies.

```
 // KNNVectorFieldType -- thin orchestrator, no encoder knowledge
 public void validateSupportRadialSearch(KNNEngine engine) {
     if (!engine.supportsRadialSearch()) throw ...;
     if (vectorDataType == BINARY) throw ...;
     if (!encoder.supportsRadialSearch(encoderCtx)) throw ...;
 }
 
```

When someone adds a 4-bit encoder that supports radial natively, they implement `supportsRadialSearch() -> true` on their encoder class. `KNNVectorFieldType` is untouched.

I'm working on this as part of the 32x default refactoring (#3290), specifically making the Encoder interface declare its capabilities rather than having query-time code infer them through compression-level proxies.

I am not sure if it makes sense to refactor it right now - but a TODO should call this out as temporary code. As is - this is adding in debt with the merge.

Vikasht34: that's very right way to think about @kotwanikunal , that's what I am proposing ..

navneet1v: Let call this as a temporary for now, since @kotwanikunal is working on the refactor to ensure that we can better capture the capabilities. @kotwanikunal The direction you are suggesting I partially agree on that. Something like rescore can be part of encoder, right now we make it as part of compression level(which in a way is proxy to all the different encoders[engine specific encoders]). but a specific search capability is always part of Engine/method. IMO Encoder should not include this. We can talk about this in your refactor PR more.


But for now just lets just close out on this discussion on refactoring and move on in the PR, I hope everyone is aligned on this thought(@kotwanikunal , @VijayanB, @Vikasht34 ). @0ctopus13prime other concerns related to ordering of the function still remains open even though we are closing out on this discussion.

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
navneet1v: @VijayanB yes you are right, its little disconnected right now. But to keep scope of PR to radial search we are doing this. With @kotwanikunal refactoring the code for default 32x compression we should be able to bring consistency in topK path with those PRs.

src/main/java/org/opensearch/knn/index/query/RNNQueryFactory.java
navneet1v: I would suggest to avoid defensive programming here. You can add a @NonNull check directly on the member variable of the class.

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldType.java
navneet1v: @0ctopus13prime 
```
mappingConfig.getCompressionLevel() == CompressionLevel.x32
                    && METHOD_FLAT.equals(methodContext.get().getMethodComponentContext().getName());
```

having compression as 32x is not a necessary condition for Method being flat. I agree we may have made it mandatory in the mapping, but that is not a necessary condition for search. Its a way for us to block users adding other compression type. but in the code we should not assume that isFlatMethod would mean we should have 32x compression + method name as flat.

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldType.java
navneet1v: > it has been almost 2 weeks

I won't suggest this to be a reason to move it to next rev. I know you will take care of it in next PR, but as a reviewer its hard to keep track of these things. Moreover there will be another revision for this PR, so lets ensure that we can close out as many comments as possible.

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldType.java
navneet1v: I think I understand your point here. Since we are in the process of building the Radial Search query and to complete the build the process we would need to do 2 things:
1. validate if query is supported. 
2. If its is supported then check if rescoring required or not. 

Since both of them are called by `RNNQueryBuilderFactory`, then it make sense.

src/main/java/org/opensearch/knn/index/query/RNNQueryFactory.java
0ctopus13prime: That's the choice to take, I would rather do explicit check whether it's non null

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldType.java
0ctopus13prime: Will do
But let's think about what is the practical way in this situation. We can do things much smater. You can delegate author of this PR to track of things. With this development velocity, this PR will be closed out in next Christmas .

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldType.java
0ctopus13prime: @navneet1v 
It's necessary condition for now.
`flat` method always uses 32x compression underlying, you agreed upon this. Therefore it must be the necessary condition for the search as this is indexing parameter validation, we impose 32x compression during the index. 

I really don't understand why this is blocker. `flat` with non 32x compression should be blocked for radial search.

navneet1v: > flat with non 32x compression should be blocked for radial search.

Would like to know why we think that? if you say that hey `Flat` with 16x compression right now fallback to BQ hence we should block it then I am with you.

0ctopus13prime: Will remove 32x compression.

src/main/java/org/opensearch/knn/index/query/RNNQueryFactory.java
0ctopus13prime: Will add @NonNull

0ctopus13prime: Turns out, not all path is setting field type in the code. 
Radial search, it's setting field type in the flow, but for the rest, it's not setting it. 
Will keep it as it is.

src/main/java/org/opensearch/knn/index/query/RescoreRadialSearchQuery.java
0ctopus13prime: Discussed in offline. I could not find a clear path to use ExactSearcher with FunctionScoreQuery.

navneet1v (approved):

Vikasht34 (approved):

PR #3302, opened 2026-04-30
Use KNN1040ScalarQuantizedVectorsFormat for Faiss SQ flat format

VijayanB:
### Description
The Faiss SQ format was using Lucene's `Lucene104ScalarQuantizedVectorsFormat` directly, which lacks the prefetch-enabled raw vector reader that `KNN1040ScalarQuantizedVectorsFormat` provides. This meant exact search rescoring was missing I/O prefetch during graph traversal.

### Changes
- Switch `faissSqFlatFormat` from `Lucene104ScalarQuantizedVectorsFormat` to `KNN1040ScalarQuantizedVectorsFormat` in `Faiss1040ScalarQuantizedKnnVectorsFormat`
- Add `@VisibleForTesting` package-private accessors: `getFaissSqFlatFormat()` on `Faiss1040ScalarQuantizedKnnVectorsFormat` and `getFlatVectorsReader()` on `Faiss1040ScalarQuantizedKnnVectorsReader`

### Testing
Verified the test catches the bug: when `KNN1040ScalarQuantizedVectorsFormat` is replaced with `Lucene104ScalarQuantizedVectorsFormat`, the test fails with:
```
Scorer from Faiss SQ reader should be PrefetchableRandomVectorScorer, but was: ScalarQuantizedRandomVectorScorer
```

### Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 * Update Gradle wrapper to 9.4.1 and Jacoco to 0.8.14 to match core OpenSearch [#3308](https://github.com/opensearch-project/k-NN/pull/3308)
 
 ### Bug Fixes
+* Use KNN1040ScalarQuantizedVectorsFormat for Faiss SQ flat format to enable prefetch [#3302](https://github.com/opensearch-project/k-NN/pull/3302)
 
 ### Refactoring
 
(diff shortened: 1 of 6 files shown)


navneet1v:
Please add the UTs that validate that right reader is getting used and the flat vector format is using right scoer

VijayanB:
Addressing the bot's review suggestions:

**1. Null Codec in `SegmentInfo`** — This is standard practice in Lucene unit tests. The `codec` parameter is only accessed when the full codec pipeline is invoked (e.g., via `SegmentReader`). Here we only construct `SegmentWriteState`/`SegmentReadState` to pass directly to the flat vectors format writer/reader — the codec field is never dereferenced. Lucene's own test suite uses the same pattern.

**2. Unclosed Writer** — The writer is already in a try-with-resources block, so it's always closed even if `flush()`/`finish()` throws. This is a test utility writing to a temp directory that gets cleaned up by the test framework. No additional error handling needed.

**3. Static Accessor Visibility** — `KNN1040ScalarQuantizedVectorsFormat` is stateless (it only holds an encoding enum value). The `@VisibleForTesting` annotation correctly communicates intent — tests need the shared format instance to write vectors using the same configuration as production. There's nothing mutable to corrupt.

VijayanB:
The CI failure (`Build and Test k-NN Plugin on MacOS (25)`) is unrelated to this PR. OpenSearch `build-tools:3.7.0-SNAPSHOT` now requires Gradle 9.4.1+ via `GlobalBuildInfoPlugin`, but the k-NN wrapper was still at 9.2.0. Fix: #3307 — rebase after it merges.

navneet1v (approved):

naveentatikonda (approved):

PR #3313, opened 2026-05-07
Fix derived source for mixed-case vector fields

### LINKED ISSUE #3312: k-NN derived source placeholder can fail replica recovery for knn_vector fields
### Description

When `index.knn.derived_source.enabled=true` is enabled on an index with `knn_vector` fields, replica shard recovery can fail during translog replay. The primary shard can continue indexing/searching successfully, but a replica remains unassigned and the cluster/index stays yellow.

The recovery failure appears to happen because a derived-source placeholder value such as `<object>.<vectorField>: 1` is replayed through the normal `knn_vector` mapper as if it were the actual vector.

### Environment

Observed with:

- OpenSearch 3.6.0
- opensearch-knn 3.6.0.0
- Multi-node cluster with replicas
- Index includes one or more `knn_vector` fields with dimension 384
- Relevant index settings included:

```json
{
  "index.knn": "true",
  "index.knn.derived_source.enabled": "true",
  "index.replication.type": "DOCUMENT",
  "index.derived_source.enabled": "false",
  "index.derived_source.translog.enabled": "false"
}
```

### Observed behavior

Replica allocation fails during recovery/translog replay with an error shape like:

```text
RecoveryEngineException[Phase[2] failed to send/replay operations]
internal:index/shard/recovery/translog_ops
MapperParsingException[failed to parse field [<object>.<vectorField>] of type [knn_vector]]
IllegalArgumentException[Vector dimension mismatch. Expected: 384, Given: 1]
```

Application indexing did not send `1` for the vector field. The `1` appears to be the k-NN derived-source placeholder, and it seems to be treated as the real vector during replica recovery.

### Expected behavior

Replica recovery should be able to replay/recover documents for indexes that use k-NN derived source, or should reconstruct the vector before the value reaches the `knn_vector` mapper. At minimum, if this setting combination is unsupported, the index should fail fast or the limitation should be documented.

### Workaround

Setting `index.knn.derived_source.enabled=false` for new indexes avoids the recovery failure. Existing affected indexes need to be recreated/reindexed with that setting disabled before replicas recover cleanly.

### Additional notes

This does not look like a general indexing or search failure: search can continue to work against the primary shard. The visible symptom is shard allocation failure/yellow health caused by replica recovery failing on the derived-source placeholder.

leewjae:
## Summary

- Preserve exact k-NN vector field names when writing derived-source segment attributes.
- Resolve previously-written lowercase segment attributes through an unambiguous case-insensitive `FieldInfo` fallback.
- Add unit and integration regression coverage for mixed-case object vector fields.

## Issue

Fixes #3312

## Testing

- `./gradlew :spotlessCheck`
- `./gradlew :test -x buildJniTest -x buildJniLib -x cmakeJniLib --tests org.opensearch.knn.index.codec.KNN10010Codec.DerivedSourceStoredFieldsWriterTests --tests org.opensearch.knn.index.codec.KNN10010Codec.DerivedSourceStoredFieldsFormatTests`
- `./gradlew :integTest -x buildJniLib -x cmakeJniLib -x integTestMultiNode --tests org.opensearch.knn.integ.DerivedSourceIT.testDerivedSource_withMixedCaseObjectVectorField`

### DIFF
--- a/src/main/java/org/opensearch/knn/index/codec/KNN10010Codec/KNN10010DerivedSourceStoredFieldsFormat.java
+++ b/src/main/java/org/opensearch/knn/index/codec/KNN10010Codec/KNN10010DerivedSourceStoredFieldsFormat.java
@@ -104,13 +104,19 @@ public class KNN10010DerivedSourceStoredFieldsFormat extends StoredFieldsFormat 
                 continue;
             }
             if (matchedFieldInfo != null) {
-                log.warn(
-                    "Skipping derived vector field [{}] because field infos [{}] and [{}] both match case-insensitively",
-                    field,
-                    matchedFieldInfo.name,
-                    candidate.name
-                );
-                return null;
+                if (matchedFieldInfo.hasVectorValues() && candidate.hasVectorValues()) {
+                    log.warn(
+                        "Skipping derived vector field [{}] because field infos [{}] and [{}] both match case-insensitively",
+                        field,
+                        matchedFieldInfo.name,
+                        candidate.name
+                    );
+                    return null;
+                }
+                if (candidate.hasVectorValues()) {
+                    matchedFieldInfo = candidate;
+                }
+                continue;
             }
             matchedFieldInfo = candidate;
         }
--- a/src/main/java/org/opensearch/knn/index/codec/KNN10010Codec/KNN10010DerivedSourceStoredFieldsFormat.java
+++ b/src/main/java/org/opensearch/knn/index/codec/KNN10010Codec/KNN10010DerivedSourceStoredFieldsFormat.java
@@ -11,6 +11,7 @@ import org.apache.lucene.codecs.Codec;
 import org.apache.lucene.codecs.StoredFieldsFormat;
 import org.apache.lucene.codecs.StoredFieldsReader;
 import org.apache.lucene.codecs.StoredFieldsWriter;
+import org.apache.lucene.index.FieldInfo;
 import org.apache.lucene.index.FieldInfos;
 import org.apache.lucene.index.SegmentInfo;
 import org.apache.lucene.index.SegmentReadState;
@@ -48,12 +49,10 @@ public class KNN10010DerivedSourceStoredFieldsFormat extends StoredFieldsFormat 
         List<DerivedFieldInfo> derivedVectorFields = Stream.concat(
             DerivedSourceSegmentAttributeParser.parseDerivedVectorFields(segmentInfo, false)
                 .stream()
-                .filter(field -> fieldInfos.fieldInfo(field) != null)
-                .map(field -> new DerivedFieldInfo(fieldInfos.fieldInfo(field), false)),
+                .flatMap(field -> getDerivedFieldInfo(fieldInfos, field, false)),
             DerivedSourceSegmentAttributeParser.parseDerivedVectorFields(segmentInfo, true)
                 .stream()
-                .filter(field -> fieldInfos.fieldInfo(field) != null)
-                .map(field -> new DerivedFieldInfo(fieldInfos.fieldInfo(field), true))
+                .flatMap(field -> getDerivedFieldInfo(fieldInfos, field, true))
         ).toList();
 
         // If no fields have it enabled, we can just short-circuit and return the delegate's fieldReader
@@ -79,6 +78,33 @@ public class KNN10010DerivedSourceStoredFieldsFormat extends StoredFieldsFormat 
         } else {
             return delegate;
         }
+    }
+
+    private Stream<DerivedFieldInfo> getDerivedFieldInfo(FieldInfos fieldInfos, String field, boolean isNested) {
+        FieldInfo fieldInfo = getFieldInfo(fieldInfos, field);
+        if (fieldInfo == null) {
+            return Stream.empty();
+        }
+        return Stream.of(new DerivedFieldInfo(fieldInfo, isNested));
+    }
+
+    static FieldInfo getFieldInfo(FieldInfos fieldInfos, String field) {
+        FieldInfo fieldInfo = fieldInfos.fieldInfo(field);
+        if (fieldInfo != null) {
+            return fieldInfo;
+        }
+
+        FieldInfo matchedFieldInfo = null;
+        for (FieldInfo candidate : fieldInfos) {
+            if (candidate.name.equalsIgnoreCase(field) == false) {
+                continue;
+            }
+            if (matchedFieldInfo != null) {
+                return null;
+            }
+            matchedFieldInfo = candidate;
+        }
+        return matchedFieldInfo;
     }
 
     @Override
--- a/src/main/java/org/opensearch/knn/index/codec/KNN10010Codec/KNN10010DerivedSourceStoredFieldsFormat.java
+++ b/src/main/java/org/opensearch/knn/index/codec/KNN10010Codec/KNN10010DerivedSourceStoredFieldsFormat.java
@@ -5,7 +5,9 @@
 
 package org.opensearch.knn.index.codec.KNN10010Codec;
 
+import com.google.common.annotations.VisibleForTesting;
 import lombok.AllArgsConstructor;
+import lombok.extern.log4j.Log4j2;
 
 import org.apache.lucene.codecs.Codec;
 import org.apache.lucene.codecs.StoredFieldsFormat;
@@ -30,6 +32,7 @@ import java.util.List;
 import java.util.stream.Stream;
 
 @AllArgsConstructor
+@Log4j2
 public class KNN10010DerivedSourceStoredFieldsFormat extends StoredFieldsFormat {
     // Stores the delegate codec name (in case it is different from the default one)
     static final String KNN_DELEGATE_CODEC_NAME = "knn_delegate_stored_fields_codec_key";
@@ -88,6 +91,7 @@ public class KNN10010DerivedSourceStoredFieldsFormat extends StoredFieldsFormat 
         return Stream.of(new DerivedFieldInfo(fieldInfo, isNested));
     }
 
+    @VisibleForTesting
     static FieldInfo getFieldInfo(FieldInfos fieldInfos, String field) {
         FieldInfo fieldInfo = fieldInfos.fieldInfo(field);
         if (fieldInfo != null) {
@@ -100,6 +104,12 @@ public class KNN10010DerivedSourceStoredFieldsFormat extends StoredFieldsFormat 
                 continue;
             }
             if (matchedFieldInfo != null) {
+                log.warn(
+                    "Skipping derived vector field [{}] because field infos [{}] and [{}] both match case-insensitively",
+                    field,
+                    matchedFieldInfo.name,
+                    candidate.name
+                );
                 return null;
             }
             matchedFieldInfo = candidate;
--- a/src/main/java/org/opensearch/knn/index/codec/KNN10010Codec/KNN10010DerivedSourceStoredFieldsFormat.java
+++ b/src/main/java/org/opensearch/knn/index/codec/KNN10010Codec/KNN10010DerivedSourceStoredFieldsFormat.java
@@ -102,9 +102,14 @@ public class KNN10010DerivedSourceStoredFieldsFormat extends StoredFieldsFormat 
             if (candidate.name.equalsIgnoreCase(field) == false) {
                 continue;
             }
-            if (matchedFieldInfo == null || candidate.hasVectorValues() && !matchedFieldInfo.hasVectorValues()) {
+            if (matchedFieldInfo == null) {
                 matchedFieldInfo = candidate;
-            } else if (matchedFieldInfo.hasVectorValues() && candidate.hasVectorValues()) {
+                continue;
+            }
+
+            boolean matchedHasVectorValues = matchedFieldInfo.hasVectorValues();
+            boolean candidateHasVectorValues = candidate.hasVectorValues();
+            if (matchedHasVectorValues == candidateHasVectorValues) {
                 log.warn(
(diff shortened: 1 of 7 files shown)


src/main/java/org/opensearch/knn/index/codec/KNN10010Codec/KNN10010DerivedSourceStoredFieldsWriter.java
```diff
@@ -140,7 +140,7 @@ private void refreshKnownVectorFields() {
         for (MappedFieldType fieldType : mapperService.fieldTypes()) {
             if (fieldType instanceof KNNVectorFieldType knnVectorFieldType) {
                 if (IndexUtil.isDerivedEnabledForField(knnVectorFieldType, mapperService)) {
-                    String fieldName = fieldType.name().toLowerCase();
+                    String fieldName = fieldType.name();
```
navneet1v, line 143: will this cause any BWC issue? we should add a BWC test here to ensure that there is no BWC issue.

src/main/java/org/opensearch/knn/index/codec/KNN10010Codec/KNN10010DerivedSourceStoredFieldsFormat.java
```diff
@@ -81,6 +80,33 @@ private StoredFieldsFormat getStoredFieldsFormat(final SegmentInfo segmentInfo)
...
 
+    private Stream<DerivedFieldInfo> getDerivedFieldInfo(FieldInfos fieldInfos, String field, boolean isNested) {
+        FieldInfo fieldInfo = getFieldInfo(fieldInfos, field);
+        if (fieldInfo == null) {
+            return Stream.empty();
+        }
+        return Stream.of(new DerivedFieldInfo(fieldInfo, isNested));
+    }
+
+    static FieldInfo getFieldInfo(FieldInfos fieldInfos, String field) {
```
navneet1v, line 94: why this is package private>?

src/main/java/org/opensearch/knn/index/codec/KNN10010Codec/KNN10010DerivedSourceStoredFieldsWriter.java
leewjae: Added restart-upgrade BWC coverage in `DerivedSourceBWCRestartIT.testMixedCaseDerivedSourceField_indexOnOld_reconstructOnNew`. The test creates the mixed-case derived-source field on the old cluster, refreshes/flushes that segment, then validates after upgrade that `_source` reconstructs `vectorSearch.nameVector` from the old lowercase segment metadata.

src/main/java/org/opensearch/knn/index/codec/KNN10010Codec/KNN10010DerivedSourceStoredFieldsFormat.java
leewjae: Good point. This is package-private only to unit-test the field resolution behavior directly, so I marked it with `@VisibleForTesting`. I also added a warning when the case-insensitive fallback is ambiguous, and tightened the writer unit test to assert the mocked field is actually derived-source eligible.

src/main/java/org/opensearch/knn/index/codec/KNN10010Codec/KNN10010DerivedSourceStoredFieldsFormat.java
```diff
@@ -81,6 +83,40 @@ private StoredFieldsFormat getStoredFieldsFormat(final SegmentInfo segmentInfo)
         }
     }
 
+    private Stream<DerivedFieldInfo> getDerivedFieldInfo(FieldInfos fieldInfos, String field, boolean isNested) {
```
shatejas: nit : I wouldn't recommend wrapping this with stream, small stream often have an overhead

src/main/java/org/opensearch/knn/index/codec/KNN10010Codec/KNN10010DerivedSourceStoredFieldsFormat.java
```diff
@@ -81,6 +83,40 @@ private StoredFieldsFormat getStoredFieldsFormat(final SegmentInfo segmentInfo)
...
+        if (fieldInfo != null) {
+            return fieldInfo;
+        }
+
+        FieldInfo matchedFieldInfo = null;
+        for (FieldInfo candidate : fieldInfos) {
+            if (candidate.name.equalsIgnoreCase(field) == false) {
+                continue;
+            }
+            if (matchedFieldInfo != null) {
```
shatejas: In case of conflicts you can have an added check if `fieldInfo.hasVectorValues()` and return null only if both are true

leewjae: Implemented in 219e72f5. We now return null only when both case-insensitive matches have vector values; when only one match has vector values, we select that one. Also added a unit test covering the vector-vs-nonvector conflict case.

src/main/java/org/opensearch/knn/index/codec/KNN10010Codec/KNN10010DerivedSourceStoredFieldsFormat.java
leewjae: Addressed in 53083cfd by replacing the small stream wrappers with direct list/loop logic for derived field collection.

src/main/java/org/opensearch/knn/index/codec/KNN10010Codec/KNN10010DerivedSourceStoredFieldsFormat.java
```diff
@@ -81,6 +82,46 @@ private StoredFieldsFormat getStoredFieldsFormat(final SegmentInfo segmentInfo)
...
+                if (matchedFieldInfo.hasVectorValues() && candidate.hasVectorValues()) {
+                    log.warn(
+                        "Skipping derived vector field [{}] because field infos [{}] and [{}] both match case-insensitively",
+                        field,
+                        matchedFieldInfo.name,
+                        candidate.name
+                    );
+                    return null;
+                }
+                if (candidate.hasVectorValues()) {
```
shatejas: This might not work for pre 2.16 indices, where vectors were stored in binary doc values. Having said that I am not sure if any index will have vectors in binary doc values considering its supported starting 2.19. @navneet1v can you help verify if BDV is supported here.

If we need support for binary doc values, then we need to remove this if condition and do a best attempt with `hasVectorValues` before returning null.

leewjae: Implemented in 845df71b. Updated `getFieldInfo` to do a best-effort selection with `hasVectorValues` and return `null` when both candidates report the same vector hint (both true or both false). This avoids guessing when there are no vector hints (e.g., potential BDV-era metadata). Added `testGetFieldInfoDoesNotGuessWhenCaseInsensitiveMatchHasNoVectorHints` to cover the ambiguous both-false path.

src/main/java/org/opensearch/knn/index/codec/KNN10010Codec/KNN10010DerivedSourceStoredFieldsFormat.java
```diff
@@ -81,6 +82,48 @@ private StoredFieldsFormat getStoredFieldsFormat(final SegmentInfo segmentInfo)
...
+                    return null;
+                }
+                if (candidateHasVectorValues) {
+                    matchedFieldInfo = candidate;
+                }
+                continue;
+            }
+            matchedFieldInfo = candidate;
+        }
+        return matchedFieldInfo;
```
shatejas, line 117: This seems a bit complex to read, can we please simplify this, let me know what you think
```suggestion
        for (FieldInfo candidate : fieldInfos) {
            if (candidate.name.equalsIgnoreCase(field) == false) {
                continue;
            }
            if (matchedFieldInfo == null || candidate.hasVectorValues() && !matchedFieldInfo.hasVectorValues()) {
                matchedFieldInfo = candidate;
             } else if (matchedFieldInfo.hasVectorValues() && candidate.hasVectorValues()) { 
                    log.warn(
                        "Skipping derived vector field [{}] because field infos [{}] and [{}] both match case-insensitively",
                        field,
                        matchedFieldInfo.name,
                        candidate.name
                    );
                    return null;
            }
        }
        return matchedFieldInfo;
```

leewjae: Sure.

shatejas:
Thanks @leewjae! This looks good, waiting for CI runs to approve

shatejas:
@leewjae please run `./gradlew :spotlessApply to fox formatting`

leewjae:
> @leewjae please run `./gradlew :spotlessApply to fox formatting`
@shatejas 
done thanks!

leewjae:
Correction: a2b85d0 restores getFieldInfo ambiguous both-false handling (returns null when case-insensitive matches have equal vector hints). Commit 2509f18 removed that path, which is why DerivedSourceStoredFieldsFormatTests.testGetFieldInfoDoesNotGuessWhenCaseInsensitiveMatchHasNoVectorHints failed in CI. The fix still keeps best-effort preference when exactly one candidate has vector hints.

shatejas:
> Correction: [a2b85d0](https://github.com/opensearch-project/k-NN/commit/a2b85d0ee068168370f41d1b6046163cee38fa0c) restores getFieldInfo ambiguous both-false handling (returns null when case-insensitive matches have equal vector hints). Commit [2509f18](https://github.com/opensearch-project/k-NN/commit/2509f182dc49cde0079913ddb2957d9e137eafe8) removed that path, which is why DerivedSourceStoredFieldsFormatTests.testGetFieldInfoDoesNotGuessWhenCaseInsensitiveMatchHasNoVectorHints failed in CI. The fix still keeps best-effort preference when exactly one candidate has vector hints.

```else if (matchedFieldInfo.hasVectorValues() == candidate.hasVectorValues()) { ``` instead of && would have done the trick I think

**Comment by leewjae** [comment]
> > Correction: [a2b85d0](https://github.com/opensearch-project/k-NN/commit/a2b85d0ee068168370f41d1b6046163cee38fa0c) restores getFieldInfo ambiguous both-false handling (returns null when case-insensitive matches have equal vector hints). Commit [2509f18](https://github.com/opensearch-project/k-NN/commit/2509f182dc49cde0079913ddb2957d9e137eafe8) removed that path, which is why DerivedSourceStoredFieldsFormatTests.testGetFieldInfoDoesNotGuessWhenCaseInsensitiveMatchHasNoVectorHints failed in CI. The fix still keeps best-effort preference when exactly one candidate has vector hints.
> 
> `else if (matchedFieldInfo.hasVectorValues() == candidate.hasVectorValues()) { ` instead of && would have done the trick I think

@shatejas I have corrected the pr PTAL! Thanks

**Comment by shatejas** [comment]
@leewjae can you do force push to trigger all CIs

**Review by shatejas — DISMISSED** [review]

**Comment by leewjae** [comment]
Hey @shatejas Do I have to force push for running tests?

**Comment by heemin32** [comment]
> Hey @shatejas Do I have to force push for running tests?

Triggered github CI. Only maintainers have a permission to trigger github CI.

**Comment by leewjae** [comment]
Thanks @heemin32

**Comment by leewjae** [comment]
@shatejas all CI checks are green on the latest head commit (b9a6edc). Could you please take a quick look and re-approve when you get a chance? Thank you!

File: src/main/java/org/opensearch/knn/index/codec/KNN10010Codec/KNN10010DerivedSourceStoredFieldsFormat.java [hunk 7]
```diff
@@ -81,6 +82,41 @@ private StoredFieldsFormat getStoredFieldsFormat(final SegmentInfo segmentInfo)
...
+        if (fieldInfo != null) {
+            return fieldInfo;
+        }
+
+        FieldInfo matchedFieldInfo = null;
+        for (FieldInfo candidate : fieldInfos) {
+            if (candidate.name.equalsIgnoreCase(field) == false) {
+                continue;
+            }
+            if (matchedFieldInfo == null || candidate.hasVectorValues() && !matchedFieldInfo.hasVectorValues()) {
```
martin-gaievski, line 105: have you tested that `hasVectorValues` is true for native engines? For Lucene that's the native vector format so that works, but in tests you only checking lucene engine. Ideally that should be part of the tests, but I'm not sure if current CI allows fields that are based on faiss or nmslib

leewjae: Added native-engine coverage in `fd3a8e1`. The new tests build FAISS- and NMSLIB-style `FieldInfo` metadata, assert `hasVectorValues()` is true for those native vector field infos, and verify the mixed-case fallback chooses the native vector field over the non-vector case-insensitive match.

shatejas (dismissed):

navneet1v (dismissed):

shatejas (approved):

leewjae:
@shatejas thank you for the approval and for helping keep the PR moving. CI is now green and all review threads are resolved. GitHub still shows REVIEW_REQUIRED; could you please advise who else needs to review/approve, or help loop in the remaining required reviewer?

shatejas:
@leewjae been merging and rerunning the workflows, please update the branch if there is an update branch requirement till the CI runs

navneet1v (approved):

PR #3319, opened 2026-05-11
Skips excludes and includes for derived source codec layer

### LINKED ISSUE #3303: [BUG]  Inner hits returns vector mask instead of vectors — fields returns wrong values when _source excludes are configured for derived source
**What is the bug?**
When _source excludes are configured on an index to hide vector fields from the default _source response (a common pattern to reduce payload size), inner hits on nested knn queries return corrupted vector data in two ways:

   1. _source in inner hits returns the mask value 1
   2. fields in inner hits are not returned in inner hits

The _source excludes filtering is being applied too aggressively in the inner hits context, corrupting vector data even when explicitly requested.

```
PUT /nested-knn-test
   {
     "settings": { "index.knn": true },
     "mappings": {
       "properties": {
         "title": { "type": "text" },
         "nested_field": {
           "type": "nested",
           "properties": {
             "my_vector": { "type": "knn_vector", "dimension": 4 },
             "color": { "type": "keyword" }
           }
         }
       }
     }
   }
```
```
   POST /nested-knn-test/_doc/1
   {
     "title": "test doc",
     "nested_field": [
       { "my_vector": [1.0, 2.0, 3.0, 4.0], "color": "red" },
       { "my_vector": [5.0, 6.0, 7.0, 8.0], "color": "blue" }
     ]
   }
```

**Case 1**: _source in inner hits returns 1 instead of array
```
   GET /nested-knn-test/_search
   {
     "_source" : {
          "excludes": ["nested_field.my_vector"]
      },
     "query": {
       "nested": {
         "path": "nested_field",
         "query": {
           "knn": { "nested_field.my_vector": { "vector": [1.0, 2.0, 3.0, 4.0], "k": 5 } }
         },
         "inner_hits": { "_source": true }
       }
     }
   }
```
Expected inner hit _source:
```
{ "color": "red", "my_vector": [1.0, 2.0, 3.0, 4.0] }
```
Actual inner hit _source:
```  
 { "color": "red", "my_vector": 1 }
```

**Case 2**: fields in inner hits does not return any value
```
   GET /nested-knn-test/_search
   {
     "_source" : {
          "excludes": ["nested_field.my_vector"]
      },
     "query": {
       "nested": {
         "path": "nested_field",
         "query": {
           "knn": { "nested_field.my_vector": { "vector": [1.0, 2.0, 3.0, 4.0], "k": 5 } }
         },
         "inner_hits": {
           "fields": ["color", "my_vector"]
         }
       }
     }
   }
```

Expected: fields includes my_vector as  ```[1.0, 2.0, 3.0, 4.0]```.
Actual: my_vector is missing or returns incorrect values in the fields response.

**Impact**

   - Both failure modes are silent — no error is returned, just wrong data.
   - Users who set _source.excludes on vector fields (a recommended practice to reduce response payload size) and then explicitly request vector data back via inner hits will silently receive incorrect data.

shatejas:
### Description

Honoring includes and excludes at codec layer breaks the contract in opensearch core where source is read as is on codec layer. This has unintended side effects on fetchSubporcessor layer causing source to have mask values for vectors when derived source is enabled

### Related Issues
Resolves https://github.com/opensearch-project/k-NN/issues/3303

### Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ### Bug Fixes
 * Use KNN1040ScalarQuantizedVectorsFormat for Faiss SQ flat format to enable prefetch [#3302](https://github.com/opensearch-project/k-NN/pull/3302)
+* Skips excludes and includes for derived source [#3319](https://github.com/opensearch-project/k-NN/pull/3319)
 
 ### Refactoring
 
--- a/src/main/java/org/opensearch/knn/index/codec/KNN10010Codec/KNN10010DerivedSourceStoredFieldsReader.java
+++ b/src/main/java/org/opensearch/knn/index/codec/KNN10010Codec/KNN10010DerivedSourceStoredFieldsReader.java
@@ -24,7 +24,7 @@ public class KNN10010DerivedSourceStoredFieldsReader extends StoredFieldsReader 
     private final DerivedSourceReaders derivedSourceReaders;
     private final SegmentReadState segmentReadState;
     private final boolean shouldInject;
-    private boolean closed;
+    private volatile boolean closed;
 
     private final DerivedSourceVectorTransformer derivedSourceVectorTransformer;
 
@@ -98,7 +98,7 @@ public class KNN10010DerivedSourceStoredFieldsReader extends StoredFieldsReader 
     }
 
     @Override
-    public void close() throws IOException {
+    public synchronized void close() throws IOException {
         if (closed == false) {
             IOUtils.close(delegate, derivedSourceReaders);
             closed = true;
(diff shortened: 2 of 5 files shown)


shatejas:
> ❌ Patch coverage is 75.00000% with 4 lines in your changes missing coverage. Please review.
✅ Project coverage is 83.37%. Comparing base 

The 4 lines are for ensureOpen() which is hard to test. The code infact adds unit tests which are missing cases in transformer for inject vectors

navneet1v:
@shatejas adding this change will mean that we are going to read vectors regardless of vectors is required or not at query time. Which will mean that in upcoming versions of OpenSearch this will become a regression for the users. 

I don't think we can merge this change, we need to figure out a better way to fix the problem at hand.

shatejas:
> I don't think we can merge this change, we need to figure out a better way to fix the problem at hand.

https://github.com/opensearch-project/k-NN/issues/3303 The plugin broke the contract with core when excludes was added at codec layer, this is just correcting it. Would be open to suggestions but I haven't found a good way to fix it without removing it. Any other solution would break the architectural contract with core

@sohami Any thoughts here?

PR #3321, opened 2026-05-12
Added capability to retrieve float data type vectors using doc_values

### LINKED ISSUE #3315: [FEATURE] Support docvalue_fields for retrieving KNN vectors without _source parsing
## Description

Enable KNN vector fields to be returned via `docvalue_fields` in search requests, bypassing `_source` decompression entirely. This provides a significant performance
improvement when retrieving vectors alongside search results — up to **7x faster** with binary response formats (CBOR/SMILE) for high-dimensional vectors.

## Motivation

Today, retrieving vectors from search results with derived source means that read full `_source ` from `.fdt` file, decompress, deserialize and then inject the vectors at right place after reading vectors from `.vec` file and then again serializing back to bytes. When we have to get large number vectors say k and size=1000, and _source is too large then `deserialize and then inject the vectors at right place` adds latency and JVM pressure(view small scale benchmarks).

One better way could be users can exclude the vectors from _source during search and then use `docvalue_fields` to get the vectors. Which will avoid whole `deserialize and then inject the vectors at right place`. We would still need to do serialization since we need to send the results back.

## Proposed Change

Implement the OpenSearch `docvalue_fields` contract in the KNN plugin's field mapper:

1. **`KNNVectorFieldType.docValueFormat()`** — Declare support for doc value formatting (returns `DocValueFormat.RAW`)
2. **`KNNVectorDVLeafFieldData.getLeafValueFetcher()`** — Provide a `DocValueFetcher.Leaf` implementation that reads vectors directly from the KNN vector index with a single
seek per document

**Usage:**

    {
      "query": { "knn": { "embedding": { "vector": [...], "k": 10 } } },
      "docvalue_fields": ["embedding"],
      "_source": false
    }

## Benchmark Results JSON Array [Approach 1]

**Quick Setup to validate:** 768-dimensional Cohere vectors, k=1000, 1000 queries, single node on my macbook. JVM: 4gb. 

| Retrieval Method | Format | p50 Latency (ms) | Throughput (QPS) | Speedup vs _source |
| _source (default) | JSON | 56.6 | 17.1 | baseline |
| docvalue_fields | JSON | 40.2 | 23.6 | **1.4x** |
| _source (default) | CBOR | 112.1 | 8.8 | baseline |
| docvalue_fields | CBOR | 15.8 | 60.5 | **6.9x** |
| _source (default) | SMILE | 113.3 | 8.7 | baseline |
| docvalue_fields | SMILE | 15.1 | 62.5 | **7.2x** |
| OSB style (no vector) | JSON | 4.3 | 212.0 | reference |

**Key observations:**
- Binary formats (CBOR/SMILE) benefit most because they encode floats as 4 bytes each vs ~8 chars in JSON text
- `docvalue_fields` with CBOR/SMILE adds only ~11ms over the no-vector baseline for 1000 vectors × 768 dimensions
- The `_source` path is disproportionately expensive with binary formats because it still decompresses the full stored source, then re-serializes

## Benchmark Results Base64 encoded byte array [Approach 2]
Instead of passing the array to serialization layer we do something similar to what CBOR/Smile formats are doing but only vector field. We can encode the whole float[] array to byte[] -> base64 encoded string. I did some benchmarks and here is the observations

```
--- Summary ---
Scenario                                 p50      p95      p99     mean      min      max   stddev      qps
-------------------------------------------------------------------------------------------------------------------
OSB style / JSON / urllib              4.39     6.72    11.34     4.76     2.94    33.29     1.90    210.1
_source default / JSON / urllib       63.46    74.97   103.14    65.39    59.70   145.45     6.95     15.3
docvalue_fields / JSON / urllib       44.49    64.19   110.16    48.73    41.78   278.98    15.71     20.5
docvalue_fields (binary) / JSON / urllib   21.71    37.67    54.08    24.44    19.13   129.53     7.97     40.9
OSB style / CBOR / urllib              4.98     6.54     7.90     5.14     3.45    31.89     1.14    194.7
_source default / CBOR / urllib      115.85   151.48   220.60   122.25   110.35   390.16    20.30      8.2
docvalue_fields / CBOR / urllib       19.94    29.93    38.78    21.83    17.76   107.37     5.92     45.8
docvalue_fields (binary) / CBOR / urllib   19.98    31.99    39.06    21.90    17.32    99.89     6.10     45.7
OSB style / SMILE / urllib             4.85     6.82    10.51     5.10     3.35    30.32     1.83    195.9
_source default / SMILE / urllib     118.24   142.99   193.28   122.87   113.98   277.48    14.54      8.1
docvalue_fields / SMILE / urllib      19.67    28.98    39.34    21.05    17.16    50.15     4.14     47.5
docvalue_fields (binary) / SMILE / urllib   21.17    33.65    48.20    23.82    19.05   310.83    11.03     42.0
OSB style / JSON / opensearch-py       5.42     7.73    21.22     5.93     3.70   105.61     4.21    168.6
_source default / JSON / opensearch-py  122.51   137.69   161.68   124.82   113.59   324.95    11.48      8.0
docvalue_fields / JSON / opensearch-py  113.18   132.53   173.73   116.07   101.38   240.25    12.45      8.6
docvalue_fields (binary) / JSON / opensearch-py   32.69    47.05    78.24    35.13    26.55   273.31    12.93     28.5

```

**Key observations**
```
Speedup (docvalue_fields vs _source default, JSON/urllib): 1.34x
Speedup (docvalue_fields vs _source default, CBOR/urllib): 5.60x
Speedup (docvalue_fields vs _source default, SMILE/urllib): 5.84x
Speedup (docvalue_fields binary vs _source default, JSON/urllib): 2.68x
Speedup (docvalue_fields binary vs _source default, CBOR/urllib): 5.58x
Speedup (docvalue_fields binary vs _source default, SMILE/urllib): 5.16x
Binary vs Array (docvalue_fields, JSON/urllib): 1.99x
Binary vs Array (docvalue_fields, CBOR/urllib): 1.00x
Binary vs Array (docvalue_fields, SMILE/urllib): 0.88x
Speedup (docvalue_fields vs _source default, JSON/opensearch-py): 1.08x
Transport overhead (opensearch-py vs urllib, docvalue_fields/JSON): 2.38x

```

```
Array format response:

{
  "fields": {
    "my_vector": [[1.0, 2.0, 3.0, 4.0]]
  }
}

Binary format response:

{
  "fields": {
    "my_vector": ["AACAPwAAAEAAAEBAAACAQA=="]
  }
}

```

## Implementation Details

The architecture follows the standard OpenSearch `docvalue_fields` contract:
```
    SearchRequest
      → FetchDocValuesPhase
        → KNNVectorFieldType.docValueFormat()
        → SearchLookup → KNNVectorIndexFieldData.load()
          → KNNVectorDVLeafFieldData.getLeafValueFetcher()
            → DocValueFetcher iterates per doc
```
The `getLeafValueFetcher()` implementation uses the existing `KNNVectorValues` iterator (the same one that powers ANN search), so no additional data structures or memory
copies are needed.

We will be adding both the approaches to KNN with default as binary encoded vectors.

## Things to Look for:
1. How to support the same with nested fields.
2. What should be the behavior for byte and binary vectors.

navneet1v:
### Description
Added capability to retrieve float data type vectors using doc_values. More details on small scale benchmarks can be found here: https://github.com/opensearch-project/k-NN/issues/3315

#### Next Steps:
1. Added the support for byte and binary data types.

### When this is useful?
1. If a user wants to get the vector field in the response, but don't want to pay for reading of whole source, then they can user this way to get the vectors.
2. They can also combine with source.excludes since during source excludes for vector field we don't read vectors and inject them back in _source. 

### Related Issues
Partially resolves: https://github.com/opensearch-project/k-NN/issues/3315

### Example
```
POST /my_index/_search
  {
    "query": {
      "knn": {
        "my_vector": {
          "vector": [1.0, 2.0, 3.0, 4.0],
          "k": 5
        }
      }
    },
    "docvalue_fields": ["my_vector"],
    "_source": false
  }
```
  This returns the vectors directly from doc values without reading _source, which is faster — especially for large vectors. The response looks like:
```
  {
    "hits": {
      "hits": [
        {
          "_id": "1",
          "_score": 0.95,
          "fields": {
            "my_vector": [[1.0, 2.0, 3.0, 4.0]]
          }
        }
      ]
    }
  }
```

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/KNNVectorDVLeafFieldData.java
+++ b/src/main/java/org/opensearch/knn/index/KNNVectorDVLeafFieldData.java
@@ -10,13 +10,30 @@ import org.apache.lucene.index.FieldInfo;
 import org.apache.lucene.index.KnnVectorValues;
 import org.apache.lucene.index.LeafReader;
 import org.apache.lucene.search.DocIdSetIterator;
+import org.opensearch.common.lucene.Lucene;
 import org.opensearch.index.fielddata.LeafFieldData;
 import org.opensearch.index.fielddata.ScriptDocValues;
 import org.opensearch.index.fielddata.SortedBinaryDocValues;
+import org.opensearch.index.mapper.DocValueFetcher;
 import org.opensearch.knn.common.FieldInfoExtractor;
+import org.opensearch.knn.index.vectorvalues.KNNVectorValues;
+import org.opensearch.knn.index.vectorvalues.KNNVectorValuesFactory;
+import org.opensearch.search.DocValueFormat;
 
 import java.io.IOException;
 
+/**
+ * Per-segment leaf field data for KNN vector fields. Provides access to vector values
+ * for scripting ({@code doc['field']}) and {@code docvalue_fields} in search responses.
+ *
+ * <p><b>Lifecycle and threading:</b> A new instance is created per segment per search request
+ * via {@link KNNVectorIndexFieldData#load(org.apache.lucene.index.LeafReaderContext)}.
+ * Each search thread owns its own instance — there is no sharing across threads.
+ *
+ * <p><b>Iterator isolation:</b> Each call to {@link #getLeafValueFetcher(DocValueFormat)}
+ * creates its own {@link KNNVectorValues} iterator, ensuring that multiple fetchers
+ * (if ever created from the same instance) cannot corrupt each other's iteration state.
+ */
 public class KNNVectorDVLeafFieldData implements LeafFieldData {
 
     private final LeafReader reader;
@@ -71,4 +88,90 @@ public class KNNVectorDVLeafFieldData implements LeafFieldData {
     public SortedBinaryDocValues getBytesValues() {
         throw new UnsupportedOperationException("knn vector field '" + fieldName + "' doesn't support sorting");
     }
+
+    /**
+     * Returns a {@link DocValueFetcher.Leaf} that reads vector values directly from the
+     * KNN vector index, bypassing {@code _source} entirely.
+     *
+     * <p>This powers {@code "docvalue_fields": ["my_vector_field"]} in search requests.
+     * Unlike the k-NN DerivedSource path (which deserializes the entire {@code _source}, injects
+     * the vector, and re-serializes), this reads the vector with a single seek — zero
+     * {@code _source} parsing overhead.
+     *
+     * <p><b>Iterator isolation:</b> Each invocation creates a fresh {@link KNNVectorValues}
+     * iterator so that multiple {@code Leaf} instances obtained from the same
+     * {@code KNNVectorDVLeafFieldData} cannot interfere with each other's state.
+     *
+     * <p><b>Return type:</b> {@code float[]} for FLOAT vectors —
+     * {@link org.opensearch.core.xcontent.XContentBuilder} serializes this as a JSON numeric array.
+     *
+     * @param format the doc value format — currently unused but required by the interface
+     * @return a leaf fetcher that yields vector values per document, or an empty fetcher
+     *         if the field has no vectors in this segment
+     * @throws UnsupportedOperationException if the vector data type is BYTE or BINARY
+     */
+    @Override
+    public DocValueFetcher.Leaf getLeafValueFetcher(final DocValueFormat format) {
+        if (vectorDataType == VectorDataType.BYTE || vectorDataType == VectorDataType.BINARY) {
+            throw new UnsupportedOperationException(
+                "docvalue_fields is not supported for [" + vectorDataType + "] vector field '" + fieldName + "'"
+            );
+        }
+
+        final FieldInfo fieldInfo = FieldInfoExtractor.getFieldInfo(reader, fieldName);
+        // This is important because if for a segment there is no vector field present and then customer still ask for
+        // the vector field values for the docs in that segment, we should return empty array since vector are not present.
+        if (fieldInfo == null) {
+            return EMPTY_DOCVALUE_FETCHER_LEAF;
+        }
+
+        final KNNVectorValues<?> vectorValues;
+        try {
+            vectorValues = KNNVectorValuesFactory.getVectorValues(fieldInfo, Lucene.segmentReader(reader));
+        } catch (IOException e) {
+            throw new IllegalStateException("Cannot load vector values for field: " + fieldName, e);
+        }
+
+        return new DocValueFetcher.Leaf() {
+            private int count;
+
+            @Override
+            public boolean advanceExact(int docId) throws IOException {
+                if (vectorValues.advance(docId) == docId) {
+                    count = 1;
+                    return true;
+                }
+                count = 0;
+                return false;
+            }
+
+            @Override
+            public int docValueCount() {
+                return count;
+            }
+
+            @Override
+            public Object nextValue() throws IOException {
+                return vectorValues.conditionalCloneVector();
+            }
+        };
+    }
+
+    private static final DocValueFetcher.Leaf EMPTY_DOCVALUE_FETCHER_LEAF = new DocValueFetcher.Leaf() {
+
+        @Override
+        public boolean advanceExact(int docId) {
+            return false;
+        }
+
+        @Override
+        public int docValueCount() {
+            return 0;
+        }
+
+        @Override
+        public Object nextValue() {
+            return null;
+        }
+    };
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,3 +21,4 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ### Enhancements
 * Add the bulkscore logic in MOS when K is greater than number of docs in segment [#3285](https://github.com/opensearch-project/k-NN/pull/3285)
+* Added capability to retrieve float data type vectors using doc_values [#3321](https://github.com/opensearch-project/k-NN/pull/3321)
--- a/src/test/java/org/opensearch/knn/index/KNNVectorDVLeafFieldDataTests.java
+++ b/src/test/java/org/opensearch/knn/index/KNNVectorDVLeafFieldDataTests.java
@@ -5,20 +5,21 @@
 
 package org.opensearch.knn.index;
 
-import org.apache.lucene.util.BytesRef;
-import org.opensearch.knn.KNNTestCase;
-import org.apache.lucene.tests.analysis.MockAnalyzer;
-import org.apache.lucene.document.BinaryDocValuesField;
 import org.apache.lucene.document.Document;
+import org.apache.lucene.document.KnnFloatVectorField;
 import org.apache.lucene.document.NumericDocValuesField;
 import org.apache.lucene.index.DirectoryReader;
 import org.apache.lucene.index.IndexWriter;
 import org.apache.lucene.index.IndexWriterConfig;
 import org.apache.lucene.index.LeafReaderContext;
+import org.apache.lucene.index.VectorSimilarityFunction;
 import org.apache.lucene.store.Directory;
+import org.apache.lucene.tests.analysis.MockAnalyzer;
 import org.opensearch.index.fielddata.ScriptDocValues;
+import org.opensearch.index.mapper.DocValueFetcher;
+import org.opensearch.knn.KNNTestCase;
+import org.opensearch.search.DocValueFormat;
 import org.junit.Before;
-import org.opensearch.knn.index.codec.util.KNNVectorAsCollectionOfFloatsSerializer;
 
 import java.io.IOException;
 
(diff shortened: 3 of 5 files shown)


src/main/java/org/opensearch/knn/index/KNNVectorDVLeafFieldData.java
```diff
@@ -71,4 +96,63 @@ public ScriptDocValues<?> getScriptValues() {
...
+     *   <li>{@code short[]} for BYTE/BINARY vectors — converted from {@code byte[]} because
+     *       {@code XContentBuilder} would serialize raw {@code byte[]} as base64 binary data
+     *       instead of a numeric array.</li>
+     * </ul>
+     *
+     * @param format the doc value format (unused for vectors, but required by the interface)
+     * @return a leaf fetcher that yields vector values per document
+     */
+    @Override
+    public DocValueFetcher.Leaf getLeafValueFetcher(final DocValueFormat format) {
```
Vikasht34, line 114: Naive Question :- How this function will be used upstream?

src/main/java/org/opensearch/knn/index/KNNVectorDVLeafFieldData.java
```diff
@@ -71,4 +96,63 @@ public ScriptDocValues<?> getScriptValues() {
...
+     * <p><b>Threading:</b> The returned {@link DocValueFetcher.Leaf} captures the
+     * {@code vectorValues} iterator from this instance. Since this instance is created
+     * per segment per search request and owned by a single thread, the iterator is never
+     * shared across threads.
+     *
+     * <p><b>Return types:</b>
+     * <ul>
+     *   <li>{@code float[]} for FLOAT vectors — {@link org.opensearch.core.xcontent.XContentBuilder}
+     *       has a registered writer that serializes this directly as a JSON numeric array.</li>
+     *   <li>{@code short[]} for BYTE/BINARY vectors — converted from {@code byte[]} because
```
Vikasht34: nit :- Comment Says , it's returns but function says , it's not supported?

src/main/java/org/opensearch/knn/index/KNNVectorDVLeafFieldData.java
```diff
@@ -71,4 +96,63 @@ public ScriptDocValues<?> getScriptValues() {
...
+    @Override
+    public DocValueFetcher.Leaf getLeafValueFetcher(final DocValueFormat format) {
+        if (vectorDataType == VectorDataType.BYTE || vectorDataType == VectorDataType.BINARY) {
+            throw new UnsupportedOperationException(
+                "docvalue_fields is not supported for [" + vectorDataType + "] vector field '" + fieldName + "'"
+            );
+        }
+
+        return new DocValueFetcher.Leaf() {
+            private int count;
```
Vikasht34, line 136: What is this count used for , it's always 1 .

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldType.java
```diff
@@ -152,6 +154,17 @@ public IndexFieldData.Builder fielddataBuilder(String fullyQualifiedIndexName, S
         return new KNNVectorIndexFieldData.Builder(name(), CoreValuesSourceType.BYTES, this.vectorDataType);
     }
 
+    @Override
+    public DocValueFormat docValueFormat(String format, ZoneId timeZone) {
```
Vikasht34: I am trying to understand why this function is needed?

src/main/java/org/opensearch/knn/index/KNNVectorDVLeafFieldData.java
navneet1v: it is always 1 because in 1 vector field we store only 1 vector. In doc values, you can have more than 1 value. Hence the interface is like this. 

So for vectors it will always be 1

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldType.java
navneet1v: This function is needed to tell what should be the format for the docValues. In my next PR I am going to change this. Let me remove this function for now, since its not needed in this PR.

src/main/java/org/opensearch/knn/index/KNNVectorDVLeafFieldData.java
navneet1v: It was supported earlier, then I removed the support. Let me fix it

src/main/java/org/opensearch/knn/index/KNNVectorDVLeafFieldData.java
navneet1v: This is the call trace.
```
SearchRequest
      → FetchDocValuesPhase
        → KNNVectorFieldType.docValueFormat()
        → SearchLookup → KNNVectorIndexFieldData.load()
          → KNNVectorDVLeafFieldData.getLeafValueFetcher()
            → DocValueFetcher iterates per doc
```

navneet1v: Let me know if you need some other details. So idea is simple, during fetchDocValue phase in Opensearch, if user is asking for vectors in doc values we call `KNNVectorIndexFieldData.load` to get and instance of `KNNVectorDVLeafFieldData`. which then call `getLeafValueFetcher` to get the `DocValueFetcher` which is then used to get the data per document. This looks to be a lot of indirection, but since the interfaces are coming from core, didn't have much of an option to change it.

Vikasht34: Thanks @navneet1v for explanation, that's fine ..I am good with this , Just wanted to understand how this works.

Vikasht34 (dismissed):

src/main/java/org/opensearch/knn/index/KNNVectorDVLeafFieldData.java
Vikasht34: make sense Thanks @navneet1v

src/main/java/org/opensearch/knn/index/KNNVectorDVLeafFieldData.java
```diff
@@ -71,4 +96,61 @@ public ScriptDocValues<?> getScriptValues() {
...
+
+            @Override
+            public boolean advanceExact(int docId) throws IOException {
+                if (vectorValues.advance(docId) == docId) {
```
kotwanikunal, line 140: > Since this instance is created
>     * per segment per search request and owned by a single thread, the iterator is never
>   * shared across threads.

I wanted some more clarity on this for my understanding. I know there should not be a case where the iterator is shared. But if that contract changes, is there a mechanism we can follow where the iterator is reset before advancing to prevent potential issues? Or instead re-initialize the iterator?

navneet1v: if the iterator is shared then in that case the interfaces on how to get doc values would be change. So when that happens we would need to change our code. The reason I wanted to keep a single instance of `vectorValues` so that I can avoid the penalty of vectorvalues creation per doc read.

navneet1v: Another point when we will start to do multi-threading 

we will start to parallelize the doc values fetch, we will parallelize at segment level. If we are parallelizing within segment then we will go with intra segment approach where the leaf reader context per thread will be different. Hence the class KNNVectorDVLeafFieldData will be created per thread. So no issue will happen.

kotwanikunal (approved):

src/main/java/org/opensearch/knn/index/KNNVectorDVLeafFieldData.java
```diff
@@ -71,4 +88,90 @@ public ScriptDocValues<?> getScriptValues() {
...
         throw new UnsupportedOperationException("knn vector field '" + fieldName + "' doesn't support sorting");
     }
+
+    /**
+     * Returns a {@link DocValueFetcher.Leaf} that reads vector values directly from the
+     * KNN vector index, bypassing {@code _source} entirely.
+     *
+     * <p>This powers {@code "docvalue_fields": ["my_vector_field"]} in search requests.
+     * Unlike the DerivedSource path (which deserializes the entire {@code _source}, injects
+     * the vector, and re-serializes), this reads the vector with a single seek — zero
```
shatejas:
```suggestion
     * Unlike the k-NN DerivedSource path (which deserializes the entire {@code _source}, injects
     * the vector, and re-serializes), this reads the vector with a single seek — zero
```

src/main/java/org/opensearch/knn/index/KNNVectorDVLeafFieldData.java
```diff
@@ -71,4 +88,90 @@ public ScriptDocValues<?> getScriptValues() {
...
+            }
+
+            @Override
+            public int docValueCount() {
+                return count;
+            }
+
+            @Override
+            public Object nextValue() throws IOException {
+                return vectorValues.getVector();
```
shatejas: This seems to be storing it in an array list https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/mapper/DocValueFetcher.java#L74

Are you sure this doesn't need `conditionalClonedVector`? 

Most of the ITs are just equaling dimensions which isn't giving me much confidence and the one test     `public void testDocValueFields_vectorValuesMatchSource() {` which matches vector just has 1 doc in it. Can we make sure we have assertArrayEquals and match all the vectors?

navneet1v: Lets me fix the array to validate the actual results

navneet1v: Good point, actually this was in back of my mind but somehow got skipped. Thanks for pointing this out.

navneet1v: Fixed and improved UTs and ITs to ensure that we check vectors properly

navneet1v:
@shatejas improved the tests please check

shatejas (approved):

Vikasht34 (approved):

PR #3324, opened 2026-05-17
Add base64 binary encoding as default format for knn_vector docvalue_fields

### LINKED ISSUE #3315: [FEATURE] Support docvalue_fields for retrieving KNN vectors without _source parsing

navneet1v:
### Description
Add base64 binary encoding as default format for knn_vector docvalue_fields

### Summary

  Adds binary (base64) encoding as the default output format for `docvalue_fields` on `knn_vector` fields. When no format is specified in the request, vectors are now returned as
  base64-encoded little-endian float byte strings instead of JSON numeric arrays.

  This builds on #3321 which added `docvalue_fields` support with array format only. The binary default provides **~2x throughput improvement** over the array format for JSON
  transport (the most common case), while maintaining the same performance for binary transports (CBOR/SMILE) where both formats are already efficient.

#### Benchmark highlights (768D Cohere vectors, k=1000)

Quick Setup to validate: 768-dimensional Cohere vectors, k=1000, 1000 queries, single node on my macbook. JVM: 4gb.

  | Scenario (JSON transport) | p50 latency | QPS | Speedup vs `_source` |
  |---|---|---|---|
  | `_source` (baseline) | 63.5 ms | 15.3 | — |
  | `docvalue_fields` (array) | 44.5 ms | 20.5 | 1.34x |
  | `docvalue_fields` (binary, default) | 21.7 ms | 40.9 | **2.68x** |
  
Binary format also reduces JSON response payload by ~30-40% for high-dimensional vectors since base64 encodes 768 floats in ~4 KB vs ~6 KB for a JSON numeric array.
  
#### Format selection
  
Users can explicitly choose the format:
  ```json
  "docvalue_fields": [{"field": "my_vector", "format": "binary"}]   // base64 string (default)
  "docvalue_fields": [{"field": "my_vector", "format": "array"}]    // JSON numeric array
  "docvalue_fields": ["my_vector"]                                   // base64 string (default)
```  

### Related Issues
Resolves https://github.com/opensearch-project/k-NN/issues/3315
### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/KNNVectorDVLeafFieldData.java
+++ b/src/main/java/org/opensearch/knn/index/KNNVectorDVLeafFieldData.java
@@ -102,16 +102,29 @@ public class KNNVectorDVLeafFieldData implements LeafFieldData {
      * iterator so that multiple {@code Leaf} instances obtained from the same
      * {@code KNNVectorDVLeafFieldData} cannot interfere with each other's state.
      *
-     * <p><b>Return type:</b> {@code float[]} for FLOAT vectors —
-     * {@link org.opensearch.core.xcontent.XContentBuilder} serializes this as a JSON numeric array.
+     * <p><b>Return type:</b> Depends on the format:
+     * <ul>
+     *   <li>Array format ({@link KNNVectorDocValueFormat#ARRAY_FORMAT}): returns {@code float[]},
+     *       serialized by {@link org.opensearch.core.xcontent.XContentBuilder} as a JSON numeric array.</li>
+     *   <li>Binary format ({@link KNNVectorDocValueFormat#BINARY_FORMAT}, the default): returns a
+     *       base64-encoded {@link String} of little-endian float bytes.</li>
+     * </ul>
      *
-     * @param format the doc value format — currently unused but required by the interface
+     * @param format the {@link KNNVectorDocValueFormat} that determines output encoding (array or binary)
      * @return a leaf fetcher that yields vector values per document, or an empty fetcher
      *         if the field has no vectors in this segment
      * @throws UnsupportedOperationException if the vector data type is BYTE or BINARY
+     * @throws IllegalArgumentException if format is not an instance of {@link KNNVectorDocValueFormat}
      */
     @Override
     public DocValueFetcher.Leaf getLeafValueFetcher(final DocValueFormat format) {
+        if (!(format instanceof KNNVectorDocValueFormat knnFormat)) {
+            throw new IllegalArgumentException(
+                "Unsupported DocValueFormat [" + format + "] for knn_vector field '" + fieldName + "'. Expected KNNVectorDocValueFormat."
+            );
+        }
+        final boolean isBinary = knnFormat.isBinary();
+
         if (vectorDataType == VectorDataType.BYTE || vectorDataType == VectorDataType.BINARY) {
             throw new UnsupportedOperationException(
                 "docvalue_fields is not supported for [" + vectorDataType + "] vector field '" + fieldName + "'"
@@ -152,6 +165,12 @@ public class KNNVectorDVLeafFieldData implements LeafFieldData {
 
             @Override
             public Object nextValue() throws IOException {
+                // We don't need a conditional clone here since encodeToBinary will convert the array to string.
+                if (isBinary) {
+                    return KNNVectorDocValueFormat.encodeToBinary((float[]) vectorValues.getVector());
+                }
+                // We need a conditional clone since vector returned from here will be added in a map, so we do the clone
+                // since vectorValues keep a single copy of vector array for all docs.
                 return vectorValues.conditionalCloneVector();
             }
         };
--- a/src/main/java/org/opensearch/knn/index/KNNVectorDocValueFormat.java
+++ b/src/main/java/org/opensearch/knn/index/KNNVectorDocValueFormat.java
@@ -1,0 +1,96 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.knn.index;
+
+import lombok.Getter;
+import org.opensearch.core.common.io.stream.StreamInput;
+import org.opensearch.core.common.io.stream.StreamOutput;
+import org.opensearch.search.DocValueFormat;
+
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+import java.util.Arrays;
+import java.util.Base64;
+
+/**
+ * DocValueFormat for knn_vector fields. Supports two modes:
+ * <ul>
+ *   <li>{@code array} — vectors are returned as JSON numeric arrays</li>
+ *   <li>{@code binary} (default) — vectors are returned as base64-encoded little-endian byte strings (with padding)</li>
+ * </ul>
+ */
+@Getter
+public enum KNNVectorDocValueFormat implements DocValueFormat {
+
+    ARRAY_FORMAT("array", false),
+    BINARY_FORMAT("binary", true);
+
+    public static final String NAME = "knn_vector";
+
+    private final String formatName;
+    private final boolean binary;
+    private static final Base64.Encoder BASE64_ENCODER = Base64.getEncoder();
+
+    KNNVectorDocValueFormat(final String formatName, boolean binary) {
+        this.formatName = formatName;
+        this.binary = binary;
+    }
+
+    public static KNNVectorDocValueFormat fromStream(final StreamInput in) throws IOException {
+        return in.readBoolean() ? BINARY_FORMAT : ARRAY_FORMAT;
+    }
+
+    /**
+     * Resolves the format string to the corresponding enum constant.
+     * Returns {@link #BINARY_FORMAT} when format is null (the default).
+     *
+     * @param format the format string from the docvalue_fields request, or null for default
+     * @return the matching {@link KNNVectorDocValueFormat}
+     * @throws IllegalArgumentException if the format string is not recognized
+     */
+    public static KNNVectorDocValueFormat fromFormatString(final String format) {
+        if (format == null || BINARY_FORMAT.formatName.equals(format)) {
+            return BINARY_FORMAT;
+        }
+        if (ARRAY_FORMAT.formatName.equals(format)) {
+            return ARRAY_FORMAT;
+        }
+        throw new IllegalArgumentException(
+            "Unsupported knn_vector docvalue_fields format ["
+                + format
+                + "]. Supported formats are "
+                + Arrays.toString(KNNVectorDocValueFormat.values())
+        );
+    }
+
+    @Override
+    public String getWriteableName() {
+        return NAME;
+    }
+
+    @Override
+    public void writeTo(final StreamOutput out) throws IOException {
+        out.writeBoolean(binary);
+    }
+
+    /**
+     * Encodes a float[] vector as a base64 string with little-endian byte order.
+     * Each float is written as 4 bytes in little-endian format (native byte order on x86/ARM),
+     * then the resulting byte array is base64-encoded.
+     */
+    public static String encodeToBinary(final float[] vector) {
+        final ByteBuffer buffer = ByteBuffer.allocate(vector.length * Float.BYTES).order(ByteOrder.LITTLE_ENDIAN);
+        buffer.asFloatBuffer().put(vector); // Bulk operation optimized by the JVM
+        final byte[] bytes = buffer.array();
+        return BASE64_ENCODER.encodeToString(bytes);
+    }
+
+    @Override
+    public String toString() {
+        return "knn_vector(" + formatName + ")";
+    }
+}
--- a/src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldType.java
+++ b/src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldType.java
@@ -20,6 +20,7 @@ import org.opensearch.index.mapper.TextSearchInfo;
 import org.opensearch.index.mapper.ValueFetcher;
 import org.opensearch.index.query.QueryShardContext;
 import org.opensearch.index.query.QueryShardException;
+import org.opensearch.knn.index.KNNVectorDocValueFormat;
 import org.opensearch.knn.index.KNNVectorIndexFieldData;
 import org.opensearch.knn.index.VectorDataType;
 import org.opensearch.knn.index.engine.KNNEngine;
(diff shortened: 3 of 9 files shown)

### CODE REVIEW

File: src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldType.java [hunk 1]
```diff
@@ -152,6 +155,35 @@ public IndexFieldData.Builder fielddataBuilder(String fullyQualifiedIndexName, S
         return new KNNVectorIndexFieldData.Builder(name(), CoreValuesSourceType.BYTES, this.vectorDataType);
     }
 
+    @Override
+    public DocValueFormat docValueFormat(final String format, final ZoneId timeZone) {
+        if (timeZone != null) {
+            throw new IllegalArgumentException("Field [" + name() + "] of type [" + typeName() + "] does not support custom time zones");
+        }
```
VijayanB, line 162: Just double checking that if it timeZone is not null, we want to throw  exception? can we throw UnSupportedOperation if we think support is not available?

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldType.java
```diff
@@ -152,6 +155,35 @@ public IndexFieldData.Builder fielddataBuilder(String fullyQualifiedIndexName, S
...
+        // Using binary format as default here since binary format provide better performance over array format.
+        // Ref gh issue: https://github.com/opensearch-project/k-NN/issues/3315
+        if (format == null || KNNVectorDocValueFormat.FORMAT_BINARY.equals(format)) {
+            return KNNVectorDocValueFormat.BINARY_FORMAT;
```
VijayanB: can we add log.debug?

src/main/java/org/opensearch/knn/index/KNNVectorDocValueFormat.java
```diff
@@ -0,0 +1,73 @@
...
+
+/**
+ * DocValueFormat for knn_vector fields. Supports two modes:
+ * <ul>
+ *   <li>{@code array} — vectors are returned as JSON numeric arrays</li>
+ *   <li>{@code binary} (default) — vectors are returned as base64-encoded little-endian byte strings (with padding)</li>
+ * </ul>
+ */
+@Getter
+public class KNNVectorDocValueFormat implements DocValueFormat {
```
VijayanB: Have you considered using Enum for KNNVectorDocValueFormat  instead of static singleton instances? It will be easy to extend imo.

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldType.java
navneet1v: sure

src/main/java/org/opensearch/knn/index/KNNVectorDocValueFormat.java
navneet1v: We might not be able to use enum since we need to give a new instance here:
```
entries.add(new NamedWriteableRegistry.Entry(DocValueFormat.class, KNNVectorDocValueFormat.NAME, KNNVectorDocValueFormat::new));
```

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldType.java
navneet1v: isn't that same thing saying that argument is not supported. :P

src/main/java/org/opensearch/knn/index/KNNVectorDocValueFormat.java
kotwanikunal: +1 on the enum. It reads a lot more naturally.
I think we can still achieve it - 

```
public enum KNNVectorDocValueFormat implements DocValueFormat {
    ARRAY("array"),
    BINARY("binary");

    public static final String NAME = "knn_vector";
    ...
    public static KNNVectorDocValueFormat fromStream(StreamInput in) throws IOException {
        return in.readBoolean() ? BINARY : ARRAY;
    }
...
}
```

ANd for the registry - 
```
entries.add(new NamedWriteableRegistry.Entry(DocValueFormat.class, KNNVectorDocValueFormat.NAME, KNNVectorDocValueFormat::fromStream));

```

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldType.java
VijayanB: Fair enough. I was just looking at error message and see that it gives impression that it is unsupported than illegal argument. may be can we change error message to say that is not valid argument. it is not deal breaker. I will leave it to you.

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldType.java
```diff
@@ -152,6 +155,33 @@ public IndexFieldData.Builder fielddataBuilder(String fullyQualifiedIndexName, S
...
+                    + "] of type ["
+                    + typeName()
+                    + "] does not support format ["
+                    + format
+                    + "]. "
+                    + "Supported formats are ["
+                    + KNNVectorDocValueFormat.ARRAY_FORMAT.getFormatName()
+                    + ", "
+                    + KNNVectorDocValueFormat.BINARY_FORMAT.getFormatName()
+                    + "]"
```
VijayanB: nit: Instead of leaking the formats, can we move it to Array/method to give those details? i am trying to see if we add new enum tomm, that we don't have to touch this class at all. or if you think that we can re-use error message from the exception that will be great too.

src/main/java/org/opensearch/knn/index/KNNVectorDocValueFormat.java
```diff
@@ -0,0 +1,98 @@
...
+    public static String encodeToBinary(final float[] vector) {
+        final ByteBuffer buffer = ByteBuffer.allocate(vector.length * Float.BYTES).order(ByteOrder.LITTLE_ENDIAN);
+        buffer.asFloatBuffer().put(vector); // Bulk operation optimized by the JVM
+        final byte[] bytes = buffer.array();
+        return BASE64_ENCODER.encodeToString(bytes);
+    }
+
+    @Override
+    public String toString() {
+        return binary ? "knn_vector(binary)" : "knn_vector(array)";
```
VijayanB: can this be simplified as `return "knn_vector(" + formatName + ")";`

navneet1v: I don't think that makes any diff. we can keep it like this

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldType.java
navneet1v: sure. doesn't make much of a difference.

kotwanikunal (approved):

VijayanB (dismissed):

VijayanB (dismissed):

src/main/java/org/opensearch/knn/index/KNNVectorDVLeafFieldData.java
```diff
@@ -132,6 +137,8 @@ public DocValueFetcher.Leaf getLeafValueFetcher(final DocValueFormat format) {
             throw new IllegalStateException("Cannot load vector values for field: " + fieldName, e);
         }
 
+        final boolean isBinary = (format instanceof KNNVectorDocValueFormat == false) || ((KNNVectorDocValueFormat) format).isBinary();
```
shatejas: shouldn't `(format instanceof KNNVectorDocValueFormat == false)` throw an error

navneet1v: Good point. Idea here is since this code:
```
@Override
    public DocValueFormat docValueFormat(final String format, final ZoneId timeZone) {
        if (timeZone != null) {
            throw new IllegalArgumentException("Field [" + name() + "] of type [" + typeName() + "] does not support custom time zones");
        }
        return KNNVectorDocValueFormat.fromFormatString(format);
    }
```

will avoid sending the wrong format. I was thinking to just default it. But I see the other side of this too. Lets throw an error from here which will be better looks like.

Vikasht34 (approved):
Changes Looks Good , I don't have any concern on Changes or Code.

VijayanB (approved):

PR #3330, opened 2026-05-19
Introduce ResolvedIndexSpec for pre-computed index behavior

kotwanikunal:
### Description

Introduces `ResolvedIndexSpec` (RIS), an immutable object computed once at index mapping resolution time that consolidates all engine, compression, mode, encoder, and behavioral decisions into a single pre-computed source of truth.

**Problem:** Behavioral questions about an index (codec format selection, rescore eligibility, memory-optimized search, remote build support) were answered by scattered logic across `CompressionLevel`, `FaissHNSWMethod`, `FaissSQEncoder`, and `KNNVectorFieldType`. Each call site re-derived answers from raw parameters, leading to duplication and inconsistency risk.

**Solution:** RIS is built once during method resolution and flows through:
- **Write path:** `KNNLibraryIndexingContext.getResolvedSpec()` (source of truth)
- **Query path:** `KNNVectorFieldType.getResolvedSpec()` (same object, cached reference)

Callers now ask behavioral methods (`usesSQ1BitCodecFormat()`, `getRescoreContext()`, `isMemoryOptimizedEligible()`, `supportsRemoteIndexBuild()`) instead of inspecting raw encoder parameters.

**What reviewers should expect:**

| Area | Change pattern |
|------|---------------|
| `ResolvedIndexSpec.java` (new) | Core abstraction. Lombok `@Builder @Getter`. Methods for now - can be cached properties. |
| `Encoder.java` | New inner enums (`EncoderType`, `QuantizationBits`) + `validate()` default method |
| `AbstractKNNMethod.java` | Builds RIS during `getKNNLibraryIndexingContext()` |
| `KNNVectorFieldType.java` | Holds RIS reference, delegates rescore to `resolvedSpec.getRescoreContext()` |
| `FaissFieldStrategy.java` / `LuceneFieldStrategy.java` (new) | Extracted from `EngineFieldMapper` for engine-specific field type construction |
| `FaissCodecFormatResolver` / `LuceneCodecFormatResolver` | Accept RIS in `resolve()` signature, prefer spec-based check over raw param inspection |
| `CompressionLevel.java` | `getDefaultRescoreContext()` deprecated (logic moved to RIS) |
| `FaissMethodResolver` / `LuceneHNSWMethodResolver` | `getDefaultCompressionLevel()` extracted to `AbstractMethodResolver`, mode deprecated |
| `FaissSQEncoder` / `LuceneSQEncoder` | Implement `validate()` for create-index-time encoder validation |
| Tests | New coverage for RIS, wiring, consumer migration. Existing tests migrated to new APIs. |

**Key design decisions:**
- RIS uses methods for derived properties for evolution, can be optimized to be cached in the object instance
- Backward compatible: no behavioral change for any existing index or query path

### Related Issues
Related #3290

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/engine/AbstractKNNMethod.java
+++ b/src/main/java/org/opensearch/knn/index/engine/AbstractKNNMethod.java
@@ -153,12 +153,11 @@ public abstract class AbstractKNNMethod implements KNNMethod {
         Map<String, Object> methodParams = knnMethodContext.getMethodComponentContext().getParameters();
         if (methodParams != null && methodParams.containsKey(METHOD_ENCODER_PARAMETER)) {
             Object encoderObj = methodParams.get(METHOD_ENCODER_PARAMETER);
-            if (encoderObj instanceof MethodComponentContext) {
-                MethodComponentContext encoderCtx = (MethodComponentContext) encoderObj;
+            if (encoderObj instanceof MethodComponentContext encoderCtx) {
                 encoderType = Encoder.EncoderType.fromName(encoderCtx.getName());
                 Object bitsObj = encoderCtx.getParameters().get(KNNConstants.SQ_BITS);
-                if (bitsObj instanceof Integer) {
-                    quantizationBits = Encoder.QuantizationBits.fromValue((Integer) bitsObj);
+                if (bitsObj instanceof Integer bitsInt) {
+                    quantizationBits = Encoder.QuantizationBits.fromValue(bitsInt);
                 } else if (encoderType == Encoder.EncoderType.FLAT) {
                     quantizationBits = Encoder.QuantizationBits.FULL_PRECISION;
                 } else if (encoderType == Encoder.EncoderType.BQ) {
@@ -167,6 +166,7 @@ public abstract class AbstractKNNMethod implements KNNMethod {
             }
         }
 
+        // Dimension is null during model training flows (resolved from training data, not mapping params)
         Integer dimension = knnMethodConfigContext.getDimension();
         return ResolvedIndexSpec.builder()
             .engine(knnMethodContext.getKnnEngine())
--- a/src/main/java/org/opensearch/knn/index/engine/AbstractKNNMethod.java
+++ b/src/main/java/org/opensearch/knn/index/engine/AbstractKNNMethod.java
@@ -24,6 +24,8 @@ import java.util.Map;
 import java.util.Set;
 import java.util.function.Function;
 
+import static org.opensearch.knn.common.KNNConstants.METHOD_ENCODER_PARAMETER;
+
 /**
  * Abstract class for KNN methods. This class provides the common functionality for all KNN methods.
  * It defines the common attributes and methods that all KNN methods should implement.
@@ -140,6 +142,42 @@ public abstract class AbstractKNNMethod implements KNNMethod {
             .perDimensionProcessor(doGetPerDimensionProcessor(knnMethodContext, knnMethodConfigContext))
             .vectorTransformer(getVectorTransformer(knnMethodContext.getSpaceType()))
             .trainingConfigValidationSetup(doGetTrainingConfigValidationSetup())
+            .resolvedSpec(buildResolvedIndexSpec(knnMethodContext, knnMethodConfigContext))
+            .build();
+    }
+
+    private ResolvedIndexSpec buildResolvedIndexSpec(KNNMethodContext knnMethodContext, KNNMethodConfigContext knnMethodConfigContext) {
+        Encoder.EncoderType encoderType = Encoder.EncoderType.FLAT;
+        Encoder.QuantizationBits quantizationBits = Encoder.QuantizationBits.FULL_PRECISION;
+
+        Map<String, Object> methodParams = knnMethodContext.getMethodComponentContext().getParameters();
+        if (methodParams != null && methodParams.containsKey(METHOD_ENCODER_PARAMETER)) {
+            Object encoderObj = methodParams.get(METHOD_ENCODER_PARAMETER);
+            if (encoderObj instanceof MethodComponentContext) {
+                MethodComponentContext encoderCtx = (MethodComponentContext) encoderObj;
+                encoderType = Encoder.EncoderType.fromName(encoderCtx.getName());
+                Object bitsObj = encoderCtx.getParameters().get(KNNConstants.SQ_BITS);
+                if (bitsObj instanceof Integer) {
+                    quantizationBits = Encoder.QuantizationBits.fromValue((Integer) bitsObj);
+                } else if (encoderType == Encoder.EncoderType.FLAT) {
+                    quantizationBits = Encoder.QuantizationBits.FULL_PRECISION;
+                } else if (encoderType == Encoder.EncoderType.BQ) {
+                    quantizationBits = Encoder.QuantizationBits.ONE;
+                }
+            }
+        }
+
+        Integer dimension = knnMethodConfigContext.getDimension();
+        return ResolvedIndexSpec.builder()
+            .engine(knnMethodContext.getKnnEngine())
+            .methodName(knnMethodContext.getMethodComponentContext().getName())
+            .encoderType(encoderType)
+            .quantizationBits(quantizationBits)
+            .compressionLevel(knnMethodConfigContext.getCompressionLevel())
+            .mode(knnMethodConfigContext.getMode())
+            .vectorDataType(knnMethodConfigContext.getVectorDataType())
+            .dimension(dimension != null ? dimension : 0)
+            .indexVersionCreated(knnMethodConfigContext.getVersionCreated())
             .build();
     }
 
--- a/src/main/java/org/opensearch/knn/index/engine/Encoder.java
+++ b/src/main/java/org/opensearch/knn/index/engine/Encoder.java
@@ -17,6 +17,14 @@ import static org.opensearch.knn.common.KNNConstants.ENCODER_SQ;
 
 /**
  * Interface representing an encoder. An encoder generally refers to a vector quantizer.
+ *
+ * <p>Serves a dual role:
+ * <ul>
+ *   <li>Type classification (via {@link EncoderType} and {@link QuantizationBits}) that drives
+ *       ResolvedIndexSpec behavioral decisions.</li>
+ *   <li>Method component integration (via {@link #getMethodComponent()}) for the existing
+ *       resolution framework.</li>
+ * </ul>
  */
 public interface Encoder {
 
@@ -59,8 +67,8 @@ public interface Encoder {
         FOUR(4, CompressionLevel.x8),
         SEVEN(7, CompressionLevel.x4),
         SIXTEEN(16, CompressionLevel.x2),
-        FULL_PRECISION(32, CompressionLevel.x1),
-        NOT_APPLICABLE(-1, CompressionLevel.NOT_CONFIGURED);
+        /** Identity value for FLAT encoders: full precision float32 with no quantization applied. */
+        FULL_PRECISION(32, CompressionLevel.x1);
 
         private final int value;
         private final CompressionLevel compressionLevel;
@@ -132,6 +140,7 @@ public interface Encoder {
 
     /**
      * @return the set of bit widths this encoder supports
+     * TODO: Evaluate removal -- zero production callers, risks inconsistency with getQuantizationBits().
      */
     Set<QuantizationBits> getSupportedBits();
 
--- a/src/main/java/org/opensearch/knn/index/engine/Encoder.java
+++ b/src/main/java/org/opensearch/knn/index/engine/Encoder.java
@@ -7,12 +7,89 @@ package org.opensearch.knn.index.engine;
 
 import org.opensearch.knn.index.mapper.CompressionLevel;
 
+import java.util.Locale;
+import java.util.Set;
+
+import static org.opensearch.knn.common.KNNConstants.ENCODER_BINARY;
+import static org.opensearch.knn.common.KNNConstants.ENCODER_FLAT;
+import static org.opensearch.knn.common.KNNConstants.ENCODER_PQ;
+import static org.opensearch.knn.common.KNNConstants.ENCODER_SQ;
+
 /**
  * Interface representing an encoder. An encoder generally refers to a vector quantizer.
  */
 public interface Encoder {
+
     /**
-     * The name of the encoder does not have to be unique. Howevwer, when using within a method, there cannot be
+     * Identifies the encoder type. Anchored to KNNConstants encoder name strings.
+     */
+    enum EncoderType {
+        FLAT(ENCODER_FLAT),
+        SQ(ENCODER_SQ),
+        PQ(ENCODER_PQ),
+        BQ(ENCODER_BINARY);
+
+        private final String name;
+
+        EncoderType(String name) {
+            this.name = name;
+        }
+
+        public String getName() {
+            return name;
+        }
+
+        public static EncoderType fromName(String name) {
+            for (EncoderType type : values()) {
+                if (type.name.equals(name)) {
+                    return type;
+                }
+            }
+            throw new IllegalArgumentException(String.format(Locale.ROOT, "Unsupported encoder type: [%s]", name));
+        }
+    }
+
+    /**
+     * Unified quantization bits enum. Maps bit widths to compression levels.
+     * Serialization-compatible: getValue() returns the int stored in FieldInfo/SQConfig.
+     */
+    enum QuantizationBits {
+        ONE(1, CompressionLevel.x32),
+        TWO(2, CompressionLevel.x16),
+        FOUR(4, CompressionLevel.x8),
+        SEVEN(7, CompressionLevel.x4),
+        SIXTEEN(16, CompressionLevel.x2),
+        FULL_PRECISION(32, CompressionLevel.x1),
+        NOT_APPLICABLE(-1, CompressionLevel.NOT_CONFIGURED);
+
+        private final int value;
+        private final CompressionLevel compressionLevel;
+
+        QuantizationBits(int value, CompressionLevel compressionLevel) {
+            this.value = value;
+            this.compressionLevel = compressionLevel;
+        }
+
+        public int getValue() {
+            return value;
+        }
+
+        public CompressionLevel getCompressionLevel() {
+            return compressionLevel;
+        }
+
+        public static QuantizationBits fromValue(int value) {
+            for (QuantizationBits bits : values()) {
+                if (bits.value == value) {
+                    return bits;
+                }
+            }
+            return FULL_PRECISION;
+        }
+    }
+
+    /**
+     * The name of the encoder does not have to be unique. However, when using within a method, there cannot be
      * 2 encoders with the same name.
      *
      * @return Name of the encoder
@@ -22,16 +99,16 @@ public interface Encoder {
     }
 
     /**
-     *
      * @return Method component associated with the encoder
      */
     MethodComponent getMethodComponent();
 
     /**
-     * Calculate the compression level for the give params. Assume float32 vectors are used. All parameters should
+     * Calculate the compression level for the given params. Assume float32 vectors are used. All parameters should
      * be resolved in the encoderContext passed in.
      *
      * @param encoderContext Context for the encoder to extract params from
+     * @param knnMethodConfigContext method config context
      * @return Compression level this encoder produces. If the encoder does not support this calculation yet, it will
      *          return {@link CompressionLevel#NOT_CONFIGURED}
      */
@@ -40,10 +117,28 @@ public interface Encoder {
     /**
      * Validates config of encoder
      *
+     * @param validationInput input for validation
      * @return Validation output of encoder parameters
      */
     default TrainingConfigValidationOutput validateEncoderConfig(TrainingConfigValidationInput validationInput) {
         TrainingConfigValidationOutput.TrainingConfigValidationOutputBuilder builder = TrainingConfigValidationOutput.builder();
         return builder.build();
+    }
+
+    /**
+     * @return the encoder type classification
+     */
+    EncoderType getEncoderType();
+
+    /**
+     * @return the set of bit widths this encoder supports
+     */
+    Set<QuantizationBits> getSupportedBits();
+
+    /**
+     * @return the resolved quantization bits for this encoder instance
+     */
+    default QuantizationBits getQuantizationBits() {
+        return QuantizationBits.FULL_PRECISION;
     }
 }
--- a/src/main/java/org/opensearch/knn/index/engine/KNNLibraryIndexingContext.java
+++ b/src/main/java/org/opensearch/knn/index/engine/KNNLibraryIndexingContext.java
@@ -11,6 +11,7 @@ import org.opensearch.knn.index.mapper.PerDimensionValidator;
 import org.opensearch.knn.index.mapper.VectorTransformer;
 import org.opensearch.knn.index.mapper.VectorValidator;
 
+import org.opensearch.common.Nullable;
 import java.util.Map;
 import java.util.function.Function;
 
@@ -63,4 +64,16 @@ public interface KNNLibraryIndexingContext {
      * @return VectorTransformer
      */
     VectorTransformer getVectorTransformer();
+
+    /**
+     * Get the resolved index spec containing pre-computed behavioral decisions.
+     * May be null for model-based indices, training indices, or flat-index paths
+     * where full resolution does not run.
+     *
+     * @return ResolvedIndexSpec or null
+     */
+    @Nullable
+    default ResolvedIndexSpec getResolvedSpec() {
+        return null;
+    }
 }
--- a/src/main/java/org/opensearch/knn/index/engine/KNNLibraryIndexingContext.java
+++ b/src/main/java/org/opensearch/knn/index/engine/KNNLibraryIndexingContext.java
@@ -11,7 +11,6 @@ import org.opensearch.knn.index.mapper.PerDimensionValidator;
 import org.opensearch.knn.index.mapper.VectorTransformer;
 import org.opensearch.knn.index.mapper.VectorValidator;
 
-import org.opensearch.common.Nullable;
 import java.util.Map;
 import java.util.function.Function;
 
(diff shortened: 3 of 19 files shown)


src/main/java/org/opensearch/knn/index/engine/ResolvedIndexSpec.java
```diff
@@ -0,0 +1,127 @@
...
+    private final VectorDataType vectorDataType;
+    private final int dimension;
+    private final Version indexVersionCreated;
+
+    /**
+     * Whether this configuration uses the SQ 1-bit codec format (Faiss only).
+     */
+    public boolean usesSQ1BitCodecFormat() {
+        return engine == KNNEngine.FAISS && isSQOneBit();
+    }
```
navneet1v: why this is for faiss? and one big question I have is when Sq2 bit comes in what would we do?

src/main/java/org/opensearch/knn/index/mapper/EngineFieldMapper.java
```diff
@@ -78,6 +79,7 @@ public static EngineFieldMapper createFieldMapper(
         KNNLibraryIndexingContext libraryContext = methodContext.getKnnEngine()
             .getKNNLibraryIndexingContext(methodContext, knnMethodConfigContext);
         boolean isLuceneEngine = KNNEngine.LUCENE.equals(methodContext.getKnnEngine());
+        ResolvedIndexSpec resolvedSpec = libraryContext != null ? libraryContext.getResolvedSpec() : null;
```
navneet1v: can we figure out a way to not have ResolvedIndexSpec being null? so that downstream code can easily use `ResolvedIndexSpec` always

src/main/java/org/opensearch/knn/index/engine/ResolvedIndexSpec.java
kotwanikunal: I expected this question - and this is across the interface methods currently. The Faiss check is there because codec format routing for SQ is Faiss specific, given lucene is handled internally. 

For SQ n-bit: this method would broaden to something like `usesFaissSQCodecFormat()` or evolve into `getCodecFormat()` returning a typed value. The engine guard stays regardless.

RIS is intentionally minimal right now - narrow methods for specific decisions, generalized as patterns emerge. The current goal is a pluggable replacement for the scattered checks. We can evolve the abstraction (e.g. `EnumSet<Capability>` for boolean flags, typed methods for config-carrying decisions) through subsequent PRs once the second and third cases land. Open to suggestions - but this was my thought process when implementing.

src/main/java/org/opensearch/knn/index/mapper/EngineFieldMapper.java
kotwanikunal: Agreed - the nullable path exists right now because this PR only wires RIS for the standard engine paths.

The plan is to cover those paths in follow-up commits and remove the @Nullable annotation + null guards once all creation paths produce a RIS. At that point downstream code can rely on it always being present.

src/main/java/org/opensearch/knn/index/engine/Encoder.java
```diff
@@ -7,12 +7,89 @@
...
+            return compressionLevel;
+        }
+
+        public static QuantizationBits fromValue(int value) {
+            for (QuantizationBits bits : values()) {
+                if (bits.value == value) {
+                    return bits;
+                }
+            }
+            return FULL_PRECISION;
```
MrFlap, line 95: Why are we not returning an error here?

src/main/java/org/opensearch/knn/index/engine/Encoder.java
```diff
@@ -7,12 +7,89 @@
 
 import org.opensearch.knn.index.mapper.CompressionLevel;
 
+import java.util.Locale;
+import java.util.Set;
+
+import static org.opensearch.knn.common.KNNConstants.ENCODER_BINARY;
+import static org.opensearch.knn.common.KNNConstants.ENCODER_FLAT;
+import static org.opensearch.knn.common.KNNConstants.ENCODER_PQ;
+import static org.opensearch.knn.common.KNNConstants.ENCODER_SQ;
```
shatejas, line 16: Do you really need public constants for these? I see you are using that in unit tests - personally I don't use public ones in unit test, the idea is unit tests should fail if the constant changes

src/main/java/org/opensearch/knn/index/engine/Encoder.java
```diff
@@ -7,12 +7,89 @@
...
+     * Serialization-compatible: getValue() returns the int stored in FieldInfo/SQConfig.
+     */
+    enum QuantizationBits {
+        ONE(1, CompressionLevel.x32),
+        TWO(2, CompressionLevel.x16),
+        FOUR(4, CompressionLevel.x8),
+        SEVEN(7, CompressionLevel.x4),
+        SIXTEEN(16, CompressionLevel.x2),
+        FULL_PRECISION(32, CompressionLevel.x1),
+        NOT_APPLICABLE(-1, CompressionLevel.NOT_CONFIGURED);
```
shatejas: nit: we can try and avoid NOT_APPLICABLE, a null can be represented as the absence

src/main/java/org/opensearch/knn/index/engine/Encoder.java
```diff
@@ -7,12 +7,89 @@
...
+     * Unified quantization bits enum. Maps bit widths to compression levels.
+     * Serialization-compatible: getValue() returns the int stored in FieldInfo/SQConfig.
+     */
+    enum QuantizationBits {
+        ONE(1, CompressionLevel.x32),
+        TWO(2, CompressionLevel.x16),
+        FOUR(4, CompressionLevel.x8),
+        SEVEN(7, CompressionLevel.x4),
+        SIXTEEN(16, CompressionLevel.x2),
+        FULL_PRECISION(32, CompressionLevel.x1),
```
shatejas: is it quantization if its full precision?

src/main/java/org/opensearch/knn/index/engine/Encoder.java
```diff
@@ -7,12 +7,89 @@
...
+import java.util.Set;
+
+import static org.opensearch.knn.common.KNNConstants.ENCODER_BINARY;
+import static org.opensearch.knn.common.KNNConstants.ENCODER_FLAT;
+import static org.opensearch.knn.common.KNNConstants.ENCODER_PQ;
+import static org.opensearch.knn.common.KNNConstants.ENCODER_SQ;
+
 /**
  * Interface representing an encoder. An encoder generally refers to a vector quantizer.
  */
```
shatejas, line 28: nit: it would be nice to formally define the purpose of the interface, standalone this feels like a internal request object that holds valid values - this helps with future modifications if any

Also some methods like getMethodComponent in here are a but confusing

src/main/java/org/opensearch/knn/index/engine/faiss/FaissSQEncoder.java
shatejas, line 86: Do we still need this with introduction of QuantizedBits?

shatejas, line 100: conflates with getSupportedBits, there is a risk of inconsistencies here

src/main/java/org/opensearch/knn/index/engine/ResolvedIndexSpec.java
```diff
@@ -0,0 +1,127 @@
...
+
+    private boolean isMethodFlat() {
+        return METHOD_FLAT.equals(methodName);
+    }
+
+    private boolean isQuantizedIndex() {
+        return CompressionLevel.isConfigured(compressionLevel)
+            && compressionLevel != CompressionLevel.x1
+            && compressionLevel != CompressionLevel.x2;
```
shatejas, line 127: nit: 2x is non quantized? can we have comment of what is considered quantized

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldType.java
```diff
@@ -75,6 +76,7 @@ public class KNNVectorFieldType extends MappedFieldType {
      */
     boolean memoryOptimizedSearchAvailable;
     Version indexCreatedVersion;
+    ResolvedIndexSpec resolvedSpec;
```
shatejas, line 79: nit: do we need indexVersion in both resolvedSpec and here?

src/main/java/org/opensearch/knn/index/engine/ResolvedIndexSpec.java
```diff
@@ -0,0 +1,127 @@
...
+import org.opensearch.knn.index.VectorDataType;
+import org.opensearch.knn.index.mapper.CompressionLevel;
+import org.opensearch.knn.index.mapper.Mode;
+import org.opensearch.knn.index.query.rescore.RescoreContext;
+
+import static org.opensearch.knn.common.KNNConstants.METHOD_FLAT;
+import static org.opensearch.knn.common.KNNConstants.METHOD_HNSW;
+
+/**
+ * Immutable value object holding resolved index configuration.
```
shatejas, line 20: nit: Lets be a bit more clear in terms of semantics - what is index configuration?

The fields, except for index version is related to mappings. My mental model of configuration would be 
```
ResolvedIndexSpec {
    ResolvedVectorMappings {
      ......all mapping fields
   }
  
   version,
   //maybe nondynamic settings 
}
```

Let me know what you think

shatejas (commented):
Just for some context, 

How/Where is the ResolveSpec stored, is in inside EngineMapper? Will we be able to use this in query path?

Vikasht34:
I briefly looked at this PR refactoring , couple of good merits for this 
1. I really liked ResolvedIndexSpec , which is kind of imutable and will server a source of truth 
2. We are removing all the external checks and putting into methods , which is quite good move.


I don't have any major concern on this PR. There are some nits which I will post. Overeall this PR kills it. Thanks @kotwanikunal

Vikasht34:
> Just for some context,
> 
> How/Where is the ResolveSpec stored, is in inside EngineMapper? Will we be able to use this in query path?

As far I can see this is In memeory only and not serilized to disk, It's part of KNNVectorFieldType and stored as part of mapper service , It's reconstructed from the mapping metadata every time the index is opened.

src/main/java/org/opensearch/knn/index/engine/faiss/FaissSQEncoder.java
kotwanikunal: Yes for now. Multiple callers operate on raw parameter maps before RIS exists (FaissCodecFormatResolver, EngineFieldMapper, RemoteIndexBuildStrategy, KNNVectorFieldType). Gets removed as subsequent PRs wire RIS into those paths. TODO already in FaissHNSWMethod.

kotwanikunal: Actual bits resolution happens in AbstractKNNMethod. Added a todo - will clean up with the next set of changes.

src/main/java/org/opensearch/knn/index/engine/Encoder.java
kotwanikunal: EncoderType anchors to KNNConstants which are the wire-format strings stored in `FieldInfo`. If the constant changes, the enum tracks it. I am intentionally tying it up with these constants. We can revisit the location of these if you like - I think that is what creates the confusion. We do reference these through the same constants right now.

src/main/java/org/opensearch/knn/index/engine/Encoder.java
kotwanikunal: Agreed, added structured javadoc. Two main purposes with the current design: 

1. Type classification (EncoderType/QuantizationBits driving RIS decisions)
2. Method component integration (getMethodComponent/calculateCompressionLevel for the existing resolution framework). 
 
Latter is legacy but needed for BwC.

src/main/java/org/opensearch/knn/index/engine/Encoder.java
kotwanikunal: Fair point. It is the identity value: original 32-bit float, no quantization. Added javadoc. I can rename it no NO_QUANTIZATION for context and readibility - but we do need some centralization of this logic.

src/main/java/org/opensearch/knn/index/engine/Encoder.java
kotwanikunal: NOT_APPLICABLE is for PQ which uses codebook quantization, which is not null. Also null would break the enum usage and force null checks everywhere. Matches existing conventions (CompressionLevel.NOT_CONFIGURED, Mode.NOT_CONFIGURED). Added a clarifying javadoc.

src/main/java/org/opensearch/knn/index/engine/ResolvedIndexSpec.java
kotwanikunal: RIS is the resolved vector mapping. Wrapping it just to separate version creates ambiguity when RIS methods need version internally (`getRescoreContext` uses it for threshold decisions). Flat is simpler to work through currently. But we can evolve it over time as the decisions will be centralized.

src/main/java/org/opensearch/knn/index/engine/ResolvedIndexSpec.java
kotwanikunal: Added javadoc.

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldType.java
kotwanikunal: Version duplication is transitional. KNNVectorFieldType is used by paths that don't have RIS yet (model-based indices, training). Removed once RIS is non-null everywhere in follow-up PRs.

src/main/java/org/opensearch/knn/index/engine/ResolvedIndexSpec.java
kotwanikunal: Also related: https://github.com/opensearch-project/k-NN/pull/3330/changes#r3276753591

kotwanikunal:
@navneet1v / @shatejas / @Vikasht34 - Mind taking another look?

src/main/java/org/opensearch/knn/index/engine/Encoder.java
shatejas: using NOT_APPLICABLE in place of PQ doesn't sound right, moreover PQ is also a quantization and it should be supported here, if the wiring for code is complex we should just add a validation and throw for now

Moreover not configured is different from not applicable, one is indicating absence from user other is a replacement. I am fine with it for now but we should get the semantics right in terms of mapping compression not configured to not applicable

src/main/java/org/opensearch/knn/index/engine/Encoder.java
shatejas: I was hinting towards dropping it completely but not sure what the implications will be, will rely on your judgement

src/main/java/org/opensearch/knn/index/engine/AbstractKNNMethod.java
```diff
@@ -140,6 +142,42 @@ public KNNLibraryIndexingContext getKNNLibraryIndexingContext(
...
+
+    private ResolvedIndexSpec buildResolvedIndexSpec(KNNMethodContext knnMethodContext, KNNMethodConfigContext knnMethodConfigContext) {
+        Encoder.EncoderType encoderType = Encoder.EncoderType.FLAT;
+        Encoder.QuantizationBits quantizationBits = Encoder.QuantizationBits.FULL_PRECISION;
+
+        Map<String, Object> methodParams = knnMethodContext.getMethodComponentContext().getParameters();
+        if (methodParams != null && methodParams.containsKey(METHOD_ENCODER_PARAMETER)) {
+            Object encoderObj = methodParams.get(METHOD_ENCODER_PARAMETER);
+            if (encoderObj instanceof MethodComponentContext) {
+                MethodComponentContext encoderCtx = (MethodComponentContext) encoderObj;
```
shatejas:
```suggestion
            if (encoderObj instanceof MethodComponentContext encoderCtx) {
```

src/main/java/org/opensearch/knn/index/engine/AbstractKNNMethod.java
```diff
@@ -140,6 +142,42 @@ public KNNLibraryIndexingContext getKNNLibraryIndexingContext(
...
+                encoderType = Encoder.EncoderType.fromName(encoderCtx.getName());
+                Object bitsObj = encoderCtx.getParameters().get(KNNConstants.SQ_BITS);
+                if (bitsObj instanceof Integer) {
+                    quantizationBits = Encoder.QuantizationBits.fromValue((Integer) bitsObj);
```
shatejas:
```suggestion
                if (bitsObj instanceof Integer bitsInt) {
                    quantizationBits = Encoder.QuantizationBits.fromValue(bitsInt);
```

src/main/java/org/opensearch/knn/index/engine/ResolvedIndexSpec.java
shatejas: Got it
nit: maybe rename this to ResolvedVectorMappingSpec to be specific

src/main/java/org/opensearch/knn/index/engine/AbstractKNNMethod.java
```diff
@@ -140,6 +142,42 @@ public KNNLibraryIndexingContext getKNNLibraryIndexingContext(
...
+                } else if (encoderType == Encoder.EncoderType.FLAT) {
+                    quantizationBits = Encoder.QuantizationBits.FULL_PRECISION;
+                } else if (encoderType == Encoder.EncoderType.BQ) {
+                    quantizationBits = Encoder.QuantizationBits.ONE;
+                }
+            }
+        }
```
navneet1v, line 167: @kotwanikunal question here: If index doesn't have encoder present and only compression level present then why we are not setting the encoder here?

src/main/java/org/opensearch/knn/index/engine/AbstractKNNMethod.java
```diff
@@ -140,6 +142,42 @@ public KNNLibraryIndexingContext getKNNLibraryIndexingContext(
...
+        Integer dimension = knnMethodConfigContext.getDimension();
+        return ResolvedIndexSpec.builder()
+            .engine(knnMethodContext.getKnnEngine())
+            .methodName(knnMethodContext.getMethodComponentContext().getName())
+            .encoderType(encoderType)
+            .quantizationBits(quantizationBits)
+            .compressionLevel(knnMethodConfigContext.getCompressionLevel())
+            .mode(knnMethodConfigContext.getMode())
+            .vectorDataType(knnMethodConfigContext.getVectorDataType())
+            .dimension(dimension != null ? dimension : 0)
```
navneet1v, line 179: why we think dimension here will be null? we should not take any assumption here. If its null then we should throw error here.

src/main/java/org/opensearch/knn/index/engine/AbstractKNNMethod.java
```diff
@@ -140,6 +142,42 @@ public KNNLibraryIndexingContext getKNNLibraryIndexingContext(
...
+            }
+        }
+
+        Integer dimension = knnMethodConfigContext.getDimension();
+        return ResolvedIndexSpec.builder()
+            .engine(knnMethodContext.getKnnEngine())
+            .methodName(knnMethodContext.getMethodComponentContext().getName())
+            .encoderType(encoderType)
+            .quantizationBits(quantizationBits)
+            .compressionLevel(knnMethodConfigContext.getCompressionLevel())
```
navneet1v, line 176: where do we make sure that compression level is matching the Encoder config?

src/main/java/org/opensearch/knn/index/engine/KNNLibraryIndexingContext.java
```diff
@@ -63,4 +64,16 @@ public interface KNNLibraryIndexingContext {
...
+    /**
+     * Get the resolved index spec containing pre-computed behavioral decisions.
+     * May be null for model-based indices, training indices, or flat-index paths
+     * where full resolution does not run.
+     *
+     * @return ResolvedIndexSpec or null
+     */
+    @Nullable
+    default ResolvedIndexSpec getResolvedSpec() {
+        return null;
```
navneet1v: lets not give default implementation of this. Since we want every library to provide the implementation of this function. We should avoid returning nulls here so that we don't fallback to old methods of figuring out encoders and compression levels

src/main/java/org/opensearch/knn/index/engine/KNNLibraryIndexingContextImpl.java
```diff
@@ -31,6 +32,8 @@ public class KNNLibraryIndexingContextImpl implements KNNLibraryIndexingContext
     @Builder.Default
     private QuantizationConfig quantizationConfig = QuantizationConfig.EMPTY;
     private Function<TrainingConfigValidationInput, TrainingConfigValidationOutput> trainingConfigValidationSetup;
+    @Nullable
```
navneet1v: No Nullable here. This object lets guarantee it will be there.

src/main/java/org/opensearch/knn/index/engine/ResolvedIndexSpec.java
navneet1v: given it is specific to faiss, and its temporary lets make sure function name reflects that its for faiss. Then Java docs should reflect that it is temporary and should not be used. This way atleast we know this should be removed.

src/main/java/org/opensearch/knn/index/mapper/EngineFieldMapper.java
navneet1v: @kotwanikunal can we remove the null path in this PR only ?

src/main/java/org/opensearch/knn/index/engine/Encoder.java
kotwanikunal: This defaults to `FULL_PRECISION` for unknown values for compatibility during rolling upgrades. If a newer version writes a bits value that an older node doesn't recognize, defaulting to full precision is safe.

src/main/java/org/opensearch/knn/index/engine/AbstractKNNMethod.java
kotwanikunal: By this point, method resolution has already completed upstream (MethodResolver) and the MethodComponentContext has both encoder and compression consistently resolved.

src/main/java/org/opensearch/knn/index/engine/AbstractKNNMethod.java
kotwanikunal: Same answer as above. `MethodResolver.resolveMethod()` handles the cross-validation between compression_level and encoder params before this code runs. If a user specifies conflicting values - the resolver rejects it at mapping time. By the time we reach buildResolvedIndexSpec, the values are already validated as consistent.

src/main/java/org/opensearch/knn/index/engine/AbstractKNNMethod.java
kotwanikunal: Dimension is null during model training flows where the actual dimension comes from training data, not mapping params. Throwing here breaks model training. Will add a comment.

src/main/java/org/opensearch/knn/index/engine/ResolvedIndexSpec.java
kotwanikunal: Thought about this. The fields are mostly mapping-derived (engine, method, encoder, compression, mode, datatype, dimension) but the behavioral methods it exposes are index concerns (codec format routing, rescore thresholds, MOS eligibility). `indexVersionCreated` specifically is an index property, not a mapping property and the methods use it for version-gated decisions that span mapping + index config.

Keeping ResolvedIndexSpec since "index" is the right scope: codec is per field for an index, version is for the index, and the object lives on KNNVectorFieldType which is also directly for the index.

src/main/java/org/opensearch/knn/index/engine/Encoder.java
kotwanikunal: Makes sense. Thought through again - removed it.

src/main/java/org/opensearch/knn/index/engine/ResolvedIndexSpec.java
shatejas: Just based on the discussion, it still doesn't seem that the semantics are clear. We are conflating index and mapping. For instance if RIS is created per mapping then isMemoryOptimizedEligible behavior living inside doesn't seem right. that mapping information isn't enough to decide if its MOS eligible or not. what if there is another RIS which is nmslib or lucene, should the index be eligible?

We can probably defer the discussion for now since I do believe this is simplifying the code base right now and is a step in right direction considering current code base. Having said that do think about this a bit more since having fuzzy semantics will probably not ensure that the code base remains simplified in future

shatejas (dismissed):

Vikasht34 (dismissed):

Vikasht34 (approved):

kotwanikunal:
>Just based on the discussion, it still doesn't seem that the semantics are clear. We are conflating index and mapping. For instance if RIS is created per mapping then isMemoryOptimizedEligible behavior living inside doesn't seem right. that mapping information isn't enough to decide if its MOS eligible or not. what if there is another RIS which is nmslib or lucene, should the index be eligible?

>We can probably defer the discussion for now since I do believe this is simplifying the code base right now and is a step in right direction considering current code base. Having said that do think about this a bit more since having fuzzy semantics will probably not ensure that the code base remains simplified in future


Fair points. RIS is scoped per knn_vector field (one per mapping definition). isMemoryOptimizedEligible answers whether this field's configuration is compatible with MOS, not whether MOS should actually run. The execution decision lives in KNNWeight/MemoryOptimizedSearchSupportSpec which considers cluster settings, graph availability, and the specific field being searched.

An index with both a Faiss field and a Lucene field would have separate RIS instances, each answering for their own field.

Agreed the boundary between "field compatibility" and "execution decisions" needs to be clearer as this class evolves. For now, RIS centralizes what was previously scattered across 6+ call sites. As we wire more consumers through it, the separation will either solidify naturally or we split it.

src/main/java/org/opensearch/knn/index/engine/AbstractKNNMethod.java
navneet1v: Lets make sure that we add such things in java docs. Please take that as a followup of this PR.

navneet1v (approved):

navneet1v:
Approved the code, since @shatejas has earlier approved it and he is OOTO.

PR #3331, opened 2026-05-19
Integrated proper ef_search functionality into MOS and Lucene with oversample_factor

### LINKED ISSUE #3272: [BUG] Memory optimized Search doesn't honor EF Search
## Description
I was going through the code for MOS, and what I observed was we are not using ef_search any where. When we create our collector in MemoryOptimizedSearch weight class we pass the K, which eventually gets used as ef_search in HNSWGraphSearcher.

## Expectation
Ideally we should do something similar we did for Lucene engine in KNN plugin where we use EF_Search has a value of K.

Ref: https://github.com/opensearch-project/k-NN/issues/2940

MrFlap:
### Description
Implementation of MOS does not honor ef_search, and implementation of lucene engine rescores on all ef_search docs if ef_search > oversampledK. Fixes both issues.

1. Lucene : k=100, ef_search=10000 on quantized index
From the 1st phase ANN search, result size would be 10000. Then the 10k vectors will be passed to exact searcher. Ideally, we should cut down the result to 200 (assuming over sampling factor is 2, hence 100 * 2 = 200), then only running exact search on 200 vectors. 

2. MOS with Faiss : k=100, ef_search=1000, but we're using 100 as a ANN heap size.
We're not honoring ef_search that was given. We should 1000 as a heap size during ANN search instead of 100.

### Related Issues
Resolves #3272 

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/query/KNNQueryFactory.java
+++ b/src/main/java/org/opensearch/knn/index/query/KNNQueryFactory.java
@@ -148,7 +148,7 @@ public class KNNQueryFactory extends BaseQueryFactory {
                 expandNested,
                 vectorDataType,
                 k,
-                needsRescore
+                needsRescore ? overSampledK : RescoreContext.NO_RESCORE_NEEDED
             )
         );
 
@@ -196,12 +196,12 @@ public class KNNQueryFactory extends BaseQueryFactory {
         final boolean expandNested,
         @NonNull final VectorDataType vectorDataType,
         final int k,
-        final boolean needsRescore
+        final int rescoreK
     ) {
         if (parentFilter == null) {
             assert expandNested == false : "expandNested is allowed to be true only for nested fields.";
             return vectorDataType == VectorDataType.FLOAT
-                ? new OSKnnFloatVectorQuery(fieldName, floatQueryVector, luceneK, filterQuery, k, needsRescore)
+                ? new OSKnnFloatVectorQuery(fieldName, floatQueryVector, luceneK, filterQuery, k, rescoreK)
                 : new OSKnnByteVectorQuery(fieldName, byteQueryVector, luceneK, filterQuery, k);
         }
         // If parentFilter is not null, it is a nested query. Therefore, we delegate creation of query to {@link
@@ -216,7 +216,7 @@ public class KNNQueryFactory extends BaseQueryFactory {
                 parentFilter,
                 expandNested,
                 k,
-                needsRescore
+                rescoreK
             )
             : NestedKnnVectorQueryFactory.createNestedKnnVectorQuery(
                 fieldName,
--- a/src/main/java/org/opensearch/knn/index/query/lucenelib/OSDiversifyingChildrenFloatKnnVectorQuery.java
+++ b/src/main/java/org/opensearch/knn/index/query/lucenelib/OSDiversifyingChildrenFloatKnnVectorQuery.java
@@ -9,6 +9,7 @@ import org.apache.lucene.search.Query;
 import org.apache.lucene.search.TopDocs;
 import org.apache.lucene.search.join.BitSetProducer;
 import org.apache.lucene.search.join.DiversifyingChildrenFloatKnnVectorQuery;
+import org.opensearch.knn.index.query.rescore.RescoreContext;
 
 /**
  * OpenSearch wrapper around Lucene's DiversifyingChildrenFloatKnnVectorQuery that customizes
@@ -19,7 +20,7 @@ import org.apache.lucene.search.join.DiversifyingChildrenFloatKnnVectorQuery;
  */
 public final class OSDiversifyingChildrenFloatKnnVectorQuery extends DiversifyingChildrenFloatKnnVectorQuery {
     private final int k;
-    private final boolean needsRescore;
+    private final int rescoreK;
     private final boolean expandNestedDocs;
 
     public OSDiversifyingChildrenFloatKnnVectorQuery(
@@ -29,9 +30,9 @@ public final class OSDiversifyingChildrenFloatKnnVectorQuery extends Diversifyin
         final int luceneK,
         final BitSetProducer parentFilter,
         final int k,
-        final boolean needsRescore
+        final int rescoreK
     ) {
-        this(fieldName, vector, filterQuery, luceneK, parentFilter, k, needsRescore, false);
+        this(fieldName, vector, filterQuery, luceneK, parentFilter, k, rescoreK, false);
     }
 
     public OSDiversifyingChildrenFloatKnnVectorQuery(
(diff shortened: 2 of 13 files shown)


src/main/java/org/opensearch/knn/index/query/lucenelib/OSDiversifyingChildrenFloatKnnVectorQuery.java
```diff
@@ -41,22 +41,22 @@ public OSDiversifyingChildrenFloatKnnVectorQuery(
...
     }
 
     @Override
     protected TopDocs mergeLeafResults(TopDocs[] perLeafResults) {
         // TODO: Fix this if condition after adding rescoring logic inside ExpandNestedDocsQuery when rescoring is enabled
-        if (needsRescore && !expandNestedDocs) {
-
-            // When rescoring is enabled, we need to return all the oversampled k results to rescore which are later reduced to k
-            return super.mergeLeafResults(perLeafResults);
+        if (rescoreK > 0 && !expandNestedDocs) {
```
0ctopus13prime: Could you separate the magic number `0` as a constant?
I think it would be easier to read out `if (rescoreK > NO_NEED_RESCORE ...)` some thing like this

src/main/java/org/opensearch/knn/index/query/lucenelib/OSKnnFloatVectorQuery.java
```diff
@@ -18,27 +18,27 @@
...
+        this.rescoreK = rescoreK;
     }
 
     @Override
     protected TopDocs mergeLeafResults(TopDocs[] perLeafResults) {
-        if (needsRescore) {
-
-            // When rescoring is enabled, we need to return all the oversampled k results to rescore which are later reduced to k
-            return super.mergeLeafResults(perLeafResults);
+        if (rescoreK > 0) {
```
0ctopus13prime: Same here, could you have a constant for `0`?

src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java
```diff
@@ -204,26 +213,56 @@ private Integer getFirstPassK(final boolean isShardLevelRescoringDisabled) {
...
      *
      * @param topDocs the top documents
-     * @return the total number of documents in the topDocs
+     * @param k the user's requested result count
+     * @return the topN value to pass to TopDocs.merge
      */
-    private int getTotalTopDoc(TopDocs[] topDocs) {
-        if (knnQuery.isMemoryOptimizedSearch() || expandNestedDocs) {
+    private int getMergeTopN(TopDocs[] topDocs, int k) {
+        if (expandNestedDocs) {
```
0ctopus13prime: Can we simplify as below

```
if (expandNestedDocs || (knnQuery.isMemoryOptimizedSearch() && getEffectiveK(k) == k) {
            int sum = 0;
            for (TopDocs topDoc : topDocs) {
                sum += topDoc.scoreDocs.length;
            }
            return sum;
}

return k;
```

src/main/java/org/opensearch/knn/index/query/KNNQueryFactory.java
```diff
@@ -148,7 +148,7 @@ public static Query create(CreateQueryRequest createQueryRequest) {
                 expandNested,
                 vectorDataType,
                 k,
-                needsRescore
+                needsRescore ? overSampledK : 0
```
0ctopus13prime: Same here. Can we use constant over raw `0`?

src/test/java/org/opensearch/knn/index/query/lucenelib/NestedKnnVectorQueryFactoryTests.java
```diff
@@ -41,7 +41,7 @@ public void testCreate_whenCalled_thenCreateQuery() {
         );
 
         ExpandNestedDocsQuery expectedFloatQuery = new ExpandNestedDocsQuery.ExpandNestedDocsQueryBuilder().internalNestedKnnVectorQuery(
-            new InternalNestedKnnFloatVectorQuery(fieldName, floatVectors, queryFilter, luceneK, parentFilter, k, false)
+            new InternalNestedKnnFloatVectorQuery(fieldName, floatVectors, queryFilter, luceneK, parentFilter, k, 0)
```
0ctopus13prime: Also please change all `0` to constant in tests

0ctopus13prime:
Curious, with FP32 in Lucene engine, when k=100, ef_search=1000 was given, then will we see 1000 results?

MrFlap:
> Curious, with FP32 in Lucene engine, when k=100, ef_search=1000 was given, then will we see 1000 results?

@0ctopus13prime We will search for 1000 and then only take top 100 scores from there.

src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java
```diff
@@ -204,26 +213,56 @@ private Integer getFirstPassK(final boolean isShardLevelRescoringDisabled) {
...
-     * The theoretical maximum value this method could return is Integer.MAX_VALUE,
-     * as a single shard cannot have more documents than Integer.MAX_VALUE.
+     * For memory-optimized search, returns max(k, ef_search) so the HNSW graph explores a wider
+     * candidate set. Results are trimmed back to k without a full-precision rescore.
+     * For non-memory-optimized search, ef_search is passed via methodParameters to the native engine
+     * directly, so no expansion is needed here.
+     *
+     * @return expanded k if ef_search > k for memory-optimized search, otherwise the original k
+     */
+    private int getEffectiveK(final int k) {
```
naveentatikonda, line 223: getEffectiveK() is invoked at lines 115, 190, and conditionally inside getMergeTopN (line 224). Each call invokes IndexHyperParametersUtil.getHNSWEFSearchValue, which performs an index-setting lookup. Can you compute it once at the start of createWeight() and reuse it ?

src/main/java/org/opensearch/knn/index/query/nativelib/NativeEngineKnnVectorQuery.java
```diff
@@ -204,26 +213,56 @@ private Integer getFirstPassK(final boolean isShardLevelRescoringDisabled) {
...
+     * For non-MOS without expandNestedDocs: uses k (already trimmed by reduceToTopK).
      *
      * @param topDocs the top documents
-     * @return the total number of documents in the topDocs
+     * @param k the user's requested result count
+     * @return the topN value to pass to TopDocs.merge
      */
-    private int getTotalTopDoc(TopDocs[] topDocs) {
-        if (knnQuery.isMemoryOptimizedSearch() || expandNestedDocs) {
+    private int getMergeTopN(TopDocs[] topDocs, int k) {
```
naveentatikonda: pls simplify this, same loop repeated twice

naveentatikonda (approved):
LGTM!

0ctopus13prime (approved):

0ctopus13prime (dismissed):

naveentatikonda (dismissed):

0ctopus13prime (approved):

naveentatikonda (approved):

PR #3337, opened 2026-05-20
Rescore radial search quantized complete

0ctopus13prime:
### Description
Radial search (max_distance/min_score) was previously blocked for all quantized indices due to scoring inaccuracy from quantization error. This PR enables radial search on 1-bit SQ (32x compression) indices by introducing RescoreRadialSearchQuery — a wrapper query that runs quantized radial search as a first pass, then rescores
candidates using full-precision vectors to filter out false positives (vectors whose quantized score fell within the radius but
whose true score does not). 
Results are capped at 10k per segment. Supports both Faiss and Lucene engines across all space types (L2, cosine, inner product).

### Related Issues
N/A

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/query/RescoreRadialSearchQuery.java
+++ b/src/main/java/org/opensearch/knn/index/query/RescoreRadialSearchQuery.java
@@ -5,6 +5,7 @@
 
 package org.opensearch.knn.index.query;
 
+import com.google.common.annotations.VisibleForTesting;
 import lombok.EqualsAndHashCode;
 import lombok.Getter;
 import org.apache.lucene.index.LeafReaderContext;
@@ -14,19 +15,15 @@ import org.apache.lucene.search.Explanation;
 import org.apache.lucene.search.IndexSearcher;
 import org.apache.lucene.search.Query;
 import org.apache.lucene.search.QueryVisitor;
-import org.apache.lucene.search.ScoreDoc;
 import org.apache.lucene.search.ScoreMode;
 import org.apache.lucene.search.Scorer;
 import org.apache.lucene.search.ScorerSupplier;
 import org.apache.lucene.search.TopDocs;
-import org.apache.lucene.search.TotalHits;
+import org.apache.lucene.search.TopKnnCollector;
 import org.apache.lucene.search.Weight;
 import org.opensearch.knn.index.query.exactsearch.ExactSearcher;
-import org.opensearch.knn.indices.ModelDao;
 
 import java.io.IOException;
-import java.util.ArrayList;
-import java.util.List;
 import java.util.Objects;
 
 /**
@@ -52,6 +49,7 @@ import java.util.Objects;
 @Getter
 @EqualsAndHashCode(callSuper = false)
 public class RescoreRadialSearchQuery extends Query {
+    private static ExactSearcher EXACT_SEARCHER_SINGLETON;
 
     /** The inner radial search query that operates on quantized vectors. */
     private final Query innerQuery;
@@ -78,6 +76,14 @@ public class RescoreRadialSearchQuery extends Query {
     private final boolean memoryOptimizedSearchEnabled;
 
     /**
+     * Maximum number of results to retain after rescoring.
+     * Derived from the index-level {@code max_result_window} setting when available,
+     * otherwise defaults to {@code MAX_RESULTS_RADIAL_RESCORING}.
+     * All first-pass candidates are still scored, but only the top results up to this cap are kept.
+     */
+    private final int maxResultsSize;
+
+    /**
      * Constructs a new rescoring wrapper for radial search on a quantized index.
      *
      * @param innerQuery                   the inner radial search query (must not be null)
@@ -91,13 +97,21 @@ public class RescoreRadialSearchQuery extends Query {
         final String field,
         final float[] queryVector,
         float radius,
-        boolean memoryOptimizedSearchEnabled
+        final boolean memoryOptimizedSearchEnabled,
+        final int maxResultsSize
     ) {
         this.innerQuery = Objects.requireNonNull(innerQuery);
         this.field = Objects.requireNonNull(field);
         this.queryVector = Objects.requireNonNull(queryVector);
         this.radius = radius;
         this.memoryOptimizedSearchEnabled = memoryOptimizedSearchEnabled;
+        this.maxResultsSize = maxResultsSize;
+        Objects.requireNonNull(EXACT_SEARCHER_SINGLETON, "Exact searcher was not initialized.");
+    }
+
+    @VisibleForTesting
+    public static void initialize(final ExactSearcher exactSearcher) {
+        EXACT_SEARCHER_SINGLETON = exactSearcher;
     }
 
     /**
@@ -121,13 +135,7 @@ public class RescoreRadialSearchQuery extends Query {
     public Query rewrite(final IndexSearcher indexSearcher) throws IOException {
         final Query rewritten = innerQuery.rewrite(indexSearcher);
         if (rewritten != innerQuery) {
-            return new RescoreRadialSearchQuery(
-                innerQuery.rewrite(indexSearcher),
-                field,
-                queryVector,
-                radius,
-                memoryOptimizedSearchEnabled
-            );
+            return new RescoreRadialSearchQuery(rewritten, field, queryVector, radius, memoryOptimizedSearchEnabled, maxResultsSize);
         } else {
             return this;
         }
@@ -173,6 +181,7 @@ public class RescoreRadialSearchQuery extends Query {
         private final float[] queryVector;
         private final float radius;
         private final boolean memoryOptimizedSearchEnabled;
+        private final int maxResultsSize;
 
         /**
          * @param query       the parent query (for Lucene's Weight contract)
@@ -188,6 +197,7 @@ public class RescoreRadialSearchQuery extends Query {
             this.queryVector = rescoreQuery.queryVector;
             this.radius = rescoreQuery.radius;
             this.memoryOptimizedSearchEnabled = rescoreQuery.memoryOptimizedSearchEnabled;
+            this.maxResultsSize = rescoreQuery.maxResultsSize;
         }
 
         @Override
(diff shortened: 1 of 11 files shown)


src/main/java/org/opensearch/knn/index/query/RescoreRadialSearchQuery.java
```diff
@@ -61,19 +71,35 @@ public class RescoreRadialSearchQuery extends Query {
      */
     private final float radius;
 
+    /**
+     * Whether memory-optimized search is enabled for this field.
+     * Determines how {@code radius} is interpreted during rescoring:
+     * when true, radius is already a Lucene-normalized score;
+     * when false, radius is a raw distance requiring conversion via {@code KNNEngine.score()}.
+     */
+    private final boolean memoryOptimizedSearchEnabled;
```
kotwanikunal, line 76: This flag is a property of the index configuration (engine + encoder + mode), not the query. Threading it through `RNNQueryFactory.CreateQueryRequest`-> `RescoreRadialSearchQuery` -> `RescoreWeight` -> `ExactSearcherContext` pollutes the query layer. I think we are mixing concerns again.

src/main/java/org/opensearch/knn/index/query/RescoreRadialSearchQuery.java
```diff
@@ -202,5 +266,22 @@ public long cost() {
...
+         */
+        private TopDocs collectTopDocs(final Scorer scorer) throws IOException {
+            // Preallocate array to prevent array doubling
+            final int initialCapacity = (int) Math.min(scorer.iterator().cost(), MAX_RESULTS_RADIAL_RESCORING);
+            final List<ScoreDoc> scoreDocs = new ArrayList<>(initialCapacity);
+
+            // Pull score docs and create TopDocs
+            final DocIdSetIterator iterator = scorer.iterator();
+            int docId;
+            while ((docId = iterator.nextDoc()) != DocIdSetIterator.NO_MORE_DOCS) {
```
kotwanikunal, line 300: We are not checking for capacity here - what if we have more than 10k documents?

src/main/java/org/opensearch/knn/index/query/RescoreRadialSearchQuery.java
```diff
@@ -177,11 +212,40 @@ public ScorerSupplier scorerSupplier(final LeafReaderContext context) throws IOE
...
+                        // All candidates are still scored, but only the top results are kept.
+                        .maxResultWindow((int) Math.min(MAX_RESULTS_RADIAL_RESCORING, matchedDocs.cost()))
+                        .radius(radius)
+                        .field(field)
+                        .floatQueryVector(queryVector)
+                        .isMemoryOptimizedSearchEnabled(memoryOptimizedSearchEnabled)
+                        .build();
+
+                    // 4. Rescore — ExactSearcher handles radius → minScore conversion internally
+                    final ExactSearcher exactSearcher = new ExactSearcher(ModelDao.OpenSearchKNNModelDao.getInstance());
```
kotwanikunal: Should the exact searcher be instantiated per get call? I have noticed it is a singleton usually.

src/main/java/org/opensearch/knn/index/query/RescoreRadialSearchQuery.java
0ctopus13prime: I have different ideas on 'polluting' thing in here as we're already passing the flag through KNNQuery in other flow. If we are passing the flag via KNNQuery, why not in here.
Also we need to pass `memoryOptimizedSearchEnabled` for exact searcher for score conversion.

src/main/java/org/opensearch/knn/index/query/RescoreRadialSearchQuery.java
0ctopus13prime: Good catch, will fix it

src/main/java/org/opensearch/knn/index/query/RescoreRadialSearchQuery.java
0ctopus13prime: No, will update it

Vikasht34:
Is this PR is ready for Review , I see in Draft? @0ctopus13prime

0ctopus13prime:
@Vikasht34 
Yeah, overall it's ready!

src/main/java/org/opensearch/knn/index/query/RescoreRadialSearchQuery.java
```diff
@@ -177,11 +218,39 @@ public ScorerSupplier scorerSupplier(final LeafReaderContext context) throws IOE
...
+                        // All candidates are still scored, but only the top results are kept.
+                        .maxResultWindow((int) Math.min(MAX_RESULTS_RADIAL_RESCORING, matchedDocs.cost()))
+                        .radius(radius)
+                        .field(field)
+                        .floatQueryVector(queryVector)
+                        .isMemoryOptimizedSearchEnabled(memoryOptimizedSearchEnabled)
+                        .build();
+
+                    // 4. Rescore — ExactSearcher handles radius → minScore conversion internally
+                    final TopDocs rescored = EXACT_SEARCHER_SINGLETON.searchLeaf(context, exactSearcherContext);
```
VijayanB, line 267: Can we add null check for EXACT_SEARCHER_SINGLETON? If never initialized this will throws an NPE with no helpful message

src/main/java/org/opensearch/knn/index/query/RescoreRadialSearchQuery.java
```diff
@@ -177,11 +218,39 @@ public ScorerSupplier scorerSupplier(final LeafReaderContext context) throws IOE
...
+                        return KNNScorer.emptyScorer();
+                    }
+
+                    // 3. Build ExactSearcherContext — rescore with full-precision vectors
+                    final DocIdSetIterator matchedDocs = new TopDocsDISI(firstPassResults);
+                    final ExactSearcher.ExactSearcherContext exactSearcherContext = ExactSearcher.ExactSearcherContext.builder()
+                        .matchedDocsIterator(matchedDocs)
+                        .numberOfMatchedDocs(matchedDocs.cost())
+                        .useQuantizedVectorsForSearch(false)
+                        .k(0)
```
VijayanB: nit: I don't think you need to explicitly set k as 0

src/main/java/org/opensearch/knn/index/query/RescoreRadialSearchQuery.java
0ctopus13prime: Let me add that check in constructor to fail fast. Then we can save ann cost at least

src/main/java/org/opensearch/knn/index/query/RescoreRadialSearchQuery.java
Vikasht34: @0ctopus13prime  Thanks for putting this PR, I feel we can we derive this information already then we should not add extra consutrtor paramemter else it next time if derivable paramenter would be added to constructor and in my opinion that should be we avoid, @kotwanikunal  is already doing lots of refactoring and this is will again add backlog to refactor, why do u think we should not derive isMemorysearchEnabled from already persent info and what makes it pass all the way from createrequest to here.

src/main/java/org/opensearch/knn/index/query/RescoreRadialSearchQuery.java
```diff
@@ -177,11 +218,39 @@ public ScorerSupplier scorerSupplier(final LeafReaderContext context) throws IOE
...
+                        // Cap the maximum number of results retained after rescoring.
+                        // All candidates are still scored, but only the top results are kept.
+                        .maxResultWindow((int) Math.min(MAX_RESULTS_RADIAL_RESCORING, matchedDocs.cost()))
```
Vikasht34: @VijayanB was mentioning that there is alreqady index setting for that  maxResulltWindow, Do u think we can use that and have default?

src/main/java/org/opensearch/knn/index/query/RescoreRadialSearchQuery.java
0ctopus13prime: Yeah, but the only change is to change 

```
return new RescoreRadialSearchQuery(
    innerQuery,
    fieldName,
    vector,
    radius,
    createQueryRequest.isMemoryOptimizedSearchEnabled(),
    maxResultsSize
);
```

to 

```
return new RescoreRadialSearchQuery(
    innerQuery,
    fieldName,
    vector,
    radius,
    MemoryOptimizedSearchSupportSpec.isSupportedFieldType(
      createQueryRequest.getVectorFieldType(), createQueryRequest.getIndexName()),
    maxResultsSize
);
```

But `createQueryRequest` already has the flag, wouldn't be easier just to pass `createQueryRequest.isMemoryOptimizedSearchEnabled()` directly to RescoreRadialSearchQuery?

Is it a part of plan getting rid of `createQueryRequest.isMemoryOptimizedSearchEnabled()`?

src/main/java/org/opensearch/knn/index/query/RescoreRadialSearchQuery.java
0ctopus13prime: Yeah, we can. I'm woking on it, adding tests.

src/main/java/org/opensearch/knn/index/query/RescoreRadialSearchQuery.java
```diff
@@ -177,11 +218,39 @@ public ScorerSupplier scorerSupplier(final LeafReaderContext context) throws IOE
...
+                    }
+
+                    // 2. Collect first-pass results into TopDocs
+                    final TopDocs firstPassResults = collectTopDocs(innerScorer);
+                    if (firstPassResults.scoreDocs.length == 0) {
+                        return KNNScorer.emptyScorer();
+                    }
+
+                    // 3. Build ExactSearcherContext — rescore with full-precision vectors
+                    final DocIdSetIterator matchedDocs = new TopDocsDISI(firstPassResults);
```
shatejas: Was curious why do we need collecting TopDocs again? 

The scorer is obligated to return the results in sorted docId order as per lucene contract so can't we just do?
```
matchedDocs = firstPassScorer.iterator()
```

As far as MAX_RESULTS_RADIAL_RESCORING is concerned innerQuery should handle it. Let me know if I am missing something here

0ctopus13prime: Sure, will update it.

For `MAX_RESULTS_RADIAL_RESCORING`, innerQuery is not handling it, at the extreme case when 0 minScore is given, it's returning more than 10k results. 
So I've imposed that responsibility on its upper level to cap on the maximum possible computation cost.

Vikasht34 (approved):
Looks Good !!

src/main/java/org/opensearch/knn/index/query/RescoreRadialSearchQuery.java
```diff
@@ -177,11 +226,48 @@ public ScorerSupplier scorerSupplier(final LeafReaderContext context) throws IOE
...
+                        .field(field)
+                        .floatQueryVector(queryVector)
+                        .isMemoryOptimizedSearchEnabled(memoryOptimizedSearchEnabled)
+                        .build();
+
+                    // 5. Rescore — ExactSearcher handles radius → minScore conversion internally
+                    final TopDocs rescored = EXACT_SEARCHER_SINGLETON.searchLeaf(context, exactSearcherContext);
+
+                    // 6. Return scorer over rescored results
+                    return new KNNScorer(rescored, boost);
```
shatejas, line 270: One of the potential optimizations here is to have exact searcher return a scorer rather than topdocs. Returning topDocs here is redundant because [TopScoreDocCollector](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/TopScoreDocCollector.java#L94) will do it again. Collecting topDocs forces to [loop through](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/query/exactsearch/ExactSearcher.java#L275) the entire results unnecessarily. This breaks the lazy behavior of lucene and bypasses early termination and leap frogging if at all its needed.

The idea is simple here, wrap the bulkScoring logic in an iterator and return a scorer from exact search level removing the redundant loop for creating TopDocs and packaging it again in KNNScorer. I have been thinking through this idea and see where this scorer is beneficial apart from radial search because exactsearch does not need to have topk either

shatejas (approved):
Looks good for now, we can take scorer optimization as followup

src/main/java/org/opensearch/knn/index/query/RescoreRadialSearchQuery.java
0ctopus13prime: Agree. We can directly pull and sort docs from min-heap without having to convert them to TopDocs since as you mentioned, TopScoreDocCollector will do it again anyway.
And I think the lazy behavior will only be matter when it was used as a sub query, like `Conjunction(TermQuery, Radial)` for example. And in most case I believe radial search will be used as a top level query.

I believe we should find a way to get the entire query tree then decide whether to return iterator wrapper having exact searcher in it. But for that, we probably need core's support.

shatejas: > We can directly pull and sort docs from min-heap without having to convert them to TopDocs

We don't even need a minHeap its additional JVM heap use followed by GC cleanup. Just a scorer which tells how to compute should be enough. We just need to make sure that the DocIdSetIterator of the scorer iterates the docs in sequence. No sorting, No looping through

> And in most case I believe radial search will be used as a top level query.

Correct, it will still avoid the extra minHeap and looping through all results to get TopDocs again to hold those in memory, I understand these are minor things but might matter in a high throughput environments

> I believe we should find a way to get the entire query tree then decide whether to return iterator wrapper having exact searcher in it. But for that, we probably need core's support.

Not necessarily if we abide by the lucene scorer contract, traversing a query tree is not needed.

0ctopus13prime: Hm, I think it's better to have a quick sync on this

But I think believe we still need the min heap when the intermediate result size > 10k though to extract top-10k vectors. If the result from 1st phase < 10k, then yes, it's not needed and this PR already has it.

Also be able to look at the entire query tree is essential in the optimization, as in my view, if radial search is at the top level, then we should run exact search aggressively otherwise we should get benefit from lazy evaluation.

PR #3343, opened 2026-05-27
Fixes RescoreParser to pass the rescore flag

shatejas:
For coordinator-data node setup, rescore set to false is not passed through streams. This causes rescoring to execute even when its not disabled explicitly by user

### Related Issues
N/A

### Testing
`./gradlew run -PnumNodes=2`
`./gradlew startCoordinator`

Request:
```
curl -s 'localhost:9202/test-rescore/_search?explain=true' -H 'Content-Type: application/json' -d '{
    "query": {"knn": {"my_vector": {"vector": [0.5, 1.0, 1.5, 2.0], "k": 3, "rescore": false}}}
  }'
```
Response:
```
Before fix (via coordinator node port 9202):
  
  {
    "_explanation": {
      "value": 0.11764706,
      "description": "the type of knn search executed was Disk-based and the first pass k was 100 with vector dimension of 4, over sampling factor of 1.0, shard level rescoring enabled",
      "details": [
        {
          "value": 0.11764706,
          "description": "the type of knn search executed at leaf was Approximate-NN with vectorDataType = FLOAT, spaceType = l2",
          "details": []
        }
      ]
    }
  }
  
  After fix (via coordinator node port 9202):
  
  {
    "_explanation": {
      "value": 0.11764706,
      "description": "the type of knn search executed was Disk-based and the first pass k was 100 with vector dimension of 4, over sampling factor of 1.0, shard level rescoring disabled",
      "details": [
        {
          "value": 0.11764706,
          "description": "the type of knn search executed at leaf was Approximate-NN with vectorDataType = FLOAT, spaceType = l2",
          "details": []
        }
      ]
    }
  }
```

### Check List

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -807,7 +807,7 @@ run {
 // Task to start a dedicated coordinator node that joins the running test cluster.
 // Start the cluster first: ./gradlew run -PnumNodes=2
 // Then in another terminal: ./gradlew startCoordinator
-// The coordinator node will be available at localhost:9202
+// The coordinator node will be available at localhost:9200+numNodes (e.g. 9202 for 2 nodes)
 task startCoordinator(type: Exec) {
     def distroDir = "${buildDir}/testclusters/integTest-0/distro/${opensearch_version_no_snapshot}-ARCHIVE"
     def coordDir = "${buildDir}/testclusters/coordinator-0"
@@ -824,14 +824,16 @@ task startCoordinator(type: Exec) {
         }
 
         def seedHosts = (0..<_numNodes).collect { "127.0.0.1:${9300 + it}" }.join('", "')
+        def coordHttpPort = 9200 + _numNodes
+        def coordTransportPort = 9300 + _numNodes
         file("${coordDir}/config/opensearch.yml").text = """\
 cluster.name: integTest
 node.name: coordinator-0
 node.roles: []
 discovery.seed_hosts: ["${seedHosts}"]
 network.host: 127.0.0.1
-http.port: 9202
-transport.port: 9302
+http.port: ${coordHttpPort}
+transport.port: ${coordTransportPort}
 path.data: ${coordDir}/data
 path.logs: ${coordDir}/logs
 """
--- a/build.gradle
+++ b/build.gradle
@@ -804,6 +804,45 @@ run {
     }
 }
 
+// Task to start a dedicated coordinator node that joins the running test cluster.
+// Start the cluster first: ./gradlew run -PnumNodes=2
+// Then in another terminal: ./gradlew startCoordinator
+// The coordinator node will be available at localhost:9202
+task startCoordinator(type: Exec) {
+    def distroDir = "${buildDir}/testclusters/integTest-0/distro/${opensearch_version_no_snapshot}-ARCHIVE"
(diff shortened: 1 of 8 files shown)


navneet1v:
@shatejas why this is a problem with coordinator and data node setup? can this not happen with 2 node setup?

shatejas:
> @shatejas why this is a problem with coordinator and data node setup? can this not happen with 2 node setup?

Wasn't able to replicate on local with -PnumNodes=2 setup on local, ideally it should be

Vikasht34 (dismissed):
Looks Good

navneet1v:
> > @shatejas why this is a problem with coordinator and data node setup? can this not happen with 2 node setup?
> 
> Wasn't able to replicate on local with -PnumNodes=2 setup on local, ideally it should be

Did we create multiple shards in the index? because this problem should be replicated there. And reason why I saying this is because we do have multi-node test and we can combine this test in that suite.

src/test/java/org/opensearch/knn/index/query/parser/RescoreParserTests.java
```diff
@@ -27,6 +28,29 @@ public void testStreams() {
         RescoreContext rescoreContext = RescoreContext.builder().oversampleFactor(RescoreContext.DEFAULT_OVERSAMPLE_FACTOR).build();
         validateStreams(rescoreContext);
         validateStreams(null);
+        validateStreams(RescoreContext.EXPLICITLY_DISABLED_RESCORE_CONTEXT);
+    }
+
+    @SneakyThrows
+    public void testStreams_withOlderVersion() {
```
navneet1v, line 35: @shatejas can we have the test for happy path too. and also test where rescore is false and we get false.

build.gradle
```diff
@@ -804,6 +804,47 @@ run {
     }
 }
 
+// Task to start a dedicated coordinator node that joins the running test cluster.
+// Start the cluster first: ./gradlew run -PnumNodes=2
+// Then in another terminal: ./gradlew startCoordinator
```
navneet1v, line 809: can we not combine all this in 1 gradle run? there is already a multinode integ test wondering if we can add it there,

src/test/java/org/opensearch/knn/index/query/parser/RescoreParserTests.java
shatejas: Line 31 is testing the happy path

build.gradle
shatejas: That would be ideal but KNN plugin doesn't control it, plugin plugin uses the OpensearchCluster construct from core via
```
  run {
      useCluster project.testClusters.integTest
  }
```
The numNodes parameter is set on the cluster directly in testClusters.integTest task. This needs change in core gradle plugin if coordinator needs to be supported. To be able to test it I have added it, we can take this up in core as a follow up

shatejas:
> Did we create multiple shards in the index? because this problem should be replicated there. And reason why I saying this is because we do have multi-node test and we can combine this test in that suite.

I checked again and I was able to repro it, my bad, didn't look at the response correctly

navneet1v (dismissed):

navneet1v (approved):

Vikasht34 (approved):

PR #3354, opened 2026-06-02
Check to see if Lucene's search budget has exhausted when deciding to exact search

MrFlap:
### Description
Lucene engine has a check such that it falls back to exact search if we hit a visit limit before finding qualified candidates. This change takes the conditional from Lucene and applies the same logic to Memory Optimized Search.

After this change we will see a more aggressive exact search strategy in MOS when filtered docs are sparse in the graph. We should see a recall gain in low filtering percentages (~1-5%).

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/query/KNNWeight.java
+++ b/src/main/java/org/opensearch/knn/index/query/KNNWeight.java
@@ -198,16 +198,19 @@ public abstract class KNNWeight extends Weight {
             }
         }
         final Integer annResult = knnExplanation.getAnnResult(context.id());
+        final Boolean exhaustedSearch = knnExplanation.getExhaustedSearch(context.id());
         if (annResult != null && annResult == 0 && isMissingNativeEngineFiles(context)) {
             sb.append(KNNConstants.EXACT_SEARCH).append(" since no native engine files are available");
         }
-        if (annResult != null && isFilteredExactSearchRequireAfterANNSearch(cardinality, annResult)) {
+        if (annResult != null && isFilteredExactSearchRequireAfterANNSearch(cardinality, exhaustedSearch, annResult)) {
             boolean isExactSearchDisabled = KNNSettings.isKnnIndexFaissEfficientFilterExactSearchDisabled(knnQuery.getIndexName());
             if (isExactSearchDisabled) {
                 sb.append(KNNConstants.ANN_SEARCH)
                     .append(", it is not falling back to exact search after ")
                     .append(KNNConstants.ANN_SEARCH)
                     .append(" search since exact search is disabled,");
+            } else if (exhaustedSearch) {
+                sb.append(KNNConstants.EXACT_SEARCH).append(" since lucene vector search has exhausted number of steps.");
             } else {
                 sb.append(KNNConstants.EXACT_SEARCH)
                     .append(" since the number of documents returned are less than K = ")
@@ -216,7 +219,7 @@ public abstract class KNNWeight extends Weight {
                     .append(cardinality);
             }
         }
-        if (annResult != null && annResult > 0 && !isFilteredExactSearchRequireAfterANNSearch(cardinality, annResult)) {
+        if (annResult != null && annResult > 0 && !isFilteredExactSearchRequireAfterANNSearch(cardinality, exhaustedSearch, annResult)) {
             sb.append(KNNConstants.ANN_SEARCH);
         }
         sb.append(" with vectorDataType = ").append(knnQuery.getVectorDataType());
@@ -345,13 +348,17 @@ public abstract class KNNWeight extends Weight {
         final TopDocs topDocs = approximateSearch(context, filterBitSet, filterCardinality, k);
         stopStopWatchAndLog(log, annStopWatch, "ANN search", knnQuery.getShardId(), segmentName, knnQuery.getField());
 
+        int annResultsCount = topDocs.scoreDocs.length;
+        boolean annSearchBudgetExhausted = topDocs.totalHits.relation() != TotalHits.Relation.EQUAL_TO;
+
         if (knnQuery.isExplain()) {
-            knnExplanation.addLeafResult(context.id(), topDocs.scoreDocs.length);
+            knnExplanation.addLeafResult(context.id(), annResultsCount);
+            knnExplanation.addExhaustedSearch(context.id(), annSearchBudgetExhausted);
         }
         // See whether we have to perform exact search based on approx search results
         // This is required if there are no native engine files or if approximate search returned
         // results less than K, though we have more than k filtered docs
-        if (isExactSearchRequire(context, filterCardinality, topDocs.scoreDocs.length)) {
+        if (isExactSearchRequire(context, filterCardinality, annSearchBudgetExhausted, annResultsCount)) {
             final BitSetIterator docs = filterWeight != null ? new BitSetIterator(filterBitSet, filterCardinality) : null;
             final TopDocs result = doExactSearch(context, docs, filterCardinality, k);
             return new PerLeafResult(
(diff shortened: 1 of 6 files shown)


navneet1v:
@MrFlap what is the impact of this change? can you please add some details on that part

MrFlap:
@navneet1v the impact of the change is that we will do exact search more aggressively on queries where filtered documents are sparse. Adding this to the PR description. Keeping it as a draft since I still need to write tests and do benchmarking on 1% case to make sure this actually closes the gap.

MrFlap:
> ## PR Code Suggestions ✨
> 
> Latest suggestions up to [fb4a05c](https://github.com/opensearch-project/k-NN/commit/fb4a05c1549525dc63cfb023b5ca674b0b4655f3) Explore these optional code suggestions:
> **Category**	**Suggestion                                                                                                                                    **	**Impact**
> Possible issue 	
> Track exhaustedSearch regardless of explain mode
> 
> **The `exhaustedSearch` flag is only populated when `isExplain()` is true, but it's used unconditionally in `isExactSearchRequire`. This creates a critical bug where exact search decisions depend on whether explain mode is enabled, leading to inconsistent search behavior.**
> 
> [src/main/java/org/opensearch/knn/index/query/KNNWeight.java [354-361]](https://github.com/opensearch-project/k-NN/pull/3354/files#diff-b0eb527e944c912b803db37ed9dd3dfa7689864b77d7146a0d45d7b73138d7b8R354-R361)
> 
> ```diff
> -if (knnQuery.isExplain()) {
> -    knnExplanation.addLeafResult(context.id(), annResultsCount);
> -    knnExplanation.addExhaustedSearch(context.id(), exhaustedSearch);
> -}
> -...
> +knnExplanation.addLeafResult(context.id(), annResultsCount);
> +knnExplanation.addExhaustedSearch(context.id(), exhaustedSearch);
> +
>  if (isExactSearchRequire(context, filterCardinality, exhaustedSearch, annResultsCount)) {
> ```
> 
> Suggestion importance[1-10]: 10
> 
> 	High
> #### Previous suggestions
> Suggestions up to commit [a71fc1f](https://github.com/opensearch-project/k-NN/commit/a71fc1f7b3ec8e0b7099785fb00ac71e28784ec1)

This is just not true. We only insert to knnExplanation for the purpose of explain. That function does not alter search behavior.

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -345,13 +348,17 @@ public PerLeafResult searchLeaf(LeafReaderContext context, int k) throws IOExcep
         final TopDocs topDocs = approximateSearch(context, filterBitSet, filterCardinality, k);
         stopStopWatchAndLog(log, annStopWatch, "ANN search", knnQuery.getShardId(), segmentName, knnQuery.getField());
 
+        int annResultsCount = topDocs.scoreDocs.length;
+        boolean exhaustedSearch = topDocs.totalHits.relation() != TotalHits.Relation.EQUAL_TO;
```
navneet1v: can we rename this to `annSearchBudgetExhausted`.

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -667,13 +674,18 @@ private boolean isExactSearchThresholdSettingSet(int filterThresholdValue) {
      * @param annResultCount Count of Nearest Neighbours we got after doing filtered ANN Search.
      * @return boolean - true if exactSearch needs to be done after ANNSearch.
      */
-    private boolean isExactSearchRequire(final LeafReaderContext context, final int filterIdsCount, final int annResultCount) {
+    private boolean isExactSearchRequire(
```
navneet1v, line 678: can we update the java doc to tell details about what `exhaustedSearch` search means. I understand what it is doing now, but for future lets add some details

navneet1v:
Code looks good to me.

navneet1v (approved):

src/main/java/org/opensearch/knn/index/query/explain/KnnExplanation.java
```diff
@@ -20,6 +20,8 @@ public class KnnExplanation {
 
     private final Map<Object, Integer> annResultPerLeaf;
 
+    private final Map<Object, Boolean> exhaustedSearchPerLeaf;
```
0ctopus13prime, line 23: Can we consider using concurrent Set<Object> instead?
I think not all segments will always add the flag, having `Set` would make things simple, we can just call `contains` in getExhaustedSearch.

src/main/java/org/opensearch/knn/index/query/explain/KnnExplanation.java
```diff
@@ -51,6 +58,10 @@ public Integer getAnnResult(Object leafId) {
         return this.annResultPerLeaf.get(leafId);
     }
 
+    public Boolean getExhaustedSearch(Object leafId) {
+        return this.exhaustedSearchPerLeaf.get(leafId);
```
0ctopus13prime, line 62: With concurrent set, we can just call `exhaustedSearchPerLeaf.contains(leafId)`;
We can save null Boolean hassel

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -198,16 +198,19 @@ private StringBuilder getLeafLevelExplanation(LeafReaderContext context) {
             }
         }
         final Integer annResult = knnExplanation.getAnnResult(context.id());
+        final Boolean exhaustedSearch = knnExplanation.getExhaustedSearch(context.id());
```
0ctopus13prime, line 201: With concurrent set, we can have `exhaustedSearch` as `boolean` instead which I believe will save NPE hassels

0ctopus13prime, line 201: Also could you move this above line 205?

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -345,13 +348,17 @@ public PerLeafResult searchLeaf(LeafReaderContext context, int k) throws IOExcep
         final TopDocs topDocs = approximateSearch(context, filterBitSet, filterCardinality, k);
         stopStopWatchAndLog(log, annStopWatch, "ANN search", knnQuery.getShardId(), segmentName, knnQuery.getField());
 
+        int annResultsCount = topDocs.scoreDocs.length;
+        boolean annSearchBudgetExhausted = topDocs.totalHits.relation() != TotalHits.Relation.EQUAL_TO;
+
         if (knnQuery.isExplain()) {
-            knnExplanation.addLeafResult(context.id(), topDocs.scoreDocs.length);
+            knnExplanation.addLeafResult(context.id(), annResultsCount);
+            knnExplanation.addExhaustedSearch(context.id(), annSearchBudgetExhausted);
```
0ctopus13prime, line 356: hmm.. confusing, shouldn't we always add the flag regardless of `isExplain`?

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
```diff
@@ -345,13 +348,17 @@ public PerLeafResult searchLeaf(LeafReaderContext context, int k) throws IOExcep
         final TopDocs topDocs = approximateSearch(context, filterBitSet, filterCardinality, k);
         stopStopWatchAndLog(log, annStopWatch, "ANN search", knnQuery.getShardId(), segmentName, knnQuery.getField());
 
+        int annResultsCount = topDocs.scoreDocs.length;
+        boolean annSearchBudgetExhausted = topDocs.totalHits.relation() != TotalHits.Relation.EQUAL_TO;
```
0ctopus13prime, line 352: Help me understand, could you point me to the early termination logic in collector?
It's my understanding that collector counts #vector to visit, and if it exceeds filter cardinality, then marks early termination as true to stop neighbor exploration. https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/KnnCollector.java#L39
Is my understanding correct? 
Could you help me understand how this termination logic is enabled?

src/main/java/org/opensearch/knn/index/query/explain/KnnExplanation.java
MrFlap: We will never call null boolean because it is added only with annResultPerLeaf. Kept as map to keep style consistent.

0ctopus13prime: This is not about `style`. If having `Set` is better, then we should have `Set`
Also, are you sure?
What if this line is evaluated as true?
https://github.com/MrFlap/k-NN/blob/29d324ece43da8f48d529cc197ecdc59089ac99d/src/main/java/org/opensearch/knn/index/query/KNNWeight.java#L337

0ctopus13prime: For 5 segments, I think there can be a case where only 3 segments added flags, and 2 segments can bypass it.
As a result exhaustedSearchPerLeaf only has 3 leaf ids. Then we will have null for Boolean right? 🤔

0ctopus13prime: Discussed in offline, the flags will be evaluated only if annResultPerLeaf has ann result for a leaf segment.

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
0ctopus13prime: if annResult != null, then exhaustedSearch will never be null

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
0ctopus13prime: Discussed offline, it's for explain feature.

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
0ctopus13prime: We are setting visit limit in here https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/query/memoryoptsearch/MemoryOptimizedKNNWeight.java#L191

src/main/java/org/opensearch/knn/index/query/KNNWeight.java
0ctopus13prime: It's up to you, won't block this PR

0ctopus13prime (approved):

PR #3361, opened 2026-06-12
Introduces BulkVectorScorer for ExactSearch

### LINKED ISSUE #3348: [RFC] Simplify exact search to return a lucene Scorer
## Description
Currently, ExactSearcher - a construct that performs brute force k-NN search returns [TopDocs](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/query/exactsearch/ExactSearcher.java#L94) which holds the top-k documents. This breaks the lazy flow of lucene's query->weight->scorer approach.

One of the drawbacks of returning TopDocs is that it needs to eagerly iterate through all the documents while maintaining the data structures to hold the topDocs to again wrap it in a scorer via either KNNScorer or DocAndScoreQueryScorer. 

For N segments in a shard there is
1. It create N+1 priority queues (Rescoring needs N+1 while KNNquery might not need it)
2. ScoreDoc[] arrays allocated per segment, then merged into another array
3. Callable + List<Callable> + task executor overhead per query

This creates a GC overhead of short lived objects. This is redundant because lucenes [TopScoreDocCollector](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/TopScoreDocCollector.java#L94) collects the results again with the same priority queue approach with an added benefit of early termination. 

## Proposal
The approach recommended is simple - Drop top-k calculations for exact search and return a scorer instead

### Consequences
**Rescoring**
Exact search is heavily used in rescoring, this will move all rescoring to lazy approach rather than eager score computations.
This means we loose aggressive parallelism and honor slice counts; both of which will have pros and cons

**Won't honor k value**
Current approach of k results won't be honored for any exact search. Theoretically this makes sense since all exact search is expected to loop though all results, having an additional loop to limit it to k value does not makes sense. Note that it will directly honor size values

**Refactoring**
This can involve significant refactoring

Sample implementation for RescoreKNNVectorQuery
- [BulkVectorScorer](https://github.com/shatejas/k-NN/commit/efe2a6e5aa3ad70a99f484c0178c34958562d058)
- [ExactSearcher](https://github.com/shatejas/k-NN/commit/7999c40fed18669b49354f01d8fd00c256755384)

Currently I don't have any benchmarks around this. Needed initial feed back while I benchmark in parallel

shatejas:
### Description

Introduces BulkVectorScorer, a Lucene Scorer implementation that wraps the bulk vector scoring API and encapsulates score filtering logic. This replaces the manual buffer-iteration loops previously spread across ExactSearcher methods, consolidating all exact-search scoring into a single reusable abstraction.

#### Key changes
- **New BulkVectorScorer class** — A Scorer backed by VectorScorer.Bulk that iterates documents in batches, applies a score predicate, and exposes standard 
- DocIdSetIterator semantics (nextDoc, advance, cost). Two factory methods express intent:
    - forKSearch — accepts all scores (used for top-k collection)
    - forRadialSearch — filters to scores ≥ a minimum threshold
- **Simplified ExactSearcher** — Removed searchTopK, searchWithMinScore, and the raw buffer loop in scoreAllDocs. All paths now construct the appropriate BulkVectorScorer and pass it to collectTopK or the simplified scoreAllDocs(Scorer). This eliminates duplicated iteration logic and makes the scoring strategy explicit at the call site.

### Test plan

  - [x] New BulkVectorScorerTests — covers k-search scoring, radial search filtering, filtered doc iteration, advance() semantics (including at/past target),
  cost, initial state, empty docs, and edge cases
  - [x] New ExactSearcherTests — unit tests for exact searcher flows
  - [x] Existing integration tests (MOSFaissFloatIndexIT, FilteredSearchANNSearchIT) validate end-to-end behavior

### Related Issues
Partially Resolves [#3348](https://github.com/opensearch-project/k-NN/issues/3348)

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/query/exactsearch/BulkVectorScorer.java
+++ b/src/main/java/org/opensearch/knn/index/query/exactsearch/BulkVectorScorer.java
@@ -35,11 +35,11 @@ public class BulkVectorScorer extends Scorer {
         this.cost = matchedDocs != null ? matchedDocs.cost() : vectorScorer.iterator().cost();
     }
 
-    public static BulkVectorScorer fullPrecision(VectorScorer vectorScorer, DocIdSetIterator matchedDocs) throws IOException {
+    public static BulkVectorScorer forKSearch(VectorScorer vectorScorer, DocIdSetIterator matchedDocs) throws IOException {
         return new BulkVectorScorer(vectorScorer, matchedDocs, score -> true);
     }
 
-    public static BulkVectorScorer fullPrecision(VectorScorer vectorScorer, DocIdSetIterator matchedDocs, float minScore)
+    public static BulkVectorScorer forRadialSearch(VectorScorer vectorScorer, DocIdSetIterator matchedDocs, float minScore)
         throws IOException {
         return new BulkVectorScorer(vectorScorer, matchedDocs, score -> score >= minScore);
     }
@@ -78,6 +78,9 @@ public class BulkVectorScorer extends Scorer {
 
             @Override
             public int advance(int target) throws IOException {
+                if (currentDocId >= target) {
+                    return currentDocId;
+                }
                 while (true) {
                     int doc = nextDoc();
                     if (doc == NO_MORE_DOCS) {
@@ -108,9 +111,10 @@ public class BulkVectorScorer extends Scorer {
 
     private int scanBufferForMatch() {
         while (currentBatchIdx < buffer.size) {
-            if (scoreFilter.test(buffer.features[currentBatchIdx])) {
+            float score = buffer.features[currentBatchIdx];
+            if (scoreFilter.test(score)) {
                 currentDocId = buffer.docs[currentBatchIdx];
-                currentScore = buffer.features[currentBatchIdx];
+                currentScore = score;
                 currentBatchIdx++;
                 return currentDocId;
             }
--- a/src/main/java/org/opensearch/knn/index/query/exactsearch/BulkVectorScorer.java
+++ b/src/main/java/org/opensearch/knn/index/query/exactsearch/BulkVectorScorer.java
@@ -28,7 +28,7 @@ public class BulkVectorScorer extends Scorer {
     private int currentBatchIdx = 0;
     private float currentScore;
 
-    public BulkVectorScorer(final VectorScorer vectorScorer, final DocIdSetIterator matchedDocs, final Predicate<Float> scoreFilter)
+    private BulkVectorScorer(final VectorScorer vectorScorer, final DocIdSetIterator matchedDocs, final Predicate<Float> scoreFilter)
         throws IOException {
         this.bulkScorer = vectorScorer.bulk(matchedDocs);
         this.scoreFilter = scoreFilter;
--- a/src/main/java/org/opensearch/knn/index/query/exactsearch/BulkVectorScorer.java
+++ b/src/main/java/org/opensearch/knn/index/query/exactsearch/BulkVectorScorer.java
@@ -1,0 +1,121 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.knn.index.query.exactsearch;
+
+import org.apache.lucene.search.DocAndFloatFeatureBuffer;
+import org.apache.lucene.search.DocIdSetIterator;
+import org.apache.lucene.search.Scorer;
+import org.apache.lucene.search.VectorScorer;
+
+import java.io.IOException;
+import java.util.function.Predicate;
+
+/**
+ * A {@link Scorer} that scores documents using bulk vector scoring, yielding only those
+ * whose score satisfies the provided {@link Predicate}.
+ */
+public class BulkVectorScorer extends Scorer {
+
+    private final DocAndFloatFeatureBuffer buffer = new DocAndFloatFeatureBuffer();
+    private final VectorScorer.Bulk bulkScorer;
+    private final Predicate<Float> scoreFilter;
+    private final long cost;
+
+    private int currentDocId = -1;
+    private int currentBatchIdx = 0;
+    private float currentScore;
+
+    public BulkVectorScorer(final VectorScorer vectorScorer, final DocIdSetIterator matchedDocs, final Predicate<Float> scoreFilter)
+        throws IOException {
+        this.bulkScorer = vectorScorer.bulk(matchedDocs);
+        this.scoreFilter = scoreFilter;
+        this.cost = matchedDocs != null ? matchedDocs.cost() : vectorScorer.iterator().cost();
+    }
+
+    public static BulkVectorScorer fullPrecision(VectorScorer vectorScorer, DocIdSetIterator matchedDocs) throws IOException {
+        return new BulkVectorScorer(vectorScorer, matchedDocs, score -> true);
+    }
+
+    public static BulkVectorScorer fullPrecision(VectorScorer vectorScorer, DocIdSetIterator matchedDocs, float minScore)
+        throws IOException {
+        return new BulkVectorScorer(vectorScorer, matchedDocs, score -> score >= minScore);
+    }
+
+    @Override
+    public int docID() {
+        return currentDocId;
+    }
+
+    @Override
+    public DocIdSetIterator iterator() {
+        return new DocIdSetIterator() {
+
+            @Override
+            public int docID() {
+                return currentDocId;
+            }
+
+            @Override
+            public int nextDoc() throws IOException {
+                while (true) {
+                    int result = scanBufferForMatch();
+                    if (result != -1) {
+                        return result;
+                    }
+                    float maxBatchScore = bulkScorer.nextDocsAndScores(DocIdSetIterator.NO_MORE_DOCS, null, buffer);
+                    currentBatchIdx = 0;
+                    if (buffer.size == 0) {
+                        return currentDocId = NO_MORE_DOCS;
+                    }
+                    if (!scoreFilter.test(maxBatchScore)) {
+                        currentBatchIdx = buffer.size;
+                    }
+                }
+            }
+
+            @Override
+            public int advance(int target) throws IOException {
+                while (true) {
+                    int doc = nextDoc();
+                    if (doc == NO_MORE_DOCS) {
+                        return NO_MORE_DOCS;
+                    }
+                    if (doc >= target) {
+                        return doc;
+                    }
+                }
+            }
+
+            @Override
+            public long cost() {
+                return cost;
+            }
+        };
+    }
+
+    @Override
+    public float getMaxScore(int upTo) {
+        return Float.MAX_VALUE;
+    }
+
+    @Override
+    public float score() {
+        return currentScore;
+    }
+
+    private int scanBufferForMatch() {
+        while (currentBatchIdx < buffer.size) {
+            if (scoreFilter.test(buffer.features[currentBatchIdx])) {
+                currentDocId = buffer.docs[currentBatchIdx];
+                currentScore = buffer.features[currentBatchIdx];
+                currentBatchIdx++;
+                return currentDocId;
+            }
+            currentBatchIdx++;
+        }
+        return -1;
+    }
+}
--- a/src/main/java/org/opensearch/knn/index/query/exactsearch/BulkVectorScorer.java
+++ b/src/main/java/org/opensearch/knn/index/query/exactsearch/BulkVectorScorer.java
@@ -27,6 +27,7 @@ public class BulkVectorScorer extends Scorer {
     private int currentDocId = -1;
     private int currentBatchIdx = 0;
     private float currentScore;
+    private float minCompetitiveScore = 0f;
 
     private BulkVectorScorer(final VectorScorer vectorScorer, final DocIdSetIterator matchedDocs, final Predicate<Float> scoreFilter)
         throws IOException {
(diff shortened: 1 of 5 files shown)


Vikasht34 (commented):
The refactoring is clean and the BulkVectorScorer abstraction makes the code much more readable Thanks @shatejas , I have one concern though , our Old code https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/query/exactsearch/ExactSearcher.java#L227-L248 , we had optimization to skip block if score is less than max score 

```
if (maxScore < topDoc.score) {
    continue;  
}
```

Seems like now we are just using wrapper againt Lucene VectorScorer.Bulk interface that logic is missed . @navneet1v that skiping logic for meant for optimizating latency if I am nit wrong?

src/main/java/org/opensearch/knn/index/query/exactsearch/ExactSearcher.java
```diff
@@ -116,6 +119,34 @@ public TopDocs searchLeaf(final LeafReaderContext leafReaderContext, final Exact
...
+    public Scorer exactSearchScorer(final LeafReaderContext leafReaderContext, final ExactSearcherContext context) throws IOException {
+        final SegmentReader reader = Lucene.segmentReader(leafReaderContext.reader());
+        final FieldInfo fieldInfo = FieldInfoExtractor.getFieldInfo(reader, context.getField());
+        if (fieldInfo == null) {
+            log.debug("[KNN] FieldInfo is null for field [{}] in segment [{}]", context.getField(), reader.getSegmentName());
+            return null;
+        }
+
+        final VectorScorer vectorScorer = createVectorScorer(reader, fieldInfo, leafReaderContext, context);
+        assert vectorScorer != null;
```
navneet1v: any change vectorScorer will be null? if no then do we need this assert?

src/main/java/org/opensearch/knn/index/query/exactsearch/ExactSearcher.java
```diff
@@ -116,6 +119,34 @@ public TopDocs searchLeaf(final LeafReaderContext leafReaderContext, final Exact
...
+
+        // When nested, matchedDocsIterator is already consumed inside NestedBestChildVectorScorer,
+        // so pass null to avoid double consumption of the same iterator.
+        final DocIdSetIterator matchedDocs = context.getParentsFilter() != null ? null : context.getMatchedDocsIterator();
+
+        if (context.getRadius() != null) {
+            assert extractKNNEngine(fieldInfo) == KNNEngine.FAISS : "Exact searcher for Radial search is only used by FAISS engine";
+            final float minScore = context.isMemoryOptimizedSearchEnabled
+                ? context.getRadius()
+                : KNNEngine.FAISS.score(context.getRadius(), getSpaceType(modelDao, fieldInfo));
```
navneet1v, line 144: is this change algined with what we have been doing in the radial search bug fix for IP space type? Ref: https://github.com/opensearch-project/k-NN/pull/3369

src/main/java/org/opensearch/knn/index/query/exactsearch/ExactSearcher.java
```diff
@@ -116,6 +119,34 @@ public TopDocs searchLeaf(final LeafReaderContext leafReaderContext, final Exact
         return exactNearestNeighborSearch(context, vectorScorer, matchedDocs);
     }
 
+    public Scorer exactSearchScorer(final LeafReaderContext leafReaderContext, final ExactSearcherContext context) throws IOException {
```
navneet1v, line 122: I am not seeing use of this function, any reason for keeping it public like you want to use it in next PRs?

src/test/java/org/opensearch/knn/index/query/exactsearch/BulkVectorScorerTests.java
```diff
@@ -0,0 +1,288 @@
...
+import org.apache.lucene.util.FixedBitSet;
+import org.opensearch.knn.KNNTestCase;
+import org.opensearch.knn.index.vectorvalues.TestVectorValues;
+
+import java.io.IOException;
+import java.util.List;
+
+public class BulkVectorScorerTests extends KNNTestCase {
+
+    private static final VectorSimilarityFunction SIMILARITY = VectorSimilarityFunction.EUCLIDEAN;
```
navneet1v: can we please add it for COSINE and inner product?

src/main/java/org/opensearch/knn/index/query/exactsearch/ExactSearcher.java
shatejas: correct, this is for integration with knn weight

src/main/java/org/opensearch/knn/index/query/exactsearch/ExactSearcher.java
shatejas: searchLeaf has sort of a[ silent failure here ](https://github.com/opensearch-project/k-NN/pull/3361/changes#diff-506dbbeb9385167fbc3e3943521b79b918b294deda5764246804fbdd307a56e1R108) I can return an empty scorer to keep status quo

src/main/java/org/opensearch/knn/index/query/exactsearch/ExactSearcher.java
shatejas: Its the same as https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/query/exactsearch/ExactSearcher.java#L180 which is not changed in #3369

shatejas:
> we had optimization to skip block if score is less than max score

[This is where ](https://github.com/opensearch-project/k-NN/pull/3361/changes#diff-3a2a3a5e0a3cd3ff857b92e07602d1c7184a320aa42d62de222a63399d2db2acR73) its being handled, where it will reset the currentBatchIndes to the size, so it skips scanBufferForMatch() and bulk scores the next batch

src/test/java/org/opensearch/knn/index/query/exactsearch/BulkVectorScorerTests.java
shatejas: Will do

Vikasht34:
> > we had optimization to skip block if score is less than max score
> 
> [This is where ](https://github.com/opensearch-project/k-NN/pull/3361/changes#diff-3a2a3a5e0a3cd3ff857b92e07602d1c7184a320aa42d62de222a63399d2db2acR73) its being handled, where it will reset the currentBatchIndes to the size, so it skips scanBufferForMatch() and bulk scores the next batch

Got it ... Thanks

Vikasht34 (dismissed):
LGTM ...

src/main/java/org/opensearch/knn/index/query/exactsearch/BulkVectorScorer.java
```diff
@@ -0,0 +1,125 @@
...
+                    }
+                    float maxBatchScore = bulkScorer.nextDocsAndScores(DocIdSetIterator.NO_MORE_DOCS, null, buffer);
+                    currentBatchIdx = 0;
+                    if (buffer.size == 0) {
+                        return currentDocId = NO_MORE_DOCS;
+                    }
+                    if (!scoreFilter.test(maxBatchScore)) {
+                        currentBatchIdx = buffer.size;
+                    }
+                }
```
navneet1v, line 77: @shatejas can we add the logic of skipping the whole set of vectors if the max score < minScore we have seen till now during the iteration?

navneet1v:
> The refactoring is clean and the BulkVectorScorer abstraction makes the code much more readable Thanks @shatejas , I have one concern though , our Old code https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/query/exactsearch/ExactSearcher.java#L227-L248 , we had optimization to skip block if score is less than max score
> 
> ```
> if (maxScore < topDoc.score) {
>     continue;  
> }
> ```
> 
> Seems like now we are just using wrapper againt Lucene VectorScorer.Bulk interface that logic is missed . @navneet1v that skiping logic for meant for optimizating latency if I am nit wrong?

Some clarification all that logic is doing was it will skip the block of vectors to be passed through topK queue. @Vikasht34 good catch. I added a comment here: https://github.com/opensearch-project/k-NN/pull/3361/changes#r3483691142

src/main/java/org/opensearch/knn/index/query/exactsearch/BulkVectorScorer.java
shatejas: https://github.com/opensearch-project/k-NN/pull/3361#issuecomment-4812413522

navneet1v: @shatejas I don't see that `scoreFilter` getting updated when maxscore is change. I might be missing the code where you are updating the score. can you please point me to that?

shatejas: I misunderstood when @Vikasht34 pointed out, I thought it was for radial. Added support for minCompetitive score

Vikasht34 (dismissed):

src/main/java/org/opensearch/knn/index/query/exactsearch/ExactSearcher.java
navneet1v: I didn't get it. If this is for integration with KNNWeight I don't see that integration. Little confused here.

src/main/java/org/opensearch/knn/index/query/exactsearch/ExactSearcher.java
```diff
@@ -172,121 +203,56 @@ private TopDocs doRadialSearch(
...
+        for (int doc = iter.nextDoc(); doc != DocIdSetIterator.NO_MORE_DOCS; doc = iter.nextDoc()) {
+            float score = scorer.score();
+            if (score > topDoc.score) {
+                topDoc.score = score;
+                topDoc.doc = doc;
+                topDoc = queue.updateTop();
+                collectedCount++;
+                if (updateMinCompetitiveScore && collectedCount >= heapSize) {
+                    scorer.setMinCompetitiveScore(topDoc.score);
                 }
```
navneet1v, line 259: why not bulkVectorScorer encapsulating this score change. Won't that be better if bulkVectorScorer has this and ensure that only return docs which are better than the minCompetitive score it has seen.

shatejas: It won't work across segments, The collector is kept responsible for minCompetitive score to be able to have it updated for scorers across segments and not keep it localized, within exact searcher its being honored locally but if a top level collector wants to update the minCompetitive score it won't be able to 

The Scorable interface has setMinCompetitiveScore to be able to leverage it across segments and not just locally; its designed keeping multi-segment use case in mind

src/main/java/org/opensearch/knn/index/query/exactsearch/ExactSearcher.java
shatejas: Yeah it will be in consecutive PRs

src/main/java/org/opensearch/knn/index/query/exactsearch/ExactSearcher.java
navneet1v: @shatejas I am not sure if I got the explanation here. 

Let me check with you if my understanding is correct:

BulkScorer is at a segment level right? It should track its own minCompetitive Score which it is doing right now. So is the idea being Scorer will keep the track, but it will never take a decision whether to collect a doc or not. The collection logic based on minCompetitive score will be handled by the collector, in this case it ExactSearcher class.

If its correct then, I am good here.

src/main/java/org/opensearch/knn/index/query/exactsearch/ExactSearcher.java
navneet1v: I think its better log the error if its truely an error and return an empty scorer rather than failing on asserts.

src/main/java/org/opensearch/knn/index/query/exactsearch/ExactSearcher.java
shatejas: > So is the idea being Scorer will keep the track, but it will never take a decision whether to collect a doc or not. The collection logic based on minCompetitive score will be handled by the collector, in this case it ExactSearcher class.

Correct. The idea here is apart from ExactSearcher class, [TopScoreDocCollector](https://github.com/apache/lucene/blob/8fe701c0873de37b95fabd4a0feafa2e16f30097/lucene/core/src/java/org/apache/lucene/search/TopScoreDocCollector.java#L162) or any other collector, should be able to leverage it, the minCompetitve score for TopScoreDocCollector is updated locally as well as [gloabaly](https://github.com/apache/lucene/blob/8fe701c0873de37b95fabd4a0feafa2e16f30097/lucene/core/src/java/org/apache/lucene/search/TopScoreDocCollector.java#L89C11-L89C42) that is across segments

src/main/java/org/opensearch/knn/index/query/exactsearch/ExactSearcher.java
shatejas: I changed it to throw IllegalState to avoid silent failures, there is no case where vector scorer is null

navneet1v:
@shatejas can you please use `./gradlew spotlessApply`

navneet1v (approved):

Vikasht34 (approved):

PR #3364, opened 2026-06-15
Add expression scripting unsupported integration test

AdibaAdi:
### Description

Adds an integration test that documents and guards the current behavior for Lucene Expression scripting over `knn_vector`. The new test creates a `knn_vector` index, runs a `script_score` query with `lang: "expression"`, and asserts that the request is rejected with HTTP 400 and an error indicating the field must be numeric, date, or geopoint.

While investigating #459, I focused on the Expression scripting path. Painless scripting over `knn_vector` is already covered by `PainlessScriptScoreIT`, but Expression scripting currently rejects `knn_vector` because the field does not expose numeric doc values. This test captures that observable behavior so the rejection remains explicit if the scripting paths change later.

This PR does not implement full Expression support for `knn_vector`. It adds focused integration coverage for the current unsupported behavior and provides a concrete baseline for maintainer guidance on whether Expression support should be handled in this plugin, documented separately, or pursued in OpenSearch core.

### Issues Resolved

Refs #459

### Check List

* [x] New functionality includes testing.
* [x] Commits are signed per the DCO using `--signoff`.

### DIFF
--- a/src/test/java/org/opensearch/knn/integ/ExpressionScriptScoreIT.java
+++ b/src/test/java/org/opensearch/knn/integ/ExpressionScriptScoreIT.java
@@ -1,0 +1,85 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.knn.integ;
+
+import org.apache.hc.core5.http.io.entity.EntityUtils;
+import org.opensearch.client.Request;
+import org.opensearch.client.Response;
+import org.opensearch.client.ResponseException;
+import org.opensearch.index.query.MatchAllQueryBuilder;
+import org.opensearch.index.query.QueryBuilder;
+import org.opensearch.knn.KNNRestTestCase;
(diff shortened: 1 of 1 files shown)


kotwanikunal:
Thanks for the PR. The behavior being tested here isn't owned by k-NN. Expression scripting fails on `knn_vector` because the field doesn't expose the appropriate DocValues, which is what field resolution requires. That rejection lives in OpenSearch core's scripting layer, not in our plugin.

If core changes how Expression field resolution works (breaking change), this test breaks for reasons outside k-NN's scope. I'd suggest moving this to an OpenSearch core test (where Expression field resolution is owned).

AdibaAdi:
@kotwanikunal Thanks for the clarification. That makes sense.

I started in k-NN because issue #459 is filed here and the visible behavior involves the knn_vector field type, but I understand now that the Expression field-resolution behavior is owned by OpenSearch core rather than the k-NN plugin.
I will close this PR and move the test to the OpenSearch core repo instead.

PR #3296, opened 2026-04-28
Feature/sandbox module

rcpeken:
### Description

Adds the structural skeleton for a new `sandbox` Gradle sub-module that serves as a first-class incubation environment for experimental algorithms, optimizations, and architectural components.

**What's included:**
- **`settings.gradle`** — Registers `:sandbox` as a Gradle sub-project
- **`sandbox/build.gradle`** — Isolated build configuration (modeled after `remote-index-build-client`) with `opensearch.build` and `opensearch.java-agent` plugins
- **`ExperimentalAlgorithm.java`** — Marker annotation (`@Retention(RUNTIME)`, `@Target(TYPE)`) with `description()` and `since()` metadata fields to label experimental components
- **`ExperimentalAlgorithmTests.java`** — 6 unit tests covering annotation presence, metadata values, defaults, retention policy, and target type
- **`sandbox/README.md`** — Module documentation with lifecycle, graduation criteria, and usage guidelines
- **`package-info.java`** — Package-level Javadoc

**Build isolation:**
The root project has no dependency on the sandbox module. This means `scripts/build.sh`, `bundlePlugin`, and the standard `./gradlew build` naturally exclude it from release artifacts — no script modifications were needed.

**Design note:** I used an `@interface` (annotation) instead of a plain marker interface because it can carry metadata (`description`, `since`) and doesn't pollute the class hierarchy. Happy to switch to a simple marker interface if preferred.

### Related Issues
Resolves #3283 

### Check List
- [x] New functionality includes testing.
- [x] New functionality has been documented.
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
- [x] Commits are signed per the DCO using `--signoff`.
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/k-NN/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

---

Issue #3224, opened 2026-03-27
[RFC] Add Intel SVS (Scalable Vector Search) as a new Faiss index type


izaakk:
## Overview

This RFC proposes adding Intel [SVS (Scalable Vector Search)](https://github.com/intel/ScalableVectorSearch) as a new index type within the existing Faiss engine in the OpenSearch k-NN plugin. SVS combines the Vamana graph-based search algorithm, introduced by [Subramanya et al.](https://proceedings.neurips.cc/paper_files/paper/2019/file/09853c7fb1d3f8ee67a61b6bf4a7f8e6-Paper.pdf), with Intel's LVQ and LeanVec compression. Unlike HNSW's multi-layer hierarchy, Vamana constructs a single-layer proximity graph. When combined with vector compression (LVQ and LeanVec), SVS reduces memory footprint while improving search throughput at comparable recall.

The integration adds a new `svs_vamana` method name alongside the existing `hnsw` and `ivf` methods. No new engine is introduced. SVS indexes follow the standard Faiss Index lifecycle (add, search, I/O, factory strings), so the integration reuses the existing JNI bridge and codec infrastructure.

SVS has been [upstreamed into Faiss](https://github.com/facebookresearch/faiss/wiki/CPU-Faiss---Intel-SVS-%E2%80%90-Overview) and is already integrated in [Redis Query Engine](https://redis.io/blog/tech-dive-comprehensive-compression-leveraging-quantization-and-dimensionality-reduction/).

## Problem Statement

The Faiss engine currently offers two methods (HNSW and IVF) with `flat`, `sq` (FP16), and `pq` encoders. These encoders trade memory for accuracy at different points on the curve. As vector workloads scale in dataset size and embedding dimensionality, there is room for compression techniques that push further on this tradeoff, reducing memory while preserving or improving throughput at comparable recall.

### Functional

- **FR-1**: Support the Vamana graph algorithm as a new Faiss index type (`svs_vamana`) alongside `hnsw` and `ivf`
- **FR-2**: Support encoders: `flat`, `sq` (FP16, SQ8), `lvq` (4x0, 4x4, 4x8), and `leanvec` (4x4, 4x8, 8x8)
- **FR-3**: Support construction parameters: `degree`, `construction_window_size`, `alpha`
- **FR-4**: Support search-time parameters via `method_parameters`: `search_window_size`, `search_buffer_capacity`
- **FR-5**: Support L2, inner product, and cosinesimil distance metrics
[...]

PR #3370, opened 2026-06-17
Intel SVS as a fully JNI-isolated :sandbox engine

### LINKED ISSUE #3283: [FEATURE] Sandbox Incubator: Experimental Algorithms & Optimizations

izaakk:
## Description

Adds Intel SVS (Scalable Vector Search) Vamana as an experimental `:sandbox` tenant, isolated from the core plugin all the way down to its own JNI binary. SVS is exposed as its own engine (`engine: "svs"`) and runs in `libopensearchknn_svs*.so`, which embeds its own SVS-enabled Faiss build. The core `libopensearchknn_faiss*.so` contains zero SVS symbols.

The entire tenant (Java module, native library, packaging, and engine availability) is gated behind a single build flag, `-Pknn.sandbox.enabled=true`. Without it, the plugin builds and behaves exactly as it does today.

`svs_vamana` is a graph-based ANN method with optional LVQ or scalar quantization for memory-efficient search. This PR is the engine integration described in [RFC #3224](https://github.com/opensearch-project/k-NN/issues/3224).

## Related Issues

Resolves [#3283](https://github.com/opensearch-project/k-NN/issues/3283).

Relates to [RFC #3224](https://github.com/opensearch-project/k-NN/issues/3224). Builds on the `:sandbox` skeleton in [#3296](https://github.com/opensearch-project/k-NN/pull/3296).

## Design principles

1. One switch, the whole tenant. A single flag (`knn.sandbox.enabled`) flips the Java module, the native `.so`, packaging, and engine availability together. No second knob, no partial state, no runtime toggle.
2. Maximum isolation, down to a separate JNI binary. The tenant has its own `.so` embedding its own Faiss; the core `faiss.so` stays clean. Some native helper code is duplicated into the SVS library, a deliberate trade for a core binary that carries zero tenant code.
3. No compile-time coupling from core to the tenant. The core exposes one generic SPI behind a generic `EXPERIMENTAL` engine slot; the tenant is discovered at runtime via `ServiceLoader`. The dependency arrow is sandbox to core only (`compileOnly project(':')`).
4. Inert by default. With the flag off, the generic engine slot is dormant: it carries an inert placeholder library whose accessors are benign for iteration and throw `UnsupportedOperationException` for any real use. No SVS Java compiles, no SVS `.so` builds, and the plugin zip is identical to a build without this change.

## How the tenant plugs in

A registered, searchable engine must be discovered, resolved, searched (JNI), and loaded, all of which live in the core. This PR keeps compile-time isolation (root has no compile dependency on `:sandbox`) and makes every core change additive and tenant-agnostic:

1. `compileOnly project(':')` in `sandbox/build.gradle`, so sandbox sees core types while root does not depend on sandbox.
2. A generic engine slot, `KNNEngine.EXPERIMENTAL`, whose name, library, and native ops are supplied at runtime by a discovered provider. There is no `"svs"` literal in `KNNEngine`.
3. A generic SPI, `SandboxEngineProvider` (engine name, `KNNLibrary`, and native index lifecycle), discovered once via `ServiceLoader` by a small `SandboxEngine` holder. This is the JDK-standard `ServiceLoader` / `META-INF/services` pattern already used in this repo for `org.apache.lucene.codecs.Codec`, `org.apache.lucene.codecs.KnnVectorsFormat`, `QueryBuilderProtoConverter`, and `PainlessExtension`.
4. Engine-based routing. The engine is stored on the field (as it already is for FAISS, NMSLIB, and LUCENE) and its segment files carry the engine's own `.svs` extension, so `JNIService` routes each native op to the provider with one branch per method, and the codec reads SVS files back by extension. Nothing extra is persisted on the field to route the call.
5. A generic native loader, `KNNLibraryLoader.loadLibraryByVariant(baseName)`, the only place permitted to call `System.loadLibrary`, so the tenant loads its own AVX-variant-selected `.so`. The existing `loadFaissLibrary` and `loadSimdLibrary` delegate to it, so the variant-selection logic exists once. The base name `opensearchknn_svs` lives in the sandbox, not the core.
6. Method resolution stays tenant-side. `SvsLibrary` resolves `svs_vamana` with its own sandbox `SvsMethodResolver` (extending the public `AbstractMethodResolver`); the core `FaissMethodResolver` is untouched.

Hooks 1 through 6 are all tenant-agnostic; a future tenant reuses them unchanged. Only the sandbox-side classes are SVS-specific.

## How a sandbox SVS index loads

Loading happens in four ordered steps:

0. Build flag. Decides whether the SVS `.so` and the provider declaration are present in the artifact at all.
1. Class load. `SandboxEngine` runs `ServiceLoader.load(SandboxEngineProvider.class)` once. In a flagged build it finds the sandbox's provider, so `EXPERIMENTAL` gains the name `"svs"` and its `KNNLibrary`. Discovery does not touch native code.
2. First native use. The first `JNIService` op for an `EXPERIMENTAL` field reaches `SvsService`, whose static initializer calls `loadLibraryByVariant("opensearchknn_svs")` and loads the AVX-appropriate `.so` exactly once (an `$ORIGIN` rpath resolves the co-located `libsvs_runtime.so`).
3. Steady state. The engine value and the `.svs` file extension route create, load, query, and free from then on.

## Functional scope

| Aspect | This PR |
|---|---|
| Engine | `svs` (`KNNEngine.EXPERIMENTAL`), own `libopensearchknn_svs*.so` |
| Method | `svs_vamana` |
| Encoders | `flat` (FP32), `sq` (`type: fp16` \| `sq8`), `lvq` |
| Construction params | `degree`, `construction_window_size`, `alpha`, `search_window_size` / `search_buffer_capacity` (index-time defaults) |
| Query-time param | `search_window_size` (the recall and latency knob, analogous to `ef_search`) |
| Distance | `l2`, `innerproduct`, `cosinesimil` |
| File extension | `.svs` |
| Build flag | `-Pknn.sandbox.enabled=true` (default OFF) |
| Platform guard | `lvq` rejected with a clear error on non-Intel hosts (AVX-512 Sapphire Rapids or newer required) |

## Building and deploying

SVS is opt-in. By default `:sandbox` is excluded from the build graph entirely, so no SVS Java compiles, no SVS `.so` builds, and a normal build and deploy is unchanged.

```bash
# Default build (unchanged plugin):
./gradlew assemble

# SVS-enabled build (one flag):
./gradlew assemble -Pknn.sandbox.enabled=true \
  -Psvs.prefix=<dir with lib/cmake/svs_runtime>   # optional; otherwise the build fetches libsvs-runtime=0.3.0
```

The flagged build compiles `:sandbox`, builds a separate, unmodified upstream Faiss (SVS-capable commit, `FAISS_ENABLE_SVS=ON`, zero k-NN patches) into `libopensearchknn_svs*.so`, links `libsvs_runtime`, and bundles the SVS `.so` and the sandbox jar into the plugin zip. `-Psvs.prefix` is optional and only tells the already-enabled build where the SVS runtime lives (for offline builds); otherwise the build fetches the conda-forge `libsvs-runtime=0.3.0` package. LVQ requires an Intel AVX-512 (Sapphire Rapids or newer) host.

## Impact on the existing (HNSW / IVF / Lucene) path: none

* The default build excludes `:sandbox` from the build graph: no SVS Java compiles, no SVS `.so` builds, and `unzip -l` of the plugin zip shows no sandbox or SVS entries.
* The core `libopensearchknn_faiss*.so` contains zero SVS symbols (`nm` verified) and does not link `libsvs_runtime`.
* The one new enum value (`EXPERIMENTAL`) is inert without a bundled tenant: it carries an inert placeholder library whose iteration-safe accessors return benign values and whose use-implying methods throw `UnsupportedOperationException`, so even code that iterates `KNNEngine.values()` (tests, diagnostics) has no NPE path. Name and path resolution are additionally guarded by `SandboxEngine.isPresent()`; with the flag off, `engine: "svs"` is rejected with `Invalid engine type: svs`. The full core unit-test suite passes on the default build.
* Every core edit is additive (a generic engine slot, a generic SPI, one branch per `JNIService` op, one line in the build-strategy factory). No existing HNSW, IVF, or Lucene code path is modified.
* Two Faiss builds (the core submodule pin and the SVS library's SVS-capable commit) coexist safely in one JVM via hidden symbol visibility plus a JNI-only linker version script; a coexistence integration test exercises an HNSW index and an SVS index in the same node.

## Testing

The SVS tests are opt-in, mirroring the feature: they run only with `-Pknn.sandbox.enabled=true`. So default CI proves the change is inert and causes no regression; SVS correctness is covered by the flagged tests below, which a reviewer can run by building with the flag (LVQ and the native runtime require an Intel AVX-512 Sapphire Rapids host).

Default build (no flag):
* The full core unit-test suite passes (`./gradlew test`).
* `:sandbox` is absent from `./gradlew projects`; `nm` of `faiss.so` is SVS-symbol-free; no SVS artifacts are produced; `engine: "svs"` is rejected with a 400.

Flagged build (`-Pknn.sandbox.enabled=true`):
* `libopensearchknn_svs*.so` builds against a pristine upstream Faiss, links `libsvs_runtime`, exports only JNI symbols (linker version script, so no Faiss-symbol interposition), and carries the SVS and Vamana symbols while `faiss.so` stays SVS-symbol-free.
* `:sandbox` Java unit tests cover the method/encoder resolution and `compression_level` mapping.
* The REST `FaissSVSVamanaIT` exercises the native path end-to-end (index, merge, query, filter, and the validation rejections) and passes 7/7. It is excluded from the default-build `integTest` run.

Native C++ unit tests (`jni_test`) for the SVS wrapper are a tracked follow-up: the SVS test target links a separate runtime that requires Intel AVX-512, so it needs its own gated harness and runner rather than the standard `jni_test` flow. The native path is exercised end-to-end today through `FaissSVSVamanaIT`.

Beyond the above, the change was manually validated on an Intel AVX-512 host across the encoders, distances, `compression_level` mappings, build/query parameters, filtering, and faiss-HNSW coexistence.

## Relationship to the `:sandbox` skeleton ([#3296](https://github.com/opensearch-project/k-NN/pull/3296))

This branch includes the `:sandbox` module skeleton so it builds and CI runs standalone; once #3296 merges, that skeleton commit will be rebased or dropped and this PR retargeted. The generic seam introduced here (the `SandboxEngineProvider` SPI, the `EXPERIMENTAL` engine slot, and `loadLibraryByVariant`) is reusable sandbox infrastructure that may belong in #3296 itself; happy to relocate it if maintainers prefer.

## Follow-up

LeanVec, a learned dimensionality-reduction encoder that further cuts memory for high-dimensional vectors, is a planned follow-up. It needs a per-shard training step that this PR does not yet wire up; the encoder set here (`flat`, `sq`, `lvq`) is the in-memory, no-training set.

## Check List
* [x] New functionality includes testing.
* [ ] New functionality has been documented.
* [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
* [x] Commits are signed per the DCO using `--signoff`.
* [x] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).

### DIFF
--- a/sandbox/src/main/java/org/opensearch/knn/sandbox/svs/SvsMethodResolver.java
+++ b/sandbox/src/main/java/org/opensearch/knn/sandbox/svs/SvsMethodResolver.java
@@ -1,0 +1,221 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.knn.sandbox.svs;
+
+import org.opensearch.common.ValidationException;
+import org.opensearch.knn.index.SpaceType;
+import org.opensearch.knn.index.engine.AbstractMethodResolver;
+import org.opensearch.knn.index.engine.Encoder;
+import org.opensearch.knn.index.engine.KNNEngine;
+import org.opensearch.knn.index.engine.KNNMethodConfigContext;
+import org.opensearch.knn.index.engine.KNNMethodContext;
+import org.opensearch.knn.index.engine.MethodComponent;
+import org.opensearch.knn.index.engine.MethodComponentContext;
+import org.opensearch.knn.index.engine.ResolvedMethodContext;
+import org.opensearch.knn.index.engine.TrainingConfigValidationInput;
+import org.opensearch.knn.index.engine.TrainingConfigValidationOutput;
+import org.opensearch.knn.index.mapper.CompressionLevel;
+import org.opensearch.knn.index.mapper.Mode;
+import org.opensearch.knn.sandbox.ExperimentalAlgorithm;
+
+import java.util.HashMap;
+import java.util.Locale;
+import java.util.Map;
+import java.util.Set;
+
+import static org.opensearch.knn.common.KNNConstants.METHOD_ENCODER_PARAMETER;
+import static org.opensearch.knn.sandbox.svs.SVSConstants.FAISS_SVS_ENCODER_LVQ;
+import static org.opensearch.knn.sandbox.svs.SVSConstants.FAISS_SVS_SQ_TYPE;
+import static org.opensearch.knn.sandbox.svs.SVSConstants.FAISS_SVS_SQ_TYPE_FP16;
+import static org.opensearch.knn.sandbox.svs.SVSConstants.METHOD_PARAMETER_LVQ_PRIMARY_BITS;
+import static org.opensearch.knn.sandbox.svs.SVSConstants.METHOD_PARAMETER_LVQ_RESIDUAL_BITS;
+import static org.opensearch.knn.sandbox.svs.SVSConstants.METHOD_SVS_VAMANA;
+import static org.opensearch.knn.common.KNNConstants.ENCODER_SQ;
+
+/**
+ * {@link org.opensearch.knn.index.engine.MethodResolver} for the experimental SVS engine. The sole method is
+ * {@code svs_vamana}; this resolver fills in its defaults, maps a user-supplied {@code compression_level} to
+ * the SVS encoders ({@code 2x} -> {@code sq}(fp16), {@code 4x} -> {@code lvq}(4,4), {@code 8x} ->
+ * {@code lvq}(4,0)) and rejects configurations SVS cannot serve ({@code on_disk} mode, training contexts,
+ * unsupported compression levels). It lives entirely in the sandbox: the core Faiss resolver knows nothing
+ * about SVS.
+ */
+@ExperimentalAlgorithm(description = "Intel SVS Vamana method resolver", since = "3.7.0")
+public class SvsMethodResolver extends AbstractMethodResolver {
+
+    // x4 maps to lvq(4,4); the core engines top out differently (no x4) and support higher levels SVS does not.
+    private static final Set<CompressionLevel> SUPPORTED_COMPRESSION_LEVELS = Set.of(
+        CompressionLevel.x1,
+        CompressionLevel.x2,
+        CompressionLevel.x4,
+        CompressionLevel.x8
+    );
+
+    @Override
+    public ResolvedMethodContext resolveMethod(
+        KNNMethodContext knnMethodContext,
+        KNNMethodConfigContext knnMethodConfigContext,
+        boolean shouldRequireTraining,
+        final SpaceType spaceType
+    ) {
+        validateConfig(knnMethodConfigContext, shouldRequireTraining);
+
+        KNNMethodContext resolvedKNNMethodContext = initResolvedKNNMethodContext(
+            knnMethodContext,
+            KNNEngine.EXPERIMENTAL,
+            spaceType,
+            METHOD_SVS_VAMANA
+        );
+
+        String methodName = resolvedKNNMethodContext.getMethodComponentContext().getName();
+        if (METHOD_SVS_VAMANA.equals(methodName) == false) {
+            ValidationException validationException = new ValidationException();
+            validationException.addValidationError(
+                String.format(
+                    Locale.ROOT,
+                    "Invalid method name [%s] for engine [%s]. The only supported method is [%s].",
+                    methodName,
+                    KNNEngine.EXPERIMENTAL.getName(),
+                    METHOD_SVS_VAMANA
+                )
+            );
+            throw validationException;
+        }
+
+        Map<String, Encoder> encoderMap = FaissSVSVamanaMethod.SUPPORTED_ENCODERS;
+
+        // Fill in parameters for the encoder and then the method.
+        resolveEncoder(resolvedKNNMethodContext, knnMethodConfigContext, encoderMap);
+        CompressionLevel resolvedCompressionLevel = resolveCompressionLevelFromMethodContext(
+            resolvedKNNMethodContext,
+            knnMethodConfigContext,
+            encoderMap
+        );
+
+        validateEncoderConfig(resolvedKNNMethodContext, knnMethodConfigContext, encoderMap);
+        validateCompressionConflicts(knnMethodConfigContext.getCompressionLevel(), resolvedCompressionLevel);
+        knnMethodConfigContext.setCompressionLevel(resolvedCompressionLevel);
+        resolveMethodParams(
+            resolvedKNNMethodContext.getMethodComponentContext(),
+            knnMethodConfigContext,
+            FaissSVSVamanaMethod.METHOD_COMPONENT
+        );
+
+        return ResolvedMethodContext.builder()
+            .knnMethodContext(resolvedKNNMethodContext)
+            .compressionLevel(resolvedCompressionLevel)
+            .build();
+    }
+
+    private void validateConfig(KNNMethodConfigContext knnMethodConfigContext, boolean shouldRequireTraining) {
+        ValidationException validationException = validateNotTrainingContext(shouldRequireTraining, KNNEngine.EXPERIMENTAL, null);
+        validationException = validateCompressionSupported(
+            knnMethodConfigContext.getCompressionLevel(),
+            SUPPORTED_COMPRESSION_LEVELS,
+            KNNEngine.EXPERIMENTAL,
+            validationException
+        );
+        if (knnMethodConfigContext.getMode() == Mode.ON_DISK) {
+            validationException = validationException == null ? new ValidationException() : validationException;
+            validationException.addValidationError(
+                String.format(
+                    Locale.ROOT,
+                    "mode=on_disk is not supported with %s; SVS is an in-memory execution path. "
+                        + "Use mode=in_memory or a different method.",
+                    METHOD_SVS_VAMANA
+                )
+            );
+        }
+        if (validationException != null) {
+            throw validationException;
+        }
+    }
+
+    /**
+     * Resolves the SVS encoder for a user-supplied {@code compression_level} when no explicit encoder was
+     * given. Other levels require an explicit {@code encoder} block.
+     */
+    private void resolveEncoder(
+        KNNMethodContext resolvedKNNMethodContext,
+        KNNMethodConfigContext knnMethodConfigContext,
+        Map<String, Encoder> encoderMap
+    ) {
+        if (shouldEncoderBeResolved(resolvedKNNMethodContext, knnMethodConfigContext) == false) {
+            return;
+        }
+
+        CompressionLevel resolvedCompressionLevel = CompressionLevel.isConfigured(knnMethodConfigContext.getCompressionLevel())
+            ? knnMethodConfigContext.getCompressionLevel()
+            : CompressionLevel.x1;
+        if (resolvedCompressionLevel == CompressionLevel.x1) {
+            return;
+        }
+
+        MethodComponentContext encoderComponentContext;
+        Encoder encoder;
+        if (CompressionLevel.x2 == resolvedCompressionLevel) {
+            encoderComponentContext = new MethodComponentContext(ENCODER_SQ, new HashMap<>());
+            encoderComponentContext.getParameters().put(FAISS_SVS_SQ_TYPE, FAISS_SVS_SQ_TYPE_FP16);
+            encoder = encoderMap.get(ENCODER_SQ);
+        } else if (CompressionLevel.x4 == resolvedCompressionLevel) {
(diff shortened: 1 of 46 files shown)


sandbox/src/main/java/org/opensearch/knn/sandbox/svs/SvsMethodResolver.java
```diff
@@ -0,0 +1,221 @@
...
+
+/**
+ * {@link org.opensearch.knn.index.engine.MethodResolver} for the experimental SVS engine. The sole method is
+ * {@code svs_vamana}; this resolver fills in its defaults, maps a user-supplied {@code compression_level} to
+ * the SVS encoders ({@code 2x} -> {@code sq}(fp16), {@code 4x} -> {@code lvq}(4,4), {@code 8x} ->
+ * {@code lvq}(4,0)) and rejects configurations SVS cannot serve ({@code on_disk} mode, training contexts,
+ * unsupported compression levels). It lives entirely in the sandbox: the core Faiss resolver knows nothing
+ * about SVS.
+ */
+@ExperimentalAlgorithm(description = "Intel SVS Vamana method resolver", since = "3.7.0")
```
navneet1v: lets make it 3.8.0

sandbox/src/test/java/org/opensearch/knn/sandbox/svs/FaissSVSVamanaIT.java
```diff
@@ -0,0 +1,225 @@
...
+import static org.opensearch.knn.common.KNNConstants.NAME;
+import static org.opensearch.knn.common.KNNConstants.PARAMETERS;
+
+/**
+ * REST integration tests for the experimental {@code svs} engine ({@code svs_vamana} method), requiring a
+ * node built with {@code -Pknn.sandbox.enabled=true} (sandbox jar bundled + the isolated SVS native lib);
+ * excluded from default-build integTest runs. SVS names are written as literals to avoid a test-time
+ * dependency on the sandbox module.
+ */
+public class FaissSVSVamanaIT extends KNNRestTestCase {
```
navneet1v, line 35: we should move this also in sandbox module

src/main/java/org/opensearch/knn/index/engine/SandboxEngine.java
```diff
@@ -0,0 +1,157 @@
...
+/**
+ * Discovers the optional {@link SandboxEngineProvider} (contributed by an opt-in {@code :sandbox} module)
+ * once at class load, and exposes what {@link KNNEngine#EXPERIMENTAL} and {@code JNIService} need from it.
+ * When no provider is bundled (the default build), {@link #isPresent()} is {@code false}, {@code EXPERIMENTAL}
+ * is inert ({@link #library()} returns an {@link InertSandboxLibrary} that never resolves anything), and the
+ * plugin behaves exactly as upstream.
+ */
+public final class SandboxEngine {
+
+    private static final Logger logger = LogManager.getLogger(SandboxEngine.class);
```
navneet1v: lets use log4j2 annotation instead of this.

src/main/java/org/opensearch/knn/index/engine/SandboxEngine.java
```diff
@@ -0,0 +1,157 @@
...
+
+    private static final SandboxEngineProvider PROVIDER = load();
+    private static final KNNLibrary INERT_LIBRARY = new InertSandboxLibrary();
+
+    private SandboxEngine() {}
```
navneet1v: use lombok annotaton for this

src/main/java/org/opensearch/knn/index/query/request/MethodParameter.java
```diff
@@ -69,6 +71,25 @@ public ValidationException validate(Object value) {
             validationException.addValidationError(METHOD_PARAMETER_NPROBES + " should be greater than 0");
             return validationException;
         }
+    },
+
+    SEARCH_WINDOW_SIZE(METHOD_PARAMETER_SEARCH_WINDOW_SIZE, Version.V_3_7_0, SEARCH_WINDOW_SIZE_FIELD) {
```
navneet1v, line 76: we should build an extension point for this in sandbox module.

navneet1v (commented):
@izaakk Thanks for raising the PR.

Some high level thoughts,
1. From this PR we should remove all the interactions with main KNN module for now. So that we can decouple the SVS integration from Sandbox module with Experimental engine implementations. This will give us faster PR velocity.
2. Lets start to put thoughts for extension points in different layers like Query, JNIService, KNNEngine current implementation is little bit coupled in my opinion.
3. Tests be it IT or UTs should be part of Sandbox module.

izaakk:
Thanks @navneet1v. Before I push a large refactor I'd like to align on what reads as coupled to you at each layer.

The overall aim is for core to carry no knowledge of the sandbox tenant, with the tenant-specific logic living entirely in `:sandbox`. Everything plugs into one core SPI, `SandboxEngineProvider` (engine name + `KNNLibrary` + native ops), discovered at runtime via `ServiceLoader`, and the three layers below build on it. (Some minimal glue stays in core since `KNNEngine` is an enum and `getEngine` lives there.) Does that shape, and the `ServiceLoader` discovery, match what you had in mind, or would you structure it differently?

**1. KNNEngine.** The engine gets the generic `EXPERIMENTAL` slot and is added to the capability sets it
participates in (filters, custom-segment-files).
```java
// engine/KNNEngine.java
EXPERIMENTAL(SandboxEngine.engineName(), SandboxEngine.library()),
private static Set<KNNEngine> withExperimental(KNNEngine... base) {
    ... if (SandboxEngine.isPresent()) builder.add(EXPERIMENTAL); ...   // core adds it to filters + custom-files sets
}
```
One thing I'd change is that core decides the engine's traits; the engine could declare its own capabilities
instead. But what reads as coupling to you here, this, the slot itself, or something else?

**2. JNIService.** Each native op routes to the engine that owns it, alongside the FAISS/NMSLIB branches, again
keyed on the generic `EXPERIMENTAL` slot.
```java
// jni/JNIService.java  (one of 8 such branches)
if (KNNEngine.EXPERIMENTAL == knnEngine) {
    return SandboxEngine.provider().initIndex(...);
}
```
This branch repeats in all 8 native ops; a generic dispatch would route to the owning engine without a
per-op path. Same question: is that what feels coupled to you, or something else?

**3. Query.** `search_window_size` was added to the core `MethodParameter` enum so it would flow through.
```java
// query/request/MethodParameter.java
SEARCH_WINDOW_SIZE(METHOD_PARAMETER_SEARCH_WINDOW_SIZE, ...) { ... }
```
This one's clear to me: an SVS-specific param sitting in core, when the engine already declares it in `:sandbox`
via `KNNLibrarySearchContext`. It's only in core because the REST parser rejects any param name not in the enum,
before the engine is ever consulted. Following your suggestion to make this an extension point, we'd drop it
from the enum and instead have the parser pass unrecognized names through to the engine to validate.

But it's not as simple as deleting the enum entry. The coordinator serializes `method_parameters` positionally
against the enum, so a param that isn't in the enum is silently dropped on the hop to the data node, a
wrong-results bug I confirmed on a multi-node cluster. So to propagate the param to the data nodes at all, the
wire format has to change too; one option could be a generic map (version-gated so older nodes still parse it).
I've put that whole candidate up on my fork so you can see what it touches, without pushing to
this PR: **[a separate demo PR](https://github.com/izaakk/k-NN/pull/9)**. Would love your thoughts there, or any other approach you'd suggest.

I've also pushed the smaller agreed items as a separate commit: moving the SVS tests into `:sandbox` and the
`since` 3.8.0 bump.

PR #3381, opened 2026-06-23
Fix FAISS SQ merge failure when segment has no live vectors

### LINKED ISSUE #3379: [BUG] FAISS SQ merge can fail when empty scalar-quantized vector values lack quantizedVectorValues
**What is the bug?**
For native FAISS scalar-quantized vector fields, segment merge can fail when Lucene returns an empty vector-values implementation for a field.

The affected path reopens merged flat vector files and calls `KNN1040ScalarQuantizedUtils.extractQuantizedByteVectorValues(...)`. That utility reflects a private Lucene field named `quantizedVectorValues` from the concrete runtime class. When a merge input segment has no live vectors for the field, Lucene may return an empty/off-heap wrapper that does not declare that field. This can throw `NoSuchFieldException`, fail `QuantizedByteVectorValues` extraction, and abort the merge.

Observed failure signature:
```text
merge_exception: this writer hit an unrecoverable error; cannot merge
Caused by: IOException: Failed to extract QuantizedByteVectorValues from floatVectorValues [EmptyOffHeapVectorValues/...]
Caused by: NoSuchFieldException: quantizedVectorValues
```

A minimal reproduction should create a FAISS scalar-quantized knn_vector field, produce multiple segments, ensure at least one merge input segment has zero live values for that field, then trigger a merge.

Example shape:
1. Create an index with a FAISS HNSW knn_vector field using an SQ encoder.
2. Index documents across multiple segments, including documents that omit the vector field or later delete all documents containing that field in one segment.
3. Trigger merge or force-merge.
4. The merge path may receive empty FloatVectorValues for the SQ field and fail while extracting quantizedVectorValues.

If a merge input has no live vectors for a scalar-quantized field, k-NN should treat it as empty and skip quantized extraction/native build work for that field instead of failing the merge.

For non-empty values, quantized extraction should continue to work. If reflection is still needed, lookup should be robust to Lucene wrappers whose private field is declared in a superclass.

- OpenSearch: 3.6.x
- k-NN plugin: 3.6.x
- Engine: FAISS
- Vector field type: knn_vector
- Method: HNSW
- Encoder: SQ / scalar quantization

No screenshots.

Relevant code paths:

- `Faiss1040ScalarQuantizedKnnVectorsWriter.flush`
- `Faiss1040ScalarQuantizedKnnVectorsWriter.mergeOneField`
- `KNN1040ScalarQuantizedUtils.extractQuantizedByteVectorValues`

A defensive fix would:
- short-circuit when FloatVectorValues is null or size() == 0,
- search the class hierarchy for quantizedVectorValues instead of only the concrete runtime class,
- add test coverage for empty values and superclass field lookup.

Example sketch: 
``` java

    log.debug("No scalar-quantized vectors found for field {}", fieldInfo.getName());
    return;
}

final QuantizedByteVectorValues quantizedValues =
    KNN1040ScalarQuantizedUtils.extractQuantizedByteVectorValues(floatVectorValues);
```

``` java
private static Field findFieldInClassHierarchy(final Class<?> candidateClass, final String fieldName) {
    Class<?> currentClass = candidateClass;
    while (currentClass != null) {
        try {
            return currentClass.getDeclaredField(fieldName);
        } catch (NoSuchFieldException ignored) {
            currentClass = currentClass.getSuperclass();
        }
    }
    return null;
}
```

Gaurav12058:
### Description
When merging segments with scalar-quantized vector fields, Lucene may return an empty wrapper (e.g. `EmptyOffHeapVectorValues`) for segments where all documents with the vector field were deleted. The reflection-based extraction of `quantizedVectorValues` failed on these wrappers because they lack that field, aborting the merge with `NoSuchFieldException`.

This fix:
- Guards `flush()` and `mergeOneField()` to skip native build when `FloatVectorValues` is null or has size 0
- Walks the class hierarchy when looking up the private field via reflection, so subclass wrappers are handled correctly

### Related Issues
Resolves #3379

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss1040ScalarQuantizedKnnVectorsWriter.java
+++ b/src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss1040ScalarQuantizedKnnVectorsWriter.java
@@ -14,6 +14,7 @@ import org.apache.lucene.codecs.hnsw.FlatVectorsWriter;
 import org.apache.lucene.codecs.lucene104.QuantizedByteVectorValues;
 import org.apache.lucene.index.FieldInfo;
 import org.apache.lucene.index.FieldInfos;
+import org.apache.lucene.index.FloatVectorValues;
 import org.apache.lucene.index.MergeState;
 import org.apache.lucene.index.SegmentReadState;
 import org.apache.lucene.index.SegmentWriteState;
@@ -145,7 +146,7 @@ class Faiss1040ScalarQuantizedKnnVectorsWriter extends AbstractNativeEnginesKnnV
         // and pass it to the build strategy. The writer owns the reader lifecycle.
         final FlatVectorsReader flatVectorsReader = openFlatVectorsReader();
         try {
-            final var floatVectorValues = flatVectorsReader.getFloatVectorValues(fieldInfo.getName());
+            final FloatVectorValues floatVectorValues = flatVectorsReader.getFloatVectorValues(fieldInfo.getName());
             if (floatVectorValues == null || floatVectorValues.size() == 0) {
                 log.debug("No scalar-quantized vectors found for field [{}], skipping native build", fieldInfo.getName());
                 return;
--- a/src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss1040ScalarQuantizedKnnVectorsWriter.java
+++ b/src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss1040ScalarQuantizedKnnVectorsWriter.java
@@ -110,8 +110,13 @@ class Faiss1040ScalarQuantizedKnnVectorsWriter extends AbstractNativeEnginesKnnV
         // and pass it to the build strategy. The writer owns the reader lifecycle.
         final FlatVectorsReader flatVectorsReader = openFlatVectorsReader();
         try {
+            final var floatVectorValues = flatVectorsReader.getFloatVectorValues(fieldInfo.getName());
+            if (floatVectorValues == null || floatVectorValues.size() == 0) {
+                log.debug("No scalar-quantized vectors found for field [{}], skipping native build", fieldInfo.getName());
+                return;
+            }
             final QuantizedByteVectorValues quantizedValues = KNN1040ScalarQuantizedUtils.extractQuantizedByteVectorValues(
-                flatVectorsReader.getFloatVectorValues(fieldInfo.getName())
+                floatVectorValues
             );
             doFlush(
                 fieldInfo,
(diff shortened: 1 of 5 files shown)


src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss1040ScalarQuantizedKnnVectorsWriter.java
```diff
@@ -110,8 +110,13 @@ public void flush(int maxDoc, Sorter.DocMap sortMap) throws IOException {
         // and pass it to the build strategy. The writer owns the reader lifecycle.
         final FlatVectorsReader flatVectorsReader = openFlatVectorsReader();
         try {
+            final var floatVectorValues = flatVectorsReader.getFloatVectorValues(fieldInfo.getName());
+            if (floatVectorValues == null || floatVectorValues.size() == 0) {
+                log.debug("No scalar-quantized vectors found for field [{}], skipping native build", fieldInfo.getName());
+                return;
+            }
```
shatejas: why is this needed for flush? It will always have atleast 1 vector via addField() which is only called if there is atleast 1 document with that field, lets remove this from flush?

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/KNN1040ScalarQuantizedUtils.java
```diff
@@ -47,7 +47,10 @@ class KNN1040ScalarQuantizedUtils {
 
     public static QuantizedByteVectorValues extractQuantizedByteVectorValues(final KnnVectorValues floatVectorValues) throws IOException {
         try {
-            final Field f = floatVectorValues.getClass().getDeclaredField(QUANTIZED_VECTOR_VALUES_FIELD_NAME);
+            final Field f = findFieldInClassHierarchy(floatVectorValues.getClass(), QUANTIZED_VECTOR_VALUES_FIELD_NAME);
+            if (f == null) {
+                throw new NoSuchFieldException(QUANTIZED_VECTOR_VALUES_FIELD_NAME);
+            }
```
shatejas: Why is this change needed, this should be exoecting the exact type, there is no need to traverse the class hierarchy right?

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss1040ScalarQuantizedKnnVectorsWriter.java
Gaurav12058: I originally added it because the issue suggested guarding both paths, but flush can't hit this. Removed.

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/KNN1040ScalarQuantizedUtils.java
Gaurav12058: Agreed, the upstream empty-values guard in mergeOneField() ensures only ScalarQuantizedVectorValues reaches extraction, which has the field directly. Reverted.

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss1040ScalarQuantizedKnnVectorsWriter.java
```diff
@@ -145,8 +145,13 @@ public void mergeOneField(FieldInfo fieldInfo, MergeState mergeState) throws IOE
         // and pass it to the build strategy. The writer owns the reader lifecycle.
         final FlatVectorsReader flatVectorsReader = openFlatVectorsReader();
         try {
+            final var floatVectorValues = flatVectorsReader.getFloatVectorValues(fieldInfo.getName());
```
navneet1v: better to not use var. lets use proper types here.

navneet1v (approved):
Looks good. just 1 small comment

shatejas (approved):

PR #3382, opened 2026-06-23
Ignore flaky test testFilteredSearch_whenVisitBudgetExhausted_thenFallbackToExactSearch

MrFlap:
### Description
Flaky test is failing when the index is built on remote build. Ignore for now to unblock PRs. Also adding github changes from https://github.com/opensearch-project/k-NN/pull/3295 because they are blocking eachother.

### Related Issues
https://github.com/opensearch-project/k-NN/issues/3383

### Check List

### DIFF
--- a/.github/workflows/backwards_compatibility_tests_workflow.yml
+++ b/.github/workflows/backwards_compatibility_tests_workflow.yml
@@ -63,9 +63,11 @@ jobs:
           git config --global user.email "github-actions[bot]@users.noreply.github.com"
 
       - name: Setup Java ${{ matrix.java }}
-        uses: actions/setup-java@b6e674f4b717d7b0ae3baee0fbe79f498905dfde # v1
+        uses: actions/setup-java@c1e323688fd81a25caa38c78aa6df2d33d3e20d9 # v4
         with:
           java-version: ${{ matrix.java }}
+          distribution: temurin
+          cache: gradle
 
       - if: startsWith(matrix.os,'ubuntu')
         name: Install dependencies on ubuntu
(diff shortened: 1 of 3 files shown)


naveentatikonda:
please create an issue to track and fix it

src/test/java/org/opensearch/knn/memoryoptsearch/MOSFaissFloatIndexIT.java
```diff
@@ -73,6 +74,7 @@ public void testWhenNoIndexBuilt() {
      * contains the exhausted budget message.
      */
     @SneakyThrows
+    @Ignore
```
kotwanikunal: Can we use AwaitsFix with the issue url instead? We will lose track of flaky tests otherwise.

MrFlap:
Added the issue and changed to AwaitsFix @naveentatikonda @kotwanikunal

naveentatikonda (approved):

shatejas (approved):

src/test/java/org/opensearch/knn/memoryoptsearch/MOSFaissFloatIndexIT.java
```diff
@@ -73,6 +73,7 @@ public void testWhenNoIndexBuilt() {
      * contains the exhausted budget message.
      */
     @SneakyThrows
+    @AwaitsFix(bugUrl = "https://github.com/opensearch-project/k-NN/issues/3383")
```
navneet1v: rather than doing this, can we just not ignore this test from the RemoteIndexBuild integ test task? It will atleast ensure that code is tested on CPU based builds and we don't cause any regression

@shatejas , @naveentatikonda , @MrFlap WDYT?

navneet1v (changes requested):

src/test/java/org/opensearch/knn/memoryoptsearch/MOSFaissFloatIndexIT.java
MrFlap: @navneet1v , I spoke with @rchitale7 about this and we decided on ignore because we shouldn't have a test that breaks on Remote Index Build and not others. I'm fine with disabling it for remote index build only so that we can still have this test, though. Definitely will still need a fix.

MrFlap: Also I'm not sure if the pattern to skip remote build exists other than just doing something like `assumeFalse("This test is not compatible with remote index build", System.getProperty("test.remoteBuild") != null);`

kotwanikunal: You can follow https://github.com/opensearch-project/k-NN/blob/main/build.gradle#L669-L679

navneet1v (approved):

PR #3341, opened 2026-05-26
Accelerating FP16 Bulk Similarity (Follow-up to #3172)

mulugetam:
This PR is a follow-up to #3172. While #3172 demonstrated significant performance gains (based on the Google Benchmark harness), precomputing the tail mask for the tail portion of both IP and L2 does not provide benefits and can sometimes introduce minor overhead. This PR addresses that by computing the mask inline.

Below are the performance gains measured on `r8i.4xlarge`, giving a speedup of up to `1.55x`. `dim` (dimension), `nV` (number of vectors). 

```
MaxIP (inner product)
+--------+-------+-------+-------+-------+-------+-------+-------+-------+-------+
| dim\nV |     1 |     4 |     8 |    16 |    24 |    32 |    64 |   128 |   256 |
+========+=======+=======+=======+=======+=======+=======+=======+=======+=======+
|     64 | 1.00x | 1.06x | 1.33x | 1.31x | 1.35x | 1.34x | 1.31x | 1.30x | 1.31x |
|    384 | 1.00x | 1.00x | 1.48x | 1.46x | 1.49x | 1.49x | 1.50x | 1.49x | 1.50x |
|    512 | 1.04x | 1.00x | 1.51x | 1.50x | 1.53x | 1.51x | 1.51x | 1.52x | 1.46x |
|    768 | 1.00x | 1.01x | 1.47x | 1.44x | 1.46x | 1.44x | 1.46x | 1.45x | 1.52x |
|   1024 | 1.00x | 1.00x | 1.54x | 1.52x | 1.55x | 1.54x | 1.54x | 1.53x | 1.43x |
|   1536 | 0.99x | 0.99x | 1.48x | 1.46x | 1.43x | 1.44x | 1.44x | 1.44x | 1.55x |
+--------+-------+-------+-------+-------+-------+-------+-------+-------+-------+

L2 (squared distance)
+--------+-------+-------+-------+-------+-------+-------+-------+-------+-------+
| dim\nV |     1 |     4 |     8 |    16 |    24 |    32 |    64 |   128 |   256 |
+========+=======+=======+=======+=======+=======+=======+=======+=======+=======+
|     64 | 1.17x | 1.12x | 1.32x | 1.30x | 1.32x | 1.32x | 1.31x | 1.32x | 1.32x |
|    384 | 1.05x | 1.00x | 1.46x | 1.44x | 1.46x | 1.45x | 1.46x | 1.46x | 1.46x |
[...]

PR #3386, opened 2026-06-24
Add native SIMD cosine scoring for FP16 and SQ 

mulugetam:
Currently, cosine similarity in the memory-optimized search (MOS) path was handled differently depending on the storage format. For SQ, the SIMD bulk scorer computed a MaxIP score and `MemoryOptimizedKNNWeight` then post-converted it to a cosine score via `MemoryOptimizedSearchScoreConverter.convertToCosineScore()`. For FP16, cosine fell through to the Lucene delegate entirely, missing the SIMD fast path. In neither case did the scorer itself emit a cosine score.

This patch collapses cosine scoring into the SIMD kernel by introducing dedicated `FP16_COSINE` and `SQ_COSINE` similarity function types. Each reuses the underlying inner-product computation and applies a `(1 + dot) / 2` transform in-kernel, so callers receive correctly-ranged `[0, 1]` scores directly with no post-processing.

On AVX-512-FP16-capable hardware, `FP16_COSINE` additionally binds to a new native kernel, `AVX512NativeFP16IP`, which uses `_mm512_fmadd_ph` at 32 FP16 elements per 512-bit register. The FP16 accumulator is mathematically safe here specifically because cosine guarantees L2-normalized vectors and therefore bounded dot products, so this kernel is opt-in via the COSINE route only.

**C++ / JNI layer:**

- Add `FP16_COSINE` and `SQ_COSINE` to `NativeSimilarityFunctionType`.
- Add `cosineTransform` / `cosineTransformBulk` to `FaissScoreToLuceneScoreTransform`, with output clamped to `[0, 1]` to absorb SQ quantization noise that can push the raw dot product slightly outside `[-1, 1]`.
- Replace the `bool IsMaxIP` template parameter on the SQ similarity structs with `enum class SQMetricMode {MAX_IP, L2, COSINE}`. `SQ_COSINE` shares the quantized inner-product math with `SQ_IP`; only the transform differs.
- Add `AVX512NativeFP16IP`, gated on `__AVX512FP16__`, used only for `FP16_COSINE`.
- Wire `FP16_COSINE` and `SQ_COSINE` through the AVX-512, ARM NEON, and scalar fallback similarity-function tables, and through `saveSearchContext` in `simd_similarity_function_common.cpp`.

**Java layer:**

- Add `FP16_COSINE` and `SQ_COSINE` to `SimdVectorComputeService.SimilarityFunctionType`.
- Route `VectorSimilarityFunction.COSINE -> SQ_COSINE` in `KNN1040ScalarQuantizedVectorScorer` and `-> FP16_COSINE` in `NativeEngines990KnnVectorsScorer.getNativeFunctionType()`, so `cosinesimil` fields reach the new kernels on both the SQ and FP16 paths.
- `FaissMemoryOptimizedSearcher` resolves the similarity function from the user-declared space type in `FieldInfo` (via the new `resolveKnnVectorSimilarityFunction`) instead of the Faiss header metric. Faiss has no cosine metric — `cosinesimil` indices are written as inner product on L2-normalized vectors, so `faissIndex.getVectorSimilarityFunction()` reports `MAXIMUM_INNER_PRODUCT` for them. Recovering the declared COSINE space is what routes `cosinesimil` fields to the new `FP16_COSINE` / `SQ_COSINE` kernels; the override is intentionally scoped to cosine, since every other space agrees with the Faiss-reported metric.
- Expose `NativeRandomVectorScorer.create()` as a static factory with a package-private constructor. The constructor invokes the native `saveSearchContext`, which cannot be intercepted by Mockito's bytecode instrumentation, so tests now mock the factory via `Mockito.mockStatic` instead.

**Cleanup:**

- Remove the `convertToCosineScore()` call from `MemoryOptimizedKNNWeight` for the FP16 and SQ paths; cosine scores are now produced at the scorer level. The ADC (binary-quantized) path still scores a float query against 1-bit vectors and emits MaxIP-format scores, so it retains the MaxIP -> cosine post-conversion on the ADC branch only.
- Drop the manual MaxIP -> cosine conversion in `FaissScalarQuantizedBulkSimdScorerTests`; the scorer's output is already correctly cosine-scaled.

### Check List

### DIFF
--- a/jni/include/simd/similarity_function/similarity_function.h
+++ b/jni/include/simd/similarity_function/similarity_function.h
@@ -14,7 +14,11 @@ namespace knn_jni::simd::similarity_function {
         // L2 for FP16
         FP16_L2,
         SQ_IP,
-        SQ_L2
+        SQ_L2,
+        // Cosine for FP16. Vectors are L2-normalized so cosine = inner product with score = (1 + dot) / 2.
+        FP16_COSINE,
+        // Cosine for SQ. Same IP intermediate math as SQ_IP but with cosine score transform.
+        SQ_COSINE
     };
 
     struct SimilarityFunction;
--- a/jni/include/simd/similarity_function/similarity_function.h
+++ b/jni/include/simd/similarity_function/similarity_function.h
@@ -19,7 +19,11 @@ namespace knn_jni::simd::similarity_function {
         // L2 for FP16
         FP16_L2,
         SQ_IP,
-        SQ_L2
+        SQ_L2,
+        // Cosine for FP16. Vectors are L2-normalized so cosine = inner product with score = (1 + dot) / 2.
+        FP16_COSINE,
+        // Cosine for SQ. Same IP intermediate math as SQ_IP but with cosine score transform.
+        SQ_COSINE
     };
 
     struct SimilarityFunction;
@@ -44,8 +48,13 @@ namespace knn_jni::simd::similarity_function {
         SimilarityFunction* similarityFunction;
         // Faiss distance computation function.
         std::unique_ptr<faiss::DistanceComputer> faissFunction;
-        // Temp buffer which is reset per search.
+        // Temp buffer which is reset per search. Also used by getVectorPointer() to reassemble a
+        // vector that straddles two mmap regions, so it may be resized (and reallocated) mid-scan.
         std::vector<uint8_t> tmpBuffer;
+        // Dedicated buffer holding the FP16-converted query for the native FP16 kernel. Kept
+        // separate from tmpBuffer so that a cross-region vector reassembly (which resizes
+        // tmpBuffer) can never invalidate the query pointer held by the kernel.
+        std::vector<uint8_t> queryFP16Buffer;
 
         ~SimdVectorSearchContext();
 
--- a/jni/src/simd/similarity_function/avx512_simd_similarity_function.cpp
+++ b/jni/src/simd/similarity_function/avx512_simd_similarity_function.cpp
@@ -10,7 +10,6 @@
 #include "faiss_score_to_lucene_transform.cpp"
 
 
-
 //
 // FP16
 //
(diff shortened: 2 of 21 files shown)


ManasviGoyal:
@mulugetam can you address the failing IT tests? thanks!

ManasviGoyal:
@mulugetam Tried to run this code with the FP16 Exact Search Feature I am working on and there are the benchmark numbers across the differenct architectures on `cohere-1m`:

| Metric | AVX512 | ARM Neon | AVX512 SPR | AVX2 |
|---|---|---|---|---|
| Min Throughput | 4.51 ops/s | 2.63 ops/s | 4.91 ops/s | 3.98 ops/s |
| Mean Throughput | 4.57 ops/s | 2.65 ops/s | 5.38 ops/s | 4.04 ops/s |
| Median Throughput | 4.57 ops/s | 2.65 ops/s | 5.41 ops/s | 4.04 ops/s |
| Max Throughput | 4.58 ops/s | 2.81 ops/s | 5.48 ops/s | 4.04 ops/s |
| 50th pct latency | 217.07 ms | 385.86 ms | 187.47 ms | 246.15 ms |
| 90th pct latency | 219.11 ms | 411.14 ms | 198.86 ms | 247.01 ms |
| 99th pct latency | 225.09 ms | 419.91 ms | 207.12 ms | 247.78 ms |
| 99.9th pct latency | 229.50 ms | 426.00 ms | 213.87 ms | 251.08 ms |
| 99.99th pct latency | 233.05 ms | 430.71 ms | 222.47 ms | 259.70 ms |
| 100th pct latency | 233.66 ms | 430.92 ms | 231.83 ms | 464.15 ms |
| Mean recall@k | 1.00 | 1.00 | 0.99 | 1.00 |
| Mean recall@1 | 1.00 | 1.00 | 0.99 | 1.00 |

I see that with with AVX512 SPR there is a loss due to the FP16 accumulator which does increase the throughput but at the cost of recall drop.

mulugetam:
> I see that with with AVX512 SPR there is a loss due to the FP16 accumulator which does increase the throughput but at the cost of recall drop

@ManasviGoyal Thanks! Yes, the recall drop is expected since AVX-512 FP16 uses half-precision. The ~9% gain is good enough IMOH, since the memory-optimized search is already DRAM-bound, which limits the gains we can achieve with FP16 SIMD. Will check the failing IT tests and update.

src/test/java/org/opensearch/knn/memoryoptsearch/faiss/FaissScalarQuantizedBulkSimdScorerTests.java
ManasviGoyal: I think this function can be removed now as all the calls to this function are removed in this PR. Also need to remove the tests for this function

src/main/java/org/opensearch/knn/memoryoptsearch/faiss/NativeRandomVectorScorer.java
ManasviGoyal: I think we can make this package-private now since `create()` is now the entry point

jni/src/simd/similarity_function/avx512_simd_similarity_function.cpp
```diff
@@ -262,6 +261,251 @@ struct AVX512SPRFP16L2 final : BaseSimilarityFunction<BulkScoreTransformFunc, Sc
     }
 };
 
+//
+// Native AVX512-FP16 Inner Product for Cosine Similarity
+//
+// Uses native AVX512-FP16 FMA instructions (_mm512_fmadd_ph) at 32 FP16 elements per
+// 512-bit register, with on-the-fly FP32->FP16 query conversion. To bound FP16
+// accumulation rounding error across long vectors, the FP16 accumulator is drained
+// into an FP32 accumulator every FP16_DRAIN_INTERVAL iterations -- trading a small
```
ManasviGoyal: NIT: Is this outdated? dont see FP16_DRAIN_INTERVAL used anywhere, looks like it drains once per vector at the end, not periodically. can you update the comment?

jni/src/simd/similarity_function/avx512_simd_similarity_function.cpp
```diff
@@ -262,6 +261,251 @@ struct AVX512SPRFP16L2 final : BaseSimilarityFunction<BulkScoreTransformFunc, Sc
...
+//
+// PRECONDITION: query and data vectors are L2-normalized. This is the cosine
+// invariant. Running partial sums stay in [-1, +1] throughout, where FP16 ULP
+// is ~5e-4. Cumulative rounding error over dim=768 is ~sqrt(N)*ulp ~= 2e-3
+// expected, ~6e-3 worst-case — well below cosine ranking noise. Therefore we
+// accumulate entirely in FP16 and drain to FP32 only once per vector at the
+// end, before horizontal reduce.
+//
+// DO NOT route any unnormalized similarity through this kernel. The previous
+// version of this code drained to FP32 every FP16_DRAIN_INTERVAL iterations
```
ManasviGoyal: NIT: same here. FP16_DRAIN_INTERVAL is not used

ManasviGoyal (changes requested):
Made some comments for minor fixes

ManasviGoyal:
@mulugetam Do you have some benchmarking numbers that you can share here? curious to know the performance numbers you observed.

mulugetam:
> @mulugetam Do you have some benchmarking numbers that you can share here? curious to know the performance numbers you observed.

I’ll run Cohere-10M and share the performance numbers. Meanwhile, could you also review the BF16 PR (#3190)? Thanks!

ManasviGoyal:
@mulugetam Also can you look into the agent suggestions? I have the same concern regarding the `tmpBuffer` issue it mentions. Thanks!

jni/include/simd/similarity_function/similarity_function.h
```diff
@@ -39,8 +43,13 @@ namespace knn_jni::simd::similarity_function {
...
         // Faiss distance computation function.
         std::unique_ptr<faiss::DistanceComputer> faissFunction;
-        // Temp buffer which is reset per search.
+        // Temp buffer which is reset per search. Also used by getVectorPointer() to reassemble a
+        // vector that straddles two mmap regions, so it may be resized (and reallocated) mid-scan.
         std::vector<uint8_t> tmpBuffer;
+        // Dedicated buffer holding the FP16-converted query for the native FP16 kernel. Kept
+        // separate from tmpBuffer so that a cross-region vector reassembly (which resizes
+        // tmpBuffer) can never invalidate the query pointer held by the kernel.
+        std::vector<uint8_t> queryFP16Buffer;
```
0ctopus13prime, line 57: Curious, then what `queryVectorSimdAligned` will have for FP16?

mulugetam: `queryVectorSimdAligned` holds the query as `FP32 floats`.

0ctopus13prime: Right, do you think we can assign the query FP16 vector in there?

ManasviGoyal (approved):
LGTM!

0ctopus13prime:
## Performance (Cosine)
Data set : Cohere-10m
Space : Cosine

```
+----------------------------+------------+------------+------------+--------------------+
|           Metric           |  baseline  | candidate  | Difference | Percent Difference |
+----------------------------+------------+------------+------------+--------------------+
| Min Throughput             | 257.125667 | 297.262500 |  40.136833 |             15.61% |
| Mean Throughput            | 299.723333 | 329.868500 |  30.145167 |             10.06% |
| Median Throughput          | 303.701333 | 332.923500 |  29.222167 |              9.62% |
| Max Throughput             | 310.683333 | 337.560000 |  26.876667 |              8.65% |
| 50th percentile latency    |  50.630171 |  46.665527 |  -3.964644 |             -7.83% |
| 90th percentile latency    |  54.226149 |  50.308743 |  -3.917406 |             -7.22% |
| 99th percentile latency    |  90.888529 |  80.989479 |  -9.899050 |            -10.89% |
| 99.9th percentile latency  | 183.224205 | 176.110197 |  -7.114008 |             -3.88% |
| 99.99th percentile latency | 207.590601 | 194.125549 | -13.465052 |             -6.49% |
+----------------------------+------------+------------+------------+--------------------+
```

0ctopus13prime:
## Performance (Cosine, SQ 32x)
Data set : Cohere 10m
Space : Cosine

```
+----------------------------+------------+------------+------------+--------------------+
|           Metric           |  baseline  | candidate  | Difference | Percent Difference |
+----------------------------+------------+------------+------------+--------------------+
| Min Throughput             | 110.654615 | 112.966667 |   2.312051 |              2.09% |
| Mean Throughput            | 131.454250 | 133.087778 |   1.633528 |              1.24% |
| Median Throughput          | 133.468250 | 134.453704 |   0.985454 |              0.74% |
| Max Throughput             | 136.620500 | 136.091852 |  -0.528648 |             -0.39% |
| 50th percentile latency    |  77.317499 |  77.715661 |   0.398163 |              0.51% |
| 90th percentile latency    |  96.039391 |  95.668046 |  -0.371345 |             -0.39% |
| 99th percentile latency    | 144.314373 | 143.945664 |  -0.368710 |             -0.26% |
| 99.9th percentile latency  | 171.231137 | 163.216413 |  -8.014725 |             -4.68% |
| 99.99th percentile latency | 204.783170 | 206.783902 |   2.000732 |              0.98% |
+----------------------------+------------+------------+------------+--------------------+
```

0ctopus13prime:
@mulugetam 
I can see meaningful performance jump on cosine FP16, but not on SQ 32x. 
Is this expected? Does it align with your micro benchmark against SQ 32x?

mulugetam:
> I can see meaningful performance jump on cosine FP16, but not on SQ 32x.
> Is this expected?

@0ctopus13prime Yes, this is expected. The FP16 and SQ changes aren't symmetric. FP16 cosine previously had no SIMD fast path at all; it fell through to the Lucene delegate. This PR gives it both SIMD entry and the new AVX512NativeFP16IP kernel (_mm512_fmadd_ph, 32 FP16 lanes/register), and that hits the hot inner loop; hence the meaningful jump. SQ was already running its inner product in SIMD. All this PR changes for SQ is moving the (1 + dot)/2 cosine transform in-kernel and dropping the post-hoc convertToCosineScore().

0ctopus13prime:
Hi @mulugetam 
Can you rebase it? LGTM, I think we're good to go

mulugetam:
> Hi @mulugetam Can you rebase it? LGTM, I think we're good to go

@0ctopus13prime Will rebase this after #3341 merges (since I will get a merge conflict).

0ctopus13prime:
Hi @mulugetam 
Can you rebase this to main? #3341 just got merged

mulugetam:
@0ctopus13prime rebased and all checks pass.

0ctopus13prime (approved):

naveentatikonda (approved):

Issue #3316, opened 2026-05-09
Support `_source.exclude` with index mapping for derived source


shatejas:
## Description

When a user specifies any field in `_source.exclude`, the segment size becomes unexpectedly large. This occurs regardless of the field type. As long as `_source.exclude` is not empty, the segment ends up containing the float vector JSON string, even when KNN derived source is enabled.

Ideally, when KNN derived source is enabled, the float vector JSON string should not be stored in the segment. However, this expected behavior does not hold when `_source.exclude` is used.

## Root Cause

This happens for two reasons:

### 1. `_recovery_source` contains the full vector during flush

During the flush phase, KNN derived source works correctly for `_source`, which stores a masked value `1` instead of the full float vector JSON string. However, `_recovery_source` still contains the float vector JSON string, which increases the segment size.

### 2. `_source` becomes bloated after merge

During the merge phase, `_recovery_source` is correctly removed by `RecoverySourcePruneMergePolicy`. However, `_source` ends up containing the float vector JSON string again because `KNN10010DerivedSourceStoredFieldsReader`'s merge-specific instantiation does not work with `RecoverySourcePruneMergePolicy`.

#### Detailed explanation

`RecoverySourcePruneMergePolicy` wraps the stored fields reader in a `RecoverySourcePruningStoredFieldsReader`, which delegates to `KNN10010DerivedSourceStoredFieldsReader`. When `MergeState` calls `getMergeInstance()`, it gets back a `RecoverySourcePruningStoredFieldsReader` — not a `KNN10010DerivedSourceStoredFieldsReader`.

In `KNN10010DerivedSourceStoredFieldsWriter.merge()`, the code calls `wrapForMerge()` on each stored fields reader:

```java
public static StoredFieldsReader wrapForMerge(StoredFieldsReader storedFieldsReader) {
    if (storedFieldsReader instanceof KNN10010DerivedSourceStoredFieldsReader) {
        return ((KNN10010DerivedSourceStoredFieldsReader) storedFieldsReader).cloneForMerge();
    }
    return storedFieldsReader; // <-- RecoverySourcePruningStoredFieldsReader hits this path
}
```

Since the actual type is `RecoverySourcePruningStoredFieldsReader`, the `instanceof` check fails, `cloneForMerge()` is never called, and `shouldInject` remains `true`. This causes the reader to always restore the float vector string into `_source` during merge — even though the pre-merge `_source` did not contain it.

## Expected Behavior

When KNN derived source is enabled:
[...]

PR #3399, opened 2026-06-29
Fix routing loss in by-query with derived source

MdTanwer:
## Summary

- Skip derived-source visitor wrapping when the stored-fields fetch does not request `_source`, so metadata-only scroll fetches used by `_delete_by_query` / `_update_by_query` still return `_routing`.
- Override `getMergeInstance()` so scroll/sequential stored-field reads use a derived-source reader with injection enabled; segment merges continue to use `wrapForMerge()` with injection disabled.

Fixes routing_missing_exception on join-field child documents that contain `knn_vector` when `_routing.required` is true and `index.knn.derived_source.enabled` 

Fix opensearch-project/k-NN#3395


src/main/java/org/opensearch/knn/index/codec/KNN10010Codec/KNN10010DerivedSourceStoredFieldsReader.java
```diff
@@ -69,14 +71,33 @@ public void document(int docId, StoredFieldVisitor storedFieldVisitor) throws IO
...
+     * {@code _update_by_query} fetch metadata fields like {@code _routing} with {@code fetch_source=false},
+     * so wrapping those visitors can prevent metadata from being returned.
+     */
+    private boolean visitorRequestsSource(StoredFieldVisitor storedFieldVisitor) throws IOException {
+        if (storedFieldVisitor instanceof FieldsVisitor fieldsVisitor) {
+            fieldsVisitor.reset();
+            FieldInfo sourceField = segmentReadState.fieldInfos.fieldInfo(SourceFieldMapper.NAME);
+            if (sourceField == null) {
+                return false;
+            }
```
shatejas: - why is this resetting, core should be handling this the codec reader should not mutate fieldvisitor?
- There is no need to bypass source here derivedSourceVisitor already takes care of it in binaryField

Also can you explain how is this solving the routing issue? to my understanding missing routing was fixed by https://github.com/opensearch-project/k-NN/pull/2925/changes, am I missing something here?

src/main/java/org/opensearch/knn/index/codec/KNN10010Codec/KNN10010DerivedSourceStoredFieldsReader.java
```diff
@@ -118,10 +139,28 @@ public void close() throws IOException {
         IOUtils.close(delegate, derivedSourceReaders);
     }
 
+    /**
+     * Scroll and other sequential stored-field fetches use {@link StoredFieldsReader#getMergeInstance()}.
+     * Segment merges explicitly call {@link #wrapForMerge(StoredFieldsReader)} to disable injection.
+     */
+    @Override
+    public StoredFieldsReader getMergeInstance() {
```
shatejas: > Scroll and other sequential stored-field fetches use {@link StoredFieldsReader#getMergeInstance()}.

The merge instance is used because it needs sequential read access, core should not be using [mergeInstance](https://github.com/opensearch-project/OpenSearch/issues/21583) is search path anyway. So if I understand correctly are knn indices skipping the madvise and thats wise we need this?

MdTanwer:
Thanks for the review — agreed on all points, and the PR has been updated accordingly.

On **`reset()` / mutating `FieldVisitor`**: I removed `visitorRequestsSource()` and the `FieldsVisitor.reset()` calls; the codec reader should not probe or mutate the caller's visitor.

On **bypassing source / `binaryField`**: `DerivedSourceStoredFieldVisitor` already only rewrites `_source` in `binaryField` and delegates `_routing` via `stringField` after #2925, so the KNN10010 reader bypass was unnecessary and has been reverted.

On **routing / #2925**: you're right for KNN10010 — #2925 already fixed routing there; the remaining gap was `KNN9120DerivedSourceStoredFieldVisitor`, which still lacked the same `stringField`/primitive delegation, so this PR backports that fix to the legacy codec only.

On **`getMergeInstance()` / madvise**: overriding `getMergeInstance()` isn't needed for routing and conflicts with our merge-path design (#3316), so I've dropped it; k-NN isn't skipping madvise, and core reusing merge instances .

VijayanB:
@MdTanwer Can we add integration tests?

shatejas:
@MdTanwer Understood so this was missed for 2.19, Thanks for this! Can we also have a backport PR for 2.19 branch so it can go in next release.

Also please take care of the CIs

VijayanB:
Please add change log as well.

shatejas:
> @MdTanwer Can we add integration tests?

@VijayanB is there a good way to integ test this? we can only do BWC I think. let me know if I am missing something

VijayanB:
> > @MdTanwer Can we add integration tests?
> 
> @VijayanB is there a good way to integ test this? we can only do BWC I think. let me know if I am missing something

@shatejas I haven't looked deep into this, but just checking from the GitHub issue https://github.com/opensearch-project/k-NN/issues/3395, there were reproducible steps provided. Can we use those repro steps as the basis for an integration test?

MdTanwer:
> Please add change log as well.

Done

MdTanwer:
> > > @MdTanwer Can we add integration tests?
> > 
> > 
> > @VijayanB is there a good way to integ test this? we can only do BWC I think. let me know if I am missing something
> 
> @shatejas I haven't looked deep into this, but just checking from the GitHub issue #3395, there were reproducible steps provided. Can we use those repro steps as the basis for an integration test?

We can add an integration test based on the repro steps — it doesn’t have to be BWC-only — the key is to write it as a REST IT using our existing `KNNRestTestCase/DerivedSourceIT` harness (like other derived-source ITs), and focus on the observable behavior: join-field child doc + _routing.required + populated knn_vector must let `_delete_by_query / `_update_by_query succeed when request routing is provided.

If the concern is specifically validating the legacy codec path (KNN9120), then yes: that part is best covered via BWC/rolling-upgrade (or a dedicated BWC test that opens an index created with the older version and runs the same by-query assertions), because we can’t reliably force the runtime to produce KNN9120 segments on a fresh index in a single-version IT.

So: use the repro steps for a normal integration test for the functional behavior, and use BWC to prove the legacy-codec regression/fix path.

MdTanwer:
@VijayanB @shatejas   have look this PR when you get time thanks in advance ):

shatejas:
> If the concern is specifically validating the legacy codec path (KNN9120), then yes: that part is best covered via BWC/rolling-upgrade (or a dedicated BWC test that opens an index created with the older version and runs the same by-query assertions)

Exactly the integ test won't create an older version, so it won't test the change in this PR. If there is an integ test missing for routing we should add it. But for this specific change BWC seems to be the only way

MdTanwer:
@navneet1v @shatejas, if there’s anything needed from my side, please let me know.

Issue #2880, opened 2025-09-17
[BUG] Derived Source deserialization exception on doc 4xx


jed326:
**What is the bug?**
Whenever derived fields is enabled on an index and an invalid document comes in, we will encounter a deserialization exception and cause the engine + shard to fail, leading to red index.

**How can one reproduce the bug?**
Create index:
```
curl --location --request PUT 'localhost:9200/ds-enabled' \
--header 'Content-Type: application/json' \
--data '{
    "settings": {
        "index": {
            "replication": {
                "type": "DOCUMENT"
            },
            "mapping": {
                "coerce": "false"
            },
            "refresh_interval": "3h",
            "knn": "true",
            "analysis": {
                "analyzer": {
                    "delimited_tf": {
                        "filter": [
                            "delimited_term_freq"
                        ],
                        "index_options": "freqs",
                        "tokenizer": "whitespace"
                    }
                }
            },
            "number_of_replicas": "0",
            "number_of_shards": "1",
            "knn.derived_source": {
                "enabled": "true"
            }
        }
    },
    "mappings": {
        "dynamic": "strict",
        "properties": {
            "text_field": {
                "type": "text",
                "index_options": "freqs",
                "analyzer": "delimited_tf"
            },
            "vector_field": {
                "type": "knn_vector",
                "dimension": 2,
                "method": {
                    "engine": "faiss",
                    "space_type": "innerproduct",
                    "name": "hnsw",
                    "parameters": {}
                }
            }
        }
    }
}'
```

**What is the expected behavior?**
OpenSearch should gracefully reject invalid documents without red index.

**What is your host/environment?**
 - Plugins

**Do you have any screenshots?**

**Do you have any additional context?**
Add any other context about the problem.

PR #2882, opened 2025-09-18
Fix derived source deserialization bug on invalid documents

jed326:
### Description
This commits makes the `KNN10010DerivedSourceStoredFieldsWriter` skip writing whenever it encounters a `_source` field that is not in JSON format. 

This protects 2 things:
1.  The de-serialization exception in #2880 which is causing red index
2. `KNN10010DerivedSourceStoredFieldsReader` also attempts to de-serialize the `_source`, so this makes sure no non-JSON content is ever written into the `_source`.

Added a new derived source test which checks that derived source enabled and derived source disabled handle this in the same way.

### Related Issues
Resolves https://github.com/opensearch-project/k-NN/issues/2880

### Check List
- [x] New functionality includes testing.
- [x] New functionality has been documented.
- [x] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
- [x] Commits are signed per the DCO using `--signoff`.
- [x] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/k-NN/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #3402, opened 2026-06-29
Preserve non-XContent source in derived source

### LINKED ISSUE #3401: [BUG] Derived source should preserve non-XContent _source fields
**What is the bug?**

Derived source currently treats every stored `_source` field as parseable XContent. However, OpenSearch core can store internal documents whose `_source` is raw bytes rather than JSON/XContent. One important example is a no-op tombstone: OpenSearch stores the no-op reason string directly in `_source`, such as `filling gaps` or an exception string.

This interacts badly with k-NN derived source:

* The write path in `KNN10010DerivedSourceStoredFieldsWriter.writeField(...)` intercepts `_source` and attempts to parse it.
* PR #2882, which fixed #2880, catches `NotXContentException`, but the catch block returns without delegating the stored field. That means the raw `_source` is dropped.
* If the raw `_source` is preserved, the read path can still attempt to parse it in `DerivedSourceVectorTransformer.injectVectors(...)` when `KNN10010DerivedSourceStoredFieldsReader` injects derived vector fields.

So #2882 appears to have changed the failure mode for non-XContent `_source`: instead of failing while writing the stored field, the stored field can be omitted, and later consumers may see a missing or unparsable no-op `_source`.

At the code-path level:

1. Create or use an index with k-NN derived source enabled and at least one derived vector field.
2. Produce an OpenSearch no-op tombstone in that index. OpenSearch creates these for sequence-number gap filling and for some failed indexing operations after a sequence number has been assigned.
3. The no-op tombstone stores a raw reason string in `_source`, not JSON/XContent.
4. During stored field writing, `KNN10010DerivedSourceStoredFieldsWriter.writeField(...)` receives `_source`, attempts `XContentHelper.convertToMap(...)`, catches `NotXContentException`, logs, and returns without `delegate.writeField(...)`.
5. The stored `_source` is then absent for that no-op tombstone. If the writer is changed to preserve the field, the read-side transformer also needs to pass non-XContent bytes through unchanged instead of parsing them.

Related public context:

* #2880 reported the original write-time `NotXContentException` with derived source.
* #2882 fixed #2880 by catching `NotXContentException`, but the current catch block drops the stored `_source` instead of preserving it.
* opensearch-project/OpenSearch#19330 discusses OpenSearch no-op tombstones storing non-JSON `_source`.

Derived source should only mask or inject vector fields when `_source` is parseable XContent. If `_source` is not XContent, k-NN should preserve the original stored field unchanged.

Concretely:

* On write, `KNN10010DerivedSourceStoredFieldsWriter` should delegate the original `_source` bytes when parsing throws `NotXContentException`.
* On read, `DerivedSourceVectorTransformer` should return non-XContent `_source` bytes unchanged rather than attempting vector injection.

This keeps derived source behavior safe for OpenSearch internal stored fields while preserving the existing vector masking/injection behavior for normal document `_source`.

* OS: Linux
* Version: observed on the current `main` code path and relevant to versions that include the #2882 behavior
* Plugins: k-NN with derived source enabled

No.

The codec layer sees stored Lucene fields and does not distinguish between user documents and OpenSearch internal tombstone documents. Since no-op tombstones use the same `_source` field name, they can hit both the derived-source writer and reader paths.

The safe behavior is pass-through for non-XContent `_source`: derived source cannot transform vector fields without parsing XContent, and dropping the field loses information needed by other OpenSearch recovery/history readers.

aneesh-db:
### Description
Preserve raw non-XContent `_source` fields when k-NN derived source is enabled.

OpenSearch engine-generated documents can store `_source` as raw bytes rather than XContent. For example, no-op tombstones store their reason string directly in `_source`. Derived source can only mask or inject vector fields after parsing XContent, so this change makes non-XContent `_source` a pass-through case:

* `KNN10010DerivedSourceStoredFieldsWriter` delegates the original `_source` bytes when parsing throws `NotXContentException`.
* `DerivedSourceVectorTransformer` returns non-XContent `_source` bytes unchanged on read instead of attempting vector injection.
* Unit tests cover both the writer and reader-side transformer behavior.

This addresses the follow-up failure mode introduced by #2882 while preserving the original intent of #2880.

### Related Issues
Resolves #3401

Related:
* #2880
* #2882
* opensearch-project/OpenSearch#19330
* opensearch-project/OpenSearch#22317

### Check List
- [ ] New functionality has been documented. Not applicable; this is an internal bug fix.
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md). Not applicable; there are no API changes.
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose). Not applicable; this is an internal bug fix.

### DIFF
--- a/src/main/java/org/opensearch/knn/index/codec/derivedsource/DerivedSourceVectorTransformer.java
+++ b/src/main/java/org/opensearch/knn/index/codec/derivedsource/DerivedSourceVectorTransformer.java
@@ -12,6 +12,7 @@ import org.opensearch.common.io.stream.BytesStreamOutput;
 import org.opensearch.common.xcontent.XContentHelper;
 import org.opensearch.common.xcontent.support.XContentMapValues;
 import org.opensearch.core.common.bytes.BytesReference;
+import org.opensearch.core.compress.NotXContentException;
 import org.opensearch.core.xcontent.MediaType;
 import org.opensearch.core.xcontent.MediaTypeRegistry;
 import org.opensearch.core.xcontent.XContentBuilder;
@@ -140,11 +141,23 @@ public class DerivedSourceVectorTransformer {
         // Reference:
         // https://github.com/opensearch-project/OpenSearch/blob/2.18.0/server/src/main/java/org/opensearch/index/mapper/SourceFieldMapper.java#L322
         // Deserialize the source into a modifiable map
-        Tuple<? extends MediaType, Map<String, Object>> mapTuple = XContentHelper.convertToMap(
-            BytesReference.fromByteBuffer(ByteBuffer.wrap(sourceAsBytes)),
-            true,
-            MediaTypeRegistry.getDefaultMediaType()
-        );
+        Tuple<? extends MediaType, Map<String, Object>> mapTuple;
+        try {
+            mapTuple = XContentHelper.convertToMap(
+                BytesReference.fromByteBuffer(ByteBuffer.wrap(sourceAsBytes)),
+                true,
+                MediaTypeRegistry.getDefaultMediaType()
+            );
+        } catch (NotXContentException e) {
+            // Some OpenSearch internal documents, such as no-op tombstones, store _source as raw bytes rather than XContent.
+            // Derived source can only inject vector fields after parsing XContent, so preserve non-XContent _source unchanged.
+            log.warn(
+                "Encountered NotXContent while deserializing _source field. Instead found String: [{}]",
+                new String(sourceAsBytes, 0, Math.min(sourceAsBytes.length, 512)),
+                e
+            );
+            return sourceAsBytes;
+        }
         // Have to create a copy of the map here to ensure that is mutable
         Map<String, Object> sourceAsMap = mapTuple.v2();
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 * Upgrade Lucene to 10.5.0 [#3411](https://github.com/opensearch-project/k-NN/pull/3411)
 
 ### Bug Fixes
+* Preserve raw non-XContent `_source` fields when derived source is enabled [#3402](https://github.com/opensearch-project/k-NN/pull/3402)
 * Fix NPE in nested kNN search when index contains documents without nested object [#3368](https://github.com/opensearch-project/k-NN/pull/3368)
 * Turn off ACORN for MOS to match default Lucene HNSW behavior [#3346](https://github.com/opensearch-project/k-NN/pull/3346)
 * Preserve mixed-case derived source vector field names and add backward-compatible field resolution for previously lowercased segment metadata [#3313](https://github.com/opensearch-project/k-NN/pull/3313)
--- a/src/main/java/org/opensearch/knn/index/codec/KNN10010Codec/KNN10010DerivedSourceStoredFieldsWriter.java
+++ b/src/main/java/org/opensearch/knn/index/codec/KNN10010Codec/KNN10010DerivedSourceStoredFieldsWriter.java
@@ -242,13 +242,14 @@ public class KNN10010DerivedSourceStoredFieldsWriter extends StoredFieldsWriter 
                 MediaTypeRegistry.JSON
             );
         } catch (NotXContentException e) {
-            // If the content is not XContent, skip writing altogether. See:
-            // https://github.com/opensearch-project/k-NN/issues/2880
+            // Some OpenSearch internal documents, such as no-op tombstones, store _source as raw bytes rather than XContent.
+            // Derived source can only mask vector fields after parsing XContent, so preserve non-XContent _source unchanged.
             log.warn(
                 "Encountered NotXContent while deserializing _source field. Instead found String: [{}]",
-                new String(bytesRef.bytes, 0, Math.min(bytesRef.bytes.length, 512)),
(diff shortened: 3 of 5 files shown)


src/main/java/org/opensearch/knn/index/codec/derivedsource/DerivedSourceVectorTransformer.java
```diff
@@ -140,11 +141,23 @@ public byte[] injectVectors(int docId, byte[] sourceAsBytes) throws IOException
...
+        } catch (NotXContentException e) {
+            // Some OpenSearch internal documents, such as no-op tombstones, store _source as raw bytes rather than XContent.
+            // Derived source can only inject vector fields after parsing XContent, so preserve non-XContent _source unchanged.
+            log.warn(
+                "Encountered NotXContent while deserializing _source field. Instead found String: [{}]",
+                new String(sourceAsBytes, 0, Math.min(sourceAsBytes.length, 512)),
+                e
+            );
+            return sourceAsBytes;
+        }
```
shatejas, line 160: This was discussed in the previous PR https://github.com/opensearch-project/k-NN/pull/2882/changes#r2360498165. There is no value on in storing this since its a soft delete. if LuceneChangeSnapshot is not handling null case shouldn't it be fixed to handle NPE instead, is there any other case apart from no-op tombstone that we are missing here?

Vikasht34:
So As far as I understand this issue , Documents like replication tombstones would lose their _source field entirely during segment merges on k-NN indices with derived source enabled? and is this causing replication failures?

src/main/java/org/opensearch/knn/index/codec/derivedsource/DerivedSourceVectorTransformer.java
Vikasht34: @shatejas  but  StoredFieldsWriter that silently drops fields is dangerous, The write path should be transparent , it's like if I don't get it what it is , I will pass through.  I agree fixing LuceneChangeSnapshot should be long term fix but our writer shouldn't silently discard data it doesn't understand

aneesh-db:
> So As far as I understand this issue , Documents like replication tombstones would lose their _source field entirely during segment merges on k-NN indices with derived source enabled? and is this causing replication failures?

@Vikasht34 Yes, this downstream failure is during recovery/readback rather than normal document replication: OpenSearch's Lucene changes snapshot may read the no-op tombstone back and
  reconstruct a `Translog.NoOp` from its stored `_source`. If k-NN dropped that field, the source is missing and the core recovery path can fail, tracked in [opensearch-
  project/OpenSearch#22317](https://github.com/opensearch-project/OpenSearch/issues/22317).

src/main/java/org/opensearch/knn/index/codec/derivedsource/DerivedSourceVectorTransformer.java
aneesh-db: yes, this matches my thinking.
I agree that OpenSearch core should defensively handle a missing no-op tombstone source, and that is a separate fix I have opened https://github.com/opensearch-project/OpenSearch/pull/22318. But I do not think k-NN’s stored-fields writer should silently drop a field just because derived source cannot parse it as XContent.

This is also the safer version of the option discussed in #2882: preserve non-XContent bytes and add a read-side guard. The read-side guard in this PR is what avoids shifting the failure from write time to read time.

shatejas: tombstones have just the error message in source which is not used in any code flow - it was ack in the previous PR that there is no value in storing it. I am not denying the fact that we shouldn't be dropping the data, at the same time we shouldn't be storing the data thats not useful

I am okay with the change overall but if there is no identified use case why not fix the root cause of NPE. so the question was more about use case identification

shatejas:
@aneesh-db please fix the CIs overall I am good with the change, just highlight any other case apart from tombstones so we can make a judgement call on if there is anything else needed for derived source in the fix

src/main/java/org/opensearch/knn/index/codec/derivedsource/DerivedSourceVectorTransformer.java
Vikasht34: Discussed offline with @shatejas  Make Sense ..

aneesh-db:
> @aneesh-db please fix the CIs overall I am good with the change, just highlight any other case apart from tombstones so we can make a judgement call on if there is anything else needed for derived source in the fix

Thanks @shatejas for the fast review,  I have checked code for non-XContent `_source`, and no other producers besides NoOp tombstones.

On CI: the compile failures were not caused by my changes. They came from the recent Lucene 10.5 API changes that current OpenSearch snapshots now pull in. k-NN still had a few Lucene 10.4 assumptions around quantized vector APIs, flat vector scorer access, HNSW writer construction, merge callbacks, and `BitSet`. I updated the affected files in this PR as well.


PS: given the main bug is on opensearch, appreciate if u can ask someone from opensearch team to review https://github.com/opensearch-project/OpenSearch/pull/22318 on priority.

navneet1v:
@aneesh-db can you please rebase your branch.

aneesh-db:
@navneet1v rebased the branch.

navneet1v:
Will review it today once all CI passes

navneet1v (approved):

shatejas (approved):

PR #3406, opened 2026-06-30
Update maven2 mirror repository url order

peterzhuamazon:
### Description
Update maven2 mirror repository url order

### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/6278

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #3411, opened 2026-07-01
Upgrade to lucene 10.5

kotwanikunal:
### Description

Upgrades Lucene from 10.4.0 to 10.5.0.

Key Changes

1. **Merge API:** `KnnVectorsWriter.mergeOneField()` now returns `IORunnable` (was `void`). `FlatVectorsWriter.mergeOneField()` is `final` and always returns `null` (work done inline). Our writers do all work synchronously and return `null` -- the caller null-checks, so the flow is unchanged.

2. **Radial search:** `traversalSimilarity` replaced by `decay` parameter. We pass `decay=1.0f` (max quality, no traversal pruning) to preserve the wide-traversal behavior of the old `0.95 * resultSimilarity`.

3. **Import relocations:** `QuantizedByteVectorValues` moved to `org.apache.lucene.util.quantization`; `ScalarEncoding` moved to `QuantizedByteVectorValues.ScalarEncoding`. Backward-compat inner classes now extend `LegacyQuantizedByteVectorValues`.

4. **Other:** `FlatVectorsReader.getFlatVectorScorer()` now takes a field name param. `BitSet` added abstract `nextClearBit(int, int)`. `Lucene99HnswVectorsWriter` constructor takes an additional `FlatVectorsFormat` param.

### Related Issues
Resolves failing builds on main

### Check List

### DIFF
--- a/build-tools/repositories.gradle
+++ b/build-tools/repositories.gradle
@@ -5,8 +5,8 @@
 
 repositories {
     mavenLocal()
-    maven { url "https://ci.opensearch.org/ci/dbc/snapshots/maven/" }
     maven { url "https://ci.opensearch.org/maven2/" }
+    maven { url "https://ci.opensearch.org/ci/dbc/snapshots/maven/" }
     mavenCentral()
     maven { url "https://plugins.gradle.org/m2/" }
 }
--- a/src/main/java/org/apache/lucene/backward_codecs/lucene99/Lucene99ScalarQuantizedVectorsWriter.java
+++ b/src/main/java/org/apache/lucene/backward_codecs/lucene99/Lucene99ScalarQuantizedVectorsWriter.java
@@ -54,12 +54,15 @@ import org.apache.lucene.search.VectorScorer;
 import org.apache.lucene.store.IndexInput;
 import org.apache.lucene.store.IndexOutput;
 import org.apache.lucene.util.IOUtils;
+import org.apache.lucene.util.IORunnable;
 import org.apache.lucene.util.InfoStream;
 import org.apache.lucene.util.RamUsageEstimator;
 import org.apache.lucene.util.VectorUtil;
 import org.apache.lucene.util.hnsw.CloseableRandomVectorScorerSupplier;
 import org.apache.lucene.util.hnsw.RandomVectorScorerSupplier;
 import org.apache.lucene.util.hnsw.UpdateableRandomVectorScorer;
+import org.apache.lucene.util.quantization.BaseQuantizedByteVectorValues;
+import org.apache.lucene.util.quantization.LegacyQuantizedByteVectorValues;
 import org.apache.lucene.util.quantization.QuantizedByteVectorValues;
 import org.apache.lucene.util.quantization.QuantizedVectorsReader;
 import org.apache.lucene.util.quantization.ScalarQuantizer;
@@ -214,8 +217,9 @@ public final class Lucene99ScalarQuantizedVectorsWriter extends FlatVectorsWrite
     }
 
     @Override
-    public void mergeOneField(FieldInfo fieldInfo, MergeState mergeState) throws IOException {
-        rawVectorDelegate.mergeOneField(fieldInfo, mergeState);
+    public void mergeOneFlatVectorField(FieldInfo fieldInfo, MergeState mergeState) throws IOException {
+        IORunnable rawMerge = rawVectorDelegate.mergeOneField(fieldInfo, mergeState);
+        if (rawMerge != null) rawMerge.run();
         // Since we know we will not be searching for additional indexing, we can just write the
         // the vectors directly to the new segment.
         // No need to use temporary file as we don't have to re-open for reading
@@ -244,18 +248,19 @@ public final class Lucene99ScalarQuantizedVectorsWriter extends FlatVectorsWrite
         }
     }
 
-    @Override
     public CloseableRandomVectorScorerSupplier mergeOneFieldToIndex(FieldInfo fieldInfo, MergeState mergeState) throws IOException {
         if (fieldInfo.getVectorEncoding().equals(VectorEncoding.FLOAT32)) {
             // Simply merge the underlying delegate, which just copies the raw vector data to a new
             // segment file
-            rawVectorDelegate.mergeOneField(fieldInfo, mergeState);
+            IORunnable rawMerge = rawVectorDelegate.mergeOneField(fieldInfo, mergeState);
+            if (rawMerge != null) rawMerge.run();
             ScalarQuantizer mergedQuantizationState = mergeAndRecalculateQuantiles(mergeState, fieldInfo, confidenceInterval, bits);
             return mergeOneFieldToIndex(segmentWriteState, fieldInfo, mergeState, mergedQuantizationState);
         }
         // We only merge the delegate, since the field type isn't float32, quantization wasn't
         // supported, so bypass it.
-        return rawVectorDelegate.mergeOneFieldToIndex(fieldInfo, mergeState);
+        // In Lucene 10.5, mergeOneFieldToIndex is removed. Just return null for non-float32.
+        return null;
     }
 
     @Override
@@ -696,7 +701,7 @@ public final class Lucene99ScalarQuantizedVectorsWriter extends FlatVectorsWrite
      */
     public static DocsWithFieldSet writeQuantizedVectorData(
         IndexOutput output,
-        QuantizedByteVectorValues quantizedByteVectorValues,
+        LegacyQuantizedByteVectorValues quantizedByteVectorValues,
         byte bits,
         boolean compress
     ) throws IOException {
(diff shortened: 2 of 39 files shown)


Vikasht34 (approved):
LGTM

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss1040ScalarQuantizedKnnVectorsWriter.java
```diff
@@ -149,7 +152,7 @@ public void mergeOneField(FieldInfo fieldInfo, MergeState mergeState) throws IOE
             final FloatVectorValues floatVectorValues = flatVectorsReader.getFloatVectorValues(fieldInfo.getName());
             if (floatVectorValues == null || floatVectorValues.size() == 0) {
                 log.debug("No scalar-quantized vectors found for field [{}], skipping native build", fieldInfo.getName());
-                return;
+                return null;
```
VijayanB, line 155: can you confirm that this will be checked by the caller that it is safe to pass null?

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss1040ScalarQuantizedKnnVectorsWriter.java
```diff
@@ -133,12 +134,14 @@ public void flush(int maxDoc, Sorter.DocMap sortMap) throws IOException {
...
-    public void mergeOneField(FieldInfo fieldInfo, MergeState mergeState) throws IOException {
+    public IORunnable mergeOneField(FieldInfo fieldInfo, MergeState mergeState) throws IOException {
         // Setting field info
         this.fieldInfo = fieldInfo;
 
         // Merge, finish, and close the flat writer so that files are readable.
-        flatVectorsWriter.mergeOneField(fieldInfo, mergeState);
+        IORunnable mergeRunnable = flatVectorsWriter.mergeOneField(fieldInfo, mergeState);
+
+        if (mergeRunnable != null) mergeRunnable.run();
```
VijayanB, line 144: nit
```suggestion
        if (mergeRunnable != null) {
               mergeRunnable.run();
        }
```

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss1040ScalarQuantizedKnnVectorsWriter.java
```diff
@@ -158,6 +161,7 @@ public void mergeOneField(FieldInfo fieldInfo, MergeState mergeState) throws IOE
         } finally {
             IOUtils.close(flatVectorsReader);
         }
+        return null;
```
VijayanB, line 164: And here as well, and we don't need to implement IORunnable?

src/main/java/org/opensearch/knn/index/codec/KNN990Codec/NativeEngines990KnnVectorsWriter.java
```diff
@@ -100,11 +101,14 @@ public void flush(int maxDoc, final Sorter.DocMap sortMap) throws IOException {
...
-    public void mergeOneField(final FieldInfo fieldInfo, final MergeState mergeState) throws IOException {
+    public IORunnable mergeOneField(final FieldInfo fieldInfo, final MergeState mergeState) throws IOException {
         // This will ensure that we are merging the FlatIndex during force merge.
-        flatVectorsWriter.mergeOneField(fieldInfo, mergeState);
+        IORunnable mergeRunnable = flatVectorsWriter.mergeOneField(fieldInfo, mergeState);
+
+        if (mergeRunnable != null) mergeRunnable.run();
 
         doMergeOneField(fieldInfo, mergeState, this::train, approximateThreshold, segmentWriteState, nativeIndexBuildStrategyFactory, null);
+        return null;
```
VijayanB, line 111: same as above

src/main/java/org/opensearch/knn/index/query/RNNQueryFactory.java
```diff
@@ -180,13 +179,7 @@ private static Query getFloatVectorSimilarityQuery(
...
         final Query filterQuery
     ) {
-        return new FloatVectorSimilarityQuery(
-            fieldName,
-            floatVector,
-            DEFAULT_LUCENE_RADIAL_SEARCH_TRAVERSAL_SIMILARITY_RATIO * resultSimilarity,
-            resultSimilarity,
-            filterQuery
-        );
+        return new FloatVectorSimilarityQuery(fieldName, floatVector, resultSimilarity, 1.0f, filterQuery);
```
VijayanB, line 182: Do we need to run experiments to choose decay value to keep this backward compatible with respect to performance? IIUC, 1.0 gives good recall but may have higher latency?

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss1040ScalarQuantizedKnnVectorsWriter.java
VijayanB: nvm. i see it is checked

build-tools/repositories.gradle
```diff
@@ -5,8 +5,8 @@
 
 repositories {
     mavenLocal()
-    maven { url "https://ci.opensearch.org/ci/dbc/snapshots/maven/" }
     maven { url "https://ci.opensearch.org/maven2/" }
+    maven { url "https://ci.opensearch.org/ci/dbc/snapshots/maven/" }
```
navneet1v, line 9: any reason for moving it below?

src/main/java/org/apache/lucene/backward_codecs/lucene99/Lucene99ScalarQuantizedVectorsWriter.java
```diff
@@ -244,18 +248,19 @@ public void mergeOneField(FieldInfo fieldInfo, MergeState mergeState) throws IOE
...
+            IORunnable rawMerge = rawVectorDelegate.mergeOneField(fieldInfo, mergeState);
+            if (rawMerge != null) rawMerge.run();
             ScalarQuantizer mergedQuantizationState = mergeAndRecalculateQuantiles(mergeState, fieldInfo, confidenceInterval, bits);
             return mergeOneFieldToIndex(segmentWriteState, fieldInfo, mergeState, mergedQuantizationState);
         }
         // We only merge the delegate, since the field type isn't float32, quantization wasn't
         // supported, so bypass it.
-        return rawVectorDelegate.mergeOneFieldToIndex(fieldInfo, mergeState);
+        // In Lucene 10.5, mergeOneFieldToIndex is removed. Just return null for non-float32.
+        return null;
```
navneet1v, line 263: why null here?

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss1040ScalarQuantizedKnnVectorsWriter.java
kotwanikunal: `KnnVectorsWriter.merge()` is final in 10.5. It calls `mergeOneField()` null-checks the return, and collects non-null runnables for a deferred execution pass. Returning null signals work was done inline. Our writers do flat merge + native index build synchronously, so null is correct. Same execution flow as pre-10.5.

src/main/java/org/opensearch/knn/index/query/RNNQueryFactory.java
kotwanikunal: The old `traversalSimilarity` was removed in Lucene 10.5 and replaced with an adaptive decay parameter. The two mechanisms are fundamentally different so there is no exact equivalent. decay=1.0 gives widest traversal which preserves recall but may have latency implications vs the old behavior. Default decay (0.5) fails existing tests with reduced recall, so that is not viable as-is.

For now 1.0 unblocks the build. I will open a follow-up issue to benchmark decay values and find the right tradeoff between recall and latency for our radial search path.

kotwanikunal: Created tracking: https://github.com/opensearch-project/k-NN/issues/3413

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss1040ScalarQuantizedKnnVectorsWriter.java
navneet1v: @kotwanikunal what the reason to do it async? and how does that help?

src/main/java/org/apache/lucene/backward_codecs/lucene99/Lucene99ScalarQuantizedVectorsWriter.java
navneet1v: nvm I saw the lucene code.

navneet1v (approved):

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss1040ScalarQuantizedKnnVectorsWriter.java
kotwanikunal: This comes in from https://github.com/apache/lucene/pull/15732. From what I can understand, the two stage process helps avoid redundant writes during flat vector creation especially with multiple field types under a segment. Our architecture with custom codecs is different - we have a separate file for each field type - the runnable would not be as helpful in our case.

build-tools/repositories.gradle
kotwanikunal: The order has some throttling issues still. Cherry picking from https://github.com/opensearch-project/k-NN/pull/3406
It is affecting builds otherwise. This PR completely unblocks main build issues.

Issue #3084, opened 2026-01-26
[BUG] null_pointer_exception on knn query with rescore and explain (engine=lucene)


telendt:
### Describe the bug

I get a `null_pointer_exception` on knn query on knn_vector with rescore and explain when Lucene engine is used.

The NPE doesn't happen when (either one):
- the query doesn't ask for an explanation (no `explain=true`)
- no rescore is used
- engine other that Lucene (e.g. Faiss) is used for index construction

### Related component

Search

### To Reproduce

1. Create a test index:
```
PUT test
{
  "settings": {
    "index": {
      "knn": true
    }
  },
  "mappings": {
    "properties": {
      "embedding": {
        "type": "knn_vector",
        "dimension": 2,
        "method": {
          "name": "hnsw",
          "engine": "lucene"
        }
      }
    }
  }
}
```
2. Index a document
```
PUT test/_doc/1?refresh=true
{
  "embedding": [1, 2]
}
```
3. Try to perform **knn** search with **rescore** and **explain**
```
GET test/_search?explain=true
{
  "query": {
    "knn": {
      "embedding": {
        "vector": [1,2],
        "k": 1,
        "rescore": {
          "oversample_factor": 3
        }
      }
    }
  }
}
```
4. Observe the error:
```
{
  "error": {
    "root_cause": [
      {
        "type": "null_pointer_exception",
        "reason": "Cannot invoke \"org.opensearch.knn.index.query.KNNWeight.explain(org.apache.lucene.index.LeafReaderContext, int, float)\" because \"this.this$0.knnWeight\" is null"
      }
    ],
    "type": "search_phase_execution_exception",
    "reason": "all shards failed",
    "phase": "query",
    "grouped": true,
    "failed_shards": [
      {
        "shard": 0,
        "index": "test",
        "node": "0-TLQAxdS6Sbon3RR-Rkyg",
        "reason": {
          "type": "null_pointer_exception",
          "reason": "Cannot invoke \"org.opensearch.knn.index.query.KNNWeight.explain(org.apache.lucene.index.LeafReaderContext, int, float)\" because \"this.this$0.knnWeight\" is null"
        }
      }
    ],
    "caused_by": {
      "type": "null_pointer_exception",
      "reason": "Cannot invoke \"org.opensearch.knn.index.query.KNNWeight.explain(org.apache.lucene.index.LeafReaderContext, int, float)\" because \"this.this$0.knnWeight\" is null",
      "caused_by": {
[...]

PR #3412, opened 2026-07-02
Fix NPE on k-NN explain with rescore

MdTanwer:
## Summary
- Fixes a `null_pointer_exception` when running a k-NN search with `rescore` and `explain=true` on Lucene engine indices ([#3084](https://github.com/opensearch-project/k-NN/issues/3084))
- `DocAndScoreQuery` created during shard-level rescore does not carry a `KNNWeight`, but `explain()` unconditionally delegated to `knnWeight.explain()`
- Return a basic score explanation when `knnWeight` is null instead of throwing

### DIFF
--- a/src/main/java/org/opensearch/knn/index/query/common/DocAndScoreQuery.java
+++ b/src/main/java/org/opensearch/knn/index/query/common/DocAndScoreQuery.java
@@ -67,6 +67,10 @@ final class DocAndScoreQuery extends Query {
                     throw new RuntimeException(e);
                 }
 
+                if (knnWeight == null) {
+                    return Explanation.match(score, "k-NN score");
+                }
+
                 return knnWeight.explain(context, doc, score);
             }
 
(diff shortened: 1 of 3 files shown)


src/main/java/org/opensearch/knn/index/query/common/DocAndScoreQuery.java
```diff
@@ -67,6 +67,10 @@ public Explanation explain(LeafReaderContext context, int doc) {
                     throw new RuntimeException(e);
                 }
 
+                if (knnWeight == null) {
+                    return Explanation.match(score, "k-NN score");
+                }
```
navneet1v, line 72: I don't think this is the right solution. We should give a valid response when rescore is true with Lucene query, similar to what we do in the Faiss engine.

navneet1v (changes requested):
Thank you for raising the PR. But the fix is not a correct fix. Lets do a proper fix for this rather than returning a dummy explanation.

Also we should add ITs for this change similar to what we have other other Engines. See ExplainIT class on how to add the ITs for explain usecase.

PR #3416, opened 2026-07-06
Parameterize tests based on compression level for 32x defaults

kotwanikunal:
### Description

Introduces compression-aware test parameterization across all integration tests to ensure test stability after the 32x default compression flip ([RFC #3290](https://github.com/opensearch-project/k-NN/issues/3290)).

Every IT class that creates a float HNSW index without explicit compression will silently resolve to 32x (SQ 1-bit) after the default flip. Tests with exact-score assertions would break. This PR adds a `KNNCompressionRestTestCase` base class that parameterizes tests by compression level, so each test class explicitly declares which compression configurations it supports and gets config-appropriate recall thresholds.

**Infrastructure:**
- `KNNCompressionRestTestCase`: base class with `@ParametersFactory`, index creation helpers that inject compression from config, and recall assertion helpers
- `CompressionTestConfig`: enum carrying per-space-type recall thresholds, mode, and eligibility flags

### Related Issues
Related #3290

### Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 * Add rescoring phase after radial search on quantized index [#3347](https://github.com/opensearch-project/k-NN/pull/3347)
 * Add base64 encoded vector indexing support for knn_vector fields [#3350](https://github.com/opensearch-project/k-NN/pull/3350)
 * Introduce system-generated search pipeline processor to automatically exclude knn_vector fields from _source in search responses [#3152](https://github.com/opensearch-project/k-NN/pull/3152)
+* Parameterize integration test framework for compression level [#3416](https://github.com/opensearch-project/k-NN/pull/3416)
 
 ### Maintenance
 * Upgrade Lucene to 10.5.0 [#3411](https://github.com/opensearch-project/k-NN/pull/3411)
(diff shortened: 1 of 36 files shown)


Vikasht34 (approved):
LGTM !!

navneet1v (commented):
@kotwanikunal when test fails with this new parameterization can we re-run the tests with same seed?

navneet1v (approved):
Overall looks good. Just 1 clarifying question.

shatejas (commented):
On initial look we shouldn't be adding another layer in the hierarchy of polymorphism to resuse methods IMO. Its hard to maintain overall with tests, one change can affect multiple tests

Vikasht34:
> On initial look we shouldn't be adding another layer in the hierarchy of polymorphism to resuse methods IMO. Its hard to maintain overall with tests, one change can affect multiple tests

@shatejas Could you help me understand the concern here? I believe this approach aligns with standard polymorphism patterns — whether through interface-based or abstract class-based polymorphism.

I acknowledge the initial changeset touches multiple classes, but this is expected when introducing polymorphic behavior into an existing hierarchy. The upfront refactoring cost should pay off in extensibility and 
reduced coupling going forward.

Happy to discuss further if there are specific maintainability or design pattern I'm missing.

kotwanikunal:
> @kotwanikunal when test fails with this new parameterization can we re-run the tests with same seed?

Verified this and it works with the reproduce command.
Logged the random seed with 2 different seeds - across multiple runs - the seeds were consistent with the one provided to the runner.

kotwanikunal:
> On initial look we shouldn't be adding another layer in the hierarchy of polymorphism to resuse methods IMO. Its hard to maintain overall with tests, one change can affect multiple tests

I thought through the same problem when I was designing it and had a similar concern. The base class is intentionally minimal right now, just the parameterization plumbing needed for the compression flip (factory + field helper + thresholds). No shared test logic.

If we need engine or mode parameterization later, this evolves into a generic `KNNParameterizedRestTestCase` with selectable axes. Keeping it scoped to the immediate problem for now.

PR #3421, opened 2026-07-10
Wire up ResolvedIndexSpec usage across classes

kotwanikunal:
### Description
Wire ResolvedIndexSpec consumers through spec-driven resolution flow

Wires codec format selection, memory-optimized search, rescore context, radial search validation, remote index build, and field attribute mapping through ResolvedIndexSpec instead of independently inspecting method parameters at each call site.

Key changes:
- KNNQueryBuilder resolves rescore and radial validation via ResolvedIndexSpec
- FaissCodecFormatResolver and EngineFieldMapper use ResolvedIndexSpec for SQ 1-bit routing
- RemoteIndexBuildStrategy uses ResolvedIndexSpec for data type and skip-storage decisions
- Added engine-agnostic isSQOneBit() for radial search and MOS gating

### Related Issues
Related #3290 

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/codec/KNN1040BasePerFieldKnnVectorsFormat.java
+++ b/src/main/java/org/opensearch/knn/index/codec/KNN1040BasePerFieldKnnVectorsFormat.java
@@ -13,6 +13,7 @@ import org.opensearch.knn.index.codec.nativeindex.NativeIndexBuildStrategyFactor
 import org.opensearch.knn.index.engine.CodecFormatResolver;
 import org.opensearch.knn.index.engine.KNNEngine;
 import org.opensearch.knn.index.engine.KNNMethodContext;
+import org.opensearch.knn.index.engine.ResolvedIndexSpec;
 import org.opensearch.knn.index.mapper.KNNMappingConfig;
 import org.opensearch.knn.index.mapper.KNNVectorFieldType;
 
@@ -90,13 +91,14 @@ public abstract class KNN1040BasePerFieldKnnVectorsFormat extends PerFieldKnnVec
         nativeIndexBuildStrategyFactory.setKnnLibraryIndexingContext(knnMappingConfig.getKnnLibraryIndexingContext());
         final KNNEngine engine = knnMethodContext.getKnnEngine();
         final Map<String, Object> params = knnMethodContext.getMethodComponentContext().getParameters();
+        final ResolvedIndexSpec resolvedSpec = mappedFieldType.getResolvedSpec();
 
         if (engine == KNNEngine.LUCENE) {
-            return luceneFormatResolver.resolve(field, knnMethodContext, params, defaultMaxConnections, defaultBeamWidth);
+            return luceneFormatResolver.resolve(field, knnMethodContext, params, defaultMaxConnections, defaultBeamWidth, resolvedSpec);
         }
 
-        // Native engines — pass params so the resolver can detect SQ encoder
-        return nativeFormatResolver.resolve(field, knnMethodContext, params, defaultMaxConnections, defaultBeamWidth);
+        // Native engines — pass params so the resolver can detect SQ encoder when the spec is null
+        return nativeFormatResolver.resolve(field, knnMethodContext, params, defaultMaxConnections, defaultBeamWidth, resolvedSpec);
     }
 
     @Override
--- a/src/main/java/org/opensearch/knn/index/engine/ResolvedIndexSpec.java
+++ b/src/main/java/org/opensearch/knn/index/engine/ResolvedIndexSpec.java
@@ -36,16 +36,11 @@ public final class ResolvedIndexSpec {
     private final int dimension;
     private final Version indexVersionCreated;
 
-    /** Faiss-specific: routes field to Faiss1040ScalarQuantizedKnnVectorsFormat. Temporary -- to be replaced by generalized codec format resolver. */
-    public boolean usesFaissSQ1BitCodecFormat() {
-        return engine == KNNEngine.FAISS && isSQOneBit();
-    }
-
     /**
      * Whether this configuration always uses memory optimized search.
      */
     public boolean alwaysUseMemoryOptimizedSearch() {
-        return engine == KNNEngine.FAISS && isSQOneBit();
+        return isSQOneBit();
     }
 
     /**
@@ -80,7 +75,6 @@ public final class ResolvedIndexSpec {
         if (encoderType == Encoder.EncoderType.BQ) {
             return false;
         }
-        // Among quantized indices, only flat method or SQ 1-bit supports radial
         if (isQuantizedIndex()) {
             return isMethodFlat() || isSQOneBit();
         }
@@ -109,8 +103,26 @@ public final class ResolvedIndexSpec {
         return getRescoreContext() != null;
     }
 
-    private boolean isSQOneBit() {
+    /**
+     * Whether this is an SQ encoder producing fp16 quantized output. Distinct from native halfFloat data type storage.
+     * Equivalent to FaissHNSWMethod.isFloat16Index:
+     * encoder is SQ, dataType is FLOAT, and bits is SIXTEEN or legacy (null/FULL_PRECISION defaults to fp16).
+     */
+    public boolean isFP16QuantizedIndex() {
+        // FULL_PRECISION represents the legacy path where SQ encoder is specified without explicit bits param.
+        // buildResolvedIndexSpec defaults to FULL_PRECISION in this case, and the old FaissHNSWMethod.isFloat16Index()
+        // treated SQ-with-no-bits as fp16 by convention.
+        return vectorDataType == VectorDataType.FLOAT
+            && encoderType == Encoder.EncoderType.SQ
+            && (quantizationBits == Encoder.QuantizationBits.SIXTEEN || quantizationBits == Encoder.QuantizationBits.FULL_PRECISION);
+    }
+
+    public boolean isSQOneBit() {
         return encoderType == Encoder.EncoderType.SQ && quantizationBits == Encoder.QuantizationBits.ONE;
+    }
+
+    public boolean isFaissSQOneBit() {
+        return engine == KNNEngine.FAISS && isSQOneBit();
     }
 
     private boolean isMethodFlat() {
--- a/src/main/java/org/opensearch/knn/index/engine/faiss/FaissCodecFormatResolver.java
+++ b/src/main/java/org/opensearch/knn/index/engine/faiss/FaissCodecFormatResolver.java
@@ -6,6 +6,7 @@
 package org.opensearch.knn.index.engine.faiss;
 
 import org.apache.lucene.codecs.KnnVectorsFormat;
+import org.opensearch.common.Nullable;
 import org.opensearch.index.IndexSettings;
 import org.opensearch.index.mapper.MapperService;
 import org.opensearch.knn.index.KNNSettings;
@@ -14,6 +15,7 @@ import org.opensearch.knn.index.codec.KNN990Codec.NativeEngines990KnnVectorsForm
 import org.opensearch.knn.index.codec.nativeindex.NativeIndexBuildStrategyFactory;
 import org.opensearch.knn.index.engine.CodecFormatResolver;
 import org.opensearch.knn.index.engine.KNNMethodContext;
+import org.opensearch.knn.index.engine.ResolvedIndexSpec;
 
 import java.util.Map;
 import java.util.Optional;
@@ -41,7 +43,8 @@ public class FaissCodecFormatResolver implements CodecFormatResolver {
 
     /**
      * Resolves the format for a specific field. Returns {@link Faiss1040ScalarQuantizedKnnVectorsFormat} when
-     * the encoder is sq with bits=1, otherwise falls back to the default native format.
+     * the encoder is sq with bits=1, otherwise falls back to the default native format. Prefers the resolved
+     * index spec when non-null; otherwise inspects the method component parameters.
      */
     @Override
     public KnnVectorsFormat resolve(
@@ -49,9 +52,11 @@ public class FaissCodecFormatResolver implements CodecFormatResolver {
         KNNMethodContext methodContext,
         Map<String, Object> params,
         int defaultMaxConnections,
-        int defaultBeamWidth
+        int defaultBeamWidth,
+        @Nullable ResolvedIndexSpec resolvedSpec
     ) {
-        if (isSQOneBitEncoder(params)) {
+        final boolean isSQOneBit = resolvedSpec != null ? resolvedSpec.isFaissSQOneBit() : isSQOneBitEncoder(params);
+        if (isSQOneBit) {
             return new Faiss1040ScalarQuantizedKnnVectorsFormat(nativeIndexBuildStrategyFactory);
         }
         return resolve();
--- a/src/main/java/org/opensearch/knn/index/engine/lucene/LuceneCodecFormatResolver.java
+++ b/src/main/java/org/opensearch/knn/index/engine/lucene/LuceneCodecFormatResolver.java
@@ -7,10 +7,12 @@ package org.opensearch.knn.index.engine.lucene;
 
 import lombok.extern.log4j.Log4j2;
 import org.apache.lucene.codecs.KnnVectorsFormat;
+import org.opensearch.common.Nullable;
 import org.opensearch.knn.index.codec.KnnVectorsFormatContext;
 import org.opensearch.knn.index.codec.LuceneVectorsFormatType;
 import org.opensearch.knn.index.codec.params.KNNScalarQuantizedVectorsFormatParams;
 import org.opensearch.knn.index.engine.CodecFormatResolver;
+import org.opensearch.knn.index.engine.ResolvedIndexSpec;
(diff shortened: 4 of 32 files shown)


src/main/java/org/opensearch/knn/index/engine/ResolvedIndexSpec.java
```diff
@@ -109,10 +108,24 @@ public boolean requiresRescore() {
         return getRescoreContext() != null;
     }
 
-    private boolean isSQOneBit() {
+    /**
+     * Whether this is an fp16 index. Equivalent to FaissHNSWMethod.isFloat16Index:
+     * encoder is SQ, dataType is FLOAT, and bits is SIXTEEN or legacy (null/FULL_PRECISION defaults to fp16).
+     */
+    public boolean isFloat16Index() {
```
Vikasht34: Could you clarify this , FULL_PRECISION here means bits was null/not specified which defaults to fp16 in the old path. But QuantizationBits.fromValue(32) == FULL_PRECISION would also match, which is wrong — a 32-bit SQ is not fp16. However, if somehow quantizationBits = FULL_PRECISION AND encoderType = SQ for a non-fp16 reason, this would be a false positive

navneet1v:
> * Added engine-agnostic isSQOneBit() for radial search and MOS gating

thinking when can we get of this function. :D

src/main/java/org/opensearch/knn/index/codec/KNN1040BasePerFieldKnnVectorsFormat.java
```diff
@@ -90,12 +91,19 @@ public KnnVectorsFormat getKnnVectorsFormatForField(final String field) {
         nativeIndexBuildStrategyFactory.setKnnLibraryIndexingContext(knnMappingConfig.getKnnLibraryIndexingContext());
         final KNNEngine engine = knnMethodContext.getKnnEngine();
         final Map<String, Object> params = knnMethodContext.getMethodComponentContext().getParameters();
+        final ResolvedIndexSpec resolvedSpec = mappedFieldType.getResolvedSpec();
 
         if (engine == KNNEngine.LUCENE) {
+            if (resolvedSpec != null) {
+                return luceneFormatResolver.resolve(field, knnMethodContext, params, defaultMaxConnections, defaultBeamWidth, resolvedSpec);
+            }
```
navneet1v: in what cases this will be null? I think in some PR we discussed that we should never return ResolveIndexSpec as null. Can you help me remind what decision we took over there.

src/main/java/org/opensearch/knn/index/engine/faiss/FaissCodecFormatResolver.java
```diff
@@ -57,6 +59,21 @@ public KnnVectorsFormat resolve(
...
+        Map<String, Object> params,
+        int defaultMaxConnections,
+        int defaultBeamWidth,
+        @NonNull ResolvedIndexSpec resolvedSpec
+    ) {
+        if (resolvedSpec.usesFaissSQ1BitCodecFormat()) {
+            return new Faiss1040ScalarQuantizedKnnVectorsFormat(nativeIndexBuildStrategyFactory);
+        }
+        return resolve();
+    }
```
navneet1v, line 63: can we not avoid this function and just use one single resolve function which has all the details on how to resolve the KNNVectorFormat

src/main/java/org/opensearch/knn/index/engine/lucene/LuceneCodecFormatResolver.java
```diff
@@ -63,6 +64,21 @@ public KnnVectorsFormat resolve(
         return factory.apply(new KnnVectorsFormatContext(field, methodContext, params, defaultMaxConnections, defaultBeamWidth));
     }
 
+    /**
+     * Resolve with ResolvedIndexSpec available for future use.
+     * Currently delegates to the standard resolve method.
+     */
+    public KnnVectorsFormat resolve(
```
navneet1v: same as above

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
```diff
@@ -478,7 +479,10 @@ protected Query doToQuery(QueryShardContext context) {
         }
 
         if (this.maxDistance != null || this.minScore != null) {
-            knnVectorFieldType.validateSupportRadialSearch(knnEngine);
+            ResolvedIndexSpec spec = knnVectorFieldType.getResolvedSpec();
+            if (spec != null && !spec.supportsRadialSearch()) {
```
Vikasht34: When spec == null (model-based indices), validation is skipped entirely. The old knnVectorFieldType.validateSupportRadialSearch(knnEngine) ran unconditionally — it checked engine support, binary datatype, and BQ quantization config for model-based indices too.

This means a quantized model-based index can now do radial search without any check silent regression from the old behavior?

src/main/java/org/opensearch/knn/index/engine/ResolvedIndexSpec.java
navneet1v: shouldn't this be is2xCompressedIndex or isFp16QuantizedIndex? Since we are thinking to add a new datatype called as halfFloat. So that will also be a fp16index

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldType.java
```diff
@@ -116,16 +116,12 @@ public KNNVectorFieldType(
...
-            knnMappingConfig.getKnnMethodContext()
-        );
-        this.memoryOptimizedSearchAvailable = MemoryOptimizedSearchSupportSpec.isSupportedFieldType(
-            knnMappingConfig.getKnnMethodContext(),
-            annConfig.getQuantizationConfig(),
-            annConfig.getModelId()
-        );
-        this.indexCreatedVersion = indexCreatedVersion;
         this.resolvedSpec = resolvedSpec;
+        if (resolvedSpec != null) {
```
Vikasht34, line 120: When resolvedSpec == null, alwaysUseMemoryOptimizedSearch and memoryOptimizedSearchAvailable stay at their field-level default (false). The old constructor called MemoryOptimizedSearchSupportSpec.isAlwaysUseMemoryOptimizedSearch() and isSupportedFieldType() unconditionally to set these.
Model-based indices that pass null spec now silently lose MOS eligibility detection. Should the else-branch fall back to the old MemoryOptimizedSearchSupportSpec calls?

src/main/java/org/opensearch/knn/index/engine/ResolvedIndexSpec.java
```diff
@@ -109,10 +108,24 @@ public boolean requiresRescore() {
...
+            && (quantizationBits == Encoder.QuantizationBits.SIXTEEN || quantizationBits == Encoder.QuantizationBits.FULL_PRECISION);
+    }
+
+    public boolean isSQOneBit() {
         return encoderType == Encoder.EncoderType.SQ && quantizationBits == Encoder.QuantizationBits.ONE;
     }
 
+    public boolean isFaissSQOneBit() {
+        return engine == KNNEngine.FAISS && encoderType == Encoder.EncoderType.SQ && quantizationBits == Encoder.QuantizationBits.ONE;
+    }
```
navneet1v, line 126: any reason for having both of them? 

and please use this simplification

`engine == KNNEngine.FAISS && isSQOneBit()`

navneet1v (commented):
completed 50% of the PR. Will review another 505 by EOD.

src/main/java/org/opensearch/knn/index/codec/KNN1040BasePerFieldKnnVectorsFormat.java
kotwanikunal: Null when the field is model-based where they don't go through AbstractKNNMethod.buildResolvedIndexSpec(). We discussed making it non-null everywhere but it will need a follow-up since model-based fields require lazy initialization. The next PR will have those changes.

src/main/java/org/opensearch/knn/index/engine/lucene/LuceneCodecFormatResolver.java
kotwanikunal: Agreed. The overload with ResolvedIndexSpec currently just delegates to the original for Lucene, and adds one SQ 1-bit check for Faiss. We can collapse to a single method that takes an @Nullable ResolvedIndexSpec and uses it when available, falling back to param inspection when null. Will consolidate for both.

src/main/java/org/opensearch/knn/index/engine/ResolvedIndexSpec.java
kotwanikunal: `FULL_PRECISION` here represents the legacy path where SQ encoder is specified without an explicit bits parameter. In that case, `buildResolvedIndexSpec` defaults to `FULL_PRECISION` (the 32 value), and the old `FaissHNSWMethod.isFloat16Index()` treated this as fp16 (SQ encoder with no bits = fp16 by convention). This PR preserves that exact behavior.

A 32-bit SQ with explicitly set bits=32 is not a valid user configuration today. The method resolver rejects it during validation.

kotwanikunal: Fair point. The method currently checks vector data type, encoder and bits. This identifies the fp16 codec path, not a general "2x compression" concept. With a future halfFloat data type, that would be a different code path.

Renaming to `isFP16QuantizedIndex()` is more precise in this case. Updating it.

src/main/java/org/opensearch/knn/index/engine/ResolvedIndexSpec.java
kotwanikunal: Yep. Missed updating it with the new method.
This would ideally sit best as a typed structure - i'll think through it as a follow up once we have the flow all routed through RIS.

src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldType.java
kotwanikunal: Added the else-branch that falls back to the original MemoryOptimizedSearchSupportSpec calls when resolvedSpec is null. Model-based fields retain MOS eligibility detection through the legacy path. Marked with a TODO for removal once all field mapper paths supply a ResolvedIndexSpec.

Ties in with https://github.com/opensearch-project/k-NN/pull/3421/changes#r3590011674

src/main/java/org/opensearch/knn/index/codec/KNN1040BasePerFieldKnnVectorsFormat.java
navneet1v: make sense. thanks for providing the info.

src/main/java/org/opensearch/knn/index/engine/ResolvedIndexSpec.java
navneet1v: make sense.

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
```diff
@@ -478,7 +479,15 @@ protected Query doToQuery(QueryShardContext context) {
         }
 
         if (this.maxDistance != null || this.minScore != null) {
-            knnVectorFieldType.validateSupportRadialSearch(knnEngine);
+            ResolvedIndexSpec spec = knnVectorFieldType.getResolvedSpec();
+            if (spec != null) {
+                if (!spec.supportsRadialSearch()) {
+                    throw new UnsupportedOperationException("Radial search is not supported for indices which have quantization enabled");
```
navneet1v: I think you mean to say here, binary quantization doesn't support radial search, rather than quantization enabled doesn't support radial search

src/test/java/org/opensearch/knn/index/engine/ResolvedIndexSpecTests.java
```diff
@@ -210,6 +210,63 @@ public void testGetRescoreContext_NullVersionDefaults() {
...
+    public void testAlwaysUseMemoryOptimizedSearch_FaissSQ1Bit() {
+        ResolvedIndexSpec spec = baseFaissSQ1Bit().build();
+        assertTrue(spec.alwaysUseMemoryOptimizedSearch());
+    }
+
+    public void testAlwaysUseMemoryOptimizedSearch_LuceneSQ1Bit() {
+        ResolvedIndexSpec spec = baseFaissSQ1Bit().engine(KNNEngine.LUCENE).build();
+        assertTrue(spec.alwaysUseMemoryOptimizedSearch());
+    }
+
```
navneet1v: shouldn't this be `testNotAlwaysUseMemoryOptimizedSearch_FaissSQ16Bit`

src/test/java/org/opensearch/knn/index/engine/ResolvedIndexSpecTests.java
```diff
@@ -210,6 +210,63 @@ public void testGetRescoreContext_NullVersionDefaults() {
...
+    public void testAlwaysUseMemoryOptimizedSearch_FaissSQ16Bit() {
+        ResolvedIndexSpec spec = baseFaiss().encoderType(Encoder.EncoderType.SQ)
+            .quantizationBits(Encoder.QuantizationBits.SIXTEEN)
+            .compressionLevel(CompressionLevel.x2)
+            .build();
+        assertFalse(spec.alwaysUseMemoryOptimizedSearch());
+    }
+
+    public void testAlwaysUseMemoryOptimizedSearch_FaissFlat() {
```
navneet1v: same as above. One recommendation I would suggest is rather than having these different test cases, can we not have a single or double test that can test different engines and compression level for this function `alawysUseMemoryOptizedSearch`

navneet1v (commented):
completed my review.

src/test/java/org/opensearch/knn/index/engine/ResolvedIndexSpecTests.java
kotwanikunal: Makes sense. Will consolidate into a parameterized style test that covers multiple engine/compression combos in one method/

src/test/java/org/opensearch/knn/index/engine/ResolvedIndexSpecTests.java
kotwanikunal: Ahh. Good catch. Will fix it.

src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java
kotwanikunal: The supportsRadialSearch gate blocks BQ, any quantized index (compression > x2) that is not SQ 1-bit. The current message is intentionally broader. I think I'll update it to "Radial search is not supported for this quantization configuration" to read cleaner

src/main/java/org/opensearch/knn/index/engine/ResolvedIndexSpec.java
```diff
@@ -36,16 +36,11 @@ public final class ResolvedIndexSpec {
...
-    public boolean usesFaissSQ1BitCodecFormat() {
-        return engine == KNNEngine.FAISS && isSQOneBit();
-    }
-
     /**
      * Whether this configuration always uses memory optimized search.
      */
     public boolean alwaysUseMemoryOptimizedSearch() {
-        return engine == KNNEngine.FAISS && isSQOneBit();
+        return isSQOneBit();
```
Vikasht34, line 43: alwaysUseMemoryOptimizedSearch() now returns true for Lucene SQ 1-bit (engine-agnostic), but isMemoryOptimizedEligible() returns false for Lucene (checks engine == FAISS). this is confusing where alwaysUseMemoryOptimizedSearch = true and memoryOptimizedSearchAvailable = false on the same field type

kotwanikunal: `MemoryOptimizedSearchSupportSpec.isSupportedFieldType()` checks `alwaysUseMemoryOptimizedSearch `first and short-circuits:

https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/engine/MemoryOptimizedSearchSupportSpec.java#L53-L55

When `alwaysUseMemoryOptimizedSearch=true`, `memoryOptimizedSearchAvailable` is never consulted. The flags have a ordering and are not to be used combined.

This PR does not change this logic. The pre RIS code already had `isAlwaysUseMemoryOptimizedSearch()` returning true for Lucene SQ 1-bit via `FaissSQEncoder.isSQOneBit(params)` - https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/engine/MemoryOptimizedSearchSupportSpec.java#L163-L166

What this PR does is centralize that same check onto `ResolvedIndexSpec.alwaysUseMemoryOptimizedSearch()` so it is computed once at resolution time rather than inspected from params at every query. No behavioral change, just centralized access.

kotwanikunal:
@navneet1v / @Vikasht34 - Can you please take a look? I have worked through your comments

Vikasht34:
> @navneet1v / @Vikasht34 - Can you please take a look? I have worked through your comments

Yes .. Will have look after SQ PRS

Vikasht34 (approved):
LGTM .. Thanks 
@kotwanikunal

navneet1v (approved):

PR #3424, opened 2026-07-12
Simplifiy byte vector handling in exact search by removing an unnecessary float-to-byte conversion step.

navneet1v:
### Description
Simplifies byte vector handling in exact search by removing an unnecessary float-to-byte conversion step.

##### Changes:
  - KNNQueryBuilder: Always populate byteQueryVector for BYTE data type regardless of engine (previously only set for Lucene/memory-optimized paths). Removes the isUsingLuceneQuery guard that left byteQueryVector null for FAISS/NMSLIB.
  - ExactSearcher: For BYTE vector data type, use context.getByteQueryVector() directly instead of reading floatQueryVector and converting it back to bytes. Eliminates redundant float→byte conversion on every segment
  - ExactSearcherTests: Added unit test verifying that exactSearchScorer works correctly when only byteQueryVector is provided on the context.

Apart from this:
1. Fix the flakiness in the 32x compression tests.
2. Removed some unused variables.

#### Motivation
This is more of an inconsistencies in ExactSearch. This will ensure that if we use ExactSearcher next time in other path we should not see any failures.

### Related Issues
NA

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/common/KNNConstants.java
+++ b/src/main/java/org/opensearch/knn/common/KNNConstants.java
@@ -42,7 +42,6 @@ public class KNNConstants {
     public static final String TOP_LEVEL_PARAMETER_SPACE_TYPE = METHOD_PARAMETER_SPACE_TYPE;
     public static final String TOP_LEVEL_PARAMETER_ENGINE = KNN_ENGINE;
     public static final String COMPOUND_EXTENSION = "c";
-    public static final String MODEL = "model";
     public static final String MODELS = "models";
     public static final String MODEL_ID = "model_id";
     public static final String MODEL_BLOB_PARAMETER = "model_blob";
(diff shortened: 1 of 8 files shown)


navneet1v:
> ## PR Code Suggestions ✨
> Explore these optional code suggestions:
> 
> **Category**	**Suggestion                                                                                                                                    **	**Impact**
> Possible issue	
> Guard against missing byte query vector
> **The change now relies on `context.getByteQueryVector()` being populated for BYTE vector data types, but previously it was derived from `getFloatQueryVector()`. If any caller path still only sets the float query vector (not the byte one) for BYTE data type, this will cause a NullPointerException in the scorer. Verify all callers populate `byteQueryVector` on the context, or fall back to converting from float when null.**
> 
> [src/main/java/org/opensearch/knn/index/query/exactsearch/ExactSearcher.java [340-348]](https://github.com/opensearch-project/k-NN/pull/3424/files#diff-506dbbeb9385167fbc3e3943521b79b918b294deda5764246804fbdd307a56e1R340-R348)
> 
> ```diff
>  if (VectorDataType.BYTE == vectorDataType) {
> +    byte[] byteQueryVector = context.getByteQueryVector();
> +    if (byteQueryVector == null && context.getFloatQueryVector() != null) {
> +        final float[] floatQueryVector = context.getFloatQueryVector();
> +        byteQueryVector = new byte[floatQueryVector.length];
> +        for (int i = 0; i < byteQueryVector.length; i++) {
> +            byteQueryVector[i] = (byte) floatQueryVector[i];
> +        }
> +    }
>      return VectorScorers.createScorer(
>          iteratorValues,
> -        context.getByteQueryVector(),
> +        byteQueryVector,
>          scorerMode,
>          spaceType,
>          fieldInfo,
>          context.getMatchedDocsIterator(),
> ```
> 
> Suggestion importance[1-10]: 5
> __
> 
> Why: The suggestion raises a valid concern about potential NPE if callers don't populate `byteQueryVector`, but it's primarily a "verify" type suggestion and the PR presumably updates all callers to populate the byte vector. The fallback conversion may mask real bugs.
> 
> Low

Validated that this code path is not getting used.

kotwanikunal (approved):

naveentatikonda (approved):

PR #3426, opened 2026-07-14
Validate inverted lists type in template index deserialization

shatejas:
When creating an index from a trained model template, validate that the deserialized faiss index only contains ArrayInvertedLists. Reject any other InvertedLists implementation (e.g. OnDiskInvertedLists, BlockInvertedLists) which could perform unintended filesystem operations during index construction.

The validation is applied to all three CreateIndexFromTemplate variants: float, byte, and binary.

### Check List

### DIFF
--- a/jni/src/faiss_wrapper.cpp
+++ b/jni/src/faiss_wrapper.cpp
@@ -27,6 +27,7 @@
 #include "commons.h"
 #include "faiss/IndexBinaryIVF.h"
 #include "faiss/IndexBinaryHNSW.h"
+#include "faiss/invlists/InvertedLists.h"
 
 #include <algorithm>
 #include <jni.h>
@@ -80,6 +81,51 @@ void InternalTrainIndex(faiss::Index * index, faiss::idx_t n, const float* x);
 
 // Train a binary index with data provided
 void InternalTrainBinaryIndex(faiss::IndexBinary * index, faiss::idx_t n, const uint8_t* x);
+
+// Validates that a deserialized template index is one of the types that the
+// k-NN train path legitimately produces. For IVF types, additionally validates
+// that inverted lists use only ArrayInvertedLists.
+//
+// Allowed types:
+//   IndexIVFFlat, IndexIVFPQ, IndexIVFScalarQuantizer (all are IndexIVF)
+//   IndexHNSWPQ
+static void validateTemplateIndex(faiss::Index* index) {
+    if (dynamic_cast<faiss::IndexIVF*>(index) != nullptr) {
+        auto* ivf = dynamic_cast<faiss::IndexIVF*>(index);
+        if (ivf->invlists != nullptr &&
+            dynamic_cast<faiss::ArrayInvertedLists*>(ivf->invlists) == nullptr) {
+            throw std::runtime_error(
+                "Template index contains unsupported inverted lists type. "
+                "Only ArrayInvertedLists is allowed.");
+        }
+        return;
+    }
+    if (auto* hnswpq = dynamic_cast<faiss::IndexHNSWPQ*>(index)) {
+        // With HNSWPQ no other storage type is supported
+        if (dynamic_cast<faiss::IndexPQ*>(hnswpq->storage) == nullptr) {
+            throw std::runtime_error(
+                "Template index HNSWPQ has unsupported storage type. "
+                "Only IndexPQ is allowed.");
+        }
+        return;
+    }
+    throw std::runtime_error(
+        "Template index is not a supported index type.");
+}
+
+static void validateTemplateIndex(faiss::IndexBinary* index) {
+    auto* ivf = dynamic_cast<faiss::IndexBinaryIVF*>(index);
+    if (ivf == nullptr) {
+        throw std::runtime_error(
+            "Template index is not a valid IVF index type.");
+    }
+    if (ivf->invlists != nullptr &&
+        dynamic_cast<faiss::ArrayInvertedLists*>(ivf->invlists) == nullptr) {
+        throw std::runtime_error(
+            "Template index contains unsupported inverted lists type. "
+            "Only ArrayInvertedLists is allowed.");
+    }
+}
 
 // Converts the int FilterIds to Faiss ids type array.
 void convertFilterIdsToFaissIdType(const int* filterIds, int filterIdsLength, faiss::idx_t* convertedFilterIds);
@@ -307,9 +353,13 @@ void knn_jni::faiss_wrapper::CreateIndexFromTemplate(knn_jni::JNIUtilInterface *
     }
     jniUtil->ReleaseByteArrayElements(env, templateIndexJ, indexBytesJ, JNI_ABORT);
 
-    // Create faiss index
+    // IO_FLAG_READ_ONLY is an added guard to make sure only on-heap inverted lists
+    // are written since k-NN only supports ArrayInvertedLists. This does not affect
+    // ArrayInvertedLists which has no read_only check and writes to in-memory
+    // std::vectors unconditionally.
     std::unique_ptr<faiss::Index> indexWriter;
-    indexWriter.reset(faiss::read_index(&vectorIoReader, 0));
+    indexWriter.reset(faiss::read_index(&vectorIoReader, faiss::IO_FLAG_READ_ONLY));
+    validateTemplateIndex(indexWriter.get());
 
     auto idVector = jniUtil->ConvertJavaIntArrayToCppIntVector(env, idsJ);
     faiss::IndexIDMap idMap =  faiss::IndexIDMap(indexWriter.get());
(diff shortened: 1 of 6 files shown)


kotwanikunal (dismissed):

jni/src/faiss_wrapper.cpp
```diff
@@ -81,6 +82,22 @@ void InternalTrainIndex(faiss::Index * index, faiss::idx_t n, const float* x);
 // Train a binary index with data provided
 void InternalTrainBinaryIndex(faiss::IndexBinary * index, faiss::idx_t n, const uint8_t* x);
 
+// Validates that a deserialized template index only uses ArrayInvertedLists.
+// Rejects any other InvertedLists type (e.g. OnDiskInvertedLists) that could
+// perform unsafe filesystem operations.
+template <typename IndexType, typename IVFType>
+static void validateTemplateIndex(IndexType* index) {
```
Vikasht34: One thing about validateTemplateIndex — I think it only covers the case where the deserialized index is itself an IVF, and misses the case where the IVF is nested inside a wrapper index. The check does dynamic_cast<IVFType*>(index) on the top-level object and early-returns if that's null. But read_index builds these trees recursively, and several index types just hold an inner index rather than being an IVF themselves — IndexPreTransform, IndexIDMap, IndexRefine, IndexIVFIndependentQuantizer. For those, the top-level dynamic_cast<IndexIVF*> returns null, so we return early and never look at the nested index's invlists.

This matters because it's a shape we produce ourselves — e.g. an OPQ-encoded model (OPQ..,IVF..,PQ..) deserializes as an IndexPreTransform wrapping the IVF. So a template in that form skips the invlists check entirely, even though add_with_ids still flows down into the nested IVF and its inverted lists

So the validation holds for a flat top-level IVF (which the current tests cover) but not for the wrapped variants.

shatejas: added a throw for all those types

Vikasht34: Thanks , I see but with this legitimate trained HNSWPQ model would be rejected at build time.

Vikasht34: I see more egde cases filled in , Would u Cosnider this 

```
// Maximum index-tree depth we will traverse. Faiss index trees are shallow
// (wrappers nest at most a handful deep); this is a defensive bound against a
// pathologically nested template rather than an expected limit.
static constexpr int MAX_TEMPLATE_INDEX_DEPTH = 16;

// Forward declaration so the float and binary walkers can call each other
// (IndexBinaryFromFloat bridges a binary index to a float sub-index).
static void validateTemplateIndexTree(const faiss::IndexBinary* index, int depth);

// Recursively validates a deserialized template index. The only property we
// enforce is: every IVF node in the tree must use ArrayInvertedLists. Any other
// InvertedLists implementation (e.g. OnDiskInvertedLists, BlockInvertedLists)
// can trigger filesystem operations during index construction and is rejected.
//
// Non-IVF indices (HNSW, Flat, PQ, SQ, ...) have no inverted lists and are
// allowed; we still descend into their sub-indices because a wrapper or graph
// index can hold an IVF as a nested member (e.g. IndexHNSW2Level storage).
static void validateTemplateIndexTree(const faiss::Index* index, int depth) {
    if (index == nullptr) {
        return;
    }
    if (depth > MAX_TEMPLATE_INDEX_DEPTH) {
        throw std::runtime_error(
            "Template index nesting is too deep to validate.");
    }

    // IVF node: enforce ArrayInvertedLists, then descend into its quantizer.
    if (const auto* ivf = dynamic_cast<const faiss::IndexIVF*>(index)) {
        if (ivf->invlists != nullptr &&
            dynamic_cast<const faiss::ArrayInvertedLists*>(ivf->invlists) == nullptr) {
            throw std::runtime_error(
                "Template index contains unsupported inverted lists type. "
                "Only ArrayInvertedLists is allowed.");
        }
        validateTemplateIndexTree(ivf->quantizer, depth + 1);
        return;
    }

    // Wrapper / graph indices: descend into every sub-index they own.
    if (const auto* pt = dynamic_cast<const faiss::IndexPreTransform*>(index)) {
        validateTemplateIndexTree(pt->index, depth + 1);
        return;
    }
    if (const auto* idmap = dynamic_cast<const faiss::IndexIDMap*>(index)) {
        validateTemplateIndexTree(idmap->index, depth + 1);
        return;
    }
    if (const auto* refine = dynamic_cast<const faiss::IndexRefine*>(index)) {
        validateTemplateIndexTree(refine->base_index, depth + 1);
        validateTemplateIndexTree(refine->refine_index, depth + 1);
        return;
    }
    if (const auto* indep =
            dynamic_cast<const faiss::IndexIVFIndependentQuantizer*>(index)) {
        validateTemplateIndexTree(indep->quantizer, depth + 1);
        validateTemplateIndexTree(indep->index_ivf, depth + 1);
        return;
    }
    if (const auto* hnsw = dynamic_cast<const faiss::IndexHNSW*>(index)) {
        validateTemplateIndexTree(hnsw->storage, depth + 1);
        return;
    }

    // Leaf index with no inverted lists (Flat, PQ, SQ, ...): nothing to check.
}

// Binary index walker. Mirrors the float one; IndexBinaryFromFloat bridges
// back into the float tree.
static void validateTemplateIndexTree(const faiss::IndexBinary* index, int depth) {
    if (index == nullptr) {
        return;
    }
    if (depth > MAX_TEMPLATE_INDEX_DEPTH) {
        throw std::runtime_error(
            "Template index nesting is too deep to validate.");
    }

    if (const auto* ivf = dynamic_cast<const faiss::IndexBinaryIVF*>(index)) {
        if (ivf->invlists != nullptr &&
            dynamic_cast<const faiss::ArrayInvertedLists*>(ivf->invlists) == nullptr) {
            throw std::runtime_error(
                "Template index contains unsupported inverted lists type. "
                "Only ArrayInvertedLists is allowed.");
        }
        validateTemplateIndexTree(ivf->quantizer, depth + 1);
        return;
    }
    if (const auto* idmap = dynamic_cast<const faiss::IndexBinaryIDMap*>(index)) {
        validateTemplateIndexTree(idmap->index, depth + 1);
        return;
    }
    if (const auto* hnsw = dynamic_cast<const faiss::IndexBinaryHNSW*>(index)) {
        validateTemplateIndexTree(hnsw->storage, depth + 1);
        return;
    }
    if (const auto* fromFloat =
            dynamic_cast<const faiss::IndexBinaryFromFloat*>(index)) {
        validateTemplateIndexTree(fromFloat->index, depth + 1);
        return;
    }

    // Leaf binary index (e.g. IndexBinaryFlat): nothing to check.
}

// Public entry points used by the CreateIndexFromTemplate variants.
static void validateTemplateIndex(const faiss::Index* index) {
    validateTemplateIndexTree(index, 0);
}
static void validateTemplateIndex(const faiss::IndexBinary* index) {
    validateTemplateIndexTree(index, 0);
}
```

shatejas: This seems over-engineered for the use case. When k-NN trains the index it never produces IndexPreTransform, IndexRefine.. etc, if ever the support is added we can figure out what needs to be added. 

While creating a training index the user can only do IVF and HNSW, no other index type can be created and trained. So if the model is updated to have an index other than these knn can simply throw.

The approach I have is much more stricter - k-NN allows deserialization if the case is supported, if the model is wrapped in any other index k-NN simply throws

What am I missing here?

Vikasht34: yeah here is the thing , IndexHNSW is not a leaf , it owns a storage sub-index that read_index builds recursively (idxhnsw->storage = read_index(f, io_flags)), and IndexHNSW::add writes into it (storage->add(n, x)). faiss allows that storage to be an IVF — specifically IndexHNSW2Level (fourcc IHN2) requires its storage to be Index2Layer or IndexIVFPQ (index_read.cpp enforces exactly that). IndexIVFPQ is an IndexIVF, so it has invlists.

So the bypass is:

User submits a top-level IndexHNSW2Level. IndexHNSW2Level : IndexHNSW, so dynamic_cast<IndexHNSW*>(index) != nullptr → your check returns early and allows it.
Its storage is an IndexIVFPQ whose invlists fourcc is "ilod" (OnDiskInvertedLists) with an controlled filename.
The check never looks at storage, so the ilod is never seen.
add_with_ids → IndexHNSW::add → storage->add(...) → OnDiskInvertedLists::add_entries → update_totsize → fopen(filename, "w") + truncate + write.
No IndexPreTransform/IndexIDMap involved — the IVF hides in the HNSW's storage member, and top-level HNSW is exactly what you're allowing.

jni/src/faiss_wrapper.cpp
```diff
@@ -81,6 +83,44 @@ void InternalTrainIndex(faiss::Index * index, faiss::idx_t n, const float* x);
...
+    if (dynamic_cast<faiss::IndexFlatCodes*>(index) != nullptr) {
+        return;
+    }
+    throw std::runtime_error(
+        "Template index is not a supported index type.");
+}
+
+static void validateTemplateIndex(faiss::IndexBinary* index) {
+    auto* ivf = dynamic_cast<faiss::IndexBinaryIVF*>(index);
+    if (ivf == nullptr) {
```
0ctopus13prime, line 118: Curious, for template index, couldn't we have IndexBinaryHNSW as index? 🤔 
Hoping to understand HNSW index type check only exists for non-binary index.

jni/src/faiss_wrapper.cpp
shatejas: The solution is to simply do`dynamic_cast<const faiss::IndexHNSWPQ*>` if thats the case I am still not onboard with the approach of recursion and validate things which Opensearch does not support

jni/src/faiss_wrapper.cpp
shatejas: @0ctopus13prime binary encoder in hnsw does not require training so it will never use template index code path

0ctopus13prime: Sounds good

jni/src/faiss_wrapper.cpp
Vikasht34: Sounds Good !!

Vikasht34 (dismissed):
LGTM

0ctopus13prime (approved):

navneet1v (approved):

Issue #3427, opened 2026-07-15
[FEATURE] Support 2-bit and 4-bit scalar quantization (SQ)


naveentatikonda:
**Overview**
Extend the FAISS scalar-quantization (SQ) memory-optimized-search (MOS) path — which today ships only 1-bit quantization (SINGLE_BIT_QUERY_NIBBLE, x32) — to also support 2-bit (DIBIT_QUERY_NIBBLE, x16) and 4-bit (PACKED_NIBBLE, x8) quantization. All three encodings use a 4-bit quantized query.

The end state: on the HNSW method, users can pick the compression/recall trade-off that fits their workload by specifying encoder=sq, bits ∈ {1, 2, 4} explicitly, or by using mode=on_disk at compression=x32 / x16 / x8. Also, achieve full feature parity: remote index build (GPU/RVIB), method=flat, radial search, bulk-SIMD scoring on both the architectures, and every other feature that works on sq 1-bit today.

navneet1v:
@naveentatikonda can you share some benchmark results for this

naveentatikonda:
# SQ 2 bit and 4 bit Benchmarking Results

  | # | Config | Value |
  |---|---|---|
  | 1 | data node instance type | r7gd.2xl |
  | 2 | data node count | 2 |
  | 3 | primary shards | 5 |
  | 4 | replica | 1 |
  | 5 | Memory optimized search | TRUE |
  | 6 | GPU enabled | FALSE |
  | 7 | Dataset | Cohere-10M |

  ### Notes

  **Bulk SIMD-2 optimizations** (NEON, `arm_neon_simd_similarity_function.cpp`): reduces prefetch pressure in the B=2 (`simd4bitDibitDotProductBatch`) and B=4 (`simd4bitPackedNibbleDotProductBatch`) hot loops.
  Prefetch distance bumped from 16 bytes to 128 bytes (2 cache lines ahead, covering L2 miss latency), fired once per 64-byte cache-line boundary instead of every 16-byte iteration (~4× fewer prefetches), and
  software prefetches on the sequential query streams are dropped since the hardware prefetcher already handles them.

  **HNSW graph-build NEON SIMD (B=2 only)**, `symmetricDibitDp` in `jni/include/sq/faiss_sq_flat.h`: fuses the naïve 2×2 popcount-plane loop into a single 16-byte sweep that loads all four plane operands
[...]

PR #3428, opened 2026-07-15
[SQ PR-1] Refactor(SQ): Add ScalarEncodingResolver and parameterize Faiss SQ format by encoding

naveentatikonda:
### Description
Adds `ScalarEncodingResolver` — a single decision point mapping doc bit width (1/2/4) to the Lucene `ScalarEncoding` used by the Faiss SQ memory-optimized-search path — and parameterizes `Faiss1040ScalarQuantizedKnnVectorsFormat`
  by that encoding.

  **No functional change.** Existing 1-bit path is preserved: default constructor resolves to `SINGLE_BIT_QUERY_NIBBLE`, and all current call sites get the same cached flat-format instance as before.

### Related Issues

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss1040ScalarQuantizedKnnVectorsFormat.java
+++ b/src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss1040ScalarQuantizedKnnVectorsFormat.java
@@ -17,17 +17,21 @@ import org.opensearch.knn.index.codec.nativeindex.NativeIndexBuildStrategyFactor
 import org.opensearch.knn.index.engine.KNNEngine;
 
 import java.io.IOException;
+import java.util.Map;
+import java.util.Objects;
+import java.util.concurrent.ConcurrentHashMap;
 
 /**
  * Dedicated format for Faiss SQ vector fields.
  *
- * <p>Uses {@link KNN1040ScalarQuantizedVectorsFormat} with 1-bit quantization
- * ({@link ScalarEncoding#SINGLE_BIT_QUERY_NIBBLE}) for flat vector storage (.vec/.veq files),
- * while HNSW graph construction is delegated to the native Faiss engine (.faiss files).
+ * <p>Uses {@link KNN1040ScalarQuantizedVectorsFormat} for flat vector storage (.vec/.veq files),
+ * while HNSW graph construction is delegated to the native Faiss engine (.faiss files). The
+ * {@link ScalarEncoding} determines the document bit width (1, 2, or 4) and is resolved from the
+ * field's configured bits via {@link ScalarEncodingResolver}; it defaults to 1-bit
+ * ({@link ScalarEncoding#SINGLE_BIT_QUERY_NIBBLE}) for backward compatibility.
  *
- * <p>A field is routed to this format when its method parameters contain
- * {@code "encoder": {"name": "sq", "bits": 1}}. See {@code FaissCodecFormatResolver} for the
- * routing logic in {@code BasePerFieldKnnVectorsFormat.getKnnVectorsFormatForField}.
+ * <p>A field is routed to this format when its method parameters contain an {@code sq} encoder
+ * with a supported bit width. See {@code FaissCodecFormatResolver} for the routing logic.
  *
  * @see Faiss1040ScalarQuantizedKnnVectorsWriter
  * @see Faiss1040ScalarQuantizedKnnVectorsReader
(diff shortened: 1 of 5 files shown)


src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/ScalarEncodingResolver.java
```diff
@@ -0,0 +1,108 @@
...
+ *   <li>4-bit document → {@code PACKED_NIBBLE} (x8)</li>
+ * </ul>
+ *
+ * <p>The mapping is resolved dynamically from {@link ScalarEncoding#getBits()} and
+ * {@link ScalarEncoding#getQueryBits()} rather than by hardcoding enum names, so it stays correct
+ * if Lucene renames constants and so adding a new supported bit width only requires updating
+ * {@link #SUPPORTED_DOC_BITS}. This keeps the {@code SINGLE_BIT_QUERY_NIBBLE} reference out of the
+ * codec format, the build strategy, and the scorer (Requirement 11.4).
+ */
+@UtilityClass
```
kotwanikunal: nit: We can skip this annotation

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/ScalarEncodingResolver.java
```diff
@@ -0,0 +1,108 @@
...
+public class ScalarEncodingResolver {
+
+    /** The FAISS SQ path always quantizes the query to a 4-bit nibble. */
+    public static final int QUERY_BITS = 4;
+
+    /** Document bit widths supported by the FAISS SQ memory-optimized-search path. */
+    private static final int[] SUPPORTED_DOC_BITS = { 1, 2, 4 };
```
kotwanikunal, line 45: More work for me :( 
I am trying to not have bits spread out across classes and have it directly on the base class. I'll refactor with your changes.

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/ScalarEncodingResolver.java
```diff
@@ -0,0 +1,108 @@
...
+
+    /** docBits -> ScalarEncoding, resolved once from the enum metadata. */
+    private static final Map<Integer, ScalarEncoding> DOC_BITS_TO_ENCODING = buildDocBitsToEncoding();
+
+    private static Map<Integer, ScalarEncoding> buildDocBitsToEncoding() {
```
kotwanikunal, line 71: Can we document the current map here for reference? Context will help track deviations over time

naveentatikonda: Added

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/ScalarEncodingResolver.java
```diff
@@ -0,0 +1,131 @@
...
+        return encoding;
+    }
+
+    /**
+     * Inverse of {@link #forDocBits(int)} — the document bit width carried by {@code encoding}.
+     *
+     * @param encoding a Lucene scalar encoding
+     * @return the document bit width ({@link ScalarEncoding#getBits()})
+     */
+    public static int docBits(final ScalarEncoding encoding) {
```
Vikasht34, line 125: we are documeting docBits()  as the Inverse of forDocBits, but it isn't one over the full enum docBits(SEVEN_BIT) returns 7 and docBits(UNSIGNED_BYTE) returns 8, yet forDocBits(7) / forDocBits(8) throw. So the round-trip only holds for the three supported widths. we should validate becuase one to one mapping is not true for the inverse and we should add more test around it.

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss1040ScalarQuantizedKnnVectorsFormat.java
```diff
@@ -37,26 +40,42 @@ public class Faiss1040ScalarQuantizedKnnVectorsFormat extends KnnVectorsFormat {
...
+        this(nativeIndexBuildStrategyFactory, DEFAULT_ENCODING);
+    }
+
+    public Faiss1040ScalarQuantizedKnnVectorsFormat(
+        final NativeIndexBuildStrategyFactory nativeIndexBuildStrategyFactory,
+        final ScalarEncoding encoding
+    ) {
         super(FORMAT_NAME);
         this.nativeIndexBuildStrategyFactory = nativeIndexBuildStrategyFactory;
+        this.faissSqFlatFormat = flatFormatFor(encoding);
```
Vikasht34, line 80: Do we need null check on encoding? and Can we validate what happens when we pass null as encoding.

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/ScalarEncodingResolver.java
```diff
@@ -0,0 +1,131 @@
...
+        return encoding.getBits();
+    }
+
+    private static String supportedDocBitsString() {
```
Vikasht34, line 142: nit :- we can just return return Arrays.toString(SUPPORTED_DOC_BITS);

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/ScalarEncodingResolver.java
Vikasht34: I think we should keep it ... with thing class is by default static and no need to mention private construct.

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/ScalarEncodingResolver.java
naveentatikonda: makes sense..added validation and tests for it

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss1040ScalarQuantizedKnnVectorsFormat.java
naveentatikonda: added it

Vikasht34 (approved):
Looks Good !! Thanks for addressing .. Let's merge and Let's rebase with others

kotwanikunal (approved):

PR #3429, opened 2026-07-15
[SQ PR-2] Accept SQ multibit quantization at mapping and codec layers

naveentatikonda:
### Description
Adds 2-bit (`DIBIT_QUERY_NIBBLE`, x16) and 4-bit (`PACKED_NIBBLE`, x8) as valid values for `sq.bits` alongside the existing 1-bit and fp16 paths, and threads the bit width through the mapping, codec-format, MOS, scorer, and
  build-strategy layers.

  **No user-visible behavior change on the existing 1-bit path.** Explicit encoder config (`encoder: {name: sq, bits: 2|4}`) is now honored end-to-end at the Java layer. Native HNSW build and bulk-SIMD scoring for multi-bit land in
  follow-up PRs.

Fixed a bug where Lucene's SPI would reload `Faiss1040ScalarQuantizedKnnVectorsFormat` via its no-arg constructor and produce a writer with 1-bit encoding hardcoded, silently writing 2/4-bit segments as 1-bit. The format is now
  encoding-agnostic — encoding is resolved per-field from the `SQ_CONFIG` field attribute on the first `addField()` / `mergeOneField()` call for writes, and at `fieldsReader(state)` time for reads. Throws a clear error on missing or
  invalid `SQ_CONFIG` instead of falling back to 1-bit silently.

### Related Issues

### Check List

### DIFF
--- a/src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss1040ScalarQuantizedKnnVectorsFormat.java
+++ b/src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss1040ScalarQuantizedKnnVectorsFormat.java
@@ -9,29 +9,45 @@ import lombok.extern.log4j.Log4j2;
 import org.apache.lucene.codecs.KnnVectorsFormat;
 import org.apache.lucene.codecs.KnnVectorsReader;
 import org.apache.lucene.codecs.KnnVectorsWriter;
-import org.apache.lucene.util.quantization.QuantizedByteVectorValues.ScalarEncoding;
+import org.apache.lucene.index.FieldInfo;
 import org.apache.lucene.index.SegmentReadState;
 import org.apache.lucene.index.SegmentWriteState;
-import com.google.common.annotations.VisibleForTesting;
+import org.apache.lucene.util.quantization.QuantizedByteVectorValues.ScalarEncoding;
+import org.opensearch.knn.common.FieldInfoExtractor;
 import org.opensearch.knn.index.codec.nativeindex.NativeIndexBuildStrategyFactory;
 import org.opensearch.knn.index.engine.KNNEngine;
 
 import java.io.IOException;
+import java.util.Locale;
 import java.util.Map;
-import java.util.Objects;
 import java.util.concurrent.ConcurrentHashMap;
+
+import static org.opensearch.knn.common.KNNConstants.SQ_CONFIG;
 
 /**
  * Dedicated format for Faiss SQ vector fields.
  *
  * <p>Uses {@link KNN1040ScalarQuantizedVectorsFormat} for flat vector storage (.vec/.veq files),
  * while HNSW graph construction is delegated to the native Faiss engine (.faiss files). The
- * {@link ScalarEncoding} determines the document bit width (1, 2, or 4) and is resolved from the
- * field's configured bits via {@link ScalarEncodingResolver}; it defaults to 1-bit
- * ({@link ScalarEncoding#SINGLE_BIT_QUERY_NIBBLE}) for backward compatibility.
+ * {@link ScalarEncoding} is resolved <b>per-field at write time</b> from the segment's
+ * {@link FieldInfo} attributes ({@code SQ_CONFIG}, populated by {@code EngineFieldMapper}) —
+ * <b>not</b> from a value baked into the format instance at construction time.
  *
- * <p>A field is routed to this format when its method parameters contain an {@code sq} encoder
- * with a supported bit width. See {@code FaissCodecFormatResolver} for the routing logic.
+ * <p>This matters because Lucene's SPI machinery ({@link KnnVectorsFormat#forName(String)},
+ * invoked by {@code PerFieldKnnVectorsFormat.FieldsReader}) reopens codecs via the mandatory
+ * no-arg constructor whenever segments are opened on a new reader (index open, shard relocation,
+ * node bootup). If encoding were carried on the instance, an SPI-instantiated format would hold
+ * a constructor default (previously {@code SINGLE_BIT_QUERY_NIBBLE}) and any writer produced by
+ * it would silently miswrite non-1-bit fields as 1-bit. Resolving encoding per-field from
+ * {@code SQ_CONFIG} makes the format instance encoding-agnostic.
+ *
+ * <p>On the <b>read path</b>, {@link org.apache.lucene.codecs.lucene104.Lucene104ScalarQuantizedVectorsFormat#fieldsReader} never
+ * threads its format-instance encoding through to
+ * {@code Lucene104ScalarQuantizedVectorsReader} — the reader resolves per-field encoding by
+ * reading the wire number from each field's {@code .vemq} meta sidecar (scoped by
+ * {@code SegmentReadState.segmentSuffix} at the Lucene layer). So the encoding we pass to
+ * {@link KNN1040ScalarQuantizedVectorsFormat} on the read path is unused — we intentionally use
+ * a default-encoding instance rather than pretend to resolve it.
  *
  * @see Faiss1040ScalarQuantizedKnnVectorsWriter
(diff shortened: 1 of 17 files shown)


src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexBuildStrategyFactory.java
```diff
@@ -60,12 +60,12 @@ public NativeIndexBuildStrategy getBuildStrategy(
...
-        final boolean isFaissSQOneBitField = FieldInfoExtractor.isSQField(fieldInfo)
-            && FieldInfoExtractor.extractSQConfig(fieldInfo).getBits() == FaissSQEncoder.Bits.ONE.getValue();
+        final boolean isFaissSQMosField = FieldInfoExtractor.isSQField(fieldInfo)
+            && FaissSQEncoder.isMosBits(FieldInfoExtractor.extractSQConfig(fieldInfo).getBits());
 
         // Determine build strategy
         final NativeIndexBuildStrategy strategy;
-        if (isFaissSQOneBitField) {
+        if (isFaissSQMosField) {
             strategy = MemOptimizedScalarQuantizedIndexBuildStrategy.getInstance();
```
Vikasht34, line 69: Just Curious to know , based on 2 and 4 bit now , does this startegy has to changed becuase when I was looking at method , it is asuming D/8

Vikasht34 (commented):
Don't have major commnet on this PR .. Things looks good.

src/main/java/org/opensearch/knn/index/codec/nativeindex/NativeIndexBuildStrategyFactory.java
naveentatikonda: Good catch — the comment (and the stale TODO) was written back when only B=1 was supported. The strategy itself is already generic: line 123 reads quantizedVecBytes from QuantizedByteVectorValues.vectorValue(0).length, so we consume whatever byte layout Lucene wrote — ceil(D/8) for B=1, 2 * planeBytes for B=2, packedBytes for B=4. Nothing here assumes D/8.

Updated the comment to document all three layouts and dropped the TODO. No code change needed.

Vikasht34 (dismissed):
Looks Good, Please re-base with you previous PR.

src/main/java/org/opensearch/knn/index/engine/faiss/FaissSQEncoder.java
```diff
@@ -252,26 +256,62 @@ public TrainingConfigValidationOutput validateEncoderConfig(TrainingConfigValida
...
-     * Works with both the method component parameters (from KNNMethodContext) and the
-     * per-field params map (from codec format resolver).
+     * Returns true if {@code bits} is a document bit width served by the memory-optimized SQ (MOS) path.
+     * These are the non-fp16 widths {1, 2, 4} that delegate HNSW construction to native Faiss over
+     * scalar-quantized codes. fp16 (16) is excluded — it uses the standard Faiss SQ description.
+     *
+     * @param bits the configured sq encoder bit width
+     * @return true for bits in {1, 2, 4}
+     */
+    public static boolean isMosBits(final int bits) {
```
navneet1v: [nit-pick] we should see if we can find a better name for these SQ : 1, 2, 4 bits rather than calling them as isMosBits.

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss1040ScalarQuantizedKnnVectorsFormat.java
```diff
@@ -41,51 +50,98 @@ public class Faiss1040ScalarQuantizedKnnVectorsFormat extends KnnVectorsFormat {
...
+     * in that case (see {@link #resolveEncodingForField(FieldInfo)}).
+     */
+    private static FieldInfo findSQField(final FieldInfos fieldInfos) {
+        if (fieldInfos == null) {
+            return null;
+        }
+        for (final FieldInfo fieldInfo : fieldInfos) {
+            if (FieldInfoExtractor.isSQField(fieldInfo)) {
+                return fieldInfo;
+            }
```
navneet1v: what if there are multiple fields with vectors + SQ. How this will work?

Vikasht34 (dismissed):

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss1040ScalarQuantizedKnnVectorsFormat.java
naveentatikonda: Good catch. Does this approach looks good if we try to match it against segmentsuffix ?

```
if (fieldInfos == null) {
            return null;
        }
        final boolean filterBySuffix = segmentSuffix != null && segmentSuffix.isEmpty() == false;
        for (final FieldInfo fieldInfo : fieldInfos) {
            if (FieldInfoExtractor.isSQField(fieldInfo) == false) {
                continue;
            }
            if (filterBySuffix) {
                final String fieldSuffix = fieldInfo.getAttribute(PerFieldKnnVectorsFormat.PER_FIELD_SUFFIX_KEY);
                if (fieldSuffix == null || segmentSuffix.endsWith(FORMAT_NAME + "_" + fieldSuffix) == false) {
                    continue;
                }
            }
            return fieldInfo;
        }
        return null;
    }
```

If not pls let me know if there are any other suggestions

src/main/java/org/opensearch/knn/index/engine/faiss/FaissSQEncoder.java
naveentatikonda: Shall we name it `isSQCodedBits`  ?

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss1040ScalarQuantizedKnnVectorsFormat.java
naveentatikonda: As discussed offline, I'm passing the default constructor to the fieldsReader because anyways we will read the encoder from the segment file

Vikasht34 (dismissed):

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss1040ScalarQuantizedKnnVectorsFormat.java
```diff
@@ -41,51 +57,73 @@ public class Faiss1040ScalarQuantizedKnnVectorsFormat extends KnnVectorsFormat {
...
         return FLAT_FORMAT_CACHE.computeIfAbsent(encoding, KNN1040ScalarQuantizedVectorsFormat::new);
     }
 
-    public Faiss1040ScalarQuantizedKnnVectorsFormat() {
-        this(new NativeIndexBuildStrategyFactory(), DEFAULT_ENCODING);
+    /**
+     * Returns a fresh {@link KNN1040ScalarQuantizedVectorsFormat} with the default encoding. Used
+     * on the read path, where the encoding is unused downstream (see {@link #fieldsReader}).
+     */
+    private static KNN1040ScalarQuantizedVectorsFormat flatFormatFor() {
```
navneet1v, line 74: I really don't want to add such comment, but `flatFormatFor` function name sounds very weird to me, I hoping this is be a typo. and blocking PR because of this comment really looks like nit picking, but we should avoid this kind of function names.

navneet1v (approved):

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss1040ScalarQuantizedKnnVectorsFormat.java
naveentatikonda: Ack, will fix it later in next PR

Vikasht34 (approved):

PR #3431, opened 2026-07-16
[SQ PR-3] Build SQ B-bit HNSW graph with multi-bit symmetric distance

naveentatikonda:
### Description
Generalize the native FAISS HNSW build path from 1-bit to `bits ∈ {1, 2, 4}` scalar quantization.
  Before this PR, `FaissSQDistanceComputer` (the FAISS `DistanceComputer` the HNSW builder queries
  millions of times) hard-coded a single-plane `popcount(a & b)`; after it, the same class computes
  symmetric distances at any of the three supported doc bit widths.
  
  The change threads a new `docBits` int through:
  - Java: `MemOptimizedScalarQuantizedIndexBuildStrategy` → `JNIService.initFaissSQIndex`
    → `FaissService.initFaissSQIndex`. `docBits` is derived from the on-disk `ScalarEncoding`
    (materialised by the codec writer, source of truth vs. field mapping) via
    `ScalarEncodingResolver.docBits(enc)`.
  - JNI: descriptor `(IILjava/util/Map;FI)J` → `(IILjava/util/Map;FII)J`.
  - C++: `InitFaissSQIndex` → `BinaryIndexService::initFaissSQIndex` → `FaissSQFlat` → per-computer.

  #### Two kernel families, dispatched by `docBits` in `multiBitDp`

  | docBits | Kernel | Notes |
  |---|---|---|
  | 1 | `popcountAndPlane` | single-plane popcount — bit-identical to the legacy path |
  | 2 | 2×2 `popcountAndPlane` over bit planes, `<< (i+j)` | matches the DIBIT layout |
  | 4 | `bothPackedNibbleDp` (byte-wise `aHi*bHi + aLo*bLo`) | mirrors Lucene's `int4DotProductBothPacked` for PACKED_NIBBLE |

  PACKED_NIBBLE isn't a bit-plane layout, so bit-plane popcount is arithmetically wrong for B=4 —
  byte-wise nibble multiply is the correct kernel and lets Java-side ingest stay as a single
  `arraycopy` for all three widths (no repack). Modern clang/gcc/msvc auto-vectorise the
  branch-free nibble loop to `PMULLW` / NEON `MUL`.

  Correction-factor scaling: `intervalLength *= 1 / (2^B - 1)` is applied once in
  `setCorrectionFactors`; for B=1 it's a no-op (`dataScale = 1.0`).

  #### Correctness invariant

  For `docBits == 1` the new path produces exactly the same numeric result as the legacy inline
  popcount. Existing 1-bit graphs are byte-for-byte unchanged. `DocBits1MatchesLegacyPopcount` is
  the regression guard.

### Related Issues

### Check List

### DIFF
--- a/jni/include/sq/faiss_sq_flat.h
+++ b/jni/include/sq/faiss_sq_flat.h
@@ -41,10 +41,54 @@ namespace knn_jni {
         quantizedComponentSum = static_cast<float>(componentSum);
     }
 
+    // Only bit widths supported by the memory-optimized-search (MOS) path.
+    // Kept in sync with FaissSQEncoder.isMosBits on the Java side and with
+    // ScalarEncodingResolver.SUPPORTED_DOC_BITS.
+    static inline bool isMosDocBits(int32_t bits) {
+        return bits == 1 || bits == 2 || bits == 4;
+    }
+
+    // Validates a docBits value and returns it unchanged when supported. Called from
+    // FaissSQDistanceComputer's initializer list so the throw beats the `qvb / docBits`
+    // computation in `planeBytes`, which would otherwise trigger a SIGFPE for docBits == 0.
+    static inline int32_t validateDocBits(int32_t bits) {
+        if (!isMosDocBits(bits)) {
+            throw std::runtime_error(
+                "FaissSQDistanceComputer: unsupported docBits=" + std::to_string(bits)
+                + ". Supported: 1, 2, 4."
+            );
+        }
+        return bits;
+    }
+
+    // Validates that quantizedVectorBytes is compatible with the given docBits and returns it
+    // unchanged. For docBits == 2 the bit-plane kernel requires equal-length planes (qvb must
+    // be even). For docBits ∈ {1, 4} any positive qvb is valid — B=1 is a single plane, and
+    // B=4 uses PACKED_NIBBLE where each byte independently carries two elements.
+    static inline uint64_t validateQvbForDocBits(uint64_t qvb, int32_t bits) {
+        if (bits == 2 && qvb % 2 != 0) {
+            throw std::runtime_error(
+                "FaissSQDistanceComputer: quantizedVectorBytes=" + std::to_string(qvb)
+                + " must be even for docBits=2 (two bit planes of equal length)"
+            );
+        }
+        return qvb;
+    }
+
     template <bool IsMaxIP, bool IsBytesMultipleOf8>
     struct FaissSQDistanceComputer final : faiss::DistanceComputer {
         const int64_t oneElementByteSize;
         const uint64_t quantizedVectorBytes;
+        // Number of bits used to quantize each document dimension (1, 2, or 4).
+        const int32_t docBits;
+        // Byte length of a single bit plane. Only meaningful for the bit-plane popcount path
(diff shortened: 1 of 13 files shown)


jni/include/sq/faiss_sq_flat.h
```diff
@@ -69,6 +87,98 @@ namespace knn_jni {
             y1(),
             dimension(_dimension),
             numVectors(_numVectors) {
+            // Validate docBits early. Without this, a bad value from the caller would:
+            //   - docBits == 0    -> planeBytes computes as `qvb / 0` (SIGFPE / UB)
+            //   - docBits > 30    -> `1 << _docBits` is signed shift UB before we cast to float
+            //   - docBits ∉ {1,2,4} -> silent wrong reconstruction / graph corruption
+            // Reject at construction so the failure is loud at build time, not at query time.
+            if (!isMosDocBits(_docBits)) {
```
Vikasht34: I am just curious to know , this validation is too late and would it not break , becuase at line number 78 , 79 , if docs bit is zero and planbytes will throw devide by zero?

jni/include/sq/faiss_sq_flat.h
```diff
@@ -55,11 +70,14 @@ namespace knn_jni {
...
         int32_t numVectors;
 
-        FaissSQDistanceComputer(int32_t _oneElementByteSize, const void* _data, float _centroidDp, int32_t _dimension, int32_t _numVectors)
+        FaissSQDistanceComputer(int32_t _oneElementByteSize, const void* _data, float _centroidDp, int32_t _dimension, int32_t _numVectors, int32_t _docBits)
           : faiss::DistanceComputer(),
             oneElementByteSize(_oneElementByteSize),
             // Memory layout : [Quantized Vector | lowerInterval (float) | upperInterval (float) | additionalCorrection (float) | quantizedComponentSum (int)]
             quantizedVectorBytes(_oneElementByteSize - (sizeof(float) * 3 + sizeof(int32_t))),
+            docBits(_docBits),
+            planeBytes((_oneElementByteSize - (sizeof(float) * 3 + sizeof(int32_t))) / _docBits),
```
Vikasht34: Are we using this for B= 4?

jni/include/sq/faiss_sq_flat.h
naveentatikonda: Good catch. You're right — the constructor's initializer list runs before the body, so planeBytes = qvb / _docBits at line 79 would divide by zero before the guard at line 95 could fire. Fixed by moving the check into a small 
static helper (validateDocBits) that throws early and is called from the initializer list, so validation now happens before planeBytes is computed.

jni/include/sq/faiss_sq_flat.h
naveentatikonda: No — planeBytes is only used by the bit-plane popcount path (B=1 and B=2). For B=4 we use PACKED_NIBBLE, and its kernel (bothPackedNibbleDp) reads quantizedVectorBytes directly without touching planeBytes. I updated the field comment to say this explicitly.

ManasviGoyal:
LGTM

0ctopus13prime (approved):
LGTM

Vikasht34 (approved):

PR #3433, opened 2026-07-16
Fix FAISS SQ merges with vectorless input segments

### LINKED ISSUE #3379: [BUG] FAISS SQ merge can fail when empty scalar-quantized vector values lack quantizedVectorValues

JeremyRitchie:
### Description

FAISS scalar-quantized segment merges can fail when a segment that declares the vector field contains documents but no vector values. Lucene returns `EmptyOffHeapVectorValues` for that source segment; the type does not contain the private `quantizedVectorValues` field expected by `Faiss1040ScalarQuantizedFlatVectorsReader`, so reflective extraction throws before the merged-output guard added in #3381 can run.

This change preserves the reader's existing `ScalarQuantizedFloatVectorValues` contract by supplying an empty `QuantizedByteVectorValues` implementation whenever the delegate reports zero values, before attempting reflective extraction. Non-empty values retain the original extraction path, so incompatibilities still propagate instead of silently omitting the native index. Invalid vector access on the empty implementation follows Lucene's empty off-heap behavior by throwing `UnsupportedOperationException`.

The codec regression test writes a real one-document segment with no vector value, verifies its wrapped empty read path and warmup behavior, merges it with a segment containing 150 vectors, and verifies every surviving vector remains readable. An integration test also force-merges ten non-vector documents to one segment, adds ten vector documents in a second segment, force-merges again, and verifies k-NN search.

This is a Lucene 10.5 adaptation of the system-tested fix developed with @alanshieh.

### Related Issues

Resolves #3379

Follow-up to #3381.

### Check List

- [x] New functionality has been documented (code-level behavior; no public API change).
- [x] API changes companion pull request (not applicable; no API change).
- [x] Public documentation issue/PR (not applicable; internal merge correctness fix).

### Testing

- [x] `./gradlew spotlessApply spotlessJavaCheck`
- [x] Focused wrapper, reader, warmup, and writer tests: 44 tests, 0 failures
- [x] Requested force-merge integration regression: 1 test, 0 failures (seed `4F233982CC831F4D`)
- [x] `./gradlew precommit`
- [x] Initial patch: `./gradlew build -Dnproc.count=4` (`BUILD SUCCESSFUL` in 2h 10m 59s; seed `8C02CBE5BB25AF28`)


src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss1040ScalarQuantizedFlatVectorsReader.java
```diff
@@ -67,14 +70,32 @@ public void checkIntegrity() throws IOException {
...
-        );
+        if (floatVectorValues == null) {
+            return null;
+        }
+
+        final QuantizedByteVectorValues quantizedValues;
+        try {
+            quantizedValues = KNN1040ScalarQuantizedUtils.extractQuantizedByteVectorValues(floatVectorValues);
+        } catch (IOException e) {
+            if (e.getCause() instanceof NoSuchFieldException && floatVectorValues.size() == 0) {
```
navneet1v: rather than getting and exception can we just add a top level if condition if `floatVectorValues.size() == 0` then just return `return new ScalarQuantizedFloatVectorValues(floatVectorValues, new EmptyQuantizedByteVectorValues(floatVectorValues));` and rest of the flow which was there originally can stay?

navneet1v (commented):
Can we also add an IT like this:
1. create a vector index 
2. Ingest 10 docs without vectors in the index.
3. Force merge to 1 segment.
4. Then index 10 more new docs with vectors. which will create a new segment.
5. Do force merge. and then do search.

This should help validate the bug and also validate the fix is actually helping.

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/EmptyQuantizedByteVectorValues.java
```diff
@@ -0,0 +1,98 @@
...
+
+    @Override
+    public int size() {
+        return floatVectorValues.size();
+    }
+
+    @Override
+    public byte[] vectorValue(int ord) throws IOException {
+        return emptyVector;
+    }
```
navneet1v: is this the same contract which lucene has for EmptyFloatVectorValues?

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss1040ScalarQuantizedFlatVectorsReader.java
JeremyRitchie: Done

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/EmptyQuantizedByteVectorValues.java
JeremyRitchie: Good catch.
I checked Lucene 10.5, and our behavior wasn’t quite the same. Lucene’s empty off-heap vector values throw UnsupportedOperationException if vectorValue() is called, while ours returned an empty byte array. I’ve updated our implementation to match Lucene and adjusted the unit test accordingly.

src/test/java/org/opensearch/knn/integ/ModeAndCompressionIT.java
```diff
@@ -583,6 +583,56 @@ public void testMixedSegmentsWithNonVectorDoc_whenValid_ThenSucceed() {
         }
     }
 
+    /**
+     * Regression test for merging a vectorless segment with a FAISS 1-bit scalar-quantized segment.
+     * First force-merges non-vector documents into one segment, then adds vector documents in a
+     * second segment and force-merges both segments before searching.
+     */
+    @SneakyThrows
+    public void testForceMergeWithVectorlessSegment_whenFaissSQ_thenSucceed() {
```
navneet1v: curious to know without your fix this IT is able to capture the issue or not?

src/test/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss1040ScalarQuantizedKnnVectorsWriterTests.java
```diff
@@ -434,8 +437,131 @@ public void testMergeOneField_whenFlatReaderReturnsEmptyVectors_thenSkipsNativeB
...
+    public void testMergeOneField_whenInputSegmentHasNoVectors_thenSucceeds() {
+        final int vectorsPerSegment = 150;
+        final int vectorlessSegmentDocs = 1;
+        final float[][] vectors = generateRandomVectors(vectorsPerSegment, DIMENSION);
+        final FieldInfo fi = createRealFieldInfo();
+        final FieldInfos fieldInfos = new FieldInfos(new FieldInfo[] { fi });
+        final Faiss1040ScalarQuantizedKnnVectorsFormat format = new Faiss1040ScalarQuantizedKnnVectorsFormat();
+        final Lucene104ScalarQuantizedVectorsFormat luceneSqFormat = new Lucene104ScalarQuantizedVectorsFormat(SINGLE_BIT_QUERY_NIBBLE);
+
+        try (org.apache.lucene.store.Directory directory = newDirectory()) {
```
navneet1v: [nit-pick] can we just export this class?

navneet1v (dismissed):
Overall looks good. Minor questions

navneet1v:
added skip change log label since changelog verifier was failing even when changelog entry is present

navneet1v:
@JeremyRitchie why did we remove the IT?

src/test/java/org/opensearch/knn/integ/ModeAndCompressionIT.java
JeremyRitchie: I checked this by temporarily removing the size() == 0 reader guard and running the IT. It still passed, so it wasn’t actually reproducing this failure. A segment containing only documents that never had vectors doesn’t create the empty SQ vector metadata that causes Lucene to return EmptyOffHeapVectorValues. I removed the IT rather than leave misleading regression coverage. 

The real-I/O codec test `testMergeOneField_whenInputSegmentHasNoVectors_thenSucceeds` **does** reproduce the issue: without the guard it fails with an IOException caused by NoSuchFieldException: quantizedVectorValues, and with the fix it passes.

src/test/java/org/opensearch/knn/index/codec/KNN1040Codec/Faiss1040ScalarQuantizedKnnVectorsWriterTests.java
JeremyRitchie: import? Yup, done.

src/test/java/org/opensearch/knn/integ/ModeAndCompressionIT.java
navneet1v: @JeremyRitchie but in when you were doing testing in your distribution you were able to repro it, can we not build a similar thing in the IT?

JeremyRitchie:
> @JeremyRitchie why did we remove the IT?


My thinking was that we wanted test coverage to ensure this fix actually works. This was the success criteria as i understood it.

I was able to validate that we have that without the IT, and that the IT was actually ineffective at validating the fix so it seemed rather pointless to me. 

I'm happy to add an IT which will validate this fix - just wanted to keep this lean.

navneet1v:
> > @JeremyRitchie why did we remove the IT?
> 
> My thinking was that we wanted test coverage to ensure this fix actually works. This was the success criteria as i understood it.
> 
> I was able to validate that we have that without the IT, and that the IT was actually ineffective at validating the fix so it seemed rather pointless to me.
> 
> I'm happy to add an IT which will validate this fix - just wanted to keep this lean.

We want a test that can repro the error and the fix shows the failing test worked.

navneet1v:
> > > @JeremyRitchie why did we remove the IT?
> > 
> > 
> > My thinking was that we wanted test coverage to ensure this fix actually works. This was the success criteria as i understood it.
> > I was able to validate that we have that without the IT, and that the IT was actually ineffective at validating the fix so it seemed rather pointless to me.
> > I'm happy to add an IT which will validate this fix - just wanted to keep this lean.
> 
> We want a test that can repro the error and the fix shows the failing test worked.

Yes lets have it as an IT also. IT generally survives more than UTs in the repo as per experience working in KNN repo. But its not a blocker for merge

JeremyRitchie:
> Yes lets have it as an IT also. IT generally survives more than UTs in the repo as per experience working in KNN repo. But its not a blocker for merge

Added the fixed IT.
- Without the reader guard, it reproduces the quantizedVectorValues merge failure
- With the fix restored, the IT passes

src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/EmptyQuantizedByteVectorValues.java
```diff
@@ -0,0 +1,97 @@
...
+/**
+ * An empty implementation of {@link QuantizedByteVectorValues} used when Lucene returns empty
+ * vector values that do not expose the {@code quantizedVectorValues} field.
+ *
+ * <p>This class delegates size and iteration operations to the underlying {@link FloatVectorValues},
+ * rejects vector access, and returns empty defaults for quantization-specific operations. It maintains the
+ * {@link ScalarQuantizedFloatVectorValues} return type expected by callers without fabricating
+ * quantized data for a segment that contains no vectors.
+ */
+class EmptyQuantizedByteVectorValues extends QuantizedByteVectorValues {
```
kotwanikunal: Why not use the existing `ScalarQuantizedFloatVectorValues` to do null checks instead? 

https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/ScalarQuantizedFloatVectorValues.java#L54-L56
 and https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/codec/KNN1040Codec/ScalarQuantizedFloatVectorValues.java#L64-L66 can have null guards instead. 

We can reuse the existing class structure and just return null with 0 floatVectorValues


Within the reader - 

```
if (floatVectorValues == null) {
    return null;
}
if (floatVectorValues.size() == 0) {
    return new ScalarQuantizedFloatVectorValues(floatVectorValues, null);
}
```

JeremyRitchie: Sounds good - i've made that change now. :)

kotwanikunal (approved):

navneet1v (approved):

navneet1v:
Will merge the change once CI passes

